PR #919, opened 2017-01-24
Move properties in IdentifierManagerTrait and fix CS.

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Makes SensioLabs Insight happy.

### DIFF
--- a/src/Bridge/Doctrine/Orm/ItemDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/ItemDataProvider.php
@@ -32,17 +32,16 @@ use Doctrine\ORM\QueryBuilder;
  */
 class ItemDataProvider implements ItemDataProviderInterface
 {
-    private $managerRegistry;
-    private $propertyNameCollectionFactory;
-    private $propertyMetadataFactory;
-    private $itemExtensions;
-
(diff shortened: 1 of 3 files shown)


teohhanhui:
I'm really not sure if it's a good idea for traits to define properties...

Simperfit:
@dunglas it seems that the memory_limit part is not enough for travis :/

soyuka:
> I'm really not sure if it's a good idea for traits to define properties... In order for traits to stay true to their horizontal reuse (i.e. just like copy-and-paste) nature, they should not be defining things.

It needs to be a class not a trait anyway. This is a good temporary solution no?

dunglas:
As they are using those properties and the class itself is not, IMO they belong to the trait.

dunglas:
Both analysis tools prefer properties in traits as well: https://scrutinizer-ci.com/g/api-platform/core/inspections/69c0ff4c-11dd-4a00-ba62-2c0d97f5434d/issues/

PR #921, opened 2017-01-25
Prevent phpcov segfault on PHP 7.0 (Travis)

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,14 @@ cache:
 matrix:
   include:
     - php: '7.0'
+    - php: '7.1'
       env: coverage=1 lint=1
     - php: '7.1'
       env: deps='low'
 
 before_install:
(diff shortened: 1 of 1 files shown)


soyuka:
Just a question/remark, this is only a problem when we test the Configuration, isn't there something that we could optimize there to reduce the memory usage / increase the test speed?

dunglas:
It's not related with memory usage after all (just calling `merge` even with a small file segfault). I've not identified precisely the bug but upgrading to 7.1 fixes it.

soyuka:
Okay, good to now thanks. I've memory issues with the api-platform test suite on my machine though.

Issue #924, opened 2017-01-27
Can't serialize Ramsey\Uuid\Uuid identifier


fbourigault:
When api-platform try to serialize my entity, I get the `Resource "Ramsey\Uuid\Uuid" not found.` error. It seems to only happen on deep serialization. In the following stack trace, I'm exposing `AppBundle\Entity\Theme` entities which have a bidirectional ManyToOne to a `AppBundle\Entity\Category`.

Also, I tried to expose my `AppBundle\Entity\User` entities which use the same identifier configuration and serialization works well when using `application/json` format but not `application/ld+json` format.

Stack trace:
```
[1] ApiPlatform\Core\Exception\ResourceClassNotFoundException: Resource "Ramsey\Uuid\Uuid" not found.
    at n/a
        in /var/www/vendor/api-platform/core/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php line 72

    at ApiPlatform\Core\Metadata\Resource\Factory\ExtractorResourceMetadataFactory->handleNotFound(null, 'Ramsey\\Uuid\\Uuid')
        in /var/www/vendor/api-platform/core/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php line 50

    at ApiPlatform\Core\Metadata\Resource\Factory\ExtractorResourceMetadataFactory->create('Ramsey\\Uuid\\Uuid')
        in /var/www/vendor/api-platform/core/src/Metadata/Resource/Factory/ShortNameResourceMetadataFactory.php line 35

    at ApiPlatform\Core\Metadata\Resource\Factory\ShortNameResourceMetadataFactory->create('Ramsey\\Uuid\\Uuid')
        in /var/www/vendor/api-platform/core/src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php line 35

    at ApiPlatform\Core\Metadata\Resource\Factory\OperationResourceMetadataFactory->create('Ramsey\\Uuid\\Uuid')
        in /var/www/vendor/api-platform/core/src/Metadata/Resource/Factory/CachedResourceMetadataFactory.php line 53

    at ApiPlatform\Core\Metadata\Resource\Factory\CachedResourceMetadataFactory->create('Ramsey\\Uuid\\Uuid')
        in /var/www/vendor/api-platform/core/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php line 156

[...]

Issue #916, opened 2017-01-21
No resource class found for object of type


ufik:
Hello,

I have entity Role which has collection of RoleResource entities, but this one (RoleResource) is not registered as an API resource. It's giving me back `No resource class found for object of type Resource`, which is an attribute in RoleResource and is registered as an API resource.

```
    /**
     * Resources of the role and their permissions.
     *
     * @ORM\OneToMany(targetEntity="RoleResource", mappedBy="role", cascade={"remove"}, fetch="EAGER")
     **/
    private $resources;
```

If I add @ApiProperty(readable=false) on resources, it's working fine, but I need this property to be propagated.

Another strange thing is, when I remove ApiResource from Resource entity, it works as well. (By it works, I mean Role API endpoint)

Any idea, what's wrong?

teohhanhui:
This seems similar to api-platform/api-platform#201

If you can help us add a failing test case it'd be great.

ufik:
Hi, I have a project on github which has an example.

[security bundle](https://travis-ci.org/ufik/webcookcms-security-bundle)

It's about Roles and Resources as I explained above.

You can see entities here [Entities](https://github.com/ufik/webcookcms-security-bundle/tree/master/src/Webcook/Cms/SecurityBundle/Entity).

And tests which are failing are here [role controller test](https://github.com/ufik/webcookcms-security-bundle/blob/master/tests/Controller/RoleControllerTest.php)

if you run for example first test, there is already var dump, with error message.

Simperfit:
@ufik does the PR fix the issue ?

ufik:
I will try, but it seems PR was about to create tests not fix the issue.

soyuka:
If test pass, it means there was no issue :p.

ufik:
Ok then you can close it, I won't have time to get closer to this problem any time soon.

PR #925, opened 2017-01-30
Add a test for non-resource object normalization

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | not yet
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

This a try to reproduce #924 and #916 but it seems to work well... @fbourigault @ufik do you have an idea of how to reproduce your issue?

### DIFF
--- a/features/main/non_resource.feature
+++ b/features/main/non_resource.feature
@@ -1,0 +1,30 @@
+Feature: Non-resources handling
+  In order to handle use non-resource types
+  As a developer
+  I should be able serialize types not mapped to an API resource.
+
+  Scenario: Get a resource containing a raw object
+    When  I send a "GET" request to "/contain_non_resources/1"
+    Then the JSON should be equal to:
+    """
+    {
+        "@context": "/contexts/ContainNonResource",
+        "@id": "/contain_non_resources/1",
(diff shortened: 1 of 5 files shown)


ufik:
Hi, I have a project on github which has an example.

[security bundle](https://travis-ci.org/ufik/webcookcms-security-bundle)

It's about Roles and Resources as I explained above.

You can see entities here [Entities](https://github.com/ufik/webcookcms-security-bundle/tree/master/src/Webcook/Cms/SecurityBundle/Entity).

And tests which are failing are here [role controller test](https://github.com/ufik/webcookcms-security-bundle/blob/master/tests/Controller/RoleControllerTest.php)

if you run for example first test, there is already var dump, with error message.

Flexmail-Thomas:
I had a very similar issue:

I used @Groups annotation in my resource class but did not add an @Groups annotation in my embedded non-resource class. 
This made the attributes from the embedded non-resource class not allowed by the getAllowedAttributes function in the AbstractNormalizer.

Solution is to a the @Groups annotation to the non-resource class as well.

PR #927, opened 2017-01-30
add test for object as primary key

fbourigault:
| Q             | A
| ------------- | ---
| Bug fix?      | not yet
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

I think it's a simpler reproduction of #924. The issue seems to be about using objects (Ramsey\Uuid\Uuid, DateTime) as primary key. Stack traces looks the same.

### DIFF
--- a/features/main/object_primary_key.feature
+++ b/features/main/object_primary_key.feature
@@ -1,0 +1,24 @@
+Feature: Planning handling
+  In order to use an object as primary key
+  As a developer
+  I should be able serialize types with object primary key.
+
+  @createSchema
+  @dropSchema
+  Scenario: Get a resource containing a raw object
+    When  I send a "GET" request to "/object_primary_keys/2017-01-30"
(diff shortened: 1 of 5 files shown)


dunglas:
I've pushed the start of a fix. It doesn't work for `DateTime` because it cannot be casted to a string.
I suggest to prevent using a non-resource as an identifier. Supporting this edge case introduces a lot of complexity for no real benefit. WDYT @api-platform/core-team

teohhanhui:
Yes, I don't think we should support this.

fbourigault:
I talked about primary keys with @dunglas on slack and agreed to not support objects as primary key. In my real case (using ramsey\uuid), I could easly revert to use of strings. Feel free to close this PR.

Issue #930, opened 2017-02-02
[Hydra] Normalization error in DocumentationNormalizer for ConstraintViolationList


fmata:
In DocumentationNormalizer.php:389 : 

```php
        // Constraint violation list
        $classes[] = [
            '@id' => '#ConstraintViolationList',
            '@type' => 'hydra:Class',
            'subClassOf' => 'hydra:Error',
            'hydra:title' => 'A constraint violation list',
            'hydra:supportedProperty' => [
                [
                    '@type' => 'hydra:SupportedProperty',
                    'hydra:property' => [
                        '@id' => '#ConstraintViolationList/violation',
                        '@type' => 'rdf:Property',
                        'rdfs:label' => 'violation',
                        'domain' => '#ConstraintViolationList',
                        'range' => '#ConstraintViolation',
                    ],
                    'hydra:title' => 'violation',
                    'hydra:description' => 'The violations',
                    'hydra:readable' => true,
                    'hydra:writable' => false,
                ],
            ],
        ];
```

`@type` must be `hydra:Link` to specify it's a relation with ConstraintViolation and `@id` of the property is singular but when a ConstraintViolationList is normalized, this property is plural in ConstraintViolationListNormalizer.php:62 : 

```php
       return [
            '@context' => $this->urlGenerator->generate('api_jsonld_context', ['shortName' => 'ConstraintViolationList']),
            '@type' => 'ConstraintViolationList',
            'hydra:title' => $context['title'] ?? 'An error occurred',
            'hydra:description' => $messages ? implode("\n", $messages) : (string) $object,
            'violations' => $violations,
        ];
```

So the documentation does not reflect the reality : with a generic Hydra client it's a bug which do not let denormalize a ConstraintViolationList.

PR is coming.

dunglas:
Just by curiosity, what do you use as Hydra client?

fmata:
@dunglas I use a custom library I developed. I think I can open it but I prefer to polish some things before.

dunglas:
Nice!

Simperfit:
@fmata This has been fixed by the PR right ?

fmata:
@Simperfit yep

PR #931, opened 2017-02-02
[Hydra] ConstraintViolationList documentation fix

fmata:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #930
| License       | MIT

### DIFF
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -396,13 +396,13 @@ final class DocumentationNormalizer implements NormalizerInterface
                 [
                     '@type' => 'hydra:SupportedProperty',
                     'hydra:property' => [
-                        '@id' => '#ConstraintViolationList/violation',
+                        '@id' => '#ConstraintViolationList/violations',
                         '@type' => 'rdf:Property',
-                        'rdfs:label' => 'violation',
+                        'rdfs:label' => 'violations',
                         'domain' => '#ConstraintViolationList',
                         'range' => '#ConstraintViolation',
                     ],
-                    'hydra:title' => 'violation',
+                    'hydra:title' => 'violations',
                     'hydra:description' => 'The violations',
                     'hydra:readable' => true,
                     'hydra:writable' => false,
--- a/tests/Hydra/Serializer/DocumentationNormalizerTest.php
+++ b/tests/Hydra/Serializer/DocumentationNormalizerTest.php
@@ -248,13 +248,13 @@ class DocumentationNormalizerTest extends \PHPUnit_Framework_TestCase
                         0 => [
                             '@type' => 'hydra:SupportedProperty',
                             'hydra:property' => [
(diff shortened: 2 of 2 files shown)


src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -396,13 +396,13 @@ private function getClasses(array $entrypointProperties, array $classes): array
                 [
                     '@type' => 'hydra:SupportedProperty',
                     'hydra:property' => [
-                        '@id' => '#ConstraintViolationList/violation',
-                        '@type' => 'rdf:Property',
-                        'rdfs:label' => 'violation',
+                        '@id' => '#ConstraintViolationList/violations',
+                        '@type' => 'hydra:Link',
```
dunglas: Are you sure about this change? We include the document so it's not a link (a link should refers to an IRI, not to an embedded document IIRC).

dunglas:
Can you target the 2.0 branch please as it's a bug fix?

src/Hydra/Serializer/DocumentationNormalizer.php
fmata: @dunglas I looked for more informations but I found nothing. May be I am wrong but I don't know how to process a property as relation if `@type` is not `hydra:Link`

dunglas: You can detect that it's an embedded relation because children documents contain an `@id` key.

dunglas: @lanthaler sorry to annoy you again but can you confirm that `Link` should not be used when the relation is embedded as a sub document?

fmata: @dunglas you are right, I can check if the targeted range is a supported class or not to guess if the property is a relation, and no more depend on `@type` of the property. It works fine. Tell me when you want I update my PR to keep `rdf:Property`.

dunglas: Yes please.

fmata:
@dunglas done. Failed tests are unrelated.

dunglas:
Tests fixed in #937. Can you rebase please?

fmata:
@dunglas tests green :+1:

dunglas:
Thanks @fmata

lanthaler:
That’s fine. The embedded node can be used to warm up a cache for instance. It's a non-authoritative optimization though that the client might choose to ignore.

src/Hydra/Serializer/DocumentationNormalizer.php
lanthaler: Using `Link` when the resource is embedded is fine. The embedded node can be used to warm up a cache for instance. As it is a non-authoritative optimization the client might choose to ignore though.

PR #234, opened 2015-08-19
[Proposal] Implement per-operation access control rules

teohhanhui:
Using [Security expressions](http://symfony.com/doc/current/cookbook/expression/expressions.html#book-security-expressions) to control access to each operation.

```yaml
# app/config/services.yml
services:
    resource.product:
        parent: api.resource
        arguments:
            - AppBundle\Entity\Product
        calls:
            - method: initAccessControlRules
              arguments:
                  - collection:
                        POST: "has_role('ROLE_ADMIN')"
                    item:
                        PUT: "has_role('ROLE_ADMIN')"
                        DELETE: "has_role('ROLE_ADMIN')"
        tags:
            - name: api.resource
```

PR #938, opened 2017-02-08
Allow to configure auth access from the resource class

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | api-platform/api-platform#109 (partially), #234
| License       | MIT
| Doc PR        | todo

Edit: now with expression language support

This PR allows to setup authorization controls directly from the resource.

In the following exemple, the only logged in admins can access all operations related to the `Secured` resource. However, regular users owning a specific resource can also access to it.

All variables available in [Symfony's access control expressions](http://symfony.com/doc/current/expressions.html#security-complex-access-controls-with-expressions) are available. The `object` variable value is the current resource (or collection of resources).

```php
<?php

// src/AppBundle/Entity/Secured.php

namespace AppBundle\Entity;

/**
 * @ApiResource(
 *     attributes={"is_granted"="has_role('ROLE_ADMIN')"},
 *     itemOperations={
 *         "get"={"method"="GET", "is_granted"="object.getOwner() == user"}
 *     }
 * )
 * @ORM\Entity
 */
class Secured
{
    /**
     * @ORM\Column(type="integer")
     * @ORM\Id
     */
    public $id;

    /**
     * @ORM\Column(type="text")
     */
    public $owner;
}
```

### DIFF
--- a/src/EventListener/DenyAccessListener.php
+++ b/src/EventListener/DenyAccessListener.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ApiPlatform\Core\EventListener;
+
+use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\ExpressionLanguage\Expression;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+/**
+ * Denies access to the current resource if the logged user doesn't have sufficient permissions.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+final class DenyAccessListener
+{
+    private $resourceMetadataFactory;
+    private $authorizationChecker;
+
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, AuthorizationCheckerInterface $authorizationChecker = null)
+    {
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->authorizationChecker = $authorizationChecker;
+    }
+
+    /**
+     * Sets the applicable format to the HttpFoundation Request.
+     *
+     * @param GetResponseEvent $event
+     *
+     * @throws AccessDeniedException
+     */
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        try {
+            $attributes = RequestAttributesExtractor::extractAttributes($request);
+        } catch (RuntimeException $e) {
+            return;
+        }
+
+        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
+
+        if (isset($attributes['collection_operation_name'])) {
+            $isGranted = $resourceMetadata->getCollectionOperationAttribute($attributes['collection_operation_name'], 'is_granted', null, true);
+        } else {
+            $isGranted = $resourceMetadata->getItemOperationAttribute($attributes['item_operation_name'], 'is_granted', null, true);
+        }
+
+        if (null === $isGranted) {
+            return;
+        }
+
+        if (null === $this->authorizationChecker) {
+            throw new \LogicException(sprintf('The "symfony/security" library must be installed to use the "is_granted" attribute on class "%s".', $attributes['resource_class']));
+        }
+
+        if (!class_exists(Expression::class)) {
+            throw new \LogicException(sprintf('The "symfony/expression-language" library must be installed to use the "is_granted" attribute on class "%s".', $attributes['resource_class']));
+        }
+
+        if (!$this->authorizationChecker->isGranted(new Expression($isGranted), $request->attributes->get('data'))) {
+            throw new AccessDeniedException();
+        }
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,7 @@
         "symfony/config": "^3.2",
         "symfony/dependency-injection": "^2.7 || ^3.0",
         "symfony/doctrine-bridge": "^2.8 || ^3.0",
+        "symfony/expression-language": "^2.8 || ^3.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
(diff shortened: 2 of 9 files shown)


desmax:
Great feature!
But what if you pass to isGranted() not only attribute, but api resource object as well?
This way we could use advantage of symfony voters.

dunglas:
@desmax good idea.

teohhanhui:
(Old suggestion: api-platform/core#234)

I think we should use the Security expression language:

```php
"access_control"="is_granted('ROLE_ADMIN')"
```

Basically to have the same effect as http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/security.html

teohhanhui:
And this feature must be disabled when the Symfony Security component is not present. (Additionally, the Symfony Expression component too, if we go with that.)

dunglas:
PR updated:

* Now use expression language
* Give access to the current object
* Throw an exception if (and only if) a security rule is used but the security bundle or the expression language component isn't installed

src/EventListener/DenyAccessListener.php
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            $isGranted = $resourceMetadata->getItemOperationAttribute($attributes['item_operation_name'], 'is_granted', null, true);
+        }
+
+        if (null === $isGranted) {
+            return;
+        }
+
+        if (null === $this->authorizationChecker) {
+            throw new \LogicException('The "symfony/security" library must be installed to use the "is_granted" attribute.');
```
lyrixx: Does it worth it to add something like "on property X on class Y" ? to ease debug ? (same for next exception)

src/EventListener/DenyAccessListener.php
```diff
@@ -0,0 +1,85 @@
...
+
+        if (!class_exists(Expression::class)) {
+            throw new \LogicException('The "symfony/expression-language" library must be installed to use the "is_granted" attribute.');
+        }
+
+        $data = $request->attributes->get('data');
+
+        foreach ($isGranted as $expression) {
+            if ($this->authorizationChecker->isGranted(new Expression($expression), $data)) {
+                return;
```
lyrixx: This logic is a bit weird. If at least one expression is valid, then it's OK.
What if I want an `and` between all my expressions? And In order to avoid dealing with this kind of decission, why not allowing only one `is_granted`?

(Actually, I don't know, I'm just asking)

dunglas: Indeed, this is a rest of the old logic (without expression language) but it's better to only allow one expression now.

I'll update the PR.

src/EventListener/DenyAccessListener.php
dunglas: done

src/EventListener/DenyAccessListener.php
dunglas: fixed

wadjeroudi:
@dunglas Thanks for your work. This feature will be released on next version or do we need to use dev-master ?

dunglas:
It will be in 2.1.

piotrbrzezina:
Hi 
@dunglas is there any way to use  with custom voters ?

` attributes={"is_granted"="is_granted('view',object)"}`

desmax:
@piotrbrzezina define your operation smth like this `"get"={"method"="GET", "is_granted"="get"},`
then register voter, that supports that class, but give it a priority 246, so it would be ran before ExpessionVoter.

larkarvin:
I have implemented mine as
```
/**
 * @ApiResource(
 *     attributes={"is_granted"="has_role('ROLE_ADMIN')"},
 *     itemOperations={
 *         "get"={"method"="GET", "is_granted"="object.isAllowed(user)"}
 *     }
 * )
 * @ORM\Entity
 */
class Book
{
    // code omitted
    public function isAllowed($user)
    {
        if($user->hasRole('ROLE_ADMIN')){
            return TRUE;
        }

        if($this->getOwner() == $user){
            return TRUE;
        }
        return FALSE;
    }
}
```
I needed the extra check that if the logged user is an admin, they will be allowed to access the item operation.

wcoppens:
@desmax Thanks for the info. It seems that using it the way you suggest makes it possible to go through the custom security voter, however the `get` in your case is still being seen as a expression by Symfony and so finally goes through the ExressionVoter in which it can't parse the expression because it's invalid.

Any suggestion how we can prevent this becoming a Expression?

playmice:
Also, there's some useful stuff at [https://api-platform.com/docs/core/extensions](https://api-platform.com/docs/core/extensions).

synchr0n123r:
I implementing complex authorization, i have 4 level configuration:
1. Basic | nothing to do here
2. Medium | i configure via attributes base on access level.
 ```
*      attributes={
 *          "authorization" = "api.firewall.comment",
 *          "permissions" = {
 *                  "collection" = ["GUEST", "ROLE_ADMIN"],
 *                  "item" = ["GUEST", "ROLE_ADMIN"],
 *                  "create" = ["GUEST", "ROLE_ADMIN"],
 *                  "update" = ["GUEST", "ROLE_ADMIN"],
 *                  "delete" = ["GUEST", "ROLE_ADMIN"],
 *                  "getCollection" = ["GUEST", "ROLE_ADMIN"],
 *          },
```
3. Custom Authorization based Roles, Entity & Operations. 
4. Custom Authorization based on Role and Object. Like "ROLE_USER" only able change status

My solutions is creating custom event subscriber, then define "authorization" class on which @ApiResource definition.

I create specific class for each entity who need custom authorization. And inject class into custom event subscriber class via addMethod call:
 ```
$firewall = $container->getDefinition('api.firewall.voter');
        foreach ($container->findTaggedServiceIds('api.firewall.entity') as $id => $attributes)
        {
            $firewall->addMethodCall('addVoter', array($id, new Reference($id)));
        }
```
it work great and flexible. I think we can implement this into core system

ksom:
@larkarvin What do you think about:

```php
/**
 * @ApiResource(
 *     attributes={"is_granted"="has_role('ROLE_ADMIN')"},
 *     itemOperations={
 *         "get"={"method"="GET", "is_granted"="has_role('ROLE_ADMIN') || object.getOwner()
 == user"}
 *     }
 * )
 * @ORM\Entity
 */
class Book
{
}

```

ksom:
@synchr0n123r Why do you not create compiler pass to populate the tagged services? Inject the container isn't a good practice. In Symfony 3.4 you can do: 

```yaml
services:
    api.firewall.voter:
        class: Tstudio\ApiBundle\Firewall\Pool
        arguments: ['@api_platform.metadata.resource.metadata_factory', !tagged api.firewall.entity]
        tags:
            - { name: kernel.event_subscriber}
```
CF: http://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services

For your project not using Sf3.4 you can refer to the compiler pass documentation (anchor on the section you need): http://symfony.com/doc/current/service_container/tags.html#create-a-compiler-pass
Thanks for your sharing :)

PR #939, opened 2017-02-10
Fix prepend configuration of serializer and property info

magarzon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -38,11 +38,24 @@ public function prepend(ContainerBuilder $container)
...
+            if (isset($frameworkParameters['serializer'])) {
+                $serializerConfig = $serializerConfig ?: $frameworkParameters['serializer'];
+            }
+
+            if (isset($frameworkParameters['property_info'])) {
+                $propertyInfoConfig = $propertyInfoConfig ?: $frameworkParameters['property_info'];
+            }
+        }
+
+        if (null === $serializerConfig || !isset($serializerConfig['enabled'])) {
```
dunglas: The `null` test is useless, isset will already take care of that. It can be removed.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -38,11 +38,24 @@ public function prepend(ContainerBuilder $container)
...
             $container->prependExtensionConfig('framework', ['serializer' => ['enabled' => true]]);
         }
 
-        if (!isset($frameworkConfiguration['property_info'], $frameworkConfiguration['property_info']['enabled'])) {
+        if (null === $propertyInfoConfig || !isset($propertyInfoConfig['enabled'])) {
```
dunglas: Same here, the `null === ...` part can be removed.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -38,11 +38,24 @@ public function prepend(ContainerBuilder $container)
             return;
         }
 
-        if (!isset($frameworkConfiguration['serializer'], $frameworkConfiguration['serializer']['enabled'])) {
+        $serializerConfig = null;
```
dunglas: And you can remove those 2 new variables too (still thanks of `isset`).

magarzon: If I remove the two variables, I have a compiler error, because are used in line 46 and 50.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -38,11 +38,24 @@ public function prepend(ContainerBuilder $container)
...
+        $propertyInfoConfig = null;
+
+        foreach ($frameworkConfiguration as $frameworkParameters) {
+            if (isset($frameworkParameters['serializer'])) {
+                $serializerConfig = $serializerConfig ?: $frameworkParameters['serializer'];
```
dunglas: You can do `$serializerConfig = $serializerConfig ?? $frameworkParameters['serializer']`. And the 2 vars can be deleted.

magarzon: I forgot that v2.x has PHP 7.0 as minimum version

PR #941, opened 2017-02-12
Fix prepend configuration

magarzon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
@@ -76,7 +76,7 @@ class ApiPlatformExtensionTest extends \PHPUnit_Framework_TestCase
     public function testNotPrependSerializerWhenConfigExist()
     {
         $containerBuilderProphecy = $this->prophesize(ContainerBuilder::class);
-        $containerBuilderProphecy->getExtensionConfig('framework')->willReturn(['serializer' => ['enabled' => false]])->shouldBeCalled();
+        $containerBuilderProphecy->getExtensionConfig('framework')->willReturn([0 => ['serializer' => ['enabled' => false]]])->shouldBeCalled();
         $containerBuilderProphecy->prependExtensionConfig('framework', Argument::any())->willReturn(null)->shouldBeCalled();
         $containerBuilderProphecy->prependExtensionConfig('framework', Argument::that(function (array $config) {
             return array_key_exists('serializer', $config);
@@ -89,7 +89,7 @@ class ApiPlatformExtensionTest extends \PHPUnit_Framework_TestCase
     public function testNotPrependPropertyInfoWhenConfigExist()
     {
         $containerBuilderProphecy = $this->prophesize(ContainerBuilder::class);
-        $containerBuilderProphecy->getExtensionConfig('framework')->willReturn(['property_info' => ['enabled' => false]])->shouldBeCalled();
+        $containerBuilderProphecy->getExtensionConfig('framework')->willReturn([0 => ['property_info' => ['enabled' => false]]])->shouldBeCalled();
         $containerBuilderProphecy->prependExtensionConfig('framework', Argument::any())->willReturn(null)->shouldBeCalled();
         $containerBuilderProphecy->prependExtensionConfig('framework', Argument::that(function (array $config) {
             return array_key_exists('property_info', $config);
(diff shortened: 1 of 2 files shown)


magarzon:
I messed up my branch when I did de amend to change the author of a commit, so I have to remove it, and so the pull request. In this "new" pull request there is only 1 commit with all the changes requested

dunglas:
Thank you for fixing this @magarzon

teohhanhui:
I don't think this is the correct way to fix it. See http://symfony.com/doc/current/bundles/prepend_extension.html

magarzon:
@teohhanhui and what do you see different from that doc? (That btw is not updated)

dunglas:
@magarzon in the doc they explicitly whitelist some extensions using a switch statement. Here we blindly retrieve the config of any extension having a `serializer` or `property_info` section.

magarzon:
@dunglas no, in the doc they check if two extension are available, and in that case a configuration is set. Here we're getting the configuration of the framework extension directly, and if it exists, we check if the serializer is enabled in the framework extension configuration, not any other extension.

The problem was that the framework configuration was handled as if the getExtensionConfig returned the parameters directly, but what getExtensionConfig returns is an array of arrays, because configuration can be splitted in several files (like config_dev.yml and config.yml).

teohhanhui:
It looks like we should call `processConfiguration` (defined here: https://github.com/symfony/symfony/blob/v3.2.3/src/Symfony/Component/DependencyInjection/Extension/Extension.php#L95-L100), instead of trying to work with the raw configuration parameters.

magarzon:
But we don't need to process the bundle configuration (it'll be processed later), we don't need any parameter from the bundle configuration, we're changing the configuration of another bundle (framework).

In the doc example, it needs to process the configuration of the bundle itself, because it needs a parameter to configure the others.

teohhanhui:
No, processConfiguration is still necessary. Because it ensures that the configs are properly merged. The last config should take precedence.

The current code we have is too naive and does the wrong thing when there is more than one config. (We only take the first.)

magarzon:
No, processConfiguration is necessary to merge the configuration of the bundle, but in this case we don't need it (yet, it'll be done later in the call to load method).

Only in the case that we need any parameter from the bundle configuration to configure the other/s bundle/s we need to process configuration, in order to validate, get default values, etc. those parameters that we need to use. As the example in the documentation does.

The second point: the code is not too naive. As the configuration process works, with the prepend configuration we can add configuration to another bundle, but we can NOT overwrite explicit configuration. The precedence order is the last config/file loaded, that's why parameters in config_dev.yml takes precedence over parameters defined in config.yml (for the dev environment) and that's why prependExtensionConfig method use array_unshift to put our configuration first (less precedence) in the array of configurations.

So, checking the first (not one, the first) configuration that has serializer configuration set is enough, because THAT first configuration already has more precedence that ours (and also the other ones), and it's not worth to add a configuration that is going to be overwritten.

In fact, what it's checked is that the enabled is not set, because if it's set, we can't do anything to revert/overwrite it.

I mean, if the user explicitly set enabled to false in the config.yml, we can't do anything to change it.
This code is only to enable serializer (and property info) if nothing is set in the configuration.

teohhanhui:
@magarzon You're right about `processConfiguration`. We should not call it: https://github.com/symfony/symfony/issues/12353#issuecomment-65891644

But can you modify the test to have more than one configs, to show that we're indeed using the correct config?

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -89,7 +89,7 @@ public function testNotPrependSerializerWhenConfigExist()
     public function testNotPrependPropertyInfoWhenConfigExist()
     {
         $containerBuilderProphecy = $this->prophesize(ContainerBuilder::class);
-        $containerBuilderProphecy->getExtensionConfig('framework')->willReturn(['property_info' => ['enabled' => false]])->shouldBeCalled();
+        $containerBuilderProphecy->getExtensionConfig('framework')->willReturn([0 => ['property_info' => ['enabled' => false]]])->shouldBeCalled();
```
teohhanhui: The `0 => ` part can be removed, right?

dunglas: indeed

dunglas: It's weird that PHP CS Fixer doesn't cache this.

PR #950, opened 2017-02-16
Fix #944 - allows to filter by relation by keeping EagerLoading enabled

### LINKED ISSUE #944: Filtering by a relation's property
Hi,

This is affecting version 2.0.1 to current (2.0.3)

Having the following entities:
```php
<?php

namespace AppBundle\Entity;

use Doctrine\Common\Collections\ArrayCollection;
use Symfony\Component\Serializer\Annotation as Serializer;

/**
 * @ApiResource(
 *     attributes={
 *          "normalization_context"={"groups"={"foo"}},
 *          "filters"={"foo.search_filter"}
 *     }
 * )
 * @ORM\Entity
 */
class Foo
{
    /**
     * @var int The entity Id
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var string Something else
     *
     * @ORM\OneToMany(targetEntity="Bar", mappedBy="foo")
     *
     * @Serializer\Groups({"foo"})
     */
    private $bars;

    public function __construct()
    {
        $this->bars = new ArrayCollection();
    }

    public function getId()
    {
        return $this->id;
    }

    /**
     * @return string
     */
    public function getBars()
    {
        return $this->bars;
    }

    /**
     * @param string $bars
     * @return static
     */
    public function setBars($bars)
    {
        $this->bars = $bars;
        return $this;
    }
}
```

```php
<?php

namespace AppBundle\Entity;

use Symfony\Component\Serializer\Annotation as Serializer;

/**
 * @ApiResource
 * @ORM\Entity
 */
class Bar
{
    /**
     * @var int The entity Id
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var Foo
     *
     * @ORM\ManyToOne(targetEntity="Foo", inversedBy="bars")
     * @ORM\JoinColumn(nullable=false, onDelete="CASCADE")
     * @Assert\NotBlank
     */
    private $foo;

    /**
     * @var string
     *
     * @ORM\Column(nullable=false)
     * @Assert\NotBlank
     *
     * @Serializer\Groups({"foo"})
     */
    private $prop = '';

    public function getId()
    {
        return $this->id;
    }

    /**
     * @return Foo|null
     */
    public function getFoo()
    {
        return $this->foo;
    }

    /**
     * @param Foo $foo
     * @return static
     */
    public function setFoo(Foo $foo)
    {
        $this->foo = $foo;
        return $this;
    }

    /**
     * @return string
     */
    public function getProp()
    {
        return $this->prop;
    }

    /**
     * @param string $prop
     * @return static
     */
    public function setProp($prop)
    {
        $this->prop = $prop;
        return $this;
    }
}
```

Services defined as:
```yaml
services:
  app.entity.filter.foo:
      parent:    'api_platform.doctrine.orm.search_filter'
      arguments:
          - { bars.prop: 'ipartial' }
      tags:
          - { name: 'api_platform.filter', id: 'foo.search_filter' }
```

Having the following data:
```yaml
AppBundle\Entity\Foo:
  foo1:
    id: 1
    bars: ["@bar1", "@bar2"]
  
AppBundle\Entity\Bar:
  bar1:
    id: 1
    prop: "toto"
    foo: "@foo1"
  bar2:
    id: 2
    prop: "tata"
    foo: "@foo1"
```

Having set the proper filters in order to run that simple query:
```http://localhost/foos?bar.prop=toto```

Will result:
```json
{
  "@context": "/contexts/Foo",
  "@id": "/foos",
  "@type": "hydra:Collection",
  "hydra:member": [
    {
      "@id": "/foos/1",
      "@type": "Foo",
      "bars": [
        {
          "@id": "/bars/1",
          "@type": "Bar",
          "prop": "toto"
        }
      ]
    }
  ],
  "hydra:totalItems": 1,
  "...otherProps...": "...otherValues...."
}
```

I understand that the *second Bar* attached to `foo1` gets filtered through the doctrine query, but since we're on the Foo endpoint:
* is it still a valid behaviour ?
* if so, how to tell api-platform/doctrine to fetch all relations upon filtering on them ?

Regards,
Yohann

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #944
| License       | MIT
| Doc PR        | na

/!\ This is probably the worse code I've ever produced here and I hope we can find how to improve it together.

To fix #944 I needed to alter the query generated by the filters and the eager loading extension. To do so, I've added a `FilterEagerLoadingExtension` that patches the Query generated by the `FilterExtension`. 

It runs directly after the `FilterExtension` and transforms the query so that it ends up doing something like:

```
SELECT o, partial bars_a1_p.{id,prop} 
FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Foo o 
LEFT JOIN o.bars bars_a1_p 
WHERE o IN(
  SELECT p 
  FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Foo p 
  LEFT JOIN p.bars bars_a1 
  WHERE LOWER(bars_a1.prop) LIKE LOWER(CONCAT('%', :prop_p1, '%'))
)
```

TODO:

- [x] Rename test class Foo/Bar too generic
- [x] Unit tests
- [x] Allow the user to skip this extension so that we can get the relation items filtered (previous behavior that is considered as a bug)
- [x] Prevent the extension from running when `forceEager` is not set
- [x] Performances might be improved slightly by selecting the wanted id with the were clause (removes the JOIN inside the IN and select from that entity directly, but it's complex)
- [x] Test OneToMany
- [x] Test ManyToMany

~~Apologizes to Scrutinizer, that complexity index will be very high -_-.~~

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -13,8 +13,12 @@ use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\CompositeItem;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\CompositeLabel;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\CompositeRelation;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCar;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCarColor;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyFriend;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedToDummyFriend;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelationEmbedder;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\UuidIdentifierDummy;
 use Behat\Behat\Context\Context;
@@ -135,18 +139,21 @@ class FeatureContext implements Context, SnippetAcceptingContext
     }
 
     /**
-     * @Given there is :nb dummy objects with relatedDummies
+     * @Given there is :nb dummy objects having each :nbrelated relatedDummies
      */
-    public function thereIsDummyObjectsWithRelatedDummies($nb)
+    public function thereIsDummyObjectsWithRelatedDummies($nb, $nbrelated)
     {
         for ($i = 1; $i <= $nb; ++$i) {
-            $relatedDummy = new RelatedDummy();
-            $relatedDummy->setName('RelatedDummy #'.$i);
-
             $dummy = new Dummy();
             $dummy->setName('Dummy #'.$i);
             $dummy->setAlias('Alias #'.($nb - $i));
-            $dummy->addRelatedDummy($relatedDummy);
+
+            for ($j = 1; $j <= $nbrelated; ++$j) {
+                $relatedDummy = new RelatedDummy();
+                $relatedDummy->setName('RelatedDummy'.$j.$i);
+                $this->manager->persist($relatedDummy);
+                $dummy->addRelatedDummy($relatedDummy);
+            }
 
             $this->manager->persist($relatedDummy);
             $this->manager->persist($dummy);
@@ -317,6 +324,62 @@ class FeatureContext implements Context, SnippetAcceptingContext
         $fileConfigDummy->setFoo('Foo');
 
         $this->manager->persist($fileConfigDummy);
+        $this->manager->flush();
+    }
+
+    /**
+     * @Given there is a DummyCar entity with related colors
+     */
+    public function thereIsAFooEntityWithRelatedBars()
+    {
+        $foo = new DummyCar();
+        $this->manager->persist($foo);
+
+        $bar1 = new DummyCarColor();
+        $bar1->setProp('red');
+        $bar1->setCar($foo);
+        $this->manager->persist($bar1);
+
+        $bar2 = new DummyCarColor();
+        $bar2->setProp('blue');
+        $bar2->setCar($foo);
+        $this->manager->persist($bar2);
+
+        $foo->setColors([$bar1, $bar2]);
+        $this->manager->persist($foo);
+
+        $this->manager->flush();
+    }
+
+    /**
+     * @Given there is a RelatedDummy with :nb friends
+     */
+    public function thereIsARelatedDummyWithFriends($nb)
+    {
+        $relatedDummy = new RelatedDummy();
+        $relatedDummy->setName('RelatedDummy with friends');
+        $this->manager->persist($relatedDummy);
+
+        for ($i = 1; $i <= $nb; ++$i) {
+            $friend = new DummyFriend();
+            $friend->setName('Friend-'.$i);
+
+            $this->manager->persist($friend);
+
+            $relation = new RelatedToDummyFriend();
+            $relation->setName('Relation-'.$i);
+            $relation->setDummyFriend($friend);
+            $relation->setRelatedDummy($relatedDummy);
+
+            $relatedDummy->addRelatedToDummyFriend($relation);
+
+            $this->manager->persist($relation);
+        }
+
+        $relatedDummy2 = new RelatedDummy();
+        $relatedDummy2->setName('RelatedDummy without friends');
+        $this->manager->persist($relatedDummy2);
+
         $this->manager->flush();
     }
 }
--- a/features/doctrine/date_filter.feature
+++ b/features/doctrine/date_filter.feature
@@ -395,18 +395,18 @@ Feature: Date filter on collections
     And the JSON should be equal to:
     """
     {
-          "@context": "\/contexts\/Dummy",
-          "@id": "\/dummies",
+          "@context": "/contexts/Dummy",
+          "@id": "/dummies",
           "@type": "hydra:Collection",
           "hydra:member": [],
           "hydra:totalItems": 0,
           "hydra:view": {
-              "@id": "\/dummies?relatedDummy.dummyDate%5Bafter%5D=2015-04-28",
+              "@id": "/dummies?relatedDummy.dummyDate%5Bafter%5D=2015-04-28",
               "@type": "hydra:PartialCollectionView"
           },
           "hydra:search": {
               "@type": "hydra:IriTemplate",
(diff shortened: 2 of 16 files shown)


soyuka:
Query is now better:

```
SELECT o, partial bars_a1.{id,prop} FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Foo o LEFT JOIN o.bars bars_a1 WHERE o IN(
    SELECT IDENTITY(bars_a1_a2.foo) 
    FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Bar bars_a1_a2 
    WHERE LOWER(bars_a1_a2.prop) LIKE LOWER(CONCAT('%', :prop_p1, '%'))
)
```

soyuka:
WDYT? I'll add unit tests but I want to know first if this solution is fine to you or if you want to try another approach :). 

@api-platform/core-team

dunglas:
Looks better, indeed. @ymarillet can you try if this patch fix the problem?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -94,6 +94,7 @@ private function handleConfig(ContainerBuilder $container, array $config, array
         $container->setParameter('api_platform.eager_loading.enabled', $config['eager_loading']['enabled']);
         $container->setParameter('api_platform.eager_loading.max_joins', $config['eager_loading']['max_joins']);
         $container->setParameter('api_platform.eager_loading.force_eager', $config['eager_loading']['force_eager']);
+        $container->setParameter('api_platform.eager_loading.filter_relations', $config['eager_loading']['filter_relations']);
```
soyuka: @teohhanhui suggestions about that parameter name ? \o/

teohhanhui: fetch_partial?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -49,6 +49,7 @@ public function getConfigTreeBuilder()
                         ->booleanNode('enabled')->defaultTrue()->info('To enable or disable eager loading')->end()
                         ->integerNode('max_joins')->defaultValue(30)->info('Max number of joined relations before EagerLoading throws a RuntimeException')->end()
                         ->booleanNode('force_eager')->defaultTrue()->info('Force join on every relation. If disabled, it will only join relations having the EAGER fetch mode.')->end()
+                        ->booleanNode('filter_relations')->defaultFalse()->info('Allow to filter the relation collection when the filter is on a nested property. By default, it will retrieve the full relation of the filtered resource. If enabled, only relations matching the filter will be returned.')->end()
```
teohhanhui: Remind me to rewrite this for you. :smile:

soyuka:
Haha I will <3

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka: !RemindMe

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
soyuka: fetch_partial to true would mean we want to fetch a partial relation?

teohhanhui: Yes. So the default should be false.

Or perhaps `partial_relations`?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
teohhanhui: When filtering on a nested property through a to-many relation, do not fetch the other items in the collection. Warning: this breaks round trip, as you cannot PUT the returned collection back to the parent property.

p/s: @soyuka Does this only happen when filtering on a nested property? Or whenever the other side of the relation is fetch-joined?

teohhanhui: Also, we need to warn that this would apply to every level in a multi-level nested property.

soyuka: Only when filtering on a nested property because it only joins the result matching that filter.

teohhanhui: When filtering by e.g. `products?brands=/brands/1`, doesn't it result in SQL something like:

```sql
SELECT ...
FROM products
INNER JOIN products_brands
    ON products.id = products_brands.product_id
INNER JOIN brands
    ON products_brands.brand_id = brands.id
WHERE brands.id = 1
```

and thus:

```json
{
  "@id": "/products/1",
  "brands": [
    "/brands/1"
  ]
}
```

even if product 1 has other brands?

teohhanhui: I really think we should not have this option, because it results in returning invalid data. ("Product 1 has 1 brand which is Brand 1.")

soyuka: Not sure I understood, what is the relation type on products_brands and brands?

teohhanhui: Many-to-many. products_brands is the join table.

soyuka: Yeah your right I think that this option is too much confusing, ManyToMany relations are working fine, I've added a test case anyway. 

I'm removing the availability to filter on the relation. I think that the best to filter a relation would be to use `subresource` + filter.

soyuka:
@teohhanhui I refactored the fix, it's much cleaner and simple now. It clones the query and uses a subquery with the where clause to find the root entity identifiers.

Origin query:

```
SELECT o, partial relatedToDummyFriend_a1.{name,dummyFriend,relatedDummy}, partial dummyFriend_a2.{id,name} 
FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy o 
LEFT JOIN o.relatedToDummyFriend relatedToDummyFriend_a1 
LEFT JOIN relatedToDummyFriend_a1.dummyFriend dummyFriend_a2 
WHERE dummyFriend_a2.id = :dummyFriend_p1
``` 

Fixed query:

```
SELECT o, partial relatedToDummyFriend_a1.{name,dummyFriend,relatedDummy}, partial dummyFriend_a2.{id,name} 
FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy o 
LEFT JOIN o.relatedToDummyFriend relatedToDummyFriend_a1 
LEFT JOIN relatedToDummyFriend_a1.dummyFriend dummyFriend_a2 
WHERE o IN(
  SELECT o_2 FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy o_2 
  LEFT JOIN o_2.relatedToDummyFriend relatedToDummyFriend_a1_a3 
  LEFT JOIN relatedToDummyFriend_a1_a3.dummyFriend dummyFriend_a2_a4 
  WHERE dummyFriend_a2_a4.id = :dummyFriend_p1
)
```

This will work with any relation (manytomany, manytoone etc.). I've removed the parameter to skip this extension.

soyuka:
ping @api-platform/core-team 

I think that this (as it's a bug fix) should target 2.0. Lmk.

teohhanhui:
Yes, this is a bug fix. `2.0` it should be then.

dunglas:
Thank you very much @soyuka

vasilvestre:
filter_relations parameter seem not documented, it's still exist ?

soyuka:
> filter_relations parameter seem not documented, it's still exist ?

Nope.

qDL7:
is it possible then to filter relations without a custom filters? @soyuka

PR #952, opened 2017-02-20
Builtin cache invalidation system aka make API Platform fast as hell

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

The usual quote:

> There are only two hard things in Computer Science: cache invalidation and naming things.
>
> -- Phil Karlton

Well, API Platform is an awesome name, so this PR try to solve the other hard thing: caching to make your API as fast as possible.

It introduces a builtin mechanism to always serve API responses from a cache (Varnish and CloudFlare are targeted), and invalid stale data in real time when a resource is updated, deleted or created.
With this mechanism, on my computer and using Docker, large and complex responses are served in **~15ms instead of ~700ms** without cache.

The API Platform serializer has been tweaked to store the list of all resources included in a given response (the root document, but also embedded documents and documents appearing in lists).

The response is marked with all resources it contains in the `Cache-Tags` HTTP header. A md5 hash of all included IRIs is generated to prevent collisions and reduce the header size.
Then, all API's responses are stored with a high expiration time in the proxy cache. On all subsequent (read) requests from a client, the response is served from the proxy, the PHP application is not touched.

When a resource is modified (only changes made to Doctrine entities are supported for now), all responses containing or referencing it are purged from the cache.
A class to purge Varnish is provided in this PR, and a class to purge CloudFlare (enterprise plans only) will be provided later. An interface allows to add support for other cache providers.

Paged collections are also handled: if a resource is added or removed, all collection pages are purged. If a resource is edited, pages (and all other API responses, including those embedding it as a nested document) are purged.

Enabling this new feature **doesn't require to change existing code**. The following config enable the mechanism and mades your API instantly blazing fast:

```yaml
api_platform:
    http_cache:
        enable_tags: true
        varnish_url: 'http://my-varnish-proxy'
        shared_max_age: 3600
```

I've also opened a PR to add Varnish with a compatible setup to the API Platform Docker setup: api-platform/api-platform#238.
Last but not least, this PR introduces some config options to set global cache settings. Example:

```yaml
api_platform:
    http_cache:
        max_age: 60
        shared_max_age: 3600
        vary: ['Content-Type', 'Cookie']
```

Note: for advanced needs, prefer the awesome [FosHttpCache library](foshttpcache.readthedocs.io).

As stated in the quote, cache invalidation **is** a hard thing and this PR probably contains bugs and edge cases. Please test it and report any problem.

TODO:
* [x] Add unit tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
         "doctrine/orm": "^2.5",
         "doctrine/annotations": "^1.2",
         "friendsofsymfony/user-bundle": "^2.0",
+        "guzzlehttp/guzzle": "^6.0",
         "nelmio/api-doc-bundle": "^2.11.2",
         "php-mock/php-mock-phpunit": "^1.1",
         "phpdocumentor/reflection-docblock": "^3.0",
@@ -63,6 +64,7 @@
     },
     "suggest": {
         "friendsofsymfony/user-bundle": "To use the FOSUserBundle bridge.",
+        "guzzlehttp/guzzle": "To use the HTTP cache invalidation system.",
         "phpdocumentor/reflection-docblock": "To support extracting metadata from PHPDoc.",
         "psr/cache-implementation": "To use metadata caching.",
         "symfony/cache": "To have metadata caching when using Symfony integration.",
--- a/features/http_cache/tags.feature
+++ b/features/http_cache/tags.feature
@@ -1,0 +1,104 @@
+Feature: Cache invalidation through HTTP Cache tags
+  In order to have a fast API
+  As an API software developer
+  I need to store API responses in a cache
+
+  @createSchema
+  Scenario: Create some embedded resources
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/relation_embedders" with body:
+    """
+    {
+      "anotherRelated": {
+        "name": "Related",
+        "thirdLevel": {}
+      }
+    }
+    """
+    Then the response status code should be 201
+    And the header "Cache-Tags" should not exist
+    And "/relation_embedders,/related_dummies,/third_levels" IRIs should be purged
+
+  Scenario: Tags must be set for items
+    When I send a "GET" request to "/relation_embedders/1"
+    Then the response status code should be 200
+    And the header "Cache-Tags" should be equal to "/relation_embedders/1,/related_dummies/1,/third_levels/1"
+
+  Scenario: Create some more resources
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/relation_embedders" with body:
+    """
+    {
+      "anotherRelated": {
+        "name": "Another Related",
+        "thirdLevel": {}
+      }
+    }
+    """
+    Then the response status code should be 201
+    And the header "Cache-Tags" should not exist
+
+  Scenario: Tags must be set for collections
+    When I send a "GET" request to "/relation_embedders"
+    Then the response status code should be 200
+    And the header "Cache-Tags" should be equal to "/relation_embedders/1,/related_dummies/1,/third_levels/1,/relation_embedders/2,/related_dummies/2,/third_levels/2,/relation_embedders"
+
+  Scenario: Purge item on update
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/relation_embedders/1" with body:
+    """
+    {
+      "paris": "France"
+    }
+    """
+    Then the response status code should be 200
+    And the header "Cache-Tags" should not exist
+    And "/relation_embedders,/relation_embedders/1,/related_dummies/1" IRIs should be purged
+
+  Scenario: Purge item and the related collection on update
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "DELETE" request to "/relation_embedders/1"
+    Then the response status code should be 204
+    And the header "Cache-Tags" should not exist
+    And "/relation_embedders,/relation_embedders/1,/related_dummies/1" IRIs should be purged
+
+  Scenario: Create two Relation2
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/relation2s" with body:
+    """
+    {
+    }
+    """
+    And I send a "POST" request to "/relation2s" with body:
+    """
+    {
+    }
+    """
+    Then the response status code should be 201
+
+  Scenario: Embedded collection must be listed in cache tags
+    When I send a "GET" request to "/relation2s/1"
+    Then the header "Cache-Tags" should be equal to "/relation2s/1"
+
+  Scenario: Create a Relation1
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/relation1s" with body:
+    """
+    {
+      "relation2": "/relation2s/1"
+    }
+    """
+    Then the response status code should be 201
+    And "/relation1s,/relation2s/1" IRIs should be purged
+
+  @dropSchema
+  Scenario: Update a Relation1
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/relation1s/1" with body:
+    """
+    {
+      "relation2": "/relation2s/2"
+    }
+    """
+    Then the response status code should be 200
+    And "/relation1s,/relation1s/1,/relation2s/2,/relation2s/1" IRIs should be purged
--- a/src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
+++ b/src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
@@ -1,0 +1,156 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Doctrine\EventListener;
+
+use ApiPlatform\Core\Api\IriConverterInterface;
+use ApiPlatform\Core\Api\ResourceClassResolverInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\HttpCache\PurgerInterface;
+use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Event\OnFlushEventArgs;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
+use Doctrine\ORM\PersistentCollection;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+
+/**
+ * Purges responses containing modified entities from the proxy cache.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @experimental
+ */
+final class PurgeHttpCacheListener
+{
+    private $purger;
+    private $iriConverter;
+    private $resourceClassResolver;
+    private $propertyAccessor;
+    private $tags = [];
+
+    public function __construct(PurgerInterface $purger, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
+    {
+        $this->purger = $purger;
+        $this->iriConverter = $iriConverter;
+        $this->resourceClassResolver = $resourceClassResolver;
+        $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
+    }
+
+    /**
+     * Collects tags from the previous and the current version of the updated entities to purge related documents.
+     */
+    public function preUpdate(PreUpdateEventArgs $eventArgs)
+    {
+        $object = $eventArgs->getObject();
+        $this->gatherResourceAndItemTags($object, true);
+
+        $changeSet = $eventArgs->getEntityChangeSet();
+        $associationMappings = $eventArgs->getEntityManager()->getClassMetadata(ClassUtils::getClass($eventArgs->getObject()))->getAssociationMappings();
+
+        foreach ($changeSet as $key => $value) {
+            if (!isset($associationMappings[$key])) {
+                continue;
+            }
+
+            $this->addTagsFor($value[0]);
+            $this->addTagsFor($value[1]);
+        }
+    }
+
+    /**
+     * Collects tags from inserted and deleted entities, including relations.
+     */
+    public function onFlush(OnFlushEventArgs $eventArgs)
+    {
+        $em = $eventArgs->getEntityManager();
+        $uow = $em->getUnitOfWork();
+
+        foreach ($uow->getScheduledEntityInsertions() as $entity) {
+            $this->gatherResourceAndItemTags($entity, false);
+            $this->gatherRelationTags($em, $entity);
+        }
+
+        foreach ($uow->getScheduledEntityUpdates() as $entity) {
+            $this->gatherResourceAndItemTags($entity, true);
+            $this->gatherRelationTags($em, $entity);
+        }
+
+        foreach ($uow->getScheduledEntityDeletions() as $entity) {
+            $this->gatherResourceAndItemTags($entity, true);
+            $this->gatherRelationTags($em, $entity);
+        }
+    }
+
+    /**
+     * Purges tags collected during this request, and clears the tag list.
+     */
+    public function postFlush()
+    {
+        $this->purger->purge($this->tags);
+        $this->tags = [];
+    }
+
+    private function gatherResourceAndItemTags($entity, bool $purgeItem)
+    {
+        try {
+            $resourceClass = $this->resourceClassResolver->getResourceClass($entity);
+        } catch (InvalidArgumentException $e) {
+            return;
+        }
+
+        $iri = $this->iriConverter->getIriFromResourceClass($resourceClass);
+        $this->tags[$iri] = $iri;
+        if ($purgeItem) {
+            $iri = $this->iriConverter->getIriFromItem($entity);
+            $this->tags[$iri] = $iri;
+        }
+    }
+
+    private function gatherRelationTags(EntityManagerInterface $em, $entity)
+    {
+        $associationMappings = $em->getClassMetadata(ClassUtils::getClass($entity))->getAssociationMappings();
+        foreach (array_keys($associationMappings) as $property) {
+            $this->addTagsFor($this->propertyAccessor->getValue($entity, $property));
+        }
+    }
+
+    private function addTagsFor($value)
+    {
+        if (!is_array($value) && !$value instanceof \Traversable) {
+            $this->addTagForItem($value);
+
+            return;
+        }
+
+        if ($value instanceof PersistentCollection) {
+            $value = clone $value;
+        }
+
+        foreach ($value as $v) {
+            $this->addTagForItem($v);
+        }
+    }
+
+    private function addTagForItem($value)
+    {
+        try {
+            $iri = $this->iriConverter->getIriFromItem($value);
+            $this->tags[$iri] = $iri;
+        } catch (InvalidArgumentException $e) {
+        } catch (RuntimeException $e) {
+        }
+    }
+}
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -19,9 +19,11 @@ use phpDocumentor\Reflection\DocBlockFactoryInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\DirectoryResource;
+use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
@@ -87,15 +89,18 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             $loader->load('security.xml');
         }
 
+        $useDoctrine = isset($bundles['DoctrineBundle']) && class_exists(Version::class);
+
         $this->registerMetadataConfiguration($container, $loader, $bundles, $config['loader_paths']);
         $this->registerOAuthConfiguration($container, $config, $loader);
         $this->registerSwaggerConfiguration($container, $config, $loader);
         $this->registerJsonLdConfiguration($formats, $loader);
         $this->registerJsonHalConfiguration($formats, $loader);
         $this->registerJsonProblemConfiguration($errorFormats, $loader);
-        $this->registerBundlesConfiguration($bundles, $config, $loader);
+        $this->registerBundlesConfiguration($bundles, $config, $loader, $useDoctrine);
         $this->registerCacheConfiguration($container);
-        $this->registerDoctrineExtensionConfiguration($container, $config);
+        $this->registerDoctrineExtensionConfiguration($container, $config, $useDoctrine);
+        $this->registerHttpCache($container, $config, $loader, $useDoctrine);
     }
 
     /**
@@ -129,6 +134,11 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.collection.pagination.page_parameter_name', $config['collection']['pagination']['page_parameter_name']);
         $container->setParameter('api_platform.collection.pagination.enabled_parameter_name', $config['collection']['pagination']['enabled_parameter_name']);
         $container->setParameter('api_platform.collection.pagination.items_per_page_parameter_name', $config['collection']['pagination']['items_per_page_parameter_name']);
+        $container->setParameter('api_platform.http_cache.etag', $config['http_cache']['etag']);
+        $container->setParameter('api_platform.http_cache.max_age', $config['http_cache']['max_age']);
+        $container->setParameter('api_platform.http_cache.shared_max_age', $config['http_cache']['shared_max_age']);
+        $container->setParameter('api_platform.http_cache.vary', $config['http_cache']['vary']);
+        $container->setParameter('api_platform.http_cache.public', $config['http_cache']['public']);
 
         $container->setAlias('api_platform.operation_path_resolver.default', $config['default_operation_path_resolver']);
 
@@ -308,11 +318,12 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
      * @param string[]      $bundles
      * @param array         $config
      * @param XmlFileLoader $loader
+     * @param bool          $useDoctrine
      */
-    private function registerBundlesConfiguration(array $bundles, array $config, XmlFileLoader $loader)
+    private function registerBundlesConfiguration(array $bundles, array $config, XmlFileLoader $loader, bool $useDoctrine)
     {
         // Doctrine ORM support
-        if (isset($bundles['DoctrineBundle']) && class_exists(Version::class)) {
+        if ($useDoctrine) {
             $loader->load('doctrine_orm.xml');
         }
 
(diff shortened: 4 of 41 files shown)


src/HttpCache/EventListener/PurgeListener.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->purger = $purger;
+        $this->iriConverter = $iriConverter;
+    }
+
+    /**
+     * Stores the original IRI of the object to be deleted because it may not be available anymore at purge time (e.g. when using Doctrine).
+     *
+     * @param GetResponseForControllerResultEvent $event
+     */
+    public function onKernelView(GetResponseForControllerResultEvent $event)
```
dunglas: Probably useless because it is already done for Doctrine and not reliable enough for non-Doctrine.

src/HttpCache/EventListener/PurgeListener.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $this->addResources($event->getRequest());
+
+        if ($iris = $this->resourceManager->getResources()) {
+            $this->purger->purge($iris);
+            $this->resourceManager->clear();
+        }
+    }
+
+    private function addResources(Request $request)
```
dunglas: Probably useless because it is already handled for Doctrine and not reliable enough for non-Doctrine data providers.

fbourigault:
This looks amazing, but does providing a solution coupled to specific cache solution as first class citizen a good idea?

Maybe an intermediate solution build using standard Http headers would improve interoperability.

bendavies:
Nice work @dunglas.

This is a very similar strategy to what i've implemented previously.  
A few comments:

1. why not just use FosHttpCache instead of implementing your own tagging/purgers etc?  
2. It does no look like this will work with authenticated apis that don't use cookies (FosHttpCache has provisions for user specific caching)
3. why do you md5 the tags?

composer.json
```diff
@@ -34,6 +34,7 @@
         "doctrine/orm": "^2.5",
         "doctrine/annotations": "^1.2",
         "friendsofsymfony/user-bundle": "^2.0@dev",
+        "guzzlehttp/guzzle": "^6.0",
```
fbourigault: You should use `php-http/httplug` for better interop.

dunglas:
@fbourigault there is an abstraction layer and and implementation (Varnish). Basically, anyone can add support for any cache provider supporting cache invalidation (it's a matter of implementing an interface). I plan to add support CloudFlare in core too.

There is no standard (HTTP headers) for cache invalidation (only expiration is supported in RFCs) but I choose to use `Cache-Tags` because it's the one used by CloudFlare.

@bendavies regarding FosHttpCache, for 2 reasons:

* my implementation is a bit different (the md5 hash) and tight to the concept of "resources" and "IRIs", not present in FosHttpCache
* the implementation is trivial (and it's possible to bridge it with FosHttpCache if wanted), it will ease our maintenance process to not have a dependency to a 3rd party library we don't maintain (our soft dependency to FosUser is a pain to maintain...).

It can work with any mechanism of authentification (it's why I've introduced a way to configure the `Vary` HTTP header). If the resource varies depending of the `Authorization` header, just set `api_platform.vary` to `['Content-Type', 'Authorization']` (you'll need to tweak the Varnish config too). A key concept of REST is being explicit, a resource identified by an IRI should not vary depending of the current logged in user (in this case, the URL should not be the same).

3/ To avoid collision (if the a resource is tagged with `/foos/,/foos1` and another with only `/foos/2`, sending a BAN request with `/foos` as parameter will ban both, and it's a bug. Using md5 hashes prevent this problem. A CRC32 hash should do the trick too, but md5 is usually faster on modern servers.

composer.json
dunglas: Maybe when it will be a PSR. But for now it has nos advantage over using Guzzle 6 for this use case: it is harder to setup (requires to install an extra Symfony bundle), adds complexity and a performance overhead.
Guzzle 6 is already an abstraction layer, with builtin curl and streams implementations. We only depend of the `ClientInterface`, it's easy to bridge other implementations (but I think it is not worth it).

As we support PHP 7 only, we don't target older libraries such as Guzzle 5.

fbourigault: I didn't catch the last point. :+1:

fbourigault:
> There is no standard (HTTP headers) for cache invalidation (only expiration is supported in RFCs) but I choose to use Cache-Tags because it's the one used by CloudFlare.

By standard HTTP headers, I mean revalidation. But maybe in such case, caching is not efficient.

teohhanhui:
> There is no standard (HTTP headers) for cache invalidation (only
expiration is supported in RFCs) but I choose to use Cache-Tags because
it's the one used by CloudFlare.

Custom HTTP headers should be prepended by `X-`. CloudFlare is flaunting
the spec (bad).

> If the resource varies depending of the Authorization header, just
set api_platform.vary to ['Content-Type', 'Authorization'] (you'll need to
tweak the Varnish config too).

It's not so simple. An authorization context needs to be provided and
hashed in Varnish. Blindly adding `Vary: Authorization` would kill hit
rates, if your intention is to keep a cache for each user, for example.

> A key concept of REST is being explicit, a resource identified by an IRI
should not vary depending of the current logged in user (in this case, the
URL should not be the same).

False. A resource can vary by any header as declared in the Vary header.

teohhanhui:
> To avoid collision (if the a resource is tagged with /foos/,/foos1 and
another with only /foos/2, sending a BAN request with /foos as parameter
will ban both, and it's a bug. Using md5 hashes prevent this problem. A
CRC32 hash should do the trick too, but md5 is usually faster on modern
servers.

Just use regex. Varnish has good regex support. Using a hash is a nightmare
when trying to debug cache tags. :)

dunglas:
I'll make the header configurable but I'll keep CloudFlare compatibility by default, it's an invaluable feature. By the way this header should not be exposed to the end client.

Regarding Vary, it's exactly what I explain in my post 🙂

Regarding the hit rate, doing more advanced things like hashing cannot be automated on the API Platform side, it requires a custom development and it's easy to implement using the new vary option.
Using FosHTTPCache for such usages is also an option, it's up to the developer.

dunglas:
Regex (at least simple regexes) doesn't fix the issue in my example. /foos will match /foos/*, not only the collection response.

teohhanhui:
/^\/foos$/

dunglas:
But maybe a more complex regex can do the trick, I'll give it a try (I agree that plain tags are better than hashes for debug).

dunglas:
The comma must be handled too.

features/http_cache/tags.feature
```diff
@@ -0,0 +1,64 @@
...
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/relation_embedders/1" with body:
+    """
+    {
+      "paris": "France"
+    }
+    """
+    Then the response status code should be 200
+    And the header "Cache-Tags" should not exist
+    And "/relation_embedders/1,/related_dummies/1,/third_levels/1" IRIs should be purged
```
jderusse: A PUT may change the value of an attribute used to order a responses. You should invalidate collection too

features/http_cache/tags.feature
```diff
@@ -0,0 +1,64 @@
...
+
+  @dropSchema
+  Scenario: Purge item and the related collection on update
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "DELETE" request to "/relation_embedders/1"
+    Then the response status code should be 204
+    And the header "Cache-Tags" should not exist
+    And "/relation_embedders/1,/relation_embedders" IRIs should be purged
```
jderusse: What's about `related_dummies` who was previously linked to this entity? They should be invalidated too

features/http_cache/tags.feature
```diff
@@ -0,0 +1,64 @@
...
+    {
+      "anotherRelated": {
+        "name": "Related",
+        "thirdLevel": {}
+      }
+    }
+    """
+    Then the response status code should be 201
+    And the header "Cache-Tags" should not exist
+    And "/relation_embedders,/related_dummies,/third_levels,/relation_embedders/1,/related_dummies/1,/third_levels/1" IRIs should be purged
```
jderusse: I don't understand in you code how you make the link between `relation_embedders` and `third_levels`? Who flaged it?

features/http_cache/tags.feature
```diff
@@ -0,0 +1,64 @@
...
+
+  Scenario: Tags must be set for items
+    When I send a "GET" request to "/relation_embedders/1"
+    Then the response status code should be 200
+    And the header "Cache-Tags" should be equal to "aa9e2bee5be20590f7dcc520ce2dffca,12a0c94f947a680d68bd6f65e025457d,91774d67418192a057e25dae00345572"
```
jderusse: md5 obfuscate tests... :( 
I wonder if a dedicated context could be more readable: `the cache-tag header should be equals to "/relation_embedders, ...."`

bendavies:
@dunglas 
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/Varnish.php#L59-L61
https://regex101.com/r/vqmNLw/2

composer.json
lyrixx: @dunglas IMHO, you should consider using buzz. And It's not a joke:

* Guzzle have so many version, and these versions are not compatible each others. So people that are using a version that is incompatible with guzzle 6 will be blocked.
* Buzz is very simple and works well

features/http_cache/tags.feature
dunglas: They will be, because they are marked with `/relation_embedders/1`.

features/http_cache/tags.feature
dunglas: Good catch, I add this.

features/http_cache/tags.feature
dunglas: It's because of this line: https://github.com/api-platform/core/pull/952/files#diff-0362bfae91549bcd08af0d06bd806b37R14

features/http_cache/tags.feature
dunglas: True, I'm working on a fix to remove the hash and use plain IRIs. I'll update the PR when it's ready.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -92,6 +92,23 @@ public function getConfigTreeBuilder()
...
+                        ->integerNode('max_age')->defaultNull()->info('Default value for the response max age.')->end()
+                        ->integerNode('shared_max_age')->defaultNull()->info('Default value for the response shared (proxy) max age.')->end()
+                        ->arrayNode('vary')
+                            ->defaultValue(['Content-Type'])
+                            ->prototype('scalar')->end()
+                            ->info('Default values of the "Vary" HTTP header.')
+                        ->end()
+                        ->booleanNode('public')->defaultNull()->info('To make all responses public by default.')->end()
+                        ->booleanNode('enable_tags')->defaultFalse()->info('Add cache tags to the response.')->end()
+                        ->scalarNode('varnish_url')->defaultNull()->info('URL of the Varnish server to purge using cache tags when a resource is updated.')->end()
```
jderusse: Don't forget user with several varnish servers

dunglas:
Thank you for the reviews everyone!

* Plain IRIs are now used instead of md5 hashes (and the regex have been updated accordingly)
* Collections are purged even when a PUT is done

composer.json
dunglas: AFAIK it doesn't support async requests... and it has not been updated since 2015.

fbourigault: When guzzle guys will release the 7.0 version, you will be stuck with the 6.0 version. Http clients are a widely common dependency and so conflicts may appears. But ATM you could go with guzzle 6. Using httplug could be added later.

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    public function __construct($container)
+    {
+        $this->container = $container;
+    }
+
+    public function onFlush(OnFlushEventArgs $eventArgs)
+    {
+        $iriConverter = $this->container->get('api_platform.iri_converter');
+        $resourceManager = $this->container->get('api_platform.http_cache.resource_manager');
```
soyuka: Circular reference is only because of IriConverter no? Can't you inject the ResourceManager properly though?

src/HttpCache/VarnishPurger.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function purge(array $iris)
+    {
+        if (!$iris) {
+            return;
+        }
+
+        $parts = array_map(function ($iri) {
```
soyuka: A small comment to explain what goes below would be great for future readers!

src/Bridge/Symfony/Bundle/Resources/config/http_cache_tags.xml
```diff
@@ -0,0 +1,27 @@
...
+
+            <tag name="kernel.event_listener" event="kernel.response" method="onKernelResponse" />
+        </service>
+
+        <!-- The onKernelView method must be executed after the validation (32) but before the write (64) -->
+        <service id="api_platform.http_cache.listener.response.purge" class="ApiPlatform\Core\HttpCache\EventListener\PurgeListener">
+            <argument type="service" id="api_platform.http_cache.resource_manager" />
+            <argument type="service" id="api_platform.http_cache.purger" />
+
+            <tag name="kernel.event_listener" event="kernel.terminate" method="onKernelTerminate" />
```
jderusse: I suggest to move the purge as close as possible to the transaction (doctrine postFlush IMO)

I understand your point of view to return a response to the user as soon as possible, but this introduce cross concurrency bugs.

Example:
- list all blogs (cache mis)
- post a blog
- list all blog (cache hit when called before the kernel terminate even)

I already have this issue in the my CI (agreed, this is a particular case with a lot of stress tests). but the probability of cross concurrency is higher when:
- you have several varnishes to purge
- you have other kernel.terminate events with an higher priority (sending an email for instance)

src/HttpCache/EventListener/ConfigureListener.php
```diff
@@ -0,0 +1,59 @@
...
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+final class ConfigureListener
+{
+    private $maxAge;
+    private $sharedMaxage;
+    private $vary;
+    private $public;
+
+    public function __construct(int $maxAge = null, int $sharedMaxAge = null, array $vary, bool $public = null)
```
jderusse: I suggest to add a LastModified header too initialized to the default value `now()`.
As a result, Varnish will handled it for free (by default), and returns 304 code stuff.

This small change should not interfer in your App and how you invalid cache. But will save bandwidth between the user and varnish.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -288,6 +296,41 @@ private function registerDoctrineExtensionConfiguration(ContainerBuilder $contai
         }
     }
 
+    private function registerHttpCache(ContainerBuilder $container, array $config, XmlFileLoader $loader)
+    {
+        $loader->load('http_cache.xml');
+
+        if (true !== $config['http_cache']['enable_tags']) {
```
jderusse: When `enable_tags` is true and `maxAge` > 0. We should warn the developer that the browser will cache responses and remove the benefits of cache invalidation.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -92,6 +92,27 @@ public function getConfigTreeBuilder()
...
+                    ->children()
+                        ->integerNode('max_age')->defaultNull()->info('Default value for the response max age.')->end()
+                        ->integerNode('shared_max_age')->defaultNull()->info('Default value for the response shared (proxy) max age.')->end()
+                        ->arrayNode('vary')
+                            ->defaultValue(['Content-Type'])
+                            ->prototype('scalar')->end()
+                            ->info('Default values of the "Vary" HTTP header.')
+                        ->end()
+                        ->booleanNode('public')->defaultNull()->info('To make all responses public by default.')->end()
+                        ->booleanNode('enable_tags')->defaultFalse()->info('Add cache tags to the response.')->end()
```
jderusse: it's not only about adding tags in response, it's also about cache invalidation.

src/HttpCache/VarnishPurger.php
```diff
@@ -0,0 +1,46 @@
...
+            return sprintf('(^|\,)%s($|\,)', preg_quote($iri));
```
jderusse: You may store the tag with a `,` at the beginning and ending to simplify this RegExp and optimizeinvalidation (especially when you invalidate several tags at the time).
Regexp would became `,%s,`

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -0,0 +1,71 @@
...
+        $uow = $eventArgs->getEntityManager()->getUnitOfWork();
+
+        foreach ($uow->getScheduledEntityInsertions() as $entity) {
+            $this->purge($iriConverter, $resourceManager, $entity, false);
```
jderusse: are you sure to be able to retrieves an auto-increment id in the `onFlush` eventListener? I thought, this ID was injected *after* the insert operation

src/HttpCache/EventListener/ConfigureListener.php
teohhanhui: Varnish already returns 304 response properly. We should not send bogus `Last-Modified` headers.

src/Bridge/Symfony/Bundle/Resources/config/http_cache_tags.xml
teohhanhui: ^ That does not actually address the race condition, but only trying to avoid it.

jderusse: indeed, it fixes one case.

BTW (not related to the RC): A small improvment for the user would be to start async request on the `postFlush` event, and wait on the `kernelFinish` event.

src/HttpCache/EventListener/ConfigureListener.php
jderusse: I thought this listener was used only with "tag" thing. I apologize, agreed, this header don't make sens here.

But I think, you missed my point, the idea, when using tagged response, is to add a validation header (could be etag or last-modified, don't care). To save bandwidth between users and varnish when cache is hit.

dunglas: IMO we should provide a good default config for this (in the standard edition) but let the user configure it as he wants.

teohhanhui: @jderusse:
> the idea, when using tagged response, is to add a validation header (could be etag or last-modified, don't care). To save bandwidth between users and varnish when cache is hit.

That's incorrect. Varnish already sends 304 response with proper `E-Tag` in that case. There is no need for the backend to send an `E-Tag` or `Last-Modified` header unless the backend intends to return 304 on `vcl_backend_fetch`.

composer.json
teohhanhui: I have a very strong (I think) argument for using HTTPlug: you do not get into hell like FOSHttpCache did (and still is) with their hard coupling to a specific Guzzle version.

src/HttpCache/EventListener/ConfigureListener.php
jderusse: Of course varnish deals with validation headers and return 304 responses. I'd never said to handle the validation in the app. BUT if you don't send validation headers in the response, clients will never send validations headers back in the next request (if-none-match and if-modified-since) and varnish will never have the possibility to use it. Here is my point.
It's free, it don't change the behavior of your application (if you don't deal invalidation on app side) and it save bandwidth.

teohhanhui: > BUT if you don't send validation headers in the response, clients will never send validations headers back in the next request (if-none-match and if-modified-since) and varnish will never have the possibility to use it.

Varnish does it. :smile:

jderusse: Let explain with a schema (my last try)...
Here is the actual implementation
now

My objective is to transform the last response in step **6** into a 304.

* To do it, varnish need a `if-none-match` header in step **5**
* To do it, client need a `etag` header in step **4**
* To do it, varnish needs nothing, it just forward the response from the app from step **3**
* To do it, app must send this header <= this is my point

And Here is the target sequence
then



Your last comment suggest that varnish add this header by it self, with something like that
other

Is it what you mean? Because I doubt varnish add such header by itself (unless you implement this behavior in a VCL which is fine to me too)

teohhanhui: Varnish does it by default. Try it. At least it has been doing so since 4.0 :smile:

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
jderusse: nevermind, if missed the option `false` which means: don't purge the item itself :)

jderusse:
beware, this PR sends cache headers on error 500 too
```
curl test2_nginx_1.docker/foos -I
HTTP/1.1 500 Internal Server Error
Content-Type: application/ld+json; charset=utf-8
Cache-Control: public, s-maxage=3600
Link: <http://test2_nginx_1.docker/docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"
Vary: Content-Type
Cache-Tags: /foos
Date: Wed, 22 Mar 2017 07:20:44 GMT
```

src/HttpCache/EventListener/ConfigureListener.php
jderusse: I bootstraped an empty api-plateform from this PR https://github.com/api-platform/api-platform/pull/238

```
curl test2_varnish_1.docker/foos -i
HTTP/1.1 200 OK
Server: nginx/1.11.10
Content-Type: application/ld+json; charset=utf-8
X-Powered-By: PHP/7.1.3
Vary: Content-Type
X-Content-Type-Options: nosniff
X-Frame-Options: deny
Cache-Control: public, s-maxage=3600
Link: <http://test2_varnish_1.docker/docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"
Cache-Tags: /foos
Date: Wed, 22 Mar 2017 07:23:50 GMT
X-Varnish: 26 32789
Age: 3
Via: 1.1 varnish-v4
Accept-Ranges: bytes
Content-Length: 111
Connection: keep-alive

{"@context":"\/contexts\/Foo","@id":"\/foos","@type":"hydra:Collection","hydra:member":[],"hydra:totalItems":0}
```

varnish don't add validation header :confused:

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param FilterResponseEvent $event
+     */
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        $request = $event->getRequest();
+        if (
+            !$request->isMethodCacheable()
+            || !$attributes = RequestAttributesExtractor::extractAttributes($request)
+        ) {
```
teohhanhui: To address @jderusse's [comment](https://github.com/api-platform/core/pull/952#issuecomment-288318323), we'd add another condition here:

```php
!in_array($response->getStatusCode(), [
    Response::HTTP_OK,
    Response::HTTP_NO_CONTENT,
])
```

src/HttpCache/EventListener/ConfigureListener.php
teohhanhui: @jderusse You're right. I overlooked that FOSHttpCacheBundle was the one adding it. My apologies...

I think we should use the same approach. Just use the MD5 hash of the `Response` content as `ETag`. MD5 is cheap, so no reason to use a meaningless `Last-Modified`. :smile:

dunglas:
I'll refactor this PR to get rid of the stateful service.

dunglas:
About

> Custom HTTP headers should be prepended by `X-`. CloudFlare is flaunting
the spec (bad).

This convention has been deprecated in RFC 6648. See https://specs.openstack.org/openstack/api-wg/guidelines/headers.html. I'll remove all `X-` prefix from this PR.

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->container = $container;
+    }
+
+    public function onFlush(OnFlushEventArgs $eventArgs)
+    {
+        if (!$request = $this->container->get('request_stack')->getCurrentRequest()) {
+            return;
+        }
+
+        $resources = $request->attributes->get('_resources', []);
```
jderusse: This tight couple the purge system with http requests context which is used like a databag.

dunglas: Yes I'll try to find a better solution but I'm not sure there is one. Anyway its called `purge HTTP cache`, it's not a big deal to couple it with the HTTP request.

src/HttpCache/EventListener/ConfigureListener.php
dunglas: Doing a md5 of the response is not ok (the response contain the date). We may do a md5 of the response's body.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
dunglas: It can be intended (I often do that). It allows to always serve fresh data for the first request but reduce the server load/bandwidth using expiration. It's often not a problem when the TTL is low. I would let the user do what he wants without warning.

src/HttpCache/VarnishPurger.php
dunglas: I'm not a fond of this, the header become ugly. It may have other use cases for this headers (e.g. : building a smart client-side cache).

src/HttpCache/EventListener/AddTagsListener.php
dunglas: done

dunglas:
All comments handled. Can you make a last review?

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -46,13 +46,16 @@ public function normalize($object, $format = null, array $context = [])
...
+        $context['api_normalize'] = true;
 
         $rawData = parent::normalize($object, $format, $context);
         if (!is_array($rawData)) {
             return $rawData;
         }
 
-        $data = ['_links' => ['self' => ['href' => $this->iriConverter->getIriFromItem($object)]]];
+        $data = ['_links' => ['self' => ['href' => $context['iri']]]];
+        $context['debug'] = true;
```
sroze: Is there any reason for this property in the context?

features/http_cache/tags.feature
```diff
@@ -0,0 +1,64 @@
+Feature: Cache invalidation trough HTTP Cache tags
```
soyuka: through?

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -0,0 +1,71 @@
...
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        $request = $event->getRequest();
+        $response = $event->getResponse();
+
+        if (
+            !$request->isMethodCacheable()
+            || !$response->isCacheable()
+            || (!$attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !$resources = $request->attributes->get('_resources')
```
soyuka: no parenthesis here, but the line above yes?

tests/Serializer/ItemNormalizerTest.php
```diff
@@ -94,7 +95,7 @@ public function testNormalize()
         );
         $normalizer->setSerializer($serializerProphecy->reveal());
 
-        $this->assertEquals(['name' => 'hello'], $normalizer->normalize($dummy));
+        $this->assertEquals(['name' => 'hello'], $normalizer->normalize($dummy, null, ['resources' => []]));
```
soyuka: Shouldn't we avoid changing those tests? It feels like the test won't pass without 2 more arguments (eg, `null, ['resources' => []]`). Is it the case / Isn't this breaking things somehow?

src/Hal/Serializer/ItemNormalizer.php
dunglas: Good catch, it has nothing to do here!

src/HttpCache/EventListener/AddTagsListener.php
dunglas: The parenthesis on the previous line are mandatory because of operator priorities.

tests/Serializer/ItemNormalizerTest.php
dunglas: It passes without the arguments and without change, but I've modified it to test the new behavior.

Simperfit:
I'm going to test this with a project, ill tell if there are some bugs

tests/Serializer/ItemNormalizerTest.php
soyuka: Okay as long as there is a test that doesn't adds up those arguments.

patkar:
May you use http://httplug.io/ instead of guzzle directly?

teohhanhui:
> May you use http://httplug.io/ instead of guzzle directly?

It has been raised before, and I'd like to once again echo this.

dunglas:
HTTPPlug introduces a lot of complexity (including an extra bundle to configure... until we get Flex) for no gain here. I'm still 👎 for now.
I understand the problem with Guzzle 3, but API Platform is younger than Guzzle 3, every bundle I use are compatible with Guzzle 6, and AFAIK, there is no projet to develop Guzzle 7 soon.

dunglas:
And by the way, Guzzle is a soft dependency here. So someone wanting to use another client can do it, he just have to implement himself the `PurgerInterface`, not a big deal.

theofidry:
As long as Guzzle is a soft dependency it's ok

PR #875, opened 2016-12-09
Allow use of pagination_maximum_items_per_page per resource

fmata:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | pending approbation of this PR to create it

Now it's only possible to indicate the maximum number of results in a collection for all resources.
Depends on complexity of one resource, we want to limit the maximum number of results in a collection to avoid leak of system resources.

This PR adds the attribute "pagination_maximum_items_per_page" to allow this customization.
If it's ok, I can create the doc PR.

PR #955, opened 2017-02-24
Added PaginationHelper class

jonadoe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #875 
| License       | MIT
| Doc PR        | none yet

The main intention for this PR is to make it easier to implement pagination for custom data providers and to make it more independent from the ORM. 
This is my first contribution to a public project, so please be patient if some things are not perfect yet. I will add tests and doc pr if you think the code changes are ok. 
I'm not sure if the name and namespace of the ApiPlatform\Core\Util\PaginationHelper class really fits.

Changes in this PR:
- Added PaginationHelper class to make it easier to implement pagination for custom data providers
- Added pagination fields to swagger api doc
- Changed PaginationExtension to use PaginationHelper class
- Fixed bug in PaginationExtension->isPaginationEnabled (code is now in PaginationHelper class. Use $enabled instead of $this->enabled as default when getting clientEnabled var) (is clientEnabled really useful? I think it should be dropped)
- Allow use of pagination_maximum_items_per_page per resource (Issue #875)

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -52,7 +52,7 @@ final class PaginationExtension implements QueryResultCollectionExtensionInterfa
             return;
         }
 
-        $this->paginationHelper->setResourceAndOperation( $resourceClass, $operationName );
+        $this->paginationHelper->setResourceAndOperation($resourceClass, $operationName);
         if (!$this->paginationHelper->isPaginationEnabled()) {
             return;
         }
@@ -75,7 +75,7 @@ final class PaginationExtension implements QueryResultCollectionExtensionInterfa
             return false;
         }
 
-        $this->paginationHelper->setResourceAndOperation( $resourceClass, $operationName );
+        $this->paginationHelper->setResourceAndOperation($resourceClass, $operationName);
 
         return $this->paginationHelper->isPaginationEnabled();
     }
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -15,7 +15,7 @@ use ApiPlatform\Core\Bridge\Doctrine\Orm\Paginator;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryChecker;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
-use ApiPlatform\Core\Util\PaginationHelper;
+use ApiPlatform\Core\Util\PaginationHelperFactory;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\Paginator as DoctrineOrmPaginator;
@@ -32,14 +32,18 @@ final class PaginationExtension implements QueryResultCollectionExtensionInterfa
     private $managerRegistry;
     private $requestStack;
     private $resourceMetadataFactory;
-    private $paginationHelper;
+    private $paginationHelperFactory;
 
-    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, PaginationHelper $paginationHelper)
+    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, bool $enabled = true, bool $clientEnabled = false, bool $clientItemsPerPage = false, int $itemsPerPage = 30, string $pageParameterName = 'page', string $enabledParameterName = 'pagination', string $itemsPerPageParameterName = 'itemsPerPage', int $maximumItemPerPage = null, PaginationHelperFactory $paginationHelperFactory = null)
     {
         $this->managerRegistry = $managerRegistry;
         $this->requestStack = $requestStack;
         $this->resourceMetadataFactory = $resourceMetadataFactory;
-        $this->paginationHelper = $paginationHelper;
+
+        if (!$paginationHelperFactory) {
+            $paginationHelperFactory = new PaginationHelperFactory($requestStack, $resourceMetadataFactory, $enabled, $clientEnabled, $clientItemsPerPage, $itemsPerPage, $pageParameterName, $enabledParameterName, $itemsPerPageParameterName, $maximumItemPerPage);
+        }
+        $this->paginationHelperFactory = $paginationHelperFactory;
     }
 
     /**
@@ -52,16 +56,15 @@ final class PaginationExtension implements QueryResultCollectionExtensionInterfa
             return;
         }
 
-        $this->paginationHelper->setResourceAndOperation($resourceClass, $operationName);
-        if (!$this->paginationHelper->isPaginationEnabled()) {
+        $paginationHelper = $this->paginationHelperFactory->create($resourceClass, $operationName);
+        if (!$paginationHelper->isPaginationEnabled()) {
             return;
         }
 
-        $itemsPerPage = $this->paginationHelper->getItemsPerPage();
-        $page = $this->paginationHelper->getPage();
+        $itemsPerPage = $paginationHelper->getItemsPerPage();
 
         $queryBuilder
-            ->setFirstResult(($page - 1) * $itemsPerPage)
+            ->setFirstResult(($paginationHelper->getPage() - 1) * $itemsPerPage)
             ->setMaxResults($itemsPerPage);
     }
 
@@ -75,9 +78,7 @@ final class PaginationExtension implements QueryResultCollectionExtensionInterfa
             return false;
         }
 
-        $this->paginationHelper->setResourceAndOperation($resourceClass, $operationName);
-
-        return $this->paginationHelper->isPaginationEnabled();
+        return $this->paginationHelperFactory->create($resourceClass, $operationName)->isPaginationEnabled();
     }
 
     /**
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -15,11 +15,10 @@ use ApiPlatform\Core\Bridge\Doctrine\Orm\Paginator;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryChecker;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
-use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Util\PaginationHelper;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\Paginator as DoctrineOrmPaginator;
-use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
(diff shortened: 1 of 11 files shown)


src/Util/PaginationHelper.php
```diff
@@ -0,0 +1,197 @@
...
+     * @return null
+     */
+    public function getConfig($name)
+    {
+        if (!in_array($name, ['enabled', 'clientEnabled', 'clientItemsPerPage', 'itemsPerPage', 'maximumItemsPerPage', 'parameterNamePage', 'parameterNameEnabled', 'parameterNameItemsPerPage'])) {
+            return null;
+        }
+
+        return $this->$name;
+    }
```
soyuka: Having getters for those would be prefered IMO.

src/Util/PaginationHelper.php
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * @param string $resourceClass
+     * @param string|null $operationName
+     */
+    public function setResourceAndOperation(string $resourceClass, string $operationName = null)
+    {
+        $this->resourceClass = $resourceClass;
+        $this->operationName = $operationName;
+    }
```
soyuka: Not sure that this is the best way to do it, especially if other methods rely on those two variables.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -33,28 +32,14 @@
...
-    private $itemsPerPage;
-    private $pageParameterName;
-    private $enabledParameterName;
-    private $itemsPerPageParameterName;
-    private $maximumItemPerPage;
-
-    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, bool $enabled = true, bool $clientEnabled = false, bool $clientItemsPerPage = false, int $itemsPerPage = 30, string $pageParameterName = 'page', string $enabledParameterName = 'pagination', string $itemsPerPageParameterName = 'itemsPerPage', int $maximumItemPerPage = null)
+    private $paginationHelper;
+
+    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, PaginationHelper $paginationHelper)
```
soyuka: Not sure but I think that this might break things.

jonadoe: Yes, this would not be BC. But on the other side it would not make sense to keep these variables in the constructor.

src/Util/PaginationHelper.php
jonadoe: The other way would be to supply these two parameters in every other method.
Should I change it that way?

src/Util/PaginationHelper.php
jonadoe: You're right. I will change this

src/Util/PaginationHelper.php
soyuka: I'm not sure, let's hear someone else's advice on this first.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
dunglas: We cannot change the constructor until API Platform 3, and it's not planned yet. We need to add a compatibility layer (we can deprecate the old usage).

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -67,19 +52,16 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
...
             return;
         }
 
-        $itemsPerPage = $resourceMetadata->getCollectionOperationAttribute($operationName, 'pagination_items_per_page', $this->itemsPerPage, true);
-        if ($resourceMetadata->getCollectionOperationAttribute($operationName, 'pagination_client_items_per_page', $this->clientItemsPerPage, true)) {
-            $itemsPerPage = (int) $request->query->get($this->itemsPerPageParameterName, $itemsPerPage);
-            $itemsPerPage = (null !== $this->maximumItemPerPage && $itemsPerPage >= $this->maximumItemPerPage ? $this->maximumItemPerPage : $itemsPerPage);
-        }
+        $itemsPerPage = $this->paginationHelper->getItemsPerPage();
+        $page = $this->paginationHelper->getPage();
```
dunglas: There is no need to introduce this var as it is used only 1 time.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -208,6 +211,11 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
                 $pathOperation['parameters'] = $parameters;
             }
 
+            $paginationParameters = $this->getPaginationParameters($resourceClass, $operationName);
+            if (!empty($paginationParameters)) {
```
dunglas: Empty is useless here, can be `if (!$paginationParameters)`.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -208,6 +211,11 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+                $pathOperation['parameters'] = empty($pathOperation['parameters']) ? $pathOperation['parameters'] : array_merge($pathOperation['parameters'], $paginationParameters);
```
dunglas: IIUC, the test should be inverted. `empty` should be removed here too.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -230,6 +238,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
     }
 
     /**
+     * @param string $resourceClass
+     * @param string $operationName
+     *
+     * @return array
+     */
+    protected function getPaginationParameters(string $resourceClass, string $operationName)
```
dunglas: Missing return type, and should be `private` (we use `private` by default to ease the maintenance).

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -230,6 +238,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
     }
 
     /**
+     * @param string $resourceClass
+     * @param string $operationName
```
dunglas: Useless dockblock, can be removed.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -230,6 +238,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+    {
+        $this->paginationHelper->setResourceAndOperation($resourceClass, $operationName);
+
+        if (!$this->paginationHelper->isResourcePaginationEnabled()) {
+            return [];
+        }
+
+        $paginationParameters = [];
+
+        if ($this->paginationHelper->isClientPaginationEnabledByDefault()) {
```
dunglas: I'll double check but it looks weird, it should be `isClientPaginationEnabled`.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -230,6 +238,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+        $paginationParameters[] = [
+            'name' => $this->paginationHelper->getParameterNamePage(),
+            'in' => 'query',
+            'type' => 'integer',
+            'description' => 'Page number',
+            'default' => 1,
+            'minimum' => 1,
+        ];
+
+        if ($this->paginationHelper->isClientItemsPerPageEnabledByDefault()) {
```
dunglas: Same here, the name looks weird.

src/Util/PaginationHelper.php
```diff
@@ -0,0 +1,238 @@
...
+    private $clientEnabled = false;
+    private $clientItemsPerPage = false;
+    private $itemsPerPage = 30;
+    private $maximumItemsPerPage = null;
+
+    private $parameterNamePage = 'page';
+    private $parameterNameEnabled = 'pagination';
+    private $parameterNameItemsPerPage = 'itemsPerPage';
+
+    private $resourceClass;
```
dunglas: A service must not have mutable states like that. This class shouldn't be a service, it should be instantiated directly in the extension, it will allow to get rid of those states, and it's a cleaner design (better, it would be easier to be backward compatible).

src/Util/PaginationHelper.php
```diff
@@ -0,0 +1,238 @@
...
+    private $operationName;
+
+    /**
```
dunglas: Useless docblock.

src/Util/PaginationHelper.php
```diff
@@ -0,0 +1,238 @@
...
+     * @param RequestStack                     $requestStack
+     * @param ResourceMetadataFactoryInterface $resourceMetadataFactory
+     */
+    public function __construct(RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory)
+    {
+        $this->requestStack = $requestStack;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+    }
+
+    /**
```
dunglas: Useless dockblock.

src/Util/PaginationHelper.php
dunglas: See my previous comment, this should definitely be avoided.

teohhanhui:
Can you elaborate how this is related to #875?

src/Util/PaginationHelper.php
jonadoe: Class is now modified to be immutable and can be instantiated using a factory

src/Swagger/Serializer/DocumentationNormalizer.php
jonadoe: What is your evaluation criteria if a docblock is useful or not? Missing description?

dunglas: If it adds nothing more than existing type hints and return types, it is useless and should be removed to ease maintenance.

jonadoe:
@teohhanhui 
The changes regarding #875 were in fact missing. It is now included in PaginationHelper->getMaximumItemsPerPage

jonadoe:
@dunglas : Your reviews are applied, please have a look.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -33,28 +32,18 @@
...
-        $this->enabled = $enabled;
-        $this->clientEnabled = $clientEnabled;
-        $this->clientItemsPerPage = $clientItemsPerPage;
-        $this->itemsPerPage = $itemsPerPage;
-        $this->pageParameterName = $pageParameterName;
-        $this->enabledParameterName = $enabledParameterName;
-        $this->itemsPerPageParameterName = $itemsPerPageParameterName;
-        $this->maximumItemPerPage = $maximumItemPerPage;
+
+        if (!$paginationHelperFactory) {
```
dunglas: I would write: `$this->paginationHelperFactory = $paginationHelperFactory ?? new PaginationHelperFactory($requestStack, $resourceMetadataFactory, $enabled, $clientEnabled, $clientItemsPerPage, $itemsPerPage, $pageParameterName, $enabledParameterName, $itemsPerPageParameterName, $maximumItemPerPage);`

src/Util/Factory/PaginationHelperFactory.php
```diff
@@ -0,0 +1,56 @@
...
+namespace ApiPlatform\Core\Util\Factory;
+
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\PaginationHelper;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * @author Jonathan Doelfs <jd@sodatech.com>
+ */
+class PaginationHelperFactory
```
dunglas: Can you make this class final?

src/Util/PaginationHelper.php
```diff
@@ -0,0 +1,161 @@
...
+    private $maximumItemsPerPage = null;
+
+    private $parameterNamePage = 'page';
+    private $parameterNameEnabled = 'pagination';
+    private $parameterNameItemsPerPage = 'itemsPerPage';
+
+    private $resourceClass;
+    private $operationName;
+
+    public function __construct(RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, string $resourceClass, string $operationName = null, bool $enabled = true, bool $clientEnabled = false, bool $clientItemsPerPage = false, int $itemsPerPage = 30, string $parameterNamePage = 'page', string $parameterNameEnabled = 'pagination', string $parameterNameItemsPerPage = 'itemsPerPage', int $maximumItemsPerPage = null)
```
dunglas: May I suggest an alternative approach?

* Remove the factory
* Register `PaginationHelper` (maybe can we find a more explicit name) as a service
* Pass `enabled`, `clientEnabled` and so on has parameters of methods (instead of the constructor) to make this service stateless

WDYT?

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
jonadoe: ok, will change that

src/Util/Factory/PaginationHelperFactory.php
jonadoe: Sure

src/Util/PaginationHelper.php
jonadoe: The goal of this class was to make it as easy as possible to use it for custom data providers. If the enabled, clientEnabled, ... parameters have to be passed to the different methods you also have to inject them to the constructor of your data provider, which makes it more complicated.
I would go with the factory so you just have to inject the pagination factory into your data provider.

Regarding the name and maybe also the namespace, I think we can find something better

dunglas: Let me some time to think about it and review more deeply but the current proposal hurts me from a design POV (the use is very valid).

dunglas: Why not just introducing a trait?

jonadoe: When using a trait, how should you get all the needed parameters (enabled, ...) without injecting all of them into the constructor?

dunglas: I don't get the problem of injecting them (direct injection is almost always better than indirect injection). If it's for concision, we can provide an `abstract` service definition.

jonadoe: It just bloats up the code of a custom data provider and makes it overly complicated if you have to inject the additional config parameters and also have to pass them to the paging class in the right order. 
Why not abstract this away to make a developers life easier?

dunglas: Because it is a bad design.

soyuka:
Why did you close this @jonadoe ? I still think that this is a good enhancement.

PR #956, opened 2017-02-24
Allow blank nodes in JSON-LD

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | TODO
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | TODO

TODO:

- [ ] Tests
- [ ] Docs

### DIFF
--- a/src/Bridge/Symfony/Routing/IriConverter.php
+++ b/src/Bridge/Symfony/Routing/IriConverter.php
@@ -57,7 +57,7 @@ final class IriConverter implements IriConverterInterface
         $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
 
         if (null === $identifiersExtractor) {
-            @trigger_error('Not injecting ItemIdentifiersExtractor is deprecated since API Platform 2.1 and will not be possible anymore in API Platform 3');
+            @trigger_error('Not injecting IdentifiersExtractorInterface is deprecated since API Platform 2.1 and will not be possible anymore in API Platform 3');
             $this->identifiersExtractor = new IdentifiersExtractor($this->propertyNameCollectionFactory, $this->propertyMetadataFactory, $this->propertyAccessor);
         } else {
             $this->identifiersExtractor = $identifiersExtractor;
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -13,10 +13,12 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\JsonLd\Serializer;
 
+use ApiPlatform\Core\Api\IdentifiersExtractorInterface;
 use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\JsonLd\ContextBuilderInterface;
+use ApiPlatform\Core\JsonLd\Util\BlankNodeIdentifiersGenerator;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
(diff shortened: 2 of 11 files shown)


src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -34,21 +33,18 @@
...
 
-    private $propertyNameCollectionFactory;
-    private $propertyMetadataFactory;
     private $itemDataProvider;
     private $routeNameResolver;
     private $router;
-    private $propertyAccessor;
+    private $itemIdentifiersExtractor;
 
     public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null)
```
soyuka: We should put a reminder somewhere that (when we can break things again) we need to remove those injections and rather directly inject ItemIdentifiersExtractor.

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -64,12 +72,16 @@ public function normalize($object, $format = null, array $context = [])
...
+
         $rawData = parent::normalize($object, $format, $context);
         if (!is_array($rawData)) {
             return $rawData;
         }
 
-        $data['@id'] = $this->iriConverter->getIriFromItem($object);
+        $identifiers = $this->itemIdentifiersExtractor->getIdentifiersFromItem($object);
+
+        $data['@id'] = count(array_filter($identifiers)) > 0 ? $this->iriConverter->getIriFromItem($relatedObject) : $this->blankNodeIdentifiersGenerator->getBlankNodeIdentifier($relatedObject, $context['jsonld_document_root']);
```
soyuka: Shouldn't we add a method that takes identifiers and resolves to one of those to avoid repetition?
Is `$context['jsonld_document_root']` used only to cache the `spl_object_hash`?

dunglas: `count(array_filter($identifiers)) > 0` can be replaced by `array_filter($identifiers)` (faster, easier to read).

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -101,4 +119,33 @@ public function denormalize($data, $class, $format = null, array $context = [])
...
+     */
+    protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relatedObject, string $resourceClass, string $format = null, array $context)
+    {
+        if ($propertyMetadata->isReadableLink()) {
+            return $this->serializer->normalize($relatedObject, $format, $this->createRelationSerializationContext($resourceClass, $context));
+        }
+
+        $identifiers = $this->itemIdentifiersExtractor->getIdentifiersFromItem($relatedObject);
+
+        return count(array_filter($identifiers)) > 0 ? $this->iriConverter->getIriFromItem($relatedObject) : $this->blankNodeIdentifiersGenerator->getBlankNodeIdentifier($relatedObject, $context['jsonld_document_root']);
```
dunglas: Same here `array_filter($identifiers)` is enough.

src/JsonLd/Util/BlankNodeIdentifiersGenerator.php
```diff
@@ -0,0 +1,49 @@
...
+     * Gets a blank node identifier for an object, scoped to a JSON-LD document.
+     *
+     * @param object $object
+     * @param string $documentRootHash
+     *
+     * @return string
+     */
+    public function getBlankNodeIdentifier($object, string $documentRootHash): string
+    {
+        $objectHash = spl_object_hash($object);
```
dunglas: I'm not a fond of exposing such low level data from the system publicly. Is it really safe? Another potential problem is that this ID will be reused later by PHP. WDYT about using a UUID instead?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -415,7 +415,7 @@ protected function getAttributeValue($object, $attribute, $format = null, array
      *
      * @return string|array
      */
-    private function normalizeRelation(PropertyMetadata $propertyMetadata, $relatedObject, string $resourceClass, string $format = null, array $context)
+    protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relatedObject, string $resourceClass, string $format = null, array $context)
```
dunglas: Can you mark this method as `@internal`?

src/JsonLd/Serializer/ItemNormalizer.php
teohhanhui: > Shouldn't we add a method that takes identifiers and resolves to one of those to avoid repetition?

Sure. I'll move it to a private method.

> Is $context['jsonld_document_root'] used only to cache the spl_object_hash?

We need to keep track of which JSON-LD document we're in. But of course there are many ways of doing it. I agree this is not ideal.

src/JsonLd/Util/BlankNodeIdentifiersGenerator.php
teohhanhui: It's not being exposed in any way. The hash is only used to keep track of which objects we've already generated a blank node identifier for, and always return that same identifier for the same object (the scope is the JSON-LD document).

We cannot use UUID for this, as the point is to detect the same object instance.

dunglas: Nevermind got it.

src/JsonLd/Serializer/ItemNormalizer.php
soyuka: Yeah, nothing wrong with how you did it, just wanted to make sure I understood ;)

dunglas:
Can you rebase? It should be ok to be merged.

teohhanhui:
@dunglas There are no tests yet.

soyuka:
You do you own tests :smiling_imp: 

Joking, if you don't have the time lmk

teohhanhui:
@dunglas I need some help rebasing this... There seems to have been a proliferation of `IriConverterInterface::getIriFromItem` to more places. Perhaps we need a different approach?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -86,7 +96,7 @@ public function normalize($object, $format = null, array $context = [])
         $context = $this->initContext($resourceClass, $context);
         $context['api_normalize'] = true;
 
-        if (isset($context['resources'])) {
+        if (isset($context['resources']) && $this->hasIri($object)) {
             $resource = $context['iri'] ?? $this->iriConverter->getIriFromItem($object);
```
soyuka: Is the condition still needed? if it `hasIri` `context['iri']` is defined no?

teohhanhui: No. `$context['iri']` is only set from the JsonLd `ItemNormalizer` (I think). I don't know what's the reasoning behind, but I did not change it...

soyuka:
kinda fixed by #2495

PR #963, opened 2017-02-28
Fix normalizing 3rd level objects

desmax:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

I experience an exception `No resource class found for object of type ...` when trying to normalize some object on 3rd level from initial resource. 
For example:
Object1 Resource has collection of Object2 and Object2 has relation to Object3.
All of them are ApiResources and api-platform fails to normalize Object3

Changes proposed solve the problem and tests still pass.
I did not write new test case though, not sure where to put test for this thing.
Please advice.

### DIFF
--- a/src/Api/ResourceClassResolver.php
+++ b/src/Api/ResourceClassResolver.php
@@ -50,7 +50,7 @@ final class ResourceClassResolver implements ResourceClassResolverInterface
         }
 
         if (($strict && isset($type) && $resourceClass !== $type) || !$this->isResourceClass($typeToFind)) {
-            if (is_subclass_of($type, $resourceClass) && $this->isResourceClass($resourceClass)) {
+            if ($this->isResourceClass($resourceClass)) {
                 return $type;
             }
 
--- a/src/Serializer/ContextTrait.php
+++ b/src/Serializer/ContextTrait.php
@@ -28,10 +28,6 @@ trait ContextTrait
      */
     private function initContext(string $resourceClass, array $context): array
     {
-        if (isset($context['api_sub_level'])) {
-            return $context;
-        }
-
         return array_merge($context, [
             'api_sub_level' => true,
             'resource_class' => $resourceClass,


desmax:
Can somebody confirm issue at least?

soyuka:
I'd need to test this, can you reproduce this in a test case at least?

I think that we're testing three level relationships already in the behat tests.

src/Api/ResourceClassResolver.php
```diff
@@ -50,7 +50,7 @@ public function getResourceClass($value, string $resourceClass = null, bool $str
         }
 
         if (($strict && isset($type) && $resourceClass !== $type) || !$this->isResourceClass($typeToFind)) {
-            if (is_subclass_of($type, $resourceClass) && $this->isResourceClass($resourceClass)) {
+            if ($this->isResourceClass($resourceClass)) {
```
dunglas: It looks not OK to me. We lose information, we don't want only a resource here, but a resource of the given type.

src/Serializer/ContextTrait.php
```diff
@@ -28,10 +28,6 @@
      */
     private function initContext(string $resourceClass, array $context): array
     {
-        if (isset($context['api_sub_level'])) {
```
dunglas: Same here, is this check was here there is a reason for.

dunglas:
Yes, I've looked recently and this is tested. Can you explain the exact problem you encounter or provide a failing test please?

dunglas:
Closing for now because we don't get any feedback.

desmax:
My problem seems to be gone when I use master branch. Just tested it.

dunglas:
Great, thanks for the feedback.

desmax:
Found exactly what was wrong: I have doctrine single table inheritance and only base class was marked as `@ApiResource`

PR #964, opened 2017-03-02
Boolean filter integration tests improvements

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

I've been reading the [boolean filter documentation](https://api-platform.com/docs/core/filters#boolean-filter) and found that using `on|off` as documented was not working. So I've been searching in the code and saw that only `true|false|1|0` was implemented and tested and also figured out that most of integration tests were not working properly.

It seems that it was working at some point and has been removed [here](https://github.com/api-platform/core/commit/e4eca423d32720bc0cba3fd9a0e47675c1fa21c1#diff-31a7356474220b422a684ad542968342) on purpose ?

Should I open a PR to the doc to remove it or should I re-add this possibility into the core ?

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -235,6 +235,14 @@ class FeatureContext implements Context, SnippetAcceptingContext
      */
     public function thereIsDummyObjectsWithDummyBoolean($nb, $bool)
     {
+        if (in_array($bool, ['true', '1', 1], true)) {
+            $bool = true;
+        } elseif (in_array($bool, ['false', '0', 0], true)) {
+            $bool = false;
+        } else {
+            $expected = ['true', 'false', '1', '0'];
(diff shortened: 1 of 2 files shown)


antograssiot:
I've also found a bunch of similar cases where the `pattern` property in the schema can be anything and should be fixed and I can provide another PR to fixes them but I wanted your opinion on this beforehand.

teohhanhui:
Yes, the docs should be updated. We do not support on|off anymore.

antograssiot:
rebased onto 2.0 and PR target updated

dunglas:
Can you rebase? Tests should be green now.

antograssiot:
@dunglas rebase done, all green now 😉

dunglas:
Thanks @antograssiot!

PR #906, opened 2017-01-12
Add ExistsFilter

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | TODO
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | TODO

IS NULL / IS NOT NULL (field / single-valued association)
or
IS EMPTY / IS NOT EMPTY (collection-valued association)

PR #967, opened 2017-03-05
hydra:view schema validation 

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I found that the hydra:view schema validation pattern were not correctly checked in the testsuite. Any value would let the tests passing as the schema is malformed.

I've also updated the Behatch namespace as they changed it in 2.6 (call it deprecated) without providing a decent BC version.

PR #970, opened 2017-03-08
hydra:view schema validation  (#967)

soyuka:
* Fix Behatch namespace 
* Fix all hydra:view pattern check in json schema validation
* Bump Behatch requirement

ping @teohhanhui just doing this on github so that we all know I did it and to see if tests are okay \o/.

This is a prerequisite to fix #906 so that we finally can merge it!

### DIFF
--- a/behat.yml
+++ b/behat.yml
@@ -6,8 +6,8 @@ default:
         - 'HydraContext'
         - 'SwaggerContext'
         - 'Behat\MinkExtension\Context\MinkContext'
-        - 'Sanpi\Behatch\Context\RestContext'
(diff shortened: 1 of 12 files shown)


soyuka:
Wow I think I can't do that this way because it's out-of-date with the base branch, any clue @teohhanhui ? I don't want to break everything :p but I think that I should do :

```
git checkout master
git merge 2.0
```

dunglas:
Sorry I was not available but it is exactly what you need to do :)

soyuka:
:D done and I updated the existsFilter along the way.

Issue #973, opened 2017-03-08
Customize Doctrine Exception Message


jordscream:
Hello,
I was working on configuration of status code for exception.

For the moment, I define this in config yml file:
    exception_to_status:
        ApiPlatform\Core\Exception\InvalidArgumentException: 400
        Doctrine\DBAL\Exception\UniqueConstraintViolationException: 409

For the first exception, the message return is correct and I do not need to modify it
(Ex: The type of the "my_field" attribute must be \"int\", \"double\" given.)

For the second Exception (Doctrine UniqueConstraintViolationException), I have a nice SQL request returned (which is not good for security and in general in production server)

Have you something to catch it or modify the Doctrine message ? 
Or is it something it can be developped in API platform ?
At the moment, any Doctrine Exception will return a SQL message:
- NotNullConstraintViolationException
- ForeignKeyConstraintViolationException
- UniqueConstraintViolationException
- SyntaxErrorException
- NonUniqueFieldNameException
- InvalidFieldNameException
- TableNotFoundException
- TableExistsException

Proposed Solution:

In API Platform, the event class responsible to send the data to doctrine (ApiPlatform\Core\Bridge\Doctrine\EventListener\WriteListener), flush the data without catching the Doctrine Exception (line 68)

We need to develop the exception classes above in API Platform and catch the flush operation with theses classes and a better message, am I right ?

Thanks,
Jordan

dunglas:
You're right. But those exceptions should always be catched by your code and never reach the end user level. Why do you have such exceptions?

You're right about the default message. We should probably do something to not display it in production by default and use generic messages like "Internal Server Error" to not expose sensitive information.

jordscream:
@dunglas Thanks for your response.

I have this exception, for example, when I post 2 times an Object with a same UUID (We use UUID field for unique identifier and we put it in the POST method)

Entity 
    /**
     * @var guid The id of this object
     * @ORM\Id
     * @ORM\Column(type="guid")
     * @Assert\Uuid
     */
     private $id;

If I do this, I have a Request SQL displayed (constraint integrity by doctrine).

soyuka:
#929

jordscream:
@dunglas @soyuka @meyerbaptiste @teohhanhui 

In my Team, for an internal reason, we call API POST with a predefined ID.
{
 id: 'ea62c86c-d20f-4ce9-a502-441ca96b9434',
 name: 'test',
}
[...]

Issue #929, opened 2017-02-02
[Security, RFC] Streamline exception handling, prevent disclosure


dkarlovi:
I'm testing error conditions and default exception handling for my app. Note, all of this is in **production** environment. This is primarily an RFC so feel free to comment on the behaviour before any action is required.

Things to note:
- Hydra error only for SQL exception, not for DB connection issue
- Exception messages include complete SQL or networking errors

Triggering an SQL exception + `Accept: application/ld+json`:
- Status 500
- `Content-Type: application/ld+json`
- `@type: error`
- `hydra:description: <Complete SQL Exception message>` (way too verbose)

Triggering an SQL exception + `Accept: application/json`:
- Status 500
- `Content-Type: application/problem+json`
- `detail: <Complete SQL Exception message>` (way too verbose)

DB host down + `Accept: application/ld+json`:
- Status 500
- `Content-Type: application/problem+json`
- `detail: <Complete SQL Exception message>` (way too verbose)

DB host down + `Accept: application/json`:
- Status 500
- `Content-Type: application/problem+json`
- `detail: <Complete SQL Exception message>` (way too verbose)

dunglas:
Can you review @dkarlovi #978?
Should we mark this issue as a security vulnerability @api-platform/core-team?

dkarlovi:
@dunglas will try his branch and comment on #978.

IMO you should mark this as a security issue, it's a typical disclosure issue, giving the attacker a lot to work with (ie, identifying the database system and maybe even version, Doctrine, etc), it's a large attack vector.

Simperfit:
@dkarlovi This has been merged, is it ok fo you ?

dkarlovi:
@Simperfit I believe this was completed in #1024, yes. Haven't tested it since.

PR #797, opened 2016-10-11
Add the ability to configure and match exceptions with an HTTP status code

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #405 
| License       | MIT
| Doc PR        | N/A

This PR adds the ability to configure and match exceptions with an HTTP status code and to override default exceptions (i.e., `Symfony\Component\Serializer\Exception\UnexpectedValueException` and `Symfony\Component\Serializer\Exception\ExceptionInterface` to 400 HTTP error).

Example of configuration:
```yaml
api_platform:
  exception_to_status:
    Foo\Bar\BazException: 400
    Baz\Bar\FooException: "HTTP_BAD_REQUEST" # or use a constant defined in the 'Symfony\Component\HttpFoundation\Response' class
```

PR #978, opened 2017-03-10
Do not expose sensitive information

jordscream:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #973 , #929 
| License       | MIT
| Doc PR        | N/A

This PR fixes the display of Sensitive information (SQL request displayed, Database information) when the project is in production.
No BC according to me

### DIFF
--- a/src/Action/ExceptionAction.php
+++ b/src/Action/ExceptionAction.php
@@ -66,9 +66,11 @@ final class ExceptionAction
     public function __invoke(FlattenException $exception, Request $request): Response
     {
         $exceptionClass = $exception->getClass();
+        $statusCode = $exception->getStatusCode();
+
         foreach ($this->exceptionToStatus as $class => $status) {
             if (is_a($exceptionClass, $class, true)) {
-                $exception->setStatusCode($status);
+                $statusCode = $status;
 
                 break;
             }
@@ -80,6 +82,6 @@ final class ExceptionAction
         $headers['X-Content-Type-Options'] = 'nosniff';
         $headers['X-Frame-Options'] = 'deny';
 
-        return new Response($this->serializer->serialize($exception, $format['key']), $exception->getStatusCode(), $headers);
+        return new Response($this->serializer->serialize($exception, $format['key'], ['statusCode' => $statusCode]), $statusCode, $headers);
     }
 }
--- a/src/Hydra/Serializer/ErrorNormalizer.php
+++ b/src/Hydra/Serializer/ErrorNormalizer.php
@@ -13,6 +13,7 @@ namespace ApiPlatform\Core\Hydra\Serializer;
 
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
 use Symfony\Component\Debug\Exception\FlattenException;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
@@ -42,6 +43,11 @@ final class ErrorNormalizer implements NormalizerInterface
         $message = $object->getMessage();
         if ($this->debug) {
             $trace = $object->getTrace();
+        } elseif ($object instanceof FlattenException) {
+            $statusCode = $context['statusCode'] ?? $object->getStatusCode();
+            if ($statusCode >= 500 && $statusCode < 600) {
+                $message = Response::$statusTexts[$statusCode];
+            }
         }
 
         $data = [
--- a/tests/Action/ExceptionActionTest.php
+++ b/tests/Action/ExceptionActionTest.php
@@ -33,7 +33,7 @@ class ExceptionActionTest extends \PHPUnit_Framework_TestCase
         $flattenException = FlattenException::create($serializerException->reveal());
 
         $serializer = $this->prophesize(SerializerInterface::class);
-        $serializer->serialize($flattenException, 'jsonproblem')->willReturn();
+        $serializer->serialize($flattenException, 'jsonproblem', ['statusCode' => Response::HTTP_BAD_REQUEST])->willReturn();
 
         $exceptionAction = new ExceptionAction($serializer->reveal(), ['jsonproblem' => ['application/problem+json'], 'jsonld' => ['application/ld+json']], [ExceptionInterface::class => Response::HTTP_BAD_REQUEST, InvalidArgumentException::class => Response::HTTP_BAD_REQUEST]);
 
@@ -57,7 +57,7 @@ class ExceptionActionTest extends \PHPUnit_Framework_TestCase
         $flattenException = FlattenException::create($serializerException->reveal());
 
         $serializer = $this->prophesize(SerializerInterface::class);
-        $serializer->serialize($flattenException, 'jsonproblem')->willReturn();
+        $serializer->serialize($flattenException, 'jsonproblem', ['statusCode' => $flattenException->getStatusCode()])->willReturn();
 
         $exceptionAction = new ExceptionAction($serializer->reveal(), ['jsonproblem' => ['application/problem+json'], 'jsonld' => ['application/ld+json']]);
 
--- a/tests/Hydra/Serializer/ErrorNormalizerTest.php
+++ b/tests/Hydra/Serializer/ErrorNormalizerTest.php
@@ -14,6 +14,7 @@ namespace ApiPlatform\Core\Tests\Hydra\Serializer;
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
 use ApiPlatform\Core\Hydra\Serializer\ErrorNormalizer;
 use Symfony\Component\Debug\Exception\FlattenException;
+use Symfony\Component\HttpFoundation\Response;
 
 /**
  * @author Kévin Dunglas <dunglas@gmail.com>
@@ -33,6 +34,47 @@ class ErrorNormalizerTest extends \PHPUnit_Framework_TestCase
         $this->assertTrue($normalizer->supportsNormalization(new FlattenException(), ErrorNormalizer::FORMAT));
         $this->assertFalse($normalizer->supportsNormalization(new FlattenException(), 'xml'));
         $this->assertFalse($normalizer->supportsNormalization(new \stdClass(), ErrorNormalizer::FORMAT));
+    }
+
+    /**
+     * @dataProvider providerStatusCode
+     *
+     * @param $status            http status code of the Exception
+     * @param $originalMessage   original message of the Exception
+     * @param $debug             simulates kernel debug variable
+     */
+    public function testErrorServerNormalize($status, $originalMessage, $debug)
+    {
+        $urlGeneratorProphecy = $this->prophesize(UrlGeneratorInterface::class);
+        $urlGeneratorProphecy->generate('api_jsonld_context', ['shortName' => 'Error'])->willReturn('/context/foo')->shouldBeCalled();
+
+        $normalizer = new ErrorNormalizer($urlGeneratorProphecy->reveal(), $debug);
+        $exception = FlattenException::create(new \Exception($originalMessage), $status);
+
+        $expected = [
+            '@context' => '/context/foo',
+            '@type' => 'hydra:Error',
+            'hydra:title' => 'An error occurred',
+            'hydra:description' => ($debug || $status < 500) ? $originalMessage : Response::$statusTexts[$status],
+        ];
+
+        if ($debug) {
+            $expected['trace'] = $exception->getTrace();
+        }
+
+        $this->assertEquals($expected, $normalizer->normalize($exception, null, ['statusCode' => $status]));
+    }
+
+    public function providerStatusCode()
+    {
+        return [
+            [Response::HTTP_INTERNAL_SERVER_ERROR, 'Sensitive SQL error displayed', false],
+            [Response::HTTP_GATEWAY_TIMEOUT, 'Sensitive server error displayed', false],
+            [Response::HTTP_BAD_REQUEST, 'Bad Request Message', false],
+            [Response::HTTP_INTERNAL_SERVER_ERROR, 'Sensitive SQL error displayed', true],
+            [Response::HTTP_GATEWAY_TIMEOUT, 'Sensitive server error displayed', true],
+            [Response::HTTP_BAD_REQUEST, 'Bad Request Message', true],
+        ];
     }
 
     public function testNormalize()


src/Action/ExceptionAction.php
```diff
@@ -42,17 +42,20 @@
...
+    private $debug;
 
     /**
      * @param SerializerInterface $serializer
      * @param array               $errorFormats      A list of enabled formats, the first one will be the default
      * @param array               $exceptionToStatus A list of exceptions mapped to their HTTP status code
+     * @param bool                $debug             The kernel debug parameter
      */
-    public function __construct(SerializerInterface $serializer, array $errorFormats, array $exceptionToStatus = [])
+    public function __construct(SerializerInterface $serializer, array $errorFormats, array $exceptionToStatus, bool $debug)
```
dunglas: The `$exceptionToStatus` default value should be kept and `debug` default to false to preserve backward compatibility.

src/Action/ExceptionAction.php
```diff
@@ -80,6 +83,10 @@ public function __invoke(FlattenException $exception, Request $request): Respons
         $headers['X-Content-Type-Options'] = 'nosniff';
         $headers['X-Frame-Options'] = 'deny';
 
+        if ($exception->getStatusCode() == Response::HTTP_INTERNAL_SERVER_ERROR && !$this->debug) {
```
dunglas: Please use `===` and Yoda style. I would also invert the order because `!$this->debug` is cheaper.

`if (!$this->debug && Response::HTTP_INTERNAL_SERVER_ERROR === $exception->getStatusCode() ) {`

tests/Action/ExceptionActionTest.php
```diff
@@ -59,7 +59,7 @@ public function testActionWithUncatchableException()
         $serializer = $this->prophesize(SerializerInterface::class);
         $serializer->serialize($flattenException, 'jsonproblem')->willReturn();
 
-        $exceptionAction = new ExceptionAction($serializer->reveal(), ['jsonproblem' => ['application/problem+json'], 'jsonld' => ['application/ld+json']]);
+        $exceptionAction = new ExceptionAction($serializer->reveal(), ['jsonproblem' => ['application/problem+json'], 'jsonld' => ['application/ld+json']], [], true);
```
dunglas: I would let it is it was previously (to test that there is no BC breaks and that the previous signature is still accepted).

dunglas:
Thanks for this PR, it's greatly appreciated. I left some minor comment but :+1: to change this behavior when comments will be fixed.

src/Action/ExceptionAction.php
dkarlovi: Would it make sense to catch all 5xx here?

dunglas: Good catch!

jordscream:
@dunglas : reviews applied 👍 . You can check

src/Action/ExceptionAction.php
dkarlovi: @jordscream what do you think?

teohhanhui:
Is this the correct way to fix it? I hope someone who's familiar with Symfony exception handling can comment...

jordscream:
@teohhanhui , check Symfony\Bundle\TwigBundle\Controller\ExceptionController.php
I tried to do the same approach... I guess :-)

teohhanhui:
Ok, sounds good then.

src/Action/ExceptionAction.php
jordscream: @dunglas @dkarlovi , good point. I made a new commit to catch all server errors according the method isServerError of the Response Object.

jordscream:
@teohhanhui , as suggest @dunglas @dkarlovi , I modify the code to catch all 5XX exceptions.

jordscream:
@teohhanhui @soyuka , any news ?

soyuka:
@jordscream looks fine to me, just waiting for more approval (ping @teohhanhui @meyerbaptiste ?). Anyway I'd be able to merge this but I won't publish a new release without @dunglas. Are you in a hurry to get this patch?

jordscream:
@soyuka we need it to push in production. We have a temporary solution without that. ;-)
Let me know when you want ^^

jordscream:
@soyuka , I did not know about min-stability ! Let's do that
Do you want a target rebase on master ?

soyuka:
No, I meant 2.0 same thing. Leave this one like this it's fine! Still, if I merge it won't be released (new tag) and so `composer update` won't have the patch.

src/Action/ExceptionAction.php
```diff
@@ -80,6 +83,14 @@ public function __invoke(FlattenException $exception, Request $request): Respons
...
-        return new Response($this->serializer->serialize($exception, $format['key']), $exception->getStatusCode(), $headers);
+        $response = new Response(null, $exception->getStatusCode(), $headers);
+
+        if (!$this->debug && $response->isServerError()) {
+            $exception->setMessage(Response::$statusTexts[$response->getStatusCode()]);
+        }
+
+        $response->setContent($this->serializer->serialize($exception, $format['key']));
+
+        return $response;
```
meyerbaptiste: `return $response->setContent(...)`?

jordscream: @soyuka , Yes in fact, I need the Response object to use the method isServerError() in the Error condition just above.
It was a point I have to discuss with you. No other way to deal it

soyuka: Yes but still you can return `$response->setContent()` (returns Response)

jordscream: @meyerbaptiste good catch, I just pushed the modification

soyuka:
Nice one! Merging this tomorrow if noone has comments until then :).

meyerbaptiste:
Before merging this, what about https://github.com/api-platform/core/issues/929#issuecomment-285852049?

cc @dkarlovi

jordscream:
@meyerbaptiste it's effectively a security issue. Somethong to do in particular?

src/Action/ExceptionAction.php
```diff
@@ -80,6 +83,12 @@ public function __invoke(FlattenException $exception, Request $request): Respons
         $headers['X-Content-Type-Options'] = 'nosniff';
         $headers['X-Frame-Options'] = 'deny';
 
-        return new Response($this->serializer->serialize($exception, $format['key']), $exception->getStatusCode(), $headers);
+        $response = new Response(null, $exception->getStatusCode(), $headers);
+
+        if (!$this->debug && $response->isServerError()) {
+            $exception->setMessage(Response::$statusTexts[$response->getStatusCode()]);
```
teohhanhui: We should not modify the exception.

teohhanhui: Which means we need to look for a different way of doing this.

src/Action/ExceptionAction.php
```diff
@@ -42,17 +42,20 @@
...
+    private $debug;
 
     /**
      * @param SerializerInterface $serializer
      * @param array               $errorFormats      A list of enabled formats, the first one will be the default
      * @param array               $exceptionToStatus A list of exceptions mapped to their HTTP status code
+     * @param bool                $debug             The kernel debug parameter
      */
-    public function __construct(SerializerInterface $serializer, array $errorFormats, array $exceptionToStatus = [])
+    public function __construct(SerializerInterface $serializer, array $errorFormats, array $exceptionToStatus = [], bool $debug = false)
```
teohhanhui: We should move `$debug` before `$exceptionToStatus` and remove its default value.

src/Action/ExceptionAction.php
jordscream: @teohhanhui , you mean create a new Exception object ?

src/Action/ExceptionAction.php
jordscream: @teohhanhui , what about BC breaks if we do this ? I do this to prevent the bc break
thanks for your feeback

src/Action/ExceptionAction.php
jordscream: @teohhanhui, or à simply response with message and 500 status code?

soyuka: I'd say that we could create a new `HTTPException` if we shouldn't change the exception message. I didn't though about this but indeed it'd be better if (in production) you'd like to log the origin exception.

src/Action/ExceptionAction.php
soyuka: Because this is a security issue and needs to be fixed in 2.0, I think we should maintain the debug flag here.

src/Action/ExceptionAction.php
meyerbaptiste: I thought the same thing but currently the `FlattenException` is already altered ([ExceptionAction.php#L71](https://github.com/api-platform/core/blob/master/src/Action/ExceptionAction.php#L71)). But in the same time, the original exception is not altered because `FlattenException` is just _a wrap to be able to serialize it_. The original one is logged and collected by the profiler before the request is duplicated and the `FlattenException` is created.

soyuka: What would be your suggestion then @meyerbaptiste?

meyerbaptiste: So my suggestion is... why not to deal with the message in the [`ErrorNormalizer` (L40)](https://github.com/api-platform/core/blob/master/src/Problem/Serializer/ErrorNormalizer.php#L40)? And avoid the [`FlattenException::setStatusCode()`](https://github.com/api-platform/core/blob/master/src/Action/ExceptionAction.php#L71) in the same time. WDYT?

soyuka: Indeed would be much cleaner. WDYT @teohhanhui?

jordscream: @meyerbaptiste @soyuka , Woooh, it seems better ^^, waiting for @teohhanhui approval before modification if needed

src/Action/ExceptionAction.php
teohhanhui: There is no BC break because this class is `final`. :smile:

src/Action/ExceptionAction.php
teohhanhui: Yes, let's avoid modifying the `FlattenException`. We should treat it as immutable.

src/Action/ExceptionAction.php
jordscream: @teohhanhui oHo ^^

src/Action/ExceptionAction.php
jordscream: @meyerbaptiste @teohhanhui @soyuka , what about the comment of @meyerbaptiste "And avoid the FlattenException::setStatusCode() in the same time" (means move it in ErrorNormalizer) ? 
The Response with status code is dealed by the ExceptionAction, not by ErrorNormalizer.... 

I feel confused, cause in the same time, We alter the exception status code in ExceptionAction and the exception message in ErrorNormalizer...

src/Action/ExceptionAction.php
dunglas: @teohhanhui there is. People can (and do) instantiate this class directly.

src/Action/ExceptionAction.php
teohhanhui: @jordscream Have a look at https://github.com/symfony/symfony/blob/4d48b58d197cfa76046bb23ce9c2fe7bec8fbbb2/src/Symfony/Bundle/TwigBundle/Controller/ExceptionController.php#L41-L73

I see no compelling reason to modify the exception in any way whatsoever.

For example, https://github.com/api-platform/core/blob/8053c55311b713153179f412ea294362a9997bea/src/Action/ExceptionAction.php#L58-L84 can be rewritten as:

```php
/**
 * Converts a an exception to a JSON response.
 *
 * @param FlattenException $exception
 * @param Request          $request
 *
 * @return Response
 */
public function __invoke(FlattenException $exception, Request $request): Response
{
    $exceptionClass = $exception->getClass();
    $statusCode = $exception->getStatusCode();

    foreach ($this->exceptionToStatus as $class => $status) {
        if (is_a($exceptionClass, $class, true)) {
            $statusCode = $status;

            break;
        }
    }

    $headers = $exception->getHeaders();
    $format = ErrorFormatGuesser::guessErrorFormat($request, $this->errorFormats);
    $headers['Content-Type'] = sprintf('%s; charset=utf-8', $format['value'][0]);
    $headers['X-Content-Type-Options'] = 'nosniff';
    $headers['X-Frame-Options'] = 'deny';

    return new Response($this->serializer->serialize($exception, $format['key']), $statusCode, $headers);
}
```
with no change in behaviour and no BC.

> The Response with status code is dealed by the ExceptionAction, not by ErrorNormalizer....

The `ErrorNormalizer` is responsible for the body / content of the `Response`. (It basically replaces the content set when creating the `Response`.) How is it called? https://github.com/api-platform/core/blob/8053c55311b713153179f412ea294362a9997bea/src/EventListener/SerializeListener.php

So in our case, if we want to change the message, we should change in in the respective `ErrorNormalizer`s.

src/Action/ExceptionAction.php
teohhanhui: Uhh, yes. :disappointed:

src/Action/ExceptionAction.php
dunglas: 👍 to do it directly in the normalizer. It's cleaner.

jordscream:
@dunglas @teohhanhui , I made changes, waiting for your reviews
Also, the behat functionnal test fails on 2.0 branch... no relation with my code

teohhanhui:
> Also, the behat functionnal test fails on 2.0 branch... no relation with my code
failed causing by #950

That's not true. Please check Travis. There is 1 error in PHPUnit test. All Behat tests passed.

teohhanhui:
My previous suggestion doesn't work, at least with our current way of handling exceptions. The big question is, is what we're doing correct? Can we do it better?

src/Hydra/Serializer/ErrorNormalizer.php
```diff
@@ -42,6 +43,10 @@ public function normalize($object, $format = null, array $context = [])
         $message = $object->getMessage();
         if ($this->debug) {
             $trace = $object->getTrace();
+        } else {
```
dunglas: I suggest:

```php
if ($this->debug) {
    $trace = $object->getTrace();
} elseif (Response::HTTP_INTERNAL_SERVER_ERROR <= ($code = $object->getStatusCode())) {
    $message = Response::$statusTexts[$code];
}
```

teohhanhui: We need to check `$object instanceof FlattenException` first. And I think it's a bad idea to use `<=` for this.

meyerbaptiste: We lose the mapping between exceptions and their HTTP status code (#797). What about passing [the correct HTTP status code](https://github.com/api-platform/core/blob/master/src/Action/ExceptionAction.php#L71) inside the serializer context?

jordscream:
@teohhanhui , oups, done....

dunglas:
@meyerbaptiste is right, we must pass the status code in the context to be reliable.

src/Hydra/Serializer/ErrorNormalizer.php
teohhanhui: If we're okay with modifying the status code in the `FlattenException`, like we currently do, then this works just fine.

teohhanhui: But perhaps we should determine the status code here. That may be more proper.

jordscream:
@dunglas , just to be sure, I just need to pass the **originalStatusCode** in the Serializer Context. And use it to compare in ErrorNormalizer ? If a User define in ExceptionToStatus configuration a 500 ErrorException mapped to status code 400 (for example), in production, the project will return 400 statusCode, with a message "Internal Server Error" ? Does it make sense ?

jordscream:
@dunglas @meyerbaptiste , I did the modifications, let me know if it is okay for you

src/Action/ExceptionAction.php
```diff
@@ -66,6 +66,8 @@ public function __construct(SerializerInterface $serializer, array $errorFormats
     public function __invoke(FlattenException $exception, Request $request): Response
     {
         $exceptionClass = $exception->getClass();
+        $statusCode = $exception->getStatusCode();
+
         foreach ($this->exceptionToStatus as $class => $status) {
             if (is_a($exceptionClass, $class, true)) {
                 $exception->setStatusCode($status);
```
meyerbaptiste: You have to change the value of your `$statusCode` var here to treat the `FlattenException` as immutable.

```diff
- $exception->setStatusCode($status);
+ $statusCode = $status;
```

src/Hydra/Serializer/ErrorNormalizer.php
```diff
@@ -42,6 +42,8 @@ public function normalize($object, $format = null, array $context = [])
         $message = $object->getMessage();
         if ($this->debug) {
             $trace = $object->getTrace();
+        } elseif ($object instanceof FlattenException && isset($context['statusCode']) && $context['statusCode'] >= Response::HTTP_INTERNAL_SERVER_ERROR) {
```
meyerbaptiste: According to `Response::isServerError()`:
```php
$context['statusCode'] >= 500 && $context['statusCode'] < 600
```
And IMO if `$context['statusCode']` isn't set, you should use `$exception->getStatusCode()`.

src/Action/ExceptionAction.php
```diff
@@ -80,6 +82,6 @@ public function __invoke(FlattenException $exception, Request $request): Respons
         $headers['X-Content-Type-Options'] = 'nosniff';
         $headers['X-Frame-Options'] = 'deny';
 
-        return new Response($this->serializer->serialize($exception, $format['key']), $exception->getStatusCode(), $headers);
+        return new Response($this->serializer->serialize($exception, $format['key'], ['statusCode' => $statusCode]), $exception->getStatusCode(), $headers);
```
meyerbaptiste: Same here:
```diff
- return new Response($this->serializer->serialize($exception, $format['key'], ['statusCode' => $statusCode]), $exception->getStatusCode(), $headers);
+ return new Response($this->serializer->serialize($exception, $format['key'], ['statusCode' => $statusCode]), $statusCode, $headers);
```

jordscream:
@meyerbaptiste @dunglas done, let me know ^^. (I learn a lot of things with us, guys, do not hesitate to feeback me ^^)

src/Hydra/Serializer/ErrorNormalizer.php
```diff
@@ -42,6 +43,11 @@ public function normalize($object, $format = null, array $context = [])
         $message = $object->getMessage();
         if ($this->debug) {
             $trace = $object->getTrace();
+        } elseif ($object instanceof FlattenException) {
+            $statusCode = (isset($context['statusCode'])) ? $context['statusCode'] : $object->getStatusCode();
```
meyerbaptiste:
```php
$statusCode = $context['statusCode'] ?? $object->getStatusCode();
```

src/Hydra/Serializer/ErrorNormalizer.php
```diff
@@ -42,6 +43,11 @@ public function normalize($object, $format = null, array $context = [])
...
+            if ($statusCode >= 500 && $statusCode < 600) {
+                $message = Response::$statusTexts[$context['statusCode']];
```
meyerbaptiste: It should be `$statusCode` instead of `$context['statusCode']`.

jordscream: @meyerbaptiste , I push too fast, it's fixed

src/Hydra/Serializer/ErrorNormalizer.php
jordscream: @meyerbaptiste  PHP 7 , not yet the reflex, dixed

teohhanhui:
I believe we do not have to pass it in the serialization context, but instead moving the logic into the `ErrorNormalizer` itself. It'll be cleaner.

jordscream:
rebase branch with 2.0

soyuka:
May you also squash the commits? Use `git rebase -i HEAD~5` for interactive and change `pick` by `fixup` for every commit that's not the first one. Thanks!

jordscream:
@teohhanhui , for me is totally the opposite.

ErrorNormalizer->normalize is only here to normalize, not to modify... and only responsible to send a serialize data. (IMO)
ExceptionAction is here to catch and to call the normalize with the right data and headers (IMO)

So always IMO, I would let the satusCode and message treatment in ExceptionAction without alter FlattenException

Again, You have more experience. Tell me the right way to take. :-)

jordscream:
@soyuka done ^^

dunglas:
I agree that the normalizer should not change the state of the underlying exception.

teohhanhui:
An exception being *normalized* into a different error message is not modifying anything in any way. :smile:

jordscream:
@dunglas @teohhanhui @meyerbaptiste @soyuka 

Need to continue this PR with your advice:

1/ Let the code like this (statusCode in Action and Message in ErrorNormalizer) => implies that anybody who uses his own normalizer or a Symfony normalizer cannot benefit of the error hidding in production.

2/ Move all the logic in Action. (statusCode + Message)

3/ Move all the logic in ErrorNormalizer (satusCode + Message): same remark as 1/ 

4/ Other choices ?

Your choice ? :-)

dunglas:
I'm for merging the code as is. Can you also add a unit test checking that the message is hidden when the status code is 500 (and not hidden when it's another status)?

jordscream:
@dunglas , done tell me if you want more.
I used dataProvider and test differents status code associated to a fake original message. 
If debug is true or status < 500, I should have the original message in any case, otherwise, I should have a compliant error message from the Response Foundation Class

dunglas:
Thank you very much for fixing this @jordscream!

jordscream:
@dunglas , my pleasure ^^

dkarlovi:
@jordscream great work following this through, thank you!

dkarlovi:
Houston, we have a problem. :) /cc @dunglas @jordscream 

This works as expected with `Accept: application/ld+json`, but not with `Accept: application/json`.

    Accept: application/ld+json:
    {
        "@context": "/api/contexts/Error",
        "@type": "hydra:Error",
        "hydra:description": "Internal Server Error"
    }

    Accept: application/json:
    {
        "type": "https://tools.ietf.org/html/rfc2616#section-10",
        "title": "An error occurred",
        "detail": "An exception occurred while executing 'INSERT INTO images <TLDR> SQLSTATE[42S02]: Base table or view not found: 1146 Table 'images' doesn't exist"
    }

dkarlovi:
Also, if I trigger a connection-type error (I turn off the DB container), the same full error is thrown with either content type:

    {
        "type": "https://tools.ietf.org/html/rfc2616#section-10",
        "title": "An error occurred",
        "detail": "An exception occured in driver: SQLSTATE[HY000] [2002] php_network_getaddresses: getaddrinfo failed: Try again"
    }

meyerbaptiste:
Yes, we forget the `ApiPlatform\Core\Problem\Serializer\ErrorNormalizer` in this PR...

jordscream:
@dunglas @meyerbaptiste @dkarlovi oups!

PR #979, opened 2017-03-11
[WIP] Fix Travis setup

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

PR #980, opened 2017-03-11
Remove useless count in ValidatorListener

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
+++ b/src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
@@ -74,7 +74,7 @@ final class ValidateListener
         }
 
         $violations = $this->validator->validate($data, null, $validationGroups);
-        if (0 !== count($violations)) {
+        if ($violations) {
             throw new ValidationException($violations);
         }
     }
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
```diff
@@ -74,7 +74,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
         }
 
         $violations = $this->validator->validate($data, null, $validationGroups);
-        if (0 !== count($violations)) {
+        if ($violations) {
```
soyuka: This evaluates to `true` even if there are no violations. `count` isn't that useless here.

teohhanhui:
I really think we should prefer to be explicit, instead of trying to micro-optimize (we should do that where it matters).

soyuka:
I think that @dunglas just saw that while trying to fix the segfault and therefore changed it (in the train with 3G). No big deal, closing this one.

teohhanhui:
Someone needs to send a debug core dump to the PHP team. I'm running into segfault too with phpdbg + PHPUnit in our project. :cry:

soyuka:
https://github.com/api-platform/core/pull/979 ^^

I'm having segfaults because of either the ulimit either the memory limit when covering the Configuration test case.

Issue #969, opened 2017-03-08
Filters not working with @ORM\Embedded 


elazer:
My entity has a "partySize", an embedded class with "value", "minValue", "maxValue".

But trying to filter with partySize.value or any of the other 2 properties would have no effects.

dunglas:
It's indeed not currently supported. I don't know if it can be easily be added but PR about this are very welcome!

antograssiot:
I have a version for the boolean filter that supports embeddable.  I can try to find some time to open a PR but I guess that you're expecting the same level of support on every filters whigh will take much more time.
I know my implementation is far away from ideal but it could beasily a starting point for discussion. WDYT ?

soyuka:
@antograssiot +1 about the starting point!

jordscream:
Hi @antograssiot, @soyuka : the same for me ^^. I have a branch I develop last week with an implementation of Embed Filter for Boolean. Functionnal and  ready for a PR. (except I did not write unit test and need discussions about my implementation). @antograssiot , Do you want I push for a PR if you have not time ?

antograssiot:
Sure go for it.

antograssiot:
So my naive implementation looks like that

https://gist.github.com/antograssiot/e971410aa740ac18c54ba999cc513143

Still needs refractoring, I didn't had a chance to spend more time on it

jordscream:
@antograssiot , thanks. I did a PR. We had the same logic to deal that. Do not hesitate to comment What I did in the PR

PR #984, opened 2017-03-13
Manage embedded fields for API Platform Filter

jordscream:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #969 
| License       | MIT
| Doc PR        | N/A

This PR is to aliment a discussion about how to manage embedded fields in API Platform Filter
This PR works form Boolean Filter Only. It works recursively Entity->Embed Fields -> Embed Fields..
No Tests is developped for the moment

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -25,6 +25,8 @@ use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyGroup;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyOffer;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyProduct;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyProperty;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\EmbeddableDummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\EmbeddedDummy;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Node;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Question;
@@ -171,6 +173,25 @@ class FeatureContext implements Context, SnippetAcceptingContext
     }
 
     /**
+     * @Given there are :nb embedded dummy objects
+     */
+    public function thereIsEmbeddedDummyObjects($nb)
+    {
+        for ($i = 1; $i <= $nb; ++$i) {
+            $dummy = new EmbeddedDummy();
+            $dummy->setName('Dummy #'.$i);
+
+            $embeddableDummy = new EmbeddableDummy();
+            $embeddableDummy->setDummyName('Dummy #'.$i);
+            $dummy->setEmbeddedDummy($embeddableDummy);
+
+            $this->manager->persist($dummy);
+        }
+
+        $this->manager->flush();
+    }
+
+    /**
      * @Given there is :nb dummy objects with relatedDummy
      */
     public function thereIsDummyObjectsWithRelatedDummy($nb)
@@ -185,6 +206,25 @@ class FeatureContext implements Context, SnippetAcceptingContext
             $dummy->setRelatedDummy($relatedDummy);
 
             $this->manager->persist($relatedDummy);
+            $this->manager->persist($dummy);
+        }
+
+        $this->manager->flush();
+    }
+
+    /**
+     * @Given there is :nb dummy objects with embeddedDummy
+     */
+    public function thereIsDummyObjectsWithEmbeddedDummy($nb)
+    {
+        for ($i = 1; $i <= $nb; ++$i) {
+            $embeddableDummy = new EmbeddableDummy();
+            $embeddableDummy->setDummyName('EmbeddedDummy #'.$i);
+
+            $dummy = new EmbeddedDummy();
+            $dummy->setName('Dummy #'.$i);
+            $dummy->setEmbeddedDummy($embeddableDummy);
+
             $this->manager->persist($dummy);
         }
 
@@ -306,6 +346,32 @@ class FeatureContext implements Context, SnippetAcceptingContext
     }
 
     /**
+     * @Given there is :nb embedded dummy objects with dummyDate and embeddedDummy
+     */
+    public function thereIsDummyObjectsWithDummyDateAndEmbeddedDummy($nb)
+    {
+        for ($i = 1; $i <= $nb; ++$i) {
+            $date = new \DateTime(sprintf('2015-04-%d', $i), new \DateTimeZone('UTC'));
+
+            $embeddableDummy = new EmbeddableDummy();
+            $embeddableDummy->setDummyName('Embeddable #'.$i);
+            $embeddableDummy->setDummyDate($date);
+
+            $dummy = new EmbeddedDummy();
+            $dummy->setName('Dummy #'.$i);
+            $dummy->setEmbeddedDummy($embeddableDummy);
+            // Last Dummy has a null date
+            if ($nb !== $i) {
+                $dummy->setDummyDate($date);
+            }
+
+            $this->manager->persist($dummy);
+        }
+
+        $this->manager->flush();
+    }
+
+    /**
      * @Given there is :nb dummy objects with dummyPrice
      */
     public function thereIsDummyObjectsWithDummyPrice($nb)
@@ -348,6 +414,66 @@ class FeatureContext implements Context, SnippetAcceptingContext
             $dummy->setDescription($descriptions[($i - 1) % 2]);
             $dummy->setDummyBoolean($bool);
 
+            $this->manager->persist($dummy);
+        }
+
+        $this->manager->flush();
+    }
+
+    /**
+     * @Given there is :nb embedded dummy objects with embeddedDummy.dummyBoolean :bool
+     */
+    public function thereIsDummyObjectsWithEmbeddedDummyBoolean($nb, $bool)
+    {
+        if (in_array($bool, ['true', '1', 1], true)) {
+            $bool = true;
+        } elseif (in_array($bool, ['false', '0', 0], true)) {
+            $bool = false;
+        } else {
+            $expected = ['true', 'false', '1', '0'];
+            throw new InvalidArgumentException(sprintf('Invalid boolean value for "%s" property, expected one of ( "%s" )', $bool, implode('" | "', $expected)));
+        }
+
+        for ($i = 1; $i <= $nb; ++$i) {
+            $dummy = new EmbeddedDummy();
+            $dummy->setName('Embedded Dummy #'.$i);
+            $embeddableDummy = new EmbeddableDummy();
+            $embeddableDummy->setDummyName('Embedded Dummy #'.$i);
+            $embeddableDummy->setDummyBoolean($bool);
+            $dummy->setEmbeddedDummy($embeddableDummy);
+            $this->manager->persist($dummy);
+        }
+
+        $this->manager->flush();
+    }
+
+    /**
+     * @Given there is :nb embedded dummy objects with relatedDummy.embeddedDummy.dummyBoolean :bool
+     */
+    public function thereIsDummyObjectsWithRelationEmbeddedDummyBoolean($nb, $bool)
+    {
+        if (in_array($bool, ['true', '1', 1], true)) {
+            $bool = true;
+        } elseif (in_array($bool, ['false', '0', 0], true)) {
+            $bool = false;
+        } else {
+            $expected = ['true', 'false', '1', '0'];
+            throw new InvalidArgumentException(sprintf('Invalid boolean value for "%s" property, expected one of ( "%s" )', $bool, implode('" | "', $expected)));
+        }
+
+        for ($i = 1; $i <= $nb; ++$i) {
+            $dummy = new EmbeddedDummy();
+            $dummy->setName('Embedded Dummy #'.$i);
+            $embeddableDummy = new EmbeddableDummy();
+            $embeddableDummy->setDummyName('Embedded Dummy #'.$i);
+            $embeddableDummy->setDummyBoolean($bool);
+
+            $relationDummy = new RelatedDummy();
+            $relationDummy->setEmbeddedDummy($embeddableDummy);
+
+            $dummy->setRelatedDummy($relationDummy);
+
+            $this->manager->persist($relationDummy);
             $this->manager->persist($dummy);
         }
 
--- a/features/doctrine/boolean_filter.feature
+++ b/features/doctrine/boolean_filter.feature
@@ -166,6 +166,210 @@ Feature: Boolean filter on collections
     """
     And the JSON node "hydra:totalItems" should be equal to 10
 
+  Scenario: Get collection by embeddedDummy.dummyBoolean true
+    Given there is "15" embedded dummy objects with embeddedDummy.dummyBoolean true
+    And there is "10" embedded dummy objects with embeddedDummy.dummyBoolean false
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyBoolean=true"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/EmbeddedDummy$"},
+        "@id": {"pattern": "^/embedded_dummies$"},
+        "@type": {"pattern": "^hydra:Collection$"},
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@id": {
+                "oneOf": [
+                  {"pattern": "^/embedded_dummies/1$"},
+                  {"pattern": "^/embedded_dummies/2$"},
+                  {"pattern": "^/embedded_dummies/3$"}
+                ]
+              }
+            }
+          }
+        },
+        "hydra:view": {
+          "type": "object",
+          "properties": {
+            "@id": {"pattern": "^/embedded_dummies\\?embeddedDummy\\.dummyBoolean=true"},
+            "@type": {"pattern": "^hydra:PartialCollectionView$"}
+          }
+        }
+      }
+    }
+    """
+    And the JSON node "hydra:totalItems" should be equal to 15
+
+  Scenario: Get collection by embeddedDummy.dummyBoolean true
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyBoolean=1"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/EmbeddedDummy$"},
+        "@id": {"pattern": "^/embedded_dummies$"},
+        "@type": {"pattern": "^hydra:Collection$"},
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@id": {
+                "oneOf": [
+                  {"pattern": "^/embedded_dummies/1$"},
+                  {"pattern": "^/embedded_dummies/2$"},
+                  {"pattern": "^/embedded_dummies/3$"}
+                ]
+              }
+            }
+          }
+        },
+        "hydra:view": {
+          "type": "object",
+          "properties": {
+            "@id": {"pattern": "^/embedded_dummies\\?embeddedDummy\\.dummyBoolean=1"},
+            "@type": {"pattern": "^hydra:PartialCollectionView$"}
+          }
+        }
+      }
+    }
+    """
+    And the JSON node "hydra:totalItems" should be equal to 15
+
+  Scenario: Get collection by embeddedDummy.dummyBoolean false
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyBoolean=false"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/EmbeddedDummy$"},
+        "@id": {"pattern": "^/embedded_dummies$"},
+        "@type": {"pattern": "^hydra:Collection$"},
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@id": {
+                "oneOf": [
+                  {"pattern": "^/embedded_dummies/16$"},
+                  {"pattern": "^/embedded_dummies/17$"},
+                  {"pattern": "^/embedded_dummies/18$"}
+                ]
+              }
+            }
+          }
+        },
+        "hydra:view": {
+          "type": "object",
+          "properties": {
+            "@id": {"pattern": "^/embedded_dummies\\?embeddedDummy\\.dummyBoolean=false"},
+            "@type": {"pattern": "^hydra:PartialCollectionView$"}
+          }
+        }
+      }
+    }
+    """
+    And the JSON node "hydra:totalItems" should be equal to 10
+
+  Scenario: Get collection by embeddedDummy.dummyBoolean false
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyBoolean=0"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/EmbeddedDummy$"},
+        "@id": {"pattern": "^/embedded_dummies$"},
+        "@type": {"pattern": "^hydra:Collection$"},
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@id": {
+                "oneOf": [
+                  {"pattern": "^/embedded_dummies/16$"},
+                  {"pattern": "^/embedded_dummies/17$"},
+                  {"pattern": "^/embedded_dummies/18$"}
+                ]
+              }
+            }
+          }
+        },
+        "hydra:view": {
+          "type": "object",
+          "properties": {
+            "@id": {"pattern": "^/embedded_dummies\\?embeddedDummy\\.dummyBoolean=0"},
+            "@type": {"pattern": "^hydra:PartialCollectionView$"}
+          }
+        }
+      }
+    }
+    """
+    And the JSON node "hydra:totalItems" should be equal to 10
+
+  Scenario: Get collection by association with embed relatedDummy.embeddedDummy.dummyBoolean true
+    Given there is "15" embedded dummy objects with relatedDummy.embeddedDummy.dummyBoolean true
+    And there is "10" embedded dummy objects with relatedDummy.embeddedDummy.dummyBoolean false
+    When I send a "GET" request to "/embedded_dummies?relatedDummy.embeddedDummy.dummyBoolean=true"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/EmbeddedDummy$"},
+        "@id": {"pattern": "^/embedded_dummies$"},
+        "@type": {"pattern": "^hydra:Collection$"},
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@id": {
+                "oneOf": [
+                  {"pattern": "^/embedded_dummies/26$"},
+                  {"pattern": "^/embedded_dummies/27$"},
+                  {"pattern": "^/embedded_dummies/28$"}
+                ]
+              }
+            }
+          }
+        },
+        "hydra:view": {
+          "type": "object",
+          "properties": {
+            "@id": {"pattern": "^/embedded_dummies\\?relatedDummy.embeddedDummy\\.dummyBoolean=true"},
+            "@type": {"pattern": "^hydra:PartialCollectionView$"}
+          }
+        }
+      }
+    }
+    """
+    And the JSON node "hydra:totalItems" should be equal to 15
+
   @dropSchema
   Scenario: Get collection ordered by a non valid properties
     When I send a "GET" request to "/dummies?unknown=0"
--- a/features/doctrine/date_filter.feature
+++ b/features/doctrine/date_filter.feature
@@ -395,14 +395,269 @@ Feature: Date filter on collections
     And the JSON should be equal to:
     """
     {
-          "@context": "/contexts/Dummy",
-          "@id": "/dummies",
-          "@type": "hydra:Collection",
-          "hydra:member": [],
-          "hydra:totalItems": 0,
-          "hydra:view": {
-              "@id": "/dummies?relatedDummy.dummyDate%5Bafter%5D=2015-04-28",
-              "@type": "hydra:PartialCollectionView"
+      "@context": "/contexts/Dummy",
+      "@id": "/dummies",
+      "@type": "hydra:Collection",
+      "hydra:member": [],
+      "hydra:totalItems": 0,
+      "hydra:view": {
+        "@id": "/dummies?relatedDummy.dummyDate%5Bafter%5D=2015-04-28",
+        "@type": "hydra:PartialCollectionView"
+      },
+      "hydra:search": {
+        "@type": "hydra:IriTemplate",
+        "hydra:template": "/dummies{?dummyBoolean,relatedDummy.embeddedDummy.dummyBoolean,dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],description[exists],relatedDummy.name[exists],dummyBoolean[exists],dummyFloat,dummyPrice,order[id],order[name],order[relatedDummy.name],order[relatedDummy.symfony],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,relatedDummies.name}",
+        "hydra:variableRepresentation": "BasicRepresentation",
+        "hydra:mapping": [
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyBoolean",
+            "property": "dummyBoolean",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummy.embeddedDummy.dummyBoolean",
+            "property": "relatedDummy.embeddedDummy.dummyBoolean",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyDate[before]",
+            "property": "dummyDate",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyDate[after]",
+            "property": "dummyDate",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummy.dummyDate[before]",
+            "property": "relatedDummy.dummyDate",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummy.dummyDate[after]",
+            "property": "relatedDummy.dummyDate",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "description[exists]",
+            "property": "description",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummy.name[exists]",
+            "property": "relatedDummy.name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyBoolean[exists]",
+            "property": "dummyBoolean",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyFloat",
+            "property": "dummyFloat",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyPrice",
+            "property": "dummyPrice",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "order[id]",
+            "property": "id",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "order[name]",
+            "property": "name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "order[relatedDummy.name]",
+            "property": "relatedDummy.name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "order[relatedDummy.symfony]",
+            "property": "relatedDummy.symfony",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyFloat[between]",
+            "property": "dummyFloat",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyFloat[gt]",
+            "property": "dummyFloat",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyFloat[gte]",
+            "property": "dummyFloat",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyFloat[lt]",
+            "property": "dummyFloat",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyFloat[lte]",
+            "property": "dummyFloat",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyPrice[between]",
+            "property": "dummyPrice",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyPrice[gt]",
+            "property": "dummyPrice",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyPrice[gte]",
+            "property": "dummyPrice",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyPrice[lt]",
+            "property": "dummyPrice",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummyPrice[lte]",
+            "property": "dummyPrice",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "id",
+            "property": "id",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "id[]",
+            "property": "id",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "name",
+            "property": "name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "alias",
+            "property": "alias",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "description",
+            "property": "description",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummy.name",
+            "property": "relatedDummy.name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummy.name[]",
+            "property": "relatedDummy.name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummies",
+            "property": "relatedDummies",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummies[]",
+            "property": "relatedDummies",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "dummy",
+            "property": "dummy",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "relatedDummies.name",
+            "property": "relatedDummies.name",
+            "required": false
+          }
+        ]
+      }
+    }
+    """
+
+  @dropSchema
+  @createSchema
+  Scenario: Get collection filtered by embedded date
+    Given there is "2" embedded dummy objects with dummyDate and embeddedDummy
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyDate[after]=2015-04-28"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/EmbeddedDummy$"},
+        "@id": {"pattern": "^/embedded_dummies$"},
+        "@type": {"pattern": "^hydra:Collection$"},
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@id": {
+                "oneOf": [
+                  {"pattern": "^/embedded_dummies/28$"},
+                  {"pattern": "^/embedded_dummies/29$"}
+                ]
+              }
+            }
           },
           "hydra:search": {
               "@type": "hydra:IriTemplate",
@@ -614,6 +869,15 @@ Feature: Date filter on collections
                       "required": false
                   }
               ]
+          },
+          "hydra:view": {
+            "type": "object",
+            "properties": {
+              "@id": {"pattern": "^/embedded_dummies\\?embeddedDummy\\.dummyDate%5Bafter%5D=2015-04-28$"},
+              "@type": {"pattern": "^hydra:PartialCollectionView$"}
+            }
           }
+        }
       }
+    }
     """
--- a/features/doctrine/order_filter.feature
+++ b/features/doctrine/order_filter.feature
@@ -331,7 +331,7 @@ Feature: Order filter on collections
               "properties": {
                 "@id": {
                   "type": "string",
-                  "pattern": "^/dummies/2$"
+                  "pattern": "^/dummies/2"
                 }
               }
             },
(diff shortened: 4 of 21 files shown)


src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -116,7 +116,7 @@ protected function isPropertyEnabled(string $property): bool
      */
     protected function isPropertyMapped(string $property, string $resourceClass, bool $allowAssociation = false): bool
     {
-        if ($this->isPropertyNested($property)) {
+        if ($this->isPropertyNested($property) && !$this->isPropertyNestedEmbedded($property, $resourceClass)) {
```
soyuka: Is this condition valid everytime? I mean, will it break current implementations if we put this inside `isPropertyNested` so that other filters also benefits from this? Maybe it's just the boolean filter that's impacted, but I doubt that.

jordscream:
To explain a little bit:
- Embed field is a nested property. So it makes sense according to me to recognize this afield as a nested property. However, we need to dissociate nestedAssociation versus nestedEmbedded.
- So I create the method isPropertyNestedEmbedded in AbstractFilter and apply it when necessary. This functions just needs to know if the Metadata owns this fields. An Embed field is managed like a simple field
- I modify a little bit the method isBooleanField in BooleanFilter to retrieve metadata in the same way as the function isPropertyMapped in AbstractFilter
- I modify the QueryNameGenerator of API platform to replace . by _ cause it generates an error when we set the parameter (not pretty)
- I did not develop isPropertyAssociation cause it costs performance so I prefer to use isPropertyNested and use Negation in the code to know if the property is a nested association

soyuka:
Is this issue limited to the `booleanFilter`?

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
jordscream: Yes it will breaks 👍 
If we put the code in isPropertyNested, it disturbs me for differents reasons:
- 1/ it means that a embed field is not considered as a nested field. In fact it is by definition :-)
- 2/ the function extractProperties in AbstractFilter use isPropertyNested and for an Embed field, it has to be needsFixing to true. Otherwise, it fails for the request.

This code is in AbstractFilter, so any Filter will benefit the code.
Normannly the only thing to change for each filter is isBooleanField, isDateField, isNumericField...

jordscream:
@soyuka , for the moment yes. I'm waiting your feedback before to apply it in any filter and write **STRONG tests** about this

soyuka:
What about modifying `isPropertyNested` directly then? I mean if the condition always apply, it'd be cleaner than changing every `isPropertyNested` by `isPropertyNested && !isPropertyNestedEmbedded`

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
soyuka: Okay got it. I'd add a method that ends up doing `$this->isPropertyNested($property) && !$this->isPropertyNestedEmbedded($property, $resourceClass)` though. Feel free to add tests.

teohhanhui: I do not agree that an `Embedded` field should be considered a "nested property". `isNestedProperty` is to help us find out when we need to join.

soyuka: Agreed ofc. Smth like `isPropertyNestedFilterable`

teohhanhui: Ideally we can just treat embedded fields as normal fields. That'd avoid complicating the code, and limit the changes to only `AbstractFilter` itself. We'd need to change `splitPropertyParts` and `isPropertyNested`.

jordscream: @teohhanhui , @soyuka , to be sure beacuse I'm french :-D . When we add an embed field, we create a "nested" property access ? 
Example:
- Entity A contains Embed B -> we access like entity_a.entity_b.field. It is considered as a nested access ? (difficult to explain with words :-D)

teohhanhui: `splitPropertyParts('entity_a.embed_b.field')` should return:

```php
[
    'associations' => [
        'entity_a',
    ],
    'field' => 'embed_b.field',
]
```

jordscream: @teohhanhui , you're right for that. So We confirm that a nested property concerns only join ?

jordscream: @teohhanhui @soyuka , what about extractProperties method in the class AbstracFilter ? The property $needsFixing to true. Otherwise the embed field embed_b.field is transformed to embed_b_field and generates an error in the request

teohhanhui: Yes, that's the approach that I'd take.

Embedded fields are inlined to the field mapping, apparently (see https://github.com/doctrine/doctrine2/blob/v2.5.6/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L3230-L3264), so there should be no further changes necessary.

teohhanhui: Yes, we also need to fix `extractProperties`. `$needsFixing = true` for embedded fields. :smile:

teohhanhui: So you may still find it valuable to extract the logic into, say, `isPropertyEmbedded`.

jordscream:
@teohhanhui @soyuka , second commit for discussion (apply the code in all filters). No unit and functional tests are developped for the moment. But existing tests success

soyuka:
Better, I'd need @teohhanhui to confirm but I think that we break backward changes by adding `$resourceClass`. This'll have to target master. 

@jordscream would be nice to have a functional test involving those embedded properties. For unit tests, coverage looks fine not sure you'd be able to add much.

jordscream:
@soyuka , no problem for functionnal test, writing in progress

antograssiot:
In unit test, I might be wrong but I wanted to make sure that the filter description worked ok.

jordscream:
@soyuka @teohhanhui Why we call dropSchema to execute an insert just after ?
 https://github.com/api-platform/core/blob/2.0/features/doctrine/date_filter.feature#L267-L269

soyuka:
`dropSchema` executes after the scenario is complete (https://github.com/api-platform/core/blob/2.0/features/bootstrap/FeatureContext.php#L88)

jordscream:
My apologize....

jordscream:
@soyuka @teohhanhui I had functionnal tests. 

Tell me if it is enough. The most important to me is to test of at least on filter a model like this entity.relation.embed.field.

jordscream:
@soyuka @teohhanhui any news ? :-)

features/doctrine/date_filter.feature
```diff
@@ -394,7 +393,7 @@ Feature: Date filter on collections
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
     And the JSON should be equal to:
     """
-    {
+      {
```
soyuka: Indentation looks wrong here but it's maybe github failing

features/doctrine/date_filter.feature
```diff
@@ -593,3 +598,262 @@ Feature: Date filter on collections
...
+
+  @dropSchema
+  @createSchema
+  Scenario: Get collection filtered by embedded date
+    Given there is "2" embedded dummy objects with dummyDate and embeddedDummy
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyDate[after]=2015-04-28"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
```
soyuka: Would it be possible to use a schema here instead? Or you can play with `JSON node should equal`. It's a pain to maintain when we add more filters. Also, we already tested the full JSON object above.

features/main/crud.feature
```diff
@@ -92,223 +92,229 @@ Feature: Create-Retrieve-Update-Delete
...
+                      "@type": "IriTemplateMapping",
+                      "variable": "dummyPrice",
+                      "property": "dummyPrice",
+                      "required": false
+                  }
+              ]
           }
-        ]
       }
-    }
```
soyuka: Same indentation looks weird here from github POV. This is the maintain mess I'm talking above :p.

soyuka:
Shit failed the review thing \o/. Anyway great job there! However, this has a BC break and should target `master`.

jordscream:
@soyuka @teohhanhui , thanks for your feedback. I gonna check indentation and I will rebase to master.

Just 2 questions:
- Why it is a BC breaks ? 
- @soyuka , can you give me more details about your comment ? I am not sure to understand

Thanks API platform Team !

soyuka:
> Why it is a BC breaks ?

Because it alters methods that custom filters might use by extending our AbstractFilter. 

For example, I use it with `extractProperties`, if I update it'll break my code because the second argument is mandatory.

> can you give me more details about your comment ? I am not sure to understand

which one?

jordscream:
@soyuka Ok , thanks for the explanation

This comment:
Would it be possible to use a schema here instead? Or you can play with JSON node should equal. It's a pain to maintain when we add more filters.

soyuka:
Oh yes, just like you did the other tests, you're using `And the JSON should be valid according to this schema:` instead of `And the JSON should be equal to:`.

When I speak about `nodes` you can write tests [like this as well](https://github.com/api-platform/core/pull/950/files#diff-534f9c66d94968377ac76a8bc211da4dR13):

```
And the JSON node "_embedded.item" should have 1 element 
```

jordscream:
@soyuka , not sure of my modification accorgin your comment but I did it in boolean_filter.feature. Can you tell me if it is ok for you before I modify my other functionnal test ?

thanks

features/doctrine/boolean_filter.feature
```diff
@@ -237,11 +279,927 @@ Feature: Boolean filter on collections
...
     }
     """
+    And the JSON node "hydra:totalItems" should be equal to 10
+
+  Scenario: Get collection by embeddedDummy.dummyBoolean false
+    When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyBoolean=0"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
```
soyuka: `And the JSON should be valid according to this schema:`

features/doctrine/boolean_filter.feature
```diff
@@ -237,11 +279,927 @@ Feature: Boolean filter on collections
...
+    And the JSON node "hydra:totalItems" should be equal to 10
+
+  Scenario: Get collection by association with embed relatedDummy.embeddedDummy.dummyBoolean true
+    Given there is "15" embedded dummy objects with relatedDummy.embeddedDummy.dummyBoolean true
+    And there is "10" embedded dummy objects with relatedDummy.embeddedDummy.dummyBoolean false
+    When I send a "GET" request to "/embedded_dummies?relatedDummy.embeddedDummy.dummyBoolean=true"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
```
soyuka: `And the JSON should be valid according to this schema:`

jordscream:
@soyuka 
ah, according your comments , you say to me to do the opposite :-)

> Oh yes, just like you did the other tests, you're using And the JSON should be valid according to this schema: instead of And the JSON should be equal to

So I need to replace all my the JSON should be equal to by And the JSON should be valid according to this schema ?

soyuka:
Oh sorry about this, rephrasing :

use schema instead of full json equality for the sake of maintainability.

Really sorry that you misunderstood this I was just doing too many things at once yesterday and speaking 3 languages at the same time makes it hard to write comprehensible things sometimes :|.

jordscream:
@soyuka done ^^

soyuka:
Really nice job thanks! @antograssiot if you've some time would you mind taking a quick look at the tests?

features/bootstrap/FeatureContext.php
```diff
@@ -13,6 +13,8 @@
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\CompositeLabel;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\CompositeRelation;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\EmbeddableDummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\EmbeddedDummy;
```
teohhanhui: 1. Add the property to `Dummy` instead of having another `EmbeddedDummy`.
2. Rename `EmbeddableDummy` to `EmbeddedDummy`.

antograssiot:
I'll have a look tonight to the code and will test against my application tomorrow

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -131,12 +132,28 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
      *
      * @return bool
      */
-    protected function isPropertyNested(string $property): bool
+    protected function isPropertyEmbedded(string $property, string $resourceClass): bool
     {
-        return false !== strpos($property, '.');
+        $metadata = $this->getClassMetadata($resourceClass);
+
+        return $metadata->hasField($property);
```
teohhanhui: This is wrong. The check should be:

```php
if (false === $pos = strrpos($property, '.', -1)) {
    return false;
}

$metadata = $this->getClassMetadata($resourceClass);

return $metadata->hasField(substr($property, $pos + 1));
```

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -131,12 +132,28 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
      * Determines whether the given property is nested.
      *
      * @param string $property
+     * @param string $resourceClass
+     *
+     * @return bool
+     */
+    protected function isPropertyNested(string $property, string $resourceClass): bool
+    {
+        return false !== strpos($property, '.') && false === $this->isPropertyEmbedded($property, $resourceClass);
```
teohhanhui: This is wrong. The check should be:

```php
if (false === $pos = strpos($property, '.')) {
    return false;
}

$metadata = $this->getClassMetadata($resourceClass);

return $metadata->hasAssociation(substr($property, 0, $pos));
```

features/bootstrap/FeatureContext.php
jordscream: @teohhanhui , it means, I have to fix a lot of functionnals tests cause of a new field in Dummy entity :-/ . If you think it is the right way....

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
jordscream: @teohhanhui , no pb

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
jordscream: @teohhanhui , the metadata refers embed field like with the "." hasField("embed.field") works only. Do you want I override the Symfony method ?
thanks for your feedback

features/bootstrap/FeatureContext.php
soyuka: Agreed with @jordscream on this one, sadly adding properties to tested entities is a pain with the current tests implementation.

features/doctrine/boolean_filter.feature
```diff
@@ -166,6 +166,210 @@ Feature: Boolean filter on collections
...
+                  {"pattern": "^/embedded_dummies/3$"}
+                ]
+              }
+            }
+          }
+        },
+        "hydra:view": {
+          "type": "object",
+          "properties": {
+            "@id": {"pattern": "^/embedded_dummies\\?embeddedDummy.dummyBoolean=true"},
```
antograssiot: Not a big deal but I guess the strict pattern would be `"^/embedded_dummies\\?embeddedDummy\\.dummyBoolean=true"` here and in other places.
Not sure it is worth updating it though, it's quite unlikely to be biaised

features/doctrine/order_filter.feature
```diff
@@ -360,6 +360,66 @@ Feature: Order filter on collections
     }
     """
 
+  Scenario: Get collection ordered in ascending order on an embedded and on which order filter has been enabled in whitelist mode
+    Given there is "30" dummy objects with embeddedDummy
+    When I send a "GET" request to "/embedded_dummies?order[embeddedDummy]=asc"
```
antograssiot: Just a question @jordscream  as I don't see test for that, would sending a request to `"/embedded_dummies?order[embeddedDummy.dummyName]=asc"`  work ?

tests/Fixtures/app/config/config.yml
```diff
@@ -94,17 +94,17 @@ services:
...
     app.my_dummy_resource.search_filter:
         parent:    'api_platform.doctrine.orm.search_filter'
-        arguments: [ { 'id': 'exact', 'name': 'partial', 'alias': 'start', 'description': 'word_start', 'relatedDummy.name': 'exact', 'relatedDummies': 'exact', 'dummy': 'ipartial' } ]
+        arguments: [ { 'id': 'exact', 'name': 'partial', 'alias': 'start', 'description': 'word_start', 'relatedDummy.name': 'exact', 'relatedDummies': 'exact', 'dummy': 'ipartial', 'embeddedDummy.dummyName': 'partial' } ]
         tags:      [ { name: 'api_platform.filter', id: 'my_dummy.search' } ]
 
     app.my_dummy_resource.order_filter:
         parent:    'api_platform.doctrine.orm.order_filter'
-        arguments: [ { 'id': ~, 'name': 'desc', 'relatedDummy.symfony': ~ } ]
+        arguments: [ { 'id': ~, 'name': 'desc', 'relatedDummy.symfony': ~, 'embeddedDummy.symfony': ~ } ]
```
antograssiot: I might be wrong but you didn't used it from what I can see. The property doesn't even exists, is it on purpose ?

jordscream: @antograssiot , you're right I test nothing about this. I should order[embeddedDummy.symfony]=asc to respond to your comment above.

features/doctrine/order_filter.feature
jordscream: @antograssiot I will do with embeddedDummy.symfony

features/doctrine/boolean_filter.feature
jordscream: I'll check @antograssiot

jordscream:
@antograssiot @soyuka @teohhanhui , I applied modifications you suggest to me
Do not hesitate for any changes

Thanks 4 all

tests/Fixtures/app/config/config.yml
```diff
@@ -94,17 +94,17 @@ services:
...
     app.my_dummy_resource.search_filter:
         parent:    'api_platform.doctrine.orm.search_filter'
-        arguments: [ { 'id': 'exact', 'name': 'partial', 'alias': 'start', 'description': 'word_start', 'relatedDummy.name': 'exact', 'relatedDummies': 'exact', 'dummy': 'ipartial' } ]
+        arguments: [ { 'id': 'exact', 'name': 'partial', 'alias': 'start', 'description': 'word_start', 'relatedDummy.name': 'exact', 'relatedDummies': 'exact', 'dummy': 'ipartial', 'embeddedDummy.dummyName': 'partial' } ]
         tags:      [ { name: 'api_platform.filter', id: 'my_dummy.search' } ]
 
     app.my_dummy_resource.order_filter:
         parent:    'api_platform.doctrine.orm.order_filter'
-        arguments: [ { 'id': ~, 'name': 'desc', 'relatedDummy.symfony': ~ } ]
+        arguments: [ { 'id': ~, 'name': 'desc', 'relatedDummy.name': ~, 'embeddedDummy.dummyName': 'desc' } ]
```
antograssiot: @jordscream the change from `relatedDummy.symfony` to `relatedDummy.name` is good I think but causes other tests to fail.

here is the diff for master of what needs to be updated if it can help.
```diff
diff --git a/features/doctrine/date_filter.feature b/features/doctrine/date_filter.feature
index a5cc528..54b70ad 100644
--- a/features/doctrine/date_filter.feature
+++ b/features/doctrine/date_filter.feature
@@ -406,7 +406,7 @@ Feature: Date filter on collections
           },
           "hydra:search": {
               "@type": "hydra:IriTemplate",
-              "hydra:template": "\/dummies{?id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,order[id],order[name],order[relatedDummy.symfony],dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],dummyBoolean,dummyFloat,dummyPrice}",
+              "hydra:template": "\/dummies{?id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,order[id],order[name],order[relatedDummy.name],dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],dummyBoolean,dummyFloat,dummyPrice}",
               "hydra:variableRepresentation": "BasicRepresentation",
               "hydra:mapping": [
                   {
@@ -483,8 +483,8 @@ Feature: Date filter on collections
                   },
                   {
                       "@type": "IriTemplateMapping",
-                      "variable": "order[relatedDummy.symfony]",
-                      "property": "relatedDummy.symfony",
+                      "variable": "order[relatedDummy.name]",
+                      "property": "relatedDummy.name",
                       "required": false
                   },
                   {
diff --git a/features/main/crud.feature b/features/main/crud.feature
index 2c534c7..2a8793c 100644
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -123,7 +123,7 @@ Feature: Create-Retrieve-Update-Delete
       "hydra:totalItems": 1,
       "hydra:search": {
         "@type": "hydra:IriTemplate",
-        "hydra:template": "/dummies{?id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,order[id],order[name],order[relatedDummy.symfony],dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],dummyBoolean,dummyFloat,dummyPrice}",
+        "hydra:template": "/dummies{?id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,order[id],order[name],order[relatedDummy.name],dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],dummyBoolean,dummyFloat,dummyPrice}",
         "hydra:variableRepresentation": "BasicRepresentation",
         "hydra:mapping": [
           {
@@ -200,8 +200,8 @@ Feature: Create-Retrieve-Update-Delete
           },
           {
             "@type": "IriTemplateMapping",
-            "variable": "order[relatedDummy.symfony]",
-            "property": "relatedDummy.symfony",
+            "variable": "order[relatedDummy.name]",
+            "property": "relatedDummy.name",
             "required": false
           },
           {

Comment (jordscream) [inline tests/Fixtures/app/config/config.yml]: @antograssiot , thks for that!
fixed now

File: src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php [hunk 14]
```diff
@@ -131,12 +132,38 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+     * @param string $property
+     * @param string $resourceClass
      *
      * @return bool
      */
-    protected function isPropertyNested(string $property): bool
+    protected function isPropertyEmbedded(string $property, string $resourceClass): bool
     {
-        return false !== strpos($property, '.');
+        if (false === $pos = strrpos($property, '.', -1)) {
```
meyerbaptiste:
```php
return false !== strrpos($property, '.', -1) && $this->getClassMetadata($resourceClass)->hasField($property);
```

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -131,12 +132,38 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+     */
+    protected function isPropertyNested(string $property, string $resourceClass): bool
+    {
+        if (false === $pos = strpos($property, '.')) {
+            return false;
+        }
+
+        $metadata = $this->getClassMetadata($resourceClass);
+
+        return $metadata->hasAssociation(substr($property, 0, $pos));
```
meyerbaptiste:
```php
$pos = strpos($property, '.');

return false !== $pos && $this->getClassMetadata($resourceClass)->hasAssociation(substr($property, 0, $pos));
```

features/bootstrap/FeatureContext.php
teohhanhui: Do you think we can get another PR merged first to modernize all the Behat test scenarios? :)

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
jordscream: @meyerbaptiste , thanks for reviews. @teohhanhui already proposed me a code changes. What is the right way to do it ? Just to prevent a minimal change, which way I have to take when I propose this kind of code ?

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
jordscream: @meyerbaptiste , thanks for reviews. @teohhanhui already proposed me a code changes. What is the right way to do it ? Just to prevent a minimal change, which way I have to take when I propose this kind of code ?

features/bootstrap/FeatureContext.php
jordscream: @teohhanhui @soyuka , just my opinion by working a little bit in functionnal test as a small contributor. It's really hard cause any change of existing tested entities, you have to fix a lots a functionnal tests (maybe due to comparaison with JSON should be equal to... and comparaison with entity field definition...) . An Issue about that would be good ^^

soyuka: > Do you think we can get another PR merged first to modernize all the Behat test scenarios? :)
  
Modernize how so? IMO we should avoid strict equality in the behat tests and anyway we can fix this later. I think that @dunglas wants to find/build something better than behat for our api tests anyway.

dunglas: Please do not add new properties to existing entities :) It's better to have a dedicated entity for every class of functional tests.
And they are just... tests. We'll not stop merging feature because tests can be modernized :)

dunglas:
There are a lot of conflicts. Is it intended?

dunglas:
And, as it's a new feature, it should target master.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -92,14 +92,15 @@ protected function getClassMetadata(string $resourceClass): ClassMetadata
      * Determines whether the given property is enabled.
      *
      * @param string $property
+     * @param string $resourceClass
      *
      * @return bool
      */
-    protected function isPropertyEnabled(string $property): bool
+    protected function isPropertyEnabled(string $property, string $resourceClass): bool
```
dunglas: We cannot change the signature of a `protected` method like this. It will cause a BC break. You can handle this like that: https://github.com/symfony/symfony/pull/19371/files#diff-eda6922ceec0d4a52894dd37d7772dc1R180

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
dunglas: I prefer the @meyerbaptiste way but it's just a matter of taste IMO.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -173,33 +200,45 @@ protected function getNestedMetadata(string $resourceClass, array $associations)
      *   - field: string holding the actual field (leaf node)
      *
      * @param string $property
+     * @param string $resourceClass
      *
      * @return array
      */
-    protected function splitPropertyParts(string $property): array
+    protected function splitPropertyParts(string $property, string $resourceClass): array
```
dunglas: Same here regarding BC.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -173,33 +200,45 @@ protected function getNestedMetadata(string $resourceClass, array $associations)
...
     /**
      * Extracts properties to filter from the request.
      *
      * @param Request $request
+     * @param string  $resourceClass
      *
      * @return array
      */
-    protected function extractProperties(Request $request): array
+    protected function extractProperties(Request $request, string $resourceClass): array
```
dunglas: Same here for BC.

soyuka:
> And, as it's a new feature, it should target master.

To me it's more a fix then a feature, there's no reason for the filters not to work with embedded fields no?

> We cannot change the signature of a protected method like this. It will cause a BC break. 

We discussed the BC break above, can't we break things in master :D? If this is not considered as a fix (so target master) it still needs to be break-free?

dunglas:
It has never been supported so it's a new feature :)
It's an incentive for users to upgrade an I'm not confident to include a path of 1k LOC in a patch release.

teohhanhui:
We cannot have BC breaks even in `master` :disappointed:

But, I agree this is a bug fix. Support for embedded fields was broken, because they just weren't handled.

dunglas:
We know how to deprecate things easily (there is a solution here). There is no reason make the upgrade painful for users with the introduction of BC. At some point we'll release a new major version and remove the deprecated code (probably some months after the release of Symfony 4, and we'll support only SF 4+ at the same time).

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
teohhanhui: @jordscream: @meyerbaptiste is right. My code was wrong. :smile:

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
teohhanhui: We don't need to make code hard to read just to save on a variable.

teohhanhui: If we're going that way, why not make it one-liner?

```php
return (false !== $pos = strpos($property, '.')) && $this->getClassMetadata($resourceClass)->hasAssociation(substr($property, 0, $pos));

**Comment by soyuka** [comment]
This means that because I broke a lot of things in #904 it won't be available until symfony 4 support? :cry:

**Comment by teohhanhui** [comment]
> We know how to deprecate things easily (there is a solution here). There is no reason to make the upgrade painful for users with the introduction of BC breaks.

There is a good reason: it makes for (really) bad design. Cruft builds up and bad design leads to more bad design. But this discussion can happen elsewhere. :smile:

**Comment by jordscream** [comment]
@dunglas , So what is your decision ? no BC breaks in master ?

**Comment by dunglas** [comment]
@jordscream no BC breaks. Master is actually 2.1.

**Comment by jordscream** [comment]
@dunglas @teohhanhui @soyuka  , rebase to master, fix conflicts + apply your suggestion to prevent BC break in master

**Comment by jordscream** [comment]
@dunglas tell me when you'll have time to reviews your suggest modification about no BC break in master. ;-)

File: src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php [hunk 19]
```diff
@@ -95,11 +95,23 @@ protected function getClassMetadata(string $resourceClass): ClassMetadata
...
-    protected function isPropertyEnabled(string $property): bool
+    protected function isPropertyEnabled(string $property/*, string $resourceClass*/): bool
     {
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version Symfony 4.0. Not defining it is deprecated since 3.3.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
dunglas: `API Platform 3.0`, `API Platform 2.1`.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -134,9 +146,34 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+    protected function isPropertyNested(string $property/*, string $resourceClass*/): bool
     {
-        return false !== strpos($property, '.');
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version Symfony 4.0. Not defining it is deprecated since 3.3.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
dunglas: `API Platform 3.0`, `API Platform 2.1`.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -176,13 +213,40 @@ protected function getNestedMetadata(string $resourceClass, array $associations)
...
     {
         $parts = explode('.', $property);
 
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version Symfony 4.0. Not defining it is deprecated since 3.3.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
dunglas: `API Platform 3.0`, `API Platform 2.1`.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -193,13 +257,25 @@ protected function splitPropertyParts(string $property): array
...
-    protected function extractProperties(Request $request): array
+    protected function extractProperties(Request $request/*, string $resourceClass*/): array
     {
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version Symfony 4.0. Not defining it is deprecated since 3.3.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
dunglas: `API Platform 3.0`, `API Platform 2.1`.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -226,9 +302,21 @@ protected function extractProperties(Request $request): array
...
+    protected function addJoinsForNestedProperty(string $property, string $rootAlias, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator/*, string $resourceClass*/): array
     {
-        $propertyParts = $this->splitPropertyParts($property);
+        if (func_num_args() > 4) {
+            $resourceClass = func_get_arg(4);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version Symfony 4.0. Not defining it is deprecated since 3.3.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
dunglas: `API Platform 3.0`, `API Platform 2.1`.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -134,9 +146,34 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+                }
+            }
+            $resourceClass = null;
+        }
+
+        return false !== ($pos = strpos($property, '.')) && !$resourceClass || (false !== $pos && $this->getClassMetadata($resourceClass)->hasAssociation(substr($property, 0, $pos)));
```
dunglas: May I suggest the following to make the code faster and easier to read?

```php
if (false === $pos = strpos($property, '.')) {
    return false;
}

return null === $resourceClass || $this->getClassMetadata($resourceClass)->hasAssociation(substr($property, 0, $pos));
```

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -134,9 +146,34 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+     * Determines whether the given property is embedded.
+     *
+     * @param string $property
+     * @param string $resourceClass
+     *
+     * @return bool
+     */
+    protected function isPropertyEmbedded(string $property, string $resourceClass): bool
+    {
+        return false !== ($pos = strpos($property, '.')) && $this->getClassMetadata($resourceClass)->hasField($property);
```
dunglas: The assignation of `$pos` is useless:

```php
return false !== strpos($property, '.') && $this->getClassMetadata($resourceClass)->hasField($property);
```

jordscream:
@dunglas , will make some changes soon

jordscream:
@soyuka , I rebase to master to fix conflict with your recent new feature but a behat test do not pass and I really do not why. Can you help me ? :-)

jordscream:
@dunglas , I made changes as you suggest

features/bootstrap/FeatureContext.php
```diff
@@ -118,6 +120,25 @@ public function thereIsDummyObjects($nb)
     }
 
     /**
+     * @Given there is :nb embedded dummy objects
+     */
+    public function thereIsEmbeddedDummyObjects($nb)
```
soyuka: should be "there are :nb embedded dummy objects" :D

soyuka:
screen shot 2017-04-06 at 08 39 10

And raw logs look like shit because of color chars :|. I'll clone your fork asap to try it out.

features/bootstrap/FeatureContext.php
jordscream: oho @soyuka  good catch, stupid merge ^^

jordscream:
@soyuka , we keep in touch when you can have a look

jordscream:
@soyuka fixed @dunglas you can review also if u want

features/doctrine/order_filter.feature
```diff
@@ -322,7 +322,7 @@ Feature: Order filter on collections
               "properties": {
                 "@id": {
                   "type": "string",
-                  "pattern": "^/dummies/1$"
+                  "pattern": "^/dummies/31$"
```
soyuka: Why the change?

features/doctrine/order_filter.feature
```diff
@@ -302,7 +302,7 @@ Feature: Order filter on collections
 
   Scenario: Get collection ordered in ascending order on an association and on which order filter has been enabled in whitelist mode
     Given there is "30" dummy objects with relatedDummy
-    When I send a "GET" request to "/dummies?order[relatedDummy]=asc"
+    When I send a "GET" request to "/dummies?order[relatedDummy.name]=asc"
```
soyuka: why the change? We should do a new test if we want to test the name property no?

features/doctrine/order_filter.feature
```diff
@@ -331,7 +331,7 @@ Feature: Order filter on collections
               "properties": {
                 "@id": {
                   "type": "string",
-                  "pattern": "^/dummies/2$"
+                  "pattern": "^/dummies/40"
```
soyuka: Same I don't get how this change is related to the current patch

jordscream:
@soyuka , I revert my change about functional test to not alter the existing functional test
Let me know if it is okay for you

tests/Fixtures/app/config/config.yml
```diff
@@ -94,18 +94,18 @@ services:
...
-        arguments: [ { 'id': 'exact', 'name': 'partial', 'alias': 'start', 'description': 'word_start', 'relatedDummy.name': 'exact', 'relatedDummies': 'exact', 'dummy': 'ipartial', 'relatedDummies.name': 'start' } ]
+        arguments: [ { 'id': 'exact', 'name': 'partial', 'alias': 'start', 'description': 'word_start', 'relatedDummy.name': 'exact', 'relatedDummies': 'exact', 'dummy': 'ipartial', 'relatedDummies.name': 'start', 'embeddedDummy.dummyName': 'partial' } ]
         tags:      [ { name: 'api_platform.filter', id: 'my_dummy.search' } ]
 
     # Tests if the id default to the service name, do not add id attributes here
     my_dummy.order:
         parent:    'api_platform.doctrine.orm.order_filter'
-        arguments: [ { 'id': ~, 'name': 'desc', 'relatedDummy.symfony': ~ } ]
-        tags:      [ { name: 'api_platform.filter' } ]
+        arguments: [ { 'id': ~, 'name': 'desc', 'relatedDummy.name': ~, 'embeddedDummy.dummyName': 'desc' } ]
```
soyuka: you removed `relatedDummy.symfony` order, I assume this is wanted? As long as tests pass it's ok to me 👍

soyuka:
php-cs-fixer failed, you can check the diff on travis

jordscream:
@soyuka done (normally)

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -136,9 +148,38 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+    protected function isPropertyNested(string $property/*, string $resourceClass*/): bool
     {
-        return false !== strpos($property, '.');
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
meyerbaptiste: Your `sprintf()` has too many args.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -97,11 +97,23 @@ protected function getClassMetadata(string $resourceClass): ClassMetadata
...
-    protected function isPropertyEnabled(string $property): bool
+    protected function isPropertyEnabled(string $property/*, string $resourceClass*/): bool
     {
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
meyerbaptiste: Your `sprintf()` has too many args.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -178,13 +219,43 @@ protected function getNestedMetadata(string $resourceClass, array $associations)
...
     {
         $parts = explode('.', $property);
 
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
meyerbaptiste: Your `sprintf()` has too many args.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -195,13 +266,25 @@ protected function splitPropertyParts(string $property): array
...
-    protected function extractProperties(Request $request): array
+    protected function extractProperties(Request $request/*, string $resourceClass*/): array
     {
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
meyerbaptiste: Your `sprintf()` has too many args.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -228,9 +311,21 @@ protected function extractProperties(Request $request): array
...
+    protected function addJoinsForNestedProperty(string $property, string $rootAlias, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator/*, string $resourceClass*/): array
     {
-        $propertyParts = $this->splitPropertyParts($property);
+        if (func_num_args() > 4) {
+            $resourceClass = func_get_arg(4);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', get_class($this), __FUNCTION__), E_USER_DEPRECATED);
```
meyerbaptiste: Your `sprintf()` has too many args.

jordscream:
@soyuka travis/appveyor failed is really linked to what i did ?

soyuka:
@jordscream yeah you added an `embeddedDummy` property I think. It fails on the following equality:

```json
          {
            "@id": "/related_dummies/1",
            "@type": "https://schema.org/Product",
            "id": 1,
            "name": "Hello",
            "symfony": "symfony",
            "dummyDate": null,
            "thirdLevel": "/third_levels/1",
            "relatedToDummyFriend": [],
            "dummyBoolean": null,
            "age": null
          },
```

```json
              {
                  "@id": "\/related_dummies\/1",
                  "@type": "https:\/\/schema.org\/Product",
                  "id": 1,
                  "name": "Hello",
                  "symfony": "symfony",
                  "dummyDate": null,
                  "thirdLevel": "\/third_levels\/1",
                  "relatedToDummyFriend": [],
                  "dummyBoolean": null,
                  "embeddedDummy": [],
                  "age": null
              },
```

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -136,9 +148,38 @@ protected function isPropertyMapped(string $property, string $resourceClass, boo
...
+    protected function isPropertyNested(string $property/*, string $resourceClass*/): bool
     {
-        return false !== strpos($property, '.');
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
```
teohhanhui: will have a ~~two~~ second argument

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -97,11 +97,23 @@ protected function getClassMetadata(string $resourceClass): ClassMetadata
...
-    protected function isPropertyEnabled(string $property): bool
+    protected function isPropertyEnabled(string $property/*, string $resourceClass*/): bool
     {
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
```
teohhanhui: will have a ~~two~~ second argument

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -178,13 +219,43 @@ protected function getNestedMetadata(string $resourceClass, array $associations)
...
     {
         $parts = explode('.', $property);
 
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
```
teohhanhui: will have a ~~two~~ second argument

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -228,9 +311,21 @@ protected function extractProperties(Request $request): array
...
+    protected function addJoinsForNestedProperty(string $property, string $rootAlias, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator/*, string $resourceClass*/): array
     {
-        $propertyParts = $this->splitPropertyParts($property);
+        if (func_num_args() > 4) {
+            $resourceClass = func_get_arg(4);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
```
teohhanhui: will have a ~~two~~ second argument

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -195,13 +266,25 @@ protected function splitPropertyParts(string $property): array
...
-    protected function extractProperties(Request $request): array
+    protected function extractProperties(Request $request/*, string $resourceClass*/): array
     {
+        if (func_num_args() > 1) {
+            $resourceClass = func_get_arg(1);
+        } else {
+            if (__CLASS__ !== get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a two `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
```
teohhanhui: will have a ~~two~~ second argument

tests/Fixtures/TestBundle/Entity/EmbeddableDummy.php
```diff
@@ -0,0 +1,173 @@
+<?php
+
+declare(strict_types=1);
+/*
```
teohhanhui: There should be a blank line here.

tests/Fixtures/TestBundle/Entity/EmbeddedDummy.php
```diff
@@ -0,0 +1,144 @@
+<?php
+
+declare(strict_types=1);
```
teohhanhui: There should be a blank line here.

blaues0cke:
Just FYI: Pointed a project I worked for over 4 months on to this pull request and everything (including the filters on embedded entities) worked like a charm without any error at all!

dunglas:
@jordscream do you think you can handle the last few comments?

jordscream:
@dunglas , yes will do this week :-)

soyuka:
@jordscream just so you know (and you don't force push :stuck_out_tongue:)  I fixed the last comments and rebased against master :).

features/main/crud.feature
```diff
@@ -92,7 +92,7 @@ Feature: Create-Retrieve-Update-Delete
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
     And the JSON should be equal to:
     """
-    {
+		{
```
meyerbaptiste: Indentation!

dunglas:
Thanks @jordscream and @soyuka!

features/bootstrap/FeatureContext.php
```diff
@@ -355,6 +421,66 @@ public function thereIsDummyObjectsWithDummyBoolean($nb, $bool)
     }
 
     /**
+     * @Given there is :nb embedded dummy objects with embeddedDummy.dummyBoolean :bool
+     */
+    public function thereIsDummyObjectsWithEmbeddedDummyBoolean($nb, $bool)
+    {
+        if (in_array($bool, ['true', '1', 1], true)) {
```
soyuka: This wasn't part of this PR or was it? I think there was a reason we didn't use that. I'll look into this.

meyerbaptiste: It was and I don't see any reason 😛

soyuka: Okay so did some digging. This actually comes from https://github.com/api-platform/core/pull/1010 and I used the same code as we already use in the `BooleanFilter` for consistency.

antograssiot: "on" / "off" support in boolean filter was removed. That's where this modification is comming from IIRC

features/main/crud.feature
soyuka: Thanks, rebasing behat features is a mess. Fixed in #1137

features/bootstrap/FeatureContext.php
soyuka: Note that I'm sure we talked about `FILTER_VALIDATE_BOOLEAN` when working on the boolean filter, IIRC it was with @teohhanhui.

jordscream:
@soyuka @dunglas sorry for my late
Cool to have merged :-)

features/bootstrap/FeatureContext.php
meyerbaptiste: Hmm, I don't see what can block here. We're not talking about a filter here but a setter. It's just a step definition with an argument which is converted to a pure boolean in all cases! So why couldn't we use `FILTER_VALIDATE_BOOLEAN`?🤔

soyuka: Sure, can add it to my todolist.

teohhanhui: Can we all stop using `filter_var`, please? :disappointed:

meyerbaptiste: Maybe with some explanations! 😛

PR #985, opened 2017-03-13
Fixed duplicated tag in ValidaterListener service definition

magarzon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

There is a duplicated tag in the definition of ValidationListener service, and so the listener is called twice in the kernel.view event


soyuka:
Good catch thx @magarzon

Issue #988, opened 2017-03-14
[Question] Related entities validation


kate-kate:
Hello. I have applied rules to denormalize embedded relation, like it was explained in docs. And it works well, except validation in related entity is totally ignored. It means, that if I provide fields of related entity, they actually will be saved, but if I don't provide fields with ```@Assert\NotBlank()``` annotations, everything will be saved as well (or cause MySQL violation). Can I force validation error in that case?

Also it isn't quite clear for me, how could I make some related entity mandatory. Should I mark relational fields with ```@Assert\NotBlank()```? Is there some recipe for that variant? 
(```@Assert\Valid()``` unfortunately doesn't work with groups and entity wouldn't be validated at all)

kate-kate:
I've tried different strategies to make it work the closest way to what I aimed to get. If I mark property with related object with ```@Assert\NotBlank()``` - it will shows that object as mandatory. And if I try to send embedded object without required fields - they will be also shown in validation errors. That's ok! But didn't see them in "Example Value" to post, and until I send any related data I wouldn't know, which fields are required. For example:

I have entity Job with embedded Campaign. "Example Value" I see is following:
```json
{
  "firstName": "string",
  "campaign": {}
}
```

While I expecting to retrieve something like:
```json
{
  "firstName": "string",
  "campaign": {
    "name": "string"
  }
}
```
because "name" is annotated as NotBlank and has the same validation group.

If I send only first name - I will see, that campaign should not be blank. If I send campaign without name - I finally will know, that I should have provided "campaign.name". But I would like to see that in "Example Value" or in the first validation

Is that possible?

Issue #986, opened 2017-03-14
Swagger documentation of related entities isn't proper


kate-kate:
Hello everyone. 
I have a Job entity with related Campaign. According to [https://api-platform.com/docs/core/serialization-groups-and-relations](url) I configured them like this:

```php
<?php

namespace AppBundle\Entity;

use Doctrine\ORM\Mapping as ORM;
use ApiPlatform\Core\Annotation\ApiResource;
use Symfony\Component\Serializer\Annotation\Groups;

/**
 * Job
 * @ApiResource(attributes={"normalization_context"={"groups"={"dashboard"}}})
 * @ORM\HasLifecycleCallbacks
 * @ORM\Table(name="job")
 * @ORM\Entity
 */
class Job
{
    /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer", nullable=false)
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="IDENTITY")
     */
    private $id;

    /**
     * @Groups({"dashboard"})
     * @ORM\Column(name="status", type="integer", nullable=false)
     */
    private $status;

    /**
     * @Groups({"dashboard"})
     * @ORM\Column(name="postcode", type="string", length=255, nullable=true)
     */
    private $postcode;

    /**
     * @Groups({"dashboard"})
     * @ORM\ManyToOne(targetEntity="Campaign")
     * @ORM\JoinColumns({
     *   @ORM\JoinColumn(name="campaign_id", referencedColumnName="id")
     * })
     */
    private $campaign;
 
    // ...
}
```

```php
<?php
namespace AppBundle\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Serializer\Annotation\Groups;

/**
 * Campaign
 * @ORM\HasLifecycleCallbacks
 * @ApiResource(attributes={"normalization_context"={"groups"={"dashboard"}}})
 * @ORM\Table(name="campaign")
 * @ORM\Entity
 */
class Campaign
{
    /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer", nullable=false)
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="IDENTITY")
     */
    private $id;

    /**
     * @var string
     * @ORM\Column(name="external_id", type="string", length=255, nullable=true)
     */
    private $externalId;

    /**
     * @var string
     * @Groups({"dashboard"})
[...]

Issue #940, opened 2017-02-10
[Swagger] Wrong serializing in docs


aistis-:
Hi,

I am using a default bundle of API platform v2 and run into small, but annoying docs problem.

I have two entities - `ExpenseCategory` which hold a reference to `Expense` under `expense` property. The problem is, that it is serialized incorrectly `ExpenseCategory` hold a reference to itself instead of `Expense`.

Any ideas? The config is more or less default - I have only registered `ExpenseCategory`

![screenshot](https://s30.postimg.org/d866l7vn5/API_Platform.png)
![screenshot](https://s29.postimg.org/6s4vvdz2f/API_Platform2.png)

Simperfit:
Could you please do the following ? 

- Fork the [API Platform Standard Edition](https://github.com/api-platform/api-platform)
- Do the minimum modification to reproduce the bug
- Post the link in here

We will test that and confirm.

PR #996, opened 2017-03-17
Changed Swagger Normalizer to correctly support nested serialization …

jonadoe:
…groups

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #986 #988 #940 
| License       | MIT
| Doc PR        | not needed

Fixed the bug from Issue #986 and also fixed general handling of serialization groups in swagger normalizer.

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\Api\OperationMethodResolverInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
 use ApiPlatform\Core\Documentation\Documentation;
+use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
@@ -99,7 +100,9 @@ final class DocumentationNormalizer implements NormalizerInterface
      */
     private function addPaths(\ArrayObject $paths, \ArrayObject $definitions, string $resourceClass, string $resourceShortName, ResourceMetadata $resourceMetadata, array $mimeTypes, bool $collection)
     {
-        if (null === $operations = $collection ? $resourceMetadata->getCollectionOperations() : $resourceMetadata->getItemOperations()) {
+        $operations = $collection ? $resourceMetadata->getCollectionOperations() : $resourceMetadata->getItemOperations();
+
+        if (!$operations) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -49,6 +49,10 @@
     private $filterCollection;
     private $nameConverter;
 
+    private $mimeTypes;
```
dunglas: Can you get rid of those states please? A service should not have states like those ones.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -67,36 +71,33 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
      */
     public function normalize($object, $format = null, array $context = [])
```
dunglas: All changes to this method must be reverted.

src/Swagger/Serializer/DocumentationNormalizer.php
jonadoe: The variables get reset on every call of the normalize function, so they are not really states.
The only other way to solve this is to pass through these variables (at least the $definitions) to nearly all other methods in the call. This makes the code much harder to read and does not make any sense at all.
If you like I can implement it that way, but I think it's better like it is now

dunglas: I would prefer to pass those variables as arguments like it was before.

src/Swagger/Serializer/DocumentationNormalizer.php
dunglas: WDYT @api-platform/core-team?

src/Swagger/Serializer/DocumentationNormalizer.php
jonadoe: Ok, I think this will make the code more complicated, but I will change it.

jonadoe:
@dunglas 
The requested changes are now implemented

jonadoe:
I'm currently short on time, but I will add a test and rebase the branch soon.

dunglas:
@jonadoe do you think that you'll be able to add a test soon? I plan to release a new bug fix version in the beginning of the next week.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -333,25 +341,24 @@ private function updateDeleteOperation(\ArrayObject $pathOperation, string $reso
...
         if (isset($serializerContext['groups'])) {
             $definitionKey = sprintf('%s_%s', $resourceMetadata->getShortName(), md5(serialize($serializerContext['groups'])));
         } else {
             $definitionKey = $resourceMetadata->getShortName();
         }
-        $definitions[$definitionKey] = $this->getDefinitionSchema($resourceClass, $resourceMetadata, $serializerContext, $definitionKey);
+
+        if (!isset($definitions[$definitionKey])) {
+            $definitions[$definitionKey] = [];  // Initialize first to prevent infinite loop
+            $definitions[$definitionKey] = $this->getDefinitionSchema($resourceClass, $resourceMetadata, $definitions, $serializerContext);
```
soyuka: This shouldn't be outside of the condition?

jonadoe:
@dunglas I will be on vacation for one week and can't do it before 24/25th, sorry

soyuka:
I'll add the test asap.

jonadoe:
Thank you @soyuka !!

dunglas:
Thank you very much @jonadoe and @soyuka as always :)

Issue #212, opened 2015-08-06
Let's talk about performances


dupuchba:
Hi everyone,

I was playing with the bundle and I did a backfire.io profiling which was very interesting. 
I just did a GET /poeple/3 API call with php 5.5.18 (no cache).

Note that Nelmio doc generation takes almost 30% of CPU time but I wasn't asking for it (we load it everytime).

I was curious **to see your point of view about that**.

Thank's everyone.

link here => https://blackfire.io/profiles/74457ce7-817a-4266-ad11-87c498d18762/graph

dunglas:
He heavy operation is the PHPDoc parsing but this not especially related to NelmioApiDoc. It occurs for populating metadata :

- property type guessing (using the property info library)
- textual description for the Hydra documentation (so used by NelmioApiDoc of installed)

This an expensive operation but in production you should use the cache (see the performance chapter in the doc).

dupuchba:
@dunglas I know (and I warned ^^) that caching metadata is a must have but I am still curious about nelmio optimization. 

- Do we really need it on each request ?

dunglas:
Have you enabled the Nelmio cache? Can you make another profile to show the difference when Nelmio is not installed?

dupuchba:
@dunglas I'll do it next week, like you I really need holidays ^_^

rolebi:
related to #198

rolebi:
related to #198 ?

theofidry:
@rolebi not really, #198 is really more on guidelines and doc. This case is an optimization of the bundle.

Simperfit:
What is the state of this issue ?

theofidry:
@Simperfit we need a good benchmark showing we can handle 2000 req/s just fine :P

teohhanhui:
Usually, serialization takes up the bulk of the time. If we have generated normalizers (symfony/symfony#17516), that'll be a huge win. Or perhaps it may be possible to introduce some cache layer to prevent re-generating the serialized string each time?

theofidry:
@teohhanhui make custom serializers and a property accessor for the benchmark :P

soyuka:
> Usually, serialization takes up the bulk of the time.

+1, for example (development mode): 

![screenshot-localhost-2016-11-09-14-25-43](https://cloud.githubusercontent.com/assets/1321971/20139854/74e0c532-a688-11e6-85a7-f89f72c94861.png)

soyuka:
[...]

PR #997, opened 2017-03-17
Cache identifiers for better performance on getIriFromItem

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | Not yet (still experimenting)
| Related tickets | #212
| License       | MIT
| Doc PR        | ø

## Add `getIriFromItem` cache

The collection here has ~8k items, no relations. Dev mode active but caching with apcu is forced.

[Blackfire before](https://blackfire.io/profiles/325404d0-c782-4f05-88ed-4676b6c3368b/graph):
- 2.87s spent in `getIriFromItem`
- 2.35 s for `getIdentifiersFromItem`

[Blackfire after](https://blackfire.io/profiles/7215dbe1-efb1-4835-aa8f-5296761c280a/graph):
- 898 ms spent in `getIriFromItem`
- 371 ms for `getIdentifiersFromItem`

When discussing my performance issue with @Simperfit we noticed that `getIdentifiersFromItem` was called a lot (previous method is `getIriFromItem`). This is called for every item in the collection.

### Profiler:

Don't look at the overall time, only the time spent serializing :p.

#### before
serializer3

#### after
cacheiri

In this PR, we cache the **properties** that are **identifiers** for the given class.


teohhanhui:
I like the idea, but the implementation is wrong. We should decorate the IriConverter.

soyuka:
👍 indeed

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -112,6 +118,34 @@ private function getIdentifiersFromItem($item): array
         $identifiers = [];
         $resourceClass = $this->getObjectClass($item);
 
+        $cacheKey = self::CACHE_KEY_PREFIX.md5(serialize([$resourceClass]));
```
dunglas: Why not using `$resourceClass` directly as it's a string?

soyuka: Indeed, copy paste you know :dancing_men: 

I made this in a hurry just to see if my first assumption was correct! I'll refactor this and decorate the `IriConverter` as suggested by @teohhanhui. Will be done this week!

soyuka:
Okay so I did this with a decorator, but because what's slow is `getIdentifiersFromItem` (which is a private method) I'm not sure it's ideal. Indeed, I had to copy the code from `IriConverter` (trying to avoid BC break at the same time), therefore I see no point in decorating :/.

You can check the code to understand what I mean... 

A solution would be to have a new service for `getIdentifiersFromItem` that we can cache without changing the IriConverter. Lmk if you've a better idea.

----- 

From my benchmarks, I know that we can still win a lot by caching the identifiers getters instead of calling the `PropertyAccessor` (`$identifiers[$propertyName] = $item->{$accessor}();`.

Some links (note that I'm using the normalizer dumper, 8000 items collection):

- [IriConverter without cache](https://blackfire.io/profiles/beab0865-7e50-40ed-823c-bc743e69edb3/graph) 8.36s
- [IriConverter with this PR cache](https://blackfire.io/profiles/51b859bd-956e-4f35-b5c5-79d3968717b7/graph) 6.89s
- [IriConverter with direct getter instead of PropertyAccessor + cache](https://blackfire.io/profiles/e9fda268-d964-4692-85cf-c85ed32391c9/graph) 4.78s

teohhanhui:
@soyuka Okay, since what we're caching is a private method, then decoration indeed makes no sense.

> A solution would be to have a new service for getIdentifiersFromItem that we can cache without changing the IriConverter.

`ItemIdentifiersResolverInterface`?

soyuka:
`ItemIdentifiersResolverInterface`? Extract is a bit strong no?

Would `get${ucfirst($propertyName)}` be acceptable to retrieve identifiers instead of using the property accessor?

teohhanhui:
I think "extract" describes better what we're doing here: extracting identifiers from an item.

> Would get${ucfirst($propertyName)} be acceptable to retrieve identifiers instead of using the property accessor?

No, we cannot make such assumptions. Plus, it'd be a BC break.

dunglas:
What we can do however is provide implementation of the `PropertyAccessorInterface` doing that. I usually do this for perf in my project. Such class should be included directly in Symfony.

soyuka:
@dunglas You mean having a kind of IdentifierPropertyAccessor witch would end up doing something simpler than the original class (still no assumption could be made regarding `get{$attribute}` ? We also don't need `setValue` in this context.

dunglas:
@soyuka yes, that you can register for your project if you stick to a particular convention (like public properties, or getters and setters). It prevents to trigger the introspection system, so it improves performances.

soyuka:
Oh I'm doing this already on my project for this particular class :). So, I'll just add a `ItemIdentifiersExtractorInterface` so that identifiers can be cached here but I won't touch the propertyAccessor thing.

soyuka:
ping @teohhanhui @dunglas 

I tried to avoid breaking changes lmk what you think. 
Note that Behat tests are green, just waiting for implementation approval to add tests (and #1014 comes first obviously).

src/Bridge/Symfony/Routing/ItemIdentifiersExtractor.php
```diff
@@ -0,0 +1,82 @@
...
+namespace ApiPlatform\Core\Bridge\Symfony\Routing;
+
+use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
+use ApiPlatform\Core\Util\ClassInfoTrait;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+
+class ItemIdentifiersExtractor implements ItemIdentifiersExtractorInterface
```
teohhanhui: Can be moved under `ApiPlatform\Core\Metadata\Property` namespace then?

src/Bridge/Symfony/Routing/CachedItemIdentifiersExtractor.php
```diff
@@ -0,0 +1,91 @@
...
+{
+    use ClassInfoTrait;
+
+    const CACHE_KEY_PREFIX = 'iri_converter';
+
+    private $cacheItemPool;
+    private $propertyAccessor;
+    private $decorated;
+
+    public function __construct(CacheItemPoolInterface $cacheItemPool, PropertyAccessorInterface $propertyAccessor = null, ItemIdentifiersExtractorInterface $decorated = null)
```
Simperfit: should $decorated be null ?

src/Bridge/Symfony/Routing/ItemIdentifiersExtractor.php
Simperfit: I agree

Simperfit:
👍  Great work @soyuka

src/Bridge/Symfony/Routing/ItemIdentifiersExtractor.php
dunglas: Indeed

src/Api/IdentifiersExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+            $relatedItem = $identifiers[$propertyName];
+
+            unset($identifiers[$propertyName]);
+
+            foreach ($this->propertyNameCollectionFactory->create($relatedResourceClass) as $relatedPropertyName) {
+                $propertyMetadata = $this->propertyMetadataFactory->create($relatedResourceClass, $relatedPropertyName);
+
+                if ($propertyMetadata->isIdentifier()) {
+                    if (isset($identifiers[$propertyName])) {
+                        throw new RuntimeException(sprintf('Composite identifiers not supported in "%s" through relation "%s" of "%s" used as identifier', $relatedResourceClass, $propertyName, $resourceClass));
```
soyuka: @teohhanhui I'm having a hard time trying to add test coverage on this one. IMO this is not a valid case, but maybe you can give me some more context/informations about the Exception? [Origin commit you authored](https://github.com/api-platform/core/commit/6117fe9786819f6b35ce0c9371bb30ab20a2637f).

soyuka:
I did a great job covering this part as it wasn't tested anywhere (except through IriConverter, but not fully covered). 

There's still a point to discuss: the location of the `IdentifiersExtractor`. 

I'm not sure that it should actually be under `ApiPlatform\Core\Metadata` because it doesn't interfere with metadata, rather uses the metadata to extract identifiers. Therefore, IMHO this class should be in `ApiPlatform\Core\Api` (so is `ResourceClassResolver` for example). Let me know if you disagree we can always move it elsewhere :).

soyuka:
All green ping @api-platform/core-team

Simperfit:
@soyuka could you please update with master ?

src/Api/CachedIdentifiersExtractor.php
```diff
@@ -0,0 +1,101 @@
...
+    const CACHE_KEY_PREFIX = 'iri_identifiers';
+
+    private $cacheItemPool;
+    private $propertyAccessor;
+    private $decorated;
+
+    public function __construct(CacheItemPoolInterface $cacheItemPool, PropertyAccessorInterface $propertyAccessor = null, IdentifiersExtractorInterface $decorated)
+    {
+        $this->cacheItemPool = $cacheItemPool;
+        $this->propertyAccessor = $propertyAccessor ?: PropertyAccess::createPropertyAccessor();
```
dunglas: `$propertyAccessor ?? PropertyAccess::createPropertyAccessor();`?

src/Api/IdentifiersExtractorInterface.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
```
dunglas: Missing blank line

src/Api/IdentifiersExtractorInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ApiPlatform\Core\Api;
+
+/**
+ * Extracts identifiers for a given Resource according to the retrieved Metadata.
+
```
dunglas: Line to remove

src/Api/IdentifiersExtractorInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+interface IdentifiersExtractorInterface
+{
+    /**
+     * Find identifiers from an Item (Object).
```
dunglas: Finds

src/Exception/CacheException.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
```
dunglas: Blank line

tests/Api/CachedIdentifiersExtractorTest.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+declare(strict_types=1);
```
dunglas: Blank line

src/Api/CachedIdentifiersExtractor.php
```diff
@@ -0,0 +1,101 @@
...
+                        continue;
+                    }
+
+                    $relatedItem = $identifiers[$propertyName];
+                    $relatedCacheKey = self::CACHE_KEY_PREFIX.md5($this->getObjectClass($relatedItem));
+
+                    $relatedCacheItem = $this->cacheItemPool->getItem($relatedCacheKey);
+
+                    if (!$relatedCacheItem->isHit()) {
+                        throw new CacheException('No relation cache item founded, we need more cache to continue.');
```
dunglas: Can't we avoid to use exceptions? Exception handling is slow.

tests/Api/IdentifiersExtractorTest.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+declare(strict_types=1);
```
dunglas: blank line

soyuka:
@dunglas this should be good now!

src/Api/IdentifiersExtractor.php
```diff
@@ -0,0 +1,89 @@
...
+
+    private $propertyNameCollectionFactory;
+    private $propertyMetadataFactory;
+    private $propertyAccessor;
+
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, PropertyAccessorInterface $propertyAccessor = null)
+    {
+        $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
+        $this->propertyMetadataFactory = $propertyMetadataFactory;
+        $this->propertyAccessor = $propertyAccessor ?: PropertyAccess::createPropertyAccessor();
```
dunglas: `??`

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -42,15 +43,23 @@
...
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null, IdentifiersExtractorInterface $identifiersExtractor = null)
     {
         $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
         $this->propertyMetadataFactory = $propertyMetadataFactory;
         $this->itemDataProvider = $itemDataProvider;
         $this->routeNameResolver = $routeNameResolver;
         $this->router = $router;
         $this->propertyAccessor = $propertyAccessor ?: PropertyAccess::createPropertyAccessor();
```
dunglas: Can you change this one too? `??`

PR #993, opened 2017-03-15
Deprecate xml operations configuration and provide new way to make it…

sawmurai:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/252
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/185

PR #1000, opened 2017-03-20
Attempt to fix tests (segfault php 7)

soyuka:
Trying the dunglas magic thing on deps=low

PR #1000 is for this 😢


teohhanhui:
Shouldn't we use 7.1.3 now?

soyuka:
See https://github.com/api-platform/core/pull/993#issuecomment-287870561

dunglas:
Thanks for this @soyuka

PR #1007, opened 2017-03-24
Make the CS stricter; fix some bugs / issues

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Stricter CS = better code, especially `declare(strict_types=1);`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -16,7 +16,7 @@ return PhpCsFixer\Config::create()
         'braces' => [
             'allow_single_line_closure' => true,
         ],
-        // 'declare_strict_types' => true,
+        'declare_strict_types' => true,
         'modernize_types_casting' => true,
         // 'native_function_invocation' => true,
         'no_extra_consecutive_blank_lines' => [
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -6,11 +6,43 @@ $finder = PhpCsFixer\Finder::create()
 ;
 
 return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
     ->setRules([
         '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => [
+            'syntax' => 'short',
+        ],
+        'braces' => [
+            'allow_single_line_closure' => true,
+        ],
+        // 'declare_strict_types' => true,
+        'modernize_types_casting' => true,
+        // 'native_function_invocation' => true,
+        'no_extra_consecutive_blank_lines' => [
+            'break',
+            'continue',
+            'curly_brace_block',
+            'extra',
+            'parenthesis_brace_block',
+            'return',
+            'square_brace_block',
+            'throw',
+            'use',
+        ],
+        'no_unreachable_default_argument_value' => true,
+        'no_useless_else' => true,
+        'no_useless_return' => true,
         'ordered_imports' => true,
+        // 'phpdoc_add_missing_param_annotation' => [
+        //     'only_untyped' => false,
+        // ],
         'phpdoc_order' => true,
-        'array_syntax' => array('syntax' => 'short'),
+        'psr4' => true,
+        'semicolon_after_instruction' => true,
+        'strict_comparison' => true,
+        'strict_param' => true,
+        'ternary_to_null_coalescing' => true,
     ])
     ->setFinder($finder)
 ;
--- a/src/Bridge/Symfony/Routing/RouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/RouteNameResolver.php
@@ -9,6 +9,8 @@
  * file that was distributed with this source code.
  */
 
+declare(strict_types=1);
+
 namespace ApiPlatform\Core\Bridge\Symfony\Routing;
 
 use ApiPlatform\Core\Exception\InvalidArgumentException;
--- a/src/Bridge/Symfony/Routing/RouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/RouteNameResolver.php
@@ -40,7 +40,7 @@ final class RouteNameResolver implements RouteNameResolverInterface
             $operation = $route->getDefault(sprintf('_api_%s_operation_name', $operationType));
             $methods = $route->getMethods();
 
-            if ($resourceClass === $currentResourceClass && null !== $operation && (empty($methods) || in_array('GET', $methods))) {
+            if ($resourceClass === $currentResourceClass && null !== $operation && (empty($methods) || in_array('GET', $methods, true))) {
                 return $routeName;
             }
         }
--- a/tests/Bridge/Doctrine/Orm/Filter/DateFilterTest.php
+++ b/tests/Bridge/Doctrine/Orm/Filter/DateFilterTest.php
@@ -9,6 +9,8 @@
  * file that was distributed with this source code.
  */
 
+declare(strict_types=1);
+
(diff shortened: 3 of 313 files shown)


tests/Bridge/Doctrine/Orm/Filter/DateFilterTest.php
```diff
@@ -124,14 +126,14 @@ public function provideApplyTestData(): array
             ],
             'after' => [
                 [
-                    'dummyDate' => true,
+                    'dummyDate' => null,
```
soyuka: why the change here?

.php_cs.dist
```diff
@@ -6,11 +6,43 @@ $finder = PhpCsFixer\Finder::create()
...
+            'throw',
+            'use',
+        ],
+        'no_unreachable_default_argument_value' => true,
+        'no_useless_else' => true,
+        'no_useless_return' => true,
         'ordered_imports' => true,
+        // 'phpdoc_add_missing_param_annotation' => [
+        //     'only_untyped' => false,
+        // ],
```
soyuka: should we remove those?

tests/Bridge/Doctrine/Orm/Filter/DateFilterTest.php
teohhanhui: Because this is `$nullManagement`, which can only be `string` or `null`, not `bool`.

.php_cs.dist
teohhanhui: I plan to get these in next... :smile:

.php_cs.dist
```diff
@@ -6,11 +6,43 @@ $finder = PhpCsFixer\Finder::create()
...
+            'syntax' => 'short',
+        ],
+        'braces' => [
+            'allow_single_line_closure' => true,
+        ],
+        'declare_strict_types' => true,
+        'modernize_types_casting' => true,
+        // 'native_function_invocation' => true,
+        'no_extra_consecutive_blank_lines' => [
+            'break',
```
teohhanhui: This one may be a bit controversial, but I think it's good that it encourages us to not have complex logic nested inside a switch-case, or even avoid switch-case altogether.

dunglas: I agree with the intent, but it makes the code (even simple statements) less readable (IMO).

src/Bridge/Symfony/Routing/RouteNameResolver.php
```diff
@@ -40,7 +42,7 @@ public function getRouteName(string $resourceClass, bool $collection): string
             $operation = $route->getDefault(sprintf('_api_%s_operation_name', $operationType));
             $methods = $route->getMethods();
 
-            if ($resourceClass === $currentResourceClass && null !== $operation && (empty($methods) || in_array('GET', $methods))) {
+            if ($resourceClass === $currentResourceClass && null !== $operation && (empty($methods) || in_array('GET', $methods, true))) {
```
dunglas: The call to `empty` can be replaced by the `!` operator.

dunglas:
Awesome. Can you just fix StyleCI ? Putting the `declare` line before the license header should do the trick.

teohhanhui:
> Can you just fix StyleCI ? Putting the declare line before the license header should do the trick.

No. StyleCI is buggy here. See FriendsOfPHP/PHP-CS-Fixer#2590

Also, declare should go after file-level docblock, according to PSR-12 draft.

.php_cs.dist
teohhanhui: Yeah, personally I prefer the blank line too, but for the sake of consistency...

src/Bridge/Symfony/Routing/RouteNameResolver.php
teohhanhui: I really don't like such implicit conversions... And I don't like `empty` too...

My preference would be `0 === count($methods)`, but I understand there is a  performance overhead (how much?) for the function call.

dunglas:
So what do we do in the meantime? Disable StyleCI? Or disable the header check?

src/Bridge/Symfony/Routing/RouteNameResolver.php
dunglas: I prefer to keep in sync with what we do in Symfony (we remove `empty`, `count` and so on).

.php_cs.dist
dunglas: Consistency with what?

teohhanhui:
Disabling the header check sounds reasonable. But it might escape our watchful eyes. :smile:

dunglas:
We'll restore the check when PHP CS Fixer will be updated.

.php_cs.dist
teohhanhui: We cannot enforce a blank line before/after `break;`, but we can enforce `break;` always sticking to the next `case` statement:

```php
switch ($str) {
    case 'A':
        $this->doSomething();

        $this->doSomethingElse();

        break;
    case 'B':
        $this->doSomethingAwesome();
        break;
}
```

dunglas: But we can ignore this rule too and check it manually (it's not a big deal after all).

teohhanhui: Do you need me to revert this? I'll have to do it manually...

dunglas: Don't worry about that, keep it as is.

dunglas:
StyleCI config updated. Can you rebase?

.php_cs.dist
teohhanhui: @dunglas I want to get this in too... But this is gonna require quite a bit of manual work. :laughing:

.php_cs.dist
```diff
@@ -6,11 +6,43 @@ $finder = PhpCsFixer\Finder::create()
...
+        '@Symfony:risky' => true,
+        'array_syntax' => [
+            'syntax' => 'short',
+        ],
+        'braces' => [
+            'allow_single_line_closure' => true,
+        ],
+        'declare_strict_types' => true,
+        'modernize_types_casting' => true,
+        // 'native_function_invocation' => true,
```
teohhanhui: What do you guys think about this? I remember @dunglas talking about this before... We'll need to do Blackfire profiling for this to see the performance difference.

teohhanhui:
Rebased. But StyleCI still has a problem with `return null;` when Symfony has already changed its CS in view of the introduction of the `void` return type. I suspect they are using outdated PHP-CS-Fixer? :disappointed:

teohhanhui:
It could make sense for us to disable StyleCI (at least for now?), because frankly I don't see what's the added value of having it.

dunglas:
Are you sure Symfony has changed this CS?

dunglas:
Anyway StyleCI is useless since we run our own PHP CS Fixer instance. I've just disabled it. Can you rebase again?

soyuka:
Styleci would make sense if it prevented the other CI from running on failure.

teohhanhui:
Yes, since symfony/symfony-docs#6614

> * Use `return null;` when a function explicitly returns `null` values and use `return;` when the function returns `void` values;
>
> http://symfony.com/doc/current/contributing/code/standards.html

teohhanhui:
The only merit I see for StyleCI is the nice output, and the ability to send a PR. :laughing:

dunglas:
Thank you @teohhanhui

PR #1009, opened 2017-03-24
Avoid to truncate WHERE clauses when clonning the querybuilder

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This PR aims to fix a bug introduced by #950 
Actually when the querybuilder is cloned, some where parts are stripped.

The issue comes from 

```php
    // Change where aliases
    foreach ($wherePart->getParts() as $where) {
        $queryBuilderClone->add('where', str_replace($aliases, $replacements, $where));
    }
```
because the `add()` function doesn't append.

There's probably a cleaner way of fixing this, please let me know and i'll update


antograssiot:
Closing in favor of #1010

PR #1010, opened 2017-03-24
Fix #1008 multiple filters on eager loading

### LINKED ISSUE #1008: QueryException with custom data provider and filters
After updating the api-platform/core from version 2.0.3 to 2.0.4 I get the "tooManyParameters" QueryException _(Too many parameters: the query defines 1 parameters and you bound 2_)

I have an custom collection data provider with the following code:

```
public function getCollection(string $resourceClass, string $operationName = null)
{
    ....
    $alias = 'o';
    $queryBuilder = $repository->createQueryBuilder($alias);
    $queryNameGenerator = new QueryNameGenerator();

    $this->applyVerifiedFilter($alias, $queryBuilder, $queryNameGenerator);

    foreach ($this->collectionExtensions as $extension) {
        $extension->applyToCollection($queryBuilder, $queryNameGenerator, $resourceClass, $operationName);

        if ($extension instanceof QueryResultCollectionExtensionInterface
            && $extension->supportsResult($resourceClass, $operationName)
        ) {
            return $extension->getResult($queryBuilder);
        }
    }

    return $queryBuilder->getQuery()->getResult();
}

private function applyVerifiedFilter(string $alias, QueryBuilder $queryBuilder, QueryNameGenerator $queryNameGenerator)
{
    $field = 'verified';
    $valueParameter = $queryNameGenerator->generateParameterName($field);

    $queryBuilder
        ->andWhere(sprintf('%s.%s = :%s', $alias, $field, $valueParameter))
        ->setParameter($valueParameter, true)
    ;
}
```

For the collection I've defined some additional filters that can be applied as parameters:

```
generic.search_filter:
    parent: 'api_platform.doctrine.orm.search_filter'
    arguments: [ { id: 'exact' } ]
    tags:      [ { name: 'api_platform.filter', id: 'generic.search' } ]
```

After the api-platform update the Doctrine Query now got a mismatch when checking `if ($paramCount > $mappingCount) {`

When checking the DQL now only the id filter is applied and not the verified filter.

api-platform 2.0.3:
`WHERE o.verified = :verified_p1 AND o.id = :id_p2`

api-platform 2.0.4:
`WHERE o_2.id = :id_p2`

I don't know why this is happening after the update. When reverting to 2.0.3 it works fine.

The code in the doctrine query class:
```
$paramMappings = $this->_parserResult->getParameterMappings();
$paramCount = count($this->parameters);
$mappingCount = count($paramMappings);
```

In both versions (2.0.3 and 2.0.4) $this->parameters has two elements but $paramMappings has only one item in 2.0.4 but two items in 2.0.3

Does anyone know why this is happening and how to solve this issue?

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1008
| License       | MIT
| Doc PR        | na

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -189,6 +189,42 @@ class FeatureContext implements Context, SnippetAcceptingContext
     }
 
     /**
+     * @Given there is :nb dummy objects with dummyDate and dummyBoolean :bool
+     */
+    public function thereIsDummyObjectsWithDummyDateAndDummyBoolean($nb, $bool)
+    {
+        $descriptions = ['Smart dummy.', 'Not so smart dummy.'];
+
+        if (in_array($bool, ['true', '1', 1], true)) {
+            $bool = true;
+        } elseif (in_array($bool, ['false', '0', 0], true)) {
+            $bool = false;
+        } else {
+            $expected = ['true', 'false', '1', '0'];
+            throw new InvalidArgumentException(sprintf('Invalid boolean value for "%s" property, expected one of ( "%s" )', $bool, implode('" | "', $expected)));
+        }
+
+        for ($i = 1; $i <= $nb; ++$i) {
+            $date = new \DateTime(sprintf('2015-04-%d', $i), new \DateTimeZone('UTC'));
+
+            $dummy = new Dummy();
+            $dummy->setName('Dummy #'.$i);
+            $dummy->setAlias('Alias #'.($nb - $i));
+            $dummy->setDescription($descriptions[($i - 1) % 2]);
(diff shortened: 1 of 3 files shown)


soyuka:
@bschultz @antograssiot may you try this patch? I added a behat test (with date filter that does an `orWhere` and a boolean filter `andWhere`) so this should be good to go, just want to be sure :). 

Thanks!

bschultz:
@soyuka Yes with this patch it works :) 👍

dunglas:
Thanks for the fix!

PR #1013, opened 2017-03-27
Fix the generated XML Schema type for float

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

`xmls:number` doesn't even exist...

### DIFF
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -303,7 +303,7 @@ final class DocumentationNormalizer implements NormalizerInterface
                 return 'xmls:integer';
 
             case Type::BUILTIN_TYPE_FLOAT:
-                return 'xmls:number';
+                return 'xmls:decimal';
 
             case Type::BUILTIN_TYPE_BOOL:
                 return 'xmls:boolean';


teohhanhui:
Can we stick to the usually used prefix:

```json
"xsd": "http://www.w3.org/2001/XMLSchema#"
```

instead of `xmls`

dunglas:
It would be a BC break...

teohhanhui:
Why would it be a BC break? If a client uses string matching to identify type, that's doing it wrong in the first place. Changing the prefix should not break anything. (Okay, I'm not sure about this edge case, but: unless someone added in their own `xsd` prefix too?)

dunglas:
Most parsers don't expand the JSON-LD and just do something like:

`fetch('/docs.jsonld').then(response => reponse.json()).then(json => console.log(json['my:key']));`

Such parsers will fail.

teohhanhui:
Non-standards compliant parsing... If that's considered a BC break, then we're in trouble. :disappointed:

dunglas:
We may not be strict about this, however this change has no real benefits but can break things.

teohhanhui:
Okay, let's keep it for when we can actually break things again.

PR #1014, opened 2017-03-28
Normalize identifiers to the correct type

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Improve compatibility with `declare(strict_types=1);`

Should be backported to `2.0`

### DIFF
--- a/src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
+++ b/src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
@@ -13,6 +13,8 @@ namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Util;
 
 use ApiPlatform\Core\Exception\PropertyNotFoundException;
 use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\DBAL\Types\Type as DBALType;
+use Doctrine\ORM\EntityManagerInterface;
 
 /**
  * @internal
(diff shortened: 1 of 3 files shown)


src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -33,19 +35,21 @@
...
      * @return array
      */
-    public function normalizeIdentifiers($id, ObjectManager $manager, string $resourceClass): array
+    private function normalizeIdentifiers($id, ObjectManager $manager, string $resourceClass): array
     {
         $identifierValues = [$id];
-        $doctrineMetadataIdentifier = $manager->getClassMetadata($resourceClass)->getIdentifier();
+        $doctrineClassMetadata = $manager->getClassMetadata($resourceClass);
+        $doctrineIdentifierFields = $doctrineClassMetadata->getIdentifier();
+        $isOrm = interface_exists(EntityManagerInterface::class) && $manager instanceof EntityManagerInterface;
```
dunglas: The call to `interface_exists` is useless.

teohhanhui: I think PHPStan will complain:

> - Existence of classes and interfaces in `instanceof`, `catch`, typehints, other language constructs and even annotations. PHP does not do this and just stays silent instead.

soyuka: Can't we skip the check?

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -33,19 +35,21 @@
...
-            $identifiers = explode(';', $id);
+        if (count($doctrineIdentifierFields) > 1) {
             $identifiersMap = [];
 
             // first transform identifiers to a proper key/value array
-            foreach ($identifiers as $identifier) {
-                $keyValue = explode('=', $identifier);
-                $identifiersMap[$keyValue[0]] = $keyValue[1];
+            foreach (explode(';', $id) as $identifier) {
+                $identifierPair = explode('=', $identifier);
```
soyuka: nice name!

dunglas:
@teohhanhui can you merge it in 2.0?

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
teohhanhui: I think PHPStan is trying to help us avoid typos and missing imports.

dunglas:
@teohhanhui do you want me to do the merge or can you handle it?

dunglas:
After double thinking about it, as it's a large patch and this problem doesn't impact users directly, I propose to merge it only in 2.0. WDYT?

soyuka:
@dunglas we can't merge this in 2.0 as the IdentifierManagerTrait was a feature introduced recently in master ;)

soyuka:
thanks @teohhanhui

teohhanhui:
This needs to be backported to 2.0

PR #1022, opened 2017-03-30
Fix #1021 order by hidden eager loading

### LINKED ISSUE #1021: FilterEagerLoadingExtension messes up `ORDER BY` a `HIDDEN` field
FilterEagerLoadingExtension does not handle `ORDER BY` a `HIDDEN` field properly.

Before:
```dql
SELECT o, CASE WHEN o.dateCreated IS NULL THEN 0 ELSE 1 END AS HIDDEN _o_dateCreated_null_rank
FROM AppBundle\Entity\Tour o
INNER JOIN o.offers offers_a1
INNER JOIN offers_a1.seller seller_a2
WHERE seller_a2.id = :seller_p1 AND o.privacy IN (:privacy_p2)
ORDER BY _o_dateCreated_null_rank DESC, o.dateCreated DESC
```

After:
```dql
SELECT o, CASE WHEN o.dateCreated IS NULL THEN 0 ELSE 1 END AS HIDDEN _o_dateCreated_null_rank
FROM AppBundle\Entity\Tour o
INNER JOIN o.offers offers_a1
INNER JOIN offers_a1.seller seller_a2
WHERE o IN(
    SELECT o_2
    FROM AppBundle\Entity\Tour o_2
    INNER JOIN o_2.offers offers_a1_a3
    INNER JOIN offers_a1_a3.seller seller_a2_a4
    WHERE seller_a2_a4.id = :seller_p1
        AND o_2.privacy IN (:privacy_p2)
    ORDER BY _o_dateCreated_null_rank DESC, o.dateCreated DESC
)
ORDER BY _o_dateCreated_null_rank DESC, o.dateCreated DESC
```

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1021
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -80,6 +80,9 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
         //reset parts
         $queryBuilderClone->resetDQLPart('join');
         $queryBuilderClone->resetDQLPart('where');
+        $queryBuilderClone->resetDQLPart('orderBy');
+        $queryBuilderClone->resetDQLPart('groupBy');
+        $queryBuilderClone->resetDQLPart('having');
 
         //Change from alias
         $from = $queryBuilderClone->getDQLPart('from')[0];
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -48,14 +48,31 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
         }
 
         $joinParts = $queryBuilder->getDQLPart('join');
+        $originAlias = 'o';
 
-        if (!$joinParts || !isset($joinParts['o'])) {
+        if (!$joinParts || !isset($joinParts[$originAlias])) {
             return;
         }
 
         $queryBuilderClone = clone $queryBuilder;
         $queryBuilderClone->resetDQLPart('where');
-        $queryBuilderClone->andWhere($queryBuilderClone->expr()->in('o', $this->getQueryBuilderWithNewAliases($queryBuilder, $queryNameGenerator)->getDQL()));
+
+        $classMetadata = $queryBuilder->getEntityManager()->getClassMetadata($resourceClass);
+
+        if (!$classMetadata->isIdentifierComposite) {
+            $replacementAlias = $queryNameGenerator->generateJoinAlias($originAlias);
+            $in = $this->getQueryBuilderWithNewAliases($queryBuilder, $queryNameGenerator, $originAlias, $replacementAlias);
+            $in->select($replacementAlias);
+            $queryBuilderClone->andWhere($queryBuilderClone->expr()->in($originAlias, $in->getDQL()));
+        } else {
+            // Because Doctrine doesn't support WHERE ( foo, bar ) IN () (https://github.com/doctrine/doctrine2/issues/5238), we are building as many subqueries as they are identifiers
+            foreach ($classMetadata->identifier as $identifier) {
+                $replacementAlias = $queryNameGenerator->generateJoinAlias($originAlias);
+                $in = $this->getQueryBuilderWithNewAliases($queryBuilder, $queryNameGenerator, $originAlias, $replacementAlias);
+                $in->select("IDENTITY($replacementAlias.$identifier)");
+                $queryBuilderClone->andWhere($queryBuilderClone->expr()->in("$originAlias.$identifier", $in->getDQL()));
+            }
+        }
 
         $queryBuilder->resetDQLPart('where');
         $queryBuilder->add('where', $queryBuilderClone->getDQLPart('where'));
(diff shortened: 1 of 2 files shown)


teohhanhui:
I'm curious about `GROUP BY` and `HAVING` clauses too. We should add tests for them.

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
```diff
@@ -78,6 +78,7 @@ private function getQueryBuilderWithNewAliases(QueryBuilder $queryBuilder, Query
         $wherePart = $queryBuilder->getDQLPart('where');
 
         //reset parts
+        $queryBuilderClone->resetDQLPart('orderBy');
```
teohhanhui: Logically the sequence should be: join - where - orderBy \o/

soyuka:
Your right, I'll add those in ~30min ;)

soyuka:
W8 do not merge I'm adding a fix for #1023

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
```diff
@@ -123,10 +126,150 @@ public function testApplyCollection()
...
+    INNER JOIN o_2.compositeItem item_2 
+    INNER JOIN o_2.compositeLabel label_2 
+    WHERE item_2.field1 = :foo
+) AND o.label IN(
+    SELECT IDENTITY(o_2.label) FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\CompositeRelation o_2 
+    INNER JOIN o_2.compositeItem item_2 
+    INNER JOIN o_2.compositeLabel label_2 
+    WHERE item_2.field1 = :foo
+)
+SQL;
```
soyuka: WDYT @teohhanhui ? A better fix would be to fix doctrine but this works fine as a temporary fix (will take a long time to get that fixed in doctrine :p). Indeed, it should be `WHERE (o.item, o.label) IN (subquery)`.

dunglas: Maybe can you add a comment with this rationale? We'll remove this temporary when (if) the patch will be merged in Doctrine.

dunglas:
Feel free to merge this one @soyuka. I'll tag a new patch release after that.

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
teohhanhui: I didn't even know there's such a syntax in SQL. Obviously I don't really use composite keys :stuck_out_tongue:

Anyway, I think this is fine.

teohhanhui: Oh, but I think the tests are confusing, because there's no eager loading going on (how do we tell?). If so, we can skip query rewriting in such cases, right?

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
```diff
@@ -123,10 +126,150 @@ public function testApplyCollection()
...
+
+        $this->assertEquals($this->toDQLString($expected), $qb->getDQL());
+    }
+
+    private function toDQLString(string $dql): string
+    {
+        return preg_replace('/\\r\\n|\\n/', '', str_replace('  ', '', $dql));
```
teohhanhui: If we improve the regex then there'd be no need for the `str_replace` :smile:

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
soyuka: The fix will get applied only if the conditions are a match (where clause + eager + composite identifier in this case). There will be no rewriting if those conditions don't match.
I force those conditions with the test.

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
soyuka: haha yeah appveyor behavior was driving me crazy, intent was first to use `PHP_EOL` then I ended up with the regexp. Anyway, no big deal this is just to make tests more readable.

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
teohhanhui: What I mean is that even though `$forceEager` is true, the query itself might not need to be rewritten if there is no eager loading going on, no? (Not specific to this test case.)

soyuka: Eager loading is active when there are joins [this condition](https://github.com/soyuka/core/blob/7a5903572c29370043cdf1f8979724a30c130829/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L53) checks that. The rewrite is needed only when there is a [where clause](https://github.com/soyuka/core/blob/7a5903572c29370043cdf1f8979724a30c130829/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L46).

teohhanhui: But it wouldn't run till there, because of https://github.com/api-platform/core/blob/7a5903572c29370043cdf1f8979724a30c130829/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L39-L41

PR #1028, opened 2017-04-03
Hotfix/filter eager loading fetch eager

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

ping @teohhanhui

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -13,6 +13,8 @@ namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Extension;
 
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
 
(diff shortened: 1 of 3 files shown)


tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
```diff
@@ -270,6 +347,6 @@ public function testCompositeIdentifiers()
 
     private function toDQLString(string $dql): string
     {
-        return preg_replace('/\\r\\n|\\n/', '', str_replace('  ', '', $dql));
+        return preg_replace(['/\s+/', '/\(\s/', '/\s\)/'], [' ', '(', ')'], $dql);
```
soyuka: This should be better \o/

teohhanhui: Actually this is unsafe... It might change whitespace within a SQL literal (string). Lol...

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
```diff
@@ -143,4 +146,27 @@ private function isForceEager(string $resourceClass, array $options): bool
 
         return is_bool($forceEager) ? $forceEager : $this->forceEager;
     }
+
+    private function hasFetchEagerAssociation(EntityManager $em, ClassMetadataInfo $classMetadata, &$checked = [])
```
teohhanhui: Is it possible to extract this logic out for reuse between `EagerLoadingExtension` and `FilterEagerLoadingExtension`?

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
soyuka: Of course, as said before it makes tests more readable :).

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
soyuka: I don't want to to this to avoid a 2-pass recursive call in `EagerLoadingExtension`. If we want to extract this logic we need to add cache. I obviously agree that we should add a new class with the `shouldEagerLoad` logic.

dunglas: Can't this be extracted in another PR?

soyuka: Yeah it can and should, +1 to do this in another PR as it's not a priority.

soyuka:
Feel free to merge I have a note that I should refactor the `shouldForceEager` thing.

dunglas:
Thank you @soyuka

Issue #1026, opened 2017-04-02
Impossible to set identifier on upper cased attribute


insekticid:
Upper Case attribute name example, that doesn't work:
Btw, there is no mention about `@ApiProperty(identifier=true)` in docs :/ ! 

```
    /**
     * @ApiProperty(identifier=true)
     */
    protected $Id;
```
Resource:
```
     AppBundle\Entity\Project:
         itemOperations:
                get:
                    method: 'GET'
                    path: '/project/{id}'
```
Request
curl -X GET --header 'Accept: application/ld+json' 'http://127.0.0.1:81/project/48'
Response:
```
{
  "@context": "/contexts/Error",
  "@type": "hydra:Error",
  "hydra:title": "An error occurred",
  "hydra:description": "Parameter \"id\" for route \"api_projects_get_item\" must match \"[^/]++\" (\"\" given) to generate a corresponding URL.",
"trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
      "file": "/srv/api-platform/vendor/symfony/symfony/src/Symfony/Component/Routing/Generator/UrlGenerator.php",
      "line": 163,
```

temporary solution is to define lower case attribute :
```
    /**
     * @var int
     */
    protected $Id = null;

    /**
     * @ApiProperty(identifier=true)
     * @var int
     */
    protected $id = null;
```

Why I had to use upper case name? Because of bad SOAP service naming convention from vendor

What about to define identifier in resource.yml?
```
AppBundle\Entity\Project:
         identifier: id
         itemOperations:
                get:
                    method: 'GET'
                    path: '/project/{id}'
```

teohhanhui:
I'd consider this a bug.

meyerbaptiste:
This problem comes from Symfony and its PropertyInfo component. You can see it in the [`ReflectionExtractor::getProperties()` method, L86-88](https://github.com/symfony/symfony/blob/313fec952f2593509c65a74410c54fcc2eb2f972/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php#L86-L88).

To fix this problem you can change the visibility of your `Id` attribute from `protected` to `public`.
Another solution is to implement your own extractor class for your specific use case.

You should report it using the [Symfony bug tracker](https://github.com/symfony/symfony/issues).

teohhanhui:
[...]

PR #1030, opened 2017-04-04
Allow Upper Case property names

insekticid:
ReflectionExtractor::getProperties() returns $id instead of $Id. It is bad naming convention, but is possible

```
class Entity {
    protected $Id;

    public function getId()
    {
        return $this->Id;
    }
}
```

https://github.com/symfony/symfony/pull/22265

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1026
| License       | MIT


dunglas:
Can you cover this with an unit test?

meyerbaptiste:
Is this PR really useful? https://github.com/api-platform/core/issues/1026#issuecomment-291451046

PR #1034, opened 2017-04-06
Skip colors on travis job

soyuka:
They disabled logs when nb_lines > 10000, which is our case, this helps
debugging

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes \o/
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,11 +31,11 @@ install:
   - if [[ $deps = 'low' ]]; then composer update --prefer-dist --no-progress --no-suggest --prefer-stable --prefer-lowest --ansi; fi
 
 script:
-  - if [[ $coverage = 1 ]]; then phpdbg -qrr vendor/bin/phpunit --coverage-php build/cov/coverage-phpunit.cov; else vendor/bin/phpunit; fi
-  - if [[ $coverage = 1 ]]; then for feature in $(ls -d features/*/ | cut -f2 -d'/' | grep -v bootstrap); do FEATURE=$feature phpdbg -qrr vendor/bin/behat --profile coverage features/$feature; done; else vendor/bin/behat; fi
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks @soyuka

teohhanhui:
Probably a better fix would be to reduce the verbosity of some of the installation logs. Colours are good...

soyuka:
Behat alone is >10k lines no?

teohhanhui:
Okay, now I'm not sure I get what this PR is supposed to do... \o/

soyuka:
I don't remember what PR it was but the author was asking for help debugging. I had to fork his branch to execute the tests on my device because travis output was unreadable because colors where stripped. Can't find the reference, but travis clearly stated that above 10 000 lines, colors won't be outputted and we'll instead get ASCII characters.

teohhanhui:
I think Travis stops showing log output when it hits 10000 lines (it's still available in the raw log). The way mozilla/fxa-content-server#488 handled it (reducing verbosity where unneeded) seems better to me.

teohhanhui:
See https://twitter.com/travisci/status/174904509754118144 and travis-ci/docs-travis-ci-com#281

soyuka:
Damn, I need to get my hand on the screenshot again. Anyway, if you really want colors back we can revert this ^^ but it's not a priority imo.

teohhanhui:
Symfony uses colors too... :stuck_out_tongue:

https://github.com/symfony/symfony/blob/59f3f34cc84e93e287e0971d05bfe0edee8e20e2/.travis.yml

soyuka:
> No colours = hard to read.

Many unix users would actually tell you that it is more a distraction 😛 . Thing is, it's way more readable than when colors are ASCII outputed no?

PR #1037, opened 2017-04-07
Allow Upper Case property names

insekticid:
ReflectionExtractor::getProperties() returns $id instead of $Id. It is bad naming convention, but is possible

```

class Entity {
    /**
     * @var string The dummy Guid
     *
     * @ORM\Column(nullable=true)
     * @ApiProperty()
     */
    protected $Id;

    public function getId()
    {
        return $this->Id;
    }
}
```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1026
| License       | MIT

### DIFF
--- a/src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
+++ b/src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
@@ -74,7 +74,7 @@ final class AnnotationPropertyNameCollectionFactory implements PropertyNameColle
             }
 
             $propertyName = $this->reflection->getProperty($reflectionMethod->name);
-            if (!preg_match('/^[A-Z]{2,}/', $propertyName)) {
+            if (!$reflectionClass->hasProperty($propertyName) && !preg_match('/^[A-Z]{2,}/', $propertyName)) {
                 $propertyName = lcfirst($propertyName);
             }
 
--- a/tests/Fixtures/TestBundle/Entity/UpperCaseIdentifierDummy.php
+++ b/tests/Fixtures/TestBundle/Entity/UpperCaseIdentifierDummy.php
@@ -15,14 +15,12 @@ use ApiPlatform\Core\Annotation\ApiResource;
 use Doctrine\ORM\Mapping as ORM;
 
 /**
- * Custom identifier dummy.
- *
- * @author Exploit.cz <insekticid@exploit.cz>
+ * UpperCaseIdentifier dummy.
  *
  * @ApiResource
  * @ORM\Entity
  */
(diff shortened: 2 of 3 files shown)


insekticid:
I forgot to modify Behat scenarios

dunglas:
I strongly encourage you to test it with a new entity instead of modifying an existing one. It will be less painful than editing Behat tests.

tests/Fixtures/TestBundle/Entity/UuidIdentifierDummy.php
```diff
@@ -17,6 +17,8 @@
 /**
  * Custom identifier dummy.
  *
+ * @author Exploit.cz <insekticid@exploit.cz>
```
meyerbaptiste: Why?

insekticid: oh, cool. someone is reading my code

because of this
https://github.com/api-platform/core/blob/master/tests/Fixtures/TestBundle/Entity/CircularReference.php#L22
and this
https://github.com/api-platform/core/blob/master/tests/Fixtures/TestBundle/Entity/AbstractDummy.php#L22
etc

Simperfit:
Shouldn't we add a behat test case to check ? @dunglas

soyuka:
@Simperfit no need to add a behat test here IMO, our behat suite is heavy enough (and painful to maintain too). The unit test covers the use case just fine!

tests/Fixtures/TestBundle/Entity/UuidIdentifierDummy.php
soyuka: lol no I wondered the same

IMO we should keep as `@author` the one who created the class not uppon each edit :). Should be fixed in other classes too. Still, what's the point of adding `@author` when we can `git blame`?

meyerbaptiste: @insekticid I have no problems with the `@author` tag but you made no changes on this class! 😄 
Add it to the `UpperCaseIdentifierDummy` class instead, it's more relevant! 😉

insekticid:
Sorry. I do not have any more time solving this simple stupid issue with
authoring this fix. I lost about 15hours of investigating bug with
xdebug+fix+phpunit. It is now up to you. You are allowed to change my pull.

dunglas:
Don't worry I can do the change during the merge.
Thanks you very much for fixing this.

soyuka:
Thanks @insekticid !

PR #1038, opened 2017-04-07
Refresh routes when an entity is created

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | a lot
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -160,7 +160,7 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
 
             $related = $em->getClassMetadata($mapping['targetEntity']);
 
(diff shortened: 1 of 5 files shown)


soyuka:
I'm wondering, what does this fix? Why do we need this? Has this an impact on performances? I remember an issue on the subject but can't put my hand on it anymore :|

dunglas:
When you download the distribution and and add an entity you got a 404 error if you don't clear the cache. I need to test if this patch fix this.

soyuka:
Hmm, clearing the cache should be the developer's work no? I'm not sure that it's a good idea to do it on the fly.

dunglas:
@soyuka we already do that (here and in Symfony) in dev mode for everything. It's definitely a bug to miss this specific case.

dunglas:
I've just tried it on a real world project, and it works as expected.

Issue #1039, opened 2017-04-08
Eager Loading Extension does not use dynamic serialization context


bwegrzyn:
I use a custom context builder to enable a resource expansion type feature. However, when looking at the context that the eager loading extension uses, it does not use the dynamic serialization groups I have defined.

Looks like the getSerializerGroups() method in EagerLoadingExtension gets them groups directly from the resource metadata object instead of the context builder.

dunglas:
ping @soyuka

soyuka:
I think that's because the `SerializerContextBuilder` relies on `Request` that I didn't injected it to the extensions. 

Which one of our contexts does handle the so called dynamic serialization group?

dunglas:
We may inject the `request_stack` service and use the `SerializerContextBuilder` only if a request is available?

soyuka:
So it's [this class](https://github.com/api-platform/core/blob/master/src/Serializer/SerializerContextBuilder.php) that handles dynamic serialization?

Yeah 👍  we have to do that because our eager loading will only fetch what needs to be serialized, and those groups are missing.

PR #1041, opened 2017-04-08
Eager Loading Extension does not use dynamic serialization context

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1039
| License       | MIT
| Doc PR        | na

ping @bwegrzyn may you try my branch to see if this fixes the issue? 
Could you please also give me an example of your dynamic context so that I can add a proper test for this use case?

Tyvm!

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -20,8 +20,10 @@ use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\QueryBuilder;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
  * Eager loads relations.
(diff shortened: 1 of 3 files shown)


bwegrzyn:
@soyuka Thanks, it looks like this will fix the issue. I will test it tonight and let you know.

My context builder looks for a "expand" parameter in the request (/users?expand=role). Based on the values in the expand parameter, it will verify and dynamically add groups to the context so that the serializer serializes those properties. In this case, the User resource has a Role resource (User->role). The serializer would not serialize the Role resource unless the serializer context contains the "role" group.

bwegrzyn:
@soyuka Just tested your PR, it works as expected. Thanks!

bwegrzyn:
I am curious about the requirements for it to use the groups from the ContextBuilder. 

Why do you require isMethodSafe() or for it to be DELETE? DELETE doesn't return any content anyways, and isMethodSafe() prevents eager loading during POST/PUT requests.

soyuka:
@bwegrzyn I need to refactor, indeed those are unneeded bits! I wanted to see if it actually fixed your issue before going further :).

soyuka:
@bwegrzyn May you share your ContextBuilder (or an extract)? Thanks!

bwegrzyn:
@soyuka This is what I'm using right now. WIP so far, but it seems to work fairly well despite certain short comings.

Maybe api-platform should support such a feature out of the box? It would be nice to support nested resource expansions using dot notation, but that may be something that the serializer component would need to support first.

```php
use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;

class ContextBuilder implements SerializerContextBuilderInterface
{

    /**
     * @var SerializerContextBuilderInterface
     */
    private $decorated;

    public function __construct(SerializerContextBuilderInterface $decorated)
    {
    }

    public function createFromRequest(Request $request, bool $normalization, array $extractedAttributes = null) : array
    {
        $context = $this->decorated->createFromRequest($request, $normalization, $extractedAttributes);

        // Only allow expansions on normalization
        if (!$normalization) {
            return $context;
        }

        $permittedExpansions = $context['permitted_expansions'] ?? [];

        $requestedExpansions = (array) $request->query->get('@expand', []);

        // Support CSV format for single parameter
        if (count($requestedExpansions) === 1) {
            $requestedExpansions = explode(',', $requestedExpansions[0]);
        }

        foreach ($requestedExpansions as $expansion) {
            $expansion = trim($expansion);
            if (in_array($expansion, $permittedExpansions)) {
                if (!isset($context['groups'])) {
                    $context['groups'] = [];
                }

                $context['groups'][] = $expansion;
            }
        }

        return $context;
    }

}
```

dunglas:
Thank you very much @soyuka

PR #1055, opened 2017-04-13
Fix strict typing exception when the propertyName is a number

jocel1:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/282
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
+++ b/src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
@@ -65,7 +65,7 @@ final class AnnotationPropertyNameCollectionFactory implements PropertyNameColle
         // Properties
         foreach ($reflectionClass->getProperties() as $reflectionProperty) {
             if (null !== $this->reader->getPropertyAnnotation($reflectionProperty, ApiProperty::class)) {
-                $propertyNames[$reflectionProperty->name] = true;
+                $propertyNames[$reflectionProperty->name] = $reflectionProperty->name;
             }
         }
 
@@ -81,7 +81,7 @@ final class AnnotationPropertyNameCollectionFactory implements PropertyNameColle
             }
 
             if (null !== $propertyName && null !== $this->reader->getMethodAnnotation($reflectionMethod, ApiProperty::class)) {
-                $propertyNames[$propertyName] = true;
+                $propertyNames[$propertyName] = (string) $propertyName;
             }
         }
 
@@ -92,6 +92,6 @@ final class AnnotationPropertyNameCollectionFactory implements PropertyNameColle
             }
         }
 
-        return new PropertyNameCollection(array_keys($propertyNames));
+        return new PropertyNameCollection(array_values($propertyNames));
     }
 }
--- a/src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
+++ b/src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
@@ -81,7 +81,7 @@ final class AnnotationPropertyNameCollectionFactory implements PropertyNameColle
             }
 
             if (null !== $propertyName && null !== $this->reader->getMethodAnnotation($reflectionMethod, ApiProperty::class)) {
-                $propertyNames[$propertyName] = (string) $propertyName;
+                $propertyNames[$propertyName] = $propertyName;
             }
         }
 
--- a/src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php
+++ b/src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php
@@ -41,7 +41,7 @@ final class InheritedPropertyNameCollectionFactory implements PropertyNameCollec
 
         // Inherited from parent
         foreach ($this->decorated->create($resourceClass, $options) as $propertyName) {
-            $propertyNames[$propertyName] = true;
+            $propertyNames[$propertyName] = (string) $propertyName;
         }
 
         foreach ($this->resourceNameCollection->create() as $knownResourceClass) {
@@ -51,11 +51,11 @@ final class InheritedPropertyNameCollectionFactory implements PropertyNameCollec
 
             if (is_subclass_of($knownResourceClass, $resourceClass)) {
                 foreach ($this->create($knownResourceClass) as $propertyName) {
-                    $propertyNames[$propertyName] = true;
+                    $propertyNames[$propertyName] = $propertyName;
                 }
             }
         }
 
-        return new PropertyNameCollection(array_keys($propertyNames));
+        return new PropertyNameCollection(array_values($propertyNames));
     }
 }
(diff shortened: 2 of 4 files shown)


src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
```diff
@@ -63,7 +63,7 @@ public function create(string $resourceClass, array $options = []): PropertyName
         // Properties
         foreach ($reflectionClass->getProperties() as $reflectionProperty) {
             if (null !== $this->reader->getPropertyAnnotation($reflectionProperty, ApiProperty::class)) {
-                $propertyNames[$reflectionProperty->name] = true;
+                $propertyNames[$reflectionProperty->name] = (string)$reflectionProperty->name;
```
soyuka: Can't we keep keys? `$propertyNames[(string) $reflectionProperty->name]` should work no?

soyuka:
For the test, I'd recommend to create a new fixture class (avoid editing `Dummy`, too many tests are based on it at one change will require lots of changes :p). 
You can add a test in [this class](https://github.com/api-platform/core/blob/master/tests/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactoryTest.php) by using the new fixture, would be just fine!

src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
soyuka: I studied this a bit more, looks like the fix should be in `PropertyNameCollection`:

```diff
--- a/src/Metadata/Property/PropertyNameCollection.php
+++ b/src/Metadata/Property/PropertyNameCollection.php
@@ -30,7 +30,9 @@ final class PropertyNameCollection implements \IteratorAggregate, \Countable
      */
     public function __construct(array $properties = [])
     {
-        $this->properties = $properties;
+        $this->properties = array_map(function($property) {
+            return (string) $property;
+        }, $properties);
     }
```

jocel1: For the key unfortunately php is converting the key to int, even when (string) is forced

soyuka: Indeed, still:

```
<?php

$t = [];
$t[118] = 'foo';
$t['bar'] = 'foo';

var_dump(array_map(function($v) {
    return (string) $v;
}, array_keys($t)));
```

Outputs:

```
array(2) {
  [0]=>
  string(3) "118"
  [1]=>
  string(3) "bar"
}
```

jocel1: yup a centralized array_map is definitely better to handle this issue

dunglas: `$reflectionProperty->name` is a always a string, so the cast isn't necessary (same everywhere).

dunglas: I'm not a fond of the centralized `array_map` solution. It looks hackish... This is not the reponsibility of `PropertyNameCollection` to fix the data provided by friend classes. Other classes should provide data respecting the contract (an array of strings).

soyuka: So we can make this even faster by using `$propertyNames[] = $reflectionProperty->name;` if we rely on values :).

src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
```diff
@@ -63,7 +63,7 @@ public function create(string $resourceClass, array $options = []): PropertyName
         // Properties
         foreach ($reflectionClass->getProperties() as $reflectionProperty) {
             if (null !== $this->reader->getPropertyAnnotation($reflectionProperty, ApiProperty::class)) {
-                $propertyNames[$reflectionProperty->name] = true;
+                $propertyNames[$reflectionProperty->name] = $reflectionProperty->name;
```
Simperfit: Key may be useless now ?

src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
dunglas: Not really because we must ensure that properties are unique. If you use values, you'll need to call `in_array` and it's an O(n) operation instead of O(1).

src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
dunglas: It's not, it ensures the uniqueness.

jocel1:
I've just pushed the required tests

soyuka:
@jocel1 may you change the target branch to `2.0` and rebase? (If you're not sure about how to rebase lmk).

jocel1:
rebase & squash done

src/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactory.php
```diff
@@ -81,7 +81,7 @@ public function create(string $resourceClass, array $options = []): PropertyName
             }
 
             if (null !== $propertyName && null !== $this->reader->getMethodAnnotation($reflectionMethod, ApiProperty::class)) {
-                $propertyNames[$propertyName] = true;
+                $propertyNames[$propertyName] = (string) $propertyName;
```
dunglas: This cast can be removed too.

src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php
```diff
@@ -41,7 +41,7 @@ public function create(string $resourceClass, array $options = []): PropertyName
 
         // Inherited from parent
         foreach ($this->decorated->create($resourceClass, $options) as $propertyName) {
-            $propertyNames[$propertyName] = true;
+            $propertyNames[$propertyName] = (string) $propertyName;
```
dunglas: This cast can be removed.

jocel1: mm no, this is this one in my case which was causing the issue : the symfony component returns an int instead of a string

dunglas: It looks like the Symfony component suffers from the same bug. We may keep this fix temporarily but we should fix this behavior in Symfony as well then bump the minimal version required by API Platform.

Do you want to handle this? If you're out of time I can open the Symfony PR.

dunglas:
Travis and AppVeyor seem to have some troubles. Can you rebase and push force again?

src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php
jocel1: If you can handle this one, I'm definitely out of my time ;)

dunglas:
Thank you very much @jocel1!

Issue #763, opened 2016-09-21
Notice: Undefined index: subType


dagrinchi:
Notice: Undefined index: subType
in vendor/dunglas/api-bundle/src/Bridge/NelmioApiDoc/Parser/ApiPlatformParser.php at line 220

teohhanhui:
Which version are you using? 1.1?

dagrinchi:
Hello, 

First, I installed from here:

`composer require dunglas/api-bundle:dev-master`

Then, I removed and install again from:

`composer require api-platform/core:v2.0.0-rc.1`

And problem solved 👍 

Thanks for your reply!

tacman:
I'm getting this as well now, I think the issue may be that api-platform/core is on RC6 and api-bundle is a few versions back.

dunglas:
You should use api-platform/core or dunglas/api-bundle but not both.

PR #1056, opened 2017-04-13
fix undefined index: subType in NelmioApiDoc/ApiPlatformParser

karser:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | ?
| Fixed tickets | #763 
| License       | MIT

I think it has to do with the entity that uses table inheritance.


dunglas:
Thanks @karser

PR #1058, opened 2017-04-14
Fix the bundle when soft deps aren't installed

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

This PR fixes some edge cases. It's require to be compatible with Symfony Flex.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,12 +46,12 @@
         "symfony/config": "^3.2",
         "symfony/dependency-injection": "^2.7 || ^3.0",
         "symfony/doctrine-bridge": "^2.8 || ^3.0",
-        "symfony/phpunit-bridge": "^2.7 || ^3.0",
-        "symfony/security": "^2.7 || ^3.0",
-        "symfony/validator": "^2.7 || ^3.0",
         "symfony/finder": "^2.7 || ^3.0",
         "symfony/framework-bundle": "^3.1",
-        "symfony/twig-bundle": "^2.8 || ^3.1"
+        "symfony/phpunit-bridge": "^2.7 || ^3.0",
+        "symfony/security": "^2.7 || ^3.0",
+        "symfony/twig-bundle": "^2.8 || ^3.1",
+        "symfony/validator": "^2.7 || ^3.0"
     },
     "suggest": {
         "friendsofsymfony/user-bundle": "To use the FOSUserBundle bridge.",
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection;
 
+use Doctrine\Common\Annotations\Annotation;
 use Doctrine\ORM\Version;
 use phpDocumentor\Reflection\DocBlockFactoryInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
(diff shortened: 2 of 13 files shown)


src/Metadata/Property/Factory/AnnotationPropertyMetadataFactory.php
```diff
@@ -131,8 +131,11 @@ private function createMetadata(ApiProperty $annotation, PropertyMetadata $paren
 
     private function createWith(PropertyMetadata $propertyMetadata, array $property, $value): PropertyMetadata
     {
-        $getter = $property[0].ucfirst($property[1]);
+        if (null === $value) {
+            return $propertyMetadata;
+        }
```
soyuka: Why this change?

composer.json
```diff
@@ -46,12 +46,12 @@
...
-        "symfony/phpunit-bridge": "^2.7 || ^3.0",
-        "symfony/security": "^2.7 || ^3.0",
-        "symfony/validator": "^2.7 || ^3.0",
         "symfony/finder": "^2.7 || ^3.0",
         "symfony/framework-bundle": "^3.1",
-        "symfony/twig-bundle": "^2.8 || ^3.1"
+        "symfony/phpunit-bridge": "^2.7 || ^3.0",
+        "symfony/security": "^2.7 || ^3.0",
+        "symfony/twig-bundle": "^2.8 || ^3.1",
+        "symfony/validator": "^2.7 || ^3.0"
```
dunglas: It is in alphabetical order isn't it?

src/Metadata/Property/Factory/AnnotationPropertyMetadataFactory.php
dunglas: Because it was leading to a bug if the value is null (the default)

composer.json
soyuka: :laughing: omg friday you know (and I've apparently issues with the alphabet).

src/Metadata/Property/Factory/AnnotationPropertyMetadataFactory.php
soyuka: Okay, I'd have kept in from calling this method in the above loop instead of adding the condition here. Weird that we had no issues about this though oO.

dunglas: Done (it removes some function calls in a loop, it's always appreciable).

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -129,16 +134,71 @@ private function handleConfig(ContainerBuilder $container, array $config, array
...
+
+        if (class_exists(Yaml::class)) {
+            $loader->load('metadata/yaml.xml');
+            $container->getDefinition('api_platform.metadata.extractor.yaml')->addArgument($yamlResources);
+        }
 
-        if (!interface_exists(DocBlockFactoryInterface::class)) {
-            $container->removeDefinition('api_platform.metadata.resource.metadata_factory.php_doc');
+        if (interface_exists(DocBlockFactoryInterface::class)) {
+            $loader->load('metadata/php_doc.xml');
```
teohhanhui: Why `php_doc` instead of `phpdoc`? For example the filename of the PSR draft is https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md

teohhanhui: Just like PHPUnit is always `phpunit`, not `php_unit` or `php-unit` etc.

PR #1046, opened 2017-04-11
Tried to fix https://github.com/api-platform/core/issues/1025

nik-lampe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | n/a
| Deprecations? | no
| Tests pass?   | n/a
| Fixed tickets | #1234, #5678
| License       | MIT
| Doc PR        | api-platform/doc#1234

This is my WIP Pull Request adressing issue https://github.com/api-platform/core/issues/1025
I am not sure, if this does not break something else...

Issue #1025, opened 2017-04-01
Wrong attributes serialized when embedding multiple different inherited entities


nik-lampe:
When serializing an related entity which hast some kind of entity with inheritance  embedded (doesn't actually matter if it's single table or joined).
The properties of the parent class are serialized, and ONLY the properties of one of the child classes. But for all entites, regardless of their type. 

This is a sample setup:
```

/**
 * @ApiResource()
 * @ORM\Entity
 * @ORM\InheritanceType(value="JOINED")
 */
class ParentClass
{
    /**
     * @var int The entity Id
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     * @Groups({"get"})
     */
    protected $id;

    /**
     * @var string Something else
     *
     * @ORM\Column
     * @Assert\NotBlank
     * @Groups({"get"})
     */
    protected $baz = '';
}
```
```
/**
 * @ApiResource
 * @ORM\Entity
 */
class BarChild extends ParentClass
{
    /**
     * @var int The entity Id
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    protected $id;

    /**
     * @var string Something else
     *
     * @ORM\Column
     * @Assert\NotBlank
     * @Groups({"get"})
     */
    private $bar = '';
}
```
```
/**
 * @ApiResource
 * @ORM\Entity
 */
class FooChild extends ParentClass
{
    /**
     * @var int The entity Id
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    protected $id;

    /**
     * @var string Something else
     *
     * @ORM\Column
     * @Assert\NotBlank
     * @Groups({"get"})
     */
    private $foo = '';
```
```
/**
 * @ApiResource(attributes={"normalization_context"={"groups"={"get"}}})
 * @ORM\Entity()
 */
class RelatedEntity
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     * @ORM\Column(type="integer")
     * @Groups({"get"})
     */
    private $id;

    /**
     * @ORM\Column(type="string")
     * @Groups({"get"})
     */
    private $attribute;

    /**
     * @ORM\OneToMany(targetEntity="AppBundle\Entity\ParentClass", mappedBy="relatedEntity")
     * @Groups({"get"})
     */
    private $relatedEntites;
}
```

[...]

Issue #1043, opened 2017-04-11
Custom filtered collection as property throws error


nik-lampe:
I also posted this question on stackoverflow, but I am not sure if it is a bug or a question...
https://stackoverflow.com/questions/43339263/api-platform-custom-filtered-collection-as-property

I am trying to serialize a collection of an entity multiple times with different criteria. I have created the following example entities:

```
namespace AppBundle\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Serializer\Annotation\Groups;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * @ApiResource(attributes={"normalization_context"={"groups"={"get"}}})
 * @ORM\Entity()
 */
class EntityWithFilteredCollection
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     * @ORM\Column(type="integer")
     * @Groups({"get"})
     */
    private $id;

    /**
     * @ORM\OneToMany(targetEntity="AppBundle\Entity\FilteredCollectionRelatedEntity", mappedBy="relatedEntity")
     */
    private $relatedEntities;

    /**
     * @Groups({"get"})
     */
    private $filteredEntities;

    public function __construct()
    {
        $this->relatedEntities = new ArrayCollection();
    }

    public function getId()
    {
        return $this->id;
    }

    public function getRelatedEntities()
    {
        return $this->relatedEntities;
    }

    public function setRelatedEntities($relatedEntities)
    {
        $this->relatedEntities = $relatedEntities;
    }

    public function getFilteredEntities()
    {
        return $this->filteredEntities;
    }

    public function setFilteredEntities($filteredEntities)
    {
        $this->filteredEntities = $filteredEntities;
    }
}
```

```
namespace AppBundle\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Serializer\Annotation\Groups;

/**
 * @ApiResource(attributes={"normalization_context"={"groups"={"get"}}})
 * @ORM\Entity()
 */
class FilteredCollectionRelatedEntity
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     * @ORM\Column(type="integer")
     * @Groups({"get"})
     */
    private $id;

    /**
[...]

PR #1063, opened 2017-04-17
Multiple inheritance collection serialization

bwegrzyn:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1025
| License       | MIT
| Doc PR        | 

As I am having the same issues as #1025, I figured I'd contribute a failing test case to help move things along on PR #1046. 

@nik-lampe @dunglas @soyuka Can you take a look at this test case and make sure it matches your failing use case?

### DIFF
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -66,6 +66,9 @@ final class ItemNormalizer extends AbstractItemNormalizer
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
         $data = $this->addJsonLdContext($this->contextBuilder, $resourceClass, $context);
 
+        // Use resolved resource class instead of given resource class to support multiple inheritance child types
+        $context['resource_class'] = $resourceClass;
+
         $rawData = parent::normalize($object, $format, $context);
         if (!is_array($rawData)) {
             return $rawData;
--- a/features/main/table_inheritance.feature
+++ b/features/main/table_inheritance.feature
@@ -80,3 +80,142 @@ Feature: Table inheritance
       "required": ["hydra:member"]
     }
     """
+
+  @createSchema
+  Scenario: Create a table inherited resource
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/dummy_table_inheritance_children" with body:
+    """
+    {"name": "foo", "nickname": "bar"}
+    """
+    Then the response status code should be 201
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@type": {
+          "type": "string",
+          "pattern": "^DummyTableInheritanceChild$"
+        },
+        "@context": {
+          "type": "string",
+          "pattern": "^/contexts/DummyTableInheritanceChild$"
+        },
+        "@id": {
+          "type": "string",
+          "pattern": "^/dummy_table_inheritance_children/1$"
+        },
+        "name": {
+          "type": "string",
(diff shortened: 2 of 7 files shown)


bwegrzyn:
Curiosity got the better of me, and I found a super simple fix while debugging the issue. It fixed all my use cases and all behat tests, including the ones I added, are green.

However, there is one failing phpunit tests. It looks like it just might have to be updated, but I'm not sure. ping @dunglas @soyuka can you take a look?

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -66,6 +66,9 @@ public function normalize($object, $format = null, array $context = [])
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
         $data = $this->addJsonLdContext($this->contextBuilder, $resourceClass, $context);
 
+        // Use resolved resource class instead of given resource class to support multiple inheritance child types
+        $context['resource_class'] = $resourceClass;
```
soyuka: Is this true for classes without inheritance?

soyuka:
This fixes test, but it's kinda weird that we now call `getResourceClass` twice with different parameters.

```diff
index e474a5b..fdc57b5 100644
--- a/tests/Hydra/Serializer/ItemNormalizerTest.php
+++ b/tests/Hydra/Serializer/ItemNormalizerTest.php
@@ -102,6 +102,7 @@ class ItemNormalizerTest extends \PHPUnit_Framework_TestCase

         $resourceClassResolverProphecy = $this->prophesize(ResourceClassResolverInterface::class);
         $resourceClassResolverProphecy->getResourceClass($dummy, null, true)->willReturn(Dummy::class)->shouldBeCalled();
+        $resourceClassResolverProphecy->getResourceClass($dummy, Dummy::class, true)->willReturn(Dummy::class)->shouldBeCalled();

         $serializerProphecy = $this->prophesize(SerializerInterface::class);
         $serializerProphecy->willImplement(NormalizerInterface::class);
```

bwegrzyn:
Yes, I've noticed that and that's how I fixed the unit test as well, but I have not committed it yet.

I agree that it is a little weird, but I'm not sure how to go about fixing it another way since ItemNormalizer->normalize is where we first are able to figure out the proper resource class to normalize. Any ideas?

soyuka:
I think modifying the test is fine. Indeed, we force the `context['resource_class']` on the next `normalize` passes.

soyuka:
@nik-lampe does this patch fixes your issue?

nik-lampe:
Yes, looks pretty good so far. It is a similar change like I made in my code, but much better.

nik-lampe:
I just came across one issue, that was solved with my dirty hacked fix but came back when using the fix from @bwegrzyn. Can't pin it down right now, but I'm on it

bwegrzyn:
@nik-lampe Do you have a failing test case for it or an example? I tried to cover my changes with all the use cases I could think of, but I may have missed some.

nik-lampe:
@bwegrzyn I took a quick look and the issue I described in #1043 might be a case. At least it fails with your changes but works with mine

bwegrzyn:
@nik-lampe It looks like that issue is separate from this issue and not related to inheritance. I think this PR should be merged if it is approved, and the other issue should be taken care of in a new PR. 

I'll take a look at #1043 and see if I can help you with it.

nik-lampe:
That's true, it has nothing to do with inheritance. But it results from having the wrong resource class in the context when trying to access the property metadata. I agree with you, that it should be taken care of in a new PR.

bwegrzyn:
@dunglas I'm having some trouble figuring out how to do that properly. Can you help me out?

soyuka:
Thanks @bwegrzyn !

Issue #1064, opened 2017-04-17
Issue with FOSUserBundle 2


pierredup:
I'm still investigating the issue, but just wanted to mention it here so long, in case someone else has come across this.

I'm building a new api with FOSUserBundle 2, and the issue that I'm experiencing is that my user passwords are never hashed upon save. This is due to the `UserManager::updateUser` method never being called.

What I've found so far, in [ApiPlatform\Core\Bridge\FosUser\EventListener#L49](https://github.com/api-platform/core/blob/master/src/Bridge/FosUser/EventListener.php#L49), when dumping the result of the `$event->getControllerResult()` call, I actually get a JSON string instead of the user object:

```
string(89) "{"id":"1bc3b33a9d3b1e09","email":"foo@bars12.com","fullName":null,"username":"foobars12"}"
```

The next line then checks `$user instanceof UserInterface`, and returns if it's false, which means the `switch` is never reached below to call the `updateUser` method.

pierredup:
Since #597, the `ApiPlatform\Core\EventListener\SerializerListener` (formerly `SerializerViewListener`) class changed priorty from 10 to 16.
The `ApiPlatform\Core\Bridge\FosUser\EventListener` class has a priority of 11, which means it now runs after the `SerializerListener`.

I have created #1065 to fix this issue

PR #597, opened 2016-07-06
Move data retrieving and deserialization to kernel.request listeners

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

PR #1065, opened 2017-04-17
Update priority of FosUser\EventListener

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1064
| License       | MIT
| Doc PR        | N/A

Since #597, the `ApiPlatform\Core\EventListener\SerializerListener` (formerly `SerializerViewListener`) class changed priorty from 10 to 16.
The `ApiPlatform\Core\Bridge\FosUser\EventListener` class has a priority of 11, which means it now runs after the `SerializerListener`. This causes the controller result to be already serialized, which fails the checks in the FOSUser listener.

This PR updates the priority of `ApiPlatform\Core\Bridge\FosUser\EventListener` to 24, so that it can run before the `SerializerListener` class.

### DIFF
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
@@ -170,7 +170,7 @@ class ApiPlatformExtensionTest extends \PHPUnit_Framework_TestCase
         $this->assertGreaterThan(
             $viewListener->getTag('kernel.event_listener')[0]['priority'],
             $fosListener->getTag('kernel.event_listener')[0]['priority'],
-            "api_platform.fos_user.event_listener priority needs to be greater than that of api_platform.listener.view.serialize"
+            'api_platform.fos_user.event_listener priority needs to be greater than that of api_platform.listener.view.serialize'
         );
     }
 
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
@@ -19,6 +19,7 @@ use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use FOS\UserBundle\FOSUserBundle;
 use Nelmio\ApiDocBundle\NelmioApiDocBundle;
 use Prophecy\Argument;
+use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\ResourceInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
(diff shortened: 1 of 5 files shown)


dunglas:
Great! Thank you very much for finding and fixing the problem.
Can you just add a regression est to prevent future similar problems?

soyuka:
Can you rebase on 2.0 also? Thanks!

pierredup:
I added a unit test, but it feels a bit ugly (especially the path mapping to load the configs), let me know if there is a better way.
I'll do the rebase shortly

dunglas:
IMO a Behat (integration) test will be easier to do and more useful.

soyuka:
Isn't this enough to test the priority? IMO, the less behat tests we have the better.

dunglas:
But we need to be sure that this feature actually work :)

pierredup:
The integration is already tested separately, so IMO the priority is the only thing that still needs to be tested because that is what is broken. So I have kept the unit test to ensure the priority is set correctly, and added another step to the integration test to ensure the password hashing works when creating users

pierredup:
Rebase done

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -153,6 +155,25 @@ public function testEnableFosUser()
...
+        $loader->load('fos_user.xml');
+
+        $fosListener = $builder->getDefinition('api_platform.fos_user.event_listener');
+        $viewListener = $builder->getDefinition('api_platform.listener.view.serialize');
+
+        // Ensure FOSUser event listener priority is always greater than the view serialize listener
+        $this->assertGreaterThan(
+            $viewListener->getTag('kernel.event_listener')[0]['priority'],
+            $fosListener->getTag('kernel.event_listener')[0]['priority'],
+            "api_platform.fos_user.event_listener priority needs to be greater than that of api_platform.listener.view.serialize"
```
soyuka: Needs simple quotes, lint failed :)

dunglas:
Thanks @pierredup!

ad3n:
@dunglas :

```
"kernel.view" event
-------------------

 ------- ------------------------------------------------------------------------------------------ ---------- 
  Order   Callable                                                                                   Priority  
 ------- ------------------------------------------------------------------------------------------ ---------- 
  #1      ApiPlatform\Core\Bridge\Symfony\Validator\EventListener\ValidateListener::onKernelView()   64        
  #2      ApiPlatform\Core\Bridge\Doctrine\EventListener\WriteListener::onKernelView()               32        
  #3      ApiPlatform\Core\EventListener\SerializeListener::onKernelView()                           16        
  #4      Bisnis\Subscriber\ResponseCacheSubscriber::onKernelView()                                  9         
  #5      ApiPlatform\Core\EventListener\RespondListener::onKernelView()                             8         
  #6      Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener::onKernelView()          0         
 ------- ------------------------------------------------------------------------------------------ ---------- 
```
The listener is not registered, why? I try to add my own with priority 9 and that is called

Issue #1071, opened 2017-04-20
Collection of Doctrine Single table inheritance entities


desmax:
I have an object, that has collection containing different entities, that use Single inheritance.
Problem is that only base fields are fetched, entity specific are null. As I understand its happening because Doctrine partial query is used, but how can I force load all fields for given collection?

soyuka:
#1025 can you try the `2.0` branch?

desmax:
@soyuka I am sorry for duplicate. And I am on master already.

soyuka:
Hmm this patch hasn't been merged back to master yet :|.

soyuka:
@desmax done through #1072 can you update?

desmax:
@soyuka I don't seem to see any difference...

soyuka:
ping @bwegrzyn 

Quoting some related issues:

- #1069
- https://github.com/api-platform/api-platform/issues/277
- https://github.com/api-platform/api-platform/issues/266
- https://github.com/api-platform/api-platform/issues/263
- https://github.com/api-platform/api-platform/issues/272

I'll investigate this and hopefully find a clean fix.

bwegrzyn:
@desmax Can you try to disabled eager loading and try again?

```yml
api_platform:
    eager_loading:
        enabled: false
```

Also, can you provide a sample of your entities?

soyuka:
Even better disable it on the resource:

```
/*
 * @ApiResource(attributes={"force_eager"=false})
 */
```

bwegrzyn:
@soyuka Interestingly... that doesn't disabled eager loading on a resource for me. I have to force it through config.yml. That might be a separate issue.

soyuka:
This is the condition:
```
            if (false === $forceEager && ClassMetadataInfo::FETCH_EAGER !== $mapping['fetch']) {
```

Anyway, I need to refactor this fetchEager thing #1035

desmax:
@bwegrzyn disabling eager loading via config does help.
On resource object I have one to many relation to base single inherited master class, marked as abstract.
All non abstract entities extend base class, some of them have additional fields, that are not present in base class. Base class is marked with `@ApiResource()` annotation.
```    
oneToMany:
  history:
     targetEntity: AppBundle\Entity\HistoryAction\HistoryAction
     mappedBy: vacancy
```

````
AppBundle\Entity\HistoryAction\HistoryAction:
    type: entity

    inheritanceType: SINGLE_TABLE

    discriminatorColumn:
[...]

Issue #1069, opened 2017-04-20
Eager Loading Extension partial objects causing properties to not be available


bwegrzyn:
It seems like the eager loading extension loads partial objects from Doctrine through the use of the `@Groups` annotation. While this seems like it would be a good idea in theory, proves to be quite a hassle in practice.

I've started using @dunglas https://github.com/dunglas/doctrine-json-odm to store events with a bunch of data in them as opposed to using inheritance with several children. Because these json_document fields contain quite a bit of info that I don't need to expose in the API, I don't want to expose the entire field. Instead, I create custom getters and setters and expose those via the `@Groups` annotation. 

However, because the json_document field is not exposed through the `@Groups` annotation, it is always NULL when the eager loading extension is enabled, and thus the custom getters that I have exposed return NULL as well. 

In the example below, you can see the `context` field being using to the store the `json_document`. Individual properties of this property are exposed through the `getMessage()` and `setMessage()` methods. This does not work when eager loading is enabled as the `context` field is always NULL, but works correctly when eager loading is disabled.

I'd like to propose a few methods on how to handle this:

1. Remove partial objects from the eager loading extension as they prove to be more painful than helpful.
2. Maybe add a property to the `@ApiProperty` annotation that will cause the eager loading extension to load the property regardless of the `@Group` annotation. 
3. Any other ideas?

```php
/**
 * @ApiResource(
 *     attributes={
 *          "normalization_context"={"groups"={"api-read"}},
 *          "denormalization_context"={"groups"={"api-write"}}
 *     }
 * )
 *
 * @ORM\Entity
 */
class Event
{
    
    /**
     * @ORM\Column(type="json_document", options={"jsonb": true})
     *
     * @var EventContextInterface
     */
    protected $context;

    /**
     * @return EventContextInterface
     */
    public function getContext()
    {
        return $this->context;
    }

    /**
     * @param EventContextInterface $context
     */
    protected function setContext(EventContextInterface $context)
    {
        $this->context = $context;
    }
    
    /**
     * @Groups({"api-read"})
     *
     * @return string
     */
    public function getMessage()
    {
        return $this->getContext()->getMessage();
    }

    /**
     * @Groups({"api-write"})
     *
     * @param $message
     *
     * @return $this
     */
    public function setMessage($message)
    {
        $this->getContext()->setMessage($message);

        return $this;
    }
}
```

soyuka:
I disagree about removing partials, it's neat to fetch only the data you need. 

[...]

PR #1073, opened 2017-04-21
EagerLoadingExtension: Disable partial fetching if entity has subclasses

desmax:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1069, #1071
| License       | MIT
| Doc PR        | none

Tried to fix this issue "Eager Loading Extension partial objects causing properties to not be available" #1069 So I basically select all fields for entities that have subclasses.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -170,22 +170,25 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
         $select = [];
         $entityManager = $queryBuilder->getEntityManager();
         $targetClassMetadata = $entityManager->getClassMetadata($entity);
+        if ($targetClassMetadata->subClasses) {
+            $queryBuilder->addSelect($associationAlias);
+        } else {
+            foreach ($this->propertyNameCollectionFactory->create($entity) as $property) {
+                $propertyMetadata = $this->propertyMetadataFactory->create($entity, $property, $propertyMetadataOptions);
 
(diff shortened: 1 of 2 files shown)


dunglas:
Can you add a non-regression test?

soyuka:
The perfect solution would be to call `joinRelations` on every `subclasses` instead of just doing this. 

The issue with this is that it won't eager load subclasses properties that might have a joined serialization group. Doctrine will then lazy-load them.

WDYT @desmax ?

desmax:
@soyuka Looks like this is impossible. If you join with base class, doctrine would not allow to join relation of subclass: "Error: Class X has no association named Y".

soyuka:
> Looks like this is impossible. 

Nothing is impossible.

But yes it might not be easy to do, and I don't know if it has a real gain. 

Can you add a test on this?

desmax:
@soyuka I will try.
why am I getting this?
[Behat\Testwork\ServiceContainer\Exception\ExtensionInitializationException]
  `Behatch\Extension` extension file or class could not be located.
never worked with Behat before..

soyuka:
@desmax do a rebase on `master`, we added php stan and a few errors were already fixed in #1074. IMO your solution is fine, just add a test and we'll merge ;)!

Simperfit:
I agree with @soyuka and @dunglas a non-regression test should be added :)

@desmax Thank you for contributing to Api-Platform !

desmax:
@soyuka I updated pull request. Added phpunit test, is it good enough?

dunglas:
Thank you @desmax!

PR #1076, opened 2017-04-21
Fix the resource class resolver (and collection normalizers) with Traversable values

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

This PR allows the `ResourceClassResolver::getResourceClass()` method to return the given resource class when value is a `\Traversable`.
So, `CollectionNormalizer` classes no longer throw an `InvalidArgumentException` exception with `\Traversable` objects during the normalization.

### DIFF
--- a/src/Api/ResourceClassResolver.php
+++ b/src/Api/ResourceClassResolver.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Api;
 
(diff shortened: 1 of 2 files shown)


dunglas:
Thanks @meyerbaptiste

Issue #1070, opened 2017-04-20
Assert\NotBlank() doesn't prevent AbstractItemNormalizer error


kate-kate:
Hi guys! I have an entity with related field like this:
```php
    /**
     * @Groups({"dashboard","create_manual"})
     * @Assert\NotBlank(groups={"create_manual"})
     *
     * @ORM\ManyToOne(targetEntity="Campaign")
     */
    private $campaign;
```
where ```"create_manual"``` is a denormalization context group.

And then comes PUT request with ```"campaign":""``` (because there is a select in the frontend, which is sending campaign IRI from dropdown or empty string if user doesn't choose any).

 I expected validation error for this variant, but I get error: ```Type error: Argument 2 passed to ApiPlatform\Core\Metadata\Property\Factory\CachedPropertyMetadataFactory::create() must be of the type string, integer given, called in /path-to-my-project/vendor/api-platform/core/src/Serializer/AbstractItemNormalizer.php on line 150```.

 I saw exactly the same error, once I tried to create embedded entity and have forgotten to use correct ```denormalization_context```. 

But I don't expect any objects here. I'm waiting for correct IRI or at least empty string, which may cause validation error. 

So, what's wrong?

bwegrzyn:
I'm guessing the error is caused by strict types being declared on all files in api platform recently. I'm not sure why its trying to pass an integer though.

Can you post the full error?

kate-kate:
@bwegrzyn of course, validation groups are not the same as groups on the normalization/denormalization context:) But I named them the same (it seems more simple for me). This is full ```@ApiResource``` annotation of my entity:
```php
/**
 * @ApiResource(attributes={
 *     "validation_groups"={"create_manual"},
 *     "normalization_context"={"groups"={"dashboard"}},
 *     "denormalization_context"={"groups"={"create_manual"}},
 *     "pagination_client_enabled"=false,
 *     "filters"={"job.search", "job.date", "job.order"}
 * })
 */
```

And here is the full error response:
```
{
  "@context": "/api/contexts/Error",
  "@type": "hydra:Error",
  "hydra:title": "An error occurred",
  "hydra:description": "Deserialization for the format html is not supported",
  "trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
      "file": "/Users/kate/htdocs/api_tryoup/vendor/symfony/symfony/src/Symfony/Component/Serializer/Serializer.php",
      "line": 127,
[...]

PR #1077, opened 2017-04-22
test and fix empty string relation validation

kate-kate:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1070 
| License       | MIT

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -280,7 +280,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             } catch (ItemNotFoundException $e) {
                 throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
             } catch (InvalidArgumentException $e) {
-                if ("" === $value && ($propertyMetadata->isWritableLink() || $this->resourceClassResolver->isResourceClass($className))) {
+                if ('' === $value && ($propertyMetadata->isWritableLink() || $this->resourceClassResolver->isResourceClass($className))) {
                     return null;
                 }
                 // Give a chance to other normalizers (e.g.: DateTimeNormalizer)
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -280,6 +280,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             } catch (ItemNotFoundException $e) {
                 throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
             } catch (InvalidArgumentException $e) {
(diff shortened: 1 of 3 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -280,6 +280,9 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
             } catch (ItemNotFoundException $e) {
                 throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
             } catch (InvalidArgumentException $e) {
+                if ($value === "" && ($propertyMetadata->isWritableLink() || $this->resourceClassResolver->isResourceClass($className))) {
```
soyuka: "" === $value

dunglas:
I need to review this carefully but this change looks wrong to me. Give me some time to dive into the issue.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -280,6 +280,9 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
             } catch (ItemNotFoundException $e) {
                 throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
             } catch (InvalidArgumentException $e) {
+                if ('' === $value && ($propertyMetadata->isWritableLink() || $this->resourceClassResolver->isResourceClass($className))) {
+                    return null;
+                }
                 // Give a chance to other normalizers (e.g.: DateTimeNormalizer)
```
kate-kate: @dunglas here is this place. We normalizing relation, we got string, so it must definitely be IRI and no any chances for other. I've just tried to fix the symptom for exception didn't crush the application

amenophis:
@dunglas some news about this issue ??

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -280,6 +280,9 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
             } catch (ItemNotFoundException $e) {
                 throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
             } catch (InvalidArgumentException $e) {
+                if ('' === $value && ($propertyMetadata->isWritableLink() || $this->resourceClassResolver->isResourceClass($className))) {
```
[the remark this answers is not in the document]
kate-kate: Well, this causes the validation problem. Because empty strings are often used as a null value for dropdown lists in web, for example. So you think, that you provided an analogue variant of null, you expect the validation error, as I described in #1070 but you getting the exception and it is a bit confusing. Also I've stacked with it in a few projects and I had to use my own fork in all of them due to frontend get unexpected exceptions instead of validation errors. And you just can't send "null" from the select. And I see in #1070 that I'm not the only one who had a problem with it.

soyuka:
see https://github.com/api-platform/core/issues/1070#issuecomment-480512622

PR #1078, opened 2017-04-24
Add support for customizable api resources folder's name.

omissis:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/195

This PR adds the possibility to customise the name of the folder in which the api resources (as in the classes tagged using the `@ApiResource` annotation) are stored, rather than the hardcoded "Entity" folder.
Such change allows for easier separation between the entities used for persisting the data and the classes used for describing the api resources.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -113,6 +113,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.exception_to_status', $config['exception_to_status']);
         $container->setParameter('api_platform.formats', $formats);
         $container->setParameter('api_platform.error_formats', $errorFormats);
+        $container->setParameter('api_platform.api_resources_directory', $config['api_resources_directory']);
         $container->setParameter('api_platform.eager_loading.enabled', $config['eager_loading']['enabled']);
         $container->setParameter('api_platform.eager_loading.max_joins', $config['eager_loading']['max_joins']);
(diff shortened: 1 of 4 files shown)


Simperfit:
I think this should be on master branch

omissis:
@Simperfit should I open a second PR for that?

dunglas:
Hi, and thanks for contributing! But as (non public) bundles will be removed from Symfony 4 (and not recommended with 3.3 + Flex), I'm not sure we need this change (we will probably drop automatic bundle discovery in API Platform 3).

omissis:
hi @dunglas, I haven't followed Symfony's developments too closely lately, so I wasn't aware of that. Is such decision something that will impact the 2.0 branch as well(if I understood well, it shouldn't)? As for now this PR only targets V2. Thanks for the heads up!

soyuka:
@dunglas until then we could merge this no?

dunglas:
Anyway it should be merged in master (it's a new feature). As 2.1 will be released almost at the same time than Symfony 3.3, I'm not sure it is worth it (there is already a feature in master allowing to customize the full path).

Simperfit:
ping @teohhanhui @vincentchalamon WDYT ?

Simperfit:
@omissis Could you please change the target to master and rebase to master if needed ?

omissis:
@Simperfit sure, no problem. I will do it in the next days.

omissis:
@Simperfit rebased and changed the target to master as you suggested. There seems to be a memory issue on travis though :(

dunglas:
Thanks @omissis!

teohhanhui:
This doesn't really provide enough flexibility as it impacts all bundles.

PR #1080, opened 2017-04-24
Add attributes to metadata XSD file

scollado:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | TBD

Not a critical thing, but as I tend to prefer XML for configuration files I'm sure other will enjoy the change :
Added attributes allow to validate resources.xml files as developers can do with other Symfony config files (ex: routing, services, ...)
This would ease auto-completion by editors supporting XML structure hints from the XSD (like PHPStorm) by defining the namespace in the XML header, as other XML config files do in Symfony : 
```xml
<!-- src/AppBundle/Resources/config/api_resources/resources.xml -->

<resources xmlns="https://api-platform.com/schema/metadata"
           xsi:schemaLocation="https://api-platform.com/schema/metadata
           https://api-platform.com/schema/metadata/metadata-2.0.xsd">
   [...]
</resources>
```
The proposed URI for the schema (`https://api-platform.com/schema/metadata/metadata-2.0.xsd`) is subject of discussion as it implies that the last stable version of the XSD file for the choosen branch is publicly available.

I'll make the PR to the Getting-Started page of the documentation once those details will be sorted out.
WHat do you think about this ?

### DIFF
--- a/src/Metadata/schema/metadata.xsd
+++ b/src/Metadata/schema/metadata.xsd
@@ -1,7 +1,9 @@
 <?xml version="1.0" ?>
 
 <xsd:schema
+        xmlns="https://api-platform.com/schema/metadata"
         xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+        targetNamespace="https://api-platform.com/schema/metadata"
         elementFormDefault="qualified">
 
     <xsd:element name="resources" type="resources"/>
--- a/tests/Fixtures/FileConfigurations/propertyinvalid.xml
+++ b/tests/Fixtures/FileConfigurations/propertyinvalid.xml
@@ -1,6 +1,9 @@
 <?xml version="1.0" ?>
 
-<resources>
+<resources xmlns="https://api-platform.com/schema/metadata"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="https://api-platform.com/schema/metadata
+           https://api-platform.com/schema/metadata/metadata-2.0.xsd">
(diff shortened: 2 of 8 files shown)


scollado:
I did not commit updated fixtures as the XSD is not yet available from a public URL, I'll update the PR when the final location will be chosen (or is this not required ? I tried locally by defining a valid internal URL for the xsd file).

dunglas:
Great. But as with Symfony, it should not rely on nor request an external server in production. So you can add fixtures without the file being publicly available, it will be a good test that it works.

scollado:
@dunglas Updated fixtures & exception message regexp in tests to include proposed namespace (after messing up with my git remotes while updating 😕 ).
Forking the documentation right-now to prepare a PR including the changes.

dunglas:
Thank you @scollado

dunglas:
This PR breaks tests: https://travis-ci.org/api-platform/core/jobs/226752482

Do you have an idea of the problem @scollado?

soyuka:
ping @scollado would you be able to patch the documentation about XML declarations there?

soyuka:
@dunglas fixed, it was only an issue on master where we had more tests on XML than on 2.0.

jakesyl:
As per #1229, does this need to be done to the `filter.xml` file as well?

jakesyl:
Nevermind.  #1229 is fixed.

PR #1088, opened 2017-04-28
Add groups to subresource

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

Just to prepare documentation needed to check if it actually worked with groups on subresources (it does \o/).

### DIFF
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -173,7 +173,7 @@ final class ApiLoader extends Loader
                     '_controller' => self::DEFAULT_ACTION_PATTERN.'get_subresource',
                     '_format' => null,
                     '_api_resource_class' => $subresource,
-                    '_api_subresource_operation_name' => 'get',
+                    '_api_subresource_operation_name' => 'get_subresource_'.$operation['property'],
                     '_api_subresource_context' => [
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -173,7 +173,7 @@ private function computeSubresourceOperations(RouteCollection $routeCollection,
                     '_controller' => self::DEFAULT_ACTION_PATTERN.'get_subresource',
                     '_format' => null,
                     '_api_resource_class' => $subresource,
-                    '_api_subresource_operation_name' => 'get',
+                    '_api_subresource_operation_name' => 'get_'.$operation['property'],
```
soyuka: We need to agree on this because it sets the operation name to override when we want to add a `normalization_context`. 

Maybe `get_subresource_$property`? WDYT ? ping @teohhanhui @dunglas

Simperfit: `get_subresource_$property` feels right to me

dunglas:
Thank you @soyuka

Issue #1082, opened 2017-04-26
Add a filter to pass dynamic serialization groups in the URL


dunglas:
`http://example.com/books?groups=group1,group2` should add the passed groups to the list of applied groups. The groups configured tru annotations should be preserved or not depending of a configuration flag.

The filter should **not** be Doctrine dependent. It must implement the `FilterInterface` (to appear in the documentation) and decorate the `SerializerContextBuilder` to apply groups.

antograssiot:
I love this feature, we implemented something close to that in our app but we also have a `Group`   entity si wé décider to go with `normalization_groups` as parameter name.

teohhanhui:
I don't like this. This reeks of implementation details being leaked.

meyerbaptiste:
Can be closed.

Issue #1083, opened 2017-04-26
Add a filter to select specific fields


dunglas:
`http://example.com/books?fields=title,description` should serialize only the specified properties.

The filter should **not** be Doctrine dependent. It must implement the `FilterInterface` (to appear in the documentation) and decorate the `SerializerContextBuilder` to use [the new `attributes` context option](https://github.com/symfony/symfony/pull/18834) available in Symfony 3.3.

This filter should throw an exception if the Serializer version is < to 3.3.

meyerbaptiste:
How does this feature deal with serialization groups @dunglas?

dunglas:
In the Symfony serializer, groups are applied first, then fields are applied. A field cannot be serialized if it is hidden by a group. I think we respect this behavior.

bwegrzyn:
Would this work with nested properties? For example `?fields=user.email`

dunglas:
Yes, the Symfony serializer supports this.

mitalcoi:
Great feature, but maybe we should to avoid using key names without "_" prefix?
It will be more safe design:
```http://example.com/books?_fields=title,description```
It is also related to #1082  
```http://example.com/books?_groups=group1,group2```

dunglas:
Existing keys such as page are not prefixed. We cannot use this convention now, it would not be consistent. However all keys are already configurable. It will be the same for this one.

Simperfit:
This has been merged

erikkubica:
Sorry for being stupid, i am searching for like a hour if this is possible accross related issues, but still don´t know if its possible. 

I make a get request /api/schedules?fields[]=id&fields[]=form&fields[]=status&fields[]=sent and it returns all the fields. I have tried using _field, field, _property, property, _properties, properties none of them worked.

The situation is, that for listing I don´t need to request all fields/properties, and as result I end up pulling 10 times more data because it also contains XtoY relationships which are connected using groups, since I need them when I pull a single entity.

There would be no issues if i can just /api/schedules?groups[]=schedules.list (or the example above, but groups seems to be more effective) and i will only pull fields that I need.

PR #1090, opened 2017-04-29
Add dynamic serialization groups filter

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | NO
| Fixed tickets | #1082
| License       | MIT
| Doc PR        | n/a

Should this filter be enabled per resource or will act on a global scope?
Should we add a way to disable it globally?

If we want it to show up on swagger it needs to be enabled in the resource filters (`src/Swagger/Serializer/DocumentationNormalizer.php+:588`). Do we want this in swagger?

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -113,6 +113,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.exception_to_status', $config['exception_to_status']);
         $container->setParameter('api_platform.formats', $formats);
         $container->setParameter('api_platform.error_formats', $errorFormats);
+        $container->setParameter('api_platform.group_parameter_name', $config['group_parameter_name']);
         $container->setParameter('api_platform.eager_loading.enabled', $config['eager_loading']['enabled']);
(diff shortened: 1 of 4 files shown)


soyuka:
Because #1083 works the same, when I have all the elements from above it'll be a matter of minutes to set it up ;).

antograssiot:
I would allow it per resource only. Having a whitelist of additional allowed groups per resource would also make sense IMO to still be alllowed to control the range of exposed properties.

meyerbaptiste:
I had started working on this subject too (#1082, #1083), I have a draft [here](https://github.com/api-platform/core/compare/master...meyerbaptiste:add_group_filter).
I haven't tested anything yet but I know it does not work because of a circular dependency from the `FilterCollection` class.
So with @dunglas we found a solution by refactoring the `FilterCollection` class with an implementation of PSR11. I have the code (not finished) on my business laptop only.

soyuka:
>  I haven't tested anything yet but I know it does not work because of a circular dependency from the FilterCollection class.

Oh, this implementation here works fine, how did you get a circular dependency?

I like the `overrideDefaultGroups` thing. Though, why didn't you simply decorate the `SerializerContextBuilder`?

soyuka:
Oh okay I get it, this is because doctrine calls the filter stack. I'm closing this one and will wait for your PR!

PR #1099, opened 2017-05-04
Add filter locator and deprecate filter collection

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

First step before #1082 and #1083, we have to refactor the `FilterCollection` to avoid circular references with new filters.
Our solution is to use the PSR11 and a service locator. Thanks to http://symfony.com/blog/new-in-symfony-3-3-service-locators.


src/Api/FilterCollectionFactory.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function createFilterCollectionFromLocator(ContainerInterface $filterLocator): FilterCollection
+    {
+        $filters = [];
+
+        foreach ($this->filters as $filter) {
+            try {
+                $filters[$filter] = $filterLocator->get($filter);
+            } catch (\Throwable $e) {
+            }
```
soyuka: Shouldn't we use `if ($filterLocator->has($filter)) {}` instead of catching?

src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
```diff
@@ -27,13 +27,15 @@
  */
 final class FilterExtension implements QueryCollectionExtensionInterface
 {
+    use FilterLocatorTrait;
+
     private $resourceMetadataFactory;
-    private $filters;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, FilterCollection $filters)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, $filterLocator)
```
soyuka: There is no such typing as `ContainerInterface | FilterCollection` in php right? Maybe we could document this new argument though.

src/Metadata/Extractor/YamlExtractor.php
```diff
@@ -32,7 +32,7 @@
     protected function extractPath(string $path)
     {
         try {
-            $resourcesYaml = Yaml::parse(file_get_contents($path));
+            $resourcesYaml = Yaml::parse(file_get_contents($path), Yaml::PARSE_KEYS_AS_STRINGS);
```
soyuka: Hmm why this change? Did you hit a bug? This might need to be backported to 2.0 if it's the case.

src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
soyuka: same goes in the other classes with the same change

meyerbaptiste: This works only for try/catch with PHP 7.1. I will add some PHPDoc.

src/Metadata/Extractor/YamlExtractor.php
meyerbaptiste: I have deprecation warnings from the component without this change and tests fail...

soyuka: oh okay this is because symfony 3.3 then alright! Thanks for the information.

meyerbaptiste: https://github.com/symfony/symfony/pull/21774

src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
meyerbaptiste: I just made the change. Is it good for you like this?

soyuka: :+1: I hope that phpstan is fine with it and yeah this is better for future developers IMO. Thanks!

meyerbaptiste:
Tests fail because the master branch fails too.

src/Bridge/Symfony/Bundle/Resources/config/filter.xml
```diff
@@ -0,0 +1,21 @@
...
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="api_platform.filter_locator" class="Symfony\Component\DependencyInjection\ServiceLocator" public="false">
+            <tag name="container.service_locator" />
+        </service>
+
+        <service id="api_platform.filter_collection_factory" class="ApiPlatform\Core\Api\FilterCollectionFactory" public="false" />
+
+        <service id="api_platform.filters" class="ApiPlatform\Core\Api\FilterCollection" public="false">
```
meyerbaptiste: Oh, nice, I missed this feature! Thanks!

soyuka:
> Tests fail because the master branch fails too.

Indeed, the `Accept` header isn't changing the format. [This](https://github.com/api-platform/core/blob/master/features/doctrine/search_filter.feature#L10) returns JSON-LD. It's weird, everything was green, then I just did `composer update` and it now fails.

Investigating but maybe @dunglas has a clue on this?

composer.json
```diff
@@ -23,7 +23,7 @@
         "symfony/property-access": "^2.7 || ^3.0",
         "symfony/property-info": "^3.3@beta",
         "symfony/serializer": "^3.3@beta",
-        "willdurand/negotiation": "^2.0.3"
+        "willdurand/negotiation": "~2.2.0"
```
soyuka: nice one maybe we can patch this in a separated pr?

meyerbaptiste:
Ok, I found it. The problem comes from the new release of `willdurand/negotiation` (2.3.0) and this commit: https://github.com/willdurand/Negotiation/pull/92.

composer.json
dunglas: Should be `~2.2.3` right?

meyerbaptiste: What about `>=2.0.3 <2.3.0`?

teohhanhui: `^2.0.3 <2.3`

meyerbaptiste:
Does anyone have an idea to fix/avoid that?
```
Generating code coverage report in PHP format ... done
Remaining deprecation notices (1)
The ApiPlatform\Core\Api\FilterCollection class is deprecated since version 2.1 and will be removed in 3.0. Implement the Psr\Container\ContainerInterface interface as service locator instead: 1x
    1x in CodeCoverage::start from SebastianBergmann\CodeCoverage
```

chalasr:
@meyerbaptiste CodeCoverage loads classes which causes the FilterCollection deprecation to be triggered. What you need is to move the deprecation to the constructor:

```diff
diff --git a/src/Api/FilterCollection.php b/src/Api/FilterCollection.php
index 32b1157..76af317 100644
--- a/src/Api/FilterCollection.php
+++ b/src/Api/FilterCollection.php
@@ -15,8 +15,6 @@ namespace ApiPlatform\Core\Api;
 
 use Psr\Container\ContainerInterface;
 
-@trigger_error(sprintf('The %s class is deprecated since version 2.1 and will be removed in 3.0. Implement the %s interface as service locator instead.', FilterCollection::class, ContainerInterface::class), E_USER_DEPRECATED);
-
 /**
  * A list of filters.
  *
@@ -26,4 +24,10 @@ use Psr\Container\ContainerInterface;
  */
 final class FilterCollection extends \ArrayObject
 {
+    public function __construct($input = [], $flags = 0, $iterator_class = \ArrayIterator::class)
+    {
+        @trigger_error(sprintf('The %s class is deprecated since version 2.1 and will be removed in 3.0. Implement the %s interface as service locator instead.', self::class, ContainerInterface::class), E_USER_DEPRECATED);
+
+        parent::__construct($input, $flags, $iterator_class);
+    }
 }

```

soyuka:
Also you can mark specific tests as `@group legacy`  and use `@expectedDeprecation`

meyerbaptiste:
It's what I made @soyuka but the problem here comes from coverage and not tests themselves.
But I think I will follow the @chalasr's suggestion and move my `trigger_error()` inside my controller to avoid the deprecation notice during the coverage.

chalasr:
@soyuka comment made me think that maybe moving the `@group legacy`+`@expectedDeprecation` from the test methods to the top of the test class would solve it, unfortunately it does not. Either you have to trigger deprecations from constructors in order to don't have it triggered before construction, or the whole class should be excluded from the coverage, which I think is not a good idea as the class is still usable and well tested.

src/Api/IdentifiersExtractor.php
```diff
@@ -76,6 +76,7 @@ public function getIdentifiersFromItem($item): array
                 throw new RuntimeException(sprintf('No identifier found in "%s" through relation "%s" of "%s" used as identifier', $relatedResourceClass, $propertyName, $resourceClass));
             }
         }
+
```
soyuka: Was fixed by https://github.com/api-platform/core/commit/7dfdcb1715210f12160919d92975e15d8da7e81f

features/main/crud.feature
```diff
@@ -123,73 +123,67 @@ Feature: Create-Retrieve-Update-Delete
...
         "@type": "hydra:IriTemplate",
-        "hydra:template": "/dummies{?id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,relatedDummies.name,order[id],order[name],order[relatedDummy.symfony],dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],dummyBoolean,dummyFloat,dummyPrice,description[exists],relatedDummy.name[exists],dummyBoolean[exists]}",
+        "hydra:template": "\/dummies{?dummyBoolean,dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],description[exists],relatedDummy.name[exists],dummyBoolean[exists],dummyFloat,dummyPrice,order[id],order[name],order[relatedDummy.symfony],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,relatedDummies.name}",
```
soyuka: Not sure what went wrong here? If this is due to random sorting maybe we should sort everything to avoid such issues in the feature?

dunglas: (we discussed this IRL with @meyerbaptiste).

Before: filters were in the same order than declared in `services.yml`
Now: filters are in the order the configuration in the resource.

It's not a BC break (the JSON parser should not depend of the order).
I prefer to not sort the filters, it allows to detect when we change the output, it's a desirable behavior.

src/Api/FilterCollection.php
```diff
@@ -13,11 +13,21 @@
...
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @deprecated since version 2.1, to be removed in 3.0. Use a service locator {@see \Psr\Container\ContainerInterface}.
  */
 final class FilterCollection extends \ArrayObject
 {
+    public function __construct($input = [], $flags = 0, $iterator_class = 'ArrayIterator')
+    {
+        @trigger_error(sprintf('The %s class is deprecated since version 2.1 and will be removed in 3.0. Implement the %s interface as service locator instead.', self::class, ContainerInterface::class), E_USER_DEPRECATED);
```
dunglas: I would say `Provide and implementation of "%s" instead`.

src/Api/FilterCollectionFactory.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Baptiste Meyer <baptiste.meyer@gmail.com>
+ *
+ * @internal
+ */
+class FilterCollectionFactory
+{
+    private $filtersIds;
+
+    public function __construct(string ...$filtersIds)
```
dunglas: Why not using an array?

meyerbaptiste: Just a hack to have a typed array...

dunglas: I prefer to use a raw array and use `string[]` in the PHPDoc, PHPStan will be able to check types using the PHPDoc.

meyerbaptiste: Bad day for variadic functions 😭

theofidry: if you really want a check you could also force it _after_ in the code:

```php
$filterIds = (function (string ...$filterIds) { return $filterIds; })(...$filterIds);
```

Or use a lib like https://github.com/beberlei/assert

dunglas: We don't have to enforce this. The PHPDoc is the contract and it's an `@internal` class.

theofidry: I'm just giving alternatives to enforce it, not saying you have to enforce it :P

meyerbaptiste:
All comments are fixed.

dunglas:
Thanks @meyerbaptiste

teohhanhui:
Unfortunately, this must be reverted because bumping the required versions in composer.json is a major BC break. It means we suddenly drop support for a lot of people... :disappointed:

dunglas:
@teohhanhui it's not, its allowed by semver (and we do this everytime in Symfony). Note that we don't force to upgrade to new major versions, just to the last minor.

teohhanhui:
But if we're thinking in semver, 2.1 is not a major version. It's a minor version.

dunglas:
Semver is about code API. Not about dependencies. Every major project (including Symfony, Doctrine...) do this and this is totally acceptable. It's also the only way to no being stuck with very old and versions.
It's not a big deal for users to upgrade to a new minor version of a project. If they use the dependency themselves, it must be explicitly defined in their `composer.json` to prevent problems.

teohhanhui:
I don't know if we'll ever reach agreement about the technicality of whether bumping dependencies is a BC break, but in the *spirit* of semver, it'd be best if we can tag a major release when we drop support for an older *minor* version of Symfony.

In any case, [stability flags](https://getcomposer.org/doc/04-schema.md#package-links) only work at the project root level. So when we use `@beta` in `require`, that does nothing and will result in unfulfillable dependencies (because Symfony 3.3 is not released yet).

dunglas:
You're right about stability flags, but it's just a temporary solution. We'll not release 2.1 before the release of Symfony 3.3.

teohhanhui:
Looks like you're right about semver:

http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

:smile:

PR #1100, opened 2017-05-04
Disable eager loading partial fetch by default #1069

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1069 
| License       | MIT
| Doc PR        | na

As discussed in #1069 sometimes you still want to fetch a property, even though it will not be serialized. This new metadata attribute adds a flag that will prevent the eager loader to check the `readable` state. 
ping @bwegrzyn

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -44,13 +44,17 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
     private $serializerContextBuilder;
     private $requestStack;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null)
+    /**
+     * @TODO move $fetchPartial after $forceEager (@soyuka) in 3.0
+     */
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null, bool $fetchPartial = false)
     {
         $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
         $this->propertyMetadataFactory = $propertyMetadataFactory;
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->maxJoins = $maxJoins;
         $this->forceEager = $forceEager;
+        $this->fetchPartial = $fetchPartial;
         $this->serializerContextBuilder = $serializerContextBuilder;
         $this->requestStack = $requestStack;
     }
@@ -67,10 +71,11 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
         }
 
         $forceEager = $this->shouldOperationForceEager($resourceClass, $options);
+        $fetchPartial = $this->shouldOperationFetchPartial($resourceClass, $options);
 
         $groups = $this->getSerializerGroups($resourceClass, $options, 'normalization_context');
 
-        $this->joinRelations($queryBuilder, $queryNameGenerator, $resourceClass, $forceEager, $queryBuilder->getRootAliases()[0], $groups);
+        $this->joinRelations($queryBuilder, $queryNameGenerator, $resourceClass, $forceEager, $fetchPartial, $queryBuilder->getRootAliases()[0], $groups);
     }
 
     /**
@@ -86,6 +91,7 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
         }
 
         $forceEager = $this->shouldOperationForceEager($resourceClass, $options);
+        $fetchPartial = $this->shouldOperationFetchPartial($resourceClass, $options);
 
         if (isset($context['groups'])) {
             $groups = ['serializer_groups' => $context['groups']];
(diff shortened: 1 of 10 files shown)


teohhanhui:
I think we should just use an attribute, since this is a "hint" for the eager loading extension. We do not want to be adding metadata fields that are not well thought out.

soyuka:
hmm yeah you're right, I'm changing this thanks! To me this is not a bug though, more a feature.

bwegrzyn:
Looks good to me! :+1:

soyuka:
Wow about the tests, as @meyerbaptiste said master is broken. At about 14h gmt + 1 is was okay :stuck_out_tongue:

teohhanhui:
Partial loading is dangerous because it breaks entities. It must definitely be opt-in. Having partial loading on by default is thus a bug.

teohhanhui:
I don't believe that this needs to be configurable. We should instead make sure that partial loading is only activated on an opt-in basis.

src/Metadata/Property/PropertyMetadata.php
```diff
@@ -306,6 +306,23 @@ public function withAttributes(array $attributes): self
...
+    public function getAttribute(string $key, $defaultValue = null)
+    {
+        if (isset($this->attributes[$key])) {
+            return $this->attributes[$key];
+        }
+
+        return $defaultValue;
+    }
+
+    /**
```
soyuka: I should remove this, not used.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -44,13 +44,14 @@
...
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, bool $fetchPartial = false, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null)
     {
         $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
         $this->propertyMetadataFactory = $propertyMetadataFactory;
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->maxJoins = $maxJoins;
         $this->forceEager = $forceEager;
+        $this->fetchPartial = $fetchPartial;
```
soyuka: name is fine @teohhanhui 😛 ?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -50,6 +50,7 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->booleanNode('enabled')->defaultTrue()->info('To enable or disable eager loading')->end()
+                        ->booleanNode('fetch_partial')->defaultFalse()->info('Fetch only partial data according to serialization groups')->end()
```
teohhanhui: I think we should add:

"If enabled, Doctrine ORM entities will **NOT** work as expected if any of the other fields are used."

:stuck_out_tongue_closed_eyes:

soyuka: Done

soyuka:
ping @api-platform/core-team this one should be good to go.

Should we merge this on `2.0` instead? It's not a breaking change but it removes a feature that was enabled by default before. Results won't change but performances may be impacted.

teohhanhui:
This is a bugfix so it should go into `2.0`.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -44,13 +44,14 @@
     private $serializerContextBuilder;
     private $requestStack;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, bool $fetchPartial = false, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null)
```
dunglas: Isn't this a BC break?

teohhanhui: Oops... Another ugly hack required?

p/s: When can we actually break things again? In the Symfony model (deprecations only), the answer seems to be "never, unless you create new interfaces / classes". \o/

soyuka: Kinda yes. I'll try to handle backward compatibility! I thought about setting this new argument as last one, do you think it'd be a solution?

soyuka: @teohhanhui I can also put this as last argument. Then, no ugly hacks nor BC break no?

teohhanhui: Yeah, sure. But then there's still the underlying problem of piling on layers upon layers of bad design (because we don't really have a choice).

soyuka: Yeah sure, when we can break again we will refactor those don't worry :).

soyuka:
LGTM, can someone do a last review? Thanks!

PR #1101, opened 2017-05-04
fix stringifyable identifier

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #924 
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Bridge/Symfony/Routing/IriConverter.php
+++ b/src/Bridge/Symfony/Routing/IriConverter.php
@@ -122,14 +122,17 @@ final class IriConverter implements IriConverterInterface
                 continue;
             }
 
-            $identifiers[$propertyName] = $this->propertyAccessor->getValue($item, $propertyName);
+            $identifier = $identifiers[$propertyName] = $this->propertyAccessor->getValue($item, $propertyName);
 
-            if (!is_object($identifiers[$propertyName])) {
+            if (!is_object($identifier)) {
+                continue;
+            } elseif (method_exists($identifier, '__toString')) {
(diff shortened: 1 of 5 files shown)


src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -126,6 +126,9 @@ private function getIdentifiersFromItem($item): array
 
             if (!is_object($identifiers[$propertyName])) {
                 continue;
+            } else if (method_exists($identifiers[$propertyName], '__toString')) {
```
eps90: If projects keeps PSR-2 rules, there should be `elseif` instead of `else if`

teohhanhui:
We might want to refactor the code a bit, since we keep using
```php
$identifiers[$propertyName]
```
lol...

src/Bridge/Symfony/Routing/IriConverter.php
dunglas: `elseif`

soyuka:
Refactored, fixed, and added a test.

dunglas:
Thanks @soyuka

dkarlovi:
@soyuka awesome, thanks a lot! \o/

teohhanhui:
@soyuka The test is incorrect. The entity's `getId` method should return an object which implements `__toString`.

soyuka:
Hmm, wasn't sure about this one. Should have left a note. I'm opening a new pr to patch this.

PR #1102, opened 2017-05-04
Fix Behat

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

See https://github.com/willdurand/Negotiation/pull/92#issuecomment-299248503

PR #1104, opened 2017-05-08
Upgrade the Negotiation library

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

See willdurand/Negotiation#94

TODO:

- [x] Wait for a new tag

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "symfony/property-access": "^2.7 || ^3.0",
         "symfony/property-info": "^3.1",
         "symfony/serializer": "^3.1",
-        "willdurand/negotiation": "^2.0.3 <2.3"
+        "willdurand/negotiation": "^2.3.1"
     },
     "require-dev": {
         "behat/behat": "^3.1",
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,7 +10,5 @@ parameters:
         - '#Call to an undefined method PHPUnit_Framework_MockObject_MockObject::[a-zA-Z0-9_]+\(\)#'
 
         # False positives
-        - '#scalar#' # Fixed in PHPStan 0.7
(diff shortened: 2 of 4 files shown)


insekticid:
create new tag ASAP, because after composer update you get willdurand/negotiation v 2.3 and after that will work only first output normalizer with same format (e.g. json)

```
api_platform:
    formats:
        jsonld:  ['application/ld+json']
        jsonhal: ['application/hal+json']
```

request:
`
curl -X GET --header 'Accept: application/hal+json' 
`
response is in 'ld+json' format (the first one in config)

soyuka:
We're aware of that @insekticid this is why the version should have been patched in #1102! See the root cause [here](https://github.com/willdurand/Negotiation/pull/92#issuecomment-299248503). We're waiting for [this](https://github.com/willdurand/Negotiation/issues/94) to be merged to fix the issue :).

insekticid:
I understand, but you should add warning to Api Platform website, or create new tag with '"willdurand/negotiation": "^2.0.3 <2.3",' since this will be fixed

meyerbaptiste:
The new tag is here, this PR can be updated!

soyuka:
Hmm we have the same phpstan errors then in #1113

soyuka:
@dunglas could you set up a new tag (also in api-platform/api-platform)? Thanks!

dunglas:
@soyuka why?

soyuka:
Composer should take the last available version of the negotation library? If that's the case, never mind ;).

dunglas:
@soyuka yep. And when doing `composer install`,  it will take an older one.

teohhanhui:
@dunglas Bumping the required version of `willdurand/negotation` is a BC break. :disappointed:

I believe it should be:

```json
"willdurand/negotiation": "^2.0.3 <2.3 || ^2.3.1",
```

dunglas:
I haven't seen this PR was merged. There is no need to do that as a new version of a new version of 
`willdurand/negotiation` has been released. I've opened this PR just to run the test suite with the patched version.
I need to revert this PR.

soyuka:
Yeah sorry my bad, thought it was needed. I've reverted it in #1115

Issue #1107, opened 2017-05-09
Possible conflict between subresource collection and nested subresource collection


teohhanhui:
Given `Tour -[:offers]-> AggregateOffer -[:offers]-> Offer`

`GET /app_dev.php/aggregate_offers/8339/offers.jsonld` failed with:

```
[1] ApiPlatform\Core\Exception\InvalidArgumentException: Unable to generate an IRI for "AppBundle\Entity\Offer".
    at n/a
        in /srv/tripviss-api/vendor/api-platform/core/src/Bridge/Symfony/Routing/IriConverter.php line 133

    at ApiPlatform\Core\Bridge\Symfony\Routing\IriConverter->getSubresourceIriFromResourceClass('AppBundle\\Entity\\Offer', array('id' => '8339'))
        in /srv/tripviss-api/vendor/api-platform/core/src/Hydra/Serializer/CollectionNormalizer.php line 79

    at ApiPlatform\Core\Hydra\Serializer\CollectionNormalizer->normalize(object(Paginator), 'jsonld', array('groups' => array('offer_read'), 'subresource_operation_name' => 'get_subresource_offers', 'operation_type' => 'subresource', 'resource_class' => 'AppBundle\\Entity\\Offer', 'request_uri' => '/app_dev.php/aggregate_offers/8339/offers.jsonld', 'subresource_identifiers' => array('id' => '8339'), 'subresource_resources' => array('AppBundle\Entity\AggregateOffer' => array('id' => '8339')), 'jsonld_has_context' => true, 'api_sub_level' => true))
        in /srv/tripviss-api/vendor/api-platform/core/src/Hydra/Serializer/PartialCollectionViewNormalizer.php line 48

    at ApiPlatform\Core\Hydra\Serializer\PartialCollectionViewNormalizer->normalize(object(Paginator), 'jsonld', array('groups' => array('offer_read'), 'subresource_operation_name' => 'get_subresource_offers', 'operation_type' => 'subresource', 'resource_class' => 'AppBundle\\Entity\\Offer', 'request_uri' => '/app_dev.php/aggregate_offers/8339/offers.jsonld', 'subresource_identifiers' => array('id' => '8339'), 'subresource_resources' => array('AppBundle\Entity\AggregateOffer' => array('id' => '8339'))))
        in /srv/tripviss-api/vendor/api-platform/core/src/Hydra/Serializer/CollectionFiltersNormalizer.php line 59

[...]

PR #1110, opened 2017-05-10
Fix route name resolving with subresources

soyuka:
Also resolves conflicts with existing subresources on a property having
the same name

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | behat is okay, unit tests are nogo
| Fixed tickets | #1107 
| License       | MIT
| Doc PR        | ma

Unit tests don't pass for now, waiting for an answer to the below comment.

### DIFF
--- a/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
@@ -37,7 +37,7 @@ final class CachedRouteNameResolver implements RouteNameResolverInterface
     /**
      * {@inheritdoc}
      */
-    public function getRouteName(string $resourceClass, $operationType): string
+    public function getRouteName(string $resourceClass, $operationType /**, array $context = []**/): string
     {
         $cacheKey = self::CACHE_KEY_PREFIX.md5(serialize([$resourceClass, $operationType]));
 
@@ -51,7 +51,14 @@ final class CachedRouteNameResolver implements RouteNameResolverInterface
             // do nothing
         }
 
-        $routeName = $this->decorated->getRouteName($resourceClass, $operationType);
+        if (func_num_args() > 2) {
+            $context = func_get_arg(2);
+        } else {
+            $context = [];
+            @trigger_error(sprintf('Method %s() will have a third `$context = []` argument in version 3.0. Not defining it is deprecated since 2.1.', __METHOD__), E_USER_DEPRECATED);
+        }
+
+        $routeName = $this->decorated->getRouteName($resourceClass, $operationType, $context);
 
         if (!isset($cacheItem)) {
             return $routeName;
--- a/src/Bridge/Symfony/Routing/RouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/RouteNameResolver.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Routing;
 
+use ApiPlatform\Core\Api\OperationType;
 use ApiPlatform\Core\Api\OperationTypeDeprecationHelper;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use Symfony\Component\Routing\RouterInterface;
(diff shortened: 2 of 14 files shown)


src/Bridge/Symfony/Routing/RouteNameResolverInterface.php
```diff
@@ -32,5 +32,5 @@
      *
      * @return string
      */
-    public function getRouteName(string $resourceClass, $operationType): string;
+    public function getRouteName(string $resourceClass, $operationType, array $context = []): string;
```
soyuka: I think that this is a breaking change. Should I avoid changing this interface? I need more informations when retrieving the route name for a subresource because multiple routes may find a wrong match on the same resource class. It has to take into consideration the subresource tree. See `isSameSubresource` in `RouteNameResolver`.

teohhanhui: Yes, this is a BC break.

tests/Fixtures/TestBundle/Entity/DummyTour.php
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Dummy Tour.
+ * https://github.com/api-platform/core/issues/1107.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @ApiResource
+ * @ORM\Entity
+ */
+class DummyTour
```
teohhanhui: I suggest renaming this to `DummyProduct` (`Tour` is from my own project), since the `offers` property is defined by https://schema.org/offers

teohhanhui: Or we might want to take a page out of how Doctrine keeps test cases. For example: https://github.com/doctrine/doctrine2/tree/v2.5.6/tests/Doctrine/Tests/Models/DDC3699

soyuka:
I've done some dirty fix to avoid breaking, let me know your thoughts or if you have another idea on how to do this. 
Obviously, phpstan won't be happy with this:

```
 ------ ----------------------------------------------------------------------------------
  Line   tests/Bridge/Symfony/Routing/RouteNameResolverTest.php
 ------ ----------------------------------------------------------------------------------
  135    Method ApiPlatform\Core\Bridge\Symfony\Routing\RouteNameResolver::getRouteName()
         invoked with 3 parameters, 2 required.
 ------ ----------------------------------------------------------------------------------
```

teohhanhui:
Hehe... We can ignore the PHPstan error for now. :smile:

soyuka:
ping @dunglas is also good to merge if you agree with the routename patch!

src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
```diff
@@ -51,7 +51,7 @@ public function getRouteName(string $resourceClass, $operationType): string
             // do nothing
         }
 
-        $routeName = $this->decorated->getRouteName($resourceClass, $operationType);
+        $routeName = $this->decorated->getRouteName($resourceClass, $operationType, @func_get_arg(2));
```
dunglas: Can you do it like that instead https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Serializer/Serializer.php#L182-L190

src/Bridge/Symfony/Routing/RouteNameResolver.php
```diff
@@ -34,8 +35,14 @@ public function __construct(RouterInterface $router)
     /**
      * {@inheritdoc}
      */
-    public function getRouteName(string $resourceClass, $operationType): string
+    public function getRouteName(string $resourceClass, $operationType /**, array $context = [] **/): string
     {
+        $context = @func_get_arg(2);
```
dunglas: Same here

soyuka: done thanks

dunglas:
Thanks @soyuka

PR #1111, opened 2017-05-10
Add a filter for serialization groups

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1082 
| License       | MIT
| Doc PR        | N/A


src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,90 @@
...
+namespace ApiPlatform\Core\Serializer\Filter;
+
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Group filter.
+ *
+ * @author Baptiste Meyer <baptiste@les-tilleuls.coop>
+ */
+class GroupFilter implements FilterInterface
```
dunglas: final

dunglas: `final`

src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    protected $overrideDefaultGroups;
```
dunglas: `private`

src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->overrideDefaultGroups = $overrideDefaultGroups;
+        $this->parameterName = $parameterName;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(Request $request, bool $normalization, array $attributes, array &$context)
+    {
+        if (empty($groups = $this->extractGroups($request))) {
```
dunglas: `empty` is useless.

src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,90 @@
...
+            return;
+        }
+
+        if (!$this->overrideDefaultGroups && !empty($context['groups'])) {
```
dunglas: Same here

src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Extracts groups from a request.
+     *
+     * @param Request $request
+     *
+     * @return array
+     */
+    protected function extractGroups(Request $request): array
```
dunglas: `private`

src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $groups = [];
+
+        if (null !== $groupsParameter = $request->query->get($this->parameterName)) {
+            if (is_string($groupsParameter)) {
```
dunglas: Why supporting both syntax? We only support the second one in other filters.

src/Serializer/SerializerContextBuilderFilter.php
```diff
@@ -0,0 +1,76 @@
...
+                $attributes['collection_operation_name'] ?? $attributes['subresource_operation_name'],
+                'filters',
+                [],
+                true
+            );
+        } else {
+            $resourceFilters = $resourceMetadata->getItemOperationAttribute($attributes['item_operation_name'], 'filters', [], true);
+        }
+
+        if (empty($resourceFilters)) {
```
dunglas: `empty` is useless

src/Serializer/Filter/GroupFilter.php
soyuka: I don't think we should `explode` or do any transformation on query parameters. Just let symfony do the work here.

I'd propose:

```php
$groups = $request->query->get($this->parameterName);
```

This method can then be removed as this one line should be enough.

src/Serializer/Filter/GroupFilter.php
meyerbaptiste: I followed the same standards as ORM filters: not final + protected methods/attributes.

src/Serializer/Filter/GroupFilter.php
meyerbaptiste: I followed what you described in your issue: `?groups=foo,bar,baz` and I added the array syntax.

dunglas: After a second though, I was wrong :) We should just support the array syntax, just like for other fields (consistency...).

src/Serializer/Filter/GroupFilter.php
meyerbaptiste: So @dunglas, should we keep consistency with ORM filters or should I follow your requested changes?

dunglas: ORM filters support only the `array` syntax. So, same for this new one.

meyerbaptiste: No, I talk about `final` + `protected` @dunglas 😋

src/Serializer/Filter/GroupFilter.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(Request $request, bool $normalization, array $attributes, array &$context)
+    {
+        if (!is_array($groups = $request->query->get($this->parameterName))) {
+            return;
+        }
+
+        if (!$this->overrideDefaultGroups && isset($context['groups']) && is_array($context['groups'])) {
```
meyerbaptiste: About `is_array($context['groups'])`, I'm not really sure how we should handle the case where `$context['groups']` exists and is not an array:
* do not apply the filter when `$overrideDefaultGroups` is true or false
* apply the filter when `$overrideDefaultGroups` is true
* apply the filter and overwrite the value of `$context['groups']` in all cases (current behavior)

WDYT?

soyuka: What is `$context['groups']` if not an array? I think we should cast is as one.

meyerbaptiste: > What is $context['groups'] if not an array?

I don't know, just an assumption!

soyuka: Should be a string, therefore we could cast it as array.

src/Serializer/SerializerContextBuilderFilter.php
```diff
@@ -0,0 +1,76 @@
...
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * {@inheritdoc}
+ *
+ * @author Baptiste Meyer <baptiste.meyer@gmail.com>
+ */
+final class SerializerContextBuilderFilter implements SerializerContextBuilderInterface
```
dunglas: `SerializerFilterContextBuilder`?

teohhanhui:
To restate what I've raised in https://github.com/api-platform/core/issues/1082#issuecomment-299400557, I think we should not introduce misfeatures like this...

teohhanhui:
If we are to have a "property groups" filter, then the "property groups" MUST be exposed as resources. That's the way to keep things as hypermedia.

We need to always keep these principles in mind, as we claim that API Platform is a hypermedia framework / library. :smile:

dunglas:
@teohhanhui I'm not sure to get exactly what you have in mind? Can you give us some detail?
For me, this filter is a purely technical feature to save bandwidth. It's one of the most wanted  filter of our customers and I often see projects having this kind of filters (as custom filters).

It's also a 100% optin filter.

dunglas:
Regarding the implementation details leak, this is not true. As stated just before, filter is 100% optional and must be enabled manually by the developper. When doing this, you know that the groups will be exposed publicly, and you need to take care of this.

teohhanhui:
Yeah, I guess it must be very hard to expose the property groups as resources.

But I wish we could change "groups" to "property_groups"?

teohhanhui:
When I said it's a leak of implementation detail, I meant that these groups are not defined anywhere. They are not exposed as resources, they are not in the Swagger doc. So they're magical as far as the client is concerned... Out-of-band information which is a big nono.

dunglas:
I agree about the out of band information. However this is not the case the `fields` filter, so it should be privileged when developing hypermedia API.

About the name, I prefer `groups` (shorter is better), but it should be configurable.

meyerbaptiste:
Merged with #1123, see https://github.com/api-platform/core/commit/e3c17c885377d73bb94bd6a728b039080b7d31f7.

PR #1113, opened 2017-05-15
Move the metadata from validator factory to the validator.xml file

sroze:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ø
| License       | MIT
| Doc PR        | ø

As the validator as a dev dependency, it won't necessary be installed when using ApiPlatform's "pack" and Symfony Flex. Still, the validator is a dependency to get the metadata from its component: moving this definition to the `validator.xml` file that is only included when the validator component exists fixes the problem.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
@@ -64,11 +64,6 @@
             <argument type="service" id="api_platform.metadata.property.metadata_factory.serializer.inner" />
         </service>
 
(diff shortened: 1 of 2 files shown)


soyuka:
Hmm phpstan errors are not related. I'll fix those in another patch. They may be the consequence of a phpstan upgrade.

soyuka:
@sroze you can rebase if you want, should fix phpstan :)

meyerbaptiste:
IMO you should change the base branch to `2.0`.

dunglas:
Btw, the API pack includes the validator: https://github.com/api-platform/api-pack/blob/master/composer.json#L17

Anyway this PR must be merged.

sroze:
Rebased and targeted on 2.0 😉

dunglas:
Thank you @sroze!

PR #1115, opened 2017-05-15
Do not restrict negotiation library to 2.3.1

soyuka:
ping @teohhanhui

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "symfony/property-access": "^2.7 || ^3.0",
         "symfony/property-info": "^3.1",
         "symfony/serializer": "^3.1",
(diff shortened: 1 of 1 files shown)


teohhanhui:
Needs rebase though... \o/

dunglas:
I would just let `^2.0.3`, we'll not manually exclude all buggy patch versions of our deps.

soyuka:
@dunglas agreed, so the patch was not needed :p. Rebased.

teohhanhui:
@soyuka Still needs a rebase. :laughing:

dunglas:
Thanks @soyuka.
@teohhanhui it's ok, I've just merged another PR in the meantime :)

Issue #1033, opened 2017-04-06
Proposition: using annotation for filters


pierre-H:
Thank you again for your great work !

It would be easier to add a filter in the entity with annotations instead of adding a service.
For example : 
```php
<?php

// src/AppBundle/Entity/Offer.php

namespace AppBundle\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use ApiPlatform\Core\Annotation\ApiFilter;          // for example

/**
 * @ApiResource()
 */
class Offer
{
    /**
     * @ApiFilter('type="search_filter", precision="exact")
     */
    private $id;

    /**
     * @ApiFilter('type="date_filter")
     */
    private $date;

    ...
}

```

The goal here is to simplify the use of filters with Api-Platform, and I think it's important because the great success of Api-Platform is it's simplicity.

soyuka:
I think this is a good idea. We should wait for @api-platform/core-team approval first, but would you ike to work on a PR for this feature?

dunglas:
I like the idea, may I suggest an alternative syntax?

```php
<?php

// src/AppBundle/Entity/Offer.php

namespace AppBundle\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use ApiPlatform\Core\Annotation\ApiFilter;          // for example
use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\SearchFilter;
ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\DateFilter;

/**
 * @ApiResource
 */
class Offer
{
    /**
     * @ApiFilter(SearchFilter::class, strategy="exact")
     */
    private $id;

    /**
     * @ApiFilter(DateFilter::class)
     */
    private $date;

    ...
}
```

I would also allow to add the `@ApiFilter` on the class to enable the filter for all properties.

vincentchalamon:
I would love to see this feature! But we should also think about filters applied to multiple fields (i.e. `search_filter`)

dunglas:
We'll need a bridge with the Symfony DI component:

1. collect all annotations for the same filter in a given class
2. create a service extending the `abstract` one with all properties of this class configured

soyuka:
```
     * @ApiFilter(SearchFilter::class, strategy="exact")
```

This is really neat, it'll help using custom filters too :).

> I would also allow to add the @ApiFilter on the class to enable the filter for all properties.

[...]

PR #1117, opened 2017-05-16
Implement Annotation filters @ApiFilter

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1033 
| License       | MIT
| Doc PR        | n/a

- [x] add tests

Please, let me know if I'm on the right track or if I should have done some things differently :).

Docs:

```
 * @ApiFilter(DateFilter::class, strategy=DateFilter::EXCLUDE_NULL)
 * @ApiFilter(BooleanFilter::class)
 * @ApiFilter(PropertyFilter::class, arguments={"parameterName"="foobar"})
 * @ApiFilter(SearchFilter::class, properties={"colors.prop": "ipartial"})
 * @ApiFilter(SearchFilter::class, strategy="partial")
```

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -616,6 +616,9 @@ class FeatureContext implements Context, SnippetAcceptingContext
     public function thereIsAFooEntityWithRelatedBars()
     {
         $foo = new DummyCar();
+        $foo->setName('mustli');
+        $foo->setCanSell(true);
+        $foo->setAvailableAt(new \DateTime());
         $this->manager->persist($foo);
 
         $bar1 = new DummyCarColor();
--- a/src/Annotation/ApiFilter.php
+++ b/src/Annotation/ApiFilter.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Annotation;
+
+use ApiPlatform\Core\Api\FilterInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+
+/**
+ * Filter annotation.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @Annotation
+ * @Target({"PROPERTY", "CLASS"})
+ */
+final class ApiFilter
+{
+    /**
+     * @var string
+     */
+    public $strategy;
+
+    /**
+     * @var string
+     */
+    public $filterClass;
+
+    /**
+     * @var array
+     */
+    public $properties = [];
+
+    /**
+     * @var array raw arguments for the filter
+     */
+    public $arguments = [];
+
+    public function __construct($options = [])
+    {
+        if (!isset($options['value'])) {
+            throw new InvalidArgumentException('This annotation needs a value representing the filter class.');
+        }
+
+        if (!is_subclass_of($options['value'], FilterInterface::class)) {
+            throw new InvalidArgumentException(sprintf('The filter class "%s" does not implement "%s".', $options['value'], FilterInterface::class));
+        }
+
+        $this->filterClass = $options['value'];
+        unset($options['value']);
+
+        foreach ($options as $key => $value) {
+            if (!property_exists($this, $key)) {
+                throw new InvalidArgumentException(sprintf('Property "%s" does not exist on the ApiFilter annotation.', $key));
+            }
+
+            $this->$key = $value;
+        }
+    }
+}
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
+
+use ApiPlatform\Core\Util\AnnotationFilterExtractorTrait;
+use ApiPlatform\Core\Util\ReflectionClassRecursiveIterator;
+use Doctrine\Common\Annotations\Reader;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+/**
+ * Injects filters.
+ *
+ * @internal
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class AnnotationFilterPass implements CompilerPassInterface
+{
+    use AnnotationFilterExtractorTrait;
+
+    const TAG_FILTER_NAME = 'api_platform.filter';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $resourceClassDirectories = $container->getParameter('api_platform.resource_class_directories');
+        $reader = $container->get('annotation_reader');
+
+        foreach (ReflectionClassRecursiveIterator::getReflectionClassesFromDirectories($resourceClassDirectories) as $className => $reflectionClass) {
+            $this->createFilterDefinitions($reflectionClass, $reader, $container);
+        }
+    }
+
+    private function createFilterDefinitions(\ReflectionClass $reflectionClass, Reader $reader, ContainerBuilder $container)
+    {
+        foreach ($this->readFilterAnnotations($reflectionClass, $reader) as $id => list($arguments, $filterClass)) {
+            if ($container->hasDefinition($id)) {
+                continue;
+            }
+
+            $definition = new Definition();
+            $definition->setClass($filterClass);
+            $definition->addTag(self::TAG_FILTER_NAME);
+            $definition->setAutowired(true);
+
+            foreach ($arguments as $key => $value) {
+                $definition->setArgument("$$key", $value);
+            }
+
+            $container->setDefinition($id, $definition);
+        }
+    }
+}
--- a/src/Metadata/Resource/Factory/AnnotationResourceFilterMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/AnnotationResourceFilterMetadataFactory.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Resource\Factory;
+
+use ApiPlatform\Core\Annotation\ApiFilter;
+use ApiPlatform\Core\Exception\ResourceClassNotFoundException;
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Util\AnnotationFilterExtractorTrait;
+use Doctrine\Common\Annotations\Reader;
+
+/**
+ * Adds filters to the resource metadata {@see ApiFilter} annotation.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class AnnotationResourceFilterMetadataFactory implements ResourceMetadataFactoryInterface
+{
+    use AnnotationFilterExtractorTrait;
+
+    private $reader;
+    private $decorated;
+
+    public function __construct(Reader $reader, ResourceMetadataFactoryInterface $decorated = null)
+    {
+        $this->reader = $reader;
+        $this->decorated = $decorated;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(string $resourceClass): ResourceMetadata
+    {
+        $parentResourceMetadata = null;
+        if ($this->decorated) {
+            try {
+                $parentResourceMetadata = $this->decorated->create($resourceClass);
+            } catch (ResourceClassNotFoundException $resourceNotFoundException) {
+                // Ignore not found exception from decorated factories
+            }
+        }
+
+        if (null === $parentResourceMetadata) {
+            return $this->handleNotFound($parentResourceMetadata, $resourceClass);
+        }
+
+        try {
+            $reflectionClass = new \ReflectionClass($resourceClass);
+        } catch (\ReflectionException $reflectionException) {
+            return $this->handleNotFound($parentResourceMetadata, $resourceClass);
+        }
+
+        $filters = array_keys($this->readFilterAnnotations($reflectionClass, $this->reader));
+
+        if (!$filters) {
+            return $parentResourceMetadata;
+        }
+
+        $parentFilters = $parentResourceMetadata->getAttribute('filters', []);
+
+        if ($parentFilters) {
+            $filters = array_merge($parentFilters, $filters);
+        }
+
+        $attributes = $parentResourceMetadata->getAttributes();
+
+        if (!$attributes) {
+            $attributes = [];
+        }
+
+        return $parentResourceMetadata->withAttributes(array_merge($attributes, ['filters' => $filters]));
+    }
+
+    /**
+     * Returns the metadata from the decorated factory if available or throws an exception.
+     *
+     * @param ResourceMetadata|null $parentPropertyMetadata
+     * @param string                $resourceClass
+     *
+     * @throws ResourceClassNotFoundException
+     *
+     * @return ResourceMetadata
+     */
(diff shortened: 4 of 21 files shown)


src/Util/ReflectionClassRecursiveIterator.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Gets reflection classes for php files in the given directories.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @internal
+ */
+final class ReflectionClassRecursiveIterator
+{
+    public static function getReflectionClassesFromDirectories(array $directories): array
```
soyuka: naming sucks here

src/Api/FilterInterface.php
```diff
@@ -36,4 +36,13 @@
...
     public function getDescription(string $resourceClass): array;
+
+    /**
+     * setter for properties.
+     *
+     * used in the dependency injection to allow independent constructor arguments.
+     *
+     * @param array $properties
+     */
+    public function setProperties(array $properties);
```
soyuka: Whoops, breaking change. Any ideas on how I could avoid this? I've added this because I create services on the fly in the Dependency injection, and because the constructor arguments aren't predictable I can't do `$def->replaceArgument`.

src/Metadata/Resource/Factory/AnnotationResourceFilterMetadataFactory.php
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        $attributes = $parentResourceMetadata->getAttributes();
+
+        if (!$attributes) {
+            $attributes = [];
+        }
+
+        return $parentResourceMetadata->withAttributes(array_merge($attributes, ['filters' => $filters]));
+    }
```
soyuka: Code duplication with `Compiler/AnnotationFilterPass.php`. Needs more work.

src/Util/FilterGeneratorTrait.php
```diff
@@ -0,0 +1,51 @@
...
+use Doctrine\Common\Inflector\Inflector;
+
+/**
+ * Generates a service id for a generic filter.
+ *
+ * @internal
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+trait FilterGeneratorTrait
```
soyuka: name sux

src/Annotation/ApiFilter.php
```diff
@@ -0,0 +1,60 @@
...
+use ApiPlatform\Core\Api\FilterInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+
+/**
+ * Filter annotation.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @Annotation
+ * @Target({"PROPERTY", "CLASS"})
```
soyuka: Not sure that `METHOD` target's are a good thing here. Also, they add complexity.

src/Api/FilterInterface.php
soyuka: Fixed this by adding a new Interface (should be merged with Api\FilterInterface if we can break this later).

soyuka:
Any idea for relational properties?

```
@ApiFilter(SearchFilter::class, strategy="exact", properties=["rel.bar", "rel.foo"])
```

WDYT?

src/Api/AnnotatedFilterInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Api;
+
+/**
+ * Filters applicable on a resource.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
Simperfit: it is him ?

src/Metadata/Resource/Factory/AnnotationResourceFilterMetadataFactory.php
```diff
@@ -0,0 +1,103 @@
...
+        $parentResourceMetadata = null;
+        if ($this->decorated) {
+            try {
+                $parentResourceMetadata = $this->decorated->create($resourceClass);
+            } catch (ResourceClassNotFoundException $resourceNotFoundException) {
+                // Ignore not found exception from decorated factories
+            }
+        }
+
+        if (!$parentResourceMetadata) {
```
Simperfit: can't we do null === $parentResourceMetadata?

soyuka:
I really don't understand of this coverage calculation works... I've no coverage decrease, even added some, but still lower then before -_-.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -0,0 +1,76 @@
...
+        foreach (ReflectionClassRecursiveIterator::getReflectionClassesFromDirectories($resourceClassDirectories) as $className => $reflectionClass) {
+            $this->createFilterDefinitions($reflectionClass, $reader, $container);
+        }
+    }
+
+    private function createFilterDefinitions(\ReflectionClass $reflectionClass, Reader $reader, ContainerBuilder $container)
+    {
+        foreach ($this->readFilterAnnotations($reflectionClass, $reader) as $id => list($arguments, $filterClass)) {
+            if (!is_subclass_of($filterClass, AnnotatedFilterInterface::class)) {
+                trigger_error(sprintf('The filter "%s" does not implement the "%s" interface. Therefore, we can not create a filter service for this annotation.', $filterClass, AnnotatedFilterInterface::class));
```
meyerbaptiste: Why an `E_USER_NOTICE` here and below?

soyuka: Seems cleaner then throwing because this is not blocking. Do you have suggestions?

src/Util/AnnotationFilterExtractorTrait.php
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * @return array only ApiFilter annotations
+     */
+    private function getFilterAnnotations(array $miscAnnotations)
+    {
+        $annotations = [];
+
+        foreach ($miscAnnotations as $miscAnnotation) {
+            if (get_class($miscAnnotation) === ApiFilter::class) {
+                $annotations[] = $miscAnnotation;
```
meyerbaptiste: `yield $miscAnnotation;` (this method is only used with `foreach`)? 😃

src/Util/ReflectionClassRecursiveIterator.php
```diff
@@ -0,0 +1,63 @@
...
+                $includedFiles[$sourceFile] = true;
+            }
+        }
+
+        $declared = get_declared_classes();
+        foreach ($declared as $className) {
+            $reflectionClass = new \ReflectionClass($className);
+            $sourceFile = $reflectionClass->getFileName();
+            if (isset($includedFiles[$sourceFile])) {
+                $classes[$className] = $reflectionClass;
```
meyerbaptiste: `yield $className => $reflectionClass;`?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
meyerbaptiste: IMO, I think throwing is correct for this case and `continue` is enough for the second case (L59-62).

src/Util/ReflectionClassRecursiveIterator.php
```diff
@@ -0,0 +1,63 @@
...
+namespace ApiPlatform\Core\Util;
+
+/**
+ * Gets reflection classes for php files in the given directories.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @internal
+ */
+final class ReflectionClassRecursiveIterator
```
meyerbaptiste: You should add a private constructor.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
soyuka: Hmm okay but, even if it's not likely to happen, won't the user want to know that we're not adding the filter?

src/Util/AnnotationFilterExtractorTrait.php
soyuka: yeah, is there some typings we may use for such a case (when using `yield`)?

meyerbaptiste: You can use `: \Iterator` or `: iterable` but with PHP7.1+

soyuka:
HOLY COW! Coverage is green! I'm not changing this PR ever again :laughing:

pierre-H:
Will this PR be in the 2.1.0 version ?

soyuka:
@pierre-H I think it'll be in 3.0 directly with the JSON-api PR. 

I'm using it on my own fork and it works great, we just need a bit more time to review and agree on the feature proposal :).

pierre-H:
@soyuka Ok thank you !

soyuka:
ping @meyerbaptiste could you check again, I've fixed the requested changes.

features/bootstrap/FeatureContext.php
```diff
@@ -569,6 +569,9 @@ public function thereIsAFileConfigDummyObject()
     public function thereIsAFooEntityWithRelatedBars()
     {
         $foo = new DummyCar();
+        $foo->setName('mustli');
```
Simperfit: A car called mustli ?

soyuka: so?

meyerbaptiste: Your nickname is Simperfit, so why not Mustli for a car? 😄

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -0,0 +1,76 @@
...
+                continue;
+            }
+
+            $definition = new Definition();
+            $definition->setClass($filterClass);
+            $definition->addTag(self::TAG_FILTER_NAME, ['id' => $id]);
+            $definition->setAutowired(true);
+
+            if ($arguments) {
+                $definition->addMethodCall('configureFilter', [$arguments]);
```
meyerbaptiste: Why not something like this?
```php
for ($arguments as $key => $value) {
    $definition->setArgument('$'.$key, $value);
}
``` 
`AnnotatedFilterInterface` would then become useless.

soyuka: Those would get through the Filter constructor? Seems messy, how do I get back every arguments then? 

This uses the final `Filter::class` as a service (ie `GroupFilter` or `AbstractFilter` or whatever). Not sure we could do this.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -0,0 +1,76 @@
...
+            $this->createFilterDefinitions($reflectionClass, $reader, $container);
+        }
+    }
+
+    private function createFilterDefinitions(\ReflectionClass $reflectionClass, Reader $reader, ContainerBuilder $container)
+    {
+        foreach ($this->readFilterAnnotations($reflectionClass, $reader) as $id => list($arguments, $filterClass)) {
+            if (!is_subclass_of($filterClass, FilterInterface::class)) {
+                throw new InvalidArgumentException(sprintf('The filter "%s" does not implement the "%s" interface. Therefore, we can not create a filter service for this annotation.', $filterClass, FilterInterface::class));
+                continue;
```
dunglas: Can be removed

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -0,0 +1,76 @@
...
+    const TAG_FILTER_NAME = 'api_platform.filter';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $resourceClassDirectories = $container->getParameter('api_platform.resource_class_directories');
+        $reader = $container->get('annotation_reader');
+        $includedFiles = [];
```
dunglas: Unused.

src/Annotation/ApiFilter.php
```diff
@@ -0,0 +1,70 @@
...
+        foreach ($options as $key => $value) {
+            if (!property_exists($this, $key)) {
+                throw new InvalidArgumentException(sprintf('Property "%s" does not exist on the ApiFilter annotation.', $key));
+            }
+
+            $this->$key = $value;
+        }
+    }
+
+    /**
```
dunglas: Properties should be before methods.

src/Util/AnnotationFilterExtractorTrait.php
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+        return $properties;
+    }
+
+    /**
+     * Reads filter annotations from a ReflectionClass.
+     *
+     * @param \ReflectionClass $reflectionClass
```
dunglas: Useless `@param`s, can be removed.

src/Util/AnnotationFilterExtractorTrait.php
```diff
@@ -0,0 +1,146 @@
...
+            if (get_class($miscAnnotation) === ApiFilter::class) {
+                yield $miscAnnotation;
+            }
+        }
+    }
+
+    /**
+     * Given a filter annotation and reflection elements, find out the properties where the filter is applied.
+     *
+     * @param ApiFilter           $filterAnnotation
```
dunglas: Useless `@param`s, can be dropped. Same for `@return`

src/Util/AnnotationFilterExtractorTrait.php
```diff
@@ -0,0 +1,146 @@
...
+     * @param \ReflectionClass    $reflectionClass
+     * @param \ReflectionProperty $reflectionClass
+     *
+     * @return array
+     */
+    private function getFilterProperties(ApiFilter $filterAnnotation, \ReflectionClass $reflectionClass, \ReflectionProperty $reflectionProperty = null)
```
dunglas: Missing return type.

src/Util/AnnotationFilterExtractorTrait.php
```diff
@@ -0,0 +1,146 @@
...
+     * Generates a unique, per-class and per-filter identifier prefixed by `annotated_`.
+     *
+     * @param \ReflectionClass $reflectionClass the reflection class (a ResourceClass)
+     * @param string           $filterClass     the filter class
+     *
+     * @return string
+     */
+    private function generateFilterId(\ReflectionClass $reflectionClass, string $filterClass): string
+    {
+        return 'annotated_'.Inflector::tableize(str_replace('\\', '', $reflectionClass->getName()).(new \ReflectionClass($filterClass))->getShortName());
```
dunglas: Shouldn't you use FQN to avoid conflicts? Can you use `sprintf` too to make the code more readable?

soyuka: You mean FQN on the filter name? 
About `sprintf` there'd be only one substitution so I don't think it'd change much:

```
'annotated_'.Inflector::tableize(str_replace('\\', '', $reflectionClass->getName().(new \ReflectionClass($filterClass))->getName()));

sprintf('annotated_%s, Inflector::tableize(str_replace('\\', '', $reflectionClass->getName().(new \ReflectionClass($filterClass))->getName())));
```

src/Annotation/ApiFilter.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        if (!is_subclass_of($options['value'], FilterInterface::class)) {
+            throw new InvalidArgumentException(sprintf('The filter class "%s" does not implement "%s".', $options['value'], FilterInterface::class));
+        }
+
+        $this->filterClass = $options['value'];
+        unset($options['value']);
+
+        foreach ($options as $key => $value) {
```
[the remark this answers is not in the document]
soyuka: Either this or my test `AnnotationFilterPassTest` is wrong. If I remove this, the filter arguments are empty (see `AnnotationFilterExtractorTrait->readFilterAnnotations`).

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -0,0 +1,68 @@
...
+    private function createFilterDefinitions(\ReflectionClass $reflectionClass, Reader $reader, ContainerBuilder $container)
+    {
+        foreach ($this->readFilterAnnotations($reflectionClass, $reader) as $id => list($arguments, $filterClass)) {
+            if ($container->hasDefinition($id)) {
+                continue;
+            }
+
+            $definition = new Definition();
+            $definition->setClass($filterClass);
+            $definition->addTag(self::TAG_FILTER_NAME, ['id' => $id]);
```
meyerbaptiste: `['id' => $id]` is useless here, the compiler pass takes the service name by default and here it's also `$id`.

src/Util/ReflectionClassRecursiveIterator.php
dunglas: Really necessary? This class is already marked as `@internal`.

soyuka: has already be fixed anyway

soyuka:
<3 thanks for the review guys

bendavies:
remember docs!

sroze:
Nice @soyuka 👏

Simperfit:
That's nice !

PR #1123, opened 2017-05-17
Add a filter for serialization attributes

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1083 
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -24,6 +24,7 @@ use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyFriend;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyGroup;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyOffer;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyProduct;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyProperty;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Node;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Question;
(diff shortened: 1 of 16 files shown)


src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -37,11 +34,6 @@ class ExistsFilter extends AbstractFilter
 {
     const QUERY_PARAMETER_KEY = 'exists';
 
-    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, LoggerInterface $logger = null, array $properties = null)
```
soyuka: I've seen this in other filters I think. Do this need to be part of this commit?

meyerbaptiste: No this is the only filter where there is this useless constructor! I will open another PR.

soyuka: Okay, we have other useless constructors (QueryChecker for example)

meyerbaptiste: Not totally useless, the `QueryChecker`'s constructor is private! 😛

teohhanhui:
Perhaps "properties" instead of "fields" to be consistent in our terminology?

teohhanhui:
But I guess people are more used to "fields"... Maybe we should just keep it this way. :smile:

dunglas:
I'm for `property`. It's why what we use in API Platform (`@ApiProperty`) and it's the Hydra terminology.

meyerbaptiste:
Done.

teohhanhui:
> It's why what we use in API Platform (`@ApiProperty`) and it's the Hydra terminology.

Yes, but what would people be looking for, when they want that feature? "Fields". It might be worth using the term most people are likely most familiar with.

dunglas:
> Yes, but what would people be looking for, when they want that feature? "Fields". It might be worth using the term most people are likely most familiar with.

It's a doc issue. We need to make it easy to find on the website using both terms, but in the code, consistency is the key. Keep it as is.

dunglas:
Good job!!

Issue #1122, opened 2017-05-17
Swagger documentation throws an error if a PATCH action is defined


magarzon:
Swagger documentation throws an error Method PATCH is not supported if there is PATCH action defined.

The issue is added in this commit: https://github.com/api-platform/core/commit/0333f7c764602c2bf07e8e32363a8f5a36d5ac1f

IMHO PATCH should be allowed even if there is nothing to show in the documentation (like it was until this commit), specially if there are requests to add support to JSON PATCH

dunglas:
Definitely a bug. @soyuka this change should be reverted https://github.com/api-platform/core/commit/0333f7c764602c2bf07e8e32363a8f5a36d5ac1f#diff-32d63273c86a7e4c14c3fa6bb4131194R173 WDYT?

Simperfit:
I agree that this change should be reverted.

soyuka:
I think I asked about this when I added the `RuntimeException` :p. Reverted.

PR #1124, opened 2017-05-17
Swagger documentation normalizer revert throw when method is not supp…

soyuka:
…orted

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1122
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -18,7 +18,6 @@ use ApiPlatform\Core\Api\OperationMethodResolverInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
 use ApiPlatform\Core\Documentation\Documentation;
-use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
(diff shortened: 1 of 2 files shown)


src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -171,7 +171,6 @@ private function getPathOperation(string $operationName, array $operation, strin
             case 'DELETE':
                 return $this->updateDeleteOperation($pathOperation, $resourceShortName);
             default:
```
dunglas: this line should be reverted and the `return` statement on the last line added back.

tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -1111,6 +1111,8 @@ public function testNoOperations()
      */
     public function testMethodNotSupported()
```
dunglas: We may want to use custom methods (standard ones like PATCH, but also custom ones like BAN or PURGE).

soyuka: Yeah I skipped the test, I'll add one with custom methods.

tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -1151,6 +1153,65 @@ public function testMethodNotSupported()
...
+            $propertyMetadataFactoryProphecy->reveal(),
+            $resourceClassResolverProphecy->reveal(),
+            $operationMethodResolverProphecy->reveal(),
+            $operationPathResolver,
+            $urlGeneratorProphecy->reveal()
+        );
+
+        $expected = [
+            'swagger' => '2.0',
+            'basePath' => null,
```
soyuka: ping @Simperfit is this correct? I think that this custom method behavior isn't correctly supported no?

dunglas:
The patch looks good to me. Can you add the missing test?

soyuka:
I added the missing test already, but the result seems weird.

soyuka:
Okay, tests is fine after second look. Rebased feel free to merge once green.

dunglas:
Thanks @soyuka!

PR #768, opened 2016-09-27
Use composer for swagger ui

soyuka:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets |  |
| License | MIT |
| Doc PR |  |

PR #1138, opened 2017-05-23
Upgrade Swagger UI to v3 (React inside)

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

* [x] Fix the CSS
* [x] Recreate the system to display the current request
* [x] Recreate the OAuth support


soyuka:
Those diffs :/. Reminds me of the https://github.com/api-platform/core/pull/768 reason.

dunglas:
Yes but the new version improve things, we delete all previous files and there is only 2 JS files now.

soyuka:
Oh yeah that's nice, I didn't saw that on first look. We may be able to reduce those to 1 file in the `update-js.sh` script.

src/Bridge/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig
```diff
@@ -4,44 +4,46 @@
...
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/handlebars-4.0.5.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/lodash.min.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/backbone-min.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/swagger-ui.min.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/highlight.9.1.0.pack.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/jsoneditor.min.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/marked.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/swagger-ui/lib/swagger-oauth.js') }}"></script>
-    <script src="{{ asset('bundles/apiplatform/init-swagger-ui.js') }}"></script>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Source+Code+Pro:300,600|Titillium+Web:400,600,700">
```
meyerbaptiste: Outch, are we obliged?

Simperfit: I think we should import it... we shouln't need internet to load the Ui

teohhanhui: It'd still work when offline, the fallback font would just be used.

teohhanhui:
Why are we digging ourselves a new hole?

As I've mentioned in https://github.com/api-platform/core/issues/1149#issuecomment-305520797, I believe this should live in a separate project.

dunglas:
Well, I'm not against splitting this piece of code in another project later. But for 2.1 it will stay in core.
We need to have a good documentation system builtin installed with the default distribution. It's one on the added value of API Platform.

dunglas:
Btw this PR removes ~ 12k lines of JS and CSS, we're on the good path isn't it?

Issue #1118, opened 2017-05-16
Pagination: Division by zero 


grachevko:
On get collection with itemsPerPage=0 platform crash with exception "Warning: Division by zero"

PR #1141, opened 2017-05-26
hotfix: add an exception for item per page equal 0

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1118
| License       | MIT


grachevko:
What if send negative number? :)

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '-1 OFFSET 0' at line 1"
```

Simperfit:
@soyuka done :)

soyuka:
bug fix no? Shouldn't this target 2.0?

dunglas:
@Simperfit you need to rebase ;)

Issue #1151, opened 2017-06-01
Register services for autoconfiguration (Symfony 3.3 autowiring)


vincentchalamon:
Using Symfony 3.3 autowiring, it's possible to add tags on services by directories. But if some services in the same directory should have different tags, it's not possible to autowire it properly.

Like EventSubscriberInterface, API Platform should register its services for autoconfiguration. For example with data providers:
```php
$container->registerForAutoconfiguration(ItemDataProviderInterface::class)
            ->addTag('api_platform.item_data_provider');
$container->registerForAutoconfiguration(CollectionDataProviderInterface::class)
            ->addTag('api_platform.collection_data_provider');
```
In your opinion, which interfaces should be autoconfigured?:
- ItemDataProviderInterface: api_platform.item_data_provider
- CollectionDataProviderInterface: api_platform.collection_data_provider
- FilterInterface: api_platform.filter

dunglas:
Excellent idea! Can you open a PR?

vincentchalamon:
@dunglas PR in progress. I just need to know if the interface list is enough or need to be completed

dunglas:
I would also suggest to add:

* `QueryItemExtensionInterface` => `api_platform.doctrine.orm.query_extension.item`
* `QueryCollectionExtensionInterface` => * `api_platform.doctrine.orm.query_extension.collection`

dunglas:
And do you think you'll be able to create a [MVCE](https://stackoverflow.com/help/mcve) without including the secret elements?

soyuka:
thanks @vincentchalamon !

PR #1152, opened 2017-06-01
[RFR] Improve autowiring

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1151
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/214

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection;
 
-use ApiPlatform\Core\Api\FilterInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryCollectionExtensionInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryItemExtensionInterface;
 use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -13,6 +13,11 @@
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection;
 
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryCollectionExtensionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryItemExtensionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\FilterInterface;
```
meyerbaptiste: It should be `ApiPlatform\Core\Api\FilterInterface`.

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -13,14 +13,20 @@
 
 namespace ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\DependencyInjection;
 
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryCollectionExtensionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryItemExtensionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\FilterInterface;
```
meyerbaptiste: It should be `ApiPlatform\Core\Api\FilterInterface`.

vincentchalamon:
I'm just thinking about something: filters require `id` option on tag, but I though it would be removed. Is it too soon to auto-configure filters?

dunglas:
It doesn't require anymore because of #862.
However the name of the service must be defined explicilty, I'm not sure if it's compatible or not with the autoconfigure feature.

vincentchalamon:
Maybe autoconfigure is not the right way to declare filters. We should still declare them manually.

BTW @dunglas, #862 requires documentation to be updated.

teohhanhui:
Most of the filters you'd declare in your app are from API Platform anyway, so PSR-4 service discovery and autowiring (there's nothing to be wired up) doesn't really help.

dunglas:
Thank you @vincentchalamon!

PR #1154, opened 2017-06-02
fix #1153

### LINKED ISSUE #1153: Deprecation in the form of new arguments not handled properly
Bug introduced in #984:

> TypeError: Argument 1 passed to ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\AbstractFilter::getClassMetadata() must be of the type string, null given

We must ensure that things work properly even when the new arguments are not provided.

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1153
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
@@ -233,7 +233,9 @@ abstract class AbstractFilter implements FilterInterface
                     @trigger_error(sprintf('Method %s() will have a second `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
                 }
             }
+        }
 
+        if (!isset($resourceClass)) {
             return [
                 'associations' => array_slice($parts, 0, -1),
                 'field' => end($parts),
--- a/tests/Bridge/Doctrine/Orm/Filter/AbstractFilterTest.php
+++ b/tests/Bridge/Doctrine/Orm/Filter/AbstractFilterTest.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Bridge\Doctrine\Orm\Filter;
+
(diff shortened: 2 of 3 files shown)


dunglas:
Can you add a non regression test?

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -233,7 +235,9 @@ protected function splitPropertyParts(string $property/*, string $resourceClass*
                     @trigger_error(sprintf('Method %s() will have a second `$resourceClass` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.1.', __FUNCTION__), E_USER_DEPRECATED);
                 }
             }
+        }
 
+        if (null === $resourceClass) {
```
teohhanhui: We could just use:

```php
if (!isset($resourceClass)) {
```

teohhanhui: Oh, wait... I think this should be fixed at the calling site instead?

soyuka:
> Oh, wait... I think this should be fixed at the calling site instead?

Nope the other one handles his stuff just fine IMO.

> Can you add a non regression test?

Done.

phpstan.neon
```diff
@@ -14,3 +14,4 @@ parameters:
         - '#Parameter \#2 \$dqlPart of method Doctrine\\ORM\\QueryBuilder::add\(\) expects Doctrine\\ORM\\Query\\Expr\\Base, Doctrine\\ORM\\Query\\Expr\\Join\[\] given#' # Fixed in Doctrine's master
         - '#Call to an undefined method Doctrine\\Common\\Persistence\\ObjectManager::getConnection\(\)#'
         - '#Parameter \#1 \$callable of static method Doctrine\\Common\\Annotations\\AnnotationRegistry::registerLoader\(\) expects callable, mixed\[\] given#'
+        - '#Class ApiPlatform\\Core\\Tests\\Bridge\\Doctrine\\Orm\\Filter\\DummyFilter was not found while trying to analyse it - autoloading is probably not configured properly#'
```
soyuka: I like the below test because it's simple, though I feel it's overkill to add a file for my dummy filter which is only used in the test. Is there a way I can avoid adding this? Or maybe it's bad practice and I should move this to a file?

Simperfit: I think it should be moved to a Class.

dunglas:
Needs `php-cs-fixer fix` before the merge.

soyuka:
Only scrutinizer failed but was okay last push. Thx for the review @teohhanhui.

PR #1156, opened 2017-06-02
Fix the Varnish purger

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

* Allows non indexed arrays
* Use `request` instead of `requestAsync` to throw an error in the purge doesn't work

### DIFF
--- a/src/HttpCache/VarnishPurger.php
+++ b/src/HttpCache/VarnishPurger.php
@@ -48,10 +48,10 @@ final class VarnishPurger implements PurgerInterface
             return sprintf('(^|\,)%s($|\,)', preg_quote($iri));
         }, $iris);
 
-        $regex = isset($parts[1]) ? sprintf('(%s)', implode(')|(', $parts)) : $parts[0];
+        $regex = count($parts) > 1 ? sprintf('(%s)', implode(')|(', $parts)) : array_shift($parts);
 
         foreach ($this->clients as $client) {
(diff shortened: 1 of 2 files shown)


src/HttpCache/VarnishPurger.php
```diff
@@ -48,10 +48,10 @@ public function purge(array $iris)
             return sprintf('(^|\,)%s($|\,)', preg_quote($iri));
         }, $iris);
 
-        $regex = isset($parts[1]) ? sprintf('(%s)', implode(')|(', $parts)) : $parts[0];
+        $regex = count($parts) > 1 ? sprintf('(%s)', implode(')|(', $parts)) : array_shift($parts);
```
soyuka: isset may still work with array shift no?

dunglas: No because the array has the following format: `['/foo' => '/foo', '/bar' => '/bar']`

soyuka: :| nice way to avoid counting otherwise.

soyuka: `current` might work, not sure about the perf reward though. `array_shift` does put cursor in position 0 every call.

meyerbaptiste: Perf are better with `current()` (or `array_pop()` for example). But I prefer `reset()`, it's more safe.

PR #1168, opened 2017-06-09
fix deprecation notice from in IriConvertor::getIriFromItem

sh41:
Per [OperationTypeDeprecationHelper.php](https://github.com/api-platform/core/blob/42940e3f48c3450d6c193f6bca6b8c5041821ffa/src/Api/OperationTypeDeprecationHelper.php#L18) comment: `false` should be replaced with `OperationType::ITEM`. 

Per deprecation notice in [RouteNameResolver](https://github.com/api-platform/core/blob/e66ed60aa0ea4065f75cbd8c24ef07d8f2b14abb/src/Bridge/Symfony/Routing/RouteNameResolver.php#L44) third argument also becomes mandatory and is empty array when not in use. 

| Q             | A
| ------------- | ---
| Bug fix?      | yes - fix deprecation notice
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

I've tried, but I cannot figure out how to write a test to cover this change. The deprecation comes out of RouteNameResolver::getRouteName which has coverage, but the deprecated call is made from IriConverter::GetIriFromItem which doesn't appear to have coverage at the moment.

I ended up in a situation where I was attempting to mock/prophesize Router returning a mocked RouteCollection returning a mocked Route and began to suspect that I was following the wrong path. If anyone can provide pointers I'll be eternally grateful.

### DIFF
--- a/src/Bridge/Symfony/Routing/IriConverter.php
+++ b/src/Bridge/Symfony/Routing/IriConverter.php
@@ -92,7 +92,7 @@ final class IriConverter implements IriConverterInterface
     public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
     {
         $resourceClass = $this->getObjectClass($item);
-        $routeName = $this->routeNameResolver->getRouteName($resourceClass, false);
+        $routeName = $this->routeNameResolver->getRouteName($resourceClass, OperationType::ITEM, []);
 
         try {
             $identifiers = $this->generateIdentifiersUrl($this->identifiersExtractor->getIdentifiersFromItem($item));
--- a/src/Bridge/Symfony/Routing/IriConverter.php
+++ b/src/Bridge/Symfony/Routing/IriConverter.php
@@ -92,7 +92,7 @@ final class IriConverter implements IriConverterInterface
     public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
     {
         $resourceClass = $this->getObjectClass($item);
-        $routeName = $this->routeNameResolver->getRouteName($resourceClass, OperationType::ITEM, []);
+        $routeName = $this->routeNameResolver->getRouteName($resourceClass, OperationType::ITEM);
 
         try {
             $identifiers = $this->generateIdentifiersUrl($this->identifiersExtractor->getIdentifiersFromItem($item));
--- a/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
@@ -55,7 +55,6 @@ final class CachedRouteNameResolver implements RouteNameResolverInterface
             $context = func_get_arg(2);
         } else {
             $context = [];
-            @trigger_error(sprintf('Method %s() will have a third `$context = []` argument in version 3.0. Not defining it is deprecated since 2.1.', __METHOD__), E_USER_DEPRECATED);
         }
 
         $routeName = $this->decorated->getRouteName($resourceClass, $operationType, $context);
--- a/src/Bridge/Symfony/Routing/RouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/RouteNameResolver.php
@@ -41,7 +41,6 @@ final class RouteNameResolver implements RouteNameResolverInterface
             $context = func_get_arg(2);
         } else {
             $context = [];
-            @trigger_error(sprintf('Method %s() will have a third `$context = []` argument in version 3.0. Not defining it is deprecated since 2.1.', __METHOD__), E_USER_DEPRECATED);
         }
 
         $operationType = OperationTypeDeprecationHelper::getOperationType($operationType);


src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -92,7 +92,7 @@ public function getItemFromIri(string $iri, array $context = [])
     public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
     {
         $resourceClass = $this->getObjectClass($item);
-        $routeName = $this->routeNameResolver->getRouteName($resourceClass, false);
+        $routeName = $this->routeNameResolver->getRouteName($resourceClass, OperationType::ITEM, []);
```
soyuka: Hmm, this is supposed to work fine with false. It'll raise an exception but should not be blocking.

sh41: Yes, that's right - it functions perfectly well, but logs two deprecation notices. The only effect of the change is that the two deprecations notices are no longer logged.

meyerbaptiste: To prevent deprecation notices, you have to change these lines too:
* [IriConverter.php#L114](https://github.com/api-platform/core/blob/2139c802abd52a81af7f6f16132eea55ad081e14/src/Bridge/Symfony/Routing/IriConverter.php#L114)
* [IriConverter.php#L126](https://github.com/api-platform/core/blob/2139c802abd52a81af7f6f16132eea55ad081e14/src/Bridge/Symfony/Routing/IriConverter.php#L126)

But yes, it's a little weird to have a `trigger_error()` with an optional argument... 😄

soyuka: Hmm yeah you're right this last argument is optional. I added this because the interface will change but maybe it's not appropriate and could be silenced. WDYT?

meyerbaptiste: IMO we should because `Not defining it is deprecated since 2.1` and the `E_USER_DEPRECATED` are wrong if it's an optional argument.

soyuka: :+1:, @sh41 may you update the pull request by removing the `trigger_error` instead of changing files where it's being used? Thanks!

sh41: Thanks @soyuka I have removed the trigger error from both the CachedRouteNameResolver and  RouteNameResolver. 

I'm not familiar with the syntax of the optional argument in this interface/class - if there is no deprecation to handle can the method signatures also be refactored?

```php
public function getRouteName(string $resourceClass, $operationType /**, array $context = [] **/): string
```
changed to 
```php
public function getRouteName(string $resourceClass, $operationType , array $context = [] ): string
```
and drop this bit?
```php
        if (func_num_args() > 2) {
            $context = func_get_arg(2);
        } else {
            $context = [];
        }
```

In CachedRouteNameResolver, RouteNameResolver and RouteNameResolverInterface?

soyuka: Haha no we can't do this because it'd need to change the interface, which becomes then a breaking change (ie major version). Leave the hack alone, it'll be removed in 3.x (hopefully).

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -92,7 +92,7 @@ public function getItemFromIri(string $iri, array $context = [])
     public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
     {
         $resourceClass = $this->getObjectClass($item);
-        $routeName = $this->routeNameResolver->getRouteName($resourceClass, false);
+        $routeName = $this->routeNameResolver->getRouteName($resourceClass, OperationType::ITEM);
```
soyuka: I don't think we need this change, anyway as both argument types are supported yet I'd say we can keep it like this.

sh41: There's a deprecation notice at [OperationTypeDeprecationHelper.php#L39](https://github.com/api-platform/core/blob/42940e3f48c3450d6c193f6bca6b8c5041821ffa/src/Api/OperationTypeDeprecationHelper.php#L39) that suggests that although the boolean type works at the moment it will stop working in v3. Happy to revert the change if it is premature.

meyerbaptiste: 👍 for this change, we should not use our own deprecated code.

soyuka: If we don't how would we now they're deprecated? ^^ Anyway, [there are a bunch of places where we still use `false`](https://github.com/api-platform/core/search?utf8=%E2%9C%93&q=getRouteName&type=), though we were going to fix those only when breaking.

meyerbaptiste: For me, deprecated notices should be for the users of our public API and should be avoided in our internals. And in our tests we should only have one for the deprecation test (i.e. `@group legacy` + `@expectedDeprecation`).

soyuka:
thanks @sh41

PR #1172, opened 2017-06-12
Check if operation_type exists in AbstractItemNormalizer

karser:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

When I directly serialize an entity which is ApiPlatform resource, `$context['operation_type']` isn't set (usually it's set by SerializerContextBuilder::createFromRequest) and it throws and error
```
Undefined index: operation_type in AbstractItemNormalizer
```
For consistency, CollectionNormalizer [has `isset($context['operation_type'])` check](https://github.com/survos/core/blob/e66ed60aa0ea4065f75cbd8c24ef07d8f2b14abb/src/Hydra/Serializer/CollectionNormalizer.php#L78-L78).


soyuka:
Thanks @karser !

PR #1182, opened 2017-06-16
Fix a problem when a composite identifier is missing

raoulclais:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This PR fixes a bug when a resource is requested with a wrong composite identifier, like 'ida=1;' instead of 'ida=1;idb=2'.

### DIFF
--- a/features/main/composite.feature
+++ b/features/main/composite.feature
@@ -120,6 +120,11 @@ Feature: Retrieve data with Composite identifiers
     }
     """
 
+  Scenario: Get the first composite relation with a missing identifier
+    Given there are Composite identifier objects
+    When I send a "GET" request to "/composite_relations/compositeLabel=1;"
+    Then the response status code should be 400
+
   @dropSchema
   Scenario: Get first composite item
     Given there are Composite identifier objects
--- a/src/Bridge/Doctrine/Orm/ItemDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/ItemDataProvider.php
@@ -18,6 +18,8 @@ use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryResultItemExtensionInter
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\IdentifierManagerTrait;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGenerator;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\PropertyNotFoundException;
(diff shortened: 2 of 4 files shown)


src/Bridge/Doctrine/Orm/ItemDataProvider.php
```diff
@@ -67,7 +69,11 @@ public function getItem(string $resourceClass, $id, string $operationName = null
             throw new ResourceClassNotSupportedException();
         }
 
-        $identifiers = $this->normalizeIdentifiers($id, $manager, $resourceClass);
+        try {
+            $identifiers = $this->normalizeIdentifiers($id, $manager, $resourceClass);
+        } catch (PropertyNotFoundException $e) {
+            throw new BadRequestHttpException($e->getMessage());
```
meyerbaptiste: The exception should be `InvalidArgumentException`.

soyuka:
Failures are not related, I'll fix those in a new PR.

soyuka:
Thanks for the patch @raoulclais ! May you rebase on 2.0 instead of merging?

For example:

```
# first reset the merge
git fetch origin # assuming origin remote points to api-platform/core not your fork
git rebase origin/2.0
git push -f upstream hotfix/missing-composite-identifier # assuming upstream remote points to your fork
```

You may have to reset to skip the merge commit from above.

Another solution is to submit your patch without merging and we will take care of rebasing :). Thanks!

Simperfit:
Thanks @raoulclais

features/main/composite.feature
```diff
@@ -120,6 +120,11 @@ Feature: Retrieve data with Composite identifiers
     }
     """
 
+  Scenario: Get the first composite relation with a missing identifier
+    Given there are Composite identifier objects
+    When I send a "GET" request to "/composite_relations/compositeLabel=1;"
+    Then the response status code should be 400
```
teohhanhui: It should be 404, not 400...

Simperfit: The resource could never be found because the composite is invalid, why 404 ?

soyuka: Because it's "not found"

Simperfit: The server did not even search it in the DB ^^

soyuka: > 10.4.5 404 Not Found
   The server has not found anything matching the Request-URI. 

This is by definition what happened here no? Though, one could argue that the url is malformed (hence 400).

Simperfit: This is totally my point ;)

Simperfit: so @teohhanhui @soyuka what do we do ? :o

PR #1183, opened 2017-06-16
Fix/2.0 tests

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

### DIFF
--- a/tests/Fixtures/FileConfigurations/resources.yml
+++ b/tests/Fixtures/FileConfigurations/resources.yml
@@ -32,7 +32,7 @@ resources:
                 attributes:
                     'foo': ['Foo']
                     'bar':
-                        0: ['Bar']
+                        '0': ['Bar']
                         'baz': 'Baz'
                     'baz': 'Baz'
             'name':
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
@@ -19,6 +19,7 @@ use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use FOS\UserBundle\FOSUserBundle;
 use Nelmio\ApiDocBundle\NelmioApiDocBundle;
 use Prophecy\Argument;
+use Prophecy\Exception\Doubler\MethodNotFoundException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\ResourceInterface;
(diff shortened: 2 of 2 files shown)


src/Metadata/Extractor/YamlExtractor.php
```diff
@@ -32,7 +32,7 @@
     protected function extractPath(string $path)
     {
         try {
-            $resourcesYaml = Yaml::parse(file_get_contents($path));
+            $resourcesYaml = Yaml::parse(file_get_contents($path), Yaml::PARSE_KEYS_AS_STRINGS);
```
soyuka: Not sure about this one, @meyerbaptiste is this correct? Thanks!

meyerbaptiste: Yes it's correct, but this constant was introduced in `symfony/yaml` 3.3. So tests will fail with the `--prefer-lowest` argument!

soyuka: Hmm, what solution do we have if we want tests to pass on 2.0 then?

soyuka: I could mark tests as legacy but doesn't feel clean enough.

dunglas: IIRC, using this constant is deprecated, the user should use quotes instead.

soyuka: hmm okay maybe I need to fix yaml files instead then.

tests/Fixtures/FileConfigurations/resources.yml
```diff
@@ -32,7 +32,7 @@ resources:
                 attributes:
                     'foo': ['Foo']
                     'bar':
-                        0: ['Bar']
+                        '0': ['Bar']
```
soyuka: thanks @dunglas :smile:

src/Metadata/Extractor/YamlExtractor.php
meyerbaptiste: > IIRC, using this constant is deprecated

Are you sure? I can't find any information on this subject and I don't see anything in the code of the component.

soyuka: Anyway, fixing the actual issue in the yaml seems to be the way to go.

dunglas: @meyerbaptiste https://github.com/symfony/symfony/pull/22948

Issue #1165, opened 2017-06-08
CircularReferenceException in custom ItemDataProvider


vincentchalamon:
Hello,

I'm implementing 2 custom data providers, connecting on WordPress API: 1 for collection, 1 for item. For the first one, I can easily inject the `serializer` service. But for the item, I get a CircularReferenceException.

A solution would be to create an event subscriber which listens to `kernel.request` event and serialize the controller result, but it's not generic at all (and I don't find it very clean…).

A better solution would be to listen to `SerializerAwareInterface` on my ItemDataProvider from API Platform, and if so call `setSerializer` method.

What do you think of it @api-platform/core-team?

teohhanhui:
A data provider isn't supposed to do serialization?

vincentchalamon:
@teohhanhui So how can I convert my WordPress API JSON result to my entity?

soyuka:
Just to understand the context you're parsing the resulting JSON (from WP API) and then populating some entities?

vincentchalamon:
@soyuka Yep. I have a `PostCollectionDataProvider` which calls WordPress API (through guzzle) to get the list of posts in JSON, I need to serialize this list to an array of `AppBundle\WordPress\Post` objects. Same for `PostItemDataProvider`

vincentchalamon:
Then, I'll have the same context with `Tag` & `Category` objects

teohhanhui:
> I need to serialize this list to an array of AppBundle\WordPress\Post objects. 

Do you mean deserializing the JSON data into the `Post` objects? I guess it makes sense then...

> A better solution would be to listen to SerializerAwareInterface on my ItemDataProvider from API Platform, and if so call setSerializer method.

Or some other tricks like lazy services which should help...

soyuka:
to deserialize then not serialize :p. 

I don't get the issue here, you should be able to use your own serializer/normalizer.

vincentchalamon:
To deserialize of course, my bad 😉

vincentchalamon:
> Or some other tricks like lazy services which should help...

@teohhanhui Like a service locator?

```yml
# services.yml
services:
    app.serializer.locator:
        class: Symfony\Component\DependencyInjection\ServiceLocator
        arguments:
            -
                serializer: '@serializer'
        tags:
            - { name: container.service_locator }

    App\DataProvider\PostItemDataProvider:
        class: App\DataProvider\PostItemDataProvider
[...]

PR #1184, opened 2017-06-16
[RFR] Add SerializerAwareInterface supports on ItemDataProvider

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1165
| License       | MIT
| Doc PR        | WIP

Calling any external API from a custom ItemDataProvider, it may return a JSON which must be deserialized to my object. To do so, I must inject the `serializer` service in my custom ItemDataProvider, which creates a CircularReferenceException. A solution could be to use [service locators](http://symfony.com/blog/new-in-symfony-3-3-service-locators), but the following solution can manage it more easily for final users.

This PR allows to support `SerializerAwareInterface` on ItemDataProvider to prevent this exception.

- [x] Fix DoctrineBridge min version (bug on detecting non-entity classes)
- [x] Add interface + trait
- [x] Detect interface on data providers (collection/item) to add `api_platform.serializer` dependency through service locator
- [x] Add Behat tests
- [x] Add unit tests

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
@@ -53,14 +53,14 @@ final class DataProviderPass implements CompilerPassInterface
         $queue = new \SplPriorityQueue();
 
         foreach ($services as $serviceId => $tags) {
+            $definition = $container->getDefinition($serviceId);
+            if (is_subclass_of($definition->getClass(), SerializerAwareDataProviderInterface::class)) {
+                $definition->addMethodCall('setSerializerLocator', [$container->getDefinition('api_platform.serializer_locator')]);
+            }
+
             foreach ($tags as $attributes) {
                 $priority = $attributes['priority'] ?? 0;
                 $queue->insert(new Reference($serviceId), $priority);
-
-                $definition = $container->getDefinition($serviceId);
-                if (is_subclass_of($definition->getClass(), SerializerAwareDataProviderInterface::class)) {
-                    $definition->addMethodCall('setSerializerLocator', [$container->getDefinition('api_platform.serializer_locator')]);
-                }
             }
         }
 
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
 
 use ApiPlatform\Core\Api\OperationType;
+use ApiPlatform\Core\DataProvider\SerializerAwareDataProviderInterface;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
@@ -55,6 +56,11 @@ final class DataProviderPass implements CompilerPassInterface
             foreach ($tags as $attributes) {
                 $priority = $attributes['priority'] ?? 0;
                 $queue->insert(new Reference($serviceId), $priority);
+
+                $definition = $container->getDefinition($serviceId);
+                if (is_subclass_of($definition->getClass(), SerializerAwareDataProviderInterface::class)) {
+                    $definition->addMethodCall('setSerializerLocator', [$container->getDefinition('api_platform.serializer_locator')]);
+                }
             }
         }
 
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
 use ApiPlatform\Core\Api\OperationType;
 use ApiPlatform\Core\DataProvider\SerializerAwareDataProviderInterface;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 
@@ -25,9 +26,12 @@ use Symfony\Component\DependencyInjection\Reference;
  * @internal
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
  */
 final class DataProviderPass implements CompilerPassInterface
 {
+    use PriorityTaggedServiceTrait;
+
     /**
      * {@inheritdoc}
      */
@@ -48,22 +52,15 @@ final class DataProviderPass implements CompilerPassInterface
      */
     private function registerDataProviders(ContainerBuilder $container, string $type)
     {
-        $services = $container->findTaggedServiceIds("api_platform.{$type}_data_provider", true);
+        $services = $this->findAndSortTaggedServices("api_platform.{$type}_data_provider", $container);
 
-        $queue = new \SplPriorityQueue();
-
-        foreach ($services as $serviceId => $tags) {
-            $definition = $container->getDefinition($serviceId);
+        foreach ($services as $reference) {
+            $definition = $container->getDefinition((string)$reference);
             if (is_a($definition->getClass(), SerializerAwareDataProviderInterface::class, true)) {
-                $definition->addMethodCall('setSerializerLocator', [$container->getDefinition('api_platform.serializer_locator')]);
-            }
-
-            foreach ($tags as $attributes) {
-                $priority = $attributes['priority'] ?? 0;
-                $queue->insert(new Reference($serviceId), $priority);
+                $definition->addMethodCall('setSerializerLocator', [new Reference('api_platform.serializer_locator')]);
             }
         }
 
-        $container->getDefinition("api_platform.{$type}_data_provider")->addArgument(iterator_to_array($queue, false));
+        $container->getDefinition("api_platform.{$type}_data_provider")->addArgument($services);
     }
 }
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
@@ -54,7 +54,7 @@ final class DataProviderPass implements CompilerPassInterface
 
         foreach ($services as $serviceId => $tags) {
             $definition = $container->getDefinition($serviceId);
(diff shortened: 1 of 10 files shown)


soyuka:
Got some requests on this feature, would be lovely if you had time to fix tests!

vincentchalamon:
@soyuka I'm working on it, but I'm blocked: the easiest way should be to inject the `@api_platform.serializer` in all services calling `ChainItemDataProvider::getItem` method, and injecting the serializer in it, but I really don't like the idea of setting a dependence from another one…

soyuka:
Though, I have no issues in injecting `serializer` in a DataProvider. What was your CircularReferenceException about?

vincentchalamon:
@soyuka Check https://github.com/api-platform/core/issues/1165 for more details. Did you try injecting the serializer in an ItemDataProvider?

soyuka:
Ohhh, yeah I remember now. 

I'm not sure that injecting the dependency from so many files is the best solution :/. It may look over-engineered but a service locator might be worth it here.

If we want this to be in the bundle for future users, we could add a `SerializerAwareDataProviderInterface` with a trait (similar to SerializerAware):

```php
trait SerializerAwareDataProviderTrait
{
  private $serializer;
  public function setSerializer(Locator $serializerLocator) {
    $this->serializer = $serializerLocator->get(Serializer::class);
  }
}
``` 

```yaml
    app.serializer_locator:
        class: Symfony\Component\DependencyInjection\ServiceLocator
        tags: ['container.service_locator']
        arguments:
            -
                Symfony\Component\Serializer\Serializer: '@serializer'
```

And then in the `CompilerPass` we can add a check:

```
            if (is_subclass_of($def->getClass(), SerializerAwareDataProviderInterface::class)) {
                $def->addMethodCall('setSerializer', [$container->getDefinition('app.serializer_locator')]);
            }
```

As you've managed to do it in #1165 service locators are working in this case. 

WDYT? I'm still not 100% convinced by the above but this case is not an easy one ^^.

vincentchalamon:
I was already thinking about service locators, but as it's only available since Symfony 3.3, it will cause a BC break.
Otherwise, we could enable this feature only if service locators are available. And if the final user is using Symfony <3.3, he will have to upgrade `symfony/dependency-injection` dependency, or manage it by itself.

soyuka:
IIRC, we already use service locators for the new filters (property filter and groups filter).

vincentchalamon:
OK I'll use a service locator to inject the serializer in the ChainItemDataProvider.

soyuka:
Why does it need to be specific to the `ChainItemDataProvider`? We can make this work on every `ItemDataProviderInterface` no?

vincentchalamon:
@soyuka Yep, better idea 😃 I'm on it

composer.json
```diff
@@ -50,7 +50,7 @@
         "symfony/console": "^3.3 || ^4.0",
         "symfony/debug": "^2.8 || ^3.0 || ^4.0",
         "symfony/dependency-injection": "^3.3 || ^4.0",
-        "symfony/doctrine-bridge": "^2.8 || ^3.0 || ^4.0",
+        "symfony/doctrine-bridge": "^2.8.2 || ^3.0 || ^4.0",
```
soyuka: why?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
```diff
@@ -55,6 +56,11 @@ private function registerDataProviders(ContainerBuilder $container, string $type
             foreach ($tags as $attributes) {
                 $priority = $attributes['priority'] ?? 0;
                 $queue->insert(new Reference($serviceId), $priority);
+
+                $definition = $container->getDefinition($serviceId);
```
soyuka: we should do this in the above loop no? I think this loops on the tags because a data provider can be both collection and item. There you're changing the definition twice, but once is enough.

teohhanhui:
Haven't looked into the implementation, but let's make sure that this does not shift the potential for circular reference from client code to API Platform code. :)

vincentchalamon:
@soyuka Great idea for your article, it will sure help people. I have more & more projects using API Platform without doctrine, as you said: api to api. Don't hesitate to contact me if I can help :-)

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
```diff
@@ -52,6 +53,11 @@ private function registerDataProviders(ContainerBuilder $container, string $type
         $queue = new \SplPriorityQueue();
 
         foreach ($services as $serviceId => $tags) {
+            $definition = $container->getDefinition($serviceId);
+            if (is_subclass_of($definition->getClass(), SerializerAwareDataProviderInterface::class)) {
+                $definition->addMethodCall('setSerializerLocator', [$container->getDefinition('api_platform.serializer_locator')]);
```
dunglas: Would it not be better to inject the locator into the constructor?

vincentchalamon: How would you impose this locator in the constructor through an interface?

dunglas:
What do you think about this other strategy:

* create a new instance of the serializer component dedicated to API Platform
* don't add the API Platform's normalizer to the `serializer` service but to ours (`api_platform.serialzier` I guess)

It will fix almost all recurrent problems with circular references (it will be possible to inject the `serializer` service in a data provider without problem) and it will prevent "polluting" the general Symfony serializer with our (more and more complex) own services.

ping @api-platform/core-team

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
dunglas: You can't, but setter injection should be limited to optional dependencies or you can't assume the object is fully initialized after the call to the constructor. Setter injection should generally be avoided IMO.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
```diff
@@ -52,6 +53,11 @@ private function registerDataProviders(ContainerBuilder $container, string $type
         $queue = new \SplPriorityQueue();
 
         foreach ($services as $serviceId => $tags) {
+            $definition = $container->getDefinition($serviceId);
+            if (is_subclass_of($definition->getClass(), SerializerAwareDataProviderInterface::class)) {
```
dunglas: Btw why not `is_a($definition->getClass(), SerializerAwareDataProviderInterface::class, true)`?

dunglas:
Btw, I'm probably missing something but what's the benefit of providing this interface and the related compiler pass instead of just adding a doc entry explaining how to use a service locator to use the serializer in a data provider? It will get rid of the setter injection (in the docs we can promote constructor injection). WDYT @vincentchalamon?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
vincentchalamon: I agree about the required dependencies in the constructor, but how can I properly set the locator in the constructor? It can be at any position or named with any variable name…
To do so, I did the same feature than the [Serializer](https://github.com/symfony/serializer/blob/master/Serializer.php#L65).

vincentchalamon:
@dunglas I've found more easy if I could just implement an interface & a trait (3 lines and autowiring support) to use the serializer in my ItemDataProvider instead of using my own service locator in my project (2 services declaration and no autowiring support).

About your [other serializer](https://github.com/api-platform/core/pull/1184#issuecomment-311166174), you want to create a custom serializer with just API Platform normalizers/denormalizers?

> (...) it will prevent "polluting" the general Symfony serializer with our (more and more complex) own services.

But we'll finally have 2 serializers with the same number of normalizers/denormalizers, instead of 1 serializer now.

theofidry:
@dunglas the issue with having two serializers in Symfony is that you must implement the normalizer/encoder discovery by yourself. Or do you consider in that specific case creating a serializer manually from scratch should be enough?

vincentchalamon:
Ping @dunglas: need your answer in previous comments, or could it be related in another PR (so this one could be merged)?

Simperfit:
ping @dunglas ;)

src/DataProvider/SerializerAwareDataProviderTrait.php
```diff
@@ -0,0 +1,46 @@
...
+
+use Psr\Container\ContainerInterface;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * Inject serializer in data providers.
+ *
+ * @author Vincent Chalamon <vincent@les-tilleuls.coop>
+ */
+trait SerializerAwareDataProviderTrait
```
Simperfit: I'm thinking of something about this trait,

Couln't we add the serialize and deserialize method directly in it ?

Since it's already some king of a proxy, that would avoid a function call to getSereliazer() IMO.

vincentchalamon: This is related to https://github.com/api-platform/core/pull/1184#issuecomment-311166174 as a serializer alias or alternative.

philippecarle:
Tested on my project using Symfony & API Platform as a REST Services orchestrator, it works fine and sounds quite good to me from a design point of view. I'm currently using my fork but would ❤️ to see this merged (or improved and merged) ! 👍 
_(could be synthesised by a "up!")_

src/DataProvider/SerializerAwareDataProviderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataProvider;
+
+use Psr\Container\ContainerInterface;
+
+/**
+ * Inject serializer in data providers.
```
dunglas: Injects

src/DataProvider/SerializerAwareDataProviderTrait.php
```diff
@@ -0,0 +1,42 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataProvider;
+
+use Psr\Container\ContainerInterface;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * Inject serializer in data providers.
```
dunglas: Injects

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataProviderPass.php
```diff
@@ -47,17 +52,15 @@ public function process(ContainerBuilder $container)
...
-
-        $queue = new \SplPriorityQueue();
+        $services = $this->findAndSortTaggedServices("api_platform.{$type}_data_provider", $container);
 
-        foreach ($services as $serviceId => $tags) {
-            foreach ($tags as $attributes) {
-                $priority = $attributes['priority'] ?? 0;
-                $queue->insert(new Reference($serviceId), $priority);
+        foreach ($services as $reference) {
+            $definition = $container->getDefinition((string)$reference);
```
dunglas: Missing space after `(string)`

dunglas:
LGTM

dunglas:
Thanks @vincentchalamon

PR #1188, opened 2017-06-20
Add Subresource to swagger docs

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Display SubResources in the swagger docs

screen shot 2017-06-20 at 16 00 36

### DIFF
--- a/src/Metadata/Resource/SubResourceMetadata.php
+++ b/src/Metadata/Resource/SubResourceMetadata.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Resource;
+
+final class SubResourceMetadata
+{
+    private $parent;
+    private $property;
+    private $isCollection;
+    private $parentResourceClass;
+
+    public function __construct(ResourceMetadata $parent, string $property, bool $isCollection, string $parentResourceClass)
+    {
+        $this->parent = $parent;
+        $this->property = $property;
+        $this->isCollection = $isCollection;
+        $this->parentResourceClass = $parentResourceClass;
+    }
+
+    /**
+     * @return ResourceMetadata
+     */
+    public function getParent(): ResourceMetadata
+    {
+        return $this->parent;
+    }
+
+    /**
+     * @return string
+     */
+    public function getProperty(): string
+    {
+        return $this->property;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isCollection(): bool
+    {
+        return $this->isCollection;
+    }
+
+    /**
+     * @return string
+     */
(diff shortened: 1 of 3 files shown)


src/Metadata/Resource/SubResourceMetadata.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Resource;
+
+final class SubResourceMetadata
```
soyuka: Hmm this is a good idea. I didn't wanted to introduce a new Metadata class when I worked on subresources but this will give more power to subresource declarations (ping @teohhanhui  ^^).

WDYT about going a bit further, so that `PropertyMetadataFactory` will create the `SubResourceMetadata` when it's declared as such? This is what the outcome may be in the code below:

```php
$propertyMetadata = $this->propertyMetadataFactory->create($parentResourceClass, $property);
if ($propertyMetadata->hasSubresource()) {
        $subResourceMetadata = $propertyMetadata->getSubresource(); // this is an instance of SubResourceMetadata
}
```

With something like this we can avoid repeating the following code (for example also in `ApiLoader`): 

```
$isCollection = $propertyMetadata->getType()->isCollection();
$resourceClass = $isCollection ? $propertyMetadata->getType()->getCollectionValueType()->getClassName() : $propertyMetadata->getType()->getClassName();
```

ping @api-platform/core-team 

IMO we need to agree on this change first because it may involve many changes ^^. Anyway, if we introduce this new metadata class, it'd be a shame not to use it everywhere we can :).

teohhanhui: I don't think it's worth it to introduce yet another metadata class.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -636,4 +649,29 @@ private function getSerializerContext(string $operationType, bool $denormalizati
...
+     * @param ResourceMetadata $parentResourceMetadata
+     * @param array            $mimeTypes
+     */
+    private function processSubResources(\ArrayObject $paths, \ArrayObject $definitions, string $parentResourceClass, ResourceMetadata $parentResourceMetadata, array $mimeTypes)
+    {
+        $serializerContext = $this->getSerializerContext(OperationType::SUBRESOURCE, false, $parentResourceMetadata, 'get');
+        $options = isset($serializerContext['groups']) ? ['serializer_groups' => $serializerContext['groups']] : [];
+        foreach ($this->propertyNameCollectionFactory->create($parentResourceClass, $options) as $property) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($parentResourceClass, $property);
+            if ($propertyMetadata->hasSubresource()) {
```
dunglas: Can you use a guard clause to make the code easier to read?

```php
if (!$propertyMetadata->hasSubresource()) {
    continue;
}
```

pierredup:
Closing in favor of #1225

Issue #1170, opened 2017-06-10
Dependency on Symfony 3.3 not specified ?


tseho:
When trying **v2.1.0-beta.1**, there is a missing dependency on symfony/dependency-injection:^3.3 when using the ApiPlatformBundle. It's only specified inside require-dev.

When installing it on symfony/symfony:3.2.9, the package will be installed but the DependencyInjection will produce an error on runtime like this:

`Unable to parse file "/app/vendor/api-platform/core/src/Bridge/Symfony/Bundle/DependencyInjection/../Resources/config/filter.xml".`

I had to upgrade to symfony:3.3 and it worked.

Should the requirement be explicit ?

meyerbaptiste:
No, because as you can see it, `symfony/dependency-injection` is used as a bridge inside the core, so it isn't required.

tseho:
@meyerbaptiste Thanks for the response. 

I understand the core/bridge difference, is adding a composer conflict to symfony/dependency-injection:<3.3 an option ?

This, in theory, should not affect standalone projects using core, only full symfony applications using the bundle.

meyerbaptiste:
To sum up the problem, see this comment: https://github.com/api-platform/core/issues/1121#issuecomment-302354562

But yes, `conflict` could be an idea, this should work.

dunglas:
👍 for a `conflict` section.

Simperfit:
@tseho could you submit a PR ?

tseho:
@Simperfit sure.

Do we need to add other conflicts or `symfony/dependency-injection:<3.3` will be enough ?

meyerbaptiste:
IMO, all the others dev dependencies except tests dependencies.

Simperfit:
Like @meyerbaptiste said, we don't want it to be symfony/symfony btw, but only the dev dependencies.

Thanks about that @tseho

Nek-:
This issue can be close @tseho

PR #1190, opened 2017-06-20
Adds conflict on symfony/dependency-injection < 3.3

tseho:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Fixed tickets | #1170 
| License       | MIT

As discussed in #1170, I'm adding composer conflict for symfony/dependency-injection lower than 3.3, since the bundle packed in the repository only works for symfony ^3.3.

There was discussion about adding more conflicts, I'm not sure if this is a good idea. Since api-platform/core is not only a bundle, it could totally be installed with doctrine/orm or any other require-dev package without using the api-platform bundle (in theory).

I would rather add only real conflicts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,6 +64,9 @@
         "symfony/validator": "^3.3 || ^4.0",
         "symfony/yaml": "^3.3 || ^4.0"
     },
+    "conflict": {
+        "symfony/dependency-injection": "<3.3"
+    },
     "suggest": {
         "friendsofsymfony/user-bundle": "To use the FOSUserBundle bridge.",
         "guzzlehttp/guzzle": "To use the HTTP cache invalidation system.",


teohhanhui:
This is not a real solution. The conflict only exists when using the bundle. We need a separate `composer.json`.

tseho:
@teohhanhui This is a partial solution. Since **v2.1.0-beta.1**, the included bundle uses features that have been added in symfony/dependency-injection:3.3.

It will not work with symfony 3.2 but still can be installed because the `require` is minimal.

If you have a better solution, I'm open to it :)

teohhanhui:
What about someone who uses an older version of `symfony/dependency-injection`, but does not use the `ApiPlatformBundle`? :laughing:

teohhanhui:
We indeed have some missing dependencies that can only be expressed by having a separate `composer.json` for the bundle.

tseho:
How do you specify a separate composer.json ? Do you mean splitting the package in 2 repositories (library/bundle) ?

dunglas:
Thanks @tseho

PR #1192, opened 2017-06-21
Fix path for custom operation with Swagger UI

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/254
| License       | MIT
| Doc PR        | N/A

A fix while waiting for Swagger 3 which will permit overriding the base path at the path item level.
See:
* https://github.com/OAI/OpenAPI-Specification/issues/562
* https://github.com/OAI/OpenAPI-Specification/pull/604

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/swagger.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/swagger.xml
@@ -13,7 +13,7 @@
             <argument type="service" id="api_platform.resource_class_resolver" />
             <argument type="service" id="api_platform.operation_method_resolver" />
             <argument type="service" id="api_platform.operation_path_resolver" />
-            <argument type="service" id="api_platform.router" />
+            <argument>null</argument>
             <argument type="service" id="api_platform.filter_locator" />
             <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
             <argument>%api_platform.oauth.enabled%</argument>
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -38,6 +38,9 @@ use Symfony\Component\Routing\RouteCollection;
  */
 final class ApiLoader extends Loader
 {
+    /**
+     * @deprecated since version 2.1, to be removed in 3.0. Use {@see RouteNameGenerator::ROUTE_NAME_PREFIX} instead.
+     */
     const ROUTE_NAME_PREFIX = 'api_';
     const DEFAULT_ACTION_PATTERN = 'api_platform.action.';
     const SUBRESOURCE_SUFFIX = '_get_subresource';
@@ -166,13 +169,13 @@ final class ApiLoader extends Loader
                 $resourceRouteName = $this->routeNameResolver($rootShortname);
 
                 $operation['identifiers'] = [['id', $rootResourceClass]];
-                $operation['route_name'] = sprintf('%s%s_%s%s', self::ROUTE_NAME_PREFIX, $resourceRouteName, $propertyName, self::SUBRESOURCE_SUFFIX);
-                $operation['path'] = $this->operationPathResolver->resolveOperationPath($rootShortname, $operation, OperationType::SUBRESOURCE);
+                $operation['route_name'] = sprintf('%s%s_%s%s', RouteNameGenerator::ROUTE_NAME_PREFIX, $resourceRouteName, $propertyName, self::SUBRESOURCE_SUFFIX);
+                $operation['path'] = $this->operationPathResolver->resolveOperationPath($rootShortname, $operation, OperationType::SUBRESOURCE, $operation['route_name']);
             } else {
                 $operation['identifiers'] = $parentOperation['identifiers'];
                 $operation['identifiers'][] = [$parentOperation['property'], $resourceClass];
                 $operation['route_name'] = str_replace(self::SUBRESOURCE_SUFFIX, "_$propertyName".self::SUBRESOURCE_SUFFIX, $parentOperation['route_name']);
-                $operation['path'] = $this->operationPathResolver->resolveOperationPath($parentOperation['path'], $operation, OperationType::SUBRESOURCE);
+                $operation['path'] = $this->operationPathResolver->resolveOperationPath($parentOperation['path'], $operation, OperationType::SUBRESOURCE, $operation['route_name']);
             }
 
             $route = new Route(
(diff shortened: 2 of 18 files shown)


src/Bridge/Symfony/Routing/RouteNameGenerator.php
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Generates the Symfony route name associated with an operation name and a resource short name.
+ *
+ * @author Baptiste Meyer <baptiste.meyer@gmail.com>
+ */
+class RouteNameGenerator
+{
+    const ROUTE_NAME_PREFIX = 'api_';
+
+    private function __construct()
```
Simperfit: why ?

meyerbaptiste: Static class. Should this class be moved to `Util`?

Simperfit: I don't think it should be moved to Util, it's a right place here ;)

soyuka: private constructor does prevent the class from being instantiated no? I think it's the point here :p.

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -35,6 +34,10 @@
  */
 final class ApiLoader extends Loader
 {
+    /**
+     * @deprecated since version 2.0.10, to be removed in 3.0.
+     *             Use {@see RouteNameGenerator::ROUTE_NAME_PREFIX} instead.
```
soyuka: indentation

src/Bridge/Symfony/Routing/RouteNameGenerator.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Generates a Symfony route name.
+     *
+     * @param string $operationName
+     * @param string $resourceShortName
+     * @param bool   $collection
+     *
+     * @return string
+     */
+    public static function generate(string $operationName, string $resourceShortName, bool $collection): string
```
soyuka: Oh okay we're on `2.0` here. We need to change the `collection` flag to an `OperationType` when merging this on master IMO.

src/Bridge/Symfony/Routing/RouteNameGenerator.php
meyerbaptiste: Yep, right! Recipe for static classes:
* final class
* private constructor
* static attributes/methods

src/Bridge/Symfony/Bundle/Resources/config/swagger.xml
```diff
@@ -13,7 +13,7 @@
             <argument type="service" id="api_platform.resource_class_resolver" />
             <argument type="service" id="api_platform.operation_method_resolver" />
             <argument type="service" id="api_platform.operation_path_resolver" />
-            <argument type="service" id="api_platform.router" />
+            <argument>null</argument>
```
dunglas: `<argument />` is enough IIRC

src/Bridge/Symfony/Routing/ApiLoader.php
dunglas: I would merge this patch in 2.1, WDYT? I don't like to introduce deprecations in patch releases.

src/Bridge/Symfony/Routing/RouteNameGenerator.php
```diff
@@ -0,0 +1,50 @@
...
+namespace ApiPlatform\Core\Bridge\Symfony\Routing;
+
+use Doctrine\Common\Util\Inflector;
+
+/**
+ * Generates the Symfony route name associated with an operation name and a resource short name.
+ *
+ * @author Baptiste Meyer <baptiste.meyer@gmail.com>
+ */
+class RouteNameGenerator
```
dunglas: Can we mark this class `@internal`, I don't like public static classes.

src/Bridge/Symfony/Routing/RouteNameGenerator.php
dunglas: See my previous comment (we should merge this only in master IMO)

src/Bridge/Symfony/Routing/ApiLoader.php
meyerbaptiste: As you want, but this bug seems critic for many people (cf issue).

src/Bridge/Symfony/Routing/RouteNameGenerator.php
dunglas: So they will upgrade to 2.1 :P

src/Bridge/Symfony/Bundle/Resources/config/swagger.xml
meyerbaptiste: An empty string is dumped with `<argument />`...

src/Bridge/Symfony/Routing/RouteNameGenerator.php
meyerbaptiste: :smile: Okay, I'll update the base branch, and my PR.

soyuka: Oh okay @dunglas didn't saw your comment, and I'm :+1: on merging this in master! So yes, please drop the `bool $collection` for an `OperationType` :D.

meyerbaptiste: Not so easy with sub resources 😢

soyuka: I can take care of it if you want, anyway I want to refactor some parts and introduce Subresource metadata which should make things easier. For now, you can just support Item and Collection!

meyerbaptiste: Are sub resources not exposed by Swagger UI? 😮

soyuka: Not yet but we've an open pr here: https://github.com/api-platform/core/pull/1188

meyerbaptiste:
Okay, the PR is ready with the master branch and ready for review.

**I didn't take care of sub-resources. IMO we can add the support of these ones when the `DocumentationNormalizer` will support them.**

dunglas:
Amazing work! Thank you very much for this fix and this cleanup @meyerbaptiste!!

src/Bridge/Symfony/Routing/RouteNameGenerator.php
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @throws InvalidArgumentException
+     *
+     * @return string
+     */
+    public static function generate(string $operationName, string $resourceShortName, $operationType): string
+    {
+        if (OperationType::SUBRESOURCE === $operationType = OperationTypeDeprecationHelper::getOperationType($operationType)) {
+            throw new InvalidArgumentException(sprintf('%s::SUBRESOURCE is not supported as operation type by %s().', OperationType::class, __METHOD__));
+        }
```
soyuka: @meyerbaptiste I have to remove this basically and use this class in the RouterOperationPathResolver?

meyerbaptiste: Yes, ideally...

Issue #1177, opened 2017-06-14
Swagger ui 


soyuka:
We have an issue on the new swagger ui with `identifier` fields. 

First, it's always forced to `integer` and always has the `id` name (even though the id name can be different).

![swager-ui-bug](https://user-images.githubusercontent.com/1321971/27125645-3e01be50-50f5-11e7-9efd-2cf41f73d85e.png)

On the screenshot for example, the resource has a composite identifier. In the old swagger I could send a string `foo=a;bar=b` and it worked. Because there is a client-side validation on `integer` it's not possible anymore :|. 

Same goes for resources that are not identified by an `integer`, field is still client-validating and doesn't send the request.

I'm not sure where it could be fixed (but if it could be fixed by Friday it'd be lovely :p).

ping @ginifizz @Simperfit

soyuka:
Oh and another thing. Every time we want to test we have to:
1. click on try it out
2. then only we can fill the fields

I think that fields should be visible the moment we open a "request" so that we can directly click on `Execute` without having to do 2 manipulations.

dunglas:
For the other thing, I agree but it should be reported to the Swagger team... It's the default behavior of the new version.

backbone87:
big problem when one is using UUIDs as PKs

Simperfit:
@backbone87 could you please create  a new issue describing the problem ?

backbone87:
i dont see why? its exactly the problem @soyuka described in the opening post. you cant use anything but integer IDs with the current sandbox UI. i was just replying to stress that this is a hugh DX problems for ppl using UUIDs.

dunglas:
It's a bug in Swagger UI isn't it? If it's the case it should be fixed upstream.

backbone87:
idk, i just searched for an open issue, because i experienced the problem described in the OP. afaik swagger ui is configured via a swagger/open api config? then maybe the generated config isnt correct?

backbone87:
Ok i have looked into the swagger api spec generated by api platform and this is what i get for a string ID field (UUID).

```json
                    "parameters": [
                        {
                            "name": "id",
                            "in": "path",
                            "required": true,
                            "type": "integer"
                        }
                    ],
```

so this is not an upstream problem

dunglas:
Thanks for investigating and finding the underlying problem. We'll fix that!

backbone87:
Here are the relevant code sections:
[...]

PR #1197, opened 2017-06-23
Swagger: change the type of IDs in URL to string

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1177
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -217,7 +217,7 @@ final class DocumentationNormalizer implements NormalizerInterface
             'name' => 'id',
             'in' => 'path',
             'required' => true,
-            'type' => 'integer',
+            'type' => 'string',
         ]];
         $pathOperation['responses'] ?? $pathOperation['responses'] = [
             '200' => [
@@ -290,7 +290,7 @@ final class DocumentationNormalizer implements NormalizerInterface
             [
                 'name' => 'id',
                 'in' => 'path',
-                'type' => 'integer',
+                'type' => 'string',
                 'required' => true,
             ],
             [
@@ -333,7 +333,7 @@ final class DocumentationNormalizer implements NormalizerInterface
         $pathOperation['parameters'] ?? $pathOperation['parameters'] = [[
             'name' => 'id',
             'in' => 'path',
-            'type' => 'integer',
+            'type' => 'string',
             'required' => true,
         ]];
 
--- a/tests/Swagger/Serializer/DocumentationNormalizerTest.php
+++ b/tests/Swagger/Serializer/DocumentationNormalizerTest.php
@@ -139,7 +139,7 @@ class DocumentationNormalizerTest extends \PHPUnit_Framework_TestCase
                             [
                                 'name' => 'id',
                                 'in' => 'path',
-                                'type' => 'integer',
+                                'type' => 'string',
                                 'required' => true,
                             ],
                         ],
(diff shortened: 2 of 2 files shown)


backbone87:
While i think taht this fixes the problem wouldnt it be better to set the Parameters according to the API-Platform Metadata?

soyuka:
+1 @backbone87 why is this hardcoded?

dunglas:
By definition, a part of an URL is necessarily a string isn't it?

There are other cases where the current behavior is faulty (for insrance when using composite identifiers).
Why should we introduce a heavier logic if this tiny patch fixes all cases?

backbone87:
for composite PKs i would rly like a spec like this:
```
"/books/{isbn}/chapters/{chapter_index}": {
  "get": {
    "parameters": [
      {
        "name": "isbn",
        "in": "path",
        "required": true,
        "type": "string"
      },
      {
        "name": "chapter_index",
        "in": "path",
        "required": true,
        "type": "integer"
      }
    ]
  }
}
```

but yes, just making it a string would allow anything allowed in URLs. but that doesnt provide nice inputs on GUIs that take Open API Specs as a base (like Swagger UI)

dunglas:
The rationale behind this is related to how JSON-LD works. A JSON-LD document is identified by an URL (included in the `@id` property). There is no native concept of "composite keys" in a JSON-LD and in hypermedia APIs in general. There is a single identifier for a resource: its URL.

We imported this concept in API Platform, composite keys are basically just a hack to support this Doctrine feature (and its why it's in the data provider and not in the IRI converter). I think we shouldn't change that. Composite keys shouldn't escape from the Doctrine layer, outside key are just a single identifier.

> but yes, just making it a string would allow anything allowed in URLs. but that doesnt provide nice inputs on GUIs that take Open API Specs as a base (like Swagger UI)

Do you have an example? I don't understand how it can change something because an URL is by definition a string, to include a int in an URL, you must cast it to a string at some point.

dunglas:
> but yes, just making it a string would allow anything allowed in URLs.

Yes, but the validation still occurs server-side. If you pute an invalid ID, you'll get a 404. Not a big deal to me.

backbone87:
> The rationale behind this is related to how JSON-LD works. A JSON-LD document is identified by an URL (included in the @id property). There is no native concept of "composite keys" in a JSON-LD and in hypermedia APIs in general. There is a single identifier for a resource: its URL.

But the only interface that json-ld has in that sense is "give me the resource for IRI", which is the `IriConverter`. The interface of the `IriConverter` is good. i am talking how it works internally. And internally we do use the router to parse the IRI, but insteadof doin ALL the parsing in the router, there is a subparsing process within the `ItemDataProvider`, which violates SRP imho. An interface only sets requirements on the outside, not on a specific implementation and the concept of multiple identifier values of an entity is well known inside the default implementation of the `IriConverter` (note: i explicitly not use the RDB terminology here)

backbone87:
> Yes, but the validation still occurs server-side. If you pute an invalid ID, you'll get a 404. Not a big deal to me.

That ensures integrety, but not best usability of client side tools using the Open API spec to build the UI.

soyuka:
@dunglas in the UI there's client side validation according to types. It's therefore cleaner to have correct properties and types. Anyway, I'm 👍 on this patch because it fixes the blocking bug. Though I'd be 👍 if someone wanted to return the correct types. About composite identifiers I'm not sure that they could be supported as 2 fields in swagger (`string` would be fine).

dunglas:
I agree with @soyuka proposal. @backbone87 I get your point and indeed documenting a precise type would be better from an UX point of view. If you want to create a patch to do a better type detection, I'll be happy to merge it!

backbone87:
i am also perfectly fine with this PR being a quick fix. also tested it and it works

backbone87:
this needs to be merged into master

dunglas:
@backbone87 done

PR #1206, opened 2017-06-27
Implement subresource metadata and annotation

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

For now the @ApiSubresource only does the same thing as `@ApiProperty(subresource=true)`. It'll be better with metadata and an annotation class if we want to add features to it in the future.

Related to https://github.com/api-platform/core/pull/1188. ping @pierredup

### DIFF
--- a/src/Annotation/ApiSubresource.php
+++ b/src/Annotation/ApiSubresource.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Annotation;
+
+/**
+ * Property annotation.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @Annotation
+ * @Target({"METHOD", "PROPERTY"})
+ */
+final class ApiSubresource
+{
+}
--- a/src/Metadata/Property/PropertyMetadata.php
+++ b/src/Metadata/Property/PropertyMetadata.php
@@ -35,7 +35,7 @@ final class PropertyMetadata
     private $attributes;
     private $subresource;
 
-    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, bool $subresource = null)
+    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null)
     {
         $this->type = $type;
         $this->description = $description;
@@ -347,12 +347,34 @@ final class PropertyMetadata
         return $metadata;
     }
 
-    public function hasSubresource()
+    /**
+     * Represents whether the property has a subresource.
+     *
+     * @return bool
+     */
+    public function hasSubresource(): bool
+    {
+        return $this->subresource !== null;
+    }
+
+    /**
+     * Gets the subresource metadata.
+     *
+     * @return SubresourceMetadata|null
+     */
+    public function getSubresource()
     {
         return $this->subresource;
     }
 
-    public function withSubresource(bool $subresource = null): self
+    /**
+     * Returns a new instance with the given subresource.
+     *
+     * @param SubresourceMetadata $subresource
+     *
+     * @return self
+     */
+    public function withSubresource(SubresourceMetadata $subresource = null): self
     {
         $metadata = clone $this;
         $metadata->subresource = $subresource;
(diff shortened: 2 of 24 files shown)


tests/Fixtures/TestBundle/Entity/Container.php
```diff
@@ -33,7 +34,7 @@ class Container
     private $id;
 
     /**
-     * @ApiProperty(subresource=true)
+     * @ApiProperty(subresource=@ApiSubresource())
```
soyuka: Should it be a `PROPERTY` annotation? I don't think we need it, and there is no value in adding metadata factories for subresources IMO. Lmk.

dunglas: I agree, `@ApiSubresource()` directly on the property is easier to type, to read and to understand.

src/Annotation/ApiSubresource.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace ApiPlatform\Core\Annotation;
+
+/**
+ * Property annotation.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @Annotation
+ * @Target({"ANNOTATION"})
```
dunglas: Should be `@Target({"METHOD", "PROPERTY"})`

dunglas: Nevermind, my bad.

teohhanhui:
I suggest we take this opportunity to make it:

```php
/**
 * @ApiSubresource(operations=true)
 */
protected $children;
```

(open to better name)

soyuka:
What does `operations` trigger?

teohhanhui:
`@ApiSubresource` is for configuring a subresource, just like how `@ApiProperty` is for configuring a property. It also implies that the property is a subresource (*what does this do?*).The actual configuration is therefore whether to expose the operations (only `GET`, for now).

teohhanhui:
On second thought, we should add `operations` when we actually support overriding the operations. :laughing:

soyuka:
Okay but I suggest we expose operations without having a `operations` flag. It doesn't make sense to declare `ApiSubresource` if you don't want operations, does it?

teohhanhui:
Okay, maybe we need to have a clear definition of what we call a subresource (for our purposes):

> A "subresource" is a resource exposed under a path with parameters dependent on another resource.

tests/Metadata/Property/Factory/ExtractorPropertyMetadataFactoryTest.php
```diff
@@ -116,7 +119,63 @@ public function testCreateWithParentPropertyMetadataFactoryYaml(PropertyMetadata
...
+    /**
+     * @dataProvider decoratedPropertyMetadataProvider
+     */
+    public function testCreateWithCollectionTypedParentPropertyMetadataFactoryYaml(PropertyMetadata $expectedPropertyMetadata)
+    {
+        $configPath = __DIR__.'/../../../Fixtures/FileConfigurations/resources.yml';
+
+        $collectionType = new Type(Type::BUILTIN_TYPE_OBJECT,
+                    false,
+                    ArrayCollection::class,
```
dunglas: Missing `use` statement (it should made Travis happy btw).

soyuka: last minute test \o/ fixed thx.

src/Metadata/Property/PropertyMetadata.php
```diff
@@ -347,12 +347,34 @@ public function withChildInherited(string $childInherited): self
...
+    {
+        return $this->subresource !== null;
+    }
+
+    /**
+     * Gets the subresource metadata.
+     *
+     * @return SubresourceMetadata|null
+     */
+    public function getSubresource(): SubresourceMetadata
```
meyerbaptiste: You have to remove the return type here.

src/Metadata/Property/SubresourceMetadata.php
```diff
@@ -0,0 +1,57 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Property;
+
+/**
+ * Subresource metadata.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class SubresourceMetadata
```
teohhanhui: I'm -1 for this. It duplicates information already available from the `Type` via `PropertyMetadata`.

soyuka: It does and in the end we'll have less function calls to retrieve the same information.

```php
           $isCollection = $propertyMetadata->getType()->isCollection();
           $subresource = $isCollection ? $propertyMetadata->getType()->getCollectionValueType()->getClassName() : $propertyMetadata->getType()->getClassName();
```

This will also ease the maintenance to have a SubresourceMetadata class if we need to support subresource related cases in the future no?

dunglas:
Thanks @soyuka!

jorgeluisacostaalonso:
Hi!!

soyuka:
Please open a new issue, and if it's a collection it should already be paginated.

jorgeluisacostaalonso:
@soyuka It's already implemted! It was that i didnt see it in the subResource url api_doc.
thanks

PR #1208, opened 2017-06-27
Replace AccessDeniedException by AccessDeniedHttpException

thomasglachant:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT
| Doc PR        | 

Replace "AccessDeniedException" by "AccessDeniedHttpException" in `DenyAccessListener` to have a status code 403 when using *is_granted* attribute on ApiResource.

Example : 
```
@ApiResource(
      attributes={"is_granted"="has_role('ROLE_ADMIN')"}
}
```

### DIFF
--- a/src/EventListener/DenyAccessListener.php
+++ b/src/EventListener/DenyAccessListener.php
@@ -17,8 +17,8 @@ use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\ExpressionLanguage\Expression;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
-use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 
 /**
  * Denies access to the current resource if the logged user doesn't have sufficient permissions.
@@ -41,7 +41,7 @@ final class DenyAccessListener
      *
      * @param GetResponseEvent $event
      *
-     * @throws AccessDeniedException
+     * @throws AccessDeniedHttpException
      */
     public function onKernelRequest(GetResponseEvent $event)
     {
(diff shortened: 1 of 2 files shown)


meyerbaptiste:
This PR break all the `Symfony\Component\Security\Http\Firewall\ExceptionListener` mechanism. For example, when the user is not fully authenticated, the authentication entry point process doesn't work anymore (cf. Travis).

I suspect a problem with your code if you have this issue.

dunglas:
I suspect that a better fix would be to add `AccessDeniedException` to the default exception mapping: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L174-L175

meyerbaptiste:
Mapping should be useless here. `AccessDeniedException` already has a code (403) and we use it in our `ExceptionAction`: https://github.com/api-platform/core/blob/8c0664a798c9ce241b26eafd3dd40daeef1c2423/src/Action/ExceptionAction.php#L71. IMO the problem comes from elsewhere.

dunglas:
@meyerbaptiste you're right, we even have a test for this: https://github.com/api-platform/core/blob/master/features/authorization/deny.feature#L32

Looks like another issue.

meyerbaptiste:
What do you use as authentication system @thomasglachant? What status code do you have instead of 403?

thomasglachant:
When I call an api resource with "is_granted" attribute, and a non-granted user I had a status 500 because the listener returns an `AccessDeniedException` (and not an `HttpException`).

I look other listeners (ex : ReadListener) and I saw it returns a `NotFoundHttpException`, so I thought we can normalize this and always throw HttpException in listeners. 

It break an other usecase in test because now it returns a 403 instead of a 401 when no authorization header is provide ... but I did not have the time to search why from now...

meyerbaptiste:
> It break an other usecase in test because now it returns a 403 instead of a 401 when no authorization header is provide ... but I did not have the time to search why from now...

I explained why in my first comment 😉. tl;dr Removing the `AccessDeniedException` breaks a lot of mechanism of the Symfony security component (i.e `ExceptionListener`).

> and a non-granted user I had a status 500 because the listener returns an AccessDeniedException (and not an HttpException)

The problem is here. 401 or 403 should be returned because it's provided by API Platform. If you have a 500, you probably have a problem or a specific use case with your authentication's implementation. So, what do you use?

dunglas:
@thomasglachant Any news? It looks like there is another problem in your code.

thomasglachant:
@dunglas Yes, I try to reproduced this with @clementtalleu yesterday, but we can't. So I think too for a problem in our code.

PR #1209, opened 2017-06-28
Extra try-catch in DenyAccessListener to keep ExpressionVoter from failing if is_granted expression is invalid

desmax:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

The problem I wanted to address:
I have custom Voters  that handle my custom operations(apply, get, delete)  from `is_granted` field. That works, but if I deny access then ExpressionVoter tries to process `is_granted`, which fails, since its not a valid expression according to Expression Language.

### DIFF
--- a/src/EventListener/DenyAccessListener.php
+++ b/src/EventListener/DenyAccessListener.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\EventListener;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\ExpressionLanguage\Expression;
+use Symfony\Component\ExpressionLanguage\SyntaxError;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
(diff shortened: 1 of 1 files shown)


meyerbaptiste:
I'm against, IMO developers should be informed if the expression is not a valid one. So, the `SyntaxError` exception should be thrown.

meyerbaptiste:
If you have this specific use case, maybe you should decorate this listener yourself instead.

desmax:
My use case is very common as for me. I think solution has to be in the core. Not sure if this is best implementation though. Please advice.

dunglas:
SensioFrameworkExtraBundle fixed this issue by adding a new `is_granted` function to call custom voters. Maybe should we follow the same path?

Here is the file we should copy: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/0c88c0b057aceb3ff09488b2a35d8ba25ad11ab8/Security/ExpressionLanguage.php#L21

WDYT @desmax @meyerbaptiste?

dunglas:
Closing in favor of #1216

PR #1212, opened 2017-06-29
OrderFilter::extractProperties now always return an array

marcw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

This PR fixes a bug related to https://github.com/api-platform/admin/issues/25 where wrongly formed order parameters causes the api to crash.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
@@ -108,6 +108,11 @@ class OrderFilter extends AbstractFilter
      */
     protected function extractProperties(Request $request): array
     {
-        return $request->query->get($this->orderParameterName, []);
+        $properties = $request->query->get($this->orderParameterName, []);
+        if (!is_array($properties)) {
+            return [];
+        }
+
+        return $properties;
     }
 }
--- a/features/doctrine/order_filter.feature
+++ b/features/doctrine/order_filter.feature
@@ -360,6 +360,65 @@ Feature: Order filter on collections
     }
     """
 
+  Scenario: Get a collection even if the order parameter is not well-formed
+    When I send a "GET" request to "/dummies?sort=id&order=asc"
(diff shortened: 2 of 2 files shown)


src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -143,6 +143,11 @@ protected function filterProperty(string $property, $direction, QueryBuilder $qu
      */
     protected function extractProperties(Request $request/*, string $resourceClass*/): array
     {
-        return $request->query->get($this->orderParameterName, []);
+        $properties = $request->query->get($this->orderParameterName, []);
+        if (!is_array($properties)) {
+            return [];
```
soyuka: Hmm this means that the parameter will not be correctly parsed. Shouldn't this be seen as an error? I mean, if nothing shows up, the end user will not know that this isn't giving the expecting results.

marcw: I'm not sure that we should be *that* severe by returning a 400 error code for a wrongly formed parameters.

marcw: Plus the built-in API documentation is pretty clear on how to send order parameters.

dunglas: Without this patch, does it throws a 400 or a 500?

marcw: If the parameter is not an array (ie only ?order=foobar), PHP throws a type error because the returned value is a string, not an array

dunglas:
Thank you for fixing this @marcw!

teohhanhui:
What about the other filters? :D

PR #1219, opened 2017-07-03
fix varnish content negotiation

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/349
| License       | MIT
| Doc PR        | 

Fixes https://github.com/api-platform/api-platform/issues/349. The vary header for varnish was wrong.

### DIFF
--- a/features/http_cache/headers.feature
+++ b/features/http_cache/headers.feature
@@ -1,5 +1,5 @@
 Feature: Default values of HTTP cache headers
-  In order to make API responses caheable
+  In order to make API responses cacheable
   As an API software developer
   I need to be able to set default cache headers values
 
@@ -10,4 +10,4 @@ Feature: Default values of HTTP cache headers
     Then the response status code should be 200
     And the header "Etag" should be equal to '"21248afbca1f242fd3009ac7cdf13293"'
     And the header "Cache-Control" should be equal to "max-age=60, public, s-maxage=3600"
-    And the header "Vary" should be equal to "Content-Type, Cookie"
+    And the header "Vary" should be equal to "Accept, Cookie"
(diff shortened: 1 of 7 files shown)


teohhanhui:
Can you also change https://github.com/api-platform/core/blob/fe5d7dd946ff94f4a24a304883db78044761435b/src/HttpCache/EventListener/AddHeadersListener.php#L65

from

```php
$response->setVary($this->vary);
```

to 

```php
$response->setVary($this->vary, false);
```

so that we do not replace any existing `Vary` header(s).

bendavies:
@teohhanhui done

bendavies:
@teohhanhui i'm not getting that headers feature. where does "cookie" come from in
`And the header "Vary" should be equal to "Accept, Cookie"`?

teohhanhui:
It's from the test app bundle's configuration.

bendavies:
of course, thanks.

teohhanhui:
You need to fix the test here:

https://github.com/api-platform/core/blob/1bc08379451582c6894eb6054e711650b4dfbfb7/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php#L310

bendavies:
(Y)

teohhanhui:
We have a duplicate value in the `Vary` header because we've already set `Vary: Accept` in https://github.com/api-platform/core/blob/1bc08379451582c6894eb6054e711650b4dfbfb7/src/EventListener/RespondListener.php#L51

What we could do is:

```php
foreach (array_diff($this->vary, $response->getVary()) as $vary) {
    $response->setVary($vary, false);
}
```

(Not the most efficient, but I don't see another way for now...)

bendavies:
@teohhanhui done with test

teohhanhui:
Thanks @bendavies! And thanks for bearing with us.

PR #1221, opened 2017-07-03
404 instead of 400 for bad identifiers fix #1191

### LINKED ISSUE #1191: When composite identifier is missing should return 404 instead of 400
See https://github.com/api-platform/core/pull/1182/files#r123189130

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1191 
| License       | MIT
| Doc PR        | na

Another solution for this would be to catch the Exception in the ReadListener and return null. Lmk wdyt about this.

ping @teohhanhui  @Simperfit

### DIFF
--- a/features/main/composite.feature
+++ b/features/main/composite.feature
@@ -123,7 +123,7 @@ Feature: Retrieve data with Composite identifiers
   Scenario: Get the first composite relation with a missing identifier
     Given there are Composite identifier objects
     When I send a "GET" request to "/composite_relations/compositeLabel=1;"
-    Then the response status code should be 400
+    Then the response status code should be 404
 
   @dropSchema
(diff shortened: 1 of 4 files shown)


teohhanhui:
> Another solution for this would be to catch the Exception in the ReadListener and return null

Yes, no need to duplicate the throwing code. :smile:

teohhanhui:
> Another solution for this would be to catch the Exception in the ReadListener and return null

So in the end this wasn't changed eh?

soyuka:
omg thought I patched this https://github.com/api-platform/core/pull/1231

PR #1225, opened 2017-07-04
Add subresources to hydra

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

ping @pierredup 

Update of #1188

### DIFF
--- a/src/Bridge/Symfony/Routing/RouteNameGenerator.php
+++ b/src/Bridge/Symfony/Routing/RouteNameGenerator.php
@@ -28,6 +28,7 @@ use Doctrine\Common\Util\Inflector;
 class RouteNameGenerator
 {
     const ROUTE_NAME_PREFIX = 'api_';
+    const SUBRESOURCE_SUFFIX = '_subresource';
 
     private function __construct()
     {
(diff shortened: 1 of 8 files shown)


src/Metadata/Property/SubresourceMetadata.php
```diff
@@ -54,4 +58,30 @@ public function withCollection(bool $collection): self
...
+        $metadata = clone $this;
+        $metadata->parentResourceClass = $parentResourceClass;
+
+        return $metadata;
+    }
+
+    public function getParentResourceClass(): string
+    {
+        return $this->parentResourceClass;
+    }
```
soyuka: I'm not sure we need those informations here. Usually when using a `SubresourceMetadata` we already have those informations. I want to update the PR to try and remove those adds.

tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -262,6 +265,131 @@ public function testNormalize()
...
+                                'description' => 'Question resource response',
+                                'schema' => ['$ref' => '#/definitions/Question'],
+                            ],
+                            404 => ['description' => 'Resource not found'],
+                        ],
+                    ]),
+                ],
+                '/questions/{id}/answer' => [
+                    'get' => new \ArrayObject([
+                        'tags' => ['Answer'],
```
soyuka: Shouldn't there be `Answer` and `Question` here?

meyerbaptiste:
We have to fix paths of sub-resources too: https://github.com/api-platform/core/pull/1192/files#diff-000bd9bbe078960ebd85ec9cbd642bcdR54

soyuka:
@meyerbaptiste could you remind me what needs to be done there?

meyerbaptiste:
We have to use the router to resolve paths (because of route prefix), so the way to find the route of a sub-resource.

src/Bridge/Symfony/Routing/RouteNameGenerator.php
```diff
@@ -44,18 +45,39 @@ private function __construct()
      *
      * @return string
      */
-    public static function generate(string $operationName, string $resourceShortName, $operationType): string
+    public static function generate(string $operationName, string $resourceShortName, $operationType, string $property = null, bool $subresourceCollection = false): string
```
soyuka: ping @meyerbaptiste I have to add `property` and `collection` information everywhere :|. 

I'm not satisfied with this, WDYT? 

I haven't changed the `OperationPathResolver`s yet because I'd need to add so many hacks to add those 2 new arguments... Let me know if you've a better idea in mind, I'm kinda stuck in the no-breaking-change-loop here :smile:.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -134,6 +137,37 @@ private function addPaths(\ArrayObject $paths, \ArrayObject $definitions, string
...
+        }
+
+        foreach ($operations as $operationName => $operation) {
+            $operation['identifiers'] = [['id', $resourceClass]];
+            $operation['property'] = $property;
+            $operation['collection'] = $subresourceMetadata->isCollection();
+            $operation['swagger_context']['tags'] = [$resourceMetadata->getShortName(), $parentResourceMetadata->getShortName()];
+
+            $path = $this->getPath($parentResourceMetadata->getShortName(), $operationName, $operation, OperationType::SUBRESOURCE);
+            $method = 'GET'; // we only support GET for subresources
```
soyuka: TODO: No need to declare variable.

src/Bridge/Symfony/Routing/RouteNameGenerator.php
meyerbaptiste: Yes that's why I said it's not so easy! I had encountered the same problem... 😞

soyuka:
@meyerbaptiste hmm in fact is was easier as expected because we have everything we need in `$operation`! 

One thing I didn't get though is that we're getting the `$operationName` before calling `resolveOperationPath`. Therefore, we go through the `RouterOperationPathResolver` even though we didn't created the route yet (for example in Swagger or in ApiLoader). 

Why is the `RouterOperationPathResolver` not deferring if it can't find the `route_name`?

soyuka:
I've removed subresources on swagger for now. There is more work to be done before we can actually do this. See #1239

dunglas:
Thanks @pierredup @soyuka!

Issue #1233, opened 2017-07-06
Identifier date doctrine type converts to value with time that broke item query result


cve:
Hi,

when I have a Post with date field as doctrine date type, and I want to use that field as ApiProperty identifier then in ItemProvider there is a trait that converts a parameter value into doctrine type which for example transforms "2017-01-01" into "2017-01-01 00:00:00" that fails on query result.

soyuka:
We had the same issue and I think that #1478 is a slightly better fix, indeed it allows you to add a normalizer for any identifer (uuid or datetime), for example:

```php
<?php
// DateIdentifierNormalizer

namespace Core\ApiBundle\Identifier\Normalizer;
use AppBundle\Orm\Types\IdxDate; // using a custom Doctrine Type here

use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;

class DateNormalizer implements DenormalizerInterface
{
    public function denormalize($data, $class, $format = null, array $context = [])
    {
        return (string) (new IdxDate($data));
    }

    public function supportsDenormalization($data, $type, $format = null)
    {
        return $type === \DateTime::class; //I'm working on this there's an issue on PropertyInfo regarding \DateTime extended classes
    }
}
```

```yaml
services:
    bender_api.date_identifier_normalizer:
        public: false
        class: Core\ApiBundle\Identifier\Normalizer\DateNormalizer
        tags: [ { name: 'api_platform.identifier_normalizer' } ]
```

```php
<?php
namespace AppBundle\Orm\Types;

class IdxDate extends \DateTime
{
    public function __toString()
    {
        return $this->format('d-m-Y');
    }

    public static function createFromFormat($format, $time, $object = null)
    {
        $d = \DateTime::createFromFormat($format, $time);

        $t = new self();
        $t->setTimestamp($d->getTimeStamp());

        return $t;
    }
}
```

Keeping this here for docs

Issue #1167, opened 2017-06-09
Bad query parameters when the identifier is a Doctrine Type in Doctrine ItemDataProvider:getItem


magarzon:
Hi.

I found an issue in doctrine ItemDataProvider introduced in v2.0.8.

In the doctrine ItemDataProvider, when is getting an item, is building a query using the identifiers in the request.

To get the identifiers, first it normalizes them, because can be another than id and even can be two fields

This normalization is performed now by a trait, IdentifierManagerTrait. So, in this line: [https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php#L76](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php#L76 ) is converting the value of the identifier, if it has a Doctrine type, to its PHP value.

I don't know if this it's needed, because eventually this value is used to query the database, and it should be the database value for that type, not the PHP value, that can be even an object.

If the reason to add this is some kind of conversion of the input to the PHP value, there is another step missing: In [the addWhereForIdentifiers method in ItemDataProvider](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/ItemDataProvider.php#L104), when is setting the parameter in the query builder, it should check again the type (as it's done in the trait), and add it as the third parameter:
`            $queryBuilder->setParameter($placeholder, $value, $type);`

If you don't do that, the default type is string, and then doctrine doesn't call Type::convertToDatabaseValue and instead of the right value for that parameter/type, the value could be (as in my case, that is an object), the classname.

So, or either remove the conversion in the trait if it's not really needed, or the type must be added as I explained.

Simperfit:
@magarzon I think this is a bug, do you mind opening a PR to fix it ?

magarzon:
@Simperfit ok, but I don't know which solution it's better, because I don't know what was the final purpose of this change

soyuka:
May I ask what Type are identifiers in your case?

magarzon:
[...]

PR #1234, opened 2017-07-06
Add an option for skip type conversion

cve:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1167, #1233
| License       | MIT
| Doc PR        | None

I'm added an option to ApiProperty for attributes with "convert"=false, which skip converting doctrine type

@dunglas @sroze @theofidry


src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -86,4 +88,24 @@ private function normalizeIdentifiers($id, ObjectManager $manager, string $resou
...
+    /**
+     * Check if attributes contains convert to false.
+     *
+     * @param array $attributes
+     *
+     * @return bool
+     */
+    private function shouldConvert(array $attributes = null): bool
+    {
+        foreach ((array) $attributes as $attribute) {
```
dunglas: Use less cast because of the typehint.

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -74,10 +74,12 @@ private function normalizeIdentifiers($id, ObjectManager $manager, string $resou
                 throw new PropertyNotFoundException(sprintf('Invalid identifier "%s", "%s" has not been found.', $id, $propertyName));
             }
 
-            $doctrineTypeName = $doctrineClassMetadata->getTypeOfField($propertyName);
+            if ($this->shouldConvert($propertyMetadata->getAttributes())) {
```
dunglas: Why don't you use `The `getAttribute` method directly?

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
cve: it seems to me, If attributes will be null then invalid argument supplied to foreach will be thrown

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
cve: cause I didn't know about that :), I will fix it, thanks.

Simperfit:
Could you finish this please ?

cve:
#1478 is not it better?

soyuka:
:+1: #1478 should cover that use case, thanks @cve !

Issue #1239, opened 2017-07-07
Subresources tree computation & refactor OperationPathResolver


soyuka:
When working on adding subresources to swagger, we hit a wall with the `OperationPathResolver`. Indeed, because subresources are computed recursively, it's not possible to generate it's path through the `OperationPathResolver`.

We talked about it and came to this solution:
1. Implement a "SubresourceTree" that will compute the whole tree (+ cache). This gives us the ability to retrieve a path/route name from a class and the properties tree.
2. Deprecate the current `OperationPathResolver::resolveOperationPath` method
3. Create a new method with a more useful signature:

```
public function getOperationPath(ResourceMetadata $resourceMetadata, string $operationType, string $operation, string $operationName, ? $treeLikeBelow, maybe PropertyMetadata here): string;
```

The goal is to be able to retrieve an operation path. For this to work with subresources we need the following information: 

```
Root => property => property => property
```

so that we know the beginning and the end of the subresource.

The following classes would need refactoring:
- OperationPathResolver
- ApiLoader
- RouteNameGenerator
- Swagger\DocumentationNormalizer

PR #1245, opened 2017-07-11
Swagger subresource + OperationPathResolver refactor proposal

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | not really
| New feature?  | yes but subresources aren't part of a stable release yet
| BC breaks?    | maybe, we'll get to that below
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #1239
| License       | MIT
| Doc PR        | na

### 1. `OperationPathResolverInterface`

Okay, after lot of thinking here's my proposal about the whole OperationPathResolver mess. 

Initially OperationPathResolver was implemented because some users want to use `-` instead of `_` in their paths. Because of this simple need, we introduced a complicated decoration pattern.

My proposal would be to deprecate the use of the `OperationPathResolverInterface` if all we want is to use a dash instead of an underscore. 

Instead, I've introduced a simple `PathNameGenerator` who's signature looks like this:

```php
    public function getName(string $name, bool $pluralize = true): string;
```

This class will be configurable the same way as is the `OperationPathResolver` but it won't compute paths. Instead, it's only used to name things.

What about `UnderscoreOperationPathResolver` and `DashOperationPathResolver` then? They're still working and there's no breaking change here but I think we should deprecate them in favor of a this new "naming service". Indeed, we don't need to generate a path, all we want is to use `-` or `_` to format our path.

### 2. `SubresourceOperationFactory`

With subresources complexity (recursion, start/end-points) we need to be able to retrieve a Subresource `path` or `route_name` from a given `ResourceClass` with ease. To do so, I've implemented a `SubresourceOperationFactory` (thanks @Simperfit) that will generate the operations from a given `resourceClass`.
This is better then using a `OperationPathResolver` based on `OperationType` and is less complex to implement.

This new class is currently used :
- in `ApiLoader` to add routes to the symfony router
- in `Swagger\DocumentationNormalizer` to provide swagger docs

Note that `Subresources` are NOT compatible with `OperationPathResolver`s! Still, it's using `PathNameGenerator` and subresources paths could be written with dashes instead of default underscores. 

Because the subresource feature is not in the stable release yet I think it's fine. Still, if one wants dashes, he would've to use `PathNameGenerator`.

Thanks at @meyerbaptiste @Simperfit for the talks around this issue.

@dunglas this refactor is needed for a stable release because the current state of handling subresource paths is too messy IMO.

### TODO:

- [x] Add tests for SubresourceOperationFactory
- [x] Add SubresourceOperationFactoryCache - @teohhanhui finally got a clean "visiting" identifier :D
- [x] Validate PathNameGenerator
- [x] Make the PathNameGenerator configurable + test
- [x] Deprecate DashOperationPathResolver UnderscoreOperationPathResolver
- [x] Remove commented tests on subresources and operationpathresolver
- [x] PathSegmentGenerator => PathSegmentNameGenerator


Simperfit:
I like that idea ;p

Do you want some help on this ?

soyuka:
It's almost done, just few tests to add! I just want to validate the `PathNameGenerator` thing before continuing.

src/Bridge/Symfony/Routing/SubresourceOperationFactory.php
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $property) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $property);
+
+            if (!$propertyMetadata->hasSubresource()) {
+                continue;
+            }
+
+            $subresource = $propertyMetadata->getSubresource();
```
soyuka: Nope it's `subresource` everywhere :p.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -43,8 +43,9 @@ public function getConfigTreeBuilder()
                 ->scalarNode('title')->defaultValue('')->info('The title of the API.')->end()
                 ->scalarNode('description')->defaultValue('')->info('The description of the API.')->end()
                 ->scalarNode('version')->defaultValue('0.0.0')->info('The version of the API.')->end()
-                ->scalarNode('default_operation_path_resolver')->defaultValue('api_platform.operation_path_resolver.underscore')->info('Specify the default operation path resolver to use for generating resources operations path.')->end()
+                ->scalarNode('default_operation_path_resolver')->defaultValue('api_platform.operation_path_resolver.generator')->info('[Deprecated] Specify the default operation path resolver to use for generating resources operations path.')->end()
```
soyuka: Is it possible to deprecate a configuration?

src/PathResolver/OperationPathResolver.php
```diff
@@ -0,0 +1,60 @@
...
+            @trigger_error(sprintf('Method %s() will have a 4th `string $operationName` argument in version 3.0. Not defining it is deprecated since 2.1.', __METHOD__), E_USER_DEPRECATED);
+        }
+
+        $operationType = OperationTypeDeprecationHelper::getOperationType($operationType);
+
+        if (OperationType::SUBRESOURCE === $operationType) {
+            throw new InvalidArgumentException('Subresource operations are not supported by the OperationPathResolver.');
+        }
+
+        $path = '/'.$this->pathNameGenerator->getPathName($resourceShortName, true);
```
soyuka: @TODO use some kind of backward compatibility here to call `UnderscoreOperationPathResolver`.

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -91,14 +90,31 @@
...
+            <argument type="service" id="api_platform.path_name_generator" />
+        </service>
+
+        <service id="api_platform.operation_path_resolver.underscore" class="ApiPlatform\Core\PathResolver\UnderscoreOperationPathResolver" public="false">
+                <deprecated>The "%service_id%" service is deprecated since ApiPlatform 2.1 and will be removed in 3.0. Use PathNameGenerator instead.</deprecated>
+        </service>
+
+        <service id="api_platform.operation_path_resolver.dash" class="ApiPlatform\Core\PathResolver\DashOperationPathResolver" public="false">
+                <deprecated>The "%service_id%" service is deprecated since ApiPlatform 2.1 and will be removed in 3.0. Use PathNameGenerator instead.</deprecated>
+        </service>
```
soyuka: Note that the `operation_path_resolver` is still valid. I just want to deprecate those:

1. `operation_path_resolver.underscore`
2. `operation_path_resolver.dash`

soyuka:
Almost there. The current code is backward compatible with the pre-2.1 OperationPathResolver's! Tests will be red because of deprecations (will fix later).

I'd like your help on naming things :man_dancing: :

- is `PathNameGenerator` fine?
- Where should I put `SubresourceOperationFactory`? I've put it under `ApiPlatform\Core\Operation\Factory` but maybe you've a better idea

src/Bridge/Symfony/Bundle/Resources/config/api.xml
Simperfit: Agreed

src/Bridge/Symfony/Routing/SubresourceOperationFactory.php
Simperfit: but it feels awkward

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -43,8 +43,9 @@ public function getConfigTreeBuilder()
                 ->scalarNode('title')->defaultValue('')->info('The title of the API.')->end()
                 ->scalarNode('description')->defaultValue('')->info('The description of the API.')->end()
                 ->scalarNode('version')->defaultValue('0.0.0')->info('The version of the API.')->end()
-                ->scalarNode('default_operation_path_resolver')->defaultValue('api_platform.operation_path_resolver.underscore')->info('Specify the default operation path resolver to use for generating resources operations path.')->end()
+                ->scalarNode('default_operation_path_resolver')->defaultValue('api_platform.operation_path_resolver.underscore')->info('[Deprecated] Specify the default operation path resolver to use for generating resources operations path.')->end()
```
dunglas: It's being added, but it's not merged yet: https://github.com/symfony/symfony/pull/22382. For now you can use this  trick: https://github.com/symfony/symfony/pull/21051#issuecomment-269208657

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -91,14 +90,31 @@
         <service id="api_platform.operation_path_resolver.router" class="ApiPlatform\Core\Bridge\Symfony\Routing\RouterOperationPathResolver" public="false">
             <argument type="service" id="api_platform.router" />
             <argument type="service" id="api_platform.operation_path_resolver.custom" />
+            <argument type="service" id="api_platform.subresource_operation_factory" />
         </service>
 
         <service id="api_platform.operation_path_resolver.custom" class="ApiPlatform\Core\PathResolver\CustomOperationPathResolver" public="false">
+            <!-- @TODO should be "api_platform.operation_path_resolver.generator" when the default is removed -->
```
dunglas: Can you remove the `@todo` annotation? It triggers annoying warnings in SensioLabs Insight.

src/Bridge/Symfony/Routing/RouteNameGenerator.php
```diff
@@ -28,7 +28,6 @@
 class RouteNameGenerator
```
dunglas: Can you make this make this class `final` btw?

src/Operation/DashPathNameGenerator.php
```diff
@@ -0,0 +1,39 @@
...
+namespace ApiPlatform\Core\Operation;
+
+use Doctrine\Common\Util\Inflector;
+
+/**
+ * Generate a path name with an underscore separator according to a string and whether it needs pluralization.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+class DashPathNameGenerator implements PathNameGeneratorInterface
```
dunglas: `final`

src/Operation/Factory/CachedSubresourceOperationFactory.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->cacheItemPool = $cacheItemPool;
+        $this->decorated = $decorated;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(string $resourceClass): array
+    {
+        $cacheKey = self::CACHE_KEY_PREFIX.md5(serialize([$resourceClass]));
```
dunglas: Why not just using `$resourceClass` as the cache key? It will be faster (if `/` must be escaped, you can use `str_replace`).

src/Operation/Factory/CachedSubresourceOperationFactory.php
```diff
@@ -0,0 +1,63 @@
...
+
+        try {
+            $cacheItem = $this->cacheItemPool->getItem($cacheKey);
+
+            if ($cacheItem->isHit()) {
+                return $cacheItem->get();
+            }
+        } catch (CacheException $e) {
+            // do nothing
```
dunglas: Maybe can you `return $subresourceOperations` here to prevent the call to `isset` later?

src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -0,0 +1,136 @@
...
+            $subresourceMetadata = $this->resourceMetadataFactory->create($subresourceClass);
+
+            if (null === $parentOperation) {
+                $visiting = "$rootResourceClass-$property-$subresourceClass";
+            } else {
+                $prefix = '';
+                $visiting = "{$parentOperation['property']}-{$parentOperation['resource_class']}-$property-$subresourceClass";
+
+                foreach ($parentOperation['identifiers'] as $key => list($param, $class)) {
+                    $prefix .= 0 === $key ? "$class" : "-$param-$class";
```
dunglas: Useless quotes around `"$class"`.

src/Operation/PathNameGeneratorInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Operation;
+
+/**
+ * Generate a path name according to a string and whether it needs pluralization.
```
dunglas: Generates

src/PathResolver/DashOperationPathResolver.php
```diff
@@ -29,35 +28,16 @@
...
+        if (func_num_args() >= 4) {
+            $operationName = func_get_arg(3);
         } else {
-            $path = '/'.Inflector::pluralize($this->dashize($resourceShortName));
-        }
-
-        if ($operationType === OperationType::ITEM) {
-            $path .= '/{id}';
-        }
+            @trigger_error(sprintf('Method %s() will have a 4th `string $operationName` argument in version 3.0. Not defining it is deprecated since 2.1.', __METHOD__), E_USER_DEPRECATED);
```
dunglas: Is this deprecation necessary if the class itself is deprecated?

src/PathResolver/UnderscoreOperationPathResolver.php
```diff
@@ -29,30 +28,16 @@
...
+        if (func_num_args() >= 4) {
+            $operationName = func_get_arg(3);
         } else {
-            $path = '/'.RouteNameGenerator::routeNameResolver($resourceShortName, true);
-        }
-
-        if ($operationType === OperationType::ITEM) {
-            $path .= '/{id}';
-        }
+            @trigger_error(sprintf('Method %s() will have a 4th `string $operationName` argument in version 3.0. Not defining it is deprecated since 2.1.', __METHOD__), E_USER_DEPRECATED);
```
dunglas: same here

teohhanhui:
`PathNameGenerator` -> `PathSegmentGenerator` (what it really does)

src/Operation/PathNameGeneratorInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+interface PathNameGeneratorInterface
+{
+    /**
+     * Transforms a given string to a tableized, pluralized string.
```
teohhanhui: PHPDoc summary is wrong here.

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandTest.php
```diff
@@ -19,6 +19,7 @@
 
 /**
  * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ * @group legacy
```
soyuka: Insight on this?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -38,13 +38,22 @@ public function getConfigTreeBuilder()
         $treeBuilder = new TreeBuilder();
         $rootNode = $treeBuilder->root('api_platform');
 
+        // $rootNode->beforeNormalization()->always(function ($v) {
+        //     if (isset($v['default_operation_path_resolver'])) {
+        //         @trigger_error('The use of the `default_operation_path_resolver` has been deprecated in 2.1 and will be removed in 3.0. Use `path_name_generator` instead.');
+        //     }
+        // })->end();
```
soyuka: Should this hack be added to the `rootNode` or elsewhere?

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandTest.php
dunglas: Can you paste the error?

soyuka: Just a lot of deprecations because of the UnderscoreOperationResolver class deprecation.

dunglas: Our own code shouldn't trigger deprecations. Can you paste the stack trace?

soyuka:
I found some bugs while trying this on the framework package, still working on improving the `toOne` relations:

Get the `foobars` collection on `bar` belonging to `foo/1`:

```
/foos/1/bar/foobars
```

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandTest.php
soyuka:
```
Remaining deprecation notices (220)

The use of ApiPlatform\Core\PathResolver\UnderscoreOperationPathResolver is deprecated since 2.1. Please use PathSegmentNameGenerator instead: 219x
    219x in SwaggerCommandTest::testExecute from ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Command

The "api_platform.operation_path_resolver.default" service is deprecated since ApiPlatform 2.1 and will be removed in 3.0. Use PathSegmentNameGenerator instead: 1x
    1x in SwaggerCommandTest::testExecute from ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Command

Legacy deprecation notices (62)
```

This makes phpunit exit with code `1`. We're still using `UnderscoreOperationPathResolver`, should I already remove the service use in the code? If I do so, it will lead to breaking changes. 

Or I can just change the `api_platform.operation_path_resolver.default` but then again, BC break.

dunglas: Ok got it. It's fine this way.

soyuka:
Hmm, I'm not fond of the last commit, I think we want to avoid injecting the Router in this class (`DocumentationNormalizer`). 

Should I try to use `RouterOperationPathResolver` instead?

src/Operation/DashPathSegmentNameGenerator.php
```diff
@@ -0,0 +1,39 @@
...
+ * Generate a path name with an underscore separator according to a string and whether it needs pluralization.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class DashPathSegmentNameGenerator implements PathSegmentNameGeneratorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSegmentName(string $name, bool $pluralize = true): string
```
teohhanhui: I don't like the `$pluralize` argument. It should be up to the implementation to decide whether they want to pluralize or do other things. What we need to provide here is the context (e.g. whether this is for a collection or item).

dunglas:
Thank you very much @soyuka!

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandTest.php
tseho: I'm also having a lot of this deprecation notices. The library shouldn't trigger it, is there any way to use the new version without BC ?

dunglas: @tseho thanks for the report. Can you try if #1259 fixes the problem for you?

wlalele:
@soyuka What method would you recommend using to handle access authorization to a subresource ?
Is there any way I can use a « is_granted » property like in the operations ?
Id like to know what would you recommend doing on that subject.
BTW Great work on the project!

desmax:
@soyuka @wlalele Im also confused how to handle access control, please help

soyuka:
@wlalele @desmax it should not be different from other operations.

desmax:
@soyuka please give an example. where to define it? parent or child entity? maybe on property?

desmax:
@soyuka when I define it like this on child entity it seems it does not call my voters at all
```/**
 * @ApiResource(
 *  collectionOperations={
 *      "get"={
 *          "method"="GET",
 *          "access_control"="is_granted('get', object)"
 *     }
 *  },
 * )
 */

**Comment by desmax** [comment]
@soyuka But if I do
```
/**
 * @ApiResource(
 *  collectionOperations={
 *      "api_users_user_project_settings_get_subresource"={
 *          "method"="GET",
 *          "access_control"="is_granted('get', object)"
 *     }
 *  },
 * )
 */
```
It does work. A bit not obvious, I would say. 
P.S. User is my parent entity, UserProjectSettings is a child.

soyuka:
Yes you have to give the correct operation name and it's not easy to find/guess... Your second example is the good one. Sorry about the delay I'm on vacation.

PR #1198, opened 2017-06-24
Allow to use a service for dynamic validation groups

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

Usage:

```php
<?php

// src/ValidatorGroupsGenerator.php

class ValidatorGroupsGenerator
{
    private $s;

    public function __construct(AService $s)
    {
        $this->s = $s;
    }

    public function __invoke(): array
    {
        return $this->s->isSomething() ? ['a', 'b'] : ['a'];
    }
}
```

```yaml
services:
    App\ValidatorGroupsGenerator: ['@a_service']
```

```php
<?php

// src/Entity/Book.php

use ApiPlatform\Core\Annotation\ApiResource;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * @ApiResource(attributes={"validation_groups"="App\ValidatorGroupsGenerator"})
 */
class Book
{
    /**
     * @Assert\NotBlank(groups={"a"})
     */
    private $name;

    /**
     * @Assert\NotNull(groups={"b"})
     */
    private $author;

    // ...
}
```

TODO:

* [x] Add tests

PR #1249, opened 2017-07-17
Fix BC break in ValidateListener

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/1198#issuecomment-314716934
| License       | MIT
| Doc PR        | N/A

* Fix the bug/BC break introduced by #1198 (see https://github.com/api-platform/core/pull/1198#issuecomment-314716934)
* Allow to use a simple scalar instead of an array as validation group in attributes.

### DIFF
--- a/src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
+++ b/src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
@@ -74,6 +74,7 @@ final class ValidateListener
 
         if (
             $this->container &&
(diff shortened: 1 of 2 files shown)


dunglas:
Thanks @meyerbaptiste

PR #1251, opened 2017-07-17
Replace the old exception to status config with the new Symfony's YAML extension for PHP constants

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -186,6 +186,8 @@ final class Configuration implements ConfigurationInterface
                                 }
 
                                 if (defined($httpStatusCodeConstant = sprintf('%s::%s', Response::class, $httpStatusCode))) {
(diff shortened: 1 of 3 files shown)


dunglas:
What do you think about not deprecating the previous way (I really don't like the `!php...` notation). We can still use it internally.

teohhanhui:
The old way should be deprecated.

And the `!php/const` is YAML syntax (almost - pending symfony/symfony#22913). Users who don't like YAML syntax can already choose not to use YAML as the configuration file format.

meyerbaptiste:
I'm agree with @teohhanhui, we should only support the most official way of doing it and not also a kind of "homemade hack".

dunglas:
Let's do this so. Can you update the doc please @meyerbaptiste?

PR #1256, opened 2017-07-19
Improve performance of the metadata system

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

By default, the Symfony's `cache.system` setup uses apcu. Our metadata component relies on this cache, however, as we retrieve the cache many time in a single request `apcu_fetch` is called many times and do a lot of IO. This fix introduce a local memory cache to avoid the IO operations.

### DIFF
--- a/src/Metadata/Property/Factory/CachedPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/CachedPropertyMetadataFactory.php
@@ -28,6 +28,7 @@ final class CachedPropertyMetadataFactory implements PropertyMetadataFactoryInte
 
     private $cacheItemPool;
     private $decorated;
+    private $memoryCache = [];
 
     public function __construct(CacheItemPoolInterface $cacheItemPool, PropertyMetadataFactoryInterface $decorated)
     {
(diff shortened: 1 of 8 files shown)


tests/Metadata/Resource/Factory/CachedResourceMetadataFactoryTest.php
```diff
@@ -86,6 +86,6 @@ public function testCreateWithGetCacheItemThrowsCacheException()
 
     private function generateCacheKey(string $resourceClass = Dummy::class)
     {
-        return CachedResourceMetadataFactory::CACHE_KEY_PREFIX.md5(serialize([$resourceClass]));
```
soyuka: @teohhanhui you had a good reason to serialize here? (I think it comes from you)

Do we really need the `md5` thing?

soyuka:
Nice one!
Is there no added value when doing the same on the `NameCollectionFactories`?

We may want to add the same thing to `CachedIdentifiersExtractor`, IIRC it was called a lot too. IMO any micro optimization is good here.

dunglas:
I missed NameCollectionFactoty. I'll add it. I'll do it for the identifier too.

tests/Metadata/Resource/Factory/CachedResourceMetadataFactoryTest.php
dunglas: This is necessary because the key must not exceed 64 chars (PSR-6). A long class name may exceed this.

src/Metadata/Property/Factory/CachedPropertyMetadataFactory.php
```diff
@@ -28,6 +28,7 @@
 
     private $cacheItemPool;
     private $decorated;
+    private $memoryCache = [];
```
teohhanhui: `$memoryCache` is not a very descriptive name... :disappointed:

Perhaps `$localCache` and `$localCacheKey`. Descriptive is better.

dunglas: @teohhanhui however it's exactly what it is, a memory cache.

teohhanhui: APCu is also a (shared) memory cache. This is the local memory cache. :laughing:

teohhanhui: The word memory is just redundant and meaningless since all the variables are in memory anyway...

sstok: How about `loaded` ?

PR #1257, opened 2017-07-19
Improve the performance of CachedIdentifiersExtractor

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT

Follows #1256.

### DIFF
--- a/src/Api/CachedIdentifiersExtractor.php
+++ b/src/Api/CachedIdentifiersExtractor.php
@@ -33,6 +33,7 @@ final class CachedIdentifiersExtractor implements IdentifiersExtractorInterface
     private $cacheItemPool;
     private $propertyAccessor;
     private $decorated;
+    private $memoryCache = [];
 
     public function __construct(CacheItemPoolInterface $cacheItemPool, IdentifiersExtractorInterface $decorated, PropertyAccessorInterface $propertyAccessor = null)
     {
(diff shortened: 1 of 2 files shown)


soyuka:
hmm tests failed though

dunglas:
I'm on it

meyerbaptiste:
Could we not do the same for all `Cached*` classes?

src/Api/CachedIdentifiersExtractor.php
```diff
@@ -33,6 +33,7 @@
     private $cacheItemPool;
     private $propertyAccessor;
     private $decorated;
+    private $memoryCache = [];
```
teohhanhui: Naming... See https://github.com/api-platform/core/pull/1256#discussion_r128330279

dunglas:
I've tried something but I would like a review @soyuka. I'm not sure I respect your initial intent.

src/Api/CachedIdentifiersExtractor.php
```diff
@@ -46,60 +47,66 @@ public function __construct(CacheItemPoolInterface $cacheItemPool, IdentifiersEx
...
-        $identifiers = [];
-        $resourceClass = $this->getObjectClass($item);
-
-        $cacheKey = self::CACHE_KEY_PREFIX.md5($resourceClass);
-
-        // This is to avoid setting the cache twice in the case where the related item cache doesn't exist
-        $cacheIsHit = false;
+        $keys = $this->getKeys($item, function ($item) use (&$identifiers) {
+            return $identifiers = $this->decorated->getIdentifiersFromItem($item);
+        });
```
soyuka: this is cool stuff!

teohhanhui: :man_shrugging:

soyuka:
@meyerbaptiste this one is slightly different and more complex then metadata classes. Initial PR https://github.com/api-platform/core/pull/997

src/Api/CachedIdentifiersExtractor.php
soyuka: Using a reference to an unknown variable and declaring it in a closure looks cool to me haha :sweat_smile:

meyerbaptiste:
Yep @soyuka, I posted my message before he refactored his PR. My previous comment is now more relevant for #1256.

soyuka:
Yeah I agree we have a lot of repetition in those "cached" classes, finally it's always the same:
1. check cache (key is PREFIX + md5(resourceClass))
2. if cached return cached value
3. if no cache call decorator and cache

Do you have an idea to remove the duplicated code (only diff is the injected decorator and the cache key prefix)?

teohhanhui:
Is it safe for us to use `Symfony\Component\Cache\Adapter\ArrayAdapter` with `$storeSerialized = false`? Since our metadata classes are all immutable... And `Symfony\Component\PropertyInfo\Type` is immutable too.

teohhanhui:
In the case of `CachedIdentifiersExtractor` we are storing an array of strings, so it's "immutable" too.

dunglas:
`ArrayAdapter` will be slower (more function calls) and - as pointed out by Nicolas Grekas on Twitter - is not part of `cache.cache` because it's always slower than a local cache. Symfony components use a memory cache too.

dunglas:
We can probably go one step further for this class: it should be possible to cache (in a memory cache) the results as well.

tests/Api/CachedIdentifiersExtractorTest.php
```diff
@@ -45,7 +45,9 @@ public function testFirstPass()
 
         $identifiersExtractor = new CachedIdentifiersExtractor($cacheItemPool->reveal(), $decoration->reveal(), null);
 
-        $this->assertEquals(['id' => 1], $identifiersExtractor->getIdentifiersFromItem($dummy));
+        $expectedResult = ['id' => 1];
+        $this->assertEquals($expectedResult, $identifiersExtractor->getIdentifiersFromItem($dummy));
+        $this->assertEquals($expectedResult, $identifiersExtractor->getIdentifiersFromItem($dummy), 'Trigger the local cache');
```
teohhanhui: How does this test verify that we are triggering the local cache? Also, it should not matter in a unit test.

soyuka: He's hacking coverage to please @Simperfit :laughing:

soyuka:
@dunglas I don't think that caching values is worth it. Also cache invalidation will be an issue. This is already improving performances a lot because we avoid calling the whole metadataFactory thing from `IdentifierExtractor` to find identifiers (simple and sufficient IMO).

tests/Api/CachedIdentifiersExtractorTest.php
teohhanhui: Actually, this second call is crucial to test correct behaviour, i.e. that we return the correct value when cache is used. But it's not testing whether we're triggering the local cache.

teohhanhui: Oh wait, we already have different test cases for that... Lol

dunglas: Not really for the coverage. You're right @teohhanhui, it doesn't verify that the local cache actually works (it's very hard to test reliably), however it tests if the local cache don't introduce a bug.

For instance, if some one contribute this patch:

```patch
-return $this->memoryCache[$resourceClass];
+return $this->memoryCache[$resourceC];
```

The test will detect the problem. Without this line, it will not.

teohhanhui: @dunglas Doesn't `testSecondPass` already do that?

dunglas: Good catch. `testSecondPass` should be update to bypass the local cache IMO (by creating a new instance).

teohhanhui: `$localCache` is unambiguous. #preach

PR #1258, opened 2017-07-19
Fix insight notice

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a


src/Security/EventListener/DenyAccessListener.php
```diff
@@ -106,7 +107,7 @@ private function getVariables(Request $request): array
             'user' => $token->getUser(),
             'object' => $request->attributes->get('data'),
             'request' => $request,
-            'roles' => array_map(function ($role) {return $role->getRole(); }, $roles),
+            'roles' => array_map(function (Role $role) {return $role->getRole(); }, $roles),
```
teohhanhui:
```php
'roles' => array_map(function (Role $role) {
    return $role->getRole();
}, $roles),
```

Maybe it's common practice in Symfony CS, but PSR-2 is very clear on this:

> The opening brace MUST go on the same line, and **the closing brace MUST go on the next line following the body**.

(emphasis mine)

Issue #1210, opened 2017-06-28
EagerLoadingExtension denormalization operation scope issue


soyuka:
When denormalizing, `IriConverter::getItemFromIri` may get called, and therefore calls the `ItemDataProvider`. When doing this, we loose the `context`.

I saw this while trying to use `force_eager = false` on an Operation. When calling this Operation, the first call to `ItemDataProvider` has the correct `operation_name`. But then, when the call comes from `getItemFromIri`, the `operation_name` is `null`. Therefore, `force_eager` is not fetched from the initial Operation attributes.

soyuka:
Cause: https://github.com/api-platform/core/blob/master/src/Serializer/AbstractItemNormalizer.php#L362 

@dunglas may I ask why you're removing those context properties when serializing a relation? 

Because some configurations rely on `operation_name` we should propagate those to the `(de)normalization` of relations no? 

For example:

```yaml
Product:
  itemOperations:
     put:
        method: 'PUT'
        attributes: ['force_eager' => false]
```

Assuming `Product->Relation->Foo`, when `Relation` will get `denormalize`'d the `force_eager` flag will be `true` because we lost the `operation_name` information.

dunglas:
I don't remember exactly :( But if you remove this line, I bet some tests will fail and you'll get an explanation.

Simperfit:
This seems to be fixed by PR #1264

Issue #1222, opened 2017-07-03
Excpetion during normalization of embedded entities where not all are ApiResources


backbone87:
I have a problem with normalization.

Datamodel: Fall (Case) <- 1:n -> Tag <- n:1 -> Tagdefinition

All 3 models are doctrine entities, but only Fall and Tagdefinition are ApiResources. Tags should be serialized within the Fall output, something like this:
```json
{
  "@context": "/contexts/Fall",
  "@id": "/faelle",
  "@type": "hydra:Collection",
  "hydra:member": [
    {
      "@id": "/faelle/691bcf98-5c41-11e7-b7d5-046d49573cb1",
      "@type": "Fall",
      "id": "691bcf98-5c41-11e7-b7d5-046d49573cb1",
      "tags": [
        {
          "erstellungsdatum": "2017-06-30T15:09:47+00:00",
          "beschriftung": "Test"
        }
      ]
    }
  ],
  "hydra:totalItems": 1
}
```
This works fine up until now. I have a `getTags` on the Fall entity and the beschriftung (label) and erstellungsdatum (created date) are within the normalization group and get printed. however i dont know which tag it is. i want to serialize the Tagdefinition too (IRI and/or embedded within each tag). if i add the serialization group to the foreignkey "tagdefinition" in the Tag entity, i get the following error:
```
ApiPlatform\Core\Exception\InvalidArgumentException:
No resource class found for object of type "App\Entity\Tagdefinition".

  at vendor/api-platform/core/src/Api/ResourceClassResolver.php:58
  at ApiPlatform\Core\Api\ResourceClassResolver->getResourceClass(object(Tagdefinition), 'App\\Entity\\Fall', true)
     (vendor/api-platform/core/src/JsonLd/Serializer/ItemNormalizer.php:65)
[...]

PR #1264, opened 2017-07-20
Keep propagating the correct context to relations when (de)normalizing

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1210 
| License       | MIT
| Doc PR        | n/a

And...... Nothing happened :D.

Though, really unsure on how I could test this to avoid regression. There's no point in unit-testing it because it's not impacting the normalization, instead it may impact features that rely on the same context. Also, we can't test if a feature has been executed (like `force_eager=false`) with behat...

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -292,7 +292,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         }
 
         if (!$this->resourceClassResolver->isResourceClass($className) || $propertyMetadata->isWritableLink()) {
-            return $this->serializer->denormalize($value, $className, $format, $this->createRelationSerializationContext($className, $context));
+            $context['resource_class'] = $className;
+
+            return $this->serializer->denormalize($value, $className, $format, $context);
         }
 
         if (!is_array($value)) {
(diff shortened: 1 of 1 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -349,22 +350,6 @@ protected function getFactoryOptions(array $context): array
...
 
     /**
-     * Creates the context to use when serializing a relation.
-     *
-     * @param string $resourceClass
-     * @param array  $context
-     *
-     * @return array
-     */
-    protected function createRelationSerializationContext(string $resourceClass, array $context): array
```
dunglas: We cannot remove this method as it is protected.

dunglas: However you can stop using it and deprecate it.

backbone87:
This may be related to #1222

soyuka:
exactly what I thought @backbone87 while reviewing your issues. I'm not sure that this will be enough to fix your problem though :|.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -358,10 +360,7 @@ protected function getFactoryOptions(array $context): array
      */
     protected function createRelationSerializationContext(string $resourceClass, array $context): array
     {
-        $context['resource_class'] = $resourceClass;
-        unset($context['item_operation_name'], $context['collection_operation_name']);
-
-        return $context;
```
dunglas: This line should be kept

dunglas:
Thanks @soyuka

PR #805, opened 2016-10-17
Gpenverne fix/allow empty content on post and put requests

dunglas:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #781, #782 |
| License | MIT |
| Doc PR | n/a |

Follows #782. All credit go to @gpenverne.

PR #782, opened 2016-09-30
Fix/allow empty content on post and put requests

gpenverne:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #781
| License       | MIT

According issue with @kbsali and @teohhanhui ( #781 ) , this PR should fix error on PUT request with empty body.

PR #1274, opened 2017-07-21
Authorize empty content put in custom operations

thomasglachant:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT
| Doc PR        | NA

This PR allows to send empty request content for PUT in a custom operation.

Use case : 
I have a custom Action (ex : "reset") associated to an ApiResource "Job". 
I don't have any request content to send when I send a PUT request to `/jobs/1234/reset`.
But I've a SerializerError because the request content is required.

### DIFF
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -395,6 +395,41 @@ Feature: Create-Retrieve-Update-Delete
     }
     """
 
+  Scenario: Update a resource with empty body
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/dummies/1"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/Dummy",
+      "@id": "/dummies/1",
+      "@type": "Dummy",
+      "description": null,
+      "dummy": null,
+      "dummyBoolean": null,
+      "dummyDate": "2015-03-01T10:00:00+00:00",
+      "dummyFloat": null,
+      "dummyPrice": null,
+      "relatedDummy": null,
+      "relatedDummies": [],
+      "jsonData": [
+        {
+          "key": "value1"
+        },
+        {
+          "key": "value2"
+        }
+      ],
+      "name_converted": null,
+      "id": 1,
+      "name": "A nice dummy",
+      "alias": null
+    }
+    """
+
   @dropSchema
   Scenario: Delete a resource
     When I send a "DELETE" request to "/dummies/1"
--- a/src/EventListener/DeserializeListener.php
+++ b/src/EventListener/DeserializeListener.php
@@ -52,7 +52,7 @@ final class DeserializeListener
             || $request->isMethod(Request::METHOD_DELETE)
             || !($attributes = RequestAttributesExtractor::extractAttributes($request))
             || !$attributes['receive']
-            || ($request->isMethod(Request::METHOD_PUT) &&  '' === $requestContent)
+            || ($request->isMethod(Request::METHOD_PUT) && '' === $requestContent)
         ) {
             return;
         }
--- a/src/EventListener/DeserializeListener.php
+++ b/src/EventListener/DeserializeListener.php
@@ -51,6 +51,7 @@ final class DeserializeListener
             || $request->isMethod(Request::METHOD_DELETE)
             || !($attributes = RequestAttributesExtractor::extractAttributes($request))
             || !$attributes['receive']
+            || ($request->isMethod(Request::METHOD_PUT) &&$request->getContent() === "")
         ) {
             return;
         }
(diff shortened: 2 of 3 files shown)


src/EventListener/DeserializeListener.php
```diff
@@ -51,6 +51,7 @@ public function onKernelRequest(GetResponseEvent $event)
             || $request->isMethod(Request::METHOD_DELETE)
             || !($attributes = RequestAttributesExtractor::extractAttributes($request))
             || !$attributes['receive']
+            || ($request->isMethod(Request::METHOD_PUT) && $request->getContent() === '')
```
dunglas: Can you store the result of `$request->getContent()` in a variable to avoid 2 (costly) method calls (the next one is line 70)?
Also, please use Yoda styl (`'' === $content`).

src/EventListener/DeserializeListener.php
```diff
@@ -52,7 +52,7 @@ public function onKernelRequest(GetResponseEvent $event)
             || $request->isMethod(Request::METHOD_DELETE)
             || !($attributes = RequestAttributesExtractor::extractAttributes($request))
             || !$attributes['receive']
-            || ($request->isMethod(Request::METHOD_PUT) &&  '' === $requestContent)
+            || ($request->isMethod(Request::METHOD_PUT) && '' === $requestContent)
```
dunglas: May I suggest:

`|| ($request->isMethod(Request::METHOD_PUT) && '' === ($requestContent = $request->getContent()))`

It will avoid to always retrieve the body (that can be huge) in most cases.

meyerbaptiste:
> in custom operation

But in reality it will work for all `PUT` operations, custom or not. Do we really want this?

dunglas:
What's the behavior with this patch when an empty request is sent to a standard PUT route?

src/EventListener/DeserializeListener.php
thomasglachant: Ok, I can do it, but it's a little more complicated because it's a "&&" and for a POST we don't initialize the variable. 
So I need to add `$requestContent ?? $request->getContent()` in the deserialize line.

It will be faster but less clear.

meyerbaptiste:
We have a 200 with the unmodified resource instead of a 400.

thomasglachant:
> What's the behavior with this patch when an empty request is sent to a standard PUT route?

Nothing (200) ... I wrote a behat test to check. It's like a PUT request with an empty array in body.

src/EventListener/DeserializeListener.php
dunglas: As this listener will be executed for every request (even if its a page not managed by API Platform), let's use the fastest alternative.

dunglas:
It looks acceptable to me. WDYT @api-platform/core-team?

soyuka:
Ok to me.

src/EventListener/DeserializeListener.php
thomasglachant: done ! I add a behat test with the empty body ... don't know if it's necessary ...

meyerbaptiste:
It looks like a BC break to have a 200 instead of a 400 now. Nop?

soyuka:
Lot of overhead if we need to check that it's a custom operation no?

dunglas:
It's borderline, but it looks acceptable. Who would rely on an empty request throwing a 400?

src/EventListener/DeserializeListener.php
```diff
@@ -46,11 +46,13 @@ public function __construct(SerializerInterface $serializer, SerializerContextBu
     public function onKernelRequest(GetResponseEvent $event)
     {
         $request = $event->getRequest();
+        $requestContent = null;
         if (
             $request->isMethodSafe(false)
             || $request->isMethod(Request::METHOD_DELETE)
             || !($attributes = RequestAttributesExtractor::extractAttributes($request))
             || !$attributes['receive']
+            || ($request->isMethod(Request::METHOD_PUT) && '' === ($requestContent = $request->getContent()))
```
dunglas: If you invert the conditions (`'' === ($requestContent = $request->getContent())` && $request->isMethod(Request::METHOD_PUT)`), you don't need the trick with `??` anymore because the variable will always be defined. WDYT?

thomasglachant: Good idea, clean and fast 👍

features/main/crud.feature
```diff
@@ -395,6 +395,41 @@ Feature: Create-Retrieve-Update-Delete
     }
     """
 
+  Scenario: Update a resource with empty body
```
thomasglachant: @dunglas Do you think it's necessary to add this test ? I'm not convinced ...

soyuka: There are no useless integration test ;).

Simperfit:
It's not like someone should rely on a 400 return. LGTM

soyuka:
Thanks @thomasglachant !

teohhanhui:
I really do not understand why this needs to be specially handled. What happened to the PR(s) on allowing empty request body?

soyuka:
Empty request body does make sense only with PUT / DELETE requests no?

teohhanhui:
No. See #782 and #805.

PR #1279, opened 2017-07-24
Refactor cycle breaking (visited tracking) in SubresourceOperationFactory

teohhanhui:
…tory

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Refactor with visited tracking logic which is much easier to follow.

### DIFF
--- a/src/Operation/Factory/SubresourceOperationFactory.php
+++ b/src/Operation/Factory/SubresourceOperationFactory.php
@@ -58,8 +58,9 @@ final class SubresourceOperationFactory implements SubresourceOperationFactoryIn
      * @param array  $tree
      * @param string $rootResourceClass null on the first iteration, it then keeps track of the origin resource class
      * @param array  $parentOperation   the previous call operation
+     * @param array  $visited
      */
-    private function computeSubresourceOperations(string $resourceClass, array &$tree, string $rootResourceClass = null, array $parentOperation = null, $parentVisiting = null)
+    private function computeSubresourceOperations(string $resourceClass, array &$tree, string $rootResourceClass = null, array $parentOperation = null, array $visited = [])
     {
         if (null === $rootResourceClass) {
             $rootResourceClass = $resourceClass;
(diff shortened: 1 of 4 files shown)


teohhanhui:
PHPStan failure should be addressed in a separate PR.

src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -137,9 +131,9 @@ private function computeSubresourceOperations(string $resourceClass, array &$tre
                 $operation['path'] .= sprintf('/%s%s', $this->pathSegmentNameGenerator->getSegmentName($property, $operation['collection']), self::FORMAT_SUFFIX);
             }
 
-            $tree[$visiting] = $operation;
+            $tree[$operation['route_name']] = $operation;
 
-            $this->computeSubresourceOperations($subresourceClass, $tree, $rootResourceClass, $operation, $visiting);
+            $this->computeSubresourceOperations($subresourceClass, $tree, $rootResourceClass, $operation, array_merge($visited, [$visiting]));
```
meyerbaptiste:
```php
$visited[] = $visiting;

$this->computeSubresourceOperations(..., $visited);
```

Is it not better for performance?

teohhanhui: No, that would give the wrong result. We could of course use another variable.

meyerbaptiste: Hum, I think I missed something here... Why is the result wrong? It's not an associative array here, right?

teohhanhui: The result would be wrong because we have a loop, if we modify the `$visited` array in place it'd affect the other properties.

meyerbaptiste: Oh sorry, I missed the loop! BTW it's better like this with an associative array! 👍

soyuka:
Merged in #1283 Thanks @teohhanhui

PR #1283, opened 2017-07-24
Fix/psr 6 cache subresource

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Operation/Factory/CachedSubresourceOperationFactory.php
+++ b/src/Operation/Factory/CachedSubresourceOperationFactory.php
@@ -37,7 +37,7 @@ final class CachedSubresourceOperationFactory implements SubresourceOperationFac
      */
     public function create(string $resourceClass): array
     {
-        $cacheKey = self::CACHE_KEY_PREFIX.str_replace('\\', '', $resourceClass);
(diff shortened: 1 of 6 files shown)


src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -36,7 +36,7 @@
             <argument type="service" id="service_container" />
             <argument>%api_platform.formats%</argument>
             <argument>%api_platform.resource_class_directories%</argument>
-            <argument type="service" id="api_platform.subresource_operation_factory" />
+            <argument type="service" id="api_platform.subresource_operation_factory.cached" />
```
meyerbaptiste: Why? `api_platform.subresource_operation_factory` is decorated so you'll have an instance of `CachedSubresourceOperationFactory`, nop?

soyuka: Hmm yeah you're right, what was I thinking... Monday.

PR #1296, opened 2017-07-31
Fix #1224 json-ld context should not be prefixed by #

### LINKED ISSUE #1224: [JSON-LD] Context IRIs are assembled wrong
There is a `#` too much:
```json
"@context": {
  "@vocab": "http://localhost/docs.jsonld#",
  "version": "#Fall/version",
  "erstellungsdatum": "#Fall/erstellungsdatum",
}
```

But it should be:
```json
"@context": {
  "@vocab": "http://localhost/docs.jsonld#",
  "version": "Fall/version",
  "erstellungsdatum": "Fall/erstellungsdatum",
}
```

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1224
| License       | MIT
| Doc PR        | na

According to the specs I think that the `#` character is there to separate the URL from the given element, for example:

```
    "@context": {
      "name": "http://example.com/organization#name"
    },
```

I'm really no json-ld expert, let me know if removing `#` doesn't seem correct to you.

### DIFF
--- a/src/JsonLd/ContextBuilder.php
+++ b/src/JsonLd/ContextBuilder.php
@@ -89,7 +89,7 @@ final class ContextBuilder implements ContextBuilderInterface
     {
         $context = $this->getBaseContext($referenceType);
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
-        $prefixedShortName = sprintf('#%s', $resourceMetadata->getShortName());
+        $shortName = $resourceMetadata->getShortName();
 
         foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $propertyName) {
             $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
@@ -102,7 +102,7 @@ final class ContextBuilder implements ContextBuilderInterface
             $jsonldContext = $propertyMetadata->getAttributes()['jsonld_context'] ?? [];
 
             if (!$id = $propertyMetadata->getIri()) {
-                $id = sprintf('%s/%s', $prefixedShortName, $convertedName);
+                $id = sprintf('%s/%s', $shortName, $convertedName);
             }
 
             if (true !== $propertyMetadata->isReadableLink()) {
--- a/features/jsonld/context.feature
+++ b/features/jsonld/context.feature
@@ -25,24 +25,24 @@ Feature: JSON-LD contexts generation
             "@vocab": "http://example.com/docs.jsonld#",
             "hydra": "http://www.w3.org/ns/hydra/core#",
             "description": "https://schema.org/description",
-            "dummy": "#Dummy/dummy",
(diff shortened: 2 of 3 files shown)


src/JsonLd/ContextBuilder.php
```diff
@@ -89,7 +89,7 @@ public function getResourceContext(string $resourceClass, int $referenceType = U
     {
         $context = $this->getBaseContext($referenceType);
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
-        $prefixedShortName = sprintf('#%s', $resourceMetadata->getShortName());
+        $prefixedShortName = sprintf('%s', $resourceMetadata->getShortName());
```
meyerbaptiste: `sprintf` is now useless 😛

soyuka: hahaha omg

soyuka:
Thanks @meyerbaptiste 

http://thecodinglove.com/post/126895236416/first-bugfix-on-monday

dunglas:
Please check if our frontend tools still work with this change.

soyuka:
> Please check if our frontend tools still work with this change.

:+1: this is one of my concerns, ping @mysiar

soyuka:
@dunglas I checked in neither `admin` nor `crud-generate` are we using the JSON-LD `/context` endpoint. Something else we'd need to check?

For release-safety I can rebase this on `master` instead of `2.0` and we flag it in the changelog for the next stable. WDYT?

dunglas:
I'm sure we are using it when we normalize the JSON-LD document (the jsonld library used by api-doc-parser does it under the hood).

mysiar:
please check https://www.w3.org/TR/json-ld/#the-context Example 24 and description, it might be relevant

soyuka:
Got it. We should add a test on the [json-ld repository](https://github.com/digitalbazaar/jsonld.js/blob/master/tests/contexts/context-1.jsonld) to see if it's ok.

soyuka:
In fact according to the json-ld tests this patch will work just fine as is closer to the specs. 

Or we may include the api url instead of using relative paths.

dunglas:
Ok let's merge this when the build will be green.

soyuka:
@dunglas tests pass, just appveyor having issues with php not being installed?!

meyerbaptiste:
Try to restart your build, looks like this happens occasionally with chocolatey.org...

soyuka:
:+1: I'm not able to do it I think @dunglas can.

meyerbaptiste:
Force push 😛

dunglas:
Looks good, feel free to merge @soyuka.

PR #1297, opened 2017-08-01
Allow to fetchEager a non-serializable property fix #1290

### LINKED ISSUE #1290: Eager loading a relation without serializing it
Hey there,

I'd like to eager load an association *without* having it embedded, as I serialize a computed value from the association. Example:

```php
/**
 * @ApiResource(attributes={
 *     "normalization_context" = {"groups" = {"foo"}},
 * })
 */
class Foo
{
    /**
     * @var Collection|Amount[]
     *
     * @ORM\OneToMany(
     *     targetEntity="Amount",
     *     mappedBy="foo"
     * )
     */
    protected $amounts;

    /**
     * @return Money
     * @Groups({"foo"})
     */
    public function getTotalAmount()
    {
        $total = new Money(0, new Currency('USD'));

        foreach ($this->amounts as $amount) {
            $total = $total->add($amount->getAmount());
        }

        return $total;
    }
}
```
Clearly i want `amounts` eager loaded via a left join.
Is there a way to achieve this?

I realise that I could do this in an Collection/Item Extension:
```php
final class FooLoadedExtension implements QueryCollectionExtensionInterface, QueryItemExtensionInterface
{
    private function addWhere(QueryBuilder $queryBuilder, string $resourceClass): void
    {
        if (Foo::class === $resourceClass) {
            $rootAlias = $queryBuilder->getRootAliases()[0];
            $queryBuilder->addSelect('amounts');
            $queryBuilder->leftJoin(sprintf('%s.amounts', $rootAlias), 'amounts');
        }
    }
}
```

But this is dangerous, as I could easily end up with a double left join if:
- started filtering on nested amounts or
- i later decided to embed the `$amounts` but forgot to remove my Extension left join, or

So really i don't find the extension an acceptable solution.

Thanks!

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1290
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/233

The following will bypass `readable` and `readableLink`:

```
@ApiProperty(attributes={"fetchEager": true})
```

ping @bendavies could you give this a try?

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -156,7 +156,7 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
                 continue;
             }
 
-            if (false === $propertyMetadata->isReadableLink() || false === $propertyMetadata->isReadable()) {
+            if ((false === $propertyMetadata->isReadableLink() || false === $propertyMetadata->isReadable()) && false === $propertyMetadata->getAttribute('fetchEager', false)) {
                 continue;
             }
 
--- a/tests/Bridge/Doctrine/Orm/Extension/EagerLoadingExtensionTest.php
+++ b/tests/Bridge/Doctrine/Orm/Extension/EagerLoadingExtensionTest.php
@@ -665,4 +665,49 @@ class EagerLoadingExtensionTest extends \PHPUnit_Framework_TestCase
         $eagerExtensionTest = new EagerLoadingExtension($propertyNameCollectionFactoryProphecy->reveal(), $propertyMetadataFactoryProphecy->reveal(), $resourceMetadataFactoryProphecy->reveal(), 30);
         $eagerExtensionTest->applyToCollection($queryBuilder, new QueryNameGenerator(), Dummy::class);
     }
+
+    public function testApplyToCollectionWithANonRedableButFetchEagerProperty()
+    {
+        $resourceMetadataFactoryProphecy = $this->prophesize(ResourceMetadataFactoryInterface::class);
+        $resourceMetadataFactoryProphecy->create(Dummy::class)->willReturn(new ResourceMetadata());
+
+        $propertyNameCollectionFactoryProphecy = $this->prophesize(PropertyNameCollectionFactoryInterface::class);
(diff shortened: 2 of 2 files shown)


bendavies:
@soyuka thanks for the quick work!
There is a bug now, however. your new if statment for `$readable` will exit early, where as the old one checks all conditions. This causes all relations to eager join.

soyuka:
Hmm yes indeed updated ^^.

bendavies:
Perfect, works for me!

bendavies:
Obviously the docs should be updated, but also clarified, as at the moment the eager loading page (https://api-platform.com/docs/core/performance) doesn't really say that eager loading is link to the serialization process.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -157,7 +157,9 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
             }
 
             if (false === $propertyMetadata->isReadableLink() || false === $propertyMetadata->isReadable()) {
-                continue;
+                if (false === $propertyMetadata->getAttribute('fetchEager', false)) {
```
meyerbaptiste: You can merge this condition with the previous one!

soyuka: not really :|. I tried things in my head but it doesn't seem to end up well because it's not a condition to break but rather a condition to not break when we should have.

bendavies: plus this is far more readable.

soyuka: okay got it I think:

```
if ((false === $propertyMetadata->isReadableLink() || false === $propertyMetadata->isReadable()) && false === $propertyMetadata->getAttribute('fetchEager', false)) {
    continue;
}
```

bendavies:
good to merge @soyuka ?

Simperfit:
Needs rebase but good to merge.

PR #1305, opened 2017-08-03
Add hydra items per page

marcw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

I publish this PR to raise (and maybe solve at the same time) a question : 
How can we put the number of items per page in a hydra response? The official spec doesn't say anything but this is a very needed feature for every app developers out there. 

This PR tries to solve this answer by adding an "itemsPerPage" element in the JSON response (same as an HAL answer), next to the `hydra:totalItems` element.

I am clearly not an expert on Hydra, so comments are very welcome.

### DIFF
--- a/src/Hydra/Serializer/CollectionNormalizer.php
+++ b/src/Hydra/Serializer/CollectionNormalizer.php
@@ -90,6 +90,7 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
 
         if (is_array($object) || $object instanceof \Countable) {
             $data['hydra:totalItems'] = $object instanceof PaginatorInterface ? $object->getTotalItems() : count($object);
+            $data['itemsPerPage'] = $object instanceof PaginatorInterface ? $object->getItemsPerPage() : null;
         }
 
         return $data;


meyerbaptiste:
Huum, this is really not Hydra compliant. `itemsPerPage` was removed from the Hydra spec, see https://github.com/HydraCG/Specifications/issues/42#issuecomment-151644982.

marcw:
Is hydra extensible ?

meyerbaptiste:
The question is: why do you need this value?

marcw:
I want this value because it is needed in my API client (a web UI). Correct me if I'm wrong but this is not a value you can compute easily from a response as it depend from a lot of different factors.

Simperfit:
@marcw You could be able to compute this value by using the totalItem and the last.

totalItem divided by maxPage should be equal to itemPerPage.

marcw:
Not really. If you got 55 items on 2 pages you cannot compute the right itemsPerPage (in this case 30)

Simperfit:
Yeah, was thinking of a simple case, with simple value, my bad ^^'.

I don't know what could be better, but this is not the right way to do it since we should follow the spec. But you could decorate it to return it. But the value will be the same every time. So maybe you should specify it in the API and in the Admin Client.

soyuka:
> But the value will be the same every time.

We can override this per-resource/per-operation basis so not necessarily.

marcw:
My question still stands: Is there a way to extend hydra and still be hydra-compliant?

soyuka:
Still why do you want this value if you have first/next/last/total ? 

I'm not seeing anything in the specs that would allow custom metadata to be added.

marcw:
Alright, closing this then. Thanks

soyuka:
Anyway people may still be interesting in this, especially when it comes to custom metadata! @marcw maybe worth opening an issue in HydraCG/Specifications or asking the mailing list ?

marcw:
@soyuka I am honestly lacking time to push things forward. Maybe in a few weeks ! If you want to handle this, feel free to do so :)

dunglas:
Hydra is a JSON-LD vocabulary. You can mix as vocabularies you want (including custom ones) in the same document. If you want to include custom metadata, just add extra keys (not starting by hydra:) and you're done. You can also introduce your own custom prefix and namespace if you want (it's what API Platform does for properties not mapped with an IRI).

marcw:
@dunglas So in that case, is there a way to include that value in the JSON returned by the API? Looking at the PR, what would be the best key for that?

Issue #1303, opened 2017-08-03
Performance regression in 2.1 beta 2


raziel057:
I just tried to upgrade to 2.1 beta 2 where I didn't found any bugs but a regression in execution time when returning a lot of results. Memory also increases a lot. 

As you can see when I get 9200 results (id, name):

In 2.0.10 it takes about 600ms

![api-platform-2 0 1](https://user-images.githubusercontent.com/652505/28912531-ec9533e6-7834-11e7-8b49-31da165fcc81.png)

With 2.1.0 beta 2: 7s

![api-platform-2 1 0-beta-2](https://user-images.githubusercontent.com/652505/28912551-fdfbe2d8-7834-11e7-9061-e054244abcd5.png)

I tried to downgrade, upgrade multiple times and I always have the same results. The regression seems to be linked to the serialisation part. Could you please check?

Simperfit:
Is it in dev / prod ?

is it at first run ?

raziel057:
I have the same penalty in DEV and PROD env. Not only at first run, it's always the case.

Simperfit:
@raziel057 Could you please try with current master ? Some cache improvement have been done in master.

meyerbaptiste:
Did you try to run Blackfire on your project to localize the problem? It could help us to have a shared profile.

raziel057:
@Simperfit I just tried to upgrade to dev-master:

Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Updating api-platform/core (v2.0.10 => dev-master 4fffea6):  Checking out 4fffea6519

But I have the same (bad perf) results.

@meyerbaptiste I will create a Blackfire profile this afternoon.

soyuka:
Can you also check the `Cache` section?

raziel057:
Each time I upgrade / downgrade I clean the cache and restart my server. Each time I have a perf issue in 2.1.0 and not in 2.0.10

raziel057:
Please find here the Blackfire profile:
https://blackfire.io/profiles/compare/5a87c426-1908-485d-aed4-7ae05d5f7c99/graph

2.0.10 (env DEV):

```
[...]

PR #1307, opened 2017-08-04
	Avoid serializing "resources" in the normalization context

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1303 
| License       | MIT
| Doc PR        | na

Because [getCacheKey](https://github.com/symfony/serializer/blob/master/Normalizer/AbstractObjectNormalizer.php#L361) serializes the `context`, let's avoid serializing the `$context['resources']` that can get pretty big.

Solves the performance issue.

### DIFF
--- a/src/EventListener/SerializeListener.php
+++ b/src/EventListener/SerializeListener.php
@@ -59,13 +59,17 @@ final class SerializeListener
         }
 
         $context = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
-        $resources = [];
+        $resources = new class() extends \ArrayObject {
+            public function serialize()
+            {
+            }
+        };
         $context['resources'] = &$resources;
 
         $event->setControllerResult($this->serializer->serialize($controllerResult, $request->getRequestFormat(), $context));
 
         $request->attributes->set('_api_respond', true);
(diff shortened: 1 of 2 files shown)


src/EventListener/SerializeListener.php
```diff
@@ -60,7 +60,9 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
         $context = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
         $resources = [];
-        $context['resources'] = &$resources;
+        $context['resources'] = function ($iri) use (&$resources) {
```
Simperfit: Do you have the same output has before ? if yes then it seems ok to me.

soyuka: Yes but it's fucked up that a Closure is faster then a reference in this case. There may be a simple explanation with the internal memory pointer doing less work, I would love to understand :p.

Simperfit: Did you try to compare the opcode ?

soyuka: opcode?

PR #1311, opened 2017-08-08
fix #1310 join condition aliases in FilterEagerLoadingExtension

### LINKED ISSUE #1310: Wrong alias in join with explicit condition
Using a custom filter, when I add a join with an explicit condition to the query builder, it ends up generating wrong sql.

I believe that the problem is in the ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\FilterEagerLoadingExtension class in the method getQueryBuilderWithNewAliases.

The actual code is:

```
//Change join aliases
        foreach ($joinParts[$originAlias] as $joinPart) {
            $aliases[] = "{$joinPart->getAlias()}.";
            $alias = $queryNameGenerator->generateJoinAlias($joinPart->getAlias());
            $replacements[] = "$alias.";
            $join = new Join($joinPart->getJoinType(), str_replace($aliases, $replacements, $joinPart->getJoin()), $alias, $joinPart->getConditionType(), $joinPart->getCondition(), $joinPart->getIndexBy());

            $queryBuilderClone->add('join', [$join], true);
        }
```
The method is recreating the joins and replacing the aliases with new aliases, but is not doing the replacement in the possible conditions of the join, so the join ends up having a new alias but using the old alias in the conditions, and the generated sql is wrong (it doesn't even compile)

I solved locally by replacing the line 121:

`$join = new Join($joinPart->getJoinType(), str_replace($aliases, $replacements, $joinPart->getJoin()), $alias, $joinPart->getConditionType(), $joinPart->getCondition(), $joinPart->getIndexBy());`

with:

`$join = new Join($joinPart->getJoinType(), str_replace($aliases, $replacements, $joinPart->getJoin()), $alias, $joinPart->getConditionType(), str_replace($aliases, $replacements, $joinPart->getCondition()), $joinPart->getIndexBy());`

The only change is the use of the same str_replace in the conditions

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1310
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -118,7 +118,7 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
             $aliases[] = "{$joinPart->getAlias()}.";
             $alias = $queryNameGenerator->generateJoinAlias($joinPart->getAlias());
             $replacements[] = "$alias.";
-            $join = new Join($joinPart->getJoinType(), str_replace($aliases, $replacements, $joinPart->getJoin()), $alias, $joinPart->getConditionType(), $joinPart->getCondition(), $joinPart->getIndexBy());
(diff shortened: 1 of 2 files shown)


Simperfit:
Thanks @soyuka

soyuka:
I did almost nothing, thanks @fedetemoli02 !

fedetemoli02:
😀
Thank you all for the quick merge

Issue #1246, opened 2017-07-13
Configure default order


vincentchalamon:
By default, API Platform order list by resource identifier. The only way to default order list by another field would be to create a custom extension, and if the filter is not defined in url apply a default order.

For example:
```php
$request = $this->requestStack->getCurrentRequest();
--
if (null === $request || !$request->query->has('foo')) {
    // Default order
    $queryBuilder->orderBy($queryBuilder->getRootAliases()[0].'.bar ASC');

    return;
}

// Custom order from request
$queryBuilder->orderBy($queryBuilder->getRootAliases()[0].'.bar DESC');
```

But it could be interesting to configure it directly from the `@ApiResource` annotation declaration, like:
```php
/**
 * @ApiResource(attributes={"order"={"bar"="DESC"}})
 * @ORM\Entity()
 */
class Foo
{
    /**
     * @var int
     *
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @var int
     *
     * @ORM\Column()
     */
    private $bar;
}
```

@api-platform/core-team do you think it could be interesting?

PR #1321, opened 2017-08-18
Fix 1246: configure default order on ApiResource annotation

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1246
| License       | MIT
| Doc PR        | api-platform/doc#235

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Extension;
 
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Doctrine\ORM\QueryBuilder;
 
 /**
@@ -21,13 +22,16 @@ use Doctrine\ORM\QueryBuilder;
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Samuel ROZE <samuel.roze@gmail.com>
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
  */
 final class OrderExtension implements QueryCollectionExtensionInterface
 {
     private $order;
+    private $resourceMetadataFactory;
 
-    public function __construct(string $order = null)
+    public function __construct(string $order = null, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
     {
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->order = $order;
     }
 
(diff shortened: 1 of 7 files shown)


src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
```diff
@@ -14,20 +14,24 @@
...
  * @author Samuel ROZE <samuel.roze@gmail.com>
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
  */
 final class OrderExtension implements QueryCollectionExtensionInterface
 {
     private $order;
+    private $resourceMetadataFactory;
 
-    public function __construct(string $order = null)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, string $order = null)
```
dunglas: This a BC break, can you add it as an optional last parameter?

Simperfit:
Could you please rebase instead of merging ?

dunglas:
Thank you @vincentchalamon

tests/Fixtures/TestBundle/Entity/Foo.php
```diff
@@ -0,0 +1,65 @@
...
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * Foo.
+ *
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
+ *
+ * @ApiResource(attributes={
+ *     "order"={"name", "DESC"}
```
dunglas: I probably have been too fast to merge. Wouldn't be better to allow complex ordering? I suggest the following:

```php
@ApiResource(attributes={"name": "DESC", "firstname": "ASC"})
```

(Order by name then by firstname). WDYT @vincentchalamon? Can you do a follow up PR?

vincentchalamon: I though about adding the multiple default ordering, but I wasn't sure it was very useful.
I would prefer a syntax like that:
```php
@ApiResource(attributes={
    "order"={"name": "DESC", "firstname": "ASC"}
})
```
WDYT @dunglas? If so, I could open another PR to fix it.

dunglas: IMO it's useful (ordering by name then firstname is very common for instance). You're right for the syntax, my example is faulty (missing `order` key). A PR would be great!

vincentchalamon: https://github.com/api-platform/core/pull/1323

ahmed-bhs:
Guys thank you for this feature, does this support yaml configuration, I tried to add order to my resource, seems nothing work ! Any idea? @vincentchalamon 
```
resources:
    App\Entity\Repos\EventFamily:
        collectionOperations:
            get:
                order: ["name", "DESC"]
                attributes:
                    order: ["name", "DESC"]
```

vincentchalamon:
@ahmed-bhs Did you try the following configuration?:
```yaml
resources:
    App\Entity\Repos\EventFamily:
        order: ["name", "DESC"]
        collectionOperations:
            get: ~
```

I didn't try this configuration in YAML since many years (this PR is older than 3 years).

PR #1323, opened 2017-08-22
Issues/1246

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1246
| License       | MIT
| Doc PR        | api-platform/doc#235

### DIFF
--- a/tests/Fixtures/TestBundle/Entity/Foo.php
+++ b/tests/Fixtures/TestBundle/Entity/Foo.php
@@ -23,7 +23,7 @@ use Symfony\Component\Validator\Constraints as Assert;
  * @author Vincent Chalamon <vincentchalamon@gmail.com>
  *
  * @ApiResource(attributes={
- *     "order"={"bar": "ASC", "name": "DESC"}
+ *     "order"={"bar", "name": "DESC"}
  * })
  * @ORM\Entity
  */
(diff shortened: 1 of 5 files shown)


tests/Fixtures/TestBundle/Entity/Foo.php
```diff
@@ -24,7 +23,7 @@
  * @author Vincent Chalamon <vincentchalamon@gmail.com>
  *
  * @ApiResource(attributes={
- *     "order"={"name", "DESC"}
+ *     "order"={"bar": "ASC", "name": "DESC"}
```
dunglas: Just some nitpicking 😬: it would be great to also allow `"order"={"bar", "name"}` (that would be equivalent to ` "order"={"bar": "ASC", "name": "ASC"}`. It's just some syntactic sugar but it improves the overall DX.

vincentchalamon: @dunglas Could it also be interesting to order by association property, like `dummy.name`?

dunglas: Definitely. In another PR?

dunglas:
Thanks @vincentchalamon

vincentchalamon:
@dunglas Still a [comment](https://github.com/api-platform/core/pull/1323#discussion_r134468290) pending… You merged too fast !

PR #1324, opened 2017-08-22
[RFR] Issues/1246

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1246, https://github.com/api-platform/core/pull/1323#discussion_r134468290
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/269

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -122,8 +122,9 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
         foreach ($joinParts[$originAlias] as $joinPart) {
             /** @var Join $joinPart */
             $joinString = str_replace($aliases, $replacements, $joinPart->getJoin());
-            $alias = substr($joinString, 0, strpos($joinString, '.'));
-            $association = substr($joinString, strpos($joinString, '.')+1);
+            $pos = strpos($joinString, '.');
+            $alias = substr($joinString, 0, $pos);
+            $association = substr($joinString, $pos+1);
             $condition = str_replace($aliases, $replacements, $joinPart->getCondition());
             $newAlias = QueryBuilderHelper::addJoinOnce($queryBuilderClone, $queryNameGenerator, $alias, $association, $joinPart->getJoinType(), $joinPart->getConditionType(), $condition);
             $aliases[] = "{$joinPart->getAlias()}.";
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -124,7 +124,7 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
             $joinString = str_replace($aliases, $replacements, $joinPart->getJoin());
             $pos = strpos($joinString, '.');
             $alias = substr($joinString, 0, $pos);
-            $association = substr($joinString, $pos+1);
+            $association = substr($joinString, $pos + 1);
             $condition = str_replace($aliases, $replacements, $joinPart->getCondition());
             $newAlias = QueryBuilderHelper::addJoinOnce($queryBuilderClone, $queryNameGenerator, $alias, $association, $joinPart->getJoinType(), $joinPart->getConditionType(), $condition);
             $aliases[] = "{$joinPart->getAlias()}.";
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Extension;
 
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\EagerLoadingTrait;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryBuilderHelper;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Doctrine\ORM\Query\Expr\Join;
@@ -119,12 +120,14 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
 
         //Change join aliases
         foreach ($joinParts[$originAlias] as $joinPart) {
+            /** @var Join $joinPart */
+            $joinString = str_replace($aliases, $replacements, $joinPart->getJoin());
+            $alias = substr($joinString, 0, strpos($joinString, '.'));
+            $association = substr($joinString, strpos($joinString, '.')+1);
+            $condition = str_replace($aliases, $replacements, $joinPart->getCondition());
+            $newAlias = QueryBuilderHelper::addJoinOnce($queryBuilderClone, $queryNameGenerator, $alias, $association, $joinPart->getJoinType(), $joinPart->getConditionType(), $condition);
             $aliases[] = "{$joinPart->getAlias()}.";
-            $alias = $queryNameGenerator->generateJoinAlias($joinPart->getAlias());
-            $replacements[] = "$alias.";
-            $join = new Join($joinPart->getJoinType(), str_replace($aliases, $replacements, $joinPart->getJoin()), $alias, $joinPart->getConditionType(), str_replace($aliases, $replacements, $joinPart->getCondition()), $joinPart->getIndexBy());
-
-            $queryBuilderClone->add('join', [$join], true);
+            $replacements[] = "$newAlias.";
         }
 
         $queryBuilderClone->add('where', str_replace($aliases, $replacements, (string) $wherePart));
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Extension;
 
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryBuilderHelper;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Doctrine\ORM\QueryBuilder;
@@ -47,10 +48,18 @@ final class OrderExtension implements QueryCollectionExtensionInterface
             if (null !== $defaultOrder) {
                 foreach ($defaultOrder as $field => $order) {
                     if (is_int($field)) {
+                        // Default direction
                         $field = $order;
                         $order = 'ASC';
                     }
-                    $queryBuilder->addOrderBy('o.'.$field, $order);
+                    if (false === ($pos = strpos($field, '.'))) {
+                        // Configure default filter with property
+                        $field = 'o.'.$field;
+                    } else {
+                        $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, 'o', substr($field, 0, $pos));
+                        $field = sprintf('%s.%s', $alias, substr($field, $pos+1));
+                    }
+                    $queryBuilder->addOrderBy($field, $order);
                 }
 
                 return;
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -57,7 +57,7 @@ final class OrderExtension implements QueryCollectionExtensionInterface
                         $field = 'o.'.$field;
                     } else {
                         $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, 'o', substr($field, 0, $pos));
-                        $field = sprintf('%s.%s', $alias, substr($field, $pos+1));
+                        $field = sprintf('%s.%s', $alias, substr($field, $pos + 1));
                     }
                     $queryBuilder->addOrderBy($field, $order);
                 }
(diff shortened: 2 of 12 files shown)


src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
```diff
@@ -47,10 +47,22 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
...
-                    $queryBuilder->addOrderBy('o.'.$field, $order);
+                    if (false === ($pos = strpos($field, '.'))) {
+                        // Configure default filter with property
+                        $field = 'o.'.$field;
+                    } else {
+                        // Configure default filter with association
+                        // Ensure query join is done on association
+                        $associationName = substr($field, 0, $pos);
+                        if (!in_array($associationName, $queryBuilder->getRootAliases(), true)) {
+                            $queryBuilder->join('o.'.$associationName, $associationName);
```
meyerbaptiste: No conflicts with [src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L117-L124](https://github.com/api-platform/core/blob/babe7bbb3a4ca5c40a111cbe41ce81f309677b5d/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L117-L124) here (extensions priorities)?

vincentchalamon: @dunglas WDYT?

soyuka: Can you use the JoinOnce thing? This may lead to multiple joins indeed.

vincentchalamon: @soyuka What do you mean? Do you have any example please?

meyerbaptiste: Maybe this: https://github.com/api-platform/core/blob/0e1af77ee629b4f3d4841e172a67bfb6eda0db8b/src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php#L382-L401

soyuka: Thanks @meyerbaptiste ! This way we always use the same method to join only once :).

Simperfit: @vincentchalamon does it feels right for you ?

vincentchalamon: I have to check how to prevent code duplication. I may have to create another service common with filters & extensions

soyuka: Oh you're in an extension... maybe this could be a static method not sure.

vincentchalamon: What do you think about a QueryBuilderHelper, which can be shared with filters & extensions?

soyuka: Would be nice yeah, maybe we can do this in another PR!

dunglas: The helper (as long as it is marked `@internal`) looks good to me. The QueryBuilder is provided by Doctrine, it's not really wanted to modify it.

soyuka: I'd keep static helpers instead of extending the QueryBuilder.

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param QueryNameGeneratorInterface $queryNameGenerator
+     * @param string                      $alias
+     * @param string                      $association        the association field
+     * @param string|null                 $joinType           the join type (left join / inner join)
+     * @param string|null                 $conditionType
+     * @param string|null                 $condition
+     *
+     * @return string the new association alias
+     */
+    public static function addJoinOnce(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $association, ?string $joinType = null, ?string $conditionType = null, ?string $condition = null): string
```
dunglas: You can't use the `?string` syntax because we ensure compatibility with PHP 7.0 (this syntax was introduced in 7.1).

dunglas: You can't use the `?string` syntax because we ensure compatibility with PHP 7.0 (this syntax was introduced in 7.1).

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
```diff
@@ -115,12 +116,14 @@ private function getQueryBuilderWithNewAliases(QueryBuilder $queryBuilder, Query
 
         //Change join aliases
         foreach ($joinParts[$originAlias] as $joinPart) {
+            /** @var Join $joinPart */
+            $joinString = str_replace($aliases, $replacements, $joinPart->getJoin());
+            $alias = substr($joinString, 0, strpos($joinString, '.'));
+            $association = substr($joinString, strpos($joinString, '.')+1);
```
soyuka: can we avoid multiple calls on `strpos`?

soyuka: can we avoid multiple calls on strpos?

Do you think it's possible that strpos returns `false` here? Maybe we kinda need a test?

tests/Fixtures/TestBundle/Entity/Foo.php
```diff
@@ -42,15 +41,13 @@ class Foo
      * @var string The foo name
      *
      * @ORM\Column
-     * @Assert\NotBlank
```
soyuka: Really not needed? I'd have think those are for a given behat test, if not np just remove them though it's not part of the fix (or is it?).

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
vincentchalamon: @soyuka As it's from a cloned query builder, I don't think it's even possible that `strpos` could return `false`. It would mean that the original query builder is invalid, which report the bug to doctrine.

tests/Fixtures/TestBundle/Entity/Foo.php
vincentchalamon: Originally, I created this `Foo` entity for my tests. It was copied/pasted from another one. But I think the assertion is useless on this entity as I only create the objects by myself without validation.
This modification (in this PR) is a clean :-)

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
```diff
@@ -0,0 +1,85 @@
...
+
+use Doctrine\ORM\Query\Expr\Join;
+use Doctrine\ORM\QueryBuilder;
+
+/**
+ * @author Vincent Chalamon <vincent@les-tilleuls.coop>
+ *
+ * @internal
+ */
+final class QueryBuilderHelper
```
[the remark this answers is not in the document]
vincentchalamon: Why for?

vincentchalamon:
> It would be nice to compare SQL queries before and after these changes to verify that they are identical. These ones seem to me very sensitive...

@meyerbaptiste on which class: FilterEagerLoadingExtension or QueryBuilderHelper?

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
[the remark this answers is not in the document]
soyuka: Because he likes private constructors :P.

Nah just that it's a convention when classes have no instances (ie are never called with `new` because they only have statics).

soyuka:
@meyerbaptiste He didn't changed the tests and we do test queries there.

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param string       $alias
+     * @param string       $association  the association field
+     *
+     * @return Join|null
+     */
+    private static function getExistingJoin(QueryBuilder $queryBuilder, string $alias, string $association)
+    {
+        $parts = $queryBuilder->getDQLPart('join');
+
+        if (!isset($parts['o'])) {
```
soyuka: Maybe we could put this `o` alias in some constant somewhere (in another PR).

vincentchalamon: It could be easily retried by `$queryBuilder->getRootAliases()[0]`. Should I replace it here?

soyuka: don't IMO it's fine like this. It's just that we have a lot of these `'o'` alias hard-coded in the bridge.

vincentchalamon:
@dunglas @meyerbaptiste @Simperfit PR updated. Should I update/rebase it from 2.2?

soyuka:
Would you mind fixing phpstan?

vincentchalamon:
@soyuka I don't understand where the error comes from in my code:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Error                                                                                                                                                                                                                    
 ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Ignored error pattern #Parameter \#2 \$dqlPart of method Doctrine\\ORM\\QueryBuilder::add\(\) expects Doctrine\\ORM\\Query\\Expr\\Base, Doctrine\\ORM\\Query\\Expr\\Join\[\] given# was not matched in reported errors.  
 ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
                                                                                
 [ERROR] Found 1 error                                                          
                                                                                
The command "if [[ $lint = 1 ]]; then phpstan analyse -c phpstan.neon -l5 --ansi src tests; fi" exited with 1.
```

soyuka:
Just remove this line: 
https://github.com/api-platform/core/blob/master/phpstan.neon#L14

tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php
```diff
@@ -264,8 +264,8 @@ public function testCompositeIdentifiers()
             ->setParameter('foo', 1);
 
         $queryNameGenerator = $this->prophesize(QueryNameGeneratorInterface::class);
-        $queryNameGenerator->generateJoinAlias('item')->shouldBeCalled()->willReturn('item_2');
-        $queryNameGenerator->generateJoinAlias('label')->shouldBeCalled()->willReturn('label_2');
+        $queryNameGenerator->generateJoinAlias('compositeItem')->shouldBeCalled()->willReturn('item_2');
+        $queryNameGenerator->generateJoinAlias('compositeLabel')->shouldBeCalled()->willReturn('label_2');
```
soyuka: This change is wrong, `item` and `label` are correct I think (at least according to failing tests).

vincentchalamon: Nop, cause this call is made through [QueryBuilderHelper::addJoinOnce](https://github.com/api-platform/core/pull/1324/files#diff-44f9dbe5365de635dbe64677eb371d9cR48) from association name.

I don't understand why travis is failing…on a non-existing unit test 😮 (failing method `testCompositeIdentifiersWithoutAssociation` does not exist in `FilterEagerLoadingExtensionTest`)

vincentchalamon: I'll try to squash my changes to force travis to rebuild its environment

soyuka: I added this recently maybe a bad rebase? 
https://github.com/api-platform/core/blob/d18ec0da1ac52f92ca755ca275cd0853d0972878/tests/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtensionTest.php

vincentchalamon: Rebase missing, I fixed it, thanks @soyuka

soyuka:
:+1:  waiting for some more review but this LGTM!

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
```diff
@@ -0,0 +1,89 @@
...
+final class QueryBuilderHelper
+{
+    private function __construct()
+    {
+    }
+
+    /**
+     * Adds a join to the queryBuilder if none exists.
+     *
+     * @param QueryBuilder                $queryBuilder
```
dunglas: You can remove most of these `@param`.

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param string                      $association        the association field
+     * @param string|null                 $joinType           the join type (left join / inner join)
+     * @param string|null                 $conditionType
+     * @param string|null                 $condition
+     *
+     * @return string the new association alias
+     */
+    public static function addJoinOnce(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $association, $joinType = null, $conditionType = null, $condition = null): string
+    {
+        $join = self::getExistingJoin($queryBuilder, $alias, $association);
```
dunglas: What about:

```php
if (null !== $join) {
    return $join->getAlias();
}

$associationAlias = $queryNameGenerator->generateJoinAlias($association);
$query = "$alias.$association";

if (Join::LEFT_JOIN === $joinType || QueryChecker::hasLeftJoin($queryBuilder)) {
    $queryBuilder->leftJoin($query, $associationAlias, $conditionType, $condition);
} else {
    $queryBuilder->innerJoin($query, $associationAlias, $conditionType, $condition);
}

return $associationAlias;
```

To reduce the complexity of this method.

src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php
```diff
@@ -0,0 +1,89 @@
...
+            $associationAlias = $join->getAlias();
+        }
+
+        return $associationAlias;
+    }
+
+    /**
+     * Get the existing join from queryBuilder DQL parts.
+     *
+     * @param QueryBuilder $queryBuilder
```
dunglas: All `@param` can be removed

dunglas:
Thanks @vincentchalamon!

PR #1333, opened 2017-08-25
[2.1] Allow to specify a whitelist of serialization groups for Group Filter

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This PR aims to give the possibility to restrict the Group Filter application.
The use case would be for example when exposing books with author, to protect calling `/books?goups[]=author_sensitive_data`.

If this PR is welcome, I'll do it for the Property Filter and provide a doc PR

### DIFF
--- a/src/Serializer/Filter/GroupFilter.php
+++ b/src/Serializer/Filter/GroupFilter.php
@@ -24,11 +24,13 @@ final class GroupFilter implements FilterInterface
 {
     private $overrideDefaultGroups;
     private $parameterName;
+    private $whitelist;
 
-    public function __construct(string $parameterName = 'groups', bool $overrideDefaultGroups = false)
+    public function __construct(string $parameterName = 'groups', bool $overrideDefaultGroups = false, array $whitelist = null)
     {
         $this->overrideDefaultGroups = $overrideDefaultGroups;
         $this->parameterName = $parameterName;
+        $this->whitelist = $whitelist;
     }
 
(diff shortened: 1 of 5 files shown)


src/Serializer/Filter/GroupFilter.php
```diff
@@ -24,11 +24,13 @@
 {
     private $overrideDefaultGroups;
     private $parameterName;
+    private $whitelist;
 
-    public function __construct(string $parameterName = 'groups', bool $overrideDefaultGroups = false)
+    public function __construct(string $parameterName = 'groups', bool $overrideDefaultGroups = false, array $whitelist = [])
```
dunglas: I suggest to use `array $whitelist = null` and to check if the value if not `null` instead. It will be more intuitive, we have a special value for a special case.

meyerbaptiste:
👍 for this feature

src/Serializer/Filter/GroupFilter.php
antograssiot: Sure I'll update tonight

antograssiot:
Updated as requested @dunglas

dunglas:
Thank you very much @antograssiot. Great contribs!

Issue #1338, opened 2017-08-28
PUT & nested properties with validation : always null


pierre-H:
Hello,

Thanks for Api Platform : a great tool !!!

I am trying to update an object with sub-objects : 

```php

/**
 * @ApiResource(
 *     collectionOperations={"get": {"method": "GET"} },
 *     itemOperations={
 *         "get": {
 *             "method": "GET"
 *         },
 *         "put": {
 *             "method": "PUT",
 *             "denormalization_context"={"groups"={"book_edit"}}
 *         }
 *     }
 * )
 */
class Book {
    /**
     * @var Author
     * @Groups({"book_edit"})
     * @Assert\NotNull
     */
    private $author;
}

/**
 * @ApiResource
 */
class Author {
    /**
     * @var string
     * @Groups({"book_edit"})
     */
    private $name;

    /**
     * @var sting
     *
     * @Assert\NotBlank
     */
    private $aProperty;
}
```

And like the doc [says](https://api-platform.com/docs/core/serialization-groups-and-relations#denormalization), I added in the json the **@id** key in order to point to the author but all author properties are null so it doesn't retrieve the real object and it causes validation error.

I discovered few things : 
- if I remove the validation, it tries do to an INSERT on the database instead of UPDATE (for author)
- with Xdebug, I find that it uses `src\Serializer\ItemNormalizer::denormalize` and this function uses the **id** property, not **@id**. So I tried with **id** and I have the same result.

Could you please help me ?

Thank you

soyuka:
Are you using doctrine? Could we have the mapping?

Do you have getters and setters for those relations?

pierre-H:
@soyuka  yes I use Doctrine.

This is the original mapping : 

```php
/**
 * @ORM\Entity()
 * @ApiResource(
 *     collectionOperations={"get": {"method": "GET", "normalization_context"={"groups"={"sejour_list"} } } },
 *     itemOperations={
 *         "get": {
 *             "method": "GET",
 *             "normalization_context"={"groups"= {"fiche_patient_general", "fiche_patient", "rens-ss", "dossier-medical" } }
 *         },
 *         "put": {
 *              "method": "PUT",
 *              "denormalization_context"={"groups"={"sejour_edit"}},
 *         },
 *         "info_sejour": {
 *             "method": "GET",
 *             "normalization_context"={"groups"={"sejour_info"}}
 *         } 
 *    }
 * )
 */
class Sejour {
    /**
     * @var Patient
     *
[...]

PR #1340, opened 2017-08-29
fetch_data -> true when using json

pierre-H:
fix https://github.com/api-platform/core/issues/1338#issuecomment-325624277

The JsonLd Normalizer has `fetch_data = true` but not the default one. See : 
https://github.com/api-platform/core/blob/12871f38fdadf79e696526ac492be4ccfd82001c/src/JsonLd/Serializer/ItemNormalizer.php#L105

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #1338
| License       | MIT

### DIFF
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -44,7 +44,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
     private function updateObjectToPopulate(array $data, array &$context)
     {
         try {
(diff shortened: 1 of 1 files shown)


soyuka:
`tests/Serializer/ItemNormalizerTest.php` needs an update too

dunglas:
👍 when tests will be green. This PR should target the`2.0` branch isn't it?

pierre-H:
Thank you @soyuka for the better pull https://github.com/api-platform/core/pull/1342

PR #1343, opened 2017-08-31
Fix #1326 (DateFilter before/after filters on before or equals/after or equals)

### LINKED ISSUE #1326: DateFilter before/after filters on before or equals/after or equals
Hi, 

The `DateFilter` allows to filter with `before` or `after` a date
> ?property[before]=date
> ?property[after]=date

But in fact it filters on `before or equals` and `after or equals`:
> DateFilter::addWhere, line 13
> 
> $baseWhere = sprintf('%s.%s %s :%s', $alias, $field, self::PARAMETER_BEFORE === $operator ? '<=' : '>=', $valueParameter);

This is causing me issues with this scenario : The client wants to retrieve only the new items, so he filters on `?creationDate[after]=<creation date of the last item retrieved>` but the API will also return the previous last item.

Would il be possible to :
- Change the behaviour to filter on `>` instead of `>=` and `<` instead of `<=`
or 
- Add new parameters like  `before but not equals` or `after but not equals` (I'm not really inspired for the naming...)

I'll try to make a pull request when I will have some time for the second solution.

Regards

blesenechal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1326
| License       | MIT
| Doc PR        | api-platform/doc#244

The DateFilter `?property[after]=date` should filter on `>`, not on `>=` to have only the items with date after the value

### DIFF
--- a/features/doctrine/date_filter.feature
+++ b/features/doctrine/date_filter.feature
@@ -406,7 +406,7 @@ Feature: Date filter on collections
       },
       "hydra:search": {
         "@type": "hydra:IriTemplate",
-        "hydra:template": "/dummies{?dummyBoolean,relatedDummy.embeddedDummy.dummyBoolean,dummyDate[before],dummyDate[after],relatedDummy.dummyDate[before],relatedDummy.dummyDate[after],description[exists],relatedDummy.name[exists],dummyBoolean[exists],dummyFloat,dummyPrice,order[id],order[name],order[relatedDummy.name],order[relatedDummy.symfony],dummyFloat[between],dummyFloat[gt],dummyFloat[gte],dummyFloat[lt],dummyFloat[lte],dummyPrice[between],dummyPrice[gt],dummyPrice[gte],dummyPrice[lt],dummyPrice[lte],id,id[],name,alias,description,relatedDummy.name,relatedDummy.name[],relatedDummies,relatedDummies[],dummy,relatedDummies.name}",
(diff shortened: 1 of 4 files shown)


norkunas:
This would be not expected if i wanted to filter items between 2017-01-01 00:00:00 and 2017-01-01 23:59:59 :disappointed:

blesenechal:
In this case we should add an other filter like `?property[after_or_equals]=date` ?

norkunas:
Good point, but I think the after/before should be left as is and (after|before)_neq should be added as people already uses this filter

blesenechal:
Yes tou're right, doing this would not break anything for other users .
I'll update the PR.

blesenechal:
@norkunas I've made the changes

blesenechal:
If you release 2.1 as stable, please include it in 2.1 :)

soyuka:
Don't worry @blesenechal every patch on `2.0` is merged on `master` (and therefore will be available in the `2.1` release!). It's just that usually features are merged on `master` instead of `2.0`.

dunglas:
I also prefer to merge it in 2.1 (not in 2.0).

dunglas:
Just a question about the naming. We don't use acronyms anywhere in filter. `after_not_equals` would be better no?

blesenechal:
@dunglas OK I can rename to `before_not_equals` / `after_not_equals` and update the PR to merge in master instead of 2.0 if you want ?

antograssiot:
`strictly_after` may also work if I may suggest something

soyuka:
> OK I can rename to before_not_equals / after_not_equals and update the PR to merge in master instead of 2.0 if you want ?

Please do! I'm also :+1: for `strictly_before`/`strictly_after` as it's shorter and means the same :).

soyuka:
I think that if you rebase the branch against master it'll fix the tests ;).

Simperfit:
Could you please update to master again ? and we can merge it ;)

soyuka:
Thanks @blesenechal !

Simperfit:
Thanks for the contribution @blesenechal ;)

PR #1346, opened 2017-09-01
Correctly retrieve resource class of an object if it exists even when…

antograssiot:
… it is not the given resource class

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        | 

Correctly retrieve resource class of an object if it exists even when it is not the given resource class

### DIFF
--- a/src/Api/ResourceClassResolver.php
+++ b/src/Api/ResourceClassResolver.php
@@ -54,6 +54,9 @@ final class ResourceClassResolver implements ResourceClassResolverInterface
             if (is_subclass_of($type, $resourceClass) && $this->isResourceClass($resourceClass)) {
                 return $type;
             }
+            if ($this->isResourceClass($typeToFind)) {
+                return $typeToFind;
+            }
 
             throw new InvalidArgumentException(sprintf('No resource class found for object of type "%s".', $typeToFind));
         }
--- a/src/Api/ResourceClassResolver.php
+++ b/src/Api/ResourceClassResolver.php
@@ -50,11 +50,11 @@ final class ResourceClassResolver implements ResourceClassResolverInterface
             $typeToFind = $type = $resourceClass;
         }
 
-        if (($strict && isset($type) && $resourceClass !== $type) || !$this->isResourceClass($typeToFind)) {
+        if (($strict && isset($type) && $resourceClass !== $type) || false === $isResourceClass = $this->isResourceClass($typeToFind)) {
             if (is_subclass_of($type, $resourceClass) && $this->isResourceClass($resourceClass)) {
                 return $type;
             }
-            if ($this->isResourceClass($typeToFind)) {
+            if ($isResourceClass ?? $this->isResourceClass($typeToFind)) {
                 return $typeToFind;
             }
 
--- a/tests/Api/ResourceClassResolverTest.php
+++ b/tests/Api/ResourceClassResolverTest.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\DataProvider\PaginatorInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceNameCollection;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCar;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyTableInheritance;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyTableInheritanceChild;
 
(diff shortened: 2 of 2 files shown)


antograssiot:
appveyor failure doesn't seems related

src/Api/ResourceClassResolver.php
```diff
@@ -54,6 +54,9 @@ public function getResourceClass($value, string $resourceClass = null, bool $str
             if (is_subclass_of($type, $resourceClass) && $this->isResourceClass($resourceClass)) {
                 return $type;
             }
+            if ($this->isResourceClass($typeToFind)) {
+                return $typeToFind;
+            }
```
soyuka: May you cache the previous `$this->isResourceClass($typeToFind)` call in a variable instead as this is heavy? Indeed this method can have a big impact on performances.

I'm always kinda scared to modify this file anyway as it can lead to weird behavior. I'm not sure to understand the use case so far, may you give some more explanation? Thanks!

tests/Api/ResourceClassResolverTest.php
```diff
@@ -38,6 +39,18 @@ public function testGetResourceClassWithIntendedClassName()
...
+    {
+        $dummy = new Dummy();
+        $dummy->setName('Smail');
+        $resourceNameCollectionFactoryProphecy = $this->prophesize(ResourceNameCollectionFactoryInterface::class);
+        $resourceNameCollectionFactoryProphecy->create()->willReturn(new ResourceNameCollection([Dummy::class]))->shouldBeCalled();
+
+        $resourceClassResolver = new ResourceClassResolver($resourceNameCollectionFactoryProphecy->reveal());
+        $resourceClass = $resourceClassResolver->getResourceClass($dummy, DummyCar::class, true);
+        $this->assertEquals($resourceClass, Dummy::class);
+    }
```
soyuka: For example I don't see how `$resourceClassResolver->getResourceClass($dummy, DummyCar::class, true)` would be called as `DummyCar` isn't a part of your `ResourceNameCollection` and isn't the `Dummy` object. I may miss something.

src/Api/ResourceClassResolver.php
antograssiot: I don't see easy way of caching the value without being obliged to call the method each time.
I mean that I can totally enter the and quit the `if` statement without calling this method...
Do you have something to suggest @soyuka ?

tests/Api/ResourceClassResolverTest.php
antograssiot: The actual behavior that I have is that I have a resource `Book` linked to a  `File` entity which is not a resource but the file entity is linked to an `Author` which is a resource. When I call the route I got an exceptio nfrom the ResourceClassresolver saying that Author is not a resource.
The JsonLd ItemNormalizer call get resourceClass [here](https://github.com/api-platform/core/blob/master/src/JsonLd/Serializer/ItemNormalizer.php#L65), the first param is my Author object and the second one would be the Book.

I was hesitant to send the PR as it does fix my issue, doesn't break any existing tests but I understand that it might come from somewhere else....

src/Api/ResourceClassResolver.php
soyuka: Hmm, indeed this is why I think this patch is kinda weird (we have `!$this->isResourceClass($typeToFind)` in the condition which means we're in the first part of the condition and that the second needs to be computed afterwards. Maybe this needs a refactor but it is very sensitive ^^.

tests/Api/ResourceClassResolverTest.php
soyuka: So (`R` represents a resource):

```
Book (R) => File
File => Author (R)
```

You request `Book`, then it normalizes the `File` followed by the `Author`? So call would be `getResourceClass(Author::class (the class being denormalized), Book::class (the context resource class is correct), true);`.

Am I correct?

I'm not sure it comes from elsewhere so fixing this here would be correct. I just want to be sure I get this correctly to give this a try.

antograssiot: True that what is called

src/Api/ResourceClassResolver.php
antograssiot: @soyuka I think I found a way to avoid extra call to `isResourceClass($typeToFind)` at least when it is false. WDYT about it ?

src/Api/ResourceClassResolver.php
```diff
@@ -50,10 +50,13 @@ public function getResourceClass($value, string $resourceClass = null, bool $str
...
         }
 
-        if (($strict && isset($type) && $resourceClass !== $type) || !$this->isResourceClass($typeToFind)) {
+        if (($strict && isset($type) && $resourceClass !== $type) || false === $isResourceClass = $this->isResourceClass($typeToFind)) {
             if (is_subclass_of($type, $resourceClass) && $this->isResourceClass($resourceClass)) {
                 return $type;
             }
+            if ($isResourceClass ?? $this->isResourceClass($typeToFind)) {
+                return $typeToFind;
+            }
```
soyuka: Indeed this saves a call when falsy :p.

dunglas:
Thanks @antograssiot

PR #1240, opened 2017-07-08
feature: Implement SwaggerUI api key authentication functionality

bartwesselink:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | api-platform/docs#232

This PR integrates api key functionality for Swagger UI into api-platform.

PR #1351, opened 2017-09-06
Fix Configuration Bug for Api Keys

Engerim:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1240 
| License       | MIT

This fix a configuration bug which was added with #1240.

The `Configuration` was implement in a way that the config for api keys looks like this.
```yaml
swagger:
    api_keys:
      name: "key"
      type: "query"
```

But the implementation expects an array for api keys so the config should look like 
```yaml
swagger:
    api_keys:
      -
        name: "key"
        type: "query"
```


Simperfit:
Thanks @Engerim

Issue #1309, opened 2017-08-07
[UUID] Bug with Ramsey\UUID uuid_binary Doctrine type


dkarlovi:
If using Ramsey Doctrine UUID `uuid_binary` datatype, identifiers in motion are classic UUID strings (ie. `5e619188-b5c8-4797-905e-05f4ba6dfc67`) but they're binary at rest (ie. `0xFEFDF85691AE48E2888D39C3EBEC352F`).

[In IdentifierManagerTrait, a statement tries to convert the user-supplied identifier to a PHP value](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php#L80), which means it tries to convert the input as if it's binary, NOT string. This fails, obviously. If I change it to `convertToDatabaseValue()`, it works as expected.

This brings up another question, if the ID provided is not decodable (not a valid UUID, for example), how should that be treated? I guess as a 404, not as a 500.

/cc @teohhanhui 

TLDR
```
GET http://127.0.0.1:8000/api/images/5e619188-b5c8-4797-905e-05f4ba6dfc67

{
    "type": "https://tools.ietf.org/html/rfc2616#section-10",
    "title": "An error occurred",
    "detail": "Could not convert database value \"5e619188-b5c8-4797-9...\" to Doctrine Type uuid_binary"
}
```

theofidry:
>if the ID provided is not decodable (not a valid UUID, for example), how should that be treated? I guess as a 404, not as a 500

For a URL, if the ID is not a valid UUID or the entity of that ID doesn't exists it should be a 404.

meyerbaptiste:
I'm not familiar with this part of the core but this bug makes sense to me.

`Type::convertToPHPValue()` method allows Doctrine to normalize the data in PHP from the output of the database, therefore it expects in our case to receive many bytes. 
So, IMO, we don't use properly this method because we should not use it to normalize data from a client.

@api-platform/core-team why don't we use the serializer to normalize IDs here?

dkarlovi:
@meyerbaptiste a proper way to fix this might be to use the serializer, yes. Especially since [the instance supports it](https://github.com/ramsey/uuid/blob/master/src/UuidInterface.php#L24).

soyuka:
[...]

PR #1358, opened 2017-09-11
GraphQL Query support

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/230
| License       | MIT
| Doc PR        |

Support of GraphQL Queries by using https://github.com/webonyx/graphql-php.
Mutations will be done later.
Based on the work of @raoulclais.
Endpoint is `/graphql`.
GraphiQL is used for the UI.


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -142,6 +143,7 @@ private function handleConfig(ContainerBuilder $container, array $config, array
         $container->setParameter('api_platform.eager_loading.max_joins', $config['eager_loading']['max_joins']);
         $container->setParameter('api_platform.eager_loading.fetch_partial', $config['eager_loading']['fetch_partial']);
         $container->setParameter('api_platform.eager_loading.force_eager', $config['eager_loading']['force_eager']);
+        $container->setParameter('api_platform.enable_graphql', $config['enable_graphql']);
```
soyuka: I'd prefer:

```
graphdql:
  enabled: true
```

So that we try not to face the same issue as with swagger when adding more features.

composer.json
```diff
@@ -23,7 +23,8 @@
         "symfony/property-access": "^2.7 || ^3.0 || ^4.0",
         "symfony/property-info": "^3.1 || ^4.0",
         "symfony/serializer": "^3.1 || ^4.0",
-        "willdurand/negotiation": "^2.0.3"
+        "willdurand/negotiation": "^2.0.3",
+        "webonyx/graphql-php": "^0.10.2"
```
dunglas: This dep should be moved in `require-dev` and added to the `suggest` section.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -253,7 +253,7 @@ private function getSerializerContext(string $resourceClass, string $contextType
             $request = $this->requestStack->getCurrentRequest();
         }
 
-        if (null !== $this->serializerContextBuilder && null !== $request) {
+        if (null !== $this->serializerContextBuilder && null !== $request && 'api_graphql_entrypoint' !== $request->attributes->get('_route')) {
```
dunglas: I suggest to check for the `application/graphql` MIME type instead. It will be more reliable.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -44,8 +44,9 @@
     private $enabledParameterName;
     private $itemsPerPageParameterName;
     private $maximumItemPerPage;
+    private $enableGraphql;
 
-    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, bool $enabled = true, bool $clientEnabled = false, bool $clientItemsPerPage = false, int $itemsPerPage = 30, string $pageParameterName = 'page', string $enabledParameterName = 'pagination', string $itemsPerPageParameterName = 'itemsPerPage', int $maximumItemPerPage = null)
+    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, bool $enabled = true, bool $clientEnabled = false, bool $clientItemsPerPage = false, int $itemsPerPage = 30, string $pageParameterName = 'page', string $enabledParameterName = 'pagination', string $itemsPerPageParameterName = 'itemsPerPage', int $maximumItemPerPage = null, bool $enableGraphql)
```
dunglas: You must set a default value (`false`?) to `$enableGraphql` or it's a BC break.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -76,6 +78,10 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
         }
 
         $itemsPerPage = $resourceMetadata->getCollectionOperationAttribute($operationName, 'pagination_items_per_page', $this->itemsPerPage, true);
+        if ($this->enableGraphql) {
+            // @TODO Retrieve GraphQL 'first' argument using the request?
```
dunglas: Maybe can you create another `PaginationExtension` class specific to GraphQL if the handling is really different than this one.

src/Bridge/Graphql/Executor.php
```diff
@@ -0,0 +1,37 @@
...
+
+use GraphQL\Executor\ExecutionResult;
+use GraphQL\GraphQL;
+
+/**
+ * Wrapper for the GraphQL facade.
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+class Executor
```
dunglas: `final`

src/Bridge/Graphql/Executor.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
dunglas: Why do we need this class? Can't we call `GraphQL::executeQuery` directly?

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+    private $propertyMetadataFactory;
+    private $resourceNameCollectionFactory;
+    private $resourceMetadataFactory;
+    private $collectionDataProvider;
+    private $itemDataProvider;
+    private $normalizer;
+    private $resourceTypesCache = [];
+
+    /**
+     * @param PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory
```
dunglas: You can remove this whole docblock, types are already documented by typehints.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+/**
+ * Builder of the GraphQL schema.
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+class SchemaBuilder
```
dunglas: `final` (and introduce an interface to allow service decoration)

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+        $this->propertyMetadataFactory = $propertyMetadataFactory;
+        $this->resourceNameCollectionFactory = $resourceNameCollectionFactory;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->collectionDataProvider = $collectionDataProvider;
+        $this->itemDataProvider = $itemDataProvider;
+        $this->normalizer = $normalizer;
+    }
+
+    /**
+     * Gets the schema.
```
dunglas: Useless docblock, can be removed.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+                'name' => 'Query',
+                'fields' => $queryFields,
+            ]),
+        ]);
+    }
+
+    /**
+     * Gets the query fields of the schema.
+     *
+     * @param string           $resource
```
dunglas: All the annotations can be remove (already documented by typehints), same everywhere.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+        $queryFields = [];
+        $shortName = $resourceMetadata->getShortName();
+        $resourceType = $this->getResourceType($resource, $resourceMetadata);
+
+        foreach ($this->getOperations($resourceMetadata, true, true) as $operationName => $queryItemOperation) {
+            $fieldNamePrefix = 'get' === $operationName ? '' : $operationName;
+            $queryFields[lcfirst($fieldNamePrefix.$shortName)] = [
+                'type' => $resourceType,
+                'args' => ['id' => $this->getResourceIdentifierFieldArguments($resource)],
+                // @TODO Create a schema resolver class?
```
dunglas: 👍, with an interface. It will be a good extension point for the end user.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+     * @param string           $resource
+     * @param ResourceMetadata $resourceMetadata
+     *
+     * @return ObjectType
+     */
+    private function getResourceType(string $resource, ResourceMetadata $resourceMetadata): ObjectType
+    {
+        $shortName = $resourceMetadata->getShortName();
+
+        if (!isset($this->resourceTypesCache[$shortName])) {
```
dunglas: Can you use a guard clause (invert the test)?

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+    private function getResourceIdentifierFieldArguments(string $resource): array
+    {
+        foreach ($this->propertyNameCollectionFactory->create($resource) as $property) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resource, $property);
+
+            if (!$propertyMetadata->isIdentifier()) {
+                continue;
+            }
+
+            // @TODO Manage multiple identifiers?
```
dunglas: Maybe can you use just reuse this method: https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php#L40

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+                    $className = $type->isCollection() ? $type->getCollectionValueType()->getClassName() : $type->getClassName();
+                    $resourceMetadata = $this->resourceMetadataFactory->create($className);
+                } catch (ResourceClassNotFoundException $e) {
+                    throw new InvalidTypeException();
+                }
+
+                $graphqlType = $this->getResourceType($className, $resourceMetadata);
+                break;
+
+            case Type::BUILTIN_TYPE_ARRAY:
```
dunglas: You can remove all those useless cases.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -68,6 +68,7 @@ public function getConfigTreeBuilder()
                 ->booleanNode('enable_nelmio_api_doc')->defaultValue(false)->info('Enable the Nelmio Api doc integration.')->end()
                 ->booleanNode('enable_swagger')->defaultValue(true)->info('Enable the Swagger documentation and export.')->end()
                 ->booleanNode('enable_swagger_ui')->defaultValue(class_exists(TwigBundle::class))->info('Enable Swagger ui.')->end()
+                ->booleanNode('enable_graphql')->defaultValue(false)->info('Enable GraphQL.')->end()
```
dunglas: I suggest `defaultValue(class_exists(GraphQL::class))` (as for Swagger UI the line before).

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -48,17 +48,19 @@
     private $operationPathResolver;
     private $container;
     private $formats;
+    private $enableGraphql;
     private $resourceClassDirectories;
     private $subresourceOperationFactory;
 
-    public function __construct(KernelInterface $kernel, ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, OperationPathResolverInterface $operationPathResolver, ContainerInterface $container, array $formats, array $resourceClassDirectories = [], SubresourceOperationFactoryInterface $subresourceOperationFactory = null)
+    public function __construct(KernelInterface $kernel, ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, OperationPathResolverInterface $operationPathResolver, ContainerInterface $container, array $formats, bool $enableGraphql, array $resourceClassDirectories = [], SubresourceOperationFactoryInterface $subresourceOperationFactory = null)
```
dunglas: `bool $enableGraphql` must be the last parameter and have a default value or it's a BC break.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+                'args' => [
+                    'first' => [
+                        'type' => GraphQLType::int(),
+                        'description' => 'Returns the first n elements from the list.'
+                    ],
+                ],
+                'resolve' => function ($root, $args, $context, ResolveInfo $resolveInfo) use ($resource, $operationName) {
+                    // @TODO Pagination
+                    $collection = $this->collectionDataProvider->getCollection($resource, $operationName);
+                    return $this->normalizer->normalize($collection, null, ['graphql_selected_fields' => $resolveInfo->getFieldSelection()]);
```
dunglas: Cant' we use https://github.com/symfony/symfony/pull/18834 instead of `graphql_selected_fields`? I've introduced this option exactly for GraphQL support :)

dunglas:
I'm wondering if we should not totally separate our current (REST) "operations" subsystem from GraphQL. They are not the same concept and aren't really compatible. CRUD fit well with REST (resource based API), but I don't think it fits well with GraphQL. Maybe should we provide in the future a different way to implement REST operations than mutations. I guess it will be very difficult to implement generic operations that work for both REST and GraphQL.

WDYT @alanpoulain @raoulclais @api-platform/core-team?

dunglas:
On the other end, it should be easy to provide basic CRUD mutations for GraphQL... Maybe can we just add a new option to operations to define if it's a rest operation, a graphql mutation, or both.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: Forget about that, content-type can also be `application/json`, so it's a bad idea. I'm still not a found of checking the route name (it creates a hard coupling with the Symfony Routing component).

Maybe can you just set a new `_graphql` attributes to true in the action, and check it here. It will be easier for end users to hook into this.

src/Action/GraphqlEntrypointAction.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->schemaBuilder = $schemaBuilder;
+        $this->executor = $executor;
+    }
+
+    public function __invoke(): JsonResponse
+    {
+        $request = $this->requestStack->getCurrentRequest();
+
+        // @TODO Manage GET request.
+        $input = json_decode($request->getContent(), true);
```
dunglas: You should check that the incoming `Content-Type` is `application/json` or `application/graphql`:

* if it's `application/json`, the current behavior is OK
* if it's `application/graphql`, the raw content is a GraphQL query (so no `json_decode` needed)
* throw an error for all other content types (maybe can you also serve https://github.com/graphql/graphiql if the content type is `text/html`)?

theofidry:
@dunglas in my experience what I find painfully laborious with GraphQL is to re-define those crudish operations. It's not like you need all the CRUD endpoints all the time, even with a REST API, but they are a PITA to re-define again and again. Another thing that API Platform nailed right is to provide a standardised structure whereas with a GraphQL API, you tend to have a bit of everything simply because there is no standard "enforced".

src/Bridge/Symfony/Routing/ApiLoader.php
alanpoulain: Could you explain why? It seems to me this is an internal service. I don't see what it could break.

soyuka: not marked as `@internal`

dunglas: Some people (including me :D) use API Platform as a reusable library in other contexts (Silex, Laravel...). This change will break this class of uses.

src/Bridge/Graphql/Executor.php
alanpoulain: It's mainly for testability. To be able to mock it.

soyuka: You should be able to mock GraphQL directly then?

dunglas: Indeed, the `GraphQL` class is not final, so you can mock it.

src/Bridge/Graphql/Type/SchemaBuilder.php
alanpoulain: I can replace `graphql_selected_fields` by `attributes`, yes. But it will not change the initial issue: to replace the generated URI (call to `normalizeRelation`) by the subresource data.

alanpoulain: I'm not even sure this is a good idea. If I replace it, it will have consequences when using `attributes` in an SQL context. See its use here: https://github.com/alanpoulain/core/blob/2ed4e8c54ac8579857fc401ad32b18ece0f962a4/src/Serializer/AbstractItemNormalizer.php#L416

dunglas: Ok, keep it as is for now :)

alanpoulain: Using `attributes` is not useful in this context because the fields are filtered by graphql-php. Maybe it can improves performance though?

dunglas: It will definitely improve performances, serialization is a costly operation, we should definitely prevent serializing useless data.

src/Bridge/Graphql/Executor.php
alanpoulain: Even a static method?

src/Bridge/Symfony/Routing/ApiLoader.php
alanpoulain: OK thanks for the explanation.

src/Bridge/Graphql/Executor.php
dunglas: Indeed.

mvrhov:
Watch out. the server part of GraphQL has a patent on it! I don't care for software patents as I'm from WU, but others using this inside Us should be careful.

dunglas:
We must indeed make that clear in the documentation. And it's why it will be an optional disablable feature.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public function createCollectionResolver(string $resourceClass, string $operationName): callable
+    {
+        return function ($root, $args, $context, ResolveInfo $resolveInfo) use ($resourceClass, $operationName) {
+            // @TODO Pagination.
+            $request = $this->requestStack->getCurrentRequest();
+            $collectionArgs = $request->attributes->get('_graphql_collections_args', []);
+            $collectionArgs[$resourceClass] = $args;
+            $request->attributes->set('_graphql_collections_args', $collectionArgs);
```
meyerbaptiste:
```php
$request->attributes->set('_graphql_collections_args', [$resourceClass => $args] + $request->attributes->get('_graphql_collections_args', []));

File: src/Bridge/Graphql/Resolver/CollectionResolverFactory.php [hunk 21]
```diff
@@ -0,0 +1,59 @@
...
+        $this->collectionDataProvider = $collectionDataProvider;
+        $this->normalizer = $normalizer;
+        $this->requestStack = $requestStack;
+    }
+
+    public function createCollectionResolver(string $resourceClass, string $operationName): callable
+    {
+        return function ($root, $args, $context, ResolveInfo $resolveInfo) use ($resourceClass, $operationName) {
+            // @TODO Pagination.
+            $request = $this->requestStack->getCurrentRequest();
```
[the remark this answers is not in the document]
meyerbaptiste: You have to test `RequestStack::getCurrentRequest()` is not null.

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -0,0 +1,48 @@
...
+    public function __construct(ItemDataProviderInterface $itemDataProvider, NormalizerInterface $normalizer)
+    {
+        $this->itemDataProvider = $itemDataProvider;
+        $this->normalizer = $normalizer;
+    }
+
+    public function createItemResolver(string $resourceClass, string $operationName): callable
+    {
+        return function ($root, $args, $context, ResolveInfo $resolveInfo) use ($resourceClass, $operationName) {
+            $item = $this->itemDataProvider->getItem($resourceClass, $args['id'], $operationName);
```
meyerbaptiste: See my previous comment about `$this` in an anonymous function.

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -0,0 +1,48 @@
...
+            if (null === $item) {
+                return null;
+            }
+            $selectedFields = $resolveInfo->getFieldSelection();
+
+            return $this->normalizer->normalize($item, null, ['graphql_selected_fields' => $selectedFields]);
```
meyerbaptiste: One line:
```php
return ($item = $itemDataProvider->getItem($resourceClass, $args['id'], $operationName))
    ? $normalizer->normalize($item, null, ['graphql_selected_fields' => $resolveInfo->getFieldSelection()])
    : null;
```

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -76,6 +78,10 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
         }
 
         $itemsPerPage = $resourceMetadata->getCollectionOperationAttribute($operationName, 'pagination_items_per_page', $this->itemsPerPage, true);
+        if ($this->graphqlEnabled) {
```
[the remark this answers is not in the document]
alanpoulain: Yes... Nice catch!

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
alanpoulain: In which cases it could occur?

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
alanpoulain: Sometimes, I prefer to be more verbose to avoid potential bugs and to modify more easily the code. But if it is preferred in API Platform, no problem.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,237 @@
...
+                $graphqlType = GraphQLType::int();
+                break;
+            case Type::BUILTIN_TYPE_FLOAT:
+                $graphqlType = GraphQLType::float();
+                break;
+            case Type::BUILTIN_TYPE_STRING:
+                $graphqlType = GraphQLType::string();
+                break;
+            case Type::BUILTIN_TYPE_OBJECT:
+                if (\DateTime::class === $type->getClassName()) {
```
meyerbaptiste:
```php
if (is_subclass_of($type->getClassName(), \DateTimeInterface::class)) {
```

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,237 @@
...
+
+        return $type->isCollection() ? GraphQLType::listOf($graphqlType) : $graphqlType;
+    }
+
+    /**
+     * Get the available operations for a resource.
+     */
+    private function getOperations(ResourceMetadata $resourceMetadata, bool $isQuery, bool $isItem): array
+    {
+        return array_filter($isItem ? $resourceMetadata->getItemOperations() : $resourceMetadata->getCollectionOperations(), function ($operation) use ($isQuery) {
```
meyerbaptiste: Prefer a `foreach` with a generator here:
```php
$operations = $isItem ? $resourceMetadata->getItemOperations() : $resourceMetadata->getCollectionOperations();

foreach ($operations as $operationName => $operation) {
    if (isset($operation['controller']) || !isset($operation['method'])) {
        continue;
    }

    if ($isQuery && Request::METHOD_GET !== $operation['method'] || !$isQuery && Request::METHOD_GET === $operation['method']) {
        continue;
    }

    yield $operationName => $operation;
}

File: src/Bridge/Graphql/Type/SchemaBuilder.php [hunk 27]
```diff
@@ -0,0 +1,237 @@
...
+
+            // @TODO Manage multiple identifiers?
+            return [
+                'name' => $property,
+                'type' => GraphQLType::nonNull($this->convertType($propertyType)),
+                'description' => $propertyMetadata->getDescription(),
+            ];
+        }
+
+        throw new \LogicException(sprintf('Missing identifier field for resource "%s"', $resource));
```
meyerbaptiste:
```php
"Missing identifier field for resource \"$resource\"."
```

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,237 @@
...
+     * Gets the field arguments of the identifier of a given resource.
+     */
+    private function getResourceIdentifierFieldArguments(string $resource): array
+    {
+        foreach ($this->propertyNameCollectionFactory->create($resource) as $property) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resource, $property);
+            $propertyIsIdentifier = $propertyMetadata->isIdentifier() ?? false;
+            $propertyType = $propertyMetadata->getType();
+
+            if (!$propertyIsIdentifier || null === $propertyType) {
```
meyerbaptiste: `|| !$propertyType`

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,237 @@
...
+                    continue;
+                }
+            }
+        }
+
+        return $fields;
+    }
+
+    /**
+     * Gets the field arguments of the identifier of a given resource.
```
meyerbaptiste: Missing `@throws` annotation.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,237 @@
...
+                'type' => GraphQLType::nonNull($this->convertType($propertyType)),
+                'description' => $propertyMetadata->getDescription(),
+            ];
+        }
+
+        throw new \LogicException(sprintf('Missing identifier field for resource "%s"', $resource));
+    }
+
+    /**
+     * Converts a built-in type to its GraphQL equivalent.
```
meyerbaptiste: Missing `@throws` annotation.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
meyerbaptiste: When? We don't know but it could occur this is what the contract says and we have to respect the method signature!

dunglas: It definitely occurs in when running this class in command line.

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -0,0 +1,237 @@
...
+     */
+    private function getResourceTypeFields(string $resource): array
+    {
+        $fields = [];
+
+        foreach ($this->propertyNameCollectionFactory->create($resource) as $property) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resource, $property);
+
+            $propertyType = $propertyMetadata->getType();
+            if (null !== $propertyType) {
```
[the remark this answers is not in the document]
dunglas: Are you sure @meyerbaptiste? We need an explicit check for null values (not for boolean).

src/Bridge/Graphql/Type/SchemaBuilder.php
dunglas: Sructinizr (legitimately) add warnings when not checking explicitly the type when dealing with `null`values. I would keep this as is.

src/Bridge/Graphql/Type/SchemaBuilder.php
dunglas: We can even nitpick with `is_a($type->getClassName(), \DateTimeInterface::class, true)`

src/Bridge/Graphql/Type/SchemaBuilder.php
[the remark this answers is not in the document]
meyerbaptiste: Even if the return type is `object|null`?

`null !== $property = $propertyMetadata->getType()` then.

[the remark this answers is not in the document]
dunglas: We should double check.

alanpoulain:
Added: pagination support.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,96 @@
...
+ */
+final class CollectionResolverFactory extends AbstractResolverFactory implements CollectionResolverFactoryInterface
+{
+    private $collectionDataProvider;
+    private $normalizer;
+    private $identifiersExtractor;
+    private $requestStack;
+    private $paginationEnabled;
+
+    public function __construct(CollectionDataProviderInterface $collectionDataProvider, SubresourceDataProviderInterface $subresourceDataProvider, NormalizerInterface $normalizer, IdentifiersExtractorInterface $identifiersExtractor, RequestStack $requestStack, bool $paginationEnabled)
```
dunglas: IMO the request should not be injected that deep. The low level GraphQL implementation should be agnostic of the transport layer (HTTP, AMQP...).

Can't we extract the offset and other data needed in the controller layer and pass the base64-encoded values to `createCollectionResolver`?

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,96 @@
...
+            } else {
+                $collection = $this->collectionDataProvider->getCollection($resourceClass, $operationName);
+            }
+
+            if ($this->paginationEnabled) {
+                $offset = 0;
+                if (isset($args['after'])) {
+                    $after = base64_decode($args['after'], true);
+                    $offset = (int) $after;
+                    $offset = false === $after ? $offset : ++$offset;
```
dunglas: Shouldn't we throw an error if `base64_decode` returns an error?

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,96 @@
...
+                }
+                $data = ['edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];
+                if ($collection instanceof PaginatorInterface) {
+                    $totalItems = $collection->getTotalItems();
+                    if ($totalItems > 0) {
```
dunglas: Can be merge with the previous if:

```
if ($collection instanceof PaginatorInterface && $totalItems = $collection->getTotalItems() > 0)
```

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,96 @@
...
+            $rootProperty = $info->fieldName;
+            if (isset($root[$rootProperty])) {
+                $rootIdentifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($rootClass);
+                $subresource = $this->getSubresource($rootClass, $root, $rootIdentifiers, $rootProperty, $resourceClass, $operationName, true);
+                $collection = $subresource ?? [];
+            } else {
+                $collection = $this->collectionDataProvider->getCollection($resourceClass, $operationName);
+            }
+
+            if ($this->paginationEnabled) {
```
dunglas: I would move all the following block in a new method to improve readability and decrease the cyclomatic complexity.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
alanpoulain: This resolver does not need the request to resolve the query. It's the PaginationExtension who needs to retrieve the offset and the limit.
The resolver adds the offset / limit argument of the GraphQL query to the request attributes to be used by the PaginationExtension.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
alanpoulain: I can yes, if you prefer. I thought it was the responsibility of the client to use the cursors we give them.

dunglas: Yes but we cannot trust client input.

.travis.yml
```diff
@@ -11,7 +11,7 @@ matrix:
   include:
     - php: '7.0'
     - php: '7.1'
-      env: coverage=1 lint=1
+      env: coverage=0 lint=1
```
dunglas: We need to restore this later, but I've no idea how to fix the problem for now.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -253,7 +253,7 @@ private function getSerializerContext(string $resourceClass, string $contextType
             $request = $this->requestStack->getCurrentRequest();
         }
 
-        if (null !== $this->serializerContextBuilder && null !== $request) {
+        if (null !== $this->serializerContextBuilder && null !== $request && !$request->attributes->get('_graphql')) {
```
soyuka: why?

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -92,8 +96,18 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
             throw new InvalidArgumentException('Item per page parameter should not be less than or equal to 0');
         }
 
+        $firstResult = ($request->query->get($this->pageParameterName, 1) - 1) * $itemsPerPage;
+        if ($request->attributes->get('_graphql')) {
+            $collectionArgs = $request->attributes->get('_graphql_collections_args', []);
+            if (isset($collectionArgs[$resourceClass]['after'])) {
+                $after = base64_decode($collectionArgs[$resourceClass]['after'], true);
```
soyuka: `\base64_decode` as below? Don't remember the policy on this but we need it to be consistent.

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -0,0 +1,87 @@
...
+            } else {
+                $identifiers = [];
+                $uniqueIdentifier = [];
+                foreach ($rootIdentifiers as $rootIdentifier) {
+                    if (!isset($args[$rootIdentifier])) {
+                        continue;
+                    }
+
+                    if (\is_array($args[$rootIdentifier])) {
+                        if (\count($args[$rootIdentifier]) > 1) {
```
soyuka: Same on these, we currently are not prefixing with `\` in the base code.

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -0,0 +1,87 @@
...
+
+                    $identifiers[] = $rootIdentifier.'='.$args[$rootIdentifier];
+                    $uniqueIdentifier[] = $args[$rootIdentifier];
+                }
+
+                if (!$identifiers) {
+                    return null;
+                }
+
+                $item = $this->itemDataProvider->getItem($resourceClass, \count($identifiers) > 1 ? implode(';', $identifiers) : $uniqueIdentifier[0]);
```
soyuka: Just a note to myself but this would need refactoring if we can implement some kind of identifiers denormalization (https://github.com/api-platform/core/issues/1309) not linked to any bridge (code duplication).

src/DataProvider/ChainSubresourceDataProvider.php
```diff
@@ -35,7 +35,7 @@ public function __construct(array $dataProviders)
     /**
      * {@inheritdoc}
      */
-    public function getSubresource(string $resourceClass, array $identifiers, array $context, string $operationName)
+    public function getSubresource(string $resourceClass, array $identifiers, array $context, string $operationName = null)
```
soyuka: will conflict haha

update-js.sh
```diff
@@ -10,4 +10,43 @@ mkdir -p $dest
...
+dest=src/Bridge/Symfony/Bundle/Resources/public/graphiql/
+
+yarn add --production --no-lockfile graphiql
+if [ -d $dest ]; then
+  rm -Rf $dest
+fi
+mkdir -p $dest
+cp node_modules/graphiql/graphiql.min.js $dest
+cp node_modules/graphiql/graphiql.css $dest
+
```
soyuka: gosh :cat: maybe we should kinda create a js repository for all the graphql stuff in the future and just une one module for all these.

This also means that we're now publishing react sources in the source code of api-platform, totally :-1: on this personally...

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
alanpoulain: Because of https://github.com/api-platform/core/blob/master/src/Serializer/SerializerContextBuilder.php#L42. There is no serialization context in the request.

soyuka: Shouldn't there be one? I mean, usually `_api_resource_class` is declared on every route no?

Also maybe that the `SerializerContextBuilder` shouldn't throw at all.

src/Action/GraphqlEntrypointAction.php
```diff
@@ -0,0 +1,107 @@
...
+        $executionResult = $this->executor->executeQuery($this->schemaBuilder->getSchema(), $query, null, null, $variables, $operation);
+
+        return new JsonResponse($executionResult->toArray($this->debug));
+    }
+
+    private function parseRequest(Request $request): array
+    {
+        $query = $request->query->get('query');
+        $operation = $request->query->get('operation');
+        if ([] !== $variables = $request->query->get('variables', [])) {
```
meyerbaptiste: `if ($variables = $request->...) {`

src/Api/IdentifiersExtractor.php
```diff
@@ -43,6 +43,24 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
     /**
      * {@inheritdoc}
      */
+    public function getIdentifiersFromResourceClass(string $resourceClass): array
+    {
+        $identifiers = [];
+        foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $property) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $property);
+            $propertyIsIdentifier = $propertyMetadata->isIdentifier() ?? false;
+            if (!$propertyIsIdentifier) {
```
meyerbaptiste:
```php
if ($propertyMetadata->isIdentifier() ?? false) {
    $identifiers[] = $property;
}
```

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -83,6 +83,10 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
         }
 
         $itemsPerPage = $resourceMetadata->getCollectionOperationAttribute($operationName, 'pagination_items_per_page', $this->itemsPerPage, true);
+        if ($request->attributes->get('_graphql')) {
```
meyerbaptiste: `has()` instead of `get()`?

src/Bridge/Graphql/Resolver/AbstractResolverFactory.php
```diff
@@ -0,0 +1,52 @@
...
+use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+
+/**
+ * Base resolver factory.
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ *
+ * @internal
+ */
+class AbstractResolverFactory
```
meyerbaptiste: Missing `abstract` here!

src/Bridge/Graphql/Resolver/AbstractResolverFactory.php
```diff
@@ -0,0 +1,52 @@
...
+        $identifiers = [];
+        foreach ($rootIdentifiers as $rootIdentifier) {
+            if (isset($rootResolvedFields[$rootIdentifier])) {
+                $identifiers[$rootIdentifier] = $rootResolvedFields[$rootIdentifier];
+            }
+        }
+
+        return $this->subresourceDataProvider->getSubresource($subresourceClass, $identifiers, [
+            'property' => $rootProperty,
+            'identifiers' => array_map(function ($rootIdentifier) use ($rootClass) {return [$rootIdentifier, $rootClass]; }, $rootIdentifiers),
```
meyerbaptiste: I prefer a `foreach` for performances.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,109 @@
...
+
+        $this->collectionDataProvider = $collectionDataProvider;
+        $this->normalizer = $normalizer;
+        $this->identifiersExtractor = $identifiersExtractor;
+        $this->requestStack = $requestStack;
+        $this->paginationEnabled = $paginationEnabled;
+    }
+
+    /**
+     * @throws \InvalidArgumentException
```
meyerbaptiste: You have to remove the backslash.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,109 @@
...
+                return $data;
+            }
+
+            $offset = 0;
+            if (isset($args['after'])) {
+                $after = \base64_decode($args['after'], true);
+                if (false === $after) {
+                    throw new InvalidArgumentException(sprintf('Cursor %s is invalid', $args['after']));
+                }
+                $offset = (int) $after;
```
meyerbaptiste: `1 + (int) $after`;

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -0,0 +1,109 @@
...
+                ++$offset;
+            }
+
+            $data = ['edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];
+            if ($collection instanceof PaginatorInterface && ($totalItems = $collection->getTotalItems()) > 0) {
```
meyerbaptiste: Can't we also support `PartialPaginatorInterface`?

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -0,0 +1,87 @@
...
+                    if (!isset($args[$rootIdentifier])) {
+                        continue;
+                    }
+
+                    if (\is_array($args[$rootIdentifier])) {
+                        if (\count($args[$rootIdentifier]) > 1) {
+                            throw new InvalidArgumentException('Composite identifiers are not allowed for a resource already used as a composite identifier');
+                        }
+
+                        $identifiers[] = $rootIdentifier.'='.current($args[$rootIdentifier]);
```
meyerbaptiste: If you use `current()` to get the first element then it's not really safe!

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: Maybe can we find something expliciter, but as GraphQL isn't REST, it looks okish that `_api_resource_class` doesn't exist.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
dunglas: We should use `\` everywhere, it's faster. IIRC, there is a PHP CS Fixer rule for that.

update-js.sh
dunglas: Creating our own JS repo will not help. AFAIK, it's not possible to upload artifacts on Packagist, so the JS must be included in the repo. It's still a dependency, the way it is downloaded doesn't change anything from a legal POV.

This script can be dramatically improved but we can still handle that later

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -0,0 +1,87 @@
...
+                    if (!isset($args[$rootIdentifier])) {
+                        continue;
+                    }
+
+                    if (\is_array($args[$rootIdentifier])) {
+                        if (\count($args[$rootIdentifier]) > 1) {
+                            throw new InvalidArgumentException('Composite identifiers are not allowed for a resource already used as a composite identifier');
+                        }
+
+                        $identifiers[] = $rootIdentifier.'='.\current($args[$rootIdentifier]);
```
meyerbaptiste: If you use `current()` to get the first item then it's not really safe!

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
meyerbaptiste: So we should add and enable this fixer in our php-cs-fixer config!

.travis.yml
meyerbaptiste: What is the problem with the coverage?

dunglas: We reach the `ulimit` defined by Travis... Too many open files.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
dunglas: `get` is safer IMO (in case it is explicitly set to `false`)

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
soyuka: Okay then!

soyuka: should use `has` btw as pointed by @meyerbaptiste

src/Bridge/Graphql/Resolver/AbstractResolverFactory.php
dunglas: Btw both loops can be merged. I take care of your comments.

src/Bridge/Graphql/Resolver/CollectionResolverFactory.php
dunglas: 👍 but in another PR please

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
dunglas: Yes but it is a risky one, IIRC, it breaks our tests.

soyuka: wdyt we remove `\` for now and we will do this in another pr.

.travis.yml
soyuka: can't we `ulimit -n 4000` before install? It happens on my local computer too when testing the `Extension`.

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
meyerbaptiste: Okay, there are a lot of places where he uses `has()` so it would also be safer to replace them!

src/Bridge/Graphql/Resolver/ItemResolverFactory.php
meyerbaptiste: Agreed!

.travis.yml
dunglas: Unfortunately no, Travis disallow to increase the default value.

alanpoulain:
https://medium.com/@leeb/relicensing-the-graphql-specification-e7d07a52301b

theofidry:
🎉  I will try it out ASAP but that's unlikely to be before 2-3 weeks

jorgeluisacostaalonso:
Hello there!
There 's a little quick-start for grapqh and apiplatform!! Or something explaining how start!!
have a nice day

dunglas:
@jorgeluisacostaalonso, you need to use `api-platform/core:dev-master`. Then require `webonyx/graphql-php`, and you're down, the GraphQL API is available at the `/graphql` endpoint.

Issue #1299, opened 2017-08-01
Deprecate NelmioApiDoc in favor of Swagger


soyuka:
Ref: https://github.com/api-platform/core/issues/1131#issuecomment-319079874

Simperfit:
I will be doing it.

lvillarino:
How we will go to expose a swagger documentation for the code below after you remove the NelmioApiDoc ?

```
class DefaultController extends BaseController
{
	protected $_context = 'AppBundle.Controller.DefaultController';

	/**
	 *
	 * Example
	 *
	 * ### Response ###
	 *
	 *  Json string
	 *
	 * @ApiDoc(
	 *  views = { "default" },
	 *  section="Swagger",
	 *     statusCodes={
	 *         200="Returned Json",
	 *         400={
	 *           "error",
	 *         }
	 *     }
	 * )
	 *
	 * @Route("/custom_route/hello_julia")
	 * @Method("GET")
	 */
	public function helloJuliaAction(Request $request)
	{
			$responseData = array(
				'status' => Response::HTTP_OK,
				'msg' => "Hello Julia",
			);
			return $this->createApiResponse($responseData, Response::HTTP_OK);
		}
	}
}


Thanks


```

jdeniau:
Seems fixed and merged @soyuka

PR #1359, opened 2017-09-11
Deprecate NelmioApiDoc 2 support

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

NelmioApiDocBundle 3 has native support for API Platform. Deprecating our support for v2.

### DIFF
--- a/tests/Bridge/NelmioApiDoc/Extractor/AnnotationsProvider/ApiPlatformProviderTest.php
+++ b/tests/Bridge/NelmioApiDoc/Extractor/AnnotationsProvider/ApiPlatformProviderTest.php
@@ -32,6 +32,9 @@ use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
  * @author Teoh Han Hui <teohhanhui@gmail.com>
+ *
+ * @group legacy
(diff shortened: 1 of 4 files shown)


soyuka:
fix #1299 ?

Adding those annotations didn't work, I think we should use `trigger_error` in the constructor instead of in the file.

tests/Bridge/NelmioApiDoc/Extractor/AnnotationsProvider/ApiPlatformProviderTest.php
```diff
@@ -32,6 +32,8 @@
 
 /**
  * @author Teoh Han Hui <teohhanhui@gmail.com>
+ *
+ * @group legacy
```
meyerbaptiste: We should also add a `@expectedDeprecation`.

soyuka: done

dunglas:
Thanks @soyuka

PR #1362, opened 2017-09-12
Add swagger definition keys to changelog fix #1361

### LINKED ISSUE #1361: Can't reference issues since minor release 2.1.0
We used to reference definitions of entities with the schema: `{entity_name}_{hash}` for the swagger doc:

**src/AppBundle/Resources/config/api_resources/resources.yml**
```yml
Acme\Api\Model\Image:
    itemOperations:
        get:
             method: "GET"
             path: "/path/{id}"
             swagger_context:
                 parameters:
                  - name: "id"
                    required: true
                    type: "string"
                    in: "path"
                    description: "ID of the image resource."
                 responses:
                     200:
                        description: "Image returned"
                        schema:
                            $ref: "#/definitions/Image_45f28e6683c923d629dea6466d6b7259"
```

this worked in `2.0.x`. But since `2.1.0` the definitions are gone.

Don't know if we used an undocumented "feature" (couldn't find it in the documentation) or if this is a bug.

soyuka:


dunglas:
👍 but should target the `2.1` branch.

Issue #1357, opened 2017-09-11
Validation : support of errors levels


pierre-H:
Hello,

Thanks again for Api-Platform !

I'd like to use [different error levels](https://symfony.com/doc/current/validation/severity.html).
What I'd like is to have in the response (with all constraints message) the severity value. Is there a way to achieve this ?

Thank you !

dunglas:
I wasn't even aware of this feature. After double checking, there is nothing regarding severity directly in the Validator component, it's a custom usage of the `payload` object.

Exposing `payload` objects client-side can be implemented easily has we already do for `violations:

* For Hydra: https://github.com/api-platform/core/blob/master/src/Hydra/Serializer/ConstraintViolationListNormalizer.php#L64
Anf for RFC7807: https://github.com/api-platform/core/blob/master/src/Problem/Serializer/ConstraintViolationListNormalizer.php#L54

It is as easy as adding `'payload' => $violation->getConstraint()->getPayload()` after `violations`. However, maybe the payload defined by the user may contain sensitive data (@webmozart, do you have a hint about that?), if it's the case, we should provide a configuration option to prevent payload serialization. We should also not add a `payload` key if no payload is provided.

Would you mind opening a PR?

norkunas:
I would suggest to also serialize violation constraint error code

pierre-H:
An option where we can configure which payload fields to serialize could be better : 
```yml
api_platform:
    payload_validation_keys:
    - severity
    - other_field
```
This way, we can have sensitive and public data in the playload.

pierre-H:
@dunglas I created a [PR](https://github.com/api-platform/core/pull/1364) but I don't know why test don't pass ...

PR #1364, opened 2017-09-12
Add payload to ConstraintViolationListNormalizer

pierre-H:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1357 
| License       | MIT
| Doc  PR      | https://github.com/api-platform/docs/pull/263

see the ticket for more information


meyerbaptiste:
> I don't know why test don't pass

Run php-cs-fixer: https://travis-ci.org/api-platform/core/jobs/274610519#L1072-L1158

src/Hydra/Serializer/ConstraintViolationListNormalizer.php
```diff
@@ -45,11 +45,16 @@ public function normalize($object, $format = null, array $context = [])
...
 
         foreach ($object as $violation) {
-            $violations[] = [
+            $violationData = [
                 'propertyPath' => $violation->getPropertyPath(),
-                'message' => $violation->getMessage(),
+                'message' => $violation->getMessage()
             ];
 
+            if (! empty($violation->getConstraint()->payload)) {
```
dunglas: I suggest to store `$violation->getConstraint()->payload` in a variable to prevent an extra function call. Btw you can remove the call to `empty`, it is useless.

pierre-H: Done ! Thanks

pierre-H:
@dunglas I added the option and some tests but I am not that tests are complete for ApiPlatformExtensionTest .

src/Hydra/Serializer/ConstraintViolationListNormalizer.php
```diff
@@ -31,9 +31,15 @@
      */
     private $urlGenerator;
 
-    public function __construct(UrlGeneratorInterface $urlGenerator)
+    /**
+     * @var bool
+     */
+    private $serializePayload;
+
+    public function __construct(UrlGeneratorInterface $urlGenerator, bool $serializePayload)
```
soyuka: `serializePayload` needs a default value

pierre-H: @soyuka  Done

pierre-H:
PHP-CS-Fixer and tests are ok

src/Api/OperationTypeDeprecationHelper.php
```diff
@@ -38,7 +38,7 @@ public static function getOperationType($operationType): string
         if (is_bool($operationType)) {
             @trigger_error('Using a boolean for the Operation Type is deprecrated since API Platform 2.1 and will not be possible anymore in API Platform 3', E_USER_DEPRECATED);
 
-            $operationType = $operationType === true ? OperationType::COLLECTION : OperationType::ITEM;
+            $operationType = true === $operationType ? OperationType::COLLECTION : OperationType::ITEM;
```
meyerbaptiste: `true ===` is useless here!

soyuka: though this is my bad haha, how did this end up in the code :P.

meyerbaptiste: It looks like PHP-CS-Fixer has a new rule for Yoda style 😛

dunglas: We should probably apply it on lower branches.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -55,6 +55,7 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->scalarNode('name_converter')->defaultNull()->info('Specify a name converter to use.')->end()
                 ->scalarNode('path_segment_name_generator')->defaultValue('api_platform.path_segment_name_generator.underscore')->info('Specify a path name generator to use.')->end()
+                ->booleanNode('serialize_payload_constraint')->defaultFalse()->info('Enable the serialization of the payload field when a validation error is thrown.')->end()
```
dunglas: What do you think of something like:

```yaml
validator:
    serialize_payload: true
```

Some people are thinking about new options regarding the validator, this way we'll be feature-proof?

src/Api/OperationTypeDeprecationHelper.php
pierre-H: Yes it was PHP-CS-Fixer

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
pierre-H: Ok, I will change that tomorrow

pierre-H:
Isn't better to define in the configuration which fields we want to serialize ?

For example : 
```yml
validator:
    serialize_payload_fields:
        severity
        anotherField
```

This way we can use payload field with sensitive and public data?

dunglas:
@pierre-H good idea! and 

```
serialize_payload_fields: true
```

to serialize all fields.

soyuka:
As I've applied the cs fixes to `2.1`, @dunglas may you merge `2.1` on to master. Then @pierre-H can `git rebase origin/master` so that the patch is cleaner :). 

Thanks!

pierre-H:
@dunglas done

pierre-H:
Should I do something more ?

dunglas:
Thanks @pierre-H

pierre-H:
With pleasure !
Thank you for your great work !

Issue #1363, opened 2017-09-12
POST (create new item) with nested data


GonZOO82:
I would like POST with nested data, but don't succeed.

Sample:
```
{
    "title": "test",
    "cover": {  //ManyToOne connection to MediaItem Entity
        "id": 1 // This MediaItem is exist
    }
}
```
Response: `Update is not allowed for this operation.`

but when I update item (PUT) then process is good.

I don't want to use this code:
```
{
    "title": "test",
    "cover": "/media_item/1"
}
```

soyuka:
What format are you using (ie `Accept` header)?

GonZOO82:
I try `application/json` and `application/ld+json`

json -->  {"id":1} .  --> Update is not allowed for this operation.
ld+json --> {"@id":1} --> Update is not allowed for this operation.

json -->  {"@id":1} .  -->  try run SQL nested data INSERT command
ld+json --> {"id":1} --> try run SQL nested data INSERT command

soyuka:
What works with `json-ld`:

POST /books

```
{
  "cover": "/covers/1",
  "title": "foo"
}
```

About `json` I'm investigating.

soyuka:
https://github.com/api-platform/api-platform/issues/249 not supported yet with plain IDs.

GonZOO82:
@soyuka not supported when I POST data, but supported in PUT method?!
```
if (!$normalization && !isset($context['api_allow_update'])) {
    $context['api_allow_update'] = Request::METHOD_PUT === $request->getMethod();
}
``` 
If I PUT data with nested content then it's successful.

Video: https://youtu.be/kEuzdS4z3wk

mab05k:
I was able to get something like this to work by adding "api_allow_update=true" in my deserialization context.

Example
```yaml
AppBundle\Entity\Resource:
    attributes:
        normalization_context:
            groups: [ 'resource.read' ]
        denormalization_context:
            api_allow_update: true
            groups: [ 'resource.write' ]
```

This allowed me to POST in data with full object references when I had embedded objects. It worked for me using JSON, LD+JSON, and HAL+JSON.

Simperfit:
This seems to be resolved.

PR #1365, opened 2017-09-12
Allow plain IDs with `allow_plain_identifiers`

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no (depends on the POV)
| New feature?  | yes/no (same)
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/249 https://github.com/api-platform/core/issues/1363
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\Serializer;
 use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\OperationType;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
+use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
@@ -44,8 +45,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     protected $resourceClassResolver;
     protected $propertyAccessor;
     protected $localCache = [];
+    protected $itemDataProvider;
+    protected $allowPlainIdentifiers;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, ItemDataProviderInterface $itemDataProvider = null, bool $allowPlainIdentifiers = false)
     {
         parent::__construct($classMetadataFactory, $nameConverter);
 
@@ -54,6 +57,8 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $this->iriConverter = $iriConverter;
         $this->resourceClassResolver = $resourceClassResolver;
         $this->propertyAccessor = $propertyAccessor ?: PropertyAccess::createPropertyAccessor();
+        $this->itemDataProvider = $itemDataProvider;
+        $this->allowPlainIdentifiers = $allowPlainIdentifiers;
 
         $this->setCircularReferenceHandler(function ($object) {
             return $this->iriConverter->getIriFromItem($object);
@@ -299,6 +304,17 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         }
 
         if (!is_array($value)) {
+            // repeat the code so that IRIs keep working with the json format
+            if (true === $this->allowPlainIdentifiers && $this->itemDataProvider) {
+                try {
+                    return $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true]);
+                } catch (ItemNotFoundException $e) {
+                    throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+                } catch (InvalidArgumentException $e) {
+                    // Give a chance to other normalizers (e.g.: DateTimeNormalizer)
+                }
+            }
+
             throw new InvalidArgumentException(sprintf(
                 'Expected IRI or nested document for attribute "%s", "%s" given.', $attributeName, gettype($value)
             ));
--- a/features/main/relation.feature
+++ b/features/main/relation.feature
@@ -443,6 +443,34 @@ Feature: Relations support
     }
     """
 
+  Scenario: Create a related dummy with a relation (json)
+    When I add "Content-Type" header equal to "application/json"
+    And I send a "POST" request to "/related_dummies" with body:
+    """
+    {"thirdLevel": "1"}
+    """
+    Then the response status code should be 201
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/RelatedDummy",
+      "@id": "/related_dummies/6",
+      "@type": "https://schema.org/Product",
+      "id": 6,
+      "name": null,
+      "symfony": "symfony",
+      "dummyDate": null,
+      "thirdLevel": "/third_levels/1",
+      "relatedToDummyFriend": [],
+      "dummyBoolean": null,
+      "embeddedDummy": null,
+      "symfony": "symfony",
+      "age": null
+    }
+    """
+
   @dropSchema
   Scenario: Issue #1222
     Given there are people having pets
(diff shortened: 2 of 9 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -298,6 +300,17 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
         }
 
         if (!is_array($value)) {
+            // repeat the code so that IRIs keep working with the json format
```
soyuka: I'd like to test that `Accept` is `application/json` here, not sure if it's possible or wanted...

dunglas: you can just check the `$format` var.

soyuka: not really, `json` is the format for `hal`, `jsonld` and `json` no?

dunglas: No, it's `jsonld` for JSON-LD and `jsonhal` for HAL.

soyuka: Okay, my bad dunno why I thought so...

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -298,6 +301,17 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
...
+            if ('json' === $format && $this->itemDataProvider) {
```
dunglas: I get the idea and the use case, but it hurts me a bit:

* it doesn't fit the hypermedia philosophie of API Platform
* it doesn't work consistently across formats (XML or CSV for instance may need this feature as well)

May I suggest an alternative implementation:

* add a new flag `allow_plain_identifiers` set to false by default
* if this flag is on, execute your code in all circonstances, regardless of the format

dunglas: And then it will be merged in master.

soyuka: You see :p had the same concerns ^^

teohhanhui:
I don't think we need a config option. `@id` in plain JSON (both normalization and denormalization) is a bug. For BC, we should keep it but trigger a deprecation error.

soyuka:
To you we should:
1. enable `@id` only for `json-ld`
2. enable `id` only for `json`, `xml`, `csv`

@dunglas wdyt? I'm also to remove `@id` support in other formats.

teohhanhui:
There is no question that `@id` is defined in the JSON-LD spec. For all of the plain (non-hypermedia) formats (JSON, XML, etc.), there is not really a right or wrong thing to do, but using the item's ID properties seems appropriate.

soyuka:
Agreed @teohhanhui, though IRIs are such great at identifying things that it'd be a shame not to be able to use them for other things (might be a point to keep them with other formats).

soyuka:
For now I can go back to my previous commit to do:

```
if ('jsonld' !== $format && $this->itemDataProvider) {
 //use id 
}
```

dunglas:
I don't agree:

* We already decided to use IRIs as identifiers everywhere. I'm ok to add an option to change this behavior in write mode (= support both plain IDs and IRIs), but not to remove support of IRIs in other formats than JSON-LD and HAL.
* In JSON-LD (and HAL), we should give the possibility to access to the raw `id` property and to the `@id` IRI. It's mandatory to support client-side generated UUIDs for instance.

soyuka:
In that case @dunglas feel free to merge! I see no issues in keeping `@id` and `id` :)

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -43,8 +44,10 @@
     protected $iriConverter;
     protected $resourceClassResolver;
     protected $propertyAccessor;
+    protected $itemDataProvider;
+    protected $allowPlainIdentifiers;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, ItemDataProviderInterface $itemDataProvider = null, $allowPlainIdentifiers = false)
```
dunglas: `bool $allowPlainIdentifiers = false`

meyerbaptiste:
Should we use IRIs as identifiers with JSON API too?

Simperfit:
@dunglas could you re-review it ? @soyuka could you rebase ?

soyuka:
rebased

sroze:
I really don't think this is going a correct choice to do so. What's the point? It's just adding another way to do something already possible... If you look at #1363 it's just because "I don't want to use this code" or in api-platform/api-platform#249 the reason is "I think plain IDs will be used for many years to come".

We are convinced that semantic web is the way to go. The way to properly go towards to this direction is to have people using IRIs as their identifiers... let's educate for this instead of allow "plain IDs" :(

soyuka:
@sroze totally agree when it comes to formats like `json-ld`. Though with standard `json`, to me, accepting an `id` (not an IRI) should be valid.

glnemeth:
@sroze This is needed for example in our project, where we are currently using JSON (not LD-JSON), so having IRIs is not required and forcing IRIs on our developers is unneccessary and adds to the work. Also our routing is not fixed as the project is still evolving, so it's easier to send IDs, instead of constantly following the changes of routing on frontend and backend too.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -299,6 +304,17 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
         }
 
         if (!is_array($value)) {
+            // repeat the code so that IRIs keep working with the json format
+            if (true === $this->allowPlainIdentifiers && $this->itemDataProvider) {
```
sroze: What is `$this->itemDataProvider === null` and `true === $this->allowPlainIdentifiers` ? Instead of silently not doing anything, should we throw an exception instead?

sroze:
> Also our routing is not fixed as the project is still evolving, so it's easier to send IDs, instead of constantly following the changes of routing on frontend and backend too.

That's an interesting argument because that's actually where using IRIs should help you :)

But fair point for the JSON format. If it helps... 👍 then...

src/Serializer/AbstractItemNormalizer.php
soyuka: We could, maybe a bit overkill, this dependency should be a hard dep, it's just nullable to avoid breaking.

antoine64:
Hello guys. Is it possible to merge this in 2.1 branch? I think this is a really important feature and I hope to see it in stable version ASAP. Thanks!

dunglas:
Sorry we respect semver and new features are always merged in minor releases. Patch releases only contain bug and security fixes.

julienpillias:
According to this PR, I should be able to update (PUT) nested document with plain id ? 
Just tried, a, dstille have this error : 

```
Nested documents for attribute "groups" are not allowed. Use IRIs instead
```

```json
{
  "id":1,
  "name": "foo",
  "groups": [
       {
           "id":1,
           "name": "bar"
       }
   ]
}
```

Composer.json : 
```
"api-platform/core": "2.2.x-dev"
```
(composer update done)

config.sjon
```
api_platform:
    allow_plain_identifiers: true
```

Forget something ? Yet not ready  maybe ?

soyuka:
@julienpillias  you have a serialization issue fix your groups and this will indeed work.

julienpillias:
OK, had to put denormalization_context into the nested document too.
But know, having this error :

```
No route matches "/app_dev.php/groups/1"
```

This route exists, but should be http://api.xxxxx.project.local/app_dev.php/groups/1, because of my routing I think : 
```yaml
api:
    resource: '.'
    type:     'api_platform'
    host:     'api.{project_name}'
    defaults:
        project_name: xxxx.project.local
    requirements:
        project_name: xxxx.project.local|xxxx.fr|preprod.xxxx.fr
```

I don't see any option in apiPlatform to specify a route parameter. At least, is it possible to use the default parameter ?

julienpillias:
So just dig into the code, and found this in https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Routing/IriConverter.php (Line 141) : 
```
public function getSubresourceIriFromResourceClass(string $resourceClass, array $context, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
```
Change it to 
```
public function getSubresourceIriFromResourceClass(string $resourceClass, array $context, int $referenceType = UrlGeneratorInterface::ABS_URL): string
```
(ABS_PATH => ABS_URL).
No more problem ! 
I do not know if this could have some side effect ! 
Should we use ABS_URL every where in iriConverter to avoid problem like mine ?

soyuka:
> (ABS_PATH => ABS_URL).
this should be configurable I think but it is not actually :|.

julienpillias:
could try a PR for this.

Something in config.json ? 
config.json
```yaml
api_platform:
    use_absolute_url: true # default will be false
```

soyuka:
Not sure, I think that this is a configuration issue on your side but I have no idea what (something linked to symfony routing configuration) :|.

julienpillias:
On my project, I have a "classic" web site (SF + twig), using mysite.com (or www.mysite.com) AND have the API on api.mysite.com for mobile application for example.

So if apiPlatform check route /entity/id, it would not be an API ressource. It's why I need ABS_URL their, so ressource will be found with http://api.mysite.com/entity/id

Don't think I have a routing problem here, i've checked router:debug, and do not have any problem with api doc, or my "classic" web site".

soyuka:
Can you open an issue about this with all the details? Thanks.

julienpillias:
Already done few week's ago :)
https://github.com/api-platform/api-platform/issues/427

PR #1367, opened 2017-09-12
Swagger Path Identifier to reflect the annotation of ApiProperty(identifier=true)

Bwen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -35,6 +35,7 @@ use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 /**
  * Creates a machine readable Swagger API documentation.
  *
+ * @author Philippe Guilbault <philippe.guilbault@gmail.com>
  * @author Amrouche Hamza <hamza.simperfit@gmail.com>
  * @author Teoh Han Hui <teohhanhui@gmail.com>
  * @author Kévin Dunglas <dunglas@gmail.com>
@@ -302,12 +303,22 @@ final class DocumentationNormalizer implements NormalizerInterface
             return $pathOperation;
         }
 
+        $pathIdentifier = 'id';
+        $pathType = 'string';
+        foreach ($this->propertyNameCollectionFactory->create($resourceClass, []) as $propertyName) {
+            $property = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
+            if ($property->isIdentifier()) {
+                $pathIdentifier = $propertyName;
+                $pathType = $property->getType()->getBuiltinType();
+            }
+        }
+
         $pathOperation['summary'] ?? $pathOperation['summary'] = sprintf('Retrieves a %s resource.', $resourceShortName);
         $pathOperation['parameters'] ?? $pathOperation['parameters'] = [[
-            'name' => 'id',
+            'name' => $pathIdentifier,
             'in' => 'path',
             'required' => true,
-            'type' => 'string',
+            'type' => $pathType,
         ]];
         $pathOperation['responses'] ?? $pathOperation['responses'] = [
             '200' => [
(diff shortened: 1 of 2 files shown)


src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -302,12 +303,22 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
 
+        $pathIdentifier = 'id';
+        $pathType = 'string';
+        foreach ($this->propertyNameCollectionFactory->create($resourceClass, []) as $propertyName) {
+            $property = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
+            if ($property->isIdentifier()) {
+                $pathIdentifier = $propertyName;
+                $pathType = $property->getType()->getBuiltinType();
+            }
+        }
```
soyuka: We have an issue on composite identifiers here. Indeed, a composite identifier would be written as a string (for example `foo=1;bar=2`). With this code, the identifier would only be `bar`./
About the `id` name, I think it's like this so that it's being used in the path from Swagger. I'm not sure how Swagger is identifying the url parameter like this.

You could maybe reuse the same code from a new method instead of repeating it.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -321,6 +317,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
     /**
+     * Returns an array of swagger path parameters according to the provided class' identifiers
+     *
+     * @param string $resourceClass
+     * @return array
+     */
+    private function getPathParameters(string $resourceClass)
+    {
+        $parameters = [];
+        $properties = $this->getClassIdentifiers($resourceClass);
```
meyerbaptiste: `$identifiers` instead of `$properties`?

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -321,6 +317,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+        foreach ($properties as $propertyName => $property) {
+            $parameters[] = [
+                'name' => $propertyName,
+                'in' => 'path',
+                'required' => true,
+                'type' => $property->getType()->getBuiltinType(),
+            ];
+        }
+
+        if (empty($parameters)) {
```
meyerbaptiste:
```php
if (!$identifiers = $this->getClassIdentifiers($resourceClass)) {
    return [[
        'name' => "id",
        'in' => 'path',
        'required' => true,
        'type' => "string",
    ]];
}

$parameters = [];
foreach ($identifiers as $identifier => $property) {
    $parameters[] = [ ... ];
}

return $parameters;

File: src/Swagger/Serializer/DocumentationNormalizer.php [hunk 4]
```diff
@@ -321,6 +317,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+            $parameters = [[
+                'name' => "id",
```
meyerbaptiste: Use simple quotes.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -321,6 +317,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+                'in' => 'path',
+                'required' => true,
+                'type' => "string",
```
meyerbaptiste: Use simple quotes.

tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -1823,4 +1824,129 @@ public function testNormalizeWithSubResource()
 
         $this->assertEquals($expected, $normalizer->normalize($documentation));
     }
+
+    /**
+     *
+     * @group legacy
+     * @expectedDeprecation Passing an instance of ApiPlatform\Core\Api\UrlGeneratorInterface to ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer::__construct() is deprecated since version 2.1 and will be removed in 3.0.
```
meyerbaptiste: Remove these annotations and fix the constructor instead.

tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -1823,4 +1824,129 @@ public function testNormalizeWithSubResource()
...
+     */
+    public function testNormalizeWithIdentifierName()
```
meyerbaptiste: Maybe add a test with composite identifiers?

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -321,6 +317,56 @@ private function updateGetOperation(\ArrayObject $pathOperation, array $mimeType
...
+        return $parameters;
+    }
+
+    /**
+     * Returns all the identifier properties for the given class
+     *
+     * @param string $resourceClass
+     * @return PropertyMetadata[]
+     */
+    private function getClassIdentifiers(string $resourceClass)
```
meyerbaptiste: Can you follow the same strategy as https://github.com/api-platform/core/pull/1358/files#diff-6ad90c31c0386b8ef6b541a28614c002?

Simperfit:
Could you please finish this ?

flug:
@Bwen  Up is this pull request still valid?

gphilippe:
No idea... lost interest, sorry. :cry: 

I guess not?

PR #1368, opened 2017-09-13
Merge branch '2.1'

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

PR #1369, opened 2017-09-13
Fix Travis when deps are low

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

> Class "Symfony\Bridge\PhpUnit\SymfonyTestsListener" does not exist

We obtain this message with `deps=low` when we run PHPUnit, but tests are not executed and the command exits with 0.

About the `weak_vendors`, see https://github.com/api-platform/core/pull/1368#issuecomment-329133122.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "symfony/finder": "^3.3 || ^4.0",
         "symfony/form": "^3.3 || ^4.0",
         "symfony/framework-bundle": "^3.3 || ^4.0",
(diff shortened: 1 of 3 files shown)


dunglas:
Shouldn't it be merged in 2.0?

dunglas:
Perfect then!

dunglas:
Thanks Baptiste!

PR #1383, opened 2017-09-19
operationName parameter should be optional in SubresourceDataProviders

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

operationName can be null in SubresourceDataProviders like the other DataProviders.

### DIFF
--- a/src/DataProvider/ChainSubresourceDataProvider.php
+++ b/src/DataProvider/ChainSubresourceDataProvider.php
@@ -35,7 +35,7 @@ final class ChainSubresourceDataProvider implements SubresourceDataProviderInter
     /**
      * {@inheritdoc}
      */
-    public function getSubresource(string $resourceClass, array $identifiers, array $context, string $operationName)
+    public function getSubresource(string $resourceClass, array $identifiers, array $context, string $operationName = null)
     {
         foreach ($this->dataProviders as $dataProviders) {
             try {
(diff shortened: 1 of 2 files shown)


soyuka:
Hmm nope I think that this is wrong in the `SubresourceDataProvider` instead. IIRC the `operationName`  was introduced as a fix and set to `nullable` to prevent a BC break. Because SubresourceDataProvider was released after, its signature could be with a required `operationName`.

I might be wrong though but I think that in the end this parameter will become mandatory.

dunglas:
I think it must be optional. What if you use those classes in a non-HTTP context? It's perfectly legit and in this case the operation will not be provided.

soyuka:
Because it's more then just representing the HTTP context, it's directly related to the resource operation (sure it's HTTP today). More specifically, it's being used to retrieve Operation attributes that are changing the behavior of the data providers.

On a second look however, this setting has always a fallback to the Resource it belongs and is always defaulted to `null`.

soyuka:
Btw, shouldn't this target `2.1`?

dunglas:
Yes, it should target 2.1. @alanpoulain can you change the target and rebase please?

soyuka:
Rebased, just waiting for the tests to pass.

soyuka:
Thanks @alanpoulain !

Issue #1079, opened 2017-04-24
Hypermedia controls at docs.jsonld 


cr3a7ure:
Parsing the docs.jsonld, which is the Hydra description for the API, there are no links/IRIs that can be resolved in order for an automated client to perform the described operations.

The IRIs used are either local or schema.org based if we use SchemaGenerator.
So there is no information for the client as to where he should GET/POST.

One idea to address this is to change the IRIs of classes to something dereferenceable inside the API,   `/thing` for example and specify `@type: schema:Thing` in order to describe the class.
`/thing` would be similar to `contexts/Thing`.

Another idea is the use of `hydra:Link` and `hydra:TemplatedLink` for collections and instances of a class as an `rdf:Property`. Hydra do specify that both `Link` and `TemplatedLink` have a subClassof `hydra:Resource, rdf:Property`. I'll try to implement it and see where it leads.

I was reading this example [Restbucks_with_Hydra](https://www.w3.org/community/hydra/wiki/Restbucks_with_Hydra), while trying to implement a graph based automated client.

What is your views on this?

dunglas:
Hi @cr3a7ure, I've the same kind of questions. In my opinion, the best idea is to ask the Hydra mailing list: http://www.hydra-cg.com/

cr3a7ure:
Well, it's a good thing that others share my concerns :)
I will try to resolve those things (hydra:search, etc) and I hope I'll comeback with a solution-interpretation.

Thanks Dunglas for your dedication!

cr3a7ure:
I've searched through the Hydra mailing list the past days and have to make some points we should consider. Hydra does not explicitly specify the structure of the document from what I've seen. There is a proposed way from some members, but the closest thing to validation is Hydra Console.

With this in mind, I propose some changes in the api-platform that might address some recurring problems. Some points are just keeping in pace with the spec.
About the others, I will try to implement some of these either way, but opinions are more than welcome in order to be accepted as a pull request in a later time.

1. Replace `hydra:CreateResourceOperation` to `hydra:Operation` .
Since every named operation has changed to the generic `hydra:Operation`.

2. Replace `hydra:PagedCollection` to `hydra:PartialCollectionView` .
`hydra:PagedCollection` is no longer used has changed to `hydra:PartialCollectionView`. I'll refer to it as `hydra:PagedCollection` for the rest of this post. 

3. Allow someone if he wants to express the search patterns inside the `docs.jsonld`.
[...]

PR #1396, opened 2017-10-02
Prop

cr3a7ure:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1079
| License       | MIT
| Doc PR        | 

I have added the ApiResource 'type' and ApiProperty 'vocabType'.
As I've mentioned in the #1079 , when parsing a json-ld graph, using the exact vocabulary terms as '@id' for classes and properties, creates some issues.
In my test scenario I am parsing my graphs and create a new one, combining many of those classes.
'type' and 'vocabType' are used to describe that this class is of type schema:Class, with an IRI - '@id' of 'localVocab:Class'
Currently those keywords are not used. I'll try to submit the new Hydra documentation normalizer without breaking anything.
This is my first PR, I tried to follow the rules.

@dunglas


.gitignore
```diff
@@ -6,3 +6,4 @@
 /vendor/
 /tests/Fixtures/app/cache/*
 /tests/Fixtures/app/logs/*
+composer.phar
```
sroze: Shouldn't be there but in your global gitignore 😉

src/Metadata/Property/PropertyMetadata.php
```diff
@@ -30,12 +30,13 @@
...
     private $required;
     private $iri;
+    private $vocabType;
     private $identifier;
     private $childInherited;
     private $attributes;
     private $subresource;
 
-    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null)
+    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, string $vocabType = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null)
```
sroze: Changing the argument order is a backward compatible break that we can't add without a new major version release. One of way adding arguments without BC-break is to add it at the end of the constructor, with a default `null` value.

src/Metadata/Resource/ResourceMetadata.php
```diff
@@ -23,15 +23,17 @@
     private $shortName;
     private $description;
     private $iri;
+    private $type;
     private $itemOperations;
     private $collectionOperations;
     private $attributes;
 
-    public function __construct(string $shortName = null, string $description = null, string $iri = null, array $itemOperations = null, array $collectionOperations = null, array $attributes = null)
+    public function __construct(string $shortName = null, string $description = null, string $iri = null, string $type = null, array $itemOperations = null, array $collectionOperations = null, array $attributes = null)
```
sroze: Same comment that for `PropertyMetadata`'s constructor.

src/Metadata/Resource/ResourceMetadata.php
```diff
@@ -113,6 +115,31 @@ public function withIri(string $iri): self
...
+    }
+
+    /**
+     * Returns a new instance with the given type.
+     *
+     * @param string $type
+     *
+     * @return self
+     */
+    public function withType(string $type): self
```
sroze: Can't this be `null` as well?

sroze:
Thanks for your PR @cr3a7ure. What do you mean by this?
> I'll try to submit the new Hydra documentation normalizer without breaking anything.

Also, could you try to find a better name for your PR and maybe squashing your commits with a name that would helps us to understand what you've tried to do?

Simperfit:
@cr3a7ure Are you going forward with this ?

cr3a7ure:
@Simperfit I have stopped all development since November.
My test case seems live at http://149.202.40.241/schydra/playground.html

soyuka:
closing as old, lmk if I need to reopen

Issue #1394, opened 2017-09-25
Error at the installation : non-existent service api_platform.http_cache.purger


pierre-H:
Hello,

When I try to install a new version of AP 2.1.1, I've got this error : 

> [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
>   The service "api_platform.doctrine.listener.http_cache.purge" has a dependency on a non-existent service "api_platform.http_cache.purger".

In `src\Bridge\Symfony\Bundle\Resources\config` I've got only `api_platform.http_cache.purger.varnish_client` and `api_platform.http_cache.purger.varnish`.

Note : I don't want to use varnish so in parameters, I have `varnish_urls: []`

Thanks for your work !

dunglas:
Can you try to just delete those lines https://github.com/api-platform/api-platform/blob/master/app/config/config.yml#L86-L88 ?

pierre-H:
@dunglas thank you : now there is no error

PR #1400, opened 2017-10-07
Fix missing service when no Varnish URL is defined

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1394
| License       | MIT
| Doc PR        | -

When HTTP Cache invalidation is enabled, the `api_platform.doctrine.listener.http_cache.purge` service is created. This service depends on `api_platform.http_cache.purger` which is an alias that is only defined when at least one Varnish URL is defined.

With this PR, the alias is defined even if no Varnish URL is defined, which fixes the dependency issue.


sroze:
Thanks @julienfalque!

PR #1405, opened 2017-10-09
Provide a better exception message

greg0ire:
Getting "Not found" on a route where you are getting an object can get
real confusing.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/features/security/validate_response_types.feature
+++ b/features/security/validate_response_types.feature
@@ -28,11 +28,11 @@ Feature: Validate response types
     And I send a "GET" request to "/dummies/1.invalid"
     Then the response status code should be 404
     And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
-    And the JSON node "detail" should be equal to "Not Found"
+    And the JSON node "detail" should be equal to 'Format "invalid" is not supported'
 
   Scenario: Requesting an invalid format in the Accept header and in the URL should throw an error
     When I add "Accept" header equal to "text/invalid"
     And I send a "GET" request to "/dummies/1.invalid"
     Then the response status code should be 404
     And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
-    And the JSON node "detail" should be equal to "Not Found"
+    And the JSON node "detail" should be equal to 'Format "invalid" is not supported'
--- a/src/EventListener/AddFormatListener.php
+++ b/src/EventListener/AddFormatListener.php
@@ -58,7 +58,7 @@ final class AddFormatListener
         if (null === $routeFormat = $request->attributes->get('_format') ?: null) {
             $mimeTypes = array_keys($this->mimeTypes);
         } elseif (!isset($this->formats[$routeFormat])) {
-            throw new NotFoundHttpException('Not Found');
+            throw new NotFoundHttpException(sprintf('Format "%s" is not supported', $routeFormat));
         } else {
             $mimeTypes = Request::getMimeTypes($routeFormat);
         }
--- a/tests/EventListener/AddFormatListenerTest.php
+++ b/tests/EventListener/AddFormatListenerTest.php
@@ -171,7 +171,7 @@ class AddFormatListenerTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @expectedException \Symfony\Component\HttpKernel\Exception\NotFoundHttpException
-     * @expectedExceptionMessage Not Found
+     * @expectedExceptionMessage Format "invalid" is not supported
      */
     public function testInvalidRouteFormat()
     {


julienfalque:
Should't the response status be `406 Not Acceptable`?

src/EventListener/AddFormatListener.php
```diff
@@ -58,7 +58,7 @@ public function onKernelRequest(GetResponseEvent $event)
         if (null === $routeFormat = $request->attributes->get('_format') ?: null) {
             $mimeTypes = array_keys($this->mimeTypes);
         } elseif (!isset($this->formats[$routeFormat])) {
-            throw new NotFoundHttpException('Not Found');
+            throw new NotFoundHttpException(sprintf('Format "%s" is not supported', $routeFormat));
```
sroze: I'd say "415 Unsupported Media Type" or "406 Not Acceptable" but definitely not 404 👍

greg0ire:
Yeah I should have seen past the error message, good catch!

src/EventListener/AddFormatListener.php
greg0ire: https://httpstatuses.com/415 is about the payload of the request, so let's go with https://httpstatuses.com/406, which looks like a perfect fit for this

features/security/validate_response_types.feature
```diff
@@ -28,11 +28,11 @@ Feature: Validate response types
     And I send a "GET" request to "/dummies/1.invalid"
     Then the response status code should be 404
     And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
```
dunglas: Should be 406 right (same below)

features/security/validate_response_types.feature
```diff
@@ -28,11 +28,11 @@ Feature: Validate response types
     And I send a "GET" request to "/dummies/1.invalid"
     Then the response status code should be 404
```
dunglas: Should be 406 right? (same below)

features/security/validate_response_types.feature
```diff
@@ -26,13 +26,13 @@ Feature: Validate response types
 
   Scenario: Requesting an invalid format in the URL should throw an error
     And I send a "GET" request to "/dummies/1.invalid"
-    Then the response status code should be 404
+    Then the response status code should be 406
```
julienfalque: I think status 406 only applies to content negotiation using `Accept*` headers. This one should remain a 404 IMO.

features/security/validate_response_types.feature
```diff
@@ -26,13 +26,13 @@ Feature: Validate response types
...
+    And the JSON node "detail" should be equal to 'Format "invalid" is not supported'
 
   Scenario: Requesting an invalid format in the Accept header and in the URL should throw an error
     When I add "Accept" header equal to "text/invalid"
     And I send a "GET" request to "/dummies/1.invalid"
-    Then the response status code should be 404
+    Then the response status code should be 406
     And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
-    And the JSON node "detail" should be equal to "Not Found"
+    And the JSON node "detail" should be equal to 'Format "invalid" is not supported'
```
julienfalque: Isn't it possible to use the exact value of the `Accept` header, i.e. `text/invalid`?

greg0ire:
Travis is slow...

features/security/validate_response_types.feature
greg0ire: I think you're right, and this is why header fields should be preferred over extensions

dunglas: The extension support is just intended for debug/send in Slack/etc. Not for production usage.

greg0ire: 💡 good to know

greg0ire:
@julienfalque I think I did it exactly right by chance the first time: the last scenario is about a wrong format in the accept header AND in the url.

features/security/validate_response_types.feature
greg0ire: nope after all

src/EventListener/AddFormatListener.php
greg0ire: @sroze in fact, 404 was right, and extensions shouldn't be used (in production, in development it's ok although I'm starting to think this causes more issues than it solves).

greg0ire:
@sroze @julienfalque please review again

julienfalque:
Indeed, I missed the URL part in the second scenario 👍

PR #1407, opened 2017-10-10
Replace the "object_to_populate" context key by the Symfony constant.

tuanphpvn:
Use const instead string

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/EventListener/DeserializeListener.php
+++ b/src/EventListener/DeserializeListener.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\EventListener;
 
-use ApiPlatform\Core\Serializer\AbstractItemNormalizer;
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
 use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpFoundation\Request;
@@ -62,7 +62,7 @@ final class DeserializeListener
 
         $data = $request->attributes->get('data');
         if (null !== $data) {
-            $context[AbstractItemNormalizer::OBJECT_TO_POPULATE] = $data;
+            $context[AbstractNormalizer::OBJECT_TO_POPULATE] = $data;
         }
 
         $request->attributes->set(
--- a/src/EventListener/DeserializeListener.php
+++ b/src/EventListener/DeserializeListener.php
@@ -13,12 +13,12 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\EventListener;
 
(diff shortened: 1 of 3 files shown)


src/Serializer/ItemNormalizer.php
```diff
@@ -30,7 +30,7 @@
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['id']) && !isset($context['object_to_populate'])) {
+        if (isset($data['id']) && !isset($context[static::OBJECT_TO_POPULATE])) {
```
sroze: You need to define this constant then? I'd use `self` and not `static` as this class is `final`. Also, it might be worth updating the other usages as well.

tuanphpvn: Thank you for suggesting that. I know that OBJECT_TO_POPULATE already define in ```Symfony\Component\Serializer\Normalizer\AbstractNormalizer```. Also static better than self. And Yes, I will edit other place as well.

tuanphpvn: @sroze. Is there anything wrong? Please let me know. Thanks

sroze: Oh good point, it's in Symfony's class, sorry! 👍 I'm still in favour of `self` as `static` is to allow child classes.

tuanphpvn: Please review this class: ```Symfony\Component\Serializer\NormalizerAbstractNormalizer```. Symfony use static event in the same class.

meyerbaptiste: They use `static` because the class is abstract (extensible) but ours is final (not extensible) so you should use `self` instead.

tuanphpvn: Ok understood, thanks for explanation.

src/EventListener/DeserializeListener.php
```diff
@@ -61,7 +62,7 @@ public function onKernelRequest(GetResponseEvent $event)
 
         $data = $request->attributes->get('data');
         if (null !== $data) {
-            $context['object_to_populate'] = $data;
+            $context[AbstractItemNormalizer::OBJECT_TO_POPULATE] = $data;
```
meyerbaptiste: Use `use Symfony\Component\Serializer\Normalizer\AbstractNormalizer::OBJECT_TO_POPULATE` instead.

tuanphpvn: Thanks. Done

soyuka:
Shouldn't this be on 2.1?

tuanphpvn:
Hi @meyerbaptiste . I thinks I will close this PR since It is not important for now, will focus on features instead.

PR #1411, opened 2017-10-11
Add support update embed relation (many relationship) with simple object

tuanphpvn:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

### DIFF
--- a/src/Metadata/Property/PropertyNameCollection.php
+++ b/src/Metadata/Property/PropertyNameCollection.php
@@ -48,4 +48,13 @@ final class PropertyNameCollection implements \IteratorAggregate, \Countable
     {
         return count($this->properties);
     }
+
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getArrProperty()
+    {
+        return $this->properties;
+    }
 }
--- a/src/Metadata/Property/PropertyNameCollection.php
+++ b/src/Metadata/Property/PropertyNameCollection.php
@@ -49,7 +49,6 @@ final class PropertyNameCollection implements \IteratorAggregate, \Countable
         return count($this->properties);
     }
 
-
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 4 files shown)


src/Metadata/Property/PropertyNameCollection.php
```diff
@@ -48,4 +48,12 @@ public function count()
     {
         return count($this->properties);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getArrProperty()
```
soyuka: what does this mean?

tuanphpvn: Never mind. Removed.

tuanphpvn:
Just want to support an simple case:

For example, Update person entity.

with request data change from

```
{
  "name": "Update person name",
  "books": [{
    "id": "existing id"
    "name": "new book name"
   }
  ]
}
```

to

```
{
  "name": "Update person name",
  "books": {
    "id": "existing id"
    "name": "new book name"
   }
}
```

soyuka:
I find this wrong, really not sure that we should accept such request.

tuanphpvn:
Yes, I know. By example above books is array (many relationship). So this implement just give a little more flexible. 

And about definition.

I think: 

1) Array have many item => we should implement array.
2) Array just have one item => we should allow flexible change between array and item. Like

```
books: [{
    "id": "1",
    "name: "name"
}]
```

should equal to

```
books: {
    "id": "1",
    "name: "name"
}
```

If you don't like it. Just close this request. Thank you.

soyuka:
Feel free to add your own normalizer on top of our to handle your custom case. We can not accept this pull request because of the above comments but thank your for your interest in Api Platform.

PR #1412, opened 2017-10-11
Reuse PriorityTaggedServiceTrait from symfony

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

Only difference will be now that it will throw if service will be abstract, but I think this should have done already before.


norkunas:
Oh.. dependency injection component is soft dependency :/

meyerbaptiste:
But we can update our method with the one provided by the Symfony trait...

norkunas:
If you want I can submit PR for that

meyerbaptiste:
Definitively @norkunas!

sroze:
When using this compiler pass we know what the DependencyInjection component is here tho... so we could use this trait.

PR #1413, opened 2017-10-11
Align findSortedServices code with symfony PriorityTaggedServiceTrait

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1412
| License       | MIT
| Doc PR        | none


meyerbaptiste:
IMO you can directly copy/paste the trait from Symfony. IIRC we use sorting in the other compilation passes.

sroze:
When using this compiler pass we know what the DependencyInjection component is here tho... so we could use this trait.

(beautiful copy/paste on my comment on the issue)

meyerbaptiste:
But we have to bump our minimum requirements @sroze.

norkunas:
I didn't find any other occurences for sorting services

sroze:
This trait was introduced in Symfony 3.2. And we already have the following in our `composer.json`:
```
    "conflict": {
        "symfony/dependency-injection": "<3.3"
    },
```

dunglas:
@sroze is right, we should just use the trait!

meyerbaptiste:
I'm agree, but tests will fail with `deps=low`!

dunglas:
Indeed in 2.0 (which is only maintained for bug and security fix - and it's not the case here) https://github.com/api-platform/core/blob/2.0/composer.json#L47
But we can use the trait in 2.1: https://github.com/api-platform/core/blob/2.1/composer.json#L52

So I propose to use the trait and rebase this PR against 2.1.

sroze:
Sounds good to me 👍

meyerbaptiste:
I didn't pay attention to the base branch! Let's do that!

PR #1420, opened 2017-10-12
Add doctrine/common to "require" block.

tuanphpvn:
Since ApiPlatform\Core\Metadata\Resource\Factory\AnnotationResourceMetadataFactory depend on Doctrine\Common.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     "require": {
         "php": ">=7.0",
 
+        "doctrine/common": "~2.8.1",
         "doctrine/inflector": "^1.0",
         "psr/cache": "^1.0",
         "psr/container": "^1.0",


dunglas:
This dependency is optional (it's not mandatory to use annotations).

tuanphpvn:
There are two things:

1. "it's not mandatory to use annotations" => Do you mean we can use xml or yaml? If so, since api_platform also support annotation. So doctrine/common is required.

2. "it's not mandatory to use annotations" => Just use annotation when use with symfony framework. So i think must have a directory separate in ```src/Metadata/Resource/Factory``` directory. Some kind like: 
src/Metadata/Symfony/Resource/Factory or src/Metadata/Zend/Resource/Factory. Correct if me wrong. Thanks.

dunglas:
It's 1/ (you can use XML or YAML). And no it's not required, for instance: https://github.com/symfony/serializer/blob/master/composer.json#L28

dunglas:
Check https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L190-L192

tuanphpvn:
Using class but require register from outside package. Cool way!

Issue #1422, opened 2017-10-12
CollectionDataProviderInterface does not play well with generators


greg0ire:
I tried implementing it with generators, and it worked just fine... until... I implemented a second.

The problem is here:

https://github.com/api-platform/core/blob/020dbf21246bc7acbd8c417e0ab64bddafcadf3b/src/DataProvider/ChainCollectionDataProvider.php#L38-L48

Here is what my implementation looks like:

```php
    public function getCollection(string $resourceClass, string $operationName = null)
    {
        if ($resourceClass !== Person::class) {
            throw new ResourceClassNotSupportedException();
        }

       yield $aPerson;
```

The ChainCollectionDataProvider will loop over my providers, and will immediatly return the generator returned by the first data provider because no code is executed until the generator is used inside a foreach, which happens later in the code, outside the try / catch block.

The solution would be to introduce a `supports` method in the interface.

meyerbaptiste:
Fixed by #1429.

PR #1429, opened 2017-10-12
Generator compat improvements

greg0ire:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1422 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/321

This PR should enable people to use generators when implementing `DataCollectionProviderInterface`

# TODO
- [x] the same with the item interface, although it's not really needed?
- [x] implement the new interface everywhere we can
- [x] the doc PR


src/DataProvider/ChainCollectionDataProvider.php
```diff
@@ -39,8 +40,14 @@ public function getCollection(string $resourceClass, string $operationName = nul
...
         foreach ($this->dataProviders as $dataProvider) {
             try {
+                if ($dataProvider instanceof RestrictedDataProviderInterface
+                    && !$dataProvider->supports($resourceClass, $operationName)) {
+                    continue;
+                }
+
                 return $dataProvider->getCollection($resourceClass, $operationName);
             } catch (ResourceClassNotSupportedException $e) {
+                @trigger_error(sprintf('Throwing a "%s" is deprecated in favor of implementing "%s"', get_class($e), RestrictedDataProviderInterface::class), E_USER_DEPRECATED);
```
greg0ire: Following the sf style here... yuck!

greg0ire:
TODO: the same with the item interface, although it's not really needed?

dunglas:
Looks good to me.

> the same with the item interface, although it's not really needed?

We should do it at least for consistency.

greg0ire:
@dunglas done, but it's no longer possible not to support something based on `$id` (no big deal IMO), or `$context` (no idea what this is about).

dunglas:
Shouldn't the builtin Doctrine providers implement this new interface?

greg0ire:
@dunglas they should, and the build should probably not pass because of the deprecations, should it? something seems to be wrong...

greg0ire:
Oh wait no it's normal, they are not called when testing the chain providers

src/DataProvider/ChainCollectionDataProvider.php
soyuka: Why not `ResourceClassNotSupportedException::class` constant? Also, could you mention "Throwing a "..." **in a data provider** is deprecated..."?

tests/Bridge/Doctrine/Orm/CollectionDataProviderTest.php
```diff
@@ -97,17 +97,14 @@ public function testCannotCreateQueryBuilder()
         $this->assertEquals([], $dataProvider->getCollection(Dummy::class, 'foo'));
     }
 
-    /**
-     * @expectedException \ApiPlatform\Core\Exception\ResourceClassNotSupportedException
-     */
-    public function testThrowResourceClassNotSupportedException()
```
soyuka: Can we keep this test anyway to ensure backward compatibility?

soyuka: Nevermind I saw it in the chain!

soyuka:
It's not really a bug fix or is it? I think it should target master. Anyway it's a nice improvement!

greg0ire:
I should probably call `supports` inside the other methods and throw if it returns false, shouldn't I?

soyuka:
What "other methods" do you had in mind? The chain provider continues when an exception is catched so this looks fine!

greg0ire:
@soyuka `getCollection` and `getItem`... just in case someone decides to use this from their own code.

soyuka:
Nope, the service exposed is the ChainDataProvider:

```
        <service id="api_platform.item_data_provider" class="ApiPlatform\Core\DataProvider\ChainItemDataProvider" />
```

greg0ire:
Doesn't prevent people from just instanciating the class and misusing it, but I'm not sure you require that level of defensiveness

soyuka:
Maybe we can consider flagging the doctrine data providers as Internal then. Also, they may be final (I see no reason to extend them). In the docs the recommended way is to implement the DataProviderInterface.

greg0ire:
Final would be a BC break though

greg0ire:
@sroze @julienfalque @soyuka can we have your stance on this? We were wondering if there wouldn't be a better way to add support for generators, without deprecating anything / introducing a new interface. Technically, right now, you can achieve that by writing something like this:

```php
<?php
public function getCollection()
{
    // throw exception here;
    return $this->getGenerator();
}

private function getGenerator();
{
    yield new Item;
}
```

But this feels hackish to me, I don't think we should recommend that.

soyuka:
I like the clean way you added in this PR and I won't mind a deprecation on the subject especially if we remove an exception (try/catching is slow)

dunglas:
ping @api-platform/core-team

soyuka:
@dunglas do you consider this as a bug fix rather then a new feature?

dunglas:
It introduces deprecation so it should be merged in master.

greg0ire:
@dunglas rebased

greg0ire:
As well as the docs PR

src/DataProvider/ChainCollectionDataProvider.php
dunglas: Can you just make this change and we'll merge.

greg0ire: Oh sure I forgot. Let's see what those constants are

greg0ire: Oh right I understand now that I see the code in my editor, I just put the wrong class constant. I could have understood this earlier had it been legible on github :P
Fun fact: this line does not even fit on a 1920x1200 display.

greg0ire: Fixed @dunglas

soyuka:
Tests fails though:

```
-%A  Throwing a "ApiPlatform\Core\Exception\ResourceClassNotSupportedException" is deprecated in favor of implementing "ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface"
+  Throwing a "ApiPlatform\Core\Exception\ResourceClassNotSupportedException" in a data provider is deprecated in favor of implementing "ApiPlatform\Core\Exception\ResourceClassNotSupportedException"
```

src/DataProvider/ChainCollectionDataProvider.php
```diff
@@ -39,8 +39,14 @@ public function getCollection(string $resourceClass, string $operationName = nul
...
         foreach ($this->dataProviders as $dataProvider) {
             try {
+                if ($dataProvider instanceof RestrictedDataProviderInterface
+                    && !$dataProvider->supports($resourceClass, $operationName)) {
+                    continue;
+                }
+
                 return $dataProvider->getCollection($resourceClass, $operationName);
             } catch (ResourceClassNotSupportedException $e) {
+                @trigger_error(sprintf('Throwing a "%s" in a data provider is deprecated in favor of implementing "%s"', ResourceClassNotSupportedException::class, RestrictedDataProviderInterface::class), E_USER_DEPRECATED);
```
greg0ire: @soyuka I misunderstood your comment at first. I hope you all see how easier it would have been had each argument been on its own line: you could have commented directly under `get_class($e)`. BTW, the sf standard is to have the message on one line, but is the `sprintf` part of the message? If yes, that's really a stupid rule who just made us loose precious time. It just makes life harder for everyone for no good reason.

meyerbaptiste:
Thanks @greg0ire!

PR #1430, opened 2017-10-12
Add job to test upstream libs deprecations

greg0ire:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | hopefully
| Fixed tickets | none
| License       | MIT
| Doc PR        | n / a

If api-platform uses upstream libs in a deprecated way, we should be aware
of it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ matrix:
       env: coverage=1 lint=1
     - php: '7.1'
       env: deps='low'
+    - php: '7.1'
+      env: SYMFONY_DEPRECATIONS_HELPER=0
+  allow_failures:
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks @greg0ire!

meyerbaptiste:
I think this will not work because of this line: https://github.com/api-platform/core/blob/863127226ece5b03e7df0f4dfac1a3bd2acb668e/phpunit.xml.dist#L11
IIRC this line overwrites any environment variable...

greg0ire:
Oh sorry I'll try to fix this tomorrow unless you fix it before

PR #1432, opened 2017-10-12
Fix missing cache tag on empty collections

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

When a collection is empty, the cache tag of the collection itself is missing from the response, which cannot be invalidated later when the collection changes.


meyerbaptiste:
Thx @julienfalque!

PR #1436, opened 2017-10-13
fix hydra documentation normalizer with subresources

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

Fixes the hydra documentation normalization by using the `subresourceOperationFactory`

### DIFF
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -25,6 +25,7 @@ use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Metadata\Property\SubresourceMetadata;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Operation\Factory\SubresourceOperationFactoryInterface;
 use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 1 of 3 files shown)


src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -217,17 +220,12 @@ private function getHydraOperations(string $resourceClass, ResourceMetadata $res
...
+            foreach ($this->subresourceOperationFactory->create($resourceClass) as $operationId => $operation) {
+                $subresourceMetadata = $this->resourceMetadataFactory->create($operation['resource_class']);
+                $propertyMetadata = $this->propertyMetadataFactory->create(end($operation['identifiers'])[1], $operation['property']);
+                $hydraOperations[] = $this->getHydraOperation($resourceClass, $subresourceMetadata, $operation['route_name'], $operation, "#{$subresourceMetadata->getShortName()}", OperationType::SUBRESOURCE, $propertyMetadata->getSubresource());
             }
-
-            $subresourceMetadata = $this->resourceMetadataFactory->create($propertyMetadata->getSubresource()->getResourceClass());
-            $prefixedShortName = "#{$subresourceMetadata->getShortName()}";
-
-            $hydraOperations[] = $this->getHydraOperation($resourceClass, $subresourceMetadata, $operationName, $operation, $prefixedShortName, OperationType::SUBRESOURCE, $propertyMetadata->getSubresource());
```
dunglas: As `$operationName` and `$operation` aren't used for subresources, can't we just made them nullable in the `private` method instead of introducing a new dependency?

soyuka: They're used (at least `operationName` is).  IMO this is better because if we introduce other methods on subresources for example, we will use this dependency.

Whay I may do is make them nullable in `2.1` and add this dependency on `master`, wdyt?

dunglas:
Thanks @soyuka!

Issue #1404, opened 2017-10-09
Why no cache on API doc Hydra generation ?


fmata:
Hi,

In order to speed up our app which uses api-platform intensively, we try to cache everything and we use HTTP cache with Varnish.

But one question : why the API doc Hydra is not cached ? In our app it takes each times ~1s to generate in jsonld.

Could you accept a PR which adds a cache for API doc Hydra and an invalidation when the metadata are updated ?

dunglas:
Yes it's a miss, the Hydra endpoint should respect values set in the `cache` config section.

fmata:
This endpoint does not depend on a resource but on all definitions/mappings information.

Do you have an insight where I can plug the invalidation ?

Maybe in a cache warmer, WDYT ?

sroze:
I'd completely agree this should be cached. As you mentioned the invalidation is here a tricky point... should a cache warmer talk to Varnish? Not sure about this one 🤔 

Maybe just following the `cache` configuration is enough, and invalidation of this documentation is down to the user.

fmata:
@dunglas @sroze see PR #1437

PR #1437, opened 2017-10-14
add cache to Hydra endpoints

fmata:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes but not risky IMHO
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1404 
| License       | MIT

This PR adds cache management to all Hydra endpoints (doc and all contexts), according `http_cache` config.

Invalidation is done while cache clearing (`kernel.cache_clearer`).

I hope this can be merged in 2.1 despite the BC in `AddTagsListener` (2 more services are injected in the constructor) and 2 methods added in `IriConverterInterface`.


dunglas:
I think that it's not a good idea to use cache tags for docs. Docs don't change (except during deploy), a traditional Cache-Control header (expiration strategy) should be enough. And it will be way easier to implement (without BC breaks).

Simperfit:
@fmata do you think you can modify the PR according to @dunglas's suggestion or do you want to maybe explain why not ?

fmata:
@Simperfit I want do it but I don't know how to compute an expiration time for `Expires` header.

Only `Cache-Control` is not enough because the invalidation will never happen. May be can you give me an hint ?

flug:
@fmata Up is this pull request still valid?

fmata:
Not by now but if someone gives me a hint I can pursue.

PR #1450, opened 2017-10-20
add aliases for services

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | api-platform/api-platform#429
| License       | MIT
| Doc PR        | 

Maybe we should add other services too ?

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -61,6 +61,7 @@
         <service id="api_platform.serializer.context_builder" class="ApiPlatform\Core\Serializer\SerializerContextBuilder" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
         </service>
+        <service id="ApiPlatform\Core\Serializer\SerializerContextBuilder" alias="api_platform.serializer.context_builder.filter.inner"/>
 
         <service id="api_platform.serializer.context_builder.filter" class="ApiPlatform\Core\Serializer\SerializerFilterContextBuilder" decorates="api_platform.serializer.context_builder" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -61,6 +61,7 @@
         <service id="api_platform.serializer.context_builder" class="ApiPlatform\Core\Serializer\SerializerContextBuilder" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
         </service>
+        <service id="ApiPlatform\Core\Serializer\SerializerContextBuilder" alias="api_platform.serializer.context_builder.filter.inner"/>
```
dunglas: Shouldn't the alias name been the interface?

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -142,6 +143,7 @@
 
             <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="2" />
         </service>
+        <service id="ApiPlatform\Core\EventListener\DeserializeListener" alias="api_platform.listener.request.deserialize"/>
```
dunglas: When can it be necessary to inject this class?

src/Bridge/Symfony/Bundle/Resources/config/api.xml
sroze: 👍

Simperfit: Yeah you are right

src/Bridge/Symfony/Bundle/Resources/config/api.xml
Simperfit: It does trigger a notice, So we should look

norkunas:
ApiPlatform\Core\Api\IriConverterInterface also should be aliased? I use it in custom normalizer

soyuka:
I think that this has been implemented by #1516, feel free to reopen @Simperfit

PR #1453, opened 2017-10-22
[SF 4.0] Make actions explicitly public

GuilhemN:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

As of sf 4.0, controllers must be explicitly made public.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -168,19 +168,19 @@
 
         <!-- Action -->
 
-        <service id="api_platform.action.placeholder" class="ApiPlatform\Core\Action\PlaceholderAction" />
-        <service id="api_platform.action.get_collection" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.post_collection" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.get_item" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.put_item" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.delete_item" alias="api_platform.action.placeholder" />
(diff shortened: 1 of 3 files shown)


dunglas:
Can you target 2.1 as it's a bug fix?

GuilhemN:
@dunglas done ;)

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -168,27 +168,27 @@
...
-        <service id="api_platform.action.get_collection" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.post_collection" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.get_item" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.put_item" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.delete_item" alias="api_platform.action.placeholder" />
-        <service id="api_platform.action.get_subresource" alias="api_platform.action.placeholder" />
-
-        <service id="api_platform.action.entrypoint" class="ApiPlatform\Core\Action\EntrypointAction">
+        <service id="api_platform.action.placeholder" class="ApiPlatform\Core\Action\PlaceholderAction" public="true" />
+        <service id="api_platform.action.get_collection" alias="api_platform.action.placeholder" public="true" />
```
sroze: Do we really need these aliases to be public? 🤔

GuilhemN: If they're used in routing, yes. At least they were public until now so I think they should still be when using sf 4.0.

dunglas:
Thanks @GuilhemN

PR #1456, opened 2017-10-23
Fix request auto-runner

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

This feature was broken since an update of SwaggerUI... Please be careful when updating Swagger UI.


soyuka:
> Please be careful when updating Swagger UI.

Meaning? Who updated/merged it? How can we avoid this?

meyerbaptiste:
> Who updated/merged it?

Me: https://github.com/api-platform/core/pull/1330 😃

dunglas:
@soyuka I checked with Baptiste (it was him, not a big deal). We could introduce a functional test but it will require to add a WebDriver implementation... And our build will be still slower.

Instead, I think to propose to the Swagger UI team to add this feature upstream.

dunglas:
During the meantime, the solution is to test manually when runing `./update-js.sh`

PR #1460, opened 2017-10-25
[WIP] GraphQL Mutation support

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

TODO:
- [x] delete mutation (item)
- [x] put mutation (item)
- [x] post mutation (collection)
- [ ] custom mutation (item / collection)

### DIFF
--- a/src/Bridge/Graphql/Resolver/ItemMutationResolverFactory.php
+++ b/src/Bridge/Graphql/Resolver/ItemMutationResolverFactory.php
@@ -48,39 +48,18 @@ final class ItemMutationResolverFactory implements ItemMutationResolverFactoryIn
     public function createItemMutationResolver(string $resourceClass, string $mutationName): callable
     {
         return function ($root, $args, $context, ResolveInfo $info) use ($resourceClass, $mutationName) {
-            $identifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($resourceClass);
-            if (\count($identifiers) > 1) {
-                $identifierPairs = \array_map(function ($identifier) use ($args, $info) {
-                    if (\is_array($args['input'][$identifier])) {
-                        if (\count($args['input'][$identifier]) > 1) {
-                            throw Error::createLocatedError('Composite identifiers are not allowed for a resource already used as a composite identifier', $info->fieldNodes, $info->path);
-                        }
-
-                        return $identifier.'='.\reset($args['input'][$identifier]);
-                    }
-
-                    return "{$identifier}={$args['input'][$identifier]}";
-                }, $identifiers);
-                $id = \implode(';', $identifierPairs);
-            } else {
-                $id = $args['input'][$identifiers[0]];
-            }
+            $id = $this->getIdentifier($this->identifiersExtractor->getIdentifiersFromResourceClass($resourceClass), $args, $info);
+            $item = null;
 
             if ('update' === $mutationName || 'delete' === $mutationName) {
-                $item = $this->itemDataProvider->getItem($resourceClass, $id);
-                if (null === $item) {
-                    throw Error::createLocatedError("Item $resourceClass $id not found", $info->fieldNodes, $info->path);
-                }
+                $item = $this->getItem($resourceClass, $id, $info);
             }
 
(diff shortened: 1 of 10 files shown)


src/Bridge/Graphql/Resolver/ItemMutationResolverFactory.php
```diff
@@ -0,0 +1,76 @@
...
+                }, $identifiers);
+                $id = implode(';', $identifierPairs);
+            } else {
+                $id = $args['input'][$identifiers[0]];
+            }
+            $uri = $this->urlGenerator->generate($route, ['id' => $id]);
+            $request = Request::create($uri, $operation['method']);
+            $request->headers->set('Accept', 'application/json');
+
+            $response = $this->httpKernel->handle($request, HttpKernelInterface::SUB_REQUEST);
```
dunglas: This is weird but I get it, we don't actually have anything to handle object deletion, we just call `remove()` from Doctrine in the Doctrine's `WriteListener`: https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/EventListener/WriteListener.php#L62

It was OK for REST until now but it doesn't scale to GraphQL. Injecting `HttpKernel` here is bad design.

I suggest to do the following:

* Introduce a new `ObjectManagerInterface` in API Platform with `delete()` and a `persist()` method
* Implement this interface in the Doctrine bridge to delegate to Doctrine's Object Manager (it will just be a proxy class)
* Use this new interface here and in `WriteListener`

WDYT?

src/Bridge/Graphql/Type/SchemaBuilder.php
```diff
@@ -335,6 +373,10 @@ private function getOperations(ResourceMetadata $resourceMetadata, bool $isQuery
                 continue;
             }
 
+            if (Request::METHOD_PATCH === $operation['method']) {
```
dunglas: You should probably whitelist methods here instead of blacklisting them, some other can be problematic (HEAD, OPTIONS...)

src/Bridge/Graphql/Resolver/ItemMutationResolverFactory.php
alanpoulain: Ok so you want to modify API Platform to avoid this issue then. Your suggestion seems good to me.

dunglas: Yes, it's a tiny, non-breaking change IMO.

src/Bridge/Graphql/Type/SchemaBuilder.php
alanpoulain: Sure.

src/Bridge/Graphql/Resolver/ItemMutationResolverFactory.php
dunglas: The existing `Bridge\Doctrine\EventListener\WriteListener` should be deprecated (and not registered anymore) and a new one (`EventListener\WriteListener`) using the new interface should be added.

sroze: Can you please do this in another PR?

dunglas: Why not but this other PR must be merged before this one :)

sroze: Obsly :)

dunglas:
Can we merge this one?

alanpoulain:
I think @meyerbaptiste was working on a better way to do this, doesn't he?

dunglas:
It's merged #1464, but it's lower level system. This PR can now be updated to use this new service.

alanpoulain:
OK maybe I'll do that this weekend.

dunglas:
Thanks @alanpoulain! This is still a WIP, but it's a big first step!

PR #1464, opened 2017-10-27
Add a data persistence layer

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/1460#discussion_r146764601
| License       | MIT
| Doc PR        | N/A


src/Bridge/Doctrine/Common/DataPersister.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $this->managerRegistry = $managerRegistry;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports($data): bool
+    {
+        return (bool) $this->getManager($data);
```
dunglas: Wouldn't `null !== $this->getManager($data)` be less weird?

src/Bridge/Doctrine/EventListener/WriteListener.php
```diff
@@ -1,87 +0,0 @@
-<?php
```
dunglas: We cannot drop it right now (BC promise), this class must be deprecated instead.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataPersisterPass.php
```diff
@@ -0,0 +1,43 @@
...
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Registers data persisters.
+ *
+ * @internal
```
dunglas: We'll be able to drop this class when we'll upgrade to DI 3.4: http://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services

src/Bridge/Doctrine/EventListener/WriteListener.php
meyerbaptiste: So should we also keep the definition of the tagged service and depreciate it?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DataPersisterPass.php
meyerbaptiste: Yep, I saw that. Do you suggest to delete it?

src/Bridge/Doctrine/EventListener/WriteListener.php
dunglas: I'm not sure about that because it's a public service so it will always be executed.

meyerbaptiste:
Done @dunglas

dunglas:
Thanks @meyerbaptiste

dunglas:
Don't forget the docs PR :D

PR #1433, opened 2017-10-13
Add feature to update swagger context for properties

mab05k:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | api-platform/docs#315

PR #1466, opened 2017-10-30
Include swagger context even when type is null

greg0ire:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | n/a

I do not know when the type is supposed to be null, but I am using the `YamlExtractor` and it always seems to be the case with that metadata specification method.

Refs #1433

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -508,7 +508,7 @@ final class DocumentationNormalizer implements NormalizerInterface
      */
     private function getPropertySchema(PropertyMetadata $propertyMetadata, \ArrayObject $definitions, array $serializerContext = null): \ArrayObject
     {
-        $propertySchema = new \ArrayObject();
+        $propertySchema = new \ArrayObject($propertyMetadata->getAttributes()['swagger_context'] ?? []);
 
         if (false === $propertyMetadata->isWritable()) {
             $propertySchema['readOnly'] = true;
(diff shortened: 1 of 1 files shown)


greg0ire:
Using a phpdoc comment on the property is another way to fix this.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -518,8 +518,9 @@ private function getPropertySchema(PropertyMetadata $propertyMetadata, \ArrayObj
             $propertySchema['description'] = $description;
         }
 
+        $swaggerContext = $propertyMetadata->getAttributes()['swagger_context'] ?? [];
         if (null === $type = $propertyMetadata->getType()) {
-            return $propertySchema;
+            return new \ArrayObject((array) $propertySchema + $swaggerContext);
```
dunglas: Can't we directly initialize `propertySchema` as an array line 511 and create directly the final `\ArrayObject` here?

greg0ire: @dunglas I think I fixed it, not sure this is exactly what you meant so please check.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -508,7 +508,7 @@ private function getDefinitionSchema(string $resourceClass, ResourceMetadata $re
      */
     private function getPropertySchema(PropertyMetadata $propertyMetadata, \ArrayObject $definitions, array $serializerContext = null): \ArrayObject
     {
-        $propertySchema = new \ArrayObject();
+        $propertySchema = new \ArrayObject($propertyMetadata->getAttributes()['swagger_context'] ?? []);
```
dunglas: Isn't `$propertyMetadata->getAttributes()['swagger_context'] ?? [] ` enough?

greg0ire: Not if you want to return an `ArrayObject` instance on line 522 https://github.com/api-platform/core/pull/1466/files#diff-32d63273c86a7e4c14c3fa6bb4131194R522

dunglas:
Thanks @greg0ire!

PR #1469, opened 2017-10-31
Remove countable

greg0ire:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

This method is not needed, this PR makes the interface easier to implement. Either merge this or #1468, but not both.


dunglas:
It's a BC break... And even if it's not used internally, it may be useful for the developer (I think it was the reason why this was implemented).

I prefer merging #1468.

greg0ire:
@dunglas how is it a BC-break? Do you mean for consuming classes that might type hint against it and use `count()`? Or is there a bigger BC break I didn't see?

dunglas:
> Do you mean for consuming classes that might type hint against it and use count()?

Yes, this one.

PR #1474, opened 2017-11-02
Pass serialization context to GraphQL

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Use case: take into account serialization groups for example.

### DIFF
--- a/src/Bridge/Graphql/Resolver/ItemResolverFactory.php
+++ b/src/Bridge/Graphql/Resolver/ItemResolverFactory.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\Bridge\Graphql\Resolver;
 use ApiPlatform\Core\Api\IdentifiersExtractorInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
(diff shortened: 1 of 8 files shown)


src/Bridge/Graphql/Resolver/ItemResolverFactory.php
```diff
@@ -81,7 +84,13 @@ public function createItemResolver(string $resourceClass, string $rootClass): ca
                 $item = $this->itemDataProvider->getItem($resourceClass, \count($identifiers) > 1 ? \implode(';', $identifiers) : $uniqueIdentifier[0]);
             }
 
-            return $item ? $this->normalizer->normalize($item, null, ['graphql' => true]) : null;
+            return $item ? $this->normalizer->normalize(
+                $item,
+                null,
+                ['graphql' => true] + $this->resourceMetadataFactory
+                    ->create($resourceClass)
```
soyuka: Can't this be `null` if metadata hasn't been found?

meyerbaptiste: Why? Should not the `ResourceClassNotFoundException` be thrown instead?

soyuka: Monday :sleeping:

dunglas:
Thanks @meyerbaptiste

PR #1476, opened 2017-11-02
Catch dbal conversion exception

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #
| License       | MIT
| Doc PR        | #

Using [ramsey/uuid-doctrine](https://github.com/ramsey/uuid-doctrine) doctrine type for entities I've accidentally encountered `Doctrine\DBAL\Types\ConversionException` which was thrown in `ApiPlatform\Core\Bridge\Doctrine\Orm\Util\IdentifierManagerTrait` and the response was 500 instead of 400.
This is possible if user enters a different ID than uuid when fetching a resource. Rethrowing as `InvalidArgumentException` apip properly returns the 400 code.

In slack I didn't got any response from the core team so submitting here to get some feedback. If this patch will be considered to merge I'll try to add test.

Issue #1376, opened 2017-09-14
How Overwrite the GET action to access a composite primary key


lvillarino:
Hi,

I want to know how I can overwrite the GET action to accept a composite primary key.
My current GET action is /entity/{id} and I want to be /entity/{r1}/{r2}/{r3}

My table don't have an unique primary key ID at this time

```
CREATE TABLE `table_name` (
  `r1` smallint(5) unsigned NOT NULL DEFAULT '0',
  `r2` mediumint(8) unsigned NOT NULL DEFAULT '0',
  `r3` enum('A','B','C') NOT NULL DEFAULT 'A',
  `r4` varchar(255) DEFAULT NULL,
  `r5` varchar(25) DEFAULT NULL,
  PRIMARY KEY (`r1`,`r2`,`r3`),
  KEY `table_name_001` (`r2`),
  KEY `table_name_002` (`r5`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

and I can't modify the table

Thanks

soyuka:
Composite are accessible like this: `/entity/r1=1;r2=1;r3=1`

backbone87:
is there any way to change this?
i would love to create api resources for URLs like `/parent/{parent_id}/child/{child_id}`

soyuka:
Could you be more precise @backbone87 ?

backbone87:
Well it applies to basically all 1:n relations with privately owned to-many entities and other things like an attributed n:n.

Example:
- article (id)
- tagdefinition (id)
- tag (article_id, tagdefinition_id, position, created_date)

here tag has a composite foreign primary key.

When i design an API for this (by hand) i would like the following:
- /articles (includes taglist)
- /articles/{id} (includes taglist)
- /tagdefinitions
- /tagdefinitions/{id} 
- /articles/{id}/tags (primarily to insert new tags)
- /articles/{id}/tags/{tagdefinition_id} (primarily to address and delete tags)

soyuka:
Interesting!

The two last routes will not be hard to implement with subresources, I'll try to dig into it! Indeed, if there's already one of the composite key in the endpoint we don't need it twice.

```
POST /articles/{id}/tags 
PUT/DELETE /articles/{id}/tags/{tagdefinition_id}
```

Work needed:
- support `POST`, `PUT`, `DELETE` on subresources (hopefully uses the same WriteListener as usual resources) - works on the last resource part (here tags)
- support composite keys interpretation to allow smarter endpoints

[...]

PR #1478, opened 2017-11-03
Improve ":id" param normalization  

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no (depends)
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | not yet
| Fixed tickets | #1309 #1476 and maybe other (#1376 start) #1234 #1309 
| License       | MIT
| Doc PR        | n/a

## Issue

Currently, as I mentioned in #1309, our `DataProviders` get the raw `id` param coming from the request. ie https://github.com/api-platform/core/blob/183526946637ecf72ce1ddc4032ede281ee3b381/src/EventListener/ReadListener.php#L103

There is an issue here where we sometimes want our `DataProviders` to get a correct instance of the given identifier (eg `Ramsey\UUID` for an uuid or `\DateTime` for a date).

## What's being done

This is what I've come up with. Basically it adds a `IdentifierNormalizer` class (does not implement the `NormalizerInterface` because it's not one) that takes the `string` from `$request->attributes->get('id')` and returns an `array` with correct Types. 
This is not working yet because I have a train to catch haha. Going to add tests and the ability to add normalizers asap (those are `NormalizerInterface`, for example `DateTimeNormalizer` etc.)!

There is a BC break but I've added deprecation and my goal is to keep everything working as before by getting rid of the `IdentifierManagerTrait` in future versions.

## Benefits

With this fix the identifier normalization isn't linked to Doctrine anymore. This is a great improvement for future `DataProviders` that might need the correct types without forcing the developer to use Doctrine. Also, it allows the developper to add custom normalizers for identifiers (for example the UUID mentioned above). 
By having a cleaner way to parse those, it'll also help us manage errors and throw a `NotFoundException` when needed (mentioned by @theofidry in #1309).
Last but not least, I've improved the composite identifier parser as for now something like `a=test;b=bar ;foo;c=123;459;` would totally fail because of the `;` in the identifier value (nb try with `=`).

ping @dkarlovi


src/Identifier/CompositeIdentifierNormalizer.php
```diff
@@ -0,0 +1,44 @@
...
+            } else {
+                $identifiers[$match[1]] = $match[2];
+            }
+        }
+
+        return $identifiers;
+    }
+}
+
+// $id = new IdentifierNormalizer();
```
mickaelandrieu: to be removed, isn't it? :)

soyuka:
@mickaelandrieu haha of course sorry!

PR updated with tests, a compiler pass, correct deprecation, BC break coverage

I just need to update the tests for the deprecation and open a doc PR!

Let me know your thoughts @api-platform/core-team !

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -218,6 +219,19 @@
...
+            <argument type="service" id="api_platform.metadata.property.metadata_factory" />
+            <argument type="service" id="api_platform.identifier.composite_identifier_normalizer" />
+            <argument type="collection" /> <!-- normalizers -->
+        </service>
+
+        <service id="api_platform.identifier.composite_identifier_normalizer" class="ApiPlatform\Core\Identifier\CompositeIdentifierNormalizer" public="false" />
+
+        <service id="api_platform.identifier.date_normalizer" class="Symfony\Component\Serializer\Normalizer\DateTimeNormalizer" public="false">
+            <tag name="api_platform.identifier_normalizer" />
+        </service>
```
soyuka: I'm not sure that this is allowed, lmk if there's something better to use instead.

norkunas: What about adding uuid normalizer? It is very common I think.

dkarlovi:
This looks great @soyuka!

About the "Not found" when un-normalizable (that's actually a word, or my Chrome spell-checker is asleep), maybe introducing a new Exception will enable the desired behaviour, it can be hooked up to a 404 status by default.

soyuka:
TODO:
- [ ] Test uuid denormalizer
- [ ] Improve deprecations

src/Bridge/Doctrine/Orm/ItemDataProvider.php
```diff
@@ -65,11 +65,18 @@ public function supports(string $resourceClass, string $operationName = null): b
      *
      * @throws RuntimeException
      */
-    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = [])
+    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = []/** , array $identifiers **/)
```
soyuka: This was the only solution I had to avoid breaking the `DataProvider::getItem` signature. Also, in future versions `$identifier` should replace `$id` and it will be typed as `array`.

src/Bridge/Symfony/Bundle/Resources/config/api.xml
soyuka: added below :) Just need to test it!

norkunas:
Yes, I'm using it already;)

soyuka:
@norkunas could you try this branch to see if it fixes your issue? I think that my implementation is as far as it will get :).

norkunas:
I'll try this evening.

dkarlovi:
@soyuka I'm testing it now.

dkarlovi:
It works for me (as in, string gets converted to a valid `Ramsey\Uuid\Uuid` instance), but the lookup fails because of https://github.com/ramsey/uuid-doctrine/issues/18. Would probably need to create a custom normalizer which isn't that bad.

LGTM!

soyuka:
Not sure I understood the issue @dkarlovi is it something that needs to be done here? Lmk if needed and if you have a failing example.

dkarlovi:
I think it's Doctrine's fault, it's not able to create the proper value to do a lookup with so you need to supply it yourself.

To be exact, I'm using [binary columns](https://github.com/ramsey/uuid-doctrine#binary-database-columns) to represent IDs (so, they're not in their string representation in the DB). Doctrine should be able to serialize the object properly before querying, but it's not.

I don't think you should change anything here.

soyuka:
@dkarlovi IMO this should be done in the custom `UuidType` where `convertToDatabaseValue` should call `getBytes`.

dkarlovi:
I think Doctrine should fix the underlying issue as there are plenty of places that get bitten by this, but your proposed solution works in the meantime.

In any case, that's outside of scope here: your API Platform fix works as expected and IMHO it can be merged as-is.Thank you for doing it!  :+1:

src/Bridge/Symfony/Bundle/Resources/config/api.xml
dunglas: Why would not it be?

src/Identifier/CompositeIdentifierNormalizer.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Identifier;
+
+/**
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+class CompositeIdentifierNormalizer
```
dunglas: `final`

src/Identifier/IdentifierNormalizer.php
```diff
@@ -0,0 +1,90 @@
...
+
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+class IdentifierNormalizer implements IdentifierNormalizerInterface
```
dunglas: `final`

src/Exception/InvalidIdentifierException.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * Property not found exception.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+class InvalidIdentifierException extends \Exception implements ExceptionInterface
```
dunglas: `final`

src/Bridge/Ramsey/Normalizer/UuidNormalizer.php
```diff
@@ -0,0 +1,42 @@
...
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+/**
+ * Denormalizes an UUID string to an instance of Ramsey\Uuid.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+class UuidNormalizer implements DenormalizerInterface
```
dunglas: `final`

src/Identifier/IdentifierNormalizer.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    private $propertyNameCollectionFactory;
+    private $propertyMetadataFactory;
+    private $compositeIdentifierNormalizer;
```
dunglas: This class has no interface, maybe can it be initialized directly in the constructor instead of being injected?

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -39,6 +39,8 @@
      */
     private function normalizeIdentifiers($id, ObjectManager $manager, string $resourceClass): array
     {
+        @trigger_error(sprintf('Trait ApiPlatform\Core\Bridge\Doctrine\Orm\Util\IdentifierManagerTrait will be removed in version 3.0. Using it is deprecated since 2.2.'), E_USER_DEPRECATED);
```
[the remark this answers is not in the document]
soyuka: Nahh I can't as "php-cs-fixer" is using `self` instead but because it's a trait the resulting class for `self::class` is wrong :|. IMO it's a bug in `php-cs-fixer`.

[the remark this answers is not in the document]
dunglas: Can't you just disable this rule for now?

backbone87:
Sorry that i swing in so late into this PRs progress, but I think the interface should get the whole `Request` object.

Why is that? It would allow one to write a normalizer that can handle stuff like this: `/parent/{parent_id}/child/{child_id}` where childs identifier is composite.

Also for some more non-standard use-cases like taking identifier information from the query string.

dunglas:
@backbone87 you can use the `request_stack` service for that.

soyuka:
@backbone87 yes I was thinking about this use case as well while working on this. However, I'm not sure what `interface` you're talking about. You can overwrite/inject `request_stack` if needed but I doubt it'll be sufficient :|.

If I take your example `/parent/{parent_id}/child/{child_id}` there are a few things that can/should be handled before it can work.

First, today we still rely on [`$id = $request->attributes->get('id');`](https://github.com/soyuka/core/blob/2751a870f2b7ddaa41c9e20ab983edf8c813e7a5/src/EventListener/ReadListener.php#L111) to get the "identifier".
Then, the resource url is created through the `operationPathResolver` (see ApiLoader).

I see an option that could work with this PR, having a custom operation, with `:id` of format `{parent_id}/child/{child_id}` (not sure if `/` can be part of a parameter haha). Then one would need to add an `IdentifierNormalizer` that parses this format.

backbone87:
> today we still rely on $id = $request->attributes->get('id'); to get the "identifier".

this is pretty simple to solve. with this PR the resource gets an already normalized id array. for all the other places that uses `$request->attributes->get('id')` we can provide a fallback:
```php
$identifiers = $normalizer->normalize($request, $resouceClass);
$request->attributes->set('id', http_build_query($identifiers, '', ';'));
```

> then, the resource url is created through the operationPathResolver (see ApiLoader).

this maybe could be solved by providing a "partial" denormalize, which maps the composite ID's keys to the route params without actually doing the `http_build_query` (for id route param). that "partially" denormalized array can be passed to the url generator

backbone87:
> today we still rely on $id = $request->attributes->get('id'); to get the "identifier".

this is pretty simple to solve. with this PR the resource gets an already normalized id array. for all the other places that uses `$request->attributes->get('id')` we can provide a fallback:
```php
$identifiers = $normalizer->normalize($request, $resouceClass);
$request->attributes->set('id', http_build_query($identifiers, '', ';'));
```

> then, the resource url is created through the operationPathResolver (see ApiLoader).

this maybe could be solved by providing a "partial" denormalize, which maps the composite ID's keys to the route params without actually doing the `http_build_query` (for id route param). that "partially" denormalized array can be passed to the url generator

backbone87:
btw. i dont want endpoints like `/parent/{pid}/child/{cid}` to be automatically resolved, but explicitly given in the operation config

soyuka:
Maybe we could just use `$identifiers = $this->identifierNormalizer->normalize($request, $attributes['resource_class']);` this would indeed give much more power to the end user.
The default normalizer would still use `$request->attributes->get('id')` but you'll be able to overwrite the normalizer to do what you want.

WDYT?

soyuka:
ping @api-platform/core-team would love some reviews on this :)

composer.json
```diff
@@ -64,7 +64,9 @@
         "symfony/twig-bundle": "^3.1 || ^4.0",
         "symfony/validator": "^3.3 || ^4.0",
         "symfony/yaml": "^3.3 || ^4.0",
-        "webonyx/graphql-php": "^0.10.2"
+        "webonyx/graphql-php": "^0.10.2",
+        "ramsey/uuid": "^3.7",
```
dunglas: Can you order the deps alphabetically please?

composer.json
```diff
@@ -79,7 +81,8 @@
         "symfony/expression-language": "To use authorization features.",
         "symfony/security": "To use authorization features.",
         "symfony/twig-bundle": "To use the Swagger UI integration.",
-        "webonyx/graphql-php": "To support GraphQL."
+        "webonyx/graphql-php": "To support GraphQL.",
+        "ramsey/uuid": "To support uuid identifiers."
```
dunglas: Same here.

src/Bridge/Doctrine/Orm/ItemDataProvider.php
```diff
@@ -66,11 +66,18 @@ public function supports(string $resourceClass, string $operationName = null): b
      *
      * @throws RuntimeException
      */
-    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = [])
+    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = []/**, array $identifiers = []**/)
```
dunglas: Can't you pass the identifiers in the context instead?

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -106,7 +106,13 @@ public function getSubresource(string $resourceClass, array $identifiers, array
             $qb = $manager->createQueryBuilder();
             $alias = $queryNameGenerator->generateJoinAlias($identifier);
             $relationType = $classMetadata->getAssociationMapping($previousAssociationProperty)['type'];
-            $normalizedIdentifiers = isset($identifiers[$identifier]) ? $this->normalizeIdentifiers($identifiers[$identifier], $manager, $identifierResourceClass) : [];
+
+            $normalizedIdentifiers = [];
+
+            if (isset($identifiers[$identifier])) {
+                // if it's an array it's already normalized, the IdentifierManagerTrait is deprecated
+                $normalizedIdentifiers = is_array($identifiers[$identifier]) ? $identifiers[$identifier] : $this->normalizeIdentifiers($identifiers[$identifier], $manager, $identifierResourceClass);
```
dunglas: `\is_array` (opcache optim).

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -39,6 +39,8 @@
      */
     private function normalizeIdentifiers($id, ObjectManager $manager, string $resourceClass): array
     {
+        @trigger_error(sprintf('Trait "%s" will be removed in version 3.0. Using it is deprecated since 2.2.', IdentifierManagerTrait::class), E_USER_DEPRECATED);
```
dunglas: This trait is internal, so there is no need to deprecate it, you can even remove it if you want to.

src/Bridge/Ramsey/Normalizer/UuidNormalizer.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return Uuid::fromString($data);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsDenormalization($data, $type, $format = null)
+    {
+        return interface_exists(UuidInterface::class) && (UuidInterface::class === $type || is_subclass_of($type, UuidInterface::class));
```
dunglas: I think you can remove the `interface_exists` call (at least, you should check `UuidInterface::class === $type` first because it's less coastly).

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
```diff
@@ -0,0 +1,40 @@
...
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Injects identifier normalizers.
+ *
+ * @internal
```
dunglas: Can't it be part of the public API?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class IdentifierNormalizerPass implements CompilerPassInterface
```
dunglas: Maybe can you bump the dep `symfony/di` to 3.4 and use https://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services instead.

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -73,7 +76,13 @@ public function getItemFromIri(string $iri, array $context = [])
             throw new InvalidArgumentException(sprintf('No resource associated to "%s".', $iri));
         }
 
-        if ($item = $this->itemDataProvider->getItem($parameters['_api_resource_class'], $parameters['id'], null, $context)) {
+        $identifiers = [];
+
+        if ($this->identifierNormalizer) {
```
dunglas: Can't we do this directly in the data provider? It would encapsulate the logic in only one place.

src/DataProvider/ChainItemDataProvider.php
```diff
@@ -35,16 +35,21 @@ public function __construct(array $dataProviders)
     /**
      * {@inheritdoc}
      */
-    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = [])
+    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = []/**, array $identifiers = [] **/)
```
dunglas: IMO we should just normalize the identifiers directly in the `ItemDataProvider` (or use the context), but I would like to get rid of this extra parameter.

src/EventListener/ReadListener.php
```diff
@@ -33,12 +34,19 @@
...
+    private $identifierNormalizer;
 
-    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null)
+    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, IdentifierNormalizerInterface $identifierNormalizer = null)
     {
         $this->collectionDataProvider = $collectionDataProvider;
         $this->itemDataProvider = $itemDataProvider;
         $this->subresourceDataProvider = $subresourceDataProvider;
+
+        if (null === $identifierNormalizer) {
```
dunglas: Is this mandatory, can't we just do nothing if `null` (in many cases, you don't need to normalize the identifier at all).

src/EventListener/ReadListener.php
```diff
@@ -133,10 +146,18 @@ private function getSubresourceData(Request $request, array $attributes)
         }
 
         $identifiers = [];
-        foreach ($attributes['subresource_context']['identifiers'] as $key => list($id, , $hasIdentifier)) {
-            if (true === $hasIdentifier) {
-                $identifiers[$id] = $request->attributes->get($id);
+        try {
```
dunglas: Can't we have a less broad try?

src/EventListener/ReadListener.php
```diff
@@ -133,10 +146,18 @@ private function getSubresourceData(Request $request, array $attributes)
...
+            foreach ($attributes['subresource_context']['identifiers'] as $key => list($id, $resourceClass, $hasIdentifier)) {
```
dunglas:
```php
if (!$hasIdentifier) {
    continue;
}

$identifiers[$id] = $request->attributes->get($id);
if ($this->identifierNormalizer) {
    $identifiers[$id] = $this->identifierNormalizer->normalize((string) $identifiers[$id], $resourceClass);
}
```
$identifiers[$id] = $this->identifierNormalizer ?

src/Identifier/IdentifierNormalizer.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace ApiPlatform\Core\Identifier;
+
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * Identiifer normalizer.
```
dunglas: typo (I suggest: `Normalizes an indentifier`)

src/Identifier/IdentifierNormalizer.php
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class IdentifierNormalizer implements IdentifierNormalizerInterface
```
dunglas: `ChainIdentifierNormalizer` for consistency?

src/Identifier/IdentifierNormalizer.php
```diff
@@ -0,0 +1,92 @@
...
+        // Get metadata identifiers
+        foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $propertyName) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
+
+            if (!$propertyMetadata->isIdentifier()) {
+                continue;
+            }
+
+            $type = $propertyMetadata->getType();
+            $metadataIdentifiers[$propertyName] = !$type ?: (Type::BUILTIN_TYPE_OBJECT === $type->getBuiltinType() ? $type->getClassName() : null);
```
dunglas: Can you switch to if statements (nested ternaries are very hard to read and throw warning in many linters).

src/Identifier/IdentifierNormalizer.php
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        // Normalize identifiers (DateTime, UUID etc.)
+        foreach ($identifiers as $key => $identifier) {
+            foreach ($this->identifierNormalizers as $normalizer) {
+                if ($normalizer->supportsDenormalization($identifier, $metadataIdentifiers[$key])) {
+                    try {
+                        $identifiers[$key] = $normalizer->denormalize($identifier, $metadataIdentifiers[$key]);
+                    } catch (\Exception $e) {
```
dunglas: When can this occur? Catching `\Exception` is usually broader and is a pain to debug.

composer.json
dunglas: You should also add it in the suggest section.

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -220,6 +222,22 @@
...
+
+        <service id="api_platform.identifier.uuid_normalizer" class="ApiPlatform\Core\Bridge\Ramsey\Normalizer\UuidNormalizer" public="false">
```
dunglas: You must remove this service in the extension if Ramsey's UUID isn't installed.

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
soyuka: If I remove the trait, injecting the IdentifierNormalizer will be mandatory

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
soyuka: Why don't we keep this for another PR where we use this improvement on every cases we have?

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
dunglas: All the dependencies of the composite identifier normalizer are already injected in `SubresourceDataProvider`, you can just initialize the class in the constructor if `null` is passed (and trigger a deprecation notice). The other benefit of doing this is that we'll have only one implementation to maintain.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
dunglas: If we introduce a new class, we need to maintain it (and deprecate it soon), but your call.

src/Identifier/IdentifierNormalizer.php
soyuka: Agreed. In that case I'd have to improve the interface and we have to throw an `InvalidIdentifierException` in the denormalizer (fix #1476)

soyuka:
Thanks for the review @dunglas ! I changed the detection in the dataproviders by using the context! 

I don't think it's the data provider responsability to normalize identifiers. We should get the correct identifiers in the data providers as it'd simplify DX when using our own data providers !

TODO: review identifier normalizer exceptions when denormalizing (remove `\Exception`).

src/EventListener/ReadListener.php
```diff
@@ -101,10 +110,17 @@ private function getCollectionData(Request $request, array $attributes)
...
         try {
-            $data = $this->itemDataProvider->getItem($attributes['resource_class'], $id, $attributes['item_operation_name']);
-        } catch (PropertyNotFoundException $e) {
+            if ($this->identifierNormalizer) {
+                $identifiers = $this->identifierNormalizer->normalize((string) $id, $attributes['resource_class']);
+                $context = [ChainIdentifierNormalizer::HAS_IDENTIFIER_NORMALIZER => true];
+            }
+
+            $data = $this->itemDataProvider->getItem($attributes['resource_class'], $identifiers ?: $id, $attributes['item_operation_name'], $context);
+        } catch (InvalidIdentifierException $e) {
```
soyuka: @dunglas this is how I'd like things to be, although I think that this will break current DataProvider implementations. 
If we keep this in the data provider, implementing gets more complex and this path improves DX (end developper doesn't care about identifiers and gets what he expects, for example Uuid/Datetime instance instead of string).

Gounlaf:
Hi,

I've question about this PR - it's not helping for review, sorry =/
I've read the What's being done/Benefits parts, but I still don't get it clearly.

Does this PR will allow us to "identify" an entity with another field than the id/PK?
**Example**: having "regular" ids (integer) as PK, and an UUID field, this one used for identifying the entity from "outside" (explained better in api-platform/api-platform#462).

Thanks

soyuka:
@Gounlaf not really, this patch is only there to improve the "identifier" normalization (ie: ':id' part of `/api/foo/:id`). In api-platform/api-platform#462 the uuid field is the real identifier.

src/EventListener/ReadListener.php
soyuka: ping @dunglas thoughts about using `$context[ChainIdentifierNormalizer::HAS_IDENTIFIER_NORMALIZER]`?

.php_cs.dist
```diff
@@ -55,6 +55,7 @@ return PhpCsFixer\Config::create()
         'strict_comparison' => true,
         'strict_param' => true,
         'ternary_to_null_coalescing' => true,
+        'self_accessor' => false
```
dunglas: Why this change, the `+` is a bug right?

src/EventListener/ReadListener.php
dunglas: Looks good to me.

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class ChainIdentifierNormalizer implements IdentifierNormalizerInterface
+{
+    private $propertyNameCollectionFactory;
+    private $propertyMetadataFactory;
+    private $compositeIdentifierNormalizer;
+    private $identifierNormalizers;
+
+    public const HAS_IDENTIFIER_NORMALIZER = 'has_normalized_identifier';
```
dunglas: Shouldn't this const be defined directly in the interface?

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, array $identifierNormalizers)
+    {
+        $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
+        $this->propertyMetadataFactory = $propertyMetadataFactory;
+        $this->compositeIdentifierNormalizer = new CompositeIdentifierNormalizer();
```
dunglas: Why not injecting it in the existing array?

soyuka: Because it's not implementing the IdentifierNormalizerInterface but it's called pre-normalization when we have more then 1 identifier. See  https://github.com/api-platform/core/pull/1478/files/b3037d9fa3ec663a8f1dd51c771e67a3aeab23f5#diff-9622bccedd195e39c3b3da920161ef42R69

soyuka:
ping @dkarlovi

dkarlovi:
Following up the discussion from Twitter yesterday:

I've updated to 2.1.4 and now getting:

```
Ramsey\Uuid\Exception\InvalidUuidStringException: Invalid UUID string: ,urdpfiń��˗�� in /app/vendor/ramsey/uuid/src/Codec/StringCodec.php:143
Stack trace:
#0 /app/vendor/ramsey/uuid/src/Codec/StringCodec.php(80): Ramsey\Uuid\Codec\StringCodec->extractComponents(',urdpfi\xC5\x84\x9B\x15\x8B\xCB\x97\xD2...')
#1 /app/vendor/ramsey/uuid/src/UuidFactory.php(191): Ramsey\Uuid\Codec\StringCodec->decode(',urdpfi\xC5\x84\x9B\x15\x8B\xCB\x97\xD2...')
#2 /app/vendor/ramsey/uuid/src/Uuid.php(629): Ramsey\Uuid\UuidFactory->fromString(',urdpfi\xC5\x84\x9B\x15\x8B\xCB\x97\xD2...')
#3 /app/vendor/ramsey/uuid-doctrine/src/UuidBinaryType.php(94): Ramsey\Uuid\Uuid::fromString(',uRdpFI\xC5\x84\x9B\x15\x8B\xCB\x97\xD2...')
#4 /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1574): Ramsey\Uuid\Doctrine\UuidBinaryType->convertToDatabaseValue(',uRdpFI\xC5\x84\x9B\x15\x8B\xCB\x97\xD2...', Object(Doctrine\DBAL\Platforms\MySQL57Platform))
#5 /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1539): Doctrine\DBAL\Connection->getBindingInfo(',uRdpFI\xC5\x84\x9B\x15\x8B\xCB\x97\xD2...', Object(Ramsey\Uuid\Doctrine\UuidBinaryType))
#6 /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(906): Doctrine\DBAL\Connection->_bindTypedValues(Object(Doctrine\DBAL\Driver\PDOStatement), Array, Array)
#7 /app/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Exec/SingleSelectExecutor.php(50): Doctrine\DBAL\Connection->executeQuery('SELECT i0_.name...', Array, Array, NULL)
#8 /app/vendor/doctrine/orm/lib/Doctrine/ORM/Query.php(321): Doctrine\ORM\Query\Exec\SingleSelectExecutor->execute(Object(Doctrine\DBAL\Connection), Array, Array)
#9 /app/vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php(969): Doctrine\ORM\Query->_doExecute()
#10 /app/vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php(924): Doctrine\ORM\AbstractQuery->executeIgnoreQueryCache(NULL, NULL)
#11 /app/vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php(766): Doctrine\ORM\AbstractQuery->execute(NULL, NULL)
#12 /app/vendor/api-platform/core/src/Bridge/Doctrine/Orm/ItemDataProvider.php(97): Doctrine\ORM\AbstractQuery->getOneOrNullResult()
#13 /app/vendor/api-platform/core/src/DataProvider/ChainItemDataProvider.php(42): ApiPlatform\Core\Bridge\Doctrine\Orm\ItemDataProvider->getItem('ImageBundle\\Ent...', '2c755264-7046-4...', 'delete', Array)
#14 /app/vendor/api-platform/core/src/EventListener/ReadListener.php(106): ApiPlatform\Core\DataProvider\ChainItemDataProvider->getItem('ImageBundle\\Ent...', '2c755264-7046-4...', 'delete')
#15 /app/vendor/api-platform/core/src/EventListener/ReadListener.php(64): ApiPlatform\Core\EventListener\ReadListener->getItemData(Object(Symfony\Component\HttpFoundation\Request), Array)
#16 [internal function]: ApiPlatform\Core\EventListener\ReadListener->onKernelRequest(Object(Symfony\Component\HttpKernel\Event\GetResponseEvent), 'kernel.request', Object(Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher))
#17 /app/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/WrappedListener.php(104): call_user_func(Array, Object(Symfony\Component\HttpKernel\Event\GetResponseEvent), 'kernel.request', Object(Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher))
#18 [internal function]: Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke(Object(Symfony\Component\HttpKernel\Event\GetResponseEvent), 'kernel.request', Object(Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher))
#19 /app/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php(212): call_user_func(Object(Symfony\Component\EventDispatcher\Debug\WrappedListener), Object(Symfony\Component\HttpKernel\Event\GetResponseEvent), 'kernel.request', Object(Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher))
#20 /app/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php(44): Symfony\Component\EventDispatcher\EventDispatcher->doDispatch(Array, 'kernel.request', Object(Symfony\Component\HttpKernel\Event\GetResponseEvent))
#21 /app/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/TraceableEventDispatcher.php(139): Symfony\Component\EventDispatcher\EventDispatcher->dispatch('kernel.request', Object(Symfony\Component\HttpKernel\Event\GetResponseEvent))
#22 /app/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(129): Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch('kernel.request', Object(Symfony\Component\HttpKernel\Event\GetResponseEvent))
#23 /app/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(68): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#24 /app/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(169): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#25 /app/web/index.php(26): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#26 {main}
```

What I did on 2.1.3 was override `ApiPlatform\Core\Bridge\Doctrine\Orm\Util\IdentifierManagerTrait` with exactly the same copy, only changed

```php
$identifier = DBALType::getType($doctrineTypeName)->convertToPHPValue($identifier, $platform);
```

to

```php
$identifier = DBALType::getType($doctrineTypeName)->convertToDatabaseValue($identifier, $platform);
```

This hack no longer works on 2.1.4, nor does it work without it. So, my question on Twitter was: what changed in 2.1.4 to re-break this?

About your patch, I've tested it before and it works for me, so you want me to test it again?

soyuka:
> About your patch, I've tested it before and it works for me, so you want me to test it again?

Nope that's fine I know that this patch is a better replacement for the trait you have issues with.

I'm not sure I get how you modified the trait :P. 

> This hack no longer works on 2.1.4, nor does it work without it. So, my question on Twitter was: what changed in 2.1.4 to re-break this?

Might be https://github.com/api-platform/core/commit/ed4c6887c9be20945f0c012772a7b7de42032020 but it's better now then it was before.

Anyway, for identifiers that are no string/integer I think that this patch is a must have. Though I won't be able to make it pass in `2.1` :|. Would you be able to jump on 2.2 beta?

dkarlovi:
> I'm not sure I get how you modified the trait :P.

Copied to my codebase, pointed composer to it, excluded API Platform's copy. :|

> Might be ed4c688 but it's better now then it was before.

Thanks, I'll have a look.

>  Would you be able to jump on 2.2 beta?

Probably yes, will be having a round of upgrades to Symfony 3.4 / 4.0 soon anyway.

soyuka:
> Copied to my codebase, pointed composer to it, excluded API Platform's copy. :|

Doh, would probably be better to use a fork directly.

dkarlovi:
Will this get merged before 2.2.0 final?

soyuka:
> Will this get merged before 2.2.0 final?

I'm not sure, I think @dunglas froze the features releases for now :/. I'm maintaining my fork with it for the time being.

bendavies:
What's left to do here?

dkarlovi:
Wait. :(

bendavies:
Yes, my question is for for? 😁

soyuka:
> Yes, my question is what for?

Reviews? :D

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -152,11 +153,17 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
...
-            $identifiers[$identifier],
-            $manager,
-            $identifierResourceClass
-        ) : [];
+
+        $normalizedIdentifiers = [];
+
+        if (isset($identifiers[$identifier])) {
+            // if it's an array it's already normalized, the IdentifierManagerTrait is deprecated
+            $normalizedIdentifiers = ($context[IdentifierNormalizerInterface::HAS_IDENTIFIER_NORMALIZER] ?? false) ? $identifiers[$identifier] : $this->normalizeIdentifiers(
```
dunglas: Can you put everything on a single line?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
```diff
@@ -0,0 +1,44 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Injects identifier normalizers.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class IdentifierNormalizerPass implements CompilerPassInterface
```
dunglas: Can you also make this class `@internal`? When we'll bump the DI component to 4.1, we'll be able to use http://symfony.com/doc/3.4/service_container/tags.html#reference-tagged-services.
Alternatively, we can bump the version now and remove this class.

bendavies: @dunglas do you mean `"^3.4 || ^4.0"`?

dunglas: indeed

Perf:
Hi @dunglas,
do you know when this PR will be merged in `api-platform/core`?
I guess, you will make all `ramsey/uuid` and `api-platform/core` users happy!

@soyuka 
could you please update your fork from `api-platform/core`? It's `59 commits behind api-platform:master`.
As we don't know when this PR will be merged... at least your fork will be fully working and updated!

Thank you guys!

soyuka:
done

bendavies:
@soyuka now https://github.com/api-platform/core/pull/1781 is merged, can you think of a way to improve that, given the work you've done here?

soyuka:
@bendavies I'd not have merged #1781 

Instead we can fix it by adding an `IdentifierNormalizer` that supports the `__toString` method like I did [here](https://github.com/api-platform/core/pull/1478/files#diff-3c6b1ae29084aaabaf272ccf76504f0f) for ramsey uuids.

bendavies:
So we should use the `IdentifierNormalizer`s in the `IdentifiersExtractor`s?

soyuka:
Okay, I looked closer.

What I think is that we won't need your patch with this feature because the identifier can still be an Object, it'll be (de)normalized after anyway.

```
+            if (method_exists($identifiers[$propertyName], '__toString')) {
 +                $identifiers[$propertyName] = (string) $identifiers[$propertyName];
 +                continue;
 +            }
 +
```

What you do here is "denormalizing" an `UUID` normalizer to a string in a class that' responsible for caching identifiers "keys" for a given resource (because it costs a lot to find what properties are identifiers in a resource). 

What should happend with this patch:

1. Keep the object in the identifiers array
2. cast it as string with an IdentifierNormalizer after

In fact I'm going to add your test case to this patch this afternoon and see how it goes! 

> My app uses ramsey uuids for identifiers, and I noticed caching not working here. Using object identifiers effectively disabled the caching of the CachedIdentifiersExtractor, making it call the decorated IdentifiersExtractor every time.

this is something I don't get though. How is a value infering on a cache that's supposed to cache keys (identifier properties, not values) only?

bendavies:
I'm sorry i can't remember exactly, but `return $this->decorated->getIdentifiersFromItem($item);` was being called every time for items with ramsey uuids identifiers. my tests should reveal the problem if you revert the fix.

soyuka:
Updated with latest 2.2 fixes on identifier cache. We fixed it correctly so no additional changes here.

composer.json
```diff
@@ -74,6 +76,7 @@
         "guzzlehttp/guzzle": "To use the HTTP cache invalidation system.",
         "phpdocumentor/reflection-docblock": "To support extracting metadata from PHPDoc.",
         "psr/cache-implementation": "To use metadata caching.",
+        "ramsey/uuid": "To support uuid identifiers.",
```
dunglas: It's not accurate, UUID are supported out of the box if generated client-side (better) or by the DBMS. I would reword this "to support Ramsey's UUID", or just drop this line.

features/bootstrap/FeatureContext.php
```diff
@@ -870,4 +871,16 @@ public function thereAreDummyImmutableDateObjectsWithDummyDate(int $nb)
 
         $this->manager->flush();
     }
+
+    /**
+     * @Given there is a ramsey identified resource with uuid :uuid
+     */
+    public function thereIsARamseyIdentifiedResource(string $uuid)
+    {
+        $u = new RamseyUuidDummy();
```
dunglas: `$uuid`? (explicit is better than implicit 🐍)

src/Bridge/Doctrine/Orm/ItemDataProvider.php
```diff
@@ -69,8 +70,8 @@ public function supports(string $resourceClass, string $operationName = null, ar
     public function getItem(string $resourceClass, $id, string $operationName = null, array $context = [])
     {
         $manager = $this->managerRegistry->getManagerForClass($resourceClass);
-
-        $identifiers = $this->normalizeIdentifiers($id, $manager, $resourceClass);
+        $hasIdentifierNormalizer = $context[IdentifierNormalizerInterface::HAS_IDENTIFIER_NORMALIZER] ?? false;
+        $identifiers = $hasIdentifierNormalizer ? $id : $this->normalizeIdentifiers($id, $manager, $resourceClass);
```
dunglas:
```php
if ($context[IdentifierNormalizerInterface::HAS_IDENTIFIER_NORMALIZER] ?? false) {
    $id = $this->normalizeIdentifiers($id, $manager, $resourceClass);
}
```

? (more readable)

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -152,11 +153,13 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
...
-            $identifiers[$identifier],
-            $manager,
-            $identifierResourceClass
-        ) : [];
+
+        $normalizedIdentifiers = [];
+
+        if (isset($identifiers[$identifier])) {
+            // if it's an array it's already normalized, the IdentifierManagerTrait is deprecated
+            $normalizedIdentifiers = ($context[IdentifierNormalizerInterface::HAS_IDENTIFIER_NORMALIZER] ?? false) ? $identifiers[$identifier] : $this->normalizeIdentifiers($identifiers[$identifier], $manager, $identifierResourceClass);
```
dunglas: Maybe can this expression be decomposed to improve readability?

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
dunglas: What about this one @soyuka?

src/Bridge/Ramsey/Identifier/Normalizer/UuidNormalizer.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Ramsey\Identifier\Normalizer;
```
dunglas: `RamseyUuid`? (in the case we integrate another Ben's library at some point)

src/Bridge/Ramsey/Identifier/Normalizer/UuidNormalizer.php
```diff
@@ -0,0 +1,48 @@
...
+            throw new InvalidIdentifierException($e->getMessage());
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsDenormalization($data, $type, $format = null)
+    {
+        return UuidInterface::class === $type || is_subclass_of($type, UuidInterface::class);
```
dunglas: Maybe can you use `is_a` to remove one test (IMO it should be similar in term of perf).

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
```diff
@@ -0,0 +1,46 @@
...
+    use PriorityTaggedServiceTrait;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $identifierNormalizerDefinition = $container->getDefinition('api_platform.identifier_normalizer');
+
+        if (!class_exists(Uuid::class)) {
```
dunglas: Should be moved in the extension.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/IdentifierNormalizerPass.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Injects identifier normalizers.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
+ * @internal
+ */
+final class IdentifierNormalizerPass implements CompilerPassInterface
```
dunglas: What about dropping this class for https://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services and bumping the minimal version to 3.4?

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -227,6 +229,22 @@
             <argument type="service" id="api_platform.resource_class_resolver" />
         </service>
 
+        <service id="api_platform.identifier_normalizer" class="ApiPlatform\Core\Identifier\Normalizer\ChainIdentifierNormalizer" public="false">
```
dunglas: `api_platform.identifier.normalizer` for consistency with the other services that have been introduced?

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,97 @@
...
+        foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $propertyName) {
+            $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
+
+            if (!$propertyMetadata->isIdentifier()) {
+                continue;
+            }
+
+            $type = $propertyMetadata->getType();
+            if ($type && Type::BUILTIN_TYPE_OBJECT === $type->getBuiltinType()) {
+                $metadataIdentifiers[$propertyName] = $type->getClassName();
```
dunglas: `$metadataIdentifiers[$propertyName] = $type && Type::BUILTIN_TYPE_OBJECT === $type->getBuiltinType() ? $type->getClassName() : null`

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                throw new InvalidIdentifierException(sprintf('Invalid identifier "%s", "%s" has not been found.', $identifier, $propertyName));
+            }
+        }
+
+        // Normalize identifiers (DateTime, UUID etc.)
+        foreach ($identifiers as $key => $identifier) {
+            foreach ($this->identifierNormalizers as $normalizer) {
+                if ($normalizer->supportsDenormalization($identifier, $metadataIdentifiers[$key])) {
```
dunglas: Maybe changing this one in a guard clause would improve the readability.

src/Identifier/Normalizer/DateTimeIdentifierNormalizer.php
```diff
@@ -0,0 +1,29 @@
...
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;
+
+final class DateTimeIdentifierNormalizer extends DateTimeNormalizer implements IdentifierDenormalizerInterface
+{
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        try {
+            return parent::denormalize($data, $class, $format, $context);
+        } catch (\Exception $e) {
```
dunglas: Only Symfony Serializer's `InvalidArgumentException` should be caught.

src/Identifier/Normalizer/IdentifierDenormalizerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Identifier\Normalizer;
+
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+interface IdentifierDenormalizerInterface extends DenormalizerInterface
```
dunglas: Do we really need this new interface? I would use the Symfony one directly.

src/Identifier/Normalizer/IdentifierNormalizerInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Identifier\Normalizer;
+
+interface IdentifierNormalizerInterface
```
dunglas: Why not implementing the Symfony interface here?

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
dunglas: Why does it have a different behavior than other normalizers? It looks weird to me.

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,97 @@
...
+
+            $identifiersNames[] = $propertyName;
+            ++$identifiersCount;
+        }
+
+        // Convert to identifiers array
+        if (1 === $identifiersCount) {
+            $identifiers[$identifiersNames[0]] = $identifier;
+        } else {
+            $identifiers = $this->compositeIdentifierNormalizer->normalize($identifier);
```
dunglas: To me it should be a standard normalizer, that will call recursively other ones if needed. (as we do for data normalizers)

src/Identifier/Normalizer/IdentifierDenormalizerInterface.php
soyuka: Oh okay I remember. This was only to document that the `denormalize` method throws an `InvalidIdentifierException` haha. I can remove it though,

dunglas: Yes please :)

src/Bridge/Ramsey/Identifier/Normalizer/UuidNormalizer.php
soyuka: `is_a` doesn't work with `is_a(Uuid::class, UuidInterface::class)` although Uuid does implement the class.

src/Bridge/Doctrine/Orm/Util/IdentifierManagerTrait.php
```diff
@@ -39,6 +39,8 @@
      */
     private function normalizeIdentifiers($id, ObjectManager $manager, string $resourceClass): array
     {
+        @trigger_error(sprintf('Method "%s" of class "%s" is deprecated, please use identifier normalizers instead.', __METHOD__, __CLASS__), E_USER_DEPRECATED);
```
soyuka: sorry @dunglas have to keep the trait it's used in 2 classes :smile:

soyuka:
Thanks for the review I've refactored a bit:
- uses `DenormalizeInterface` on identifier normalizers
- `CompositeIdentifierNormalizer` is now `CompositeIdentifierParser` 
- tests are green (locally at least I need to debug travis :|)
- fixed IriConverter where the flag was not set

About the deprecation message, IMO we can remove it as it's internal an no-one really uses it? We can't remove the trait because the ItemDataProvider would require `$id` to be an array, for example:
- `['id' => 1]` instead of `1`
- `['id' => 1, 'foo' => 2]` instead of `id=1;foo=2`

In fact, if we could break the DataProvider signature the best would be to type `$id` to an `array`.

dunglas:
> About the deprecation message, IMO we can remove it as it's internal an no-one really uses it?
👍

soyuka:
Any ideas for the `deps=low` test suite? I used the `tagged` thing to avoid using a pass but looks like there's something missing in `3.4`:  `Error: Class 'Symfony\Component\DependencyInjection\Argument\ClosureProxyArgument' not found`

bendavies:
@soyuka `symfony/event-dispatcher v3.3.0` is being installed and using  `Symfony\Component\DependencyInjection\Argument\ClosureProxyArgument` but `symfony/dependency-injection v3.4.0` is installed where it doesn't exist any more.

bendavies:
I think the options are, 
1. restore `symfony/dependency-injection` to `^3.3...`, restore the compiler pass and dropped the `tagged` stuff.
2. raise all low deps to `^3.4`

soyuka:
Green :smile:

dunglas:
Just great! Thanks @soyuka

redecs:
Thank you all for the effort put into this. It's been an issue bugging me for a while now when trying to work with UUIDs (v4).

dkarlovi:
\o/

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+     * @throws InvalidIdentifierException
+     */
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        $keys = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
+
+        if (\count($keys) > 1) {
+            $identifiers = $this->compositeIdentifierParser->parse($data);
+        } else {
+            $identifiers = [$keys[0] => $data];
```
meyerbaptiste: What if `$keys` is empty? 💥

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        // Normalize every identifier (DateTime, UUID etc.)
+        foreach ($keys as $key) {
+            foreach ($this->identifierNormalizers as $normalizer) {
+                if (!isset($identifiers[$key])) {
+                    throw new InvalidIdentifierException(sprintf('Invalid identifier "%s", "%s" was not found.', $key, $key));
```
meyerbaptiste: `sprintf('... %1$s ... %1$s ...', $key)`

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public function supportsDenormalization($data, $type, $format = null)
+    {
+        return true;
+    }
+
+    private function getIdentifierMetadata($class, $propertyName)
+    {
+        $propertyMetadata = $this->propertyMetadataFactory->create($class, $propertyName);
+        $type = $propertyMetadata->getType();
```
meyerbaptiste: `$propertyMetadata` var is useless:
```php
$type = $this->propertyMetadataFactory->create($class, $propertyName)->getType();
```

src/Identifier/Normalizer/ChainIdentifierNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+        if (\count($keys) > 1) {
+            $identifiers = $this->compositeIdentifierParser->parse($data);
+        } else {
+            $identifiers = [$keys[0] => $data];
+        }
+
+        // Normalize every identifier (DateTime, UUID etc.)
+        foreach ($keys as $key) {
+            foreach ($this->identifierNormalizers as $normalizer) {
+                if (!isset($identifiers[$key])) {
```
meyerbaptiste: Could not this condition be moved in the first loop?

src/Identifier/Normalizer/DateTimeIdentifierNormalizer.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+final class DateTimeIdentifierNormalizer extends DateTimeNormalizer implements DenormalizerInterface
+{
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        try {
+            return parent::denormalize($data, $class, $format, $context);
+        } catch (InvalidArgumentException $e) {
+            throw new InvalidIdentifierException($e->getMessage());
```
meyerbaptiste: You should populate the previous exception argument.

.travis.yml
```diff
@@ -41,7 +41,7 @@ before_install:
   - phpenv config-rm xdebug.ini || echo "xdebug not available"
   - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - npm install -g swagger-cli
-  - if [[ $lint = 1 ]]; then wget https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2.8.2/php-cs-fixer.phar; fi
+  - if [[ $lint = 1 ]]; then wget https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2.8.4/php-cs-fixer.phar; fi
```
meyerbaptiste: I think 2.10.0 should work!

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -43,12 +45,14 @@
...
     private $identifiersExtractor;
+    private $identifierNormalizer;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null, IdentifiersExtractorInterface $identifiersExtractor = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null, IdentifiersExtractorInterface $identifiersExtractor = null, DenormalizerInterface $identifierNormalizer = null)
     {
         $this->itemDataProvider = $itemDataProvider;
         $this->routeNameResolver = $routeNameResolver;
         $this->router = $router;
+        $this->identifierNormalizer = $identifierNormalizer;
```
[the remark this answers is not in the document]
soyuka: nope we check below we don't want to break this class.

cve:
#1478 is not it better?

PR #1480, opened 2017-11-06
Use database type conversion in Filters (test postgres+mysql)

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

Stated in [Advanced field value conversion using custom mapping types](http://docs.doctrine-project.org/en/latest/cookbook/advanced-field-value-conversion-using-custom-mapping-types.html):

> When using DQL queries, the convertToPHPValueSQL and convertToDatabaseValueSQL methods only apply to identification variables and path expressions in SELECT clauses. Expressions in WHERE clauses are not wrapped!

We won't see any difference in sqlite for example ([know why here](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/known-vendor-issues.html)) but I'm definitely seeing one in Oracle where the date parameter has the wrong format and therefore leads to a bad query. Maybe we should use this also in numeric filters.

/edit: 

- [x] Add something to skip the ordering of relations while comparing values in JSON
- [x] Adding postgresql to travis:
  I wouldn't think this would be so hard haha. Few bugs discovered:
  - [x] filtering on any integer field with a string containing a decimal (eg`9.3`) will throw (fixed but changes behavior as they're ignored)
  - [x] actually sqlite `LIKE` filters are case insensitive whereas postgres is case sensitive. This renders a few tests useless on sqlite.
  - [x] in `features/numeric_filters` we DO NOT test the numeric filter but only the search filter... (renamed)
  - [x] add numeric filters tests
  - [x] there is weird stuff going on in search_filter which should not be there (mainly on `id` and `associations`) maybe this should've been called `text_filter`. For example we can't say that an identifier is always named `id` [assumption made here](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php#L209) (fixed by using metadata)
  - [x] needs more tests on AbstractFilter (convert) + test iriconvert new method
- [ ] Why does `tags.feature` fail on postgresql?
```
And the header "Cache-Tags" should not exist
And "/relation_embedders,/related_dummies,/third_levels" IRIs should be purged
IRIs "/relation_embedders,/related_dummies/1,/related_dummies,/third_levels/1,/third_levels" does not match expected "/relation_embedders,/related_dummies,/third_levels".
```
- [x] Why does `fos_user.feature` fail? My guess is bad table name?
```
An exception occurred while executing 'INSERT INTO User (username, username_canonical, email, email_canonical, enabled, salt, password, last_login, confirmation_token, password_requested_at, roles, id, fullname) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params ["dummy.user", "dummy.user", "dummy.user@example.com", "dummy.user@example.com", 0, null, "$2y$13$h11wei5V1h8MHUrolsKhCuJo9DgkEDsl62wQVVTJYiTYuncjzRPL6", null, null, null, "a:0:{}", 1, "Dummy User"]:
SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "User"
LINE 1: INSERT INTO User (username, username_canonical, email, email...
                    ^
```

Debug instructions:

```
docker run --name postgres -e POSTGRES_PASSWORD=api-platform -d postgres:9.6.5-alpine
```

```diff
# tests/Fixtures/app/config/config.yml
 doctrine:
     dbal:
-        default_connection: '%default_connection%'
+        default_connection: 'postgres'
@@ -28,10 +28,10 @@ doctrine:
             postgres:
                 driver:   'pdo_pgsql'
-                dbname:   'api_platform_test'
+                dbname:   'postgres'
                 user:     'postgres'
-                password: ''
-                host:     'localhost'
+                password: 'api-platform'
+                host:     '172.17.0.2' # docker inspect
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
 env:
   global:
     - SYMFONY_DEPRECATIONS_HELPER=weak_vendors
+    - APP_ENV=test
 
 matrix:
   include:
@@ -18,11 +19,23 @@ matrix:
     - php: '7.2'
       env: lint=1
     - php: '7.2'
-      env: deps='low'
+      env: deps=low
     - php: '7.2'
-      env: deps='beta'
+      env: deps=beta
     - php: '7.2'
       env: SYMFONY_DEPRECATIONS_HELPER=0
+    - php: '7.2'
+      services:
+          - postgresql
+      before_script:
+          - psql -c 'create database api_platform_test;' -U postgres
+      env: APP_ENV=postgres
+    - php: '7.2'
+      services:
+          - mysql
+      before_script:
+          - mysql -e 'CREATE DATABASE api_platform_test;'
+      env: APP_ENV=mysql
   allow_failures:
       env: SYMFONY_DEPRECATIONS_HELPER=0
 
@@ -41,8 +54,10 @@ install:
 
 script:
   - vendor/bin/phpunit
-  - if [[ $deps = 'beta' ]]; then vendor/bin/behat --format=progress --tags='~@nobeta'; fi
-  - if [[ $deps != 'beta' ]]; then vendor/bin/behat --format=progress; fi
+  - if [[ $APP_ENV = 'test' && $deps = 'beta' ]]; then vendor/bin/behat --format=progress --tags='~@nobeta&&~@postgres'; fi
+  - if [[ $APP_ENV = 'test' && $deps != 'beta' ]]; then vendor/bin/behat --format=progress --tags='~@postgres'; fi
+  - if [[ $APP_ENV = 'postgres' ]]; then vendor/bin/behat --tags='~@sqlite' --format=progress; fi
+  - if [[ $APP_ENV = 'mysql' ]]; then vendor/bin/behat --tags='~@postgres' --format=progress; fi
   - tests/Fixtures/app/console api:swagger:export > swagger.json && swagger-cli validate swagger.json && rm swagger.json
   - if [[ $lint = 1 ]]; then php php-cs-fixer.phar fix --dry-run --diff --no-ansi; fi
   - if [[ $lint = 1 ]]; then phpstan analyse -c phpstan.neon -l5 --ansi src tests; fi
--- a/features/doctrine/date_filter.feature
+++ b/features/doctrine/date_filter.feature
@@ -656,6 +656,15 @@ Feature: Date filter on collections
 
   @dropSchema
   @createSchema
+  Scenario: Get collection filtered by date that is not a datetime
+    Given there is "30" dummydate objects with dummyDate
+    When I send a "GET" request to "/dummy_dates?dummyDate[after]=2015-04-28"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+
+  @dropSchema
+  @createSchema
   Scenario: Get collection filtered by embedded date
     Given there is "2" embedded dummy objects with dummyDate and embeddedDummy
     When I send a "GET" request to "/embedded_dummies?embeddedDummy.dummyDate[after]=2015-04-28"
--- a/src/Bridge/Doctrine/Orm/Filter/DateFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/DateFilter.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Filter;
 
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\QueryBuilder;
 
 /**
@@ -91,6 +93,7 @@ class DateFilter extends AbstractFilter
         }
 
         $nullManagement = $this->properties[$property] ?? null;
+        $type = $this->getDoctrineFieldType($property, $resourceClass);
 
         if (self::EXCLUDE_NULL === $nullManagement) {
             $queryBuilder->andWhere($queryBuilder->expr()->isNotNull(sprintf('%s.%s', $alias, $field)));
@@ -104,7 +107,8 @@ class DateFilter extends AbstractFilter
                 $field,
                 self::PARAMETER_BEFORE,
                 $values[self::PARAMETER_BEFORE],
-                $nullManagement
+                $nullManagement,
+                $type
             );
         }
 
@@ -116,7 +120,8 @@ class DateFilter extends AbstractFilter
                 $field,
                 self::PARAMETER_STRICTLY_BEFORE,
                 $values[self::PARAMETER_STRICTLY_BEFORE],
-                $nullManagement
+                $nullManagement,
+                $type
             );
         }
 
@@ -128,7 +133,8 @@ class DateFilter extends AbstractFilter
                 $field,
                 self::PARAMETER_AFTER,
                 $values[self::PARAMETER_AFTER],
-                $nullManagement
+                $nullManagement,
+                $type
             );
         }
 
@@ -140,7 +146,8 @@ class DateFilter extends AbstractFilter
                 $field,
                 self::PARAMETER_STRICTLY_AFTER,
                 $values[self::PARAMETER_STRICTLY_AFTER],
-                $nullManagement
+                $nullManagement,
+                $type
             );
         }
     }
@@ -155,11 +162,22 @@ class DateFilter extends AbstractFilter
      * @param string                      $operator
      * @param string                      $value
      * @param string|null                 $nullManagement
+     * @param string|Type                 $type
      */
-    protected function addWhere(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, string $operator, string $value, string $nullManagement = null)
+    protected function addWhere(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, string $operator, string $value, string $nullManagement = null, $type = null)
     {
-        $valueParameter = $queryNameGenerator->generateParameterName($field);
+        try {
+            $value = new \DateTime($value);
+        } catch (\Exception $e) {
+            // Silently ignore this filter if it can not be transformed to a \DateTime
+            $this->logger->notice('Invalid filter ignored', [
+              'exception' => new InvalidArgumentException(sprintf('The field "%s" has a wrong date format. Use one accepted by the \DateTime constructor', $field)),
+          ]);
 
+            return;
+        }
+
+        $valueParameter = $queryNameGenerator->generateParameterName($field);
         $operatorValue = [
             self::PARAMETER_BEFORE => '<=',
             self::PARAMETER_STRICTLY_BEFORE => '<',
@@ -185,7 +203,7 @@ class DateFilter extends AbstractFilter
             ));
         }
 
-        $queryBuilder->setParameter($valueParameter, new \DateTime($value));
+        $queryBuilder->setParameter($valueParameter, $value, $type);
     }
 
     /**
@@ -198,10 +216,7 @@ class DateFilter extends AbstractFilter
      */
     protected function isDateField(string $property, string $resourceClass): bool
     {
-        $propertyParts = $this->splitPropertyParts($property, $resourceClass);
-        $metadata = $this->getNestedMetadata($resourceClass, $propertyParts['associations']);
-
-        return isset(self::DOCTRINE_DATE_TYPES[$metadata->getTypeOfField($propertyParts['field'])]);
+        return isset(self::DOCTRINE_DATE_TYPES[$this->getDoctrineFieldType($property, $resourceClass)]);
     }
 
     /**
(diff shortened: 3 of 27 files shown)


src/Bridge/Doctrine/Orm/Filter/DateFilter.php
```diff
@@ -185,7 +191,7 @@ protected function addWhere(QueryBuilder $queryBuilder, QueryNameGeneratorInterf
             ));
         }
 
-        $queryBuilder->setParameter($valueParameter, new \DateTime($value));
+        $queryBuilder->setParameter($valueParameter, $type->convertToDatabaseValue(new \DateTime($value), $queryBuilder->getEntityManager()->getConnection()->getDatabasePlatform()));
```
sroze: What if `$type` is null?

features/doctrine/date_filter.feature
```diff
@@ -656,6 +656,15 @@ Feature: Date filter on collections
 
   @dropSchema
   @createSchema
+  Scenario: Get collection filtered by date that is not a datetime
+    Given there is "30" dummydate objects with dummyDate
+    When I send a "GET" request to "/dummy_dates?dummyDate[after]=2015-04-28"
+    Then the response status code should be 200
```
sroze: You don't test anything in here if you don't test the content of the response. Therefore, I don't understand what is this test trying to prove

soyuka: If I run this test on my database without the patch it fails to status code `500`. This is just to ensure a non-regression.

src/Bridge/Doctrine/Orm/Filter/DateFilter.php
soyuka: I had the condition before but removed it. Indeed `Type::getType` will throw if not found (in `getDateType`).

What I can do:

1. test for `hasType`
2. Use what we had before when `null`

Though, this method will only be called when the `Type` is available in the `DOCTRINE_DATE_TYPES` constant. Therefore, it's not really useful because it'll never be `null` ([those are all defined by default](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/Type.php#L34)).

soyuka:
> Seems very weird that Doctrine isn't in charge of that

Indeed it surprised me as well but according to what I quoted from the docs the behavior is correct no?

features/doctrine/date_filter.feature
dunglas: Does it affects MySQL or Postgres? it it is the case, maybe should we also run the Behat test suite with such DBMS.

soyuka: As long as they have two types for dates (ie date vs datetime) yes. And yes I think we should.

I came across this when using a `date` on an index (it's weird to keep time on an index imo).

tests/Fixtures/app/config/config.yml
```diff
@@ -19,9 +20,18 @@ framework:
 
 doctrine:
     dbal:
-        driver:                        'pdo_sqlite'
-        path:                          '%kernel.cache_dir%/db.sqlite'
-        charset:                       'UTF8'
+        default_connection: '%default_connection%'
```
soyuka: I made an attempt with `%env(DEFAULT_CONNECTION)%` which would've been cleaner IMO but because this is an alias it doesn't work, see https://github.com/symfony/symfony/issues/21609

.travis.yml
```diff
@@ -16,37 +16,53 @@ matrix:
     - php: '7.0'
     - php: '7.1'
     - php: '7.2'
-      env: coverage=1 lint=1
```
dunglas: I would skip those variables and the corresponding tests to speed up the builds.

src/Api/IriConverterInterface.php
```diff
@@ -85,4 +85,11 @@ public function getItemIriFromResourceClass(string $resourceClass, array $identi
      * @return string
      */
     public function getSubresourceIriFromResourceClass(string $resourceClass, array $identifiers, int $referenceType = UrlGeneratorInterface::ABS_PATH): string;
+
+    /**
+     * Gets identifiers from a given IRI.
+     *
+     * @return array
+     */
+    public function getIdentifiersFromIri(string $iri, array $context = []);
```
dunglas: You can use a return type instead of the `@return` annotation.

features/serializer/group_filter_bis.feature
```diff
@@ -0,0 +1,442 @@
+Feature: Filter with serialization groups on items and collections (bis)
```
dunglas: Why (bis)?

src/Bridge/Doctrine/Orm/Filter/NumericFilter.php
```diff
@@ -80,8 +80,12 @@ public function getDescription(string $resourceClass): array
      *
      * @return string
      */
-    private function getType(string $doctrineType): string
+    private function getType(string $doctrineType = null): string
     {
+        if (null === $doctrineType) {
```
dunglas: can be merged with the next text:

```php
if (null === $doctrineType || DBALType::DECIMAL === $doctrineType) {
    return 'string';
}

```

src/Bridge/Doctrine/Orm/Filter/NumericFilter.php
```diff
@@ -118,13 +121,27 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
         if ($this->isPropertyNested($property, $resourceClass)) {
-            list($alias, $field) = $this->addJoinsForNestedProperty($property, $alias, $queryBuilder, $queryNameGenerator, $resourceClass);
+            list($alias, $field, $associations) = $this->addJoinsForNestedProperty($property, $alias, $queryBuilder, $queryNameGenerator, $resourceClass);
+            $metadata = $this->getNestedMetadata($resourceClass, $associations);
+        } else {
+            $metadata = $this->getClassMetadata($resourceClass);
         }
+
+        if (!isset(self::DOCTRINE_NUMERIC_TYPES[$metadata->getTypeOfField($field)])) {
+            $this->logger->notice('Invalid filter ignored', [
```
dunglas: Maybe should you log the filter name, the class and the property too (same in all logs).

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -347,6 +377,8 @@ protected function createWrapCase(bool $caseSensitive): \Closure
      *
      * @param string $value
      *
+     * @deprecated
```
dunglas: Can you add a message explaining what the user should use instead? A call to `trigger_error` should also be added.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -363,6 +395,44 @@ protected function getIdFromValue(string $value)
     }
 
     /**
+     * Transforms IRI to raw identifiers if possible.
+     *
+     * @return array
```
dunglas: Use a return type instead.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -363,6 +395,44 @@ protected function getIdFromValue(string $value)
...
+
+    /**
+     * When the field should be an integer, check that the given value is a valid one.
+     */
+    protected function hasValidValues(array $values, string $field, ClassMetadata $metadata): bool
+    {
+        foreach ($values as $key => $value) {
+            $type = $metadata->getTypeOfField($field);
+            if (DBALType::INTEGER === $type && null !== $value && false === filter_var($value, FILTER_VALIDATE_INT)) {
+                return false;
```
dunglas: Shouldn't we throw a `400 Bad Request` error instead of silently ignoring such errors (I'm not sure, just asking).

features/serializer/group_filter_bis.feature
soyuka: We have an issue on the ulimit again, this file has too many scenarios and need to be splited...

.travis.yml
soyuka: There is still the variable `coverage`, `lint` is not needed anymore as I'm defining the tests in the matrix instead of globally. IMO it's easier to maintain like this! See the `before_install` and `script` sections are only executed in this matrix. The other ones are using those defined globally below.

features/serializer/group_filter_bis.feature
dunglas: Ok :(. Can we found a more descriptive name?

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
soyuka: I asked myself the same but as we're ignoring silently bad properties in other filters (just logging) I kept the same behavior here. However, this is a BC break as [this test](https://github.com/api-platform/core/pull/1480/files#diff-534f9c66d94968377ac76a8bc211da4dR386) is now getting results.

src/Bridge/Doctrine/Orm/Filter/NumericFilter.php
soyuka: Those are logged in the exception.

features/serializer/group_filter_bis.feature
soyuka: `_2`? :laughing:

tests/Fixtures/app/config/config.yml
meyerbaptiste: Why don't you use different environments with multiple config files instead (`config_test.yml`, `config_test_mysql.yml`, `config_test_pgsql`)? Then you just have to set the `APP_ENV` environment variable in your Travis config instead of doing `cp`!

soyuka: would this mean that I'd have to copy the whole configuration file 3 times?

meyerbaptiste: Nop:
* `config_test.yaml`: the whole configuration + 
```yaml
doctrine:
      dbal:
         driver:                        'pdo_sqlite'
         path:                          '%kernel.cache_dir%/db.sqlite'
         charset:                       'UTF8'
```
* `config_test_mysql.yaml`:
```yaml
imports:
    - { resource: config_test.yaml }

doctrine:
    dbal:
        driver: pdo_mysql
        path: ~
        dbname:   'api_platform_test'
        user:     'travis'
        password: ''
        host:     'localhost'
```
etc.

soyuka:
All green :D. 

ping @sroze @dunglas @meyerbaptiste would love some review on this. 

I've simplified the pull request:
- the `numeric_fitler.feature` has a big diff because the old ones were put in `search_filter.feature` and I wrote new ones for filters.
- tests run on postgres, mysql and sqlite (added a `JSON should be deep equal to` feature to compare without taking the order into consideration).
- the only code change is to add the doctrine Type to `setParameter` and to skip invalid values in filters (which is our policy with Doctrine Filters instead of throwing). 

We could argue that there is a BC break where `id=9.99` will be ignored, but IMHO it's a bug and would be silly to count it as break.

Let me know your thoughts!

tests/Fixtures/app/config/config_mysql.yml
```diff
@@ -0,0 +1,10 @@
+imports:
+    - { resource: config_test.yml }
+
+doctrine:
```
dunglas: It should use environment variables to allow to run test locally (same for Postgres).

tests/Fixtures/app/console
```diff
@@ -13,8 +13,8 @@ use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
 
 $input = new ArgvInput();
-$env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
-$debug = getenv('SYMFONY_DEBUG') !== '0' && !$input->hasParameterOption(array('--no-debug', '')) && $env !== 'prod';
+$env = $input->getParameterOption(array('--env', '-e'), $_SERVER['APP_ENV'] ?? null ?: 'test');
+$debug = ($_SERVER['SYMFONY_DEBUG'] ?? null) !== '0' && !$input->hasParameterOption(array('--no-debug', '')) && $env !== 'prod';
```
dunglas: `$_SERVER['SYMFONY_DEBUG'] ?? 0` ?

tests/Fixtures/app/console
```diff
@@ -13,8 +13,8 @@ use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
 
 $input = new ArgvInput();
-$env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
-$debug = getenv('SYMFONY_DEBUG') !== '0' && !$input->hasParameterOption(array('--no-debug', '')) && $env !== 'prod';
+$env = $input->getParameterOption(array('--env', '-e'), $_SERVER['APP_ENV'] ?? null ?: 'test');
```
dunglas: `$_SERVER['SYMFONY_DEBUG'] ?? 'test'`?

soyuka: Didn't get that? This flag is to pilot the debug boolean, env is 'test' by default if no app_env is defined?

tests/Fixtures/app/AppKernel.php
```diff
@@ -47,6 +47,13 @@ public function registerBundles()
 
     public function registerContainerConfiguration(LoaderInterface $loader)
     {
-        $loader->load($this->getRootDir().'/config/config.yml');
+        $environment = $this->getEnvironment();
+
+        // patch for behat not supporting %env(APP_ENV)% in older versions
+        if ($appEnv = $_SERVER['APP_ENV'] ?? null && $appEnv !== $environment) {
```
meyerbaptiste: Should not it be `if (($appEnv = $_SERVER['APP_ENV'] ?? null) && $appEnv !== $environment) {`?

meyerbaptiste: Sorry, fixed by #1524.

PR #1492, opened 2017-11-15
Split GraphQL config from REST config

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Instead of guessing the REST operation corresponding to the GraphQL query and use its configuration, this PR decouple GraphQL from REST, and allow to use different configuration for REST and for GraphQL.

For instance, to configure the `normalization_context` for GraphQL, use the following snippet:

```php
/**
 * @ApiResource(graphqlQuery={"normalization_context"={"groups"={"a"}}})
 */
class MyResource {}
```

If an option is defined in the `attributes` but not in `graphqlQuery`, the option in `attributes` will be used as fallback.

TODO:

* [x] Fix the commented test
* [x] Allow to expose some resources in GraphQL but not in REST, and the reverse

ping @raoulclais @alanpoulain

### DIFF
--- a/src/Metadata/Resource/ResourceMetadata.php
+++ b/src/Metadata/Resource/ResourceMetadata.php
@@ -26,9 +26,10 @@ final class ResourceMetadata
     private $itemOperations;
     private $collectionOperations;
     private $subresourceOperations;
+    private $graphql;
     private $attributes;
 
-    public function __construct(string $shortName = null, string $description = null, string $iri = null, array $itemOperations = null, array $collectionOperations = null, array $attributes = null, array $subresourceOperations = null)
+    public function __construct(string $shortName = null, string $description = null, string $iri = null, array $itemOperations = null, array $collectionOperations = null, array $attributes = null, array $subresourceOperations = null, array $graphql = null)
     {
         $this->shortName = $shortName;
         $this->description = $description;
@@ -36,6 +37,7 @@ final class ResourceMetadata
         $this->itemOperations = $itemOperations;
         $this->collectionOperations = $collectionOperations;
         $this->subresourceOperations = $subresourceOperations;
+        $this->graphql = $graphql;
         $this->attributes = $attributes;
     }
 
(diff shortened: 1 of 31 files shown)


alanpoulain:
Seems good to me but why `graphqlQuery` though and not `graphql`? Even if these options are only applied for the query for now, maybe there will be useful options for the mutations later on (not necessarily in this notation but still).

src/Metadata/Resource/ResourceMetadata.php
```diff
@@ -25,15 +25,17 @@
...
 
-    public function __construct(string $shortName = null, string $description = null, string $iri = null, array $itemOperations = null, array $collectionOperations = null, array $attributes = null)
+    public function __construct(string $shortName = null, string $description = null, string $iri = null, array $itemOperations = null, array $collectionOperations = null, array $attributes = null, array $graphqlQuery = null)
     {
         $this->shortName = $shortName;
         $this->description = $description;
         $this->iri = $iri;
         $this->itemOperations = $itemOperations;
         $this->collectionOperations = $collectionOperations;
+        $this->graphqlQuery = $graphqlQuery;
```
meyerbaptiste: That disturbs me to have references to GraphQL here when it's supposed to be a bridge.

soyuka:
Clearly makes sense to not use operation-related attributes here. 

About the attribute name, I indeed prefer the `graphql` attribute over the `graphqlQuery` one.

However, I'm :+1: with @meyerbaptiste on the "Bridge" thing. Can't we use `attributes->graphql` instead of adding a "Bridged" property to the resource itself?

dunglas:
IMO the GraphQL support should not be in bridge. It's a core feature, and it's on my todolist to move it outside the "bridge" namespace. Not a top priority thing anyway as it's `@experimental`.

alanpoulain:
@dunglas Even if we have a dependency to a third-party library?

dunglas:
Yes IMO it's not a problem (for instance we have dependencies on some Symfony component, including HttpFoundation in core classes, same for willdurand's Negotiation).

theofidry:
👍 IMO it makes much more sense as the idea is to take advantage of GraphQL features not being limited to a RESTful API with GraphQL

PR #1501, opened 2017-11-20
Add tests for Numeric Filter

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Tests were not correctly testing the filter.
The filter on `Dummy` is configured to filter the attributes `dummyPrice` and `dummyFloat`: https://github.com/api-platform/core/blob/master/tests/Fixtures/app/config/config.yml#L147
But the tests were testing only the `id`.
And this condition was true:
https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/NumericFilter.php#L102

### DIFF
--- a/features/doctrine/numeric_filter.feature
+++ b/features/doctrine/numeric_filter.feature
@@ -5,7 +5,7 @@ Feature: Numeric filter on collections
 
   @createSchema
   Scenario: Get collection by id equals 9.99 which is not possible
-    Given there is "30" dummy objects
+    Given there is "30" dummy objects with dummyPrice
     When I send a "GET" request to "/dummies?id=9.99"
     Then the response status code should be 200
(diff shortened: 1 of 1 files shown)


soyuka:
I've done this already in https://github.com/api-platform/core/pull/1480 :|

soyuka:
By the way I think that we should keep the existing tests and add new ones because they're covering some SearchFilter use cases, at least in #1480 they do because some failed on other DBMS :p.

alanpoulain:
Ho my bad :/ I am not following the work on API Platform enough.

PR #1502, opened 2017-11-20
Update typo in interface

sroze:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ø
| License       | MIT

So IDEs are happy with the `@see` and we can Ctrl+Click :)


dunglas:
@soyuka, not before the release of Symfony 4 (we want to be sure that we'll be compatible when it will be out).

soyuka:
Thanks @sroze

PR #1512, opened 2017-11-23
WIP - Subresource maxDepth

Nightbr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1316
| License       | MIT
| Doc PR        | NOT YET - WIP

Hey,

I just made my first PR to api-platform! 

It adds the possibility to pass a maxDepth parameter to ApiSubresources in order to stop the route generation.

Because it's my first PR on api-platform, I don't know how all this works so I need some help.

I'm not sure about the annotation system, I think I am good with my code but if anyone have some advice it would be appreciated. Add to that, is there a modification for yaml and xml configuration to take into account the new parameter?

Finally, someone can validate my code on SubresourcesOperationFactory. Seems good to me and I created a test and it doesn't break other tests but like I said first PR on Api platform and I think I have a lots of thing to learn on this!

After this, I will update the documentation and create more test case.

Thanks in advance.

@soyuka

Issue #1316, opened 2017-08-14
Swagger Look Messy with Subresource


bitcoinporn:
I working with sub resources, and now swagger page look very messy.

Subresource should not appear on target groups. Any setting to prevent this? 

![screen shot 2017-08-14 at 8 39 41 am](https://user-images.githubusercontent.com/25856838/29255627-646c5b42-80cd-11e7-86b7-c10d4650cdd5.png)

synchr0n123r:
![screen shot 2017-08-14 at 8 39 41 am](https://user-images.githubusercontent.com/25856838/29255613-51a2fa0c-80cd-11e7-9790-13139084fcb5.png)

soyuka:
Looks fine 😆

synchr0n123r:
also consider to add depth settings.
I found we can have same naming on sub resource

soyuka:
I'd like a case reproduction for name collision if you have one!

synchr0n123r:
sorry typo...
i mean we can't have same sub resource name.
example:
i have 2 entities, that 2 entities have "likes" sub resources. 
/contents/{id}/likes
/events/{id/likes

when i want to make both as "user" sub resources, its not allowed.
/users/{id}/events/{id}**/likes**
/users/{id}/contents/{id}**/likes**

we should allow this practices.

synchr0n123r:
@soyuka : do you have suggestion for quick fix for this?

soyuka:
@synchr0n123r this look like a bug, I need to take a closer look.

soyuka:
@synchr0n123r I don't get it, this should be good. All you have to do is add a `Subresource` on `Events` and `Contents` in the `User` entity. 

See the referenced commit, that's exactly what the test is trying to do and I'm getting the correct results.

Please give me a reproduction case in a `gist` if you can.

soyuka:
Any updates?

Nightbr:
Hey same issue here, I think it should be nice to control the depth of subresources.

For example, If I have /api/publications/{id}/comments and /api/users/{id}/publications, I don't want to generate the route /api/users/{id}/publications/{publications}/likes. 

I tried by adding @MaxDepth(1) annotation but no effect (I'm pretty sure it's because MaxDepth is related to the serializer and not the route generator but I give it a try ^^) 

Thanks!

soyuka:
[...]

PR #1520, opened 2017-11-24
Subresource maxDepth

Nightbr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1316
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/351

Proper PR from original: https://github.com/api-platform/core/pull/1512

### DIFF
--- a/src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
+++ b/src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
@@ -82,7 +82,8 @@ final class AnnotationSubresourceMetadataFactory implements PropertyMetadataFact
         $type = $propertyMetadata->getType();
         $isCollection = $type->isCollection();
         $resourceClass = $isCollection ? $type->getCollectionValueType()->getClassName() : $type->getClassName();
+        $maxDepth = $annotation->maxDepth;
 
-        return $propertyMetadata->withSubresource(new SubresourceMetadata($resourceClass, $isCollection));
+        return $propertyMetadata->withSubresource(new SubresourceMetadata($resourceClass, $isCollection, $maxDepth));
     }
 }
--- a/src/Metadata/Property/SubresourceMetadata.php
+++ b/src/Metadata/Property/SubresourceMetadata.php
@@ -22,11 +22,13 @@ final class SubresourceMetadata
 {
     private $resourceClass;
     private $collection;
+    private $maxDepth;
 
-    public function __construct(string $resourceClass, bool $collection = false)
+    public function __construct(string $resourceClass, bool $collection = false, int $maxDepth = null)
     {
         $this->resourceClass = $resourceClass;
         $this->collection = $collection;
+        $this->maxDepth = $maxDepth;
     }
 
     public function getResourceClass(): string
@@ -53,5 +55,13 @@ final class SubresourceMetadata
         $metadata->collection = $collection;
 
         return $metadata;
+    }
+
+    /**
+     * @return int|null
+     */
+    public function getMaxDepth()
+    {
+        return $this->maxDepth;
     }
 }
--- a/src/Operation/Factory/SubresourceOperationFactory.php
+++ b/src/Operation/Factory/SubresourceOperationFactory.php
@@ -60,8 +60,10 @@ final class SubresourceOperationFactory implements SubresourceOperationFactoryIn
      * @param string $rootResourceClass null on the first iteration, it then keeps track of the origin resource class
      * @param array  $parentOperation   the previous call operation
      * @param array  $visited
+     * @param int    $depth             the number of visited
+     * @param int    $maxDepth
      */
-    private function computeSubresourceOperations(string $resourceClass, array &$tree, string $rootResourceClass = null, array $parentOperation = null, array $visited = [])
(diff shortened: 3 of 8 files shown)


Nightbr:
Hey @soyuka,

I tried to create a clean PR and rebase with `git fetch origin && git rebase origin/master` but event after I resolve all conflict it keeps create all this mess... I don't understand why this is happening :/

soyuka:
Don't worry about rebasing we can do it later ;). 

I think that you were close from a correct rebase but then you did a `git merge` (probably from a `git pull`) that broke everything. When rebased, you need to force push because it rewrites your history.

To remove the last merge: `git reset --hard HEAD~1` then `git push -f`

soyuka:
Arrgh I fixed it, why did you push again? ^^

Nightbr:
humm, I think it's Gitkraken... It make automatic git pull... I'm sorry

Nightbr:
I think its ok now 😅

soyuka:
@Nightbr no problem haha, FYI `git pull` does `git fetch + git merge` which can be an unwanted thing.

src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
```diff
@@ -82,7 +82,8 @@ private function updateMetadata(ApiSubresource $annotation, PropertyMetadata $pr
         $type = $propertyMetadata->getType();
         $isCollection = $type->isCollection();
         $resourceClass = $isCollection ? $type->getCollectionValueType()->getClassName() : $type->getClassName();
+        $maxDepth = $annotation->maxDepth ? $annotation->maxDepth : null;
```
meyerbaptiste: Do we really want to turn 0 into null?

src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -79,6 +81,15 @@ private function computeSubresourceOperations(string $resourceClass, array &$tre
...
 
             $visiting = "$resourceClass $property $subresourceClass";
+
+            // Handle maxDepth
+            if ($rootResourceClass === $resourceClass) {
+                $maxDepth = $subresource->getMaxDepth();
+            }
+
+            if (null !== $maxDepth && $depth >= $maxDepth) {
+                continue;
```
meyerbaptiste: Should not it be `break` here?

src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
Nightbr: depth 0 has no sens I think. If it's 0 or null, we do not apply maxDepth. If maxDepth=1 we create one subresource operation, if maxDepth=2 we create two subresources depth, ...

soyuka: null is correct here yes

src/Operation/Factory/SubresourceOperationFactory.php
soyuka: :+1:

Nightbr: done! ty

src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
meyerbaptiste: Right, the behaviour should be the same but IMO we should not transform the configured value: 0 is not null (it's important in terms of extension point). What do you think about keeping 0? The behavior will remain the same with your condition in `SubresourceOperationFactory::computeSubresourceOperations()`.

Nightbr: Good point! fixed

src/Metadata/Property/SubresourceMetadata.php
```diff
@@ -54,4 +56,12 @@ public function withCollection(bool $collection): self
 
         return $metadata;
     }
+
+    /**
+     * @return int
```
dunglas: `@return int|null`

src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
```diff
@@ -82,7 +82,8 @@ private function updateMetadata(ApiSubresource $annotation, PropertyMetadata $pr
         $type = $propertyMetadata->getType();
         $isCollection = $type->isCollection();
         $resourceClass = $isCollection ? $type->getCollectionValueType()->getClassName() : $type->getClassName();
+        $maxDepth = $annotation->maxDepth;
```
sroze: Shouldn't we check that `maxDepth` is `int` here?

dunglas: Doctrine Annotation will handle that because of the PHPDock IIRC.

sroze: Oh, I didn't know that. Nice.

src/Metadata/Property/SubresourceMetadata.php
Nightbr: fixed!

soyuka:
Thanks @Nightbr !

Nightbr:
Thanks for your help!

PR #1525, opened 2017-11-28
reduce log level to "error" of exceptions that map to non server error codes

backbone87:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | maybe
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Critical error level may trigger dev notification (email), which is not desired for exceptions that map to anything but http server error codes.

This PR reduces the log level of all exceptions that map to a status code < 500 and is in line with symfonys default exception listeners handling of `HttpException`s.

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\EventListener;
 
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\EventListener\ExceptionListener as BaseExceptionListener;
+use Psr\Log\LoggerInterface;
 
 /**
  * Handles requests errors.
@@ -24,6 +25,23 @@ use Symfony\Component\HttpKernel\EventListener\ExceptionListener as BaseExceptio
  */
 final class ExceptionListener extends BaseExceptionListener
 {
+    /**
+     * @var array
+     */
+    private $exceptionToStatus;
+
+    /**
+     * @param mixed $controller
+     * @param LoggerInterface $logger
+     * @param array $exceptionToStatus
+     */
+    public function __construct($controller, LoggerInterface $logger = null, array $exceptionToStatus = null)
+    {
+        parent::__construct($controller, $logger);
+
+        $this->exceptionToStatus = $exceptionToStatus ?? [];
+    }
+
     public function onKernelException(GetResponseForExceptionEvent $event)
     {
         $request = $event->getRequest();
@@ -36,5 +54,36 @@ final class ExceptionListener extends BaseExceptionListener
         }
 
         parent::onKernelException($event);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function logException(\Exception $exception, $message)
+    {
+        if (null === $this->logger) {
+            parent::logException($exception, $message);
+
+            return;
+        }
+
+        $exceptionClass = get_class($exception);
+        $statusCode = null;
+
+        foreach ($this->exceptionToStatus as $class => $status) {
+            if (is_a($exceptionClass, $class, true)) {
+                $statusCode = $status;
+
+                break;
+            }
+        }
+
+        if (null === $statusCode || $statusCode >= 500) {
+            parent::logException($exception, $message);
+
+            return;
+        }
+
+        $this->logger->error($message, ['exception' => $exception]);
     }
 }
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -13,9 +13,9 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\EventListener;
 
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\EventListener\ExceptionListener as BaseExceptionListener;
-use Psr\Log\LoggerInterface;
 
 /**
  * Handles requests errors.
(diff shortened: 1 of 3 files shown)


src/EventListener/ExceptionListener.php
```diff
@@ -24,6 +25,23 @@
  */
 final class ExceptionListener extends BaseExceptionListener
 {
+    /**
+     * @var array
```
dunglas: Useless docblock, can be removed

src/EventListener/ExceptionListener.php
```diff
@@ -24,6 +25,23 @@
...
+    /**
+     * @param mixed           $controller
+     * @param LoggerInterface $logger
+     * @param array           $exceptionToStatus
+     */
+    public function __construct($controller, LoggerInterface $logger = null, array $exceptionToStatus = null)
+    {
+        parent::__construct($controller, $logger);
+
+        $this->exceptionToStatus = $exceptionToStatus ?? [];
```
dunglas: Can't you set the default value to `[]` directly?

src/EventListener/ExceptionListener.php
```diff
@@ -24,6 +25,23 @@
...
 {
+    /**
+     * @var array
+     */
+    private $exceptionToStatus;
+
+    /**
+     * @param mixed           $controller
+     * @param LoggerInterface $logger
+     * @param array           $exceptionToStatus
```
dunglas: Can be removed too.

src/EventListener/ExceptionListener.php
```diff
@@ -37,4 +55,35 @@ public function onKernelException(GetResponseForExceptionEvent $event)
 
         parent::onKernelException($event);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function logException(\Exception $exception, $message)
```
dunglas: You can typehint `$message` with `string`

src/EventListener/ExceptionListener.php
```diff
@@ -37,4 +55,35 @@ public function onKernelException(GetResponseForExceptionEvent $event)
...
+            parent::logException($exception, $message);
+
+            return;
+        }
+
+        $exceptionClass = get_class($exception);
+        $statusCode = null;
+
+        foreach ($this->exceptionToStatus as $class => $status) {
+            if (is_a($exceptionClass, $class, true)) {
```
dunglas: You can pass `$exception` directly, remove the call to `get_class` and remove the last parameter.

src/EventListener/ExceptionListener.php
```diff
@@ -37,4 +55,35 @@ public function onKernelException(GetResponseForExceptionEvent $event)
...
+                $statusCode = $status;
+
+                break;
+            }
+        }
+
+        if (null === $statusCode || $statusCode >= 500) {
```
dunglas: WDYT about making the error level that bubbles configurable?

dunglas: Btw, I think that the current behavior should be preserved when `APP_DEBUG` is set to true.

src/EventListener/ExceptionListener.php
backbone87: hm, i always use `null` for optional arguments. here is why:
`function test(array $required, array $opt1 = [], array $opt2 = [])` if i want to call `test` without the first optional parameter i have to do it like this: `test([], [], [])`. that means, i have to know that the 2nd parameter has to be an array and that it is an empty array. i cant just "skip" it, by passing `null` like so `test([], null, [])`.
i agree that with arrays that is not such a big deal, but consider the following:
`test2(int $a = SOME_LONG_CONSTANT, int $b = 0)`, if i want to set `b`, i have to use `SOME_LONG_CONSTANT` insteadof just "skipping" the param with `null`.

dunglas: The IDE helps for that (but yes, we need named arguments in PHP!).

We use arrays, boolean and so on everywhere else in the framework, so I would prefer you change it for consistency.

src/EventListener/ExceptionListener.php
backbone87: This behavior is aligned with how symfony deals with exceptions that map to specific HTTP status codes. see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/EventListener/ExceptionListener.php#L88
also the level is still "error", which should be logged in dev **and** prod anyway. the error log level is used for "expected" expections/errors resulting from bad usage (at least that is my interpretation), therefore you may want to log it (to identifiy "bad" users), but you dont want to escalate it (via email, SMS) like critical and above (which denote app errors, that are not expected).

So for my part i dont think, that adjusting the behavior via `APP_DEBUG` is needed. but it could be easily dont in the service config of the listener (via expression not passing the `$exceptionToStatus` array when `APP_DEBUG` is set).

> WDYT about making the error level that bubbles configurable?

I think that this is beyond the scope of this PR, and i would rather use custom listeners for these cases. the reason for that is, that switching log level by class is very coarse grained and what you dont want, is that uncaught exception "slip" through, because the class happens to be configured with "warning" or "info" from one of the exceptions sources, but in another throwing part it is much more severe.

src/EventListener/ExceptionListener.php
backbone87: I agree with removing the block on the constructor, but the type infered for the property is still mixed, because it can be set from anywhere in the class (no `final` for properties like in java)

src/EventListener/ExceptionListener.php
backbone87:
```
Declaration of ApiPlatform\Core\EventListener\ExceptionListener::logException(Exception $exception, string $message) should be compatible with Symfony\Component\HttpKernel\EventListener\ExceptionListener::logException(Exception $exception, $message)
```
cant do that

backbone87:
@dunglas whats the status on this?

src/EventListener/ExceptionListener.php
Simperfit: I agree with dunglas

flug:
@backbone87 Up is this pull request still valid?

backbone87:
cant tell, i currently dont work with api-platform anymore

teohhanhui:
The way I see it, it's not our job to reduce the severity / log level. Note that we don't even deal with the log directly, so this change was wrong.

PR #1527, opened 2017-11-28
Allow another query builder alias

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

This PR allow creating our own query builder and apply filter to it.

There was a "bug" forcing the main alias to be `o` but it's not necessarily the case when creating our own query builder.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -41,6 +41,8 @@ final class OrderExtension implements QueryCollectionExtensionInterface
      */
     public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
     {
+        $rootAlias = $queryBuilder->getRootAliases()[0];
+
         $classMetaData = $queryBuilder->getEntityManager()->getClassMetadata($resourceClass);
         $identifiers = $classMetaData->getIdentifier();
         if (null !== $this->resourceMetadataFactory) {
@@ -54,9 +56,9 @@ final class OrderExtension implements QueryCollectionExtensionInterface
                     }
                     if (false === ($pos = strpos($field, '.'))) {
                         // Configure default filter with property
-                        $field = 'o.'.$field;
+                        $field = "{$rootAlias}.{$field}";
                     } else {
-                        $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, 'o', substr($field, 0, $pos));
+                        $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $rootAlias, substr($field, 0, $pos));
                         $field = sprintf('%s.%s', $alias, substr($field, $pos + 1));
                     }
                     $queryBuilder->addOrderBy($field, $order);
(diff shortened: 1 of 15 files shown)


src/Bridge/Doctrine/Orm/ItemDataProvider.php
```diff
@@ -108,10 +108,11 @@ public function getItem(string $resourceClass, $id, string $operationName = null
      */
     private function addWhereForIdentifiers(array $identifiers, QueryBuilder $queryBuilder, ClassMetadata $classMetadata)
     {
+        $alias = $queryBuilder->getRootAliases()[0];
         foreach ($identifiers as $identifier => $value) {
             $placeholder = ':id_'.$identifier;
             $expression = $queryBuilder->expr()->eq(
-                'o.'.$identifier,
+                $alias.'.'.$identifier,
```
dunglas: Can be `"$alias.$identifier"` (PHP 7 style)

jdeniau:
You are right for the identifier, I will fix it this morning.

For the named createQueryBuilder, I think that the 'o' alias is useless but removing it may break some users using it in their filters.

src/Bridge/Doctrine/Orm/ItemDataProvider.php
jdeniau: not really a PHP 7 feature, no ?

I personnaly avoid double-quoted string as it may easily contain bugs:

```php
$alias = 'foo';
$identier = 'bar'
echo "$alas.$identifier" // will output "$alas.bar"
```

But it's up to you maintainers to decide

soyuka: Please use:
```
"{$alias}.{$identifier}"
```

soyuka:
@dunglas did you mean this?

https://github.com/api-platform/core/blob/ad14ac08f857048c1414addfacfe9bdef76fa34e/src/Bridge/Doctrine/Orm/CollectionDataProvider.php#L63

Then I agree it could be configurable (and could default to `o` to avoid breaking).

src/Bridge/Doctrine/Orm/ItemDataProvider.php
jdeniau: Done

jdeniau:
I do not know why the MYSQL build does not pass, I just rebased on the master branch :/

soyuka:
> I do not know why the MYSQL build does not pass, I just rebased on the master branch :/

To debug this may you add `And print last JSON response` to this feature?

jdeniau:
Well tests are OK if I run them on my machine. Can you try re-starting travis build ?

soyuka:
With mysql? to restart travis you can force push

jdeniau:
@soyuka Fixed by push-forcing ;)

src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
```diff
@@ -68,7 +70,7 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
 
         if (null !== $this->order) {
             foreach ($identifiers as $identifier) {
-                $queryBuilder->addOrderBy('o.'.$identifier, $this->order);
+                $queryBuilder->addOrderBy($rootAlias.'.'.$identifier, $this->order);
```
meyerbaptiste: `"{$rootAlias}.{$identifier}"`

jdeniau:
Fixed @meyerbaptiste @dunglas

meyerbaptiste:
Thanks @jdeniau!

Issue #1130, opened 2017-05-18
Tree : use IRI instead of normalization context


pierre-H:
Hello !

I've got an entity with a tree : 
```php
/**
 * @ApiResource(
 *     itemOperations={
 *          "get"={"method"="GET", "normalization_context"={"groups"={"read"}}},
 *     }
 *  )
 */
class Comment {
    /**
     * @var string
     * @Groups({"read"})
     */
    private $title;

    /**
     * @var Comment
     * @Groups({"read"})
     */
    private $parent;
}
```

I'd like to have the IRI parent, not the content.
The problem here is that I get :
```json
{
    "title": "My first post",
    "parent": {
        "title": "My second post"
    }
}
```

instead of : 
```json
{
    "title": "My first post",
    "parent": "/posts/2"
}
```

I know that there is the `@MaxDepth` annotation, but it doesn't generate IRI. Is there another solution ?

soyuka:
Dooh, I had the same issue yesterday. The solution would be to implement the `circular_reference_handler` configuration in symfony. 

I can't give you my implementation because I use another normalization implementation :/.

pierre-H:
@soyuka `circular_reference_handler` is when the serializer finds a circle. But in my case, there is no circle.

soyuka:
Oh my bad, thought you had children too. For this to work you need to use some kind of dynamic group.

Or maybe we're missing something for the MaxDepth implementation to do what you said, I'll look into this becaused I also expected that result and it didn't happend.

Another solution is to use some kind of dynamic group. Indeed you serialize with `read` on `Comment` and because `parent` is also an instance of `Comment` you'll get the same fields.

teohhanhui:
Related (same?): api-platform/core#1001

pierre-H:
@teohhanhui Yes you're right : it's the same question. I still don't have a good solution for this ...

teohhanhui:
Unfortunately, I don't have a good solution for this either... We'll need to put our minds (and fingers) together to come up with something...

silverbackdan:
This can be closed as https://github.com/api-platform/core/pull/1528 is merged

pierre-H:
Since alwaysIdentifier` doesn't exist now and the Max Depth Handler is only a [function ](https://symfony.com/blog/new-in-symfony-4-1-serializer-improvements) and there is no parameter, how can we solve this problem ?

pierre-H:
For history : 
[...]

PR #1528, opened 2017-11-28
alwaysIdentifier annotation option implementation

adelplace:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1130 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/373

The goal here is to allow the users to display only the IRI of a nested resource. It's very usefull on trees or on big entities which have multiple links on a same other entity.

As the MaxDepth annotation doesn't allow us to do that I have created a new annotation option :
```php
/**
 * @ApiResource(attributes={
 *     "normalization_context"={"groups": {"alwaysIdentifierGroup"}, "enable_always_identifier"=true}
 * })
 * @ORM\Entity
 */
class AlwaysIdentifierDummy 
{
    /**
     * @var ArrayCollection children dummies
     *
     * @ORM\ManyToMany(targetEntity="AlwaysIdentifierDummy")
     * @ApiProperty(alwaysIdentifier=true)
     * @Groups({"alwaysIdentifierGroup"})
     */
    private $children;
}
```

Will be displayed as :

```json
{
      "children": [
          "/always_identifier_dummies/1"
      ]
}
```

I don't know if this work fit with the project guidelines but i would be happy to discuss about it ;)

### DIFF
--- a/src/Metadata/Property/Factory/AnnotationPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/AnnotationPropertyMetadataFactory.php
@@ -117,12 +117,14 @@ final class AnnotationPropertyMetadataFactory implements PropertyMetadataFactory
                 $annotation->identifier,
                 $annotation->iri,
                 null,
-                $annotation->attributes
+                $annotation->attributes,
+                null,
+                $annotation->alwaysIdentifier
             );
         }
 
         $propertyMetadata = $parentPropertyMetadata;
-        foreach ([['get', 'description'], ['is', 'readable'], ['is', 'writable'], ['is', 'readableLink'], ['is', 'writableLink'], ['is', 'required'], ['get', 'iri'], ['is', 'identifier'], ['get', 'attributes']] as $property) {
+        foreach ([['get', 'description'], ['is', 'readable'], ['is', 'writable'], ['is', 'readableLink'], ['is', 'writableLink'], ['is', 'required'], ['get', 'iri'], ['is', 'identifier'], ['get', 'attributes'], ['is', 'alwaysIdentifier']] as $property) {
             if (null !== $value = $annotation->{$property[1]}) {
                 $propertyMetadata = $this->createWith($propertyMetadata, $property, $value);
             }
--- a/src/Metadata/Property/PropertyMetadata.php
+++ b/src/Metadata/Property/PropertyMetadata.php
@@ -34,8 +34,9 @@ final class PropertyMetadata
     private $childInherited;
     private $attributes;
     private $subresource;
+    private $alwaysIdentifier;
 
-    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null)
+    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null, bool $alwaysIdentifier = null)
     {
         $this->type = $type;
         $this->description = $description;
@@ -49,6 +50,7 @@ final class PropertyMetadata
         $this->childInherited = $childInherited;
         $this->attributes = $attributes;
         $this->subresource = $subresource;
+        $this->alwaysIdentifier = $alwaysIdentifier;
     }
 
     /**
@@ -380,5 +382,26 @@ final class PropertyMetadata
         $metadata->subresource = $subresource;
 
         return $metadata;
+    }
+
+    /**
+     * Returns a new instance with the given alwaysIdentifier flag.
+     */
+    public function withAlwaysIdentifier(bool $alwaysIdentifier): self
+    {
+        $metadata = clone $this;
+        $metadata->alwaysIdentifier = $alwaysIdentifier;
+
+        return $metadata;
+    }
+
+    /**
+     * Is the property should be serialized as identifier  ?
+     *
+     * @return bool|null
+     */
+    public function isAlwaysIdentifier()
+    {
+        return $this->alwaysIdentifier;
     }
 }
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -39,6 +39,8 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 {
     use ContextTrait;
 
+    const ALWAYS_IDENTIFIER = 'always_identifier';
+
     protected $propertyNameCollectionFactory;
     protected $propertyMetadataFactory;
     protected $iriConverter;
@@ -95,6 +97,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         if (isset($context['resources'])) {
             $resource = $context['iri'] ?? $this->iriConverter->getIriFromItem($object);
             $context['resources'][$resource] = $resource;
+        }
+
+        if ($context[self::ALWAYS_IDENTIFIER] ?? false) {
+            return $context['iri'] ?? $this->iriConverter->getIriFromItem($object);
         }
 
         return parent::normalize($object, $format, $context);
@@ -396,6 +402,8 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     protected function getAttributeValue($object, $attribute, $format = null, array $context = [])
     {
         $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $attribute, $this->getFactoryOptions($context));
+
+        $context[self::ALWAYS_IDENTIFIER] = $propertyMetadata->isAlwaysIdentifier();
 
         try {
             $attributeValue = $this->propertyAccessor->getValue($object, $attribute);
--- a/tests/Fixtures/TestBundle/Entity/AlwaysIdentifierDummy.php
+++ b/tests/Fixtures/TestBundle/Entity/AlwaysIdentifierDummy.php
@@ -1,0 +1,139 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiProperty;
+use ApiPlatform\Core\Annotation\ApiResource;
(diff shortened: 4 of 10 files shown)


soyuka:
We're successfully doing this by using groups and max depth.

adelplace:
> What about using an attribute of @ApiProperty instead of a new annotation? For example @ApiProperty(skip_serialization=true) (not sure about the key).

Yes it would surely be a better solution.

> I'm not fond of the word IriOnly by the way because we have to keep in mind that IRI's are related to JSON-ld and sometimes we might just want id's properties (ie force to skip serialization of the property).

I'll look to implement it on the other formats.

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -81,6 +88,18 @@ public function normalize($object, $format = null, array $context = [])
...
     }
 
+    /**
+     * Is IRI only enabled on this item ?
+     *
+     * @param array $context
+     *
+     * @return bool
+     */
+    private function isIriOnlyItem(array &$context)
```
meyerbaptiste: Why is `$context` passed by reference?

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -89,6 +108,35 @@ public function supportsDenormalization($data, $type, $format = null)
...
 
+    /**
+     * Is Iri Only enabled for this attribute ?
+     *
+     * @param $propertyMetadata
+     * @param array $context
+     *
+     * @return bool
+     */
+    private function isIriOnlyAttribute($propertyMetadata, array &$context)
```
meyerbaptiste: Why is `$context` passed by reference?

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -89,6 +108,35 @@ public function supportsDenormalization($data, $type, $format = null)
...
+    {
+        return
+            isset($context[static::ENABLE_IRI_ONLY]) &&
```
meyerbaptiste: This class is final, use `self` instead of `static`.

adelplace:
@soyuka @meyerbaptiste 
The changes have been done, tell me if it's better. I also added tests for json and hal formats.

src/Metadata/Property/Factory/AnnotationPropertyMetadataFactory.php
```diff
@@ -117,12 +117,14 @@ private function createMetadata(ApiProperty $annotation, PropertyMetadata $paren
                 $annotation->identifier,
                 $annotation->iri,
                 null,
-                $annotation->attributes
+                $annotation->attributes,
+                null,
+                $annotation->skipSerialization
```
soyuka: I think that the word "serializable" would be better:

```
isSerializable
withSerializable
serializable
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -387,6 +406,21 @@ protected function createRelationSerializationContext(string $resourceClass, arr
...
 
+    /**
+     * Is normalization skipping enabled for this attribute ?
+     *
+     * @param $propertyMetadata
+     * @param array $context
+     *
+     * @return bool
+     */
+    private function isSkippedSerializationAttribute($propertyMetadata, array $context)
```
soyuka: `isPropertySerializable`

Also missing typehints and you can remove the block comments (`@param/@return`):

```
function isPropertySerializable(PropertyMetadata $propertyMetadata, array $context): bool 
{}
```
Then negate: `!$this->isPropertySerializable(...)`

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -97,9 +100,25 @@ public function normalize($object, $format = null, array $context = [])
...
     }
 
+    /**
+     * Is normalization should be skipped on this item ?
+     *
+     * @param array $context
+     *
+     * @return bool
+     */
+    protected function isItemSerializationShouldBeSkipped(array $context)
```
soyuka: `shouldSkipSerialization` same you can add a return type `:bool` and remove `@param/@return`.

dunglas:
What about `always_serialize_as_id` instead? I don't like the "skip serialization" term because the serialization is not skipped (to me it would mean "just skip this property", here it is serialized in a special type: an IRI (btw, IRI is a generic term; not specific to JSON-LD: https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier

soyuka:
👍 @dunglas , I'd propose in the metadata:

```
isAlwaysIdentifier
withAlwaysIdentifier
alwaysIdentifier
```

adelplace:
@soyuka  @dunglas

features/main/skip_serialization.feature
```diff
@@ -0,0 +1,118 @@
+@skip
```
soyuka: why skip?

src/Metadata/Property/PropertyMetadata.php
```diff
@@ -381,4 +383,29 @@ public function withSubresource(SubresourceMetadata $subresource = null): self
 
         return $metadata;
     }
+
+    /**
+     * Returns a new instance with the given alwaysIdentifier flag.
+     *
+     * @param bool $alwaysIdentifier
```
dunglas: You can remove both `@param` and `@return`, they are useless because of the typehints.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -97,9 +100,21 @@ public function normalize($object, $format = null, array $context = [])
             $context['resources'][$resource] = $resource;
         }
 
+        if ($this->isAlwaysIdentifierItem($context)) {
```
dunglas: Can you inline this one? We try to reduce the number of function calls in the serialization process (which is recursive) because a function call is costly:

```php
if (($context[self::ALWAYS_IDENTIFIER_KEY] ?? false) && ($context[self::ALWAYS_IDENTIFIER_KEY] ?? false))
```

Same for `isAlwaysIdentifierProperty`, I would drop these methods.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -39,6 +39,9 @@
 {
     use ContextTrait;
 
+    const ENABLE_ALWAYS_IDENTIFIER = 'enable_always_identifier';
+    const ALWAYS_IDENTIFIER_KEY = 'always_identifier_key';
```
dunglas: We use the `_key` suffix nowhere. Can you drop it?

tests/Fixtures/TestBundle/Entity/AlwaysIdentifierDummy.php
```diff
@@ -0,0 +1,133 @@
...
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+/**
+ * AlwaysIdentifierDummy.
+ *
+ * @author Alexandre Delplace <alexandre.delplacemille@gmail.com>
+ *
+ * @ApiResource(attributes={
+ *     "normalization_context"={"groups"={"alwaysIdentifierGroup"}, "enable_always_identifier"=true}
```
dunglas: Why do we need this `enable_always_identifier` flag? IMO we can remove it, if the `alwaysIdentifier` is set on a property, we use it, that's all, it doesn't make sense to allow to disable this feature. If you don't want to use it, just don't set the flag.

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
+Feature: AlwaysIdentifier support
```
dunglas: May  I suggest: `Feature: allow resources in properties to always be serialized as identifiers`

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
...
+  In order to use a hypermedia API
+  As a client software developer
+  I need to be able to force identifier display of related entities
```
dunglas: I need to be able to force to serialize only the identifier of the related resource

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
...
+
+  @createSchema
+  Scenario: Create a set of dummy with alwaysIdentifier attributes and check the jsonld formatted response
```
dunglas: Create a set of Dummy with the alwaysIdentifier attributes and check the JSON-LD formatted response

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
...
+                "children": [],
+                "related": [],
+                "parent": null
+            }
+        ],
+        "parent": "/always_identifier_dummies/1"
+    }
+    """
+
+    Scenario: Check the json formatted response
```
dunglas: JSON

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
...
+      When I add "Accept" header equal to "application/json"
+      And I add "Content-Type" header equal to "application/json"
+      And I send a "GET" request to "/always_identifier_dummies/2"
+      Then the JSON should be equal to:
+      """
+      {
+          "children": [
+              "\/always_identifier_dummies\/1"
```
dunglas: Can you drop the `\` please (for readability), same for all others.

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
...
+                  "children": [],
+                  "related": [],
+                  "parent": null
+              }
+          ],
+          "parent": "\/always_identifier_dummies\/1"
+      }
+      """
+
+  Scenario: Check the hal formatted response
```
dunglas: HAL

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
...
+                      }
+                  }
+              ],
+              "parent": "\/always_identifier_dummies\/1"
+          }
+      }
+      """
+
+  @dropSchema
+  Scenario: Delete a resource
```
dunglas: This is not necessary (this is not related to this feature and already tested). `@dropSchema` will already destroy the database.

features/main/always_identifier.feature
```diff
@@ -0,0 +1,117 @@
+Feature: AlwaysIdentifier support
+  In order to use a hypermedia API
```
dunglas: You can create a new context method in `FeatureContext` like https://github.com/api-platform/core/blob/master/features/bootstrap/FeatureContext.php#L125 to create new `AlwaysIdentifiersDummies`, remove the create and delete scenarios, and move the scenarios for each format (JSON-LD, HAL, JSON) in the corresponding directories (we prefer to group tests per format instead of per feature, it will help when we will create a subtree split).

tests/Fixtures/TestBundle/Entity/AlwaysIdentifierDummy.php
adelplace: I took example from the max_depth implementation. I'll remove it

adelplace:
@dunglas done. I also added a json folder in the features because I didn't find a specific one for this format.

dunglas:
Thanks @adelplace

PR #1544, opened 2017-12-05
Restore the Postgres tests

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      |  no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ install:
 script:
   - vendor/bin/phpunit
   - if [[ $APP_ENV != 'postgres' ]]; then vendor/bin/behat --format=progress; fi
-  - if [[ $APP_ENV = 'postgres' ]]; then vendor/bin/behat --tags='~@sqlite||@postgres' --format=progress; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -52,7 +52,7 @@ install:
 script:
   - vendor/bin/phpunit
   - if [[ $APP_ENV != 'postgres' ]]; then vendor/bin/behat --format=progress; fi
-  - if [[ $APP_ENV = 'postgres' ]]; then vendor/bin/behat --tags='~@sqlite||@postgres' --format=progress; fi
+  - if [[ $APP_ENV = 'postgres' ]]; then vendor/bin/behat --tags='~@sqlite' --format=progress; fi
```
soyuka: note that postgres tags are filtered by default in the `behat.yml.dist` now.

dunglas: You mean I need to restore `~@sqlite||@postgres`?

soyuka: I think it's good like this because CLI defined tags are replacing the ones from the config !

PR #1546, opened 2017-12-05
Trigger eager loading when using the attributes serializer option

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

TODO:

* [x] Add a test

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -144,6 +144,13 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
                 unset($context[AbstractNormalizer::ATTRIBUTES]);
             }
 
+            if (
+                ((!$inAttributes && false === $propertyMetadata->isReadableLink()) || false === $propertyMetadata->isReadable()) &&
+                false === $propertyMetadata->getAttribute('fetchEager', false)
+            ) {
+                continue;
+            }
+
             $isNullable = $mapping['joinColumns'][0]['nullable'] ?? true;
             if (false !== $wasLeftJoin || true === $isNullable) {
                 $method = 'leftJoin';
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -137,8 +137,11 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -137,7 +137,17 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
...
 
-            if ((false === $propertyMetadata->isReadableLink() || false === $propertyMetadata->isReadable()) && false === $propertyMetadata->getAttribute('fetchEager', false)) {
+            if ($inAttributes = isset($context[AbstractNormalizer::ATTRIBUTES][$association])) {
+                // prepare the child context
+                $context[AbstractNormalizer::ATTRIBUTES] = $context[AbstractNormalizer::ATTRIBUTES][$association];
+            } else {
+                unset($context[AbstractNormalizer::ATTRIBUTES]);
+            }
+
+            if (
```
Simperfit: cs is strange here

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -137,7 +137,17 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
...
+                ((!$inAttributes && false === $propertyMetadata->isReadableLink()) || false === $propertyMetadata->isReadable()) &&
+                false === $propertyMetadata->getAttribute('fetchEager', false)
```
soyuka: ` false === $propertyMetadata->getAttribute('fetchEager', false)` should not be needed here as it should've been tested https://github.com/dunglas/core/blob/master/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php#L142 

Did you start you PR on master? I can't recognize the replaced code which may have lead to weird copy/paste?

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: Why?

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: Can I let you handle that? I've changed this part (only the CS) and I'm not really comfortable with it.

PR #1547, opened 2017-12-06
Fix: Passing nor an unknown iri nor an array in an embedded relation throws an error

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ~
| License       | MIT
| Doc PR        | ~

`POST /relation_embedders`
with body:
```json
{
  "related": "foo"
}
```

will throw an error:
```Type error: Argument 2 passed to ApiPlatform\Core\Metadata\Property\Factory\CachedPropertyMetadataFactory::create() must be of the type string, integer given, called in src/Serializer/AbstractItemNormalizer.php on line 163 (Behat\Testwork\Call\Exception\FatalThrowableError)```

This is because "related" is marked as a "writable link"

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -285,7 +285,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             }
         }
 
-        if (!$this->resourceClassResolver->isResourceClass($className) || $propertyMetadata->isWritableLink()) {
+        if (
+            !$this->resourceClassResolver->isResourceClass($className) ||
+            ($propertyMetadata->isWritableLink() && is_array($value))
+        ) {
             return $this->serializer->denormalize($value, $className, $format, $this->createRelationSerializationContext($className, $context));
         }
 
--- a/features/main/relation.feature
+++ b/features/main/relation.feature
@@ -442,3 +442,28 @@ Feature: Relations support
       "related": null
     }
     """
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by backbone87** [comment]
`foo` can be a valid IRI

**Comment by jdeniau** [comment]
It can be a valid IRI but it's not knwn by our application

**Comment by jdeniau** [comment]
Well tests are passing except the symfony deprecated (because I am based on 2.0).

I do not know why travis does not report status: https://travis-ci.org/api-platform/core/builds/312259197

File: src/Serializer/AbstractItemNormalizer.php [hunk 1]
```diff
@@ -285,7 +285,9 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
             }
         }
 
-        if (!$this->resourceClassResolver->isResourceClass($className) || $propertyMetadata->isWritableLink()) {
+        $isResourceClass = $this->resourceClassResolver->isResourceClass($className);
```
dunglas: If I understand correctly, the variables are useless, can you just add the `is_array` check without introducing new vars?

jdeniau:
Yes they are just for the readability

dunglas:
@jdeniau can you remove them please, we try to make as few assignation as possible (especially in the normalizers).

jdeniau:
done

dunglas:
Thanks @jdeniau!

PR #1552, opened 2017-12-07
Improved joins when using the properties filter

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Follows #1546. This PR allows API Platform to add useful SQL `JOIN` clauses and to remove the useless ones according to properties selected by the client using the [property filter](https://api-platform.com/docs/core/filters#property-filter).
That's a massive performance improvement when using this filter.

I'll also use this feature later to improve the generated SQL queries when using GraphQL. It will allow to only fetch what is required for a given GraphQL query while to reducing dramatically the number of issued SQL queries for a whole GraphQL request.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -137,16 +137,22 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
                 continue;
             }
 
-            if ($inAttributes = isset($context[AbstractNormalizer::ATTRIBUTES][$association])) {
-                // prepare the child context
-                $context[AbstractNormalizer::ATTRIBUTES] = $context[AbstractNormalizer::ATTRIBUTES][$association];
-            } else {
-                unset($context[AbstractNormalizer::ATTRIBUTES]);
+            if ($useAttributes = isset($context[AbstractNormalizer::ATTRIBUTES])) {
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -137,16 +137,22 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
...
             }
 
+            $isNotReadableLink = false === $propertyMetadata->isReadableLink();
             if (
-                ((!$inAttributes && false === $propertyMetadata->isReadableLink()) || false === $propertyMetadata->isReadable()) &&
-                false === $propertyMetadata->getAttribute('fetchEager', false)
+                false === $propertyMetadata->getAttribute('fetchEager', false) &&
+                (
+                    false === $propertyMetadata->isReadable() ||
+                    ((!$useAttributes && $isNotReadableLink) || ($useAttributes && !$inAttributes))
```
soyuka: linter has an issue with `inAttributes` anyway it's fine to me. Glad to see the EagerLoading extension serve it's purpose :D.

dunglas: Fixed!

PR #1560, opened 2017-12-13
Allow passing `itemsPerPage=0`

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | Well really really minor :)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ~
| License       | MIT
| Doc PR        | ~

Specifying `itemsPerPage=0` is quite a handy feature if you just need a total count of the items (particularly if you use a filter), so I added this possibility.

There is a really small BC break (but you might see it as a bugfix ?): 

  * If you specify `itemsPerPage=0`, you will not have a 400 status code but a 200 with `hydra:member` set to `[]` and the `hydra:totalItems` set to the total number of items.

I kept the 400 error when you specify `itemsPerPage=0&page=2` (page > 1) but I can easily remove it, it's just weird to make this call but the `page` parameter is just ignored if I remove the exception.

PR #1561, opened 2017-12-13
Allow passing `itemsPerPage=0`

jdeniau:
See https://github.com/api-platform/core/pull/1560#issuecomment-351359773


dunglas:
Thanks @jdeniau

PR #1563, opened 2017-12-14
[SymfonyBundle] Fixed the configuration

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | I guess
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

---

The value of title/description/version should not be empty.
If there are empty, we got the following exception:

```
Type error: Argument 8 passed to ApiPlatform\Core\Bridge\Symfony\Bundle\Action\SwaggerUiAction::__construct() must be of the type string, null given, called in /var/www/server/backend/var/cache/dev/appDevDebugProjectContainer.php on line 1155
```

How to reproduce:

```
#config.yml
api_platform:
    title: #nothing explicitly 
```

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -40,9 +40,21 @@ final class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
-                ->scalarNode('title')->defaultValue('')->info('The title of the API.')->end()
-                ->scalarNode('description')->defaultValue('')->info('The description of the API.')->end()
-                ->scalarNode('version')->defaultValue('0.0.0')->info('The version of the API.')->end()
+                ->scalarNode('title')
+                    ->info('The title of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
+                ->end()
+                ->scalarNode('description')
+                    ->info('The description of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
+                ->end()
+                ->scalarNode('version')
+                    ->info('The version of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('0.0.0')
+                ->end()
                 ->scalarNode('default_operation_path_resolver')
                     ->beforeNormalization()->always(function ($v) {
                         @trigger_error('The use of the `default_operation_path_resolver` has been deprecated in 2.1 and will be removed in 3.0. Use `path_segment_name_generator` instead.', E_USER_DEPRECATED);


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -42,9 +42,21 @@ public function getConfigTreeBuilder()
...
+                    ->info('The title of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
+                ->end()
+                ->scalarNode('description')
+                    ->info('The description of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
+                ->end()
+                ->scalarNode('vesion')
```
meyerbaptiste: `version`

lyrixx: Oups. Goog catch. Thanks (We should never edit code on Github directly :p)

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -42,9 +42,21 @@ public function getConfigTreeBuilder()
...
+                ->end()
+                ->scalarNode('description')
+                    ->info('The description of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
+                ->end()
+                ->scalarNode('version')
+                    ->info('The version of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
```
meyerbaptiste: Is there any reason to delete the old default value (`0.0.0`)?

lyrixx: Shame on me ! 
The reason: I copy paste the previous block. It's not fixed.
So Sorry

dunglas:
Can you rebase against 2.1, as it's a bug fix? Thanks!

lyrixx:
@dunglas Here we go ;)

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -40,9 +40,21 @@ public function getConfigTreeBuilder()
 
         $rootNode
             ->children()
-                ->scalarNode('title')->defaultValue('')->info('The title of the API.')->end()
-                ->scalarNode('description')->defaultValue('')->info('The description of the API.')->end()
-                ->scalarNode('version')->defaultValue('0.0.0')->info('The version of the API.')->end()
+                ->scalarNode('title')
+                    ->info('The title of the API.')
+                    ->cannotBeEmpty()
+                    ->defaultValue('')
```
chalasr: isn't it useless given the node that cannot be empty?

dunglas: In fact we must allow empty strings, but not `null`. Maybe should be just normalize empty to `''`?

lyrixx: IMHO, the config here is simpler than normalizing everything ...

soyuka:
I don't think failing test is related, maybe a rebase would fix CI.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
chalasr: but forbids empty strings

lyrixx:
I just rebased.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
dunglas: I agree we must allow empty strings or it will be a big BC break (for instance the Flex recipe doesn’t set those keys).

chalasr:
https://github.com/api-platform/core/pull/1563#discussion_r157848790

PR #1569, opened 2017-12-18
 Use IRIs as GraphQL's ids and improve perf 

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

* Refactor (simplify) the GraphQL subsystem
* Use IRIs as id (Relay compatibility, better compat with REST endpoints - I'll blog about this -, easier handling of composed identifiers)
* Leverage the `attributes` feature of the Symfony serializer (will allow to reduce the number of SQL queries in a following PR)

TODO:

* [x] Add some more unit tests


src/Graphql/Resolver/ItemMutationResolverFactory.php
```diff
@@ -72,8 +72,8 @@ public function createItemMutationResolver(string $resourceClass, string $mutati
 
                     return $this->normalizer->normalize(
                         $item,
-                        null,
-                        ['graphql' => true] + $resourceMetadata->getGraphqlAttribute($mutationName, 'normalization_context', [], true)
+                        'graphql',
```
alanpoulain: `ItemNormalizer::FORMAT`

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -170,53 +168,30 @@ private function getResourceFieldConfiguration(string $fieldDescription = null,
...
+                $resolverFactory = $type->isCollection() ? $this->collectionResolverFactory : $this->itemResolverFactory;
+                $resolve = $resolverFactory($className, $rootResource);
             }
 
             return [
                 'type' => $graphqlType,
                 'description' => $fieldDescription,
                 'args' => $args,
                 'resolve' => $resolve,
+                'context' => ['foo' => 'bar'],
```
alanpoulain: What is this?

src/HttpCache/EventListener/AddHeadersListener.php
```diff
@@ -27,15 +27,15 @@
...
+    private $sharedMaxAge;
     private $vary;
     private $public;
 
     public function __construct(bool $etag = false, int $maxAge = null, int $sharedMaxAge = null, array $vary = null, bool $public = null)
     {
         $this->etag = $etag;
         $this->maxAge = $maxAge;
-        $this->sharedMaxage = $sharedMaxAge;
+        $this->sharedMaxAge = $sharedMaxAge;
```
alanpoulain: Bad rebase?

src/Graphql/Resolver/CollectionResolverFactory.php
```diff
@@ -96,18 +97,52 @@ public function createCollectionResolver(string $resourceClass, string $rootClas
...
+
+        return $fields;
+    }
+
+    /**
+     * @throws ResourceClassNotSupportedException
+     *
+     * @return object|null
+     */
+    private function getSubresource(string $rootClass, array $rootResolvedFields, array $rootIdentifiers, string $rootProperty, string $subresourceClass, bool $isCollection)
```
soyuka: this code is also done in the `SubresourceOperationFactory`, maybe we should use the service instead to prevent side effects in the future?

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -146,7 +144,7 @@ private function getResourceFieldConfiguration(string $fieldDescription = null,
         try {
             $graphqlType = $this->convertType($type, $isInput, $isMutation, $mutationName);
             $graphqlWrappedType = $graphqlType instanceof WrappingType ? $graphqlType->getWrappedType() : $graphqlType;
-            $isInternalGraphqlType = in_array($graphqlWrappedType, GraphQLType::getInternalTypes(), true);
+            $isInternalGraphqlType = \in_array($graphqlWrappedType, GraphQLType::getInternalTypes(), true);
```
soyuka: `\` not needed

tests/Graphql/Resolver/CollectionResolverFactoryTest.php
```diff
@@ -123,66 +161,55 @@ public function testCreatePaginatorCollectionResolver()
...
+        $collectionDataProviderProphecy->getCollection(Dummy::class)->willReturn($collection);
+        $collectionDataProviderProphecy->getCollection(RelatedDummy::class)->willReturn($collection);
 
         $subresourceDataProviderProphecy = $this->prophesize(SubresourceDataProviderInterface::class);
-        $subresourceDataProviderProphecy->getSubresource('subresourceClass', $identifiers, [
-            'property' => 'rootProperty',
-            'identifiers' => [['rootIdentifier', 'rootClass']],
+        $subresourceDataProviderProphecy->getSubresource(RelatedDummy::class, $identifiers, [
+            'property' => 'relatedDummies',
+            'identifiers' => [['id', Dummy::class]],
```
soyuka: Sometimes `id` isn't called `id` but is another property. Maybe we should add such test to prevent failing cases.

tests/Graphql/Resolver/CollectionResolverFactoryTest.php
```diff
@@ -123,66 +161,55 @@ public function testCreatePaginatorCollectionResolver()
...
             'collection' => true,
         ])->willReturn($subcollection);
 
         $normalizerProphecy = $this->prophesize(NormalizerInterface::class);
 
-        if (!is_array($collection)) {
-            $normalizerProphecy->normalize($collection->current(), Argument::cetera())->willReturn('normalized'.$collection->current());
-        } else {
+        if (\is_array($collection)) {
```
soyuka: `\` not needed

tests/Graphql/Resolver/ItemResolverFactoryTest.php
```diff
@@ -13,122 +13,87 @@
...
-        $resolver = $mockedItemResolverFactory->createItemResolver('resourceClass', 'rootClass');
+        $resolveInfo = new ResolveInfo([]);
+        $resolveInfo->fieldName = 'name';
+        $resolveInfo->fieldNodes = [];
 
-        $resolveInfoProphecy = $this->prophesize(ResolveInfo::class);
-        $resolveInfoProphecy->fieldName = 'rootProperty';
-
-        $this->assertNull($resolver(null, ['id' => 3], null, $resolveInfoProphecy->reveal()));
+        $this->assertNull($resolver(null, ['id' => '/related_dummies/3'], null, $resolveInfo));
```
soyuka: Should I assume that `['id' => 3]` isn't valid anymore (ie: not using IRIs?)

tests/Graphql/Resolver/ResourceFieldResolverTest.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $iriConverterProphecy = $this->prophesize(IriConverterInterface::class);
+
+        $resolveInfo = new ResolveInfo([]);
+        $resolveInfo->fieldName = 'foo';
+        $resolveInfo->fieldNodes = [];
+
+        $resolver = new ResourceFieldResolver($iriConverterProphecy->reveal());
+        $this->assertEquals('bar', $resolver(['foo' => 'bar'], [], [], $resolveInfo));
+        $this->assertEquals('bar', $resolver((object) ['foo' => 'bar'], [], [], $resolveInfo));
```
soyuka: Okay in fact I think that most of my questions are answered here abouts ids.

src/Graphql/Type/SchemaBuilder.php
dunglas: It is, it trigger an opcode optimization (PHP EA Inspection reports it).

tests/Graphql/Resolver/CollectionResolverFactoryTest.php
dunglas: `id` is mandatory according to the Relay spec.

tests/Graphql/Resolver/CollectionResolverFactoryTest.php
dunglas: It is.

tests/Graphql/Resolver/ItemResolverFactoryTest.php
dunglas: Indeed. IRIs are now mandatory (it's valid according to both the GraphQL and the Relay spec).

src/Graphql/Resolver/CollectionResolverFactory.php
dunglas: Can you point me to the duplicated snippet? And I didn't manage to find it.

src/Graphql/Type/SchemaBuilder.php
soyuka: Okay but we discussed it before and if we add it here shouldn't we put this everywhere? Oh okay so we actually came to the decision that we will haha nvm.

src/Graphql/Resolver/CollectionResolverFactory.php
soyuka: If I get that correctly you want to find the subresources identifiers etc (ie: everything you need for the `subresourceDataprovider->getSubresource` context). 

I'd recommend the use of this: https://github.com/api-platform/core/blob/a5dc79534f19db6359d6fa9b14c1e9d07434392c/src/Operation/Factory/SubresourceOperationFactory.php#L47

It's cached, and we use it in `ApiLoader` and `DocumentationNormalizer` already. In ApiLoader we build a per-operation context: https://github.com/api-platform/core/blob/091865fecf69e9c60bd77b8c6c8eb1c31bce4249/src/Bridge/Symfony/Routing/ApiLoader.php#L112

src/Graphql/Type/SchemaBuilder.php
meyerbaptiste: Why only for the `in_array` function then?

soyuka: https://github.com/Roave/FunctionFQNReplacer

PR #1570, opened 2017-12-18
Always return a float in Paginator::getLastPage()

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n:a
| License       | MIT
| Doc PR        | n/a

Follows #1561.


backbone87:
Why exactly do we use floats insteadof ints for paginations?

dunglas:
To support big numbers, maybe will we change that in 3.0, but it cannot be before.

PR #1577, opened 2017-12-20
Fix null title/description configuration (#1563 followup)

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/1563#discussion_r157848790
| License       | MIT
| Doc PR        | na

ping @chalasr @dunglas @lyrixx

### DIFF
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/ConfigurationTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/ConfigurationTest.php
@@ -234,4 +234,17 @@ class ConfigurationTest extends TestCase
         $this->assertTrue(isset($config['swagger']['api_keys']));
         $this->assertSame($exampleConfig, $config['swagger']['api_keys'][0]);
     }
+
+    /**
+     * Test config for empty title and description.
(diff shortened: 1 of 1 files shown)


lyrixx:
It's useless IMHO. I adds more code for nothing.
If a user wants an empty title: he does not set the key.
If a he wants a title, he add `title: something`. Just adding `title:~` is useless.

About this comment:

> I agree we must allow empty strings or it will be a big BC break (for instance the Flex recipe doesn’t set those keys).

If the key is not present in the YAML, then the value will be de default value and so `''`.

soyuka:
> If the key is not present in the YAML, then the value will be de default value and so ''.

Oh okay we had a doubt on `cannotBeEmpty()` and `defaultValue('')`. WDYT @dunglas?

dunglas:
It would be nice to have a test to be sure :)

dunglas:
Thanks @soyuka

PR #1557, opened 2017-12-08
Add service aliases for autowiring in symfony>3.4

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1556 
| License       | MIT
| Doc PR        | no

PR #1581, opened 2017-12-20
fix error when disabling eager_loading

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ~
| License       | MIT
| Doc PR        | ~

Introduced by d3b690ecb2944717faeea659f2b14692de20b361

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection;
 
 use ApiPlatform\Core\Api\FilterInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\EagerLoadingExtension;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\FilterEagerLoadingExtension;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryCollectionExtensionInterface;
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -412,8 +412,10 @@ private function registerDoctrineExtensionConfiguration(ContainerBuilder $contai
             return;
         }
 
-        $container->removeDefinition('api_platform.doctrine.orm.query_extension.eager_loading');
-        $container->removeDefinition('api_platform.doctrine.orm.query_extension.filter_eager_loading');
+        $container->removeAlias('api_platform.doctrine.orm.query_extension.eager_loading');
```
dunglas: It should be the inverse IIUC.

soyuka: :+1: this is because of https://github.com/api-platform/core/pull/1557 !

You should remove the alias `EagerLoadingExtension::class` (please use this instead of the string) and remove the definition by it's id indeed.

Nice catch!

jdeniau: You are right, I misread the error message, the problem is on the alias. I'm fixing it right now

soyuka:
I'm merging 2.1 there are a bunch of conflicts I'll let you know when ready.

jdeniau:
Well it's fixed @soyuka @dunglas 

Let me know if I need to do something after your merge

soyuka:
@jdeniau you'll just need to rebase so that tests pass, they'll fail because of doctrine

soyuka:
Though you were on master my bad!

PR #1585, opened 2017-12-21
Implement the Relay Global Object Identification Specification

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

Implementation of the GraphQL's [Relay Global Object Identification Specification](https://facebook.github.io/relay/graphql/objectidentification.htm) using hypermedia IRIs.

You can now execute this kind of queries on the GraphQL endpoint:

```graphql
    {
      node(id: "/product/ABC") {
        id
        ... on Product {
          name
        }
      }
    }
```

Note: I'll probably extract the GraphQL type builder in a from the `SchemaBuilder` in a following PR.

### DIFF
--- a/src/Graphql/Type/SchemaBuilder.php
+++ b/src/Graphql/Type/SchemaBuilder.php
@@ -106,7 +106,6 @@ final class SchemaBuilder implements SchemaBuilderInterface
             return $this->graphqlTypes['#node'];
         }
 
-
         return $this->graphqlTypes['#node'] = new InterfaceType([
             'name' => 'Node',
             'description' => 'A node, according to the Relay specification.',
--- a/src/Graphql/Type/SchemaBuilder.php
+++ b/src/Graphql/Type/SchemaBuilder.php
@@ -14,14 +14,17 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Graphql\Type;
 
 use ApiPlatform\Core\Exception\ResourceClassNotFoundException;
-use ApiPlatform\Core\Graphql\Resolver\ResolverFactoryInterface;
+use ApiPlatform\Core\Graphql\Resolver\Factory\ResolverFactoryInterface;
+use ApiPlatform\Core\Graphql\Serializer\ItemNormalizer;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Util\ClassInfoTrait;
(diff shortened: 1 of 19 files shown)


src/Graphql/Type/SchemaBuilder.php
```diff
@@ -40,47 +42,49 @@
...
+        $this->itemResolver = $itemResolver;
         $this->itemMutationResolverFactory = $itemMutationResolverFactory;
         $this->defaultFieldResolver = $defaultFieldResolver;
         $this->paginationEnabled = $paginationEnabled;
     }
 
     public function getSchema(): Schema
     {
-        $queryFields = [];
+        $nodeInterface = $this->getNodeInterface();
```
alanpoulain: Why not storing the type in `$graphqlTypes` instead of passing it in all the methods?

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -96,20 +100,54 @@ public function getSchema(): Schema
...
+            'name' => 'Node',
+            'description' => 'A node, according to the Relay specification.',
+            'fields' => [
+                'id' => [
+                    'type' => GraphQLType::nonNull(GraphQLType::id()),
+                    'description' => 'The id of this node.',
+                ],
+            ],
+            'resolveType' => function ($value) {
+                if (!isset($value['#item'])) {
```
alanpoulain: Maybe `#item` should be a const.

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -179,17 +226,20 @@ private function getResourceFieldConfiguration(string $fieldDescription = null,
                 'resolve' => $resolve,
             ];
         } catch (InvalidTypeException $e) {
-            return null;
+            // return null
```
alanpoulain: Not deleting it?

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -209,55 +263,66 @@ private function convertType(Type $type, bool $isInput = false, bool $isMutation
...
     }
 
     /**
      * Gets the fields of the type of the given resource.
      */
-    private function getResourceObjectTypeFields(string $resource, bool $isInput = false, bool $isMutation = false, string $mutationName = null): array
+    private function getResourceObjectTypeFields(string $resource, bool $input = false, string $mutationName = null): array
     {
         $fields = [];
         $idField = ['type' => GraphQLType::id()];
```
dkozickis: This should be wrapped by GraphQLType::nonNull() as it implements Relay Node interface now, and Node.id required to be nonNull. Commenting here as I'm not sure how much you have progressed after this commit, will submit PR if required. Thank you for all your work!

dunglas: This has not be fixed. Can I let you do the PR? Thanks for the review!

dkozickis: @dunglas saw you already did that, though it would be faster for you anyway, as you were still working in that file :) Thank you for all your work once again! Waiting for your blog entry on GraphQL :) have a good weekend 🎉

PR #1597, opened 2017-12-22
Implement the Relay specification for mutations

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

Implement the [Relay Input Object Mutations Specification](https://facebook.github.io/relay/graphql/mutations.htm).

When this PR will be merged, we'll respect 100% of the Relay spec!

### DIFF
--- a/features/graphql/mutation.feature
+++ b/features/graphql/mutation.feature
@@ -27,21 +27,22 @@ Feature: GraphQL mutation support
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "data.__type.fields[0].name" should contain "delete"
-    And the JSON node "data.__type.fields[0].description" should contain "Deletes "
-    And the JSON node "data.__type.fields[0].type.name" should contain "DeleteMutation"
+    And the JSON node "data.__type.fields[0].description" should match '/^Deletes a [a-zA-Z]+\.$/'
+    And the JSON node "data.__type.fields[0].type.name" should match "/^delete[a-zA-Z]+Payload$/"
     And the JSON node "data.__type.fields[0].type.kind" should be equal to "OBJECT"
     And the JSON node "data.__type.fields[0].args[0].name" should be equal to "input"
-    And the JSON node "data.__type.fields[0].args[0].type.name" should contain "InputDeleteMutation"
+    And the JSON node "data.__type.fields[0].args[0].type.name" should match "/^delete[a-zA-Z]+Input$/"
     And the JSON node "data.__type.fields[0].args[0].type.kind" should be equal to "INPUT_OBJECT"
 
   Scenario: Create an item
     When I send the following GraphQL request:
     """
     mutation {
-      createFoo(input: {name: "A new one", bar: "new"}) {
-        id,
-        name,
+      createFoo(input: {name: "A new one", bar: "new", clientMutationId: "myId"}) {
+        id
+        name
         bar
+        clientMutationId
       }
     }
     """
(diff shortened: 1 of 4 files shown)


src/Graphql/Type/SchemaBuilder.php
```diff
@@ -294,16 +294,18 @@ private function convertType(Type $type, bool $input = false, string $mutationNa
...
+            $shortName = $mutationName.ucfirst($shortName);
         }
-
-        if (isset($this->graphqlTypes[$resourceClass][$mutationName][$input])) {
-            return $this->graphqlTypes[$resourceClass][$mutationName][$input];
+        if ($input) {
+            $shortName .= 'Input';
+        } elseif (null !== $mutationName) {
+            $shortName .= 'Payload';
         }
```
soyuka: May this be shorter using `sprintf`?

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -326,9 +328,13 @@ private function getResourceObjectTypeFields(string $resource, bool $input = fal
     {
         $fields = [];
         $idField = ['type' => GraphQLType::id()];
+        $clientMutationId = GraphQLType::nonNull(GraphQLType::string());
```
alanpoulain: clientMutationId can be null: https://facebook.github.io/relay/graphql/mutations.htm#sec-Mutation-inputs

features/graphql/mutation.feature
```diff
@@ -27,21 +27,22 @@ Feature: GraphQL mutation support
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "data.__type.fields[0].name" should contain "delete"
-    And the JSON node "data.__type.fields[0].description" should contain "Deletes "
-    And the JSON node "data.__type.fields[0].type.name" should contain "DeleteMutation"
+    And the JSON node "data.__type.fields[0].description" should match '/^Deletes a [a-zA-Z]+\.$/'
```
vincentchalamon: Regex could be optimized as `[A-z]`

src/Graphql/Type/SchemaBuilder.php
vincentchalamon: Not sure it will be more clear as `$shortName` is built from conditions

src/Graphql/Type/SchemaBuilder.php
```diff
@@ -294,16 +294,18 @@ private function convertType(Type $type, bool $input = false, string $mutationNa
...
+        $shortName = $resourceMetadata->getShortName();
         if (null !== $mutationName) {
-            $shortName .= ucfirst($mutationName).'Mutation';
+            $shortName = $mutationName.ucfirst($shortName);
         }
-
-        if (isset($this->graphqlTypes[$resourceClass][$mutationName][$input])) {
-            return $this->graphqlTypes[$resourceClass][$mutationName][$input];
+        if ($input) {
+            $shortName .= 'Input';
```
vincentchalamon: What if `$mutationName` is null but `$input` is true? You will concatenate a string on a non-existing variable. But could this situation happen?

src/Graphql/Type/SchemaBuilder.php
dunglas: It can also be not null:

> That input object type may contain an argument named clientMutationId. If provided, that argument must be a String. That argument may be non‐null.

And it is in their official examples: https://facebook.github.io/relay/docs/en/graphql-server-specification.html#mutations

I propose to be strict by default (= to keep this as is).

src/Graphql/Type/SchemaBuilder.php
dunglas: `$shortName` always exists.

src/Graphql/Type/SchemaBuilder.php
dunglas: Indeed, `sprinf` doesn't help.

features/graphql/mutation.feature
dunglas: done

Issue #1590, opened 2017-12-21
Make it easier to use simple DTO


lyrixx:
Hello.

I have one simple need: A user should be able to call an endpoint to reset his password.
(Note: I could have been something totally different, but let's stick to this use case)

I have tried many thing to be the closest possible to API Platform. I mean, I want to use all the feature
of API Platform: de-serialization, validation, error management, documation, etc.

So I end up to write the following code that works, but I think we can do better.

A simple DTO, with validation and declared as an ApiRessource

```php
<?php

namespace AppBundle\Api\Dto;

use ApiPlatform\Core\Annotation\ApiResource;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * @ApiResource(
 *      collectionOperations={
 *          "post"={"route_name"="api_users_forgot_password_request"},
 *      },
 *      itemOperations={},
 * )
 */
class ForgotPasswordRequest
{
    /**
     * @Assert\NotBlank()
     * @Assert\Email()
     */
    public $email;
}

```

Then, I had to write the associated controller: 

```php
<?php

namespace AppBundle\Controller\Api2;

use AppBundle\Api\Dto\ForgotPasswordRequest;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
use Symfony\Bundle\FrameworkBundle\Controller\Controller;
use Symfony\Component\HttpFoundation\JsonResponse;
use Symfony\Component\HttpFoundation\Request;

class ForgotPasswordRequestController extends Controller
{
    /**
     * @Route(
     *     name="api_users_forgot_password_request",
     *     path="/users/forgot-password-request",
     *     defaults={"_api_resource_class"=ForgotPasswordRequest::class, "_api_collection_operation_name"="special"}
     * )
     * @Method("POST")
     */
    public function requestAction(ForgotPasswordRequest $data)
    {
        $errors = $this->get('validator')->validate($data);

        if (count($errors) > 0) {
            return $errors;
        }

        // My custom logic here

        return new JsonResponse(null, 204);
    }
}
```

---

So basically it works but:

* I had hard time to find a working solution. I tried with some events without success (always an issue because it's not a real entity) and finally with a controller. It was not easy to find the custom attribute to use. I had to open all listener to understand how it really works.
* I had to validate the entity by myself. But there is a ValidatorListener. But it is ran after the controller. Why ? cf #1538
[...]

PR #1604, opened 2017-12-25
Add validation for GraphQL mutations

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Validate data during GraphQL mutations. Also includes a refactoring of the validation system.

TODO:

* [x] Add unit tests

### DIFF
--- a/src/Bridge/Symfony/Validator/Validator.php
+++ b/src/Bridge/Symfony/Validator/Validator.php
@@ -35,10 +35,7 @@ class Validator implements ValidatorInterface
     }
 
     /**
-     * Validates an item.
-     *
-     * @param object $data
-     * @param array  $context
+     * {@inheritdoc}
      */
     public function validate($data, array $context = [])
     {
--- a/src/Bridge/Symfony/Validator/Validator.php
+++ b/src/Bridge/Symfony/Validator/Validator.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Validator;
+
+use ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException;
+use ApiPlatform\Core\Validator\ValidatorInterface;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface as SymfonyValidatorInterface;
+
+/**
+ * Validates an item using the Symfony validator component.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class Validator implements ValidatorInterface
+{
+    private $validator;
+    private $container;
+
+    public function __construct(SymfonyValidatorInterface $validator, ContainerInterface $container = null)
+    {
+        $this->validator = $validator;
+        $this->container = $container;
+    }
+
+    /**
+     * Validates an item.
+     *
+     * @param object $data
+     * @param array  $context
+     */
+    public function validate($data, array $context = [])
+    {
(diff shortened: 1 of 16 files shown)


src/Validator/ValidatorInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+interface ValidatorInterface
+{
+    /**
+     * Validates an item.
+     *
+     * @param object $data
+     * @param array  $contexy
```
alanpoulain: $context

src/Bridge/Symfony/Validator/Validator.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->validator = $validator;
+        $this->container = $container;
+    }
+
+    /**
+     * Validates an item.
+     *
+     * @param object $data
+     * @param array  $context
```
alanpoulain: Missing `@throws`?

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -98,4 +104,24 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
+
+    /**
+     * @param object $item
+     *
+     * @throws Error
+     */
+    private function validate($item, ResolveInfo $info, ResourceMetadata $resourceMetadata, string $operationName = null)
+    {
+        if (null === $this->validator) {
+            return;
```
alanpoulain: We don't want to warn the user if there is no validator?

dunglas: No (we would just make this property not nullable otherwise), it allows to create micro-frameworks on top of API Platform and it mimics the behavior of the `ValidateListener` (which is optional).

tests/Bridge/Symfony/Validator/ValidatorTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $symfonyValidator = $symfonyValidatorProphecy->reveal();
+
+        $validator = new Validator($symfonyValidator);
+        $validator->validate(new DummyEntity());
+    }
+
+    /**
+     * @expectedException \ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException
+     */
+    public function testInvalid()
```
alanpoulain: Use a common method?

dunglas: We wouldn't be able to share much code.

src/Validator/EventListener/ValidateListener.php
```diff
@@ -0,0 +1,67 @@
...
+        if (
+            $request->isMethodSafe(false)
+            || $request->isMethod(Request::METHOD_DELETE)
+            || !($attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !$attributes['receive']
+        ) {
+            return;
+        }
+
+        $data = $event->getControllerResult();
```
Koc: @dunglas consider validating controller argument `data` instead of controller result. We cann't put some logic inside controller now because request data is not validated yet.

dunglas: I'm not sure to understand, the validation occurs after the execution of the controller.

dunglas: If you need to validate the data before executing your logic, just inject `ValidatorInterface` in your controller and call it.

alanpoulain: I think it's related to this: https://github.com/api-platform/core/issues/1590

Koc: yes, @alanpoulain is right. IMHO validating input data before controller execution is much logical.

dunglas: Changing this would be a big BC break and will provide less flexibility.

theofidry:
👏

Issue #1442, opened 2017-10-17
Invalid Doctrine Query on Nested Subresources


mab05k:
I have a hierarchy of resources with OneToMany relationships between all: Large => Medium => Small => Tiny and I apply "subresource" annotations to all of them. This configuration creates all combinations of routes:

```
"Collection Operations"
/api/larges
/api/larges/{id}/mediums
/api/large/{id}/mediums/{mediums}/smalls
/api/large/{id}/mediums/{mediums}/smalls/{smalls}/tinies
/api/mediums
/api/mediums/{id}/smalls
/api/mediums/{id}/smalls/{smalls}/tinies
/api/smalls
/api/smalls/{id}/tinies
```

These extended route hierarchies return 500 errors.  The problem occurs when API Platform recursively creates the subqueries for these relationships. 

For example: /api/large/{id}/mediums/{mediums}/smalls/{smalls}/tinies creates the following query:

```mysql
SELECT o 
FROM AppBundle\\Entity\\Tiny o 
WHERE 
	o.smalls 
IN (
	SELECT smalls_a1 
	FROM AppBundle\\Entity\\Small smalls_a1 
	WHERE small_a1.id = :id_p1 AND smalls_a1.medium 
	IN(
		SELECT mediums_a2 
		FROM AppBundle\\Entity\\Medium mediums_a2 
		WHERE mediums_a2.id = :id_p2
         ) 

	AND mediums_a2.large 
	IN(
		SELECT id_a3 
		FROM AppBundle\\Entity\\Large id_a3 
		WHERE id_a3.id = :id_p3
        )
);
```

The query SHOULD look like this:
```mysql
SELECT o 
FROM AppBundle\\Entity\\Tiny o 
WHERE 
	o.smalls 
IN (
	SELECT smalls_a1 
	FROM AppBundle\\Entity\\Small smalls_a1 
	WHERE small_a1.id = :id_p1 AND smalls_a1.medium 
	IN(
		SELECT mediums_a2 
		FROM AppBundle\\Entity\\Medium mediums_a2 
		WHERE mediums_a2.id = :id_p2 AND mediums_a2.large 
        	IN(
	        	SELECT id_a3 
	        	FROM AppBundle\\Entity\\Large id_a3 
	        	WHERE id_a3.id = :id_p3
                )
         ) 
);
```

[...]

PR #1608, opened 2017-12-26
Fix Doctrine query for nested subresources

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1442
| License       | MIT
| Doc PR        |

Fix the build of the Doctrine query for subresources nested in more than 2 levels of depth.
Use a recursive function instead of a while loop.

### DIFF
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -28,6 +28,7 @@ use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInte
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\ORM\QueryBuilder;
 
 /**
  * Subresource data provider for the Doctrine ORM.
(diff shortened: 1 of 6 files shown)


tests/Fixtures/TestBundle/Entity/FourthLevel.php
```diff
@@ -0,0 +1,70 @@
...
+ * @ORM\Entity
+ */
+class FourthLevel
+{
+    /**
+     * @var int The id
+     *
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
```
soyuka: not needed

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -79,82 +80,10 @@ public function getSubresource(string $resourceClass, array $identifiers, array
...
-                $qb->andWhere("$alias.$key = :$placeholder");
-                $queryBuilder->setParameter($placeholder, $value);
-            }
-
-            // recurse queries
-            if (null === $previousQueryBuilder) {
-                $previousQueryBuilder = $qb;
-            } else {
-                $previousQueryBuilder->andWhere($qb->expr()->in($previousAlias, $qb->getDQL()));
-            }
```
soyuka: this is recursive already no? Not sure what was going on wrong here 😸

alanpoulain: No, this was not recursive (despite the comment!).

soyuka: Should've been ? 😄

alanpoulain: Yup! :D

tests/Fixtures/TestBundle/Entity/FourthLevel.php
alanpoulain: This is needed because of https://github.com/alanpoulain/core/blob/3ef7c9e556e03552597db393d5c52adc59675053/features/main/subresource.feature#L54

soyuka: Thought `strategy="AUTO"` was the default value.

alanpoulain: Oh you meant that, OK!

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -195,4 +122,77 @@ public function getSubresource(string $resourceClass, array $identifiers, array
 
         return $context['collection'] ? $query->getResult() : $query->getOneOrNullResult();
     }
+
+    /**
+     * @throws RuntimeException
+     */
+    private function buildQuery(array $identifiers, array $context, QueryNameGenerator $queryNameGenerator, QueryBuilder $previousQueryBuilder, string $previousAlias, int $remainingIdentifiers, QueryBuilder $topQueryBuilder = null): QueryBuilder
+    {
+        if (!$remainingIdentifiers) {
```
soyuka: `0 < --$remainingIdentifiers` then we can avoid `--` below and the `$remainingIdentifiers - 1` operation?

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -195,4 +122,77 @@ public function getSubresource(string $resourceClass, array $identifiers, array
...
+
+        switch ($relationType) {
+            // MANY_TO_MANY relations need an explicit join so that the identifier part can be retrieved
+            case ClassMetadataInfo::MANY_TO_MANY:
+                $joinAlias = $queryNameGenerator->generateJoinAlias($previousAssociationProperty);
+
+                $qb->select($joinAlias)
+                    ->from($identifierResourceClass, $alias)
+                    ->innerJoin("$alias.$previousAssociationProperty", $joinAlias);
+
```
soyuka: for cosmetics, new line is useless here imo \o/

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
alanpoulain: I think it's clearer this way. Because if `--` is done here it's more the position of the identifier and not the remaining ones. And you have to add the operation `$remainingIdentifiers + 1` where there is `$remainingIdentifiers`. WDYT?

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
alanpoulain: Sure :)

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
soyuka: ok then maybe `0 === $remainingIdentifiers` instead of `!`?

alanpoulain: If you prefer. Then not `0 >= $remainingIdentifiers`?

alanpoulain: Or even `$remainingIdentifiers <= 0` (I don't really like Yoda style for comparison)?

soyuka: Dunno it's fine like this as well.

dunglas:
Thanks @alanpoulain

Issue #1567, opened 2017-12-16
ManyToOne and nullable=false produce required:false in hydra response


hariszukanovic:
I expected that setting nullable=false on a ManyToOne would mark that property in hydra docs.jsonld response as required: true
Instead, I am getting required:false

Is there a correct way to mark a ManyToOne required in hydra response? 
Maybe a bug?

Entity relation:
```php
    /**
     * @ORM\ManyToOne(targetEntity="Node", inversedBy="services")
     * @ORM\JoinColumn(name="node_id", referencedColumnName="id", nullable=false, onDelete="RESTRICT")
     * @Groups({"service_list"})
     */
    private $node;
```
Hydra (part) of docs.jsonld response for the property:
```json
{
"@type": "hydra:SupportedProperty",
"hydra:property": {
"@id": "#SupervisorService/node",
"@type": "hydra:Link",
"rdfs:label": "node",
"domain": "#SupervisorService",
"owl:maxCardinality": 1,
"range": "#Node"
},
"hydra:title": "node",
**"hydra:required": false,**
"hydra:readable": true,
"hydra:writable": true
}
```
DB foreign key looks correct:
```
+---------------------------+----------------------------------+------+-----+---------+----------------+
| Field                     | Type                             | Null | Key | Default | Extra          |
+---------------------------+----------------------------------+------+-----+---------+----------------+
| node_id                   | int(11)                          | NO   | MUL | NULL    |                |
```

hariszukanovic:
It is sometimes difficult to understand which annotations affect which behaviors.
I found that adding
* @Assert\NotNull()
produces
"hydra:required": true

If you feel that nullable=false on the ManyToOne should not make the property required in the same way as adding the validator @Assert\NotNull() then please close the ticket.

soyuka:
IMO it's a non-existant feature, it can be fixed here:

https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php#L58

(Look for the property in the metadata and call `withRequired(!$metadata->nullable)`.

Simperfit:
As said in the PR, this will be closed as won't fix.

PR #1610, opened 2017-12-27
Set required Hydra property when relation is not nullable.

raoulclais:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1567
| License       | MIT
| Doc PR        |

Implement #1567 by checking if the relation is required.

### DIFF
--- a/tests/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactoryTest.php
+++ b/tests/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactoryTest.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\Bridge\Doctrine\Orm\Metadata\Property\DoctrineOrmPropertyMe
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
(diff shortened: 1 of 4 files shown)


tests/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactoryTest.php
```diff
@@ -59,10 +60,11 @@ public function testCreateNoClassMetadata()
         $this->assertEquals($doctrineOrmPropertyMetadataFactory->create(Dummy::class, 'id', []), $propertyMetadata);
     }
 
-    public function testCreateIsIdentifier()
+    public function testCreateIsIdentifierNotRequired()
     {
         $propertyMetadata = new PropertyMetadata();
         $propertyMetadata = $propertyMetadata->withIdentifier(true);
+        $propertyMetadata = $propertyMetadata->withRequired(false);
```
soyuka: Bu definition an identifier is required but I think I get what you did :p.

dunglas:
Shouldn't required properties be marked as `@NotNull` or `@NotBlank` with the validator? If we go this way, we must also mark properties with `nullable=false` as required... and enforce it somewhere because it's not the case yet (it's the responsibility of the validator, not of the ORM in a typical Symfony app).

dunglas:
I'm for closing this one as "won't fix". Ping @api-platform/core-team.

soyuka:
Also if you really need the feature, you can always add a custom MetadataPropertyFactory right?

Issue #1616, opened 2017-12-28
(de)normalization_context is not used for subresourceOperations


lyrixx:
According to the example in the doc, the following snippet do not work:

```php

/**
 * @ORM\Entity()
 * @ApiResource(
 *      subresourceOperations={
 *          "answer_get_subresource"= {
 *              "method"="GET",
 *              "path"="/questions/{id}/all-answers",
 *              "normalization_context"={"groups"={"answer_read"}},
 *          },
 *      },
 * )
 */
class Question
```

Reprocuder: https://github.com/lyrixx/test/tree/api-platform-subresource-path

soyuka:
Mhh, I think that it's taking the "normalization_context" of the required resource class, not the class where the actual subresource is declared. I'll comment in the PR.

When I implemented subresources at first, there was no `SubresourceOperation` (it was a CollectionOperation).

lyrixx:
> Mhh, I think that it's taking the "normalization_context" of the required resource class, not the class where the actual subresource is declared. I'll comment in the PR.

Yes, I noticed that. But there are no way to that in a clean way. If I add a new collectionOperationType, It adds a new route :/ IMHO, this way is much more easier and intuitive.

abysKri:
Hi,
The route (/questions/{id}/all-answers) appear in the answer swagger ui too. How to hide it ?
Thanks

lyrixx:
I opened a pr for that too :-D

Nek-:
This is not exacly the same issue but is very close. I tried to use `allow_extra_attributes` context option and this doesn't work. Only the documentation example works for the context options (which is "groups").

```php
/**
 * @ORM\Entity(repositoryClass="App\Repository\ArticleRepository")
 * @ApiResource(
 *     attributes={"denormalization_context"={"allow_extra_attributes"=false}}
 *     collectionOperations={
 *          "get",
 *          "post"={"denormalization_context"={"groups"={"post"}}}
 *     },
 *     itemOperations={
 *          "get",
 *          "put"={"denormalization_context"={"groups"={"put"}, "allow_extra_attributes"=false}}
 *     }
 * )
 */
class Article {}
```

The fun part is the group I specified on "PUT" is ok, but the "allow_extra_attributes" context key is ignored.

soyuka:
Closing in favor of https://github.com/api-platform/core/issues/2706

PR #1617, opened 2017-12-28
[Subresource] Fixed the serialization context builder

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | ???
| Fixed tickets | #1616
| License       | MIT
| Doc PR        |

Before going further, I would like to know if this is the right direction.

### DIFF
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -111,6 +111,8 @@ final class ApiLoader extends Loader
                             'property' => $operation['property'],
                             'identifiers' => $operation['identifiers'],
                             'collection' => $operation['collection'],
+                            'parent_resource_class' => $resourceClass,
+                            'parent_operation_name' => $operation['operation_name'],
                             'operationId' => $operationId,
                         ],
                     ] + $operation['defaults'] ?? [],
--- a/src/Serializer/SerializerContextBuilder.php
+++ b/src/Serializer/SerializerContextBuilder.php
@@ -42,36 +42,43 @@ final class SerializerContextBuilder implements SerializerContextBuilderInterfac
             throw new RuntimeException('Request attributes are not valid.');
         }
 
-        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
         $key = $normalization ? 'normalization_context' : 'denormalization_context';
 
-        $operationKey = null;
-        $operationType = null;
-
         if (isset($attributes['collection_operation_name'])) {
-            $operationKey = 'collection_operation_name';
-            $operationType = OperationType::COLLECTION;
-        } elseif (isset($attributes['subresource_operation_name'])) {
-            $operationKey = 'subresource_operation_name';
-            $operationType = OperationType::SUBRESOURCE;
-        }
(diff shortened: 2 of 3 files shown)


src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -111,6 +111,8 @@ public function load($data, $type = null): RouteCollection
                             'property' => $operation['property'],
                             'identifiers' => $operation['identifiers'],
                             'collection' => $operation['collection'],
+                            'parent_resource_class' => $resourceClass,
```
soyuka: We can't say that it's the "parent" because there can be an "infinity" IMO.
Anyway, we need to discuss here is whether we take the "Root" as the metadata entrypoint or the "Queue" (last).
Explanation:

```
/foo/1/bars/2
```

This path will get you an entity of type `Bar`. When this is serialized, where should we look to get the `normalization_context`?
- `Bar` metadata, after all this is a `Bar` response
- `Foo` metadata, because it's a subresource

We need to choose one. At the moment, IIRC the code looks on the `Bar` metadata to find the `normalization_context` for the given operation. 

If we want to take the "Root" (because this is where we declared the subresource), I don't think that you need the "parent_resource_class".
Take a look at the SubresourceOperationFactory: 

https://github.com/api-platform/core/blob/a5dc79534f19db6359d6fa9b14c1e9d07434392c/tests/Operation/Factory/SubresourceOperationFactoryTest.php#L84

`$rootClass = $identifiers[0][1]` is where you want to look at.
the `operation_name` should be the correct one IMO

soyuka:
Anyway I think there are edge cases that may be hard to cover. I haven't slept enough to give my more detailled opinion on this haha but I'll think about it.

src/Bridge/Symfony/Routing/ApiLoader.php
lyrixx: I vote for:
> Foo metadata, because it's a subresource

---

> $rootClass = $identifiers[0][1] is where you want to look at.

yes I noticed that. But I thought it was risky to use that ;)

lyrixx:
No problem. But for now (and for me) the subresource system is not usable yet :/ as I can not control what the API returns.

src/Bridge/Symfony/Routing/ApiLoader.php
soyuka: > yes I noticed that. But I thought it was risky to use that ;)

It is not and it's meant for that ;)

lyrixx:
So what do you think of this ? Is it the right approach?
I'm adding more sub resource in my API, and right now it's a nightmare to configure it. It's really hard to find where to put the configuration + how to name the operation. Why this PR, everything become much more simple & natural

soyuka:
Having the `RootClass` as a base metadata for the context seems indeed better.

Though, you don't need `parent_resource_class` and you should use `$identifiers[0][1]` as stated above.

lyrixx:
Sure. I will update the pr according to

soyuka:
Closing in favor of https://github.com/api-platform/core/issues/2706

PR #1619, opened 2017-12-28
GraphQL + data provider refactoring: automatically add SQL join clauses

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

This PR is mainly a refactoring of the Data Providers to get rid of the dependency to HttpFoundation (it will also allow to simplify the JSONAPI support, and maybe to support PSR-7 later).
Thanks to this new decoupling, it's now possible to automatically add appropriate `JOIN` SQL clauses according to the selected fields in a GraphQL query.

TODO:

* [x] Fix tests and bugs

### DIFF
--- a/src/Bridge/Doctrine/Orm/CollectionDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/CollectionDataProvider.php
@@ -42,7 +42,7 @@ class CollectionDataProvider implements ContextAwareCollectionDataProviderInterf
         $this->collectionExtensions = $collectionExtensions;
     }
 
-    public function supports(string $resourceClass, string $operationName = null, array $conext = []): bool
+    public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
     {
         return null !== $this->managerRegistry->getManagerForClass($resourceClass);
     }
--- a/src/Bridge/Doctrine/Orm/CollectionDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/CollectionDataProvider.php
@@ -13,10 +13,11 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
 
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\ContextAwareQueryCollectionExtensionInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryCollectionExtensionInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryResultCollectionExtensionInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGenerator;
-use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
+use ApiPlatform\Core\DataProvider\ContextAwareCollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
 use ApiPlatform\Core\Exception\RuntimeException;
 use Doctrine\Common\Persistence\ManagerRegistry;
@@ -27,14 +28,13 @@ use Doctrine\Common\Persistence\ManagerRegistry;
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Samuel ROZE <samuel.roze@gmail.com>
  */
-class CollectionDataProvider implements CollectionDataProviderInterface, RestrictedDataProviderInterface
+class CollectionDataProvider implements ContextAwareCollectionDataProviderInterface, RestrictedDataProviderInterface
 {
     private $managerRegistry;
     private $collectionExtensions;
 
     /**
-     * @param ManagerRegistry                     $managerRegistry
-     * @param QueryCollectionExtensionInterface[] $collectionExtensions
+     * @param QueryCollectionExtensionInterface[]|ContextAwareQueryCollectionExtensionInterface[] $collectionExtensions
      */
     public function __construct(ManagerRegistry $managerRegistry, array $collectionExtensions = [])
     {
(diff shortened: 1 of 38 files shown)


src/Bridge/Doctrine/Orm/CollectionDataProvider.php
```diff
@@ -27,22 +28,21 @@
...
+     * @param QueryCollectionExtensionInterface[]|ContextAwareQueryCollectionExtensionInterface[] $collectionExtensions
      */
     public function __construct(ManagerRegistry $managerRegistry, array $collectionExtensions = [])
     {
         $this->managerRegistry = $managerRegistry;
         $this->collectionExtensions = $collectionExtensions;
     }
 
-    public function supports(string $resourceClass, string $operationName = null): bool
+    public function supports(string $resourceClass, string $operationName = null, array $conext = []): bool
```
vincentchalamon: Didn't you mean `$context`?

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -67,17 +72,22 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
     /**
      * {@inheritdoc}
      */
-    public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
+    public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $options = null === $operationName ? [] : ['collection_operation_name' => $operationName];
+        if (null === $resourceClass) {
```
vincentchalamon: What if `$resourceClass` is an empty string? I would prefer `if (empty($resourceClass))`
Same in all PR

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -53,6 +54,10 @@
      */
     public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null, bool $fetchPartial = false, ClassMetadataFactoryInterface $classMetadataFactory = null)
     {
+        if (null !== $this->serializerContextBuilder) {
+            @trigger_error('Passing an instance of "%s" is deprecated since version 2.2 and will be removed in 3.0. Use the "normalization_context" of the data provider\'s context instead.', E_USER_DEPRECATED);
```
vincentchalamon: Missing `sprintf` & argument

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: It would be a BC break (`''` is currently allowed).

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -53,6 +54,10 @@
      */
     public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, int $maxJoins = 30, bool $forceEager = true, RequestStack $requestStack = null, SerializerContextBuilderInterface $serializerContextBuilder = null, bool $fetchPartial = false, ClassMetadataFactoryInterface $classMetadataFactory = null)
     {
+        //if (null !== $this->serializerContextBuilder) {
+        //    @trigger_error('Passing an instance of "%s" is deprecated since version 2.2 and will be removed in 3.0. Use the "normalization_context" of the data provider\'s context instead.', E_USER_DEPRECATED);
+        //}
```
soyuka: You want to keep this dead code?

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
```diff
@@ -37,8 +38,12 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
     /**
      * {@inheritdoc}
      */
-    public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
+    public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass = null, string $operationName = null, array $context = [])
     {
+        if (null === $resourceClass) {
+            throw new InvalidArgumentException('The "$resourceClass" parameter must not be null');
+        }
```
soyuka: Weird no? How is this possible and why do we change the interface to allow null? Especially if `supports` method does require this argument can't it be mandatory?

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: No I'll drop it, I try to debug something.

src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
dunglas: It's only possible to defined **optional** parameters when overriding the signature of method defined in a parent interface. It's why.

soyuka: Sure. Once the parent interface argument becomes mandatory we will remove this test right?

dunglas: Indeed, but I'm not sure it is worth it to make the argument mandatory in a future version (pain to upgrade for almost no gain).

PR #1625, opened 2018-01-03
Allow updating nested resource on POST

### LINKED ISSUE #1624: POST object and update nested data
I saw there are two issues opened (#1363 and #427 ) about the same issue, but they are both mixing two different issues, one about plain identifiers and the other one with a 404 issue, so I extract the issue in here to be clear.

If I try to POST an object and in the same time UPDATE a nested data, api-platform returns a 400 with description: `Update is not allowed for this operation.`

```json
POST /orders
{
    "cart": {
        "@id": "/carts/1",
        "description": "foo"
    }
}
```

This code "should" update the cart with id 1 and set the description to "foo", then create an order for cart with id 1.

It is due to [ItemNormalizer.php#L100-102](https://github.com/api-platform/core/blob/master/src/JsonLd/Serializer/ItemNormalizer.php#L100-L103)
I understand the purpose of theses lines to avoid updating a "main" resource, but isn't it a bug to "block" updating a nested resource ? 

PR will follow if you are OK with this.

Thanks

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1624, #1363, #427
| License       | MIT
| Doc PR        | ~

Allow updating nested resource on POST

### DIFF
--- a/features/hydra/error.feature
+++ b/features/hydra/error.feature
@@ -85,6 +85,7 @@ Feature: Error handling
     And the JSON node "hydra:description" should exist
     And the JSON node "trace" should exist
 
+    @dropSchema
     Scenario: Get an error during update of an existing resource with a non-allowed update operation
       When I add "Content-Type" header equal to "application/ld+json"
       And I send a "POST" request to "/dummies" with body:
@@ -103,23 +104,39 @@ Feature: Error handling
       And the JSON node "hydra:description" should be equal to "Update is not allowed for this operation."
       And the JSON node "trace" should exist
 
-    Scenario: Get an error during update of an existing relation with a non-allowed update operation
+    @createSchema
+    Scenario: Populate database with related dummies. Check that id will be "/related_dummies/1"
+      Given I add "Content-Type" header equal to "application/ld+json"
+      And I send a "POST" request to "/related_dummies" with body:
+      """
+      {
+          "@type": "https://schema.org/Product",
+          "symfony": "laravel"
+      }
+      """
+      Then the response status code should be 201
+      And the response should be in JSON
+      And the JSON node "@id" should be equal to "/related_dummies/1"
+      And the JSON node "symfony" should be equal to "laravel"
+
+    @dropSchema
+    Scenario: Do not get an error during update of an existing relation with a non-allowed update operation
       When I add "Content-Type" header equal to "application/ld+json"
       And I send a "POST" request to "/relation_embedders" with body:
       """
       {
         "anotherRelated": {
-          "@id": "/related_dummies/2",
+          "@id": "/related_dummies/1",
           "@type": "https://schema.org/Product",
           "symfony": "phalcon"
         }
       }
       """
-      Then the response status code should be 400
+      Then the response status code should be 201
       And the response should be in JSON
       And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
-      And the JSON node "@context" should be equal to "/contexts/Error"
-      And the JSON node "@type" should be equal to "hydra:Error"
-      And the JSON node "hydra:title" should be equal to "An error occurred"
-      And the JSON node "hydra:description" should be equal to "Update is not allowed for this operation."
-      And the JSON node "trace" should exist
+      And the JSON node "@context" should be equal to "/contexts/RelationEmbedder"
+      And the JSON node "@type" should be equal to "RelationEmbedder"
+      And the JSON node "@id" should be equal to "/relation_embedders/1"
+      And the JSON node "anotherRelated.@id" should be equal to "/related_dummies/1"
+      And the JSON node "anotherRelated.symfony" should be equal to "phalcon"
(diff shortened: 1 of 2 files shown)


features/hydra/error.feature
```diff
@@ -103,23 +103,38 @@ Feature: Error handling
...
+          "@type": "https://schema.org/Product",
+          "symfony": "laravel"
+      }
+      """
+      Then the response status code should be 201
+      And the response should be in JSON
+      And the JSON node "@id" should be equal to "/related_dummies/1"
+      And the JSON node "symfony" should be equal to "laravel"
+
       When I add "Content-Type" header equal to "application/ld+json"
```
dunglas: It should be split in several scenarios (a scenario should not have several `When` steps).

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -98,7 +98,9 @@ public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
         if (isset($data['@id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
-            if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
+            if (isset($context['api_allow_update']) && true !== $context['api_allow_update']
```
dunglas: WDYT about `if (!($context['api_allow_update'] ?? false) && !($context['api_resource_is_relation'] ?? false))` ?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -297,6 +297,7 @@ private function denormalizeRelation(string $attributeName, PropertyMetadata $pr
             ($propertyMetadata->isWritableLink() && \is_array($value))
         ) {
             $context['resource_class'] = $className;
+            $context['api_resource_is_relation'] = true;
```
dunglas: Wouldn't be easier to just set `api_allow_update` to true for the subcontext in this case instead of introducing a new flag?

features/hydra/error.feature
jdeniau: Well the first lines are to check that the insertion is correct (mainly that the `@id` is `/related_dummies/1`). I can change the line 109 from `When` to `Given`.

src/JsonLd/Serializer/ItemNormalizer.php
jdeniau: I think you mean `if (!($context['api_allow_update'] ?? true)  ...` because now true is the default value for `api_allow_update` (another way of saying is that an unset api_allow_update wont go if the `if`).
The only difference here is if the value is set but not boolean, but that "should" not append (except if a user set it manually in the serialization context).

I like it but can maybe be hard to read, but why not.

As a matter of fact, `api_allow_update` is always set in the `SerializerContextBuilder` so I'm wondering if someone can denormalize an item without going through the context builder ? 
If no, the code can be `true !== $context['api_allow_update']` (but it may be safer to keep the `isset` check.

src/Serializer/AbstractItemNormalizer.php
jdeniau: Apparently it works, and probably fixes the `src/Serializer/ItemNormalizer.php` that triggers the same error that I missed, I will do this.

features/hydra/error.feature
dunglas: Yes it should follow the format:

```
Given ...
And ...
When ...
And ...
```

You can also split it in two scenarios.
Then

jdeniau: @dunglas done

features/hydra/error.feature
```diff
@@ -103,23 +103,39 @@ Feature: Error handling
       And the JSON node "hydra:description" should be equal to "Update is not allowed for this operation."
       And the JSON node "trace" should exist
 
-    Scenario: Get an error during update of an existing relation with a non-allowed update operation
+    @dropSchema
```
dunglas: You can remove this one, but be sure to add `@dropSchema` on the last scenario to cleanup.

jdeniau: @dunglas I missed the fact that `@dropSchema` is `AfterScenario`.

I took the liberty of adding a new `@recreateSchema` that drop and create the schema, because I don't think relying on what did appear on the previous scenario is a good think. (mainly because someone may add another scenario between.

dunglas: Even if I agree that it's cleaner, rhe problem with recreating the schema often is that is slow down all our builds...
If someone add another scenario between, it's his build that will fail 😄. His responsibility.

Would you mind to just use create/drop?

jdeniau: @dunglas I did that, and opened #1642 to discuss about the createSchema vs recreateSchema

dunglas:
Thanks @jdeniau

PR #1633, opened 2018-01-08
GraphQL: add support for filters

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

This is the last missing part for full GraphQL support (and to release the first beta of 2.2): filters support.

This PR allows to use any API Platform filter (builtin or not) both with REST and GraphQL. It comes with a refactoring of the filter subsystem to be decoupled from the HTTP layer. This also means that it's now now easy to add support for other transport layers (e.g. RabbitMQ), and different HTTP abstractions than Symfony's HttpFoundation (e.g. PSR-7).

Example:

```php
use ApiPlatform\Core\Annotation\ApiFilter;
use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\SearchFilter;

/**
 * @ApiResource
 * @ApiFilter(SearchFilter::class, strategy="partial")
 * @ORM\Entity
 */
class Person
{
    /**
     * @ORM\Column(type="integer")
     * @ORM\Id
     */
    private $id;

    /**
     * @ORM\Column
     */
    private $name;
}
```

Then you can run this GraphQL query:

```graphql
{
  people(name: "Kévin") {
    edges {
      node {
        id
        name
      }
    }
  }
} 
```

It also works for any level of nested collections.

TODO:

* [x] Fix unit tests


features/graphql/filters.feature
```diff
@@ -0,0 +1,94 @@
...
+        }
+      }
+    }
+    """
+    Then the JSON node "data.dummies.edges" should have 1 element
+    And the JSON node "data.dummies.edges[0].node.id" should be equal to "/dummies/2"
+
+  @createSchema
+  @dropSchema
+  Scenario: Retrieve a collection filtered using the date filter
```
alanpoulain: The Scenario does not correspond to what is done.

sroze:
Nice!

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -201,17 +204,43 @@ private function getResourceFieldConfiguration(string $fieldDescription = null,
...
+                    if (!$this->filterLocator->has($filterId)) {
+                        continue;
+                    }
+
+                    foreach ($this->filterLocator->get($filterId)->getDescription($resourceClass) as $key => $value) {
+                        $nullable = isset($value['required']) ? !$value['required'] : true;
+                        $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
+                        $graphqlFilterType = $this->convertType($filterType);
+
+                        if ('[]' === $newKey = substr($key, -2)) {
```
alanpoulain: Shouldn't it be in a dedicated method (in FilterInterface?)? Or maybe a boolean returned by `getDescription` (like `$value['is_property_list']`?

dunglas: Indeed, but has it was done this way for historical reason (and we definitely need this code for existing custom filters), I propose to add the new key in a new PR.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -66,15 +71,8 @@ public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $q
...
-     *
-     * @param string                      $property
-     * @param mixed                       $value
-     * @param QueryBuilder                $queryBuilder
-     * @param QueryNameGeneratorInterface $queryNameGenerator
-     * @param string                      $resourceClass
-     * @param string|null                 $operationName
      */
-    abstract protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null);
+    abstract protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null/*, array $context = []*/);
```
karser: @dunglas The commented out `$context` in `apply` and `filterProperty` is a bit confusing. I see that `AbstractContextAwareFilter` extends from this class and overrides `apply` introducing `array $context = []`,  and then calls `$this->filterProperty(..., $context);`.
How about filterProperty, should it also be overridden in `AbstractContextAwareFilter` with adding `array $context = []`?

soyuka: oh good catch I think it should indeed

teohhanhui: As @meyerbaptiste reminds us, we need PHP 7.2 for this: https://3v4l.org/ktgkR

PR #1636, opened 2018-01-09
fix FilterEagerLoadingExtension do not remove where clause if it did …

soyuka:
…not change

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -62,12 +62,14 @@ final class FilterEagerLoadingExtension implements QueryCollectionExtensionInter
 
         $queryBuilderClone = clone $queryBuilder;
         $queryBuilderClone->resetDQLPart('where');
+        $changedWhereClause = false;
 
         if (!$classMetadata->isIdentifierComposite) {
             $replacementAlias = $queryNameGenerator->generateJoinAlias($originAlias);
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
```diff
@@ -79,9 +81,14 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
                 $in = $this->getQueryBuilderWithNewAliases($queryBuilder, $queryNameGenerator, $originAlias, $replacementAlias);
                 $in->select("IDENTITY($replacementAlias.$identifier)");
                 $queryBuilderClone->andWhere($queryBuilderClone->expr()->in("$originAlias.$identifier", $in->getDQL()));
+                $changedWhereClause = true;
             }
         }
 
+        if (false === $changedWhereClause) {
```
sroze: Can't you just check if the `where` clause of the query builder is empty instead of introducing a new variable?

soyuka: There's already a check for that [here](https://github.com/soyuka/core/blob/776f62d99219c4aa5c2ffa517310d549ee83734e/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L52). This is really an edge case.

PR #1653, opened 2018-01-16
add AccessDeniedException and InsufficientAuthentificationException c…

Rebolon:
…lass to be allowed in the configuration exception_to_status exception_to_status

https://github.com/api-platform/api-platform/issues/519

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/519
| License       | MIT

Following @dunglas recommandation here : https://github.com/api-platform/api-platform/issues/519

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -23,6 +23,8 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Serializer\Exception\ExceptionInterface;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Security\Core\Exception\InsufficientAuthenticationException;
 
 /**
  * The configuration of the bundle.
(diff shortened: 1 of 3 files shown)


Rebolon:
@dunglas i'm quite agree with you, symfony should manage this finely (which is not the cas with json_login system). But, as Symfony security let the user to override the status_code i thought it could be something wished by community of api-platform.

stof:
@dunglas the weird thing is that the firewall should have a listener converting these exception into the HTTP ones if they cannot be handled in another way (redirecting to the entry point for instance). So the right fix would first involve understanding why this conversion does not happen (it might be a bug somewhere)

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -245,6 +247,8 @@ private function addExceptionToStatusSection(ArrayNodeDefinition $rootNode)
                     ->defaultValue([
                         ExceptionInterface::class => Response::HTTP_BAD_REQUEST,
                         InvalidArgumentException::class => Response::HTTP_BAD_REQUEST,
+                        AccessDeniedException::class => Response::HTTP_FORBIDDEN,
+                        InsufficientAuthenticationException::class => Response::HTTP_FORBIDDEN,
```
Taluu: `Response::HTTP_UNAUTHORIZED` ? More like a 401 than 403. He can retry the request other credentials. 403 is really "forbidden, even if you're the all-authority".

Rebolon: Yes i agree with you

Simperfit:
could you please check why the tests are failing ?

Simperfit:
Thanks for your contribution, however I'm closing here since we got no feedback, feel free to re-open when working on it !

Rebolon:
@Simperfit the question is: does the PR has a chance to be accepted with fixed tests ? in fact it seems that the content of the original PR was not the rigth way to go on.

PR #1656, opened 2018-01-17
Update ErrorFormatGuesser to handle custom formats and improve guess…

natepage:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Wanted to implement custom error formats based on the format requested by the client, the issue was that whatever the request format the errors were always in the first error format configured. This PR has for purpose to improve the way the `ErrorFormatGuesser` is guessing the current format from the request and the configuration.

The logic is as following:
1- Check if request format has been set
2- Check if configuration for the format exist and return it if Yes
3- Otherwise guess MIME type from request headers (Accept, Content-Type)
4- Check if configuration for the MIME type exist and return it if Yes
5- Otherwise fallback to first error format configured

Example of configuration:
```yaml
api_platform:
    mapping:
        paths: ['%kernel.project_dir%/src/Database/Models']
    formats:
        json:     ['application/json']
        xml:      ['application/xml', 'text/xml']
    error_formats:
        custom_error_json: ['application/json']
        custom_error_xml:  ['application/xml', 'text/xml']
```

Return table

| Request format | Accept Header | Content-Type Header | Return
| -------------- | -------------- | --------------------- | -------
| / | / | / | `['key' => 'custom_error_xml', 'value' => ['application/xml', 'text/xml']]`
| custom_error_xml | / | / | `['key' => 'custom_error_xml', 'value' => ['application/xml', 'text/xml']]`
| / | application/json | / | `['key' => 'custom_error_json', 'value' => ['application/json']]`
| / | / | text/xml | `['key' => 'custom_error_xml', 'value' => ['application/xml', 'text/xml']]`
| custom_error_xml | application/json | text/xml | `['key' => 'custom_error_xml', 'value' => ['application/xml', 'text/xml']]`

The priority order to guess the format is: Request format > Accept Header > Content-Type Header > Fallback to first error format configured.

### DIFF
--- a/src/Util/ErrorFormatGuesser.php
+++ b/src/Util/ErrorFormatGuesser.php
@@ -46,12 +46,12 @@ final class ErrorFormatGuesser
         $defaultFormat = [];
 
         foreach ($errorFormats as $format => $errorMimeTypes) {
-            if ([] === $defaultFormat) {
-                $defaultFormat = ['key' => $format, 'value' => $errorMimeTypes];
-            }
-
             if (array_intersect($requestMimeTypes, $errorMimeTypes)) {
                 return ['key' => $format, 'value' => $errorMimeTypes];
+            }
+
+            if (!$defaultFormat) {
+                $defaultFormat = ['key' => $format, 'value' => $errorMimeTypes];
             }
         }
 
--- a/src/Util/ErrorFormatGuesser.php
+++ b/src/Util/ErrorFormatGuesser.php
@@ -37,14 +37,35 @@ final class ErrorFormatGuesser
     public static function guessErrorFormat(Request $request, array $errorFormats): array
     {
         $requestFormat = $request->getRequestFormat('');
-        if ('' !== $requestFormat && isset($errorFormats[$requestFormat])) {
+
+        if ('' !== $requestFormat && array_key_exists($requestFormat, $errorFormats)) {
             return ['key' => $requestFormat, 'value' => $errorFormats[$requestFormat]];
         }
 
-        foreach ($errorFormats as $key => $value) {
-            return ['key' => $key, 'value' => $value];
+        $mimeType = self::getMimeType($request);
+
+        foreach ($errorFormats as $format => $mimeTypes) {
+            if (in_array($mimeType, $mimeTypes, true)) {
+                return ['key' => $format, 'value' => $mimeTypes];
+            }
         }
 
-        return [];
+        return ['key' => key($errorFormats), 'value' => reset($errorFormats)];
+    }
+
+    /**
+     * Get MIME type from the request.
+     *
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return string|null
+     */
+    private static function getMimeType(Request $request): ?string
+    {
+        if ($request->headers->has('accept') && '*/*' !== $request->headers->get('accept')) {
+            return $request->headers->get('accept');
+        }
+
+        return $request->getMimeType($request->getContentType());
     }
 }
--- a/src/Util/ErrorFormatGuesser.php
+++ b/src/Util/ErrorFormatGuesser.php
@@ -38,34 +38,23 @@ final class ErrorFormatGuesser
     {
         $requestFormat = $request->getRequestFormat('');
 
-        if ('' !== $requestFormat && array_key_exists($requestFormat, $errorFormats)) {
+        if ('' !== $requestFormat && isset($errorFormats[$requestFormat])) {
             return ['key' => $requestFormat, 'value' => $errorFormats[$requestFormat]];
         }
 
-        $mimeType = self::getMimeType($request);
+        $requestMimeTypes = Request::getMimeTypes($request->getRequestFormat());
+        $defaultFormat = [];
 
-        foreach ($errorFormats as $format => $mimeTypes) {
-            if (in_array($mimeType, $mimeTypes, true)) {
-                return ['key' => $format, 'value' => $mimeTypes];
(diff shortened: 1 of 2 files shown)


natepage:
Travis check failed because their service was down yesterday, can we please restart the tests because it should pass. Thank you.

src/Util/ErrorFormatGuesser.php
```diff
@@ -37,14 +37,35 @@ private function __construct()
     public static function guessErrorFormat(Request $request, array $errorFormats): array
     {
         $requestFormat = $request->getRequestFormat('');
-        if ('' !== $requestFormat && isset($errorFormats[$requestFormat])) {
+
+        if ('' !== $requestFormat && array_key_exists($requestFormat, $errorFormats)) {
```
dunglas: Why this change? (`isset` is faster)

src/Util/ErrorFormatGuesser.php
```diff
@@ -37,14 +37,35 @@ private function __construct()
...
-            return ['key' => $key, 'value' => $value];
+        $mimeType = self::getMimeType($request);
+
+        foreach ($errorFormats as $format => $mimeTypes) {
+            if (in_array($mimeType, $mimeTypes, true)) {
+                return ['key' => $format, 'value' => $mimeTypes];
+            }
+        }
+
+        return ['key' => key($errorFormats), 'value' => reset($errorFormats)];
```
dunglas: Because we already loop over formats, you can just store the first format in a var, line 48:

```php
$requestMimeTypes = $request->getMimeTypes($request->getRequestFormat());
$defaultFormat = [];
foreach ($errorFormats as $format => $errorMimeTypes) {
    if ([] === $defaultFormat) {
        $defaultFormat = ['key' => $format, 'value' => $mimeTypes];
    }

    if (array_intersect($requestMimeTypes, $errorMimeTypes) {
        return ['key' => $format, 'value' => $mimeTypes];
    }
}

return $defaultFormat;
```

It's faster and safer in case no default format is defined.

src/Util/ErrorFormatGuesser.php
```diff
@@ -37,14 +37,35 @@ private function __construct()
...
+    }
+
+    /**
+     * Get MIME type from the request.
+     *
+     * @param \Symfony\Component\HttpFoundation\Request $request
```
dunglas: `@param` and `@return` can be removed because of the typehints.

src/Util/ErrorFormatGuesser.php
```diff
@@ -37,14 +37,35 @@ private function __construct()
...
+     *
+     * @return string|null
+     */
+    private static function getMimeType(Request $request): ?string
```
dunglas: You can remove this method and just call `$request->getMimeTypes($request->getRequestFormat())` because the content negotiation is already properly handled here: https://github.com/api-platform/core/blob/master/src/EventListener/AddFormatListener.php#L67

src/Util/ErrorFormatGuesser.php
```diff
@@ -37,14 +37,35 @@ private function __construct()
...
+        if ('' !== $requestFormat && array_key_exists($requestFormat, $errorFormats)) {
             return ['key' => $requestFormat, 'value' => $errorFormats[$requestFormat]];
         }
 
-        foreach ($errorFormats as $key => $value) {
-            return ['key' => $key, 'value' => $value];
+        $mimeType = self::getMimeType($request);
+
+        foreach ($errorFormats as $format => $mimeTypes) {
+            if (in_array($mimeType, $mimeTypes, true)) {
```
dunglas: Sometimes they are several mime types associated with one formats (it's why you should call $request->getMimeType**s**()). They all must be checked. See my alternative proposal below.

dunglas:
Can you also target the master? It's a new feature to me.

natepage:
@dunglas I've made the changes, thank you for the pieces of code you gave me. I don't know how to target master instead of 2.1, do you want me to create a new PR to master?

dunglas:
@natepage `git rebase master` should do the trick.

meyerbaptiste:
Rebased and RFR @api-platform/core-team.

src/Util/ErrorFormatGuesser.php
```diff
@@ -37,14 +37,24 @@ private function __construct()
...
             return ['key' => $requestFormat, 'value' => $errorFormats[$requestFormat]];
         }
 
-        foreach ($errorFormats as $key => $value) {
-            return ['key' => $key, 'value' => $value];
+        $requestMimeTypes = Request::getMimeTypes($request->getRequestFormat());
+        $defaultFormat = [];
+
+        foreach ($errorFormats as $format => $errorMimeTypes) {
+            if ([] === $defaultFormat) {
```
soyuka: what? this looks weird can we change to empty? or `!$defaultFormat`? Why can't we init above with the proper array?

dunglas: Because we not sure the key 0 exist, it looks ok to me to do this. We indeed may use `!$defaultFormat`.

soyuka: I'd be :+1: for `!$defaultFormat`. Anyway I panicked because `[] === []` is false in javascript :laughing:

meyerbaptiste: Fixed.

dunglas:
Thanks @natepage and @meyerbaptiste!

PR #1658, opened 2018-01-18
Fix crash when operation defaults are missing

curry684:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Given that `??` has lower precedence than `+` ([docs](http://php.net/manual/en/language.operators.precedence.php)), this snippet, that was intended to fall back to an empty array if defaults are missing, instead gave a guaranteed crash as it first tried to add an array and `null` together before coalescing the result.


dunglas:
Good catch, thank you very much!

Issue #1576, opened 2017-12-20
GraphQL - Support for `json` and `json_array` types


c410-f3r:
First of all, thanks for this awesome project!

GraphQL doesn't support any kind of "map" type, therefore, `Column` fields marked as `json` or `json_array` aren't included in the query definition.
Is there a custom scalar type that handles this problem? If not, is it possible to serialize and send these fields as string so the client can handle the conversion logic?

esistgut:
I would like a clarification about the support of the the json field. I did a proof of concept because I'm evaluting api-platform and it seems to return the data inside a json field correctly but it doesn't know about it when printing an example model.
This is a get on my test model:

Response body:
```json
{
  "@context": "/api/contexts/Product",
  "@id": "/api/products",
  "@type": "hydra:Collection",
  "hydra:member": [
    {
      "@id": "/api/products/1",
      "@type": "Product",
      "id": 1,
      "name": "prod 1",
      "attributes": {
        "asd": 12
      },
      "prices": []
    }
  ],
  "hydra:totalItems": 1
}
```
Example model:
```json
[
  {
    "id": 0,
    "name": "string",
    "prices": [
      "string"
    ]
  }
]
```
and this is the source model:
```php
<?php

namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity(repositoryClass="App\Repository\ProductRepository")
 * @ApiResource
 */
class Product
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="string", length=128)
     */
    private $name;

    /**
     * @ORM\Column(type="json")
     */
    private $attributes;

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\Price", mappedBy="product")
     */
    private $prices;

    public function __construct()
    {
        $this->prices = new ArrayCollection();
    }


    public function getId()
    {
        return $this->id;
    }


    public function setId($id): void
    {
        $this->id = $id;
    }


    public function getName()
    {
        return $this->name;
    }


    public function setName($name): void
    {
        $this->name = $name;
    }

    public function getAttributes()
    {
        return $this->attributes;
    }

    public function setAttributes($attributes)
    {
[...]

PR #1661, opened 2018-01-22
[GraphQL] Iterable type

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1576
| License       | MIT
| Doc PR        |

Add an IterableType to manage for example array types in Doctrine (http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#array-types).

TODO:
- [x] Unit tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "symfony/http-foundation": "^3.1 || ^4.0",
         "symfony/http-kernel": "^2.7 || ^3.0 || ^4.0",
         "symfony/property-access": "^2.7 || ^3.0 || ^4.0",
-        "symfony/property-info": "^3.1 || ^4.0",
+        "symfony/property-info": "^3.3.11 || ^4.0",
         "symfony/serializer": "^3.1 || ^4.0",
         "willdurand/negotiation": "^2.0.3"
     },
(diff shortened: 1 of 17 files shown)


src/GraphQl/Type/Definition/ArrayType.php
```diff
@@ -0,0 +1,105 @@
...
+use GraphQL\Utils\Utils;
+
+/**
+ * Represents an array type.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class ArrayType extends ScalarType
```
meyerbaptiste: What do you think of an `IterableType` instead and use `is_iterable()` instead of `is_array()` ?

meyerbaptiste: What do you think of an `IterableType` (`\Traversable` + `array`) for both `Type::BUILTIN_TYPE_ITERABLE` and `Type:: BUILTIN_TYPE_ARRAY` and replace `is_array()` with `is_iterable()` instead?

dunglas:
Thanks @alanpoulain

Issue #1199, opened 2017-06-25
[Question] CORS


VeeeneX:
How can I add `Access-Control-Allow-Origin` to API? The following sentense is stated in docs

_If you are starting a new project, the easiest way to get API Platform up is to install the API Platform Standard Edition. It ships with the API Platform Core library integrated with the Symfony framework, the schema generator, Doctrine ORM, NelmioApiDocBundle, **NelmioCorsBundle** and Behat. Basically, it is a Symfony edition packaged with the best tools to develop a REST API and sensible default settings._

But inside of composer.json
```json
    "require": {
        "php": ">=7.0",

        "doctrine/inflector": "^1.0",
        "psr/cache": "^1.0",
        "psr/container": "^1.0",
        "symfony/http-foundation": "^3.1 || ^4.0",
        "symfony/http-kernel": "^2.7 || ^3.0 || ^4.0",
        "symfony/property-access": "^2.7 || ^3.0 || ^4.0",
        "symfony/property-info": "^3.1 || ^4.0",
        "symfony/serializer": "^3.1 || ^4.0",
        "willdurand/negotiation": "^2.0.3"
    },
    "require-dev": {
        "behat/behat": "^3.1",
        "behat/mink": "^1.7",
        "behat/mink-browserkit-driver": "^1.3.1",
        "behat/mink-extension": "^2.2",
        "behat/symfony2-extension": "^2.1.1",
        "behatch/contexts": "^2.6",
        "doctrine/doctrine-bundle": "^1.6.3",
        "doctrine/orm": "^2.5.2",
        "doctrine/annotations": "^1.2",
        "friendsofsymfony/user-bundle": "^2.0",
        "guzzlehttp/guzzle": "^6.0",
        "nelmio/api-doc-bundle": "^2.11.2",
        "php-mock/php-mock-phpunit": "^1.1",
        "phpdocumentor/reflection-docblock": "^3.0",
        "phpdocumentor/type-resolver": "^0.2",
        "phpunit/phpunit": "^5.6.8",
        "psr/log": "^1.0",
        "sensio/framework-extra-bundle": "^3.0.11 || ^4.0",
        "symfony/asset": "^3.3 || ^4.0",
        "symfony/cache": "^3.3 || ^4.0",
        "symfony/config": "^3.3 || ^4.0",
        "symfony/console": "^3.3 || ^4.0",
        "symfony/debug": "^2.8 || ^3.0 || ^4.0",
[...]

PR #1662, opened 2018-01-23
Add a new suggestion

Saphyel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1199
| License       | MIT
| Doc PR        | none

To avoid people asking about CORS I think is better to suggest use it (or maybe should be part of the require?).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,7 +79,8 @@
         "symfony/expression-language": "To use authorization features.",
         "symfony/security": "To use authorization features.",
         "symfony/twig-bundle": "To use the Swagger UI integration.",
-        "webonyx/graphql-php": "To support GraphQL."
+        "webonyx/graphql-php": "To support GraphQL.",
+        "nelmio/cors-bundle": "To use CORS."
     },
     "autoload": {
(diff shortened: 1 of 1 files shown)


dunglas:
NelmioCors is included in the distribution: https://github.com/api-platform/api-platform/blob/master/api/composer.json#L10

We use suggest to hint about optional features that can be enabled by installing extra packages. This is not the case here. I'm 👎 to add it.

Saphyel:
@dunglas probably people that use symfony flex will end up wondering the same question (new people)... I just wanted to make it easier for them...

dunglas:
I added an official Flex recipe for NelmioCord. Run `composer req cors` And you’re all set.
What we can maybe do instead is adding NelmioCors as a dependency of the `api-platform/api-pack`.

krypton-code:
nelmio/cors-bundle should be included in the api-pack, absolutely!

PR #1663, opened 2018-01-24
Avoid to call OrderFilter on wrong properties

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

On 2.2-beta-1 if a order filter and let say a date filter are declared on the same property (`createdAt`), then calling `/dummies?createdAt[before]=2018-01-24` will trigger the date filter and cause a
`Type error: strtoupper() expects parameter 1 to be string, array given in vendor/api-platform/core/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php:134`

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
@@ -82,6 +82,7 @@ class OrderFilter extends AbstractContextAwareFilter
     public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null, array $context = [])
     {
         if (!isset($context['filters'][$this->orderParameterName]) || !\is_array($context['filters'][$this->orderParameterName])) {
+            $context['filters'] = null;
             parent::apply($queryBuilder, $queryNameGenerator, $resourceClass, $operationName, $context);
 
             return;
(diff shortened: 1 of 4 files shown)


src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -79,6 +79,7 @@ public function __construct(ManagerRegistry $managerRegistry, RequestStack $requ
     public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null, array $context = [])
     {
         if (!isset($context['filters'][$this->orderParameterName]) || !\is_array($context['filters'][$this->orderParameterName])) {
+            $context['filters'] = null;
```
soyuka: looks hackish no? Any other way we could fix that?

antograssiot: My first though @soyuka was to remove the call to parent::apply() but it was breaking phpunit test suite. The issue to solve here is to call the grand parent function without calling the parent function.
To be more explicit, in this if we would need to call  `AbstractFilter::apply()` 
without calling `AbstractContextAwareFilter::apply()`, this is the way I found.

soyuka: Mhh yeah ok I think I got it. LGTM then thanks and nice catch btw!

soyuka:
I've just rebased this (it's a master fix because context['filters'] was introduced there). Merging when status :green_apple:

soyuka:
thanks @antograssiot

Issue #1655, opened 2018-01-16
Problems with bundles that have Entity folder


giovannialbero1992:
I've noticed that api-platform scans all bundles into vendor directory, this behaviour creates problem when i use api-platform and vich-uploader without ORM installed in my project  (for example).
Is there a config to avoid this behaviour? I'd like to evaluate only folders declared into `api_platform.yml`
```
api_platform:
    mapping:
        paths: ['%kernel.project_dir%/src/Document'] 
```

jewome62:
You should override `api_platform.resource_class_directories`
I dont know if you can just add into Yaml config or if you must override by compiler

giovannialbero1992:
i must override entire `api_platform.resource_class_directories` in compiler pass . It would be nice have a config that disable this behaviour. is it welcome a PR on this?

jewome62:
Everything is welcome in PR if the request makes sense

soyuka:
So I came across the same problem. 

An alternative solution is to patch `configureContainer` of your Kernel to add the missing paths. Indeed, we currently search for `$bundle['path']/Entity`.

I'll add a patch to help configure `resource_class_directories` it'll give more power to the developer IMO. For example:

```
api_platform:
    mapping:
        paths:
          - '%kernel.project_dir%/src/Foo/api_resources.yaml'
    resource_class_directories:
        - '%kernel.project_dir%/src/Foo/Entity'
```

soyuka:
Doc PR: https://github.com/api-platform/docs/pull/380 
Closing.

PR #1665, opened 2018-01-25
feat Ability to configure resource_class_directories

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1655
| License       | MIT
| Doc PR        | need to do

It's better for flex.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -184,6 +184,12 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
         list($xmlResources, $yamlResources) = $this->getResourcesToWatch($container, $config['mapping']['paths']);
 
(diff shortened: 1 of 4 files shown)


dunglas:
What's the difference with `mappings`?

dunglas:
To be more explicit: shouldn't we set the default value of this parameter with the paths defined in `mappings`?

soyuka:
> To be more explicit: shouldn't we set the default value of this parameter with the paths defined in mappings?

Okay so at first I added `$paths['dir'] = dirname($yamlPath)` in https://github.com/api-platform/core/blob/4ea214d6aa953401552d3afecce792cf3492c13b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L256 

But in fact those are really different to me because if I add `src/App` to the directories it's not `src/App/Entity` and it'll not really parse the correct directory.

Issue #1670, opened 2018-01-31
Order by embedded field in ApiResource annotation


Awkan:
# Problem reproduction
I use an ordering on an embedded field in my entity. I want to order by embed.name field for example, then I put `embed.name` in my ApiResource->attributes->order annotation.
```php
/**
 * @ORM\Embeddable
 */
class EmbeddableDummy
{
    /*
     * @ORM\Column()
     */
    private $name;
    // ...
}

/**
 * @ApiResource(
 *     attributes={
 *         "order"={"embed.name": "DESC"}
 *     }
 * )
 */
class EmbeddedDummy
{
    /**
     * @ORM\Embedded(class="EmbeddableDummy")
     */
    public $embed;
}
```
This works well since v2.1.3 but the bug appeared with the next release. A message like this appear with a 500 : 
```
[Semantical Error] line 0, col 119 near 'embed_a1 ORDER': Error: Class AppBundle\\Entity\\EmbeddedDummy has no association named embed
```
After some research I found the "culprit": https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php#L62.

# Problem
The problem is that in case of embedded entities, a join is made.

# Resolution
Avoid join when an embedded field is used in ordering.    
Refering to the above link, we have to keep if clause (`$field = "{$rootAlias}.{$field}";`) for embedded fields and not `joinOnce` in the where clause.

If you're ok, I can make a fix

soyuka:
Seems like a valid use case :+1: on the patch!

Awkan:
Thanks. Then PR is in progress, I'll submit it soon (maybe today)

Simperfit:
The PR has been merged, thanks @Awkan!

PR #1672, opened 2018-01-31
 Fix OrderExtension to provide ordering with embedded fields

Awkan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1670
| License       | MIT
| Doc PR        | -

This PR is related to `2.1` version and would be updated too in master branch.    
To push it to `master` branch, we have to add the code `$queryBuilderProphecy->getRootAliases()->shouldBeCalled()->willReturn(['o']);` in my test.    
Do i need to open a second PR ?

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -52,7 +52,9 @@ final class OrderExtension implements QueryCollectionExtensionInterface
                         $field = $order;
                         $order = 'ASC';
                     }
-                    if (false === ($pos = strpos($field, '.'))) {
+
+                    if (false === ($pos = \strpos($field, '.'))
+                        || isset($classMetaData->embeddedClasses[\substr($field, 0, $pos)])) {
                         // Configure default filter with property
                         $field = 'o.'.$field;
                     } else {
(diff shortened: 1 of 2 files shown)


soyuka:
> Do i need to open a second PR ?

That's fine we first need to merge 2.1 to master and apply the additional patch. We will take care of this!

src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
```diff
@@ -52,7 +52,9 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
                         $field = $order;
                         $order = 'ASC';
                     }
-                    if (false === ($pos = strpos($field, '.'))) {
+
+                    if (false === ($pos = \strpos($field, '.'))
+                        || $pos && \array_key_exists(\substr($field, 0, $pos), $classMetaData->embeddedClasses)) {
```
soyuka: Could `isset($classMetaData->embeddedClasses[substr($field, 0, $pos)])` work? I think it's faster.

Also, not sure that `$pos &&` is required because the position will never be 0:

```
if (
  false === ($pos = strpos($field, '.'))
  ||
  isset($classMetaData->embeddedClasses[substr($field, 0, $pos)])
) {

}
```

Awkan: - I'll use `isset` then ;-)
- In case that `$pos = false`, `substr($field, 0, $pos)` will throw an exception since `$pos` is a boolean. Add `$pos &&` ensures us that is not boolean

soyuka: Mhh, but because the left part of the condition would be `true` with `$pos` being `false` it will never test the second part of the condition:

```
> false === ($pos = false) || false
true
```

Awkan: Yeah, my bad ! `$pos &&` is now removed.

dunglas:
Thanks @Awkan

Nek-:
@dunglas @soyuka  this should also be merge on master

PR #1675, opened 2018-02-01
Fix service declaration deprecation for api_platform.router

Zayon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -41,6 +41,8 @@
             <tag name="routing.loader" />
         </service>
 
+        <service id="ApiPlatform\Core\Api\UrlGeneratorInterface" alias="api_platform.router" />
+
         <service id="api_platform.router" class="ApiPlatform\Core\Bridge\Symfony\Routing\Router" public="false">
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -44,6 +44,8 @@
             <tag name="routing.loader" />
         </service>
 
+        <service id="ApiPlatform\Core\Api\UrlGeneratorInterface" alias="api_platform.router" public="true" />
```
dunglas: `public="true"` should be removed (there is no reason to make this alias public).

dunglas:
Can you rebase against 2.1 please? To me it's a bug fix.

dunglas:
You also to add the alias in this array to fix the tests: https://github.com/api-platform/core/blob/master/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php#L493

Zayon:
ping @dunglas. I've updated the PR.

dunglas:
Thanks @Zayon

PR #1676, opened 2018-02-04
Add a xmls:type to rdfs:range in case of an iterable type

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

This is necessary for a documentation parser like api-platform/api-doc-parser to know a field is a for example a json_array.
I am not sure this is the best way to do this, because the documentation of RDF/XML is unclear about list types, but at least it's a datatype in the XML Schema: https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#list-datatypes.

### DIFF
--- a/tests/Hydra/Serializer/DocumentationNormalizerTest.php
+++ b/tests/Hydra/Serializer/DocumentationNormalizerTest.php
@@ -44,7 +44,7 @@ class DocumentationNormalizerTest extends TestCase
         $documentation = new Documentation(new ResourceNameCollection(['dummy' => 'dummy']), $title, $desc, $version, []);
 
         $propertyNameCollectionFactoryProphecy = $this->prophesize(PropertyNameCollectionFactoryInterface::class);
-        $propertyNameCollectionFactoryProphecy->create('dummy', [])->shouldBeCalled()->willReturn(new PropertyNameCollection(['name', 'description', 'relatedDummy']));
+        $propertyNameCollectionFactoryProphecy->create('dummy', [])->shouldBeCalled()->willReturn(new PropertyNameCollection(['name', 'description', 'jsonData', 'relatedDummy']));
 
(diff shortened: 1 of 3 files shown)


tests/Hydra/Serializer/DocumentationNormalizerTest.php
```diff
@@ -168,6 +169,21 @@ public function testNormalize()
...
                             'hydra:description' => 'description',
                         ],
+                        [
+                            '@type' => 'hydra:SupportedProperty',
+                            'hydra:property' => [
+                                '@id' => '#dummy/jsonData',
+                                '@type' => 'rdf:Property',
+                                'rdfs:label' => 'jsonData',
+                                'domain' => '#dummy',
+                                'range' => 'xmls:list',
```
dunglas: According to the spec, it doesn't look ok: `·List· datatypes are always ·constructed· from some other type; they are never ·primitive·.`.

If we need to go this way, we'll need to provide our XSD with custom types, and use those custom types here.

alanpoulain: Providing our XSD with custom types doesn't seem a good idea to me, for interoperability for example.
WDYT? Do you have another solution in mind?

dunglas: XML and open data in general doesn't fit well with random, dynamic data structures... The best thing to do is that the user document correctly the expected structure of the returned data (for example using https://github.com/dunglas/doctrine-json-odm.

alanpoulain: Sure but it can also be a simple array of strings or integers. Doctrine JSON ODM does not manage this kind of JSON, doesn't it?

Issue #1677, opened 2018-02-05
Access control denied message


norkunas:
Symfony allows to configure custom message for security like `@Security("is_granted('ROLE_ADMIN')", message="Access denied because..")`, but in api-platform there is no way to configure this unless overriding/decorating the listener `ApiPlatform\Core\Security\EventListener\DenyAccessListener`.

So I would like to propose additional attribute `access_control_message` for customizing this.

soyuka:
:+1: Why aren't we propagating every attributes?

PR #1678, opened 2018-02-05
Implement access_control_message attribute

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1677
| License       | MIT
| Doc PR        | none

### DIFF
--- a/src/Security/EventListener/DenyAccessListener.php
+++ b/src/Security/EventListener/DenyAccessListener.php
@@ -76,7 +76,7 @@ final class DenyAccessListener
         $extraVariables['request'] = $request;
 
         if (!$this->resourceAccessChecker->isGranted($attributes['resource_class'], $isGranted, $extraVariables)) {
-            throw new AccessDeniedException();
+            throw new AccessDeniedException($resourceMetadata->getOperationAttribute($attributes, 'access_control_message', 'Access Denied.', true));
         }
     }
 }
--- a/tests/Security/EventListener/DenyAccessListenerTest.php
+++ b/tests/Security/EventListener/DenyAccessListenerTest.php
@@ -111,6 +111,30 @@ class DenyAccessListenerTest extends TestCase
     }
 
     /**
+     * @expectedException \Symfony\Component\Security\Core\Exception\AccessDeniedException
+     * @expectedExceptionMessage You are not admin.
+     */
+    public function testAccessControlMessage()
+    {
+        $request = new Request([], [], ['_api_resource_class' => 'Foo', '_api_collection_operation_name' => 'get']);
+
+        $eventProphecy = $this->prophesize(GetResponseEvent::class);
+        $eventProphecy->getRequest()->willReturn($request)->shouldBeCalled();
+        $event = $eventProphecy->reveal();
+
+        $resourceMetadata = new ResourceMetadata(null, null, null, null, null, ['access_control' => 'has_role("ROLE_ADMIN")', 'access_control_message' => 'You are not admin.']);
+
+        $resourceMetadataFactoryProphecy = $this->prophesize(ResourceMetadataFactoryInterface::class);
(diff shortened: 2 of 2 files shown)


src/Security/EventListener/DenyAccessListener.php
```diff
@@ -81,7 +81,17 @@ public function onKernelRequest(GetResponseEvent $event)
...
-            throw new AccessDeniedException();
+            $defaultMessage = 'Access Denied.';
+
+            if (isset($attributes['collection_operation_name'])) {
+                $message = $resourceMetadata->getCollectionOperationAttribute($attributes['collection_operation_name'], 'access_control_message', $defaultMessage, true);
+            } elseif (isset($attributes['item_operation_name'])) {
+                $message = $resourceMetadata->getItemOperationAttribute($attributes['item_operation_name'], 'access_control_message', $defaultMessage, true);
+            } else {
+                $message = $resourceMetadata->getCollectionOperationAttribute($attributes['subresource_operation_name'], 'access_control_message', $defaultMessage, true);
+            }
```
soyuka: Couldn't we add a `getOperationAttribute` method that does the if/elseif/else thing in `resourceMetadata` to avoid all this boilerplate?

norkunas: Good idea. But `getOperationAttribute` already exists as a private method :) but this would be out of PR scope, isn't it?

norkunas:
One thing looks suspicious to me.

```
} else {
    $isGranted = $resourceMetadata->getCollectionOperationAttribute($attributes['subresource_operation_name'], 'access_control', null, true);
}
```

Resource metadata has a `getSubresourceOperationAttribute`, so it should use this or I am completely don't understand this part? Because I've just copied attribute getting part but coveralls shows that i'm not covering that part with my test :/

src/Security/EventListener/DenyAccessListener.php
soyuka: yeah it would and anyway we have a lot of these in the current code I'm going to open an issue on that matter.

soyuka:
I think it should use getSubresourceOperationAttribute. Wouldn't you need to test every branch of the condition for 100% coverage? IMO it's not necessary especially if we refactor this.

norkunas:
Okay, so I will wait for further decision here.

src/Security/EventListener/DenyAccessListener.php
```diff
@@ -81,7 +81,17 @@ public function onKernelRequest(GetResponseEvent $event)
...
         if (!$this->resourceAccessChecker->isGranted($attributes['resource_class'], $isGranted, $extraVariables)) {
-            throw new AccessDeniedException();
+            $defaultMessage = 'Access Denied.';
+
+            if (isset($attributes['collection_operation_name'])) {
+                $message = $resourceMetadata->getCollectionOperationAttribute($attributes['collection_operation_name'], 'access_control_message', $defaultMessage, true);
+            } elseif (isset($attributes['item_operation_name'])) {
+                $message = $resourceMetadata->getItemOperationAttribute($attributes['item_operation_name'], 'access_control_message', $defaultMessage, true);
+            } else {
+                $message = $resourceMetadata->getCollectionOperationAttribute($attributes['subresource_operation_name'], 'access_control_message', $defaultMessage, true);
```
soyuka: In fact this should be `getSubresourceOperationAttribute`

norkunas: Updated.

dunglas:
Can now be upgraded to use #1680 :)

norkunas:
@dunglas updated.

dunglas:
Thanks! Don’t forget to open a doc PR!

soyuka:
this was fast thanks @norkunas for the update and @dunglas for the merges \o/

PR #1685, opened 2018-02-07
Allow to specify a route prefix per resource

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/391

Sometimes it's useful to be able to specify a prefix for all operations of a given resource. In that case right now you have to override the paths of all operations individually.
The goal of this PR is to allow this:

```php
/**
* @ApiResource("attributes"={"routePrefix"="/foobar-prefix"})
 */
class Entity {}
```

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -39,6 +39,11 @@ final class ApiResource
     public $iri;
 
     /**
+     * @var string
+     */
+    public $routePrefix;
+
+    /**
      * @var array
      */
     public $itemOperations;
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -39,11 +39,6 @@ final class ApiResource
     public $iri;
 
     /**
-     * @var string
-     */
-    public $routePrefix;
-
-    /**
      * @var array
      */
     public $itemOperations;
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -193,7 +193,7 @@ final class ApiLoader extends Loader
             }
         }
 
-        $path = $resourceMetadata->getAttribute('routePrefix', '');
+        $path = trim(trim($resourceMetadata->getAttribute('routePrefix', '')), '/');
         $path .= $this->operationPathResolver->resolveOperationPath($resourceShortName, $operation, $operationType, $operationName);
 
         $route = new Route(
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\Exception\InvalidResourceException;
 use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use ApiPlatform\Core\Operation\Factory\SubresourceOperationFactoryInterface;
 use ApiPlatform\Core\PathResolver\OperationPathResolverInterface;
 use Symfony\Component\Config\FileLocator;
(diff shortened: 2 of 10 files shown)


src/Annotation/ApiResource.php
```diff
@@ -38,6 +38,11 @@
      */
     public $iri;
 
+    /**
+     * @var string
+     */
+    public $routePrefix;
```
dunglas: Do we really need a new attribute for that? Can't we use an attribute?

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -197,8 +198,11 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
             }
         }
 
+        $path = $resourceMetadata->getRoutePrefix() ?: '';
```
dunglas: `??`?

src/Annotation/ApiResource.php
dunglas: Speaking about that, maybe can we just forward any top-level attribute passed to the annotation to attributes? This way it will be easier for the user. Anyway if we create a new attribute, the XML and the YAML loader must be extracted too (it's why I'm 👎).

Toflar: Well to be honest, I have no idea what qualifies for a top-level attribute and what can be part of `attributes` :-D

src/Bridge/Symfony/Routing/ApiLoader.php
Toflar: True

src/Annotation/ApiResource.php
soyuka: @dunglas we should totally freeze current top-level stuff and only allow updates with optional attributes IMO

Toflar: Yeah I found it strange that I need to update all the factories. So I'm moving this to attributes?

dunglas: Yes please, I also think it's better to use attributes.

@soyuka what do you think about the fallback mechanism I was mentioning earlier?

Toflar:
Okay, comments addressed. Makes the PR a lot smaller 😄 Will do a docs PR once this is approved.

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -170,17 +171,17 @@ private function loadExternalFiles(RouteCollection $routeCollection)
...
      *
-     * @param RouteCollection $routeCollection
-     * @param string          $resourceClass
-     * @param string          $operationName
-     * @param array           $operation
-     * @param string          $resourceShortName
-     * @param string          $operationType
-     *
-     * @throws RuntimeException
+     * @param RouteCollection  $routeCollection
```
dunglas: You can drop this block, we now don't add PHPDoc when not adding information compared to the typehints. Just keep the `@throw` if appropriate.

Toflar:
Docs PR linked. PR ready.

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -197,8 +193,11 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
             }
         }
 
+        $path = $resourceMetadata->getAttribute('routePrefix', '');
```
meyerbaptiste: What do you think about sanitizing the prefix with `'/'.trim(trim($prefix, '/'))` if it's defined (`!== ''`)?

Toflar: I thought about that too but then my idea was that in Symfony, you always have to start with `/` both, for prefixes as well as routes. And if you define them incorrectly, you see an invalid path so that's a quickfix. Not sure we should do this :)

meyerbaptiste: My idea was just to follow the same logic as [`RouteCollection::addPrefix()`](https://github.com/symfony/symfony/blob/880d7e74360f42c8eaf7484e217093e3424c7b2c/src/Symfony/Component/Routing/RouteCollection.php#L143).
IMO it can improve the UX.

antograssiot:
Just asking but should this option be applied even  if `path` is defined at operation level ? That would avoid using this option if a single operation doesn't need it, and therefore force the developer to use `path` again on every operation...

Toflar:
To me, I think chances are higher that you would want to "namespace" the whole entity. So in fact, also when you use `path` on an operation, you want to have the prefix applied as well. Your example would mean you want the prefix for 3 operations except for 2 and in cases like these I think it's okay to specify `path` on every operation.

antograssiot:
@toflar I have seen several cases of entites with 10+ operations and only 2 routes with different namespace. If you for instance need several permissions + extensions to be applied for automatic filtering of the same collection based on user "type"
If the attribute is overridable at operation level, that's enough anyway I think

Toflar:
The problem is, it's separated logic. The `OperationPathResolver` can return whatever path it wants to. I cannot find out whether the prefix should be applied or not.

src/Bridge/Symfony/Routing/ApiLoader.php
Toflar: Included 👍

dunglas:
Thanks @Toflar

PR #1689, opened 2018-02-08
Support for easier registration of own cache purgers

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | TODO
| Fixed tickets | 
| License       | MIT
| Doc PR        | TODO

This will make it pretty easy to implement and register custom cache purgers as follows:

```yaml
http_cache:
    invalidation:
        enabled: true
        purger: my_superb_cache_purger
        tags_formatter: my_superb_cache_tags_formatter
```

This will then alias the `api_platform.http_cache.purger` to `my_superb_cache_purger`.

In addition to this, I've introduced a `CacheTagsFormatterInterface` that allows to delegate the formatting of the `$iris` to the purger and a `Cache-Tags-Debug` header that contains them unformatted in case `kernel.debug` is set to `true`.


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -498,6 +498,12 @@ private function registerHttpCache(ContainerBuilder $container, array $config, X
 
         $loader->load('http_cache_tags.xml');
 
+        // TODO: Decide on how you would like to configure whether to use Varnish or Symfony (or any other?)
+        if (0 == count($config['http_cache']['invalidation']['varnish_urls'])) {
```
dunglas: We may allow to use both (to avoid such ifs). If a `varnish_urls` is defined, it's definitely to use Varnish.

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -66,6 +66,13 @@ public function onKernelResponse(FilterResponseEvent $event)
             return;
         }
 
+        // Encode tags for greater compatiblity with different proxies
+        // Some do not allow special characters like / or @ in cache tags and
+        // also it allows to use a , in a tag, if you wish to do so.
+        $resources = array_map(function($resource) {
```
dunglas: We were using md5 initially, we switched back to plain text because it is very hard to debug when reading/grepping logs. I would prefer to encode special char at the cache adapter layer.

src/HttpCache/StoreAwareKernelInterface.php
```diff
@@ -0,0 +1,36 @@
+<?php
```
dunglas: Why do we need this interface and the related trait? Can't we just use inject the PSR-6 adapter in the purger?

src/HttpCache/SymfonyPurger.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\HttpCache;
+
+use Toflar\Psr6HttpCacheStore\Psr6StoreInterface;
```
dunglas: Wouldn't it be possible to move this new implementation directly in Symfony?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
Toflar: So how would your preferred config look like? If `varnish_urls` then Varnish, else Symfony? What if we want to integrate another one?

src/HttpCache/StoreAwareKernelInterface.php
Toflar: I can do that but to remain BC I need to introduce another interface for the Purgers that implement both `PurgerInterface` and...hmmm... `CacheTagsFormatterInterface`?

src/HttpCache/SymfonyPurger.php
Toflar: No, I tried and it was denied. See https://github.com/symfony/symfony/pull/20061 for the whole discussion.

src/HttpCache/EventListener/AddTagsListener.php
Toflar: I can do that but to remain BC I need to introduce another interface for the Purgers that implement both PurgerInterface and...hmmm... CacheTagsFormatterInterface?

src/HttpCache/StoreAwareKernelInterface.php
Toflar: No that's not possible. `HttpCache` works outside the application so you cannot access it from within the container.

Toflar: We could implement the same as Varnish does. Have the purger execute a `PURGE` request and have a `kernel.request` listener that purges the data. But that's way too complicated I find.

src/HttpCache/SymfonyPurger.php
dunglas: Ok... I'm not confident with adding (soft) dependencies to non-popular yet libraries (like `Psr6HttpCacheStore`). We already had problems with bridges to 3rd party libraries not very well maintained (FOS things...).

My proposal is:
* Tweak the config in this PR to allow to easily register external Purgers from the config (using a service name for instance).
* Move the `SymfonyPurger` class and the trait in an external bundle (or directly in `Psr6HttpCacheStore`) for now

WDYT?

Toflar: Yeah I understand. In that case, it's my library and it consists of one class so I can basically guarantee it's maintained because I can't think of a lot of stuff that would need to be maintained there 😄 But anyway, I just feel it can never get popular if we don't give it the chance. I mean, I'm totally okay with putting the APIP integration into it's own bundle if you don't want it in core but then it will be just another one floating around in my personal namespace instead of APIP so chances are, people won't even notice there's such a thing.

dunglas: It's mostly a docs issue to me. I'm totally ok to add a new doc entry explaining how to install your lib (that will be required to install manually anyway) to get cache invalidation in pure PHP.

soyuka: :+1: 

@Toflar just add the SymfonyPurger and the StoreAwareKernelInterface to your repository! 
Patch api platform to allow the use of more then 1 purger in this pull request. 

Then we just add a doc entry! I'd be the first to enjoy the pure PHP cache I think it's a great idea!

soyuka: The Kernel should be in whatever namespace.

If the purger is implementing our PurgerInterface yeah it can be in a Bridge/ApiPlatform thingy

Toflar: Okay, no problem. There's one thing that needs to be clarified first though: the encoding of the tags. I see two options here:

1) I add a new interface every purger can implement that allows to ask the purger to encode the tags whereas fallback would be the current, comma-separated list. Pros: Very flexible, Cons: more complicated.

2) We use `md5` or `sha1` by default and add a `Cache-Tags-Debug` header which shows  the raw tags if `kernel.debug` is enabled. I cannot think of any cache implementation that cannot handle hashes as keys. Pros: Simple, Cons: Less flexible.

WDYT?

soyuka: I'd go for option 1

Toflar: Okay, would you still add the additional debug header? I think it might be very useful and it doesn't harm because it's hidden in production anyway.

soyuka: Sure I think it's a good idea as well!

Toflar:
Changed title, description, implementation, rebased on master branch. Ready for review round 2 😄

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -498,6 +498,14 @@ private function registerHttpCache(ContainerBuilder $container, array $config, X
 
         $loader->load('http_cache_tags.xml');
 
+        // Load purger
+        $purger = $config['http_cache']['invalidation']['purger'];
+        $container->setAlias('api_platform.http_cache.purger', 'api_platform.http_cache.purger.'.$purger);
```
dunglas: Can you use `"api_platform.http_cache.purger.$purger"` to trigger https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -25,16 +27,21 @@
...
  * @experimental
  */
 final class AddTagsListener
 {
     private $iriConverter;
+    private $purger;
+    private $debug;
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, PurgerInterface $purger, bool $debug = false)
```
dunglas: Can't we just typehint `CacheTagsFormattingPurgerInterface` directly?

src/HttpCache/CacheTagsFormattingPurgerInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace ApiPlatform\Core\HttpCache;
+
+/**
+ * Allows cache purgers to format the cache tags header according
+ * to their own needs.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+interface CacheTagsFormattingPurgerInterface extends PurgerInterface
```
dunglas: WDYT about not extending `PurgerInterface` and renaming it `CacheTagsFormatterInterface`? It will allow to extract this logic in a separated class.

src/HttpCache/EventListener/AddTagsListener.php
dunglas: Can you make the purger optional? It's useless when using the builtin one (it eases to use classes outside of Symfony / API Platform full stack).

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -66,6 +73,16 @@ public function onKernelResponse(FilterResponseEvent $event)
             return;
         }
 
-        $event->getResponse()->headers->set('Cache-Tags', implode(',', $resources));
+        if ($this->debug) {
```
dunglas: I'm not sure it is worth it. The Formatter can do it by itself by using `RequestStack` (and it's useless for the default purger).

dunglas: By the way, even in your purger, I suggest to emit non-encoded headers, and to do the encoding/decoding at the cache store level.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
Toflar: Oh nice one! I've read a lot about opcode optim but must have missed that one! Done in f4aec2ec81b4e56111465cbd2d1a7196249dfd4a.

src/HttpCache/EventListener/AddTagsListener.php
Toflar: No because the current `VarnishPurger` does not implement that and I don't know what other implementations have been made out there. Also it's completely optional to adjust the formatting of the tags :)

src/HttpCache/CacheTagsFormattingPurgerInterface.php
Toflar: Yeah, thought about that too but then again, where would you put that? It should not reside in the `HttpCache` namespace etc. so I thought I'm going to create a very specific interface for that purpose only.

src/HttpCache/EventListener/AddTagsListener.php
Toflar: It's just an easy debugging help without any negative side effects. 
It's the Symfony Cache that does not allow certain characters, that comment should probably go there :-) I still think it's fine here. Sometimes you don't have control over the cache internals and then you're lost without that. I will make both optional using setter injection though.

src/HttpCache/CacheTagsFormattingPurgerInterface.php
dunglas: Why not in `HttpCache`? This namespace looks good to me.

src/HttpCache/EventListener/AddTagsListener.php
dunglas: I don't get it. The dependency wasn't existent before. `CacheTagsFormatterInterface $formatter = null` will do the trick. Btw we try to avoid setter injection as much as possible.

Toflar:
PR is ready for another review then :)

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -207,6 +207,7 @@ public function getConfigTreeBuilder()
                             ->info('Enable the tags-based cache invalidation system.')
                             ->canBeEnabled()
                             ->children()
+                                ->scalarNode('purger')->defaultValue('varnish')->cannotBeEmpty()->info('The name of the purger to use.')->end()
```
meyerbaptiste: IMO, it should be directly the name of the service instead of a tricky concatenation (i.e `api_platform.http_cache.purger.$purger`). Or did I miss something specific?
So, create an implementation of your new interface for Varnish whose the service name will be the default value here. 

WDYT?

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -25,16 +26,21 @@
...
  * @experimental
  */
 final class AddTagsListener
 {
     private $iriConverter;
+    private $purger;
+    private $debug = false;
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, CacheTagsFormattingInterface $purger = null, bool $debug = false)
```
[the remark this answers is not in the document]
Toflar: True, fixed in a1b08deafc5054261383d87ba00b7bc735895fd7

src/HttpCache/CacheTagsFormattingInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace ApiPlatform\Core\HttpCache;
+
+/**
+ * Allows cache purgers to format the cache tags header according
+ * to their own needs.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+interface CacheTagsFormattingInterface
```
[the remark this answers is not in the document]
Toflar: I don't think that's correct. The implementation of that interface "formats the cache tags". So "CacheTagsFormatting" is correct. "CacheTaggingFormatter" sounds really weird to me.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
Toflar: Addressed in 2e5bbbf4c16a6a42caa25f53fb3abf96a1d62573.

Toflar:
Rebased onto master.

src/HttpCache/CacheTagsFormattingInterface.php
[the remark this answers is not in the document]
meyerbaptiste: > "formats the cache tags"

So... What do you think of `CacheTagsFormatterInterface`? 😄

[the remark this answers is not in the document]
Toflar: Changed 😄

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -25,16 +27,27 @@
...
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, CacheTagsFormatterInterface $tagsFormatter = null, bool $debug = false)
     {
         $this->iriConverter = $iriConverter;
+        $this->tagsFormatter = $tagsFormatter;
+        $this->debug = $debug;
+
+        // BC
+        if (null === $this->tagsFormatter) {
+            @trigger_error('Passing no implementation of the CacheTagsFormatterInterface is deprecated since version 2.3 and will be removed in 3.0.', E_USER_DEPRECATED);
```
dunglas: Please don't trigger a deprecation notice here. It makes it header to use for no benefit.

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -25,16 +27,27 @@
...
 {
     private $iriConverter;
+    private $tagsFormatter;
+    private $debug = false;
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, CacheTagsFormatterInterface $tagsFormatter = null, bool $debug = false)
     {
         $this->iriConverter = $iriConverter;
+        $this->tagsFormatter = $tagsFormatter;
```
dunglas: `$this->tagsFormatter = $tagsFormatter ?? new CsvFormatter();`

src/HttpCache/CsvFormatter.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\HttpCache;
+
+/**
+ * Formats the cache header as a comma-separated list.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class CsvFormatter implements CacheTagsFormatterInterface
```
dunglas: I'm think it is not worth it to add a new class for an implode (see my other suggestion).

src/HttpCache/CacheTagsFormatterInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace ApiPlatform\Core\HttpCache;
+
+/**
+ * Allows cache purgers to format the cache tags header according
+ * to their own needs.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+interface CacheTagsFormatterInterface
```
dunglas: What about just using `callable`. For this use case it's enough IMO, no need to introduce an interface.

src/HttpCache/EventListener/AddTagsListener.php
dunglas: Or even better:

```php
$this->tagsFormatter = $tagsFormatter ?? function (array $iris): string
{
    return implode(',', $iris);
};
```

src/HttpCache/EventListener/AddTagsListener.php
dunglas: Sorry to insist but please drop it, it brings nothing for the default setup (cache tags are already in clear text) and may cause maintenance issues (I don't like injecting the `debug` parameter in this listener). It can easily be added in an external class/bundle.

Toflar:
I've put the formatting into it's own class because of the discussion in Slack. I don't really feel like changing it yet again.

Toflar:
Debugging is gone.

src/HttpCache/CacheTagsFormatterInterface.php
Toflar: Callables are not a good choice at all here. They are very hard to add. Right now you can simply configure a different service by using the `tags_formatter` config setting and you're done.

dunglas:
I can finish this if you want @Toflar

Toflar:
No, it will just have to wait for after my holidays. And callables are the wrong choice to me 😄

src/HttpCache/CacheTagsFormatterInterface.php
dunglas: Formatting tags is very specific edge case to me, ok to allow to do it, but I'm not confortable to add so many things just for this. A `callable` looks perfect for this use case.

Toflar: I've always said it's something that's specific to the purger because the purger knows how to format the tags so it can actually purge them. It never makes sense to arbitrarily combine purgers and tag formatters. That's why I had `interface CacheTagsFormattingPurgerInterface extends PurgerInterface` and passed the purger to the `AddTagsListener`. That way, any purger can format if it needs but it can also just leave this interface alone and go with the default implementation which is just a CSV list. This is still the version that makes the most sense from a software architectural POV to me because responsibilities are put where they belong.

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -25,16 +27,25 @@
...
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, CacheTagsFormatterInterface $tagsFormatter = null)
     {
         $this->iriConverter = $iriConverter;
+        $this->tagsFormatter = $tagsFormatter;
+
+        // BC
+        if (null === $this->tagsFormatter) {
+            @trigger_error('Passing no implementation of the CacheTagsFormatterInterface is deprecated since version 2.3 and will be removed in 3.0.', E_USER_DEPRECATED);
```
Simperfit: `Passing an object that does not implement CacheTagsFormatterInterface is deprecated since version 2.3 and will be removed in 3.0.` WDYT ?

Toflar:
I would like to rework it completely as outlined in https://github.com/api-platform/core/issues/1776.

Toflar:
But haven't gotten any feedback there yet.

dunglas:
@Toflar ok for me!

soyuka:
Should I do something with this @Toflar ?

Toflar:
I think that can be closed. I've worked a lot on the FOSCache lib and bundle lately and I found that it would be better to tie tagging and purging to the same service so it would require a complete overhaul anyway.

soyuka:
Thanks for the quick response @Toflar !

Issue #1627, opened 2018-01-05
[Filters] Validation based upon filter description


jdeniau:
Hi,

If I am correct the filters are based upon [OpenApi Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject)

They accept a `required` key, and nice things like `enum`, `pattern`, etc. but I don't think it's used by api-platform for now except for documentation and `hydra:mapping` generation.

Am I right or did I miss someting ? 

If so, I will be pleased to open a PR with the start of query parameter validation.

Thank you

PR #1692, opened 2018-02-08
throw an exception if required filter is not set

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | yes
| BC breaks?    | if filter config was wrongly set to `required=true`
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1627
| License       | MIT
| Doc PR        | Doc PR will follow if PR is approved

This PR is a start of query parameter validation.

It does only check `required=true` filter for now but I will open other PR to include [the rest of OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object) on field validation.

There are some points that I am not really sure of:

  * namespace of the new `QueryParameterValidateLister`: I put it in `Core\Validator\EventListener` but maybe in `Symfony\Bridge` ?
  * I set the priority of the event listener to 16, this way, it is the first event in the queue (according to [the documentation](https://api-platform.com/docs/core/events))

Otherwise, the code is quite simple and leverage Symfony validation system.

Thanks !


dunglas:
I'm glad that you work on this! Thanks.

Do we really need a dependency to the Symfoyn validator for that? Can't we just use the filter extension? It will reusing API Platform in a non-Symfony context (PSR-7 and Laravel support is on our roadmap).

About the namespace, I would suggest to put it in the `Filter` namespace.

jdeniau:
@dunglas I used Symfony validation for several reason: 

  * api-platform automatically throw a 400 exception when a `ConstraintViolationList` is throwned
  * api-platform do already depends on a `validator` service (is it deprecated ?)
  * OpenApi declare several possible requirements that seems to work really great with Symfony constraints system

Or maybe I did not really understand your comment ?

dunglas:
The integration with the validator is 100% optional (as with Doctrine etc), in fact it's only used in the bridge for convenience for the end user, but IIRC, not for any "core" feature like this one.

Regarding the 400, it's just one line to add to support a new exception: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L251

jdeniau:
@dunglas OK for the `ContraintViolationException`, I can throw an api-platform specific exception, but don't you think it will be reinventing the wheel to re-implement every OpenApi validation ? (enum, pattern, min / max, empty value, type, etc.) ?

I think it will be a big copy/paste from symfony's constraints, don't you ?

jdeniau:
@dunglas I remove the dependency on Symfony validator service.

About using the FilterExtension, I did not, because FilterExtension is Doctrine-specific and will not work when using resources that are not Doctrine's.

I think it's good for merge. I will make other PR's implementing the rest of OpenApi parameters specification after that.

src/Exception/FilterValidationException.php
```diff
@@ -0,0 +1,46 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * Deserialization exception.
```
dunglas: Should be updated

src/Exception/FilterValidationException.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->constraintViolationList = $constraintViolationList;
+
+        parent::__construct($message ?: $this->__toString(), $code, $previous);
+    }
+
+    /**
+     * Gets constraint violations related to this exception.
+     *
+     * @return ConstraintViolationListInterface
```
dunglas: You can use a return type instead.

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,75 @@
...
+
+use ApiPlatform\Core\Api\FilterLocatorTrait;
+use ApiPlatform\Core\Exception\FilterValidationException;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+
+/**
+ * Validate query parameters depending on filter description.
```
dunglas: Validates

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,75 @@
...
+        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
+        $resourceFilters = $resourceMetadata->getCollectionOperationAttribute($operationName, 'filters', [], true);
+
+        foreach ($resourceFilters as $filterId) {
+            if (!$filter = $this->getFilter($filterId)) {
+                continue;
+            }
+
+            foreach ($filter->getDescription($attributes['resource_class']) as $name => $data) {
+                if ($data['required'] ?? false) {
```
dunglas: Can you invert this condition to reduce the complexity?

```php
if (!($data['required'] ?? false)) {
    continue;
}
```

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,75 @@
...
+                    $errorList = [];
+
+                    if (null === $request->query->get($name)) {
+                        $errorList[] = sprintf('Query parameter `%s` is required', $name);
```
dunglas: Can you change the backticks with double quotes for consistency with other exceptions?

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,75 @@
...
+                    }
+
+                    if (count($errorList) > 0) {
```
dunglas: Can be simplified in `if ($errorList) {`

src/Filter/QueryParameterValidateListener.php
jdeniau: I changed a litte this part to prepare the next step that will validate several parameter rules, but this should be more understandable though

jdeniau:
@dunglas done

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,73 @@
...
+
+        foreach ($resourceFilters as $filterId) {
+            if (!$filter = $this->getFilter($filterId)) {
+                continue;
+            }
+
+            foreach ($filter->getDescription($attributes['resource_class']) as $name => $data) {
+                $errorList = [];
+
+                if (($data['required'] ?? false) && null === $request->query->get($name)) {
```
meyerbaptiste: This condition will not work with serializer's filters (assuming they are required) for example. Indeed, `property[]` is not equal to `property`.

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,73 @@
...
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+
+/**
+ * Validates query parameters depending on filter description.
+ *
+ * @author Julien Deniau <julien.deniau@gmail.com>
+ */
+class QueryParameterValidateListener
```
meyerbaptiste: Can you add a unit test please?

jdeniau: Is it usefull ? It's fully tested with behat tests.

src/Filter/QueryParameterValidateListener.php
jdeniau: Sorry but I did not really understand what you mean here ?

Can you provide an example when this does not work ?

meyerbaptiste: Let's take the description of the property filter: https://github.com/api-platform/core/blob/63d555ae0020dc6c66feef86ff56ac5131b1f7b3/src/Serializer/Filter/PropertyFilter.php#L68-L77
And let's say `required` is equal to `true` for our example.

So, the `$name` var will be equal to `property[]` here. But with `?property['relatedDummy'][]=name` the name will be `property` in the parameter bag and `$request->query->get('property[]')` will return `null`...

src/Filter/QueryParameterValidateListener.php
meyerbaptiste: Yes, Behat is only for integration and acceptance tests. You also have to unit test your class.

jdeniau: Done

jdeniau:
@meyerbaptiste Done for the unit tests and the "array notation" (but I'm not really happy with the implementation though :| )

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            foreach ($filter->getDescription($attributes['resource_class']) as $name => $data) {
+                $errorList = [];
+
+                if (!($data['required'] ?? false)) { // property is not required
+                    continue;
+                }
+
+                if (false !== strpos($name, '[')) { // array notation of filter
```
meyerbaptiste: I have another dirty solution:
```php
parse_str($name, $query);
$name = array_keys($query)[0] ?? '';
```
It works with:
- `'foo'`
- `'foo[]'`
- `'foo[bar]'`
- `'foo[bar][]'`
- `'foo[bar][baz]'`
- etc.

😄

jdeniau: @meyerbaptiste done

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    /**
+     * Test if required filter is valid. It validates array notation too like "required[bar]"
+     */
+    private function isRequiredFilterValid($name, $request): bool
+    {
+        $matches = [];
+        if (false === mb_parse_str($name, $matches)) {
```
meyerbaptiste: Huum, this require the `mbstring` extension...

jdeniau: Oh yes, I though it was included since PHP7, my bad, I remove this

jdeniau: done

jdeniau:
@meyerbaptiste changed regex → parse_str

jdeniau:
@Simperfit done

tests/Filter/QueryParameterValidateListenerTest.php
```diff
@@ -0,0 +1,174 @@
...
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use PHPUnit\Framework\TestCase;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+
+class QueryParameterValidateListenerTest extends TestCase
+{
+    private $testedInstance;
+
```
soyuka: useless newline

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Test if required filter is valid. It validates array notation too like "required[bar]".
+     */
+    private function isRequiredFilterValid($name, $request): bool
+    {
+        $matches = [];
+        parse_str($name, $matches);
+        if (empty($matches)) {
```
soyuka: `!$matches`

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,104 @@
...
+            return false;
+        }
+
+        $rootName = array_keys($matches)[0] ?? '';
+        if (!$rootName) {
+            return false;
+        }
+
+        if (is_array($matches[$rootName])) {
```
soyuka: `\is_array`

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,104 @@
...
+            $keyName = array_keys($matches[$rootName])[0];
+
+            $queryParameter = $request->query->get($rootName);
+
+            return is_array($queryParameter) && isset($queryParameter[$keyName]);
```
soyuka: `\is_array`

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,104 @@
...
+                    continue;
+                }
+
+                if (!$this->isRequiredFilterValid($name, $request)) {
+                    $errorList[] = sprintf('Query parameter "%s" is required', $name);
+                }
+
+                if ($errorList) {
+                    throw new FilterValidationException($errorList);
+                }
```
soyuka: shouldn't we throw outside of the foreach? if not why is `$errorList` an array?

jdeniau:
@soyuka done.

About the `$errorList`, I changed it, but it was changed in #1723 anyway

jdeniau:
@soyuka @meyerbaptiste @dunglas Can you please review this PR ? Every comments you made have been treated.

After that, I would really appreciate your comment about the PR following this one https://github.com/api-platform/core/pull/1723#issuecomment-375695568

Thanks

src/Exception/FilterValidationException.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * Filter validation exception.
+ *
+ * @author Julien DENIAU <julien.deniau@gmail.com>
+ */
+class FilterValidationException extends \Exception implements ExceptionInterface
```
dunglas: `final`

src/Filter/QueryParameterValidateListener.php
dunglas: `final`

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Filter;
```
dunglas: Please move it in the `ApiPlatform\EventListener` namespace.
What about `ValidateQueryParametersListener` for the name?

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,103 @@
...
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+
+/**
+ * Validates query parameters depending on filter description.
+ *
+ * @author Julien Deniau <julien.deniau@gmail.com>
+ */
+class QueryParameterValidateListener
```
soyuka: `final`?

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -516,6 +521,7 @@ private function getPartialContainerBuilderProphecy($test = false)
             'api_platform.listener.view.respond',
             'api_platform.listener.view.serialize',
             'api_platform.listener.view.validate',
+            'ApiPlatform\Core\Filter\QueryParameterValidateListener',
```
dunglas: Please change to `api_platform.listener.view.validate_query_parameters` for consistency (we don't use id as services name yet)

tests/Fixtures/TestBundle/Entity/ArrayFilterValidator.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function setId($id)
+    {
+        $this->id = $id;
+    }
```
soyuka: For simplicity's sake you could have put public properties here, that's find as-is anyway! 
Although, are you sure you want `id` to be writable?

tests/Fixtures/TestBundle/Entity/FilterValidator.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function setId($id)
+    {
+        $this->id = $id;
+    }
```
soyuka: same comment id is auto generated

dunglas:
About the general design, wouldn't it be better to validate the "query parameters" directly in the filter instead of in a listener? It would allow to validate required filters in GraphQL too. It would also make it easier to support and maintain this feature when we'll add an integration with Laravel, or PSR-7 support.
Now that the infrastructure has been created to access to query parameters in filter through the context, it will not be very difficult. You may move the validation logic in a trait, and use this trait in every filter to validate it.

tests/Fixtures/TestBundle/Filter/ArrayRequiredFilter.php
```diff
@@ -0,0 +1,42 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Filter;
+
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\AbstractFilter;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Doctrine\ORM\QueryBuilder;
+
+class ArrayRequiredFilter extends AbstractFilter
```
soyuka: could be final

tests/Fixtures/TestBundle/Filter/RequiredFilter.php
```diff
@@ -0,0 +1,42 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Filter;
+
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\AbstractFilter;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Doctrine\ORM\QueryBuilder;
+
+class RequiredFilter extends AbstractFilter
```
soyuka: could be final

tests/Fixtures/app/config/config_test.yml
```diff
@@ -161,6 +162,14 @@ services:
         parent:    'api_platform.serializer.property_filter'
         tags:      [ { name: 'api_platform.filter', id: 'my_dummy.property' } ]
 
+    ApiPlatform\Core\Tests\Fixtures\TestBundle\Filter\RequiredFilter:
+        arguments: [ '@doctrine' ]
+        tags: [ 'api_platform.filter' ]
+
+    ApiPlatform\Core\Tests\Fixtures\TestBundle\Filter\ArrayRequiredFilter:
+        arguments: [ '@doctrine' ]
+        tags: [ 'api_platform.filter' ]
```
soyuka: we're not using spaces after `[` IIRC

src/Filter/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    /**
+     * Test if required filter is valid. It validates array notation too like "required[bar]".
+     */
+    private function isRequiredFilterValid($name, $request): bool
+    {
+        $matches = [];
+        parse_str($name, $matches);
```
teohhanhui: I've mentioned this a long time ago, but we really should stop using bad built-in PHP functions like `parse_str`. We should seriously think about using a proper URI parsing library instead: https://uri.thephpleague.com/5.0/

teohhanhui:
@dunglas I think it's better if we avoid giving the filter implementations the responsibility of checking required filters. It's not their job and it doesn't need to be. But I agree with extracting all logic into a separate class, so as to keep the event listeners really dumb.

dunglas:
Introducing a new service and using it in the listener looks good to me.

jdeniau:
@soyuka about the "light" listener, the separation between the listener and the validation has been made in #1723 ( https://github.com/api-platform/core/pull/1723/files#diff-91ffc72a7db30ec816871e7141558def exactly ) : I prefer keeping code simple first, and re-factor it when needed.
With this modification, I can extract a FilterListValidator, but the only code it will contain is a `foreach($validatorList as $validator) { doValidate(); }`

@dunglas About the namespace, we talked about it previously, but I can move the listener in EventListener and move back the Filter validator in Filter namespace.

Can we keep the listener "as-is" for this PR, and make the refactoring in #1723, as the purpose of this PR is like a POC before #1723 ? (it's quite painful to rebase each time, and I think it will be more painful this time as the code is several months old).

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
soyuka: Still, you could alias it if you think that it'd be used by developers.

src/Filter/QueryParameterValidateListener.php
dunglas: Is this really necessary for this use case? Maybe can we do that at some point in another PR and replace all calls to `parse_str` in 1 time.

dunglas:
LGTM.

tests/Fixtures/TestBundle/Entity/ArrayFilterValidator.php
soyuka: Sure thing! Just remove the id setter if it's auto generated though :)

jdeniau:
@soyuka done

src/Filter/QueryParameterValidateListener.php
soyuka: It's okay if it's refactored later as said in https://github.com/api-platform/core/pull/1692#issuecomment-391711723 I guess?

jdeniau:
@dunglas @meyerbaptiste @Simperfit @soyuka everything is OK on my part I think. Waiting for a merge on master to work on the second PR that will check all documented fields

soyuka:
Thanks @jdeniau !

PR #1695, opened 2018-02-09
Support YAML output from SwaggerCommand with --yaml

### LINKED ISSUE #1688: [Swagger] add --yaml to SwaggerCommand
It would be cool if you'd be able to say `api:swagger:export --yaml` to get valid YAML out (instead of JSON). This way, if you're storing it to a file for human consumption, you get YAML which is easier to diff etc.

Should be simple enough.

dkarlovi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1688 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/389

Closes #1688.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
+++ b/src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
@@ -66,11 +66,7 @@ final class SwaggerCommand extends Command
     {
         $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $this->apiTitle, $this->apiDescription, $this->apiVersion, $this->apiFormats);
         $data = $this->documentationNormalizer->normalize($documentation);
-        if ($input->getOption('yaml')) {
-            $content = Yaml::dump($data);
-        } else {
-            $content = json_encode($data, JSON_PRETTY_PRINT);
-        }
+        $content = $input->getOption('yaml') ? Yaml::dump($data) : json_encode($data, JSON_PRETTY_PRINT);
         $output->writeln($content);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
```diff
@@ -63,7 +66,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $this->apiTitle, $this->apiDescription, $this->apiVersion, $this->apiFormats);
         $data = $this->documentationNormalizer->normalize($documentation);
-        $content = json_encode($data, JSON_PRETTY_PRINT);
+        if ($input->getOption('yaml')) {
```
dunglas: `$input->getOption('yaml') ? Yaml::dump($data) : json_encode($data, JSON_PRETTY_PRINT`?

dkarlovi:
> Can you also add a test in the CI (like https://github.com/api-platform/core/blob/master/.travis.yml#L56) to check with the Swagger validator that the generated YAML file is valid?

Sure.

src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
dkarlovi: If you insist. :+1: I try to avoid those.

dkarlovi:
It seems like this is the output:

```yaml
swagger: '2.0'
basePath: /
info:
    title: 'My Dummy API'
    version: 0.0.0
    description: 'This is a test API.'
paths: null
definitions: null
```

while this is supposed to be the output:

```yaml
swagger: '2.0'
basePath: /
info:
    title: 'My Dummy API'
    version: 0.0.0
    description: 'This is a test API.'
paths: {}
definitions: {}
```

I've added `--no-schema` to the validator just to make it pass, should default these values to an empty object in another PR.

dkarlovi:
@dunglas Don't forget to merge this since you merged the docs describing it already. :)

dunglas:
Thanks @dkarlovi

PR #1696, opened 2018-02-09
Revert "alwaysIdentifier annotation option implementation"

dunglas:
This reverts commit 10a009fdd1d713f84059154767d9be482a8c1ab5.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Digging deeper into #1528, I think that this should be handled at the Symfony level, so I proposed https://github.com/symfony/symfony/pull/26108

I would like to revert #1528 before releasing the stable release.

ping @api-platform/core-team @adelplace


soyuka:
:+1: we will document the proper way of handling this if we can get symfony/symfony#26108 merged! Btw you can add a :+1: to show your interest in that PR!

adelplace:
>  I think that this should be handled at the Symfony level

Yes I totally agree

silverbackdan:
Is there any chance on how to implement this functionality with the PR now merged in Symfony?

Is it worth API Platform having a built-in handler that can be enabled with annotation on specific properties?

soyuka:
@silverbackdan just set this in symfony configuration:

```
serializer:
  max_depth_handler: App\MaxDepthHandler
```

Your service would look like this:

```
<?php
namespace App;

class MaxDepthHandler
{
  public function __invoke($obj) {
    return $obj->id; // or whatever
  }
}
```

silverbackdan:
Thanks @soyuka - so very simple :)

PR #1699, opened 2018-02-10
Added the verify_ssl option for the cache invalidation

mickaelprevot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

When the cache invalidation is enabled, we may encounter the problems with the BAN request and SSL certificate verification.

Indeed, by default the Guzzle client check the peer's SSL certificate but we may not have a certicate SSL on the Varnish server. If so, this produces a fatal error.

To give the possibility to developer's of disable this verification (without using a decorator and a compiler pass), I added the verify_ssl option in the invalidation parameters.

This parameter is at true by default to avoid regression.


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -442,8 +442,13 @@ private function registerHttpCache(ContainerBuilder $container, array $config, X
...
         $definitions = [];
         foreach ($config['http_cache']['invalidation']['varnish_urls'] as $key => $url) {
+            $clientConfig = [
+                'base_uri' => $url,
+                'verify' => $config['http_cache']['invalidation']['verify_ssl'],
+            ];
+
             $definition = new ChildDefinition('api_platform.http_cache.purger.varnish_client');
-            $definition->addArgument(['base_uri' => $url]);
+            $definition->addArgument($clientConfig);
```
Simperfit: You can inline this directly, it does not need to be a variable.

mickaelprevot: Indeed, I modify my PR.

Thanks.

dunglas:
Instead of introducing a new flag for every guzzle options, can't we just add a new `options` setting in API Platform's configuration and pass it straight to Guzzle? This way it will be possible to use all Guzzle features.

mickaelprevot:
Indeed ! That covers just about all cases :)

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -173,6 +173,11 @@ public function getConfigTreeBuilder()
                                     ->prototype('scalar')->end()
                                     ->info('URLs of the Varnish servers to purge using cache tags when a resource is updated.')
                                 ->end()
+                                ->arrayNode('request_options')
+                                    ->defaultValue([])
+                                    ->prototype('scalar')->end()
+                                    ->info('To pass options to the client charged with the request.')
```
dunglas: Some options aren't scalar but arrays, for instance: http://docs.guzzlephp.org/en/stable/request-options.html#allow-redirects

Can you update the config accordingly and add a test for this case please? https://gist.github.com/gergelypolonkai/30440e25f7a447730064 may help.

mickaelprevot: Ok. In your example, the attribute list is hard coding (transition and time). We using Guzzle client, tomorrow it may be another client with other attributes. We should therefore pass through this code again ?

If so, I will also declare options of scalar type in order to be consistent.

mickaelprevot: You use Guzzle client, tomorrow it may be another client with other attributes.
I thinks that the "variableNode" is perfect for this case. What do you thin

dunglas: Yes, variable is perfect.

dunglas:
Thanks @mickaelprevot. Can you open a docs PR please?

PR #1702, opened 2018-02-12
[GraphQL] Fix type for subresource

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/535
| License       | MIT
| Doc PR        |

Nested mutations are not a part of the specification. See:
- https://github.com/webonyx/graphql-php/issues/189
- https://github.com/facebook/graphql/issues/252
- https://github.com/graphql/graphql-js/issues/672
Moreover, it would be laborious to implement it. We would have to change the behavior of the ItemNormalizer and we would have to move the validation/access control of the resolver elsewhere to be reused in the normalizer.
Therefore, it seems a better solution to force IRIs for subresources in mutations.
This PR also fixes an issue with input types: pagination should not be used for them.

### DIFF
--- a/features/graphql/mutation.feature
+++ b/features/graphql/mutation.feature
@@ -54,11 +54,36 @@ Feature: GraphQL mutation support
     And the JSON node "data.createFoo.bar" should be equal to "new"
     And the JSON node "data.createFoo.clientMutationId" should be equal to "myId"
 
+  Scenario: Create an item with a subresource
+    Given there are 1 dummy objects with relatedDummy
+    When I send the following GraphQL request:
+    """
+    mutation {
+      createDummy(input: {_id: 1, name: "A dummy", foo: [], relatedDummy: "/related_dummies/1", clientMutationId: "myId"}) {
+        id
+        name
+        foo
+        relatedDummy {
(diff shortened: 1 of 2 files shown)


dunglas:
Thanks @alanpoulain

Rebolon:
Hi, I understand your choice and difficulties to implement nested mutations. But i find this rather sad. Nested mutations are a solution to solve multiple HTTP call problems from client. A project like GraphCool has choose to implement this feature in 2016/10: https://blog.graph.cool/improving-dx-with-nested-mutations-698c508339ca

Hope it will exists one day in api-platform

soyuka:
Something like batch operations may be more suitable for this particular issue, see https://github.com/api-platform/core/pull/1645

Rebolon:
i have to investigate on batch operations, but to my mind batch are for recurrent operations of the same kind, here operations can be over different object. I'll give it a try (if i find enough documentation about it :) 
)

PR #1710, opened 2018-02-16
Ensure that OrderFilter preserves the query string parameters order

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

When requesting the API with a url like `/foo?order[baz]=asc&order[bar]=asc` the order is not preserved anymore as it used to be since 1.0

The root cause is the Symfony `Request::getQueryString()` not preserving the order when re-generating the string.

For now I've ported the Symfony function and removed the sorting thing for now.

### DIFF
--- a/features/doctrine/order_filter.feature
+++ b/features/doctrine/order_filter.feature
@@ -300,6 +300,67 @@ Feature: Order filter on collections
     }
     """
 
+  Scenario: Get collection ordered collection on several property keep the order
+    # Adding 30 more data with the same name
+    Given there are 30 dummy objects
+    When I send a "GET" request to "/dummies?order[name]=desc&order[id]=desc"
+    Then the response status code should be 200
+    Then print last JSON response
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/Dummy$"},
+        "@id": {"pattern": "^/dummies$"},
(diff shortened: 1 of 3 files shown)


antograssiot:
I'm definitely not a fan of the code duplication that it generates and I'm open to any other option.

features/doctrine/order_filter.feature
```diff
@@ -300,6 +300,67 @@ Feature: Order filter on collections
     }
     """
 
+  Scenario: Get collection ordered collection on several property keep the order
```
antograssiot: @dunglas Should I open a MR to add this test to 2.1 also ?

soyuka:
> I'm definitely not a fan of the code duplication that it generates and I'm open to any other option.

Shouldn't that be fixed in symfony then? If not then :+1:

dunglas:
It should be fixed in SF, but we need a workaround in the meantime.

antograssiot:
I'll open a PR to symfony also @soyuka and remove this when an upsteam patch will be availlable. I also don't know in which version  of symfony it will end in (if it does get merged), and it might force to pin to a higher version than today which can also not be acceptable for this projet...

dunglas:
Thanks @antograssiot

antograssiot:
@soyuka @dunglas I gave a try to allow that in symfony directly as promised.
https://github.com/symfony/symfony/pull/26202

PR #1715, opened 2018-02-19
Allow to declare multiple filters with same class

mykidok:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | not yet

When using multiple filters of the same class with @ApiFilter annotation( for example `GroupFilter::class`, one with option `overrideDefaultGroups:true` and the other one with `overrideDefaultGroups:false` ), only the first one was registered. 
This PR try to fix that by adding an 'id' option to the ApiFilter filter annotation to make it unique per-class and per-filter.

### DIFF
--- a/src/Util/AnnotationFilterExtractorTrait.php
+++ b/src/Util/AnnotationFilterExtractorTrait.php
@@ -87,7 +87,7 @@ trait AnnotationFilterExtractorTrait
 
         foreach ($this->getFilterAnnotations($reader->getClassAnnotations($reflectionClass)) as $filterAnnotation) {
             $filterClass = $filterAnnotation->filterClass;
-            $id = $this->generateFilterId($reflectionClass, $filterClass);
+            $id = $this->generateFilterId($reflectionClass, $filterClass, $filterAnnotation->id);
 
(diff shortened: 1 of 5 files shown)


src/Util/AnnotationFilterExtractorTrait.php
```diff
@@ -89,9 +89,9 @@ private function readFilterAnnotations(\ReflectionClass $reflectionClass, Reader
             $filterClass = $filterAnnotation->filterClass;
             $id = $this->generateFilterId($reflectionClass, $filterClass);
```
antograssiot: @mykidok Why not updating `generateFilterId()` directly ?
The class is `@internal` so adding a default argument is not a BC break

mykidok:
appveyor faillure doesn't seems related

mykidok:
@dunglas is that ok for you ? Should I do a PR for the doc ?

dunglas:
Thanks @mykidok

PR #1716, opened 2018-02-19
throw NotFound exception when use plain identifiers

ewgRa:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

There is PR https://github.com/api-platform/core/pull/1365 that introduce "allow_plain_identifiers". But there is a problem, if you send identifier for entity that not present in database, it try to set it to object as "null", and if this field is required like "setFoo(Foo $foo)" - than we have 500 error.

Default DataProvider never throw ItemNotFound, since have "return $queryBuilder->getQuery()->getOneOrNullResult();", and AbstractItemNormalizer never have "} catch (ItemNotFoundException $e) {"

This PR fix this and behavior when we use "allow_plain_identifiers" becomes same as with iriConverter.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -315,7 +315,13 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             // repeat the code so that IRIs keep working with the json format
             if (true === $this->allowPlainIdentifiers && $this->itemDataProvider) {
                 try {
-                    return $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true]);
+                    $item = $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true]);
+
+                    if (null === $item) {
+                        throw new ItemNotFoundException(sprintf('Item with id "%s" not found for class "%s".', $value, $className));
+                    }
+
+                    return $item;
                 } catch (ItemNotFoundException $e) {
                     throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
                 } catch (InvalidArgumentException $e) {
--- a/tests/Serializer/AbstractItemNormalizerTest.php
+++ b/tests/Serializer/AbstractItemNormalizerTest.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\ItemNotFoundException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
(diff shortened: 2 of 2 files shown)


ewgRa:
travis fail looks like not related

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -311,7 +311,13 @@ protected function denormalizeRelation(string $attributeName, PropertyMetadata $
             // repeat the code so that IRIs keep working with the json format
             if (true === $this->allowPlainIdentifiers && $this->itemDataProvider) {
                 try {
-                    return $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true]);
+                    $item = $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true]);
```
soyuka: It's the `itemDataProvider` that should be throwing th ` ItemNotFoundException`  not the Normalizer.

ewgRa: I think about it, but it can brake backward compatibility. I actually don't know very well api-platform and where ItemDataProvider used and can't predict what happens with all such places if ItemDataProvider start throwing this exception. Is there everything will be fine?

If ok, I will make change tomorrow.

soyuka: Oh my bad you're right, seems we're using doing this in the [ReadListener](https://github.com/api-platform/core/blob/45c537fac12b579aea321ac2170f90abcfcba9ff/src/EventListener/ReadListener.php#L110).

ewgRa: If it is only one place, than we can change it too. Problem only with custom data providers and who use it with "allow_plain_identifiers". They need to change their DataProviders and add throw for this exception. But since nobody complain about this, looks like it will be not a problem. But anyway, here is potential brake of BC for such developers, decision up to you, I don't know how strict you are with BC.

ewgRa: ping @soyuka , so, throw exception at ItemDataProvider and change ReadListener also or leave it as is?

soyuka: Leave it as is for now.

@dunglas I thought that we could add a method in the IriConverter that would do this portion of code:

https://github.com/api-platform/core/blob/45c537fac12b579aea321ac2170f90abcfcba9ff/src/EventListener/ReadListener.php#L110-L124

We could then deduplicate some code, wdyt?

Simperfit:
Could you please rebase with master ?

soyuka:
I think that this has drastically change with the new identifier normalizer stuff. Maybe it's not even needed anymore?

ewgRa:
@soyuka rebased

If I rollback my changes in AbstractItemNormalizer then testDenormalizeNotFoundRelationWithPlainId test still fail

soyuka:
That's okay for now but we really need to improve this. Having itemDataProvider in this class is wrong, I have another fix in mind but I'm not sure if it would work.

flug:
@ewgRa Up is this pull request still valid?

ewgRa:
Actually I don't know, do not follow latest changes.

Since it is more than one year old, I think it can be closed without pain.

soyuka:
Closing let's keep track of this in #2686

PR #1687, opened 2018-02-08
remove @dropSchema and drop the schema in @createSchema instead

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1642 
| License       | MIT
| Doc PR        | ~

As told in #1642, I don't think `@dropSchema` is usefull. It is really easier to just drop the schema before creating it.

It simplifies the mental process of contributing to api-platform (especially in files [containing multiple `@dropSchema` / `@createSchema`](https://github.com/api-platform/core/blob/v2.2.0-beta.1/features/doctrine/date_filter.feature) )

I think the hint about removing the database file in `CONTRIBUTING.md` does not stands anymore as the database will be dropped on the first `@createSchema` call.

PR #1722, opened 2018-02-21
Fix wrong automatic configuration for Sf flex

Nek-:
This was an issue for some people that decided to change the way the folder
of configuration of api platform is defined.

The idea of this fix is to not interfer with the user configuration.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none, but still an issue
| License       | MIT
| Doc PR        | N/A

# Reproducer
```yaml
# api platform configuration
api_platform:
    mapping:
        paths: ['%kernel.project_dir%/config/api_platform/resources']
```

_Then imagine having the `%kernel.project_dir%/config/api_platform/routes` folder._

# Before the fix
The dump of resources folders is:

```
array:2 [
  0 => "/var/www/apps/cadi-back/config/api_platform/resources"
  1 => "/var/www/apps/cadi-back/config/api_platform"
]
```

# After the fix
The dump of resources folder is:
```
array:1 [
  0 => "/var/www/apps/cadi-back/config/api_platform/resources"
]
```

### DIFF
--- a/features/main/configurable.feature
+++ b/features/main/configurable.feature
@@ -44,6 +44,7 @@ Feature: Configurable resource CRUD
     }
     """
 
+  @dropSchema
   Scenario: Retrieve the ConfigDummy resource
     When I send a "GET" request to "/fileconfigdummies/1"
     Then the response status code should be 200
@@ -58,22 +59,5 @@ Feature: Configurable resource CRUD
       "id": 1,
       "name": "ConfigDummy",
       "foo": "Foo"
-    }
-    """
-
-  @dropSchema
-  Scenario: Entities can be configured using a Flex-like directory structure
-    When I send a "GET" request to "/flex_configs"
-    Then the response status code should be 200
-    And the response should be in JSON
-    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
-    And the JSON should be equal to:
-    """
-    {
-         "@context": "/contexts/FlexConfig",
-         "@id": "/flex_configs",
-         "@type": "hydra:Collection",
-         "hydra:member": [],
-         "hydra:totalItems": 0
     }
     """
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -246,13 +246,14 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
     private function getResourcesToWatch(ContainerBuilder $container, array $resourcesPaths): array
     {
-        // Flex structure
+        $paths = array_unique(array_merge($resourcesPaths, $this->getBundlesResourcesPaths($container)));
+
+        // Flex structure (only if nothing specified)
         $projectDir = $container->getParameter('kernel.project_dir');
-        if (is_dir($dir = "$projectDir/config/api_platform")) {
-            $resourcesPaths[] = $dir;
+        if (!$paths && is_dir($dir = "$projectDir/config/api_platform")) {
+            $paths = [$dir];
         }
 
-        $paths = array_unique(array_merge($resourcesPaths, $this->getBundlesResourcesPaths($container)));
         $resources = ['yml' => [], 'xml' => [], 'dir' => []];
 
         foreach ($paths as $path) {


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -246,13 +246,14 @@ private function getBundlesResourcesPaths(ContainerBuilder $container): array
...
     private function getResourcesToWatch(ContainerBuilder $container, array $resourcesPaths): array
     {
-        // Flex structure
+        $paths = array_unique(array_merge($resourcesPaths, $this->getBundlesResourcesPaths($container)));
+
+        // Flex structure (only if nothing specified)
         $projectDir = $container->getParameter('kernel.project_dir');
-        if (is_dir($dir = "$projectDir/config/api_platform")) {
-            $resourcesPaths[] = $dir;
+        if (is_dir($dir = "$projectDir/config/api_platform") && empty($paths)) {
```
dunglas: `if (!$paths && is_dir($dir = "$projectDir/config/api_platform"))`?
It's a micro-optim, but then...

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -246,13 +246,14 @@ private function getBundlesResourcesPaths(ContainerBuilder $container): array
...
+            $paths = array_unique(array_merge($paths, [$dir]));
```
Taluu: `$paths[] = $dir;` ? As `$paths` is already empty, calling a `array_unique` + `array_merge` seems waaay overkill.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
Nek-: On something cached :p .

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
Nek-: indeed !!

Nek-: FIxed

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
Nek-: Fixed

dunglas:
A rebase should make Travis happy.

Nek-:
I dropped a test that was not relevant anymore. (as the fix is to not override the default configuration which is needed for other tests)

Nek-:
I can't run the full testsuite on my computer (it always run out of memory). Issues in tests look not related.

dunglas:
`php -dmemory_limit=-1 ...`? It looks related because they are green on the other branches: https://travis-ci.org/api-platform/core/builds

features/main/configurable.feature
```diff
@@ -60,20 +60,3 @@ Feature: Configurable resource CRUD
       "foo": "Foo"
     }
     """
-
-  @dropSchema
```
antograssiot: Without knowing if the test you're removing is relevant, if you are removing it, this annotation should be moved to the previous one, it likely causes the testsuite to fail

dunglas:
I would like to keep this test. We need to be sure that API Platform works with Flex out of the box. What we can do is creating a new kernel for that (with no configured `paths`).

Nek-:
@dunglas I agree with you it could be testable with another kernel. But this is a huge work to make the Symfony extension work with many kernels.

Actually, I believe that the test suite of api platform should not use the Symfony extension because of that and many tests/features are designed in the way to make tests work with only one kernel in the test suite. (I have this feeling since the first look I had on the test suite)

As this work is not trivial I'm not going to add it to my PR. Feel free to use my work anyway.

features/main/configurable.feature
Nek-: Here we go : https://github.com/api-platform/core/issues/1724 :')

jdeniau: #1687 will fix that ;)

dunglas:
> Actually, I believe that the test suite of api platform should not use the Symfony extension because of that and many tests/features are designed in the way to make tests work with only one kernel in the test suite. (for the record I have this feeling since the first look I had on the test suite)

I agree, the Symfony extension (and Behatch) are limiting factors. The Behat test suite is kinda old (pre-1.0 time...) and can be modernized. I'll try to improve that.

dunglas:
Thanks @Nek-

PR #1723, opened 2018-02-22
More query parameter validation (follow #1692)

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | yes
| BC breaks?    | if filter config was wrongly configured
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1627 
| License       | MIT
| Doc PR        | Doc PR will follow if PR is approved

This follows #1692 

It implements all other validation filters from [OpenApi specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object)

This might break wrongly configured api though. I see two possible solutions here:

  * make this a `[Might break]` as we can consider wrongly configured sites a "bug"
  * add a new configuration key `validate_filters`, default to `false`, and deprecate it being to `false`, encouraging people to set it to `true`, and remove the configuration key in 3.0

### DIFF
--- a/src/EventListener/QueryParameterValidateListener.php
+++ b/src/EventListener/QueryParameterValidateListener.php
@@ -43,7 +43,8 @@ final class QueryParameterValidateListener
             !$request->isMethodSafe()
             || !($attributes = RequestAttributesExtractor::extractAttributes($request))
             || !isset($attributes['collection_operation_name'])
-            || 'get' !== ($operationName = $attributes['collection_operation_name'])
+            || !($operationName = $attributes['collection_operation_name'])
+            || 'GET' !== $request->getMethod()
         ) {
             return;
         }
--- a/src/EventListener/QueryParameterValidateListener.php
+++ b/src/EventListener/QueryParameterValidateListener.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\EventListener;
+
+use ApiPlatform\Core\Filter\QueryParameterValidator;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+
+/**
+ * Validates query parameters depending on filter description.
+ *
+ * @author Julien Deniau <julien.deniau@mapado.com>
+ */
+final class QueryParameterValidateListener
+{
+    private $resourceMetadataFactory;
+
+    private $queryParameterValidator;
+
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, QueryParameterValidator $queryParameterValidator)
+    {
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->queryParameterValidator = $queryParameterValidator;
+    }
+
+    public function onKernelRequest(RequestEvent $event)
+    {
+        $request = $event->getRequest();
+        if (
+            !$request->isMethodSafe()
+            || !($attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !isset($attributes['collection_operation_name'])
+            || 'get' !== ($operationName = $attributes['collection_operation_name'])
+        ) {
+            return;
+        }
+
+        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
+        $resourceFilters = $resourceMetadata->getCollectionOperationAttribute($operationName, 'filters', [], true);
+
+        $this->queryParameterValidator->validateFilters($attributes['resource_class'], $resourceFilters, $request);
+    }
+}
--- a/src/EventListener/QueryParameterValidateListener.php
+++ b/src/EventListener/QueryParameterValidateListener.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\EventListener;
 use ApiPlatform\Core\Filter\QueryParameterValidator;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use ApiPlatform\Core\Util\RequestParser;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
 
 /**
@@ -46,10 +47,12 @@ final class QueryParameterValidateListener
         ) {
             return;
         }
+        $queryString = RequestParser::getQueryString($request);
+        $queryParameters = $queryString ? RequestParser::parseRequestParams($queryString) : [];
 
         $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
         $resourceFilters = $resourceMetadata->getCollectionOperationAttribute($operationName, 'filters', [], true);
 
-        $this->queryParameterValidator->validateFilters($attributes['resource_class'], $resourceFilters, $request);
+        $this->queryParameterValidator->validateFilters($attributes['resource_class'], $resourceFilters, $queryParameters);
     }
 }
--- a/src/Filter/Validator/ValidatorInterface.php
+++ b/src/Filter/Validator/ValidatorInterface.php
@@ -13,9 +13,12 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Filter\Validator;
 
-use Symfony\Component\HttpFoundation\Request;
-
 interface ValidatorInterface
 {
-    public function validate(string $name, array $filterDescription, Request $request): array;
+    /**
+     * @var string        the parameter name to validate
+     * @var array<string, mixed> $filterDescription the filter descriptions as returned by `ApiPlatform\Core\Api\FilterInterface::getDescription()`
+     * @var array<string, mixed> $queryParameters the list of query parameter
+     */
+    public function validate(string $name, array $filterDescription, array $queryParameters): array;
 }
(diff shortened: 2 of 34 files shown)


jdeniau:
I'm kind of stuck with the error here : I rebased on master, but https://github.com/api-platform/core/pull/1723/commits/4cc61524f86b1544acff0993bb30be5432033e9b now throw an error `"debugMessage": "Cannot return null for non-nullable field __InputValue.type.",` in `features/graphql/introspection.feature`, apparently because of  `tests/Fixtures/TestBundle/Filter/BoundsFilter.php::getDescription` return !

Can somebody knowing GraphQL help me ?

jdeniau:
OK I found the issue, it come from the filters containing `'type' => 'number' ` like:

```
          'exclusiveMinimum' => [                                                                                                                                            
                  'property' => 'exclusiveMinimum',                                                                                                                              
                  'type' => 'number',                                                                                                                                            
                  'required' => false,                                                                                                                                           
                  'swagger' => [                                                                                                                                                 
                      'minimum' => 5,                                                                                                                                            
                      'exclusiveMinimum' => true,                                                                                                                                
                  ],                                                                                                                                                             
              ],            
```

If I set "string" or "int", the test does work (but the new tests does not work of course).

What is the real definition behind the `type` ? I mapped this PR upon the open-api specification, but apparently the GraphQL code use this to get a PHP type ?

I can "hack" and allow "int" and "float" inside this PR but I don't know if it's a good way to make this regarding api-platform point of view.
Another way to do this is to "override" the `type` in the `swagger` key, but it would be really redundant and sensible to developer error 

Thanks

soyuka:
@jdeniau could you rebase this? Thanks!

jdeniau:
@dunglas @meyerbaptiste @soyuka Everything should be good for me, except the problem with "number" type that I mentionned earlier.

jdeniau:
I finally got time to rework on that, and made the same choice as defined in https://github.com/api-platform/core/blob/5b5e6b79b7ecb705ae6a091d106e7d95a2afe8f8/src/Swagger/Serializer/DocumentationNormalizer.php#L658-L660 , thus if type is documented as `float`, it will be converted to `number` for openapi.

This way, it's more "PHP-friendly".

jdeniau:
@dunglas @meyerbaptiste @soyuka Ready for review

jdeniau:
Hi @alanpoulain ,

I just rebased this PR on master, but the behat tests for mongo are failing. I'd like to check / fix them, but can you helping on how to test with mongo ?

Actually I got this exception : 
```
No suitable servers found (`serverSelectionTryOnce` set): [connection refused calling ismaster on '127.0.0.1:27017'] (MongoDB\Driver\Exception\ConnectionTimeoutException)
```
(which is quite normal as I do not have a mongo server instance on my machine ;)

jdeniau:
This PR is ready for review @api-platform/teams/core-team

Thank you guys !

jdeniau:
This PR is ready for review @api-platform/core-team

Thank you guys !

jdeniau:
@soyuka Well, it's only seven call, and it checks the output of `getDescription` so I don't think the performance impact is huge.
The fact is that this PR may break some existing application if they are wrongly configured, so adding a parameter may make sense if it is disabled by default (and maybe deprecate the fact that it's disabled, encouraging people to either enable this feature or disable it explicitly in their configuration).

For the `final`, I'm pushing it right now

src/Filter/Validator/ValidatorInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Filter\Validator;
+
+use Symfony\Component\HttpFoundation\Request;
+
+interface ValidatorInterface
+{
+    public function validate(string $name, array $filterDescription, Request $request): array;
```
alanpoulain: I think we should use a context instead of depending of the HttpFoundation Request, like the Doctrine filters. It will more independent / maintainable this way. WDYT?

dunglas: Good catch @alanpoulain. We try to remove the hard dependency to HttpFoundation to be able to support PSR-7 at some point. The way to go is using the context.

jdeniau: As the listener is a Symfony request listener, do you think I should only call `$request->query->all()` ?

alanpoulain: You should have a look at the https://github.com/api-platform/core/blob/master/src/EventListener/ReadListener.php.

jdeniau:
@alanpoulain @dunglas changed from the symfony request instance to an array of query parameters.

@alanpoulain fixed other returns

flug:
@jdeniau Up is this pull request still valid?

jdeniau:
@flug yes! I am waiting for it to be merged!

I can spend some time to rebase and see if it's still  valid.

I externalized my work [here](https://github.com/jdeniau/api-platform-filter-validator) in the meantime

flug:
@jdeniau maybe rebase and resovle you conflicts :D

src/EventListener/QueryParameterValidateListener.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+        if (
+            !$request->isMethodSafe(false)
+            || !($attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !isset($attributes['collection_operation_name'])
+            || 'get' !== ($operationName = $attributes['collection_operation_name'])
```
TracKer: You are checking for operation name here, but probably it's better to check for method, since operation names may differ, or even there can be multiple operations for method GET.

jdeniau:
@TracKer @flug  rebase done.
The only failing test is also failing on master branch

soyuka:
Let's merge to avoid further conflicts, we can always come back to some parts of the code if there's smth to change. Thanks a lot @jdeniau for the work!

jdeniau:
@soyuka Thanks a lot :pray: 

If there is anything to change, I will try to be as quick as possible !

norkunas:
Does this allows to configure a `SearchFilter` with some property as required?

jdeniau:
@norkunas Yes you can.

See https://github.com/jdeniau/api-platform-core/blob/3ec427e85ccaeff23aca620335bdd0b7c5da3ded/tests/Fixtures/TestBundle/Filter/RequiredFilter.php to do how.

For the record this, the `required` is operational  this since [2.3.0](https://github.com/api-platform/core/releases/tag/v2.3.0). This PR adds `required allowEmptyValue` features and all swagger specifications (array items, maximum, minimum, enum, minLength, maxLength, multipleOf, pattern)

norkunas:
Looking at example I'd need to create a custom filter class which would apply the "required" as true, yes? Instead I'd expect something like:
```
/**
 * @ApiFilter(
 *     SearchFilter::class,
 *     properties={
 *         "offer.id": { "strategy": "exact", "required": true }
 *    }
 * )
 */
class Entity {}
```

soyuka:
I guess we need docs now :D

jdeniau:
Totally! I will work on it!

Le ven. 20 déc. 2019 à 10:55, Antoine Bluchet <notifications@github.com> a
écrit :

winzza:
How to use it? Is there an example?
I need `SearchFilter::class` with `required=true`. I would also like to dynamically set `required = true` (dependent on user role). How can i do this?

my temporaty bad solution:
```

/**
 * @ApiResource(
 *     // ... rules
 * )
 * @ApiFilter(SearchRequiredFilter::class, properties={"foo": "exact", "bar": "exact"})
 * @ORM\Entity(repositoryClass="App\Repository\ProjectChatRepository")
 */
class User
{ ... }
```
and SearchRequiredFilter class:
```
//src/Filter
<?php

declare(strict_types=1);

namespace App\Filter;

// ... many use ...

class SearchRequiredFilter extends SearchFilter
{
    /**
     * @var Security
     */
    private $security;

    public function __construct(
        Security $security,
        // ... parent params
    ) {
        parent::__construct(
           // parents property
        );

        $this->security = $security;
    }

    {
        $description = parent::getDescription($resourceClass);

        foreach ($description as &$options) {
            if (!$options['is_collection']
                && $this->security->isGranted('ROLE_USER')
            ) {
                $options['required'] = true;
            }
        }

        return $description;
    }
}
```
I understand this is a very bad solution, but how to do my task correctly?

jdeniau:
First point : I will work on documentation soon. It is on my to-do list.

Second point : your solution seems good to me, as it is a complicated case,
you will need to have a complex appproach for that. The only point may be
to not extend SearchFilter but create a real custom filter.

efpapado:
Would it be possible to run a custom validation callback on a filter that is **not** applied to any entity property?
So, like binding the validator class to the filter itself, and not to any entity property.

jdeniau:
@efpapado for the moment, there are only "swagger" filter validators. (See https://github.com/api-platform/core/blob/master/src/Filter/QueryParameterValidator.php#L35-L43 for more informations)

If you want to run a custom callback in a custom filter file, I think you can easily do something like that:

```php
class FooFilter implements FilterInterface
{
      public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
    {
        $this->throwIfSomethingIsWrong();

        // normal filter behaviour
    }
}
```

efpapado:
Thank you so much!!! I wasn't aware of the `apply` method because it's not in the `FilterInterface` :)

jdeniau:
@efpapado It is not available in the base [ApiPlatform\Core\Api\FilterInterface](https://github.com/api-platform/core/blob/master/src/Api/FilterInterface.php), but it is available in the [doctrine bridge ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\FilterInterface](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/FilterInterface.php).

It is mentionned in [the documentation about custom filter](https://api-platform.com/docs/core/filters/#creating-custom-doctrine-orm-filters), but in the documentation, the easier `AbstractContextAwareFilter` class is prefered for it's simplicity.

Theoretically, you can hook into both `apply` or `filterProperty` to throw an error whenever you want.

PR #1726, opened 2018-02-22
Fix initialization of the serialization context

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

According to https://github.com/api-platform/core/blob/fb7cc479046c03352d96dedc57d97f1dbc3fe323/src/EventListener/ReadListener.php#L73, the normalization context array is in the `normalization_context` index of the context array. So, if we pass the whole context array to the `EagerLoadingExtension::joinRelations()` method, we can't properly retrieve groups, max depth and attributes.

### DIFF
--- a/src/EventListener/ReadListener.php
+++ b/src/EventListener/ReadListener.php
@@ -70,8 +70,8 @@ final class ReadListener
         $context = null === $filters ? [] : ['filters' => $filters];
         if ($this->serializerContextBuilder) {
             // Builtin data providers are able to use the serialization context to automatically add join clauses
-            $context['normalization_context'] = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
-            $request->attributes->set('_api_normalization_context', $context['normalization_context']);
+            $context += $normalizationContext = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
+            $request->attributes->set('_api_normalization_context', $normalizationContext);
         }
 
         $data = [];


meyerbaptiste:
Okay, it doesn't work in the case of deserialization... I'll investigate.

meyerbaptiste:
I think we should fix this in the `ReadListener` class instead, like that it would work with deserialization too.
Indeed, when we call `IriConverterInterface::getItemFromIri()` during deserialization for example, the denormalization context is at the first level in the context array and not in a `denormalize_context` index.

WDYT @api-platform/core-team?

soyuka:
+1 about the second option. I'm not sure that this context is exposed to the public, I mean isn't it an internal thing only?

meyerbaptiste:
Huum, not really. It's exposed when you implement `ContextAwareQueryCollectionExtensionInterface` and `ContextAwareCollectionDataProviderInterface` for example...

dunglas:
👍 for the second option. It's kinda internal and it has been introduced in 2.2. It's now or never...

soyuka:
needs to target 2.2 as a bug fix then?

meyerbaptiste:
Right, I didn't see that this branch was created...

dunglas:
Thanks @meyerbaptiste

PR #1734, opened 2018-02-25
BugFix: Fix object state inconsistency after persitence

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

The state of the object presented after a POST/PUT operation doesn't reflects the actual status of the persisted object(s).

With an entity declared with `@ORM\Column(type="decimal", precision=4, scale=1, nullable=false)` for example:

# before:

POST/PUT request to the entity with value
```
{
     "value": 20.334"
}
```

would return 
```
{
   ...
   "value": 20.334"
   ...
}
```

but then the GET would return 
```
{
   ...
   "value": 20.3"
   ...
}
```

# After
The exposed values  in GET/PUT/POST are always consistent
```
{
   ...
   "value": 20.3"
   ...
}
```

# Misc
I also found that tests marked with the `@postgres` tag were never run on the CI, so I've fixed that also

# TODO
- [x] ensure postgres suite test are running on the CI
- [x] add a failling test
- [x] Update the status after persistence
- [x] Update false positives tests

### DIFF
--- a/features/main/relation.feature
+++ b/features/main/relation.feature
@@ -72,9 +72,7 @@ Feature: Relations support
       },
       "relatedToDummyFriend": [],
       "dummyBoolean": null,
-      "embeddedDummy": null,
-      "id": 1,
-      "symfony": "symfony",
+      "embeddedDummy": [],
       "age": null
     }
     """
@@ -476,8 +474,7 @@ Feature: Relations support
       },
       "relatedToDummyFriend": [],
       "dummyBoolean": null,
-      "embeddedDummy": null,
-      "symfony": "symfony",
+      "embeddedDummy": [],
       "age": null
     }
     """
--- a/tests/Fixtures/TestBundle/Entity/DummyTableInheritanceRelated.php
+++ b/tests/Fixtures/TestBundle/Entity/DummyTableInheritanceRelated.php
@@ -78,4 +78,29 @@ class DummyTableInheritanceRelated
 
         return $this;
     }
+
+    /**
+     * @param $child
+     *
+     * @return $this
+     */
+    public function addChild($child)
+    {
+        $this->children->add($child);
+        $child->setParent($this);
+
+        return $this;
+    }
+
+    /**
+     * @param $child
+     *
+     * @return $this
+     */
+    public function removeChild($child)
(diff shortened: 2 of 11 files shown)


features/main/relation.feature
```diff
@@ -72,9 +72,7 @@ Feature: Relations support
       },
       "relatedToDummyFriend": [],
       "dummyBoolean": null,
-      "embeddedDummy": null,
-      "id": 1,
-      "symfony": "symfony",
+      "embeddedDummy": [],
       "age": null
```
antograssiot: `id` and `symfony` properties where present twice, and it is execped to get an empty array on `embeddedDummy`

tests/Fixtures/TestBundle/Entity/DummyTableInheritanceRelated.php
```diff
@@ -78,4 +78,29 @@ public function setChildren($children)
 
         return $this;
     }
+
+    /**
+     * @param $child
+     *
+     * @return $this
+     */
+    public function addChild($child)
```
antograssiot: the adder and remover were missing and dur to the same bug where shown in the POST response but never really persisted, now that the state is refresh, it needed to be persisted and thus also fixed this false positive test result

dunglas:
Awesome. Thank you very much @antograssiot

emancypage:
Hi. Before I create new Issue maybe I create comment here because my problem is strongly connected with behavior described here. 
I noticed that identical problem with decimal type still exist in POST method but for sub-resources.

Example:
With an entity declared with `@ORM\Column(type="decimal", precision=4, scale=1, nullable=false)` for example:
POST:
```
{
     "value": 20.334",
     "subresource": {
           "value": 50.321"
     }
}
```
would return
```
{
     "value": 20.3",
     "subresource": {
           "value": 50.321"
     }
}
```

It should return:
```
{
     "value": 20.3",
     "subresource": {
           "value": 50.3"
     }
}
```

PR #1740, opened 2018-03-01
Add failing test for value object denormalization

Nek-:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | n
| License       | MIT
| Doc PR        | none

ApiPlatform does not support value object as a resource (which is weird, agree) but imagine you want to set all the data of your resource via the constructor (which is very similar to support value objects), it will fail.

### DIFF
--- a/tests/Serializer/ItemNormalizerTest.php
+++ b/tests/Serializer/ItemNormalizerTest.php
@@ -22,6 +22,7 @@ use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Metadata\Property\PropertyNameCollection;
 use ApiPlatform\Core\Serializer\ItemNormalizer;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyValueObject;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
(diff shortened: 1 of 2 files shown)


tests/Serializer/ItemNormalizerTest.php
```diff
@@ -188,4 +189,34 @@ public function testDenormalizeWithIdAndUpdateNotAllowed()
         $normalizer->setSerializer($serializerProphecy->reveal());
         $normalizer->denormalize(['id' => '12', 'name' => 'hello'], Dummy::class, null, $context);
     }
+
+    public function testItDenormalizeValueObjects()
+    {
+        $context = ['resource_class' => DummyValueObject::class, 'api_allow_update' => true];
+
+        $propertyNameCollection = new PropertyNameCollection(['name']);
```
antograssiot: I'm trying to understand this, but this exact same test would pass just by replacing by the following 
```php
$propertyNameCollection = new PropertyNameCollection(['foo', 'bar']);
```
what it should be if I understand this correctly, and
```php
$propertyMetadataFactoryProphecy->create(DummyValueObject::class, 'foo', [])->willReturn($propertyMetadataFactory)->shouldBeCalled();
$propertyMetadataFactoryProphecy->create(DummyValueObject::class, 'bar', [])->willReturn($propertyMetadataFactory)->shouldBeCalled();
``` 
line 203

Apologizes if I'm missing the point, sorry

Nek-: The same result is to expect. But i'll fix this little mistake :) .

Nek-:
The problem is somewhere else (there's still probably one 😄 gonna investigate)

PR #1742, opened 2018-03-01
Strip slashes from definitionKey

silverbackdan:
To allow shortNames with slashes and avoid swagger ui resolver error

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

I'd like to update some resource's short names to include a prefix - e.g. component/hero instead of just hero.

It seemed the simplest way I found in previous issues was to change the short name. It works well except in the swagger ui which results in a Resolver error where it cannot resolve the reference if it has another slash in it.

The easiest fix I could think would be to strip the slashes as I've done in this PR.

Alternatively I'd look to tag each resource to group them together is they are an instance on AbstractComponent in my case. But tagging doesn't seem so easy to do dynamically at the moment, even with the DocumentationNormalizer.

Happy for feedback if there's a better solution though. I'd personally quite like a 'route_prefix' option instead but I'm not sure how useful it'd be for others.

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -451,6 +451,7 @@ final class DocumentationNormalizer implements NormalizerInterface
         } else {
             $definitionKey = $this->getDefinitionKey($resourceMetadata->getShortName(), (array) ($serializerContext[AbstractNormalizer::GROUPS] ?? []));
         }
+        $definitionKey = str_replace('/', '-', $definitionKey);
 
         if (!isset($definitions[$definitionKey])) {
             $definitions[$definitionKey] = [];  // Initialize first to prevent infinite loop


silverbackdan:
Not sure this is ideal really - I also had custom routes/controllers setup for one of my resources and there was an error because that custom route wasn't using the same short name as I had defined on the entity.

The documentation normalizer isn't really that flexible either. The object passed in is just an array of the classes and the docs outputted never reference the full class names so I'm not sure there's a very good way of adding documentation based on the resource/entity class for a given path or definition. This makes it tricky to dynamically add tags for the docs.

I'm not sure I like the hack of using the short name to prefix the route path either. The real problem I was having was to group the resources in the Swagger UI nicely which should really be done with tags as far as I can see.

Would you consider a PR to add a `tags` attribute to the `ApiResource` annotation? Happy to work on one if it wouldn't be a waste of time. Cheers.

Simperfit:
I guess the tag feature will be better, could you take care of this ? I'm closing this one, it seems not the right fix to me too.

silverbackdan:
@Simperfit Will do, I'll look at this PR as well as the enable_max_depth improvement as soon as I'm able to. Thanks!

Issue #1714, opened 2018-02-19
Search Filter does not filter collection by nested relation field in GraphQL


k911:
Hi!
I've been recently testing GraphQL endpoint provided by this amazing framework.
So I've encountered on follwing SearchFilter issue.

<details>
<summary>
I've following entities (simplified):
</summary>

```php
// src/Entity/Article.php

/**
 * @ApiResource(
 *     attributes={
 *         "filters": {"app.article.search_filter", "app.article.group_filter"},
 *         "normalization_context": {"groups": {"ArticleRead"}},
 *     },
 *     graphql={
 *         "query": {
 *             "filters": {"app.article.search_filter"},
 *             "normalization_context": {"groups": {"ArticleRead", "CategoryRead"}},
 *         },
 *     },
 * })
 * @Entity
 */
class Article
{
    /**
     * @ORM\Id
     * @ORM\Column(type="uuid")
     * @ORM\GeneratedValue(strategy="CUSTOM")
     * @ORM\CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidGenerator")
     *
     * @Groups({"ArticleRead"})
     */
    protected $id;

    ...

    /**
     * @ORM\ManyToOne(targetEntity="Category")
     * @ORM\JoinColumn(name="category_id", referencedColumnName="id", onDelete="CASCADE", nullable=false)
     *
     * @Assert\NotBlank
     *
     * @Groups({"ArticleRead"})
     */
    protected $category;
}
```

```php
// src/Entity/Category.php

/**
 * @ApiResource
 * @Entity
 */
class Category
{
    /**
     * @ORM\Id
     * @ORM\Column(type="uuid")
     * @ORM\GeneratedValue(strategy="CUSTOM")
     * @ORM\CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidGenerator")
     *
     * @Groups({"CategoryRead"})
     */
    protected $id;

    /**
     * @ORM\Column(type="string", unique=true)
     *
     * @Gedmo\Slug(fields={"name"}, separator="-", updatable=true, unique=true)
     *
     * @Groups({"CategoryRead"})
     */
    protected $code;

    /**
     * @ORM\Column(type="string", nullable=false)
     *
     * @Assert\NotBlank
     * @Assert\Length(min="3", max="100")
     *
     * @Groups({"CategoryRead"})
     */
    protected $name;
}
```
</details>

<details>
<summary>
and I've declared following filters (old way):
</summary>

[...]

PR #1743, opened 2018-03-02
fix property name in ContextAwareFilter to support GraphQL

mneuhaus:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1714
| License       | MIT
| Doc PR        | -

GraphQL filter properties use a underscore instead of dot to separate nested fields, which causes some issues like #1714 where nested filters are not applied. This change replaces the underscore with a dot in the AbstractContextAwareFilter to have one common syntax further down.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/AbstractContextAwareFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/AbstractContextAwareFilter.php
@@ -14,7 +14,10 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Filter;
 
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Dkd\Populate\Exception;
 use Doctrine\ORM\QueryBuilder;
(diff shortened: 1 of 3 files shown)


mneuhaus:
any suggestion how/were i should fix this? :)

dunglas:
Using https://github.com/api-platform/core/blob/2.2/src/Util/RequestParser.php#L39 to retrieve the filters should do the trick.

Simperfit:
@mneuhaus Do you want to try what @dunglas suggested ?

mneuhaus:
Hey, yes i did, but sadly got stuck in there and did not manage to fix it there :/

Simperfit:
Did you push the WIP ? Ill try to help you ;).

antograssiot:
I think a better place to try fix it would be https://github.com/api-platform/core/blob/master/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php#L80

Simperfit:
This is fixed by @antograssiot. in #1868

Thanks anyway for this PR!

Issue #1747, opened 2018-03-05
Non anemic doctrine models are not supported


Nek-:
I'd like to write my model this way:

```php
class Mail
{
    private $from;
    private $to;
    private $content;
    public function __construct($from, $to, $content)
    {
        $this->from = $from;
        $this->to = $to;
        $this->content = $content;
    }
}
```

My model looks like this following DDD guidelines.

But with ApiPlatform having this model with a correct (standard) configuration will result in an error of deserialization because fields are not considered as "writable" by the [`DoctrineOrmPropertyMetadataFactory`](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php) class. Which is wrong on a `Serializer` point of view... and even on a Doctrine point of view as this entity is perfectly valid and its fields are also perfectly writable.

To me, this is an issue but @soyuka seems to disagree after a short discussion on slack.

There are many solutions for this problem inside ApiPlatform:
- ~patch the `DoctrineOrmPropertyMetadataFactory`~ 
- patch the `SerializerPropertyMetadataFactory` (after all if I say they are writable, then they are)
- probably many more as there's also a "configuration" that may set the property writable by default (adding its documentation would be a great help)

There's a last thing about adding a custom `PropertyMetadataFactoryInterface` (that still needs some documentation) but I don't think it's relevant as to me, ApiPlatform must support this by default (to be compliant with Doctrine and Symfony)

[edit] this has nothing to do with the `DoctrineOrmPropertyMetadataFactory`.

dunglas:
For the record, I'm to support this kind of objects. But it will be a tedious work.

greg0ire:
Another possibility if you're stuck might be to create a DTO to represent the writable fields, then validate it with the Validation component, and if it is valid and only then, attempt to update your entity from that.

soyuka:
> For the record, I'm to support this kind of objects. But it will be a tedious work.

You mean by using the constructor to set the properties?

dunglas:
Yes, some improvements have been done in the Symfony serializer to support this use case.

soyuka:
> Yes, some improvements have been done in the Symfony serializer to support this use case.

Though, even with this implemented, these properties will still be considered as readonly if they don't have any setters.

dunglas:
Until we patch API Platform :)

DonCallisto:
@greg0ire 

[...]

PR #1749, opened 2018-03-05
Make resource constructor parameters writables

Nek-:
This is motivated because other tools are already constructor friendly (ie Symfony serializer and Doctrine).

Also using constructors must be recommended and not supporting them is a serious feature missing.

| Q             | A
| ------------- | ---
| Bug fix?      | somehow
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/issues/1747
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
+++ b/src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
@@ -68,6 +68,22 @@ final class PropertyInfoPropertyMetadataFactory implements PropertyMetadataFacto
             $propertyMetadata = $propertyMetadata->withWritable($writable);
         }
 
+        if (method_exists($this->propertyInfo, 'isInitializable')) {
+            if (null === $propertyMetadata->isInitializable() && null !== $initializable = $this->propertyInfo->isInitializable($resourceClass, $name, $options)) {
+                $propertyMetadata = $propertyMetadata->withInitializable($initializable);
+            }
+        } else {
+            // BC layer for Symfony < 4.2
+            $ref = new \ReflectionClass($resourceClass);
+            if ($ref->isInstantiable() && $constructor = $ref->getConstructor()) {
+                foreach ($constructor->getParameters() as $constructorParameter) {
+                    if ($constructorParameter->name === $name && null === $propertyMetadata->isInitializable()) {
+                        $propertyMetadata = $propertyMetadata->withInitializable(true);
+                    }
+                }
+            }
+        }
+
         return $propertyMetadata;
     }
 }
--- a/src/Metadata/Property/PropertyMetadata.php
+++ b/src/Metadata/Property/PropertyMetadata.php
@@ -34,8 +34,9 @@ final class PropertyMetadata
     private $childInherited;
     private $attributes;
     private $subresource;
+    private $initializable;
 
-    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null)
+    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null, bool $initializable = null)
     {
         $this->type = $type;
         $this->description = $description;
@@ -49,6 +50,7 @@ final class PropertyMetadata
         $this->childInherited = $childInherited;
         $this->attributes = $attributes;
         $this->subresource = $subresource;
+        $this->initializable = $initializable;
     }
 
     /**
@@ -378,6 +380,31 @@ final class PropertyMetadata
     {
         $metadata = clone $this;
         $metadata->subresource = $subresource;
+
+        return $metadata;
+    }
+
+    /**
+     * Is initializable?
+     *
+     * @return bool|null
+     */
+    public function isInitializable()
+    {
+        return $this->initializable;
+    }
+
+    /**
+     * Returns a new instance with the given initializable flag.
+     *
+     * @param bool $initializable
+     *
+     * @return self
+     */
+    public function withInitializable(bool $initializable): self
+    {
+        $metadata = clone $this;
+        $metadata->initializable = $initializable;
 
         return $metadata;
     }
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -142,8 +142,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
             if (
                 $this->isAllowedAttribute($classOrObject, $propertyName, null, $context) &&
-                ((isset($context['api_normalize']) && $propertyMetadata->isReadable()) ||
-                (isset($context['api_denormalize']) && $propertyMetadata->isWritable()))
+                (
+                    isset($context['api_normalize']) && $propertyMetadata->isReadable() ||
+                    isset($context['api_denormalize']) && ($propertyMetadata->isWritable() || !is_object($classOrObject) && $propertyMetadata->isInitializable())
+                )
             ) {
                 $allowedAttributes[] = $propertyName;
             }
--- a/tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
+++ b/tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
@@ -1,0 +1,112 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+/**
+ * Dummy entity built with constructor.
+ * https://github.com/api-platform/core/issues/1747.
+ *
+ * @author Maxime Veber <maxime.veber@nekland.fr>
+ *
+ * @ApiResource(
+ *     itemOperations={
+ *         "get",
+ *         "put"={"denormalization_context"={"groups"={"put"}}}
+ *     }
+ * )
+ * @ORM\Entity
+ */
+class DummyEntityWithConstructor
+{
+    /**
+     * @var int The id
+     *
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column
+     */
+    private $foo;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column
+     */
+    private $bar;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(nullable=true)
+     * @Groups({"put"})
+     */
+    private $baz;
+
+    public function __construct(string $foo, string $bar)
+    {
+        $this->foo = $foo;
+        $this->bar = $bar;
+    }
+
+    /**
+     * @return int
+     */
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    /**
+     * @return string
+     */
+    public function getFoo(): string
+    {
+        return $this->foo;
+    }
+
+    /**
+     * @return string
+     */
+    public function getBar(): string
+    {
+        return $this->bar;
+    }
+
+    /**
+     * @return string
+     */
+    public function getBaz()
+    {
+        return $this->baz;
+    }
+
+    /**
+     * @param string $baz
+     */
+    public function setBaz(string $baz)
+    {
+        $this->baz = $baz;
+    }
+}
--- a/features/serializer/deserialize_objects_using_constructor.feature
+++ b/features/serializer/deserialize_objects_using_constructor.feature
@@ -1,0 +1,31 @@
+Feature: Resource with constructor deserializable
+  In order to build non anemic resource object
+  As a developer
+  I should be able to deserialize data into objects with constructors
+
+  @createSchema
+  Scenario: post a resource built with constructor
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/dummy_entity_with_constructors" with body:
+    """
+    {
+      "foo": "hello",
+      "bar": "world"
+    }
+    """
+    Then the response status code should be 201
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/DummyEntityWithConstructor",
+      "@id": "/dummy_entity_with_constructors/1",
+      "@type": "DummyEntityWithConstructor",
+      "id": 1,
+      "foo": "hello",
+      "bar": "world",
+      "baz": null
+    }
+    """
+
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -507,7 +507,7 @@ final class DocumentationNormalizer implements NormalizerInterface
             'hydra:title' => $propertyName,
             'hydra:required' => $propertyMetadata->isRequired(),
             'hydra:readable' => $propertyMetadata->isReadable(),
-            'hydra:writable' => $propertyMetadata->isWritable(),
+            'hydra:writable' => $propertyMetadata->isWritable() || $propertyMetadata->isInitializable(),
         ];
 
         if (null !== $range = $this->getRange($propertyMetadata)) {
--- a/tests/Fixtures/DummyObjectWithConstructor.php
+++ b/tests/Fixtures/DummyObjectWithConstructor.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures;
+
+/**
+ * @author Maxime Veber <maxime.veber@nekland.fr>
+ */
+class DummyObjectWithConstructor
+{
+    private $foo;
+    private $bar;
+
+    public function __construct(string $foo, \stdClass $bar)
+    {
+        $this->foo = $foo;
+        $this->bar = $bar;
+    }
+}
--- a/tests/Fixtures/DummyObjectWithoutConstructor.php
+++ b/tests/Fixtures/DummyObjectWithoutConstructor.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures;
+
+/**
+ * @author Maxime Veber <maxime.veber@nekland.fr>
+ */
+class DummyObjectWithoutConstructor
+{
+    private $foo;
+
+    public function getFoo()
+    {
+        return $this->foo;
+    }
+
+    public function setFoo($foo)
+    {
+        $this->foo = $foo;
+    }
+}
--- a/tests/Metadata/Property/PropertyMetadataTest.php
+++ b/tests/Metadata/Property/PropertyMetadataTest.php
@@ -78,6 +78,10 @@ class PropertyMetadataTest extends TestCase
         $this->assertNotSame($metadata, $newMetadata);
         $this->assertEquals(['a' => 'b'], $newMetadata->getAttributes());
         $this->assertEquals('b', $newMetadata->getAttribute('a'));
+
+        $newMetadata = $metadata->withInitializable(true);
+        $this->assertNotSame($metadata, $newMetadata);
+        $this->assertTrue($newMetadata->isInitializable());
     }
 
     public function testShouldReturnRequiredFalseWhenRequiredTrueIsSetButMaskedByWritableFalse()


src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $ref = new \ReflectionClass($resourceClass);
+        if (!$ref->isInstantiable()) {
+            return $propertyMetadata;
+        }
+
+        $constructorParameters = $ref->getConstructor()->getParameters();
+        foreach ($constructorParameters as $constructorParameter) {
+            if ($constructorParameter->getName() === $property) {
+                return $propertyMetadata->withWritable(true);
```
soyuka: Okay but this doesn't mean that the PropertyAccess component will be able to write to these properties right (denormalization process)?

dunglas: The PropertyAccessor component directly no, but the Serializer component knows how to use constructors, and constructor support has been dramatically improved in 4.1.

soyuka: Cool then :).

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,69 @@
...
+        } else {
+            try {
+                $propertyMetadata = $this->decorated->create($resourceClass, $property, $options);
+            } catch (PropertyNotFoundException $propertyNotFoundException) {
+                $propertyMetadata = new PropertyMetadata();
+            }
+        }
+
+        // Constructor arguments are obviously accessible only on post operation, put will result in an error.
+        if (!isset($options['collection_operation_name']) || 'post' !== $options['collection_operation_name']) {
```
soyuka: What if we have custom collection operations that don't have the `post` name?

src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
```diff
@@ -69,6 +69,10 @@
             <argument type="service" id="api_platform.metadata.property.metadata_factory.serializer.inner" />
         </service>
 
+        <service id="api_platform.metadata.property.metadata_factory.constructor" class="ApiPlatform\Core\Metadata\Property\Factory\ConstructorPropertyMetadataFactory" decorates="api_platform.metadata.property.metadata_factory" decoration-priority="20" public="false">
```
soyuka: Could we use 30 as a priority as well?

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
Nek-: I don't understand the use case, can you be more specific? It doesn't make sense to allow constructor in another request type than post.

src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
Nek-: Actually 35 makes even more sense (as the serializer `SerializerPropertyMetadataFactory` will potentially modify this accessibility) !

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
meyerbaptiste: He meant this:
```
collectionOperations={
    "foo" => {"method"="POST"}
}
```

Nek-: I guess the best way to fix this is to add the method to the context, what do you think about?

meyerbaptiste: Huum, `ResourceMetadata` should do the job. For example:
```php
$this->resourceMetadataFactory->create($resourceClass)->getCollectionOperationAttribute($options['collection_operation_name'], 'method')
```

Nek-: Fixed.

Nek-:
In what cases is this possible?

```php
$method = $this->resourceMetadataFactory
    ->create($resourceClass)
    ->getCollectionOperationAttribute($options['collection_operation_name'], 'method')
;
// $method === null
```

meyerbaptiste:
There is a better way to do that with the `OperationMethodResolver` class:
```php
$this->operationMethodResolver->getCollectionOperationMethod($resourceClass, $options['collection_operation_name']);
```
Sorry, I missed this class...

Nek-:
Indeed. Fixed!

dunglas:
I though it would be harder! Good job!

Would you mind an integration test to be sure we support this new feature everywhere?

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        if (!isset($options['collection_operation_name'])) {
+            return $propertyMetadata;
+        }
+
+        // Constructor arguments are obviously accessible only on post operation, put will result in an error.
+        try {
+            $method = $this->operationMethodResolver->getCollectionOperationMethod($resourceClass, $options['collection_operation_name']);
+        } catch (RuntimeException $e) {
```
soyuka: when does this throw a RuntimeException?

tests/Metadata/Property/Factory/ConstructorPropertyMetadataFactoryTest.php
```diff
@@ -0,0 +1,160 @@
...
+    private function getMethodResolverMock($method = 'post')
+    {
+        $resolver = $this->prophesize(OperationMethodResolverInterface::class);
+        $resolver->getCollectionOperationMethod(Argument::any(), Argument::any())->willReturn($method);
+
+        return $resolver->reveal();
+    }
+}
+
+class DummyObjectWithConstructor
```
soyuka: Wasn't best practice to put these in another file?

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
Nek-: It's not a problem if the method is not available.

Nek-: I also asked :) : https://github.com/api-platform/core/pull/1749#issuecomment-371136092

meyerbaptiste: Here: https://github.com/api-platform/core/blob/4fa2e4a7665dabdd29487fa907708ae5ef907c7e/src/Bridge/Symfony/Routing/OperationMethodResolver.php#L96

tests/Metadata/Property/Factory/ConstructorPropertyMetadataFactoryTest.php
Nek-: Well, that depends on your POV. IMO re-using dummy classes in many test is not a good practice because it means your test is dependent on something _outside_ (see [FIRST](https://stackoverflow.com/questions/18024785/tdd-first-principle)). I do it only in case I need doctrine entities to avoid configuration duplication.

TL;DR: I don't think it's a big deal.

soyuka: I agree but it's how we choose to do in this project and we have to follow the project rules :).

soyuka: I know for a fact that I did the same as you and I was told to move my classes out of the test file that's all :p.

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,87 @@
...
+     * @var PropertyMetadataFactoryInterface
+     */
+    private $decorated;
+
+    /**
+     * @var OperationMethodResolverInterface
+     */
+    private $operationMethodResolver;
+
+    public function __construct(OperationMethodResolverInterface $operationMethodResolver, PropertyMetadataFactoryInterface $decorated = null)
```
Taluu: Why would the decorated object be `null` ?

soyuka: so that it can be used without decoration

tests/Metadata/Property/Factory/ConstructorPropertyMetadataFactoryTest.php
Nek-: So I guess it's a rule :) .

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
Taluu: IMO it's either you decorate something or you don't, Not both at the same time (S from Solid, anyone ?).

Having a basic decorator that just returns a `new PropertyMetadata()` would be simpler.

dunglas: @Taluu loaders can be chained in any order (and sometimes you have to change the order, to change the priority). The first registered loader will not be a decorator, but you cannot guess which one will be the first when writing the code (it depends of the config).

Taluu: Then it means it is the wrong design pattern for these. Shouldn't a strategy be better for that ? of course, nothing prevents some elements in this pattern to be decorators...

Taluu: Or a chain of responsability would also do it

dunglas: It was tried, and they introduces a lof of unneeded complexity. They have been a lot of iteration on this part, I don't want to change things again. It's mostly internal, and it works well as is.

dunglas: +using a decorator makes it very easy to configure with the Symfony DI component since it supports service decoration and priority. Using a different pattern such as chain (as it was in API Platform 1) will require to add a lot of specific code to have the same flexibility.

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
Nek-: It was about how the method can be null (well, ok, the route is null, but it's not the answer).

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        // Constructor arguments are obviously accessible only on post operation, put will result in an error.
+        try {
+            $method = $this->operationMethodResolver->getCollectionOperationMethod($resourceClass, $options['collection_operation_name']);
+        } catch (RuntimeException $e) {
+            return $propertyMetadata;
+        }
+
+        if ('post' !== strtolower($method)) {
```
meyerbaptiste: With #1752, `'POST' !== $method` is enough. You just have to rebase.

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Set properties available in the constructor as writable.
+ *
+ * @author Maxime Veber <maxime.veber@nekland.fr>
+ */
+final class ConstructorPropertyMetadataFactory implements PropertyMetadataFactoryInterface
+{
+    /**
+     * @var PropertyMetadataFactoryInterface
```
dunglas: Useless (documented by the typehint), to remove.

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    private $decorated;
+
+    /**
+     * @var OperationMethodResolverInterface
```
dunglas: Useless (documented by the typehint), to remove.

tests/Metadata/Property/Factory/ConstructorPropertyMetadataFactoryTest.php
dunglas: It's a rule, because it's what recommended is by the PSRs. Anyway, we try to enforce **not** reusing dummy classes, because of first and - especially - because it's a pain to maintain such tests.

Both can be done together: a test = a dummy class in a separate file. It's what we should enforce.

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
Taluu: Not all editors / etc uses the constructor typehint for properties, so this looks adequate ?

dunglas: We do that everywhere in the project, and in Symfony as well (we follow Symfony's CS). It's less maintenance and PHPStorm supports this so... 😀

Taluu: yeah, that's why it was more a question, but then I opened a ticket on phpactor. :D

tests/Metadata/Property/Factory/ConstructorPropertyMetadataFactoryTest.php
Taluu: Couldn't an anonymous class be used ?

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,87 @@
...
+            return $propertyMetadata;
+        }
+
+        $ref = new \ReflectionClass($resourceClass);
+        if (!$ref->isInstantiable() || !$constructor = $ref->getConstructor()) {
+            return $propertyMetadata;
+        }
+
+        foreach ($constructor->getParameters() as $constructorParameter) {
+            if ($constructorParameter->getName() === $property) {
```
teohhanhui: I think we could just do `$constructorParameter->name`? It seems to be the same.

tests/Metadata/Property/Factory/ConstructorPropertyMetadataFactoryTest.php
teohhanhui: Putting them outside doesn't mean reusing them. We should not have more than 1 class in a file, as that wouldn't work for PSR-4 autoloading.

Nek-:
Fixes done. Rebased. Behat tests added. 😄

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+            return $propertyMetadata;
+        }
+
+        $ref = new \ReflectionClass($resourceClass);
+        if (!$ref->isInstantiable() || !$constructor = $ref->getConstructor()) {
+            return $propertyMetadata;
+        }
+
+        foreach ($constructor->getParameters() as $constructorParameter) {
+            if ($constructorParameter->name === $property) {
```
dunglas: We must also check that the current value of `writable` is null (to avoid overwriting a manual setting tru `@ApiProperty`).

Nek-: Hum. I think modifying the order would be better because the property accessor will set this value to "false".

Actually, this issue is also present in the `ExtractorPropertyMetadataFactory`: if the property accessor decided to set the property to "not accessible" (aka `false`), no manual configuration can reverse the value because it fills it only if the current value is "null".

What do you think of that?

teohhanhui: The user's manual configuration must always be able to override anything. And yes, I think this decorator should run after the PropertyInfo decorator.

Nek-: Ok, so this answer means that there's an issue inside the `ExtractorPropertyMetadataFactory`.

It also means as I understand that I can indeed override the `writable` value (so I don't modify this line).

Nek-:
I see that the problem I notice does not generate eruption of volunteers to check if there's actually something wrong :)... So here is a schema of what's happen (as far as I understand). The problem is more obvious if you look at it. (you may also see some other issues in it)

apiplatformpropertyfactory

Nek-:
I fixed my first commit and added a new commit for the new things pointed out.

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,80 @@
...
+
+namespace ApiPlatform\Core\Metadata\Property\Factory;
+
+use ApiPlatform\Core\Api\OperationMethodResolverInterface;
+use ApiPlatform\Core\Exception\PropertyNotFoundException;
+use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
+
+/**
+ * Set properties available in the constructor as writable.
```
dunglas: Sets

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
```diff
@@ -0,0 +1,80 @@
+<?php
```
dunglas: Do we really need this class? Can't we use https://github.com/symfony/symfony/pull/25605 directly?
https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php#L67 will do the job

Nek-: You're suggesting me to add things non-related to the property metadata factory inside of it. I don't think it's a great idea (it's about single responsibility).

It also enforces this feature to be inside the Symfony bridge (this is not a huge issue for me but all this work of decoupling for finally add things that may be decoupled easily from Symfony sounds sad to me).

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
Nek-: fixed

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
dunglas: I’m not sure to understand your last comment. To clarify: if we drop this class, and bump PropertyInfo to 4.1 it will work without further modification right?

bendavies:
I'd like to use this. Do we need help finishing it off?

src/Metadata/Property/Factory/ConstructorPropertyMetadataFactory.php
soyuka: Yes it should, @Nek- we just need to enforce `property-info` to `4.1` and it should take care of handling the constructor things for us.

dunglas: We only have to enforce it in our dev deps, with an older version, constructor support will not work but that's all.

Nek-: https://github.com/symfony/symfony/pull/25605 is about getting the type of a property via the constructor. I don't understand why you are speaking about it like it could replace the code of this class. Can you tell me more ?

soyuka:
I think that this is working already with `property-info@4.1`.

Nek-:
Just tested to be really sure.

```php
class Foo
{
    private $bar;

    public function __construct($bar)
    {
        $this->bar = $bar;
    }
}

$reflectionExtractor = new \Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor();

$propertyInfo = new \Symfony\Component\PropertyInfo\PropertyInfoExtractor([$reflectionExtractor], [$reflectionExtractor],[], [$reflectionExtractor]);

$properties = $propertyInfo->getProperties(Foo::class);
$writable = $propertyInfo->isWritable(Foo::class, 'bar');

var_dump($properties);
var_dump($writable);
```

And here is the output:
```
NULL
bool(false)
```

(With `symfony/property-info:@dev`)

bendavies:
Can the feature be made to work with the creation of relations of resources?

Nek-:
This feature is not "limited". It works for what it is designed for.

But please tell me what do you think (example of usage) because I don't understand the feature you speak about.

bendavies:
hey @Nek-, sorry, I didn't mean to offend.
What i meant was that API-Platform supports creating relations like this:
https://github.com/api-platform/core/blob/342b92a3d15cf6c9fa610872813f92d45fc36179/features/main/relation.feature#L361

Or anything in that feature really.

dunglas:
Maybe could we copy https://github.com/symfony/symfony/pull/26997 in API Platform for now and mark those classes and interfaces as internal? We'll drop them when it will be merged in Symfony. WDYT?

Nek-:
I good with that, I'll be working on a refactoring ASAP.

seb2411:
Hello, 

Can I extract the fix related to the Metadata Factory decoration priority and create a separate PR for this ?

Thx

bendavies:
what is the bug the priority change is fixing? please see https://github.com/api-platform/core/issues/1980

Nek-:
@dunglas this PR should be ready. What do you think?

src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
```diff
@@ -68,6 +68,23 @@ public function create(string $resourceClass, string $name, array $options = [])
             $propertyMetadata = $propertyMetadata->withWritable($writable);
         }
 
+        if (method_exists($this->propertyInfo, 'isInitializable')) {
+            if (true !== $propertyMetadata->isInitializable() && null !== $initializable = $this->propertyInfo->isInitializable($resourceClass, $name, $options)) {
```
meyerbaptiste: IMO it should be `null !== $propertyMetadata->isInitializable()`.

Nek-:
I think issues are non-related.

Nek-:
I removed the part that was suppose to fix order in metadata factories: it was failing. This will still be an issue but this PR should now pass tests.

soyuka:
> I removed the part that was suppose to fix order in metadata factories: it was failing. This will still be an issue but this PR should now pass tests.

I guess we can track this in https://github.com/api-platform/core/issues/1980 

You have a small indentation issue in `DummyEntityWithConstructor`, just run php-cs-fixer and we're good to go on this :).

Nek-:
Fixed 👍

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -142,8 +142,13 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
 
             if (
                 $this->isAllowedAttribute($classOrObject, $propertyName, null, $context) &&
-                ((isset($context['api_normalize']) && $propertyMetadata->isReadable()) ||
-                (isset($context['api_denormalize']) && $propertyMetadata->isWritable()))
+                (
+                    (isset($context['api_normalize']) && $propertyMetadata->isReadable()) ||
```
teohhanhui: Unnecessary brackets...

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -142,8 +142,13 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
+                    (
```
teohhanhui: Unnecessary brackets...

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -142,8 +142,13 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
 
             if (
```
teohhanhui:
```php
if (
    $this->isAllowedAttribute($classOrObject, $propertyName, null, $context) &&
    (
        isset($context['api_normalize']) && $propertyMetadata->isReadable() ||
        isset($context['api_denormalize']) && ($propertyMetadata->isWritable() || !is_object($classOrObject) && $propertyMetadata->isInitializable())
    )
) {
    $allowedAttributes[] = $propertyName;
}
```

tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
```diff
@@ -0,0 +1,112 @@
...
+     * @var int The id
+     *
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+
+    /**
+     * @var string The tour name
```
teohhanhui: Remove PHPDoc descriptions that don't make sense... :smile:

tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @ORM\Column
+     */
+    private $foo;
+
+    /**
+     * @var string The tour name
```
teohhanhui: Remove PHPDoc descriptions that don't make sense... :smile:

tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @ORM\Column
+     */
+    private $bar;
+
+    /**
+     * @var string The tour name
```
teohhanhui: Remove PHPDoc descriptions that don't make sense... :smile:

src/Serializer/AbstractItemNormalizer.php
Nek-: Looks more clear with too me.

src/Serializer/AbstractItemNormalizer.php
Nek-: Long lines should be avoided, isn't it?

komik966:
I'm closing my PR, because it's almost done here. Nice work. If you decide to add tests for relations feel free to copy it from my PR.

src/Serializer/AbstractItemNormalizer.php
teohhanhui: Split across more lines != easier to read :smile:

teohhanhui: I think many nested levels of brackets, especially with uneven indentation levels, is much worse than long lines.

Nek-:
@komik966 as it is your work it's probably better that you do the PR for the test on relations. (which is nice)

bendavies:
this can't be merged without tests for relations. let's get them in somehow! @komik966 can PR to @Nek-'s branch?

src/Serializer/AbstractItemNormalizer.php
Nek-: ✅

src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
```diff
@@ -68,6 +68,23 @@ public function create(string $resourceClass, string $name, array $options = [])
             $propertyMetadata = $propertyMetadata->withWritable($writable);
         }
 
+        if (method_exists($this->propertyInfo, 'isInitializable')) {
+            if (null === $propertyMetadata->isInitializable() && null !== $initializable = $this->propertyInfo->isInitializable($resourceClass, $name, $options)) {
+                $propertyMetadata = $propertyMetadata->withInitializable($initializable);
+            }
+        } else {
+            // BC layer for Symfony < 4.1
```
dunglas: Symfony 4.2 actually.

src/Metadata/Property/PropertyMetadata.php
```diff
@@ -34,8 +34,9 @@
     private $childInherited;
     private $attributes;
     private $subresource;
+    private $initializable;
 
-    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null)
+    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null, $initializable = null)
```
dunglas: `bool $initializable = null`

src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
Nek-: Arf, yes.

soyuka:
@Nek- @komik966 if you want me to take care of adding the tests written in #1843 in this patch let me know

Nek-:
@soyuka I prefer to not add those tests in this MR if you don't mind.

src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
```diff
@@ -68,6 +68,23 @@ public function create(string $resourceClass, string $name, array $options = [])
             $propertyMetadata = $propertyMetadata->withWritable($writable);
         }
 
+        if (method_exists($this->propertyInfo, 'isInitializable')) {
+            if (null === $propertyMetadata->isInitializable() && null !== $initializable = $this->propertyInfo->isInitializable($resourceClass, $name, $options)) {
+                $propertyMetadata = $propertyMetadata->withInitializable($initializable);
+            }
+        } else {
+            // BC layer for Symfony < 4.2
+            // To be removed in EOF of Symfony 3.4
```
teohhanhui: I don't understand this comment. If it's a BC layer for Symfony < 4.2, shouldn't it only be removed when we drop support for Symfony < 4.2?

dunglas: Yes we can just drop the last line

komik966:
added tests in #1991 
without modifications in `ApiPlatform\Core\Serializer\AbstractItemNormalizer` relations will not work

teohhanhui:
The tests should be in this PR.

komik966:
to add tests here @Nek- can rebase from my branch

src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyMetadataFactory.php
Nek-: Yes my bad!

Nek-:
Okok guys, I was more in favor of merging PRs separately. But well, if you disagree then please review tests in https://github.com/api-platform/core/pull/1991 first because they are not ok to me (and I assume also not for you).

dunglas:
Thank you very much for this @Nek-! It's very appreciated.

bendavies:
great, can't wait to try this.

@komik966 can you work on the tests now?

komik966:
In June I've limited time, so work may be little slow. But I think we are very close to fully working relations. Partially it was implemented in my PR (without handling circular references).

teohhanhui:
This is great work, but I think it's unfortunate that we've merged this without tests.

Nek-:
@teohhanhui it is actually tested.

bendavies:
Can we just clarify what needs to be done here so we don't leave work to be done, as i'm a little confused.

From reading above @Nek-, you didn't want @komik966's tests in your branch, but now you say it is actually tested?

Thanks!

Nek-:
@bendavies this is a test: https://github.com/api-platform/core/pull/1749/files#diff-b63a56b1101fe1091a9cf488e2b0b383R1

The branch of komik966 is nice because it adds many other tests, but it was not mergeable as it to me (and not only to me as I understand).

What's the problem? I don't understand.

soyuka:
@komik966's tests are covering relations from what I can see and there might be an issue there. Also, because of the red coverage, I think that this could be more tested then it is already :).

PR #1751, opened 2018-03-07
Add query extension when using custom data provider

ad3n:
When using custom data provider the query extensions is not working properly. By applying this PR, we can inject the query extension based on service tag.

But we have problem with autowire service because the compiler execute before autowire, can you help us to fix this problem?

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | [need doc change]

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DoctrineQueryExtensionPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DoctrineQueryExtensionPass.php
@@ -57,7 +57,12 @@ final class DoctrineQueryExtensionPass implements CompilerPassInterface
         }
 
         foreach ($itemDataProviders as $itemDataProvider) {
-            $container->getDefinition((string) $itemDataProvider)->replaceArgument(3, $itemExtensions);
+            $definition = $container->getDefinition((string) $itemDataProvider);
+            try {
+                $definition->replaceArgument(3, $itemExtensions);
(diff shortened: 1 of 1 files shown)


soyuka:
Custom data providers are different from the core ones.

If you want extensions just inject the one you need. If your demand is to have an injectable collection of all the available extensions we can add it.

ad3n:
@soyuka : Thanks for your comment. But how to inject the extension to custom data provider because we don't have bundle in app?

meyerbaptiste:
I'm not sure to understand your question. If you want to inject specific extensions, use the service definition, if you want to inject all the extensions use [a compiler pass](https://symfony.com/doc/current/service_container/compiler_passes.html). In any case, it works fine without bundle.

ad3n:
@soyuka : :+1: i agree to use the traversable but you need to modify the doc to mention how to apply filter and/or extension in custom data provider because it is confusing when we create a custom data provider and the filter is gone and we don't know about it until we check the compiler pass...

for new comer, it is hard to debug...

soyuka:
Indeed, it'll be easier to do and to document!

dunglas:
I don't understand the question nor the use case. Why do we need a new class? Isn't https://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services enough?

soyuka:
oh yes this is nice!

PR #1752, opened 2018-03-07
Remove Request::METHOD_* constants & uppercase HTTP methods

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Api/OperationMethodResolverInterface.php
+++ b/src/Api/OperationMethodResolverInterface.php
@@ -16,13 +16,15 @@ namespace ApiPlatform\Core\Api;
 use ApiPlatform\Core\Exception\RuntimeException;
 
 /**
- * Resolves the HTTP method associated with an operation.
+ * Resolves the uppercased HTTP method associated with an operation.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
 interface OperationMethodResolverInterface
 {
     /**
+     * Resolves the uppercased HTTP method associated with a collection operation.
+     *
      * @param string $resourceClass
      * @param string $operationName
      *
@@ -33,6 +35,8 @@ interface OperationMethodResolverInterface
     public function getCollectionOperationMethod(string $resourceClass, string $operationName): string;
 
     /**
+     * Resolves the uppercased HTTP method associated with an item operation.
+     *
      * @param string $resourceClass
      * @param string $operationName
      *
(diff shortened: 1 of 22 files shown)


meyerbaptiste:
Test failure is not related, see https://github.com/symfony/symfony/pull/26387.

dunglas:
I'm 👎 for this change. It introduces more dependencies to HttpFoundation while we try to make this dependency soft (the Doctrine filters subsystem doesn't depend of HttpFoundation anymore).

The long term goal is to also support PSR-7.

Extra arguments:

* Those constants are part of HTTP and very standard
* Even Symfony don't use them in its own code base

soyuka:
We could use our own constants? This may help us improve the codebase and get rid of things like this `strtolower`: https://github.com/api-platform/core/pull/1749/files#diff-205421573c87994453de37df3677ad2dR70

meyerbaptiste:
We need consistency on this subject. In some places we use these constants and in some other places we don't use them...

The line highlighted by @soyuka is the reason that motivated this PR.

soyuka:
> The line highlighted by @soyuka is the reason that motivated this PR.

:smile: I knew it and had the exact same thought.

Then we need to declare our own constants and use them everywhere wdyt?

dunglas:
This test is about DX, even if we were using constants, we would keep it.

The HTTP spec is enough to me. Verbs in uppercase.

meyerbaptiste:
IMO, this test is a nonsense. The method should be normalized to uppercase in our `OperationMethodResolver`.

dunglas:
Good idea.

meyerbaptiste:
RFR. Can this be considered a bug in 2.1 or do I have to change the base branch?

teohhanhui:
In the unit tests, I see that the mocks return the uppercased HTTP method anyway. So it's a bug? We should add a regression test.

src/Api/OperationMethodResolverInterface.php
```diff
@@ -16,13 +16,15 @@
...
 /**
- * Resolves the HTTP method associated with an operation.
+ * Resolves the uppercased HTTP method associated with an operation.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
 interface OperationMethodResolverInterface
 {
     /**
+     * Resolves the uppercased HTTP methods associated with a collection operation.
```
teohhanhui: methods -> method

Also, I feel we don't need to say "uppercased" explicitly. Because that should be the expectation for HTTP verbs?

meyerbaptiste: I do that because in Symfony the PHPDoc specifies `uppercased` for each method that returns an HTTP verb.

meyerbaptiste:
Oh, I missed this PR... again!

meyerbaptiste:
Rebased.

PR #1760, opened 2018-03-09
fix bc break regaring the FilterEagerLoadingExtension

ymarillet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

`$queryBuilder->getRootAliases()[0]` is not working in the case of the FilterEagerLoadingExtension because it uses subquery and the first root alias is not necessarily the one we want

### DIFF
--- a/tests/Bridge/Doctrine/Orm/Util/QueryBuilderHelperTest.php
+++ b/tests/Bridge/Doctrine/Orm/Util/QueryBuilderHelperTest.php
@@ -1,0 +1,67 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Bridge\Doctrine\Orm\Util;
+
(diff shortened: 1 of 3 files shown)


tests/Bridge/Doctrine/Orm/Util/QueryBuilderHelperTest.php
```diff
@@ -0,0 +1,74 @@
...
+    public function testAddJoinOnce(string $originAliasForJoinOnce = null, string $expectedAlias)
+    {
+        $queryBuilder = new QueryBuilder($this->prophesize(EntityManagerInterface::class)->reveal());
+        $queryBuilder->from('foo', 'f');
+        $queryBuilder->from('foo', 'f2');
+        $queryBuilder->join('f.bar', 'b');
+        $queryBuilder->join('f2.bar', 'b2');
+
+        $queryNameGenerator = $this->prophesize(QueryNameGeneratorInterface::class);
+        //$queryNameGenerator->generateJoinAlias($association)->willReturn('a3');
```
soyuka: deadline?

ymarillet: ameneded

Issue #1750, opened 2018-03-06
[GraphQL Mutation] will treat `@ORM\Embedded` as `String` type


leesiongchan:

leesiongchan:
Or `graphql` doesn't support `nested mutation` yet?

leesiongchan:
ping @alanpoulain

alanpoulain:
Yes, API Platform doesn't support nested mutations (yet?). You need to give an IRI for an embedded entity. I will document it later. See also this PR: https://github.com/api-platform/core/pull/1702

leesiongchan:
> But for an embedded entity I'm not sure about which behaviour we want.

Do you mean whether to treat it as flat or object? In the meantime, what is the temporary solution to deal with embedded type?

alanpoulain:
Yes. I think it would be better to treat it like a json_array (iterable type, see this PR https://github.com/api-platform/core/pull/1661). In the meantime have you tried to give it as a JSON encoded string?

leesiongchan:
> Yes. I think it would be better to treat it like a json_array (iterable type, see this PR #1661).

Yes, indeed.

>  In the meantime have you tried to give it as a JSON encoded string?

Tested, it still throwing errors :(

alanpoulain:
What are the errors?

leesiongchan:
![error](https://image.ibb.co/c3BcTS/Screenshot_from_2018_03_07_18_09_03.png)

alanpoulain:
I think you are done for, sorry :( You have to wait for a fix.

eradionov:
Seems like not only mutations do not work
query operations do not work as well

eradionov:
Embedded resource does not work for queries as well for graphql
Is it supported in api-platform?

PR #1765, opened 2018-03-12
[GraphQL] Embedded entities support for mutations

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1750
| License       | MIT
| Doc PR        |

TODO:
- [x] Unit tests

Add a new InputUnionType to allow using either an IRI or data for relations.
It allows to modify embedded entities or to update a related existing resource (#1363).

### DIFF
--- a/src/GraphQl/Serializer/ItemNormalizer.php
+++ b/src/GraphQl/Serializer/ItemNormalizer.php
@@ -23,13 +23,14 @@ namespace ApiPlatform\Core\GraphQl\Serializer;
 
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Serializer\AbstractItemNormalizer;
+use ApiPlatform\Core\Serializer\ItemNormalizer as GenericItemNormalizer;
 
 /**
  * GraphQL normalizer.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class ItemNormalizer extends AbstractItemNormalizer
+final class ItemNormalizer extends GenericItemNormalizer
 {
     const FORMAT = 'graphql';
     const ITEM_KEY = '#item';
@@ -39,7 +40,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        $data = parent::normalize($object, $format, $context);
+        $data = AbstractItemNormalizer::normalize($object, $format, $context);
         $data[self::ITEM_KEY] = serialize($object); // calling serialize prevent weird normalization process done by Webonyx's GraphQL PHP
 
         return $data;
--- a/src/GraphQl/Type/Definition/InputUnionType.php
+++ b/src/GraphQl/Type/Definition/InputUnionType.php
@@ -1,0 +1,187 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\GraphQl\Type\Definition;
+
+use GraphQL\Error\Error;
+use GraphQL\Error\InvariantViolation;
+use GraphQL\Type\Definition\InputObjectType;
+use GraphQL\Type\Definition\InputType;
+use GraphQL\Type\Definition\LeafType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Utils\Utils;
+
+/**
+ * Represents an union of other input types.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class InputUnionType extends Type implements InputType, LeafType
+{
+    /**
+     * @var InputObjectType[]
+     */
+    private $types;
+
+    /**
+     * @var array
+     */
+    private $config;
+
+    /**
+     * @throws InvariantViolation
+     */
+    public function __construct(array $config)
+    {
+        if (!isset($config['name'])) {
+            $config['name'] = $this->tryInferName();
+        }
+
+        Utils::assertValidName($config['name']);
+
+        $this->name = $config['name'];
+        $this->description = $config['description'] ?? null;
+        $this->config = $config;
+    }
+
+    /**
+     * @throws InvariantViolation
+     *
+     * @return InputObjectType[]
+     */
+    public function getTypes(): array
+    {
+        if (null !== $this->types) {
+            return $this->types;
+        }
+
+        if (($types = $this->config['types'] ?? null) && \is_callable($types)) {
+            $types = \call_user_func($this->config['types']);
+        }
+
+        if (!\is_array($types)) {
+            throw new InvariantViolation(
+                "{$this->name} types must be an Array or a callable which returns an Array."
+            );
+        }
+
+        return $this->types = $types;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assertValid()
+    {
+        parent::assertValid();
+
+        $types = $this->getTypes();
+        Utils::invariant(\count($types) > 0, "{$this->name} types must not be empty");
+
+        $includedTypeNames = [];
+        foreach ($types as $inputType) {
+            Utils::invariant(
+                $inputType instanceof InputType,
+                "{$this->name} may only contain input types, it cannot contain: %s.",
+                Utils::printSafe($inputType)
+            );
+            Utils::invariant(
+                !isset($includedTypeNames[$inputType->name]),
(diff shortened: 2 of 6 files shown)


src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
```diff
@@ -77,6 +77,10 @@
             <argument type="service" id="api_platform.item_data_provider" on-invalid="ignore" />
             <argument>%api_platform.allow_plain_identifiers%</argument>
 
+            <call method="setItemNormalizer">
```
dunglas: We don't use setter injection in API Platform. Can you inject it directly in the constructor please?

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -62,11 +65,24 @@ public function supportsNormalization($data, $format = null)
         return self::FORMAT === $format && parent::supportsNormalization($data, $format);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        return $this->itemNormalizer->denormalize($data, $class, $format, $context);
```
dunglas: Can't you use `parent::` directly? It feels weird to me to inject an item normalizer in another one.

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,204 @@
...
+use GraphQL\Utils\Utils;
+
+/**
+ * Represents an union of other input types.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+class InputUnionType extends Type implements InputType, LeafType
```
dunglas: `final`

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,204 @@
...
+
+
+    /**
+     * @return InputObjectType[]
+     *
+     * @throws InvariantViolation
+     */
+    public function getTypes(): array
+    {
+        if (null === $this->types) {
```
dunglas: I would use a guard clause (`if (null !== $this->types) return...`).

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,204 @@
...
+            if (!isset($this->config['types'])) {
```
dunglas:
```php
if ($types = $this->config['types'] ?? null) {
    if (\is_callable($types)) {
        $types = \call_user_func($this->config['types']);
    }
}

File: src/GraphQl/Type/Definition/InputUnionType.php [hunk 6]
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * @throws InvariantViolation
+     */
+    public function assertValid()
+    {
+        parent::assertValid();
+
+        $types = $this->getTypes();
+        Utils::invariant(
+            !empty($types),
```
dunglas: Can't empty can be removed? You can also inline this.

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function parseLiteral($valueNode)
+    {
+        foreach ($this->getTypes() as $type) {
+            if ($type instanceof LeafType) {
+                if (null !== $parsed = $type->parseLiteral($valueNode)) {
```
dunglas: can be merged with the previous line:

```php
if ($type instanceof LeafType && null !== $parsed = $type->parseLiteral($valueNode))
```

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
alanpoulain: I've done it this way because the constructor of `AbstractItemNormalizer` has a lot of dependencies and I would need to copy it in this normalizer. But I can do it if you prefer.

src/GraphQl/Serializer/ItemNormalizer.php
alanpoulain: I can't because the default item normalizer is final. And I need it because I don't want to duplicate the code needed to update related existing resources.

src/GraphQl/Type/Definition/InputUnionType.php
alanpoulain: I've copied it from here: https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/UnionType.php#L137
I think we can make a `count` instead. Do you agree?

src/GraphQl/Serializer/ItemNormalizer.php
dunglas: Remove the `final` keyword and add a `@final` PHPDoc annotation then. It will be cleaner IMHO.

dunglas:
It should target master, it's a new feature.

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,190 @@
...
+{
+    /**
+     * @var InputObjectType[]
+     */
+    private $types;
+
+    /**
+     * @var array
+     */
+    public $config;
```
meyerbaptiste: `private`?

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
alanpoulain: WDYT @dunglas?

dunglas: Yes I prefer to copy them. Please :)

alanpoulain:
@dunglas @meyerbaptiste Fixed.

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -39,7 +40,7 @@
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        $data = parent::normalize($object, $format, $context);
+        $data = AbstractItemNormalizer::normalize($object, $format, $context);
```
meyerbaptiste: Why this change?

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,183 @@
...
+        foreach ($this->getTypes() as $type) {
+            if ($type instanceof LeafType) {
+                try {
+                    return $type->serialize($value);
+                } catch (\Exception $e) {
+                }
+            }
+        }
+
+        throw new InvariantViolation(sprintf('Types in union cannot represent value: %s', Utils::printSafe($value)));
```
meyerbaptiste: `@throws` is missing.

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,183 @@
...
+        foreach ($this->getTypes() as $type) {
+            if ($type instanceof LeafType) {
+                try {
+                    return $type->parseValue($value);
+                } catch (\Exception $e) {
+                }
+            }
+        }
+
+        throw new Error(sprintf('Types in union cannot represent value: %s', Utils::printSafeJson($value)));
```
meyerbaptiste: `@throws` is missing.

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,183 @@
...
+            throw new InvariantViolation(
+                "{$this->name} types must be an Array or a callable which returns an Array."
+            );
+        }
+
+        return $this->types = $types;
+    }
+
+    /**
+     * @throws InvariantViolation
```
meyerbaptiste: Should be replaced by `{@inheritdoc}`.

alanpoulain:
@meyerbaptiste Done.

dunglas:
Thank you very much @alanpoulain

src/GraphQl/Type/Definition/InputUnionType.php
```diff
@@ -0,0 +1,187 @@
...
+{
+    /**
+     * @var InputObjectType[]
+     */
+    private $types;
+
+    /**
+     * @var array
+     */
+    private $config;
```
nmoral: in GraphQL\Type\Definition\Type class $config is declared as public and private in InputUnionType. 

This throw an fatal error with php7.1 and below.

Siregacvek:
@alanpoulain let me ask you why this fix was not merged in production branch?

alanpoulain:
Explanation about why this has been removed is here: https://github.com/api-platform/core/pull/1886

Siregacvek:
@alanpoulain thanks. Looks like it's not solved yet.

Issue #1769, opened 2018-03-13
Cache key collision when multiple subresource paths end in same entity


jmatsumura:
Using:

- API-platform 2.1.4
- Symfony 3.4.1

If you have three entities:

1. `Book`
2. `Store`
3. `Library`

With the following relationships:
- `Library` has a one-to-many relationship to `Book`
- `Store` has a one-to-many relationship to `Book`

If both `Library` and `Store` have `@ApiSubresource` annotations for `Book` then the first path (either `/store/#/books` or `/library/#/books`) hit will be set in Symfony's cache pool (using `Book` to perform the lookup) and that cache key will be assigned to the entire resource path. 

So an initial request to `/store/1/books` will lock the cache to any requests that end in `books` to be expecting something like `/store/#/books` so that if you try call `/library/1/books` it fails due to expecting a valid `Store` IRI in the request.

themizzi:
This can be repaired using the same code from 2.2 at the top of CachedRouteNameResolver::getRouteName()

```php
$context = \func_num_args() > 2 ? func_get_arg(2) : [];
$cacheKey = self::CACHE_KEY_PREFIX.md5(serialize([$resourceClass, $operationType, $context['subresource_resources'] ?? null]));
```

We would like to be able to stay on the 2.1 line as it will require us to upgrade a number of custom filters we have. (I work with @jmatsumura). I will open a pull request for this.

soyuka:
Backported. @dunglas should tag a new 2.1 by the end of the week in the mean time you can use the vcs thing! Thanks for the report!

PR #1770, opened 2018-03-14
add context subresource_resources to cache key hash

themizzi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1769 
| License       | MIT
| Doc PR        | N/A

This makes sure that if the same resource class is used as a sub resource on multiple other entities, the cache keys do not collide. Code comes from 2.2, but works in 2.1 with all tests passing.

### DIFF
--- a/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
@@ -39,7 +39,8 @@ final class CachedRouteNameResolver implements RouteNameResolverInterface
      */
     public function getRouteName(string $resourceClass, $operationType /**, array $context = []**/): string
     {
-        $cacheKey = self::CACHE_KEY_PREFIX.md5(serialize([$resourceClass, $operationType]));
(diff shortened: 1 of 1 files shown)


dunglas:
Thank you very much for reporting and fixing this @themizzi. it's very appreciated.

Don't you guys (ping @api-platform/core-team) think it's a potential security vulnerability? Cache polluting may allow an attacker to access data he shouldn't be able to?

themizzi:
Thanks for fixing this! We are planning to upgrade to 2.2 ASAP, but we have 1,000,000 things to do right now and we have routes that already use this. It would be much appreciated if api-platform (especially core for us specifically right now) could get a release/support schedule document in place with LTS releases as well. Maybe this exists already and I'm just missing it?

themizzi:
Any chance there could be a tagged release for 2.1.7 that includes this?

dunglas:
Yes it's planned because it's border line from a security point of view (can't be exploited AFAIK but...).

PR #1773, opened 2018-03-14
Fix code quality

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Api/FilterLocatorTrait.php
+++ b/src/Api/FilterLocatorTrait.php
@@ -62,5 +62,7 @@ trait FilterLocatorTrait
         if ($this->filterLocator instanceof FilterCollection && $this->filterLocator->offsetExists($filterId)) {
             return $this->filterLocator->offsetGet($filterId);
         }
+
+        return null;
     }
 }
--- a/src/Api/CachedIdentifiersExtractor.php
+++ b/src/Api/CachedIdentifiersExtractor.php
@@ -106,19 +106,18 @@ final class CachedIdentifiersExtractor implements IdentifiersExtractorInterface
 
         try {
             $cacheItem = $this->cacheItemPool->getItem(self::CACHE_KEY_PREFIX.md5($resourceClass));
-            if ($cacheItem->isHit()) {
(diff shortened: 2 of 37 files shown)


src/Api/FilterLocatorTrait.php
```diff
@@ -62,5 +62,7 @@ private function getFilter(string $filterId)
         if ($this->filterLocator instanceof FilterCollection && $this->filterLocator->offsetExists($filterId)) {
             return $this->filterLocator->offsetGet($filterId);
         }
+
+        return null;
```
soyuka: I thought that php was returning null when nothing is specified?

sroze: That's correct.

soyuka: IMO we can ditch this 2 `return null`. Apart from that these are really good code improvements.

dunglas: Since PHP 7.1, with the introduction of `void`, this behavior has changed. If you typehint with `: ?object` you **must** explicitly return null. If you typehint with `: void` you **must not** return null.

We don't use nullable nor void return types right now because we still support PHP 7.0, but one day we'll move forward. So I'm to keep these changes.

meyerbaptiste: We have to differentiate `null` (`return null`) from `void` (`return;`). In fact, that's what we actually do.

meyerbaptiste: See https://github.com/symfony/symfony/issues/17201.

soyuka: I wasn't aware of that thanks. Also TIL isset accepts multiple arguments. \o/ :heart: u

dunglas:
Thanks @meyerbaptiste

PR #1775, opened 2018-03-14
hal collection links should have a href property

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no?
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

It seems to me that json hal links for collections where missing the href property:
https://tools.ietf.org/html/draft-kelly-json-hal-08#section-5.1


dunglas:
Good catch.

We should also validate the produced HAL output using http://hyperschema.org/mediatypes/hal. As we do for Swagger and JSONAPI.

bendavies:
@dunglas happy to take a look. can you point me towards what is done currently for swagger and jsonapi?

soyuka:
> @dunglas happy to take a look. can you point me towards what is done currently for swagger and jsonapi?

1. https://github.com/api-platform/core/blob/master/.travis.yml#L43 we install swagger-cli
2. https://github.com/api-platform/core/blob/master/.travis.yml#L56 we launch a swagger-validate command on an extracted JSON file that comes from a symfony command that you can find [here](https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php)

bendavies:
@dunglas i think I've done what you wanted. Additionally, schema validation was completely broken (not validating at all) in the jsonapi features/context, so I've fixed that.

meyerbaptiste:
This makes me think that it might be nice to have a script to update our JSON Schema files:
* https://raw.githubusercontent.com/json-api/json-api/gh-pages/schema
* https://raw.githubusercontent.com/smizell/hyperschema.org/master/src/mediatypes/hal.json

features/hal/collection.feature
```diff
@@ -0,0 +1,831 @@
...
+        "prev": {
+          "href": "/dummies?itemsPerPage=1&page=1"
+        },
+        "next": {
+          "href": "/dummies?itemsPerPage=1&page=3"
+        },
+        "item": [
+          {
+            "href": "/dummies/2"
+    And the JSON should be valid according to the JSON HAL schema
```
meyerbaptiste: Oops?

bendavies: doh

dunglas:
Thank you very much @bendavies!

PR #1779, opened 2018-03-16
Allow to deactivate fetch join collection

joelwurtz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Having this option by default is a good choice since it offers better and more accurate perfomance in many situations. However there is some cases when having the fetch join collection counter productive : like many items retrieved for a simple query without join, as it forces to have 2 queries (first the disctinct, then an IN queries which is worst for performance in heavy table)

This commit allow users to set use fetch join collection to false (it still is true by default) in their resource / operation attributes for those special cases.

### Todo

 * [x] Test (not sure if it's doable)
 * [x] Docs

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -144,7 +144,7 @@ final class PaginationExtension implements ContextAwareQueryResultCollectionExte
      */
     public function getResult(QueryBuilder $queryBuilder, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $doctrineOrmPaginator = new DoctrineOrmPaginator($queryBuilder, $this->useFetchJoinCollection($queryBuilder));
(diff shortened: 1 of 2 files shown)


dunglas:
Looks legit, but let's see what @soyuka think.

soyuka:
Looks totally legit indeed! 

Please add a test in PaginationExtensionTest though. You'll also have to change the other tests because this new call to `resourceMetadataFactory` isn't expected there.

joelwurtz:
Should be good, i add a test, but don't find a way to mock the construct of the doctrine paginator (to test that true is passed to it) So not sure this test is very useful

dunglas:
Thank you @joelwurtz! Don't forget the docs PR :)

joelwurtz:
Already done in https://github.com/api-platform/docs/pull/438 for the doc

PR #1780, opened 2018-03-16
Serializer listener performance - anonymous class not serializable

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1307
| License       | MIT
| Doc PR        | 

In #1307, an anonymous class was introduced to fix a performance problem but it introduced another.

In these `ItemNormalizers`:
https://github.com/api-platform/core/blob/master/src/Hal/Serializer/ItemNormalizer.php#L219
https://github.com/api-platform/core/blob/master/src/JsonApi/Serializer/ItemNormalizer.php#L337

the context is serialized to produce a cache key, but anonymous classes are not serializable: https://3v4l.org/tvMGa.
This effectively disables the intended caching in the HAL and Json API item normalizers.

Blackfire comparison for a HAL endpoint returning 500 items
https://blackfire.io/profiles/compare/8994c0b0-8a09-4ae7-9c38-dd8ad72ecc22/graph

11% improvement.

### DIFF
--- a/src/EventListener/SerializeListener.php
+++ b/src/EventListener/SerializeListener.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\EventListener;
 
 use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Serializer\ResourceList;
 use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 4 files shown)


sroze:
Very good catch 👏

soyuka:
@dunglas is it fine to merge this on 2.2? It can be considered as a bug fix imo not sure though.

meyerbaptiste:
This is clearly a bug because of the E_WARNING.

meyerbaptiste:
Thx @bendavie!

PR #1781, opened 2018-03-17
Fix Cached identifiers extractor support for stringable identifers.

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

#1101 introduced support for string castable identifier, which was later refactored to `IdentifiersExtractorInterface`, but the support was not maintained in the `CachedIdentifiersExtractor`.

My app uses ramsey uuids for identifiers, and I noticed caching not working here. Using object identifiers effectively disabled the caching of the `CachedIdentifiersExtractor`, making it call the decorated `IdentifiersExtractor` every time.

This change results in a 20% performance boost for my app with an endpoint returning 500 resources:
https://blackfire.io/profiles/compare/0ff6fc7d-afe1-4303-80c3-5a584a933239/graph

### DIFF
--- a/src/Api/CachedIdentifiersExtractor.php
+++ b/src/Api/CachedIdentifiersExtractor.php
@@ -76,6 +76,11 @@ final class CachedIdentifiersExtractor implements IdentifiersExtractorInterface
                 continue;
             }
 
+            if (method_exists($identifiers[$propertyName], '__toString')) {
+                $identifiers[$propertyName] = (string) $identifiers[$propertyName];
+                continue;
+            }
+
             $relatedResourceClass = $this->getObjectClass($identifiers[$propertyName]);
             if (!$relatedIdentifiers = $this->localCache[$relatedResourceClass] ?? false) {
                 $relatedCacheKey = self::CACHE_KEY_PREFIX.md5($relatedResourceClass);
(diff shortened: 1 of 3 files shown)


soyuka:
ref https://github.com/api-platform/core/pull/1478 (in case I have to change something there)

bendavies:
I saw that yes. Might be something for you to improve there.
I'm not sure supporting `_toString` is the most robust method. I'd like to investigate how doctrine does it.

soyuka:
>  I'm not sure supporting _toString is the most robust method. I'd like to investigate how doctrine does it.

Maybe that we won't need to support it that way with #1478 being merged and we could do something smarter.

bendavies:
Yep. That's my thought.

teohhanhui:
I'm not sure, but the way the caching is done here just seems very odd to me.

bendavies:
It does look odd yes, but probably behond the scope of this PR. I'm fixing a clear but and difference between the cached and decorated extractor here.

bendavies:
@meyerbaptiste can i have your approval? 😄

meyerbaptiste:
Thanks @bendavies!

bendavies:
thanks!

soyuka:
I think that this patch is wrong but the bug is real. 

In fact we should test that the `relatedResourceClass` is a `ApiResource` (with `ResourceClassResolver`). The bug comes from the fact that the identifier is an object and not an ApiResource.

bendavies:
you are probably right, but the goal of the patch was to align the cached extractor with the delagate. it's clear further improvement could be made

soyuka:
Not criticizing anything @bendavies ;) I should have reviewed that more thoroughly!

PR #1783, opened 2018-03-19
[GraphQL] Fix update mutation when ID can be denormalized

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Fix an issue in update mutations, when an ID can be denormalized.
For instance, in `Dummy` entity, since there is `setId` method, an update cannot be done.
The error is the following:
`The type of the "id" attribute must be "int", "string" given.`
The first commit is the failing test.
Maybe we also need to add the possibility to change the ID (using `_id` argument)?


dunglas:
Thanks @alanpoulain

PR #1785, opened 2018-03-20
Serializer Property Metadata should not overwrite readableLink/writableLink

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | ????
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Note: I have not updated tests to discuss this first.

Background:
Take a look at this gist:
https://gist.github.com/bendavies/f1895a4089035d754b099d7b69a04c39

What I have is an Entity with a self referencing relation `parent`.

What I want to serialize (`desired.json`) is the `parent` in `_links`, but not in `_embedded`. 

I have tried to do this by explicitly setting `readableLink` to false, but this is overwritten by `SerializerPropertyMetadataFactory` because the serialization group currently takes priority.

What should take priority, `ApiResource` annotations or serializations groups?
If it is serialization groups (as it currently is), is it possible to achieve my desired output without writing a custom normalizer?

Discussed on slack with @dunglas, but bringing the discussion here.

### DIFF
--- a/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
@@ -101,9 +101,6 @@ final class SerializerPropertyMetadataFactory implements PropertyMetadataFactory
      */
     private function transformLinkStatus(PropertyMetadata $propertyMetadata, array $normalizationGroups = null, array $denormalizationGroups = null): PropertyMetadata
     {
-        $propertyMetadata = $propertyMetadata->withReadableLink(true);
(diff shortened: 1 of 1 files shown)


bendavies:
Actually the more I think about this, what I actually need is `readable`/`readableLink` to be decided based on serialisation groups, rather that just a simple boolean. But that would be a significant new feature.

bendavies:
on reflection with dunglas, this is probably not a bug.

komik966:
If readable / writable link is based on serialization groups, when may i need to use annotation?

PR #1788, opened 2018-03-22
Fallback attributes in the @ApiResource annotation

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

This PR adds the ability to configure resource attributes directly at the first level of the `@ApiResource` annotation. This improves the DX by adding autocompletion on annotation properties with modern IDEs.

Attribute keys are normally in snake_case but we use lowerCamelCase at the first level of the `@ApiResource` annotation. So I made the choice to define them also in lowerCamelCase here but they are then normalized to snake_case.

About private properties (see the second commit), they are "virtual" because of this issue: https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112. I don't like that, so if you have a better idea, just tell me.

In the same way, what about priorities? For example, actually: `validationGroups={"foo"}, attributes={"validation_groups"={"bar"}}` = `['bar']`.
Are you agree? Or do you want `['foo']` or `['foo', 'bar']` instead?

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -13,6 +13,9 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Annotation;
 
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use Doctrine\Common\Util\Inflector;
+
 /**
  * ApiResource annotation.
  *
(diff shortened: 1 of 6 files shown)


src/Annotation/ApiResource.php
```diff
@@ -62,4 +93,119 @@
      * @var array
      */
     public $attributes = [];
+
+    /**
+     * @var string
+     */
+    private $accessControl;
```
dunglas: Can you add a reference to the PHPStorm plugin issue, so we'll not forget and drop those properties later :)

src/Annotation/ApiResource.php
```diff
@@ -13,13 +13,44 @@
...
+ *     @Attribute("paginationClientPartial", type="bool"),
+ *     @Attribute("paginationEnabled", type="bool"),
+ *     @Attribute("paginationFetchJoinCollection", type="bool"),
+ *     @Attribute("paginationItemsPerPage", type="int"),
+ *     @Attribute("paginationPartial", type="bool"),
+ *     @Attribute("routePrefix", type="string"),
+ *     @Attribute("shortName", type="string"),
+ *     @Attribute("subresourceOperations", type="array"),
+ *     @Attribute("validationGroups", type="mixed")
+ * )
```
soyuka: FTR I'm using vim and all I see is noise :smiling_imp:

meyerbaptiste: Dear @soyuka,

I encourage you to download a real IDE. Here is the link to PHPStorm: https://www.jetbrains.com/phpstorm/download/.

I hope this could help you to find the truth!

Yours sincerely.

meyerbaptiste: More seriously, `@Attribute` annotations are needed when using a constructor in your annotation class. The noise here is the private properties.

src/Annotation/ApiResource.php
meyerbaptiste: Done.

src/Annotation/ApiResource.php
```diff
@@ -62,4 +93,155 @@
...
+
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var string
+     */
+    private $routePrefix;
+
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
```
Simperfit: Why not having it on the class level since it's on every property?

dunglas: Because we may introduce other properties for valid reasons later.

dunglas:
Thanks @meyerbaptiste

bendavies:
I wonder if these could be fully documented in the docs?

meyerbaptiste:
Yes it will be @bendavies, it's on my todo list 😉

teohhanhui:
Didn't see this earlier, but I also think it's bad, as the attributes were arbitrary keys we used, and many of them were specific to the Symfony integration. But now they've been elevated to a more important status.

PR #1797, opened 2018-03-28
Avoid creating potentially crashing service

greg0ire:
This is broken since sf 3.4.4, see https://github.com/symfony/symfony/pull/25823
One of api platforms listener depends on a file that now throws if it is
loaded and the expression language component is missing. The solution is
to avoid creating the service if the component is missing.
This fixes an error people will have even if they do not use the
access_control annotation attribute.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT


soyuka:
weird error in tests

meyerbaptiste:
@soyuka It's because of the base branch change.

Issue #1778, opened 2018-03-16
[Idea] Add format by resource


vincentchalamon:
Currently, default supported formats are `jsonld`, `json` & `html`. If I want to add a new format, I must add it to api_platform configuration as following:
```yaml
api_platform:
    formats:
        # ...
        pdf:
            mime_types: ['...']
```
But this format will be available for all entry-points of my API. I may want to use this format just for some entry-points.

It should be very useful to add a `format` option directly in the ApiResource annotation, which could be overriden by operation. For instance:
```php
/**
 * @ApiResource(attributes={
 *     "formats"={"jsonld", "json", "html", "csv"={"mime_types"={"text/csv"}}}
 * }, itemOperations={
 *     "put", "delete",
 *     "get"={"formats"={"jsonld", "json", "html", "pdf"={"mime_types"={"application/pdf"}}}}
 * })
 */
class Foo
```
In this example, `csv` format is available in all operations except getItem, and this one has `pdf` format. Those extra formats are only available in this ApiResource.

What do you think of this idea @api-platform/core-team?

Toflar:
I would need that too, should be pretty straight forward to add. Are you creating a PR?

vincentchalamon:
I'm looking for a time-machine to get enough time to work on it @Toflar :wink:

meyerbaptiste:
Don't waste your time looking, create the PR instead! 😄

vincentchalamon:
@Simperfit this has been released in 2.3.0 and can be closed

PR #1798, opened 2018-03-29
[RFR] Allow to specify formats per resources/operations

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/issues/1778
| License       | MIT
| Doc PR        | 
 
As suggesting in the issue, this PR will allow to use the following:

```php 
/**
 * @ApiResource(attributes={
 *     "formats"={"jsonld", "json", "html", "csv"={"text/csv"}}
 * }, itemOperations={
 *     "put", "delete",
 *     "get"={"formats"={"jsonld", "json", "html", "pdf"={"application/pdf"}}}
 * })
 */
class Foo
```

todo:
 - [x] Create a FormatsProvider service
 - [x] Add more PHPUnit tests
 - [x] Add integration tests
 - [x] Update the swagerUi available formats
 - [x] Add `formats` fallback attributes in the @ApiResource annotation

ping @vincentchalamon, @Toflar


dunglas:
I'm for keeping only `"pdf"={"application/pdf"}`, it's expliciter

antograssiot:
ok @dunglas I'll update

antograssiot:
PR updated to only support the simplest declaration form.

antograssiot:
A smaller version I have but harder to read IMO, would be 

```php
    /**
     * Finds formats from a Resource class.
     */
    public function fromRequest(Request $request): array
    {
        if ($this->formats) {
            return $this->formats;
        }

        if (!$attributes = RequestAttributesExtractor::extractAttributes($request)) {
            $this->formats = $this->configuredFormats;

            return $this->formats;
        }

        if (!$formats = (array) $resourceMetadata->getOperationAttribute($attributes, 'formats', [], true)) {
            $this->formats = $this->configuredFormats;

            return $this->formats;
        }

        $this->populateFormats($formats);

        return $this->formats;
    }

    /**
     * Filter and populate the acceptable formats
     */
    private function populateFormats(array $annotationFormats)
    {
        $resourceFormats = [];
        foreach ($annotationFormats as $format => $value) {
            if (!is_numeric($format) && !\is_array($value)) {
                throw new InvalidArgumentException(sprintf("Mime type for format '%s' must be declared as an array.", $format));
            } elseif (is_numeric($format)) {
                if (array_key_exists($value, $this->configuredFormats)) {
                    $resourceFormats[$value] = $this->configuredFormats[$value];
                    continue;
                }
                throw new InvalidArgumentException(sprintf("You either need to add the format '%s' to your project configuration or declare a mime type for it in your annotation.", $value));
            }

            $resourceFormats[$format] = $value;
        }

        $this->formats = $resourceFormats;
    }
```
let me know...

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @author Anthony GRASSIOT <antograssiot@free.fr>
+ */
+final class FormatsProvider implements FormatsProviderInterface
+{
+    private $configuredFormats;
+    private $formats;
+    private $resourceMetadataFactory;
+
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, array $formats)
```
dunglas: I suggest to renamed `formats` in `configuredFormats`, it will be mor explicit

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->configuredFormats = $formats;
+        $this->formats = [];
+    }
+
+    /**
+     * Finds formats from a Resource class.
+     */
+    public function fromRequest(Request $request): array
+    {
+        if ($this->formats) {
```
dunglas: It looks buggy: if a different format can be configured per resource, then you must test for `$this->formats[$resourceClass]` or the bad formats will be returned (for instance the ones of the previously requested resource class).

antograssiot: I'm not storing key/values just the format for one operation right now

antograssiot:
Ok I'll think about your comment @dunglas. I'm the impression that you expect this service to register all acceptables format per resource, but as the format could be defined at the operation level, that seems a bit weird to me.
I could still remove the request and only use attributes, that's what I've been starting with but when I saw the number of places where attributes were not already presents/extracted, I decided to it in a single point in the service, ok to change back if you prefer. I'll work on that next week

Toflar:
Sorry for coming late to the party and thanks for pinging me @antograssiot. I'm +1 on the fact that it should go into its own service and become part of the resource metadata rather than the request as @dunglas stated. Even if you allow to specify the formats on operation level, it is still resource meta data that can have different sources, not only annotations.

antograssiot:
@dunglas it has been updated and doesn't rely on the SF `Request` but only on attributes.

antograssiot:
@Toflar it already leverages the `getOperationAttribute()` method from the resource metadata so already doesn't rely on annotation only but would work with Yaml or XML like every other attributes. Did I missed something ?

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,75 @@
...
+final class FormatsProvider implements FormatsProviderInterface
+{
+    private $configuredFormats;
+    private $formats;
+    private $resourceMetadataFactory;
+
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, array $formats)
+    {
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->configuredFormats = $formats;
```
Toflar: Could you also rename `$formats` to `$configuredFormats` to be consistent here?

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,75 @@
...
+        $this->formats = [];
+    }
+
+    /**
+     * Finds formats from a Resource class.
+     */
+    public function getFormatsFromAttributes(array $attributes): array
```
Toflar: Not sure if I like this concept here. I mean it's a matter of how we want it to work. Right now the specified formats on the resource would **override** the generally configured formats. But if I think about it, the generally configured formats are **general**. So what I expect this annotation to do is to configure **additional** formats for this specific resource. So imho it should **merge** the configured formats with the ones defined on the resource. I understand that it can be a use case to say everything should be available in `json` **except** for this one resource so maybe how it behaves should be a configuration option too?

src/EventListener/AddFormatListener.php
```diff
@@ -29,11 +31,13 @@
...
     private $formats;
     private $mimeTypes;
+    private $formatsProvider;
 
-    public function __construct(Negotiator $negotiator, array $formats)
+    public function __construct(Negotiator $negotiator, array $formats, FormatsProviderInterface $formatsProvider = null)
     {
         $this->negotiator = $negotiator;
         $this->formats = $formats;
+        $this->formatsProvider = $formatsProvider;
```
Toflar: You could remove the typehint and do something like this:

```php
public function __construct(Negotiator $negotiator, $formatProvider)
{
    // BC
    if (\is_array($formatProvider)) {
        // deprecated notice but set anyway
        return;
    }
    if (!$formatProvider instanceof FormatsProviderInterface) {
        // throw new \InvalidArgumentException......
    }
}
```

src/EventListener/AddFormatListener.php
```diff
@@ -51,6 +55,10 @@ public function onKernelRequest(GetResponseEvent $event)
             return;
         }
 
+        if (null !== $this->formatsProvider) {
```
Toflar: Would need a comment to be removed once BC is removed for Version 3.

src/EventListener/DeserializeListener.php
```diff
@@ -31,12 +32,14 @@
     private $serializer;
     private $serializerContextBuilder;
     private $formats;
+    private $formatsProvider;
 
-    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, array $formats)
+    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, array $formats, FormatsProviderInterface $formatsProvider = null)
```
Toflar: Same here with the constructor. I just want to omit we add another 5 arguments to the constructor because it makes migration from 2 to 3 harder.

src/Annotation/ApiResource.php
```diff
@@ -129,6 +130,13 @@
      */
     private $forceEager;
 
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
```
Toflar: In my PR (https://github.com/api-platform/core/pull/1685) it was mentioned that the core team does not want to add more root annotation configuration values but instead put it into `attributes`. Just saying that this may need to be modified but I let @dunglas decide on this. I don't remember the exact reason for this anymore because I actually like when things are specific 😄

antograssiot: this just follows https://github.com/api-platform/core/pull/1788, nothing more

src/Api/FormatsProvider.php
antograssiot: That's indeed one of the concerns I have too. it could be painful if you add a general format, you'll then need to update all annotation.

Toflar: Yeah for me the general ones are the ones that should be available for **all** endpoints. So merging seems more appropriate to me.

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        if (!$attributes) {
+            return $this->configuredFormats;
+        }
+
+        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
+
+        if (!$formats = (array) $resourceMetadata->getOperationAttribute($attributes, 'formats', [], true)) {
```
meyerbaptiste: About `(array)`, is there a way to get something different from an array?

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Finds formats from a Resource class.
+     */
+    public function getFormatsFromAttributes(array $attributes): array
+    {
+        if (!$attributes) {
+            return $this->configuredFormats;
+        }
+
+        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
```
meyerbaptiste: What if `resource_class` is not set?

src/Api/FormatsProvider.php
meyerbaptiste: I like the idea to override the default formats, it's pretty useful (at the operation level) I think. IMHO, If we merge the configured formats, we also have to find a way to be able to completely override the default formats.

Toflar: Yeah, thinking about `additionalFormats` and `formats`. The first just adds and the other overrides and you're done.

src/Api/FormatsProvider.php
antograssiot: Yes I guess by using `"formats"="json"` instead of `"formats"={"json"}`

src/Api/FormatsProvider.php
antograssiot: ok let's try to get this merged with `formats` only and I'll add `additionalFormats` in a following PR if @api-platform/core-team agrees. That will be pretty easy anyway

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,75 @@
...
+        foreach ($annotationFormats as $format => $value) {
+            if (!is_numeric($format)) {
+                $resourceFormats[$format] = (array) $value;
+                continue;
+            } elseif (array_key_exists($value, $this->configuredFormats)) {
+                $resourceFormats[$value] = $this->configuredFormats[$value];
+                continue;
+            }
+
+            throw new InvalidArgumentException(sprintf("You either need to add the format '%s' to your project configuration or declare a mime type for it in your annotation.", $value));
```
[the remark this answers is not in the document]
antograssiot: done

src/Api/FormatsProvider.php
vincentchalamon: But as the key is plural, the user could easily understand it expects an array. Otherwise if the value is not an array, maybe throw an exception? WDYT @meyerbaptiste?

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    private function getOperationFormats(array $annotationFormats): array
+    {
+        $resourceFormats = [];
+        foreach ($annotationFormats as $format => $value) {
+            if (!is_numeric($format)) {
+                $resourceFormats[$format] = (array) $value;
+                continue;
+            }
+            if (array_key_exists($value, $this->configuredFormats)) {
```
vincentchalamon: According to the main feature of this PR, `$value` could be an array (`"formats"={"json", "pdf"={"application/pdf"}}`). The method `array_key_exists` does not allow an array for the first property, and PHP will throw a warning:
array_key_exists

src/Api/FormatsProvider.php
antograssiot: I can if you all prefer, it was almost costless

src/Documentation/Action/DocumentationAction.php
```diff
@@ -47,6 +67,12 @@ public function __invoke(Request $request = null): Documentation
                 $context['api_gateway'] = true;
             }
             $request->attributes->set('_api_normalization_context', $request->attributes->get('_api_normalization_context', []) + $context);
+
+            $attributes = RequestAttributesExtractor::extractAttributes($request);
+        }
+        // BC check to be removed in 3.0
+        if (null !== $this->formatsProvider) {
+            $this->formats = $this->formatsProvider->getFormatsFromAttributes($attributes ?? []);
```
vincentchalamon: There is no need to declare `$attributes` (line 71) out of this `if`, it could be:
```php
$this->formats = $this->formatsProvider->getFormatsFromAttributes(RequestAttributesExtractor::extractAttributes($request));
```

src/Api/FormatsProvider.php
vincentchalamon: The `if` statement may return false if the first argument of `array_key_exists` isn't numeric or a string. I recommend you to check `$value` type before calling `array_key_exists`

antograssiot: Indeed @vincentchalamon, I failled to see that, thanks for your time and explanation, I've added the check and a new exception message, I'm not found on the wording though, you might have a better idea...

src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -66,10 +72,30 @@ public function __construct(ResourceNameCollectionFactoryInterface $resourceName
         $this->oauthTokenUrl = $oauthTokenUrl;
         $this->oauthAuthorizationUrl = $oauthAuthorizationUrl;
         $this->oauthScopes = $oauthScopes;
+
+        if (\is_array($formatsProvider)) {
+            if ($formatsProvider) {
+                // Only trigger notification for non-default argument
+                @trigger_error('Using an array as formats provider is deprecated since API Platform 2.3 and will not be possible anymore in API Platform 3', E_USER_DEPRECATED);
```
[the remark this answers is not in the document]
antograssiot: is this about the `3` versus `3.0` ?
All other places are using this sentence @Simperfit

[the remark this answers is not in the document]
antograssiot: the only exception seems to be the AbstractFilter ... but I can update of course if necessary

src/Api/FormatsProvider.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
+
+        if (!$formats = $resourceMetadata->getOperationAttribute($attributes, 'formats', [], true)) {
+            return $this->configuredFormats;
+        }
+
+        if (!\is_array($formats)) {
+            throw new InvalidArgumentException(sprintf("The 'formats' attributes must be an array, %s given for resource class '%s'.", gettype($formats), $attributes['resource_class']));
```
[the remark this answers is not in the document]
antograssiot: done

src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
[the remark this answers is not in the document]
Simperfit: I think to be coherent it's better to update the AbstractFilter ;). let it like this

antograssiot:
I'm not sure why circle-ci would try to test the code on PHP5.6 🤔 ...
@teohhanhui is it because of https://github.com/api-platform/core/pull/1888 and 2.2 not being merged in master yet ?

soyuka:
try to force push

antograssiot:
I've rebased and updated with the last fixer rules. Anything else to be done on this one ?

src/Bridge/Symfony/Bundle/Resources/config/swagger.xml
```diff
@@ -58,7 +58,7 @@
             <argument>%api_platform.title%</argument>
             <argument>%api_platform.description%</argument>
             <argument>%api_platform.version%</argument>
-            <argument>%api_platform.formats%</argument>
+            <argument type="service" id="api_platform.formats_provider"/>
```
dunglas: missing space before `/>`

antograssiot: updated 👍

dunglas:
Thanks!

PR #1800, opened 2018-03-29
Fix identifier cache

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1781 (last comment)
| License       | MIT
| Doc PR        | na

My original fix https://github.com/api-platform/core/pull/1101 is wrong. Indeed when an relation Entity is an object AND has a `__toString()` method it won't be considered as relation in the file, where it should.

This is only a partial fix because https://github.com/api-platform/core/pull/1478 is the proper fix for the string issue on uuid identifiers for example.

### DIFF
--- a/src/Api/CachedIdentifiersExtractor.php
+++ b/src/Api/CachedIdentifiersExtractor.php
@@ -33,14 +33,20 @@ final class CachedIdentifiersExtractor implements IdentifiersExtractorInterface
     private $cacheItemPool;
     private $propertyAccessor;
     private $decorated;
+    private $resourceClassResolver;
     private $localCache = [];
     private $localResourceCache = [];
 
-    public function __construct(CacheItemPoolInterface $cacheItemPool, IdentifiersExtractorInterface $decorated, PropertyAccessorInterface $propertyAccessor = null)
+    public function __construct(CacheItemPoolInterface $cacheItemPool, IdentifiersExtractorInterface $decorated, PropertyAccessorInterface $propertyAccessor = null, ResourceClassResolverInterface $resourceClassResolver = null)
     {
         $this->cacheItemPool = $cacheItemPool;
         $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
         $this->decorated = $decorated;
+        $this->resourceClassResolver = $resourceClassResolver;
+
+        if (null === $this->resourceClassResolver) {
+            @trigger_error(sprintf('Not injecting %s in the CachedIdentifiersExtractor might introduce cache issues with object identifiers.', ResourceClassResolverInterface::class), E_USER_DEPRECATED);
+        }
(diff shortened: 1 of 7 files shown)


bendavies:
ping me when you are done and i'll try this out.

soyuka:
done @bendavies

bendavies:
This reintroduces the problem i originally fixed.
If the tests are not exposing this then my tests were deficient - sorry.

It's pretty easy to see what the problem is if you debug the request for an resource with a UUID identifier.

First Request:
1. delegate is called, keys are cached, `$identifiers` is populated:
https://github.com/soyuka/core/blob/ff5fdb37d8b40e05e8c151e6bb704064ff0d9b93/src/Api/CachedIdentifiersExtractor.php#L71
2. `$identifiers` are returned: https://github.com/soyuka/core/blob/ff5fdb37d8b40e05e8c151e6bb704064ff0d9b93/src/Api/CachedIdentifiersExtractor.php#L74-L76


Second request
1. `$keys` are retrieved from the cache: https://github.com/soyuka/core/blob/ff5fdb37d8b40e05e8c151e6bb704064ff0d9b93/src/Api/CachedIdentifiersExtractor.php#L70
2. `$identifiers` are not set because the delegate was not called: https://github.com/soyuka/core/blob/ff5fdb37d8b40e05e8c151e6bb704064ff0d9b93/src/Api/CachedIdentifiersExtractor.php#L71

This is all fine up to here. Then:

3. Loop through keys: https://github.com/soyuka/core/blob/ff5fdb37d8b40e05e8c151e6bb704064ff0d9b93/src/Api/CachedIdentifiersExtractor.php#L79
4. will fall all the way through to https://github.com/soyuka/core/blob/ff5fdb37d8b40e05e8c151e6bb704064ff0d9b93/src/Api/CachedIdentifiersExtractor.php#L102 every time, calling the delagate, where the keys and identifers are extracted after all properties have been enumerated. 

point 3 and 4 is the costly slow bit which my origin PR tried to fix.

bendavies:
ignore all of that.
this does work. I forgot to copy the services.xml change to my vendor dir. doh

It works!

The cached extractor one now returns `['id' => 'Ramsey\Uuid object']` rather than `['id' => 'uuid string'], but the tests still pass. the tests should be updated to use `assertSame` instead. 

Nice work!

meyerbaptiste:
> This change results in a 20% performance boost for my app

And no more performance issue @bendavies?

soyuka:
@meyerbaptiste I just improved the fix it's still valid.

bendavies:
@meyerbaptiste yes it's fine.

soyuka:
thanks @bendavies

PR #1803, opened 2018-03-30
Merge 2.2

soyuka:


meyerbaptiste:
Everything is broken @soyuka 😨

soyuka:
Yes I must have failed at resolving conflicts ^^

soyuka:
My guess is that the commit  https://github.com/api-platform/core/commit/d7ada1bc5a0b4f548bce061a8a9129f79c000d37 has an issue with the graphql fix https://github.com/api-platform/core/commit/4e489cd9b1b859981c4b69aaf95cfeb7a9611575

alanpoulain:
No I don't think you failed at resolving conflicts.

It's very strange. The issue is coming from these 3 lines of code: https://github.com/api-platform/core/pull/1803/commits/d2ae08de57f6f8877a09af2742038f3935b97cf4#diff-3c0fe204222ffb162653db261a74ca6aR93.

The test failing is `features/graphql/mutation.feature:57`.
It's because `Given there are 1 dummy objects with relatedDummy` is creating a Dummy object with ID 1.
And then the mutation tries to create another Dummy also using ID 1.
I think it shouldn't be possible because it's not an update (WDYT @dunglas?). But here it's working and the previous Dummy is overwritten.
That's why the test is failing because it verifies the ID should be 2.

But the strange part is how is it possible this is working in 2.2? It shouldn't be.
There have been recently modifications in master related to identifiers, maybe it can explain why.
I think we should investigate. @soyuka do you have an idea?

alanpoulain:
Strange thing: `$manager->contains($data)` is true here and it shouldn't be (case in 2.2).

alanpoulain:
Ok I think I've understood why. I will make a PR to master to fix this.

soyuka:
Many thanks @alanpoulain for fixing the graphql thing :p.

PR #1805, opened 2018-03-31
[GraphQL] Fix identifiers

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Related to https://github.com/api-platform/core/pull/1803.

The commit https://github.com/api-platform/core/commit/d7ada1bc5a0b4f548bce061a8a9129f79c000d37 is bringing issues when merging 2.2 into master.
It's because with this commit the generic item normalizer is used instead of the abstract one.
But in 2.2 this commit https://github.com/api-platform/core/commit/4e489cd9b1b859981c4b69aaf95cfeb7a9611575#diff-3c0fe204222ffb162653db261a74ca6aR89 has begun to invert `id` with `id`.
But it's not finished and when merging 2.2 into master, it highlights the issue.

This PR is fixing it by using the normalizer instead of the mutation resolver.

### DIFF
--- a/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
@@ -85,11 +85,6 @@ final class ItemMutationResolverFactory implements ResolverFactoryInterface
             switch ($operationName) {
                 case 'create':
                 case 'update':
-                    unset($args['input']['id']);
-                    if (isset($args['input']['_id'])) {
-                        $args['input']['id'] = $args['input']['_id'];
-                    }
-
                     $context = null === $item ? ['resource_class' => $resourceClass] : ['resource_class' => $resourceClass, 'object_to_populate' => $item];
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
                     $this->validate($item, $info, $resourceMetadata, $operationName);
--- a/src/GraphQl/Serializer/ItemNormalizer.php
+++ b/src/GraphQl/Serializer/ItemNormalizer.php
@@ -23,13 +23,14 @@ namespace ApiPlatform\Core\GraphQl\Serializer;
 
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Serializer\AbstractItemNormalizer;
+use ApiPlatform\Core\Serializer\ItemNormalizer as BaseItemNormalizer;
 
 /**
  * GraphQL normalizer.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class ItemNormalizer extends AbstractItemNormalizer
+final class ItemNormalizer extends BaseItemNormalizer
 {
     const FORMAT = 'graphql';
     const ITEM_KEY = '#item';
@@ -37,9 +38,17 @@ final class ItemNormalizer extends AbstractItemNormalizer
     /**
      * {@inheritdoc}
      */
+    public function supportsNormalization($data, $format = null)
+    {
+        return self::FORMAT === $format && parent::supportsNormalization($data, $format);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function normalize($object, $format = null, array $context = [])
     {
-        $data = parent::normalize($object, $format, $context);
+        $data = AbstractItemNormalizer::normalize($object, $format, $context);
         $data[self::ITEM_KEY] = serialize($object); // calling serialize prevent weird normalization process done by Webonyx's GraphQL PHP
 
         return $data;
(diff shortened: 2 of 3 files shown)


dunglas:
It is required to use « id » in the Relay spec IIRC.

soyuka:
Shouldn't we just revert the fix  https://github.com/api-platform/core/commit/4e489cd9b1b859981c4b69aaf95cfeb7a9611575#diff-3c0fe204222ffb162653db261a74ca6aR89 ?

alanpoulain:
@dunglas I don't find it anywhere. For example: https://facebook.github.io/relay/graphql/mutations.htm
@soyuka If we revert it, we can't create or modify some entities in GraphQL anymore.

dunglas:
@alanpoulain https://facebook.github.io/relay/docs/en/graphql-server-specification.html#object-identification

> The Node interface contains a single field, `id`, which is a `ID!`.

features/graphql/mutation.feature
```diff
@@ -126,7 +126,7 @@ Feature: GraphQL mutation support
     When I send the following GraphQL request:
     """
     mutation {
-      createRelationEmbedder(input: {paris: "paris", krondstadt: "Krondstadt", anotherRelated: {id: 2, symfony: "laravel"}, clientMutationId: "myId"}) {
+      createRelationEmbedder(input: {paris: "paris", krondstadt: "Krondstadt", anotherRelated: {_id: "/related_dummies/2", symfony: "laravel"}, clientMutationId: "myId"}) {
```
dunglas: Looks invalid to me. `_id` refer to the editable "id" property (so, not an IRI, something like an UUID or a random string), `id` on the other hand is the "obfuscated" global ID defined by the relay spec.

Here it can be:

* `_id: 2`
* or `id: "/related_dummies/2"` here

But not this one

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -89,7 +89,17 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             switch ($operationName) {
                 case 'create':
                 case 'update':
-                    unset($args['input']['id']);
+                    // Invert _id with id because GraphQL is using a different convention than API Platform.
```
dunglas: For the record, I think we should use this convention (id = IRI, _id = the field) everywhere when a specific id isn't defined by the implemented spec (everywhere except in JSON-LD currently).

dunglas:
@alanpoulain isn't it possible to move https://github.com/api-platform/core/commit/d2ae08de57f6f8877a09af2742038f3935b97cf4#diff-3c0fe204222ffb162653db261a74ca6aR93 in the GraphQL denormalizer?

alanpoulain:
Yes maybe it can be a solution to solve the issue with `_id` in nested resources. I will try.

features/graphql/mutation.feature
alanpoulain: Yes that's what I said in the PR's description.

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
alanpoulain: Okay for `id` (it's in the generic item normalizer), but where it is in the code for `_id`? I don't think the AbstractItemNormalizer can understand it.

alanpoulain:
@dunglas I don't understand why you had a link to https://facebook.github.io/relay/docs/en/graphql-server-specification.html#object-identification.
The Relay's specification says for the node, an `id` field is needed. But it's only for the Node object (and so only for queries).

alanpoulain:
Successful try! I have even removed the ugly unset in the mutation resolver.

alanpoulain:
I think this should target 2.2 WDYT?

soyuka:
> I think this should target 2.2 WDYT?

Definitely, we should be able to merge 2.2 back to master without changes so that tests are green.

alanpoulain:
Done.

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -23,23 +23,32 @@
 
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Serializer\AbstractItemNormalizer;
+use ApiPlatform\Core\Serializer\ItemNormalizer as GenericItemNormalizer;
```
dunglas: Can you use `BaseItemNormalizer` as alias please (we usually follow this convention).

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -57,16 +66,35 @@ protected function normalizeCollectionOfRelations(PropertyMetadata $propertyMeta
...
      */
-    public function supportsDenormalization($data, $type, $format = null)
+    protected function getAllowedAttributes($classOrObject, array $context, $attributesAsString = false)
     {
-        return self::FORMAT === $format && parent::supportsDenormalization($data, $type, $format);
+        $allowedAttributes = parent::getAllowedAttributes($classOrObject, $context, $attributesAsString);
+
+        if (($context['api_denormalize'] ?? false) && false !== ($indexId = array_search('id', $allowedAttributes, true))) {
+            $allowedAttributes[] = '_id';
+            array_splice($allowedAttributes, $indexId, 1);
```
dunglas: Wouldn't `unset($allowedAttributes[$indexId]` be explicit and faster?

alanpoulain: Yes but I don't really like it, because it doesn't preserve keys and can lead to weird bugs.

dunglas:
Thanks @alanpoulain!

PR #1766, opened 2018-03-12
Update PaginationExtension.php

remoteclient:
This PR should allow to instantiate `DoctrineOrmPaginator` with a query instead of queryBuilder which allows to set hints. Hints are needed to make translations with gedmo/translatable work.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1694 
| License       | MIT
| Doc PR        |

Issue #1802, opened 2018-03-30
Unable to use PaginationExtension's limit/offset with custom Elasticsearch extension


dkarlovi:
I'm trying to [replace the lookup part of the process with one based on Elasticsearch](https://github.com/rollerworks/search/blob/master/lib/ApiPlatform/Elasticsearch/Extension/SearchExtension.php#L89-L97).

It works like this:
1. transform the input params into an ES query
2. ES determines exactly which documents match and returns **exact set and order** of entity IDs
3. set those entity IDs directly on the query builder and fetch exactly those entities

So, for this to work, Elasticsearch needs to know the limit/offset, which is provided by the pagination extension.

Sadly, the extension does too much currently as it applies the limit/offset and the immediately fetches the result from QueryBuilder. This means I cannot reuse the functionality and this seems impossible to implement without copy/pasting the relevant code from pagination to my extension.

1. Pagination, Elasticsearch - Elasticsearch never triggered
2. Elasticsearch, Pagination - no limit/offset information

ro0NL:
@meyerbaptiste i still havent come to play with elasticsearch, but this is solved after #1999  isnt it?

soyuka:
Yes @ro0NL thanks!

PR #1808, opened 2018-04-03
Improve DX for PaginationExtension reusability

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/1766
| License       | MIT
| Doc PR        | todo

See my comment:
https://github.com/api-platform/core/pull/1766#issuecomment-372967730 

With this patch you could transform [the code](https://github.com/api-platform/core/pull/1766#issuecomment-372583372) (that has to copy almost 90% of the PaginationExtension) to something like:

```
<?php
class MyDataProvider implements CollectionDataProviderInterface {
    use PaginationTrait;

    public function getCollection() {
        foreach ($this->collectionExtensions as $extension) {
            $extension->applyToCollection($queryBuilder, $queryNameGenerator, $resourceClass, $operationName, $context);

            // This is the last extension we will call
            if ($extension instanceof PaginationExtension && $extension->supportsResult($resourceClass, $operationName)) {
                $query = $queryBuilder->getQuery();
                // do something on the query
                $this->addTranslatableQueryHints($query);

                // We can use Query to create the paginator
                $doctrineOrmPaginator = new DoctrineOrmPaginator($query, $this->useFetchJoinCollection($queryBuilder));
                $doctrineOrmPaginator->setUseOutputWalkers($this->useOutputWalkers($queryBuilder));

                // Use the PaginationExtension public method
                if ($extension->isPartialPaginationEnabled($request, $resourceMetadata, $operationName)) {
                  return $this->getPartialPaginator($doctrineOrmPaginator);
                }

                return $this->getPaginator($doctrineOrmPaginator);
            }
        }

        $query = $queryBuilder->getQuery();
        $this->addTranslatableQueryHints($query);
        return $query->getResult();
   }
}
```

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -13,16 +13,13 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Extension;
 
-use ApiPlatform\Core\Bridge\Doctrine\Orm\AbstractPaginator;
-use ApiPlatform\Core\Bridge\Doctrine\Orm\Paginator;
-use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryChecker;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\PaginationTrait;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\QueryBuilder;
-use Doctrine\ORM\Tools\Pagination\Paginator as DoctrineOrmPaginator;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -144,20 +142,17 @@ public function supportsResult(string $resourceClass, string $operationName = nu
...
-            };
+            return $this->getPartialPaginator($doctrineOrmPaginator);
         }
 
-        return new Paginator($doctrineOrmPaginator);
+        return $this->getPaginator($doctrineOrmPaginator);
     }
 
-    private function isPartialPaginationEnabled(Request $request = null, ResourceMetadata $resourceMetadata = null, string $operationName = null): bool
+    public function isPartialPaginationEnabled(Request $request = null, ResourceMetadata $resourceMetadata = null, string $operationName = null): bool
```
soyuka: Because these methods have a lot of api-platform related parameters I think it's a good idea to make them public. Anyway if not, one would have to rewrite the logic in its extension/dataprovider.

dunglas:
I'm not sure it is worth it, almost all methods are empty. Doing a copy/paste is the good solution here to me.

soyuka:
Maybe only the QueryChecker things in a trait?

dkarlovi:
This doesn't seem to solve #1802, I'd still need to copy/paste everything since I'm only after the limit/offset info.

soyuka:
copy paste FTW

PR #1819, opened 2018-04-06
bugfix: throw when a subresources id is not found

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT


src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -63,6 +64,7 @@ public function __construct(ManagerRegistry $managerRegistry, PropertyNameCollec
      * {@inheritdoc}
      *
      * @throws RuntimeException
+     * @throws ItemNotFoundException
      */
     public function getSubresource(string $resourceClass, array $identifiers, array $context, string $operationName = null)
```
soyuka: It's not the dataprovider's responsability to throw an `ItemNotFoundException`. To stay constistent with `ItemDataProvider`, please patch this inside the [`ReadListener`](https://github.com/api-platform/core/blob/master/src/EventListener/ReadListener.php#L120) instead.

Simperfit: done

soyuka:
> This could go maybe agasint 2.0 since it has been introduced in 2.0, WDYT ?

No, security patches only on 2.0. 

Also, could you add a phpunit test `expectedException` on the `ReadListenerTest`? Tyvm <3.

Simperfit:
@soyuka Unit test added.

src/EventListener/ReadListener.php
```diff
@@ -147,6 +148,10 @@ private function getSubresourceData(Request $request, array $attributes, array $
 
         $data = $this->subresourceDataProvider->getSubresource($attributes['resource_class'], $identifiers, $attributes['subresource_context'], $attributes['subresource_operation_name']);
 
+        if (null === $data) {
+            throw new ItemNotFoundException(sprintf('Object of type "%s" with identifiers %s could not be found', $attributes['resource_class'], implode(',', $identifiers)), 400);
```
meyerbaptiste: Why don't we use the `NotFoundHttpException`  here?

tests/EventListener/ReadListenerTest.php
```diff
@@ -160,6 +160,32 @@ public function testRetrieveSubresource()
...
+        $request = new Request([], [], ['id' => 1, '_api_resource_class' => 'Foo', '_api_subresource_operation_name' => 'get', '_api_format' => 'json', '_api_mime_type' => 'application/json', '_api_subresource_context' => ['identifiers' => [['id', 'Bar', true]], 'property' => 'bar']]);
+        $request->setMethod('GET');
+
+        $event = $this->prophesize(GetResponseEvent::class);
+        $event->getRequest()->willReturn($request)->shouldBeCalled();
+
+        $listener = new ReadListener($collectionDataProvider->reveal(), $itemDataProvider->reveal(), $subresourceDataProvider->reveal());
+        $listener->onKernelRequest($event->reveal());
+
+        $this->assertSame($data, $request->attributes->get('data'));
```
meyerbaptiste: `$data` is not set...

src/EventListener/ReadListener.php
Simperfit: I think it's better yeah.

src/EventListener/ReadListener.php
```diff
@@ -147,6 +147,10 @@ private function getSubresourceData(Request $request, array $attributes, array $
 
         $data = $this->subresourceDataProvider->getSubresource($attributes['resource_class'], $identifiers, $attributes['subresource_context'], $attributes['subresource_operation_name']);
 
+        if (null === $data) {
+            throw new NotFoundHttpException(sprintf('Object of type "%s" with identifiers %s could not be found', $attributes['resource_class'], implode(',', $identifiers)));
```
meyerbaptiste: Can you change the message to keep consistency with this:
https://github.com/api-platform/core/blob/d4aa84f4eac58e7900b6f4d30947d3926902bb98/src/EventListener/ReadListener.php#L175

soyuka:
thanks @Simperfit

soyuka:
It should be in 2.2.6 and later

bendavies:
it's right there
https://github.com/api-platform/core/blob/v2.2.7/src/EventListener/ReadListener.php#L102

lvillarino:
The variable $data is not null when you try to access a subresource ex: post/id/comments 

I think that we have to update the line https://github.com/Simperfit/core/blob/a65e09e354df3f5f408234e0cf8eb1e4b3bbd18b/src/EventListener/ReadListener.php#L150 with this code
`if (null === $data || 0 == $data->getTotalItems()) {
            throw new NotFoundHttpException('Not Found.');
        }
`

PR #1821, opened 2018-04-06
meta: missing throws doc property in the ReadListener

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Missing throws property on the phpdoc.


soyuka:
php-cs-fixer

Simperfit:
It is thrown at this point and it's not catched, it's possible to have the exception in here.

meyerbaptiste:
So, you have to do the same for the 425 other places:

capture d ecran 2018-04-06 a 15 45 52

Simperfit:
If we agree on this, I will do it, it seems better to me IMHO.

soyuka:
> Do we really want this? I am in favor of only annotating exceptions that are explicitly thrown in our methods.

:+1: I don't see any advantage

Simperfit:
Than ok ;) we can close this

PR #1825, opened 2018-04-08
Enhance Api Platform Profiler

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This PR aims to improve the Datacollector by : 

- Use the `profiler_dump()` function for a cleaner rendering
- Format the keys of the arrays and limit their size in the array
- Highlight the Resource class and move the Acceptable content type at the bottom
- Presents the attributes on 2 columns only
- Highlight the actual operation
- Add the resource class display from the toolbar

@jdeniau I hope this is going in the direction you initially had in mind, and that the change from 3 to 2 columns is ok, IMO mot of the time the really interesting information is/was not the method nor route name only.

I remove the request method from the collector has it was not used in the template and this information is already present in other panels...

before: 
screenshot-2018-4-8 symfony profiler 1

after:
screenshot-2018-4-8 symfony profiler

capture d ecran 2018-04-08 a 08 45 52

If people like it, I'd like then to improve the filter parts by injecting the filter locator into the dataCollector, that would allow:
 - to expose at the Filter Class at least (but maybe also attributes etc)
 - to highlight in read the filters declared that doesn't match a real filter service that are "silently ignored" today IIRC

Another idea that I have would be to display the actual result number near by the toolbar icon.
The idea would be to check if the result is a Paginator and get the total items number out of it or display 1 on an item operation. (It is sometime not quick to get this information as it stands in the bottom of the response in JSON-LD) at least

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
+++ b/src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
@@ -1,33 +1,9 @@
 {% extends '@WebProfiler/Profiler/layout.html.twig' %}
 
-{% macro operationLine(itemOrCollection, key, operation) %}
+{% macro operationLine(key, operation, actualOperationName) %}
     <tr>
-        <td>
-            {{ key }}
-        </td>
-        <td>
-            {% if operation['method'] is defined %}
-                <em>Method:</em> {{ operation['method'] }}
-            {% elseif operation['route_name'] is defined %}
-                <em>Route name:</em> {{ operation['route_name'] }}
-            {% endif %}
-        </td>
-        <td>
-            <a
-                href="#"
-                class="sf-toggle link-inverse sf-toggle-on"
-                data-toggle-selector="#api-platform-line-{{ itemOrCollection }}-{{ key }}"
-                data-toggle-alt-content="Hide"
-                data-toggle-original-content="Show"
-            >
-                Show
-            </a>
-        </td>
-    </tr>
-    <tr id="api-platform-line-{{ itemOrCollection }}-{{ key }}">
-        <td colspan="3">
-            {{- dump(operation) -}}
-        </td>
+        <th scope="row" {{ key == actualOperationName ? 'class="status-success"' : '' }}>{{ key }}</th>
+        <td {{ key == actualOperationName ? 'class="status-success"' : '' }}>{{- profiler_dump(operation, 1) -}}</td>
     </tr>
 {% endmacro %}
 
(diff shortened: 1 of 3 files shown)


src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
```diff
@@ -1,33 +1,9 @@
...
-            >
-                Show
-            </a>
-        </td>
-    </tr>
-    <tr id="api-platform-line-{{ itemOrCollection }}-{{ key }}">
-        <td colspan="3">
-            {{- dump(operation) -}}
-        </td>
+        <th scope="row" {{ key == actualOperationName ? 'class="status-success"' : '' }}>{{ key }}</th>
```
sroze: My HTML purism would say:
```patch
- <th scope="row" {{ key == actualOperationName ? 'class="status-success"' : '' }}>{{ key }}</th>
+ <th scope="row"{{ key == actualOperationName ? ' class="status-success"' : '' }}>{{ key }}</th>
```

dunglas: I would even use:

`<th scope="row"{% if key == actualOperationName %} class="status-success"{% endif %}>{{ key }}</th>`

src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
```diff
@@ -1,33 +1,9 @@
...
+        <td {{ key == actualOperationName ? 'class="status-success"' : '' }}>{{- profiler_dump(operation, 1) -}}</td>
```
dunglas: Same here btw:

```patch
-<td {{ key == actualOperationName ? 'class="status-success"' : '' }}>
+<td{% if key == actualOperationName %} class="status-success"{% endif %}>
```

antograssiot:
glad you liked it.
updated as requested

antograssiot:
rebased, anything else to be done on this one @dunglas ?

Simperfit:
Since @dunglas's comments have been changed accordingly and that we have two approuve I'm merging this one.

Thanks @antograssiot.

sroze:
It introduced a bug 🙃 See #1846.

sroze:
Great move 😉

Issue #1149, opened 2017-05-31
Drop Twig Dependency


ismailbaskin:
I think this library mostly aims API first or API only projects. Twig is an unnecessary dependency with `enable_swagger_ui: false`  but still needed for `api_platform.swagger.action.ui` service.

teohhanhui:
I believe we can drop Twig entirely even when using the Swagger UI by ensuring that it's fully client-side.

ismailbaskin:
In fact I think separating "Swagger UI for Symfony" project can be more reasonable.

teohhanhui:
Yes, I see no reason for this project to include any JS and CSS, etc.

It should live in a separate frontend project, like https://github.com/api-platform/admin and https://github.com/api-platform/generate-crud. It should be included as part of our Docker Compose setup. And with Symfony Flex it should be pretty easy to advise the user to install it separately.

teohhanhui:
@api-platform/core-team One guiding principle for web APIs is the separation of the frontend and the backend. I believe we need to set a good example ourselves.

dunglas:
I definitely agree with that, but it's easiest solution right now. We can work to do it in 2.2!

pierre-H:
What if we use EasyAdminBundle or another bundle which requires Twig ?
Personally, I don't use api-platform-admin but EasyAdminBundle because my frontend project is in typescript with react. So I can't add easily api-platform-admin. And I don't want to have two projects : one for the app in typescript and another one for the admin with js.

soyuka:
@pierre-H this doesn't mean that you can't have twig as a dependency in your project :). This is just a proposal to drop the hard dependency on twig on the core library.

dunglas:
There is no hard dependency on Twig in core. It's just a soft dep.

fgrandjean:
Actually the Symfony bridge's service "api_platform.swagger.action.ui" introduce a hard dependency with Twig (still true is the 2.2-beta).
The "cleanest" fix i could think of, would be to remove that service using a compiler pass when the twig service definition doesnt exists ?

Simperfit:
@fgrandjean Do you have some time to implement this ?

fgrandjean:
@Simperfit I can, but i do need some direction : Do we go for a clearer exception that tell end user that when used with Symfony they are required to add twig as a deps or do we silently toggle on/off the Swagger ui based on deps (as the Sf FrameworkBundle does) ?

Simperfit:
[...]

PR #1829, opened 2018-04-09
Fix. Symfony integration when starting with the micro Flex application.

fgrandjean:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1149 
| License       | MIT
| Doc PR        | 

The goal here is to make it failsafe to install api-platform/core on a symfony/skeleton project.
It remove hardcoded dependencies for doctrine/doctrine-bundle and symfony/twig-bundle.
Corresponding features get toggle back on when the package are added to the project.

### DIFF
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
@@ -490,8 +490,6 @@ class ApiPlatformExtensionTest extends TestCase
             'api_platform.swagger.normalizer.documentation',
             'api_platform.swagger.normalizer.api_gateway',
             'api_platform.swagger.command.swagger_command',
-            'api_platform.swagger.action.ui',
-            'api_platform.swagger.listener.ui',
             'api_platform.hal.encoder',
             'api_platform.hal.normalizer.collection',
             'api_platform.hal.normalizer.entrypoint',


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -218,7 +218,9 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
 
         $container->getDefinition('api_platform.metadata.extractor.xml')->addArgument($xmlResources);
 
-        if (class_exists(Annotation::class)) {
+        // The "annotation_reader" service is only available if DoctrineBundle is loaded.
+        $bundles = $container->getParameter('kernel.bundles');
+        if (class_exists(Annotation::class) && isset($bundles['DoctrineBundle'])) {
```
soyuka: Is this check on bundles really mandatory?

fgrandjean: I know it's a bit weird but in theory it's possible to require "doctrine/annotation" without "doctrine/doctrine-bundle", in that case the class exists but not the service definition.

fgrandjean: At first it does look cleaner but it's not as reliable. Again, in theory, you can install a bundle with composer but not register it in Symfony's kernel. But since it's an edge case, maybe it's acceptable to simply check for class existance?

soyuka: I'd need some expert hands to answer this as I'm not qualified enough :p

Simperfit: I agree with @fgrandjean it should not be tested alone, since it could lead to unexpected behaviours (as the bundle may not be in the Kernel). So I guess it's better to test that the bundle is loaded.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -346,7 +348,8 @@ private function registerSwaggerConfiguration(ContainerBuilder $container, array
 
         $loader->load('swagger.xml');
 
-        if ($config['enable_swagger_ui']) {
+        $bundles = $container->getParameter('kernel.bundles');
```
Simperfit: do you really need bundles ?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -432,6 +435,14 @@ private function registerGraphqlConfiguration(ContainerBuilder $container, array
         $container->setParameter('api_platform.graphql.graphiql.enabled', $config['graphql']['graphiql']['enabled']);
 
         $loader->load('graphql.xml');
+
+        $bundles = $container->getParameter('kernel.bundles');
```
Simperfit: do you really need the variable bundles ?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
dunglas: I'm for keeping the current code too.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -218,7 +218,9 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
 
         $container->getDefinition('api_platform.metadata.extractor.xml')->addArgument($xmlResources);
 
-        if (class_exists(Annotation::class)) {
+        // The "annotation_reader" service is only available if DoctrineBundle is loaded.
+        $bundles = $container->getParameter('kernel.bundles');
```
dunglas: `$bundles` is already defined here: https://github.com/fgrandjean/core/blob/a6bb41d21f37898d80daae52336418e5310d4daf/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L118

You can just pass it as parameter of all `register*` methods.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -346,7 +348,8 @@ private function registerSwaggerConfiguration(ContainerBuilder $container, array
 
         $loader->load('swagger.xml');
 
-        if ($config['enable_swagger_ui']) {
+        $bundles = $container->getParameter('kernel.bundles');
+        if ($config['enable_swagger_ui'] && isset($bundles['TwigBundle'])) {
```
dunglas: Maybe should we just throw if `enable_swagger_ui` is `true` but Twig isn't available. It would be easier to debug (the exception's message should recommend to install `twig-bundle`.

dunglas: The default value of `enable_swagger_ui` can be false if twig isn't available btw (in the `Configuration` class). As you'll not be able to access to the installed bundle list in this class, you can fallback on checking if `class_exists(TwigBundle::class)`.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -438,6 +446,13 @@ private function registerGraphqlConfiguration(ContainerBuilder $container, array
         $container->setParameter('api_platform.graphql.graphiql.enabled', $config['graphql']['graphiql']['enabled']);
 
         $loader->load('graphql.xml');
+
+        if (isset($bundles['TwigBundle'])) {
```
dunglas: It looks wrong to me. It means that you can get GraphQL support without Twig.

As for Swagger UI, you should throw if `api_platform.graphql.graphiql.enabled` is `true` but Twig isn't installed.

src/Bridge/Symfony/Bundle/Resources/config/graphql-ui.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" ?>
```
dunglas: This should be reverted. This action isn't here for GraphiQL in the first place. It is the main GraphQL entrypoint.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -218,7 +219,8 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
 
         $container->getDefinition('api_platform.metadata.extractor.xml')->addArgument($xmlResources);
 
-        if (class_exists(Annotation::class)) {
+        // The "annotation_reader" service is only available if DoctrineBundle is loaded.
+        if (class_exists(Annotation::class) && isset($bundles['DoctrineBundle'])) {
```
teohhanhui: We should use
```php
$container->hasExtension('doctrine')
```

There is no need to pass the `$bundles`.

teohhanhui: I mean, they're not strictly the same thing. But we can safely assume that the `doctrine` extension must be registered if `DoctrineBundle` is loaded.

WDYT @dunglas? I think we can simplify the existing code too...

fgrandjean: To me it's mostly the same thing, except that we check for TwigBundle using `isset($bundles['TwigBundle'])` it made sens to do it the same way for doctrine to keep the code easy to read. But I let @dunglas decide.

Issue #1833, opened 2018-04-09
[RFC] implement allow_plain_identifiers in JSON API


Simperfit:
```
Thanks @dunglas. I see the "allow_plain_identifiers" option is not implemented for the JSON API format. The JSON API spec has no opinion on whether IDs should be IRI's - in fact all their examples only have UIDs, so I think your implementation should support the "allow_plain_identifiers" option.

Would you accept a PR if I implemented it?
```

Simperfit:
from: https://github.com/api-platform/api-platform/issues/569

Ronnie-J:
👍

soyuka:
superseeded by #2686

PR #1834, opened 2018-04-09
feature: allow plain identifiers in json api

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1833
| License       | MIT
| Doc PR        | 

This PRs allows to use "allow_plain_identifiers" in JSON API.

### DIFF
--- a/features/jsonapi/jsonapi.feature
+++ b/features/jsonapi/jsonapi.feature
@@ -120,6 +120,46 @@ Feature: JSON API basic support
     And the JSON node "data.relationships.relatedDummies.data" should have 2 elements
     And the JSON node "data.relationships.relatedDummy.data.id" should be equal to "/related_dummies/2"
 
+  Scenario: Create a dummy with relations and plain identifiers
+    Given there is a RelatedDummy
+    When I send a "POST" request to "/dummies" with body:
+    """
+    {
+      "data": {
+        "type": "dummy",
+        "attributes": {
+          "name": "Dummy with relations",
+          "dummyDate": "2015-03-01T10:00:00+00:00"
+        },
+        "relationships": {
+          "relatedDummy": {
+            "data": {
+              "type": "related-dummy",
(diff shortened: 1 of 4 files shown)


soyuka:
You should've asked first. This is joining my though in https://github.com/api-platform/core/pull/1716#issuecomment-379362006 

Using `ItemDataProvider` in the normalizers (this one or the abstract one) is wrong. We should only use `IriConverter`. This is even worse with #1478 because the `ItemDataProvider` needs to be called with normalized identifiers...
Also, `IriConverter` is the class we want to be responsible for handling data in the normalizers.

@Simperfit this is the fix I want to propose, I don't have much time though so feel free to do it :).

1. Add a method in `IriConverter` that has the following signature:

```
public function getIriFromPlainIdentifier($id, $resourceClass) {
 // returns an IRI
}
```

Or this alternative (not fond of this because `IriConverter` should only do things based on `IRIs`:
```
public function getItemFromPlainIdentifier($id, $resourceClass) {}
```

2. Use that IRI in the normalization processes
3. Rewrite a proper fix for #1716 related issue

Simperfit:
Ow @soyuka I didn't know that was already a topic!

I can implement the first solution as it seem (to me too) to be the best one.

So I'm going to:
- remove ItemDataProvider from the abstract class.
- implement getIriFromPlainIdentifier and use it in the abstract class
- and use this new `getIriFromPlainIdentifier` in the JSONAPI's ItemNormalizer.

dunglas:
You should create a new interface `PlainIdentifierConverterInterface` for that (it's not the responsibility of the `IriConverterInterface`).

flug:
@Simperfit Up is this pull request still valid?

soyuka:
Let's close this for now and keep track of it in #2686

PR #1836, opened 2018-04-09
Speed-up Travis

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

The coverage now takes 10 minutes instead of 15 minutes before.

### DIFF
--- a/tests/Fixtures/app/AppKernel.php
+++ b/tests/Fixtures/app/AppKernel.php
@@ -38,6 +38,14 @@ class AppKernel extends Kernel
 {
     use MicroKernelTrait;
 
+    public function __construct(string $environment, bool $debug)
+    {
+        parent::__construct($environment, $debug);
+
+        // patch for behat/symfony2-extension not supporting %env(APP_ENV)%
+        $this->environment = $_SERVER['APP_ENV'] ?? $environment;
+    }
+
     public function registerBundles(): array
     {
         $bundles = [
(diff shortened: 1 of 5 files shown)


dunglas:
Thanks @meyerbaptiste, good job!

tests/Fixtures/app/AppKernel.php
```diff
@@ -38,6 +38,14 @@ class AppKernel extends Kernel
 {
     use MicroKernelTrait;
 
+    public function __construct(string $environment, bool $debug)
+    {
+        parent::__construct($environment, $debug);
+
+        // patch for behat/symfony2-extension not supporting %env(APP_ENV)%
+        $this->environment = $_SERVER['APP_ENV'] ?? $environment;
```
sroze: why not `getenv('APP_ENV')`?

dunglas: because it's not thread safe, and for consistency with Flex/Symfony.

teohhanhui:
The best way to speed up Travis build is to use: https://docs.travis-ci.com/user/build-stages/

meyerbaptiste:
PR @teohhanhui? 😄

teohhanhui:
@meyerbaptiste Just out of curiosity, how does this PR achieve the speed-up? :smile:

meyerbaptiste:
By fixing the cache of our test app. In fact, the Symfony2 Behat extension can't resolve `env: '%env(APP_ENV)%'`. So the test app recreated a cache directory every time we ran Behat: `app/cache/env_APP_ENV_<a random sequence of numbers>`.

PR #1843, opened 2018-04-11
value object as ApiResource

komik966:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

I'm aware that this PR still needs work, but I opened it to discuss about value objects as ApiResource.
To get the idea check behat test: https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-6af974d3c9848f846d1ee1e58d88564a and entity: https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-e1fa2952eafb66484a8db3d626ddba31
If you agree with it I'll do further work.

Found (potential) bug:
I've not hardly investigated it but it seems that `PurgeHttpCacheListener` has bug - i tries to get entity id before one is created, but not in every case - it not works on entities which I've added in this PR (`VoDummy*`). For now I commented the line which causes issue (https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-07581ee0a08a8e5193cb8a7ec890c3b7R83)

Questions:
- [ ] ConstructorExtractor (https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-71d84e71877454b3e75f141ed3b2f784) shouldn't be included in `symfony/property-access` ?
- [ ] method `instantiateObject` (https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-3e45e8e6ce08f8a7cbbbda1601d5ff29R170) is almost copied from symfony's `AbstractNormalizer` - it only adds execution of new method `prepareAttributeValue` (https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-3e45e8e6ce08f8a7cbbbda1601d5ff29R204)  - shouldn't it be included in `symfony/serializer`?

TODOS:
- [ ]  lack of required param should show violations (like validation error), not generic error (https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-6af974d3c9848f846d1ee1e58d88564aR93)
- [ ] add tests with IRI
- [ ] handle OneToMany circular (https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-6af974d3c9848f846d1ee1e58d88564aR26, https://github.com/api-platform/core/compare/master...komik966:vo-denormalizer?expand=1#diff-cb3c4ced4e06b3eb1231ca0473a03903R43)
- [ ] add documentation
- [ ] solve `PurgeHttpCacheListener` issue

### DIFF
--- a/src/PropertyInfo/ConstructorExtractor.php
+++ b/src/PropertyInfo/ConstructorExtractor.php
@@ -1,0 +1,34 @@
+<?php
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\PropertyInfo;
+
+use ReflectionClass;
+use ReflectionException;
+use Symfony\Component\PropertyInfo\PropertyAccessExtractorInterface;
+
+class ConstructorExtractor implements PropertyAccessExtractorInterface
+{
+    public function isReadable($class, $property, array $context = [])
+    {
+        return;
+    }
+
+    public function isWritable($class, $property, array $context = [])
+    {
+        $constructor = (new ReflectionClass($class))->getConstructor();
+
+        if(!$constructor) {
+            return null; // give a chance for other PropertyExtractors
+        }
+
+        $constructorParameters = $constructor->getParameters();
+        foreach ($constructorParameters as $constructorParameter) {
+            if ($property === $constructorParameter->getName()) {
+                return true;
+            }
+        }
+
+        return null; // give a chance for other PropertyExtractors
+    }
+}
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -20,6 +20,7 @@ use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\InvalidValueException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
+use ApiPlatform\Core\Exception\PropertyNotFoundException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
@@ -28,9 +29,12 @@ use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Serializer\Exception\LogicException;
+use Symfony\Component\Serializer\Exception\RuntimeException;
 use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 
 /**
  * Base item normalizer.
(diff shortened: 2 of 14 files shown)


komik966:
https://github.com/symfony/symfony/pull/25605

bendavies:
Very nice. This is what I'd be looking to use

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -159,6 +163,82 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
+    protected function instantiateObject(array &$data, $class, array &$context, \ReflectionClass $reflectionClass, $allowedAttributes, string $format = null)
+    {
+        if (null !== $object = $this->extractObjectToPopulate($class, $context, static::OBJECT_TO_POPULATE)) {
+            unset($context[static::OBJECT_TO_POPULATE]);
+
+            return $object;
+        }
+
+        $constructor = $this->getConstructor($data, $class, $context, $reflectionClass, $allowedAttributes);
+        if ($constructor) {
```
Simperfit: you can inline this

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -159,6 +163,82 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
+            $constructorParameters = $constructor->getParameters();
+
+            $params = array();
```
Simperfit: you can use short array syntax

src/PropertyInfo/ConstructorExtractor.php
```diff
@@ -0,0 +1,37 @@
...
+
+class ConstructorExtractor implements PropertyAccessExtractorInterface
+{
+    public function isReadable($class, $property, array $context = array())
+    {
+        return;
+    }
+
+    /**
+     * @throws ReflectionException
```
Simperfit: we do not document exception not thrown directly by the method

src/PropertyInfo/ConstructorExtractor.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace ApiPlatform\Core\PropertyInfo;
+
+use ReflectionClass;
+use ReflectionException;
+use Symfony\Component\PropertyInfo\PropertyAccessExtractorInterface;
+
+class ConstructorExtractor implements PropertyAccessExtractorInterface
+{
+    public function isReadable($class, $property, array $context = array())
```
Simperfit: you can use short array syntax

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -159,6 +163,82 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
+            foreach ($constructorParameters as $constructorParameter) {
+                $paramName = $constructorParameter->name;
+                $key = $this->nameConverter ? $this->nameConverter->normalize($paramName) : $paramName;
+
+                $allowed = false === $allowedAttributes || \in_array($paramName, $allowedAttributes);
+                $ignored = !$this->isAllowedAttribute($class, $paramName, $format, $context);
+                if ($constructorParameter->isVariadic()) {
+                    if ($allowed && !$ignored && (isset($data[$key]) || array_key_exists($key, $data))) {
```
Simperfit: why using both isset and array_key_exists?

komik966:
Thanks for review, I'll fix it.
The reason was the need for properties without setters - in some cases it simplifies our domain model.

komik966:
@Simperfit fixed, I leaved `AbstractItemNormalizer::instantiateObject` as is before we decide what to do with it (it is almost copied from https://github.com/symfony/serializer/blob/4.0/Normalizer/AbstractNormalizer.php#L312)

src/PropertyInfo/ConstructorExtractor.php
```diff
@@ -0,0 +1,37 @@
+<?php
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\PropertyInfo;
+
+use ReflectionClass;
+use ReflectionException;
+use Symfony\Component\PropertyInfo\PropertyAccessExtractorInterface;
+
+class ConstructorExtractor implements PropertyAccessExtractorInterface
```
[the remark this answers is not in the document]
Nek-: I don't think so. The fact that a parameter is writable is driven by the fact that you can write it for an _instantiated_ object. So this could never be inside the Symfony component.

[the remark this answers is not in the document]
dunglas: I tend to agree, this a low level and common need, it should be in Symfony, not in API Platform (history: we moved the code PropertyInfo code from API Platform to a standalone library then to Symfony for this exact same reason).

src/Serializer/AbstractItemNormalizer.php
dunglas: For performance I guess (`isset` is faster)

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -159,6 +163,82 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
+                        if (!\is_array($data[$paramName])) {
+                            throw new RuntimeException(sprintf('Cannot create an instance of %s from serialized data because the variadic parameter %s can only accept an array.',
```
dunglas: Can you add quotes around values in the message?

src/PropertyInfo/ConstructorExtractor.php
```diff
@@ -0,0 +1,34 @@
...
+    public function isReadable($class, $property, array $context = [])
+    {
+        return;
+    }
+
+    public function isWritable($class, $property, array $context = [])
+    {
+        $constructor = (new ReflectionClass($class))->getConstructor();
+
+        if(!$constructor) {
```
meyerbaptiste: You should test if the class is instantiable.

src/PropertyInfo/ConstructorExtractor.php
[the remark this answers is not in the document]
komik966: Adding `property_info.initializable_extractor` tag is better than using `property_info.access_extractor`, I like this approach

komik966:
almost done in #1749

PR #1844, opened 2018-04-11
feature: improve ids management for association and fix invalid iri/uuid in SearchFilter

### LINKED ISSUE #775: Bug ? Get collection with filters and several IRI
Hi 

If I call: 

```
GET: votes?decision=/decision/2&creator=/users/1
```

capture d ecran 2016-09-28 a 16 57 28

I got that error: 

```
An exception occurred while executing 'SELECT count(DISTINCT v0_.id) AS sclr_0 FROM vote v0_ INNER JOIN decision d1_ ON v0_.decision_id = d1_.id WHERE d1_.id = ?' with params ["\/decisions\/414&creator=\/users\/43"]:

SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "/decisions/414&creator=/users/43"
```

And the route used is: 
capture d ecran 2016-09-28 a 16 32 52

However, if I do:

```
GET: votes?decision=2&creator=1
```

It's working 
Any idea of why I can't use filters with several IRI ?

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #775
| License       | MIT
| Doc PR        | 

This PRs update the ids management for association in the SearchFilter.

I've not updated the single at the moment because I think it should be done in another PR. 
This fixes #775.

Something I'm not sure is the behaviour, do we ignore the filter (this is the current implementation) or do we return 0 elements (like by replacing the invalid values with null instead of logging + return) ?

### DIFF
--- a/features/doctrine/search_filter.feature
+++ b/features/doctrine/search_filter.feature
@@ -20,37 +20,62 @@ Feature: Search filter on collections
     When I send a "GET" request to "/dummy_cars?colors.prop=red"
     Then the response status code should be 200
     And the JSON should be deep equal to:
-    """
+		"""
     {
-        "@context": "/contexts/DummyCar",
-        "@id": "/dummy_cars",
-        "@type": "hydra:Collection",
-        "hydra:member": [
+      "@context": "/contexts/DummyCar",
+      "@id": "/dummy_cars",
+      "@type": "hydra:Collection",
+      "hydra:member": [
+        {
+          "@id": "/dummy_cars/1",
+          "@type": "DummyCar",
+          "colors": [
             {
-                "@id": "/dummy_cars/1",
-                "@type": "DummyCar",
-                "colors": [
-                    {
-                        "@id": "/dummy_car_colors/1",
-                        "@type": "DummyCarColor",
-                        "prop": "red"
-                    },
-                    {
-                        "@id": "/dummy_car_colors/2",
-                        "@type": "DummyCarColor",
-                        "prop": "blue"
-                    }
-                ]
+              "@id": "/dummy_car_colors/1",
+              "@type": "DummyCarColor",
+              "prop": "red"
+            },
+            {
+              "@id": "/dummy_car_colors/2",
+              "@type": "DummyCarColor",
+              "prop": "blue"
             }
-        ],
-        "hydra:totalItems": 1,
-        "hydra:view": {
-            "@id": "/dummy_cars?colors.prop=red",
-            "@type": "hydra:PartialCollectionView"
-        },
-        "hydra:search": {
+          ],
+          "secondColors": [
+            {
+              "@id": "/dummy_car_colors/1",
+              "@type": "DummyCarColor",
+              "prop": "red"
+            },
+            {
+              "@id": "/dummy_car_colors/2",
+              "@type": "DummyCarColor",
+              "prop": "blue"
+            }
+          ],
+          "thirdColors": [
+            {
+              "@id": "/dummy_car_colors/1",
+              "@type": "DummyCarColor",
+              "prop": "red"
+            },
+            {
+              "@id": "/dummy_car_colors/2",
+              "@type": "DummyCarColor",
+              "prop": "blue"
+            }
+          ],
+          "uuid": []
+        }
+      ],
+      "hydra:totalItems": 1,
+      "hydra:view": {
+        "@id": "/dummy_cars?colors.prop=red",
+        "@type": "hydra:PartialCollectionView"
+      },
+      "hydra:search": {
         "@type": "hydra:IriTemplate",
-        "hydra:template": "\/dummy_cars{?availableAt[before],availableAt[strictly_before],availableAt[after],availableAt[strictly_after],canSell,foobar[],foobargroups[],foobargroups_override[],colors.prop,name}",
+        "hydra:template": "/dummy_cars{?availableAt[before],availableAt[strictly_before],availableAt[after],availableAt[strictly_after],canSell,foobar[],foobargroups[],foobargroups_override[],colors.prop,colors,colors[],secondColors,secondColors[],thirdColors,thirdColors[],uuid,uuid[],name}",
         "hydra:variableRepresentation": "BasicRepresentation",
         "hydra:mapping": [
           {
@@ -85,8 +110,20 @@ Feature: Search filter on collections
           },
           {
             "@type": "IriTemplateMapping",
+            "variable": "colors",
+            "property": "colors",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
             "variable": "colors.prop",
             "property": "colors.prop",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "colors[]",
+            "property": "colors",
             "required": false
           },
           {
@@ -111,6 +148,42 @@ Feature: Search filter on collections
             "@type": "IriTemplateMapping",
             "variable": "name",
             "property": "name",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "secondColors",
+            "property": "secondColors",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "secondColors[]",
+            "property": "secondColors",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "thirdColors",
+            "property": "thirdColors",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "thirdColors[]",
+            "property": "thirdColors",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "uuid",
+            "property": "uuid",
+            "required": false
+          },
+          {
+            "@type": "IriTemplateMapping",
+            "variable": "uuid[]",
+            "property": "uuid",
             "required": false
           }
         ]
@@ -389,8 +462,7 @@ Feature: Search filter on collections
         "@id": {"pattern": "^/dummies$"},
         "@type": {"pattern": "^hydra:Collection$"},
         "hydra:member": {
-          "type": "array",
-          "maxItems": 0
+          "type": "array"
         },
         "hydra:view": {
           "type": "object",
@@ -655,3 +727,4 @@ Feature: Search filter on collections
       }
     }
     """
+
--- a/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Doctrine\MongoDbOdm\Filter;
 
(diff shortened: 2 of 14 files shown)


Simperfit:
@dunglas this needs a review

Simperfit:
@dunglas this needs a review

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -414,4 +434,18 @@ protected function hasValidValues(array $values, $type = null): bool
...
+     * @see regexp taken from https://github.com/ramsey/uuid
+     *
+     * @param string|int $identifier
+     *
+     * @return bool
+     */
+    protected function isValidUuid($identifier): bool
+    {
+        return (bool) preg_match('/^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$/', $identifier);
+    }
```
soyuka: I'm -1 for this. I'd just let the failing doctrine operation throw.

Toflar: Yeah, I agree. Also because you can create your own uuid generator, why would you validate that here?

Simperfit: Then we should not support uuid in here, why are we testing that an int is a valid int then ?

Toflar: Not exactly sure what you mean but to me it should work like this: Whatever I set to be the identifier in doctrine should be searchable and whatever the user is querying on our API, it should not validate the value of the identifier. If doctrine can find an entity based on it it's fine, otherwise not. I don't see why we should validate the value passed by the customer here. Maybe you can elaborate?

Simperfit: The thing is, when passing a non-valid uuid (but according to the type-checking it's valid because it's a valid string) the filter throws an errors 500. 

IMHO checking that you are passing a valid guid when the doctrine metadatas says that should be a guid is the right call, with this, if the guid is not valid, then it will try to search on it, which will lead to the SearchFilter, not filtering which is the expected behaviour.

AFAIK, removing this breaks how we expect filters to work.

Toflar: I have my own id generator:

```
    /**
     * @ORM\Id
     * @ORM\Column(name="voucher", type="string", length=9, options={"fixed":true})
     * @ORM\CustomIdGenerator(class="App\Doctrine\ORM\EntityManager\VoucherCodeGenerator")
     */
    private $voucher;
```

It's a 9 char, specially generated string. I could not search on that because your UUID validation would fail :(

Simperfit: no, you can, because it's not guid type.
see https://github.com/api-platform/core/pull/1844/files#diff-77ec54980605d6062780fcd93ba31c64R274

soyuka: My believe is that if someone wants a search by uuid he should do it's own because our search filter doesn't support it yet.

As we discussed before, the search filter needs a complete refactor so that it validates most of the needs and it's really hard to do without breaking the so called filter. 
I'd clearly deprecate this class and set up one or two new filters, for example:
- FieldSearchFilter (handles LIKE and such where clauses)
- IdentifierSearchFilter (handles strict equalities, IRIs and everything related to identifiers (`foo.bar=id`). This one would reuse the identifier denormalization introduced a few weeks back).

Toflar:
I really hope this can be released as bugfix because right now, the searchfilter is just completely useless if your identifier column is called `id` 😢

Toflar:
Sure, all my identifiers are uuids and so I named them `uuid` instead of `id` but I could change that.

soyuka:
> Sure, all my identifiers are uuids and so I named them uuid instead of id but I could change that.

IMHO, you should do your own filter to support uuid on identifiers. I agree it'll be closed to a `copy/paste` of the current SearchFilter but I don't think it's a bad idea. It kinda joins the idea I had above (and could maybe be proposed as a new feature for upcoming releases):

> - FieldSearchFilter (handles LIKE and such where clauses)
> - IdentifierSearchFilter (handles strict equalities, IRIs and everything related to identifiers (foo.bar=id). This one would reuse the identifier denormalization introduced a few weeks back).

Simperfit:
I've removed the UUID part.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -59,16 +60,22 @@ class SearchFilter extends AbstractContextAwareFilter
...
      * @param RequestStack|null $requestStack No prefix to prevent autowiring of this deprecated property
      */
-    public function __construct(ManagerRegistry $managerRegistry, $requestStack = null, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null)
+    public function __construct(ManagerRegistry $managerRegistry, $requestStack = null, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null, IdentifiersExtractorInterface $identifiersExtractor = null)
     {
         parent::__construct($managerRegistry, $requestStack, $logger, $properties);
 
+        if (null === $identifiersExtractor) {
+            @trigger_error('Not injecting ItemIdentifiersExtractor is deprecated since API Platform 2.3 and will not be possible anymore in API Platform 3.0.', E_USER_DEPRECATED);
+        }
```
soyuka: Shouldn't we say that not having identifiers extractor leads to bugs?

Simperfit: It leads to the existing behaviour, we can add it in the message.

Simperfit:
PR rebased with master

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -448,11 +448,12 @@ private function registerGraphqlConfiguration(ContainerBuilder $container, array
      * @param XmlFileLoader $loader
      * @param bool          $useDoctrine
      */
-    private function registerBundlesConfiguration(array $bundles, array $config, XmlFileLoader $loader, bool $useDoctrine)
+    private function registerBundlesConfiguration(array $bundles, array $config, XmlFileLoader $loader, bool $useDoctrine, ContainerBuilder $container)
     {
         // Doctrine ORM support
         if ($useDoctrine) {
             $loader->load('doctrine_orm.xml');
+            $container->getDefinition('api_platform.doctrine.orm.search_filter')->setArgument('$identifiersExtractor', $container->getDefinition('api_platform.identifiers_extractor.cached'));
```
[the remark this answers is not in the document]
Simperfit: It's the only way I found to be able to have it working it properly and without breaking the order filter. I don't know why it breaks the order filter btw.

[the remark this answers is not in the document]
meyerbaptiste: I don't get it... What is the link between our search filter and our order filter?!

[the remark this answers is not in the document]
Simperfit: I don't know :p! but try it, I've posted a message in slack about this, no one could tell me at the time what was the issue

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
```diff
@@ -48,7 +48,6 @@
         </service>
 
         <!-- Filter -->
-
```
Simperfit: to revert

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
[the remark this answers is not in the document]
meyerbaptiste: I mean, `<argument key="$identifiersExtractor" type="service" id="api_platform.identifiers_extractor.cached" />` would be better.
But what worries me is that the user may have bugs when it extends the class and it potentially renames the arguments...

soyuka:
doing this `EU-FOSSA Hackathon`

features/doctrine/search_filter.feature
```diff
@@ -389,8 +462,7 @@ Feature: Search filter on collections
         "@id": {"pattern": "^/dummies$"},
         "@type": {"pattern": "^hydra:Collection$"},
         "hydra:member": {
-          "type": "array",
-          "maxItems": 0
```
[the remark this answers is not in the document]
soyuka: This is because the test was wrong in the first place, an invalid filter should not be applied and therefore should return every items of the collection. We fixed that here.

tests/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilterTest.php
```diff
@@ -344,15 +345,6 @@ public function provideApplyTestData(): array
                                 ],
                             ],
                         ],
-                        [
-                            '$match' => [
```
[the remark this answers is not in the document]
soyuka: Same reason as above.

tests/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilterTest.php
```diff
@@ -469,17 +461,7 @@ public function provideApplyTestData(): array
                     $filterFactory,
                 ],
                 'invalid value for relation' => [
-                    [
-                        [
-                            '$match' => [
```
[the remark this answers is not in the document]
soyuka: Same reason as above.

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -40,12 +42,17 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
 
     public const DOCTRINE_INTEGER_TYPE = MongoDbType::INTEGER;
 
-    public function __construct(ManagerRegistry $managerRegistry, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null)
+    public function __construct(ManagerRegistry $managerRegistry, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null, IdentifiersExtractorInterface $identifiersExtractor = null)
     {
         parent::__construct($managerRegistry, $logger, $properties);
 
+        if (null === $identifiersExtractor) {
+            @trigger_error('Not injecting ItemIdentifiersExtractor is deprecated since API Platform 2.5 and can lead to unexpected behaviors, it will not be possible anymore in API Platform 3.0.', E_USER_DEPRECATED);
```
meyerbaptiste: Mongo is marked as `@experimental`. Not sure BC is relevant here...

PR #1846, opened 2018-04-11
Fix the profiler using undefined value

sroze:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1825
| License       | MIT
| Doc PR        | ø

Though it was a very good idea, it introduced a bug 😉


antograssiot:
Just out of curiosity, how can't it be defined, it is set a null by default if I remember correctly (I'll have a look when I'll get a computer)  ? Thank for the fix anyway

antograssiot:
Forget my comment I'm dumb

sroze:
Good point @jdeniau, I've updated it 👍

PR #1850, opened 2018-04-12
Merge 2.2

sroze:
It's a bit annoying not to be able to simply to `git merge 2.2 && git push upstream master` locally 🤔


Simperfit:
Yeah, because master is protected, only @dunglas can do this AFAIK.

soyuka:
> It's a bit annoying not to be able to simply to git merge 2.2 && git push upstream master

Would be even better if this could be automated. Although we can have conflicts sometimes or weird tests issues (see my merge 2 weeks ago) so it's not that bad to force us to do a merge request and see if tests are okay.

teohhanhui:
> It's a bit annoying not to be able to simply to git merge 2.2 && git push upstream master locally

That's too dangerous. With great power comes great responsibility. :laughing:

teohhanhui:
@sroze I believe `master` was accidentally merged into `2.2` as part of preparing this PR. Did you use the very dangerous GitHub "merge changes" button? :rofl:

For a second I was tempted to do so too in #1909, but then I realized the consequences before bad things happened...

@dunglas I think we need to push-protect the stable and old stable branches too? (Is the GitHub "merge changes" button subject to that restriction though? Or is there a way to disable the GitHub "merge changes" button? :x)

bendavies:
Huh. How did master get merged into 2.2?

teohhanhui:
See my comment right above yours. :stuck_out_tongue: GitHub shows a highly dangerous "merge changes" button when the PR branch is outdated compared to the base branch.

bendavies:
Aha! O dear!

sroze:
Outch. 🤔

sroze:
Seems like the last commit of this PR should be reverted on 2.2, WDYT? 🤔

teohhanhui:
It creates too much conflicts. I tried it last night.

I'd suggest instead to reset the working directory to its parent commit on the `2.2` side and reapply the commits since then in the working directory. A manual revert commit, sort of. (And then we'll have to take care to revert this revert commit when we merge into `master`? Or if I remember correctly, there should be a way to indicate that the branches are merged without actually touching the index.)

weaverryan:
@teohhanhui we're having a similar problem on the Symfony docs right now. Would you be able/willing to walk us through what your process was to solve this?

teohhanhui:
@weaverryan I'll take a look and let you know in a bit. :smile:

PR #1847, opened 2018-04-11
Change default priority on autoconfiguration Doctrine DataProvider extensions

Mathieudewet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1815
| License       | MIT
| Doc PR        |

PR #1852, opened 2018-04-12
Revert "Change default priority on autoconfiguration Doctrine DataProvider extensions"

Simperfit:
Reverts api-platform/core#1847


Simperfit:
Should be merged only when passing to 3.0.

Issue #1840, opened 2018-04-09
Can't use a plain id identifier for a relation denormalization


Taluu:
#1762 reverted #1547 to solve a bug, but we were kind of relying on that to be able to fetch / set an entity from its id without using iris or a nested `id` object.

```json
{
    "id": "d1890e8f-fd21-5283-9087-0df87279077d",
    "mainMedia": "d1890e8f-fd21-5283-9087-0df87279077d"
}
```

Used to work up to 2.2.3, but since #1762 (and thus 2.2.4...), it doesn't anymore and triggers an error. I know this is not standard (and kinda relying on a bug...), but it's how the front currently works (I opened a ticket on their side to solve their mess). But still, supporting that would be useful imo.

I have an `IriConverter` decorator, so I could intercept this, but the fact is that in the iri converter, I have no "context" such as the main entity, the property, or its type so I could tinker and make an iri from that before passing it to the decorated service.

Taluu:
It seems that this block : https://github.com/api-platform/core/blob/2.2/src/Serializer/AbstractItemNormalizer.php#L294-L302 doesn't take into account the allowPlainIddentifiers. Unless I read the feature wrong, this doesn't allow to use plain identifier to identify a resource.

Taluu:
So, I kinda got further on that point. It seems that my property is a writable link, So this passes into the normal serializer and never reaches the part when the `allowPlainIdentifier` is executed.

I have 2 objects : A `Product` and `Media`, and they are linked through a `mainMedia` property (which as its setter `setMainMedia`) on the `Product`.

So the request is a PUT on the product, with the body of the first message. And I end up with the `Invalid IRI ?` error.

bendavies:
can this be closed?

Taluu:
Thought this was already closed. Closing then. :}

PR #1855, opened 2018-04-12
Fix relations denormalization with plain ids

Taluu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1840 
| License       | MIT
| Doc PR        | ~

#1547 introduced an `is_array` check to prevent the normalize to act on writable relations so that we could use plain identifiers, and was reverted by #1762 because not all relations are arrays, and can be iris.

But this revert triggered also the fact that we cannot use plain ids anymore on writable relations. This PR aims to fix that ; if we encounter the invalid argument exception while trying to denormalize the relation, and that we allowed plain identifiers and we have an item provider, we should try the next block if it's not an array. Now that I think about it, I should probably add a `!is_array` check or something ?

The tests passes, but I'm not sure how I can add a test to reflect this, so any hint would be welcomed...

BTW, shouldn't we add a more specific exception for the invalid iri ? Testing the message isn't specially clean IMO... and as it's not the scope of this PR, I had to make do with this.

### DIFF
--- a/src/Exception/InvalidValueException.php
+++ b/src/Exception/InvalidValueException.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+class InvalidValueException extends InvalidArgumentException
+{
+}
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\Api\OperationType;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\InvalidValueException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
@@ -157,7 +158,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     protected function setAttributeValue($object, $attribute, $value, $format = null, array $context = [])
     {
         if (!\is_string($attribute)) {
-            throw new InvalidArgumentException('Invalid value provided (invalid IRI?).');
+            throw new InvalidValueException('Invalid value provided (invalid IRI?).');
         }
 
         $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $attribute, $this->getFactoryOptions($context));
@@ -306,11 +307,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
             try {
                 return $this->serializer->denormalize($value, $className, $format, $context);
-            } catch (InvalidArgumentException $e) {
-                if ('Invalid value provided (invalid IRI?).' !== $e->getMessage()) {
-                    throw $e;
-                }
-
+            } catch (InvalidValueException $e) {
                 if (!$this->allowPlainIdentifiers || null === $this->itemDataProvider) {
                     throw $e;
                 }
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -304,7 +304,17 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             $context['resource_class'] = $className;
             $context['api_allow_update'] = true;
 
-            return $this->serializer->denormalize($value, $className, $format, $context);
+            try {
+                return $this->serializer->denormalize($value, $className, $format, $context);
+            } catch (InvalidArgumentException $e) {
+                if ('Invalid value provided (invalid IRI?).' !== $e->getMessage()) {
+                    throw $e;
+                }
+
+                if (!$this->allowPlainIdentifiers || null === $this->itemDataProvider) {
+                    throw $e;
+                }
+            }
         }
 
         if (!\is_array($value)) {
--- a/features/main/relation.feature
+++ b/features/main/relation.feature
@@ -512,13 +512,50 @@ Feature: Relations support
     }
     """
 
+  Scenario: Passing a (valid) plain identifier on a relation
+    When I add "Content-Type" header equal to "application/json"
+    And I send a "POST" request to "/dummies" with body:
+    """
+    {
+      "relatedDummy": "1",
+      "relatedDummies": ["1"],
+      "name": "Dummy with plain relations"
+    }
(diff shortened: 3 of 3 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -308,7 +308,17 @@ protected function denormalizeRelation(string $attributeName, PropertyMetadata $
             $context['resource_class'] = $className;
             $context['api_allow_update'] = true;
 
-            return $this->serializer->denormalize($value, $className, $format, $context);
+            try {
+                return $this->serializer->denormalize($value, $className, $format, $context);
+            } catch (InvalidArgumentException $e) {
+                if ('Invalid value provided (invalid IRI?).' !== $e->getMessage()) {
```
Simperfit: Maybe it could be better to introduce a new Exception extending the `InvalidArgumentException`.
That said, we wouldn't to the check the message, WDYT ?
https://github.com/api-platform/core/blob/master/src/Serializer/AbstractItemNormalizer.php#L164

Taluu: Yep, added a comment on that on my description, but IMO this is not really the scope of this PR, right ? Or should I change it anyway ?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -308,7 +309,17 @@ protected function denormalizeRelation(string $attributeName, PropertyMetadata $
             $context['resource_class'] = $className;
             $context['api_allow_update'] = true;
 
-            return $this->serializer->denormalize($value, $className, $format, $context);
+            try {
+                return $this->serializer->denormalize($value, $className, $format, $context);
+            } catch (InvalidValueException $e) {
+                if ('invalid IRI?' !== $e->getReason()) {
```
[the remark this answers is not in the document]
Taluu: Well, wasn't sure if we should have an invalid value exception for the invalid iri, but I could alter that to use one instead ?

src/Exception/InvalidValueException.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+class InvalidValueException extends InvalidArgumentException
```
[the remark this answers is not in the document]
Taluu: It doesn't seem to extend / implement the invalid argument exception, Wouldn't that be a BC Break ?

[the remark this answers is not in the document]
soyuka: Don't think so. Anyway I'd be fine with a new exception just trying to avoid multiplying them :p. Maybe we can call this one `InvalidIRIException`?

src/Serializer/AbstractItemNormalizer.php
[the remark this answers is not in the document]
soyuka: Not sure I got that. This exception is only thrown when IRI is invalid right?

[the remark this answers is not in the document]
Taluu: Yeah, but the part between parentheses looked variable, that's why I added a reason. I can merge it if needed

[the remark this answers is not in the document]
Taluu: Well, we had `Invalid value (Invalid IRI?).`, so I thought that we could have multiple `Invalid Value (reason)`, and it could sometimes be `Invalid Value (Invalid IRI?)`. That's all, but then, I can make a `InvalidIriException` and be done with it :P

src/Exception/InvalidValueException.php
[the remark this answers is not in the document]
Taluu: I'm not against multiplying exceptions. It's just that in 2.2.4 (and 2.2.5), we should expect a `InvalidArgumentException` (or a child of it), not something else, e.g `Exception`.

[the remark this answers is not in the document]
Taluu: Unless you meant copying / pasting the code from the `InvalidIdentifierException` ?

src/Serializer/AbstractItemNormalizer.php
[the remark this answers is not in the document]
soyuka: Yes please :D

Taluu:
Updated @soyuka but not satisfied with the tests, looks like something is wrong, as the result is `null` for the `relatedDummy`. Need to check out why :/

src/Exception/InvalidIriException.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace ApiPlatform\Core\Exception;
+
+class InvalidIriException extends InvalidArgumentException
+{
+    private $reason;
+
+    public function __construct(int $code = 0, \Throwable $previous = null)
+    {
+        parent::__construct("Invalid value provided (Invalid IRI?).", $code, $previous);
```
soyuka: Can't we just remove everything in this class and call the exception with a message?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -161,7 +162,7 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
     protected function setAttributeValue($object, $attribute, $value, $format = null, array $context = [])
     {
         if (!\is_string($attribute)) {
-            throw new InvalidArgumentException('Invalid value provided (invalid IRI?).');
+            throw new InvalidIriException();
```
soyuka:
```
 throw new InvalidIriException('Invalid value provided (invalid IRI?).');
```

Taluu: What would be the point of the exception if the message is not always the same one ? It's like using a generic exception... look at the exceptions from symfony, they each have somehting customisable but a defined (and thus customizable) message (even though the message is also quite pointless and not used)

soyuka: Because we can have different reasons for an invalid IRI exception:
- bad format
- route not found
- data not found

for example

Taluu: Yeah that's why I had the `(invalid IRI?)` variable, so we would always have `Invalid value provider (REASON)`. But `¯\_(ツ)_/¯` because it's the same thing actually. :}

dunglas: I agree with @soyuka (and every other exception we have for now work like this). You can also add the message as the default value of the `$message` parameter.

Taluu: Yep, changed and removed the constructor altogether, as @soyuka suggested https://github.com/api-platform/core/pull/1855#discussion_r181410250

Taluu:
Indeed, I have some problems with the feature actually. Need some help looking into that, why it doesn't seem to take the plain id into account...

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -161,7 +162,7 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
     protected function setAttributeValue($object, $attribute, $value, $format = null, array $context = [])
     {
         if (!\is_string($attribute)) {
-            throw new InvalidArgumentException('Invalid value provided (invalid IRI?).');
+            throw new InvalidValueException('Invalid value provided (Invalid IRI?).');
```
dunglas: the message should be reverted (why an upper case I)?

Taluu:
Looks like I managed to fix the tests :tada:

dunglas:
Thanks @Taluu

Issue #1664, opened 2018-01-24
Consider splitting the IriConverter in 2 classes


lyrixx:
I got the following error with Symfony

```
Circular reference detected for service "api_platform.hydra.normalizer.entrypoint", path: 
api_platform.hydra.normalizer.entrypoint
-> api_platform.iri_converter
-> api_platform.item_data_provider
-> app.api.data_provider.agent_rule
-> app.elasticsearch.straight_rule
-> serializer
-> api_platform.hydra.normalizer.entrypoint".
```
Basically, my "agent_rule" are stored in Elasticsearch. So I have a custom DataProvider. This one depends on the serializer (because We serialize/unserialize data to/from Elasticsearch).

AFAIU, the IriConverter does 2 symmetrical things: Item <-> IRI.
I think i would ease the development if theses 2 things would be in 2 different classes to avoid such circular reference.

WDYT?

dunglas:
Good idea! 👍 on my side (at least two interfaces).

soyuka:
:+1: two interfaces seems great, this should be easy to implement

PR #1859, opened 2018-04-12
Feature/split iri converter

Mathieudewet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | 
| Fixed tickets | #1664
| License       | MIT
| Doc PR        | 


src/Api/IriConverterInterface.php
```diff
@@ -13,78 +13,15 @@
...
 namespace ApiPlatform\Core\Api;
 
-use ApiPlatform\Core\Exception\InvalidArgumentException;
-use ApiPlatform\Core\Exception\ItemNotFoundException;
-use ApiPlatform\Core\Exception\RuntimeException;
 
 /**
  * Converts item and resources to IRI and vice versa.
  *
+ * @deprecated please use ItemToIriConverterInterface or IriToItemConverterInterface to fit your needs instead
```
Simperfit: one @trigger_errors should be added.
```
IriConverterInterface is deprecated since API Platform 2.3 and will be removed in API Platform 3.0, use ItemToIriConverterInterface and/or IriToItemConverterInterface instead.
```
both class should be fully qualified.

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -36,7 +37,7 @@
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class IriConverter implements IriConverterInterface
+final class IriConverter implements IriToItemConverterInterface, ItemToIriConverterInterface
```
Simperfit: I think you need to check in the whole code base where IriConverterInterface is used.

antograssiot: Pretty sure that removing an interfave from a class is a BC break following SF rules...

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -41,7 +41,7 @@
     private $propertyAccessor;
     private $tags = [];
 
-    public function __construct(PurgerInterface $purger, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
+    public function __construct(PurgerInterface $purger, ItemToIriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
```
soyuka: Unless the class is marked as `@internal` you can't change it's signature.

Issue #102, opened 2015-06-09
Filters logging


theofidry:
It would be great to have some tools to ease the development and debug of filters, for starter:
- list all filters
- get which filters are enabled for a given resource
- for a given URI, which filters are applied, output of the DQL query if using Doctrine?
- logs in the profiler with detailed view of which filter is being used, applied and what it has done

I guess it won't be a priority but I still find it worth being mentioned.

alanpoulain:
Has been done in https://github.com/api-platform/core/pull/1865

PR #1865, opened 2018-04-13
Enhance filters description in DataCollector/WebProfiler

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

before:
capture d ecran 2018-04-13 a 21 03 50

after:
capture d ecran 2018-04-13 a 21 01 00

and a warning in the toolbar:
capture d ecran 2018-04-13 a 21 00 33
capture d ecran 2018-04-13 a 21 00 48

or when no filters are available:
capture d ecran 2018-04-14 a 08 36 46

### DIFF
--- a/src/Bridge/Symfony/Bundle/DataCollector/RequestDataCollector.php
+++ b/src/Bridge/Symfony/Bundle/DataCollector/RequestDataCollector.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DataCollector;
 
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Psr\Container\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
@@ -25,10 +26,12 @@ use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 final class RequestDataCollector extends DataCollector
 {
     private $metadataFactory;
+    private $filterLocator;
 
-    public function __construct(ResourceMetadataFactoryInterface $metadataFactory)
+    public function __construct(ResourceMetadataFactoryInterface $metadataFactory, ContainerInterface $filterLocator)
     {
         $this->metadataFactory = $metadataFactory;
+        $this->filterLocator = $filterLocator;
     }
 
     /**
@@ -36,14 +39,28 @@ final class RequestDataCollector extends DataCollector
      */
     public function collect(Request $request, Response $response, \Exception $exception = null)
     {
+        $counters = ['ignored_filters' => 0];
         $resourceClass = $request->attributes->get('_api_resource_class');
         $resourceMetadata = $resourceClass ? $this->metadataFactory->create($resourceClass) : null;
+
+        $filters = [];
+        foreach ($resourceMetadata ? $resourceMetadata->getAttribute('filters', []) : [] as $id) {
+            if ($this->filterLocator->has($id)) {
+                $filters[$id] = get_class($this->filterLocator->get($id));
+                continue;
+            }
+
+            $filters[$id] = null;
+            ++$counters['ignored_filters'];
+        }
 
         $this->data = [
             'resource_class' => $resourceClass,
             'resource_metadata' => $resourceMetadata ? $this->cloneVar($resourceMetadata) : null,
             'acceptable_content_types' => $request->getAcceptableContentTypes(),
             'request_attributes' => RequestAttributesExtractor::extractAttributes($request),
+            'filters' => $filters,
+            'counters' => $counters,
         ];
     }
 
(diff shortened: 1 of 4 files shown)


antograssiot:
To better understand, the whole page would be :
screenshot-2018-4-13 symfony profiler

src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
```diff
@@ -86,22 +93,26 @@
...
-                                {{ filter }}
-                            </td>
-                        </tr>
-                    {% endfor %}
-                {% endif %}
+                {% for id, filter in collector.filters %}
+                    <tr{% if filter is same as(null) %} class="status-warning"{% endif %}>
+                        <td>
+                            {{ id }}
+                            {% if filter is same as(null) %}<span class="newline text-muted">ignored filter</span>{% else %}{{ dump(filter) }}{% endif %}
```
IonBazan: This condition is copy-pasted above. You could bind it to a variable.
Also, make it multiline.

antograssiot: thanks @IonBazan, it has been updated

antograssiot:
Also added a message when there's no collection operation.
capture d ecran 2018-04-14 a 15 50 19

src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
```diff
@@ -60,6 +67,12 @@
             <tbody>
                 {% for key, itemOperation in collector.resourceMetadata.itemOperations %}
                     {{ apiPlatform.operationLine(key, itemOperation, collector.requestAttributes.item_operation_name|default('')) }}
+                {% else %}
```
antograssiot: maybe this block is not needed as there is always at least one item operation so the IRI can be generated, is this assumption correct ?

dunglas: Yes, you can drop it.

src/Bridge/Symfony/Bundle/DataCollector/RequestDataCollector.php
```diff
@@ -25,25 +26,44 @@
...
         $this->metadataFactory = $metadataFactory;
+        $this->filterLocator = $filterLocator;
     }
 
     /**
      * {@inheritdoc}
      */
     public function collect(Request $request, Response $response, \Exception $exception = null)
     {
+        $counters = [
```
dunglas: Can be inlined.

src/Bridge/Symfony/Bundle/DataCollector/RequestDataCollector.php
```diff
@@ -25,25 +26,44 @@
...
+            'ignored_filters' => 0,
+        ];
+
         $resourceClass = $request->attributes->get('_api_resource_class');
         $resourceMetadata = $resourceClass ? $this->metadataFactory->create($resourceClass) : null;
 
+        $filters = [];
+        foreach (($resourceMetadata ? $resourceMetadata->getAttribute('filters', []) : []) as $id) {
```
dunglas: The parenthesis are useless, right?

src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
```diff
@@ -11,6 +11,7 @@
 
 {% block toolbar %}
     {% set icon %}
+        {% set status_color = collector.counters.ignored_filters|default(0) ? 'yellow': 'default' %}
```
dunglas: Space to add between `yellow` and `:`
Can also be `default(false)`

src/Bridge/Symfony/Bundle/Resources/views/DataCollector/request.html.twig
```diff
@@ -19,9 +20,15 @@
             <b>Resource Class</b>
             <span>{{ collector.resourceClass|default('Not an API Platform resource') }}</span>
         </div>
+        {% if collector.counters.ignored_filters|default(0) %}
```
dunglas: Same here `default(false)`

tests/Bridge/Symfony/Bundle/DataCollector/RequestDataCollectorTest.php
```diff
@@ -36,6 +37,8 @@ class RequestDataCollectorTest extends TestCase
 
     private $metadataFactory;
 
+    private $filterLocator;
```
dunglas: Usually, we don't add a blank line between props when there are no doc block.

Simperfit:
Is this going to fix #102 ?

antograssiot:
Updated and rebased @dunglas.
@Simperfit it only address part of it as for now you'll get filters enabled for a given resource, a list of the ignored/badly declared filtres but not the list of the active filtres on the request for exemple.

dunglas:
Thanks @antograssiot!

PR #1866, opened 2018-04-14
The OrderFilter is still triggering bad deprecation/function calls

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Follows #1663 and fix it correctly.
The added test throws a deprecation without the fix


dunglas:
Thanks for the investigation and the fix @antograssiot. It's very appreciated!

Issue #1867, opened 2018-04-14
[GraphQL] Related/nested property filters does not work.


IonBazan:
Minimal working example:
```
  @createSchema
  Scenario: Retrieve a collection filtered using the related search filter
    Given there are 1 dummy objects having each 2 relatedDummies
    And there are 1 dummy objects having each 3 relatedDummies
    When I add "Accept" header equal to "application/hal+json"
    And I send a "GET" request to "/dummies?relatedDummies.name=RelatedDummy31"
    Then the response status code should be 200
    And the response should be in JSON
    And the JSON node "_embedded.item" should have 1 element
    When I send the following GraphQL request:
    """
    {
      dummies(relatedDummies_name: "RelatedDummy31") {
        edges {
          node {
            id
          }
        }
      }
    }
    """
    And the response status code should be 200
    And the JSON node "data.dummies.edges" should have 1 element
```

Response should contain only one element just like in JSON. The reason is probably that `relatedDummies.name` is being converted to `relatedDummies_name` for GraphQL and is not handled properly on filter resolution.

antograssiot:
It looks related to what #1743 is trying to address.

IonBazan:
Closing as it is duplicate of #1714.

PR #1868, opened 2018-04-15
Allow GraphQL to filter on nested property

### LINKED ISSUE #1714: Search Filter does not filter collection by nested relation field in GraphQL

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1714, #1867
| License       | MIT
| Doc PR        | 

Possible replacement for #1743 without updating the filter but just the way filters are passed to the data provider.

ping @alanpoulain  @IonBazan

### DIFF
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -88,3 +88,27 @@ Feature: Collections filtering
     And the JSON node "data.dummies.edges[1].node.relatedDummies.edges" should have 0 elements
     And the JSON node "data.dummies.edges[2].node.relatedDummies.edges" should have 1 element
     And the JSON node "data.dummies.edges[2].node.relatedDummies.edges[0].node.name" should be equal to "RelatedDummy13"
+
+  @createSchema
+  Scenario: Retrieve a collection filtered using the related search filter
+    Given there are 1 dummy objects having each 2 relatedDummies
+    And there are 1 dummy objects having each 3 relatedDummies
+    When I add "Accept" header equal to "application/hal+json"
+    And I send a "GET" request to "/dummies?relatedDummies.name=RelatedDummy31"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the JSON node "_embedded.item" should have 1 element
+    When I send the following GraphQL request:
+    """
+    {
+      dummies(relatedDummies_name: "RelatedDummy31") {
+        edges {
+          node {
+            id
+          }
+        }
+      }
+    }
+    """
+    And the response status code should be 200
+    And the JSON node "data.dummies.edges" should have 1 element
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -93,11 +93,6 @@ Feature: Collections filtering
   Scenario: Retrieve a collection filtered using the related search filter
     Given there are 1 dummy objects having each 2 relatedDummies
     And there are 1 dummy objects having each 3 relatedDummies
-    When I add "Accept" header equal to "application/hal+json"
-    And I send a "GET" request to "/dummies?relatedDummies.name=RelatedDummy31"
-    Then the response status code should be 200
-    And the response should be in JSON
-    And the JSON node "_embedded.item" should have 1 element
     When I send the following GraphQL request:
     """
     {
@@ -112,3 +107,27 @@ Feature: Collections filtering
     """
     And the response status code should be 200
     And the JSON node "data.dummies.edges" should have 1 element
+
+  @createSchema
+  Scenario: Retrieve a collection ordered using nested properties
+    Given there are 2 dummy objects with relatedDummy
+    When I send the following GraphQL request:
+    """
+    {
+      dummies(order: {relatedDummy_name: "DESC"}) {
+        edges {
+          node {
+            name
+            relatedDummy {
+              id
+              name
+            }
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.dummies.edges[0].node.name" should be equal to "Dummy #2"
+    And the JSON node "data.dummies.edges[1].node.name" should be equal to "Dummy #1"
--- a/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
@@ -77,14 +77,7 @@ final class CollectionResolverFactory implements ResolverFactoryInterface
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
             $dataProviderContext = $resourceMetadata->getGraphqlAttribute('query', 'normalization_context', [], true);
             $dataProviderContext['attributes'] = $this->fieldsToAttributes($info);
-            $filters = $args;
-            foreach ($filters as $name => $value) {
-                if (strpos($name, '_')) {
-                    // Gives a chance to relations/nested fields
-                    $filters[str_replace('_', '.', $name)] = $value;
-                }
-            }
-            $dataProviderContext['filters'] = $filters;
+            $dataProviderContext['filters'] = $this->getNormalizedFilters($args);
 
             if (isset($rootClass, $source[$rootProperty = $info->fieldName], $source[ItemNormalizer::ITEM_KEY])) {
                 $rootResolvedFields = $this->identifiersExtractor->getIdentifiersFromItem(unserialize($source[ItemNormalizer::ITEM_KEY]));
@@ -170,5 +163,23 @@ final class CollectionResolverFactory implements ResolverFactoryInterface
             'identifiers' => $resolvedIdentifiers,
             'collection' => $isCollection,
         ]);
+    }
+
+    private function getNormalizedFilters(array $args): array
+    {
+        $filters = $args;
+        foreach ($filters as $name => $value) {
+            if (\is_array($value)) {
+                $filters[$name] = $this->getNormalizedFilters($value);
+                continue;
+            }
+
+            if (strpos($name, '_')) {
+                // Gives a chance to relations/nested fields.
+                $filters[str_replace('_', '.', $name)] = $value;
+            }
+        }
+
+        return $filters;
     }
 }
--- a/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
@@ -77,7 +77,14 @@ final class CollectionResolverFactory implements ResolverFactoryInterface
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
             $dataProviderContext = $resourceMetadata->getGraphqlAttribute('query', 'normalization_context', [], true);
             $dataProviderContext['attributes'] = $this->fieldsToAttributes($info);
-            $dataProviderContext['filters'] = $args;
+            $filters = $args;
+            foreach ($filters as $name => $value) {
+                if (strpos($name, '_')) {
+                    // Gives a chance to relations/nested fields
+                    $filters[str_replace('_', '.', $name)] = $value;
+                }
+            }
+            $dataProviderContext['filters'] = $filters;
 
             if (isset($rootClass, $source[$rootProperty = $info->fieldName], $source[ItemNormalizer::ITEM_KEY])) {
                 $rootResolvedFields = $this->identifiersExtractor->getIdentifiersFromItem(unserialize($source[ItemNormalizer::ITEM_KEY]));
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -292,6 +292,13 @@ final class SchemaBuilder implements SchemaBuilderInterface
     private function convertFilterArgsToTypes(array $args): array
     {
         foreach ($args as $key => $value) {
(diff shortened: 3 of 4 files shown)


IonBazan:
Please also add a test for order and date filter to make sure it's ok.

antograssiot:
I will add it for date, ordering on nested property doesn't make much sense but I might be missing the point ?

IonBazan:
Consider fetching posts ordered by author's rating (stars).

alanpoulain:
I don't really like this way. The issue will be too obscure for an user if it used a custom filter with an underscore. I would prefer the other method.

antograssiot:
Sure I figured out just after commenting. this patch will need more work because the keys of value would need to be cleaned when they are arrays and for the key `order`. I'll have a look tonight maybe.

antograssiot:
So should I Spend more time on this to get also the nested filed ordering to work @alanpoulain or should I close it s the other PR is a prefered solution and should be extented for OrderFilter ?

antograssiot:
I've added a naive workaround so ordering on nesting property value would work now.
What's your opinion ?

IonBazan:
I wonder if it's possible to use nested property filters like:
`dummies(relatedDummy: {name: "foo"})` and `dummies(order: {relatedDummy: {name: "foo"}})` because in your solution, filters named: `related_name` and `related.name` are ambiguous.

antograssiot:
@IonBazan I don't know much about the graphQL part yet, but at least the solution is the same for all filters used in graphQL context without changing the way it works for standard Rest API, that's why I went with that.
Whatever the solution will be at the end, it should be properly documented I think...

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -164,4 +167,22 @@ private function getSubresource(string $rootClass, array $rootResolvedFields, ar
             'collection' => $isCollection,
         ]);
     }
+
+    private function getNormalizedFilters($args)
+    {
+        $filters = $args;
+        foreach ($filters as $name => &$value) {
+            if ($name === $this->orderParameterName && \is_array($value)) {
```
alanpoulain: Why not generalizing it for all filters (i.e. only verifying if `$value` is an array)?

antograssiot:
I do understand that for people using GraphQL only, it might be tricky to define filters using the dot notation and use and underscored notation when using it but I don't really see any other better solution for now...

features/graphql/filters.feature
```diff
@@ -88,3 +88,51 @@ Feature: Collections filtering
...
     And the JSON node "data.dummies.edges[2].node.relatedDummies.edges[0].node.name" should be equal to "RelatedDummy13"
+
+  @createSchema
+  Scenario: Retrieve a collection filtered using the related search filter
+    Given there are 1 dummy objects having each 2 relatedDummies
+    And there are 1 dummy objects having each 3 relatedDummies
+    When I add "Accept" header equal to "application/hal+json"
+    And I send a "GET" request to "/dummies?relatedDummies.name=RelatedDummy31"
+    Then the response status code should be 200
+    And the response should be in JSON
```
alanpoulain: This test shouldn't be here.

features/graphql/filters.feature
```diff
@@ -88,3 +88,51 @@ Feature: Collections filtering
...
+    And the response status code should be 200
+    And the JSON node "data.dummies.edges" should have 1 element
+
+  @createSchema
+  Scenario: Retrieve a collection ordered using nested properties
+    Given there are 2 dummy objects with relatedDummy
+    When I send the following GraphQL request:
+    """
+    {
+      dummies(order:{relatedDummy_name: "DESC"}) {
```
alanpoulain: Please add a space after `order:`.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -289,9 +291,12 @@ private function mergeFilterArgs(array $args, array $parsed, ResourceMetadata $r
         return $args;
     }
 
-    private function convertFilterArgsToTypes(array $args): array
+    private function convertFilterArgsToTypes(array $args, bool $normalizeKeys = false): array
     {
         foreach ($args as $key => $value) {
+            if ($normalizeKeys && (strpos($key, '.'))) {
```
alanpoulain: Same, why not generalizing it?

alanpoulain:
Maybe we could use another character than the underscore to avoid confusion?

IonBazan:
@alanpoulain I would go for `dummies?relatedDummy[anotherRelatedDummy][name]=foo` for REST and `dummies(relatedDummy: {anotherRelatedDummy: {name: "foo"}})` for GraphQL because it ensures consistency between the APIs but it is also a BC break.

alanpoulain:
Seems fine to me. I think there is no BC break for now because GraphQL is treated as an experimental feature. It's not really clear enough for end-users though.

alanpoulain:
I think I get it. You want to avoid characters replacement by having the same kind of notation for the REST part and the GraphQL one?

IonBazan:
@alanpoulain Exactly, but I'm not sure if there are any drawbacks of this solution. This would require a total filters refactor and a lot of testing. I think the solution proposed by @antograssiot is good enough for now.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
antograssiot: My first idea was that this is not necessary fo all filters (like DateFilter) for example and wanted to keep the modification as small as possible... but on a second thought, I don't see any issue in generalizing it.

alanpoulain: And it simplifies the code :wink:

features/graphql/filters.feature
```diff
@@ -88,3 +88,50 @@ Feature: Collections filtering
     And the JSON node "data.dummies.edges[1].node.relatedDummies.edges" should have 0 elements
     And the JSON node "data.dummies.edges[2].node.relatedDummies.edges" should have 1 element
     And the JSON node "data.dummies.edges[2].node.relatedDummies.edges[0].node.name" should be equal to "RelatedDummy13"
+
+  @createSchema
+  Scenario: Retrieve a collection filtered using the related search filter
+    Given there are 1 dummy objects having each 2 relatedDummies
+    And there are 1 dummy objects having each 3 relatedDummies
+    When I add "Accept" header equal to "application/hal+json"
```
IonBazan: I think it shouldn't be here as we are testing GraphQL endpoint, not HAL.

antograssiot:
all comments so far have been addressed.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -292,6 +292,9 @@ private function mergeFilterArgs(array $args, array $parsed, ResourceMetadata $r
     private function convertFilterArgsToTypes(array $args): array
     {
         foreach ($args as $key => $value) {
+            if (strpos($key, '.')) {
```
alanpoulain: Add a comment to explain?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -164,4 +164,22 @@ private function getSubresource(string $rootClass, array $rootResolvedFields, ar
             'collection' => $isCollection,
         ]);
     }
+
+    private function getNormalizedFilters($args)
```
alanpoulain: Add ` :array` at the end and `array` for the parameter.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -164,4 +164,22 @@ private function getSubresource(string $rootClass, array $rootResolvedFields, ar
...
+    {
+        $filters = $args;
+        foreach ($filters as $name => &$value) {
```
alanpoulain: Please don't use references.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -164,4 +164,22 @@ private function getSubresource(string $rootClass, array $rootResolvedFields, ar
...
+            if (\is_array($value)) {
+                $filters[$name] = $this->getNormalizedFilters($value);
+                continue;
+            }
+
+            if (strpos($name, '_')) {
+                // Gives a chance to relations/nested fields
```
alanpoulain: Please add a final point.

antograssiot:
done @alanpoulain

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -292,6 +292,11 @@ private function mergeFilterArgs(array $args, array $parsed, ResourceMetadata $r
...
+                // Declare relations/nested fields in a GraphQL compatible syntax.
+                $args[str_replace('.', '_', $key)] = $value;
```
alanpoulain: Is this really working? Because we are not looping with the added value.

dunglas:
Good one, thank you very much @antograssiot. Can you also open a doc PR? It looks important to have this mentioned in the documentation PR (especially for REST actually).

antograssiot:
Sure, I'll prepare a doc update this weekend or next week @dunglas

Issue #1864, opened 2018-04-13
[GraphQL] Use enum type for ORDER filters


IonBazan:
We should use Enum("DESC", "ASC") for Order Filters instead of string.

Simperfit:
@IonBazan Do you want (and have time) to provide the PR ?

Thanks for contributing to API-Platform!

PR #1869, opened 2018-04-15
[GraphQL] Use enum for OrderFilter

IonBazan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1864 
| License       | MIT

This PR changes the way `OrderFilter`'s GraphQL definition is built. It uses `EnumType` instead of `StringType` to define supported values.

### DIFF
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -4,6 +4,25 @@ Feature: Collections filtering
   I need to be able to set filters
 
   @createSchema
+  Scenario: Retrieve a collection orderes using the order filter
+    Given there are 3 dummy objects with dummyDate
+    When I send the following GraphQL request:
+    """
+    {
+      dummies(order: {dummyDate: DESC}) {
+        edges {
+          node {
+            id
+            dummyDate
+          }
+        }
+      }
+    }
+    """
+    Then the JSON node "data.dummies.edges" should have 3 elements
+    And the JSON node "data.dummies.edges[0].node.dummyDate" should be equal to "2015-04-02T00:00:00+00:00"
+
+  @createSchema
   Scenario: Retrieve a collection filtered using the boolean filter
     Given there is 1 dummy object with dummyBoolean true
     And there is 1 dummy object with dummyBoolean false
--- a/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
@@ -112,7 +112,8 @@ class OrderFilter extends AbstractContextAwareFilter
 
             $description[sprintf('%s[%s]', $this->orderParameterName, $property)] = [
                 'property' => $property,
-                'type' => 'string',
+                'type' => 'enum',
+                'values' => ['ASC', 'DESC'],
                 'required' => false,
             ];
         }
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -25,6 +25,7 @@ use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInte
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use ApiPlatform\Core\Util\ClassInfoTrait;
 use Doctrine\Common\Util\Inflector;
+use GraphQL\Type\Definition\EnumType;
 use GraphQL\Type\Definition\InputObjectType;
 use GraphQL\Type\Definition\InterfaceType;
 use GraphQL\Type\Definition\ObjectType;
@@ -226,9 +227,7 @@ final class SchemaBuilder implements SchemaBuilderInterface
                     }
 
                     foreach ($this->filterLocator->get($filterId)->getDescription($resourceClass) as $key => $value) {
-                        $nullable = isset($value['required']) ? !$value['required'] : true;
-                        $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
-                        $graphqlFilterType = $this->convertType($filterType, false, null, $depth);
+                        $graphqlFilterType = $this->getGraphQlFilterType($value, $depth);
 
                         if ('[]' === $newKey = substr($key, -2)) {
                             $key = $newKey;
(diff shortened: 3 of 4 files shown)


src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -267,6 +266,26 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
...
     }
 
+    private function getGraphQlFilterType(array $value, int $depth)
+    {
+        $nullable = isset($value['required']) ? !$value['required'] : true;
+
+        if ('enum' === $value['type']) {
+            if (!isset($this->graphqlTypes['#OrderEnumType'])) {
+                $this->graphqlTypes['#OrderEnumType'] = new EnumType([
+                    'name' => 'OrderEnumType',
```
IonBazan: This is not the most scalable solution. I would be glad for any suggestions how to improve this.

features/graphql/filters.feature
```diff
@@ -3,6 +3,25 @@ Feature: Collections filtering
   As an API consumer
   I need to be able to set filters
 
+  @createSchema
+  Scenario: Retrieve a collection orderes using the order filter
```
alanpoulain: orders

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -267,6 +266,26 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
         return null;
     }
 
+    private function getGraphQlFilterType(array $value, int $depth)
+    {
+        $nullable = isset($value['required']) ? !$value['required'] : true;
+
+        if ('enum' === $value['type']) {
+            if (!isset($this->graphqlTypes['#OrderEnumType'])) {
```
alanpoulain: This could be an enum for another filter than the order one.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -267,6 +266,26 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
...
+                $this->graphqlTypes['#OrderEnumType'] = new EnumType([
```
alanpoulain: This should be in the method `convertType`.

src/Bridge/Doctrine/Orm/Enum.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
+
+interface Enum
```
IonBazan: We could use https://github.com/eloquent/enumeration to make it more elegant but it's another dependency.
WDYT @dunglas @alanpoulain?

alanpoulain: No I don't think it's a good idea. The interfaces are defined by us (we are the ones giving a contract).

src/Bridge/Doctrine/Orm/Enum.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
```
alanpoulain: This interface shouldn't be here. Maybe in the `Api` folder?

src/Bridge/Doctrine/Orm/OrderDirection.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
```
alanpoulain: Shouldn't be here either.

src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -112,7 +113,7 @@ public function getDescription(string $resourceClass): array
 
             $description[sprintf('%s[%s]', $this->orderParameterName, $property)] = [
                 'property' => $property,
-                'type' => 'string',
+                'type' => OrderDirection::class,
```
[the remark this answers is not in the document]
IonBazan: It's up to filter to perform validation. This change should not affect REST at all.

src/Api/Enum.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Api;
+
+interface Enum
```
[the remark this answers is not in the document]
IonBazan: This interface may be used in filters to validate the input and is not specific to GraphQL.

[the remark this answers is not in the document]
alanpoulain: :+1: for renaming the interface.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -358,6 +360,19 @@ private function convertType(Type $type, bool $input = false, string $mutationNa
                     break;
                 }
 
+                if (is_a($typeClass = $type->getClassName(), Enum::class, true)) {
+                    $typeName = $typeClass::getName();
+                    if (!isset($this->graphqlTypes['#'.$typeName])) {
```
alanpoulain: You should use a variable for `'#'.$typeName`.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -358,6 +360,19 @@ private function convertType(Type $type, bool $input = false, string $mutationNa
                     break;
                 }
 
+                if (is_a($typeClass = $type->getClassName(), Enum::class, true)) {
+                    $typeName = $typeClass::getName();
```
alanpoulain: Maybe you should add: `/** @var Enum $typeClass */` before to allow IDEs to understand and autocomplete the calls.

IonBazan:
@antograssiot @alanpoulain I've added changes following your comments.

alanpoulain:
I think you're right. Maybe you should use another field.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -358,6 +360,21 @@ private function convertType(Type $type, bool $input = false, string $mutationNa
...
+                    $graphqlTypeName = '#'.$typeName;
+                    if (!isset($this->graphqlTypes[$graphqlTypeName])) {
+                        $this->graphqlTypes[$graphqlTypeName] = new EnumType([
+                            'name' => $typeName,
+                            'values' => $typeClass::getValues(),
+                        ]);
+                    }
+
+                    $graphqlType = $this->graphqlTypes['#'.$typeName];
```
dunglas: You can reuse `$graphqlTypeName`

src/Api/EnumInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Api;
+
+interface EnumInterface
```
dunglas: What do you think about using https://github.com/myclabs/php-enum instead? We already use it in the schema generator.

src/Api/EnumInterface.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    public static function getValues(): array;
```
dunglas: Can we at least get rid of the `static`?

IonBazan: Creating an instance of an Enum class is not a good idea.

src/Api/EnumInterface.php
IonBazan: We don't use it yet but we would still need to use this interface unless we have an idea to generate the name for enum without calling `getName()` method.

flug:
@IonBazan Up is this pull request still valid?

PR #1876, opened 2018-04-18
Add missing doctrine alias removed in doctrine-bundle@1.9

soyuka:


antograssiot:
Is declaring the removed alias in core an option instead @soyuka ?

PR #1877, opened 2018-04-19
Allow subresource items in the iri converter

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | depends
| New feature?  | depends
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | 

This enables subresources in the IRIConverter so that subresources can get denormalized as well.
I thought that it'd be better to use the ReadListener (avoid code duplication) instead of injecting another data provider. Let me know if you'd prefer the later (it's why I didn't added any tests yet).

-----------------------

Done:
- Move the ReadListener code to a Trait (OperationDataProviderTrait)
- Initialize a new OperationDataProvider class (uses the trait, used only in IriConverter, has a `getDataFromRouteParameters` method)
- Refactor IriConverterTests (removed duplication code)
- ReadListener uses the trait as well
- Add behat subresource denormalizeRelation test


src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -45,13 +46,15 @@
     private $router;
     private $identifiersExtractor;
     private $identifierDenormalizer;
+    private $readListener;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null, IdentifiersExtractorInterface $identifiersExtractor = null, ChainIdentifierDenormalizer $identifierDenormalizer = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ItemDataProviderInterface $itemDataProvider = null, RouteNameResolverInterface $routeNameResolver, RouterInterface $router, PropertyAccessorInterface $propertyAccessor = null, IdentifiersExtractorInterface $identifiersExtractor = null, ChainIdentifierDenormalizer $identifierDenormalizer = null, ReadListener $readListener = null)
```
dunglas: It introduces a dependency to a Symfony full stack related class in the low level infrastructure. I don't think it's a good idea, it will be very painful to support other frameworks (or PSR-7) in the future.

soyuka: Okay WDYT about this:

1. Move private methods to a trait
2. Introduce a new class (`OperationDataProvider` that uses the trait)
3. Inject the `OperationDataProvider` in the IriConverter so that it can call item and subresource operations

This way: 
- no changes in read listener
- de-duplicate code in IriConverter
- improves future updates in the IriConverter concerning data retrieval (IriToItem)

src/DataProvider/OperationDataProvider.php
```diff
@@ -0,0 +1,67 @@
...
+namespace ApiPlatform\Core\DataProvider;
+
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\ItemNotFoundException;
+use ApiPlatform\Core\Identifier\Normalizer\ChainIdentifierDenormalizer;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class OperationDataProvider
```
dunglas: Shouldn't we introduce an interface?

src/DataProvider/OperationDataProviderTrait.php
```diff
@@ -0,0 +1,128 @@
...
+                $context = [ChainIdentifierDenormalizer::HAS_IDENTIFIER_DENORMALIZER => true];
+            }
+
+            $data = $this->itemDataProvider->getItem($attributes['resource_class'], $id, $attributes['item_operation_name'], $context);
+        } catch (InvalidIdentifierException $e) {
+            $data = null;
+        }
+
+        if (null === $data) {
+            throw new NotFoundHttpException('Not Found');
```
dunglas: We shouldn't throw a HTTP error here, but a domain exception.

src/DataProvider/OperationDataProviderTrait.php
```diff
@@ -0,0 +1,128 @@
...
+     * @var ChainIdentifierDenormalizer
+     */
+    private $identifierDenormalizer;
+
+    /**
+     * Retrieves data for a collection operation.
+     *
+     * @return array|\Traversable|null
+     */
+    private function getCollectionData(Request $request, array $attributes, array $context)
```
dunglas: Don't you think we can avoid to depend of the request, but extract the relevant data from the controller/event listener and pass it through the context? It will be easier to support PSR-7 this way.

src/DataProvider/OperationDataProvider.php
soyuka: We may but I'm also to mark this as `@internal` unless we can find a global use for it? For now it'll be only used by the IriConverter and I really doubt it'll be used for anything else :p. 
I just want to avoid injecting everything the OperationDataProviderTrait depends on in the IriConverter.

src/DataProvider/OperationDataProviderTrait.php
soyuka: totally should be in ReadListener only

src/DataProvider/OperationDataProviderTrait.php
soyuka: Yes we can totally do this will be cleaner.

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -58,7 +58,7 @@
             <argument type="service" id="api_platform.router" />
             <argument type="service" id="api_platform.property_accessor" />
             <argument type="service" id="api_platform.identifiers_extractor.cached" />
-            <argument type="service" id="api_platform.identifier.denormalizer" />
+            <argument type="service" id="api_platform.operation_data_provider" />
```
soyuka: Not released yet == not BC break.

src/Bridge/Symfony/Bundle/Resources/config/data_provider.xml
```diff
@@ -26,6 +26,13 @@
             <argument type="tagged" tag="api_platform.subresource_data_provider" />
         </service>
         <service id="ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface" alias="api_platform.subresource_data_provider" />
+
+        <service id="api_platform.operation_data_provider" class="ApiPlatform\Core\DataProvider\OperationDataProvider">
```
dunglas: `public="false"`?

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -44,24 +44,23 @@
...
         if (null === $identifiersExtractor) {
             @trigger_error('Not injecting ItemIdentifiersExtractor is deprecated since API Platform 2.1 and will not be possible anymore in API Platform 3', E_USER_DEPRECATED);
             $this->identifiersExtractor = new IdentifiersExtractor($propertyNameCollectionFactory, $propertyMetadataFactory, $propertyAccessor ?? PropertyAccess::createPropertyAccessor());
-        } else {
-            $this->identifiersExtractor = $identifiersExtractor;
         }
 
-        if (null === $identifierDenormalizer) {
-            @trigger_error(sprintf('Not injecting "%s" is deprecated since API Platform 2.2 and will not be possible anymore in API Platform 3.', ChainIdentifierDenormalizer::class), E_USER_DEPRECATED);
+        if (null === $operationDataProvider) {
```
dunglas: It will make very difficult to use API Platform without a framework or with another framework than Symfony. For instance, if I want to use API Platform's classes without a framework, and I don't bother to support subresources (I've a project like that), I'm stuck.

src/DataProvider/OperationDataProvider.php
```diff
@@ -0,0 +1,51 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataProvider;
+
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Identifier\Normalizer\ChainIdentifierDenormalizer;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+
+/**
+ * @internal
```
dunglas: Has it's a mandatory dependency of `IriConverter`, and there is no related interface, it cannot be `@internal`.

dunglas: Also, please add some PHPDoc.

src/DataProvider/OperationDataProvider.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+class OperationDataProvider
```
dunglas: `final`?

src/DataProvider/OperationDataProvider.php
```diff
@@ -0,0 +1,51 @@
...
+
+    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider, ChainIdentifierDenormalizer $identifierDenormalizer)
+    {
+        $this->collectionDataProvider = $collectionDataProvider;
+        $this->itemDataProvider = $itemDataProvider;
+        $this->subresourceDataProvider = $subresourceDataProvider;
+        $this->identifierDenormalizer = $identifierDenormalizer;
+    }
+
+    public function getDataFromRouteParameters(string $iri = null, array $parameters, array $context = [])
```
dunglas: Why is `$iri` nullable?

src/Util/RequestAttributesExtractor.php
```diff
@@ -39,9 +39,14 @@ private function __construct()
      */
     public static function extractAttributes(Request $request)
     {
-        $result = ['resource_class' => $request->attributes->get('_api_resource_class')];
+        return self::extractAttributesFromParameters($request->attributes->all());
+    }
+
+    public static function extractAttributesFromParameters(array $attributes)
```
dunglas: I would introduce a new class, as it's not tight with the `Request` nor Symfony anymore.

src/DataProvider/OperationDataProviderTrait.php
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataProvider;
+
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use ApiPlatform\Core\Identifier\Normalizer\ChainIdentifierDenormalizer;
+
+trait OperationDataProviderTrait
```
dunglas: `@internal`?

src/DataProvider/OperationDataProviderTrait.php
```diff
@@ -0,0 +1,106 @@
...
+{
+    /**
+     * @var CollectionDataProviderInterface
+     */
+    private $collectionDataProvider;
```
dunglas: Please add a blank line between props (btw, it's often better to have traits without props).

src/DataProvider/OperationDataProviderTrait.php
```diff
@@ -0,0 +1,106 @@
...
+    private $subresourceDataProvider;
+    /**
+     * @var ChainIdentifierDenormalizer
+     */
+    private $identifierDenormalizer;
+
+    /**
+     * Retrieves data for a collection operation.
+     *
+     * @return array|\Traversable|null
```
dunglas: `iterable|null`?

src/EventListener/ReadListener.php
```diff
@@ -33,24 +34,21 @@
...
 
     public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, ChainIdentifierDenormalizer $identifierDenormalizer = null)
     {
         $this->collectionDataProvider = $collectionDataProvider;
         $this->itemDataProvider = $itemDataProvider;
         $this->subresourceDataProvider = $subresourceDataProvider;
         $this->serializerContextBuilder = $serializerContextBuilder;
+        $this->identifierDenormalizer = $identifierDenormalizer;
 
         if (null === $identifierDenormalizer) {
```
dunglas: Same here, I would not throw a deprecation if it's not provided. If you don't inject it, you'll just loose support for complex identifiers (not a big deal in many projects).

src/DataProvider/OperationDataProvider.php
dunglas: I would just get rid of this class, it introduces unneeded complexity and indirection. Can we not include the trait directly in the data provider? It looks cleaner to me (and allows to keep subresources support optional).

src/Util/RequestAttributesExtractor.php
dunglas: Also, please copy/adapt the PHPDoc.

dunglas:
It's a big change for a patch release. I would be more confortable to revert the previous fix and merge the new system in `master`. WDYT @soyuka?

soyuka:
> It's a big change for a patch release. I would be more confortable to revert the previous fix and merge the new system in master. WDYT @soyuka?

I removed some deprecations and the added code will work without subresources and without identifier denormalizer. It's really your call though, I don't mind! Still, this may be a big improvement for those who want to use subresources in the denormalization process (closes a bunch of issues)!

Issue #1880, opened 2018-04-21
GraphQL broke after some commit


KriBetko:
Hi. I'm wanted to check new feature for GraphQL with filters for nested property (#1868) but if i'm require on composer this commit (c1395866e8b564d27d51c9adeff1183ccc0575f9) i got "500 Internal Server Error" for any POST request on "api_graphql_entrypoint" without errors on log or profiler. 
I checked and this occurs after 2e0c64769e2472cf3bc50d19d2614b5caab0dcd0 commit (on previous b54979b9464cfa5c14022959a374fe280e619dd9 commit all is OK).

I'm not sure that I understand everything correctly, so I ask for help in solving this problem

antograssiot:
This PR was only merge in 2.2 branch but not in master. Are you sure you where not using the master branche before ?

antograssiot:
So I've been digging quickly into it and the addition on filtering for nested entity is not the root cause, it seems that introspection is broken since `InputUnionType` usage addition in the SchemaBuilder https://github.com/api-platform/core/compare/v2.2.5...d805e9a#diff-b7cf573372ef33efd3ea514a70783519L344

There are discussions already in https://github.com/api-platform/core/pull/1858 and they are trying to fix this upstream

alanpoulain:
The issue is only with introspection and it shouldn't affect the requests.
I was hoping a quick fix in graphql-php but it doesn't seem to happen soon so I will propose a dirty way to fix it soon.

PR #1886, opened 2018-04-23
[GraphQL] Fix introspection (graphql-php / GraphiQL)

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1880
| License       | MIT
| Doc PR        |

The best way to fix current issues with GraphQL introspection is to remove the problematic code.

We should be waiting for this PR to be merged https://github.com/facebook/graphql/pull/395 and changes to be affected to graphql-php and graphql-js in order to add it again.

The main lost feature is the ability to manage embedded entities.

A way to get this feature back will be to have two fields (one for an IRI, the other for the data) for relations but there is some work to be done.


alanpoulain:
This is done.

Simperfit:
Thanks @alanpoulain.

BernardA:
@alanpoulain, please inform current status concerning embedded mutations. I have a bid entity which embeds a message entity, which embeds an attachment entity. As it is now, it seems that I need to create 3 create mutations. Is that still the case? Thanks.

alanpoulain:
Yes it is: https://github.com/api-platform/core/issues/3258#issuecomment-553971707.

PR #1891, opened 2018-04-25
add a specific "documentation_formats" that may differ from the API "…

### LINKED ISSUE #1854: Add a `documentation_formats` that differs from `formats`
If you de-activate the `json` format from the configuration, you can not use like [ReDoc](https://rebilly.github.io/ReDoc/). You can not use the integrated swagger if you disable `html` format.

Personally, I disable everything except `json-ld` (less options means less bugs ;) ).

Furthermore, I don't think that enabling html format is really pertinent for an API, but it does for the documentation.

Do you think it's a good idea to split `formats` from the `documentation_formats`, this way, we could do something like:

```yaml
api_platform:
    formats: 
        jsonld: ['application/ld+json']
    documentation_formats: 
        html: ['text/html']
        json: ['application/json']
```

As usual, I will open a PR if you agree with this.

jdeniau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | See descrption
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1854 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/475

This PR allow to disable the "json" format from the api and only allow jsonld for example, without breaking the swagger documentation.

About the deprecation though: I removed "json" and "html" from the formats, to keep only "jsonld", and moved "json" and "html" to the documentation formats only.
Technically, it's a compatibility break, but as often can be seen as a bugfix (at least for the `html` format). 

If you like, I can add the `json` format back, but as I see it, api-platform default is really nicely coupled with json-jd, so I assumed that the `json` format it was for documentation only.

### DIFF
--- a/features/main/content_negotiation.feature
+++ b/features/main/content_negotiation.feature
@@ -110,8 +110,8 @@ Feature: Content Negotiation support
     Then the response status code should be 406
     And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
 
-  Scenario: If the request format is HTML, the error should be in HTML
-    When I add "Accept" header equal to "text/html"
+  Scenario: If the request format is XML, the error should be in XML
+    When I add "Accept" header equal to "application/xml"
     And I send a "GET" request to "/dummies/666"
     Then the response status code should be 404
-    And the header "Content-Type" should be equal to "text/html; charset=utf-8"
+    And the header "Content-Type" should be equal to "application/xml; charset=utf-8"
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -222,6 +222,9 @@ final class Configuration implements ConfigurationInterface
 
         $this->addFormatSection($rootNode, 'formats', [
             'jsonld' => ['mime_types' => ['application/ld+json']],
+        ]);
+        $this->addFormatSection($rootNode, 'documentation_formats', [
+            'jsonld' => ['mime_types' => ['application/ld+json']], // Hydra documentation, admin and PWA generator
             'json' => ['mime_types' => ['application/json']], // Swagger support
             'html' => ['mime_types' => ['text/html']], // Swagger UI support
         ]);
--- a/src/Bridge/Symfony/Bundle/EventListener/SwaggerUiListener.php
+++ b/src/Bridge/Symfony/Bundle/EventListener/SwaggerUiListener.php
@@ -34,17 +34,4 @@ final class SwaggerUiListener
 
         $request->attributes->set('_controller', 'api_platform.swagger.action.ui');
     }
-
-    public function addEndpointTypeToRequest(GetResponseEvent $event)
-    {
-        $request = $event->getRequest();
-        if (
-            'html' !== $request->getRequestFormat('') ||
-            (!$request->attributes->has('_api_resource_class') && !$request->attributes->has('_api_respond'))
-        ) {
-            return;
-        }
-
-        $request->attributes->set('_api_endpoint_type', 'documentation');
-    }
 }
--- a/src/Bridge/Symfony/Bundle/EventListener/SwaggerUiListener.php
+++ b/src/Bridge/Symfony/Bundle/EventListener/SwaggerUiListener.php
@@ -34,4 +34,17 @@ final class SwaggerUiListener
 
         $request->attributes->set('_controller', 'api_platform.swagger.action.ui');
     }
+
+    public function addEndpointTypeToRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+        if (
+            'html' !== $request->getRequestFormat('') ||
+            (!$request->attributes->has('_api_resource_class') && !$request->attributes->has('_api_respond'))
+        ) {
+            return;
+        }
+
+        $request->attributes->set('_api_endpoint_type', 'documentation');
+    }
 }
--- a/src/Bridge/Symfony/Bundle/Resources/config/swagger-ui.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/swagger-ui.xml
@@ -7,7 +7,8 @@
     <services>
 
         <service id="api_platform.swagger.listener.ui" class="ApiPlatform\Core\Bridge\Symfony\Bundle\EventListener\SwaggerUiListener">
-            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" />
+            <tag name="kernel.event_listener" event="kernel.request" method="addEndpointTypeToRequest" priority="8" />
+            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="0" />
         </service>
 
     </services>
--- a/src/Bridge/Symfony/Bundle/Resources/config/swagger-ui.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/swagger-ui.xml
@@ -7,8 +7,7 @@
     <services>
 
         <service id="api_platform.swagger.listener.ui" class="ApiPlatform\Core\Bridge\Symfony\Bundle\EventListener\SwaggerUiListener">
-            <tag name="kernel.event_listener" event="kernel.request" method="addEndpointTypeToRequest" priority="8" />
-            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="0" />
+            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" />
         </service>
 
     </services>
--- a/src/EventListener/AddFormatListener.php
+++ b/src/EventListener/AddFormatListener.php
@@ -28,12 +28,14 @@ final class AddFormatListener
 {
     private $negotiator;
     private $formats;
+    private $documentationFormats;
     private $mimeTypes;
 
-    public function __construct(Negotiator $negotiator, array $formats)
+    public function __construct(Negotiator $negotiator, array $formats, array $documentationFormats = [])
     {
         $this->negotiator = $negotiator;
         $this->formats = $formats;
+        $this->documentationFormats = $documentationFormats;
     }
 
     /**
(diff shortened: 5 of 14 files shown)


src/EventListener/AddFormatListener.php
```diff
@@ -28,12 +28,14 @@
 {
     private $negotiator;
     private $formats;
+    private $documentationFormats;
     private $mimeTypes;
 
-    public function __construct(Negotiator $negotiator, array $formats)
+    public function __construct(Negotiator $negotiator, array $formats, array $documentationFormats)
```
dunglas: It must have a default value (`= []`), or it's BC break.

src/EventListener/AddFormatListener.php
```diff
@@ -145,4 +151,17 @@ private function getNotAcceptableHttpException(string $accept, array $mimeTypes
             implode('", "', $mimeTypes)
         ));
     }
+
+    private function getRequestFormats(Request $request): array
+    {
+        return $this->isDocumentation($request)
```
dunglas: Can you inline this please?

src/EventListener/AddFormatListener.php
```diff
@@ -145,4 +151,17 @@ private function getNotAcceptableHttpException(string $accept, array $mimeTypes
...
+            ? $this->documentationFormats
+            : $this->formats
+        ;
+    }
+
+    private function isDocumentation(Request $request): bool
```
dunglas: Can you remove this function? (It's in the hot path, and function calls are costly).

src/EventListener/AddFormatListener.php
```diff
@@ -145,4 +151,17 @@ private function getNotAcceptableHttpException(string $accept, array $mimeTypes
...
+    {
+        return 'api_platform.action.documentation' === $request->attributes->get('_controller');
```
dunglas: Alternative proposal:

set a `_api_documentation` attribute here https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Resources/config/routing/docs.xml#L10, and test for the presence of this flag. It would be more flexible and give an extension point.

jdeniau: @dunglas What you think about a `_api_endpoint_type` set to `documentation` instead ? It may avoid anyone to add another variable later.

dunglas: I usually prefer boolean flags (almost all others are boolean). What other type can it be?

src/Bridge/Symfony/Bundle/EventListener/SwaggerUiListener.php
```diff
@@ -34,4 +34,10 @@ public function onKernelRequest(GetResponseEvent $event)
 
         $request->attributes->set('_controller', 'api_platform.swagger.action.ui');
     }
+
+    public function addEndpointTypeToRequest(GetResponseEvent $event)
```
dunglas: Is it necessary? The flag will always be set for every request even if it's not caught by Swagger UI isn't it?

jdeniau:
@dunglas done, I just changed `_api_documentation` by `_api_endpoint_type` set to `documentation`.

For the circleci test, I think it's a problem with circleci configuration

src/Bridge/Symfony/Bundle/EventListener/SwaggerUiListener.php
jdeniau: Oh, I need to set the same check as `onKernelRequest` function.

I need to separate the two listener because the AddFormatListener come between priority 0 and 7

src/EventListener/AddFormatListener.php
jdeniau: may something related to graphql :man_shrugging:

jdeniau:
@dunglas I'm kind of stuck here with a circular exception on the SwaggerUiListener, maybe you have an idea :

The SwaggerUiListener depends on the `AddFormatListener` for the `text/html` request format to be set.

The AddFormatListener need to know if we are on a documentation page or not, and thus needs to have the documentation attribute set.

The only solution I see here is to dissociate the AddFormatListener and something that throw the exception but it seems really overkill for such a small feature, don't yout think ?

dunglas:
I would handle Swagger UI in two different ways:

* For the `/docs` URL: handle it using `documentation_formats` as done in this PR
* For all others URL, consider that HTML is **not** a "documentation" format (because it is not really, it allows to serialize the resource itself as HTML). So there are no need to mark all those routes as "documentation" (they are not, they are resources routes). I would just let the hijacking mechanism working as it is currently, except that I suggest to disable the hijacking if Swagger is not present in `documentation_formats`. To clarify: disable the HTML format if Swagger is not enabled in the DI extension.

WDYT?

dunglas:
Any news on this one?

jdeniau:
Sorry, I was on holidays ;)

I will check your return soon.

dunglas:
Any news :)

jdeniau:
@dunglas should be OK.

I removed the weird listener that I previously added.

As it is very internal functionality, I am not 100% sure of what I have done though ;)

src/Bridge/Symfony/Bundle/Resources/config/swagger-ui.xml
```diff
@@ -7,7 +7,7 @@
     <services>
 
         <service id="api_platform.swagger.listener.ui" class="ApiPlatform\Core\Bridge\Symfony\Bundle\EventListener\SwaggerUiListener">
-            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" />
```
dunglas: Why this change? 0 is the default priority.

jdeniau:
Still an issue due to changes I made in the available formats

jdeniau:
@dunglas everything is good

jdeniau:
My turn to ping you @dunglas ;)

Fixes #1854

tests/EventListener/AddFormatListenerTest.php
```diff
@@ -32,7 +32,7 @@ public function testNoResourceClass()
         $eventProphecy->getRequest()->willReturn($request)->shouldBeCalled();
         $event = $eventProphecy->reveal();
 
-        $listener = new AddFormatListener(new Negotiator(), ['notexist' => 'application/vnd.notexist']);
+        $listener = new AddFormatListener(new Negotiator(), ['notexist' => 'application/vnd.notexist'], []);
```
soyuka: As the new array argument has a default value we don't really need to change this right?

features/main/content_negotiation.feature
```diff
@@ -110,8 +110,8 @@ Feature: Content Negotiation support
     Then the response status code should be 406
     And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
 
-  Scenario: If the request format is HTML, the error should be in HTML
```
dunglas: Can you keep this text (and duplicate it for XML if appropriate)? It checks that the documentation for any URL feature works (for instance when you open `http://localhost/books/1`, Swagger UI shows up and automatically runs the AJAX request).

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -222,6 +222,9 @@ public function getConfigTreeBuilder()
 
         $this->addFormatSection($rootNode, 'formats', [
             'jsonld' => ['mime_types' => ['application/ld+json']],
+        ]);
+        $this->addFormatSection($rootNode, 'documentation_formats', [
+            'jsonld' => ['mime_types' => ['application/ld+json']], // Hydra documentation, admin and PWA generator
             'json' => ['mime_types' => ['application/json']], // Swagger support
             'html' => ['mime_types' => ['text/html']], // Swagger UI support
```
dunglas: HTML must be defined also in `addFormatSection` section to have the "documentation for any URL" feature working.

src/EventListener/AddFormatListener.php
```diff
@@ -145,4 +151,16 @@ private function getNotAcceptableHttpException(string $accept, array $mimeTypes
             implode('", "', $mimeTypes)
         ));
     }
+
+    /**
+     * Retrieves the list of accepted format in the request depending on some parameters.
+     */
+    private function getRequestAcceptedFormats(Request $request): array
+    {
+        if (!$request->attributes->has('_api_endpoint_type')) {
```
dunglas: you can call `get` directly and store the result in a variable to prevent an extra function call. (get will return `null` if the attribute is not defined)

src/EventListener/AddFormatListener.php
```diff
@@ -51,13 +53,15 @@ public function onKernelRequest(GetResponseEvent $event)
             return;
         }
 
-        $this->populateMimeTypes();
-        $this->addRequestFormats($request, $this->formats);
+        $requestAcceptedFormats = $this->getRequestAcceptedFormats($request);
```
dunglas: Actually, this private function can be replaced by `'documentation' === $request->attributes->get('_api_endpoint_type') ? $this->documentationFormats : $this->formats`. It removes 2 function calls in the hot path.

src/EventListener/AddFormatListener.php
```diff
@@ -84,11 +88,11 @@ public function onKernelRequest(GetResponseEvent $event)
                 return;
             }
 
-            throw $this->getNotAcceptableHttpException($mimeType);
+            throw $this->getNotAcceptableHttpException($mimeType ?? $requestFormat);
```
dunglas: `$requestAcceptedFormats` right? Thus should be covered by a test (because the error isn't caught now)

jdeniau:
I'm kinda lost here because of #1798 : it added a format resolver but does not include error format. 

Does documentation format should be in `FormatsProvider` ? What do you thing @antograssiot ?

antograssiot:
Indeed there is no ErrorFormatsProvider (yet ?).
I can add it if there is a need, it should be pretty easy, and I guess that if you feel that document format needs in some cases to be specified at resource/operation level, you could use a DocumentFormatsProvider or something similar but it seems less likely to be used IMO.

Let's see maintainers opinion.

dunglas:
@jdeniau, do you need any help to finish this PR?

jdeniau:
Actually I think I do yes!
I am kind of lost here, because I don't know api-platform internals as you guys do. 
Maybe I should look back at #1798 and try harder but any help would be appreciated!

flug:
@jdeniau Up is this pull request still valid?

jdeniau:
I did not manage to fix this and have no motivation on that now. I am closing this one.

PR #1894, opened 2018-04-26
fix normalizer cache key generation

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Cache keys were being regenerated even if they were already present. this disabled caching when serializing relations which have the same type as the parent.

In this graph, i am serializing resources which have 3 levels of the same type: Foo->Foo->Foo

https://blackfire.io/profiles/compare/e4646b66-8700-4ac9-b36e-9de700dc5b03/graph


src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -99,8 +102,11 @@ protected function getAttributes($object, $format = null, array $context)
      */
     private function getComponents($object, string $format = null, array $context)
     {
-        if (false !== $context['cache_key'] && isset($this->componentsCache[$context['cache_key']])) {
-            return $this->componentsCache[$context['cache_key']];
+        $class = \get_class($object);
+        $cacheKey = $class.'-'.$context['cache_key'];
```
Simperfit: $class seems used only one time

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -221,8 +223,11 @@ protected function isAllowedAttribute($classOrObject, $attribute, $format = null
      */
     private function getComponents($object, string $format = null, array $context)
     {
-        if (isset($this->componentsCache[$context['cache_key']])) {
-            return $this->componentsCache[$context['cache_key']];
+        $class = \get_class($object);
```
Simperfit: $class seems used only one time

bendavies: 👍

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -99,8 +102,10 @@ protected function getAttributes($object, $format = null, array $context)
      */
     private function getComponents($object, string $format = null, array $context)
     {
-        if (false !== $context['cache_key'] && isset($this->componentsCache[$context['cache_key']])) {
-            return $this->componentsCache[$context['cache_key']];
+        $cacheKey = \get_class($object).'-'.$context['cache_key'];
```
soyuka: We know the object class in the `normalize` method right? Could we pass the information to this function to avoid calling `get_class`? Might be a micro-optimization though. (same below)

dunglas: `\get_class` is transformed in opcode, it's almost costless.

bendavies:
we good here?

dunglas:
Thanks @bendavies. All your work regarding performance is very appreciated!

src/Hal/Serializer/ItemNormalizer.php
bendavies: For info I've copied the method used in core symfony, which is to add the class here

PR #1897, opened 2018-04-26
EagerLoadingExtension: respect fetchEager=false on an association

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

`forceEager`/`fetchEager` does not already result in a performance boost, so being able to disable it per association is crucial.

This:
```
@ApiProperty(attributes={"fetchEager": false})
```
Was not being respected, and ignored if the association was `readable` or a `readableLink`.

@dunglas I don't really understand the `if` statement that this PR fixes, specifically the variable `$inAttributes`. 
You introduced it here: https://github.com/api-platform/core/pull/1552. I'd be grateful if you could check this over and explain the `if`?

Thanks


src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -170,14 +170,13 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
                 $inAttributes = null;
             }
 
+            $fetchEager = $propertyMetadata->getAttribute('fetchEager');
+            if (false === $fetchEager) {
```
antograssiot: Can be inlined

PR #1820, opened 2018-04-06
identifier normalizer code review

soyuka:
ping @meyerbaptiste

PR #1899, opened 2018-04-27
[Bug] Fix integer Identifiers denormalization

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

Right now the Identifier Denormalizer will automatically result in the identifier being casted to string.
This basicaly break the search filter on related entities with a strict typehint on the `getId()` because when getting references to the object in the [getItemFromIri()](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php#L374) the call to `getId()` will result in a TypeError.

ping @soyuka

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -733,6 +733,7 @@ final class FeatureContext implements Context, SnippetAcceptingContext
         $relatedDummy2->setName('RelatedDummy without friends');
         $this->manager->persist($relatedDummy2);
         $this->manager->flush();
+        $this->manager->clear();
     }
 
     /**
--- a/features/doctrine/search_filter.feature
+++ b/features/doctrine/search_filter.feature
@@ -10,6 +10,7 @@ Feature: Search filter on collections
     And I send a "GET" request to "/related_dummies?relatedToDummyFriend.dummyFriend=/dummy_friends/4"
     Then the response status code should be 200
     And the JSON node "_embedded.item" should have 1 element
+    And the JSON node "_embedded.item[0].id" should be equal to the number 1
     And the JSON node "_embedded.item[0]._links.relatedToDummyFriend" should have 4 elements
     And the JSON node "_embedded.item[0]._embedded.relatedToDummyFriend" should have 4 elements
 
--- a/features/main/non_resource.feature
+++ b/features/main/non_resource.feature
@@ -11,7 +11,7 @@ Feature: Non-resources handling
         "@context": "/contexts/ContainNonResource",
         "@id": "/contain_non_resources/1",
         "@type": "ContainNonResource",
-        "id": "1",
+        "id": 1,
         "nested": {
             "@id": "/contain_non_resources/1-nested",
             "@type": "ContainNonResource",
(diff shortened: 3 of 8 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
         "behatch/contexts": "^3.1",
         "doctrine/annotations": "^1.2",
         "doctrine/doctrine-bundle": "^1.8",
-        "doctrine/orm": "^2.5.2",
+        "doctrine/orm": "^2.6.0",
```
antograssiot: https://github.com/doctrine/doctrine2/issues/4884

soyuka:
Can we introduce a DoctrineIdenfitierDenormalizer instead? The string is because the identifier comes from the `Request` and type is unknown at that time.

tests/Fixtures/TestBundle/Entity/ContainNonResource.php
```diff
@@ -30,7 +30,7 @@ class ContainNonResource
     /**
      * @var mixed
      *
-     * @ORM\Column(type="integer")
+     * @ORM\Column(type="string")
```
soyuka: I don't get the change here

antograssiot: I updated the test result instead

antograssiot:
> Can we introduce a DoctrineIdenfitierDenormalizer

I can do that. But this denormalier will still only ast string to int so would be a  `DoctrineIntegerTypeDenormalizer` ? it feels a bit heavy but will allow customization, is it your concern @soyuka ?

antograssiot:
I've update the code @soyuka.
I've no clue on how to please the travis build with low-dependencies as I hit a bug with doctrine. I could be to revert the change done on the TypeHint of DummyFriend, but that was the easiest way to have a failing behat scenario...

antograssiot:
Circle-ci is drunk and waiting for config from 2.2 branch, travis should be good

soyuka:
I think that this should be done on 2.2 because it's a bug fix. May you rebase?

antograssiot:
@soyuka rebased and target branch updated. I failled to see when this got into 2.2, I was pretty sure your initial MRs (#1820) were against into master sorry.

soyuka:
thanks @antograssiot

sroze:
This definitely should have been based on master :)

soyuka:
@antograssiot could you reopen a merge request on master with this patch on master please? Really sorry for the mistake, we're all humans I guess :smile:

soyuka:
Actually, just restore your branch with the github button so that I'm sure we have the patch on your remote and I'll take care of it :).

antograssiot:
I'll have a look tonight

antograssiot:
I restored the branch, I'll rebase it onto master and open a new PR, is that ok ?

teohhanhui:
Yes, it's necessary to open a new PR because the merge functionality is gone for an already merged PR.

soyuka:
https://github.com/api-platform/core/commit/1ea6f68c82e77c4fba0ec8cd2dc9bc78e5aae261 everything back to normal :)

PR #1905, opened 2018-04-30
Simplify boolean filter

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

no need for extra query builder named parameters here. just use `$expr->literal(...)`

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/BooleanFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/BooleanFilter.php
@@ -96,11 +96,9 @@ class BooleanFilter extends AbstractContextAwareFilter
             list($alias, $field) = $this->addJoinsForNestedProperty($property, $alias, $queryBuilder, $queryNameGenerator, $resourceClass);
         }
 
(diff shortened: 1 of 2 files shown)


teohhanhui:
This has implications for the DBMS prepared statement, no? I don't see why we'd want to do this, to be honest. Does it give a sizeable performance gain?

bendavies:
The reason i did this because
1. there is no need for the parameter.
2. the sql produced by symfonys profiler is invalid before this change, (at least for postgres - it produces `where field = 1/0` which is wrong)

teohhanhui:
That's a Symfony profiler bug.

soyuka:
Isn't doctrine storing booleans as integers anyway?

teohhanhui:
> Isn't doctrine storing booleans as integers anyway?

No. Not on PostgreSQL at least. The integers 0 and 1 are not valid boolean values: https://www.postgresql.org/docs/9.1/static/datatype-boolean.html

PR #1907, opened 2018-04-30
Revert "Simplify boolean filter"

teohhanhui:
Reverts api-platform/core#1905

A value is a value, not a literal. If we go by this logic, then we'll use literals for many other things too, which would of course also be wrong.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/BooleanFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/BooleanFilter.php
@@ -96,9 +96,11 @@ class BooleanFilter extends AbstractContextAwareFilter
             list($alias, $field) = $this->addJoinsForNestedProperty($property, $alias, $queryBuilder, $queryNameGenerator, $resourceClass);
         }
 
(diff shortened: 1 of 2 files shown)


bendavies:
I'm not really sure what you mean by `A value is a value, not a literal`. 
It's always true or false in this instance.

teohhanhui:
Yes, and that true or false, despite it being only 1 of 2 possible values, are still values. My objection is for treating a value as a literal. It means the DBMS will have to generate 2 separate prepared statements for the same query that differs only by this boolean value.

teohhanhui:
You may not notice or care about the implication of that, but imagine other use cases where there are many boolean filters and the combination quickly grows out of control.

bendavies:
👍  thanks for the explanation

dunglas:
We should benchmark both, I guess not using parameters not the class generating a random one, and let the DBMS prepare 2 different requests is faster.

bendavies:
I'd guess it would be immeasurable.

PR #1908, opened 2018-04-30
RFC: Cache trait

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

cc @dunglas  as discussed.

I noticed we were doing the same thing in quite a few places.
Thoughts please?

P.S. i've put very little thought into the naming of the trait and function. Please suggest something better!

### DIFF
--- a/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
+++ b/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Routing;
 
-use Psr\Cache\CacheException;
+use ApiPlatform\Core\Cache\CachedTrait;
 use Psr\Cache\CacheItemPoolInterface;
 
 /**
@@ -23,11 +23,11 @@ use Psr\Cache\CacheItemPoolInterface;
  */
 final class CachedRouteNameResolver implements RouteNameResolverInterface
 {
+    use CachedTrait;
+
     const CACHE_KEY_PREFIX = 'route_name_';
 
-    private $cacheItemPool;
     private $decorated;
-    private $localCache = [];
 
     public function __construct(CacheItemPoolInterface $cacheItemPool, RouteNameResolverInterface $decorated)
     {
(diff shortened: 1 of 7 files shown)


teohhanhui:
-1 as it's unnecessary and only makes the code more confusing.

dunglas:
I'm +1 to factorize this code, it's an error prone one, better not duplicate it everywhere.

bendavies:
@dunglas preference on trait vs abstract class?

teohhanhui:
In that case I'm -1 for something confusing to the reader of the code like `doAOrB`, in this case `getOrSave`.

bendavies:
@teohhanhui did you even read the PR description?

teohhanhui:
Oh, I don't mean -1 for this PR. Sorry for the confusion.

src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
```diff
@@ -43,33 +44,8 @@ public function getRouteName(string $resourceClass, $operationType /**, array $c
...
-
-        try {
-            $cacheItem->set($routeName);
-            $this->cacheItemPool->save($cacheItem);
-        } catch (CacheException $e) {
-            // do nothing
-        }
-
-        return $this->localCache[$cacheKey] = $routeName;
+        return $this->getOrSave($cacheKey, $this->cacheItemPool, function () use ($resourceClass, $operationType, $context) {
```
teohhanhui: Perhaps `getCached`? I also agree with @soyuka that we shouldn't need to pass the `$cacheItemPool` here, it should be in the trait.

src/Bridge/Symfony/Routing/CachedRouteNameResolver.php
```diff
@@ -13,7 +13,7 @@
 
 namespace ApiPlatform\Core\Bridge\Symfony\Routing;
 
-use Psr\Cache\CacheException;
+use ApiPlatform\Core\Cache\CacheTrait;
```
teohhanhui: `CachedTrait`?

src/Cache/CacheTrait.php
```diff
@@ -0,0 +1,54 @@
...
+namespace ApiPlatform\Core\Cache;
+
+use Psr\Cache\CacheException;
+use Psr\Cache\CacheItemPoolInterface;
+
+trait CacheTrait
+{
+    private $localCache = [];
+
+    private function getOrSave(string $cacheKey, CacheItemPoolInterface $pool, callable $callback)
```
soyuka: naming proposal:

```php
<?php
private function cache(string $cacheKey, callable $getValue) {

}
```

With a parent class we can remove CacheItemPoolInterface from the method signature

teohhanhui: We don't need a parent class. The property could be declared in the trait as we already do elsewhere?

src/Cache/CacheTrait.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Cache;
+
+use Psr\Cache\CacheException;
+use Psr\Cache\CacheItemPoolInterface;
+
+trait CacheTrait
```
dunglas: Can you make it `@internal`?

src/Cache/CacheTrait.php
soyuka: also works yes, I just think that parent class > trait if we can :) (ability to use `instanceof parent` etc.)

teohhanhui: I think using inheritance here is... wrong. It's the kind of use case which traits are exactly designed to solve - horizontal code reuse.

bendavies:
comments addressed.
Note that has this has not been added to `CachedIdentifiersExtractor` as it looks a little more complicated, but might be possible.

src/Cache/CachedTrait.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+trait CachedTrait
+{
+    /** @var CacheItemPoolInterface */
+    private $cacheItemPool;
+    private $localCache = [];
+
+    private function getCached(string $cacheKey, callable $getValue)
+    {
+        if (isset($this->localCache[$cacheKey])) {
```
teohhanhui: Actually, I think this is a bug. We should support storing `null` values in the cache.

teohhanhui:
Thanks @bendavies! :tada:

PR #1911, opened 2018-04-30
RFC: denormalize - do not set null values on not nullable properties

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes?
| New feature?  | no
| BC breaks?    | no?
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This was discussed a little bit on slack this afternoon.
Should we *not* call setters with `null` values if the type is not nullable?

This seems to be consistent with the check on line 177.
https://github.com/api-platform/core/blob/592af017c6125928c3a25ddb2f512fa3e234967e/src/Serializer/AbstractItemNormalizer.php#L177-L180

We should respect the property info type `isNullable` value.

This is mainly to prevent denormalize errors when an api's resource may have a not nullable setter: `setFoo(string $foo)` but the api request as come in as `{"foo": null}`

Not sure what the drawbacks are here. I'm sure someone will point something out.

further work if this PR is deemed correct:
The same check should also be done when denormalizing relations.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -204,6 +204,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return;
         }
 
+        if (null === $value && !$type->isNullable()) {
+            return;
+        }
         $this->validateType($attribute, $type, $value, $format);
         $this->setValue($object, $attribute, $value);
     }
--- a/tests/Serializer/AbstractItemNormalizerTest.php
+++ b/tests/Serializer/AbstractItemNormalizerTest.php
@@ -223,12 +223,15 @@ class AbstractItemNormalizerTest extends TestCase
 
         $propertyNameCollectionFactoryProphecy = $this->prophesize(PropertyNameCollectionFactoryInterface::class);
         $propertyNameCollectionFactoryProphecy->create(Dummy::class, [])->willReturn(
-            new PropertyNameCollection(['name', 'relatedDummy', 'relatedDummies'])
+            new PropertyNameCollection(['name', 'description', 'relatedDummy', 'relatedDummies'])
         )->shouldBeCalled();
 
         $propertyMetadataFactoryProphecy = $this->prophesize(PropertyMetadataFactoryInterface::class);
         $propertyMetadataFactoryProphecy->create(Dummy::class, 'name', [])->willReturn(
             new PropertyMetadata(new Type(Type::BUILTIN_TYPE_STRING), '', false, true)
+        )->shouldBeCalled();
+        $propertyMetadataFactoryProphecy->create(Dummy::class, 'description', [])->willReturn(
+            new PropertyMetadata(new Type(Type::BUILTIN_TYPE_STRING, true), '', false, true)
         )->shouldBeCalled();
         $propertyMetadataFactoryProphecy->create(Dummy::class, 'relatedDummy', [])->willReturn(
             new PropertyMetadata(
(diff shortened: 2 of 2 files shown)


teohhanhui:
Should it fail silently? I think it should result in a HTTP 400 status code.

soyuka:
> In my API, I have a NotBlank constraint on the property which satisfies that.

This is always the same issue (I guess) :p. What if you don't have such constraint?

teohhanhui:
Either way, I don't think the normalizer should silently ignore a null value that it cannot set as requested. It should throw an exception. It's then up to the calling code to handle it, and in API Platform we should make sure that it (eventually) becomes a HTTP 400...

bendavies:
@soyuka 
> This is always the same issue (I guess) :p. What if you don't have such constraint?

then you'll have a database violation on insert.

bendavies:
@teohhanhui that seems like a nice solution.

soyuka:
> Either way, I don't think the normalizer should silently ignore a null value that it cannot set as requested. It should throw an exception. It's then up to the calling code to handle it, and in API Platform we should make sure that it (eventually) becomes a HTTP 400...

Issues stays the same when you'd expect a ConstraintValidation instead of a Normalization exception :). And we will not change the listeners order (as discussed in other issues that I'm lazy to look up right now :p).

bendavies:
either way it needs to be handled somehow. at the moment a `500` isn't really acceptable.

dunglas:
A 400 should be thrown. Actually, Symfony's `ObjectNormalizer` should already throw in this case because if this check: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L342

We need to figure why it's not the case.

dunglas:
https://github.com/bendavies/core/blob/3490493718982044420b98335787f40fabdfb4e6/src/Serializer/AbstractItemNormalizer.php#L230 should already trigger an error :|

    php > var_dump(is_int(null));
    bool(false)

bendavies:
@dunglas i'll take a look.

bendavies:
@dunglas 

1. https://github.com/symfony/symfony/blob/cb2a77b24bc01a08787fcf9b4700fd5e03884155/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L297
$this->propertyTypeExtractor is null, so the below is never evaluated. bug in service def?
2. https://github.com/bendavies/core/blob/3490493718982044420b98335787f40fabdfb4e6/src/Serializer/AbstractItemNormalizer.php#L230
the issue here is the doctrine type specifies nullable=true, but the setter does not allow nullable. the doctrine extractor takes precedence?

dunglas:
1. Yes it's intended, we overrode this logic by our own metadata system
2. Ok, so it's not really a bug, indeed the Doctrine extractor takes precedence. It should be fixed on Symfony side (but IMO the best is to fix the code).

dunglas:
Mutators should be in sync with the Doctrine mapping. To get the intended behavior I would:

* Make the setter accept null values
* Add a `@NotNull` validation assertion on the property

bendavies:
ok, sorry the actual problem is for non nullable resources relations, and passing in null.

```
setFoo(Foo $foo)
```
...
```
{"foo": null}
```

This will trigger a 500, as there is no nullable check here:
https://github.com/api-platform/core/pull/1911/files#diff-3e45e8e6ce08f8a7cbbbda1601d5ff29L197

dunglas:
Ok, we should throw in this case.

bendavies:
thanks, i'll work on this.

bendavies:
@dunglas what should i be throwing here? Is there currently a handler that converts whichever exception it is to a validation error?

bendavies:
ahh, `ApiPlatform\Core\Exception\InvalidArgumentException` returns a 400.

soyuka:
What's the status on this @bendavies ?

bendavies:
wow old one...no idea.
i'll take a look next week.

teohhanhui:
> ahh, ApiPlatform\Core\Exception\InvalidArgumentException returns a 400.

We should throw `Symfony\Component\Serializer\Exception\UnexpectedValueException` instead.

flug:
@bendavies Up is this pull request still valid?

alanpoulain:
Closing this PR because it's too old and I can't reproduce.

bendavies:
this will be resolved by https://github.com/symfony/symfony/pull/42502

PR #1913, opened 2018-05-01
Add a BC layer on data providers and deprecation errors

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

The introduction of identifier denormalization changes the type of the passed `$id` to the `getItem()` functions in custom data providers from a string to an array.

This patch allow to continue to use old data providers and will warn users that they have to update their code meanwhile.

WDYT @soyuka @dunglas ?


tests/Fixtures/TestBundle/DataProvider/ContainNonResourceItemDataProvider.php
```diff
@@ -35,10 +35,10 @@ public function getItem(string $resourceClass, $id, string $operationName = null
     {
         // Retrieve the blog post item from somewhere
         $cnr = new ContainNonResource();
-        $cnr->id = $id['id'];
+        $cnr->id = $id;
```
antograssiot: I reverted Antoine's changes in this custom dataprovider so it proves it will continue to work in behat tests "the old way"

antograssiot:
This can be reviewed but should not be merged in 2.2, it should target master when it will be cleaned.

src/DataProvider/ChainItemDataProvider.php
```diff
@@ -44,7 +44,15 @@ public function getItem(string $resourceClass, $id, string $operationName = null
                     continue;
                 }
 
-                return $dataProvider->getItem($resourceClass, $id, $operationName, $context);
+                $identifier = $id;
+                if (!$dataProvider instanceof DenormalizedIdentifiersAwareItemDataProviderInterface) {
+                    @trigger_error(sprintf('Receiving "$id" as non-array in an item data provider is deprecated, your item data providers must implement "%s".', DenormalizedIdentifiersAwareItemDataProviderInterface::class), E_USER_DEPRECATED);
```
dunglas: Is it mandatory to trigger a deprecation notice? In most case, you'll not use composed identifiers (it's not really reliable), and implementing this new interface looks like an annoyance for most people to handle specific use cases.

I suggest to trigger a deprecation only if `\count($identifier) > 1`. WDYT?

src/DataProvider/DenormalizedIdentifiersAwareItemDataProviderInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataProvider;
+
+/**
+ * Retrieves items from a persistence layer.
```
dunglas: Can you change this sentence by something like `Marks data providers able to deal with complex identifiers denormalized as an array`?

src/DataProvider/ChainItemDataProvider.php
antograssiot: @dunglas it is your call, I've no strong opinion on it.
My argument is the change affects all data providers, even if you don't use composite identifiers. In 2.2 you would get `"2"` passed as `$id` and now you will get `['id' => 2]` so in an ideal world, every data provider needs to be updated. Not triggering the deprecation error is ok, it just means that we won't be able to remove this piece of code in the next major version

dunglas: Making it easy for the user to add new data providers matter, so I suggest to trigger the deprecation only if `\count($identifier) > 1`.

antograssiot: Ok I'll update acordingly this morning

antograssiot:
Oh and I've screwed up the branch, I'll clean that :(

src/DataProvider/ChainItemDataProvider.php
```diff
@@ -44,7 +44,17 @@ public function getItem(string $resourceClass, $id, string $operationName = null
...
                 }
 
-                return $dataProvider->getItem($resourceClass, $id, $operationName, $context);
+                $identifier = $id;
+                if (!$dataProvider instanceof DenormalizedIdentifiersAwareItemDataProviderInterface && $identifier && \is_array($identifier)) {
+                    if (\count($identifier) > 1) {
+                        @trigger_error(sprintf('Receiving "$id" as non-array in an item data provider is deprecated in 2.3 in favor of implementing "%s".', DenormalizedIdentifiersAwareItemDataProviderInterface::class), E_USER_DEPRECATED);
+                        $identifier = http_build_query($identifier, '', ';');
+                    } else {
+                        $identifier = current($identifier);
```
antograssiot: Update done @dunglas, do you want me to move this statement up to avoid the `else` ?

PR #1915, opened 2018-05-01
Supports keeping the cursors for pagination

sroze:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ø
| License       | MIT
| Doc PR        | ø

The background of this pull-request is explained in the following (public but non-published) blog post:
https://medium.com/@sroze/cursor-based-pagination-with-apiplatform-74fd1d324723

This is the first step, allowing to manually use the filters to achieve a cursor-based pagination. A follow-up PR will come with an obfuscation mechanism if this one goes through :)

### DIFF
--- a/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
+++ b/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
@@ -16,7 +16,10 @@ namespace ApiPlatform\Core\Hydra\Serializer;
 use ApiPlatform\Core\DataProvider\PaginatorInterface;
 use ApiPlatform\Core\DataProvider\PartialPaginatorInterface;
 use ApiPlatform\Core\Util\IriHelper;
+use ApiPlatform\Core\JsonLd\Serializer\JsonLdContextTrait;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
+use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
@@ -29,12 +32,14 @@ use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 final class PartialCollectionViewNormalizer implements NormalizerInterface, NormalizerAwareInterface, CacheableSupportsMethodInterface
 {
     private $collectionNormalizer;
+    private $resourceMetadataFactory;
     private $pageParameterName;
     private $enabledParameterName;
 
-    public function __construct(NormalizerInterface $collectionNormalizer, string $pageParameterName = 'page', string $enabledParameterName = 'pagination')
+    public function __construct(NormalizerInterface $collectionNormalizer, ResourceMetadataFactoryInterface $resourceMetadataFactory, string $pageParameterName = 'page', string $enabledParameterName = 'pagination')
     {
         $this->collectionNormalizer = $collectionNormalizer;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->pageParameterName = $pageParameterName;
         $this->enabledParameterName = $enabledParameterName;
     }
@@ -69,12 +74,39 @@ final class PartialCollectionViewNormalizer implements NormalizerInterface, Norm
             return $data;
         }
 
+        $metadata = $this->resourceMetadataFactory->create($context['resource_class']);
+        $isPaginatedWithCursor = $paginated && null !== $cursorPaginationAttribute = $metadata->getAttribute('pagination_via_cursor');
+
         $data['hydra:view'] = [
             '@id' => IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $paginated ? $currentPage : null),
             '@type' => 'hydra:PartialCollectionView',
         ];
 
-        if ($paginated) {
+        if ($isPaginatedWithCursor) {
+            $forwardRangeOperator = 'desc' === strtolower($cursorPaginationAttribute['direction']) ? 'lt' : 'gt';
+            $backwardRangeOperator = 'gt' === $forwardRangeOperator ? 'lt' : 'gt';
+
+            $accessor = PropertyAccess::createPropertyAccessor();
+            $objects = iterator_to_array($object);
+            $firstObject = current($objects);
+            $lastObject = end($objects);
+
+            $parsed['parameters'][$cursorPaginationAttribute['field']] = [
+                $backwardRangeOperator.'e' => (string) $accessor->getValue($lastObject, $cursorPaginationAttribute['field']),
+            ];
+
+            $data['hydra:view']['@id'] = IriHelper::createIri($parsed['parts'], $parsed['parameters']);
+            $data['hydra:view']['hydra:previous'] = IriHelper::createIri($parsed['parts'], array_merge($parsed['parameters'], [
+                $cursorPaginationAttribute['field'] => [
+                    $backwardRangeOperator => (string) $accessor->getValue($firstObject, $cursorPaginationAttribute['field']),
+                ],
+            ]));
+            $data['hydra:view']['hydra:next'] = IriHelper::createIri($parsed['parts'], array_merge($parsed['parameters'], [
+                $cursorPaginationAttribute['field'] => [
+                    $forwardRangeOperator => (string) $accessor->getValue($lastObject, $cursorPaginationAttribute['field']),
+                ],
+            ]));
+        } elseif ($paginated) {
             if (null !== $lastPage) {
                 $data['hydra:view']['hydra:first'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, 1.);
                 $data['hydra:view']['hydra:last'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $lastPage);
--- a/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
+++ b/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
@@ -20,6 +20,7 @@ use ApiPlatform\Core\JsonLd\Serializer\JsonLdContextTrait;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
 use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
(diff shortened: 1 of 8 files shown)


sroze:
(failure in unrelated by the way)

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -71,12 +75,40 @@ public function normalize($object, $format = null, array $context = [])
...
+            $forwardRangeOperator = 'desc' === strtolower($cursorPaginationAttribute['direction']) ? 'lt' : 'gt';
+            $backwardRangeOperator = 'gt' === $forwardRangeOperator ? 'lt' : 'gt';
+
+            $accessor = PropertyAccess::createPropertyAccessor();
+            $objects = iterator_to_array($object);
+            $firstObject = current($objects);
+            $lastObject = end($objects);
+
+            if (false !== $lastObject) {
+                $data['hydra:view']['@id'] = IriHelper::createIri($parsed['parts'], $parsed['parameters']);
```
sroze: This... doesn't require the `$lastObject` to be not null

soyuka:
> (CircleCI failure in unrelated by the way)
 Issue with the merging stuff. Just rebase should be fine on master since #1917 .

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -71,12 +75,40 @@ public function normalize($object, $format = null, array $context = [])
             return $data;
         }
 
+        $metadata = isset($context['resource_class']) ? $this->resourceMetadataFactory->create($context['resource_class']) : null;
+        $isPaginatedWithCursor = $paginated && null !== $metadata && null !== $cursorPaginationAttribute = $metadata->getAttribute('pagination_via_cursor');
```
antograssiot: what about using `getCollectionOperationAttribute()` instead of `getAttribute()` ?
isn't `collection_operation_name` key set in the context ?

tests/Fixtures/TestBundle/Entity/SoMany.php
```diff
@@ -0,0 +1,48 @@
...
+use ApiPlatform\Core\Annotation\ApiResource;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\OrderFilter;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\RangeFilter;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ApiResource(attributes={
+ *     "pagination_partial"=true,
+ *     "pagination_via_cursor"={
```
bendavies: what about supporting multiple fields? this is often how it is done i postgres...
`SELECT * FROM table WHERE (date, id) < (prev_date, prev_id) ORDER BY date DESC, id DESC LIMIT 10;`

tests/Fixtures/TestBundle/Entity/SoMany.php
```diff
@@ -0,0 +1,48 @@
...
+ *         "field"="id",
+ *         "direction"="DESC"
+ *     }
+ * })
+ *
+ * @ApiFilter(RangeFilter::class, properties={"id"})
```
bendavies: so these filters *must* match the `pagination_via_cursor`. should there be a check for that?

sroze: Actually, shouldn't we put these filters automatically based on the `pagination_via_cursor` configuration?

tests/Fixtures/TestBundle/Entity/SoMany.php
sroze: I was thinking about adding more later but I think you're right, we can start with multiple.

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -31,12 +33,14 @@
     use JsonLdContextTrait;
 
     private $collectionNormalizer;
+    private $resourceMetadataFactory;
     private $pageParameterName;
     private $enabledParameterName;
 
-    public function __construct(NormalizerInterface $collectionNormalizer, string $pageParameterName = 'page', string $enabledParameterName = 'pagination')
+    public function __construct(NormalizerInterface $collectionNormalizer, ResourceMetadataFactoryInterface $resourceMetadataFactory, string $pageParameterName = 'page', string $enabledParameterName = 'pagination')
```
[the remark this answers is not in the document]
sroze: Well, if we apply the BC policy here I shouldn't change the argument order either 😄

tests/Fixtures/TestBundle/Entity/SoMany.php
bendavies: Yes. Think so!

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -71,12 +75,40 @@ public function normalize($object, $format = null, array $context = [])
...
+            '@id' => IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $paginated && !$isPaginatedWithCursor ? $currentPage : null),
             '@type' => 'hydra:PartialCollectionView',
         ];
 
-        if ($paginated) {
+        if ($isPaginatedWithCursor) {
+            $forwardRangeOperator = 'desc' === strtolower($cursorPaginationAttribute['direction']) ? 'lt' : 'gt';
+            $backwardRangeOperator = 'gt' === $forwardRangeOperator ? 'lt' : 'gt';
+
+            $accessor = PropertyAccess::createPropertyAccessor();
```
dunglas: You should reuse the existing service if it exists (and create it only if `null` is passed in the constructor). This way we'll benefit from the component's cache.

tests/Fixtures/TestBundle/Entity/SoMany.php
sroze: I've looked at it, will do that in another PR while refactoring the pagination actually.

tests/Fixtures/TestBundle/Entity/SoMany.php
sroze: Multiple fields added 👍

sroze:
Status: Needs review

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -71,12 +78,29 @@ public function normalize($object, $format = null, array $context = [])
             return $data;
         }
 
+        $metadata = isset($context['resource_class']) && null !== $this->resourceMetadataFactory ? $this->resourceMetadataFactory->create($context['resource_class']) : null;
+        $isPaginatedWithCursor = $paginated && null !== $metadata && null !== $cursorPaginationAttribute = $metadata->getCollectionOperationAttribute($context['collection_operation_name'], 'pagination_via_cursor', null, true);
```
soyuka: I find hese lines a bit long but apart from that 👍

sroze: If I split them, it will be longer (i.e. repeating these `metadata !== null`) 🙊

soyuka: I know

meyerbaptiste:
What about HAL (`ApiPlatform\Core\Hal\Serializer\CollectionNormalizer`)?

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -71,12 +78,29 @@ public function normalize($object, $format = null, array $context = [])
...
         $data['hydra:view'] = [
-            '@id' => IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $paginated ? $currentPage : null),
+            '@id' => IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $paginated && !$isPaginatedWithCursor ? $currentPage : null),
             '@type' => 'hydra:PartialCollectionView',
         ];
 
-        if ($paginated) {
+        if ($isPaginatedWithCursor) {
+            $objects = iterator_to_array($object);
+            $firstObject = current($objects);
```
theofidry: this will fail if there is no object

sroze: yup, what @meyerbaptiste said

sroze:
> What about HAL (ApiPlatform\Core\Hal\Serializer\CollectionNormalizer)?

Whoever needs it will add it? 🤷‍♂️

bendavies:
it would be nice if all output formats support a major feature like this. otherwise it sets a precedent that contributors only add features to the formats they use, which will result in not great framework caveats of "Awesome Feature XXXX (only supported for JSON API)"

sroze:
@bendavies I do agree with you. Do you want to add HAL support in this PR? 😉

bendavies:
i could but it would not be for a few weeks. currently on vacation

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
teohhanhui: Could be extracted into a private method?

teohhanhui:
About the format support, on the other hand, we would not want to discourage contribution because of the need to support all formats? And sometimes it's just better for people who are actually familiar with the format / who actually use it to add the feature.

teohhanhui:
> "Awesome Feature XXXX (only supported for JSON API)"

Only supported for JSON-LD / Hydra is acceptable to me, as it's the default / recommended(?) format. But not for the other formats. :stuck_out_tongue:

trsteel88:
This PR doesn't support duplicate values for something you are sorting by. e.g if you are sorting by "createdAt" and there are multiple entities it needs to check for a backup field like id.

Query would be something like:

```
entity.createdAt > [DATE from first/last collection] or (entity.createdAt = [DATE from first/last collection] and entity.id > [FIRST OR LAST ID OF COLLECTION])
```

direction of the > is going to change depending on which way you are going in the pagination.

bendavies:
it supports multiple fields. it's up to the developer to use the fields they need to do the sorting.

sroze:
What do you expect as a change in here except fixing the merge conflicts? Can we go ahead without all the formats?

dunglas:
Let’s go ahead

sroze:
Rebased on master 👍

soyuka:
ping @sroze we'd like to merge this one

bendavies:
Let's get this going again. It's an awesome feature.
 I'd like to add hal support too

tests/Fixtures/TestBundle/Entity/SoMany.php
theofidry: Rather than just supporting fields, it might be _needed_. Unlike the sorting for a regular pagination, the cursor _must_ be unique. Although not efficient performance-wise, the safest way to achieve this is to always append (when necessary as it requires what fields are already present) a sort by ID as a last order by.

I also think you need to account for the `NULL` values as they might mess up the order

soyuka:
Thanks for the feature @sroze, I merged this onto master in https://github.com/api-platform/core/pull/2532!

sroze:
My please. Thanks for taking it over the line!

Issue #669, opened 2016-08-05
[Question] include null values for date filter


aledeg:
I would like to always include null values while searching for dates.
It seems impossible for now. Correct me if I am wrong.

I tried to set the same field with different values but as expected, only one is used.

``` yaml
filter.subscription_date_filter:
    parent: "api_platform.doctrine.orm.date_filter"
    arguments:
        - validFrom: 'include_null_before'
          validFrom: 'include_null_after'
    tags: [ { name: 'api_platform.filter', id: 'filter.subscription_date' } ]
```

Did I miss something?

dunglas:
According to https://github.com/api-platform/docs/blob/master/core/filters.md#managing-null-values, it should work. You only need one line (before or after dpeending of the order you want).

Can you provide your entity mapping?

aledeg:
If I choose 'include_null_after', I have the NULL values when I filter the date with the ```after``` parameter. But I won't have a match with the ```before``` parameter.

If I choose 'include_null_before, I have the NULL values when I filter the date with the ```before``` parameter. But I won't have a match with the ```after``` parameter.

I check the code and it works as expected. But it's not what I am trying to achieve.
I would like to have NULL values displayed all the time whether I am filtering with ```before``` or ```after```

Regarding the entity mapping, I think it's irrelevant since it's not related. I include it anyway.
```php
namespace ApiBundle\Entity;

use Doctrine\ORM\Mapping as ORM;
use ApiPlatform\Core\Annotation\ApiResource;
use ApiPlatform\Core\Annotation\ApiProperty;

/**
 * @ORM\Entity
 * @ORM\Table(name="contact")
 * @ApiResource(attributes={
 *     "filters"={
 *         "filter.contact_date"
 *     }
 * })
 */
class Contact
{

    /**
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @ORM\Column(type="date", nullable=true)
     */
    private $birthDate;

    public function setId($id)
    {
        $this->id = $id;
    }

    public function getId()
    {
        return $this->id;
    }
    public function setBirthDate(\DateTime $birthDate = null)
    {
        $this->birthDate = $birthDate;
    }

    public function getBirthDate()
    {
        return $this->birthDate;
    }

}
```
The filter configuration:
```yaml
filter.contact_date_filter:
    parent: 'api_platform.doctrine.orm.date_filter'
    arguments:
[...]

PR #1928, opened 2018-05-03
feature: add a way to always include null in the date filter

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #669 
| License       | MIT
| Doc PR        | todo

Add a way to always include null values.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/DateFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/DateFilter.php
@@ -33,6 +33,7 @@ class DateFilter extends AbstractContextAwareFilter
     const EXCLUDE_NULL = 'exclude_null';
     const INCLUDE_NULL_BEFORE = 'include_null_before';
     const INCLUDE_NULL_AFTER = 'include_null_after';
+    const INCLUDE_NULL_BEFORE_AND_AFTER = 'include_null_before_and_after';
     const DOCTRINE_DATE_TYPES = [
         'date' => true,
         'datetime' => true,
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Filter/DateFilter.php
```diff
@@ -33,6 +33,7 @@ class DateFilter extends AbstractContextAwareFilter
     const EXCLUDE_NULL = 'exclude_null';
     const INCLUDE_NULL_BEFORE = 'include_null_before';
     const INCLUDE_NULL_AFTER = 'include_null_after';
+    const ALWAYS_INCLUDE_NULL = 'always_include_null';
```
antograssiot: `INCLUDE_NULL` only maybe to be more inline with the existing options name ?

src/Bridge/Doctrine/Orm/Filter/DateFilter.php
```diff
@@ -190,7 +191,8 @@ protected function addWhere(QueryBuilder $queryBuilder, QueryNameGeneratorInterf
             $queryBuilder->andWhere($baseWhere);
         } elseif (
             (\in_array($operator, [self::PARAMETER_BEFORE, self::PARAMETER_STRICTLY_BEFORE], true) && self::INCLUDE_NULL_BEFORE === $nullManagement) ||
-            (\in_array($operator, [self::PARAMETER_AFTER, self::PARAMETER_STRICTLY_AFTER], true) && self::INCLUDE_NULL_AFTER === $nullManagement)
+            (\in_array($operator, [self::PARAMETER_AFTER, self::PARAMETER_STRICTLY_AFTER], true) && self::INCLUDE_NULL_AFTER === $nullManagement) ||
+            (\in_array($operator, [self::PARAMETER_AFTER, self::PARAMETER_STRICTLY_AFTER, self::PARAMETER_BEFORE, self::PARAMETER_STRICTLY_BEFORE], true) && self::ALWAYS_INCLUDE_NULL === $nullManagement)
```
antograssiot: Maybe invert the conditions as the `===` might have a lower cost ?

teohhanhui:
I'm not sure I understand the purpose of this. Aren't null values already included if you use one of `INCLUDE_NULL_BEFORE` or `INCLUDE_NULL_AFTER`?

Simperfit:
Look at the linked issue:

The OP wants to include null values always and it seems that doesn't work for now, so it's like a merge of both after and before so it will always includes null values.

teohhanhui:
Uhh, yeah I just looked at the issue... Maybe call it `INCLUDE_NULL_BEFORE_AND_AFTER`, to be explicit about what it actually does. I think it will cause confusion otherwise, when you somehow get null values at both ends.

Simperfit:
done

soyuka:
thanks @Simperfit

Issue #767, opened 2016-09-27
Add jsonld context to resource


maechler:
It would be nice to have the possibility to split the property context into the namespace and the property itself. Here an example to illustrate what I mean:

**Entity BibliographicResource**
```php
class BibliographicResource
{
    /**
     * @var string
     *
     * @ApiProperty(iri="http://purl.org/dc/terms/title")
     */
    private $title;
}
```

**Current output**
Context:

```json
{
    "title": "http://purl.org/dc/terms/title"
}
```

Entity:

```json
{
      "title": "some title"
}
```

**Desired output**

Context:

```json
{
    "dct": "http://purl.org/dc/terms/"
}
```

Entity:

```json
{
      "dct:title": "some title"
}
```

**Possible implementation**

One way of implementing this could probably be to add context definitions to the entity class. 
e.g.

```php

/**
 * @ApiResource(
 *     "jsonld_context"={
 *         "context": {
 *             "dct":"http://purl.org/dc/terms/"
 *         }       
 *     }
 * )
 */
class BibliographicResource
{
    /**
     * @var string
     *
     * @ApiProperty(iri="dct:title")
     */
    private $title;
}
```

Simperfit:
@maechler This is fixed ?

maechler:
@Simperfit No, that is not fixed yet. I have added support to extend the json-ld context on property level: https://github.com/api-platform/core/pull/752
But to allow to modify the context completely, it should also be possible on an entity level.

PR #1930, opened 2018-05-03
[WIP] add context to resources level

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no wip
| Fixed tickets | #767
| License       | MIT
| Doc PR        | todo

@maechler As adding it to the context is easy, adding it directly to the entity seems to be more work.

What I though is maybe a nameconverter that will transform the property based on the context, I don't know if it's the right thing to do, do you have a better idea ?

### DIFF
--- a/features/main/jsonld_context.feature
+++ b/features/main/jsonld_context.feature
@@ -1,0 +1,29 @@
+Feature: Get the resource level context
+  In order to see the resource level context
+  As a client software developer
+  I need to be able to create and get the JSON LD context
+
+  @createSchema
+  Scenario: Create a resource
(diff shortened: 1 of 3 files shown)


maechler:
@Simperfit I have to admit I have not used API Platform in a while, that is why I can not come up with a better idea. But for me a name converter should work just as fine.

flug:
@Simperfit Up is this pull request still valid?

vincentchalamon:
Closing as no activity since years

PR #1931, opened 2018-05-03
Remove useless try/catch in the CachedTrait

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Sorry, I missed the #1908 review.

### DIFF
--- a/src/Cache/CachedTrait.php
+++ b/src/Cache/CachedTrait.php
@@ -33,26 +33,18 @@ trait CachedTrait
 
         try {
             $cacheItem = $this->cacheItemPool->getItem($cacheKey);
-
-            if ($cacheItem->isHit()) {
-                return $this->localCache[$cacheKey] = $cacheItem->get();
-            }
         } catch (CacheException $e) {
-            //do nothing
+            return $this->localCache[$cacheKey] = $getValue();
+        }
+
+        if ($cacheItem->isHit()) {
+            return $this->localCache[$cacheKey] = $cacheItem->get();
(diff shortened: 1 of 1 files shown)


soyuka:
thanks @meyerbaptiste

src/Cache/CachedTrait.php
```diff
@@ -33,26 +33,18 @@ private function getCached(string $cacheKey, callable $getValue)
...
-        if (!isset($cacheItem)) {
-            return $this->localCache[$cacheKey] = $value;
+        if ($cacheItem->isHit()) {
+            return $this->localCache[$cacheKey] = $cacheItem->get();
         }
 
-        try {
-            $cacheItem->set($value);
-            $this->cacheItemPool->save($cacheItem);
-        } catch (CacheException $e) {
```
bendavies: I was only copying exciting code in the previous pr so I am not sure of the original author's thoughts, but you are now keeping the exception handling above but removing this one. Should we be consistent?

teohhanhui: There is no chance of `CacheException` being thrown here. The original author (i.e. myself) was mistaken. :laughing:

soyuka: And we all copied without thinking twice :laughing:

PR #1936, opened 2018-05-08
[WIP] Allow to filter attributes of related non-api-resource entities.

michaelpetri:
Problem:
In the case we've got an api resource which relates to an entity which isn't exposed as api resource. We can not filter the attributes of the related/child entity.

Solution:
Create child context and pass it to default normalization callback to ensure correct attribute filters are passed for non api resource entities.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -420,7 +420,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return $this->normalizeCollectionOfRelations($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
         }
 
-        if ($type && ($className = $type->getClassName())) {
+        if ($type && $className = $type->getClassName()) {
             if ($this->resourceClassResolver->isResourceClass($className)) {
                 return $this->normalizeRelation($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
             }
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -420,12 +420,13 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return $this->normalizeCollectionOfRelations($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
         }
 
-        if (
-            $type &&
-            ($className = $type->getClassName()) &&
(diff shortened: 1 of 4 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -420,12 +420,13 @@ protected function getAttributeValue($object, $attribute, $format = null, array
...
-            $type &&
-            ($className = $type->getClassName()) &&
-            $this->resourceClassResolver->isResourceClass($className)
-        ) {
-            return $this->normalizeRelation($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
+        if ($type && ($className = $type->getClassName())) {
+            if ($this->resourceClassResolver->isResourceClass($className)) {
+                return $this->normalizeRelation($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
+            }
+            else {
```
dunglas: This `else` block can be removed, it is useless because of the `return` statement in the previous `if`)

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -420,12 +420,13 @@ protected function getAttributeValue($object, $attribute, $format = null, array
             return $this->normalizeCollectionOfRelations($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
         }
 
-        if (
-            $type &&
-            ($className = $type->getClassName()) &&
-            $this->resourceClassResolver->isResourceClass($className)
-        ) {
-            return $this->normalizeRelation($propertyMetadata, $attributeValue, $className, $format, $this->createChildContext($context, $attribute));
+        if ($type && ($className = $type->getClassName())) {
```
dunglas: The parenthesis are now useless, should be removed.

michaelpetri:
Hi @dunglas,

thanks for your feedback. I have removed the desired lines and extended the abstract normalizer test to cover the new line from the PR. Hope this fits your requirements.

Best regards
Michael

michaelpetri:
Okay i see extending the dummy entity/resource for the abstract normalization test wasn't that smart.

I will move everything to an isolated test szenario with it's own dummy. :)

flug:
@michaelpetri Up is this pull request still valid?

PR #1938, opened 2018-05-09
[JSON-API] Add initial spec compliance on "Inclusion of Related Resou…

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Ref http://jsonapi.org/format/#fetching-includes

This add spec compliance in the first place.
I'll try to add support for this latter if no-ones beats me to it...

### DIFF
--- a/src/JsonApi/EventListener/IncludeParametersListener.php
+++ b/src/JsonApi/EventListener/IncludeParametersListener.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\JsonApi\EventListener;
+
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @internal
+ *
+ * @see http://jsonapi.org/format/#fetching-includes
+ */
+final class IncludeParametersListener
+{
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if (
+            'jsonapi' !== $request->getRequestFormat() ||
+            !$request->attributes->get('_api_resource_class') ||
+            !$request->query->get('include')
+        ) {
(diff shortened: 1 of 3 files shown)


src/JsonApi/EventListener/IncludeParametersListener.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\JsonApi\EventListener;
+
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @see http://jsonapi.org/format/#fetching-includes
```
dunglas: Can you mark it `@internal` for now, because we'll drop it when proper support will be added.

dunglas:
The existing `PropertyFilter` has almost exactly everything we need for a proper implementation. The only difference is the naming.

I agree to merge this one as a bug fix. If we want a full implementation in 2.3 or 2.4, we should reuse `PropertyFilter`.

antograssiot:
The PropertyFilter can already be used with the `fields` parameter like `foo?fields[]=description` and will indeed be a good start to support the 'include' parameter but the main difference is that it is not only a limited attribute exposition, it is the addition of a node at root level that will include every related resources or part of them using the PropertyFilter....

antograssiot:
@dunglas I update the PR

Do you want me to add in this same PR a behat scenario that will demonstrate that the ApiProperty already works (maybe not as expected in the spec but still, it is how it works right now)

```
  Scenario: Apply property filter based on the 'fields'
    Given there are 2 dummy property objects
    When I send a "GET" request to "/dummy_properties?fields[DummyProperty]=id,foo,bar"
    And the JSON should be valid according to the JSON API schema
    And the JSON node "data" should have 2 elements
    And the JSON node "data[0].attributes._id" should be equal to "1"
    And the JSON node "data[0].attributes.foo" should be equal to "Foo #1"
    And the JSON node "data[0].attributes.bar" should be equal to "Bar #1"
    And the JSON node "data[0].attributes.group" should not exist
```

dunglas:
This one as bug fix and the implementation for master looks good to me.

dunglas:
Thanks @antograssiot

antograssiot:
:ok_hand:

Issue #857, opened 2016-11-24
PUT with embedded relation, no route matches "<int>", ItemNormalizer bug ?


VaN-dev:
Hello,

i'm facing a bug when submitting a PUT request with embedded relation :

data sample posted :

```
array:7 [
  "@context" => "/app_dev.php/contexts/Staking"
  "@id" => "/app_dev.php/stakings/10"
  "@type" => "Staking"
  "id" => 10
  "tournaments" => array:1 [
    0 => array:4 [
      "@id" => "/app_dev.php/staking_tournaments/3"
      "@type" => "StakingTournament"
      "id" => 3
      "tournament" => "/app_dev.php/tournaments/1"
    ]
  ]
  "name" => "Staking #111"
  "mode" => "simple"
]
```

api-platform returns me the following error :

```
{
    "@context":"\app_dev.php/contexts/Error",
    "@type":"Error",
    "hydra:title":"An error occurred",
    "hydra:description":"No route matches "3".", 
    ...
}
```

The error is caused by `src\Serializer\ItemNormalizer.php`, line 36, using `$data['id']` instead of `$data['@id']`. In my case, `$data['id']` is supposed to be an integer, the PK in database, not a resource URL.

Simperfit:
Could you please test the PR ?

dunglas:
I think this bug is related with how embed relations are handled. @VaN-dev, can you open a PR with a failing Behat test?

dunglas:
Can you also post the full stack trace, because `src\Serializer\ItemNormalizer.php` should not be called at all (the JSON-LD normalizer must be used).

Simperfit:
I've added the failing behat test case and this happen when sending an int in iri using the header to be application/json like in this [case](https://github.com/api-platform/core/pull/877/files#diff-bd4acc142ca94f33b7ba882cc1b52e42R52)

PR #1837, opened 2018-04-09
Add a new `getIriFromPlainIdentifier` method and remove ItemDataProvider from AbstractItemNormalizer

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | multiples need to be sure
| License       | MIT
| Doc PR        | todo

As discussed with @soyuka in #1834 we should not use `ItemDataProvider` in the normalizer and the the process works totally with IRIs.

So I'm adding a new method to convert from plainIdentifiers to IRIs.

PR #1939, opened 2018-05-09
[RFC] Refactor population of object in serializer

### LINKED ISSUE #1937: Cannot POST id to create new resources
Currently, it's not possible to create new resources while providing a custom `id` attribute because APIP will prevent from doing so saying `Update is not allowed for this operation.`.

This is caused by checking `$context['api_allow_update']` in the serializer which is set in the `SerializerContextBuilder`. Basically updating is allowed if the request was a `PUT` or `PATCH` request.

This is check nonsense for two reasons:

* First of all, it's hardcoded on `id` so if my identifier is named `uuid` this check has no effect.
* It's perfectly valid to send a POST request with an identifier, in case the resource **did not exist yet**

So I'd like to remove that check completely.

The next step would be the question if we want to validate this and imho there's only one solution to this: The serializer is the only one knowing whether an identifier was (or multiple were) provided or not. Right now it's hardcoded to `id` but instead we should use the `IdentifiersExtractor` to ask for all of them. Then we need to ask the `ItemDataProvider` if an item for these `$identifiers` already exists and **only if so, we need to throw the exception**. That would mean the serializers need both an instance of `IdentifiersExtractorInterface` and `ItemDataProviderInterface`. Also, it's actually the same for all the serializers. But I'm not really sure, the list of constructor arguments goes on and on and on for the serializers. This is usually not a good sign and maybe means we need to refactor something else first? Wdyt?

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1937
| License       | MIT
| Doc PR        | 

This is completely untested but I think it should be enough to outline the plan as described in https://github.com/api-platform/core/issues/1937

Basically this would fix two things:

* No more hardcoded dependency on `id` but using the `IdentifiersExtractor` instead.
* It would actually fix #1937 because the serializers now test if the object even exists and only then disallow populating the object.

Looking for feedback here :-)

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Serializer;
 
+use ApiPlatform\Core\Api\IdentifiersExtractorInterface;
 use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\OperationType;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
@@ -50,8 +51,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     protected $localCache = [];
     protected $itemDataProvider;
     protected $allowPlainIdentifiers;
+    protected $identifiersExtractor;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, ItemDataProviderInterface $itemDataProvider = null, bool $allowPlainIdentifiers = false)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, ItemDataProviderInterface $itemDataProvider = null, bool $allowPlainIdentifiers = false, IdentifiersExtractorInterface $identifiersExtractor = null)
     {
         parent::__construct($classMetadataFactory, $nameConverter);
 
@@ -66,6 +68,12 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $this->setCircularReferenceHandler(function ($object) {
             return $this->iriConverter->getIriFromItem($object);
         });
+
+        if (null === $identifiersExtractor) {
+            @trigger_error(sprintf('Not passing an instance of IdentifiersExtractorInterface in "%s" is deprecated and will be mandatory in version 3.0. Not passing it is deprecated since 2.2.', __METHOD__), E_USER_DEPRECATED);
+        }
+
+        $this->identifiersExtractor = $identifiersExtractor;
     }
 
     /**
@@ -114,6 +122,8 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         if (!isset($context['resource_class'])) {
             $context['resource_class'] = $class;
         }
+
+        $this->setObjectToPopulate($data, $class, $context);
 
         return parent::denormalize($data, $class, $format, $context);
     }
@@ -487,5 +497,65 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         }
 
         return $iri;
+    }
+
+    /**
+     * Sets the object to populate if allowed to and not set already. If you pass identifiers, it will search for them,
+     * otherwise it falls back to the provided IdentifiersExtractor.
+     *
+     * @param $data
+     * @param string $class
+     * @param array  $context
+     * @param array  $identifiers
+     */
+    protected function setObjectToPopulate($data, $class, array $context = [])
+    {
+        if (isset($context[self::OBJECT_TO_POPULATE]) || !\is_array($data)) {
+            return;
+        }
+
+        $identifiers = $this->getIdentifiersForDenormalization($class);
+
+        if (0 === \count(array_intersect(array_keys($data), $identifiers))) {
+            return;
+        }
+
+        $identifiersData = [];
+        foreach ($identifiers as $id) {
+            $identifiersData[] = $data[$id];
+        }
+
+        $iri = implode(';', $identifiersData); // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 got merged.
+
+        try {
+            $object = $this->iriConverter->getItemFromIri($iri, $context + ['fetch_data' => true]);
+        } catch (ItemNotFoundException $e) {
+            // IRI was valid but no item was found, do not populate
+            return;
+        } catch (InvalidArgumentException $e) {
+            // IRI was invalid
+            return;
+        }
+
+        if (null !== $object && isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
+            throw new InvalidArgumentException('Update is not allowed for this operation.');
+        }
+
+        $context[self::OBJECT_TO_POPULATE] = $object;
+    }
+
+    /**
+     * Returns the identifiers used when denormalizing an object.
+     * Usually this is just what the IdentifiersExtractor provides but you might want to override this
+     * method in your normalizer depending on the format used (e.g. json-ld uses "@id" by definition).
+     */
+    protected function getIdentifiersForDenormalization(string $class): array
+    {
+        // Backwards compatibility
+        if (null === $this->identifiersExtractor) {
+            return ['id'];
+        }
+
+        return $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
     }
 }
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -21,6 +21,7 @@ use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\InvalidValueException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
+use ApiPlatform\Core\Exception\PropertyNotFoundException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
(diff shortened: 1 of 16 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -481,4 +491,34 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+     * @param array $identifiers
+     */
+    protected function populateObjectIfAllowed($data, $class, array $context = [], array $identifiers = [])
+    {
+        if (!isset($context[self::OBJECT_TO_POPULATE])) {
+            if (0 === count($identifiers)) {
+                $identifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
+            }
+
+            if (0 !== count(array_intersect(array_keys($data), $identifiers))) {
```
Toflar: I have no clue if this works, it was just a quick hack and needs to be covered by tests. The idea here is just to check if the data contains all identifiers.

src/Serializer/ItemNormalizer.php
```diff
@@ -13,54 +13,11 @@
...
-            $this->updateObjectToPopulate($data, $context);
-        }
-
-        return parent::denormalize($data, $class, $format, $context);
-    }
-
-    private function updateObjectToPopulate(array $data, array &$context)
-    {
-        try {
-            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri((string) $data['id'], $context + ['fetch_data' => true]);
```
Toflar: This whole method was nonsense imho. Looks like it was developed before there was the `IdentifiersExtractor`.

dunglas: > Looks like it was developed before there was the `IdentifiersExtractor`.

It's the case.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -481,4 +491,44 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+    }
+
+    /**
+     * Returns the identifiers used when denormalizing an object.
+     * Usually this is just what the IdentifiersExtractor provides but you might want to override this
+     * method in your normalizer depending on the format used (e.g. json-ld uses "@id" by definition).
+     */
+    protected function getIdentifiersForDenormalization(string $class): array
+    {
+        return $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
```
antograssiot: $identifierExtractor is nullable so you can't  call a metod on it directly

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -65,6 +67,12 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
         $this->setCircularReferenceHandler(function ($object) {
             return $this->iriConverter->getIriFromItem($object);
         });
+
+        if (null === $identifiersExtractor) {
+            @trigger_error(sprintf('Not passing an instance of IdentifiersExtractorInterface in "%s" is deprecated and will be mandatory in version 3.0. Not passing it is deprecated since 2.2.', __METHOD__), E_USER_DEPRECATED);
```
Taluu: `2.3` nay ? Kinda strange to introduce a deprecated on > 2.2.0...

src/Serializer/ItemNormalizer.php
Taluu: You should add a deprecated error here saying this class is deprecated, and "unabstract" the `AbstractItemNormalizer`

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -481,4 +491,44 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+            $object = $this->itemDataProvider->getItem($class, $identifiers, null, $context);
+
+            if (null !== $object) {
+                if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
+                    throw new InvalidArgumentException('Update is not allowed for this operation.');
+                }
+
+                $context[self::OBJECT_TO_POPULATE] = $object;
+            }
+        }
```
soyuka:
```
        if (0 === count(array_intersect(array_keys($data), $identifiers))) {
            return;
        }

        $object = $this->itemDataProvider->getItem($class, $identifiers, null, $context);

        if (null === $object) {
            return;
        }

                if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
                    throw new InvalidArgumentException('Update is not allowed for this operation.');
                }

                $context[self::OBJECT_TO_POPULATE] = $object;
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -481,4 +491,44 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+     */
+    protected function populateObjectIfAllowed($data, $class, array $context = [])
+    {
+        if (isset($context[self::OBJECT_TO_POPULATE])) {
+            return;
+        }
+
+        $identifiers = $this->getIdentifiersForDenormalization($class);
+
+        if (0 !== count(array_intersect(array_keys($data), $identifiers))) {
```
soyuka: I'm really not sure about this but maybe that performances would be better with a simple foreach test:

```
foreach ($identifiers as $identifier) { if (!isset($data[$identifier])) { return false; } }
return true;
```

soyuka: Also it should be `\count`

dunglas: Can be simplified to `if (\count(array_intersect(array_keys($data), $identifiers))`

Toflar:
Hey guys, this PR is not ready to review at all 😄 I mean, I know there's loads of stuff to cleanup, null pointer exceptions etc. but all I wanted to know is if you think the general concept could work and it's worth continuing into this direction or if I'm completely wrong. What I'm e.g. missing is the use of the IriConverter. What's the idea there. Asking the IriConverter first if there's a result and if not, asking the ItemDataProvider so both is supported?

src/Serializer/ItemNormalizer.php
Toflar: I don't think that's a good plan. I'd keep it as is because the `AbstractItemNormalizer` is not used directly anywhere now and maybe there will be the day we have to add something to the `ItemNormalizer`. It can still be refactored for 3.0 even if there's no deprecation message.

src/Serializer/AbstractItemNormalizer.php
Toflar: Not really, not passing an `id` in a POST request is a bug to me but I don't care as long as it's published soon it can also go into 2.3 😄 Deprecating in bugfix releases is fine.

soyuka:
The ItemDataProvider shouldn't be useful here. IMO IriConverter should be sufficient and we would need to merge https://github.com/api-platform/core/pull/1837 before being able to fix this correctly.

Toflar:
@soyuka do you mean like so?

soyuka:
@Toflar yes this is really good!

Toflar:
Okay, phpunit is fixed. Behat left 😓

Taluu:
Sorry, couldn't help it. :D

soyuka:
Don't you run tests locally?

Toflar:
Okay guys, this is as far as I could get. It's finished. But some behat tests are left to fix which I can't get done because they are not isolated and it's mostly database schema errors. I'll leave it here again for the record: It's a horrible idea not to isolate and create a new schema for every test. You choose performance over DX.

Toflar:
@soyuka will try to help with the behat tests. I've restored them to the initial state as per his request.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +499,89 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+     * @param array  $context
+     * @param array  $identifiers
+     */
+    protected function setObjectToPopulate($data, $class, array $context = [])
+    {
+        if (isset($context[self::OBJECT_TO_POPULATE]) || !\is_array($data)) {
+            return;
+        }
+
+        $updateAllowed = isset($context['api_allow_update']) && true !== $context['api_allow_update'];
```
bendavies: should be `===` not `!==`

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +499,89 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+            $object = $this->iriConverter->getItemFromIri($iri, $context + ['fetch_data' => true]);
+        } catch (ItemNotFoundException $e) {
+            // IRI was valid but no item was found, do not populate
+            return;
+        } catch (InvalidArgumentException $e) {
+            // IRI was invalid
+            return;
+        }
+
+        if (null !== $object && $updateAllowed) {
```
bendavies: should be `!$updateAllowed`

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +499,89 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+    /**
+     * Sets the object to populate if allowed to and not set already. If you pass identifiers, it will search for them,
+     * otherwise it falls back to the provided IdentifiersExtractor.
+     *
+     * @param $data
+     * @param string $class
+     * @param array  $context
+     * @param array  $identifiers
+     */
+    protected function setObjectToPopulate($data, $class, array $context = [])
```
bendavies: you need to pass `$context` by reference as you are modifying it.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +499,89 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+    {
+        if (isset($context[self::OBJECT_TO_POPULATE]) || !\is_array($data)) {
+            return;
+        }
+
+        $updateAllowed = isset($context['api_allow_update']) && true !== $context['api_allow_update'];
+        $identifiers = $this->getIdentifiersForDenormalization($class);
+        $writeableIdentifiers = [];
```
bendavies: this logic is flawed unfortunately. You cannot validate the writability of identifiers here, as identifiers can be used to reference nested relations, to update them. example:
https://github.com/api-platform/core/blob/master/features/hal/max_depth.feature#L58

summary. having an identifier in the data does not mean it is trying to be updated.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +499,89 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+
+        $identifiersData = [];
+        foreach ($writeableIdentifiers as $id) {
+            $identifiersData[] = $data[$id];
+        }
+
+        $iri = implode(';', $identifiersData); // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 got merged.
+
+        try {
+            $object = $this->iriConverter->getItemFromIri($iri, $context + ['fetch_data' => true]);
```
bendavies: exceptions thrown here should not be caught. the old code did not catch.

src/Serializer/ItemNormalizer.php
```diff
@@ -24,43 +22,4 @@
...
-        } catch (InvalidArgumentException $e) {
```
bendavies: this has to be retained, it fixes https://github.com/api-platform/core/issues/857

bendavies:
I have all the tests passing. I'll PR in the morning.

Toflar:
Closing in favor of https://github.com/api-platform/core/pull/2022.

Issue #1934, opened 2018-05-05
No "included" object in JSON-API response


Giant-Jelly:
Been trying to get JSON-API responses to work but cannot seem to find any way to get included data returned. 

The API returns the normal JSON-API format so i assume its working but there is no included data

JSON-API Example (http://jsonapi.org/examples/): 
```
{
  "data": [{
    "type": "articles",
    "id": "1",
    "attributes": {
      "title": "JSON API paints my bikeshed!",
      "body": "The shortest article. Ever.",
      "created": "2015-05-22T14:56:29.000Z",
      "updated": "2015-05-22T14:56:28.000Z"
    },
    "relationships": {
      "author": {
        "data": {"id": "42", "type": "people"}
      }
    }
  }],
  "included": [
    {
      "type": "people",
      "id": "42",
      "attributes": {
        "name": "John",
        "age": 80,
        "gender": "male"
      }
    }
  ]
}
```

Response from my API:

```
/api/stats/22.jsonapi?include=vehicle

{
  data: {
    id: "/api/stats/22",
    type: "Stat",
    attributes: {
      _id: 22,
      engineSize: "1",
      drivetrain: "rwd",
      horsepower: "250"
    },
    relationships: {
      vehicle: {
        data: {
          type: "Vehicle",
          id: "/api/vehicles/1"
        }
      }
    }
  }
}
``` 

GraphQL request/response from my api:

Request:
```
{
  stat(id: "/api/stats/22") {
    id
    _id
    horsepower
    vehicle {
      id
      _id
      make
      model
    }
  }
}
```

Response:
```
{
  "data": {
    "stat": {
      "id": "/api/stats/22",
      "_id": 22,
      "horsepower": "250",
      "vehicle": {
        "id": "/api/vehicles/1",
        "_id": 1,
        "make": "Toyota",
        "model": "Supra"
      }
    }
  }
}
```

The API works perfectly in GraphQL which makes me assume my entities are setup correctly but I don't know why JSON-API isnt working.

I briefly looked through the normaliser and couldn't find any reference to "included"

Perhaps i'm doing something wrong?

antograssiot:
You likely need to use normalization groups to get related object attributes.
I suggest you read https://api-platform.com/docs/core/serialization/#embedding-relations

You can check in the core tests entity to get hint if you're struggling.
For example usnig existing entities I got this test to run which looks similar to what you're expecting

```
  @createSchema
  Scenario: Correctly serialize a collection
    Given there are people having pets
[...]

PR #1949, opened 2018-05-14
JSON-API - Add support to inclusion of related resources

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1934
| License       | MIT
| Doc PR        | 

Here is my first trial at implementing support for [inclusion of related resource](http://jsonapi.org/format/#fetching-includes)

It surely can be improved, thanks by advance for your feedback and tests maybe (ping @Giant-Jelly, @jtparrett )


Giant-Jelly:
image
So close!

Giant-Jelly:
Is there any way this can get more support on this or perhaps skip the 0.01% of missing coverage?

I don't understand why this isnt being treated with much priority by more contributors as it is standard functionality of JSON-API.

Without being able to return an "included" object in the response i would go as far as say that api-platform doesnt actually support JSON-API, its even in the example on the homepage of the specification - http://jsonapi.org/

Is there anything i can do to try to raise this further?

dunglas:
Don't bother with a such small coverage issue. It's only and indicator for us, it's ok as is.

> I don't understand why this isnt being treated with much priority by more contributors as it is standard functionality of JSON-API.

Because most of us are very busy right now I guess. We focus on the work we're paid for first.
But we'll try to review this ASAP, we consider it as a must have feature.

> Without being able to return an "included" object in the response i would go as far as say that api-platform doesnt actually support JSON-API.

The JSONAPI spec explicitly states that this feature is optional:

http://jsonapi.org/format/#fetching-includes

> An endpoint MAY return resources related to the primary data by default.
> An endpoint MAY also support an include request parameter to allow the client to customize which related resources should be returned.
> **If an endpoint does not support the include parameter, it MUST respond with 400 Bad Request to any requests that include it.**

> Is there anything i can do to try to raise this further?

Reviewing and testing it on real projects will definitely help! Thanks for that.

Giant-Jelly:
My apologies, I didnt notice it said that it was an optional feature, i assumed as it was on the homepage that it was expected. 

Thank you for letting me know, i appreciate that you guys are busy and im looking forward to when this gets merged.

I plan to use API-Platform in all my future API projects (cause its awesome) so hopefully can test it out as much as possible. 😄

features/bootstrap/FeatureContext.php
```diff
@@ -218,6 +218,39 @@ public function thereAreDummyPropertyObjects(int $nb)
...
+                $dummyGroup = new DummyGroup();
+
+                foreach (['foo', 'bar', 'baz'] as $property) {
+                    $dummyGroup->$property = ucfirst($property).' #'.$i.$j;
+                }
+
+                $dummyProperty->groups[] = $dummyGroup;
+
+                $this->manager->persist($dummyGroup);
+            }
```
soyuka: would put newline after bracked, and would remove empty lines between 244 and 246

antograssiot:
please note that I just develop it based on the tests I wrote as I don't use JSON-API on my projects so I would too be glad and feels more confident if it could be tested on real apps before getting merged, as some people were asking for, I hope they will come back with feedback

Giant-Jelly:
Im more than happy to test on some of my existing apps, I have been using json-api in this way for a while now.

Im not completely sure how to use the commit in this PR with my symfony4 app though as i just install api-platform with composer normally. 

Is there any simple way?

src/JsonApi/EventListener/TransformFieldsetsParametersListener.php
```diff
@@ -35,23 +36,47 @@ public function onKernelRequest(GetResponseEvent $event)
...
             'jsonapi' !== $request->getRequestFormat() ||
             !($resourceClass = $request->attributes->get('_api_resource_class')) ||
-            !($fieldsParameter = $request->query->get('fields')) ||
-            !\is_array($fieldsParameter)
+            (!($fieldsParameter = $request->query->get('fields')) && !$includeParameter)
+        ) {
+            return;
+        }
+
+        if ($includeParameter && $request->attributes->get('_api_collection_operation_name')) {
```
Giant-Jelly: Can you explain why this exception is thrown here?
I think you should be able to include related resources for collections as seen in the example here - http://jsonapi.org/

features/jsonapi/filtering.feature
```diff
@@ -23,6 +23,18 @@ Feature: JSON API filter handling
     And the JSON should be valid according to the JSON API schema
     And the JSON node "data" should have 0 elements
 
+  Scenario: Apply property filter based on the 'fields'
```
Giant-Jelly: I couldnt get the filters to work on my live example although they work in your tests, did you have any particular grouping in the entities or something to get this to work?

antograssiot: Did you declare a [Property Filter](https://api-platform.com/docs/core/filters#property-filter) on your entity ?

src/JsonApi/EventListener/TransformFieldsetsParametersListener.php
antograssiot: All example that I saw in the spec were using item only.
I'll update soon

antograssiot: @Giant-Jelly It should now work also for collection, please give it a try

Giant-Jelly:
I can confirm through using this PR on my own project that all the inclusions are working perfectly in every case i could find. Also the property filter works exactly as on the json-api spec.

Thank you so much for adding this in @antograssiot, this should make the json-api implementation much nicer for other users and i will definitely be using this in my upcoming projects

src/JsonApi/Serializer/CollectionNormalizer.php
```diff
@@ -88,6 +88,10 @@ protected function getItemsData($object, string $format = null, array $context =
             }
 
             $data['data'][] = $item['data'];
+
+            if (isset($item['included'])) {
+                $data['included'] = array_merge($data['included'] ?? [], $item['included']);
```
dunglas: `$data['included']` is always defined because of the isset the line before, `??` can be removed.

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -323,14 +345,69 @@ private function getPopulatedRelations($object, string $format = null, array $co
...
             }
         }
 
         return $data;
     }
 
+    /**
+     * Populates relationships keys.
+     *
+     * @param object      $object
```
dunglas: All the `param` annotations and the `return` one can be removed (useless, already documented by the typehints)

src/EventListener/SerializeListener.php
```diff
@@ -60,6 +60,9 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
         }
 
         $context = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
+        if ($includes = $request->attributes->get('_api_includes')) {
```
dunglas: Why include**s** here and include**d** the line below?

antograssiot: I've fixed the inconsistency and went with `included` everywhere 👍

src/JsonApi/Serializer/ItemNormalizer.php
antograssiot: done

antograssiot:
I've rebased and updated with the last fixer rules. Anything else to be done on this one ?

tests/Fixtures/TestBundle/Entity/DummyProperty.php
```diff
@@ -83,6 +83,13 @@ class DummyProperty
      */
     public $group;
 
+    /**
+     * @var DummyGroup[]
```
dunglas: blank line to add after

dunglas:
Thanks!

Issue #1952, opened 2018-05-16
[GraphQL] Exceptions are returned as HTTP 400 instead of 500


bwegrzyn:
If an exception occurs during a GraphQL request, the response is HTTP 400 with content similar to this:

```json
HTTP 400
{
   "errors":[
      {
         "debugMessage":"wtf",
         "message":"Internal server error",
         "category":"internal",
         "locations":[
            {
               "line":2,
               "column":3
            }
         ],
         "path":[
            "updateUser"
         ]
      }
   ],
   "data":{
      "updateUser":null
   }
}
```

This is quite confusing as general the UI would like to catch these types of errors and display some sort of global notification. Validation errors also produce a similar HTTP 400, so the only way to tell that the response is an issue with the server is to go looking at the content. 

By contrast, the REST API returns an HTTP 500 for exceptions so it is easy to distinguish by the client.

alanpoulain:
It is intended: https://github.com/api-platform/core/blob/master/src/GraphQl/Action/EntrypointAction.php#L70
I think we should even set a 200 instead of a 400. GraphQL is not related to HTTP like REST.

teohhanhui:
Thanks for showing me again why GraphQL is evil. :laughing:

bwegrzyn:
@alanpoulain I would expect a 200 response if this is how errors are being handled. 

I guess I will need to rethink my error handling strategy for graphql calls. Thanks everyone for the input.

alanpoulain:
@bwegrzyn I've done a PR to avoid confusion for developers by always sending a 200, even when there are errors.

PR #1954, opened 2018-05-17
[GraphQL] Fix response status code

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1952
| License       | MIT
| Doc PR        |

As discussed in #1952, the GraphQL entrypoint should always returns a 200 because GraphQL should not use HTTP status codes.

### DIFF
--- a/src/GraphQl/Action/EntrypointAction.php
+++ b/src/GraphQl/Action/EntrypointAction.php
@@ -67,7 +67,7 @@ final class EntrypointAction
             $executionResult = new ExecutionResult(null, [$e]);
         }
 
-        return new JsonResponse($executionResult->toArray($this->debug), $executionResult->errors ? Response::HTTP_BAD_REQUEST : Response::HTTP_OK);
+        return new JsonResponse($executionResult->toArray($this->debug));
     }
 
     private function parseRequest(Request $request): array
--- a/features/graphql/authorization.feature
+++ b/features/graphql/authorization.feature
@@ -15,7 +15,7 @@ Feature: Authorization checking
       }
     }
     """
-    Then the response status code should be 400
+    Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "errors[0].message" should be equal to "Access Denied."
@@ -35,7 +35,7 @@ Feature: Authorization checking
       }
     }
     """
-    Then the response status code should be 400
+    Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "errors[0].message" should be equal to "Access Denied."
@@ -51,7 +51,7 @@ Feature: Authorization checking
       }
     }
     """
-    Then the response status code should be 400
+    Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "errors[0].message" should be equal to "Access Denied."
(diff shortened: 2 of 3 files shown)


features/graphql/introspection.feature
```diff
@@ -3,15 +3,13 @@ Feature: GraphQL introspection support
   @createSchema
   Scenario: Execute an empty GraphQL query
     When I send a "GET" request to "/graphql"
-    Then the response status code should be 400
```
dunglas: Maybe can you check that the status is 200, to prevent a future regression?

src/GraphQl/Action/EntrypointAction.php
```diff
@@ -54,11 +54,11 @@ public function __invoke(Request $request): Response
         list($query, $operation, $variables) = $this->parseRequest($request);
 
         if (null === $query) {
-            return new JsonResponse(new ExecutionResult(null, [new Error('GraphQL query is not valid')]), Response::HTTP_BAD_REQUEST);
```
dunglas: Aren't this two ones legit? This time we're sure it's a bad request.

src/GraphQl/Action/EntrypointAction.php
```diff
@@ -67,7 +67,7 @@ public function __invoke(Request $request): Response
             $executionResult = new ExecutionResult(null, [$e]);
         }
 
-        return new JsonResponse($executionResult->toArray($this->debug), $executionResult->errors ? Response::HTTP_BAD_REQUEST : Response::HTTP_OK);
```
dunglas: I would just keep this change. (It looks like Apollo behave in a similar way).

Simperfit: If Apollo behave in a similar way, IMO we should keep it.

src/GraphQl/Action/EntrypointAction.php
dunglas: When there is no "mixed" result, we can safely return a 400. It's the case here.

alanpoulain: I don't agree. @bwegrzyn thought he could use status code whereas it was only for some cases. This is very misleading and I think we should not use them at all.

src/GraphQl/Action/EntrypointAction.php
alanpoulain: I think their choices are misleading for the GraphQL ecosystem.

src/GraphQl/Action/EntrypointAction.php
dunglas: From an HTTP POV, Apollo choice makes sens:

- the HTTP request is entirely invalid => 400
- the HTTP request is partially valid, but for some instructions an error is generated => 200

alanpoulain:
I've reverted some changes.

alanpoulain:
Done.

alanpoulain:
Are you sure? It's a breaking change if the client relied on the status code to know if there were errors.

dunglas:
Yes sure, it's a bug fix to mimic the reference implementation reference (BC breaks aren't allowed in master anyway). And GraphQL support is still marked as experimental.

alanpoulain:
Rebased.

soyuka:
test failure doesn't seem related

alanpoulain:
I think we could merge this PR.

soyuka:
thanks @alanpoulain

PR #1957, opened 2018-05-19
Fixes ignoring type checks during denormalization

IckleChris:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a (already explained)

This PR updates the denormalization process to honour the [ObjectNormalizer::DISABLE_TYPE_ENFORCEMENT](http://symfony.com/doc/current/components/serializer.html#recursive-denormalization-and-type-safety) flag, allowing invalid types to be handled by the validation layer instead of endpoints returning (by default) 500 status serialization errors.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -336,9 +336,11 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return $this->denormalizeRelation($attribute, $propertyMetadata, $className, $value, $format, $this->createChildContext($context, $attribute));
         }
 
-        if (empty($context[static::DISABLE_TYPE_ENFORCEMENT])) {
-            $this->validateType($attribute, $type, $value, $format);
+        if ($context[static::DISABLE_TYPE_ENFORCEMENT] ?? false) {
+            return $value;
         }
+
+        $this->validateType($attribute, $type, $value, $format);
 
         return $value;
     }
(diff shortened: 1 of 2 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -201,7 +201,10 @@ protected function setAttributeValue($object, $attribute, $value, $format = null
             return;
         }
 
-        $this->validateType($attribute, $type, $value, $format);
+        if (empty($context[AbstractObjectNormalizer::DISABLE_TYPE_ENFORCEMENT])) {
```
IckleChris: Looks like the usage of the constant from the Serializer is causing failing tests, should the constant be included within this class (despite the duplication)?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -336,7 +336,9 @@ private function createAttributeValue($attribute, $value, $format = null, array
             return $this->denormalizeRelation($attribute, $propertyMetadata, $className, $value, $format, $this->createChildContext($context, $attribute));
         }
 
-        $this->validateType($attribute, $type, $value, $format);
+        if (empty($context[static::DISABLE_TYPE_ENFORCEMENT])) {
```
alanpoulain: I would prefer an early return before, with:
```php
        if ($context[static::DISABLE_TYPE_ENFORCEMENT] ?? false) {
            return $value;
        }
```

soyuka:
Thanks @IckleChris !

PR #1960, opened 2018-05-22
[WIP] Allow oauth2 implicit flow in swagger ui.

michaelpetri:
Add missing oauth2-redirect.html for swagger-ui and provide api_platform.oauth.redirectUrl configuration to implement custom oauth2 client endpoint for swagger-ui.

This PR still needs work before reviewing! I just want share my current work in case of helping someone else. I will start work on this after finishing my other PR. 

Todos:
- [ ] PHPUnit tests
- [ ] Behat tests
- [ ] Move asset path creation to tempalte
- [ ] Update update-js.sh
- [ ] Target master branch

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
+++ b/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\Documentation\Documentation;
 use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
+use Symfony\Component\Asset\Packages;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
@@ -34,6 +35,7 @@ final class SwaggerUiAction
     private $normalizer;
     private $twig;
     private $urlGenerator;
+    private $assetPackages;
     private $title;
     private $description;
     private $version;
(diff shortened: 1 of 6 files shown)


src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -95,6 +100,13 @@ private function getContext(Request $request, Documentation $documentation): arr
             'spec' => $this->normalizer->normalize($documentation, 'json', ['base_url' => $request->getBaseUrl()]),
         ];
 
+        if (!$this->oauthRedirectUrl) {
+            $this->oauthRedirectUrl = $request->getScheme() . '://' .
+                $request->getHttpHost() .
+                $request->getBasePath() .
+                $this->assetPackages->getUrl('bundles/apiplatform/swagger-ui/oauth2-redirect.html');
```
dunglas: I suggest to generate the URL directly in the Twig template using the helpers if not defined in the config

src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -45,15 +47,17 @@
...
     private $oauthAuthorizationUrl;
+    private $oauthRedirectUrl;
     private $oauthScopes;
 
-    public function __construct(ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, NormalizerInterface $normalizer, \Twig_Environment $twig, UrlGeneratorInterface $urlGenerator, string $title = '', string $description = '', string $version = '', array $formats = [], $oauthEnabled = false, $oauthClientId = '', $oauthClientSecret = '', $oauthType = '', $oauthFlow = '', $oauthTokenUrl = '', $oauthAuthorizationUrl = '', $oauthScopes = [])
+    public function __construct(ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, NormalizerInterface $normalizer, \Twig_Environment $twig, UrlGeneratorInterface $urlGenerator, Packages $assetPackages, string $title = '', string $description = '', string $version = '', array $formats = [], $oauthEnabled = false, $oauthClientId = '', $oauthClientSecret = '', $oauthType = '', $oauthFlow = '', $oauthTokenUrl = '', $oauthAuthorizationUrl = '', $oauthRedirectUrl = NULL, $oauthScopes = [])
```
dunglas: This is a BC break, the new parameter must be the last one, and optional (but it would be better to use the Twig helper, see below)

src/Bridge/Symfony/Bundle/Resources/public/swagger-ui/oauth2-redirect.html
```diff
@@ -0,0 +1,67 @@
+<!doctype html>
```
dunglas: The `update-js.sh` script must be upgraded to copy the new version of this file every time we'll bump Swagger UI: https://github.com/api-platform/core/blob/master/update-js.sh#L12

dunglas:
Can you also target `master` as it's a new feature. Thank you very much for handling this!

src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
michaelpetri: Do you mean in your index template `src/Bridge/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig` ? 

I'm not sure if this is such a good idea, because anyone who overwrites the template without extending it has to take care of a default value, or am i missing something?

If you mean i should create a new twig template for the swagger-ui's helper (`oauth2-redirect.html`), that would mean i would have also to create a controller/route for it and we also would no longer simply copy the original file and would have to transfer any further changes made by the swagger team manually into the template. Not sure if there will be any changes soon, just a thought of me.

dunglas: > Do you mean in your index template `src/Bridge/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig` ?

Yes. I would just include `{{ asset('bundles/apiplatform/swagger-ui/oauth2-redirect.html' }}`.
It's cleaner (no need to create a dependency to the Assets component in the controller), and not a big deal to copy/paste if needed.

dunglas: If an absolute URL is mandatory: `{{ absolute_url(asset('bundles/apiplatform/swagger-ui/oauth2-redirect.html')) }}`

flug:
@michaelpetri Up is this pull request still valid?

teohhanhui:
Already fixed in https://github.com/api-platform/core/pull/3134

PR #1961, opened 2018-05-22
Fix #1959 do not join same association twice

### LINKED ISSUE #1959: Invalid DQL when using selfreferenced entity
I found the following part of code in EagerLoadingExtension

```
            } else {
                $queryBuilder->addSelect($associationAlias);
            }

            // Avoid recursion
            if ($mapping['targetEntity'] === $resourceClass) {
                $queryBuilder->addSelect($associationAlias);
                continue;
            }
```
The problem is that we add the same alias to select for the second time. It confuses Doctrine and can work only occasionally. 

The part "Avoid recursion" was added in commit https://github.com/api-platform/core/commit/5ba518014e2770e1ad5686b0124b2db45245fee5#diff-0e280fa016b79eacb061c504a6348752 with is marked as "Temporary solution". I do not know, what exactly was solved, but it seems very strange.

Below is my code, that is failed because of this.

```
<?php

namespace App\Entity;

/**
 * @ORM\Entity()
 * @ORM\InheritanceType("SINGLE_TABLE")
 * @ORM\DiscriminatorColumn(name="Type", type="string")
 * @ORM\DiscriminatorMap({"C" = "Category"})
 */
abstract class CategoryBase
{
    /**
     * @var int
     *
     * @ORM\Id
     * @Groups({"category"})
     */
    public $id;
}

////////////////

<?php

namespace App\Entity;

/**
 * Group
 *
 * @ORM\Entity
 * @ApiResource(
 *     attributes={ "normalization_context"= {"groups"={"category"} }}
 * )
 */
class Category extends CategoryBase
{
    /**
     * @var Category[]
     * @ORM\OneToMany(targetEntity="Category", mappedBy="parentCategory")
     * @ORM\JoinColumn(referencedColumnName="parentId")
     * @Groups({"category"})
     */
    public $subcategories;

    /**
     * @var Group
     * @ORM\ManyToOne(targetEntity="Category", inversedBy="subcategories")
     * @ORM\JoinColumn(name="ParentId", referencedColumnName="id")
     */
    public $parentCategory;
}
```
I tryed to fetch /category/1 , that does not have children and got  `The discriminator column \"Type\" is missing for \"App\\Entity\\Category\" using the DQL alias \"subcategories_a1\"."`
It is long to explain, but the real problem is described above. The generated DQL is
```
SELECT o, subcategories_a1, subcategories_a1 FROM App\Entity\Category o LEFT JOIN o.subcategories subcategories_a1 WHERE o.id = :id_id
```

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1959 
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -200,9 +200,13 @@ final class EagerLoadingExtension implements ContextAwareQueryCollectionExtensio
                 $queryBuilder->addSelect($associationAlias);
             }
 
-            // Avoid recursion
+            // Avoid recursive joins
             if ($mapping['targetEntity'] === $resourceClass) {
-                $queryBuilder->addSelect($associationAlias);
+                // Avoid joining the same association twice (see #1959)
+                if (!\in_array($associationAlias, $queryBuilder->getAllAliases(), true)) {
+                    $queryBuilder->addSelect($associationAlias);
+                }
+
                 continue;
             }
 
--- a/features/main/circular_reference.feature
+++ b/features/main/circular_reference.feature
@@ -61,10 +61,29 @@ Feature: Circular references handling
           "@type": "CircularReference",
           "parent": "/circular_references/1",
           "children": [
-              "/circular_references/1",
-              "/circular_references/2"
+              "/circular_references/1"
           ]
       },
       "children": []
+    }
+    """
+    And I send a "GET" request to "/circular_references/1"
+    Then the response status code should be 200
+    And the JSON should be equal to:
+    """
+    {
+        "@context": "/contexts/CircularReference",
+        "@id": "/circular_references/1",
+        "@type": "CircularReference",
+        "parent": "/circular_references/1",
+        "children": [
+            "/circular_references/1",
+            {
+                "@id": "/circular_references/2",
+                "@type": "CircularReference",
+                "parent": "/circular_references/1",
+                "children": []
+            }
+        ]
     }
     """


src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -200,9 +200,13 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
                 $queryBuilder->addSelect($associationAlias);
             }
 
-            // Avoid recursion
+            // Avoid recursion joins
             if ($mapping['targetEntity'] === $resourceClass) {
-                $queryBuilder->addSelect($associationAlias);
+                // Avoid joining the same association twice (see #1959)
+                if (!in_array($associationAlias, $queryBuilder->getAllAliases(), true)) {
```
Simperfit: can't we use \in_array().

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -200,9 +200,13 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
                 $queryBuilder->addSelect($associationAlias);
             }
 
-            // Avoid recursion
+            // Avoid recursion joins
```
teohhanhui: This is confusing... What are "recursion joins"?

soyuka: `Avoid recursive joins`, misspell :D

PR #1963, opened 2018-05-22
Add resource fallback for ApiProperty

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n:a
| License       | MIT
| Doc PR        | n/a

Same as #1788 but for `@ApiProperty`. Also add support for some missing attributes in `@ApiResource`.

### DIFF
--- a/src/Annotation/ApiProperty.php
+++ b/src/Annotation/ApiProperty.php
@@ -29,9 +29,7 @@ namespace ApiPlatform\Core\Annotation;
  */
 final class ApiProperty
 {
-    use AttributesHydratorTrait {
-        hydrateAttributes as public __construct;
-    }
+    use AttributesHydratorTrait;
 
     /**
      * @var string
@@ -100,4 +98,12 @@ final class ApiProperty
      * @var array
      */
     private $swaggerContext;
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    public function __construct(array $values = [])
+    {
+        $this->hydrateAttributes($values);
+    }
 }
--- a/src/Annotation/ApiProperty.php
+++ b/src/Annotation/ApiProperty.php
@@ -21,7 +21,7 @@ namespace ApiPlatform\Core\Annotation;
  * @Annotation
  * @Target({"METHOD", "PROPERTY"})
  * @Attributes(
- *     @Attribute("fetcheable", type="bool"),
+ *     @Attribute("fetchable", type="bool"),
  *     @Attribute("fetchEager", type="bool"),
  *     @Attribute("jsonldContext", type="array"),
  *     @Attribute("swaggerContext", type="array")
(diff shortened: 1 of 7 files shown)


src/Annotation/ApiProperty.php
```diff
@@ -20,9 +20,17 @@
  *
  * @Annotation
  * @Target({"METHOD", "PROPERTY"})
+ * @Attributes(
+ *     @Attribute("fetcheable", type="bool"),
```
meyerbaptiste:
```diff
-fetcheable
+fetchable
```

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -170,7 +170,8 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
                 $inAttributes = null;
             }
 
-            if (false === $fetchEager = $propertyMetadata->getAttribute('fetchEager')) {
+            // Calls to getAttribute are nested for compatibility with the old camelCased notation
+            if (false === $fetchEager = $propertyMetadata->getAttribute('fetchEager', $propertyMetadata->getAttribute('fetch_eager'))) {
```
meyerbaptiste: Should we add a deprecated notice if `$propertyMetadata->getAttribute('fetch_eager')` is not null?

src/Annotation/AttributesHydratorTrait.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    /**
+     * @var array
+     */
+    public $attributes = [];
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    public function __construct(array $values = [])
```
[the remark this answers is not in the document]
dunglas: Because it's one more function call and more one level of indirection for no benefit. It's unusual to add a constructor in a trait, but for an annotation like this it's legit. And it's, why this trait is marked `@internal`.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: It would force every user already using this option to change the naming, IMO it is not worth it.

src/Annotation/AttributesHydratorTrait.php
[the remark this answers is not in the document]
dunglas: @Simperfit do you prefer the strategy in the last commit? It is more explicit, SOLID and don't add an extra call.

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
Pierstoval: If it's a deprecation, it's not an enforcement 😉

dunglas: In 3.0 it will. But why not.

PR #1967, opened 2018-05-23
Allow DataPersister to return a new instance instead of mutating

### LINKED ISSUE #1799: API Platform does not play well with immutable resource
Hi :wave: 

As far as I understand, `DataPersisterInterface::persist: void` expect the resource to be mutated (at least with autocomputed identifier, the created identifier is expected).

It works great with system like doctrine, but if one is dealing with a persistence layer with some kind of immutable entities, he’s screwed (sorry, my vocabulary lacks of a less familiar term :'( ).

I don’t see the point to tight an object reference to the resource’s data.

I would rather have `DataPersisterInterface::persist` returning an object of the same resource class. The returned object would then be normalized, serialized, etc as usual.

The conceptual signature then should be something like `DataPersistierInterface::persist<T>(T $resourceClasse): T`

I don’t think this will be a huge impact on performance.

The trick may be to prevent a BC break. While don’t BC breaking the doctrine bridge will be easy, implementation in the wild could be impacted.
 
What do you think?

https://github.com/api-platform/core/blob/d2b657e2798c724ade0bd1bcd3bd089f7f143922/src/DataPersister/DataPersisterInterface.php#L37

gorghoa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | *unit tests are green, bdd are red, master too seems wrecked 🎶*
| Fixed tickets | #1799 
| License       | MIT
| Doc PR        | @todo

closes #1799

`DataPersisterInterface` can now return the result of the persistence call instead of relying on data mutation. Thus allowing persistence layer to deal with immutable data.

At least, this PR breaks the technical detail that was implicitly using objects instance reference as an identifier for an entity, and conceptually I find this a huge win :)

### DIFF
--- a/src/Bridge/Doctrine/EventListener/WriteListener.php
+++ b/src/Bridge/Doctrine/EventListener/WriteListener.php
@@ -58,7 +58,9 @@ final class WriteListener
 
         switch ($request->getMethod()) {
             case 'POST':
-                $objectManager->persist($controllerResult);
+                $event->setControllerResult(
+                    $objectManager->persist($controllerResult) ?? $controllerResult
+                );
                 break;
             case 'DELETE':
                 $objectManager->remove($controllerResult);
--- a/src/Bridge/Doctrine/EventListener/WriteListener.php
+++ b/src/Bridge/Doctrine/EventListener/WriteListener.php
@@ -58,9 +58,7 @@ final class WriteListener
 
         switch ($request->getMethod()) {
             case 'POST':
-                $event->setControllerResult(
-                    $objectManager->persist($controllerResult) ?? $controllerResult
-                );
+                $objectManager->persist($controllerResult);
                 break;
             case 'DELETE':
                 $objectManager->remove($controllerResult);
--- a/src/DataPersister/ChainDataPersister.php
+++ b/src/DataPersister/ChainDataPersister.php
@@ -51,7 +51,7 @@ final class ChainDataPersister implements DataPersisterInterface
     {
         foreach ($this->persisters as $persister) {
             if ($persister->supports($data)) {
-                $persister->persist($data);
+                return $persister->persist($data) ?? $data;
             }
         }
     }
--- a/src/DataPersister/DataPersisterInterface.php
+++ b/src/DataPersister/DataPersisterInterface.php
@@ -33,6 +33,8 @@ interface DataPersisterInterface
      * Persists the data.
      *
      * @param mixed $data
+     *
+     * @return object|void Void will not be supported in API Platform 3, an object should always be returned
      */
     public function persist($data);
 
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -50,10 +50,13 @@ class WriteListener
             case 'PUT':
             case 'PATCH':
             case 'POST':
-                $event->setControllerResult(
-                    $this->dataPersister->persist($controllerResult)
-                    ?? $controllerResult
-                );
+                $persistResult = $this->dataPersister->persist($controllerResult);
+
+                if (null === $persistResult) {
+                    @trigger_error(sprintf('Returning void from %s::persist() is deprecated since API Platform 2.3 and will not be supported in API Platform 3, an object should always be returned.', DataPersisterInterface::class), E_USER_DEPRECATED);
+                }
+
+                $event->setControllerResult($persistResult ?? $controllerResult);
                 break;
             case 'DELETE':
                 $this->dataPersister->remove($controllerResult);
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -50,7 +50,10 @@ class WriteListener
             case 'PUT':
             case 'PATCH':
             case 'POST':
-                $this->dataPersister->persist($controllerResult);
+                $event->setControllerResult(
+                    $this->dataPersister->persist($controllerResult)
+                    ?? $controllerResult
+                );
                 break;
             case 'DELETE':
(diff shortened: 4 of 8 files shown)


gorghoa:
I hope, I have not forgotten any `DataPersisterInterface` implementation in the code base. Any hint on some part of the code I may have missed is welcome!

src/Bridge/Doctrine/EventListener/WriteListener.php
```diff
@@ -58,7 +58,9 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
         switch ($request->getMethod()) {
             case 'POST':
-                $objectManager->persist($controllerResult);
+                $event->setControllerResult(
```
teohhanhui: CS: We usually don't split into multiple lines like this... Just keep it in one line.

src/EventListener/WriteListener.php
```diff
@@ -50,7 +50,10 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             case 'PUT':
             case 'PATCH':
             case 'POST':
-                $this->dataPersister->persist($controllerResult);
+                $event->setControllerResult(
```
teohhanhui: CS: We usually don't split into multiple lines like this... Just keep it in one line.

src/DataPersister/DataPersisterInterface.php
```diff
@@ -33,6 +33,8 @@ public function supports($data): bool;
      * Persists the data.
      *
      * @param mixed $data
+     *
+     * @return object|void Void will not be supported in API Platform 3, an object should always be returned
```
teohhanhui: I think it might be better if we add deprecation errors where this method is called in API Platform?

src/Bridge/Doctrine/EventListener/WriteListener.php
gorghoa: I won’t argue on CS, but in this case, given the low visual impact of `??`, I fear that relegating the `??` at the end of the line somehow hide what’s going on. 

For instance, to me, this is far less self-explanatory than the splitted version.
 
```
    case 'POST':
                $event->setControllerResult($this->dataPersister->persist($controllerResult) ?? $controllerResult));
                break;
```

As I said, this will be my only point of view on CS, I wont argue furthermore if you want to stick to the one line version ;)

gorghoa: nevermind, adding the deprecation notice forces me to assign persist result to a temp var anyway, so the one line will be of fine size :+1:

gorghoa:
Thanks for your review @teohhanhui :+1: , code updated :)

tests/EventListener/WriteListenerTest.php
```diff
@@ -49,6 +80,46 @@ public function testOnKernelViewWithControllerResultAndPersist()
...
+    /**
+     * @see https://github.com/api-platform/core/issues/1799
+     */
+    public function testOnKernelViewWithControllerResultAndPersistWithImmutableResource()
+    {
+        $dummy = new Dummy();
+        $dummy->setName('Dummyrino');
+
+        $dummy2 = new Dummy();
+        $dummy2->setName('Dummyferoce');
```
soyuka: `Dummycorn` :unicorn: ?

src/DataPersister/DataPersisterInterface.php
dunglas: Doing it in `ChainDataPersister` should be enough. If we deprecate returning `void`, please remove it from the supported types and only keep it in the comment.

dunglas: Ok you already made the changes, so keep them :)

teohhanhui: > Doing it in ChainDataPersister should be enough.

It's not enough because `ChainDataPersister` is only an implementation which might or might not be used by the user.

teohhanhui:
@gorghoa Looks like we need a rebase to merge. :+1:

teohhanhui:
Thanks @gorghoa! :tada:

gorghoa:
Thank you all for your kind reviews and making api platform always better! :muscle:

src/DataPersister/ChainDataPersister.php
```diff
@@ -51,7 +51,7 @@ public function persist($data)
     {
         foreach ($this->persisters as $persister) {
             if ($persister->supports($data)) {
-                $persister->persist($data);
+                return $persister->persist($data) ?? $data;
```
Taluu: I know it's a bit late, but wouldn't this prevent _all_ persisters to run, and wouldn't this stop at the first persister that supports this data ? what if I need to apply multiple data persisters ?

dunglas: Good catch... We may return after the loop.instead of inside. WYT? (ping @gorghoa)

gorghoa: indeed, nice catch…

> We may return after the loop instead of inside. WYT?

But from which persister shall we return? First one? Last one?  (somehow, this question is in fact valid for the former implementation. The last was the winner, but was it “safe” to assume it was the one that should win?).

 Maybe the best option in term of confidence should be not return data from persister but somehow re-trigger a read item operation. At the cost of an extra read query.

dunglas: > afterthoughts

It was exactly what i've in mind (last win)

gorghoa: https://github.com/api-platform/core/pull/2064

PR #1968, opened 2018-05-24
Fix embed

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

@dunglas backported your quality fixes as well if that's okay.

There was a bug in eager loading when groups where used on embedded entities. Indeed, we specifically selected the embedded field (ex: `embed`) instead of selecting the embedded ones (ex: `embed.foo, embed.bar`) which lead to a query parser bug.

### DIFF
--- a/features/main/operation.feature
+++ b/features/main/operation.feature
@@ -12,3 +12,24 @@ Feature: Operation support
     """
     "This is a custom action for 42."
     """
+
+  @createSchema
+  @dropSchema
+  Scenario: Select a resource and it's embedded data
+    Given there are 1 embedded dummy objects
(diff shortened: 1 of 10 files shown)


dunglas:
For information, partial select will be not possible in Doctrine 3, so I'm not sure we need to still bother with this.

soyuka:
> For information, partial select will be not possible in Doctrine 3, so I'm not sure we need to still bother with this.

Duly noted. Not sure if `root.embeded` is prefered over `root.embeded.field` though.

PR #1970, opened 2018-05-24
Check the value of enable_max_depth if defined

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Because it confuses some users that `['enable_max_depth' => false]` actually triggers the check.


soyuka:
thanks @dunglas

PR #1971, opened 2018-05-24
Max Depth does not work with HAL

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not yet
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Using `@MaxDepth` does not work with hal+json.
This is because relations are normalized separately from the parent:
https://github.com/api-platform/core/blob/master/src/Hal/Serializer/ItemNormalizer.php#L55-L63
so the context of max depth isn't checked.

I haven't looked for a nice fix yet. suggestions welcome.


src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -65,6 +66,37 @@ public function normalize($object, $format = null, array $context = [])
...
+    /**
+     * Is the max depth reached for the given attribute?
+     *
+     * @param AttributeMetadataInterface[] $attributesMetadata
+     */
+    private function isMaxDepthReached(array $attributesMetadata, string $class, string $attribute, array &$context): bool
+    {
+        if (
+            !isset($context[static::ENABLE_MAX_DEPTH]) ||
+            !isset($attributesMetadata[$attribute]) ||
```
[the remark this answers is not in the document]
bendavies: done, although there will now be a behavioral difference between hal and the other formats.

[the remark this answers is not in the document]
dunglas: Until https://github.com/symfony/symfony/pull/27365 is merged

bendavies:
don't merge. there is a bug here.

bendavies:
@dunglas this is done.  I don't think the failure is due to the PR

tests/Hal/Serializer/ItemNormalizerTest.php
```diff
@@ -207,4 +213,133 @@ public function testNormalizeWithoutCache()
         ];
         $this->assertEquals($expected, $normalizer->normalize($dummy, null, ['not_serializable' => function () {}]));
     }
+
+    public function testMaxDepth()
+    {
+        $setId = function (MaxDepthDummy $dummy, int $id): void {
```
dunglas: `void` isn't supported in PHP 7.0. It's why Travis is red.

bendavies: fixed that, but there is a strange random behat failure.

bendavies:
@dunglas yep, random behat scenario passing on travis but failing on circleci.

dunglas:
It's not random, it uses an outdated version of the scenario: https://github.com/api-platform/core/blob/master/features/jsonld/context.feature#L21
Can you try to rebase and push force?

bendavies:
done, but i'm still not sure how it passed on travis and failed on circle 😆

dunglas:
Because Travis takes `origin/2.2` and merge your branch in why Circle only checkout your branch.

bendavies:
very good!

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -232,4 +240,36 @@ private function getHalCacheKey(string $format = null, array $context)
...
+
+    /**
+     * Is the max depth reached for the given attribute?
+     *
+     * @param AttributeMetadataInterface[] $attributesMetadata
+     */
+    private function isMaxDepthReached(array $attributesMetadata, string $class, string $attribute, array &$context): bool
+    {
+        if (
+            !isset($context[static::ENABLE_MAX_DEPTH]) ||
```
dunglas: Can be `($context[static::ENABLE_MAX_DEPTH] ?? false)` (not possible in Symfony because we still support PHP 5)/

bendavies: almost, it's `!($context[static::ENABLE_MAX_DEPTH] ?? false)`.
done

PR #1986, opened 2018-05-30
Make tests green with Symfony 4.1

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/features/bootstrap/JsonContext.php
+++ b/features/bootstrap/JsonContext.php
@@ -15,6 +15,7 @@ use Behat\Gherkin\Node\PyStringNode;
 use Behatch\Context\JsonContext as BaseJsonContext;
 use Behatch\HttpCall\HttpCallResultPool;
 use Behatch\Json\Json;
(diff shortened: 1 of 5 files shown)


features/bootstrap/JsonContext.php
```diff
@@ -15,6 +15,7 @@
 use Behatch\Context\JsonContext as BaseJsonContext;
 use Behatch\HttpCall\HttpCallResultPool;
 use Behatch\Json\Json;
+use PHPUnit\Framework\Assert;
```
teohhanhui: Do we really need to use PHPUnit classes in Behat? :rofl:

dunglas: 1. We already do that in some contests
2. I don't want to bother reimplementing something that already is in PHPUnit 😅

PR #1987, opened 2018-05-30
[GraphQL] Fix iterable type

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/704
| License       | MIT
| Doc PR        |

Fix the iterable type after a BC in graphql-php.
Maybe we need to force users to require a specific version of graphql-php in order to avoid this kind of issue.
For example in the doc, instead of:
`composer require webonyx/graphql-php`
we would have:
`composer require webonyx/graphql-php:^0.12.0`
WDYT?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "symfony/twig-bundle": "^3.1 || ^4.0",
         "symfony/validator": "^3.3 || ^4.0",
         "symfony/yaml": "^3.3 || ^4.0",
-        "webonyx/graphql-php": "^0.11.5"
+        "webonyx/graphql-php": ">=0.12 <1.0"
     },
     "conflict": {
         "symfony/dependency-injection": "<3.3"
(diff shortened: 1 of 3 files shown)


alanpoulain:
Errors seem not related.

dunglas:
About the documentation, it will be a pain to update... But maybe should we change the version constraint in `require-dev` to `<1.0` to catch those errors when new versions are tagged?

dunglas:
Can you rebase? Test should be green now.

dunglas:
done @alanpoulain

alanpoulain:
Rebase done.

soyuka:
still failing for low version IMO you should constraint to `>0.12.0<1.0.0` (not sure of the format)

dunglas:
Thanks @alanpoulain!

PR #1991, opened 2018-05-31
tests for: Make resource constructor parameters writables

komik966:
relations test for #1749


bendavies:
we should try get these extra test in, no?

dunglas:
@bendavies yes, why has it be closed?

komik966:
I'll open new PR with tests and fix

PR #1996, opened 2018-06-03
Fix fatal error when subresource type is not defined

oxan:
Currently, when a property has a @ApiSubresource declaration, but no proper type declaration, this results in a "Fatal error: Call to a member function isCollection() on null", which is hard to debug. Instead throw an exception referencing the offending property, so the declaration can be fixed.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
+++ b/src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Metadata\Property\Factory;
 
 use ApiPlatform\Core\Annotation\ApiSubresource;
+use ApiPlatform\Core\Exception\InvalidResourceException;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Metadata\Property\SubresourceMetadata;
 use ApiPlatform\Core\Util\Reflection;
@@ -52,7 +53,7 @@ final class AnnotationSubresourceMetadataFactory implements PropertyMetadataFact
             $annotation = $this->reader->getPropertyAnnotation($reflectionClass->getProperty($property), ApiSubresource::class);
 
             if (null !== $annotation) {
-                return $this->updateMetadata($annotation, $propertyMetadata, $resourceClass);
+                return $this->updateMetadata($annotation, $propertyMetadata, $resourceClass, $property);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
```diff
@@ -70,16 +71,18 @@ public function create(string $resourceClass, string $property, array $options =
...
         }
 
         return $propertyMetadata;
     }
 
-    private function updateMetadata(ApiSubresource $annotation, PropertyMetadata $propertyMetadata, string $originResourceClass): PropertyMetadata
+    private function updateMetadata(ApiSubresource $annotation, PropertyMetadata $propertyMetadata, string $originResourceClass, string $propertyName): PropertyMetadata
     {
         $type = $propertyMetadata->getType();
+        if(null === $type)
```
soyuka: Add brackets and spacing :)

soyuka:
May you add a quick unit test for the exception? Thanks!

src/Metadata/Property/Factory/AnnotationSubresourceMetadataFactory.php
```diff
@@ -70,16 +71,19 @@ public function create(string $resourceClass, string $property, array $options =
...
 
         return $propertyMetadata;
     }
 
-    private function updateMetadata(ApiSubresource $annotation, PropertyMetadata $propertyMetadata, string $originResourceClass): PropertyMetadata
+    private function updateMetadata(ApiSubresource $annotation, PropertyMetadata $propertyMetadata, string $originResourceClass, string $propertyName): PropertyMetadata
     {
         $type = $propertyMetadata->getType();
+        if (null === $type) {
+            throw new InvalidResourceException(sprintf('Property %s on resource %s has no type defined, but is declared as a subresource.', $propertyName, $originResourceClass));
```
teohhanhui: I think saying that it "has no type defined" is confusing... Perhaps we could say:

Property %s on resource %s is declared as a subresource, but its type could not be determined.

soyuka: I love your suggestion indeed, your english is too good haha.

oxan: Changed.

oxan:
@soyuka Unit test added :)

dunglas:
Thanks @oxan

PR #1999, opened 2018-06-04
Reading support for Elasticsearch

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ jobs:
   include:
     - php: '7.1'
     - php: '7.2'
+    - php: '7.2'
+      services:
+        - elasticsearch
+      env: APP_ENV=elasticsearch
     - php: '7.3'
     - php: '7.3'
       env: deps=low
@@ -31,6 +35,9 @@ jobs:
   fast_finish: true
 
 before_install:
+  - if [[ $APP_ENV = 'elasticsearch' ]]; then
+      curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.5.0.deb && sudo dpkg -i --force-confnew elasticsearch-6.5.0.deb && sudo service elasticsearch restart;
+    fi
   - phpenv config-rm xdebug.ini || echo "xdebug not available"
   - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - export PATH="$PATH:$HOME/.composer/vendor/bin"
@@ -48,6 +55,8 @@ script:
   - tests/Fixtures/app/console cache:clear
   - if [[ $APP_ENV = 'postgres' ]]; then
       vendor/bin/behat --suite=postgres --format=progress;
+    elif [[ $APP_ENV = 'elasticsearch' ]]; then
+      vendor/bin/behat --suite=elasticsearch --format=progress;
     else
       vendor/bin/behat --suite=default --format=progress;
     fi
--- a/features/bootstrap/ElasticsearchContext.php
+++ b/features/bootstrap/ElasticsearchContext.php
@@ -1,0 +1,142 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+use ApiPlatform\Core\Bridge\Elasticsearch\Metadata\Document\DocumentMetadata;
+use Behat\Behat\Context\Context;
+use Elasticsearch\Client;
+use Symfony\Component\Finder\Finder;
+
+/**
+ * @experimental
+ *
+ * @author Baptiste Meyer <baptiste.meyer@gmail.com>
+ */
+final class ElasticsearchContext implements Context
+{
+    private $client;
+    private $elasticsearchMappingsPath;
+    private $elasticsearchFixturesPath;
+
+    public function __construct(Client $client, string $elasticsearchMappingsPath, string $elasticsearchFixturesPath)
+    {
+        $this->client = $client;
+        $this->elasticsearchMappingsPath = $elasticsearchMappingsPath;
+        $this->elasticsearchFixturesPath = $elasticsearchFixturesPath;
+    }
+
+    /**
+     * @BeforeScenario
+     */
+    public function initializeElasticsearch(): void
+    {
+        static $initialized = false;
+
+        if ($initialized) {
+            return;
+        }
+
+        $this->deleteIndexes();
+        $this->createIndexesAndMappings();
+        $this->loadFixtures();
+
+        $initialized = true;
+    }
+
+    /**
+     * @Given indexes and their mappings are created
+     */
+    public function thereAreIndexes(): void
+    {
+        $this->createIndexesAndMappings();
+    }
+
+    /**
+     * @Given indexes are deleted
+     */
+    public function thereAreNoIndexes(): void
+    {
+        $this->deleteIndexes();
+    }
+
+    /**
+     * @Given fixtures files are loaded
+     */
+    public function thereAreFixtures(): void
+    {
+        $this->loadFixtures();
+    }
+
+    private function createIndexesAndMappings(): void
+    {
+        $finder = new Finder();
+        $finder->files()->in($this->elasticsearchMappingsPath);
+
+        foreach ($finder as $file) {
+            $this->client->indices()->create([
+                'index' => $file->getBasename('.json'),
+                'body' => json_decode($file->getContents(), true),
+            ]);
+        }
+    }
+
+    private function deleteIndexes(): void
+    {
+        $finder = new Finder();
+        $finder->files()->in($this->elasticsearchMappingsPath)->name('*.json');
+
+        $indexClient = $this->client->indices();
+
+        foreach ($finder as $file) {
+            $index = $file->getBasename('.json');
+            if (!$indexClient->exists(['index' => $index])) {
+                continue;
+            }
+
+            $indexClient->delete(['index' => $index]);
+        }
+    }
+
+    private function loadFixtures(): void
+    {
+        $finder = new Finder();
+        $finder->files()->in($this->elasticsearchFixturesPath)->name('*.json');
+
+        $indexClient = $this->client->indices();
+
+        foreach ($finder as $file) {
+            $index = $file->getBasename('.json');
+            $bulk = [];
+
+            foreach (json_decode($file->getContents(), true) as $document) {
+                if (null === $document['id'] ?? null) {
+                    $bulk[] = ['index' => ['_index' => $index, '_type' => DocumentMetadata::DEFAULT_TYPE]];
+                } else {
+                    $bulk[] = ['create' => ['_index' => $index, '_type' => DocumentMetadata::DEFAULT_TYPE, '_id' => (string) $document['id']]];
+                }
+
+                $bulk[] = $document;
+
+                if (0 === (\count($bulk) % 50)) {
+                    $this->client->bulk(['body' => $bulk]);
+                    $bulk = [];
+                }
+            }
+
+            if ($bulk) {
+                $this->client->bulk(['body' => $bulk]);
+            }
+
+            $indexClient->refresh(['index' => $index]);
+        }
+    }
+}
--- a/features/bootstrap/ElasticsearchContext.php
+++ b/features/bootstrap/ElasticsearchContext.php
@@ -126,7 +126,7 @@ final class ElasticsearchContext implements Context
 
                 $bulk[] = $document;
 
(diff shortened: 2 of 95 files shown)


src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -35,36 +36,64 @@
...
+                $option = null;
+                if (array_key_exists($i + 3, $args)) {
+                    if (!\call_user_func('is_'.$arg['type'], $args[$i + 3]) && !(null === $arg['default'] && null === $args[$i + 3])) {
+                        throw new InvalidArgumentException(sprintf('The "$%s" argument is expected to be a %s%s.', Inflector::camelize($arg['name']), $arg['type'], null === $arg['default'] ? ' or null' : ''));
+                    }
+
+                    $option = $args[$i + 3];
+                }
+
+                $options[$arg['name']] = $option ?? $arg['default'];
```
soyuka: :see_no_evil:

src/Bridge/Elasticsearch/DataProvider/Filter/AbstractFilter.php
```diff
@@ -0,0 +1,162 @@
...
+            if ($type->isCollection() && null === $type = $type->getCollectionValueType()) {
+                return array_fill(0, 4, null);
+            }
+
+            if (Type::BUILTIN_TYPE_ARRAY === $builtinType && Type::BUILTIN_TYPE_OBJECT !== $type->getBuiltinType()) {
+                if ($totalProperties === $index) {
+                    break;
+                }
+
+                return array_fill(0, 4, null);
```
soyuka: can't we declare a `$noop` array to avoid repeating this?

src/Bridge/Elasticsearch/DataProvider/Filter/TermFilter.php
```diff
@@ -0,0 +1,189 @@
...
+
+        foreach ($context['filters'] ?? [] as $property => $values) {
+            list($type, $hasAssociation, $nestedResourceClass, $nestedProperty) = $this->getMetadata($resourceClass, $property);
+
+            if (!$type || !$values = (array) $values) {
+                continue;
+            }
+
+            if ($hasAssociation || $this->isIdentifier($nestedResourceClass, $nestedProperty)) {
+                $values = array_map([$this, 'getIdentifierValue'], $values, array_fill(0, count($values), $nestedProperty));
```
soyuka: `\count`, I've seen other `count`.

Koc:
What about using Elastica? It provides better api for building query http://elastica.io/elastica-vs-elasticsearch-php/

dunglas:
@Koc we benchmarked both solutions, but Elastica brings has no added value in our case.

src/Bridge/Elasticsearch/DataProvider/Extension/SortExtension.php
```diff
@@ -0,0 +1,88 @@
...
+            foreach ($defaultOrder as $property => $direction) {
+                if (\is_int($property)) {
+                    $property = $direction;
+                    $direction = 'asc';
+                }
+
+                $orders[] = $this->getOrder($resourceClass, $property, $direction);
+            }
+        } elseif (null !== $this->defaultDirection) {
+            foreach ($this->identifiersExtractor->getIdentifiersFromResourceClass($resourceClass) as $identifier) {
```
epourail: You are assuming that an "identifier" property is always indexed. As, each document, in elasticsearch has the attribute `_id`, you can not be sure it is true.
The following use case fails : no "identifier" property indexed and get a collection with the default filter.  

You should not apply the default condition and let elasticsearch retrieve the collection sorted by `_id` 

However, as indicated in the ElasticSearch documentation, it is better to index the "identifier" field and duplicate the identifier content 
(see https://www.elastic.co/guide/en/elasticsearch/reference/6.3/mapping-id-field.html#CO257-1)

src/Bridge/Elasticsearch/DataProvider/Filter/TermFilter.php
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(array &$clauseBody, string $resourceClass, string $operationName = null, array $context): void
+    {
+        $terms = [];
+
+        foreach ($context['filters'] ?? [] as $property => $values) {
```
epourail: This is not consistent with the "doctrine" filters.
The context of a doctrine-filter service are declared like `['foo' => 'exact', 'bar' => '~']` but this implementation supports the following syntax `['foo', 'bar']`.

er1z:
Shouldn't this be published into separate bundle with API Platform as dependency? Core is getting more and more required packages whereas Elastic is not used everywhere.

dunglas:
@er1z, they are only dev packages, if you don't use it, you don't need the dep.

But, anyway, we plan to create a subtree split at some point, so we'll be able to cherry-pick only what we need.
Having a monolithic repository makes the maintenance easier (when our CI run, we test everything), and Elastic as well as MongoDB support are ones of the most frequent feature requests.

ro0NL:
@meyerbaptiste any chance the PR can be rebased? :) would like to try it out.

src/Bridge/Elasticsearch/DataProvider/Extension/SortExtension.php
ro0NL: > The following use case fails : no "identifier" property indexed and get a collection with the default filter.

IMHO that's by design, you get the `Composite identifiers not supported.` exception right...

Currently it's consistent with the doctrine bridge: apply default order on ID fields.

src/Bridge/Elasticsearch/DataProvider/Extension/SortExtension.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $requestBody['sort'] = array_merge_recursive($requestBody['sort'] ?? [], $orders);
+    }
+
+    private function getOrder(string $resourceClass, string $property, string $direction): array
+    {
+        $order = ['order' => strtolower($direction)];
+
+        if ($this->isNestedField($resourceClass, $nestedPath = explode('.', $property)[0])) {
+            $order['nested_path'] = $nestedPath;
```
ro0NL: `nested_path` is deprecated, right?

ro0NL:
I've created https://github.com/meyerbaptiste/core/pull/4 to fix tests for now.

So far it works pretty well, nice :) would be nice if;

- write ops are disabled by default
- snake_case document fields are converted to camelCase object properties during serialization
- a more convenient configuration IMO would be

```yaml
elasticsearch:
  host: elasticsearch:9200
  index: api_documents
  mapping:
    App\Api\SomeResource: ~ # resource class as type, default index
   ...: { index: custom_index, type: custom_type }
```

meyerbaptiste:
@ro0NL, I understand your point of view about the configuration. But we made the choice to only support a single index per document type (a resource in our case) because Elasticsearch is going to drop support for index types and will allow only a single type per index.

See https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html#_index_per_document_type.

src/Bridge/Elasticsearch/DataProvider/Extension/SortExtension.php
meyerbaptiste: Good catch, I'll replace it with the new `nested` sort option.

features/bootstrap/ElasticsearchContext.php
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+
+            if ($bulk) {
+                $this->client->bulk(['body' => $bulk]);
+            }
+        }
+
+        // Waiting for indexing...
+        sleep(5);
```
meyerbaptiste: I have to replace this by https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html.

.travis.yml
```diff
@@ -31,6 +35,9 @@ jobs:
   fast_finish: true
 
 before_install:
+  - if [[ $APP_ENV = 'elasticsearch' ]]; then
+      curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.5.0.deb && sudo dpkg -i --force-confnew elasticsearch-6.5.0.deb && sudo service elasticsearch restart;
```
meyerbaptiste: From SymfonyInsight:
> Unable to parse at line 39 (near " curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.5.0.deb && sudo dpkg -i --force-confnew elasticsearch-6.5.0.deb && sudo service elasticsearch restart;").

Any idea @api-platform/core-team?

dunglas: Ignore it. Insight isn’t good at parsing Travis files (Travis’s YAML parser is very tolerant)

teohhanhui: It's valid YAML. SymfonyInsight needs to fix their parser (Symfony YAML component?)

meyerbaptiste: I fixed it by creating a `.symfony.insight.yaml` file and by adding an exclude pattern.

.travis.yml
```diff
@@ -9,6 +9,10 @@ jobs:
   include:
     - php: '7.1'
     - php: '7.2'
+    - php: '7.2'
```
meyerbaptiste: https://twitter.com/meyer_baptiste/status/1076246168919859201

meyerbaptiste:
Ready for review @api-platform/core-team.

.travis.yml
teohhanhui: I don't think we should do that...

meyerbaptiste: I'm agree but since I don't have the rights to ignore the error, it was a quick way to do it...

teohhanhui: Did you try quoting the url? The parser might be confused by the `:` and trying to treat it as a key? Lol... (I'm not sure...)

dunglas: I'll ignore the error later, don't worry about that.

meyerbaptiste: > Did you try quoting the url?

It doesn't work.

The revert is done, you can ignore the error.

src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
```diff
@@ -0,0 +1,121 @@
...
+                    'match_all' => new \stdClass(),
+                ],
+            ]);
+        }
+
+        $limit = $body['size'] ?? $this->pagination->getLimit($resourceClass, $operationName);
+        $offset = $body['offset'] ?? $this->pagination->getOffset($resourceClass, $operationName);
+
+        if (!isset($body['size'])) {
+            $body = array_merge($body, ['size' => $limit]);
```
ro0NL: merges can be avoided like `$limit = $body['size'] = $body['size'] ?? 'val'`, or just `$body['size'] = $limit;` here :thinking:

meyerbaptiste: Wow! Where was my mind? All merges can be avoided in this method... Thx.

src/Bridge/Elasticsearch/Serializer/ItemNormalizer.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return self::FORMAT === $format && parent::supportsDenormalization($data, $type, $format);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        if (\is_string($data['_id'] ?? null) && \is_array($data['_source'] ?? null)) {
```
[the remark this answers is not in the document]
meyerbaptiste: Too complicated because of the `InnerFieldsNameConverter`.

src/DataProvider/Pagination.php
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @author Baptiste Meyer <baptiste.meyer@gmail.com>
+ */
+final class Pagination
+{
+    private $options;
+    private $requestStack;
+    private $resourceMetadataFactory;
+
+    public function __construct(RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $options = [])
```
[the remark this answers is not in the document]
meyerbaptiste: It will be done in other PRs as discussed.

er1z:
BTW, what version will be this issue merged to? Any changes for 2.x or does it break the APIs according to the semver?

meyerbaptiste:
@er1z this feature is planned for the next minor release (2.4).

er1z:
@meyerbaptiste — can't wait for; feel free to ask if it needs any help I can contribute to.

meyerbaptiste:
@er1z thanks, this PR is now ready but we will definitely need help to test the 2.4.0 beta 1 :relaxed:.

PR #2000, opened 2018-06-05
Improve max join limit error

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -122,7 +122,7 @@ final class EagerLoadingExtension implements ContextAwareQueryCollectionExtensio
     private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, bool $forceEager, bool $fetchPartial, string $parentAlias, array $options = [], array $normalizationContext = [], bool $wasLeftJoin = false, int &$joinCount = 0, int $currentDepth = null)
     {
         if ($joinCount > $this->maxJoins) {
-            throw new RuntimeException('The total number of joined relations has exceeded the specified maximum. Raise the limit if necessary, or use the "max_depth" option of the Symfony serializer.');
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -122,7 +122,7 @@ private function apply(bool $collection, QueryBuilder $queryBuilder, QueryNameGe
     private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, bool $forceEager, bool $fetchPartial, string $parentAlias, array $options = [], array $normalizationContext = [], bool $wasLeftJoin = false, int &$joinCount = 0, int $currentDepth = null)
     {
         if ($joinCount > $this->maxJoins) {
-            throw new RuntimeException('The total number of joined relations has exceeded the specified maximum. Raise the limit if necessary, or use the "max_depth" option of the Symfony serializer.');
+            throw new RuntimeException('The total number of joined relations has exceeded the specified maximum. Raise the limit if necessary with the "api_platform.eager_loading.max_joins" configuration key (https://api-platform.com/docs/core/performance/#eager-loading), or limit the maximum serialization depth using the "enable_max_depth" option of the Symfony serializer (https://symfony.com/doc/current/components/serializer.html#handling-serialization-depth).');
```
teohhanhui: Too verbose for an error message. We should let the user do a Google search with the error message instead, or search in our GitHub repo. :laughing:

teohhanhui:
This could be done in `2.2`, no?

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
dunglas: It's a very difficult one to debug for newcomers IMO.

soyuka: What would render the thing useful would be to mention which property is being joined too much, could be a good lead to fix serialization groups.

soyuka:
#2524

PR #2004, opened 2018-06-05
Move lint jobs to CircleCI

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,12 @@ reusable-steps:
         - npm-cache-{{ .Revision }}
         - npm-cache-{{ .Branch }}
         - npm-cache
+  - &restore-php-cs-fixer-cache
+    restore_cache:
+      keys:
+        - php-cs-fixer-cache-{{ .Revision }}
+        - php-cs-fixer-cache-{{ .Branch }}
+        - php-cs-fixer-cache
   - &save-composer-cache-by-branch
     save_cache:
       paths:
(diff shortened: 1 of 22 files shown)


bendavies:
you can add a php-cs-fixer fixer cache

teohhanhui:
@bendavies Good catch! :D

.circleci/config.yml
```diff
@@ -45,6 +51,16 @@ reusable-steps:
       paths:
         - ~/.npm
       key: npm-cache-{{ .Revision }}-{{ .BuildNum }}
+  - &save-php-cs-fixer-cache-by-branch
+    save_cache:
+      paths:
+        - .php_cs.cache
+      key: php-cs-fixer-cache-{{ .Branch }}-{{ .BuildNum }}
```
bendavies: are you sure this is correct?
you never restore using `.BuildNum` so it seems pointless.
(i'm aware of how circle ci `restore_cache` works with substrings)

bendavies: i thought you might
save key: 
```
php-cs-fixer-cache-{{ .Branch }}-{{ .Revision }}
```
restore keys:
```
keys:
  - php-cs-fixer-cache-{{ .Branch }}-{{ .Revision }}
  - php-cs-fixer-cache-{{ .Branch }}
  - php-cs-fixer-cache
```

bendavies:
there is npm caching in here, but i can't see that npm is ever ran. am i missing it?

.circleci/config.yml
teohhanhui: It's intentional. The .BuildNum is for cache busting.

teohhanhui:
@bendavies `npx` ;)

bendavies:
ah! thanks

.circleci/config.yml
bendavies: ah, in my project i actually generate a hash of all php files they php-cs-fixer runs over, then cache on that.

something like:
`find src tests -type f -name "*.php" -exec md5sum {} + > php-cs-fixer.checksum`

```
- restore_cache:
  keys:
    - php-cs-fixer-{{ checksum "php-cs-fixer.checksum" }}
    - php-cs-fixer-
```

then there is no need for cache busting.

teohhanhui:
Overall build time should be back down to under 10 minutes (roughly). :tada:

Simperfit:
Thanks @teohhanhui

PR #2005, opened 2018-06-06
Update PHP-CS-Fixer rules

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -23,6 +23,10 @@ return PhpCsFixer\Config::create()
         '@PHPUnit60Migration:risky' => true,
         '@Symfony' => true,
         '@Symfony:risky' => true,
+        'align_multiline_comment' => [
+            'comment_type' => 'phpdocs_only',
+        ],
+        'array_indentation' => true,
         'array_syntax' => [
             'syntax' => 'short',
         ],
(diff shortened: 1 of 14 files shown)


.php_cs.dist
```diff
@@ -62,10 +65,16 @@ return PhpCsFixer\Config::create()
...
+                'class',
+                'function',
+                'const',
+            ],
+            'sortAlgorithm' => 'alpha',
+        ],
         'phpdoc_order' => true,
-        'psr4' => true,
-        'semicolon_after_instruction' => true,
+        // 'simplified_null_return' => true,
```
teohhanhui: We may be able to use this rule once we require at least PHP 7.1 and have the nullable return type declarations in place. But it might not be possible to have a return type declaration in all cases...

bendavies: note that master requires >=7.1

dunglas:
Thanks @@teohhanhui!

.php_cs.dist
teohhanhui: I'm aware of that. Will try turning this rule on in `master`.

PR #2006, opened 2018-06-07
Deactivate ItemNormalizer when not in ApiPlatform context

joelwurtz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

We use the serializer to save / load some json encoded data into our database, so we do something like this : 

```
$this->serializer->deserialize($encoded, MyObject::class . '[]', 'json');
```

However we recently added some validation on this object and use api platform to do it, so we needed to add a fake id parameter (not sure if it's still useful ?)

When doing this previous code was not working because when using this serializer it's trying to use the iri convertion tool which is really not need in our case.

Not sure if this is the right fix but for us it works.

### DIFF
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -32,7 +32,7 @@ class ItemNormalizer extends AbstractItemNormalizer
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
+        if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE]) && isset($context['resource_class'])) {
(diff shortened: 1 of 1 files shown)


dunglas:
I'm not sure to get the use case/problem, but I don't see how it may hurt to add this test, so 👍 on my side. But can you add a non-regression test to highlight the use case, and prevent that someone remove this check at some point?

ping @api-platform/core-team

lyrixx:
Ahaha, I work with @joelwurtz on the same project, and it looks like I had the same issue. But I totally missed this PR, that's why I created #2326

soyuka:
We merged #2326 so I guess this can be closed. Also https://github.com/api-platform/core/pull/2495 should improve the normalization of non-resources.

Issue #2012, opened 2018-06-11
FilterEagerLoadingExtension and leftJoin with reversed mapping fails


partageit:
When using a join with a reversed mapping, the filter eager loading extensions fails.

For example, adding this in a filter:

    $qb->leftJoin(Car::class, 'car', 'WITH', 'car.id = o.car');

leads to:

    TypeError: substr() expects parameter 3 to be integer, boolean given

The query is hooked by `FilterEagerLoadingExtension`: the first parameter expected format of `leftJoin` is a string looking like: `o.car`.

I tried this fix:

```php
private function getQueryBuilderWithNewAliases(/* [...] */)
    {
        // [...]

        //Change join aliases
        foreach ($joinParts[$originAlias] as $joinPart) {
            /** @var Join $joinPart */
            $joinString = str_replace($aliases, $replacements, $joinPart->getJoin());
            $pos = strpos($joinString, '.');
            if ($pos === false) {
                continue;
            }
            // [...]
        }

        // [...]
    }
```

It skips the alias renaming for this kind of join, but the result is correct.

I will create a PR for this soon.

PR #2013, opened 2018-06-11
FilterEagerLoadingExtension accepts joins with class name as join value

partageit:
Hello,

The cause is explained in #2012.

I have two doubts/things i am not proud of:

- the optimisation is skipped for related joins
- i am not sure of the behaviour of this fix with composite relations.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2012 
| License       | MIT
| Doc PR        | None


dunglas:
Shouldn't this patch target 2.2 as a bug fix?

partageit:
@dunglas  You're right, the target is 2.2 :-s

How can i fix it ?
Should I create a new pull request ?

dunglas:
You can change the target branch by editing your PR. Then you'll have to rebase against 2.2 (or create a new branch locally from 2.2, cherry-pick your commit, and push force in this remote branch).

partageit:
Ok, i can't change the target PR, only the source branch...

You are now allowed to change the PR if you want.

If it does not work, i will create a new PR.

Sorry for the mistake :-s

PR #2016, opened 2018-06-11
FilterEagerLoadingExtension accepts joins with class name as join value

partageit:
Hello,

The cause is explained in #2012.

I have two doubts/things i am not proud of:

- the optimisation is skipped for related joins
- i am not sure of the behaviour of this fix with composite relations.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2012 
| License       | MIT
| Doc PR        | None

(It is a reedition of the PR #2013, due to my mistake, sorry for the duplicate).

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -135,6 +135,9 @@ final class FilterEagerLoadingExtension implements ContextAwareQueryCollectionEx
             /** @var Join $joinPart */
             $joinString = str_replace($aliases, $replacements, $joinPart->getJoin());
             $pos = strpos($joinString, '.');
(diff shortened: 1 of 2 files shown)


soyuka:
Thank you for the patch!

> the optimisation is skipped for related joins

Only in the FilterEagerLoading right? This extension was originally added to fix the behavior of relations when a filter applies. In complicated cases I'd recommend to write a query from scratch, especially if optimization is needed. We can't cover every case :|.

> i am not sure of the behaviour of this fix with composite relations.

Not a problem as long as our current test suite isn't broken, currently supported use cases with composite relations will work. If not, it's again a complex behavior that we may fix in the future.

partageit:
Ok, thank you for removing my doubts, and for the merge :)

Issue #1717, opened 2018-02-19
Can't Implement maxDepth in API-Platform using YAML


fawazIbrahim:
I have an entity with the following

```
/**
 * @ORM\Table(name="category")
 * @ORM\Entity(repositoryClass="AppBundle\Repository\CategoryRepository")
 */
class Category
{
     ...
     /**
     * @ORM\OneToMany(targetEntity="AppBundle\Entity\Category", mappedBy="parent")
     */
    private $children;
    ...
}
```
and i've defined the "children" as an API-Subresource using YAML:

```
AppBundle\Entity\Category:
  properties:
    children:
      subresource:
        resourceClass: 'App\Entity\Category'
```

How to set a maxDepth for the `$children` without using annotation like

```
/**
* @ApiSubresource(maxDepth=1)
*/
```

remoteclient:
I have a similar problem with xml configuration. "maxDepth" seems to do nothing here.

soyuka:
https://github.com/api-platform/api-platform/issues/592 ?

Nightbr:
Hey, the maxDepth in ApiSubresource annotation is only for Subresource ROUTE maxDepth generation. 
i.e. If you have foo/{id}/bar and bar/{id}/other, Api platform will generate all subresource route such as foo/{id}/bar/{id}/other. You can chose the maxdepth of subresource route genration with the maxDepth parameter in ApiSubresource annotation.

If you want maxDepth for serialization, use directly @maxDepth annotation.

Note:  @ApiSubresource(maxDepth=1) is not released yet. Only on master branch.

ctrl-f5:
confirming bug with xml config. maxdepth **for route generation** does not work. 
Works as expected with annotations.

Nightbr:
Found it, We need to handle $maxDepth here https://github.com/api-platform/core/blob/master/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php#L151

I will make a PR today to fix this.

Nightbr:
@ctrl-f5 @soyuka Fix in this PR https://github.com/api-platform/core/pull/2019

Hadh:
Was this fixed?

Nightbr:
yes, fix & merged see #1717 

Do you still have this issue?

PR #2019, opened 2018-06-13
fix issue #1717 - Support Subresources MaxDepth for yaml and xml

Nightbr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1717 
| License       | MIT
| Doc PR        | -

I found it needed more integration in Extractor and Property Factory in order to use Subresources MaxDepth.

I have update related test cases.

Question about Type class which is used in ExtractorPropertyMetadataFactory:
`$type = $propertyMetadata->getType();`

It uses the built-in symfony Type class and we can't retrieve maxDepth info. So I set maxDepth to null in this case. If you have better idea.

### DIFF
--- a/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
@@ -146,13 +146,15 @@ final class ExtractorPropertyMetadataFactory implements PropertyMetadataFactoryI
         if (null !== $type) {
             $isCollection = $type->isCollection();
             $resourceClass = $isCollection ? $type->getCollectionValueType()->getClassName() : $type->getClassName();
+            $maxDepth = null; // for Type we can't configure maxDepth, maxDepth is always null
         } elseif (isset($subresource['resourceClass'])) {
             $resourceClass = $subresource['resourceClass'];
             $isCollection = $subresource['collection'] ?? true;
+            $maxDepth = $subresource['maxDepth'] ?? null;
         } else {
             return null;
         }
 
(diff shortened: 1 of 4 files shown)


Nightbr:
The code coverage seems to have an issue: https://codecov.io/gh/api-platform/core/pull/2019/diff

It seems it doesn't take into account `case 'xxx'` in the coverage... (in red). But all cases are covered because the code below each case is green... weird one...

soyuka:
Thanks for the bug fix, can you target the `2.2` branch?

antograssiot:
The diff is weird, you need to rebase on 2.2 I think

Nightbr:
It's when I change target from master to 2.2

If I rebase it messed up everything....

src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
```diff
@@ -146,13 +146,15 @@ private function createSubresourceMetadata($subresource, PropertyMetadata $prope
         if (null !== $type) {
             $isCollection = $type->isCollection();
             $resourceClass = $isCollection ? $type->getCollectionValueType()->getClassName() : $type->getClassName();
+            $maxDepth = null; // for Type we can't configure maxDepth, maxDepth is always null
```
soyuka: can you declare this var l145 instead? 

we don't need the comment, null is default anyway

soyuka:
Thanks @Nightbr !

Issue #2018, opened 2018-06-13
ID not writable when creating resource


bendavies:
Hey,

It's a pretty common use case to allow an API consumer to write a PK of a resource on creation. This allows, amongst other things, asynchronous api endpoints which the consumer can then poll, as they know the ID of their resource.

This code currently prevents this:
https://github.com/api-platform/core/blob/c0de241ab907e3ddaa857962804869963821f446/src/Serializer/ItemNormalizer.php#L35-L38

Does anyone know the intention of this code?
I'm pretty sure it can be safely adapted for the above use case, but just want some feedback before hand.

Thanks

bendavies:
No, that's not the case. Have a look at that if statement again

when posting to create a resource:
1. `$data['id']` is what is posted.
2. `$context[self::OBJECT_TO_POPULATE]` is not set
3. `$context['api_allow_update']` is false

so the exception is thrown.

soyuka:
Definitely, if you try with a property named `foo` as PK it'll work though :laughing:

bendavies:
found some related discussions:
https://github.com/api-platform/core/issues/1741
https://github.com/api-platform/api-platform/issues/343

bendavies:
https://github.com/api-platform/core/pull/1939

soyuka:
See reasoning in https://github.com/api-platform/core/pull/2022#issuecomment-399402142. You can always decorate the ItemNormalizer to implement your own logic.

Issue #1741, opened 2018-03-01
POST resource with a non-autogenerated identifier property named "id" (works with another property name)


temp:
I have an entity Receipt, with an ID-field, that is NOT auto-generated.
I need this as a local lookup for a remote microsoft sql server connection.

```php
/**
 * @ORM\Entity
 *
 * @ApiResource()
 */
class Receipt
{
    /**
     * @var int
     *
     * @ORM\Id
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var string
     *
     * @ORM\Column(type="string")
     */
    private $name;
```

Now when posting to the endpoint with this payload:
```json
{
	"id": 123,
	"name": "test"
}
```

I'm getting an error, "Update is not allowed for this operation". So I activated api_allow_update for this resource. Afterwards I receive this error: "Item not found for /api/receipts/123".

Looking through the code I found the ItemNormalizer, which checks for an index "id" and tries to populate the entity. Which, of course, fails.
I changed to field from "id" to "test", and it works like expected...
Isn't this a little bit hard-coded? Binding this functionality to an index "id"?

Or am I doing it wrong? Is there a functionality for create-or-update? I'd go with PUT /api/receipts/123, but this also throws an "Item not found" error.

soyuka:
do you have getters/setters?

temp:
Yes, getters and setters.

soyuka:
So, basically (with custom id field - not auto generated):

Create:

POST `{id: 123, name: 'test'}` /api/receipts

Update:

PUT `{name: 'test'}` /api/receipts/123

You should not need to change `api_allow_update`. If it doesn't work with `id` it's either that `id` isn't writable (no `setId()` method) or it's a bug.

temp:
Then this looks like a bug.

My entity:
```php
<?php

namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity(repositoryClass="App\Repository\ReceiptRepository")
 *
 * @ApiResource(
 * )
 */
class Receipt
{
    /**
     * @var int
     *
     * @ORM\Id
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var string
     *
     * @ORM\Column(type="string")
     */
    private $name;

    public function getId(): int
    {
        return $this->id;
    }

    public function setId(int $id): void
    {
        $this->id = $id;
    }

[...]

PR #2022, opened 2018-06-13
Fix allow creating resources with custom identifers

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1937, https://github.com/api-platform/core/issues/2018, https://github.com/api-platform/core/issues/1741, https://github.com/api-platform/api-platform/issues/343
| License       | MIT

This is a re-submission of https://github.com/api-platform/core/pull/1939 with fixes and passing tests. @Toflar's commits are retained.

### DIFF
--- a/features/main/custom_normalized.feature
+++ b/features/main/custom_normalized.feature
@@ -50,7 +50,7 @@ Feature: Using custom normalized entity
     """
 
   Scenario: Create a resource with a custom normalized dummy and an id
-    When I add "Content-Type" header equal to "application/json"
+    When I add "Content-Type" header equal to "application/ld+json"
     When I add "Accept" header equal to "application/json"
     And I send a "PUT" request to "/related_normalized_dummies/1" with body:
     """
@@ -58,7 +58,7 @@ Feature: Using custom normalized entity
       "name": "My Dummy",
       "customNormalizedDummy":[{
         "@context": "/contexts/CustomNormalizedDummy",
-        "@id": "/custom_normalized_dummies/1",
+        "@id": "app_dev.php/custom_normalized_dummies/1",
         "@type": "CustomNormalizedDummy",
         "id": 1,
         "name": "My Dummy"
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\JsonLd\Serializer;
 use ApiPlatform\Core\Api\IdentifiersExtractorInterface;
 use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\JsonLd\ContextBuilderInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
@@ -89,7 +90,34 @@ final class ItemNormalizer extends AbstractItemNormalizer
         return self::FORMAT === $format && parent::supportsDenormalization($data, $type, $format);
     }
 
-    public function getIdentifiersForDenormalization(string $class): array
+    /**
+     * {@inheritdoc}
+     */
+    protected function setObjectToPopulate($data, string $class, array &$context)
+    {
+        try {
+            parent::setObjectToPopulate($data, $class, $context);
+        } catch (InvalidArgumentException $e) {
+            // https://github.com/api-platform/core/issues/857
+            $identifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
+            $identifiersData = \array_intersect_key($data, array_flip($identifiers));
+            if (0 === \count($identifiersData)) {
+                throw $e;
+            }
+
+            // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 is merged.
+            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(
+                sprintf(
+                    '%s/%s',
+                    $this->iriConverter->getIriFromResourceClass($context['resource_class']),
+                    implode(';', $identifiersData)
+                ),
+                $context + ['fetch_data' => true]
+            );
+        }
+    }
+
+    protected function getIdentifiersForDenormalization(string $class): array
     {
         return ['@id'];
     }
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -91,6 +91,6 @@ final class ItemNormalizer extends AbstractItemNormalizer
 
     public function getIdentifiersForDenormalization(string $class): array
     {
-        return array_merge(parent::getIdentifiersForDenormalization($class), ['@id']);
+        return ['@id'];
     }
 }
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -13,9 +13,9 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\JsonLd\Serializer;
 
+use ApiPlatform\Core\Api\IdentifiersExtractorInterface;
 use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
-use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\JsonLd\ContextBuilderInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
@@ -41,9 +41,9 @@ final class ItemNormalizer extends AbstractItemNormalizer
     private $resourceMetadataFactory;
     private $contextBuilder;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, ContextBuilderInterface $contextBuilder, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, ContextBuilderInterface $contextBuilder, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, IdentifiersExtractorInterface $identifiersExtractor = null)
     {
-        parent::__construct($propertyNameCollectionFactory, $propertyMetadataFactory, $iriConverter, $resourceClassResolver, $propertyAccessor, $nameConverter, $classMetadataFactory);
+        parent::__construct($propertyNameCollectionFactory, $propertyMetadataFactory, $iriConverter, $resourceClassResolver, $propertyAccessor, $nameConverter, $classMetadataFactory, null, false, $identifiersExtractor);
 
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->contextBuilder = $contextBuilder;
@@ -91,20 +91,9 @@ final class ItemNormalizer extends AbstractItemNormalizer
 
     /**
      * {@inheritdoc}
-     *
-     * @throws InvalidArgumentException
      */
-    public function denormalize($data, $class, $format = null, array $context = [])
+    protected function getIdentifiersForDenormalization(string $class): array
     {
-        // Avoid issues with proxies if we populated the object
-        if (isset($data['@id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
-            if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
-                throw new InvalidArgumentException('Update is not allowed for this operation.');
-            }
-
-            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri($data['@id'], $context + ['fetch_data' => true]);
-        }
-
-        return parent::denormalize($data, $class, $format, $context);
+        return ['@id'];
     }
 }
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -89,11 +89,8 @@ final class ItemNormalizer extends AbstractItemNormalizer
         return self::FORMAT === $format && parent::supportsDenormalization($data, $type, $format);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function getIdentifiersForDenormalization(string $class): array
+    public function getIdentifiersForDenormalization(string $class): array
     {
-        return ['@id'];
+        return array_merge(parent::getIdentifiersForDenormalization($class), ['@id']);
     }
 }
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -97,24 +97,27 @@ final class ItemNormalizer extends AbstractItemNormalizer
     {
         try {
             parent::setObjectToPopulate($data, $class, $context);
-        } catch (InvalidArgumentException $e) {
-            // https://github.com/api-platform/core/issues/857
-            $identifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
-            $identifiersData = \array_intersect_key($data, array_flip($identifiers));
-            if (0 === \count($identifiersData)) {
-                throw $e;
-            }
 
-            // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 is merged.
-            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(
-                sprintf(
-                    '%s/%s',
-                    $this->iriConverter->getIriFromResourceClass($context['resource_class']),
-                    implode(';', $identifiersData)
-                ),
-                $context + ['fetch_data' => true]
-            );
+            return;
+        } catch (InvalidArgumentException $e) {
         }
+
+        // https://github.com/api-platform/core/issues/857
+        $identifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
+        $identifiersData = \array_intersect_key($data, array_flip($identifiers));
+        if (0 === \count($identifiersData)) {
+            throw $e;
+        }
+
+        // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 is merged.
+        $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(
+            sprintf(
+                '%s/%s',
+                $this->iriConverter->getIriFromResourceClass($context['resource_class']),
+                implode(';', $identifiersData)
+            ),
+            $context + ['fetch_data' => true]
+        );
     }
 
     protected function getIdentifiersForDenormalization(string $class): array
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -515,7 +515,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return;
         }
 
-        $updateAllowed = isset($context['api_allow_update']) && true !== $context['api_allow_update'];
+        $updateAllowed = $context['api_allow_update'] ?? false;
         $identifiers = $this->getIdentifiersForDenormalization($class);
         $writeableIdentifiers = [];
 
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -503,12 +503,11 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
      * Sets the object to populate if allowed to and not set already. If you pass identifiers, it will search for them,
      * otherwise it falls back to the provided IdentifiersExtractor.
      *
-     * @param $data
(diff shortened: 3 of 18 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +498,71 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+
+        if (0 === \count($identifiersData)) {
+            return;
+        }
+
+        if (!$updateAllowed) {
+            throw new InvalidArgumentException('Update is not allowed for this operation.');
+        }
+
+        // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 is merged.
```
soyuka: The #1837 PR is based on master.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +498,71 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+            // TODO: use $this->iriConverter->getIriFromPlainIdentifier() once https://github.com/api-platform/core/pull/1837 is merged.
+            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(
+                sprintf(
+                    '%s/%s',
+                    $this->iriConverter->getIriFromResourceClass($context['resource_class']),
+                    implode(';', $identifiersData)
+                ),
+                $context + ['fetch_data' => true]
+            );
+        }
```
soyuka: Looks like we're using twice the same logic here, can't we avoid to repeat these?

Toflar:
Nice one! Want me to close the other PR then, right?

bendavies:
@Toflar I'm happy to close this and you can cherry pick into the original PR if you like. this PR is rebased on 2.2 btw.

Toflar:
Nah, I'm perfectly fine with you taking over! Thanks for the work!

src/Serializer/AbstractItemNormalizer.php
soyuka: Was a note for myself :D

src/Serializer/AbstractItemNormalizer.php
bendavies: @soyuka if you have some ideas, please say. I can't see a very nice way atm.

bendavies:
@dunglas @teohhanhui @meyerbaptiste can i ask for a review please?

src/Serializer/AbstractItemNormalizer.php
Toflar: Can you shed some light on why this is needed? Looks like a fallback algorithm that should go into the iriconverter?

bendavies: I've added a comment a few lines up. it is a fix for https://github.com/api-platform/core/issues/857

Not sure I agree with the fix, but I have kept it as per the original as a behat feature tests it.

soyuka:
features/main/custom_normalized.feature:52 failed though

bendavies:
Hmm. I'll check it out.

bendavies:
think I've fixed this by making the feature correct, to correctly address https://github.com/api-platform/core/issues/857.

bendavies:
@dunglas this is ready

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -91,20 +92,33 @@ public function supportsDenormalization($data, $type, $format = null)
...
      */
-    public function denormalize($data, $class, $format = null, array $context = [])
+    protected function setObjectToPopulate($data, string $class, array &$context)
     {
-        // Avoid issues with proxies if we populated the object
-        if (isset($data['@id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
-            if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
-                throw new InvalidArgumentException('Update is not allowed for this operation.');
+        try {
+            parent::setObjectToPopulate($data, $class, $context);
```
meyerbaptiste: Can you add a `return;` here and move the code in the `catch` part outside?

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -91,20 +92,33 @@ public function supportsDenormalization($data, $type, $format = null)
...
+        } catch (InvalidArgumentException $e) {
+            // https://github.com/api-platform/core/issues/857
+            $identifiers = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
+            $identifiersData = \array_intersect_key($data, array_flip($identifiers));
+            if (0 === \count($identifiersData)) {
```
meyerbaptiste:
```php
if (!$identifiersData) {
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +498,51 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+    {
+        if (isset($context[self::OBJECT_TO_POPULATE]) || !\is_array($data)) {
+            return;
+        }
+
+        $updateAllowed = $context['api_allow_update'] ?? false;
+        $identifiers = $this->getIdentifiersForDenormalization($class);
+        $identifiersData = \array_intersect_key($data, array_flip($identifiers));
+
+        if (0 === \count($identifiersData)) {
```
meyerbaptiste:
```php
if (!$identifiersData)) {
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +498,51 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+
+    /**
+     * Returns the identifiers used when denormalizing an object.
+     * Usually this is just what the IdentifiersExtractor provides but you might want to override this method in your
+     * normalizer depending on the format used. Note: They also have to be writeable to serve as proper identifier.
+     */
+    protected function getIdentifiersForDenormalization(string $class): array
+    {
+        // Backwards compatibility
+        if (null === $this->identifiersExtractor) {
```
meyerbaptiste:
```php
if (!$this->identifiersExtractor) {
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -488,4 +498,51 @@ protected function normalizeRelation(PropertyMetadata $propertyMetadata, $relate
...
+        $updateAllowed = $context['api_allow_update'] ?? false;
+        $identifiers = $this->getIdentifiersForDenormalization($class);
+        $identifiersData = \array_intersect_key($data, array_flip($identifiers));
+
+        if (0 === \count($identifiersData)) {
+            return;
+        }
+
+        if (!$updateAllowed) {
+            throw new InvalidArgumentException('Update is not allowed for this operation.');
```
meyerbaptiste: Should not this exception be thrown before the previous condition (the `return;`)?

features/main/custom_normalized.feature
```diff
@@ -50,15 +50,15 @@ Feature: Using custom normalized entity
...
+    When I add "Content-Type" header equal to "application/ld+json"
     When I add "Accept" header equal to "application/json"
     And I send a "PUT" request to "/related_normalized_dummies/1" with body:
     """
     {
       "name": "My Dummy",
       "customNormalizedDummy":[{
         "@context": "/contexts/CustomNormalizedDummy",
-        "@id": "/custom_normalized_dummies/1",
+        "@id": "app_dev.php/custom_normalized_dummies/1",
```
meyerbaptiste: Why `app_dev.php`?

src/JsonLd/Serializer/ItemNormalizer.php
bendavies: is this a style adopted by API-P? i prefer to be explicit but happy to align with project style.

src/JsonLd/Serializer/ItemNormalizer.php
bendavies: you mean an empty catch right?

src/Serializer/AbstractItemNormalizer.php
bendavies: don't think so, we only throw if identifiers are provided.

features/main/custom_normalized.feature
bendavies: this is specifically a fix for #857 (#877) and i needed a way to force the custom behaviour in the Json LD ItemNormalizer. i.e the iri must be incorrect.

src/Serializer/AbstractItemNormalizer.php
bendavies: again, happy to follow style of the project, but `if (null === $variable)` is done quite a lot in this project, so i'm not sure there's any point unless consistency is introduced.

src/JsonLd/Serializer/ItemNormalizer.php
bendavies: done

jamesisaac:
Is it expected for this to only work with `Content-Type: application/ld+json`?

I've been using `Content-Type: application/json` along with `allow_plain_identifiers: true` to allow referencing other entities by ID instead of IRI.

But applying this fix still leaves me with `"Update is not allowed for this operation."` when I POST a value for ID, unless I switch to ld+json and go back to using IRIs for the other fields.

dunglas:
There is probably a misunderstanding (because of me), let me summarize:

* It **is** supported to set a custom id when using JSON-LD

    POST /books

    {"id": "<a-uuid>"}

This will work, and the result will be something like:

    {"@id": "/books/<a-uuid>", "id": "<a-uuid>"}

* `id` is a special field when using the generic `ItemNormalizer` (JSON, XML, CSV...). This special property behaves exactly like JSON-LD's `@id`. It is intended to display an IRI. It means that it's not possible to access to the internal `id` property, because of this special `id` property that takes the lead (there is a conflict, that we can solve, more about this later).
* The introduction of `allow_plain_identifiers` added some more confusion, because in the beginning only IRIs was supported, but now you can emulate internal ID thanks to this feature, but it's not real internal IDs, so there are a lot of troubles including the issues this PR tries to fix

I was reluctant to merge `allow_plain_identifiers` but as it was a feature requested very often, and a one that helps porting legacy APIs to API Platform, it has been done anyway.

The main problem is we need both an external id (the IRI), and an internal one (for instance an UUID). They cannot be the same (and it's why `@id` in JSON-LD is very practical.

In GraphQL, we had the same problem: the `id` key is reserved by the Relay specification, and contains a global obfuscated ID (that can be an IRI, a base-64 encoded composed identifier like `base64("$type-$id")`...). For GraphQL, we solved the problem by using `id` for the public (non-writable identifier), and automatically renaming any internal identifier called `id` to `_id`: https://github.com/api-platform/core/blob/e3d50fff984eeaa6a5089a135043418063c3463b/src/GraphQl/Serializer/ItemNormalizer.php#L93-L95
Having 2 fields, one for the internal id, and another for the external id is the solution.

We need to make the same thing for the generic `ItemIdentifier`. Here is my first proposal:

1. Use `@id` (as in JSON-LD) for the public identifier (the one that MUST be used for relations)
2. Use the identifier chosen by the user as a private one, (can be `id` or `code` anything he wants)
3. **deprecated** IRIs support in `id`, `@id` must be used instead
4. Deprecate the `allow_plain_identifier` totally, it means that setting relation will **require** to use an ID (`@id`) (plain identifiers will not be supported anymore). It will allow us to drop a lot of complex and fragile code from API Platform

Alternatively, we may use the same convention than GraphQL: `id` for the IRI (public id), `_id` for the internal id, but it will make port of legacy APIs harder.
Alternatively, we may also support setting relation with plain ids (is this currently supported?) using some kind of magic convention like `_relation`... I would like to prevent this.

WDYT?

dunglas:
After discussing with @bendavies, we can extend my previous proposal:

* Only support IRIs for relation. It means the generic  `ItemNormalizer` will just be the same than the JSON-LD `itemNormalizer`, we'll be able to drop one class (:D). Less confusing for the end user, less maintenance for us, same behavior everywhere.
* Provide an opt-in (disabled by default) `ItemNormalizer` decorator in API Platform to help porting legacy API that will internally convert plain IDs (provided in the JSON document) to IRIs. So the user having a legacy API sending requests like `{"books": 22}` to set a relation will be able to keep BC, but internally we'll use IRIs everywhere, (except in the opt-in decorator, that will to the conversion between plain IDs and IRIs).

WDYT?

bendavies:
Thanks for taking the time to comment and explani the current state of things. 
It's now clear where the misunderstanding came to be.

I like the proposal, especially moving the plain identifier handling to a decorator!

bendavies:
For completeness, as I said to @dunglas privately, I won't have a chance to work on this proposal for quite a while. So if anyone else has some time, please feel free :)

teohhanhui:
My stand is consistent with what I've said about the Problem RFC. We don't get any benefit by mimicking a standard. It only leads to confusion as to whether we're actually following the standard or not (we aren't, so it shouldn't look like we are).

borNfreee:
Well, a couple of days with debugger and it works now. 

* The code below is for those who find a solution right meow before anything is done officially in the repository (which I suspect may take months)
* Use it on your own risk, but it works for me and all tests are green

What we did is just replace `ItemNormalizer` with our own implementation, that allows to post the following JSON:

```js
POST /resources
{
    "id": 1234545,
    "anyOther": "fields"
}
```

<details>
<summary>Custom ItemNormalizer</summary>

```php
declare(strict_types=1);

namespace App\Serializer\Normalizer;

use ApiPlatform\Core\Exception\InvalidArgumentException;


class ItemNormalizer extends AbstractItemNormalizer
{
    private const IDENTIFIER = 'id';

    {
        $context['api_denormalize'] = true;

            $context['resource_class'] = $class;
        }

        $this->setObjectToPopulate($data, $context);

    }

    protected function setObjectToPopulate($data, array &$context): void
    {
        // in PUT request OBJECT_TO_POPULATE is already set by this moment
        if (!\is_array($data) || isset($context[self::OBJECT_TO_POPULATE])) {
            return;
        }

        [$identifierName, $identifierMetadata] = $this->getResourceIdentifierData($context);

        $isUpdateAllowed = $context['api_allow_update'] ?? false;
        $hasIdentifierInRequest = array_key_exists(self::IDENTIFIER, $data);
        $hasWritableIdentifierInRequest = $hasIdentifierInRequest && $identifierMetadata->isWritable();
        // when it is POST, update is not allowed for top level resource, but is allowed for nested resources
        $isTopLevelResourceInPostRequest = !$isUpdateAllowed
            && $context['operation_type'] === 'collection'
            && $context['collection_operation_name'] === 'post';

        // if Resource does not have an ID OR if it is writable custom id - we should not populate Entity from DB
        if (!$hasIdentifierInRequest || ($hasWritableIdentifierInRequest && $isTopLevelResourceInPostRequest)) {
            return;
        }

        if (!$isUpdateAllowed) {
            throw new InvalidArgumentException('Update is not allowed for this operation.');
        }

        try {
            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(
                (string) $data[self::IDENTIFIER],
                $context + ['fetch_data' => true]
            );
        } catch (InvalidArgumentException $e) {
            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(
                sprintf(
                    '%s/%s',
                    $this->iriConverter->getIriFromResourceClass($context['resource_class']),
                    $data[$identifierName]
                ),
                $context + ['fetch_data' => true]
            );
        }
    }

    private function getResourceIdentifierData(array $context): array
    {
        $identifierPropertyName = null;
        $identifierPropertyMetadata = null;
        $className = $context['resource_class'];

        $properties = $this->propertyNameCollectionFactory->create($className, $context);

        foreach ($properties as $propertyName) {
            $property = $this->propertyMetadataFactory->create($className, $propertyName);

            if ($property->isIdentifier()) {
                $identifierPropertyName = $propertyName;
                $identifierPropertyMetadata = $property;
                break;
            }
        }

        if ($identifierPropertyMetadata === null) {
            throw new \LogicException(
                sprintf(
                    'Resource "%s" must have an identifier. Properties: %s.',
                    $className,
                    implode(',', iterator_to_array($properties->getIterator()))
                )
            );
        }

        return [$identifierPropertyName, $identifierPropertyMetadata];
    }
}
```
</details>


To replace api-platform's analyzer with this custom one - just override `class` in `CompilerPass`

`src/Kernel.php`

```diff
+ use App\Serializer\Normalizer\ItemNormalizer;

- class Kernel extends BaseKernel
+ class Kernel extends BaseKernel implements CompilerPassInterface

+    public function process(ContainerBuilder $container)
+    {
+        $container->getDefinition('api_platform.serializer.normalizer.item')->setClass(ItemNormalizer::class);
+    }
 }
```

-----------

Still, I want to move forward this discussion. This is one of the feature that I expect to work out of the box from platform named `api-platform`.

And by the way, provided patch in this PR does not work, I literally copied code to `vendor` folder.

dunglas:
Can you rebase against 2.3 please?

bendavies:
@dunglas why? You have proposed the 'proper' solution above. This pr should not be merged, I thought we decided.

dunglas:
My bad, I read this PR too fast.

soyuka:
Let's close this then.

bendavies:
let's try not forget about your proposal @dunglas :)

borNfreee:
I suggest opening another issue with that proposal, because it will be forgotten 100% ;)

soyuka:
lol was doing too many things at once and forgot to submit the issue, done :)

clemherreman:
Hello everyone,
Are there any news on this? I noticed #2201, and I wanted to know if there is any ETA on this. Like @borNfreee I'm willing to help on that, but I guess I'd need some guidance on how :)

My use case is that I have an application, consuming my API, that needs to work offline, and still create new resources instances locally, while offline. A good solution would be to have my resources uses UUID as identifiers, and allow to specify the identifiers in `POST` calls. This way the application and the API can create new instances independently.

As this issues is closed, I'll post it on ##2201 too, because I'm not sure somebody is reading here :)

nfacciolo:
> 
> 
> Well, a couple of days with debugger and it works now.
> 
>     * The code below is for those who find a solution right meow before anything is done officially in the repository (which I suspect may take months)
> 
>     * Use it on your own risk, but it works for me and all tests are green
> 
> 
> What we did is just replaced `ItemNormalizer` with our own implementation, that allows to post the following JSON:
> 
> ```js
> POST /resources
> {
>     "id": 1234545,
>     "anyOther": "fields"
> }
> ```
> 
> Custom ItemNormalizer (click to show)
> 
> To replace api-platform's analyzer with this custom one - just override `class` in `CompilerPass`
> 
> `src/Kernel.php`
> 
> ```diff
> + use App\Serializer\Normalizer\ItemNormalizer;
> 
> - class Kernel extends BaseKernel
> + class Kernel extends BaseKernel implements CompilerPassInterface
> 
> +    public function process(ContainerBuilder $container)
> +    {
> +        $container->getDefinition('api_platform.serializer.normalizer.item')->setClass(ItemNormalizer::class);
> +    }
>  }
> ```
> 
> Still, I want to move forward this discussion. This is one of the feature that I expect to work out of the box from platform named `api-platform`.
> 
> And by the way, provided patch in this PR does not work, I literally copied code to `vendor` folder and try to debug it.
> 
> UPD: the full code for api-platform 2.4.3+ is here [api-platform/api-platform#343 (comment)](https://github.com/api-platform/api-platform/issues/343#issuecomment-525773198)

See https://github.com/api-platform/api-platform/issues/343#issuecomment-525773198 if you want make it work

loicleng:
Hello everyone, 

Here is an example I made for ApiPlatform >= 3.0 that converts plain ids sent for relation field (either for item or collection relation) to IRI's before denormalization.
It may be improved but seems to be working for me so I'll still post it here in case it may help someone as I personnally struggled to find custom normalizer examples.

```
<?php

namespace App\Serializer;

use ApiPlatform\Api\IriConverterInterface;
use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
use Symfony\Component\Serializer\Normalizer\NormalizerInterface;

// Enable to send plain identifiers as object relation or collection
final class CustomItemNormalizer implements NormalizerInterface, DenormalizerInterface
{

    public function __construct(
        private readonly NormalizerInterface $normalizer,
        private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory,
        private readonly IriConverterInterface $iriConverter
    )
    {
        if (!$normalizer instanceof DenormalizerInterface) {
            throw new \InvalidArgumentException('The normalizer must implement the DenormalizerInterface');
        }
    }

    {
        if ($class == $context["operation"]->getClass()) {
            foreach ($data as $key => $value) {
                $propertyMetadata = $this->propertyMetadataFactory->create($class, $key);
                $propertyTypeObject = $propertyMetadata->getBuiltinTypes()[0];
                $propertyType = $propertyTypeObject->getBuiltinType();

                // If property is a relation and is not a translation relation
                if ($propertyType === 'object' && $key !== 'translations') {
                    $data = $this->convertIdRelationsToIRIs($data, $key, $value, $propertyTypeObject);
                }
            }
        }

        return $this->normalizer->denormalize($data, $class, $format, $context);
    }

    private function convertIdRelationsToIRIs(array $data, string $key, mixed $value, Type $propertyTypeObject): array {

        // For collection relation: Will replace data embbed relation sent as "categories": [1, 2, 3] to "categories": ['/catalog/categories/1', '/catalog/categories/2', ...]
        if ($propertyTypeObject->isCollection() && !str_contains('Translation', $propertyTypeObject->getCollectionValueTypes()[0]->getClassName())) {
            $data = $this->handleCollectionIdsConversion($data, $key, $value, $propertyTypeObject);

        // For single relation : Converts "category": 1 to "category": "/catalog/categories/1"
        } else {
            $data = $this->handleItemIdConversion($data, $key, $value, $propertyTypeObject);
        }

        return $data;
    }

    private function handleCollectionIdsConversion(array $data, string $key, mixed $value, Type $propertyTypeObject): array
    {
        $embbedClass = new ($propertyTypeObject->getCollectionValueTypes()[0]->getClassName());
        $manyToManyIRIsList = [];
        foreach ($value as $manyToManyEmbbedObjectId) {
            if (!is_int($manyToManyEmbbedObjectId)) {
                throw new \InvalidArgumentException('Parameters for relation ' . $embbedClass::class . ' must be integers' );
            }
            $iri = $this->iriConverter->getIriFromResource(resource: $embbedClass::class, context: ['uri_variables' => ['id' => $manyToManyEmbbedObjectId]]);
            $manyToManyIRIsList[] = $iri;
        }
        if (count($manyToManyIRIsList) > 0) {
            $data[$key] = $manyToManyIRIsList;
        }

        return $data;
    }

    private function handleItemIdConversion(array $data, string $key, mixed $value, Type $propertyTypeObject): array
    {
        $embbedClass = new ($propertyTypeObject->getClassName());

        if (!is_int($value)) {
            throw new \InvalidArgumentException('Parameter for relation ' . $embbedClass::class . ' must be integer' );
        }

        $iri = $this->iriConverter->getIriFromResource(resource: $embbedClass::class, context: ['uri_variables' => ['id' => $value]]);
        $data[$key] = $iri;

        return $data;
    }

    {
        return $this->normalizer->supportsDenormalization($data, $type, $format);
    }

    {
        return $this->normalizer->normalize($object, $format, $context);
    }

    public function supportsNormalization($data, $format = null)
    {
        return $this->normalizer->supportsNormalization($data, $format);
    }
}
```

Which is registered such as follow in services.yaml :
```
services:

    'App\Serializer\CustomItemNormalizer':
        arguments: [ '@api_platform.serializer.normalizer.item' ]
```

PR #2026, opened 2018-06-16
RFC: enforce void return

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

We can add this fixer to master as it enforces php 7.1. thoughts?

further work would be adding nullable types.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -69,6 +69,7 @@ return PhpCsFixer\Config::create()
         'strict_comparison' => true,
         'strict_param' => true,
         'ternary_to_null_coalescing' => true,
+        'void_return' => true,
     ])
     ->setFinder($finder)
 ;
--- a/features/bootstrap/CoverageContext.php
+++ b/features/bootstrap/CoverageContext.php
@@ -34,7 +34,7 @@ final class CoverageContext implements Context
     /**
      * @BeforeSuite
      */
-    public static function setup()
+    public static function setup(): void
     {
         $filter = new Filter();
         $filter->addDirectoryToWhitelist(__DIR__.'/../../src');
@@ -44,7 +44,7 @@ final class CoverageContext implements Context
     /**
      * @AfterSuite
      */
-    public static function tearDown()
+    public static function tearDown(): void
     {
         $feature = getenv('FEATURE') ?: 'behat';
         (new PHP())->process(self::$coverage, __DIR__."/../../build/cov/coverage-$feature.cov");
@@ -53,7 +53,7 @@ final class CoverageContext implements Context
     /**
      * @BeforeScenario
      */
-    public function startCoverage(BeforeScenarioScope $scope)
+    public function startCoverage(BeforeScenarioScope $scope): void
     {
         self::$coverage->start("{$scope->getFeature()->getTitle()}::{$scope->getScenario()->getTitle()}");
     }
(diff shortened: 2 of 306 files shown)


dunglas:
👍 but I think we cannot add it on interfaces and on classes not marked `final`.

bendavies:
👍, but not sure we can enforce that with php-cs-fixer. would have to remove the config and remember to add voids manually.

teohhanhui:
I don't see what's the problem with adding the `void` return type. It's not a BC break if we've never specified any other return type in the first place (whether as a PHP return type or as PHPDoc). It should have been considered as an implicit return nothing anyway.

teohhanhui:
Turning on the `simplified_null_return` rule will be more tricky, as I'm pretty sure there will still be some places where we cannot use a nullable return type.

dunglas:
> I don't see what's the problem with adding the void return type.
Here is the problem: https://3v4l.org/TFDCE

We can only add it for methods on final classes and private methods, definitely not on interfaces and extendable classes.

teohhanhui:
I disagree. We're honouring the same contract. What's broken is only the implementation. I think it's reasonable to expect implementations to be updated, especially since it's due to the previously limited expressability of the language.

dunglas:
A definitive no on my side: we cannot force all projects using our interfaces or classes to upgrade all their implementations, it a huge BC break that can only be done in a major version.
And even in a major version, I'm not sure it is worth it. `void` enough benefits to counterbalance the fact that it will be a painful upgrade.

Symfony chosen to not use `void` at all. I'm ok to use it when it's safe and for new code, but not to break every existing projects using API Platform.

dunglas:
Btw, not adding `void` to our interfaces and non-final class doesn't prevent users to add it in implementations: https://3v4l.org/GbEfR

teohhanhui:
You're right, but there's no good way forward for this. Not adding it means we cannot properly enforce strict typing in our code. Do you see a way of enforcing the void return without actually using the void return type? :rofl:

teohhanhui:
My reasoning is quite simple:

- A newer version of PHP introduced a non-BC language feature. We (API Platform) have chosen to require this newer version of PHP as a minimum requirement.
- Whatever plugins / extensions for API Platform are now most likely forced to drop support for older PHP versions anyway, unless we've been really careful with our use of said new language features (I doubt it).
- So there's no problem in requiring the adoption of new language features in the plugins / extensions. Now is the best time to do it, when we're bumping the minimum PHP version supported, in order to use new language features.

teohhanhui:
Semver offers no guidance in non-BC new language features. It covers the contract between our project and client code, which has not changed.

dunglas:
It's a backward compatible feature, you can use PHP 7.1+ without using `void`.

teohhanhui:
Yes, but I'm talking about other language features, especially nullable types. It's so important that I'd even consider it more important than preventing BC breaks.

teohhanhui:
If we have to bump the major version to do it, then so be it. I'd advocate for that.

soyuka:
Let's close this, please lmk if you want to see this merged eventually.

PR #1988, opened 2018-05-30
Implement CacheableSupportsMethodInterface introduced in Symfony 4.1

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

It should improve dramatically the performance (benchmarking in progress).
It also means that we drop support for the Serializer Component <4.1, and as a direct consequence of PHP 7.0.

PR #2029, opened 2018-06-19
optimization - don't normalize scalar data

SergeyMilegov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| License       | MIT

### DIFF
--- a/src/Hydra/Serializer/CollectionNormalizer.php
+++ b/src/Hydra/Serializer/CollectionNormalizer.php
@@ -67,7 +67,11 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
         if (isset($context['api_sub_level'])) {
             $data = [];
             foreach ($object as $index => $obj) {
-                $data[$index] = $this->normalizer->normalize($obj, $format, $context);
+                if (is_scalar($obj)) {
+                    $data[$index] = $obj;
+                } else {
+                    $data[$index] = $this->normalizer->normalize($obj, $format, $context);
+                }
             }
 
             return $data;
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -447,6 +447,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
         unset($context['resource_class']);
 
+        if (is_scalar($attributeValue)) {
+            return $attributeValue;
+        }
+
         return $this->serializer->normalize($attributeValue, $format, $context);
     }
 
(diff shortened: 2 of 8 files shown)


soyuka:
tests would need a update, :+1: for the optimization though

teohhanhui:
-1

It's a feature, not a bug. With the caching (Symfony 4.1 and #1988), there should not be any overhead after the first time anyway, if there's no normalizer registered for that particular property.

soyuka:
> It's a feature, not a bug. With the caching (Symfony 4.1 and #1988), there should not be any overhead after the first time anyway, if there's no normalizer registered for that particular property.

We're still skipping some function calls no?

teohhanhui:
This will break the use case of registering normalizers for scalar types. It just cannot work.

bendavies:
@teohhanhui is right. this is why scalars are handled if a normalizer is not found: https://github.com/symfony/symfony/blob/a5a91d318ce73256b7c9677c3d46cf989383d006/src/Symfony/Component/Serializer/Serializer.php#L141

dunglas:
I agree with @teohhanhui, 👎 on my side.

SergeyMilegov:
>normalizers for scalar types...
WAT?
I made it to prevent normalization for scalar types because it is the most useless staff I can imagine.

SergeyMilegov:
> It's a feature, not a bug. With the caching (Symfony 4.1 and #1988), there should not be any overhead after the first time anyway, if there's no normalizer registered for that particular property.
What if I am using symfony 3.3...

teohhanhui:
@SergeyMilegov Normalizers for scalar types are not useless. Just because you don't use them, it doesn't mean that there are no valid use cases for them. :smile:

SergeyMilegov:
>Normalizers for scalar types are not useless. Just because you don't use them, it doesn't mean that there are no valid use cases for them. smile

But how can I disable it if I don't need it?

teohhanhui:
I think that's a good question. Do you have any suggestions? It'd be good to have in the Symfony Serializer too, I think.

SergeyMilegov:
Symfony doesn't normalize properties: Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer::getAttributeValue
It is ApiPlatform's feature.
So you can add parameter to the config and apply my changes depends on this parameter.

teohhanhui:
The AbstractObjectNormalizer is only an implementation. In fact, we could say that normalizing / denormalizing scalar values is a supported use case in Symfony Serializer, if we look at for example: https://github.com/symfony/symfony/blob/v4.1.0/src/Symfony/Component/Serializer/Serializer.php#L214

PR #2033, opened 2018-06-20
improve exception message when no identifier are defined

ronfroy:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/tests/Metadata/Property/Factory/AnnotationSubresourceMetadataFactoryTest.php
+++ b/tests/Metadata/Property/Factory/AnnotationSubresourceMetadataFactoryTest.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Tests\Metadata\Property\Factory;
 
 use ApiPlatform\Core\Annotation\ApiSubresource;
+use ApiPlatform\Core\Exception\InvalidResourceException;
 use ApiPlatform\Core\Metadata\Property\Factory\AnnotationSubresourceMetadataFactory;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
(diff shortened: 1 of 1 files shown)


teohhanhui:
I'm -1 as error messages shouldn't be verbose. We should communicate clearly what's the error encountered, but nothing else.

ronfroy:
@teohhanhui it saves time to have a useful exception message

teohhanhui:
If we want to give a better error message for the case when there are no identifiers, we should add a check after

https://github.com/api-platform/core/blob/0a69709067335fb2ccb5bd48b0850b47f3aaf72d/src/Bridge/Symfony/Routing/IriConverter.php#L118

maybe.

teohhanhui:
It also means we don't need to say "maybe". :smile:

ronfroy:
@teohhanhui done

teohhanhui:
That was not the right way to fix it. :smile:

Just add a check at the beginning of `generateIdentifiersUrl` that there are indeed any identifiers at all:

https://github.com/api-platform/core/blob/0a69709067335fb2ccb5bd48b0850b47f3aaf72d/src/Bridge/Symfony/Routing/IriConverter.php#L177-L179

And I think we could still target `2.2` since it could be considered a bug fix, and `2.3` is not released yet. :laughing:

ronfroy:
done

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -170,12 +170,22 @@ public function getSubresourceIriFromResourceClass(string $resourceClass, array
     /**
      * Generate the identifier url.
      *
-     * @param array $identifiers
+     * @param array  $identifiers
+     * @param object $resourceClass
```
teohhanhui: `$resourceClass` is a `string`. You should also add the type hint for the parameter.

teohhanhui:
Can you change this to target `2.2`?

Issue #1912, opened 2018-04-30
Better HTTP headers


teohhanhui:
For example, when a resource is created, we're not sending the correct HTTP headers in the response.

**Location**
> For 201 (Created) responses, the Location value refers to the primary
   resource created by the request.  For 3xx (Redirection) responses,
   the Location value refers to the preferred target resource for
   automatically redirecting the request.

https://tools.ietf.org/html/rfc7231#section-7.1.2

**Content-Location**
> If Content-Location is included in a 2xx (Successful) response
   message and its value refers (after conversion to absolute form) to a
   URI that is the same as the effective request URI, then the recipient
   MAY consider the payload to be a current representation of that
   resource at the time indicated by the message origination date.  For
   a GET (Section 4.3.1) or HEAD (Section 4.3.2) request, this is the
   same as the default semantics when no Content-Location is provided by
   the server.  For a state-changing request like PUT (Section 4.3.4) or
   POST (Section 4.3.3), it implies that the server's response contains
   the new representation of that resource, thereby distinguishing it
   from representations that might only report about the action (e.g.,
   "It worked!").  This allows authoring applications to update their
   local copies without the need for a subsequent GET request.
>
> If Content-Location is included in a 2xx (Successful) response
   message and its field-value refers to a URI that differs from the
   effective request URI, then the origin server claims that the URI is
   an identifier for a different resource corresponding to the enclosed
   representation.  Such a claim can only be trusted if both identifiers
   share the same resource owner, which cannot be programmatically
   determined via HTTP.
>
>   -  For a response to a GET or HEAD request, this is an indication
      that the effective request URI refers to a resource that is
      subject to content negotiation and the Content-Location
      field-value is a more specific identifier for the selected
      representation.
>
>   -  For a 201 (Created) response to a state-changing method, a
      Content-Location field-value that is identical to the Location
      field-value indicates that this payload is a current
      representation of the newly created resource.
>
>   -  Otherwise, such a Content-Location indicates that this payload is
      a representation reporting on the requested action's status and
      that the same report is available (for future access with GET) at
      the given URI.  For example, a purchase transaction made via a
      POST request might include a receipt document as the payload of
      the 200 (OK) response; the Content-Location field-value provides
      an identifier for retrieving a copy of that same receipt in the
      future.

https://tools.ietf.org/html/rfc7231#section-3.1.4.2

teohhanhui:
So, for example, if I do:

```
POST /books
Content-Type: application/ld+json
[...]

PR #2039, opened 2018-06-25
Better HTTP headers

jewome62:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes, i add one
| Tests pass?   | no, but i work on this
| Fixed tickets | #1912 
| License       | MIT
| Doc PR        | none

### DIFF
--- a/src/EventListener/RespondListener.php
+++ b/src/EventListener/RespondListener.php
@@ -40,15 +40,25 @@ final class RespondListener
             return;
         }
 
+        $headers = [
+            'Content-Type' => sprintf('%s; charset=utf-8', $request->getMimeType($request->getRequestFormat())),
+            'Vary' => 'Accept',
+            'X-Content-Type-Options' => 'nosniff',
+            'X-Frame-Options' => 'deny',
+        ];
+
+        if ($request->attributes->has('_api_write_item_iri')) {
+            $headers['Content-Location'] = $request->attributes->get('_api_write_item_iri');
+
+            if ($request->isMethod('POST')) {
+                $headers['Location'] = $request->attributes->get('_api_write_item_iri');
+            }
+        }
+
         $event->setResponse(new Response(
             $controllerResult,
             self::METHOD_TO_CODE[$request->getMethod()] ?? Response::HTTP_OK,
-            [
-                'Content-Type' => sprintf('%s; charset=utf-8', $request->getMimeType($request->getRequestFormat())),
-                'Vary' => 'Accept',
-                'X-Content-Type-Options' => 'nosniff',
-                'X-Frame-Options' => 'deny',
-            ]
+            $headers
         ));
     }
 }
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\EventListener;
 
+use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\DataPersister\DataPersisterInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
 
@@ -25,10 +26,12 @@ use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
 final class WriteListener
 {
     private $dataPersister;
+    private $iriConverter;
 
-    public function __construct(DataPersisterInterface $dataPersister)
+    public function __construct(DataPersisterInterface $dataPersister, IriConverterInterface $iriConverter = null)
     {
         $this->dataPersister = $dataPersister;
+        $this->iriConverter = $iriConverter;
     }
 
     /**
@@ -57,6 +60,10 @@ final class WriteListener
                 }
 
                 $event->setControllerResult($persistResult ?? $controllerResult);
+
+                if (null !== $this->iriConverter) {
+                    $request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($controllerResult));
+                }
                 break;
             case 'DELETE':
                 $this->dataPersister->remove($controllerResult);
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -22,6 +22,8 @@ Feature: Create-Retrieve-Update-Delete
     Then the response status code should be 201
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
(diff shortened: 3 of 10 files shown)


src/EventListener/WriteListener.php
```diff
@@ -25,10 +26,12 @@
 class WriteListener
 {
     private $dataPersister;
+    private $iriConverter;
 
-    public function __construct(DataPersisterInterface $dataPersister)
+    public function __construct(DataPersisterInterface $dataPersister, IriConverterInterface $iriConverter)
```
soyuka: this is a BC break

src/EventListener/RespondListener.php
```diff
@@ -42,15 +42,25 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
...
         }
 
+        $headers = [
+            'Content-Type' => sprintf('%s; charset=utf-8', $request->getMimeType($request->getRequestFormat())),
+            'Vary' => 'Accept',
+            'X-Content-Type-Options' => 'nosniff',
+            'X-Frame-Options' => 'deny',
+        ];
+
+        if($request->isMethod('POST') || $request->isMethod('PUT')){
```
teohhanhui: Should be the same for `PATCH`, I think...

src/EventListener/WriteListener.php
teohhanhui: The second argument must be made optional. :smile:

But we can trigger a deprecation error if it's not provided.

src/EventListener/WriteListener.php
```diff
@@ -57,6 +60,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
                 }
 
                 $event->setControllerResult($persistResult ?? $controllerResult);
+                $request->attributes->set('_iri_item', $this->iriConverter->getIriFromItem($controllerResult));
```
teohhanhui: Maybe `_api_write_item_iri`?

src/EventListener/WriteListener.php
soyuka: haha sorry yeah I should've say what you said in addition :laughing:

src/EventListener/WriteListener.php
```diff
@@ -25,10 +26,12 @@
 class WriteListener
```
teohhanhui: Also, I think we could mark this class as `final`? It was never designed to be extended in the first place...

(Strictly speaking, it'd be a BC break... But I highly doubt it'd bother anyone. I don't remember if we've done something like this before.)

jewome62:
Hands Up!

OK OK, I confess, I fucked up, I did not think and I pushed my work not finished by putting WIP

DO NOT SHOOT ME ! I HAVE A WIFE AND CHILDREN !

;)

src/EventListener/WriteListener.php
jewome62: or `@internal` ?

jewome62:
@teohhanhui 
All is good here,
The fail test became from last release of Symfony and affect master too

See: https://github.com/api-platform/core/pull/2043

src/EventListener/RespondListener.php
```diff
@@ -42,15 +42,29 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
...
         }
 
+        $headers = [
+            'Content-Type' => sprintf('%s; charset=utf-8', $request->getMimeType($request->getRequestFormat())),
+            'Vary' => 'Accept',
+            'X-Content-Type-Options' => 'nosniff',
+            'X-Frame-Options' => 'deny',
+        ];
+
+        if ($request->isMethod('POST') || $request->isMethod('PUT') || $request->isMethod('PATCH')) {
```
teohhanhui: Please use `\in_array` here.

src/EventListener/WriteListener.php
teohhanhui: Let's add `@final` in the PHPDoc.

jewome62: See https://github.com/api-platform/core/pull/2046

teohhanhui:
The error messages need to be rephrased. I'll give some suggestions, if somebody doesn't get it done before I do.

src/EventListener/RespondListener.php
```diff
@@ -42,15 +42,29 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
...
+
+        if (\in_array($request->getMethod(), ['POST', 'PUT', 'PATCH'])) {
+            if ($request->attributes->has('_api_write_item_iri')) {
+                $headers['Content-Location'] = $request->attributes->get('_api_write_item_iri');
+
+                if ($request->isMethod('POST')) {
+                    $headers['Location'] = $request->attributes->get('_api_write_item_iri');
+                }
+            } else {
+                @trigger_error(sprintf('No request attribute from `_api_write_item_iri` key is deprecated since API Platform 2.3 and will not be supported in API Platform 3, an string should always be returned. see deprecated into %s constructor for more details.', WriteListener::class), E_USER_DEPRECATED);
```
teohhanhui: Not sure if we should have this deprecation error here, but:

> Not setting the `_api_write_item_iri` request attribute is deprecated since API Platform 2.3 and will not be supported in API Platform 3. See %s for the default implementation.

jewome62:
@teohhanhui I make the fix.

Anyway, @dunglas said me that we are trying to reduce the dependencies in the Platform API classes, so normally I will remove the deprecated ones to make sure that it remains optional definitively.

src/EventListener/WriteListener.php
jewome62: done

teohhanhui: Let's leave this to #2046 then. :smile:

jewome62:
I have rebase from master

src/EventListener/WriteListener.php
```diff
@@ -13,22 +13,27 @@
...
 namespace ApiPlatform\Core\EventListener;
 
+use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\DataPersister\DataPersisterInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
 
 /**
  * Bridges persistense and the API system.
  *
+ * @final
```
teohhanhui: Should be removed.

soyuka:
@dunglas may you merge this please? I rebased it.

teohhanhui:
Thanks @jewome62! :tada:

PR #2040, opened 2018-06-25
Workflow integration

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | TODO
| Fixed tickets | na
| License       | MIT
| Doc PR        | TODO


src/Bridge/Symfony/Workflow/EventListener/WorkflowEnabledTransitionsListener.php
```diff
@@ -0,0 +1,53 @@
...
+            || 'workflow_state_get' !== $attributes['item_operation_name']
+        ) {
+            return;
+        }
+
+        $class = $request->attributes->get('data');
+        $workflow = $this->workflows->get($class);
+        $event->setResponse(new Response($this->serializer->serialize($workflow->getEnabledTransitions($class), $request->getRequestFormat())));
+    }
+}
```
soyuka: Should we keep this route? I'm not sure if it's useful, also the `Response` isn't a `class` it looks like this:

```json
[
    {
        "froms": [
            "valid"
        ],
        "name": "unvalidate",
        "tos": [
            "create"
        ]
    }
]
```

Represents the transitions available for a given resource.

src/Bridge/Symfony/Workflow/EventListener/WorkflowStateListener.php
```diff
@@ -0,0 +1,67 @@
...
+        if (!$transition = (string) $workflowTransition) {
+            throw new BadRequestHttpException('Transition is required.');
+        }
+
+        $class = $request->attributes->get('data');
+        $workflow = $this->workflows->get($class);
+
+        // @TODO replace by violation of some sort (validator?)
+        if (!$workflow->can($class, $transition)) {
+            throw new UnprocessableEntityHttpException("Can not apply transition '$transition'");
```
soyuka: Any ideas?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/WorkflowPass.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
+
+use ApiPlatform\Core\Bridge\Symfony\Workflow\Metadata\Resource\Factory\WorkflowOperationResourceMetadataFactory;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Extract classes that are binded to a workflow to build up their custom metadata.
```
teohhanhui: Grammar: binded -> bound

soyuka: :see_no_evil:

src/Bridge/Symfony/Workflow/EventListener/WorkflowStateListener.php
teohhanhui: We must not use this HTTP status code as it's a WebDAV extension. :smile:

teohhanhui: I think the most correct HTTP status code here would be 403.

jewome62: maybe  412  - Precondition Failed ?
If not 418, it always works ...

teohhanhui: No, we cannot use 412 as it's meant for the client to set the precondition: https://tools.ietf.org/html/rfc7232#section-4.2

418 :sunglasses: :tea:

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/WorkflowPass.php
```diff
@@ -0,0 +1,49 @@
...
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('workflow.registry')) {
+            return;
+        }
+
+        $registry = $container->getDefinition('workflow.registry');
+        $factory = $container->getDefinition(WorkflowOperationResourceMetadataFactory::class);
```
bendavies: won't this fail if `enable_workflow` is false?

src/Bridge/Symfony/Bundle/Resources/config/workflow.xml
```diff
@@ -0,0 +1,31 @@
...
+    </service>
+
+    <service id="ApiPlatform\Core\Bridge\Symfony\Workflow\PathResolver\WorkflowOperationPathResolver" decorates="api_platform.operation_path_resolver.generator">
+      <argument type="service" id="ApiPlatform\Core\Bridge\Symfony\Workflow\PathResolver\WorkflowOperationPathResolver.inner"/>
+    </service>
+
+    <service id="ApiPlatform\Core\Bridge\Symfony\Workflow\EventListener\WorkflowStateListener">
+      <argument type="service" id="serializer"/>
+      <argument type="service" id="workflow.registry"/>
+      <tag name="kernel.event_listener" event="kernel.request" priority="1"/>
```
bendavies: a comment describing the specific priority might be nice

src/Bridge/Symfony/Bundle/Resources/config/workflow.xml
```diff
@@ -0,0 +1,31 @@
...
+    </service>
+
+    <service id="ApiPlatform\Core\Bridge\Symfony\Workflow\EventListener\WorkflowEnabledTransitionsListener">
+      <argument type="service" id="serializer"/>
+      <argument type="service" id="workflow.registry"/>
+      <tag name="kernel.event_listener" event="kernel.view" priority="21"/>
```
bendavies: a comment describing the specific priority might be nice

src/Bridge/Symfony/Workflow/EventListener/WorkflowStateListener.php
```diff
@@ -0,0 +1,67 @@
...
+        if (!$transition = (string) $workflowTransition) {
+            throw new BadRequestHttpException('Transition is required.');
+        }
+
+        $class = $request->attributes->get('data');
+        $workflow = $this->workflows->get($class);
+
+        // @TODO replace by violation of some sort (validator?)
+        if (!$workflow->can($class, $transition)) {
+            throw new HttpException(403, "Can not apply transition '$transition'");
```
bendavies: hmmm not 403 i don't think. just 400 or [409](https://httpstatuses.com/409)

src/Bridge/Symfony/Workflow/Metadata/Resource/Factory/WorkflowOperationResourceMetadataFactory.php
```diff
@@ -0,0 +1,62 @@
...
+
+        if (!\in_array($resourceClass, $this->supportsWorkflow, true)) {
+            return $resourceMetadata;
+        }
+
+        $operations = $resourceMetadata->getItemOperations();
+
+        $operations['workflow_state_patch'] = [
+            'method' => 'PATCH',
+            '_path_suffix' => '/state',
```
bendavies: the field storing the state may not be called `state`?
can symfony workflow support multiple states per class?

src/Bridge/Symfony/Workflow/PathResolver/WorkflowOperationPathResolver.php
```diff
@@ -0,0 +1,50 @@
...
+namespace ApiPlatform\Core\Bridge\Symfony\Workflow\PathResolver;
+
+use ApiPlatform\Core\PathResolver\OperationPathResolverInterface;
+
+/**
+ * If any corresponding metadata, append the `_path_suffix` to the operation's path.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class WorkflowOperationPathResolver implements OperationPathResolverInterface
```
bendavies: add a unit test

src/Bridge/Symfony/Workflow/Metadata/Resource/Factory/WorkflowOperationResourceMetadataFactory.php
```diff
@@ -0,0 +1,62 @@
...
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+
+/**
+ * Adds two operation to resource classes supporting Workflows:
+ * - PATCH to update the workflow state
+ * - GET to get possible states for a given resource.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class WorkflowOperationResourceMetadataFactory implements ResourceMetadataFactoryInterface
```
bendavies: unit test?

src/Bridge/Symfony/Workflow/Metadata/Resource/Factory/WorkflowOperationResourceMetadataFactory.php
soyuka: > the field storing the state may not be called state?

Not a problem. Anyway this should be configurable IMO.

> can symfony workflow support multiple states per class?

I'm not sure to follow you but yes obviously because a transition brings a class from one place to another (states).

soyuka:
@bendavies I obviously need to add unit tests, I wanted some first reviews on this especially if we want to remove the WorkflowEnabledTransitionsListener for example.

tests/Fixtures/TestBundle/Entity/WorkflowOffer.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    private $id;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(type="string", length=255, nullable=true)
+     * @Assert\Choice({"create", "valid", "cancel"})
+     */
+    public $status = 'create';
```
bendavies: what is stopping the user writing to this directly with a post/patch request? it will skip the workflow `->can()` validation

bendavies:
Had a think about this.
Why does this require extra routing?
Why not just `PATCH $url` with `{"the_field_that_is_holds_the_state": "state_name"}`, and handle all this inside the item normalizers?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -93,6 +93,7 @@ public function getConfigTreeBuilder()
                 ->booleanNode('enable_entrypoint')->defaultTrue()->info('Enable the entrypoint')->end()
                 ->booleanNode('enable_docs')->defaultTrue()->info('Enable the docs')->end()
                 ->booleanNode('enable_profiler')->defaultTrue()->info('Enable the data collector and the WebProfilerBundle integration.')->end()
+                ->booleanNode('enable_workflow')->defaultTrue()->info('Enable the Workflow integration.')->end()
```
dunglas: It should default to the value of `interface_exists(WorkflowInterface::class)`.
May I suggest: 'Enable the integration with the Symfony Workflow component.'?

src/Bridge/Symfony/Bundle/Resources/config/workflow.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" ?>
```
dunglas: Extra space to remove before ` ?>`

src/Bridge/Symfony/Bundle/Resources/config/workflow.xml
```diff
@@ -0,0 +1,31 @@
...
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+  <services>
+
+    <service id="ApiPlatform\Core\Bridge\Symfony\Workflow\Metadata\Resource\Factory\WorkflowOperationResourceMetadataFactory" decorates="api_platform.metadata.resource.metadata_factory">
```
dunglas: IIRC, we don't use this convention yet.

dunglas:
Wouldn't it be cleaner to hook at the validator level? I'm not very fond of adding custom formats here. IIUC, there will not be available through Swagger and Hydra docs.

soyuka:
> Why does this require extra routing?

Because this is my first naive attempt :). I'm all ears for better ways to achieve this.

> Why not just PATCH $url with {"the_field_that_is_holds_the_state": "state_name"}, and handle all this inside the item normalizers?

At the normalizers it seems dirty but why not at the validation level as recommended by @dunglas ? 
We need to find out what field holds the state and guess if the request actually holds one, then validate through `workflow->can`.

teohhanhui:
I'd love to see the states/transitions abstracted away. The API consumer should not need to understand such concepts. Specifically, it should support dynamic state and implicit transitions. (I hope I'm making sense.)

bendavies:
Further, provided you can denormalize state along other fields:
```json
{"foo": "bar", "baz": 123, "the_field_that_is_holds_the_state": "transition_name"}
```
you would have to denormalize all the other fields first, then validate the transition somehow.

soyuka:
> I'd love to see the states/transitions abstracted away.
:+1: 

> you would have to denormalize all the other fields first, then validate the transition somehow.

The thing is that when working with a Workflow, you don't really "patch a state" but you "apply a transition" to an object. For example "Order validate" is an instruction that will move an Order from state "created" to state "valid". States are dynamic, transition is implicit :thinking: .

> Specifically, it should support dynamic state and implicit transitions. (I hope I'm making sense.)

Did I understood you?

teohhanhui:
> The thing is that when working with a Workflow, you don't really "patch a state" but you "apply a transition" to an object. For example "Order validate" is an instruction that will move an Order from state "created" to state "valid".

Then it's at odds with REST principles. In order to harmonize workflow transitions with REST principles, we'd have to have implicit transitions, that is, the API consumer should never say

```json
{
  "transition": "[draft] to [published]"
}
```

but it should be something implicit based on a property change, e.g.

```json
{
  "published": true
}
```

soyuka:
> but it should be something implicit based on a property change, e.g.

But, the entity doesn't have a property for each transition? 

I'd say that the following is even better:

`PATCH /blogpost/1/state "published"`.

teohhanhui:
> But, the entity doesn't have a property for each transition?

Those properties don't need to be real. :smile:

soyuka:
How do you link your last 2 examples to the workflow? 

> Like I've said, it seems really undesirable to me to expect the API consumer to have to deal with states/transitions.

So, we don't integrate the workflow component then?

teohhanhui:
> How do you link your last 2 examples to the workflow?

Perhaps through configuration.

> So, we don't integrate the workflow component then?

If we don't find a good way to integrate it, then I think we shouldn't.

src/Bridge/Symfony/Workflow/EventListener/WorkflowStateListener.php
```diff
@@ -0,0 +1,67 @@
...
+            || !($attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !isset($attributes['item_operation_name'])
+            || 'workflow_state_patch' !== $attributes['item_operation_name']
+        ) {
+            return;
+        }
+
+        $workflowTransition = $this->serializer->deserialize($request->getContent(), WorkflowTransition::class, $request->attributes->get('_format'));
+
+        if (!$transition = (string) $workflowTransition) {
```
toby-griffiths: If you split this into 2 lines would make it more readable…

```
$transition = (string) $workflowTransition;

if (!$transition) {
    // ...
```

src/Bridge/Symfony/Workflow/PathResolver/WorkflowOperationPathResolver.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public function resolveOperationPath(string $resourceShortName, array $operation, $operationType/*, string $operationName = null*/): string
+    {
+        $path = $this->decorated->resolveOperationPath($resourceShortName, $operation, $operationType, null);
+
+        if (!isset($operation['_path_suffix'])) {
+            return $path;
+        }
+
+        if (self::PATH_FORMAT_SUFFIX === substr($path, -10)) {
```
toby-griffiths: Using `-strlen(self::PATH_FORMAT_SUFFIX)` would avoid possible gottchas if self::PATH_FORMAT_SUFFIX change in the future, for any reason.

wesnick:
I am working on an integration of workflow component with api platform for my project, I extracted some of the code I am using to create this integration into a bundle https://github.com/wesnick/api-platform-workflow-extension (don't try to use it yet, not finished).  My approach is a bit opinionated, but I would appreciate any comments.  If it is interesting for the maintainers I could open a pull request against API Platform.

teohhanhui:
@wesnick That's not RESTful :smile:

alanpoulain:
I think it could be useful to have it at least as an extension (and to add its link in the documentation).

PR #2043, opened 2018-06-25
Fix test since last Symfony Release

jewome62:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | **no**
| Fixed tickets | 
| License       | MIT

Fail since 4.1.1 : for this commi : https://github.com/symfony/symfony/commit/146e01cb4442fbcb3b05dd8ae2569d9894374044

The test don't work for PHP 7.0 and Symfony before 3 month ago. ( because NO_AUTO_CACHE_CONTROL_HEADER is implemented on March 2018)
I don't know if this is a bug which need to be fix on Symfony

### DIFF
--- a/src/HttpCache/EventListener/AddHeadersListener.php
+++ b/src/HttpCache/EventListener/AddHeadersListener.php
@@ -74,6 +74,9 @@ final class AddHeadersListener
             $this->public ? $response->setPublic() : $response->setPrivate();
         }
 
-        $response->headers->set(SessionListener::NO_AUTO_CACHE_CONTROL_HEADER, '');
+        if(\defined(SessionListener::class.'::NO_AUTO_CACHE_CONTROL_HEADER')){
+            $response->headers->set(SessionListener::NO_AUTO_CACHE_CONTROL_HEADER, '');
+        }
+
     }
 }
--- a/src/HttpCache/EventListener/AddHeadersListener.php
+++ b/src/HttpCache/EventListener/AddHeadersListener.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\HttpCache\EventListener;
 
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\EventListener\SessionListener;
 
 /**
  * Configures cache HTTP headers for the current response.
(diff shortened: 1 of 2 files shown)


teohhanhui:
Uhh, you beat me to it. 😂

soyuka:
I don't get why php 7 is failing... Locally the same behat test is successful.

jewome62:
I currently test that locally ;)

Le mar. 26 juin 2018 à 10:32, Antoine Bluchet <notifications@github.com> a
écrit :

teohhanhui:
As I've discussed with @jewome62, there's no good fix for this now. So let's wait for this to be sorted out in Symfony.

dunglas:
There is a good fix (for our test suite at least). The problem belongs here: https://github.com/api-platform/core/blob/master/tests/Fixtures/app/AppKernel.php#L117

When routes are secured, Symfony starts a session, and when a session is started, the cache is disabled by default. It looks like a sensitive default to me. We need to do 2 different things:

* Change our `access_control` rules in the test suite to disable the security for public routes. It will fix the test suite.
* Update https://github.com/api-platform/docs/blob/2.2/core/performance.md to explain this, and to tell about the private Symfony header for advanced cases

teohhanhui:
I'm inclined to think disabling access control falls into the "not a real fix" category.

dunglas:
Why? It would be more correct on our side (the resources served by our test app are public and don't vary depending of the logged in user, so the security must be disabled).

teohhanhui:
Well, yes, it's okay for our case. But it's still bad for other apps, including apps using API Platform. Something needs to be done in Symfony.

teohhanhui:
Anyway, I forgot about this, but the session seems to be started by Symfony itself: https://github.com/symfony/symfony/blob/2b9c142f07b7dbc11d4d392b77629fa91a5dcb41/src/Symfony/Component/HttpKernel/EventListener/AbstractTestSessionListener.php#L74

teohhanhui:
https://github.com/symfony/symfony/pull/27714

dunglas:
Let's fix our test suite then. Symfony bugs must be fixed on the SF repo.

soyuka:
> Let's fix our test suite then. Symfony bugs must be fixed on the SF repo.

We can't :D.

teohhanhui:
Indeed we can't. We just have to skip that test for now.

soyuka:
thanks for the investigation @jewome62 ! Lucky for us @dunglas fixed it in symfony and we're going to temporary use the http foundation dev branch while waiting for it!

PR #2044, opened 2018-06-26
Add web server bundle for debug purpose

soyuka:


tests/Fixtures/app/public/app.php
```diff
@@ -0,0 +1,32 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+use Symfony\Component\HttpFoundation\Request;
+
+require __DIR__.'/../bootstrap.php';
+
+$env = $_SERVER['APP_ENV'] ?? 'test';
```
dunglas: We don't need the trusted proxies logic in the test app. I would remove it.

composer.json
```diff
@@ -66,6 +66,7 @@
         "symfony/twig-bundle": "^3.4 || ^4.0",
         "symfony/validator": "^3.3 || ^4.0",
         "symfony/web-profiler-bundle": "^3.3 || ^4.0",
+        "symfony/web-server-bundle": "^4.1",
```
dunglas: Can you allow lower versions? This constraint will prevent older versions of the deps to this bundle to be installed.

dunglas:
After a second though, can't we just provide `app.php` and a one liner to start the server with `php -S` instead of adding a new dependency?

tests/Fixtures/app/public/index.php
```diff
@@ -0,0 +1,21 @@
...
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
```
dunglas: Missing blank line after.

PR #2046, opened 2018-06-26
Small code tweaks

soyuka:
Add final to the WriteListener
Don't call getResponse in the cache, use the variable instead

### DIFF
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -22,7 +22,7 @@ use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Baptiste Meyer <baptiste.meyer@gmail.com>
  */
-class WriteListener
+final class WriteListener
 {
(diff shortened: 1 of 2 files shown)


src/EventListener/WriteListener.php
```diff
@@ -22,7 +22,7 @@
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Baptiste Meyer <baptiste.meyer@gmail.com>
  */
-class WriteListener
+final class WriteListener
```
teohhanhui: Can we really do this? It's a BC break, no?

soyuka: Yes! I think we did it in another file recently and we accepted it anyway.

dunglas: We considered it as a bug fix... Because our policy is to make all classes `final` by default. But it's border line I agree.

What I suggest: make it `final`, if someone complain switch to `@final`. It's a brand new class, the chances somebody actually extended it are low.

teohhanhui:
Thanks @soyuka

Issue #2049, opened 2018-06-27
ExistsFilter on inverse side of bidirectional One-To-One association has problem


sryabov:
For this two entities
```php
<?php
/** @Entity */
class Customer
{
    // ...

    /**
     * One Customer has One Cart.
     * @ORM\OneToOne(targetEntity="Cart", mappedBy="customer")
     * @ApiFilter(ExistsFilter::class)
     */
    private $cart;

    // ...
}

/** @Entity */
class Cart
{
    // ...

    /**
     * One Cart has One Customer.
     * @ORM\OneToOne(targetEntity="Customer", inversedBy="cart")
     * @ORM\JoinColumn(nullable=false)
     */
    private $customer;

    // ...
}
```
request `/customers?cart[exists]=true` will generate Doctrine exception:

_A single-valued association path expression to an inverse side is not supported in DQL queries_

But this small change in [ExistsFilter.php](https://github.com/api-platform/core/blob/0a69709067335fb2ccb5bd48b0850b47f3aaf72d/src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php#L105-L117) fix it:
```php
if ($metadata->hasAssociation($field)) {
    if ($metadata->isCollectionValuedAssociation($field)) {
        $queryBuilder
            ->andWhere(sprintf('%s.%s %s EMPTY', $alias, $field, $value ? 'IS NOT' : 'IS'));

        return;
    }

    if ($metadata->isAssociationInverseSide($field)) {
        $alias = \ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryBuilderHelper::addJoinOnce(
            $queryBuilder,
            $queryNameGenerator,
            $alias,
            $field,
            \Doctrine\ORM\Query\Expr\Join::LEFT_JOIN
        );

        $queryBuilder
            ->andWhere(sprintf('%s %s NULL', $alias, $value ? 'IS NOT' : 'IS'));

        return;
    }

    $queryBuilder
        ->andWhere(sprintf('%s.%s %s NULL', $alias, $field, $value ? 'IS NOT' : 'IS'));

    return;
}
```
I think this code is dirty but it's idea for fix error =)

alanpoulain:
Has been fixed.

PR #2051, opened 2018-06-28
Fix ExistsFilter for inverse side of OneToOne association

sryabov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #2049
| License       | MIT


soyuka:
May you add a test to prevent regression? Thanks!

sryabov:
I make something wrong with branch and don't know how to revert it =(

teohhanhui:
@sryabov I can handle that if you want.

teohhanhui:
@sryabov Reverted the incorrect merge.

sryabov:
@teohhanhui it seems to be fixed, thx =)

sryabov:
Phew, I made it =) All test are passed. Sorry for many commits, but local behat is not working correct:

```
Feature: Default values of HTTP cache headers
  In order to make API responses cacheable
  As an API software developer
  I need to be able to set default cache headers values

  @createSchema @dropSchema
  Scenario: Cache headers default value                                                   # features/http_cache/headers.feature:8
    When I send a "GET" request to "/relation_embedders"                                  # Behatch\Context\RestContext::iSendARequestTo()
    Then the response status code should be 200                                           # Behat\MinkExtension\Context\MinkContext::assertResponseStatus()
    And the header "Etag" should be equal to '"21248afbca1f242fd3009ac7cdf13293"'         # Behatch\Context\RestContext::theHeaderShouldBeEqualTo()
    And the header "Cache-Control" should be equal to "max-age=60, public, s-maxage=3600" # Behatch\Context\RestContext::theHeaderShouldBeEqualTo()
      The header 'Cache-Control' should be equal to 'max-age=60, public, s-maxage=3600', but it is: 'max-age=0, must-revalidate, private, s-maxage=3600'
      
      +--[ HTTP/1.1 200 | http://example.com/relation_embedders | KernelDriver ]
      |
      |  {"@context":"\/contexts\/RelationEmbedder","@id":"\/relation_embedders","@type":"hydra:Collection","hydra:member":[],"hydra:totalItems":0}
      |
    And the header "Vary" should be equal to "Accept, Cookie"                             # Behatch\Context\RestContext::theHeaderShouldBeEqualTo()
```

And one more thing: I think that you need to split Dummy fixture for normilize tests and for relations (include filters) tests.

dunglas:
Thanks @sryabov!

PR #2055, opened 2018-06-29
WhereFilter - Allow to build complex search query

amenophis:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not tested yet
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Hi,
I open this PR to start discuss about implementing a filter like provided in Loopback (http://loopback.io/doc/en/lb3/Where-filter.html).

The idea is to allow building complex query from client side like following examples:
```
http://localhost:8080/greetings?where[name][eq]=toto
```
```
http://localhost:8080/greetings?where[or][name][eq]=test&where[or][name][eq]=test
```
```
http://localhost:8080/greetings?where[and][name][eq]=test&where[and][id][lt]=10
```

TO-DO:
- [ ] Make allowed fields configurable
- [ ] Allow usage of sub resource fields
- [ ] Write tests
- [ ] Write documentation

I provide a first shoot implementation, wait for your comments !

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
@@ -57,11 +57,11 @@ final class WhereFilter implements ContextAwareFilterInterface
             switch ($type) {
                 case AndCondition::TYPE:
                     $conditions = $this->buildWhereFromArray($alias, $conditionArray)->getConditions();
-                    $where->addCondition(new AndCondition(...$conditions));
+                    $where->addCondition(new AndCondition($conditions));
                     break;
                 case OrCondition::TYPE:
                     $conditions = $this->buildWhereFromArray($alias, $conditionArray)->getConditions();
-                    $where->addCondition(new OrCondition(...$conditions));
+                    $where->addCondition(new OrCondition($conditions));
                     break;
                 case EqCondition::TYPE:
                     $where->addCondition(new EqCondition($alias.'.'.$conditionType, (array) $conditionArray[$type]));
--- a/src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Filter;
+
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Where;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Doctrine\ORM\QueryBuilder;
+use Symfony\Component\HttpFoundation\RequestStack;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\AndCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\ConditionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\EqCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\GtCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\GteCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\LtCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\LteCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\NeqCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\OrCondition;
+
+class WhereFilter implements FilterInterface
+{
+    private const QUERY_STRING_WHERE = 'where';
+
+    protected $requestStack;
+
+    public function __construct(RequestStack $requestStack)
+    {
+        $this->requestStack = $requestStack;
+    }
+
+    public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
+    {
+        if (null === $request = $this->requestStack->getCurrentRequest()) {
+            return;
+        }
+
+        if (null === $whereData = $request->query->get(self::QUERY_STRING_WHERE)) {
+            return;
+        }
+
+        $rootAlias = $queryBuilder->getRootAliases()[0];
+        $where = $this->buildWhereFromArray($rootAlias, $whereData);
+
+        $where->apply($queryBuilder);
+    }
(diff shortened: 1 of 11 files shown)


src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
```diff
@@ -0,0 +1,89 @@
...
+        return []; // TODO
+    }
+
+    public function buildWhereFromArray(string $alias, array $whereArray): Where
+    {
+        $where = new Where();
+        foreach ($whereArray as $conditionType => $conditionArray) {
+            $type = array_keys($conditionArray)[0];
+
+            switch (true) {
```
dunglas: `switch ($type)` would be cleaner and faster.

soyuka:
Good ideas here also: https://postgrest.readthedocs.io/en/v5.0/api.html#

src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
```diff
@@ -0,0 +1,89 @@
...
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\NeqCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\OrCondition;
+
+class WhereFilter implements FilterInterface
+{
+    private const QUERY_STRING_WHERE = 'where';
+
+    protected $requestStack;
+
+    public function __construct(RequestStack $requestStack)
```
dunglas: New filters shouldn't depend of HttpFoundation anymore. You can access the query parameters directly in the `$context` parameter now.

src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
```diff
@@ -0,0 +1,89 @@
...
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\ConditionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\EqCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\GtCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\GteCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\LtCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\LteCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\NeqCondition;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition\OrCondition;
+
+class WhereFilter implements FilterInterface
```
dunglas: `final`

src/Bridge/Doctrine/Orm/Filter/WhereFilter.php
```diff
@@ -0,0 +1,89 @@
...
+{
+    private const QUERY_STRING_WHERE = 'where';
+
+    protected $requestStack;
```
dunglas: `private`

src/Bridge/Doctrine/Orm/Filter/WhereFilter/Condition/AndCondition.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition;
+
+use Doctrine\ORM\Query\Expr;
+
+class AndCondition implements ConditionInterface
```
dunglas: `final` (same for all new classes)

src/Bridge/Doctrine/Orm/Filter/WhereFilter/Condition/AndCondition.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public const TYPE = 'and';
+
+    /**
+     * @var ConditionInterface[]
+     */
+    protected $conditions;
+
+    public function __construct(ConditionInterface ...$conditions)
```
dunglas: We usually use an array instead of variadic parameters (it eases adding new parameter in next iterations).

src/Bridge/Doctrine/Orm/Filter/WhereFilter/Condition/ConditionInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\WhereFilter\Condition;
+
+use Doctrine\ORM\Query\Expr;
+
+interface ConditionInterface
+{
+    public function getType(): string;
+
+    public function apply(Expr $expr);
```
dunglas: Can you document the return type?

dunglas:
Actually, do you think it's possible to embed any existing or future filter in the `where` clause instead of recreating new ones? It would be more powerful and easier to maintain.

amenophis:
@dunglas, thinks it's a good idea, will try to use existing filters logic in this WhereFilter

teohhanhui:
Strongly -1 for a SQL-like Frankenfilter.

dunglas:
I like the name « FrankenFilter », but +1 to include it because:

* it’s a feature frequently asked by our users since the very first release
* it’s opt-in, nobody is forced to use it
* it’s super convenient for admins (especially with GraphQL or the « properties » filter
* alternative tools such as Loopback have it

However, we must write in bold in the documentation that this filter can be dangerous (DDOS vector) and should not be used for public APIs

teohhanhui:
The fact that it's a free-form filter is already a big no-no for me. But now that you mention the security risks, that makes it even worse.

But if we really want to do something like this, I wish we would stick to a format that is at least standardized, in other words oData.

flug:
@amenophis is this pull request still valid?

soyuka:
Let's keep this as-is while we wait to reach a consensus over https://github.com/api-platform/core/issues/2400

amenophis:
Just rebased it and close outdated comments.

silverbackdan:
Hi, I was hoping to better understand #2400 composable issue and how it relates to this PR? These features seem to have been stale for just over a year.

Are there issues in this update which could not be resolved or have you just not had the time needed for this feature?

I only ask as the issue was brought to my attention by @alanpoulain while beginning another PR here https://github.com/api-platform/core/pull/3924

soyuka:
see also https://github.com/api-platform/api-platform/discussions/1724 imo we need to pick or write a filter RFC and then work on it

metaclass-nl:
I built a [WhereFilter (FilterLogic)](https://github.com/metaclass-nl/filter-bundle/tree/query-expression-generator) that reuses the existing ORM Filter classes and combines them with compound logic. If you don't like the way it manipulates the QueryBuilder (i don't): it can be combined with QueryExpressionGeneratorInterface (see [my comment on RFC Filter composition](https://github.com/api-platform/core/issues/2400#issuecomment-850821749)), resulting in [this branch](https://github.com/metaclass-nl/filter-bundle/tree/query-expression-generator).

Issue #2060, opened 2018-06-30
Graphql query fails if the collection is empty using the totalCount property


epourail:
Using a nested filter in a graphql query that returns an empty collection will fail if the totalCount property is requested. 

In the following example, I want to filter the categories according to the topcategory's title and get the collection count
```
{
  categories(topcategory_title: "UnknownTopCategory") {
    totalCount
    edges {
      node {
        title
      }
    }
  }
}

```

By querying the `totalCount` property, the following error is displayed : 
```
{
	"errors": [
		{
			"debugMessage": "Cannot return null for non-nullable field CategoryConnection.totalCount.",
			"message": "Internal server error",
			"category": "internal",
			"locations": [
				{
					"line": 3,
					"column": 3
				}
			],
			"path": [
				"categories",
				"totalCount"
			]
		}
	],
	"data": {
		"categories": null
	}
}
```

PR #2061, opened 2018-06-30
Initialize the 'totalCount' property to the default value to 0.0

epourail:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2060
| License       | MIT
| Doc PR        | ~

### DIFF
--- a/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
@@ -107,7 +107,7 @@ final class CollectionResolverFactory implements ResolverFactoryInterface
                 $offset = 1 + (int) $after;
             }
 
-            $data = ['edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];
+            $data = ['totalCount' => 0.0, 'edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];
             if ($collection instanceof PaginatorInterface && ($totalItems = $collection->getTotalItems()) > 0) {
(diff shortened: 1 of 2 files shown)


src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -107,7 +107,7 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
                 $offset = 1 + (int) $after;
             }
 
-            $data = ['edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];
+            $data = ['totalCount' => 0.0, 'edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];
```
jewome62: Why do you use a float ?

soyuka: Because every number in our pagination is a float ?

dunglas: Indeed it's ok like this. It allows to deal with large(r) numbers, and in JSON it will be a `Number` anyway.

dunglas:
Thanks @epourail

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
epourail: @jewome62 @soyuka Because, when the collection is not empty, the `getTotalItems` method called further returns a float.

PR #2068, opened 2018-07-04
Fix some SensioLabs Insight violations

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Identifier/IdentifierConverter.php
+++ b/src/Identifier/IdentifierConverter.php
@@ -11,23 +11,20 @@
 
 declare(strict_types=1);
 
-namespace ApiPlatform\Core\Identifier\Normalizer;
+namespace ApiPlatform\Core\Identifier;
 
 use ApiPlatform\Core\Api\IdentifiersExtractorInterface;
 use ApiPlatform\Core\Exception\InvalidIdentifierException;
-use ApiPlatform\Core\Identifier\CompositeIdentifierParser;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use Symfony\Component\PropertyInfo\Type;
 
 /**
- * Identifier normalizer.
+ * Identifier converter that chains identifier denormalizers.
  *
  * @author Antoine Bluchet <soyuka@gmail.com>
  */
-class ChainIdentifierDenormalizer
+final class IdentifierConverter implements IdentifierConverterInterface
 {
-    const HAS_IDENTIFIER_DENORMALIZER = 'has_identifier_denormalizer';
-
     private $propertyMetadataFactory;
     private $identifiersExtractor;
     private $identifierDenormalizers;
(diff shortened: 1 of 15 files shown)


src/Identifier/IdentifierConverter.php
```diff
@@ -45,13 +45,11 @@ public function convert(string $data, string $class): array
...
 
-        if (!$keys) {
-            throw new InvalidIdentifierException(sprintf('Resource "%s" has no identifiers.', $class));
-        }
-
-        if (\count($keys) > 1) {
+        if ($numIdentifiers = \count($keys) > 1) {
             $identifiers = CompositeIdentifierParser::parse($data);
-        } else {
+        } else if (0 === $numIdentifier) {
```
dunglas: `elseif`

src/Identifier/IdentifierConverter.php
```diff
@@ -11,22 +11,21 @@
...
- * Identifier normalizer.
+ * Identifier converter that chains identifier denormalizers.
  *
  * @author Antoine Bluchet <soyuka@gmail.com>
  */
-class ChainIdentifierDenormalizer
+final class IdentifierConverter implements IdentifierConverterInterface
 {
-    const HAS_IDENTIFIER_DENORMALIZER = 'has_identifier_denormalizer';
+    const HAS_IDENTIFIER_CONVERTER = 'has_identifier_converter';
```
dunglas: Could be moved to the interface right?

src/Identifier/IdentifierConverter.php
```diff
@@ -40,19 +39,17 @@ public function __construct(IdentifiersExtractorInterface $identifiersExtractor,
...
+    public function convert(string $data, string $class): array
     {
         $keys = $this->identifiersExtractor->getIdentifiersFromResourceClass($class);
 
-        if (!$keys) {
-            throw new InvalidIdentifierException(sprintf('Resource "%s" has no identifiers.', $class));
-        }
-
-        if (\count($keys) > 1) {
+        if ($numIdentifiers = \count($keys) > 1) {
```
dunglas: `numIdentifiers` is undefined.

src/Identifier/IdentifierConverter.php
soyuka: I'd prefer not to so that we can remove it later

src/Identifier/IdentifierConverter.php
```diff
@@ -40,18 +37,16 @@ public function __construct(IdentifiersExtractorInterface $identifiersExtractor,
...
```
dunglas: missing parenthesis: `($numIdentifiers = \count($keys))`

dunglas:
Thanks @soyuka

PR #2070, opened 2018-07-05
[graphql] Add graphql filter with multiple values

ArnoudThibaut:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/524

Using the SearchFilter with multiple values was not possible with graphql.

Using the same synthax than the query filter (`foo[]=bar`) was not possible because characters `[` and `]` in a graphql query are invalid and it seems that it's not possible to use the same key for both single value filter and multi values filter.

The idea is to prefix the field name with a special keyword to make the distinction between single and multiple values.

### DIFF
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -231,12 +231,15 @@ final class SchemaBuilder implements SchemaBuilderInterface
                         $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
                         $graphqlFilterType = $this->convertType($filterType, false, null, $depth);
 
-                        if ('[]' === $newKey = substr($key, -2)) {
-                            $key = $newKey;
+                        if ('[]' === substr($key, -2)) {
                             $graphqlFilterType = GraphQLType::listOf($graphqlFilterType);
+                            $key = 'array'.substr($key, 0, -2);
                         }
 
                         parse_str($key, $parsed);
+                        if (array_key_exists($key, $parsed) && \is_array($parsed[$key])) {
+                            $parsed = [$key => ''];
+                        }
                         array_walk_recursive($parsed, function (&$value) use ($graphqlFilterType) {
                             $value = $graphqlFilterType;
                         });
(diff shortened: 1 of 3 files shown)


dunglas:
Good one, it can even be considered a bug fix to me. Regarding the name, it will be cleaner and safer to add a separator, I would also use the term `list` instead of array, as in the GraphQL spec.

So I would rename `arrayrelatedDummy_name` to `relatedDummy_name_list`. WDYT?

ArnoudThibaut:
Yep, `list` is better than array.

For the separator, `_` is already used by nested properties, could it be a problem to use it for that too?

dunglas:
I don't think so. At worst, for a property called `list`, it would generate something like `class_list_list`. It will still work and it cannot have conflicts.

ArnoudThibaut:
Ok, the change is done.

ArnoudThibaut:
Yes sure,
I will make a PR for the doc tonight or tomorrow.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -231,12 +231,15 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
                         $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
                         $graphqlFilterType = $this->convertType($filterType, false, null, $depth);
 
-                        if ('[]' === $newKey = substr($key, -2)) {
-                            $key = $newKey;
+                        if ('[]' === substr($key, -2)) {
                             $graphqlFilterType = GraphQLType::listOf($graphqlFilterType);
+                            $key = substr($key, 0, -2).'_list';
```
toofff: `$key = "{${substr($key, 0, -2)}}_list";`?

dunglas: It's not in JS :D

toofff: Hooo shit :smile:

ArnoudThibaut:
Doc PR just here : https://github.com/api-platform/docs/pull/524

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -231,12 +231,15 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
...
                         }
 
                         parse_str($key, $parsed);
+                        if (array_key_exists($key, $parsed) && \is_array($parsed[$key])) {
```
meyerbaptiste: `isset($parsed[$key])` will stop the condition earlier with a null value.

dunglas:
Thanks @ArnoudThibaut

PR #2072, opened 2018-07-05
New "_api_persist" request attribute to skip the WriteListener

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | api-platform/docs#523

Add a new `_api_persist` flag to skip the `WriteListener`. See also https://stackoverflow.com/q/51147945/1352334 for the use case.

### DIFF
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -37,7 +37,7 @@ final class WriteListener
     public function onKernelView(GetResponseForControllerResultEvent $event)
     {
         $request = $event->getRequest();
-        if ($request->isMethodSafe(false) || !$request->attributes->has('_api_resource_class')) {
+        if ($request->isMethodSafe(false) || !$request->attributes->has('_api_resource_class') || !$request->attributes->getBoolean('_api_persist', true)) {
             return;
         }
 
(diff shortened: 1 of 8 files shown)


walva:
Thanks, very helpful ! I used it today. It was a bit hard to find in the documentation, I end up looking outside, that's why I end up here. For those who are looking for the related page in the documentation, here is the link: https://api-platform.com/docs/core/events/

Example of usage:

```
#[ApiResource(
  itemOperations: [
      "workflow_transition" => [
          "method"                  => "PUT",
          "path"                    => "/entity/{id}/{workflow}/{transition}",
          "controller"              => ApplyWorkflowTransition::class,
          "input"                   => ApplyTransitionAction::class,
          "defaults"                 => [
              "_api_persist" => false,
          ],
          'denormalization_context' => [
              'groups' => ['workflow', self::DENORMALIZATION_CONTEXT],
          ],
          'normalization_context'   => [
              'groups' => ['workflow', self::DENORMALIZATION_CONTEXT],
          ],
      ]
   ]
)
```

PR #2081, opened 2018-07-09
Unset the filter (with a dot character), in case of a relation, once it has been sanitized

epourail:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2082
| License       | MIT
| Doc PR        | 

Against some graphql client implementation (like apollo for ios), the presence of a relation/nested property with the `.` (dot character) in the schema is not supported.
So, the array of filter definitions should be cleaned by removing the nested property once it has been sanitized (the `.` character is replaced by the `_` charceter)

### DIFF
--- a/tests/GraphQl/Type/SchemaBuilderTest.php
+++ b/tests/GraphQl/Type/SchemaBuilderTest.php
@@ -95,6 +95,32 @@ class SchemaBuilderTest extends TestCase
         $this->assertArrayNotHasKey('objectProperty', $type->getFields());
     }
 
+    public function testConvertFilterArgsToTypes()
+    {
+        $propertyMetadataMockBuilder = function ($builtinType, $resourceClassName) {
+            return new PropertyMetadata();
+        };
+        $mockedSchemaBuilder = $this->createSchemaBuilder($propertyMetadataMockBuilder, false);
+
+        $reflectionClass = new \ReflectionClass(SchemaBuilder::class);
+        $method = $reflectionClass->getMethod('convertFilterArgsToTypes');
+        $method->setAccessible(true);
+        $filterArgs = [
+            'aField' => 'string',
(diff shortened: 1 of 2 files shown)


dunglas:
Good catch, thanks! Can you change the target branch to 2.3, as it's a bug fix?

epourail:
@dunglas I updated the PR to change the target branch to `2.3`

For information, concerning the non-regression test, unless I am mistaken, I did not find any unit test to validate the graphql filters within the schema builder. So, in this case, it will take times to understand and implements a non-regression test as I haven't got any basis/examples.

alanpoulain:
I think you can just add unit tests for this method (using reflection to make it public).
You could also add functional tests using GraphQL introspection.

epourail:
@dunglas @alanpoulain as you recommend, I update the PR to add a non-regression test on the concerned method (using reflection to make it public).

tests/GraphQl/Type/SchemaBuilderTest.php
```diff
@@ -95,6 +95,47 @@ public function testGetSchemaResourceClassNotFound()
         $this->assertArrayNotHasKey('objectProperty', $type->getFields());
     }
 
+    public function testConvertFilterArgsToTypes()
+    {
+        $propertyMetadataMockBuilder = function ($builtinType, $resourceClassName) {
+            return new PropertyMetadata(
```
alanpoulain: You can only do:
```php
$propertyMetadataMockBuilder = function ($builtinType, $resourceClassName) {
    return new PropertyMetadata();
}
```

epourail:
@alanpoulain Updated (oups, due to my habit, I locally rebase/squash and I lost your comment)

dunglas:
Thanks @epourail

Issue #1574, opened 2017-12-19
[RFC] Authorize interface as resource


Nek-:
## Problem

In my application I use the [KnpDictionaryBundle](https://github.com/biig-io/DictionaryBundle). This bundle provide an interface to implement dictionaries but no resource class, which is incompatible with api platform ATM.

I think api platform should authorize interface as a resource because I'm pretty sure it's not the only software to act like this.

## Solution

Modify the `ExtractorResourceMetadataFactory` to support interfaces:

```php
// before
if (!class_exists($resourceClass) || !$resource = $this->extractor->getResources()[$resourceClass] ?? false)

// After
if (!(class_exists($resourceClass) || interface_exists($resourceClass)) || !$resource = $this->extractor->getResources()[$resourceClass] ?? false)
```

What do you think?

soyuka:
They may be other interferences (like https://github.com/api-platform/core/blob/9c9f2fd9ef507aae06821dccc405f2d8df1665de/src/Api/ResourceClassResolver.php#L40)

Though I don't really get how an interface could be used as a resource without overriding some parts of the call chain. Could you elaborate?

Nek-:
If you want to use an interface as a resource you're probably going to also add a normalizer (which is not bad, right? :) ). Did you check out the [dictionary bundle](https://github.com/biig-io/DictionaryBundle)? I'm talking about the `Dictionary` interface (for my case).

According to my tests, there is no other interference. (I'll push my tests a little bit more today)

Please notice that abstract classes are supported (which make not supporting interface weirder).

soyuka:
I don't see why we should not support them ofc. But yes you're right it doesn't make sense alone!

Nek-:
There's indeed something to change. The following line is wrong:

https://github.com/api-platform/core/blob/master/src/Api/ResourceClassResolver.php#L73

```php
// This is more accurate
if (is_subclass_of($type, $resourceClass))
```

soyuka:
:smile: feel free to patch this on master!

PR #2086, opened 2018-07-10
Ignore property that does not exist for interfaces (identifiers for interfaces)

Nek-:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | Related to https://github.com/api-platform/core/issues/1574
| License       | MIT
| Doc PR        | N/A

### Motivations

TL;DR: Before: no "item" route working because we can't define an identifier. After: we can define a property for an interface and it will not be considered as "not found property".

1. It makes no sense to check if an interface exists in case it's an interface ;
2. If the property is not recognized as one and there is no identifier, then API Platform will not be able to call the DataProvider for a given item and interface as a resource is an incomplete feature.

### Usage

```yaml
Me\InterfaceAsAResource:
    properties:
         # Will work even if foo is not defined in the interface!
         foo:
             identifier: true
```

### What's next?

If you like this kind of features, I will probably work on "virtual properties" next. I see something like

```yaml
Me\InterfaceAsAResource:
    properties:
         foo:
             identifier: true
             proxy_method: 'customMethodToGetFoo'
```

What do you think? Maybe you want a separate RFC issue to discuss it?

### DIFF
--- a/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
@@ -48,9 +48,11 @@ final class ExtractorPropertyMetadataFactory implements PropertyMetadataFactoryI
             }
         }
 
+        $isInterface = interface_exists($resourceClass);
+
         if (
-            !property_exists($resourceClass, $property) ||
-            !$propertyMetadata = $this->extractor->getResources()[$resourceClass]['properties'][$property] ?? false
+            !property_exists($resourceClass, $property) && !$isInterface ||
+            null === ($propertyMetadata = $this->extractor->getResources()[$resourceClass]['properties'][$property] ?? null)
         ) {
             return $this->handleNotFound($parentPropertyMetadata, $resourceClass, $property);
         }
--- a/tests/Fixtures/FileConfigurations/interface_resource.yml
+++ b/tests/Fixtures/FileConfigurations/interface_resource.yml
@@ -1,2 +1,8 @@
 resources:
-    'ApiPlatform\Core\Tests\Fixtures\DummyResourceInterface': ~
+    'ApiPlatform\Core\Tests\Fixtures\DummyResourceInterface':
+        properties:
+            something:
+                identifier: true
+            somethingElse:
+                writable: false
+                readable: true
--- a/features/main/table_inheritance.feature
+++ b/features/main/table_inheritance.feature
@@ -327,3 +327,38 @@ Feature: Table inheritance
        "required": ["hydra:member"]
      }
      """
+
+  Scenario: Get an interface resource item
+    When I send a "GET" request to "/resource_interfaces/some-id"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "context": {
+          "type": "string",
+          "pattern": "ResourceInterface$"
+        },
+        "@id": {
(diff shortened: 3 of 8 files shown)


src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
```diff
@@ -48,9 +48,15 @@ public function create(string $resourceClass, string $property, array $options =
             }
         }
 
+        try {
+            $isInterface = (new \ReflectionClass($resourceClass))->isInterface();
+        } catch (\ReflectionException $e) {
+            $isInterface = false;
+        }
+
         if (
```
dunglas: `!$isInterface` costs less, you can check it first.

Nek-: Fixed 👍

teohhanhui:
I thought we already support virtual properties?

teohhanhui:
@Nek- I think there's no documentation, not even tests for it. :laughing: :see_no_evil:

But looking at the code (and as I've heard from other's experience), it should work. (Not when using the `Extractor` factory, obviously, as you've discovered.)

Nek-:
In any case, it doesn't invalidate this feature as I understand. Are failures on CI related? (due to inversion of condition?)

teohhanhui:
CI failures are unrelated and should be fixed after https://github.com/api-platform/core/pull/2106 is merged.

tests/Fixtures/FileConfigurations/interface_resource.yml
```diff
@@ -1,2 +1,8 @@
 resources:
-    'ApiPlatform\Core\Tests\Fixtures\DummyResourceInterface': ~
```
soyuka: shouldn't we keep a test like this?

Nek-:
It's up to date and issues non-related.

src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
```diff
@@ -48,9 +48,15 @@ public function create(string $resourceClass, string $property, array $options =
             }
         }
 
+        try {
```
dunglas: Can't this entire block be replaced by a simple `interface_exists($resourceClass)` call?

Nek-: Fixed

soyuka:
I think that https://github.com/api-platform/core/pull/2495 fixes this case lmk if it doesn't.

Going to cherry pick the tests anyway!

Nek-:
@soyuka it does not.

src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
```diff
@@ -48,9 +48,11 @@ public function create(string $resourceClass, string $property, array $options =
             }
         }
 
+        $isInterface = interface_exists($resourceClass);
+
         if (
-            !property_exists($resourceClass, $property) ||
-            !$propertyMetadata = $this->extractor->getResources()[$resourceClass]['properties'][$property] ?? false
+            !property_exists($resourceClass, $property) && !$isInterface ||
+            empty($propertyMetadata = $this->extractor->getResources()[$resourceClass]['properties'][$property] ?? [])
```
soyuka: null === (... ?? null)

Nek-: :+1: you right. (can't remember why I did it like that, probably a mistake after testing things)

PR #2087, opened 2018-07-10
Remove duplicate data from JSON API include

ilicmsreten:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

From [JSON-API](http://jsonapi.org/format/#document-compound-documents) documentation 

> A compound document MUST NOT include more than one resource object for each type and id pair.

Also in [json api schema](http://jsonapi.org/schema) or in `tests/Fixtures/JsonSchema/jsonapi.json` is definition about unique items in array.

```json
 "included": {
          "description": "To reduce the number of HTTP requests, servers **MAY** allow responses that include related resources along with the requested primary resources. Such responses are called \"compound documents\".",
          "type": "array",
          "items": {
            "$ref": "#/definitions/resource"
          },
          "uniqueItems": true
        }
```
Currently when multiple objects have relation to one object, on GET request with include parameter same object is present multiple times in include part of JSON API response.

PR #2088, opened 2018-07-11
Remove duplicate data from JSON API include

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes, but the fix is in #2087 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes once #2087 will be merged
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Just add non regression test for #2087

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -220,6 +220,31 @@ final class FeatureContext implements Context, SnippetAcceptingContext
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dunglas:
Maybe can you cherry-pick the @ilicmsreten's so I can merge both together

antograssiot:
here you are @dunglas

dunglas:
Thanks @ilicmsreten, @antograssiot!

PR #2089, opened 2018-07-11
Add subresource collection to Cache-Tags

mauchede:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

When we retrieve a subresource, `Cache-Tags` contains only IRIs of the resources shown (not the IRI collection). In that case, when we add / remove, as we will invalid IRI collection, the subresources will not be impacted.


src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -56,7 +56,7 @@ public function onKernelResponse(FilterResponseEvent $event)
         }
 
         $resources = $request->attributes->get('_resources');
-        if (isset($attributes['collection_operation_name'])) {
+        if (isset($attributes['collection_operation_name']) || (isset($attributes['subresource_context']) && $attributes['subresource_context']['collection'])) {
```
dunglas: if (isset($attributes['collection_operation_name']) || ($attributes['subresource_context']['collection'] ?? false))?

mauchede:
@dunglas: Target has been changed and your comment has been fixed.

Issue #2093, opened 2018-07-12
Mutation doesn't return related entities


Siregacvek:
Good day!

I have a question. Sorry if it was already explained, because I tried to find an answer and was unsuccessful.

I have an easy mutation:
```
mutation{
 updateProduct(input:{
    id:"api/products/33088"
    name:"Product Name 1"
    clientMutationId: "3"
  }){
    id
    storages{
      edges{
        node{
          id
        }
      }
    }
  }
}
```
It changes the data in DB successfully, but I expect that returned json will contain the list of related entities (storages), but it's not:
```
{
  "data": {
    "updateProduct": {
      "id": "/api/products/33088",
      "storages": {
        "edges": null
      }
    }
  }
}
```
When I launch a query the list is there.
Query:
```
{
  product(id: "api/products/33088") {
    id
    name
    storages {
      edges {
        node {
          id
        }
      }
    }
  }
}
```
Returned json:
```
{
  "data": {
    "product": {
      "id": "/api/products/33088",
      "name": "Product Name 1",
      "storages": {
        "totalCount": 1,
        "edges": [
          {
            "node": {
              "id": "/api/storages/105"
            }
          }
        ]
      }
    }
  }
}
```
Why? What I'm missing?

In advance thanks for the answer.

alanpoulain:
It may be related to your configuration.
Do you use serialization groups? Could you paste the configuration of the Product entity?

Siregacvek:
Yes, I use serialization groups. As far as I know for returnig value mutation uses the same serialization group  as query, so if query works fine, the mutation has to work too.
This is my Product conf (I removed unnecessary fields and left only important to save your time):

```
/**
 * @ApiResource(
 *     attributes={
 *       "access_control"="is_granted('ROLE_USER')",
 *       "normalization_context": {"groups"={"read"}},
 *       "denormalization_context": {"groups"={"write"}},
 *       "force_eager"=false
 *     },
 *     collectionOperations={"get"},
 *     itemOperations={
 *          "get"={"access_control"="is_granted('read', object)"}
 *     },
 *     graphql={
 *          "query",
 *          "delete"={"access_control"="is_granted('delete', object)"},
 *          "update"={"access_control"="is_granted('update', object)"},
 *          "create"
 *     }
 * )
 * @ApiFilter(SearchFilter::class,
 *     properties={"name": "ipartial", "productList": "exact", "positions.location": "exact"})
 */
class Product extends AbstractEntity
{
	/**
[...]

PR #2097, opened 2018-07-13
[GraphQL] Fix collection field mutation

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2093
| License       | MIT
| Doc PR        |

Fix collection fields not returned when a mutation is done.

### DIFF
--- a/features/graphql/mutation.feature
+++ b/features/graphql/mutation.feature
@@ -142,7 +142,7 @@ Feature: GraphQL mutation support
 
   @createSchema
   Scenario: Modify an item through a mutation
-    Given there are 1 dummy objects
+    Given there are 1 dummy objects having each 2 relatedDummies
     When I send the following GraphQL request:
     """
     mutation {
@@ -151,6 +151,13 @@ Feature: GraphQL mutation support
         name
         description
         dummyDate
+        relatedDummies {
+          edges {
+            node {
+              name
+            }
+          }
+        }
         clientMutationId
       }
     }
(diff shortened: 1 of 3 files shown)


alanpoulain:
Tests are failing because of unrelated deprecations (Doctrine).

Siregacvek:
Thanks for your PR. I tested it and there is still one issue. From GraphQL perspective the returned value is of `update` operation type, not `query`. So for example the security subsystem (e.g. voter) can deny this operation and dont allow return collection.

alanpoulain:
@Siregacvek I don't understand. What you explain is the expected behavior, doesn't it? You are not doing a query but a mutation. So it's logical to have the operation type as a mutation.

Siregacvek:
@alanpoulain , let me explain. Lets we have a mutation
```
mutation{
 updateEntity(input:{
    id:"entities/1"
    field:"new value"
    clientMutationId: ""
  }){
    id
    field
      relatedEntities{
      edges{
        node{
          id
        }
      }
    }
  }
}
```
I see here 2 requests (`update` and `query`) and I expect the following behavior:
1. I do update mutation on Entity, changing the property `field`. It's an `update` operation for sure.
2. Then I want to return the updated entity with some collection of related entities. Here I expect query-type request, because mutation already processed, I just want to return the current (new) state of entity with its related collection of other entities. Right now for the backend this step looks like an update operation on related collection. But it's not.
 
Am I not right?

alanpoulain:
It doesn't seem logical to me. Returning the state of the entity is part of the mutation process.
This way, you can control how you want the normalization of a mutation and it can differ from the query one (see https://api-platform.com/docs/core/graphql#serialization-groups).
Moreover, it's coherent with the REST part of API Platform: https://api-platform.com/docs/core/serialization#using-different-serialization-groups-per-operation. When you're doing a `PUT`, you can control how you want the normalized data in the `put` operation type, not in the `get` one.

Siregacvek:
@alanpoulain ok, I got you - it's other point of view on a process. Thanks for you work!

alanpoulain:
Could we merge this PR?

soyuka:
thanks @alanpoulain

PR #2098, opened 2018-07-13
Fix tests doctrine deprecation

soyuka:
@dunglas I think that the ClassLoader is used by symfony in https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/Test/DoctrineTestHelper.php#L56 if that's not the case I don't get it (see DoctrineOrmFilterTestCase).

I'm really not sure how we can fix those deprecations right now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,6 @@
     ],
     "require": {
         "php": ">=7.1",
-
-        "doctrine/inflector": "^1.0",
         "psr/cache": "^1.0",
         "psr/container": "^1.0",
         "symfony/http-foundation": "^3.1 || ^4.0",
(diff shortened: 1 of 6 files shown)


dunglas:
See https://github.com/symfony/symfony/pull/27939

composer.json
```diff
@@ -14,8 +14,6 @@
     ],
     "require": {
         "php": ">=7.1",
-
-        "doctrine/inflector": "^1.0",
```
teohhanhui: Why?

dunglas: Yes this must be reverted

TerjeBr:
Could you please merge this into branch v2.2 as well?

We are stuck with symfony 3.4 on our project, so we cannot use version 2.3 of API platform (it requires symfony serializer 4.1). And in our logs we get all the time this annoing:
` php.INFO: User Deprecated: Doctrine\Common\Util\Inflector is deprecated. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Doctrine\\Common\\Util\\Inflector is deprecated. at /var/www/allianz-portal/vendor/doctrine/common/lib/Doctrine/Common/Util/Inflector.php:8)"} []
`

soyuka:
WDYT @dunglas should we patch this on 2.2?

PR #2099, opened 2018-07-14
Inflector as a hard dependency

soyuka:
https://github.com/api-platform/core/pull/2098/files#r202432943

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,8 @@
     ],
     "require": {
         "php": ">=7.1",
+
+        "doctrine/inflector": "^1.0",
         "psr/cache": "^1.0",
         "psr/container": "^1.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -14,6 +14,8 @@
     ],
     "require": {
         "php": ">=7.1",
+
```
teohhanhui: No extra blank line please. 😂

soyuka: there was one before IIRC https://github.com/api-platform/core/blob/master/composer.json#L17

teohhanhui: I don't see why there should be...

Also, https://github.com/api-platform/core/blob/master/composer.json#L29 is in the wrong position for some reason. 😂😂

PR #2111, opened 2018-07-18
Remove superfluous PHPDoc tags

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a


src/PathResolver/OperationPathResolverInterface.php
```diff
@@ -27,9 +27,6 @@
      * @param array       $operation         The operation metadata
      * @param string|bool $operationType     One of the constants defined in ApiPlatform\Core\Api\OperationType
      *                                       If the property is a boolean, true represents OperationType::COLLECTION, false is for OperationType::ITEM
-     * @param string      $operationName     The operation name
```
dunglas: false positive, must be fixed

meyerbaptiste: If you fix it then the problem will be recurrent if we keep the `no_superfluous_phpdoc_tags` rule in our config.

Issue #1985, opened 2018-05-30
Why is a propertyName with suffix [] added to search filter when match type is exact?


Bouwdie:
https://github.com/api-platform/core/blob/e0e639efe4ad0b25f2098b875b6878268bace824/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php#L105

Can anyone explain me why a new property-name with suffix `[]` is added to the list of property descriptions when the match-type is `exact` (or when matching an association? Probably i am missing something very obvious here, but i could not deduct this from the existing documentation :1234:

teohhanhui:
I think the documentation about the support for multiple values was lost somewhere along the way... But basically, the effect is equivalent to the SQL `IN` operator.

teohhanhui:
Or perhaps it was never properly documented? But it's been around for a long time: https://github.com/api-platform/core/pull/500

Bouwdie:
@teohhanhui Thanks for your answer (and sorry for my late reply :yum: ), it does make sense to be able to send in multiple values :+1: . 

The swagger definition now contains the two definitions for the same parameter as well, where they are both the same scalar type. If a client is generated this give a pretty weird method signature (i.m.o). Your example would become getBooks(String author1, String author2); i would expect a signature getBooks(String author, List<string> authors), but if multiple values are allowed why not just use the list getBooks(List<string> authors) since they are filtering the same field it would make sense. 

* Shouldn't the generated swagger type for the multi-value scalar be an array?
* Is it an idea to make the behavior of the generation of the two parameters configurable (like list-only, single-only or both)?

dunglas:
> Shouldn't the generated swagger type for the multi-value scalar be an array?

We must fix our Swagger generator to use `collectionFormat: multi`: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterCollectionFormat

We'll need to document both parameter `foo` and `foo[]` because the brackets notation is PHP-specific. Swagger considers both parameters as different ones.

> Is it an idea to make the behavior of the generation of the two parameters configurable (like list-only, single-only or both)?

Would be nice.

alanpoulain:
Closed in favor of: https://github.com/api-platform/docs/issues/601

PR #2114, opened 2018-07-19
Fix Swagger generation for array parameters

ossinkine:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1985
| License       | MIT

Search filter adds array parameter like `key[]` to swagger documentation but type of this parameter is string. This PR fixes it.


soyuka:
As this is a bug fix, could you change the target branch to 2.3? Thanks!

dunglas:
Good catch @ossinkine! Thanks for fixing this.
Do you think that you can add a new test for the `DocumentationNormalizer` class to prevent any future reflection?

PR #2121, opened 2018-07-23
fix #2115 SwaggerCommand with --yaml now properly serializes nested objects

### LINKED ISSUE #2115: [Swagger][Bug] swagger:export --yaml doesn't work as expected
If you use `--yaml`, the nested resources will not get dumped since they're not array but objects.

If I do:

```
        $content = $input->getOption('yaml')
            ? Yaml::dump(json_decode(json_encode($data), true))
            : json_encode($data, JSON_PRETTY_PRINT);
```

as a quick hack, it works as expected. Is this a bug in `Yaml` or should we workaround this here?

dkarlovi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2115 
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
+++ b/src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
@@ -67,7 +67,7 @@ final class SwaggerCommand extends Command
     {
         $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $this->apiTitle, $this->apiDescription, $this->apiVersion, $this->apiFormats);
         $data = $this->documentationNormalizer->normalize($documentation);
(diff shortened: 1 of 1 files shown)


dkarlovi:
CI failures unrelated.

soyuka:
Do you see a way we could test this? I guess not really but we never know :p.

dkarlovi:
Well sure, we add a sample couple of entities in the test suite and Swaggerize those? :)

soyuka:
would be great!

dunglas:
Thanks @dkarlovi

PR #2141, opened 2018-08-03
Added interface for Listeners

toby-griffiths:
This is to allow for decorating of the final RespondListener, whilst maintaining the expected interface.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Feel free to merge into other version branches, or let me know if you need me to create PRs for them.  I based this on the v2.0 branch so it should merge back into that OK, but I needed it for my 2.1 project.

Any feedback welcome.

### DIFF
--- a/src/EventListener/AddFormatListener.php
+++ b/src/EventListener/AddFormatListener.php
@@ -24,7 +24,7 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class AddFormatListener
+final class AddFormatListener implements AddFormatListenerInterface
 {
     private $negotiator;
     private $formats;
--- a/src/EventListener/AddFormatListenerInterface.php
+++ b/src/EventListener/AddFormatListenerInterface.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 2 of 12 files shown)


toby-griffiths:
These failing tests appear to have been failing before I made my changes.  I was going to ask whether it was possible to merge this in just after the latest 2.1.6 tag, but it appears that the tests are failing in that tagged version as well.

dunglas:
Listeners aren’t designed to be decorated. However, we’re in the process of extracting all their business logic to dedicated services that are not depending of HttpFoundation (for a better Laravel and Expressive compatiblity). Those services can be decorated. Can you explain your use case ? It would probably be a better approach to refactor this way the listener you’re trying to decorate.

dunglas:
Btw, 2.0 and 2.1 aren’t maintained anymore. 2.2 is for security fixes only, 2.3 for regular bug fixes and master (2.4) for new features like this one.

toby-griffiths:
Hi @dunglas,
Thanks for the response.  I've implemented support for change request, rather than direct resource updates.  In order to do this, I use various listeners to modify the request and response based on the user's permissions.  This can all be done without the interfaces, except for the modifying of the response code, as I needed to override the RespondListener, so that I could amend the response code from 200 or 201 to 202.
If there's a better way, I'm open to improving our code.  Please let me know.
I respect the decision to not merge this in, as the version is now only security patches, however I've just tried to check for documentation on the upgrade path, to see what BC there are stopping me from upgrading to the later version of api-platform/core, but I can't see any obvious documentation on this.  Please can you point me in the direction of upgrade documentation if you have a moment?
Also, the release documentation is out of date as well… https://api-platform.com/docs/extra/releases/#the-release-process, as it suggests 2.2 is the current stable version.
Many thanks.

PR #1293, opened 2017-07-28
[WIP] Mongodb support

Zayon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #461 #1248
| License       | MIT
| Doc PR        | none

Filters need a rework because `FilterInterface` use `Doctrine\ORM\QueryBuilder` which is not compatible with mongodb (and other potential providers)

I removed fos_user from behat test for mongo db as didn't manage to understand the problem.

@dunglas @Simperfit @samvdb @ihr-it-projekt

PR #2144, opened 2018-08-05
MongoDB support

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Continuation of #1293.

What is left:
- [ ] Documentation

Can be done after merging;
- [ ] Register `DoctrineMongoDBExtension` in MongoDB ODM
- [ ] Add `DoctrineExtractor` to https://github.com/doctrine/DoctrineMongoDBBundle


src/Bridge/Doctrine/Common/Extension/FilterExtension.php
```diff
@@ -47,7 +46,7 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
     /**
      * {@inheritdoc}
      */
-    public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass = null, string $operationName = null, array $context = [])
+    public function applyToCollection($builder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass = null, string $operationName = null, array $context = [])
```
soyuka: Couldn't we have two interfaces to describe two different kind of filters (one for mongodb, another for sql)?

alanpoulain:
Thank you @soyuka for taking the time to read this PR :slightly_smiling_face:
Sure I think it could be a good opportunity to rework them. What have you (or @teohhanhui) considered for them to be better?
You're right, I think some code in the `AbstractFilter` is not particularly related to filters and can be moved elsewhere. Maybe having it in a service in a `Doctrine\ClassMetadata` namespace would be better? WDYT?
Having two interfaces could be done, but how do you share the code for example in the `apply` method in `AbstractFilter` and `AbstractContextAwareFilter`? And there is the same issue in `FilterExtension` (`applyToCollection`) and its related interfaces.
I agree with you, extending too many classes should be avoided but IMO using traits is even worse (I felt guilty for using them :stuck_out_tongue:).
For the adapter, I bet it's not possible to do it. I presume that `https://github.com/mongodb/mongo-php-library` can't retrieve this information whereas the true PECL legacy extension can https://pecl.php.net/package/mongo. But maybe I'm wrong, we should ask the maintainer.

src/Bridge/Doctrine/Common/Filter/ContextAwareFilterInterface.php
```diff
@@ -26,5 +25,5 @@
     /**
      * Applies the filter.
      */
-    public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null, array $context = []);
+    public function apply($builder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null, array $context = []);
```
ElectricMaxxx: why do we loose typing here?

src/Bridge/Doctrine/MongoDB/Extension/PaginationExtension.php
```diff
@@ -0,0 +1,134 @@
...
+    private $resourceMetadataFactory;
+    private $enabled;
+    private $clientEnabled;
+    private $clientItemsPerPage;
+    private $itemsPerPage;
+    private $pageParameterName;
+    private $enabledParameterName;
+    private $itemsPerPageParameterName;
+
+    public function __construct(ManagerRegistry $managerRegistry, RequestStack $requestStack, ResourceMetadataFactoryInterface $resourceMetadataFactory, bool $enabled = true, bool $clientEnabled = false, bool $clientItemsPerPage = false, int $itemsPerPage = 30, string $pageParameterName = 'page', string $enabledParameterName = 'pagination', string $itemsPerPageParameterName = 'itemsPerPage')
```
ElectricMaxxx: this should go on separate line, later.

soyuka:
> Having two interfaces could be done, but how do you share the code for example in the apply method in AbstractFilter and AbstractContextAwareFilter? And there is the same issue in FilterExtension (applyToCollection) and its related interfaces.

Let's only talk about `AbstractContextAwareFilter` the other one is deprecated. Apply has about 10 lines of code, I guess we can copy/paste it (2 abstract classes, one for ODM the other for ORM). It's only an idea to try to keep the interfaces like this though :p.  

>  Sure I think it could be a good opportunity to rework them. What have you (or @teohhanhui) considered for them to be better?

We haven't considered anything yet filters work well today but their code isn't "good" to me, matter of speak. So apart from just saying "grrr filters need a refactor" I haven't said much haha.

>  For the adapter, I bet it's not possible to do it. I presume that https://github.com/mongodb/mongo-php-library can't retrieve this information whereas the true PECL legacy extension can https://pecl.php.net/package/mongo. But maybe I'm wrong, we should ask the maintainer.

Let's open an issue there, maybe that mongodb-php maintainers can help us?

src/Bridge/Doctrine/Common/Filter/ContextAwareFilterInterface.php
soyuka: He explained it's because the QueryBuilder for the ODM isn't the same but he tried to reuse the code.

ElectricMaxxx: Indeed, i understand. it is a good sign that this interface is missing in doctrine common, when we have an equal class in each backend implementation. Just for the naming :-)

phpunit.mongo.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.6/phpunit.xsd"
```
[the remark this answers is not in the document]
alanpoulain: It should be modified I think but maybe not in this PR since this file is copied from the other one.

src/Bridge/Doctrine/MongoDB/Extension/PaginationExtension.php
alanpoulain: This style is not really used in the other parts of API Platform though.

alanpoulain:
So it's a choice between:
- keeping the old interfaces and duplicate a little of code
- introducing new interfaces to share the code
The duplicated code is in the `apply` method in `AbstractContextAwareFilter` and in the `applyToCollection` in `FilterExtension`. Which one do you prefer, others?

Maybe having separated code for shared parts and a service for the class metadata part will help a future refactoring then. WDYT of having this class metadata service @soyuka?

Done here for the issue: https://github.com/alcaeus/mongo-php-adapter/issues/239

src/Bridge/Doctrine/Common/Filter/AbstractBooleanFilter.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    protected function normalizeValue($value)
+    {
+        if (\in_array($value, [true, 'true', '1'], true)) {
+            $value = true;
+        } elseif (\in_array($value, [false, 'false', '0'], true)) {
+            $value = false;
+        } else {
+            $this->logger->notice('Invalid filter ignored', [
+                'exception' => new InvalidArgumentException(sprintf('Invalid boolean value for "%s" property, expected one of ( "%s" )', $property, implode('" | "', [
```
ElectricMaxxx: `$property` does not exist.

src/Bridge/Doctrine/Common/Filter/AbstractFilter.php
```diff
@@ -11,16 +11,13 @@
```
dunglas: As this class isn't marked as internal, we should deprecate the old one. The old class can extend the new one.

src/Bridge/Doctrine/Common/Filter/AbstractSearchFilter.php
```diff
@@ -0,0 +1,233 @@
...
+use ApiPlatform\Core\Api\IriConverterInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\DBAL\Types\Type;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+
+/**
```
dunglas: Maybe should we make these new abstract classes internal to make the maintenance easier?

src/Bridge/Doctrine/Common/Filter/AbstractSearchFilter.php
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * @var string Finds fields that are starting with the word
+     */
+    const STRATEGY_WORD_START = 'word_start';
+
+    protected $iriConverter;
+    protected $propertyAccessor;
+
+    /**
+     * @param RequestStack|null $requestStack No prefix to prevent autowiring of this deprecated property
```
dunglas: Injecting `$requestStack` is deprecated now, so if we create new classes, we shouldn't add it to the condtructor signature.

src/Bridge/Doctrine/Common/Filter/ContextAwareFilterInterface.php
dunglas: I suggest to revert this change and introduce a new interface, because QueryBuilder implementations aren't compatibles.

src/Bridge/Doctrine/Common/Filter/FilterInterface.php
```diff
@@ -27,5 +26,5 @@
     /**
      * Applies the filter.
      */
-    public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null);
+    public function apply($builder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null);
```
dunglas: same here

src/Bridge/Doctrine/MongoDB/CollectionDataProvider.php
```diff
@@ -0,0 +1,75 @@
...
+use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
+use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
+use ApiPlatform\Core\Exception\RuntimeException;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\ODM\MongoDB\Aggregation\Builder;
+
+/**
+ * Collection data provider for the Doctrine MongoDB ODM.
+ */
+class CollectionDataProvider implements CollectionDataProviderInterface, RestrictedDataProviderInterface
```
dunglas: final (same for all new classes)

src/Bridge/Doctrine/MongoDB/CollectionDataProvider.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    private $managerRegistry;
+    private $collectionExtensions;
+
+    /**
+     * @param QueryCollectionExtensionInterface[]|ContextAwareQueryCollectionExtensionInterface[] $collectionExtensions
+     */
+    public function __construct(ManagerRegistry $managerRegistry, /* iterable */ $collectionExtensions = [])
```
dunglas: you can uncomment iterable right?

src/Bridge/Doctrine/MongoDB/Extension/PaginationExtension.php
```diff
@@ -0,0 +1,134 @@
...
+        $this->enabledParameterName = $enabledParameterName;
+        $this->itemsPerPageParameterName = $itemsPerPageParameterName;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function applyToCollection(Builder $aggregationBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
+    {
+        $request = $this->requestStack->getCurrentRequest();
```
dunglas: Using the request in the data provider subsystem is deprecated. We should only support using the context for new classes.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -108,6 +110,10 @@ public function load(array $configs, ContainerBuilder $container)
             ->addTag('api_platform.doctrine.orm.query_extension.item');
         $container->registerForAutoconfiguration(QueryCollectionExtensionInterface::class)
             ->addTag('api_platform.doctrine.orm.query_extension.collection');
+        $container->registerForAutoconfiguration(MongoDbQueryItemExtensionInterface::class)
```
ElectricMaxxx: those one should depend on persistence layer decission.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -233,7 +239,12 @@ private function getBundlesResourcesPaths(ContainerBuilder $container): array
             foreach (['.yaml', '.yml', '.xml', ''] as $extension) {
                 $paths[] = "$dirname/Resources/config/api_resources$extension";
             }
-            $paths[] = "$dirname/Entity";
+            if ($config['doctrine']['enable_orm']) {
```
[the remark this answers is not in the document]
alanpoulain: I don't understand what you mean.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
alanpoulain: Why?

src/Bridge/Doctrine/MongoDB/Extension/PaginationExtension.php
alanpoulain: I think there is a lot of work to be done in the pagination extension then.
I don't think it's a good idea to do that in this PR.

alanpoulain:
New working filter: search filter.

I have revamped the architecture to keep two interfaces.
There are also less BC.
The changes (https://github.com/api-platform/core/compare/29e088f951b214e3b8168af4c54f3956656b30e6...api-platform:23c8ff967540f22145a12467f9b7e8a56e33e574) are now:
- New service `PropertyHelper`. Methods extracted from `AbstractFilter`.
- Shared code for filters moved to traits in `Common`.
- Since traits cannot contain constants, I have introduced a `SearchFilterInterface` in `Common`.
- `FilterInterface` in `Common` with the method `getPropertyHelperServiceName` (explained after).
- Interfaces using the `AggregationBuilder` in `MongoDB`.
- `MongoDB` has only `AbstractContextAwareFilter`.

The new service `property_helper` allows us to have just a little bit of duplicated code in the `AbstractFilter` classes.
Not injecting this new service has been deprecated for API Platform 2.4, and injecting the manager registry too. It's also the case for using the methods in the `Orm\AbstractFilter` directly.

There was a main issue: when using the annotation `@ApiFilter`, the filter service is created by a compiler pass (`AnnotationFilterPass`). But since there are two `property_helper` services (one for ORM, the other for MongoDB ODM), how does it know which service to inject?
That's why I had to introduce the method `getPropertyHelperServiceName` in the `FilterInterface`.
You can see how the code is used here:
https://github.com/api-platform/core/blob/23c8ff967540f22145a12467f9b7e8a56e33e574/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php#L64-L66
If you have a better way to do this, please tell me!

tests/Fixtures/TestBundle/Filter/ArrayRequiredFilter.php
```diff
@@ -19,6 +19,11 @@
 
 final class ArrayRequiredFilter extends AbstractFilter
 {
+    public static function getPropertyHelperServiceName(): string
+    {
+        return '';
```
dunglas: it's not intended right?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -60,6 +61,9 @@ private function createFilterDefinitions(\ReflectionClass $reflectionClass, Read
             $definition->setClass($filterClass);
             $definition->addTag(self::TAG_FILTER_NAME);
             $definition->setAutowired(true);
+            if ((new \ReflectionClass($filterClass))->hasMethod('getPropertyHelperServiceName')) {
+                $definition->setArgument('$propertyHelper', new Reference($filterClass::getPropertyHelperServiceName()));
```
dunglas: Why not something like this?

```php
$definition->setArgument('$propertyHelper', is_a($filterClass, \ApiPlatform\Core\Bridge\Doctrine\MongoDB\Filter::class, true) ? 'the_mongodb_propertyHelper' : 'the_orm_propertyHelper');
```

It's a cleaner design (the classes aren't aware of the DIC, and not tight to Symfony), and it's still extensible (an app or a bundle could use the same trick in its own compiler pass for another provider). 

But here, I think transforming the PropertyHelper class in a trait would be less intrusive.

src/Bridge/Doctrine/ClassMetadata/PropertyHelper.php
```diff
@@ -0,0 +1,154 @@
...
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * Helper for getting information regarding a property using the resource metadata.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ * @author Théo FIDRY <theo.fidry@gmail.com>
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class PropertyHelper
```
dunglas: As it's a final class, we must also introduce an interface.
Also, can you move it in the existing `Common` namespace, it doesn't worth introducing a new namespace for a single class.

dunglas: Another solution, that solves both the DI and the interface problem, is to transform this class in a trait.

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
```diff
@@ -0,0 +1,95 @@
...
+            ];
+        }
+
+        return $description;
+    }
+
+    /**
+     * Determines whether the given property refers to a boolean field.
+     */
+    protected function isBooleanField(string $property, string $resourceClass): bool
```
dunglas: Should be private isn't it (same for all similar methods)?

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
```diff
@@ -0,0 +1,95 @@
...
+ * Filters collection on equality of boolean properties. The value is specified
+ * as one of ( "true" | "false" | "1" | "0" ) in the query.
+ *
+ * For each property passed, if the resource does not have such property or if
+ * the value is not one of ( "true" | "false" | "1" | "0" ) the property is ignored.
+ *
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ * @author Teoh Han Hui <teohhanhui@gmail.com>
+ */
+trait BooleanFilterTrait
```
dunglas: Why not an internal abstract class? It would prevent to introduce a bunch of new interfaces.

tests/Fixtures/TestBundle/Filter/ArrayRequiredFilter.php
alanpoulain: It is :sweat_smile: Since this filter is extending the `Orm` filter but is not really an ORM one.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
alanpoulain: Sure! I don't like the idea of transforming the service into a trait though (my POV explained in the next comment).

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
alanpoulain: It is not possible. The code is shared between the `Orm` and the `Odm` boolean filter. They need to extend their own `AbstractContextAwareFilter`.

alanpoulain: Also, which interfaces are you talking about?

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
alanpoulain: It can, but since the filters are not final, I thought it was for a reason and the methods could be reused. Is there a reason for them to be not final?

src/Bridge/Doctrine/ClassMetadata/PropertyHelper.php
alanpoulain: OK for the interface and for moving it.
I don't really like the idea of having a trait though. I think this service makes sense. It exposes methods to know information about a property using the Doctrine class metadata.

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
dunglas: It was a mistake, it as been forgotten. But it was a while ago, and now it's too late (or we can mark them with `@final` now and `final` in 3.0.). But MongoDB filters, and all new filters, should be `final`.

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
dunglas: What about something like the following?

```php
/** @internal */
abstract class AbstractBooleanFilter
{
    // ...
}

class OrmBooleanFilter extends AbstractBooleanFilter implements OrmAbstractContextAwareFilter
{
    // ...
}

final class MongoBooleanFilter extends AbstractBooleanFilter implements MongoAbstractContextAwareFilter
{
    // ...
}
```

Doing so, interfaces like `SearchFilterInterface` can be removed.

src/Bridge/Doctrine/ClassMetadata/PropertyHelper.php
dunglas: Its something purely internal tight to Doctrine. Adding this service make our public API broader (so harder to maintain), and increase the overall complexity for no real benefit (ex: in the compiler pass).
It also makes it harder to use API Platform without a framework, or other frameworks than SF (more class to initialize).

To me a trait is enough, and will make the PR smaller. WDYT @api-platform/core-team?

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
alanpoulain: `OrmAbstractContextAwareFilter` and `MongoAbstractContextAwareFilter` are abstract classes not interfaces. You seem to suggest the first way I tried before but which needed to use only one interface for filters.

src/Bridge/Doctrine/ClassMetadata/PropertyHelper.php
soyuka: I'm :+1: for the trait

src/Bridge/Doctrine/Common/Filter/BooleanFilterTrait.php
dunglas: I'm not sure to understand. You may have everything in the abstract class and keep type hints:

```php
/** @internal */
abstract class AbstractBooleanFilter
{
    protected function doSomething($noTypeHint) {
        // contains the actual logic, shared
    }
}

final class MongoBooleanFilter extends AbstractBooleanFilter implements MongoAbstractContextAwareFilter
{
    public function something(TypeHint $typeHinted) {
        // preserves the typehint, delegate to the shared method
        return $this->doSomething($typeHinted);
    }
}
```

WDYT?

alanpoulain:
And all the Behat tests are now working with MongoDB! :tada:
Now begins the hard work of fixing the unit tests and adding some for the MongoDB part :cold_sweat:

dunglas:
Congrats! Can't wait to merge this one.

Dragonqos:
I am testing your branch for a few days and found one mystical bug with pagination:

Commit: 5a203ab
MongoDB version: v4.0.3

**Expected:**
When I send request with filters and expected that 0 results should be returned

**Given:**
Exception "Undefined offset: 0" was thrown
```
[2018-10-30 06:56:16] request.CRITICAL: Uncaught PHP Exception ErrorException: "Notice: Undefined offset: 0" at /var/www/ms-mapping/vendor/api-platform/core/src/Bridge/Doctrine/MongoDbOdm/Paginator.php line 67 {"exception":"[object] (ErrorException(code: 0): Notice: Undefined offset: 0 at /app/vendor/api-platform/core/src/Bridge/Doctrine/MongoDbOdm/Paginator.php:67)"} []
```

core/src/Bridge/Doctrine/MongoDbOdm/Paginator.php
line 67:         $this->totalItems = $cursor->toArray()[0]['count'][0]['count'];

Cursor may return empty count and result sets
```
[
          0 => [
                  [results] => [],
                  [count] => []
              ]
]
```

**How to fix:** 
Probably it would be better to return default zero result when offset not found?
$this->totalItems = $cursor->toArray()[0]['count'][0]['count'] ?? 0;

P.s Great job!

alanpoulain:
@Dragonqos Could you tell me which request you sent to have this error?
Because the `PaginationExtension` should always apply the count facet.

alanpoulain:
@Dragonqos There's already a test with 0 element returned here:
https://github.com/api-platform/core/blob/750e9bea5941733bb0d393aee4e7b2e58e4ee656/features/doctrine/search_filter.feature#L349-L375
I don't see why it doesn't work in your case. Could you dump the cursor in the `Paginator`'s constructor?

Dragonqos:
Same as i Posted in first message.
screen shot 2018-10-30 at 9 54 35 pm
screen shot 2018-10-30 at 9 58 38 pm

alcaeus:
Sorry for the late reply, I only noticed this PR a few days ago.

> But it was not that simple 😞 The main issue of using the `AggregationBuilder` was for pagination. Because an execution of an aggregation doesn't return the total items when you limit it...

This is not a limitation of the aggregation builder but rather the cursor interface in the new MongoDB driver. In `ext-mongodb`, a cursor exposed a `count` method, making you think it's able to return an unfiltered/filtered count of the elements in the cursor. It didn't however: it always ran a `count` command on the cursor, applying the filter and optionally limit/skip from the original query on the command. That's the reason why it was removed in the new driver version: you think it gave you the exact count when in fact it re-ran a query, thus being able to produce wrong results. `mongo-php-adapter` emulates this behaviour, but it's a stop-gap solution to run ODM on PHP 7 with the new driver. With ODM 2.0 nearing, prepare to drop it and no longer be able to count cursor results. Your facet workaround is beautiful because it preserves the original count at the time of the original query execution. Great job on that! The only downside is that you load all data into memory directly instead of fetching it from the cursor in batches. This can be an issue when having large documents and/or large amount of items on a single page.

> Using facets is bringing one more issue: you can't hydrate the query directly. And I had to do it manually in the `Paginator`:

Yes, that's an unfortunate side effect. I'm hoping to create a more dynamic way of mapping aggregation results without needing to map actual classes for libraries like this, similar to the `ResultSetMapping` in ORM. It won't be in 2.0, but I've created https://github.com/doctrine/mongodb-odm/issues/1900 to track this for future versions.

> * The user will have to store the references using an ID or a ref and **not** a DBRef.

I've been thinking of making `ref` references the default in ODM 2.0 because of the DBRef limitations when using aggregation pipeline queries. This might just be the main reason to do so.

> * Nested properties are retrieved for only one level of depth [...]. Maybe it's possible to fix it by using the `$graphLookup` stage

`$graphLookup` works recursively within a collection, not across multiple collections. If you want to work with multiple collections, you have to work with multiple `$lookup` and `$addFields` stages. `$addFields` should be used when you have a `ReferenceOne` mapping because `$lookup` always returns an array, never a single document. In that case, you can use the following pipeline stage:
```
->addFields()
    ->field('users')
    ->arrayElemAt('$users', 0)
```

From there, you can just keep adding `$lookup` stages until you have everything you need. Of course, you might want to use a `$project` stage at the very end of the pipeline to limit the result to the original data and avoid excessive memory usage.

A quick word on ODM versions: you're building this on top of ODM 1.0, which is going to cause a few problems soon. I'm going to release Beta-1 of ODM 2.0 in a couple of weeks, which will finalise its API with no further BC breaks coming. From what I can see in your PR, you shouldn't really be affected by any of its BC breaks, except for the driver swap (more on this in a separate comment). It might be worth testing this against the 2.0 version as well; not only would that improve compatibility for users but it would also help me make sure we didn't miss anything in 2.0.

alanpoulain:
Thank you for your very useful comments @alcaeus!
> The only downside is that you load all data into memory directly instead of fetching it from the cursor in batches. This can be an issue when having large documents and/or large amount of items on a single page.

Indeed, we should document this issue. I've added it in the TODO list.

> It won't be in 2.0, but I've created doctrine/mongodb-odm#1900 to track this for future versions.

Thanks, having this helper will allow us to avoid to use "tricks".

> I've been thinking of making ref references the default in ODM 2.0 because of the DBRef limitations when using aggregation pipeline queries.

Good idea! I'm a little worried about the future issues we will have if the users don't follow the documentation correctly about the references :sweat_smile:

> From there, you can just keep adding `$lookup` stages until you have everything you need. Of course, you might want to use a `$project` stage at the very end of the pipeline to limit the result to the original data and avoid excessive memory usage.

I had planned to tackle this issue later on (but I forgot to add it in the TODO list), having your explanation will provide me a simpler starting point, thanks!

> It might be worth testing this against the 2.0 version as well

Sure. Don't you think dropping the 1.0 compatibility entirely is better?

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -0,0 +1,209 @@
...
+     * @return \MongoRegex|string
+     */
+    private function addEqualityMatchStrategy(string $strategy, $value, bool $caseSensitive)
+    {
+        $addCaseFlag = $this->addCaseFlag($caseSensitive);
+
+        switch ($strategy) {
+            case null:
+            case self::STRATEGY_EXACT:
+                return $caseSensitive ? $value : new \MongoRegex($addCaseFlag("/^$value$/"));
```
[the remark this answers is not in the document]
alanpoulain: Will do.

alcaeus:
> Sure. Don't you think dropping the 1.0 compatibility entirely is better?

Yes. Heads up: I don’t have a timeline for 2.0, but I’m hoping that we can get through beta quickly.

dunglas:
> Yes. Heads up: I don’t have a timeline for 2.0, but I’m hoping that we can get through beta quickly.

Then would prefer to support just v2 from the beginning, it will reduce the maintenance burden.  As long as our code depending of MongoDB ODM is marked `@experimental`, it's not an issue than ODM is in beta.

alcaeus:
That sounds good to me. I’ll give you a ping once we’ve got beta-1, but feel free to test against dev-master in the meantime.

src/Bridge/Doctrine/MongoDbOdm/PropertyHelperTrait.php
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        $alias = $association;
+        if ($this->getClassMetadata($resourceClass)->hasReference($association)) {
+            $alias = "${association}_lkup";
+            $aggregationBuilder->lookup($association)->alias($alias);
+        }
+
+        // assocation.property => association_lkup.property
+        return [str_replace($association, $alias, $property), $propertyParts['field'], $propertyParts['associations']];
```
vincentchalamon:
```suggestion
        return [str_replace("$association.", "$alias.", $property), $propertyParts['field'], $propertyParts['associations']];
```
This changes handle the following case: `street.streetAddress` => `street_lkup.streetAddress`. Otherwise, it will change the association/field to `street_lkup.street_lkupAddress`

alanpoulain:
@strobox Don't worry I will fix it later, and you'll just have to require `doctrine/mongodb-odm` in order to enable the MongoDB integration.

alanpoulain:
All unit tests have been written! :tada:

I have upgraded MongoDB ODM to 2.0 but I have some issues with it. cc @alcaeus 

When the ODM uses the evaluating strategy for the proxies, it seems there is an issue with the cached annotation reader because it needs a true file (it uses `filemtime`):
https://github.com/doctrine/annotations/blob/d393f679ae572f4735531882b62c1bc60bf5d33c/lib/Doctrine/Annotations/CachedReader.php#L213-L230

The value of `$filename` is: `/project/vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php(54) : eval()'d code`

It works the first time, and it doesn't work anymore after.

alanpoulain:
@meyerbaptiste The `Common` traits are now more usable: I have added abstract methods and they use the `PropertyHelperTrait`.

alanpoulain:
This PR is finally RFR :tada: cc @api-platform/core-team

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -78,125 +77,35 @@ public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $q
     abstract protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null/*, array $context = []*/);
 
     /**
-     * Gets class metadata for the given resource.
-     */
-    protected function getClassMetadata(string $resourceClass): ClassMetadata
```
raoulclais: There seem to be a BC break here, because this method is missing from the OrmPropertyHelperTrait

alanpoulain: Fixed.

alanpoulain:
Rebased against master using the new Pagination configuration!

src/Bridge/Doctrine/Common/Util/IdentifierManagerTrait.php
```diff
@@ -39,6 +41,7 @@ private function normalizeIdentifiers($id, ObjectManager $manager, string $resou
         $doctrineClassMetadata = $manager->getClassMetadata($resourceClass);
         $doctrineIdentifierFields = $doctrineClassMetadata->getIdentifier();
         $isOrm = interface_exists(EntityManagerInterface::class) && $manager instanceof EntityManagerInterface;
+        $isOdm = class_exists(DocumentManager::class) && $manager instanceof DocumentManager;
```
alcaeus: Why the call to `class_exists`? If it doesn't exist, the subsequent check will return `false` anyways.

src/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function __construct(ManagerRegistry $managerRegistry, iterable $collectionExtensions = [])
+    {
+        $this->managerRegistry = $managerRegistry;
+        $this->collectionExtensions = $collectionExtensions;
+    }
+
+    public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
+    {
+        return null !== $this->managerRegistry->getManagerForClass($resourceClass);
```
alcaeus: Since you're dealing with a manager registry from `doctrine/persistence`, you'll want to check whether the manager returned for `$resourceClass` is a document manager.

src/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
```diff
@@ -0,0 +1,78 @@
...
+        $aggregationBuilder = $repository->createAggregationBuilder();
+        foreach ($this->collectionExtensions as $extension) {
+            $extension->applyToCollection($aggregationBuilder, $resourceClass, $operationName, $context);
+
+            if ($extension instanceof AggregationResultCollectionExtensionInterface && $extension->supportsResult($resourceClass, $operationName, $context)) {
+                return $extension->getResult($aggregationBuilder, $resourceClass, $operationName, $context);
+            }
+        }
+
+        $aggregationBuilder->match();
```
alcaeus: Why the empty match? If none of the extensions match, you might as well run aggregate with an empty pipeline.

src/Bridge/Doctrine/MongoDbOdm/Extension/FilterExtension.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function applyToCollection(Builder $aggregationBuilder, string $resourceClass = null, string $operationName = null, array &$context = [])
+    {
+        if (null === $resourceClass) {
+            throw new InvalidArgumentException('The "$resourceClass" parameter must not be null');
```
alcaeus: The `$resourceClass` parameter is non-nullable in the interface. Keeping with that definition makes this exception obsolete.

src/Bridge/Doctrine/MongoDbOdm/Extension/OrderExtension.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->managerRegistry = $managerRegistry;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function applyToCollection(Builder $aggregationBuilder, string $resourceClass = null, string $operationName = null, array &$context = [])
+    {
+        if (null === $resourceClass) {
+            throw new InvalidArgumentException('The "$resourceClass" parameter must not be null');
```
alcaeus: Exception is obsolete as per interface.

src/Bridge/Doctrine/MongoDbOdm/Filter/DateFilter.php
```diff
@@ -0,0 +1,148 @@
...
+            (self::INCLUDE_NULL_BEFORE === $nullManagement && \in_array($operator, [self::PARAMETER_BEFORE, self::PARAMETER_STRICTLY_BEFORE], true)) ||
+            (self::INCLUDE_NULL_AFTER === $nullManagement && \in_array($operator, [self::PARAMETER_AFTER, self::PARAMETER_STRICTLY_AFTER], true)) ||
+            (self::INCLUDE_NULL_BEFORE_AND_AFTER === $nullManagement && \in_array($operator, [self::PARAMETER_AFTER, self::PARAMETER_STRICTLY_AFTER, self::PARAMETER_BEFORE, self::PARAMETER_STRICTLY_BEFORE], true))
+        ) {
+            $aggregationBuilder->match()->addOr(
+                $aggregationBuilder->matchExpr()->field($field)->operator($operatorValue[$operator], $value),
+                $aggregationBuilder->matchExpr()->field($field)->equals(null)
+            );
+        } else {
+            $aggregationBuilder->match()->addAnd(
```
alcaeus: Multiple suggestions here:
* Fields are only evaluated if they are non-null. If a field is null, any of the comparison operators here won't match. The second expression is useless here.
* By default multiple conditions are always evaluated using `$and`, so you don't need to use `addAnd` in situations like these
* If you want to use multiple operators on the same field, call `field` once and chain the operator calls

With that in mind, this whole thing can be simplified to
```
$aggregationBuilder->match()->field($field)->operator($operatorValue[$operator], $value);
```

src/Bridge/Doctrine/MongoDbOdm/Filter/ExistsFilter.php
```diff
@@ -0,0 +1,84 @@
...
+     * {@inheritdoc}
+     */
+    protected function isNullableField(string $property, string $resourceClass): bool
+    {
+        $propertyParts = $this->splitPropertyParts($property, $resourceClass);
+        $metadata = $this->getNestedMetadata($resourceClass, $propertyParts['associations']);
+
+        $field = $propertyParts['field'];
+
+        if ($metadata instanceof ClassMetadata && $metadata->hasField($field)) {
```
alcaeus: Style nitpick: `if` condition can be inlined into the `return` below.

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -0,0 +1,207 @@
...
+
+        if (!$this->hasValidValues($values, $this->getDoctrineFieldType($property, $resourceClass))) {
+            $this->logger->notice('Invalid filter ignored', [
+                'exception' => new InvalidArgumentException(sprintf('Values for field "%s" are not valid according to the doctrine type.', $property)),
+            ]);
+
+            return;
+        }
+
+        if (1 === \count($values)) {
```
alcaeus: Let your database optimise queries for you instead of duplicating logic:
```
> db.nullTest.find({field: { $in: [1] }}).explain()
{
	"queryPlanner" : {
		"plannerVersion" : 1,
		"namespace" : "test.nullTest",
		"indexFilterSet" : false,
		"parsedQuery" : {
			"field" : {
				"$eq" : 1
			}
		},
...
```

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -0,0 +1,207 @@
...
+    private function addEqualityMatchStrategy(string $strategy, string $field, $value, bool $caseSensitive, ClassMetadata $metadata)
+    {
+        $type = $metadata->getTypeOfField($field);
+
+        switch ($strategy) {
+            case MongoDbType::STRING !== $type:
+                return MongoDbType::getType($type)->convertToDatabaseValue($value);
+            case null:
+            case self::STRATEGY_EXACT:
+                return $caseSensitive ? $value : new Regex("^$value$", $caseSensitive ? '' : 'i');
```
alcaeus: The second ternary within the `Regex` constructor is useless since it already must have evaluated to `false` in order to hit the constructor.

src/Bridge/Doctrine/MongoDbOdm/ItemDataProvider.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $this->managerRegistry = $managerRegistry;
+        $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
+        $this->propertyMetadataFactory = $propertyMetadataFactory;
+        $this->itemExtensions = $itemExtensions;
+    }
+
+    public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
+    {
+        return null !== $this->managerRegistry->getManagerForClass($resourceClass);
```
alcaeus: Same as before, you should check if the manager is a document manager.

src/Bridge/Doctrine/MongoDbOdm/ItemDataProvider.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+        $id = (array) $id;
+
+        $fetchData = $context['fetch_data'] ?? true;
+        if (!$fetchData && $manager instanceof DocumentManager) {
+            return $manager->getReference($resourceClass, reset($id));
+        }
+
+        $repository = $manager->getRepository($resourceClass);
+        if (!method_exists($repository, 'createAggregationBuilder')) {
```
alcaeus: Seeing this check multiple times makes me think that we need to enforce a few more types when it comes to repositories in ODM, what do you think?

src/Bridge/Doctrine/MongoDbOdm/Paginator.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        $this->mongoDbOdmIterator = $mongoDbOdmIterator;
+        $this->unitOfWork = $unitOfWork;
+        $this->resourceClass = $resourceClass;
+        $this->pipeline = $pipeline;
+
+        $resultsFacetInfo = $this->getFacetInfo('results');
+        $this->getFacetInfo('count');
+
+        // See https://github.com/alcaeus/mongo-php-adapter#mongocommandcursor
```
alcaeus: I'd remove the reference to `mongo-php-adapter` since it's not used at all here. Suggested wording:

> Since the \MongoDB\Driver\Cursor class does not expose information about skip/limit parameters of the query, the values set in the facet stage are used instead.

src/Bridge/Doctrine/MongoDbOdm/Paginator.php
```diff
@@ -0,0 +1,163 @@
...
+            return $this->unitOfWork->getOrCreateDocument($this->resourceClass, $result);
+        }, $this->mongoDbOdmIterator->toArray()[0]['results']));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function count(): int
+    {
+        return iterator_count($this->getIterator());
```
alcaeus: This will load and hydrate all documents in order to run a `count`. You might want to avoid that and count raw data instead:
```
return count($this->mongoDbOdmIterator->toArray()[0]['results']);
```

src/Bridge/Doctrine/MongoDbOdm/PropertyHelperTrait.php
```diff
@@ -0,0 +1,91 @@
...
+
+            if ($classMetadata->hasReference($association)) {
+                $propertyAlias = "${association}_lkup";
+                // previous_association_lkup.association
+                $localField = "$alias$association";
+                // previous_association_lkup.association_lkup
+                $alias .= $propertyAlias;
+                $referenceMapping = $classMetadata->getFieldMapping($association);
+
+                $aggregationBuilder->lookup($classMetadata->getAssociationTargetClass($association))
```
alcaeus: ODM does this for you, throwing a helpful exception if you're using a reference that can't be looked up:
```
$aggregationBuilder
    ->lookup($association)
        ->alias($alias)
```

Your code does not handle different reference storage strategies (`id`, `ref`) appropriately.

src/Bridge/Doctrine/MongoDbOdm/PropertyInfo/DoctrineExtractor.php
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getProperties($class, array $context = [])
+    {
+        try {
+            $metadata = $this->objectManager->getClassMetadata($class);
+        } catch (MappingException $exception) {
+            return null;
+        } catch (MongoDbOdmMappingException $exception) {
```
alcaeus: This exception extends the base mapping exception, so the second `catch` will never be hit.

src/Util/ClassInfoTrait.php
```diff
@@ -33,6 +31,31 @@
...
+        if (false !== $positionCg) {
+            return substr($className, $positionCg + 8);
+        }
+
+        $className = ltrim($className, '\\');
+
+        return substr(
+            $className,
+            8 + $positionPm,
+            strrpos($className, '\\') - ($positionPm + 8)
```
alcaeus: Maybe @ocramius might want to chime in here, because I'm not sure if doing this for Proxy Manager is kosher. In MongoDB ODM, we wrote a resolver that internally hits the proxy manager logic to resolve the correct parent class. See https://github.com/doctrine/mongodb-odm/blob/181edb9226d783c319bc788632ec9ce0127395c0/lib/Doctrine/ODM/MongoDB/Proxy/ClassNameResolver.php#L11 for details.

src/Bridge/Doctrine/MongoDbOdm/Filter/DateFilter.php
alanpoulain: I think I've done it like this because chaining a field with an operator in a match stage is not possible. You need to use `matchExpr`. But if I remove the line `$aggregationBuilder->matchExpr()->field($field)->notEqual(null)`, the else can be merged into the first `if`. We need Behat tests anyway for this part.

src/Bridge/Doctrine/MongoDbOdm/PropertyHelperTrait.php
alanpoulain: I had to do like this to manage to have nested lookups. It's because there is a garding `if` in the `Lookup` stage code using `$this->class` (https://github.com/doctrine/mongodb-odm/blob/55e1b9d87f6d60042878c368262d072931737b53/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php#L154-L156).
I don't understand why you say it does not handle the different strategies. It's nearly the same code as in the `Lookup` class.

src/Util/ClassInfoTrait.php
alanpoulain: The code is duplicated from the Proxy manager code. I needed it to be a trait and not a service because it has too much impacts.

dunglas:
🎆thanks @alanpoulain!

src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -34,25 +36,9 @@
...
-        self::NULLS_LARGEST => [
-            'ASC' => 'DESC',
-            'DESC' => 'ASC',
-        ],
-    ];
-
-    /**
-     * @var string Keyword used to retrieve the value
-     */
-    protected $orderParameterName;
```
bastnic: this protected value is now private.

alanpoulain: Forgot to change it, yes. Did it cause an issue?

bastnic: BC break for my app, we extended `OrderFilter` to use the work on properties and the inheritance of `orderParameterName` of the constructor as it's not a final nor an internal class I think it maybe a BC break (and for no additional value?)

alanpoulain: In fact, the `OrderFilter` _should_ be final but is not (the MongoDB one is final).
But you're right this is a BC break, could you make a PR to fix it?

teohhanhui: Sorry, I don't understand why the`OrderFilter` should be final. There are many cases where I've needed to extend it.

bastnic: I've not so much time to make PR right now, feel free @teohhanhui (hey again) or @alanpoulain.

teohhanhui: https://github.com/api-platform/core/pull/2678

PR #2145, opened 2018-08-06
OrderFilter with nullable relation uses LEFT JOIN fix #2109

### LINKED ISSUE #2109: OrderFilter on nested entity seems KO
Hi guys,

An entity _User_ has a relation with another entity _Conversation_.
If I use the **OrderBy** filter on _conversation.createdAt_ the User without a _Conversation_ will be excluded from the results.

```php 
@ApiFilter(OrderFilter::class, properties={"conversation.createdAt": "ASC" }})
```
Searching through github I've found : 

- [This first issue](https://github.com/api-platform/core/issues/363) with [the associated fix](https://github.com/api-platform/core/pull/367/files)

- The ability to use null comparison with [this PR](https://github.com/api-platform/core/pull/838) 

So I transform my filter to : 
```php 
@ApiFilter(OrderFilter::class, properties={"conversation.createdAt": { "nulls_comparison": OrderFilter::NULLS_SMALLEST, "default_direction": "ASC" }})
```

Full of trust :+1: , I press Send button on postman and... still excluded :disappointed: 

Digging a bit I see that the [OrderFilter](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php#L149) call a specific method for join relation : 
```php 
list($alias, $field) = $this->addJoinsForNestedProperty($property, $alias, $queryBuilder, $queryNameGenerator, $resourceClass);
```

Then on the [AbstractFilter](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php#L325) : 
```php 
$alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $parentAlias, $association);
```

And finally in the [QueryBuilderHelper](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Util/QueryBuilderHelper.php#L44) : 
```php 
if (Join::LEFT_JOIN === $joinType || QueryChecker::hasLeftJoin($queryBuilder)) {
    $queryBuilder->leftJoin($query, $associationAlias, $conditionType, $condition);
} else {
    $queryBuilder->innerJoin($query, $associationAlias, $conditionType, $condition);
}
```
There is no **$joinType** send and the **QueryCheck::hasLeftJoin** returns false, so an _innerJoin_ is made instead of a _leftJoin_.  
If I forced a _leftJoin_, the results are ok.

Am I missing something or is there a problem?

Thanks,
Jérémy.

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2109 
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\Bridge\Doctrine\Orm\Extension;
 use ApiPlatform\Core\Api\FilterCollection;
 use ApiPlatform\Core\Api\FilterLocatorTrait;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\FilterInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\OrderFilter;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
@@ -60,12 +61,25 @@ final class FilterExtension implements ContextAwareQueryCollectionExtensionInter
             return;
         }
 
+        $orderFilter = null;
+
         foreach ($resourceFilters as $filterId) {
             $filter = $this->getFilter($filterId);
             if ($filter instanceof FilterInterface) {
+                // Apply the OrderFilter after every other filter to avoid an edge case where OrderFilter would do a LEFT JOIN instead of an INNER JOIN
+                if ($filter instanceof OrderFilter) {
+                    $orderFilter = $filter;
+                    continue;
+                }
+
                 $context['filters'] = $context['filters'] ?? [];
                 $filter->apply($queryBuilder, $queryNameGenerator, $resourceClass, $operationName, $context);
             }
+        }
+
+        if (null !== $orderFilter) {
+            $context['filters'] = $context['filters'] ?? [];
+            $orderFilter->apply($queryBuilder, $queryNameGenerator, $resourceClass, $operationName, $context);
         }
     }
 }
--- a/src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
@@ -269,7 +269,7 @@ abstract class AbstractFilter implements FilterInterface
      *               the second element is the $field name
      *               the third element is the $associations array
      */
-    protected function addJoinsForNestedProperty(string $property, string $rootAlias, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator/*, string $resourceClass*/): array
+    protected function addJoinsForNestedProperty(string $property, string $rootAlias, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator/*, string $resourceClass, string $joinType*/): array
     {
         if (\func_num_args() > 4) {
             $resourceClass = func_get_arg(4);
@@ -283,12 +283,18 @@ abstract class AbstractFilter implements FilterInterface
             $resourceClass = null;
         }
 
+        if (\func_num_args() > 5) {
+            $joinType = func_get_arg(5);
+        } else {
+            $joinType = null;
+        }
+
         $propertyParts = $this->splitPropertyParts($property, $resourceClass);
         $parentAlias = $rootAlias;
         $alias = null;
 
         foreach ($propertyParts['associations'] as $association) {
-            $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $parentAlias, $association);
+            $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $parentAlias, $association, $joinType);
             $parentAlias = $alias;
         }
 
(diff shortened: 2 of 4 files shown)


src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -171,4 +174,38 @@ protected function extractProperties(Request $request/*, string $resourceClass*/
...
+            }
+            $resourceClass = null;
+        }
+
+        $propertyParts = $this->splitPropertyParts($property, $resourceClass);
+        $parentAlias = $rootAlias;
+        $alias = null;
+
+        foreach ($propertyParts['associations'] as $association) {
+            // Force left joins on ordering because we don't want to filter
```
teohhanhui: Hmm... Is it possible to add another argument to the function instead of duplicating it here? (I know it already suffers from too many arguments...) :rofl:

teohhanhui: The `AbstractFilter` method... How will this be different in 3.0?

soyuka: because we would remove a lot of code for this 4th argument :p. 

Anyway because it could be necessary in other filters I'll add a 5th one!

soyuka:
Dunno why coverage failed with exit code 1 :p.

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -283,12 +283,24 @@ protected function addJoinsForNestedProperty(string $property, string $rootAlias
             $resourceClass = null;
         }
 
+        if (\func_num_args() > 5) {
+            $joinType = func_get_arg(5);
+        } else {
+            if (__CLASS__ !== \get_class($this)) {
+                $r = new \ReflectionMethod($this, __FUNCTION__);
+                if (__CLASS__ !== $r->getDeclaringClass()->getName()) {
+                    @trigger_error(sprintf('Method %s() will have a sixth `$joinType` argument in version API Platform 3.0. Not defining it is deprecated since API Platform 2.3.', __FUNCTION__), E_USER_DEPRECATED);
```
teohhanhui: Actually we should not introduce a deprecation in a patch release. We should add the deprecation error in a separate PR targeting `master`. WDYT?

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -283,12 +283,24 @@ protected function addJoinsForNestedProperty(string $property, string $rootAlias
...
+            $joinType = null;
+        }
+
         $propertyParts = $this->splitPropertyParts($property, $resourceClass);
         $parentAlias = $rootAlias;
         $alias = null;
 
         foreach ($propertyParts['associations'] as $association) {
-            $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $parentAlias, $association);
+            $alias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $parentAlias, $association, $joinType);
```
mauchede: If we use another filter before `OrderFilter` on the same field, the join type will be `INNER`.

teohhanhui: Which is correct. But for such cases, we need a way to replace the join with an inner join when there was a previous left join.

soyuka: Definitely right, I'll try to fix this as well.

src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
```diff
@@ -60,12 +61,25 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
...
                 $context['filters'] = $context['filters'] ?? [];
                 $filter->apply($queryBuilder, $queryNameGenerator, $resourceClass, $operationName, $context);
             }
         }
+
+        if (null !== $orderFilter) {
+            $context['filters'] = $context['filters'] ?? [];
+            $orderFilter->apply($queryBuilder, $queryNameGenerator, $resourceClass, $operationName, $context);
+        }
     }
```
soyuka: This is the easiest fix I've found, let me know how you feel about it :) (@meyerbaptiste I think it's just fine like this, no need to get more complicated)

src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
```diff
@@ -60,12 +61,25 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
             return;
         }
 
+        $orderFilter = null;
+
         foreach ($resourceFilters as $filterId) {
             $filter = $this->getFilter($filterId);
             if ($filter instanceof FilterInterface) {
+                // Apply the OrderFilter after every other filter to avoid an edge case where OrderFilter would do a LEFT join instead of an INNER join
```
teohhanhui: left join
inner join

or

LEFT JOIN
INNER JOIN

src/Bridge/Doctrine/Orm/Extension/FilterExtension.php
```diff
@@ -60,12 +61,25 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
             return;
         }
 
+        $orderFilter = null;
+
         foreach ($resourceFilters as $filterId) {
             $filter = $this->getFilter($filterId);
             if ($filter instanceof FilterInterface) {
+                // Apply the OrderFilter after every other filter to avoid an edge case where OrderFilter would do a LEFT JOIN instead of an INNER JOIN
+                if ($filter instanceof OrderFilter) {
```
bendavies: it wouldn't be better to give filters priorities than hard code like this?

soyuka: It may be better but it'd have required more work (we currently have 2 way of instantiating filters), and there is only this case that's impacted by the order. This quick fix is enough for now imo.

PR #2148, opened 2018-08-06
WIP: Fix/router bridge match

juferchaud:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | hope yes
| Fixed tickets | #1006
| License       | MIT
| Doc PR        | -

*Please update this template with something that matches your PR*

Ref https://github.com/api-platform/core/pull/ 1620 which description really fixes #1006 (same issue for me).

PR #2150, opened 2018-08-06
[Router bridge] Keep host context

### LINKED ISSUE #1006: Problem when specifying trusted_hosts
1. `Untrusted Host "localhost"` when fixing the `Request` for filters.
2. `Untrusted Host "schema.org"`, etc. (from external IRI) when calling `IriConverter::getItemFromIri`.

juferchaud:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1006
| License       | MIT
| Doc PR        | -

fixes #1006

### DIFF
--- a/src/Bridge/Symfony/Routing/Router.php
+++ b/src/Bridge/Symfony/Routing/Router.php
@@ -71,7 +71,7 @@ final class Router implements RouterInterface, UrlGeneratorInterface
         $baseContext = $this->router->getContext();
         $pathInfo = str_replace($baseContext->getBaseUrl(), '', $pathInfo);
 
-        $request = Request::create($pathInfo);
+        $request = Request::create($pathInfo, 'GET', [], [], [], ['HTTP_HOST' => $baseContext->getHost()]);
         $context = (new RequestContext())->fromRequest($request);
         $context->setPathInfo($pathInfo);
         $context->setScheme($baseContext->getScheme());


dunglas:
Thanks for fixing this! It’s very appreciated. Can you add a non-regression test please?

teohhanhui:
@dunglas Are we sure this is the right way forward? Why do we have a router decorator in the first place? I think the problem with `app_dev.php` (the original reason mentioned for having the decorator) should be fixed by another way.

juferchaud:
@dunglas it seems complicated for me to implement a non-regression test, because it's not possible to expect a call on static methods with phpunit. If you have an idea on how to test it, do not hesitate to share it ;). (the match method could also be changed https://github.com/api-platform/core/pull/2148).

Meanwhile, to fix this issue in my project, i overrided the api_platform.router service with my own patch.

dunglas:
@teohhanhui I don't see any other straightforward solution.

Maybe can we use `$request->duplicate` instead, and override its pathinfo. WDYT @juferchaud?

src/Bridge/Symfony/Routing/Router.php
```diff
@@ -71,7 +71,7 @@ public function match($pathInfo)
         $baseContext = $this->router->getContext();
         $pathInfo = str_replace($baseContext->getBaseUrl(), '', $pathInfo);
 
-        $request = Request::create($pathInfo);
+        $request = Request::create($pathInfo, 'GET', [], [], [], ['HTTP_HOST' => $baseContext->getHost()]);
```
dunglas: You can just call `$context->setHost($baseContext->getHost());` instead of this change. It will be a bit cleaner.

dunglas:
@teohhanhui I don't have any other way in mind. We have a decorator to implement our own interface. And we must have our own interface to allow using other routers than the Symfony one. Having a PSR for routers would be better, but we haven't one now.

The only middle term solution I can see is moving the RouterInterface of Symfony in https://github.com/symfony/contracts, and use it in API Platform. But for now I suggest to merge this one.

soyuka:
@juferchaud would it be possible to check the `HTTP_HOST` from behat?

dunglas:
Thanks a lot for having fixed this @juferchaud! If you find a way to add a non-regression test, don't hesitate to open a follow up PR ;)

teohhanhui:
> We have a decorator to implement our own interface. And we must have our own interface to allow using other routers than the Symfony one.

In which case we should use composition, not decoration.

PR #2162, opened 2018-08-19
[WIP] Phpcr support

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| License       | MIT

This one is really hot-WIP. At the end i wanna get PHPCR support at the API together with mongoDB. That is the reason why i depend on @alanpoulain's branch. We also have to be carefull to devide common stuff and implementation stuff.

Issues i faced and goanna face:

- Child Assosiation are different to Reference asscociations in a phpcr 
- we do not have any mappedBy|inversedBy
- Creation and Movement of Nodes in the tree can be restricted.

@dbu please feel the ping to be informed, no hurry to invest time,but be informed :-)


composer.json
```diff
@@ -16,6 +16,7 @@
         "php": ">=7.1",
 
         "doctrine/inflector": "^1.0",
+        "jackalope/jackalope-doctrine-dbal": "^1.3",
```
dbu: does api platform not work without phpcr? i fear that this is an "optional dependency". would it be possible to do a api-platform/phpcr integration package instead?

composer.json
```diff
@@ -28,6 +29,7 @@
     "require-dev": {
         "symfony/http-foundation": "^3.1@dev || ^4.0@dev",
 
+        "alcaeus/mongo-php-adapter": "^1.1",
```
dbu: i would recommend to not mix mongodb and phpcr support into one pull request. it will be easier to review if its first one and then the other once that goes through.

features/bootstrap/FeatureContext.php
```diff
@@ -113,9 +122,10 @@ public function removeAcceptHeaderBeforeScenario()
      */
     public function createDatabase()
     {
-        $this->schemaTool->dropSchema($this->classes);
+        $this->isOrm() && $this->schemaTool->dropSchema($this->classes);
+        $this->isOdm() && $this->schemaManager->dropDatabases();
```
dbu: could it make sense to split the context into two subclasses instead?

composer.json
ElectricMaxxx: mh maybe at the end. but i do not know if we want to move this bundle in the same virtual packagage situation, we are, shall we?

features/bootstrap/FeatureContext.php
ElectricMaxxx: There is a lot to separate at the end. I think we should move the complete extension stuff into separate classes as we did it in our `sonata-admin-integration-bundle` when the branches a really separated, we should see what each implementation is needing.

ElectricMaxxx:
@dbu  is tried to do it on top of the other branch, but selected the wrong one. Now i can not change it anymore.

ElectricMaxxx:
will move that one to the right branch.

Issue #2103, opened 2018-07-15
Subresource routes missing


thasmo:
Resource:
```php

/**
 * @ORM\Entity(repositoryClass="App\Repository\UserRepository")
 * @ApiResource
 */
class User
{
    /**
     * @ORM\ManyToOne(targetEntity="App\Entity\Region", inversedBy="users", fetch="EAGER", cascade={"persist"})
     * @ApiSubresource(maxDepth=1)
     */
    private $region;

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\Player", mappedBy="user", fetch="EXTRA_LAZY", cascade={"persist", "remove"}, orphanRemoval=true)
     * @ApiSubresource(maxDepth=1)
     */
    private $players;

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\Enqueuement", mappedBy="user", fetch="EXTRA_LAZY", cascade={"persist", "remove"}, orphanRemoval=true)
     * @ApiSubresource(maxDepth=1)
     */
    private $enqueuements;
}
```

Routes:
```
  api_users_get_collection                   GET      ANY      ANY    /users.{_format}
  api_users_post_collection                  POST     ANY      ANY    /users.{_format}
  api_users_get_item                         GET      ANY      ANY    /users/{id}.{_format}
  api_users_delete_item                      DELETE   ANY      ANY    /users/{id}.{_format}
  api_users_put_item                         PUT      ANY      ANY    /users/{id}.{_format}
  api_users_region_get_subresource           GET      ANY      ANY    /users/{id}/region.{_format}
```

Why are routes for `players` and `enqueuements` missing?

thasmo:
It seems that for all entities only the first `ApiSubresource` annotation works, all subsequent annotations have no effect.

thasmo:
If I actually remove `@ApiSubresource(maxDepth=1)` from the `$region` property, a route for `$players` is available and working, but the route for `$enqueuements` is still missing.

dunglas:
It looks like a big issue to me. Do you think you can try to fix?

soyuka:
duplicate of https://github.com/api-platform/core/issues/1711 fixed in https://github.com/api-platform/core/commit/0eb9602de46073f8c3360c55914754b9e694b4cc should be available in 2.2 and 2.3

What's your version?

thasmo:
@soyuka I'm on `api-platform/api-pack:1.1.0` and `api-platform/core:2.3.0`.

[...]

Issue #1711, opened 2018-02-18
@ApiSubresource with maxDepth only works for the first property in class


niy:
When using ApiSubresource annotation with maxDepth option in release v2.2.1, if I have multiple properties in a class (all properties are collections and are ManyToMany or OneToMany), only the first property would come up in the swagger doc and the routes of other subResources are not even generated.
But if maxDepth option is omitted, then all the subResources' routes are generated correctly.

My User class:
```
...
    /**
     * @var Poet[] List of all users favorite poets.
     * @ORM\ManyToMany(targetEntity="App\Entity\Poet", fetch="EXTRA_LAZY")
     * @ApiSubresource(maxDepth=1)
     */
    protected $poets;

    /**
     * @var Book[] List of all users favorite books.
     * @ORM\ManyToMany(targetEntity="App\Entity\Book", fetch="EXTRA_LAZY")
     * @ApiSubresource(maxDepth=1)
     */
    protected $books;

    /**
     * @var Poem[] List of all users favorite poems.
     * @ORM\ManyToMany(targetEntity="App\Entity\Poem", fetch="EXTRA_LAZY")
     * @ApiSubresource(maxDepth=1)
     */
    protected $poems;
...
```

niy:
and I just figured out that if I increase the `maxDepth` of each property by 1 it will work. that is a very strange behavior.

I changed the User class to this and all the routes for these 3 subresources generated successfully:
```
...
    /**
     * @var Poet[] List of all users favorite poets.
     * @ORM\ManyToMany(targetEntity="App\Entity\Poet", fetch="EXTRA_LAZY")
     * @ApiSubresource(maxDepth=1)
     */
    protected $poets;

    /**
     * @var Book[] List of all users favorite books.
     * @ORM\ManyToMany(targetEntity="App\Entity\Book", fetch="EXTRA_LAZY")
     * @ApiSubresource(maxDepth=2)
     */
    protected $books;

    /**
     * @var Poem[] List of all users favorite poems.
     * @ORM\ManyToMany(targetEntity="App\Entity\Poem", fetch="EXTRA_LAZY")
     * @ApiSubresource(maxDepth=3)
     */
    protected $poems;
...
```

And if for example I set the `$poems` maxDepth to 2 or 1, it would not work.

Nightbr:
hey, look at this if it can help you: https://github.com/api-platform/core/issues/1717#issuecomment-370349445

Nightbr:
[...]

PR #2164, opened 2018-08-20
Fix reset on maxDepth

Nightbr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1711, #2103
| License       | MIT
| Doc PR        | -

Seems to be we have an issue with the reset of Subresource maxDepth.

We need to reset the $depth if we return to the $rootResourceClass and this before the maxDepth Check... If not, only the maxDepth of the FIRST subresource of the rootResourceClass is generating...

But I have one fear on this: the case of a self-referencing subresource. DummyEntity -subRe-> DummyEntity. Because the maxDepth will be always reset to 0.

I think I will try to create a test for this special case and How is it handle without maxDepht btw?

ping @soyuka We already have this discussion in my first PR:
image

What do you think of this?

Cheers!

### DIFF
--- a/src/Operation/Factory/SubresourceOperationFactory.php
+++ b/src/Operation/Factory/SubresourceOperationFactory.php
@@ -88,6 +88,8 @@ final class SubresourceOperationFactory implements SubresourceOperationFactoryIn
             // Handle maxDepth
             if ($rootResourceClass === $resourceClass) {
                 $maxDepth = $subresource->getMaxDepth();
+                // reset depth when we return to rootResourceClass
+                $depth = 0;
             }
 
             if (null !== $maxDepth && $depth >= $maxDepth) {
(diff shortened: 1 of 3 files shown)


soyuka:
> But I have one fear on this: the case of a self-referencing subresource. DummyEntity -subRe-> DummyEntity. Because the maxDepth will be always reset to 0.
> 
> I think I will try to create a test for this special case and How is it handle without maxDepht btw?

IMO we should definitely try that case yes! Do you think that we could simplify the writing of tests on this one (might be hard)? I wanted to look into it but time is an issue :p.

Nightbr:
Hey, time is an issue for me too... 
I think it could be nice to simplify tests around subresources (a SubresourceUtils class to create and generate mocked Dummy Subresources) but I think it's another task and need time I don't have right now...
I will just write the test with self-referencing Entity Subresource (DummyEntity -subRe-> DummyEntity) and see what's happening and if it's not an huge infinite loop of the death I think we could use this PR as a quick fix for issues #1711, #2103

Nightbr:
Test ok! 
In fact, when the first subresource operation is generated, it is added to the $visited array and that the safe guard against infinite loop on self-referencing subresource Entity.

Seems good to me :ok_hand:  let me know what do you think!

soyuka:
LGTM thanks @Nightbr

PR #2165, opened 2018-08-20
Fix tests on low dependencies mink-extension@^2.3

soyuka:


composer.json
```diff
@@ -33,7 +33,7 @@
         "behat/mink-browserkit-driver": "^1.3.1",
         "behat/mink-extension": "^2.2",
         "behat/symfony2-extension": "^2.1.1",
-        "behatch/contexts": "^3.1",
+        "behatch/contexts": "3.1.0",
```
teohhanhui: :disappointed:

soyuka: haven't found a better way to keep low deps from working :|.

PR #2174, opened 2018-08-22
Fix Circle CI

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,7 +142,7 @@ jobs:
           name: Run PHPUnit tests
           command: |-
             mkdir -p build/logs/tmp build/cov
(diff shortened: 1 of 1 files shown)


teohhanhui:
Hmm... But this means we're no longer splitting the tests.

teohhanhui:
What's the underlying cause that's causing the tests to fail when they're split?

dunglas:
@teohhanhui it's just a WIP, because on my computer the command returns 0 as expected (but I don't have the circle tool installed).

Issue #2149, opened 2018-08-06
GraphQL: can't filter nested objects


Siregacvek:
Hello.

I try to filter nested entities. Example:
I have the following schema: `Entity 1<-* SubEntity 1<-* SubSubEntity`. Each entity has string property `Name`.

When I add the following filter to `Entity`:
```
/**
 * @ApiResource()
 ...
 */
class Entity
{
    /**
     * @ApiFilter(SearchFilter::class, properties={"name": "ipartial"})
     */
    private $subEntities;
}

```
this query works fine:
```
{
  entities {
    subEntities (name: "Name") {
    ...
    }
  }
}

```
I get all the `entities` with filtered `subEntities`.

But when I add new filter:
```
...
    /**
     * @ApiFilter(SearchFilter::class, properties={"name": "ipartial"})
     * @ApiFilter(SearchFilter::class, properties={"subSubEntities": "exact"})
     */
    private $subEntities;
```
or
```
...
    /**
     * @ApiFilter(SearchFilter::class, properties={"name": "ipartial"})
     * @ApiFilter(SearchFilter::class, properties={"subSubEntities.name": "ipartial"})
     */
    private $subEntities;
```

The Schema builder even doesn't see this second filter. I suspect that the reason is that the new filter is not of scalar type, but an object. 

Am I doing something wrong or it's a bug?

Siregacvek:
@alanpoulain I explored the code and what I found.
My first filter `* @ApiFilter(SearchFilter::class, properties={"name": "ipartial"})` works only because `Entity` and `SubEntity` have the same field `Name`.
I believe the issue is here:
https://github.com/api-platform/core/blob/797cea682956871c0a848275bbaeabf1339ad615/src/GraphQl/Type/SchemaBuilder.php#L229

Your code trying to find the search property in a root class `Entity`, but my filter is given over a property `SubEntity`, and I think it must search in `SubEntity`'s properties. That's why schema builder doesn't work with the second filter `* @ApiFilter(SearchFilter::class, properties={"subSubEntities.name": "ipartial"})` - root class `Entity` doesn't have `subSubEntity` property and this filter just ignoring.

Siregacvek:
@alanpoulain regarding behat test - sorry I don't know how to launch test from your environment, but I believe that if here 
https://github.com/api-platform/core/blob/797cea682956871c0a848275bbaeabf1339ad615/features/graphql/filters.feature#L87-L108

you will filter by a unique relatedDummy's field like this:
```
[...]

PR #2177, opened 2018-08-22
[GraphQL] Fix subresource context

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2149
| License       | MIT
| Doc PR        |

When a property was corresponding to a subresource, the context given to `getResourceFieldConfiguration` was wrong.
It was creating issues like the one referenced, because the generated arguments for the collection were the parent's ones.
The following test was working because the parent entity (`Dummy`) has a search filter for its `name` field:
https://github.com/api-platform/core/blob/ebe842f3cc8d22f90f3099484b615bb183e9f8f5/features/graphql/filters.feature#L87-L113

### DIFF
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -448,9 +448,15 @@ final class SchemaBuilder implements SchemaBuilderInterface
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


alanpoulain:
Test is already in the suite. It was working before by chance.

soyuka:
thanks @alanpoulain !

Issue #2169, opened 2018-08-21
Allow Constructors with Relationships


ragboyjr:
Currently, Resource constructors will fail if they contain a relationship entity.

```php
class Book
{
    private $id;
    private $author;
    private $title;

    public function __construct(Author $author, string $title) {
        $this->author = $author;
        $this->title = $title;
    }

    public function getId(): ?int { return $this->id; }
    public function getAuthor(): Author { return $this->author; }
    public function getTitle(): string { return $this->title; }
}
```

This is because the ItemNormalizer in API Platform only denormalizes a relationship entity on the `setAttributeValue` method call. However, the `instantiateObject` method call in the normalizer is done before the relationship value has been denormalized.

I think this could be trivial to implement if we override the instantiateObject, check if any of the properties are relationships, and if so, we denormalize them ahead of time so that whether a value is set via setAttributeValue or instantiateObject, it will be denormalized properly.

Thoughts?

ragboyjr:
This is the same issue as #2123

bendavies:
I think there are extensive tests for this here from @komik966
https://github.com/api-platform/core/pull/1991

I think it's very unfortunate that https://github.com/api-platform/core/pull/1749 was merged without these tests.

ragboyjr:
I think we can close this now that @komik966's changes have been merged.

Thanks!

PR #2178, opened 2018-08-25
Constructor relations writtable

komik966:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2169
| License       | MIT
| Doc PR        | todo

Connected with: https://github.com/api-platform/core/pull/1749, https://github.com/symfony/symfony/pull/28263

TODO:
- [ ] make documentation normalizers (+ GraphQL?) aware of VOs
- [ ] add unit tests

### DIFF
--- a/features/serializer/vo_relations.feature
+++ b/features/serializer/vo_relations.feature
@@ -1,0 +1,187 @@
+Feature: Value object as ApiResource
+  In order to keep ApiResource immutable
+  As a client software developer
+  I need to be able to use class without setters as ApiResource
+
+  @createSchema
+  Scenario: Create Value object resource
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/vo_dummy_cars" with body:
+    """
+    {
+        "mileage": 1500,
+        "bodyType": "suv",
+        "make": "CustomCar",
+        "insuranceCompany": {
+            "name": "Safe Drive Company"
+        },
+        "drivers": [
+            {
+                "firstName": "John",
+                "lastName": "Doe"
+            }
+        ]
+    }
+    """
+    Then the response status code should be 201
+    And the JSON should be equal to:
+    """
+     {
+         "@context": "/contexts/VoDummyCar",
+         "@id": "/vo_dummy_cars/1",
+         "@type": "VoDummyCar",
+         "mileage": 1500,
+         "bodyType": "suv",
+         "inspections": [],
+         "make": "CustomCar",
+         "insuranceCompany": {
+             "@id": "/vo_dummy_insurance_companies/1",
+             "@type": "VoDummyInsuranceCompany",
+             "name": "Safe Drive Company"
+         },
+         "drivers": [
+             {
+                 "@id": "/vo_dummy_drivers/1",
+                 "@type": "VoDummyDriver",
+                 "firstName": "John",
+                 "lastName": "Doe"
+             }
+         ]
+     }
+    """
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+
+  Scenario: Create Value object with IRI and nullable parameter
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/vo_dummy_inspections" with body:
+    """
+    {
+        "accepted": true,
+        "car": "/vo_dummy_cars/1"
+    }
+    """
+    Then the response status code should be 201
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "required": ["accepted", "performed", "car"],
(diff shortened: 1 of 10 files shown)


komik966:
For now I've made `Swagger\DocumentationNormalizer` to not add `readOnly` label when property is initializable.

But I think that it can be more guessable, consider example:
Users can change theirs password but not email:
```php
class User
{
    private $email;
    private $password;

    public function __construct(string $email) { $this->email = $email; }

    public function getEmail(): string { return $this->email; }

    public function getPassword(): string { return $this->password; }

    public function setPassword(string $password): void { $this->password = $password; }
}
```
Then documentation should looks like this:
```
POST - input / output
{
  "email": "string",
  "password": "string"
}

PUT - input
{
  "password": "string"
}

PUT - output
{
  "email": "string",
  "password": "string"
}
```
Such documentation behavior will allow us generating CRUD clients with distinction between what should appear in create form and what in update form, without the need to define serialization groups.

If we're ok with this, I prefer to create it in different PR. For now non writable properties are included in documentation for `PUT` methods, but are ignored during request - check it: https://github.com/api-platform/core/blob/4309ffe08eeb0b5ee578f4d26885c61de250467a/features/serializer/vo_relations.feature#L83

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -127,6 +129,91 @@ public function denormalize($data, $class, $format = null, array $context = [])
         return parent::denormalize($data, $class, $format, $context);
     }
 
+    /**
```
dunglas: Please mark it as `@internal` because we'll remove it at some point.

komik966: Done

tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -299,6 +300,10 @@ public function testNormalize()
                             'type' => 'string',
                             'description' => 'This is a name.',
                         ]),
+                        'description' => new \ArrayObject([
+                            'type' => 'string',
+                            'description' => 'This is a initializable but not writable property.',
```
teohhanhui: a -> an

komik966: :+1:  Fixed

borNfreee:
please, don't forget to update documentation.

It's so painful when documentation is updated post-factum by someone who spends hours debugging, not in advance.

thanks for this feature btw!

features/serializer/vo_relations.feature
```diff
@@ -0,0 +1,187 @@
...
+          "enum": ["/contexts/Error"]
+        },
+        "type": {
+          "enum": ["hydra:Error"]
+        },
+        "hydra:title": {
+          "enum": ["An error occurred"]
+        },
+        "hydra:description": {
+          "enum": ["Cannot create an instance of ApiPlatform\\Core\\Tests\\Fixtures\\TestBundle\\Entity\\VoDummyCar from serialized data because its constructor requires parameter \"drivers\" to be present."]
```
bendavies: is there a way this can be more consumer friendly?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -127,6 +129,93 @@ public function denormalize($data, $class, $format = null, array $context = [])
...
+                    // Don't run set for a parameter passed to the constructor
+                    unset($data[$key]);
+                } elseif (isset($context[static::DEFAULT_CONSTRUCTOR_ARGUMENTS][$class][$key])) {
+                    $params[] = $context[static::DEFAULT_CONSTRUCTOR_ARGUMENTS][$class][$key];
+                } elseif ($constructorParameter->isDefaultValueAvailable()) {
+                    $params[] = $constructorParameter->getDefaultValue();
+                } else {
+                    throw new MissingConstructorArgumentsException(
+                        sprintf(
+                            'Cannot create an instance of %s from serialized data because its constructor requires parameter "%s" to be present.',
```
bendavies: can this error be more consumer friendly on serialization? it's exposing implementation detail to the consumer which they don't care about: the class name, overly verbose message etc... all the consumer needs to know is that they didn't provide the some parameter.

komik966: This code comes from `symfony/serializer` (see method comment). So I'm not sure if we should change it. But I've similar opinion. Ideally would be to show such errors the same way as validation errors (as map: property => error message). I think that non nullable and without default value constructor parameters are equivalent to class attributes with `NotBlank` constraint (with some exceptions) and errors should be presented the same way.

PR #2181, opened 2018-08-28
[GraphQL] Only exposes the mutations if any

sroze:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ø
| License       | MIT
| Doc PR        | ø

Otherwise, GraphiQL gives this lovely message:
> Mutation fields must be an object with field names as keys or a function which returns [...]

### DIFF
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -92,16 +92,21 @@ final class SchemaBuilder implements SchemaBuilderInterface
             }
         }
 
-        return new Schema([
+        $schema = [
             'query' => new ObjectType([
                 'name' => 'Query',
                 'fields' => $queryFields,
             ]),
-            'mutation' => new ObjectType([
+        ];
+
+        if ($mutationFields) {
+            $schema['mutation'] = new ObjectType([
                 'name' => 'Mutation',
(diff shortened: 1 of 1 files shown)


src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -92,16 +92,21 @@ public function getSchema(): Schema
...
-        return new Schema([
+        $schema = [
             'query' => new ObjectType([
                 'name' => 'Query',
                 'fields' => $queryFields,
             ]),
-            'mutation' => new ObjectType([
+        ];
+
+        if (\count($mutationFields) > 0) {
```
dunglas: `if ($mutationFields)`?

alanpoulain:
Maybe you could add a unit test or a Behat one (by using introspection)?

sroze:
Rebased and changed check with a simple `if`.

@alanpoulain based on the complexity of the `SchemaBuilderTest`, I don't think it's worth it. Also, I'll follow up with a fair amount of pull-requests regarding the messenger integration in the next few weeks that should cover this use-case :)

dunglas:
Thanks @sroze

PR #2187, opened 2018-09-04
Fix coverage

soyuka:
attempt at fixing coverage sorry 'bout the noise


soyuka:
I'm not sure if the coverage actually ran ?! @teohhanhui could you confirm it went according to plan?

dunglas:
Thanks @soyuka

teohhanhui:
Hmm... This has caused us to run out of memory on CircleCI.

Investigating why the exit code was wrong...

PR #799, opened 2016-10-12
Mask required when writable false in PropertyMetadata

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Follow up from https://github.com/api-platform/core/pull/764#discussion_r81112046

/cc @dunglas

PR #2188, opened 2018-09-04
Add Behat test for PR 2183

ilicmsreten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This test is related with approved [PR](https://github.com/api-platform/core/pull/2183) and reported bug [#799](https://github.com/api-platform/api-platform/issues/799).

### DIFF
--- a/features/bootstrap/DoctrineContext.php
+++ b/features/bootstrap/DoctrineContext.php
@@ -11,11 +11,13 @@
 
 declare(strict_types=1);
 
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\Address as AddressDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\Answer as AnswerDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\CompositeItem as CompositeItemDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\CompositeLabel as CompositeLabelDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\CompositePrimitiveItem as CompositePrimitiveItemDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\CompositeRelation as CompositeRelationDocument;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\Customer as CustomerDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\Dummy as DummyDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyAggregateOffer as DummyAggregateOfferDocument;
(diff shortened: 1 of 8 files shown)


soyuka:
Nice, many thanks! looks like tests are failing on `deps=LOW` though, is this feature limited to a given version of doctirne? If that's the case, we should maybe bump that dependency...

composer.json
```diff
@@ -36,7 +36,7 @@
         "behatch/contexts": "3.1.0",
         "doctrine/annotations": "^1.2",
         "doctrine/doctrine-bundle": "^1.8",
-        "doctrine/orm": "^2.5.2",
+        "doctrine/orm": "^2.6.2",
```
dunglas: Isn't 2.6.0 enough? It would be better to set as minimal version the first Doctrine version supporting this feature.

composer.json
```diff
@@ -54,7 +54,7 @@
         "symfony/console": "^3.4 || ^4.0",
         "symfony/debug": "^2.8 || ^3.0 || ^4.0",
         "symfony/dependency-injection": "^3.4 || ^4.0",
-        "symfony/doctrine-bridge": "^2.8.12 || ^3.0 || ^4.0",
+        "symfony/doctrine-bridge": "^4.0",
```
dunglas: If it is supported by Symfony 4.0, it should also be supported by Symfony 3.4. We should at least support 3.4 because it's the latest LTS version.

composer.json
ilicmsreten: @dunglas yes. I edit composer.json

soyuka:
Mhh maybe we should only push this (and https://github.com/api-platform/core/pull/2183) onto master because of versions bump @dunglas wdyt?

dunglas:
I agree with @soyuka, it may cause big issues for existing projects. @ilicmsreten do you think we can use some kind of feature detection instead of bumping the version? If I understand correctly the patch, bumping the version if only necessary for tests, not for the actual code? If true, we may just skip these tests when old versions are used. WDYT?

dunglas:
> So test will pass after bump versions, but we can skip test when old versions is used.
👍 Let's do this!

antograssiot:
Given api-platform/core#2205 I guess this can be closed ?

ilicmsreten:
@antograssiot yes cause https://github.com/api-platform/core/pull/2205, but this test need to pass after fixing bug https://github.com/api-platform/api-platform/issues/799. 
This is just test for https://github.com/api-platform/api-platform/issues/799 (after all there no needs to edit composer.json).

soyuka:
> Given api-platform/core#2205 I guess this can be closed ?

We still want the hydration feature, we just need to move it elsewhere (ie not in the eager loading extension because there's no pagination yet...)

antograssiot:
I'm working on finishing this one

soyuka:
thanks @antograssiot @ilicmsreten

PR #2193, opened 2018-09-07
[GraphQL] Allow non-writeable fields in mutation response

jamesisaac:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | 
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

It makes sense for non-writeable fields to be excluded from the GraphQL **input** types, but they should surely be accessible in the payload data read back from a mutation.

To give a simple example, `Person._id` is non-writeable, as my IDs are auto generated by the server, but I still want to be able to see the generated ID after calling `createPerson`.

```graphql
  mutation CreatePersonMutation {
    createPerson(input: {name: "Alice"}) {
      _id
      name
    }
  }
```

The above query would previously have failed (`_id` not found in `createPersonPayload`) -- this PR corrects that.

### DIFF
--- a/features/graphql/introspection.feature
+++ b/features/graphql/introspection.feature
@@ -304,11 +304,10 @@ Feature: GraphQL introspection support
     And the JSON node "data.typeCreateInput.inputFields[0].name" should be equal to "bar"
     And the JSON node "data.typeCreateInput.inputFields[1].name" should be equal to "baz"
     And the JSON node "data.typeCreateInput.inputFields[2].name" should be equal to "clientMutationId"
-    And the JSON node "data.typeCreatePayload.fields" should have 4 elements
+    And the JSON node "data.typeCreatePayload.fields" should have 3 elements
     And the JSON node "data.typeCreatePayload.fields[0].name" should be equal to "id"
     And the JSON node "data.typeCreatePayload.fields[1].name" should be equal to "bar"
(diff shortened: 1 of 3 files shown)


jamesisaac:
Hmm, sorry I'm a bit stuck on the logic here.  Not sure why it's having that unintended side effect on the Behat test.  Any help appreciated.

dunglas:
Hi @jamesisaac,

Thanks for reporting and fixing this. The patch looks good to me. It's probably the test suite that is faulty.

The 1st test should be updated (IIUC, thanks to this fix there is 1 more field available now).
I suspect the 2nd failure is related to #2194. You can add the step `And print last JSON response` to show what's going on exactly; but rebasing will probably be enough.

One last thing: as it's a bug fix, can you target the 2.3 branch please?

alanpoulain:
Thank you @jamesisaac.

jamesisaac:
Thanks for merging.  Sorry I didn't get to sorting out the changes in @dunglas's comment.  I had trouble getting the tests to run, and then became very caught up with work.

alanpoulain:
No problem, it was nice for you to contribute :slightly_smiling_face:

PR #2195, opened 2018-09-08
GraphQL: add "id" to attributes if "_id" is requested (fixes #2194)

### LINKED ISSUE #2194: [GraphQL] Requesting _id without id leads to "cannot return null for non-nullable field"
Perhaps related to #2132 and #2060

This query:

```graphql
{
  person(id: "/people/1") {
    _id
    name
  }
}
```

(Where `$id` on the model is a private field with just a getter, like in the example project) produces this error:

```json
{
  "errors": [
    {
      "debugMessage": "Cannot return null for non-nullable field Person._id.",
      "category": "internal",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "person",
        "_id"
      ]
    }
  ],
  "data": {
    "person": null
  }
}
```

Can be fixed by including `id` in the query:

```graphql
{
  person(id: "/people/1") {
    id
    _id
    name
  }
}
```

```json
{
  "data": {
    "person": {
      "id": "/people/1",
      "_id": 1,
      "name": "Alice"
    }
  }
}
```

Maybe this is not a priority until the refactoring of #2022 has been carried out.

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2194
| License       | MIT
| Doc PR        | n/a

Closes #2194.


jamesisaac:
Thanks for looking at this so quickly!  The proposed fix works on a single entity:

```graphql
{
  dummy(id: "/dummies/1") {
    _id
  }
}
```

but unfortunately still errors on a collection query:

```graphql
{
  dummies {
    _id
  }
}

**Comment by dunglas** [comment]
Ok I’ll update the PR. The same hack’must be applied in all resolvers.

File: src/GraphQl/Resolver/ItemResolver.php [hunk 1]
```diff
@@ -74,4 +74,25 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
 
         return $this->normalizer->normalize($item, ItemNormalizer::FORMAT, $normalizationContext + $baseNormalizationContext);
     }
+
+    /**
+     * Recursively replaces the key "_id" (the raw id i GraphQL) by "id" (the name of the property expected by the Serializer).
```
alanpoulain: i => in

src/GraphQl/Resolver/ItemResolver.php
```diff
@@ -74,4 +74,25 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
...
+     */
+    private function replaceIdKey(array $array): array
```
alanpoulain: I would give `$array` a better name, like `$selectedFields`.

src/GraphQl/Resolver/ItemResolver.php
```diff
@@ -74,4 +74,25 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
...
+    {
+        foreach ($array as $key => $value) {
```
alanpoulain: I would rename `$key` to `$fieldName` and `$value` to `$fieldValue`.

src/GraphQl/Resolver/ItemResolver.php
```diff
@@ -74,4 +74,25 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
...
+            if ('_id' === $key) {
+                $array['id'] = true;
```
alanpoulain: Maybe using `$array['_id']` instead of `true` would be better in case the implementation in `ResolveInfo` changes.

src/GraphQl/Resolver/ItemResolver.php
```diff
@@ -74,4 +74,25 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
...
+                unset($array['key']);
```
alanpoulain: `key`?

dunglas:
@jamesisaac can you test this updated version please?

src/GraphQl/Resolver/FieldsToAttributesTrait.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Transforms the passed GraphQL fields to the list of attributes to serialize.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+trait FieldsToAttributesTrait
+{
+    /**
+     * Recursively replaces the key "_id" (the raw id in GraphQL) by "id" (the name of the property expected by the Serializer) and flatten edge and node structures (pagination).
```
alanpoulain: This doc comment belongs to the other function.

alanpoulain: flattens

src/GraphQl/Resolver/FieldsToAttributesTrait.php
```diff
@@ -23,7 +23,7 @@
 trait FieldsToAttributesTrait
 {
     /**
-     * Recursively replaces the key "_id" (the raw id in GraphQL) by "id" (the name of the property expected by the Serializer) and flatten edge and node structures (pagination).
+     * Retrieve fields, recursively replaces the "_id" key (the raw id) by "id" (the name of the property expected by the Serializer) and flattens edge and node structures (pagination).
```
alanpoulain: Retrieves :stuck_out_tongue:

dunglas: 😭

PR #2206, opened 2018-09-18
Sligthly improved code - removed unnecessary step

Siregacvek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | didn't test

Just slight improvement - in a case if `$purgeItem === true` your code invoked `$this->tags[$iri]` twice.

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
+++ b/src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
@@ -111,12 +111,11 @@ final class PurgeHttpCacheListener
             return;
         }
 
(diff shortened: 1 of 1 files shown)


dunglas:
Unfortunately it breaks the tests, can you take a look?

Siregacvek:
I looked more closely at the code and realized that I was wrong). Sorry

Siregacvek:
@dunglas , your version was fine, we can decline my PR.

dunglas:
Thanks for contributing!

PR #2213, opened 2018-09-25
allow developer to specify iri for operations

videni:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| License       | MIT

now we can specify IRI for individual operation,  this PR resolves [issue 839](https://github.com/api-platform/api-platform/issues/839) I created,   we should let the developer decide what IRI is used for resource.
``` 
App\Bundle\OrderBundle\Entity\Order:
    collectionOperations:
        post:
            path: '/admin/orders'
            denormalization_context:
               groups: ["write"]
            access_control: is_granted('app_order_create')
        get:
            path: '/admin/orders'
            force_eager: false
            access_control: is_granted('app_order_view')
            iri:
                route: 'api_orders_get_item'
                parameters:
                    id: expr(object.getId())
        show_customer_orders:
            method: 'GET'
            force_eager: false
            path: /shop/account/orders
            iri:
                route: 'api_shop_account_order_view'
                parameters:
                    id: expr(object.getId())
    itemOperations:
        get:
            path: '/admin/orders/{id}'
            acl_enabled: true
            access_control: is_granted('app_order_view', object)
            iri:
                route: 'api_orders_get_item'
                parameters:
                    id: expr(object.getId())
        delete:
            path: '/admin/orders/{id}'
            access_control: is_granted('app_order_delete', object)
        show_customer_order_details:
            method: 'GET'
            route_name: 'api_shop_account_order_view'
            access_control: "object.getCustomer() === user"
    attributes:
        normalization_context:
            groups: ["Details", 'Default']
        denormalization_context:
            groups: ["write"]
        validation_groups:  ["app"]
```


src/Api/IriConverterInterface.php
```diff
@@ -43,7 +43,7 @@ public function getItemFromIri(string $iri, array $context = []);
      * @throws InvalidArgumentException
      * @throws RuntimeException
      */
-    public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH): string;
+    public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH, $context = []): string;
```
soyuka: We can't change this interface it breaks the backward compatibility :/.

videni: hi, there is a default value,  it should not break the compatibility, can you explain it further ?

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -50,6 +50,15 @@
             <argument type="service" id="router" />
         </service>
 
+        <service id="api_platform.expression.evaluator" class="ApiPlatform\Core\Bridge\Symfony\Routing\ExpressionEvaluator">
+            <argument type="service" id="api_platform.security.expression_language" />
+            <argument type="collection"/>
+            <call method="setContextVariable">
+                <argument>container</argument>
+                <argument type="service" id="service_container"/>
```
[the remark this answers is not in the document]
videni: you are right , the PropertyAccessor is enough for current case

src/Api/IriConverterInterface.php
soyuka: Sure, we respect the symfony backward compatibility promise and we can't add arguments:

https://symfony.com/doc/current/contributing/code/bc.html#changing-interfaces (see the `Add argument with a default value => No`.

Don't worry about this yet, I'm sure there's a solution that doesn't involve adding `context` here!

videni:
when serialize object has object properties like  an Order object has a property payment who is also an object.   this  PR won't work, because the resolved IRI for order is also passed into  payment.  the IRI is indeed a problem, see [Please let us to decide if an entity it is dereferencable or not, don't fore us to do this](https://github.com/api-platform/core/issues/2214)

videni:
has no properties that are not managed by api-platform .

src/Api/IriConverterInterface.php
dunglas: A solution is to add a new interface extending the existing one. See https://github.com/api-platform/core/blob/master/src/DataProvider/ContextAwareCollectionDataProviderInterface.php#L32 for instance.

dunglas:
It looks like this PR has been mangled with your fork. Can you restore only the appropriate changes to ease the review @videni?

For the record, I'm 👍 to allow "internal" IRI (urn for instance).

PR #2215, opened 2018-09-25
Fix: max depth not working when fetch eager is disabled

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Max depth annotation was not working in HAL when fetch eager was disabled.


dunglas:
Looks good to me, but it breaks an existing test. Can you add a new set of entities instead of modifying the existing one?

alanpoulain:
Done!

dunglas:
Thanks Alan!

PR #2219, opened 2018-09-27
[2.3] Throw an InvalidArgument when trying to get Item from a collection route

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

When using the search filter on an exact association for example, if the obviously bad value is a matching collection url, the filter IriConverter will throw `Either \"item_operation_name\" or \"collection_operation_name\" must be defined, unless the \"_api_receive\" request attribute is set to false.`

This is because `getItemFromIri()` will get a match on the router but won't match the expected the attributes.

an easy way to reproduce this on the demo is to try to filter `/reviews?book=foo` versus `/reviews?book=parchments` or `/reviews?book=book`

### DIFF
--- a/features/doctrine/search_filter.feature
+++ b/features/doctrine/search_filter.feature
@@ -374,6 +374,35 @@ Feature: Search filter on collections
     }
     """
 
+  @sqlite
+  Scenario: Search for entities with an existing collection route name
+    When I send a "GET" request to "/dummies?relatedDummies=dummy_cars"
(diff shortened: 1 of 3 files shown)


dunglas:
Nice DX improvement! Thanks @antograssiot

Issue #2155, opened 2018-08-12
[HELP NEEDED] PUT is not updating an entity


Caaliitis:
I have composer up to date for everything. Using this on Symfony 4.
Tried to do update on one of the entities 
``````

/**
 * @ORM\Entity
 * @ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")
 * @ApiResource(
 *     collectionOperations={
 *          "post_comment"={
 *               "route_name" = "api_add_comment",
 *               "method" = "POST"
 *           },
 *     "get_single"={
 *               "route_name" = "api_get_comment",
 *               "method" = "GET",
 *           },
 *     "get_all"={
 *               "route_name" = "api_get_comments",
 *               "method" = "GET",
 *           },
 *     },
 * )
 * @ORM\Entity(repositoryClass="App\Repository\CommentRepository")
 */
class Comment extends \FOS\CommentBundle\Entity\Comment implements SignedCommentInterface
{
``````

Sending `{"body":"test"}` to update the existing entity, but it returns old one with 200 OK status. No errors, no nothing. Happens to all of the entities. 
Any suggestions?

sroze:
Can you describe us how you send this HTTP request? With a reproducible
`curl` command for example?

Caaliitis:
I send {"body":"my text"} trough postman and content type as ld+json,
method put and URL 127.0.0.1:8000/api/comments/{commentId} in my case 1

jonag:
Since you have set the tracking policy of your entity to `@ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")` an explicit call to `persist()` need to be done on the entity manager in order to save the changes.

API Platform [don't call persist()](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Common/DataPersister.php#L54) if the entity is already managed, so I think that's why your changes are not saved in the database. Can you try to remove the `ChangeTrackingPolicy` on your entity to check if this fix your issue ?

soyuka:
Should be fixed by https://github.com/api-platform/core/pull/2226

PR #2226, opened 2018-10-02
Persist doctrine entities with DEFERRED_EXPLICIT change tracking policy

paxal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | maybe
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2155
| License       | MIT
| Doc PR        | ~

As reported in #2155 doctrine entities are not persisted if change tracking policy is `DEFERRED_EXPLICIT`.

This implementation relies on `ClassMetadataInfo`, which is the `ClassMetadata` implementation. As it's API might change, I've added a test on method existence.

### DIFF
--- a/src/Bridge/Doctrine/Common/DataPersister.php
+++ b/src/Bridge/Doctrine/Common/DataPersister.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\DataPersister\DataPersisterInterface;
 use ApiPlatform\Core\Util\ClassInfoTrait;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\Common\Persistence\ObjectManager as DoctrineObjectManager;
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
 
 /**
  * Data persister for Doctrine.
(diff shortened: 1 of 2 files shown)


soyuka:
Thanks @paxal !

paxal:
Thanks for your quick merge :)

backbone87:
why not just unconditionally call persist?

soyuka:
according to #2155 it does not?

backbone87:
The problem is the `contains` call, which prevents the persist call for existing (but changed) entities. Just drop the whole condition around the persist and let doctrine do the checks (which It does anyway when persist is called)

soyuka:
The `contains` was added in https://github.com/api-platform/core/pull/1782, reasons are stated there.

backbone87:
I dont get the resoning there. What is meant with „Changes from the outside“ and how does a persist call override them. This sounds to me more like a missuse of doctrine. Persist calls should never override anything: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/working-with-objects.html#working-with-objects

soyuka:
Mhh I think that it's in case of relations there were issues when persisting updated ones. Anyway, Doctrine doesn't check for `contains` and we definitely don't want to persist entities that aren't managed. From what I see in the UnitOfWork, `persist` doesn't do the check `contains` does. Also, the `persist` doc states:

> NOTE: The persist operation always considers entities that are not yet known to this EntityManager as NEW. Do not pass detached entities to the persist operation.

backbone87:
I think you missread the condition. Persist is called, when the uow does **not** contain the entity already

Issue #1801, opened 2018-03-29
Error when trying to get subresource binded by OneToOne relation


breitsmiley:
There are two entity **Users** and **Profile**. 
They binded by **OneToOne** relation via fields:

- $profile in Users entity (inverse side) 
- $user  in Profile entity (owning side)

Field $profile has **@API\ApiSubresource()** annotation. Api-platform creates route **/api/users/1/profile**. 

But when I try to GET data by this URL I  occurs an error:

> Type error: Argument 1 passed to Doctrine\\ORM\\Mapping\\DefaultQuoteStrategy::getJoinColumnName() must be of the type array, boolean given, called in C:\\TEST\\backend\\vendor\\doctrine\\orm\\lib\\Doctrine\\ORM\\Query\\AST\\Functions\\IdentityFunction.php on line 89

In case of using either ManyToOne or OneToMany associations all working perfectly.

**Users.php**
```
<?php

namespace App\Entity;

use ApiPlatform\Core\Annotation as API;
use Doctrine\ORM\Mapping as ORM;


/**
 * @ORM\Entity()
 *
 * @API\ApiResource()
 */
class Users
{
    /**
     * @var int
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var Profile
     * @ORM\OneToOne(targetEntity="Profile", mappedBy="user")
     * @API\ApiSubresource()
     */
    private $profile;

// Other properties, getters and setters next ...
}
```
**Profile.php**
```
<?php

namespace App\Entity;

use ApiPlatform\Core\Annotation as API;
use Doctrine\ORM\Mapping as ORM;


/**
 * @ORM\Entity()
 *
 * @API\ApiResource()
 */
class Profile
{
    /**
     * @var int
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var Users
     * @ORM\OneToOne(targetEntity="Users", inversedBy="profile")
     * @ORM\JoinColumn(nullable=false, onDelete="CASCADE")
     */
    private $user;

// Other properties, getters and setters next ...
}
```

Simperfit:
Could you please provide a reproducer for this ?

breitsmiley:
@Simperfit  Here you are [https://github.com/and-xdev/api-test](https://github.com/and-xdev/api-test)

1. clone repository
2. set up .env for DB etc.
[...]

PR #2228, opened 2018-10-03
fixed Bidirectional ONE_TO_ONE mapping

acabala:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1801
| License       | MIT
| Doc PR        | 

Background:
For Doctrine ONE_TO_ONE mapping there is possible to define Unidirectional and Bidirectional relations. [ref](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/association-mapping.html#one-to-many-bidirectional). For unidirectional case everything works properly, but for bidirectional an exception
```
Type error: Argument 1 passed to
Doctrine\ORM\Mapping\DefaultQuoteStrategy::getJoinColumnName() must be of the type array, boolean given, 
called in vendor\doctrine\orm\lib\Doctrine\ORM\Query\AST\Functions\IdentityFunction.php on line 89
```
is returned. The reason for that is that for this scenario syntax:
```
     * @OneToOne(targetEntity="SubEntity", mappedBy="entity")
````

My fix solves the problem by checking if there is a `mappedBy` attribute for an entity. If no, then it's processed as it used to be, and if yes, then additional query is being built.

@soyuka @alanpoulain @dunglas please take a look.

### DIFF
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -28,6 +28,7 @@ use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInte
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\QueryBuilder;
 
 /**
@@ -175,6 +176,17 @@ final class SubresourceDataProvider implements SubresourceDataProviderInterface
                     $qb->select($alias)
                         ->from($identifierResourceClass, $alias);
                     break;
+                // ONE_TO_ONE relation can be Unidirectional or Bidirectional. For Bidirectional mappedBy attribute is used.
+                case ClassMetadataInfo::ONE_TO_ONE:
+                    try {
+                        $mappedBy = $classMetadata->getAssociationMapping($previousAssociationProperty)['mappedBy'];
+                        $previousAlias = "$previousAlias.$mappedBy";
+
+                        $qb->select($alias)
+                            ->from($identifierResourceClass, $alias);
+                        break;
+                    } catch (MappingException $e) {
+                    }
                 default:
                     $qb->select("IDENTITY($alias.$previousAssociationProperty)")
                         ->from($identifierResourceClass, $alias);

### CODE REVIEW

File: src/Bridge/Doctrine/Orm/SubresourceDataProvider.php [hunk 1]
```diff
@@ -175,6 +176,17 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
                     $qb->select($alias)
                         ->from($identifierResourceClass, $alias);
                     break;
+                // ONE_TO_ONE relation can be Unidirectional or Bidirectional. For Bidirectional mappedBy attribute is used.
+                case ClassMetadataInfo::ONE_TO_ONE:
```
soyuka: If there is no `mappedBy` we should go through the `default` section right? Catching a `MappingException` doesn't look like a good option here.

soyuka:
Can you add a unit test on this?

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
acabala: yes, right. Alternative solution could be to check with `isset` for `$previousAssociationProperty` and `mappedBy` index, like

```
if (isset($classMetadata->associationMappings[$previousAssociationProperty]) 
    && isset($classMetadata->associationMappings[$previousAssociationProperty]['mappedBy'])) {
     $mappedBy = $classMetadata->getAssociationMapping($previousAssociationProperty)['mappedBy'];
    $previousAlias = "$previousAlias.$mappedBy";

     $qb->select($alias)
           ->from($identifierResourceClass, $alias);
     break;
}
```
which does not look clear here I think

soyuka: I'd prefer an isset test indeed. 

For example (it's the default section of the switch): 

```
default:
    $association = $classMetadata->getAssociationMapping($previousAssociationProperty);
    if (isset($association['mappedBy'])) {
      $previousAlias = "$previousAlias.$mappedBy";
      $qb->select($alias)
          ->from($identifierResourceClass, $alias);
      break;
    }

    $qb->select("IDENTITY($alias.$previousAssociationProperty)")
        ->from($identifierResourceClass, $alias);

```

acabala: Do you suggest to not create special `ONE_TO_ONE` case, but just put it into default?

soyuka: Mhhh with a `ONE_TO_ONE` case, you'd either repeat the code of the default section or use a `goto` instruction:

```
    case ONE_TO_ONE:
        $association = $classMetadata->getAssociationMapping($previousAssociationProperty);
        if (!isset($association['mappedBy'])) {
            goto def;
        }
        // ...
    default:
       def:
        $qb->select("IDENTITY($alias.$previousAssociationProperty)")
        ->from($identifierResourceClass, $alias);
```

I think that a `MANY_TO_ONE` would work the same way as the `ONE_TO_ONE` (these are both `to-one`) and it's why I suggested to put everything in the `default` but you made me doubt :p.

acabala:
@soyuka  I don't feel too confident with unit testing all those DQLs to be honest. But if you couldn't find time to handle this, I can try.

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
acabala: with `MANY_TO_ONE` there is only an unidirectional relation, and `mappedBy` is not used at all, so it shouldn't be the case.

Nightbr:
Hey guys, I have the same issue with bidirectional OneToOne and it's really blocking for me right now. Do you need any help here? What tests did you nee, I can write them if needed!

soyuka:
@Nightbr thanks for your help.

First thing would be to check if the behavior of `ONE_TO_ONE` and `MANY_TO_ONE` so that we can put the fix in the `default` section of the switch: https://github.com/api-platform/core/pull/2228#discussion_r222357320.

Then, it'd be great to add a unit test in the SubresourceDataProvider that checks that the DQL is correct for this particular relationship type (bi-directional one_to_one).

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
Nightbr: @acabala is right here, I test ClassMetadataInfo::MANY_TO_ONE (2) so it goes directly to default case and it works properly (the join is correct). The best part is when I dump some variable, the huge difference with OneToOne is that `$classMetadata->getAssociationMapping($previousAssociationProperty)['mappedBy'] `return NULL. Like @acabala said, in MANY_TO_ONE relation, the mappedBy is never used and is set to NULL by Doctrine. So we need to introduce a case ONE_TO_ONE for sure now. I think @soyuka  is rigth with using a goto to avoid code duplication. I am working on a solution and writing some unit test !

Nightbr:
@soyuka So, I tested the case of `MANY_TO_ONE` and @acabala is right about the fact `mappedBy` is never use (see https://github.com/api-platform/core/pull/2228#discussion_r227261438).

Here is my proposal (working for each case ONE_TO_ONE bidirectionnal foo/{id}/bar and bar/{id}/foo and doesn't alterate other relation behavior):

```
                case ClassMetadataInfo::ONE_TO_ONE:
                    $association = $classMetadata->getAssociationMapping($previousAssociationProperty);
                    if (!isset($association['mappedBy'])) {
                        goto def;
                    }
                    $mappedBy = $association['mappedBy'];
                    $previousAlias = "$previousAlias.$mappedBy";

                    $qb->select($alias)
                        ->from($identifierResourceClass, $alias);
                    break;
                default:
                    def:
                        $qb->select("IDENTITY($alias.$previousAssociationProperty)")
                            ->from($identifierResourceClass, $alias);
```

I'm working on some tests right now for this special case.

soyuka:
Perfect @Nightbr thanks! don't hesitate to push on this branch or to open a new PR

acabala:
Issue solved with other PR.

PR #2233, opened 2018-10-04
Allow to use container parameters as class name

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

Allows to use a config like this one:

```xml

<resources xmlns="https://api-platform.com/schema/metadata"
           xsi:schemaLocation="https://api-platform.com/schema/metadata
           https://api-platform.com/schema/metadata/metadata-2.0.xsd">
    <resource class="%sylius.model.currency.class%" />
</resources>
```

TODO:

* [x] Add tests


teohhanhui:
Shouldn't this be done at the metadata factory level?

src/Metadata/Extractor/AbstractExtractor.php
```diff
@@ -13,22 +13,27 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface as SymfonyContainerInterface;
+
 /**
  * Base file extractor.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
 abstract class AbstractExtractor implements ExtractorInterface
 {
+    private $container;
```
theofidry: shouldn't `private` come after `protected`?

src/Metadata/Extractor/AbstractExtractor.php
```diff
@@ -52,4 +57,64 @@ public function getResources(): array
...
+            }
+
+            if (preg_match('/^env\(\w+\)$/', $match[1])) {
+                throw new RuntimeException(sprintf('Using "%%%s%%" is not allowed in routing configuration.', $match[1]));
+            }
+
+            if ($this->container instanceof SymfonyContainerInterface) {
+                $resolved = $this->container->getParameter($match[1]);
+            } else {
+                $resolved = $this->container->get($match[1]);
```
theofidry: this will fail if no container is provided, maybe an empty container could be set in the constructor in this scenario

src/Metadata/Extractor/AbstractExtractor.php
```diff
@@ -13,6 +13,10 @@
 
 namespace ApiPlatform\Core\Metadata\Extractor;
 
+use Psr\Container\ContainerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface as SymfonyContainerInterface;
+use function array_key_exists;
```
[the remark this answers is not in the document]
theofidry: I think we should, it's way better than a leading backslash. That said it's better to apply a fixer for it so let's remove it here

PR #2235, opened 2018-10-04
Allow an input and an output for a given resource class

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | may fix some issues
| License       | MIT
| Doc PR        | TODO


src/EventListener/WriteListener.php
```diff
@@ -61,7 +61,8 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
                 $event->setControllerResult($persistResult ?? $controllerResult);
 
-                if (null !== $this->iriConverter) {
+                // Comparing the class is necessary because the input might not be readable
+                if (null !== $this->iriConverter && \get_class($controllerResult) === \get_class($event->getControllerResult())) {
                     $request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($controllerResult));
```
soyuka: ping @teohhanhui

meyerbaptiste:
You have to add these two new keys in the `ApiResource ` annotation.

src/Annotation/ApiResource.php
```diff
@@ -36,10 +36,12 @@
  *     @Attribute("filters", type="string[]"),
  *     @Attribute("graphql", type="array"),
  *     @Attribute("hydraContext", type="array"),
+ *     @Attribute("inputClass", type="string"),
  *     @Attribute("iri", type="string"),
  *     @Attribute("itemOperations", type="array"),
  *     @Attribute("maximumItemsPerPage", type="int"),
  *     @Attribute("normalizationContext", type="array"),
+ *     @Attribute("outputClass", type="string"),
```
meyerbaptiste: It should be after `order` (`ou` vs `or`) 😆

src/EventListener/WriteListener.php
```diff
@@ -61,7 +61,8 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
                 $event->setControllerResult($persistResult ?? $controllerResult);
 
-                if (null !== $this->iriConverter) {
+                // Comparing the class is necessary because the input might not be readable
+                if (null !== $this->iriConverter && \get_class($controllerResult) === \get_class($event->getControllerResult())) {
```
meyerbaptiste: Sorry but I'm not sure to understand...

teohhanhui: > `_api_write_item_iri` should return the IRI of the class pre-persist anyway. It just doesn't make sense if the classes aren't the same anymore.

I don't understand what's the problem. The IRI should not change.

Anyway, can't we just do:

```php
$request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($event->getControllerResult()));
```

soyuka: I wanted to do: `$controllerResult = $persistResult ?? $controllerResult` but I think that it breaks the functionality of `_api_write_item_iri` if I understand things correctly.

teohhanhui: I don't think it'd break anything? The new object returned by the data persister (if any) is what we will serialize. But the IRI should always remain the same.

soyuka: Nope, check the test named `WriteListenerTest::testOnKernelViewWithControllerResultAndPersistWithImmutableResource`

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -121,6 +121,8 @@ public function load($data, $type = null): RouteCollection
                         '_controller' => $controller,
                         '_format' => null,
                         '_api_resource_class' => $operation['resource_class'],
+                        '_api_input_resource_class' => $operation['input_class'],
```
dunglas: Just a naming thing, but I would call them `api_input_class` and `api_output_class` because they aren't resources anymore, but more "resources representations", or "dtos". I'm not really sure about this, thought.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -96,6 +96,10 @@ public function hasCacheableSupportsMethod(): bool
      */
     public function normalize($object, $format = null, array $context = [])
     {
+        if (isset($context['resource_class']) && isset($context['normalize_resource_class'])) {
```
dunglas: Why not `input/output_class` anymore? It will be easier to understand, track and debug.

ragboyjr:
@soyuka @dunglas Few things:

1. I don't see any changes to the Swagger Documentation Normalizer, so it's possible that the swagger docs could be out of sync with the actual api.
2. I implemented a similar feature here: https://github.com/krakphp/api-platform-extra#operation-resource-classes. The difference however is that mine is set per resource, and this is set on the class level. I'm not sure which one is better, but it may be a thing where both would valuable. Bringing this up for discussion though.

In my implementation, in order to keep the documentation in sync, I added a new attribute to `schema_only` which is used for defining resources that don't have any operations and are used just for being in the swagger definitions.

Let me know what you think.

ragboyjr:
Oh, and most important thing... Thanks for all the hard work you too do for this project. I'm like 4 api platform projects in and I'm loving it more and more!

src/EventListener/WriteListener.php
soyuka: @teohhanhui Can you confirm that the test `testOnKernelViewWithControllerResultAndPersistWithImmutableResource` is correct? If I  mutate `$controllerResult` it will fail and, to me, this indicates that I break things so any ideas of what I should do here?

dunglas:
Thanks Antoine!

sroze:
Not bad 😉

karser:
This looks great, but still not clear the transition between
```
!> The following isn't recommended anymore, please use Input/Output instead
/**
 * @ApiResource(
 *      collectionOperations={
 *          "post"={
 *              "path"="/users/forgot-password-request",
 *          },
 *      },
 *      itemOperations={},
 * )
 */
final class ForgotPasswordRequest
```
And
```
/**
 * @ApiResource(
 *   inputClass=Input::class,
 *   outputClass=Output::class,
 * )
 */
final class Data
```

Is input/output only resource level or it can be applied for operation as well?

Is it still possible to use Input/Data in the subscriber on POST_VALIDATE event or it works only in the data persister?

What would be the correct yaml syntax? This didn't work, at least wasn't relfected in swagger:
```
App\Dto\ForgotPasswordRequest:
    inputClass: 'App\Dto\ForgotPasswordRequest'
    outputClass: 'App\Dto\ForgotPasswordResponse'
    collectionOperations:
        post:
            path: '/users/forgot-password-request'
    itemOperations: {}
```

ragboyjr:
@karser, from what I can tell, these can be applied at the resource level or at the operation level (which freaking rocks)

ragboyjr:
It looks like we'll need to update the swagger documentation generator to take into account these changes as well.

soyuka:
> Is input/output only resource level or it can be applied for operation as well?

It should be available yes, if not lmk I'll fix it.

> Is it still possible to use Input/Data in the subscriber on POST_VALIDATE event or it works only in the data persister?

In fact, the input class is mainly used to know how the (de)normalizer should work. As the `validate` event comes before the `write` (data persister) you should definitely be able to use the input data there.

> What would be the correct yaml syntax? This didn't work, at least wasn't relfected in swagger:

I need to update the swagger documentation to reflect the changes. These are attributes first guess would be:

```
App\Dto\ForgotPasswordRequest:
    attributes:
        inputClass: 'App\Dto\ForgotPasswordRequest'
        outputClass: 'App\Dto\ForgotPasswordResponse'
```

Note that you can use either the input, the output or both.

karser:
I tried added outputClass in attributes and it still works (at least api tests don't fail) but swagger part is still not displaying the outputClass.
I see, the swagger part is not implemented https://github.com/api-platform/core/issues/2305
Now I realize that I can switch from setting response in subscriber to data persister and let api-platform serialize the result.

lyrixx:
I totally missed this feature. And I really love this feature. I will be able to remove custom code \o/ Thanks.

Just a side note not related: You really need a blog (I'm not talking about https://api-platform.com/news/). Something like Symfony "Living on the Edge
" > "New in Symfony 4.3: Console Hyperlinks" https://symfony.com/blog/new-in-symfony-4-3-console-hyperlinks and another category about releases.

dunglas:
Basically a RSS feed aggregator: https://en.wikipedia.org/wiki/Planet_(software)
2006, I feel old now 😅

PR #2238, opened 2018-10-05
[GraphQl] Add typeLoader option in schema constructor

ArnoudThibaut:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | ~
| License       | MIT
| Doc PR        | ~

In order to avoid performance issues with large schema, the Schema config allow to add a `typeLoader` option.
See : http://webonyx.github.io/graphql-php/type-system/schema/#lazy-loading-of-types

The `graphqlTypes` property has been changed into a simple one-dimensional array in order to retrieve fields by type name.

Note : Some behat tests are failing but this seems unrelated because they are failing on master too.

### DIFF
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -71,6 +71,7 @@ final class SchemaBuilder implements SchemaBuilderInterface
         $this->defaultFieldResolver = $defaultFieldResolver;
         $this->filterLocator = $filterLocator;
         $this->paginationEnabled = $paginationEnabled;
+        $this->graphqlTypes['Iterable'] = new IterableType();
     }
 
     public function getSchema(): Schema
@@ -97,6 +98,9 @@ final class SchemaBuilder implements SchemaBuilderInterface
                 'name' => 'Query',
                 'fields' => $queryFields,
             ]),
+            'typeLoader' => function ($name) {
+                return $this->graphqlTypes[$name];
+            },
         ];
 
         if ($mutationFields) {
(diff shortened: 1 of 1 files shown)


alanpoulain:
I think you shouldn't change the name of `#iterable` or `#node` in the `graphqlTypes` array because a user could have a resource with the same name (especially a `Node`).
It could be useful to even add a `#` in front of all the "internal" types (for example `"#{$shortName}Connection"`).

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -71,6 +71,7 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
         $this->defaultFieldResolver = $defaultFieldResolver;
         $this->filterLocator = $filterLocator;
         $this->paginationEnabled = $paginationEnabled;
+        $this->graphqlTypes['Iterable'] = new IterableType();
```
alanpoulain: I would move this line to the `getSchema` method.

ArnoudThibaut:
> I think you shouldn't change the name of `#iterable` or `#node` in the `graphqlTypes` array because a user could have a resource with the same name (especially a `Node`).
> It could be useful to even add a `#` in front of all the "internal" types (for example `"#{$shortName}Connection"`).

I tried to create a new class with name `Node` on master branch to see how the current implementation will react and I have the following error :
`Schema must contain unique named types but contains multiple types named \"Node\" (see http:\/\/webonyx.github.io\/graphql-php\/type-system\/#type-registry).`

And I think the problem will be the same with the type loader because when it will be asked for type `Node`, i see no way to know if it should return the internal type or the user defined one.

WDYT ?

alanpoulain:
You're right. It's because graphql-php uses the `name` parameter in the type. Maybe we could change it later. Could you fix the other issue?

alanpoulain:
Since it's only related to performance, why not targeting 2.3?

alanpoulain:
@dunglas why do you think it can break something? It's very internal and tests are here to make sure we didn't break anything.

dunglas:
Collisions wasn’t possible before. They are now.

Anyway performance improvement are usually not considered as bugs fix (at least in Symfony).
This not an emergency, it can wait the next minor.

ArnoudThibaut:
What are you thinking about when you say that collisions wasn't possible before and now it's ?

dunglas:
If you have a type named `Iterable` or `Node` or `*Connection`. Now it will break.

alanpoulain:
It's already the case in 2.3 as explained by @ArnoudThibaut.

dunglas:
Ok got it.

dunglas:
Thank you very much @ArnoudThibaut

Issue #2199, opened 2018-09-12
Add a test for Swagger array parameters


soyuka:
see https://github.com/api-platform/core/pull/2114#issuecomment-420355558

PR #2241, opened 2018-10-06
Add Swagger DocumentationNormalizer test for array collection filters

Pandoraaa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2199
| License       | MIT
| Doc PR        | 

There's a merged PR https://github.com/api-platform/core/pull/2114 but this part wasn't tested yet, so I added the test in the case of an array collection filters in the Swagger documentation.

### DIFF
--- a/tests/Swagger/Serializer/DocumentationNormalizerTest.php
+++ b/tests/Swagger/Serializer/DocumentationNormalizerTest.php
@@ -1229,6 +1229,13 @@ class DocumentationNormalizerTest extends TestCase
                 'required' => false,
                 'strategy' => 'partial',
             ]]),
+            'f3' => new DummyFilter(['toto' => [
+                'property' => 'name',
+                'type' => 'array',
+                'is_collection' => true,
+                'required' => true,
+                'strategy' => 'exact',
+            ]]),
         ];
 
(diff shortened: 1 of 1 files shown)


tests/Swagger/Serializer/DocumentationNormalizerTest.php
```diff
@@ -1229,14 +1229,21 @@ public function testFilters()
...
+            ]]),
         ];
 
         foreach ($filters as $filterId => $filter) {
             $filterLocatorProphecy->has($filterId)->willReturn(true)->shouldBeCalled();
             $filterLocatorProphecy->get($filterId)->willReturn($filter)->shouldBeCalled();
         }
 
-        $filterLocatorProphecy->has('f3')->willReturn(false)->shouldBeCalled();
+        $filterLocatorProphecy->has('f4')->willReturn(false)->shouldBeCalled();
```
teohhanhui: Not a fault of this PR, but `->shouldBeCalled()` should only be used when we actually want to check that a method is called. If the intention is only to mock the return value of the method, `->willReturn(...)` should suffice.

Issue #2231, opened 2018-10-03
[GraphQL] ExistsFilter and SearchFilter can not be defined on the same property


epourail:
When a resource is defined with an `ExistsFilter` and a `SearchFilter` on the same property, the graphql schema will convert both filters in the same way using the property name to identify the filter.

In the following example
```
/**
 * @ApiResource()
 * @ApiFilter(ExistsFilter::class, properties={"title"})
 * @ApiFilter(SearchFilter::class, properties={"title": "exact"})
 */
class Dummy
{
   /** @var null|string $title a nullable property */
   private $title;

   // ...
}
```
The REST collection endpoint enables both filter with the syntax `/dummies?title=shirt` and `/dummies?title[exists]=true`

But the GraphQL endpoint enables only the `SearchFilter` with the syntax `dummies(title: "shirt") { ... }`
(as the `ExistFilter` uses the syntax `dummies(title: {exists: true})`)

epourail:
Below I share my solution/workaround to enable the `ExistsFilter` on the graphql endpoint.

```
<?php

declare(strict_types=1);

namespace App\Filter;

use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\ExistsFilter as ParentExistsFilter;
use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
use Doctrine\ORM\QueryBuilder;

/**
 * Class EmptyFilter
 * @package App\Filter
 *
 */
final class ExistsFilter extends ParentExistsFilter
{
    /**
     * @inheritdoc
     */
    protected function filterProperty(
        string $property,
        $value,
        QueryBuilder $queryBuilder,
        QueryNameGeneratorInterface $queryNameGenerator,
        string $resourceClass,
        string $operationName = null
    ) {
        if ($property !== 'notempty') {
            return;
        }

        foreach ($value as $propertyName => $exists) {
            parent::filterProperty(
                $propertyName,
                ['exists' => $exists],
                $queryBuilder,
                $queryNameGenerator,
                $resourceClass,
                $operationName
            );
        }
    }

    /**
     * @inheritdoc
     */
    public function getDescription(string $resourceClass): array
    {
        $parentDescription = parent::getDescription($resourceClass);

        $description = [];

        array_walk(
            $parentDescription,
            function ($item, $key) use (&$description) {
                $item['swagger'] = [
                    'description' => 'Search elements which the property is empty (or not) : ' . $item['property'],
[...]

PR #2243, opened 2018-10-07
Improve the ExistsFilter behavior (to cohabit with the SearchFilter)

epourail:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (for GraphQL)
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #2231
| License       | MIT
| Doc PR        | api-platform/docs#613

Improve the behavior of the `ExistsFilter` filter to be declared according to the following syntax:
- on a REST collection endpoints: `/dummies?exists[<property>]=true/false`
- on the GraphQL endpoint: `dummies(exists: {<property>: "true/false"}) { ... }`

The name of the query parameter is controlled via the bundle configuration:
```yaml
api_platform:
  collection:
    exists_parameter_name: 'not_null'
``` 

It allows to use an `ExistsFilter` and a `SearchFilter` on the same property via the GraphQL endpoint.
 
- [x] Improve ExistsFilter
- [x] PHPUnit
- [x] Behat
- [x] Documentation

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
@@ -17,9 +17,13 @@ use ApiPlatform\Core\Bridge\Doctrine\Common\Filter\ExistsFilterInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Common\Filter\ExistsFilterTrait;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryBuilderHelper;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
  * Filters the collection by whether a property value exists or not.
@@ -28,7 +32,7 @@ use Doctrine\ORM\QueryBuilder;
  * the value is not one of ( "true" | "false" | "1" | "0" ) the property is ignored.
  *
  * A query parameter with key but no value is treated as `true`, e.g.:
- * Request: GET /products?brand[exists]
+ * Request: GET /products?exists[brand]
  * Interpretation: filter products which have a brand
  *
  * @author Teoh Han Hui <teohhanhui@gmail.com>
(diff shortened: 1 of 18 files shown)


alanpoulain:
Why `notnull_parameter_name` and not `exists_parameter_name` (I would call it `exists_filter_parameter_name` but I think you chose this to be consistent with the order filter, didn't you?).

alanpoulain:
You need to make sure the previous way of using the exists filter is still working, but deprecate it.

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -30,11 +34,24 @@
  * Request: GET /products?brand[exists]
  * Interpretation: filter products which have a brand
  *
- * @author Teoh Han Hui <teohhanhui@gmail.com>
+ * @author Julien Verger <julien.verger@gmail.com>
```
alanpoulain: Please add you after the previous author.

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -30,11 +34,24 @@
...
  */
 class ExistsFilter extends AbstractContextAwareFilter
 {
-    const QUERY_PARAMETER_KEY = 'exists';
+    /**
+     * @var string Keyword used to retrieve the value
+     */
+    protected $existsParameterName;
```
alanpoulain: private (this class should be final but is not).

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -30,11 +34,24 @@
...
+
+    /**
+     * @param RequestStack|null $requestStack No prefix to prevent autowiring of this deprecated property
+     */
+    public function __construct(ManagerRegistry $managerRegistry, $requestStack = null, string $existsParameterName = 'exists', LoggerInterface $logger = null, array $properties = null)
```
alanpoulain: Add `$existsParameterName` after `$logger` to prevent BC.

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -190,4 +226,15 @@ private function isAssociationNullable(array $associationMapping): bool
 
         return true;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function extractProperties(Request $request/*, string $resourceClass*/): array
```
alanpoulain: I think you don't need to add this since this filter has not been used in this version yet.

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -66,24 +83,43 @@ public function getDescription(string $resourceClass): array
     /**
      * {@inheritdoc}
      */
-    protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
+    public function apply(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null, array $context = [])
+    {
+        if (isset($context['filters']) && !isset($context['filters'][$this->existsParameterName])) {
```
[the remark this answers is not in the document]
antograssiot: the first isset is needed @dunglas if I remember correctly, that how we determine if it should be forwarded to the parent class (used before the usage of the AbstractContextAwareFilter); there is the same code in the OrderFilter (see https://github.com/api-platform/core/pull/1866)

epourail:
> Why `notnull_parameter_name` and not `exists_parameter_name` (I would call it `exists_filter_parameter_name` but I think you chose this to be consistent with the order filter, didn't you?).

@alan Yes I keep the same logic (behavior and naming) as the order filter. 
I decided to name it `notnull_parameter_name` because, at the end, it is a `IS NOTNULL`/`IS NULL` check (and keep the default value `exists` to reference the previous logic)

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
```diff
@@ -31,11 +35,27 @@
  * Interpretation: filter products which have a brand
  *
  * @author Teoh Han Hui <teohhanhui@gmail.com>
+ * @author Julien Verger <julien.verger@gmail.com>
  */
-class ExistsFilter extends AbstractContextAwareFilter
+final class ExistsFilter extends AbstractContextAwareFilter
```
antograssiot: you can only add a `@final` annotation, this is a BC break

src/Bridge/Doctrine/Orm/Filter/ExistsFilter.php
epourail: done.

alanpoulain:
@epourail I am planning to finish your PR. I am working on a PR and it takes a lot of time. That's why I can't right now. If anyone is willing to take it before me go ahead.

alanpoulain:
This PR has been rebased, tests are fixed.
I have chosen to rename the parameter `exists_parameter_name` to not lose our users (even if the comment of @epourail is true).

alanpoulain:
Thank you @epourail! Nice one.

teohhanhui:
We've come full circle then! When I did the `ExistsFilter`, it was using this syntax at first lol...

PR #2247, opened 2018-10-09
Arachnophobia mode

Deuchnord:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | WTFPL
| Doc PR        | api-platform/doc#612

One of the greatest bugs in the Human software is their fear of spiders. In some cases, the simple seeing of something big go down the screen makes them feel bad for some seconds, before realizing it's just Webby.

Problem is, the Human software is closed-source, so this PR tries to manage that issue by providing a way to disable displaying Webby on the Swagger UI pages. Of course, Webby remains visible by default.

To disable Webby's display, just add this to API Platform's config:
```yaml
# /config/api_platform.yaml
api_platform:
    show_webby: false # defaults to true
```

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
+++ b/src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
@@ -133,14 +133,9 @@ window.onload = () => {
 };
 
 function manageWebbyDisplay() {
-    const webbyClass = document.getElementsByClassName('webby');
+    const webby = document.getElementsByClassName('webby')[0];
+    if(!webby) return;
 
-    // First, check if Webby is present
-    if (webbyClass.length === 0) {
-        return;
-    }
-
-    const webby = webbyClass[0];
     const web = document.getElementsByClassName('web')[0];
     webby.classList.add('calm');
     web.classList.add('calm');
--- a/src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
+++ b/src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
@@ -1,21 +1,26 @@
 'use strict';
 
 window.onload = () => {
-    const webby = document.getElementsByClassName('webby')[0];
-    const web = document.getElementsByClassName('web')[0];
-    webby.classList.add('calm');
-    web.classList.add('calm');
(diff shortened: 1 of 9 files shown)


src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
```diff
@@ -1,21 +1,26 @@
...
-        web.classList.replace('calm', 'frighten');
-        setTimeout(() => {
-            webby.classList.replace('frighten', 'calm');
-            web.classList.replace('frighten', 'calm');
-        }, 10000);
-    });
+    const webbyClass = document.getElementsByClassName('webby');
+
+    // First, check if Webby is present
+    if(webbyClass.length > 0) {
```
alanpoulain: Prefer an early return in a function instead.

src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
```diff
@@ -145,3 +131,28 @@ window.onload = () => {
         }
     }
 };
+
+function manageWebbyDisplay() {
+    const webbyClass = document.getElementsByClassName('webby');
```
dunglas: Can be simplified:

```js
const webby = document.getElementsByClassName('webby')[0];
if (!webby) return;
```

src/Bridge/Symfony/Bundle/Resources/public/init-swagger-ui.js
```diff
@@ -133,14 +133,9 @@ window.onload = () => {
 };
 
 function manageWebbyDisplay() {
-    const webbyClass = document.getElementsByClassName('webby');
+    const webby = document.getElementsByClassName('webby')[0];
+    if(!webby) return;
```
dunglas: missing space between `if` and `(`

alanpoulain:
Can't we merge this one?

dunglas:
Thanks @Deuchnord! First PR, but not the last!

jeremyFreeAgent:
Thank you very much!!! <3

Issue #2245, opened 2018-10-08
Allow to set the sunset HTTP header


dunglas:
API Platform could provide a developer-friendly way to set [the Sunset HTTP header](https://tools.ietf.org/html/draft-wilde-sunset-header-06) on deprecated endpoints to indicate when this endpoint will not be available anymore. The syntax like could be like the following:

```php
/**
 * @ApiResource(
 *   deprecationReason="This endpoint is replaced by Bar", <- already supported
 *   sunset="2019/01/01" <- any string supported by DateTime
 * )
 */
class Foo
{
}
```

norkunas:
#2185 similar

norkunas:
Would this support the url change without replacing the resource?

soyuka:
> Would this support the url change without replacing the resource?

Like setting the Location header + 301 status?

norkunas:
If it makes sense in api.

blanky0230:
I'd like to give this a go If you don't mind ? ;)

soyuka:
> If it makes sense in api. But my primary reason is that it should generate routes for deprecated and new route and understand the IRI's from the old and new

To transform an IRI to a Resource we use the RouteCollection, not sure how we could introduce some kind of change there to transform a route to another if it has been deprecated.

PR #2249, opened 2018-10-10
WIP: Begin Introduction of new 'sunset' API-Resource annotation in order t…

blanky0230:
…o support: sunset header
I'm not quite sure this works as excpected :) May someone please take a look?

I'm also rahter new to behat. I suppose I'd need to create some Behat Tests for this as well.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Related to | #2245 
| License       | MIT
| Doc PR        | //TODO!

### DIFF
--- a/tests/Sunset/AddSunsetHeaderListenerTest.php
+++ b/tests/Sunset/AddSunsetHeaderListenerTest.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Sunset;
+
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Sunset\EventListener\AddSunsetHeaderListener;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+
+class AddSunsetHeaderListenerTest extends TestCase
+{
+    public function testSetSunsetHeader()
+    {
+        $request = new Request([], [], ['_api_resource_class' => Dummy::class, '_api_item_operation_name' => 'get']);
+        $request->setMethod('GET');
+        $response = new Response();
+        $date = new \DateTimeImmutable('tomorrow');
+
+        $eventProphecy = $this->prophesize(FilterResponseEvent::class);
+        $eventProphecy->getRequest()->willReturn($request)->shouldBeCalled();
+        $eventProphecy->getResponse()->willReturn($response)->shouldBeCalled();
+        $event = $eventProphecy->reveal();
+
+        $resourceMetadata = new ResourceMetadata(null, null, null, null,
+            null, ['sunset' => $date->format(DATE_RFC1123)]);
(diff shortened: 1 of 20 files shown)


src/Annotation/ApiResource.php
```diff
@@ -52,7 +52,8 @@
  *     @Attribute("shortName", type="string"),
  *     @Attribute("subresourceOperations", type="array"),
  *     @Attribute("swaggerContext", type="array"),
- *     @Attribute("validationGroups", type="mixed")
+ *     @Attribute("validationGroups", type="mixed"),
+ *     @Attribute("sunset", type="string")
```
dunglas: Can you move it after `subresourceOperations`? (alpha order)

src/Metadata/Resource/ResourceMetadata.php
```diff
@@ -41,6 +42,7 @@ public function __construct(string $shortName = null, string $description = null
         $this->subresourceOperations = $subresourceOperations;
         $this->graphql = $graphql;
         $this->attributes = $attributes;
+        $this->sunset = $sunset;
```
dunglas: This is not necessary IMO, an attribute would be better suited.

src/Sunset/EventListener/AddSunsetHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+/**
```
dunglas: To be removed

src/Sunset/EventListener/AddSunsetHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace ApiPlatform\Core\Sunset\EventListener;
+
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+
+class AddSunsetHeaderListener
+{
+    private $sunset;
+
+    public function __construct(string $sunset = null)
```
dunglas: The listener should be stateless. You should detect the current resource using `RequestAttributesExtractor` and call `ResourceMetadataFactory` instead.

tests/Sunset/AddSunsetHeaderListenerTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+        $response = new Response();
+
+        $event = $this->prophesize(FilterResponseEvent::class);
+        $event->getResponse()->willReturn($response)->shouldBeCalled();
+
+        $listener = new AddSunsetHeaderListener('2018-10-10');
+        $listener->onKernelResponse($event->reveal());
+
+        $this->assertEquals('2018-10-10', $response->headers->get('Sunset'));
```
dunglas: The data must follow the HTTP date format: https://tools.ietf.org/id/draft-wilde-sunset-header-03.html#sunset-header. You should call `\DateTimeImmutable($sunsetAttributeValue)` then call `format` with the appropriate value (see https://stackoverflow.com/questions/21120882/the-date-time-format-used-in-http-headers).

blanky0230: Thanks! I had some trouble figuring out what format to use and how to enforce it!

src/Annotation/ApiResource.php
blanky0230: should now be the case

src/Metadata/Resource/ResourceMetadata.php
blanky0230: removed

src/Sunset/EventListener/AddSunsetHeaderListener.php
blanky0230: Gotta set up my IDE :/ Sorry.

src/Sunset/EventListener/AddSunsetHeaderListener.php
blanky0230: Made some changes. I am still not 100% confident this is how it should be done though. Do you think there should also be some sort of enforcing the Date format?

PR #2250, opened 2018-10-11
fix deserialize relation bug

videni:
at the deserializeRelation method if the resource is managed by api-platform and it is writable, it should be array.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


dunglas:
Thanks for your contrib! Can you add a non-regression test please?

flug:
@videni Up is this pull request still valid?

soyuka:
Let's close this for now, feel free to open a new one if the bug persists

PR #2252, opened 2018-10-12
Begin Introduction of new 'sunset' API-Resource annotation in order t…

blanky0230:
I messed up with rebasing on the old PR :/

I'm sorry if this was kinda spammy. (git rebase is a bit new to me).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Related to | #2245 
| License       | MIT
| Doc PR        | //TODO!

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -51,6 +51,7 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
  *     @Attribute("routePrefix", type="string"),
  *     @Attribute("shortName", type="string"),
  *     @Attribute("subresourceOperations", type="array"),
+ *     @Attribute("sunset", type="string"),
  *     @Attribute("swaggerContext", type="array"),
  *     @Attribute("validationGroups", type="mixed")
  * )
@@ -247,6 +248,11 @@ final class ApiResource
      * @var mixed
      */
     private $validationGroups;
+
+    /**
+     * @var string
+     */
+    private $sunset;
 
     /**
      * @throws InvalidArgumentException
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -173,6 +173,7 @@
         </service>
 
         <service id="api_platform.listener.view.respond" class="ApiPlatform\Core\EventListener\RespondListener">
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
             <tag name="kernel.event_listener" event="kernel.view" method="onKernelView" priority="8" />
         </service>
 
--- a/src/EventListener/RespondListener.php
+++ b/src/EventListener/RespondListener.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\EventListener;
 
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
 
@@ -27,6 +29,13 @@ final class RespondListener
         'POST' => Response::HTTP_CREATED,
         'DELETE' => Response::HTTP_NO_CONTENT,
     ];
+
+    private $resourceMetadataFactory;
+
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
+    {
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+    }
 
     /**
      * Creates a Response to send to the client according to the requested format.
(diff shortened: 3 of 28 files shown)


src/Annotation/ApiResource.php
```diff
@@ -51,8 +51,9 @@
  *     @Attribute("routePrefix", type="string"),
  *     @Attribute("shortName", type="string"),
  *     @Attribute("subresourceOperations", type="array"),
+ *     @Attribute("sunset", type="string"),
  *     @Attribute("swaggerContext", type="array"),
- *     @Attribute("validationGroups", type="mixed")
+ *     @Attribute("validationGroups", type="mixed"),
```
dunglas: Comma to remove (the Doctrine parser will fail on this IIRC)

src/Sunset/EventListener/AddSunsetHeaderListener.php
```diff
@@ -0,0 +1,45 @@
...
+        $attributes = RequestAttributesExtractor::extractAttributes($request);
+
+        if (!$attributes) {
+            return;
+        }
+
+        $response = $event->getResponse();
+        $metadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
+
+        if (null !== $metadata->getAttribute('sunset')) {
```
dunglas: You should use the following:

```php
if ($sunset = $metadata->getOperationAttribute($attributes, 'sunset', null, true)) {
  $response->headers->set('Sunset', (new \DateTimeImmutable($sunset)->format(\DateTime::RFC1123));
}
```

It will allow to define sunset headers only for some operations, or to define different dates for different operations.

Also, the trick with `DateTime` allows the user to set the date in any format (even computations like "2019/01/01 + 1 month" will work), but to always output the valid format.

blanky0230: Wow. That is actually very smart! Thanks!

src/Annotation/ApiResource.php
blanky0230: Should be solved now.

src/Sunset/EventListener/AddSunsetHeaderListener.php
blanky0230: Done

src/Sunset/EventListener/AddSunsetHeaderListener.php
```diff
@@ -0,0 +1,45 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Sunset\EventListener;
+
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+
+class AddSunsetHeaderListener
```
dunglas: Can you make it `final` please? (sorry I forgot about that before)

src/Sunset/EventListener/AddSunsetHeaderListener.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Sunset\EventListener;
```
antograssiot: Just wondering why not directly in Core\Eventlistener and I faill to see where is the listener declared (src/Bridge/Symfony/Bundle/Resources/config/api.xml) ?

dunglas: Good catch. Should be in core, and of course registered, thank @antograssiot

src/Sunset/EventListener/AddSunsetHeaderListener.php
blanky0230: Seems like a very good idea. Feels way cleaner :) I moved it to Eventlistener/RespondListener, could you please take a look at the test, I'm not 100% certain I did it right.

src/Sunset/EventListener/AddSunsetHeaderListener.php
blanky0230: So i moved it to EventListener\RespondListner::onKernelView(), was that what you had in mind @antograssiot ?

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -173,6 +173,7 @@
         </service>
 
         <service id="api_platform.listener.view.respond" class="ApiPlatform\Core\EventListener\RespondListener">
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
```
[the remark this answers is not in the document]
dunglas: Not for internal ones, here we are sure that this service exist. I would keep it as is.

src/EventListener/RespondListener.php
```diff
@@ -55,6 +55,11 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             }
         }
 
+        if ($request->attributes->has('sunset')) {
```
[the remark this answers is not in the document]
blanky0230: I pulled your pr in :)

dunglas:
Thanks! Great improvement.
Don't forget the docs PR please :)

PR #2260, opened 2018-10-19
Use context builder for DocumentationNormalizer and SerializerPropertyMetadataFactory

desmax:
While setting up react admin I noticed that dynamic context building is not available for `/docs.jsonld` endpoint. This is an attempt to fix this.
If this code is something worth looking at then I would update tests.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | https://github.com/api-platform/api-platform/issues/866
| License       | MIT

### DIFF
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -26,6 +26,7 @@ use ApiPlatform\Core\Metadata\Property\SubresourceMetadata;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use ApiPlatform\Core\Operation\Factory\SubresourceOperationFactoryInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
 use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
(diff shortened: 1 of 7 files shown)


src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -49,15 +50,17 @@
     private $urlGenerator;
     private $subresourceOperationFactory;
     private $nameConverter;
+    private $contextBuilder;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceClassResolverInterface $resourceClassResolver, OperationMethodResolverInterface $operationMethodResolver, UrlGeneratorInterface $urlGenerator, SubresourceOperationFactoryInterface $subresourceOperationFactory = null, NameConverterInterface $nameConverter = null)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceClassResolverInterface $resourceClassResolver, OperationMethodResolverInterface $operationMethodResolver, UrlGeneratorInterface $urlGenerator, SerializerContextBuilderInterface $contextBuilder, SubresourceOperationFactoryInterface $subresourceOperationFactory = null, NameConverterInterface $nameConverter = null)
```
soyuka: To avoid a bc break the new argument should be:
- at the latest position
- nullable

see https://symfony.com/doc/current/contributing/code/bc.html

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -29,12 +30,14 @@
     private $resourceMetadataFactory;
     private $serializerClassMetadataFactory;
     private $decorated;
+    private $contextBuilder;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerClassMetadataFactoryInterface $serializerClassMetadataFactory, PropertyMetadataFactoryInterface $decorated)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerClassMetadataFactoryInterface $serializerClassMetadataFactory, PropertyMetadataFactoryInterface $decorated, SerializerContextBuilderInterface $contextBuilder)
```
soyuka: same should be nullable

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -50,7 +53,7 @@ public function create(string $resourceClass, string $property, array $options =
             $resourceClass = $childResourceClass;
         }
 
-        list($normalizationGroups, $denormalizationGroups) = $this->getEffectiveSerializerGroups($options, $resourceClass);
+        list($normalizationGroups, $denormalizationGroups) = $this->contextBuilder->createFromResourceClass($options, $resourceClass);
```
soyuka: We should deprecate the `$this->getEffectiveSerializerGroups` and propose to inject the contextbuilder instead

desmax:
In the end I decided that decorating api_platform.metadata.property.metadata_factory and api_platform.metadata.resource.metadata_factory is a better idea.

PR #2262, opened 2018-10-20
Add data providers and data persisters matches to the debug panel

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Even if I do like the data providers system, I generally find hard to debug them and end in dumping in vendor chain data providers.

This PR aims to provide a more developer friendly way to debug them by improving the Api Platform panel in Symfony debug toolbar
debugbar

Up to now I've implemented item/collection/subresources data provider and data persister
If you like the idea, I'll add tests before merging this.


antograssiot:
Add tabs to ease navigation and here is a screenshot, easier to see than the gif

capture d ecran 2018-10-20 a 21 50 54

dunglas:
Very good idea!

src/Bridge/Symfony/Bundle/Resources/config/data_provider_debug.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" ?>
```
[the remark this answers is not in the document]
antograssiot: I wanted to activate those listener only if the api platform profiler is enable **and** debug is on but I can simplify and remove extra files if you prefer, let me know.

antograssiot:
I've updated following your comments @dunglas, except the "extra" loops and the separates files waiting for more comments after my answers. thanks for your review

antograssiot:
everything has been updated here, just waiting for github webhooks to be back on the scene I guess now

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public function supports($data): bool
+    {
+        return $this->decorated->supports($data);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function persist($data)
```
dunglas: The logic in this method and in `remove` looks duplicated. Can't you extract it in a private method, and just pass the name of the decorated method to call (`persist` or `remove`) as a parameter?

src/Bridge/Symfony/Bundle/DataProvider/TraceableChainCollectionDataProvider.php
```diff
@@ -0,0 +1,75 @@
...
+        return $this->context;
+    }
+
+    public function getCollection(string $resourceClass, string $operationName = null, array $context = [])
+    {
+        $this->context = $context;
+        $results = null;
+        $match = false;
+
+        foreach ($this->dataProviders as $dataProvider) {
```
dunglas: Same here, but because it used in different classes, maybe this logic should be moved in an  `@internal` `abstract` class, or in an trait. WDYT?

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
antograssiot: Done

antograssiot: I didn't needed to pass the method name as parameter as the return is different in both case, is it ok like that ?

src/Bridge/Symfony/Bundle/DataProvider/TraceableChainCollectionDataProvider.php
antograssiot: I'll need to think more about it as the logic is not really the same in the 3 providers and severeal methods doesn't use the same args number. Not sure that the result will be more readable nor maintainable. I'll check this afternoon if I can get something better. Any ideas are welcome

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
```diff
@@ -0,0 +1,82 @@
...
+     * {@inheritdoc}
+     */
+    public function remove($data)
+    {
+        if ($match = $this->tracePersisters($data)) {
+            return $match->remove($data);
+        }
+    }
+
+    private function tracePersisters($data)
```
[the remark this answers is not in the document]
dunglas: Looks good to me (especially for a private one)

antograssiot:
branched rebased. 
A generic way to trace Extension would be great too I think, maybe something like a WrappedExtension  class that would decorate every extension so we could check if they where call (in custom data providers too). Then all wrapped extension would be collected and displayed in a new tab. WDYT ?

dunglas:
Thank you very much @antograssiot, very nice improvement

PR #2267, opened 2018-10-23
Bump PHPStan version

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

and fix a mistake in a readme

Instead of adding a new exclude rule, I would rather drop [this code](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Util/QueryJoinParser.php#L103-L111) and in other places in the QueryJoinParser. Those are public method that didn't moved since 6 years and are still there in the next 3.0, I fail to understand what it is trying to fix and why if the method is to be removed at some point, we expect the internal property to still be there.
let me know if it is ok

### DIFF
--- a/src/Bridge/Doctrine/Orm/Util/QueryJoinParser.php
+++ b/src/Bridge/Doctrine/Orm/Util/QueryJoinParser.php
@@ -100,17 +100,7 @@ final class QueryJoinParser
      */
     public static function getJoinRelationship(Join $join): string
     {
-        static $relationshipProperty = null;
-        static $initialized = false;
-
-        if (!$initialized && !method_exists(Join::class, 'getJoin')) {
-            $relationshipProperty = new \ReflectionProperty(Join::class, '_join');
-            $relationshipProperty->setAccessible(true);
-
-            $initialized = true;
-        }
-
-        return (null === $relationshipProperty) ? $join->getJoin() : $relationshipProperty->getValue($join);
+        return $join->getJoin();
     }
 
     /**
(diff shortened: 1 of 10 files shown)


dunglas:
Let's wait for @soyuka review regarding the Doctrine part.

soyuka:
We should ask @Simperfit instead it's his code and I'm also not sure to what it fixes :laughing:

antograssiot:
at least if we need to keep it I'll update the tests because even it is tested, dropping this code doesn't make the test suite failing...

soyuka:
I don't think that these are useful and they just look like copy/paste to me

antograssiot:
ok this was needed as long as we supported doctrine/orm 2.2 so basically means api-platform 1.x
i'll drop it and the associated tests was were not very reliable anyway.

src/Bridge/Doctrine/Orm/Util/QueryJoinParser.php
```diff
@@ -100,35 +100,15 @@ public static function getClassMetadataFromJoinAlias(string $alias, QueryBuilder
      */
     public static function getJoinRelationship(Join $join): string
```
antograssiot: I'll update the code to use those inner functions directly in the codebase in a future PR as they add no real benefits now

soyuka: cool ok

dunglas:
Thanks @antograssiot!

antograssiot:
@dunglas @soyuka I've a working branch based on 2.3 passing PHPStan level 6, is it something you're interested in or would it make community contributions  harder?
https://github.com/api-platform/core/compare/2.3...antograssiot:phpstan-level6

dunglas:
It's something we're definitely interested in!

teohhanhui:
@soyuka:
> We should ask @Simperfit instead it's his code and I'm also not sure to what it fixes :laughing:

Are you sure it was not my code? :stuck_out_tongue_closed_eyes:

antograssiot:
@teohhanhui the PR targeted 2.3 (which is stable to me) and will land into master when 2.3 will get merged inside or did I miss something ?

teohhanhui:
Oops... Sorry, I must have misread. :see_no_evil:

PR #2271, opened 2018-10-23
Deprecate dead code in QueryJoinParser and remove internal usage

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Follows https://github.com/api-platform/core/pull/2267

### DIFF
--- a/tests/Bridge/Doctrine/Orm/Util/QueryJoinParserTest.php
+++ b/tests/Bridge/Doctrine/Orm/Util/QueryJoinParserTest.php
@@ -43,30 +43,50 @@ class QueryJoinParserTest extends TestCase
         $this->assertEquals($metadata, $classMetadata->reveal());
     }
 
+    /**
+     * @group legacy
+     * @expectedDeprecation The use of "ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryJoinParser::getJoinRelationship()" is deprecated since 2.3 and will be removed in 3.0. Use "Doctrine\ORM\Query\Expr\Join::getJoin()" directly instead.
+     */
     public function testGetJoinRelationshipWithJoin()
     {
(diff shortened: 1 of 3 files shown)


tests/Bridge/Doctrine/Orm/Util/QueryJoinParserTest.php
```diff
@@ -43,30 +43,50 @@ public function testGetClassMetadataFromJoinAlias()
         $this->assertEquals($metadata, $classMetadata->reveal());
     }
 
+    /**
+     * @group legacy
+     * @expectedDeprecation The use of "ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryJoinParser::getJoinRelationship()" is deprecated since 2.3 and will be removed in 3.0. Use "Doctrine\ORM\Query\Expr\Join::getJoin()" directly instead.
```
teohhanhui: Actually I'm not sure why we add double quotes where they're unnecessary? (And it makes the message harder to read...)

antograssiot: I can remove them later, when I'll get a computer, it is just a copy/paste from an other error in the codebase somewhere though :stuck_out_tongue:

antograssiot: I've checked and this is also how it is done in symfony, so I guess we should stick to this convention...?

dunglas:
Thanks! Good one!

tests/Bridge/Doctrine/Orm/Util/QueryJoinParserTest.php
teohhanhui: Oh, okay...

PR #2272, opened 2018-10-24
Bump PHPStan analysis to level 6

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

ref https://github.com/api-platform/core/pull/2267#issuecomment-432523409

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/DateFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/DateFilter.php
@@ -94,7 +94,7 @@ class DateFilter extends AbstractContextAwareFilter
         }
 
         $nullManagement = $this->properties[$property] ?? null;
-        $type = $this->getDoctrineFieldType($property, $resourceClass);
+        $type = (string) $this->getDoctrineFieldType($property, $resourceClass);
 
         if (self::EXCLUDE_NULL === $nullManagement) {
             $queryBuilder->andWhere($queryBuilder->expr()->isNotNull(sprintf('%s.%s', $alias, $field)));
@@ -160,6 +160,7 @@ class DateFilter extends AbstractContextAwareFilter
      */
     protected function addWhere(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, string $operator, string $value, string $nullManagement = null, $type = null)
     {
+        $type = (string) $type;
         try {
             $value = false === strpos($type, '_immutable') ? new \DateTime($value) : new \DateTimeImmutable($value);
         } catch (\Exception $e) {
@@ -206,7 +207,7 @@ class DateFilter extends AbstractContextAwareFilter
      */
     protected function isDateField(string $property, string $resourceClass): bool
     {
-        return isset(self::DOCTRINE_DATE_TYPES[$this->getDoctrineFieldType($property, $resourceClass)]);
+        return isset(self::DOCTRINE_DATE_TYPES[(string) $this->getDoctrineFieldType($property, $resourceClass)]);
     }
 
     /**
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -191,7 +191,7 @@ final class SubresourceDataProvider implements SubresourceDataProviderInterface
         foreach ($normalizedIdentifiers as $key => $value) {
             $placeholder = $queryNameGenerator->generateParameterName($key);
             $qb->andWhere("$alias.$key = :$placeholder");
-            $topQueryBuilder->setParameter($placeholder, $value, $classMetadata->getTypeOfField($key));
+            $topQueryBuilder->setParameter($placeholder, $value, (string) $classMetadata->getTypeOfField($key));
         }
 
         // Recurse queries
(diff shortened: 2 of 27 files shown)


src/Bridge/Doctrine/Orm/Filter/DateFilter.php
```diff
@@ -94,7 +94,7 @@ protected function filterProperty(string $property, $values, QueryBuilder $query
         }
 
         $nullManagement = $this->properties[$property] ?? null;
-        $type = $this->getDoctrineFieldType($property, $resourceClass);
+        $type = (string) $this->getDoctrineFieldType($property, $resourceClass);
```
soyuka: would be great if we could directly return `string` in the AbstractFilter method instead

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -191,7 +191,7 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
         foreach ($normalizedIdentifiers as $key => $value) {
             $placeholder = $queryNameGenerator->generateParameterName($key);
             $qb->andWhere("$alias.$key = :$placeholder");
-            $topQueryBuilder->setParameter($placeholder, $value, $classMetadata->getTypeOfField($key));
+            $topQueryBuilder->setParameter($placeholder, $value, (string)   $classMetadata->getTypeOfField($key));
```
soyuka:
```suggestion
            $topQueryBuilder->setParameter($placeholder, $value, (string) $classMetadata->getTypeOfField($key));
```

src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
```diff
@@ -68,13 +68,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-            file_put_contents($input->getOption('output'), $content);
+        if (!empty($filename = $input->getOption('output')) && \is_string($filename)) {
+            file_put_contents($filename, $content);
             $output->writeln(
-                sprintf('Data written to %s', $input->getOption('output'))
+                sprintf('Data written to %s', $filename)
             );
         } else {
-            $output->writeln($content);
+            $output->writeln((string) $content);
```
soyuka: why do we have to cast here? Maybe that we should cast above in the variable declaration?

src/Filter/QueryParameterValidateListener.php
```diff
@@ -94,11 +94,11 @@ private function isRequiredFilterValid(string $name, Request $request): bool
         if (\is_array($matches[$rootName])) {
             $keyName = array_keys($matches[$rootName])[0];
 
-            $queryParameter = $request->query->get($rootName);
+            $queryParameter = $request->query->get((string) $rootName);
```
soyuka: same can't we cast above?

src/Bridge/Doctrine/Orm/Filter/DateFilter.php
antograssiot: done, thanks

src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
antograssiot: that was because of the `json_encode` that couls return false but I've coreted in a "cleaner" way I think

src/Bridge/Doctrine/Orm/Filter/AbstractFilter.php
```diff
@@ -308,13 +308,14 @@ protected function addJoinsForNestedProperty(string $property, string $rootAlias
...
+     * @return string|null
      */
     protected function getDoctrineFieldType(string $property, string $resourceClass)
     {
         $propertyParts = $this->splitPropertyParts($property, $resourceClass);
         $metadata = $this->getNestedMetadata($resourceClass, $propertyParts['associations']);
+        $type = $metadata->getTypeOfField($propertyParts['field']);
 
-        return $metadata->getTypeOfField($propertyParts['field']);
+        return null !== $type ? (string) $type : null;
```
antograssiot: I wonder if this is to be consider as a BC break though...

soyuka: We're changing the return type so I guess it is, just wondering if we use `Type` at all, if not I would say that it's no big deal.

antograssiot: We don't use Type but could be used by some custom filter maybe... In my own filters I always add to cast it so PhpStan is not compaining, so I too think it is not a big deal. Fixing this potential BC break will be easy enough also but will "cost" a release so I wanted to draw your attention on it.

teohhanhui: I think we should keep BC here. The calling code can do the cast.

teohhanhui: Also important to note: https://github.com/doctrine/dbal/blob/1b1adc2a8d4ab8dcfe5c328ccd3864a0fd36f6a7/lib/Doctrine/DBAL/Types/Type.php#L252

```php
    /**
     * @deprecated Relying on string representation is discouraged and will be removed in DBAL 3.0.
```

antograssiot: I'll revert then

soyuka: Thanks @teohhanhui !

antograssiot: ok it has been reverted.

antograssiot:
I've solved the merge conflicts and rebased, anything left to be done here ?
ping @teohhanhui

dunglas:
Thanks @antograssiot!

PR #2279, opened 2018-10-25
[EagerLoadingExtension] Remove unset attributes key in normalization context

ArnoudThibaut:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ~
| License       | MIT
| Doc PR        | ~

When an entity had more that one relation, if the first relation was not present in the attributes array in the normalization context the whole array was removed so they was no test for other relations on the next loop iteration.

The aim of this pull request is to fix this behavior and to let the chance to other relations to be eager loaded.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -167,8 +167,6 @@ final class EagerLoadingExtension implements ContextAwareQueryCollectionExtensio
                 if ($inAttributes = isset($normalizationContext[AbstractNormalizer::ATTRIBUTES][$association])) {
                     // prepare the child context
                     $normalizationContext[AbstractNormalizer::ATTRIBUTES] = $normalizationContext[AbstractNormalizer::ATTRIBUTES][$association];
-                } else {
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -167,8 +167,6 @@ private function joinRelations(QueryBuilder $queryBuilder, QueryNameGeneratorInt
                 if ($inAttributes = isset($normalizationContext[AbstractNormalizer::ATTRIBUTES][$association])) {
                     // prepare the child context
                     $normalizationContext[AbstractNormalizer::ATTRIBUTES] = $normalizationContext[AbstractNormalizer::ATTRIBUTES][$association];
-                } else {
-                    unset($normalizationContext[AbstractNormalizer::ATTRIBUTES]);
```
soyuka: @dunglas I think you added this but it does not look useful to me, I'd be :+1: for this patch, any reason we should keep this?

dunglas:
Thanks Thibaut!

PR #2282, opened 2018-10-26
Automatically dispatch updates to clients using the Mercure protocol

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

This PR adds support for [the Mercure Protocol](https://mercure.rocks) to API Platform.
It introduces a simple new attribute called `mercure`:

```php
<?php

namespace App\Entity;

/**
 * @ApiResource(mercure=true)
 */
class Book
{
    // ...
}
```

When set, every time an object of this type is created, updates or deleted through the API, the new version is sent to all connected clients through the Mercure hub

Checkout this YouTube video to see how it works with the API Platform [client generator](https://github.com/api-platform/client-generator): http://www.youtube.com/watch?v=JdlAcfTJsOI

The code of I used for this video is available here: https://github.com/api-platform/demo/pull/66

It's also possible to send the update to users having some `targets` only:

```php
<?php

namespace App\Entity;

/**
 * @ApiResource(mercure={"a-group", "another-group"})
 */
class Book
{
    // ...
}
```

Finally, it's possible to execute an expression (using the Expression Language component), to retrieve a dynamic list of targets:

```php
<?php

namespace App\Entity;

/**
 * @ApiResource(mercure="object.owners")
 */
class Book
{
    public owners = [];
   // ...
}
```

To enable Mercure support, you must install the brand new [Symfony Mercure Bundle](https://github.com/symfony/mercure-bundle), that comes with the [Symfony Mercure Component](https://github.com/symfony/mercure).

Take a look to [my Forum PHP's slide deck](https://dunglas.fr/2018/10/http-2-speed-up-your-apps-and-dispatch-real-time-updates-symfony-and-api-platforms-features-announcement/) for further information about Mercure and its support in Symfony and API Platform.


src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public function onFlush(OnFlushEventArgs $eventArgs)
+    {
+        $uow = $eventArgs->getEntityManager()->getUnitOfWork();
+
+        foreach ($uow->getScheduledEntityInsertions() as $entity) {
+            $this->storeEntityToPublish($entity, 'createdEntities');
+        }
+
+        foreach ($uow->getScheduledEntityUpdates() as $entity) {
```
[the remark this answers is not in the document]
dunglas: Hum, good question. Is it really necessary because we don't do that in https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php either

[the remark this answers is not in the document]
antograssiot: I'll wrote a test it and let you kown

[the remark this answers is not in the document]
dunglas: Commit it please! Extra tests are always good to have!

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+            $this->storeEntityToPublish($entity, 'updatedEntities');
+        }
+
+        foreach ($uow->getScheduledEntityDeletions() as $entity) {
+            $this->storeEntityToPublish($entity, 'deletedEntities');
+        }
+    }
+
+    /**
+     * Purges tags collected during this request, and clears the tag list.
```
bendavies: comment is wrong

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public function postFlush()
```
bendavies: well, i've mentioned this before, but i hate that this is not done in kernel.terminate instead. I don't see why this should delay the clients API response or possibly result in a 500.

bendavies: I see you can use messenger now, which i imagine can can flush on `kernel.terminate`

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+    private $messageBus;
+    private $publisher;
+    private $expressionLanguage;
+    private $createdEntities;
+    private $updatedEntities;
+    private $deletedEntities;
+
+    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, MessageBusInterface $messageBus = null, callable $publisher = null, ExpressionLanguage $expressionLanguage = null)
+    {
+        if (null === $messageBus && null === $publisher) {
```
bendavies: possibly ensure only 1 is provided?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+                $this->publishUpdate($iri, $targets);
+            }
+        } finally {
+            $this->reset();
+        }
+    }
+
+    private function reset(): void
+    {
+        $this->createdEntities = new \SplObjectStorage();
```
bendavies: is this the same as `removeAll`?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        $value = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('mercure', false);
+        if (false === $value) {
+            return;
+        }
+
+        if (\is_string($value)) {
+            if (null === $this->expressionLanguage) {
+                throw new RuntimeException('The Expression Language component is not installed. Try running "composer require symfony/expression-language".');
```
bendavies: is this the proper place to throw this, or can it be done in the metadata factory somehow?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,175 @@
...
+
+            $value = $this->expressionLanguage->evaluate($value, ['object' => $entity]);
+        }
+
+        if (true === $value) {
+            $value = [];
+        }
+
+        if (!\is_array($value)) {
+            throw new InvalidArgumentException(sprintf('The value of the "mercure" attribute of the "%s" resource class must be a boolean, an array of targets or a valid expression, "%s" given.', $resourceClass, \gettype($value)));
```
bendavies: as above, proper place for this?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
dunglas: `removeAll` takes a mandatory argument as parameter.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
dunglas: We currently don't handle metadata validation in the factory, but it would definitely be a better place

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
dunglas: No idea :) I would keep it as is for now because it's "safer", but if your suggestion improve GC or something similar, let me know and I'll update!

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
dunglas: Yes. I'll keep it as is because it out of this PR's scope, and I'll try to add a proper validator (for all existing attributes) in a subsequent PR. Is it ok for you?

bendavies: of course! sorry, my `?!` wasn't meant to come off as rude or aggressive, but enthusiastic!

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_http_cache_purger.xml
```diff
@@ -15,7 +15,7 @@
 
             <tag name="doctrine.event_listener" event="preUpdate" />
             <tag name="doctrine.event_listener" event="onFlush" />
-            <tag name="doctrine.event_listener" event="postFlush" />
+            <tag name="kernel.event_listener" event="kernel.terminate" method="postFlush" />
```
bendavies: oh, wow ok. has your thinking changed since https://github.com/api-platform/core/pull/1286?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
dunglas: I understood it like that no worries :)

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_http_cache_purger.xml
dunglas: Hum, you're right, I'll revert this change as it's not related to this PR. (I forgot about this discussion)

bendavies: but the discussion point is similar, i guess.

dunglas: Not really, to me it's a big deal if a cache is not flushed because you'll serve stale data. Not dispatching a Mercure update is less a problem, because it breaks nothing, and it's explicitly stated in the Mercure protocol that the client MUST call the original API if it wants to be sure to have fresh data. So we can use `kernel.terminate` safely for Mercure, I'm still unsure for Varnish.

bendavies: nicely argued.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -145,7 +145,10 @@ private function storeEntityToPublish($entity, string $property): void
         }
 
         if ('deletedEntities' === $property) {
-            $this->deletedEntities[$this->iriConverter->getIriFromItem($entity, UrlGeneratorInterface::ABS_URL)] = $value;
+            $this->deletedEntities[(object) [
```
soyuka: can we transform this to a real class? We had issues in the past with StdClass and it's better for debugging.

dunglas: Does it worth it? It's purely internal state.

dunglas: It was in the serialization context, and indeed an anonymous class. `\stdClass` is serializable: https://3v4l.org/2VWST

PR #2283, opened 2018-10-27
iri based on value from persister

wesol05:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

When using custom data persister, we don't pass id, but we want to somehow retrieve it. Unfortunately, returned object (with set id, is not taking under consideration when iri is set).

Example data persister: 
```
public function persist($data)
    {
        $core = $this->transformer->toCore($data);

        $this->entityManager->persist($core);
        $this->entityManager->flush();

        return $this->transformer->toHttp($core);
    }
```

### DIFF
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -59,7 +59,8 @@ final class WriteListener
                     @trigger_error(sprintf('Returning void from %s::persist() is deprecated since API Platform 2.3 and will not be supported in API Platform 3, an object should always be returned.', DataPersisterInterface::class), E_USER_DEPRECATED);
                 }
 
-                $event->setControllerResult($persistResult ?? $controllerResult);
+                $controllerResult = $persistResult ?? $controllerResult;
+                $event->setControllerResult($controllerResult);
 
                 if (null !== $this->iriConverter) {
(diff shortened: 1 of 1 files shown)


src/EventListener/WriteListener.php
```diff
@@ -59,7 +59,8 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
                     @trigger_error(sprintf('Returning void from %s::persist() is deprecated since API Platform 2.3 and will not be supported in API Platform 3, an object should always be returned.', DataPersisterInterface::class), E_USER_DEPRECATED);
                 }
 
-                $event->setControllerResult($persistResult ?? $controllerResult);
+                $controllerResult = $persistResult ?? $controllerResult;
+                $event->setControllerResult($controllerResult);
```
soyuka: This change has a open discussion here: https://github.com/api-platform/core/pull/2235#discussion_r223208885 and I believe it breaks a feature (see WriteListenerTest::testOnKernelViewWithControllerResultAndPersistWithImmutableResource, the resource isn't immutable anymore?) ?

flug:
@wesol05 Up is this pull request still valid?

Issue #2265, opened 2018-10-22
access_control, expression language instance


Toflar:
Can anybody elaborate on why the `DenyAccessListener` uses its own expression language instance and not the one of Symfony Security?

We're using https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Resources/config/security_expression_language.xml#L7 instead of Symfony's `security.expression_language`.

I don't see why this is done. In Symfony I can easily add new expression languages by tagging the providers as `security.expression_language_provider` which of course do not end up in our `DenyAccessListener`.
Why would we want two different instances? Both are related to security, aren't they?

Toflar:
@dunglas @soyuka @Simperfit can anyone just quickly tell me if there's any reason for it to be standalone and not the framework `security.expression_language`. I'll work on a PR then so we can get this done :)

soyuka:
I really don't know unless it's to not be dependent of another symfony-tight library.

Toflar:
I've created a PR in https://github.com/api-platform/core/pull/2289, I guess that illustrates the issue/benefits a bit better :)

PR #2289, opened 2018-10-29
[WIP/RFC] Use framework expression language service instead of custom one

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2265
| License       | MIT
| Doc PR        | 

Using the framework `security.expression_language` service rather than our own one allows us to extend the expression language with our own functions easily. Symfony already provides an easy way to register custom functions in `AddExpressionLanguageProvidersPass`. ApiPlatform currently does not. But I figured instead of providing our own compiler pass we can just use the built-in `security.expression_language`, extend that with our `is_granted()` function and thus at the same time also allow developers to add even more functions. Wdyt?

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -38,7 +38,6 @@ use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
-use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\Messenger\MessageBusInterface;
@@ -120,9 +119,6 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
         $bundles = $container->getParameter('kernel.bundles');
         if (isset($bundles['SecurityBundle'])) {
-            if (class_exists(ExpressionLanguage::class)) {
-                $loader->load('security_expression_language.xml');
-            }
             $loader->load('security.xml');
         }
 
--- a/src/Bridge/Symfony/Bundle/Resources/config/security.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/security.xml
@@ -5,6 +5,8 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
+        <service id="api_platform.security.expression_language" alias="security.expression_language" />
+
         <service id="api_platform.security.resource_access_checker" class="ApiPlatform\Core\Security\ResourceAccessChecker" public="false">
             <argument type="service" id="api_platform.security.expression_language" on-invalid="null" />
             <argument type="service" id="security.authentication.trust_resolver" on-invalid="null" />
@@ -20,6 +22,10 @@
 
             <!-- This listener must be executed only when the current object is available -->
             <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="1" />
+        </service>
+
+        <service id="api_platform.security.expression_language_provider" class="ApiPlatform\Core\Security\Core\Authorization\ExpressionLanguageProvider" public="false">
+            <tag name="security.expression_language_provider" />
         </service>
     </services>
 
--- a/src/Security/Core/Authorization/ExpressionLanguageProvider.php
+++ b/src/Security/Core/Authorization/ExpressionLanguageProvider.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Security\Core\Authorization;
+
+use Symfony\Component\ExpressionLanguage\ExpressionFunction;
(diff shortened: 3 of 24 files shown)


src/Bridge/Symfony/Bundle/Resources/config/security.xml
```diff
@@ -6,7 +6,7 @@
 
     <services>
         <service id="api_platform.security.resource_access_checker" class="ApiPlatform\Core\Security\ResourceAccessChecker" public="false">
-            <argument type="service" id="api_platform.security.expression_language" on-invalid="null" />
+            <argument type="service" id="security.expression_language" on-invalid="null" />
```
dunglas: Can you create an alias `api_platform.security.expression_language` please? It will be useful for people wanting to inject a custom service.

src/Bridge/Symfony/Bundle/Resources/config/security.xml
```diff
@@ -21,6 +21,10 @@
             <!-- This listener must be executed only when the current object is available -->
             <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="1" />
         </service>
+
+        <service id="api_platform.security.expression_language_provider" class="ApiPlatform\Core\Security\Core\Authorization\ExpressionLanguageProvider">
```
dunglas:
```suggestion
        <service id="api_platform.security.expression_language_provider" class="ApiPlatform\Core\Security\Core\Authorization\ExpressionLanguageProvider" public="false">
```

src/Security/Core/Authorization/ExpressionLanguageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Security\Core\Authorization;
+
+use Symfony\Component\ExpressionLanguage\ExpressionFunction;
+use Symfony\Component\ExpressionLanguage\ExpressionFunctionProviderInterface;
+
+/**
+ * Adding ApiPlatform specific Expression Language functions.
```
dunglas:
```suggestion
 * Registers API Platform's Language functions.
```

src/Security/Core/Authorization/ExpressionLanguageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class ExpressionLanguageProvider implements ExpressionFunctionProviderInterface
```
dunglas:
```suggestion
final class ExpressionLanguageProvider implements ExpressionFunctionProviderInterface
```

src/Security/Core/Authorization/ExpressionLanguageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function getFunctions()
```
dunglas:
```suggestion
    public function getFunctions(): array
```

src/Security/Core/Authorization/ExpressionLanguageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        return [
+            new ExpressionFunction('is_granted', function ($attributes, $object = 'null') {
```
dunglas: Wouldn't it conflict with the one registered by SensioFrameworkExtraBundle?

src/Security/ExpressionLanguage.php
```diff
@@ -1,38 +0,0 @@
-<?php
```
dunglas: Should be kept and deprecated for BC.

src/Security/Core/Authorization/ExpressionLanguageProvider.php
Toflar: One would just override the other (depending on the order they are called). And as far as I could see the idea was to provide the same logic, right? I mean, we could only register it if the bundle is not loaded if that makes you more comfortable? Or just tell the users to require the framework-extra-bundle if they want this function?

Toflar:
Comments addressed :)

src/Security/Core/Authorization/ExpressionLanguageProvider.php
dunglas: I just want to be sure it works as intended in all cases :)

Toflar: It should :)

dunglas:
@Toflar errors look related to this PR, could you take a look (I'll tag a new release soon). Thanks

Toflar:
Done!

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -117,9 +116,6 @@ public function load(array $configs, ContainerBuilder $container)
 
         $bundles = $container->getParameter('kernel.bundles');
         if (isset($bundles['SecurityBundle'])) {
-            if (class_exists(ExpressionLanguage::class)) {
-                $loader->load('security_expression_language.xml');
```
teohhanhui: Actually, `symfony/expression-language` is in `require-dev` of `symfony/security-bundle`: https://github.com/symfony/symfony/blob/b5c6892911fc7454a76528cb0d4d1af4336c7413/src/Symfony/Bundle/SecurityBundle/composer.json#L42

So we still need to check that the `ExpressionLanguage` component is actually available.

Toflar: Are you sure? `security.expression_language` would not be present so the alias would become invalid and as we use `on-invalid="null"`  on `api_platform.security.resource_access_checker` that should work. Haven't checked though.

soyuka:
Can you rebase? Going to merge it tomorrow!

Toflar:
Rebased onto master, not sure if I should've gone for the 2.4 branch? The `deps=low` test fails but apparently that's also the case for `master`.

soyuka:
Master is fine, yeah don't worry about deps=low I'll try to fix it later, thanks!

alanpoulain:
@soyuka the issue with `deps=low` comes from Composer. It will be resolved in the next version or if you do `composer selfupdate --snapshot`.

PR #2286, opened 2018-10-28
Force Yaml::dump to dump empty array as actual empty array

Cethy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Fixed tickets | #2285 
| License       | MIT
| Doc PR        | api-platform/doc#1234

Issue #2285, opened 2018-10-28
Make SwaggerCommand yml compatible w/ openapi-generator


Cethy:
As is, using `openapi-generator` after `bin/console api:swagger:export -y -o swagger.yaml` result in a failure because the empty arrays are converted as empty objects (and `openapi-generator` does not like that).

Adding the `Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE` mask to `Yaml::dump` fixes the problem :
```php
Yaml::dump($data, 6, 4, Yaml::DUMP_OBJECT_AS_MAP | Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE);
```

PR #2291, opened 2018-10-30
Add a non regression test for #2285 #2286

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Edit: Adding test decrease coverage... of course 😈

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,7 +183,7 @@ jobs:
             mkdir -p build/logs/tmp build/cov
             for f in $(find features -name '*.feature' -not -path 'features/main/exposed_state.feature' | circleci tests split --split-by=timings); do
               _f=${f//\//_}
-              FEATURE="${_f}" phpdbg -qrr vendor/bin/behat --profile=coverage --suite=default --tags=~@postgres --format=progress --out=std --format=junit --out=build/logs/tmp/"${_f}" "$f"
+              FEATURE="${_f}" phpdbg -qrr vendor/bin/behat --profile=coverage --suite=default --format=progress --out=std --format=junit --out=build/logs/tmp/"${_f}" "$f"
             done
       - run:
           name: Merge Behat test reports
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -183,7 +183,7 @@ jobs:
             mkdir -p build/logs/tmp build/cov
             for f in $(find features -name '*.feature' -not -path 'features/main/exposed_state.feature' | circleci tests split --split-by=timings); do
               _f=${f//\//_}
-              FEATURE="${_f}" phpdbg -qrr vendor/bin/behat --profile=coverage --suite=default --tags=~@postgres --format=progress --out=std --format=junit --out=build/logs/tmp/"${_f}" "$f"
```
dunglas: Why is this change needed?

dunglas: (It's probably this change that decreases the coverage)

.travis.yml
```diff
@@ -52,4 +52,4 @@ script:
       vendor/bin/behat --suite=default --format=progress;
     fi
   - tests/Fixtures/app/console api:swagger:export > swagger.json && npx swagger-cli validate swagger.json && rm swagger.json
-  - tests/Fixtures/app/console api:swagger:export --yaml > swagger.yaml && npx swagger-cli validate --no-schema swagger.yaml && rm swagger.yaml
```
dunglas: Why is this option needed?

.circleci/config.yml
antograssiot: The coverage was already decreasing before this change.
I just saw that it was useless as the tags are defined in the [behat.yml.dist](https://github.com/api-platform/core/blob/master/behat.yml.dist#L61)
that's basically what I did a while ago, see https://github.com/api-platform/core/commit/71e37cbd17b17f21d36c9e52a143ecef0db9f8f5

I can revert if you prefer

dunglas:
Thanks @antograssiot

PR #2294, opened 2018-10-30
Prefix root resource's route_prefix to sub-resources

silverbackdan:
I think this is a bug, sorry if it isn't. This will make a sub-resource have the same route_prefix as the root resource. Is this the expected behaviour?

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| Doc PR        | n/a

### DIFF
--- a/src/Operation/Factory/SubresourceOperationFactory.php
+++ b/src/Operation/Factory/SubresourceOperationFactory.php
@@ -127,8 +127,14 @@ final class SubresourceOperationFactory implements SubresourceOperationFactoryIn
                     $operation['operation_name']
                 );
 
+                $prefix = trim(trim($rootResourceMetadata->getAttribute('route_prefix', '')), '/');
+                if ('' !== $prefix) {
+                    $prefix .= '/';
+                }
+
                 $operation['path'] = $subresourceOperation['path'] ?? sprintf(
-                    '/%s/{id}/%s%s',
+                    '/%s%s/{id}/%s%s',
+                    $prefix,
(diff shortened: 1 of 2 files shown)


src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -127,8 +127,14 @@ private function computeSubresourceOperations(string $resourceClass, array &$tre
                     $operation['operation_name']
                 );
 
+                $prefix = trim(trim($rootResourceMetadata->getAttribute('route_prefix', '')), '/');
```
soyuka: we don't need the first `trim` imo

silverbackdan: This is lifted straight from here https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Routing/ApiLoader.php#L204 - do you think I should adjust both or is there another reason why you think both aren't needed. Also would you mind clarifying if you mean the inner or outer trim when you say 'first'?

antograssiot: Better keep both, but I think that the inner one is less usefull.
Pretty sure that some users use `"route_prefix"="/foo"` or `"route_prefix"="foo/"` and setting `"route_prefix"="/"`  would be strange but would not have impact at least.

silverbackdan:
Thanks both @soyuka and @antograssiot 
This will save a lot of extra config for me 😄

dunglas:
Thanks @silverbackdan!

PR #2298, opened 2018-10-31
bugfix: throw InvalidArgument on matching something that looks like an url

Simperfit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | 

If you want to reproduce here is a little insight on how to do it: 

Let's say you have this resource:

```yaml
MyNameSpace\Duration:
    shortName: 'Duration'
    itemOperations:
        get:
            method: 'GET'
            path: "/duration"
    collectionOperations:
        post:
            method: 'POST'
            path: "/duration"
            controller: 'MyDomaine/DurationAPIController'
            denormalization_context:
                dateinterval_format: 'P%dDT%hH%iM'
            normalization_context:
                dateinterval_format: 'P%dDT%hH%iM'

    properties:
        id:
            identifier: true
```

The duration is a VO that contains 3 fields.
The controller just use a librarie to calculate a start date + an interval to get an end date.

so we are sending this: 
```json
{"startDate":"28-01-2018 10:10","interval":"P0DT0H167M","business":true}
```
API-Platform tries to do `getItemFromIri` on each field that has a string, so startDate and interval, 
when the startDate is set to `28-01-2018 10:00` it does work properly and returns:

```json
{
    "interval": "P0DT0H167M",
    "id": "180128",
    "startDate": "2018-01-28T10:10:00+01:00",
    "endDate": "2018-01-29T11:47:00+01:00",
    "business": true
}
```

When adding 10 minutes to it, it does break with an `SuspiciousOperationException` since it does looks like an host from Symfony point of view.

### DIFF
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -476,6 +476,7 @@ Feature: Create-Retrieve-Update-Delete
     {
       "@id": "/dummies/1",
       "name": "A nice dummy",
+      "dummyDate": "2018-12-01 13:12",
       "jsonData": [{
           "key": "value1"
         },
@@ -498,7 +499,7 @@ Feature: Create-Retrieve-Update-Delete
       "description": null,
       "dummy": null,
       "dummyBoolean": null,
-      "dummyDate": "2015-03-01T10:00:00+00:00",
+      "dummyDate": "2018-12-01T13:12:00+00:00",
       "dummyFloat": null,
       "dummyPrice": null,
       "relatedDummy": null,
@@ -538,7 +539,7 @@ Feature: Create-Retrieve-Update-Delete
       "description": null,
       "dummy": null,
       "dummyBoolean": null,
-      "dummyDate": "2015-03-01T10:00:00+00:00",
+      "dummyDate": "2018-12-01T13:12:00+00:00",
       "dummyFloat": null,
       "dummyPrice": null,
       "relatedDummy": null,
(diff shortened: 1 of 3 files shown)


GregoireHebert:
IMO, this is more a patch than a real bugfix.  
If I understand correctly, there is two things going on. 
The first one is, using a date, valid for an IRI, but not valid as an [URI](https://tools.ietf.org/html/rfc3986#section-3.3) since it contains a colon.
But the reel problem I see here, is it should not rely on the date since, there is a identifier defined, that is correct.

soyuka:
> API-Platform tries to do getItemFromIri on each field that has a string, so startDate and interval,
when the startDate is set to 28-01-2018 10:00 it does work properly and returns:

Wait what? No it does so only on identifiers. I'm really not sure to understand the issue.

GregoireHebert:
Me neither... I tried to reproduce this, and everything went perfectly.
I think @Simperfit just broke it's project ^^'

meyerbaptiste:
Thanks to #2497, I was able to reproduce the problem. This fix works fine.

soyuka:
thanks @meyerbaptiste @Simperfit 

@kate-kate this should be fixed, I'll see with @dunglas if we backport this to 2.3 but not sure yet

dunglas:
This doesn't look to be a proper to the underlying problem to me:

1. `getItemFromIri` shouldn't be called on every string, it's slow and a non-sense. We should use our metadata to check if the call should be done or not
2. `IriConverter` must not be aware of HttpFoundation (there is probably a bug in Symfony, the router should catch this exception)

I'm in favor of reverting this patch.

soyuka:
@dunglas agreed but I thought this was critical enough to be merged. 

> `getItemFromIri` shouldn't be called on every string, it's slow and a non-sense. We should use our metadata to check if the call should be done or not

The normalizer apparently thinks that this is a relation (because it's an object) and tries to denormalize it from what I understood.

dunglas:
I think we should not merge workaround for Symfony bugs (better fix SF)

dunglas:
As pointed (privately) by @meyerbaptiste, it's not the fault of Symfony.
The catch should be moved here: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Routing/Router.php#L83

PR #2300, opened 2018-11-01
Basic rate limiting functionality

toby-griffiths:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not yet
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | api-platform/doc#637

I've started working on a rate limiter of the API & thought I'd share.  It's not quite finished yet, but I'd appreciate feedback whilst I complete the work to add request specific rate limits, and maybe a central cache storage engine.


dunglas:
Hi @toby-griffiths and thanks for your contrib. Are you aware of https://github.com/IndraGunawan/api-rate-limit-bundle? Do you think there are any room for collaboration with @IndraGunawan here?

toby-griffiths:
@dunglas No, I wasn't aware.  I see that that has a few more features, but looks like it would not handle race conditions so well.  I'll have a chat with @IndraGunawan about updating it.
I'm assuming you'd rather have this as a separate bundle, rather than integrating into the core?

SCIF:
Hi @toby-griffiths , could you get some update on this PR?

toby-griffiths:
@SCIF After @dunglas 's comment, I decided it's probably better to support the existing bundle, but I've got to find time to [update that bundle with handling of race conditions](https://github.com/IndraGunawan/api-rate-limit-bundle/issues/10)

antograssiot:
Thanks @toby-griffiths 
> I decided it's probably better to support the existing bundle

I think it is better too indeed
Closing then

norkunas:
Maybe this could be reconsidered? As it's not possible to configure rate limits per operation on resource without custom workarounds. Also that Symfony has a RateLimiter component it'd be not that hard to integrate this to core and also properly expose headers to api documentation.

PR #2302, opened 2018-11-02
[WIP] Add support for OpenAPI 3

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | not yet

This PR aims to provide a way to export in a OpenAPI 3.0 compatible format.

This is still kind of a work in progress but I wanted to get early feedback on this.
Still tests and CS should be ok and good enough.
To ensure that the new normalizer is at least as tested as the v2, I've used the same tests and updated the expectations.

Here is some topics I'd like to discuss:
 
1. I made the choice to create a separate documentation normalizer, here are the main reasons: ease v2 removal maybe at some point, do no interfere with Api Gateway normalizer, allow to be marked as experimental. I'm inclined to change this...

2. This solution doesn't "transform" a swagger 2 compliant code to a version 3 compliant code. This implies that in the current situation, any overridden swagger context might not end up valid. Is it a big issue ?

Thank you before end for your time reviewing and discussing this.

TODO:
- [ ] Check that the security parts of the generated files is ok
- [ ] Add [Links](https://swagger.io/docs/specification/links/) support
- [x] use query param to select version


src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
```diff
@@ -65,9 +70,15 @@ protected function configure()
...
+        $version = $input->getOption('spec-version');
+        if (!\in_array($version, ['2', '3'], true)) {
+            throw new InvalidOptionException(sprintf('This tool only support version 2 and 3 of the OpenAPI specification ("%s" given).', $version));
+        }
+        $documentationNormalizer = sprintf('v%ddocumentationNormalizer', $version);
         $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $this->apiTitle, $this->apiDescription, $this->apiVersion, $this->apiFormats);
-        $data = $this->documentationNormalizer->normalize($documentation);
-        $content = $input->getOption('yaml') ? Yaml::dump($data, 6, 4, Yaml::DUMP_OBJECT_AS_MAP | Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE) : (json_encode($data, JSON_PRETTY_PRINT) ?: '');
+        $data = $this->$documentationNormalizer->normalize($documentation);
+        $content = $input->getOption('yaml') ? Yaml::dump($data, 10, 2, Yaml::DUMP_OBJECT_AS_MAP | Yaml::DUMP_EMPTY_ARRAY_AS_SEQUENCE) : (json_encode($data, JSON_PRETTY_PRINT) ?: '');
```
antograssiot: I've increased the depth before inlining and reduced the indentation to 2 spaces as the latest version is even more nested

src/Serializer/AbstractDocumentationNormalizer.php
```diff
@@ -0,0 +1,389 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Serializer;
```
antograssiot: Not sure if it belongs in this namespace

src/OpenApi/Serializer/DocumentationNormalizer.php
```diff
@@ -0,0 +1,443 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\OpenApi\Serializer;
```
antograssiot: I added a namespace but they could all live in `ApiPlatform\Core\Swagger` if you prefer and be named V2 and V3 with an alias to the old DocumentationNormalizer classname to avoid BC or something like that

src/Serializer/AbstractDocumentationNormalizer.php
```diff
@@ -0,0 +1,389 @@
...
+ * Common features regarding Documentation normalization.
+ *
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ * @author Teoh Han Hui <teohhanhui@gmail.com>
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ * @author Anthony GRASSIOT <antograssiot@free.fr>
+ *
+ * @internal
+ */
+abstract class AbstractDocumentationNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface
```
antograssiot: all of the content of this file has been extracted from the current DocumentNormalizer, nothing new here, only variable/function visibility update from private to protected when necessary

src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
```diff
@@ -54,9 +57,11 @@ public function __construct(NormalizerInterface $documentationNormalizer, Resour
     protected function configure()
     {
         $this
-            ->setName('api:swagger:export')
-            ->setDescription('Dump the Swagger 2.0 (OpenAPI) documentation')
+            ->setName('api:openapi:export')
```
antograssiot: I did this change as I remembered I saw it in a ticket somewhere quite a long time ago.

antograssiot:
https://app.swaggerhub.com/apis/antograssiot/api-platform-fixtures/0.0.0
is the result

dunglas:
@antograssiot do you think we can benefit from [Links](https://swagger.io/docs/specification/links/) in this PR? It looks like a great new feature for API Platform.

src/Bridge/Symfony/Bundle/Resources/config/swagger.xml
```diff
@@ -38,14 +38,41 @@
...
+            <argument>%api_platform.oauth.scopes%</argument>
+            <argument>%api_platform.swagger.api_keys%</argument>
+            <argument type="service" id="api_platform.subresource_operation_factory" />
+            <argument>%api_platform.collection.pagination.enabled%</argument>
+            <argument>%api_platform.collection.pagination.page_parameter_name%</argument>
+            <argument>%api_platform.collection.pagination.client_items_per_page%</argument>
+            <argument>%api_platform.collection.pagination.items_per_page_parameter_name%</argument>
+            <argument>%api_platform.collection.pagination.client_enabled%</argument>
+            <argument>%api_platform.collection.pagination.enabled_parameter_name%</argument>
+            <tag name="serializer.normalizer" priority="18" />
```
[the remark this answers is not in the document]
dunglas: What do you think about allowing to choose the OpenAPI version through a query parameter? As done for API Gateway here: https://github.com/api-platform/core/blob/master/src/Documentation/Action/DocumentationAction.php#L66

Then it will be easy to test, and convenient for end users (they will be able to use OpenAPI v3 by default when supported, but still fallback to v2 for old tools).

For Swagger UI, we should to the same thing IMO (defaulting to v3, but allowing to switch to v2 by setting a query parameter).

WDYT?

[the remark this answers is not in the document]
antograssiot: looks like a good idea, I'll try to update soon

antograssiot:
Yeah adding Link seems doable, I'll try to work on that

antograssiot:
@dunglas, I've ben working a bit more on this, not pushed yet and also waiting feedback on #2317 as it will be needed to... but I have a question on how should be handle the custom documentation.
Right now the swagger doc can be overridden using the `swagger_context`. As currently the OpenAPI normalizer doesn't transform a version 2 into a version 3, should I add the possibility to use a `openapi_context` attribute so people can override both documentation, should we keep using `swagger_context` only and then it's up to the developer to choose which version he want to support ?

dunglas:
Introducing an `openapi_context` looks indeed better to me

antograssiot:
@dunglas So I've rebased and put this branch up to date (implemented #2335 and #2344), I think that every tests should be passing, I'll re-read the whole PR tomorrow, I didn't took time to do this today but it should be ready for another set of review as discussed.

PR #2307, opened 2018-11-03
[Bugfix] Remove type Error in the Debug Toolbar

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        | 

Fixes the panel display. Truly sorry about that I probably forgot to push this change after improving the context display by using `profiler_dump()`, actually causing the issue 😞 

capture d ecran 2018-11-03 a 15 38 58

PR #2314, opened 2018-11-05
Add tests on the debug bar/profiler view

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

The idea is to prevent me or someone else to break the debug bar again 😅 
I would totally understand if this is not merged though
ref #2307


tests/Metadata/Property/Factory/AnnotationPropertyNameCollectionFactoryTest.php
```diff
@@ -49,7 +49,7 @@ public function testCreate(PropertyNameCollectionFactoryInterface $decorated = n
         $this->assertEquals($results, iterator_to_array($metadata));
     }
 
-    public function getDependencies()
+    public function dependenciesProvider()
```
antograssiot: this change is to avoid conflicting with a [parent class method added in PHPUnit 7.2](https://github.com/sebastianbergmann/phpunit/blob/7.2/src/Framework/TestCase.php#L962)
_method signature weren't compatible_

composer.json
```diff
@@ -42,30 +42,32 @@
         "php-mock/php-mock-phpunit": "^2.0",
         "phpdocumentor/reflection-docblock": "^3.0 || ^4.0",
         "phpdocumentor/type-resolver": "^0.2.1 || ^0.3 || 0.4",
-        "phpunit/phpunit": "^6.1",
+        "phpunit/phpunit": "^6.1 || ^7.0",
```
soyuka: why do we need to change phpunit version?

antograssiot: I was wondering why we were not using the latest already actually. I can revert @soyuka

dunglas: Allowing both versions is better, keep it as is.

composer.json
```diff
@@ -50,22 +50,24 @@
...
+        "symfony/routing": "^3.4 || ^4.0",
         "symfony/security": "^3.0 || ^4.0",
         "symfony/security-bundle": "^3.4 || ^4.0",
         "symfony/twig-bundle": "^3.4 || ^4.0",
-        "symfony/validator": "^3.3 || ^4.0",
-        "symfony/web-profiler-bundle": "^3.3 || ^4.0",
-        "symfony/yaml": "^3.3 || ^4.0",
+        "symfony/validator": "^3.4   || ^4.0",
+        "symfony/web-profiler-bundle": "^3.4 || ^4.0",
+        "symfony/yaml": "^3.4 || ^4.0",
```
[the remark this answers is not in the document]
dunglas: I'm even fine with bumping all Symfony deps to 3.4, 3.3 is not maintained anymore.

composer.json
```diff
@@ -42,30 +42,32 @@
...
         "symfony/phpunit-bridge": "^3.3 || ^4.0",
-        "symfony/routing": "^3.3 || ^4.0",
+        "symfony/routing": "^3.4 || ^4.0",
         "symfony/security": "^3.0 || ^4.0",
         "symfony/security-bundle": "^3.4 || ^4.0",
         "symfony/twig-bundle": "^3.4 || ^4.0",
-        "symfony/validator": "^3.3 || ^4.0",
-        "symfony/web-profiler-bundle": "^3.3 || ^4.0",
-        "symfony/yaml": "^3.3 || ^4.0",
+        "symfony/validator": "^3.4   || ^4.0",
```
dunglas:
```suggestion
        "symfony/validator": "^3.4 || ^4.0",
```

composer.json
soyuka: can't this be why appveyor is not behaving correctly anymore?

antograssiot:
I got appveyor build to pass, just waiting for chocaltey to be up...

antograssiot:
rebase and merge conflict resolved. May we merge ?

appveyor.yml
```diff
@@ -23,5 +23,7 @@ install:
 
 test_script:
   - cd %APPVEYOR_BUILD_FOLDER%
+  - php tests\Fixtures\app\console cache:clear
```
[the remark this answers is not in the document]
antograssiot: I wish I know the real reason... Both suites works without issue independently but when they are run without clearing the cache, in the second suite `NelmioApiDocBundle` was enabled for now real reason and trigger deprecation notices. (`Enabling the NelmioApiDocBundle integration has been deprecated in 2.2 and will be removed in 3.0. NelmioApiDocBundle 3 has native support for API Platform.`)
I've been fighting with this for a while on appveyor and finally looked at travis config
https://github.com/api-platform/core/blob/7dbd20775f02556bd09c8715ac1055e51dc38597/.travis.yml#L45-L53

It might be related to the way that I'm booting the kernel for the test...
😞

phpunit.mongo.xml
```diff
@@ -13,7 +13,7 @@
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak_vendors" />
         <server name="KERNEL_DIR" value="tests/Fixtures/app/" />
         <server name="KERNEL_CLASS" value="AppKernel" />
-        <server name="APP_ENV" value="test_mongodb" />
+        <server name="APP_ENV" value="mongodb" />
```
antograssiot: @alanpoulain is this OK ? I had to update the env so the filters are declared in phpunit too, I didn't understood exactly why it was setup like that ?

soyuka: to me yes it should be `mongodb`

alanpoulain: It shouldn't be `mongo` IMO. It's `test_mongo` (`config_test_mongo.yml`) to match the `test` (`config_test.yml`) in `phpunit.xml`. I don't understand why you need to change it.

soyuka:
```
APP_ENV=mongodb ./vendor/bin/behat --stop-on-failure --suite=mongodb features/main/input_output.feature
```

Is how I launch mongodb tests?

soyuka: mhh check the travis config APP_ENV is set to `mongodb`, the config is set to `config_mongodb.yml`, are you sure?

alanpoulain: We need it to be `test_mongodb` to be equivalent to the ORM part. Or we change also the ORM part (to `postgres` for instance).

antograssiot: it has been reverted @alanpoulain as discussed

alanpoulain:
Thank you @antograssiot!

PR #2317, opened 2018-11-06
Correctly expose overridden formats in Swagger

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | api-platform/api-platform#886
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Api/ExtendedFormatsProviderInterface.php
+++ b/src/Api/ExtendedFormatsProviderInterface.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Api;
+
+/**
+ * Extracts formats for a given operation according to the retrieved Metadata.
+ *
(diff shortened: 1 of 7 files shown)


src/Api/ExtendedFormatsProviderInterface.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Api;
+
+/**
+ * Extracts formats for a given operation according to the retrieved Metadata.
+ *
+ * @author Anthony GRASSIOT <antograssiot@free.fr>
+ */
+interface ExtendedFormatsProviderInterface
```
antograssiot: I'm clear not a fan either on the name either on the idea.
the FormatProviderInterface as added in latest version IIRC, I did it the "clean" way regarding BC I think but I would be in inclined to update the existing interface, it is very quite unlikely that someone already implemented it... WDYT ?

dunglas: At least it could extend `FormatsProviderInterface`. For the name, may I suggest `OperationAwareFormatsProviderInterface`?

antograssiot: indeed, updated.

antograssiot:
I think this one is ready too

dunglas:
Thanks @antograssiot

PR #2288, opened 2018-10-29
Ability to modify response headers (mainly cache related headers)

silverbackdan:
The cache headers should be able to be modified and then not overwritten by the `AddHeadersListener`. Also added annotation configuration for cache header expiry times so each resource can have a different expiry time if required.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/issues/2207
| License       | MIT
| Doc PR        | (not created yet)

PR #2318, opened 2018-11-07
Added possibility to configure if a resource will obtain cache tags

Zodiarc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        | api-platform/doc#647


src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -31,10 +31,15 @@
 final class AddTagsListener
 {
     private $iriConverter;
+    /**
+     * @var array
+     */
+    private $cacheResources;
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, array $cacheResources)
```
dunglas:
```suggestion
    public function __construct(IriConverterInterface $iriConverter, array $cacheResources = [])
```

(Backward compatibility)

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -31,10 +31,15 @@
 final class AddTagsListener
 {
     private $iriConverter;
+    /**
+     * @var array
```
dunglas: Can be removed, but you should add this PHPDoc for the constructor:

```php
/**
 * @param string[] $cacheResources
 */
```

src/HttpCache/EventListener/AddTagsListener.php
dunglas: `$resources`, or `$cachedResources`?

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -54,6 +59,14 @@ public function onKernelResponse(FilterResponseEvent $event)
         }
 
         $resources = $request->attributes->get('_resources');
+        if (!empty($this->cacheResources)) {
```
dunglas: What is this block doing?

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -54,6 +59,14 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+            $cacheResourcesFlipped = array_flip($this->cacheResources);
+            $resources = array_filter($resources, function ($key) use ($cacheResourcesFlipped) {
+                $resourceParts = explode('/', trim($key, '/'));
+
+                return array_key_exists(array_shift($resourceParts), $cacheResourcesFlipped);
+            }, ARRAY_FILTER_USE_KEY);
```
dunglas: The tags are still always sent isn't it?

dunglas:
It would be better to introduce a new attribute to enable or disable the cache on a per resource basis don't you think? Something like:

```php
/**
  * @ApiResource(cacheTags=false, cacheControl="max-age=10; public")
  */

What do you think?

**Comment by antograssiot** [comment]
I would too prefer to use an attribute.
The `cacheHeaders` was added recently in #2288 so we can maybe reuse it. 
```php
/**
  * @ApiResource(cacheHeaders={"max-age"=10, "cache-tags"=false})
  */
class Foo {}
```

src/HttpCache/EventListener/AddTagsListener.php
Zodiarc: Actually resources which should be cached. But Ignore it, I'll try the annotation approach.

Zodiarc:
Is there any way that I can upgrade the api-platform core in my existing api to my customized one to be able to try my changes out before I "greenlight" them for the public? In that case it's important because I never did custom annotations before and want to be 100% sure that it works as I intended.

dunglas:
Yes:

1. Add this to your `composer.json`:

```json
{
...
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/Zodiarc/core"
        },
        "require": {
            "api-platform/core": "dev-master as 2.3"
        }
    ],
...
```

2. Remove your existing `vendor/` folder
3. Run `composer update --prefer-source`

dunglas:
> The cacheHeaders was added recently in #2288 so we can maybe reuse it.

Good idea, indeed.

dunglas:
May I suggest:

```php
/**
  * @ApiResource(cacheHeaders={"max-age"=10, "tags"=false})
  */
class Foo {}
```

(We are already in the context of the cache)

Zodiarc:
About the failing behat test: Unfortunately I have no idea about behat and the error message "expected 200 but got 500" isn't helping at all. Any directions how I can fix it?

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -64,6 +70,33 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+        $resources = $this->removeDisabledResourcesFromCacheTags($resources);
+
         $response->headers->set('Cache-Tags', implode(',', $resources));
     }
+
+    private function removeDisabledResourcesFromCacheTags($resources)
+    {
+        $resourceCacheHeadersPerResourceClass = [];
+        foreach ($resources as $resource) {
+            if (3 !== \count(explode('/', $resource))) { // simple check if it's an item or collection resource
```
antograssiot: just a quick comment, I'll review later tonight or tomorrow but quick question, what is this check doing ?
In your tests you use `/foo` and `/foo/1` so it will never be used.
Moreover as any prefix can be used for resources like `/my/prefix/foo/1` I don't see how it helps.

Zodiarc: This was supposed to be a check for item operations. I didn't take into account that resource iris can be longer. The problem is that if you give a collection iri to the IriConverter->getItemByIri() it will (rightfully) throw an exception. I need a way to also filter things like foo/1 and ID guessing isn't the way to go. Well this was working in my specific use case but you're right, it's not specific enough. My concrete use case is that I retrieve a collection of resources which have subresources which again have subresources. I reached the max header size pretty quick with it since every resource and subresource had a cache-tag.

antograssiot: so filtering  resources just after line 62 where only items are present I think, and collection/subresources in the if line 64 should do the trick no ?

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -64,6 +70,33 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+            $resourceCacheHeadersPerResourceClass[$resourceClass] = $resourceMetadata->getAttribute('cache_header', ['cache_tags' => true]);
+        }
+        $filteredResources = $resources;
+        $results = [];
+        foreach ($resourceCacheHeadersPerResourceClass as $resourceClass => $attributes) {
+            if (array_key_exists('cache_tags', $attributes) && false === $attributes['cache_tags']) {
+                $iri = $this->iriConverter->getIriFromResourceClass($resourceClass);
+                $matches = preg_grep('/^\\'.$iri.'\/{0,1}.*/', $filteredResources);
+                $filteredResources = array_diff($filteredResources, $matches);
+                $results[] = $filteredResources;
```
antograssiot: results isn't used is it ? 🤔

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -64,6 +70,33 @@ public function onKernelResponse(FilterResponseEvent $event)
...
     }
+
+    private function removeDisabledResourcesFromCacheTags($resources)
+    {
+        $resourceCacheHeadersPerResourceClass = [];
+        foreach ($resources as $resource) {
+            if (3 !== \count(explode('/', $resource))) { // simple check if it's an item or collection resource
+                continue;
+            }
+            $resourceClass = \get_class($this->iriConverter->getItemFromIri($resource));
```
antograssiot: The behat tests are failling because those classes could be Doctrine Proxies . You need to import and use the `ApiPlatform\Core\Util\ClassInfoTrait` and then modify this line like so


```suggestion
            $resourceClass =  $this->getObjectClass($this->iriConverter->getItemFromIri($resource));
```

src/HttpCache/EventListener/AddTagsListener.php
Zodiarc: Yes it's a leftover.

dunglas:
@Zodiarc, I'm trying to dive a bit deeper in this PR, but I fail to understand its use case: if you want an URL to not be in cache, can't you just use #2288 to set the appropriate `Cache-Control` header?

Zodiarc:
I could try. Is there a documentation for it?

dunglas:
Not yet (unfortunately). It would be cool to write it.

Zodiarc:
Use case satisfied by #2288 by setting max-age and shared-max-age to 0

Zodiarc:
Any idea why the travis build is failing? Those are things I never touched.

flug:
@Zodiarc Up is this pull request still valid? you can rebase and resolve conflicts ?

Zodiarc:
I'll see what I can do

Zodiarc:
@flug rebase is done

Issue #2116, opened 2018-07-19
Custom operation identifier has a different type from the entity's one but gets normalized with the entity's identifier type.


blesenechal:
I have a GET operation that retrieves an item based on a unique parameter which is a string.
It works fine with 2.2.9, but not with 2.3.0.

It looks like the query parameter value is casted to an int somewhere.

==================================

Example with the api-platform/demo project, to get the books by ISBN

1 - I create a custom operation with "requirements" (as defined in https://api-platform.com/docs/core/operations) :
```
 * @ApiResource(
 *     itemOperations={
 *         "get"={"method"="GET", "path"="/books/{id}"},
 *         "get_by_isbn"={"method"="GET", "path"="/books/by_isbn/{id}", "requirements"={"id"=".+"}},
 *     }
 * )
```

2 - I create a data provider :
```
namespace App\DataProvider;

use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
use ApiPlatform\Core\Exception\ResourceClassNotSupportedException;
use App\Entity\Book;
use Doctrine\ORM\EntityManagerInterface;

final class BookIsbnItemDataProvider implements ItemDataProviderInterface, RestrictedDataProviderInterface
{
    private $entityManager;

    public function __construct(EntityManagerInterface $entityManager)
    {
        $this->entityManager = $entityManager;
    }

    public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
    {
        return (Book::class === $resourceClass) && ('get_by_isbn' === $operationName);
    }

    public function getItem(string $resourceClass, $identifiers, string $operationName = null, array $context = []): ?BlogPost
    {
        return $this->entityManager->getRepository(Book::class)->findOneBy(['isbn' => $identifiers]);
    }
}
```

3 - When executing the request `GET /books/by_isbn/plop` :
- With 2.2.9 : In `getItem`, `$identifiers` is 'plop' (string)
- With 2.3.0 : In `getItem`, `$identifiers` is 0 (int)

soyuka:
In your parameter route `{$id}` has the type of your resource class `$id` property, which is an integer. Because of this it will go through the identifier [IntegerDenormalizer](https://github.com/api-platform/core/blob/master/src/Identifier/Normalizer/IntegerDenormalizer.php) and transform your string to a number. 
[...]

PR #2126, opened 2018-07-25
Proposal: Ability to specify identifier property of custom item operations

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | not really
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | (need to add tests)
| Fixed tickets | https://github.com/api-platform/core/issues/2116
| License       | MIT
| Doc PR        | todo

When declaring a custom item operation, you often want to use another property as identifier, for example:

```php
<?php
/**
 * @ApiResource(itemOperations={
 *     "get",
 *     "put",
 *     "get_by_isbn"={"method"="GET", "path"="/books/by_isbn/{isbn}.{_format}", "requirements"={"isbn"=".+"}}
 * })
 */
class Book {}
``` 

Today, this only works if you add a custom controller, either by using your own data or by calling the core data provider with new identifiers `['isbn' => 1]`.

My proposal is to add a property in the operation route defaults:

```php
<?php
/*
 * @ApiResource(itemOperations={
 *     "get",
 *     "put",
 *     "get_by_isbn"={"method"="GET", "path"="/books/by_isbn/{isbn}.{_format}", "requirements"={"isbn"=".+"}, "defaults"={"identifiedBy"="isbn"}}
 * })
 */
class Book {}
```

By using this you have nothing else to add (no controller, no data provider), it'll just use the `isbn` property as identifier.

This is a WIP, to be done:

- [ ] accept proposal
- [ ] add tests

PR #2323, opened 2018-11-12
Integer denormalizer is numeric

Soullivaneuh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2116 (cc @blesenechal, please confirm)
| License       | MIT
| Doc PR        | N/A

Concrete case:

```php
final class ServerDataProvider implements ItemDataProviderInterface, RestrictedDataProviderInterface
{
    /**
     * @var EntityManagerInterface
     */
    private $entityManager;

    public function __construct(EntityManagerInterface $entityManager)
    {
        $this->entityManager = $entityManager;
    }

    /**
     * {@inheritdoc}
     */
    public function getItem(string $resourceClass, $id, ?string $operationName = null, array $context = []): ?Server
    {
        $serverRepository = $this->entityManager->getRepository(Server::class);
        $server = $serverRepository->find($id);

        if (null === $server) {
            $server = $serverRepository->findOneBy([
                'name' => $id,
            ]);
        }

        return $server;
    }

    /**
     * {@inheritdoc}
     */
    {
        return Server::class === $resourceClass;
    }
}
```

This permit to try find a server by it's name if nothing is found with the id.

But in the current stable version, the string value will always be `0` because it's converted into string by the denormalizer.

Current workaround with a decorated service:

```php
/**
 * @see https://github.com/api-platform/core/issues/2116#issuecomment-437909022
 * @see https://github.com/api-platform/core/pull/2323
 */
final class IntegerDenormalizer implements DenormalizerInterface, CacheableSupportsMethodInterface
{
    /**
     * @var \ApiPlatform\Core\Identifier\Normalizer\IntegerDenormalizer
     */
    private $decorated;

    public function __construct(\ApiPlatform\Core\Identifier\Normalizer\IntegerDenormalizer $decorated)
    {
    }

    /**
     * {@inheritdoc}
     */
    public function denormalize($data, $class, $format = null, array $context = array())
    {
        return $this->decorated->denormalize($data, $class, $format, $context);
    }

    /**
     * {@inheritdoc}
     */
    {
        return $this->decorated->supportsDenormalization($data, $type, $format) && \is_numeric($data);
    }

    {
        return $this->decorated->hasCacheableSupportsMethod();
    }
}
```

### DIFF
--- a/tests/Identifier/Normalizer/IntegerDenormalizerTest.php
+++ b/tests/Identifier/Normalizer/IntegerDenormalizerTest.php
@@ -31,6 +31,8 @@ class IntegerDenormalizerTest extends TestCase
         $normalizer = new IntegerDenormalizer();
         $this->assertTrue($normalizer->supportsDenormalization('1', 'int'));
         $this->assertFalse($normalizer->supportsDenormalization([], 'int'));
+        $this->assertFalse($normalizer->supportsDenormalization('kdunglas', 'int'));
+        $this->assertFalse($normalizer->supportsDenormalization('7.2', 'int'));
         $this->assertFalse($normalizer->supportsDenormalization('1', 'foo'));
         $this->assertTrue($normalizer->hasCacheableSupportsMethod());
     }
--- a/src/Identifier/Normalizer/IntegerDenormalizer.php
+++ b/src/Identifier/Normalizer/IntegerDenormalizer.php
@@ -29,7 +29,7 @@ final class IntegerDenormalizer implements DenormalizerInterface, CacheableSuppo
      */
     public function supportsDenormalization($data, $type, $format = null): bool
     {
-        return Type::BUILTIN_TYPE_INT === $type && \is_string($data);
+        return Type::BUILTIN_TYPE_INT === $type && \ctype_digit($data);
     }
 
     /**


dunglas:
Looks legit to me, but let's wait for @soyuka's review. It should target 2.3 as it's a bug fix.

antograssiot:
@Soullivaneuh Do you still have the issue if your ItemProvider implements the [DenormalizedIdentifiersAwareItemDataProviderInterface](https://github.com/api-platform/core/blob/master/src/DataProvider/DenormalizedIdentifiersAwareItemDataProviderInterface.php#L21) ?
That's the "latest" way to do it if I remember correctly, you will get `$identifiers` as an array and should not need this modification.

blesenechal:
@Soullivaneuh : This fixes my issue (#2116)
@antograssiot : Implementing DenormalizedIdentifiersAwareItemDataProviderInterface does not fix the issue for me, this was already suggested by @soyuka (see https://github.com/api-platform/core/issues/2116#issuecomment-407684356)

soyuka:
The issue here is that the `$type` should not be `Type::BUILTIN_TYPE_INT === $type`, this way it won't get transformed to an integer right?

IMHO this use case is wrong, you can't use two different types for a given identifier. Use a custom operation if you want to fetch by `name`.

Or, you can always add another `IdentifierNormalizer` that gets called before the `Integer` one to "hack around" this.

soyuka:
Actually thought a bit more about this and I'm okay (if we can avoid it it's better) to use `is_numeric` instead of `is_string`.

Also please check out https://github.com/api-platform/core/pull/2126 which would be a better way of fixing this. 

Then again, please don't do the following:

```
 $server = $serverRepository->find($id);

        if (null === $server) {
            $server = $serverRepository->findOneBy([
                'name' => $id,
            ]);
        }
```

The `identifier` of this route is an `id` not a `name`, it just looks wrong and dirty :stuck_out_tongue:

antograssiot:
understood, it looks okish to me :+1: 
hopefully you won't get a name that will be "7.2" :wink: 

@soyuka is it still necessary to check the `is_string` here to avoid to cast values that are already integer, or it should not happen nor is a big deal ?

Soullivaneuh:
> hopefully you won't get a name that will be "7.2"

Hmm... it is not an integer either. :D

Maybe use this instead? http://php.net/manual/fr/function.ctype-digit.php

Soullivaneuh:
Updated the PR with `ctype_digit`.

soyuka:
> Do you have more details about the IdentifierNormalizer? This may be a subject for a documentation issue.

https://api-platform.com/docs/core/identifiers/#custom-identifier-normalizer

tests/Identifier/Normalizer/IntegerDenormalizerTest.php
```diff
@@ -31,6 +31,8 @@ public function testSupportsDenormalization()
         $normalizer = new IntegerDenormalizer();
         $this->assertTrue($normalizer->supportsDenormalization('1', 'int'));
         $this->assertFalse($normalizer->supportsDenormalization([], 'int'));
+        $this->assertFalse($normalizer->supportsDenormalization('kdunglas', 'int'));
```
soyuka: This is so untrue :|. How can an identifier be a string but typed as an integer?

Soullivaneuh: Stop me if I'm wrong, but I think the identified type come from PHPDoc parsing and class reflection.

Soullivaneuh:
@soyuka I read this documentation, but the sample show a single identifier related to an object. I just want to check on two field, one int and one string. I don't see how to achieve it this way.

tests/Identifier/Normalizer/IntegerDenormalizerTest.php
soyuka: Yes, it's linked to the property in your class. This is why it's wrong because if you typed `id` as `int` there are no reasons it should be a string ^^.

soyuka:
For example:


```


<?php
namespace App\Identifier;

use ApiPlatform\Core\Exception\InvalidIdentifierException;
// assumes you indeed typed your property with
// * @IntOrString
use App\Type\IntOrString;

final class IntOrStringNormalizer implements DenormalizerInterface
{
    {
        return is_numeric($data) ? (int) $data : $data;
    }

    {
        // you could also check if the $data is implementing an interface or just return `true` to override default normalizers 
        return is_a($type, IntOrString::class, true);
    }
}
```

tests/Identifier/Normalizer/IntegerDenormalizerTest.php
Soullivaneuh: Dirty solution, but works easier than the proper one. Maybe an issue may be open about OR conditions on identifier.

Soullivaneuh:
@soyuka Your solution will not work for me because I have two field. Here, you assume I have one field with the `IntOrString` type, am I right?

soyuka:
Where do you have 2 fields? It'll work as a workaround because the normalizer above won't denormalize your identifier.
The example above works with an entity of type:

```
class Entity {
  /**
   * @var IntOrString
   */
  public $id;
}
```
And you can then do (even though the following is bad design :p):

```
 $server = $serverRepository->find($id);

if (null === $server) {
    $server = $serverRepository->findOneBy([
        'name' => $id,
    ]);
}
``` 

Or even better now that you're sure that the string will be an integer:

```
if (is_string($id)) {
  return $repo->findOneByName($id);
}

return $repo->find($id);
```

Soullivaneuh:
> Where do you have 2 fields?

See https://github.com/api-platform/core/issues/2324#issuecomment-438320283

@soyuka At final, it the same thing, looking for two field with the same value.

You just advise me to create a special type with a custom normalizer just to make string and integer working. Here, the integer normalizer does not work as excpected as it should not convert anything that does not looks like an integer. Otherwise, you may accept any value.

And if nothing is able to be normalized, then I'll have the raw value. And that works for a string. ;-)

tests/Identifier/Normalizer/IntegerDenormalizerTest.php
soyuka: Because the identifier should be unique, there should not be any `or` imho.

soyuka:
> You just advise me to create a special type with a custom normalizer just to make string and integer working. 

Yes.

> Here, the integer normalizer does not work as excpected as it should not convert anything that does not looks like an integer. 

`id` in the url matches the `id` property, which is typed as an integer, so it should be transformed as one. If you take a `date` for example, it's even truer because if it's not converted it'll be error-prone.

Soullivaneuh:
So the result is the same. Having one type allowing int or string, placed on a property that should be an int, and search for a name. Same ugly thing. :trollface: :wink: 

> `id` in the url matches the `id` property, which is typed as an integer, so it should be transformed as one.

Well, in that case, why do we have this test:

```php
$this->assertFalse($normalizer->supportsDenormalization([], 'int'));
```

And why we already check value to be a string? :thinking: 

With this PR, I'm looking at the normalizer itself and nothing else. It's an integer normalizer, so it should convert intergers only. Or, rename it to `ApiPropertyIntNormalizer` in that case. :-)

Do you better see my point of view?

teohhanhui:
I think @soyuka has explained it pretty well. So I'm -1 on this.

Soullivaneuh:
I really don't understand here. In any case, why an integer normalizer should normalize something that does not represent an integer at all?

Even if my final needs does not fit your best practice, it's still an issue, and not related.

PR #2328, opened 2018-11-14
(Very small improvement) Improves exception handling in ReadListener

timonf:
This PR improves exception handling to throw also previous Exception. I spent a lot of time, to figure out what's happening there: ReadListener should not mute Exceptions. It should also pass previously caught Exception (in line 104 like this commit or in line 108). I also improved NotFoundHttpException's message.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | hopefully
| Fixed tickets | no ticket for this
| License       | MIT
| Doc PR        | no doc needed

branch/version: 2.3 (because in 2.2 we are catching an even more common exception)

### DIFF
--- a/src/EventListener/ReadListener.php
+++ b/src/EventListener/ReadListener.php
@@ -101,7 +101,7 @@ final class ReadListener
                 $data = $this->getSubresourceData($identifiers, $attributes, $context);
             }
         } catch (InvalidIdentifierException $e) {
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks @timonf

Issue #2167, opened 2018-08-20
[GraphQL] kernel.view event never fired


jamesisaac:
The API Platform [Event System Documentation](https://api-platform.com/docs/core/events) lists many places in the request/response cycle you may want to attach your own listeners, the majority of them on `kernel.view`.  This uses `GetResponseForControllerResultEvent` which is a helpful event, as you can do some useful things with the controller result.  I've had good results adding some business logic to my API here (e.g. advanced authorization, sending out automated emails).

However in my experiments with using GraphQL, the `kernel.view` event doesn't seem to be getting fired for either queries or mutations?

Is this intentional?  This makes it hard to offer a dual REST/GraphQL API that offers the same behaviour when similar actions are performed...?

Symfony: 4.0
API Platform: 2.2.8

dunglas:
It's unfortunately a known issue. There is currently no event system for GraphQL.
We have a plan to fix this:

1. create new "domain" events like "serialize", "persist", "validate"
2. throw these new events from actual REST listeners **and** from GraphQL

I'll work on this ASAP, but any help is welcome if someone want to handle this before I do it :)

jamesisaac:
Hmm I see, thanks for the quick response.  Some thought on this:

1. Might be worth having a note in the docs about this in the mean time (and if there are any other major features missing), as I presumed the state of GraphQL support was: ready to use, but a bit less that's done for you out of the box.  But having no events to hook into adds a lot of limits to what's actually possible without modifying this library itself.
2. More events will be needed the way you describe: will need "pre_\*" and "post_\*" events for everything you listed, as API platform isn't performing its logic in response to those events any more, so you can't use priorities to control whether the user's listener comes before/after.
3. Looking at how the GraphQL integration works, might it be a bit tricky mapping it to the same linear flow of events?  I'm not that familiar with it, but it seems to be more along the lines of "resolving" (carrying out most of the lifecycle) on one entity at a time.  I guess while [mutations are limited to a single entity](https://github.com/api-platform/core/pull/1702) that might not matter though.

I've made a very simple implementation here: https://github.com/api-platform/core/compare/master...jamesisaac:graphql-events which I think is enough to get what I need in my project (still listening to the old style REST events, as well as GraphQL dispatching a new event that shares a similiar interface).

But I don't think I'm familiar enough with API Platform's codebase to know how you'd really want those events designed / where exactly they should be dispatched from, sadly.

lukasluecke:
[...]

PR #2329, opened 2018-11-15
WIP: Add events

Deuchnord:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2167, api-platform/api-platform#734
| License       | MIT
| Doc PR        | _later_

Adding events to make API Platform’s actions tweaking easier.

Todos:
- [x] Add Pre- and Post- event for each action (based on the `\ApiPlatform\Core\EventListener\*` listeners)
- [x] Wire the actions to the events
- [x] Wire GraphQL to the events
- [x] Write unit tests
- [ ] Write Behat tests


alanpoulain:
Thanks for working on this :slightly_smiling_face:

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -146,6 +146,7 @@
             <argument type="service" id="api_platform.subresource_data_provider" />
             <argument type="service" id="api_platform.serializer.context_builder" />
             <argument type="service" id="api_platform.identifier.converter" />
+            <argument type="service" id="event_dispatcher" />
```
dunglas: `symfony/event-dispatcher` must then be added to the `require` section of `composer.json` to avoid a transient dependency.

src/Event/PostReadEvent.php
```diff
@@ -0,0 +1,60 @@
...
+use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+use ApiPlatform\Core\Identifier\IdentifierConverterInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class PostReadEvent extends Event
+{
+    const NAME = 'api_platform.post_read';
+
+    /** @var CollectionDataProviderInterface */
```
dunglas: Could be removed (documented in the constructor)

src/Event/PostReadEvent.php
```diff
@@ -0,0 +1,60 @@
...
+    /** @var ItemDataProviderInterface */
+    protected $itemDataProvider;
+    /** @var SubresourceDataProviderInterface */
+    protected $subresourceDataProvider;
+    /** @var SerializerContextBuilderInterface */
+    protected $serializerContextBuilder;
+    /** @var IdentifierConverterInterface */
+    protected $identifierConverter;
+
+    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null)
```
dunglas: The event must be a value object, it should have no behavior, and should not depend of services. The behaviors (and the deps to the services) must be injected directly in the subscribers/listeners.

To me events, should just contain the current value of the `data` attribute.

src/Event/PostRespondEvent.php
```diff
@@ -0,0 +1,25 @@
...
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class PostRespondEvent extends Event
+{
+    const NAME = 'api_platform.post_respond';
+
+    protected $resourceMetadataFactory;
+
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
```
dunglas: same here

src/Event/PostRespondEvent.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+
+namespace ApiPlatform\Core\Event;
+
+
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class PostRespondEvent extends Event
```
dunglas:
```suggestion
final class PostRespondEvent extends Event
```

src/Event/PostReadEvent.php
```diff
@@ -0,0 +1,60 @@
...
+namespace ApiPlatform\Core\Event;
+
+use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
+use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
+use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+use ApiPlatform\Core\Identifier\IdentifierConverterInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class PostReadEvent extends Event
```
dunglas:
```suggestion
final class PostReadEvent extends Event
```

We mark all classes final by default

src/Event/PostSerializeEvent.php
```diff
@@ -0,0 +1,35 @@
...
+
+
+namespace ApiPlatform\Core\Event;
+
+
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\Serializer\SerializerInterface;
+
+class PostSerializeEvent extends Event
```
dunglas: same

src/Event/PreReadEvent.php
```diff
@@ -0,0 +1,60 @@
...
+namespace ApiPlatform\Core\Event;
+
+use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
+use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
+use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+use ApiPlatform\Core\Identifier\IdentifierConverterInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class PreReadEvent extends Event
```
dunglas: same

src/Event/PreReadEvent.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    const NAME = 'api_platform.pre_read';
```
dunglas: You should store the names in a dedicated class, like https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/FormEvents.php

src/Event/PostReadEvent.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    const NAME = 'api_platform.post_read';
+
+    /** @var CollectionDataProviderInterface */
+    protected $collectionDataProvider;
```
dunglas: We always use `private` when possible (closed by default). Same everywhere.

src/Event/PreRespondEvent.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+
+namespace ApiPlatform\Core\Event;
+
+
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class PreRespondEvent extends Event
```
dunglas: same

src/Event/PreSerializeEvent.php
```diff
@@ -0,0 +1,35 @@
...
+
+
+namespace ApiPlatform\Core\Event;
+
+
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\Serializer\SerializerInterface;
+
+class PreSerializeEvent extends Event
```
dunglas: same

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -95,20 +103,37 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
                     $this->validate($item, $info, $resourceMetadata, $operationName);
+
+                    if (null !== $this->dispatcher) {
+                        $this->dispatcher->dispatch(PreWriteEvent::NAME, new PreWriteEvent($operationName, $item));
```
Deuchnord: Guess there will be an issue here, since the `$operationName` from GraphQL (`create`, `update`, `delete`) is not the same as the HTTP methods used in REST

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -49,8 +54,12 @@
     private $requestStack;
     private $paginationEnabled;
     private $resourceMetadataFactory;
+    /**
+     * @var EventDispatcherInterface
```
alanpoulain: I don't think the PHPDoc is needed.

Deuchnord: It’s not, indeed... It’s my PhpStorm’s default behavior, sorry for that. I’m fixing it.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -84,17 +94,38 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             if (isset($rootClass, $source[$rootProperty = $info->fieldName], $source[ItemNormalizer::ITEM_KEY])) {
                 $rootResolvedFields = $this->identifiersExtractor->getIdentifiersFromItem(unserialize($source[ItemNormalizer::ITEM_KEY]));
                 $subresource = $this->getSubresource($rootClass, $rootResolvedFields, array_keys($rootResolvedFields), $rootProperty, $resourceClass, true, $dataProviderContext);
+
+                if (null !== $this->dispatcher) {
+                    $this->dispatcher->dispatch(PreReadEvent::NAME, new PreReadEvent($subresource));
```
alanpoulain: I don't think this is a pre-read here since the subresource has been already fetched.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -84,17 +91,38 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             if (isset($rootClass, $source[$rootProperty = $info->fieldName], $source[ItemNormalizer::ITEM_KEY])) {
                 $rootResolvedFields = $this->identifiersExtractor->getIdentifiersFromItem(unserialize($source[ItemNormalizer::ITEM_KEY]));
                 $subresource = $this->getSubresource($rootClass, $rootResolvedFields, array_keys($rootResolvedFields), $rootProperty, $resourceClass, true, $dataProviderContext);
+
+                if (null !== $this->dispatcher) {
```
alanpoulain: Can't we use a `voidDispatcher` to avoid all these null checks?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -84,17 +91,38 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
                 $data = [];
                 foreach ($collection as $index => $object) {
+                    if (null !== $this->dispatcher) {
+                        $this->dispatcher->dispatch(PreSerializeEvent::NAME, new PreSerializeEvent($collection));
+                    }
+
                     $data[$index] = $this->normalizer->normalize($object, ItemNormalizer::FORMAT, $dataProviderContext);
+
+                    if (null !== $this->dispatcher) {
+                        $this->dispatcher->dispatch(PostSerializeEvent::NAME, new PostSerializeEvent($collection));
```
alanpoulain: Shouldn't we add the serialized data in this event?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -117,10 +145,18 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
+                    $this->dispatcher->dispatch(PreSerializeEvent::NAME, new PreSerializeEvent($collection));
+                }
+
                 $data['edges'][$index] = [
                     'node' => $this->normalizer->normalize($object, ItemNormalizer::FORMAT, $dataProviderContext),
                     'cursor' => base64_encode((string) ($index + $offset)),
                 ];
+
+                if (null !== $this->dispatcher) {
+                    $this->dispatcher->dispatch(PostSerializeEvent::NAME, new PostSerializeEvent($collection));
```
alanpoulain: Shouldn't we add the serialized data in this event?

src/GraphQl/Resolver/ItemResolver.php
```diff
@@ -73,6 +88,16 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
...
 
-        return $this->normalizer->normalize($item, ItemNormalizer::FORMAT, $normalizationContext + $baseNormalizationContext);
+        if (null !== $this->dispatcher) {
+            $this->dispatcher->dispatch(PreSerializeEvent::NAME, new PreSerializeEvent($item));
+        }
+
+        $normalizedObject = $this->normalizer->normalize($item, ItemNormalizer::FORMAT, $normalizationContext + $baseNormalizationContext);
+
+        if (null !== $this->dispatcher) {
+            $this->dispatcher->dispatch(PostSerializeEvent::NAME, new PostSerializeEvent($item));
```
alanpoulain: Shouldn't we use the `$normalizedObject` here?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
Deuchnord: Well, I wasn’t sure about that, but it actually seems more logical, indeed...
Do I keep the collection in the object, or will just the serialized result suffice?

alanpoulain: IMO the serialized result only is okay.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
Deuchnord: Discussed in private with @dunglas and @alanpoulain, reached a consensus: keeping the current form (`null !==`) for the following reasons:
- best performance;
- less code to maintain.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -84,17 +91,38 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
                 foreach ($collection as $index => $object) {
+                    if (null !== $this->dispatcher) {
+                        $this->dispatcher->dispatch(PreSerializeEvent::NAME, new PreSerializeEvent($collection));
+                    }
+
                     $data[$index] = $this->normalizer->normalize($object, ItemNormalizer::FORMAT, $dataProviderContext);
+
+                    if (null !== $this->dispatcher) {
+                        $this->dispatcher->dispatch(PostSerializeEvent::NAME, new PostSerializeEvent($data[$index]));
+                    }
```
soyuka: dispatching every item in the collection looks heavy, can't we jus dispatch the collection?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -84,17 +91,38 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
+            }
+
             $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $collection, $operationName ?? 'query');
 
             if (!$this->paginationEnabled) {
                 $data = [];
                 foreach ($collection as $index => $object) {
+                    if (null !== $this->dispatcher) {
+                        $this->dispatcher->dispatch(PreSerializeEvent::NAME, new PreSerializeEvent($collection));
+                    }
```
soyuka: this should be outside the foreach loop

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -117,10 +145,18 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             }
 
             foreach ($collection as $index => $object) {
+                if (null !== $this->dispatcher) {
+                    $this->dispatcher->dispatch(PreSerializeEvent::NAME, new PreSerializeEvent($collection));
+                }
```
soyuka: same the colleciton dispatch should be outside the loop

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -117,10 +145,18 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
+
                 $data['edges'][$index] = [
                     'node' => $this->normalizer->normalize($object, ItemNormalizer::FORMAT, $dataProviderContext),
                     'cursor' => base64_encode((string) ($index + $offset)),
                 ];
+
+                if (null !== $this->dispatcher) {
+                    $this->dispatcher->dispatch(PostSerializeEvent::NAME, new PostSerializeEvent($data['edges'][$index]));
+                }
```
soyuka: same comment as above, looks heavy for no real benefits?

src/GraphQl/Resolver/ItemResolver.php
```diff
@@ -62,7 +69,15 @@ public function __invoke($source, $args, $context, ResolveInfo $info)
...
+                $this->dispatcher->dispatch(PreReadEvent::NAME, new PreReadEvent(null));
+            }
+
             $item = $this->iriConverter->getItemFromIri($args['id'], $baseNormalizationContext);
+
+            if (null !== $this->dispatcher) {
+                $this->dispatcher->dispatch(PostReadEvent::NAME, new PostReadEvent($item));
+            }
         } catch (ItemNotFoundException $e) {
             return null;
```
antograssiot: wouldn't it be better to change this block to 
```php
try {
    if (null !== $this->dispatcher) {
        $this->dispatcher->dispatch(PreReadEvent::NAME, new PreReadEvent(null));
    }
    $item = $this->iriConverter->getItemFromIri($args['id'], $baseNormalizationContext);
} catch (ItemNotFoundException $e) {
    $item = null;
}

if (null !== $this->dispatcher) {
    $this->dispatcher->dispatch(PostReadEvent::NAME, new PostReadEvent($item));
}

if (null === $item) {
    return null;
}
```
to ensure that the PostReadEvent is always sent like in the ReadListener ?

soyuka:
>    read
    deserialize
    write
    validate
    serialize

do you mean that we could avoid PRE/POST events to have a single `READ` for example? I'm :+1:

Deuchnord:
> I think we can restrict the events to:
> 
>  - read
>  - deserialize
>  - write
>  - validate
>  - serialize
> 
> Other ones are Symfony specific and should be handled using Symfony kernel events. WDYT?

@dunglas Sound good, but I think we should keep the pre- and post-  events.

src/Event/Events.php
```diff
@@ -0,0 +1,38 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Event;
+
+final class Events
+{
+    const PRE_READ = 'api_platform.pre_read';
+    const POST_READ = 'api_platform.post_read';
+
+    const PRE_WRITE = 'api_platform.pre_write';
```
teohhanhui: Can we put them in the correct order? Read -> Deserialize -> Validate -> Write -> Serialize

src/Event/Events.php
```diff
@@ -0,0 +1,38 @@
...
+    const PRE_SERIALIZE = 'api_platform.pre_serialize';
+    const POST_SERIALIZE = 'api_platform.post_serialize';
+
+    const PRE_DESERIALIZE = 'api_platform.pre_deserialize';
+    const POST_DESERIALIZE = 'api_platform.post_deserialize';
+
+    const PRE_RESPOND = 'api_platform.pre_respond';
+    const POST_RESPOND = 'api_platform.post_respond';
+
+    const PRE_ADD_FORMAT = 'api_platform.pre_add_format';
```
teohhanhui: These events should be removed as they're not related to the resource.

src/Event/Events.php
```diff
@@ -0,0 +1,38 @@
...
+    const PRE_WRITE = 'api_platform.pre_write';
+    const POST_WRITE = 'api_platform.post_write';
+
+    const PRE_SERIALIZE = 'api_platform.pre_serialize';
+    const POST_SERIALIZE = 'api_platform.post_serialize';
+
+    const PRE_DESERIALIZE = 'api_platform.pre_deserialize';
+    const POST_DESERIALIZE = 'api_platform.post_deserialize';
+
+    const PRE_RESPOND = 'api_platform.pre_respond';
```
teohhanhui: These events should be removed as they're not related to the resource.

src/Event/PreWriteEvent.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Event;
+
+use Symfony\Component\EventDispatcher\Event;
+
+final class PreWriteEvent extends Event
+{
+    const NAME = Events::PRE_WRITE;
```
teohhanhui: This is a really weird thing to do. Did I miss the discussion on why this is done this way?

Deuchnord: You mean, defining a `NAME` constant in the classes?
I did it mainly for comfort purpose, actually: it makes more sense to me to use a constant attached to the concerned class, so I kept it after @dunglas asked me to create the separate `Events` class.
Do you want me to remove the `NAME` constants?

src/Event/Events.php
Deuchnord: Sorry, should I remove that event [or move it](https://github.com/api-platform/core/pull/2329#discussion_r239877920)?

src/Event/PreWriteEvent.php
teohhanhui: I don't see the value of these constants. Just use e.g. `Events::PRE_WRITE` directly.

src/Event/Events.php
teohhanhui: Only keep the events mentioned in https://github.com/api-platform/core/pull/2329#discussion_r239877920 (in the correct order).

teohhanhui: I see you've already removed the "Respond" events, so it's already correct.

src/Event/Events.php
teohhanhui: Please remove the "Add Format" events.

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
```diff
@@ -18,6 +18,7 @@
             <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="null" />
             <argument type="service" id="request_stack" />
             <argument>%api_platform.collection.pagination.enabled%</argument>
+            <argument type="service" id="event_dispatcher" />
```
teohhanhui: You need to add the same `on-invalid="null"` in this file too.

src/GraphQl/Resolver/ItemResolver.php
teohhanhui: I don't think it should be dispatched when Read has failed.

src/EventListener/ReadListener.php
```diff
@@ -90,6 +95,10 @@ public function onKernelRequest(GetResponseEvent $event)
         try {
             $identifiers = $this->extractIdentifiers($request->attributes->all(), $attributes);
 
+            if (null !== $this->dispatcher) {
+                $this->dispatcher->dispatch(PreReadEvent::NAME, new PreReadEvent($data));
```
teohhanhui: It's really weird to have data in Pre Read. I don't think it makes sense...

teohhanhui: Perhaps what we could do is to keep a reference to the event object, then check if the data is already set on the event. If it is set, we should not override it.

jamesisaac:
Very glad this is being worked on.

Is the idea with not attaching the `Request` to these events (unlike their Symfony counterparts) that the Request can be pulled out manually from the `RequestStack` service?  I don't think the current events on their own contain enough data that you'd be able to determine whether to run most real-world side-effects (normally I'm checking the method, collection/item operation name, or route name).

Also, with the current events system they're executed in a very predictable linear order.  Am I understanding that due to the way GraphQL resolvers work (one entity at a time), these events can be executed more like `1..n` times (or `0..n`) per request if the graphql request contains some nesting?  Just trying to think if that could create any awkward conditions where side-effects get executed multiple times.

Deuchnord:
Hi, sorry for the late, I forgot to answer when I saw your comment ^^

> Is the idea with not attaching the Request to these events (unlike their Symfony counterparts) that the Request can be pulled out manually from the RequestStack service? I don't think the current events on their own contain enough data that you'd be able to determine whether to run most real-world side-effects (normally I'm checking the method, collection/item operation name, or route name).

Actually, these events are supposed to be higher levelled than Symfony's ones, so they're meant to contain only the details about the data themselves. If you want to get details about the request itself, you'll probably want to intercept Symfony's events instead.

> Also, with the current events system they're executed in a very predictable linear order. Am I understanding that due to the way GraphQL resolvers work (one entity at a time), these events can be executed more like `1..n` times (or `0..n`) per request if the graphql request contains some nesting? Just trying to think if that could create any awkward conditions where side-effects get executed multiple times.

That GraphQL part is actually quite cryptic to me, I need to study that more… if you have any tips to make my integration of the events better, feel free to share them :slightly_smiling_face:

jamesisaac:
No worries!

> Actually, these events are supposed to be higher levelled than Symfony's ones, so they're meant to contain only the details about the data themselves. If you want to get details about the request itself, you'll probably want to intercept Symfony's events instead.

I thought one of the motivations here was that Symfony events aren't getting fired during the lifecycle of GraphQL requests (see #2167 )?  So do you mean intercepting a combination of the 2?  I.e. in a single listener, listen to Symfony's request event to store the route/method, then combine that with the data retrieved from this event, to execute the side effect?

I just think that individually, these events with "only the details about the data", and Symfony's events which don't have the controller result, don't contain enough information to add granular side effects in a lot of real world cases.  Both are needed together most of the time.

Deuchnord:
> I thought one of the motivations here was that Symfony events aren't getting fired during the lifecycle of GraphQL requests (see #2167 )? So do you mean intercepting a combination of the 2? I.e. in a single listener, listen to Symfony's request event to store the route/method, then combine that with the data retrieved from this event, to execute the side effect?

Yes, that's what's planned ideally: an event that's risen, whenever the action comes from REST or GraphQL :)

> I just think that individually, these events with "only the details about the data", and Symfony's events which don't have the controller result, don't contain enough information to add granular side effects in a lot of real world cases. Both are needed together most of the time.

I think I see what you mean, problem is, I'm not sure I can get that Symfony event, for the same reason we can't do it currently like described in #2167… We'll need to think more about it, maybe it would be a good thing to have an object that contain the information that may be useful and are available for both REST and GraphQL?

Deuchnord:
What about adding to all the events an object that would be common to both REST and GraphQL, and  would contain the context of the events? This way, you would get as many information as possible, without having to take care about Symfony's events being actually risen or not.

Currently, I can see at least two things this event could store:
- the origin of the event (REST or GraphQL),
- the request type (creation, read, update or delete).

Plus, we could also append the old version of the data in the `PostWriteListener`.

WDYT? Do you see any other thing that could be interesting to add in such an object?

@dunglas and @teohhanhui, I'd be interested by your opinion about that too :)

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -95,20 +100,37 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
                     $this->validate($item, $info, $resourceMetadata, $operationName);
```
alanpoulain: Missing PreValidate event.

jamesisaac:
> What about adding to all the events an object that would be common to both REST and GraphQL, and would contain the context of the events? This way, you would get as many information as possible, without having to take care about Symfony's events being actually risen or not.

That would be very helpful for me.  But I don't know if it breaks a separation of concerns or something.

jamesisaac:
> What about adding to all the events an object that would be common to both REST and GraphQL, and would contain the context of the events? This way, you would get as many information as possible, without having to take care about Symfony's events being actually risen or not.

That would be very helpful to me.  In addition to your list, the things I make use of are:

- The entity being mutated (ControllerResult)
- The route name (e.g. when there are multiple POST actions defined with different semantics)

Deuchnord:
Closing this PR since a better implementation is proposed on #2506.

PR #2337, opened 2018-11-17
return HTTP_CONFLICT 409 on Doctrine OptimisticLockException

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes?
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | =
| License       | MIT
| Doc PR        | 

Seems like a nice sane default for Doctrine. thoughts?
I needed this in my API anyway.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection;
 
 use ApiPlatform\Core\Exception\FilterValidationException;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
+use Doctrine\ORM\OptimisticLockException;
 use FOS\UserBundle\FOSUserBundle;
 use GraphQL\GraphQL;
 use Symfony\Bundle\TwigBundle\TwigBundle;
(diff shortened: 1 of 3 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -259,11 +261,13 @@ private function addExceptionToStatusSection(ArrayNodeDefinition $rootNode)
...
-                        ExceptionInterface::class => Response::HTTP_BAD_REQUEST,
-                        InvalidArgumentException::class => Response::HTTP_BAD_REQUEST,
-                        FilterValidationException::class => Response::HTTP_BAD_REQUEST,
-                    ])
+                    ->defaultValue(
+                        array_merge([
+                            ExceptionInterface::class => Response::HTTP_BAD_REQUEST,
+                            InvalidArgumentException::class => Response::HTTP_BAD_REQUEST,
+                            FilterValidationException::class => Response::HTTP_BAD_REQUEST,
+                        ], class_exists(Version::class) ? [OptimisticLockException::class => Response::HTTP_CONFLICT] : [])
```
bendavies: using `Version::class` as that is the convention.

antograssiot:
Seems a bit app specific to me ...

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
dunglas: Looks useless, PHP will allow `OptimisticLockException::class` even if the class doesn't actually exist. It doesn't hurt to have this extra key.

bendavies: done

dunglas:
Thanks @bendavies

bendavies:
thanks @dunglas

PR #2347, opened 2018-11-23
Mercure: allow to register MercureBundle after API Platform

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a


meyerbaptiste:
@dunglas this PR introduces another bug.

If the Mercure component is installed but not the bundle then we get an exception:
```
In ParameterBag.php line 100:
                                                                                                                                            
  [Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException]                                                              
  The service "api_platform.mercure.listener.response.add_link_header" has a dependency on a non-existent parameter "mercure.default_hub".
```

PR #2348, opened 2018-11-25
Allow strings for single serialization groups

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Prevent an error when using the following configuration (using strings in this case is supported in Symfony Serializer since the latest version):

```php
/**
 * @ORM\Entity(repositoryClass="App\Repository\UserRepository")
 * @ApiResource(
 *     normalizationContext={"groups": "user:read"}
 * )
 */
```

This PR fixes a performance warning thrown by PHP Inspections.

### DIFF
--- a/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
@@ -136,13 +136,14 @@ final class SerializerPropertyMetadataFactory implements PropertyMetadataFactory
      * - From metadata of the given operation ("collection_operation_name" and "item_operation_name" keys).
      * - From metadata of the current resource.
      *
-     *
      * @return (string[]|null)[]
      */
     private function getEffectiveSerializerGroups(array $options, string $resourceClass): array
     {
         if (isset($options['serializer_groups'])) {
-            return [$options['serializer_groups'], $options['serializer_groups']];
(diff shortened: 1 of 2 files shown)


src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -160,13 +161,15 @@ private function getEffectiveSerializerGroups(array $options, string $resourceCl
             $denormalizationContext = $resourceMetadata->getAttribute('denormalization_context');
         }
 
-        return [$normalizationContext['groups'] ?? null, $denormalizationContext['groups'] ?? null];
+        return [
+            isset($normalizationContext['groups']) ? (array) $normalizationContext['groups'] : null,
```
alanpoulain: Why having a variable before and not here?

dunglas: Because previously we can reuse the var, here we can't.

teohhanhui:
It seems this PR broke the build. :laughing:

antograssiot:
no @teohhanhui, the broken build cames from Symfony release (see #2350)

PR #2350, opened 2018-11-27
Add webonyx/graphql-php version 0.13 compatibility

antograssiot:
~~waiting for https://github.com/symfony/symfony/pull/29350~~

| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Allow the test suite to run normaly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
         "symfony/validator": "^3.3 || ^4.0",
         "symfony/web-profiler-bundle": "^3.3 || ^4.0",
         "symfony/yaml": "^3.3 || ^4.0",
-        "webonyx/graphql-php": ">=0.12 <1.0"
+        "webonyx/graphql-php": ">=0.13 <1.0"
     },
     "conflict": {
         "symfony/dependency-injection": "<3.4"
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
     ],
     "require": {
         "php": ">=7.1",
-
         "doctrine/inflector": "^1.0",
         "psr/cache": "^1.0",
         "psr/container": "^1.0",
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -20,6 +20,8 @@ parameters:
         - '#Parameter \#1 \$function of function call_user_func expects callable, .+ given\.#'
         - '#Parameter \#1 \$classes of class ApiPlatform\\Core\\Metadata\\Resource\\ResourceNameCollection constructor expects array<string>, array<int, int\|string> given\.#'
         - '#Method ApiPlatform\\Core\\Util\\RequestParser::parseRequestParams\(\) should return array but returns array\|false\.#'
+        - '#Parameter \#1 \$vars of class GraphQL\\Language\\AST\\(IntValue|ObjectField|ObjectValue|BooleanValue|ListValue|StringValue)Node constructor expects array<GraphQL\\Language\\AST\\Location\|GraphQL\\Language\\AST\\NameNode\|GraphQL\\Language\\AST\\NodeList\|GraphQL\\Language\\AST\\SelectionSetNode\|string\|null>, array<string, .+> given\.#'
+        - '#Parameter \#1 \$defaultContext of class Symfony\\Component\\Serializer\\Encoder\\Json(De|En)code constructor expects array, (int|true) given\.#'
         # Temporary fix while the PHPStan extension for Prophecy isn't compatible with 0.10
         - '#Parameter .* expects .*, .*object.* given\.#'
         - '#Parameter \#[0-9] \$filterLocator of class ApiPlatform\\Core\\Bridge\\Doctrine\\Orm\\Extension\\FilterExtension constructor expects ApiPlatform\\Core\\Api\\FilterCollection|Psr\\Container\\ContainerInterface(\|null)?, ArrayObject given\.#'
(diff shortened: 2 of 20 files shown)


antograssiot:
I'll have a look to the other PHPStan issue later if we are going to support several version of `webonyx/graphql-php` as I started.

alanpoulain:
Wouldn't it be better to force `webonyx/graphql-php` to be >=0.13?

alanpoulain:
The previous update was done by forcing the next version.
It's because the GraphQL part of API Platform is an experimental feature. It allows us to do some breaking changes.
In this case, I am not sure about what to do. The code will be less readable with the `method_exists` but it's mainly in the tests.
WDYT @api-platform/core-team?

bendavies:
ah that's the issue i raised:
https://github.com/symfony/symfony/issues/29340

what problem is that causing in api-platform?

dunglas:
Regarding GraphQL, I'm all for bumping the version. It is still marked as `@experimental` anyway, and it'll stay like that until webonyx/graphql releases a 1.0 version.

antograssiot:
ok I'll update tomorrow

antograssiot:
@bendavies yes that's the issue you raisd. I broke [this test](https://travis-ci.org/api-platform/core/jobs/459908728#L689) (`children` was an empty array) it also broke nelmio fixture loading in some testsuite for example (that's how I discovered it in the first place)

@alanpoulain I've bump `webonyx/graphql-php` as discussed

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -204,7 +204,7 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
             }
 
             $graphqlWrappedType = $graphqlType instanceof WrappingType ? $graphqlType->getWrappedType() : $graphqlType;
-            $isInternalGraphqlType = \in_array($graphqlWrappedType, GraphQLType::getInternalTypes(), true);
+            $isInternalGraphqlType = \in_array($graphqlWrappedType, GraphQLType::getStandardTypes(), true);
```
alanpoulain: Shouldn't we rename the variables as well?

phpstan.neon
```diff
@@ -44,3 +44,7 @@ parameters:
         - '#Method Symfony\\Component\\Serializer\\Normalizer\\AbstractObjectNormalizer::__construct\(\) invoked with 6 parameters, 0-4 required\.#'
         - '#Parameter \#1 \$bitmask of class Symfony\\Component\\Serializer\\Encoder\\JsonEncode constructor expects int, array<string, int> given\.#'
         - '#Parameter \#1 \$associative of class Symfony\\Component\\Serializer\\Encoder\\JsonDecode constructor expects bool, array<string, true> given\.#'
+
+        # Expected, due to fixture detection in webonyx/graphql-php
+        - '#Parameter \#2 \$errors of class GraphQL\\Executor\\ExecutionResult constructor expects array<GraphQL\\Error\\Error>, array<int\|string, Exception\|string> given\.#'
```
alanpoulain: I'm not sure if this line:
https://github.com/api-platform/core/blob/master/src/GraphQl/Action/EntrypointAction.php#L67
is still working.
I think it should be interesting to add a test.

antograssiot: indeed, removing it doesn't break any existing tests IIRC so I wasn't sure.
I'll dig into that

phpstan.neon
```diff
@@ -44,3 +44,7 @@ parameters:
         - '#Method Symfony\\Component\\Serializer\\Normalizer\\AbstractObjectNormalizer::__construct\(\) invoked with 6 parameters, 0-4 required\.#'
         - '#Parameter \#1 \$bitmask of class Symfony\\Component\\Serializer\\Encoder\\JsonEncode constructor expects int, array<string, int> given\.#'
         - '#Parameter \#1 \$associative of class Symfony\\Component\\Serializer\\Encoder\\JsonDecode constructor expects bool, array<string, true> given\.#'
+
+        # Expected, due to fixture detection in webonyx/graphql-php
```
alanpoulain: fixture detection?

antograssiot:
yeah, I just ran out of time this morning @alanpoulain

composer.json
```diff
@@ -20,7 +20,7 @@
         "psr/container": "^1.0",
         "symfony/http-foundation": "^3.1 || ^4.0",
         "symfony/http-kernel": "^2.7 || ^3.0 || ^4.0",
-        "symfony/property-access": "^2.7 || ^3.0 || ^4.0",
+        "symfony/property-access": "^2.7 || ^3.0 <3.4.19 || ^4.0 <4.1.8",
```
Taluu: as there has been a fix, shouldn't these be declared as conflict ?

antograssiot: I'll just remove this a 4.2 is here anyway

antograssiot:
Tests added @alanpoulain 
I've also removed the version constraint update for Symfony and update the PHPStan rule to follow the Symfony 4.2 release
It should be good this time

i did cleanup some old `@dropSchema` annotation that were still there at the same time

composer.json
Taluu: Yeah but these 2 versions still conflicts somehow. Or you can categorize them as "yup doesn't work we know" ?

alanpoulain:
Thank you @antograssiot :slightly_smiling_face:

PR #2354, opened 2018-11-29
Fix SerializeListener to respect _api_respond request attribute

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A


teohhanhui:
Hmm... Would this break `AddTagsListener`? Lol...

src/EventListener/SerializeListener.php
```diff
@@ -47,7 +47,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
         $controllerResult = $event->getControllerResult();
         $request = $event->getRequest();
 
-        if ($controllerResult instanceof Response) {
+        if ($controllerResult instanceof Response || !$request->attributes->get('_api_respond')) {
```
antograssiot: shouldn't we keep the actual default behavior like so :
```suggestion
        if ($controllerResult instanceof Response || !$request->attributes->get('_api_respond', true)) {
```

teohhanhui:
Actually, I think it should also be added here: https://github.com/api-platform/core/blob/4f12b21fa383693bb1245c2e87618b99473a8bb7/src/Util/AttributesExtractor.php#L63-L66

tests/EventListener/SerializeListenerTest.php
```diff
@@ -42,61 +42,25 @@ public function testDoNotSerializeResponse()
...
 
         $serializerContextBuilderProphecy = $this->prophesize(SerializerContextBuilderInterface::class);
-        $serializerContextBuilderProphecy->createFromRequest()->shouldNotBeCalled();
+        $serializerContextBuilderProphecy->createFromRequest(Argument::cetera())->shouldNotBeCalled();
 
         $listener = new SerializeListener($serializerProphecy->reveal(), $serializerContextBuilderProphecy->reveal());
         $listener->onKernelView($eventProphecy->reveal());
     }
 
-    public function testDoNotSerializeWhenFormatNotSet()
```
teohhanhui: These tests were removed as they are outdated (and incorrect anyway...)

tests/EventListener/SerializeListenerTest.php
```diff
@@ -32,7 +32,7 @@ class SerializeListenerTest extends TestCase
     public function testDoNotSerializeResponse()
     {
         $serializerProphecy = $this->prophesize(SerializerInterface::class);
-        $serializerProphecy->serialize()->shouldNotBeCalled();
+        $serializerProphecy->serialize(Argument::cetera())->shouldNotBeCalled();
```
teohhanhui: We should check our code base for other instances of the same problem...

alanpoulain:
We shouldn't forget to merge 2.3 in 2.4.

PR #2358, opened 2018-12-01
Graphql: item normalizer improvement

Siregacvek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

When I explored why Graphql is so slow in my project, I noticed that normalizer has significant problem - it is serializing passed object to string here:
https://github.com/api-platform/core/blob/822c78c946db1fc037eaf6e11a2b2faaa023b72b/src/GraphQl/Serializer/ItemNormalizer.php#L43
It's a very expensive operation in real systems because object graph can be very big. Especially this is expensive when the object must be deserialized from this field later. Also I noticed, that api-platform really doesn't need full object deserialized from this field - it needs only class and identifier.

So my PR fixes this problem - I do minimal copy of object. In my project it speed up system in many times.

### DIFF
--- a/src/GraphQl/Serializer/ItemNormalizer.php
+++ b/src/GraphQl/Serializer/ItemNormalizer.php
@@ -93,7 +93,7 @@ final class ItemNormalizer extends BaseItemNormalizer
 	 * Return object of passed type with all empty fields except id.
 	 * Necessary to speed up serialization/deserialization on Webonyx side.
 	 *
-	 * @param $originalObject
+	 * @param object $originalObject
 	 *
 	 * @return object
 	 */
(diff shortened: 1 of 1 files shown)


antograssiot:
[Test failure](https://travis-ci.org/api-platform/core/jobs/462222796#L736) seems related.
Thats likely because you are not preserving composite identifiers used [here](https://github.com/api-platform/core/blob/master/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php#L85) for example

dunglas:
It would be cleaner to use some kind of proxy isn'it? The best fix would be to avoid calling `serialize` entirely, because it shouldn't be needed at all.

Siregacvek:
May be the most simple way is to attach 2 fields to item (instead of #item): #class and #identifier?

dunglas:
@Siregacvek it would deserve a try, indeed. WDYT @alanpoulain?

alanpoulain:
Closed in favor of: https://github.com/api-platform/core/pull/2576.
Thank you @Siregacvek for finding and suggesting this performance improvement :slightly_smiling_face:

PR #2360, opened 2018-12-03
Fix update-composer in CircleCI configuration

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ reusable-steps:
   - &update-composer
     run:
       name: Update Composer
-      command: composer self-update
+      command: sudo composer self-update
   - &update-project-dependencies
     run:
       name: Update project dependencies


teohhanhui:
`sudo`, in a Docker container? I'm not sure what's going on here...

teohhanhui:
Maybe we should use our own Docker images for the CircleCI build, so that we could have more control and avoid weird issues like this in the future...

teohhanhui:
Thanks @alanpoulain! :1st_place_medal:

er1z:
@teohhanhui — `sudo` on composer update has to be used if its binary is outside current user scope. May be a workaround of installing under home dir + altering `PATH` could be a solution.

PR #2367, opened 2018-12-05
Update PHP-CS-Fixer and tighten CS

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no? maybe?
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/GraphQl/Resolver/ResourceAccessCheckerTrait.php
+++ b/src/GraphQl/Resolver/ResourceAccessCheckerTrait.php
@@ -11,15 +11,6 @@
 
 declare(strict_types=1);
 
-/*
- * This file is part of the API Platform project.
- *
- * (c) Kévin Dunglas <dunglas@gmail.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 namespace ApiPlatform\Core\GraphQl\Resolver;
 
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
@@ -39,7 +30,7 @@ trait ResourceAccessCheckerTrait
     /**
      * @throws Error
      */
-    public function canAccess(ResourceAccessCheckerInterface $resourceAccessChecker = null, ResourceMetadata $resourceMetadata, string $resourceClass, ResolveInfo $info, $object = null, string $operationName = null)
+    public function canAccess(ResourceAccessCheckerInterface $resourceAccessChecker = null, ResourceMetadata $resourceMetadata, string $resourceClass, ResolveInfo $info, $object = null, string $operationName = null): void
     {
         if (null === $resourceAccessChecker) {
             return;
--- a/src/Hal/Serializer/ItemNormalizer.php
+++ b/src/Hal/Serializer/ItemNormalizer.php
@@ -221,7 +221,6 @@ final class ItemNormalizer extends AbstractItemNormalizer
     /**
      * Gets the cache key to use.
      *
-     *
      * @return bool|string
      */
     private function getHalCacheKey(string $format = null, array $context)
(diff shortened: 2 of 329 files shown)


src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -81,6 +81,8 @@ public function supportsDenormalization($data, $type, $format = null)
      * {@inheritdoc}
      *
      * @throws RuntimeException
+     *
+     * @return object
```
teohhanhui: This is the only weird thing in this PR. Lol...

Because PHP-CS-Fixer only knows about what's in the current file. (It does not parse parent classes / interfaces.)

And of course when we bump to PHP 7.2, we can add the `object` return type.

src/GraphQl/Resolver/ResourceAccessCheckerTrait.php
```diff
@@ -11,15 +11,6 @@
 
 declare(strict_types=1);
 
-/*
- * This file is part of the API Platform project.
- *
- * (c) Kévin Dunglas <dunglas@gmail.com>
```
jewome62: Move up the header license message than `declare(strict_types=1);`

src/Security/ResourceAccessChecker.php
```diff
@@ -11,15 +11,6 @@
 
 declare(strict_types=1);
 
-/*
- * This file is part of the API Platform project.
- *
- * (c) Kévin Dunglas <dunglas@gmail.com>
- *
```
jewome62: move this message over `declare(strict_types=1);`

src/GraphQl/Resolver/ResourceAccessCheckerTrait.php
teohhanhui: It was duplicated. The copyright header was already there. :laughing:

src/Security/ResourceAccessChecker.php
teohhanhui: Same.

antograssiot:
I'm a bit concerned that adding `:void` return type everywhere will force user to update their own code too. It could be considered as a BC break in  filters for example as those classes are not final, anyone that extends/use a filter or the AbstractFilter as (we describe in the doc)[https://api-platform.com/docs/core/filters/#creating-custom-doctrine-orm-filters] will have to update their code.

teohhanhui:
There was no `@return` in the PHPDoc, so we're not changing the contract at all. It's only that we're letting PHP enforce it strictly now.

antograssiot:
I still think that people will get fatal errors in their app after updating based on https://3v4l.org/iN5nN
that's not a great DX, am I missing something ?

soyuka:
I tend to agree with @antograssiot could we skip this optimisation for filters so that the upgrade path keeps being smooth for our user base?

soyuka:
> But if it's a concern, I could postpone adding the void_return rule until 2.4. So that the user is at least likely to read the release notes. WDYT?

:+1: totally

teohhanhui:
Hmm... Actually I think it'd just create a lot more maintenance issues if we add `void` return for some but not all, or if we do it in `master`. (It might cause more merging conflicts?)

So for now, I'm sticking to merging this into `2.3`.

soyuka:
can you just remove the void_return setting though?

teohhanhui:
> can you just remove the void_return setting though?

As I've mentioned above:

> I think it'd just create a lot more maintenance issues if we add void return for some but not all (hard to enforce the CS)

`php-cs-fixer` check would fail when we merge `2.3` into `master` (assuming we turn it on in `master`).

teohhanhui:
Closing as it's a BC break.

PR #2371, opened 2018-12-07
Correctly expose properties[] as a collection parameter

### LINKED ISSUE #2370: PropertyFilter not exposed as collection
When applying PropertyFilter to an Entity, properties is exposed as a simple string, whereas it should be a string collection. As of now implementation is broken, ie, you can't ask for more than one property if one were to adhere to the documentation.

It's now detailed in the documentation as:
`{"name":"properties[]","in":"query","required":false,"type":"string"}`

Should be:
`{"name":"properties[]","in":"query","required":false,"type":"array","items":{"type":"string"},"collectionFormat":"multi"}`

idsplus-carlos:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2370
| License       | MIT

This fixes #2370 in which properties[] is not documented as an array of strings as expected.

### DIFF
--- a/src/Serializer/Filter/GroupFilter.php
+++ b/src/Serializer/Filter/GroupFilter.php
@@ -69,6 +69,7 @@ final class GroupFilter implements FilterInterface
             "$this->parameterName[]" => [
                 'property' => null,
                 'type' => 'string',
+                'is_collection' => true,
                 'required' => false,
             ],
         ];
--- a/src/Serializer/Filter/PropertyFilter.php
+++ b/src/Serializer/Filter/PropertyFilter.php
@@ -71,6 +71,7 @@ final class PropertyFilter implements FilterInterface
             "$this->parameterName[]" => [
                 'property' => null,
                 'type' => 'string',
+                'is_collection' => true,
                 'required' => false,
             ],
         ];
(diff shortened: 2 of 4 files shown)


antograssiot:
Would you mind update the GroupFilter too ?

idsplus-carlos:
Sure, there you go.

However, i'm not sure the fix for properties[] doesn't need additional work, but i'm unsure about how to solve the problem.

Current PR fixes properties[] as a string collection. However, the documentation states that it should accept both string[] and string[key][] for relations, but i've been reading the swagger specs and unless i'm wrong it seems to indicate that it should not be possible with swagger, ie, we have 2 types for the same parameter (string array and array of string arrays).
If that's the case, being strict we would need to create an additional parameter, like relation_properties[] in order to be able to have a correct swagger documentation.

dunglas:
I think that indeed it's not possible to document it properly with Swagger, but not a big deal IMHO (and it doesn't worth introducing a new property).

dunglas:
Thanks @idsplus-carlos!

PR #2372, opened 2018-12-07
Tighten CS without BC breaks

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no? maybe?
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Replaces #2367


src/Bridge/Doctrine/Orm/Extension/ContextAwareQueryItemExtensionInterface.php
```diff
@@ -1,28 +0,0 @@
-<?php
```
teohhanhui: Deleted as it was a (wrong) duplicate of https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Extension/QueryResultItemExtensionInterface.php, accidentally added in https://github.com/api-platform/core/pull/1619

src/DataPersister/ChainDataPersister.php
```diff
@@ -63,7 +63,9 @@ public function remove($data)
     {
         foreach ($this->persisters as $persister) {
             if ($persister->supports($data)) {
-                return $persister->remove($data);
+                $persister->remove($data);
+
+                return;
```
teohhanhui: The return type should be strict. We do not add a `void` return type for now as that'd be a BC break.

tests/Bridge/Doctrine/EventListener/WriteListenerTest.php
```diff
@@ -108,20 +108,23 @@ public function testOnKernelViewWithNoResourceClass()
...
         $event = new GetResponseForControllerResultEvent($httpKernelProphecy->reveal(), $request, HttpKernelInterface::MASTER_REQUEST, $dummy);
 
-        $this->assertNull($writeListener->onKernelView($event));
-        $this->assertNotEquals($dummy, $writeListener->onKernelView($event));
+        $writeListener->onKernelView($event);
     }
 
     /**
      * @expectedDeprecation The ApiPlatform\Core\Bridge\Doctrine\EventListener\WriteListener class is deprecated since version 2.2 and will be removed in 3.0. Use the ApiPlatform\Core\EventListener\WriteListener class instead.
+     * @doesNotPerformAssertions
```
antograssiot: Isn't this missing a `@group legacy` instead ?

composer.json
```diff
@@ -68,6 +68,7 @@
         "webonyx/graphql-php": ">=0.13 <1.0"
     },
     "conflict": {
+        "doctrine/common": "<2.7",
```
antograssiot: Why did you need this ?
We still require `^2.5.2` which could also be updated.

teohhanhui: This is `doctrine/common`, not `doctrine/orm`. See https://github.com/doctrine/common/pull/734

tests/Bridge/Doctrine/EventListener/WriteListenerTest.php
teohhanhui: The `@group legacy` is already on the class. And `@doesNotPerformAssertions` is not related to that.

composer.json
antograssiot: My bad

soyuka:
thanks @teohhanhui !

Issue #2090, opened 2018-07-11
RFC: Filters do not respect  the configured name converter.


bendavies:
Hey all,

Here's an issue I discovered today.
If your API has a configured name converter, (let's say `CamelCaseToSnakeCaseNameConverter`), filtering on fields is unintuitive for the consumer.

Given an ApiResource with a field called `createdAt`, `createdAt` will be serialized/deserialized to/from the consumer as `created_at`.

However, if the consumer wants to then filter on `createdAt`, they must use `?createdAt...`, which is very unintuitive as that is not how the field is represented.

The fix seems simple - make all filters respect the configured name converter.
However, this would be a big BC break, so will need to go in 3.0?

Cheers

dunglas:
I would be in favor of fixing this. It looks totally unexpected on my side. If someone complains... we'll provide a fix like a flag or something like that to restore the buggy behavior. WDYT?

bendavies:
Cool, while obviously a bug, I didn't expect a fix to be prioritized over the BC break. happy with that!

soyuka:
>  I didn't expect a fix to be prioritized over the BC break

Most of us are using the default name converter, therefore nothing will change.

teohhanhui:
Was it ever claimed that the filters would use the name converter? :stuck_out_tongue:

I don't know if it could be considered a bug if it's a missing feature...

slebrequier:
Sorry to have opened a duplicate of this issue.
I, for one, think of it as a bug since the client have to filter from what he knows : the response body.
If the field is called `created_at` in the response body, then it must be `created_at` that will filter or sort a list of results.

Soullivaneuh:
@dunglas To avoid BC break, we could provide a `api_platform.name_generator` parameter and deprecate the `api_platform.path_segment_name_generator` one.

Soullivaneuh:
I really need this fix and would like to help making it real, but I don't even know where to start. Could you please give us some clues? Thanks!

dunglas:
To me it’s a bug fix, so don’t bother with BC because the current behavior is not correct (this case has just been forgotten).

A PR would be very appreciated, just target master instead of 2.3 to limit the consequences in case someone was relying on the buggy behavior.

Soullivaneuh:
@dunglas Do you have any clues to give about how to fix it?

antograssiot:
[...]

PR #2377, opened 2018-12-12
Use the name converter for filters

soullivaneuh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes (but not considered as [it's a bugfix](https://github.com/api-platform/core/issues/2090#issuecomment-435949533))
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2090
| License       | MIT
| Doc PR        | N/A

I found a simpler way than https://github.com/api-platform/core/issues/2090#issuecomment-436060295 by normalize names on documentation normalizer and denormalize them on the read listener.

WDYT? cc @antograssiot

### DIFF
--- a/src/EventListener/ReadListener.php
+++ b/src/EventListener/ReadListener.php
@@ -26,6 +26,7 @@ use ApiPlatform\Core\Util\RequestParser;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
 /**
  * Retrieves data from the applicable data provider and sets it as a request parameter called data.
@@ -37,14 +38,16 @@ final class ReadListener
     use OperationDataProviderTrait;
 
     private $serializerContextBuilder;
+    private $nameConverter;
 
-    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null)
+    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null, ?NameConverterInterface $nameConverter = null)
     {
         $this->collectionDataProvider = $collectionDataProvider;
         $this->itemDataProvider = $itemDataProvider;
         $this->subresourceDataProvider = $subresourceDataProvider;
         $this->serializerContextBuilder = $serializerContextBuilder;
         $this->identifierConverter = $identifierConverter;
+        $this->nameConverter = $nameConverter;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/EventListener/ReadListener.php
```diff
@@ -64,7 +67,11 @@ public function onKernelRequest(GetResponseEvent $event)
 
         if (null === $filters = $request->attributes->get('_api_filters')) {
             $queryString = RequestParser::getQueryString($request);
-            $filters = $queryString ? RequestParser::parseRequestParams($queryString) : null;
+            if ($queryString) {
+                foreach (RequestParser::parseRequestParams($queryString) as $name => $value) {
+                    $filters[$this->nameConverter->denormalize($name)] = $value;
```
Soullivaneuh: Need a condition if the name converter is not provided.

src/EventListener/ReadListener.php
```diff
@@ -64,7 +67,12 @@ public function onKernelRequest(GetResponseEvent $event)
 
         if (null === $filters = $request->attributes->get('_api_filters')) {
             $queryString = RequestParser::getQueryString($request);
-            $filters = $queryString ? RequestParser::parseRequestParams($queryString) : null;
+            if ($queryString) {
+                foreach (RequestParser::parseRequestParams($queryString) as $name => $value) {
+                    $name = $this->nameConverter ? $this->nameConverter->denormalize($name) : $name;
```
antograssiot: Wouldn't this apply to every parameters like `itemsPerPage` or custom PropertyFilter / GroupFilter parameter name, and thus create an unexpected app behavior change and/or add more discrepencies between the app and the Swagger documentation ?

antograssiot: `itemsPerPage` is indeed not a good example, even if it will be present here, it doesn't matter much.
The point about Group Filter (or any other filter where the name can be defined in the definition) for example is that one could declare explicitly the parameter name as `group_filter_parameter_name` and it might not be respected and transformed to `groupFilterParameterName` for example. I'm not sure it is a  big deal, just wondering if it might be misleading.

Soullivaneuh: @antograssiot Mayde do you have a proof test to give me so I can add it on the PR?

antograssiot: @Soullivaneuh 
here are the easy downside of your approach:
- First you only normalize keys so ExistsFilter or OrderFilter that uses `foo[paramName]` will not be taken into account
- The hydra:mapping where the filter description leaves doesn't reflect those changes. Maybe updating `ApiPlatform\Core\Hydra\Serializer\CollectionFiltersNormalizer` could fix that

You can use the Dummy entity in the test suite, it has a `$nameConverted` property that is transformed via the `ApiPlatform\Core\Tests\Fixtures\TestBundle\Serializer\NameConverter\CustomConverter`.

You can ping me on slack for this topic if you need quick help/chat

antograssiot:
Closing, #2751 has landed in master 
Thanks @soullivaneuh

soullivaneuh:
@antograssiot I'm not sure to understand: This MR is merged onto master and not on any new release so far.

But it was considered as a [bugfix](https://github.com/api-platform/core/issues/2090#issuecomment-435949533) by @dunglas. Any chance to have this cherry-picked on a stable release?

Or I'm missing something. :-)

soyuka:
@soullivaneuh the problem is that this patch changes the behavior of existing code. Therefore, we can't put this in a patch release or it'll break implementations.

soullivaneuh:
@soyuka I know, this is exactly what @dunglas said, but also that the current behavior was not intended from the beginning and can be released as a bugfix.

This is why I was surprised, but I totally understand your point of view.

In that case, maybe this can be activated and the current stable with an option?

soyuka:
I know :|. Not really because it'd be an option that will be removed later and therefore would need to be deprecated, lot of work.

soullivaneuh:
Understood, I'll wait the next major then. :+1: 

BTW, do you have any date/plan about that?

Regards

PR #2381, opened 2018-12-14
Merge 2.3 into master

teohhanhui:


teohhanhui:
Failing test on AppVeyor is due to the time-sensitive nature of the `Response` object. It should be fixed in another PR.

All of these could fail the same way: https://github.com/api-platform/core/search?q=new+Response+path%3Atests&unscoped_q=new+Response+path%3Atests

Issue #2363, opened 2018-12-04
Documentation generator should be aware of default property values


soullivaneuh:
Example witht this property:

```php
private $externalMysql = false;
```

The documentation will display this example:

```json
"external_mysql": true,
```

Not a big deal, but it would be better to show `false` as it's the default value of `$externalMysql`.

PR #2386, opened 2018-12-16
Use default value of PHP properties to Swagger doc

aaa2000:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2363
| License       | MIT
| Doc PR        | 

Add default and example information to swagger context if php properties
have default values so the exemple documentation section use these values.

With a class

```
<?php

namespace App\Entity;

/**
 * @ORM\Entity
 * @ApiResource(
 *     normalizationContext={"groups": {"person:read"}}
 * )
 * @ApiResource
 */
class Person
{
    /**
     * @var UuidInterface
     *
     * @ApiProperty(identifier=true)
     * @Groups({"person:read"})
     *
     * @ORM\Id
     * @ORM\Column(type="uuid", unique=true)
     * @ORM\CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidGenerator")
     */
    public $code;

    /**
     * @var bool
     *
     * @Groups({"person:read", "person:write"})
     *
     * @ORM\Column(type="boolean", options={"default" : false})
     */
    public $isActive = true;

    /**
     * @var integer
     *
     * @Groups({"person:read", "person:write"})
     *
     * @ORM\Column(type="integer", nullable=true)
     */
    public $int = 100;

    /**
     * @var array
     *
     * @Groups({"person:read", "person:write"})
     *
     * @ORM\Column(type="json_array", nullable=true)
     */
    public $array = ['aa', 'bb'];

    /**
     * @var string
     *
     * @ApiProperty(
     *     attributes={
     *         "swagger_context"={
     *             "example"="example string"
     *         }
     *     }
     * )
     *
     * @Groups({"person:read", "person:write"})
     *
     * @ORM\Column(nullable=true)
     */
    public $string = 'aaaaaa';
}
```

The swagger documentation displays

api_platform_default_value

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
@@ -74,6 +74,10 @@
             <argument type="service" id="api_platform.metadata.property.metadata_factory.cached.inner" />
         </service>
 
+        <service id="api_platform.metadata.property.metadata_factory.default_property" decorates="api_platform.metadata.property.metadata_factory" decoration-priority="30" class="ApiPlatform\Core\Metadata\Property\Factory\DefaultPropertyMetadataFactory" public="false">
+            <argument type="service" id="api_platform.metadata.property.metadata_factory.default_property.inner" />
+        </service>
+
         <service id="ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface" alias="api_platform.metadata.property.metadata_factory" />
 
         <!-- Cache -->
--- a/src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
@@ -48,7 +48,7 @@ final class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInt
 
         $defaultProperties = $reflectionClass->getDefaultProperties();
 
-        if (!array_key_exists($property, $defaultProperties) || null === ($defaultProperty = $defaultProperties[$property])) {
+        if (!\array_key_exists($property, $defaultProperties) || null === ($defaultProperty = $defaultProperties[$property])) {
             return $propertyMetadata;
         }
 
--- a/src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
@@ -16,7 +16,10 @@ namespace ApiPlatform\Core\Metadata\Property\Factory;
 use ApiPlatform\Core\Exception\PropertyNotFoundException;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 
-class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInterface
+/**
+ * Populates defaults values of the ressource properties using the default PHP values of properties.
+ */
+final class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInterface
 {
     private $decorated;
 
@@ -45,16 +48,10 @@ class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInterface
 
         $defaultProperties = $reflectionClass->getDefaultProperties();
 
-        if (array_key_exists($property, $defaultProperties) && null !== ($defaultProperty = $defaultProperties[$property])) {
-            $attributes = $propertyMetadata->getAttributes() ?? [];
-            $attributes['swagger_context']['default'] = $defaultProperty;
-            if (!isset($attributes['swagger_context']['example'])) {
-                $attributes['swagger_context']['example'] = $defaultProperty;
-            }
-
-            return $propertyMetadata->withAttributes($attributes);
+        if (!array_key_exists($property, $defaultProperties) || null === ($defaultProperty = $defaultProperties[$property])) {
+            return $propertyMetadata;
         }
 
-        return $propertyMetadata;
+        return $propertyMetadata->withDefault($defaultProperty);
     }
 }
--- a/src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Property\Factory;
+
+use ApiPlatform\Core\Exception\PropertyNotFoundException;
+use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
+
+class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInterface
+{
+    private $decorated;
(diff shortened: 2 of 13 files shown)


src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Property\Factory;
+
+use ApiPlatform\Core\Exception\PropertyNotFoundException;
+use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
+
+class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInterface
```
dunglas:
```suggestion
final class DefaultPropertyMetadataFactory implements PropertyMetadataFactoryInterface
```

dunglas: Can you also add some PHPDoc please?

src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
```diff
@@ -0,0 +1,60 @@
...
+
+        try {
+            $reflectionClass = new \ReflectionClass($resourceClass);
+        } catch (\ReflectionException $reflectionException) {
+            return $propertyMetadata;
+        }
+
+        $defaultProperties = $reflectionClass->getDefaultProperties();
+
+        if (array_key_exists($property, $defaultProperties) && null !== ($defaultProperty = $defaultProperties[$property])) {
```
dunglas: Could be changed to a guard clause for clarity:

```php
if (!array_key_exists($property, $defaultProperties) || null === ($defaultProperty = $defaultProperties[$property])) {
     return $propertyMetadata;
}
```

src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
```diff
@@ -0,0 +1,60 @@
...
+            $attributes = $propertyMetadata->getAttributes() ?? [];
+            $attributes['swagger_context']['default'] = $defaultProperty;
```
dunglas: The default value could be useful in other formats (Hydra for instance). I would introduce a new, dedicated, attribute for this. The other benefit of doing this is that it would be possible to set a default value that is different than the one of the property by using the `@ApiProperty(default="foo bar")` annotation.

src/Metadata/Property/Factory/DefaultPropertyMetadataFactory.php
```diff
@@ -0,0 +1,60 @@
...
+            if (!isset($attributes['swagger_context']['example'])) {
+                $attributes['swagger_context']['example'] = $defaultProperty;
```
dunglas: Same here, I would introduce a dedicated property.

Soullivaneuh:
@dunglas Won't 2.3 be a better branch target for this?

dunglas:
@Soullivaneuh no because it's a new feature.

aaa2000:
@dunglas PR updated to introduce the `default` and `example` properties in ApiProperty

src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
```diff
@@ -74,6 +74,10 @@
             <argument type="service" id="api_platform.metadata.property.metadata_factory.cached.inner" />
         </service>
 
+        <service id="api_platform.metadata.property.metadata_factory.default_property" decorates="api_platform.metadata.property.metadata_factory" decoration-priority="30" class="ApiPlatform\Core\Metadata\Property\Factory\DefaultPropertyMetadataFactory" public="false">
```
aaa2000: `decoration-priority="30"` is correct ?

aaa2000:
@meyerbaptiste done

soyuka:
Should we also use the Doctrine default value in https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php?

aaa2000:
I also added the Doctrine default value but I'm not sure it's a good idea because with postgres, Doctrine insert a null value for the column, and there is the error 
`ERROR: null value in column "string" violates not-null constraint`

I can use Doctrine default value as example value but a doctrine annotation `@ORM\Column(name="created_at", type="datetime", options={"default": "CURRENT_TIMESTAMP"})` will display `CURRENT_TIMESTAMP` as example. Finally, I do not think that this is desirable to use the Doctrine default value

soyuka:
> I also added the Doctrine default value but I'm not sure it's a good idea because with postgres, Doctrine insert a null value for the column, and there is the error
> ERROR: null value in column "string" violates not-null constraint

Not sure I got that, what is behind this insert?

About the second point, thoughts @api-platform/core-team  ?

aaa2000:
Given an entity with Doctrine default value
```
<?php

namespace App\Entity;

/**
 * @ORM\Entity
 * @ApiResource
 */
class Bug
{
    /**
     * @var UuidInterface
     *
     * @ApiProperty(identifier=true)
     *
     * @ORM\Id
     * @ORM\Column(type="uuid", unique=true)
     * @ORM\CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidGenerator")
     */
    public $code;


    /**
     * @var string
     *
     * @ORM\Column(options={"default": "example name"})
     */
    public $name;
}

```
When I send the request `curl -X POST "http://localhost:8080/bugs" -H  "accept: application/ld+json" -H  "Content-Type: application/ld+json" -d "{}"`
Then I have the error 
```
{
  "@type": "hydra:Error",
  "hydra:description": "An exception occurred while executing 'INSERT INTO bugs (code, name) VALUES (?, ?)' with params [\"4387fe5d-ce66-4783-a310-33256cd9c294\", null]:\n\nSQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column \"name\" violates not-null constraint\nDETAIL:  Failing row contains (4387fe5d-ce66-4783-a310-33256cd9c294, null).",
```

To have the default value inserted with postgres, the query should be ``` INSERT INTO bugs (code, name) VALUES ('4387fe5d-ce66-4783-a310-33256cd9c294', default ); ```

So, it is maybe better to use the Doctrine default value as example instead of default property metadata and  change https://github.com/api-platform/core/pull/2386/files#diff-fbaaae4f33f38c0ec0e318994cb386faR82 ```$propertyMetadata = $propertyMetadata->withDefault($fieldMapping['options']['default']);``` by ``` $propertyMetadata = $propertyMetadata->withExample($fieldMapping['options']['default']); ```

soyuka:
:+1: if you can fix this

flug:
@aaa2000 you can fix et add test please for merge this pull request ?

aaa2000:
I can't fix, the fix should be done in doctrine. The doctrine ORM doesn't take into account the `options={"default": "example name"}` when it create the insertion. The options seems to be used only for creating the database schema

soyuka:
Okay nice, we'll open an issue to track this. I the meantime I'm :+1: to merge this.

soyuka:
Thanks for the work @aaa2000

teohhanhui:
> I can't fix, the fix should be done in doctrine. The doctrine ORM doesn't take into account the options={"default": "example name"} when it create the insertion. The options seems to be used only for creating the database schema

It's not a bug. That's exactly what it's intended for, the default value at the database level. It'd be a bug if Doctrine ORM actually tries to use it. :wink:

PR #2388, opened 2018-12-17
WIP: Use other keys than 'id' for associations in SearchFilter

TimoBakx:
I have an entity with a custom identifier (other than `$id`) and I'm using that in a many-to-many relationship with my main entity.
When using a SearchFilter to filter the main Entity, a query exception is thrown, because the `id` field does not exist.

This change will add a check in the doctrine metadata to figure out the correct field to use instead of `id`, with a fall-back to `id`.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

**TODO**
- [ ] Try to use identifiers instead of the doctrine metadata (they might differ)
- [X] Add unit test
- [ ] Add behat test
- [X] Test with a many-to-many relationship
- [x] Test with many-to-one relationship
- [ ] Test with a one-to-many relationship
- [ ] Test with a one-to-one relationship
- [ ] Test with multiple keys (maybe add a fallback?)
- [ ] Also use this method to retrieve the value (instead of a mandatory `getId()` method)


src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -270,7 +270,7 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
 
         if ($metadata->isCollectionValuedAssociation($association)) {
             $associationAlias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $alias, $association);
-            $associationField = 'id';
+            $associationField = ($metadata->getAssociationMapping($association)['joinTable']['inverseJoinColumns'][0]['referencedColumnName'] ?? 'id');
```
soyuka: Nice, this is indeed a long-time known bug, I'm not sure that the fix is that easy but if you make the tests pass that's already a good improvement!

antograssiot:
Thanks you very much @TimoBakx 
Closing I think that this was fixed in master in #1844 merged recently.

TimoBakx:
Perfect. Thank you. I'll keep an eye out for the release and test it against our current project to test it in a "live" environment.

PR #2389, opened 2018-12-17
Fix exclude app cache path for PHP-CS-Fixer

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

We don't notice the problem in our CI, because the PHP-CS-Fixer job is always separate.


dunglas:
Thanks @teohhanhui

PR #2391, opened 2018-12-18
Added swagger description to property filter

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I think this qualifies as a bugfix because the only thing it really only does is add some swagger docs to the `PropertyFilter`. That's not really something that has to wait for a new minor release imho 😄


dunglas:
Thanks @Toflar

PR #2392, opened 2018-12-18
Fix Doctrine DataPersister priority

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        |n/a

This allows to leverage the `autoconfigure` flag when creating custom data providers. Without this explicit priority, the default provider can be registered before custom ones, which is counter-intuitive and annoying.

This is a potential (tiny) BC break (it's why I not targeted 2.3), but I think it is worth it.


dunglas:
@antograssiot I'll add it to the changelog

PR #2393, opened 2018-12-18
Allow to disable input and output class

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

This PR allows to set the new `input_class` and `output_class` attributes to `false`. It hints API Platform that the corresponding operation(s) will respectively expect an empty input, or will return nothing.

This is especially convenient when creating an endpoint that will return nothing. My use case: I create have a `POST /reset-password` endpoint that sends a token by mail. This endpoint returns `202 Accepted` and nothing else.

The next step will be to support `input_class` and `output_class` (including this new option) in the documentation normalizers.


soyuka:
Good idea!

PR #2396, opened 2018-12-19
Allow to set the status code through an attribute

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

This patch allows to set the status code returned by an operation using a new attribute. Exemple:

```php
/**
 * @ApiResource(
 *     collectionOperations={
 *         "post"={"status"=202}
 *     },
 *     itemOperations={},
 *     outputClass=false
 * )
 *
 * @author Kévin Dunglas <dunglas@gmail.com>
 */
class ResetPasswordRequest
{
// ...
}
```

Will return a 202 Accepted instead of 201 Created.

PR #2398, opened 2018-12-19
Provides an integration with the Symfony Messenger component

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

This patch adds an easy way to dispatch messages to be handled asynchronously (or synchronously) using the Symfony Messenger component. It allows to create service-oriented endpoints very easily, but still in a restful way.

To be dispatched as a Symfony message, a resource just need to have be marked with `messenger` attribute:

```php
<?php

namespace Dunglas\UserBundle\Entity;

/**
 * @ApiResource(
*      messenger=true,
 *     collectionOperations={
 *         "post"={"status"=202}
 *     },
 *     itemOperations={},
 *     outputClass=false
 * )
 */
class ResetPasswordRequest
{
    /**
     * @var string
     *
     * @Assert\NotBlank
     */
    public $username;
}
```

The corresponding Message handler:

```php
<?php

namespace Dunglas\UserBundle\Handler;

use Symfony\Component\Messenger\Handler\MessageHandlerInterface;

final class ResetPasswordRequestHandler implements MessageHandlerInterface
{
    public function __invoke(ResetPasswordRequest $forgotPassword)
    {
        // do something with the resource
    }
}
```

It's an alternative to https://github.com/sroze/api-platform-messenger.


ragboyjr:
@dunglas With the default configuration in sf messenger, how would oneself make a message handler that handled the remove message?

ragboyjr:
Also, in the example posted above, would the swagger generated documentation properly reflect that no response is made?

src/DataPersister/RemoveStamp.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataPersister;
+
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
+/**
+ * Hints that the resource in the enveloppe must be removed.
```
borNfreee: `enveloppe` -> `envelope`

dunglas:
> @dunglas With the default configuration in sf messenger, how would oneself make a message handler that handled the remove message?

For removal, you'll need a middleware, a simple handler would just get the message.

> Also, in the example posted above, would the swagger generated documentation properly reflect that no response is made?

Not yet unfortunately, but it's something we're working on!

bendavies:
just a niggle but your example relies on https://github.com/api-platform/core/pull/2396.

https://github.com/krakphp/api-platform-extra also implemented this.

src/DataPersister/MessengerDataPersister.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return true === $this->resourceMetadataFactory->create($this->getObjectClass($data))->getAttribute('messenger');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function persist($data)
+    {
+        $this->messageBus->dispatch($data);
```
bendavies: not a return here?

src/DataPersister/MessengerDataPersister.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->messageBus = $messageBus;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports($data): bool
+    {
+        return true === $this->resourceMetadataFactory->create($this->getObjectClass($data))->getAttribute('messenger');
```
Toflar: Just double checking here: Are you sure you want to keep the `messenger` attribute a boolean? Or can you think of anything we might want to be able to configure at some point in the future? If so, I think we should turn that into an array for forward compat :)

src/DataPersister/RemoveStamp.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
Toflar: Feels like this should rather go into `DataPersister/Messenger/RemoveStamp`. It's not a data persister itself. Wdyt?

src/DataPersister/MessengerDataPersister.php
Toflar: E.g. custom attributes for a general `AttributesStamp` on the envelope?  That way we could just pass arbitrary attributes with it that can be used in the message handlers.

src/DataPersister/MessengerDataPersister.php
soyuka: It needs to return the data: https://github.com/api-platform/core/blob/master/src/DataPersister/ChainDataPersister.php#L55 dispatch returns an envelope IIRC

src/DataPersister/RemoveStamp.php
soyuka: Following symfony hierarchy I'd even say Stamp/RemoveStamp

src/DataPersister/MessengerDataPersister.php
dunglas: We'll be able to change that in the future without introducing a BC break.

src/DataPersister/RemoveStamp.php
dunglas: src/Bridge/Symfony/Messenger ^^

src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function persist($data)
+    {
+        $this->messageBus->dispatch($data);
+
+        return $data;
```
sroze: I'd return the `result` of the `HandledStamp` instead if it exists.

src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->messageBus = $messageBus;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports($data): bool
+    {
+        return true === $this->resourceMetadataFactory->create($this->getObjectClass($data))->getAttribute('messenger');
```
teohhanhui: Could we have created a new `ContextAwareDataPersisterInterface`, so that the `messenger` attribute could be set at the operation level?

PR #2399, opened 2018-12-19
fix #2158 - denormalize items without context and no Iri

### LINKED ISSUE #2158: The denormalize method in the ItemNormalizer does not set the resource_class correctly.
**Stack trace:**
In ItemNormalizer.php line 54:
                                         
  [ErrorException]                         
  Notice: Undefined index: resource_class  
                                           

Exception trace:
 ApiPlatform\Core\Serializer\ItemNormalizer->updateObjectToPopulate() at /var/www/environments/vendor/api-platform/core/src/Serializer/ItemNormalizer.php:42
 ApiPlatform\Core\Serializer\ItemNormalizer->denormalize() at /var/www/environments/vendor/symfony/serializer/Serializer.php:179
 Symfony\Component\Serializer\Serializer->denormalize() at /var/www/environments/vendor/symfony/serializer/Normalizer/AbstractNormalizer.php:370
 Symfony\Component\Serializer\Normalizer\AbstractNormalizer->instantiateObject() at /var/www/environments/vendor/symfony/serializer/Normalizer/AbstractObjectNormalizer.php:143
 Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer->instantiateObject() at /var/www/environments/vendor/symfony/serializer/Normalizer/AbstractObjectNormalizer.php:244
 Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer->denormalize() at /var/www/environments/vendor/symfony/serializer/Serializer.php:179
 Symfony\Component\Serializer\Serializer->denormalize() at /var/www/environments/vendor/symfony/serializer/Serializer.php:128
 Symfony\Component\Serializer\Serializer->deserialize() at /var/www/environments/vendor/symfony/messenger/Transport/Serialization/Serializer.php:54`

**What triggered this issue:**
Using API Platform in combination with the Symfony Message component and encoding an entity that is an APIPlatform resource.

**Root cause**
In the denormalize method in api-platform/core/src/Serializer/ItemNormalizer.php the resource_class is not set in the context. It seems this is common because it is handled in the AbstractItemNormalizer class line 123:

$context['api_denormalize'] = true;
     $context['resource_class'] = $class;
 }

Unfortunatly, this is called after we need it

            $this->updateObjectToPopulate($data, $context);  // The error is thrown in this method.
        }
        return parent::denormalize($data, $class, $format, $context); // Calls the method in the AbstractClass
 
**Potential solution**
Adding the code in the ItemNormalizer instead of the AbstractNormalizer seems to solve the issue.

rpanfili:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2158 
| License       | MIT
| Doc PR        |  - 

This PR fix an issue when `context` does not contain the `resource_class` key.

It's the same behaviour of the `AbstractItemNormalizer` that _inherit_ that option from the `$class` parameter before calling the `parent::denormalize` method. It actually has to be applied before try to update the object with `updateObjectToPopulate` method otherwise, if we are not using an IRI @id but only the object identifier, we got an 
```
[ErrorException]
  Notice: Undefined index: resource_class
```

This is very useful using *Symfony Messenger* with Api-platform since otherwise Entity normalizer fails to denormalize object and we are not able to consume messages.

### DIFF
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -31,6 +31,10 @@ class ItemNormalizer extends AbstractItemNormalizer
      */
     public function denormalize($data, $class, $format = null, array $context = [])
     {
+        if (!isset($context['resource_class'])) {
+            $context['resource_class'] = $class;
+        }
+
         // Avoid issues with proxies if we populated the object
         if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
             if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
--- a/tests/Serializer/ItemNormalizerTest.php
+++ b/tests/Serializer/ItemNormalizerTest.php
@@ -159,6 +159,63 @@ class ItemNormalizerTest extends TestCase
         $this->assertInstanceOf(Dummy::class, $normalizer->denormalize(['id' => '/dummies/12', 'name' => 'hello'], Dummy::class, null, $context));
     }
 
+    public function testDenormalizeWithIdWithoutResourceClass()
+    {
+        $context = [];
+
+        $propertyNameCollection = new PropertyNameCollection(['id']);
+        $propertyNameCollectionFactoryProphecy = $this
+            ->prophesize(PropertyNameCollectionFactoryInterface::class);
+        $propertyNameCollectionFactoryProphecy
+            ->create(Dummy::class, Argument::type('array'))
+            ->willReturn($propertyNameCollection)
+            ->shouldBeCalled();
+
+        $propertyMetadataFactory = new PropertyMetadata(null, null, true, true);
+        $propertyMetadataFactoryProphecy = $this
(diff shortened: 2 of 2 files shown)


soyuka:
Does https://github.com/api-platform/core/pull/2326 fix your issue?

rpanfili:
> Does #2326 fix your issue?

@soyuka in my specific use case (Messenger messages consumption) it works but that solution skips IRI composition (using the "guessed" class) and object population that can happens in `updateObjectToPopulate` as I tried to reproduce with the unit test case. Also the logger warning (maybe better a _notice_ instead) it's a bit too much in my opinion as it's something that the `parent` class would normally execute without triggering any message.

dunglas:
Thanks for contributing @rpanfili! We finally merged #2326 that fixes the same issue.

PR #2406, opened 2018-12-21
Added OpenAPI description to property filter

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | not really
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

follow up on #2391 that would also add the property filter in the newly supported OpenAPI spec version


dunglas:
Thanks!

PR #2413, opened 2018-12-21
Avoid type error when output_class is false

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        | 

Follow up on #2393
Those underlying functions expects string but might get boolean and 💥😄


dunglas:
Thanks @antograssiot

PR #2433, opened 2018-12-29
Allow strict matching for operation names

cedricziel:
When falling back to SwaggerUiAction, correctly case Operation ID
to enable correct matching.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |


dunglas:
Thanks @cedricziel

Issue #2364, opened 2018-12-04
Graphql permission  problem with delete mutation


SkvokeN:
Hello, 

I have permission problem with delete mutation.

`mutation {
  deleteOffice(input: {id: "/api/books/240", clientMutationId: ""}) {
    id
  }
}`

This query is valid and the book will be deleted . If I have permission to delete `office`, I can delete another entity.

I think need check `$resourceClass` and class `$item` there  https://github.com/api-platform/core/blob/v2.3.4/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php#L87. 

`            
if ($resourceClass !== get_class($item)) {
     return;
 }
`

P.S. Thank you for your work.
P.S.S. Sorry for my English )

SkvokeN:
Anybody?

alanpoulain:
Fixed in the next bugfix version (2.3.6).

SkvokeN:
thank you

dunglas:
This a severe security issue. I requested a CVE number.
It affects all versions from 2.2.0 to 2.3.5, this issue has been fixed in 2.3.6.
The patch: https://github.com/api-platform/core/pull/2441

Simperfit:
CVE number is: CVE-2019-1000011

PR #2441, opened 2019-01-09
[GraphQL] Check item resource class in mutation

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2364 
| License       | MIT
| Doc PR        | 

This prevents passing IRIs belonging to different resource classes, which would bypass access control in some instances (see #2364).

### DIFF
--- a/features/graphql/mutation.feature
+++ b/features/graphql/mutation.feature
@@ -122,6 +122,21 @@ Feature: GraphQL mutation support
     And the JSON node "data.deleteFoo.id" should be equal to "/foos/1"
     And the JSON node "data.deleteFoo.clientMutationId" should be equal to "anotherId"
 
+  Scenario: Trigger an error trying to delete item of different resource
+    When I send the following GraphQL request:
+    """
+    mutation {
+      deleteFoo(input: {id: "/dummies/1", clientMutationId: "myId"}) {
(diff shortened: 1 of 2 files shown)


lukasluecke:
Fixed php-cs failure

alanpoulain:
Nice! Could you add tests?

lukasluecke:
@alanpoulain

I added a Behat test but can't get it to succeed. I think it's something to do with the escaped quotes in the error message, any ideas? 🙈

alanpoulain:
Could you try to remove the escaping of the double quotes (or maybe to escape the non escaped ones)?

lukasluecke:
I tried almost every combination I could think of (single quotes, double quotes, double-escape, no escape). I'll have another look, but from the output of Behat it should match, shouldn't it?

lukasluecke:
@alanpoulain Fixed the test. Is there anything else needed?

alanpoulain:
LGTM :slightly_smiling_face:

alanpoulain:
Thank you @lukasluecke

PR #2442, opened 2019-01-09
Fix normalization of raw collections (not API resources)

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | possibly related to #1222
| License       | MIT
| Doc PR        | N/A

We have to support raw collections, since we have already returned `true` in `supportsNormalization` (we have no way to tell).

TODO:

- [x] Add anti-regression tests


dunglas:
Thanks @teohhanhui

Issue #2309, opened 2018-11-04
Creating custom mutations GraphQL


Giant-Jelly:
Im trying to create a custom authentication mutation with GraphQL but cant find any explanation saying how to do that?

Can someone point me in the right direction? Im happy to update the documentation with an example etc afterward.

alanpoulain:
Sorry, it's not possible nowadays. The GraphQL part of API Platform needs to be modified in order to allow it.

Giant-Jelly:
Oh I see, any plans on making this possible? I think its pretty likely that this is needed in most projects as it doesnt take long before the complexity grows past the standard functionality

leyriel:
Hello, i have a same probleme. Can i do a relationship mutation (like one to many) in graphQL request ?

serviware:
i create custom mutations with [data persisters](https://api-platform.com/docs/core/data-persisters/)

alanpoulain:
Custom mutations can be done in master: https://api-platform.com/docs/master/core/graphql/#custom-mutations

Issue #1870, opened 2018-04-15
[GraphQL] How to use DTOs?


IonBazan:
Is it possible to use a DTO without registering it as ApiResource? I would like to expose a DTO from my Entity but it is not passed to `SchemaBuilder` unless I register it as ApiResource. This however requires to add an identifier to the DTO. 
It works perfectly on JSON endpoints since `coordinates` are being serialized properly. This does not happen for GraphQL because API Platform does not know about the DTO at all and does not build schema for this.

Consider example entity:
```php
<?php

namespace App\Entity;

use App\DTO\Coordinates;
use Ramsey\Uuid\UuidInterface;

class Country
{
    /** @var UuidInterface */
    protected $id;

    /** @var float|null */
    protected $latitude;

    /** @var float|null */
    protected $longitude;

    /** @var string */
    protected $name;

    public function getId(): UuidInterface
    {
        return $this->id;
    }

    public function getLatitude(): ?float
    {
        return $this->latitude;
    }

    public function setLatitude(?float $latitude): Country
    {
        $this->latitude = $latitude;

        return $this;
    }

    public function getLongitude(): ?float
    {
        return $this->longitude;
    }

    public function setLongitude(?float $longitude): Country
    {
        $this->longitude = $longitude;

        return $this;
    }

    public function getName(): string
    {
        return $this->name;
    }

    public function setName(string $name): Country
    {
        $this->name = $name;

        return $this;
    }

    public function setCoordinates(Coordinates $coordinates): Country
    {
        return $this->setLatitude($coordinates->getLatitude())
            ->setLongitude($coordinates->getLongitude());
    }

    public function getCoordinates(): Coordinates
    {
        return new Coordinates($this->getLatitude(), $this->getLongitude());
    }

    public function __toString(): string
    {
        return $this->getName();
    }
}
```
and Coordinates DTO:
```php
<?php

namespace App\DTO;

use ApiPlatform\Core\Annotation\ApiProperty;

class Coordinates
{
    /** @var float */
    protected $latitude;

    /** @var float */
    protected $longitude;

    public function __construct(float $latitude, float $longitude)
    {
        $this->latitude = $latitude;
        $this->longitude = $longitude;
    }

    public function getLatitude(): float
    {
        return $this->latitude;
    }

    public function setLatitude(float $latitude): Coordinates
    {
        $this->latitude = $latitude;

        return $this;
    }

    public function getLongitude(): float
    {
        return $this->longitude;
    }

[...]

Issue #1700, opened 2018-02-12
GraphQL and Custom Operations 


nicodmf:
Is there a way to add custom mutation or query in graphql, like with the rest API ?

dunglas:
Yes but it's not documented yet :(
https://github.com/api-platform/core/blob/master/src/GraphQl/Type/SchemaBuilder.php#L84

AntoineBourin:
Is there some news about custom operations with GraphQL in Api platform ? :)

alanpoulain:
@AntoineBourin please see the PR above your message.

AntoineBourin:
Thanks! :+1:

Issue #473, opened 2016-03-19
Dependency to symfony/http-kernel


theofidry:
I see that `ApiPlatform\Core\EventListener\FormatRequestListener` – and `Hydra` and `JsonLd` ones as well, depends on `symfony/http-kernel`: is there any plan to remove this dependency?

We could have an interface for it instead, and then provide a bridge. In Laravel and Zend, they would be middlewares instead of listeners.

PR #2445, opened 2019-01-10
[GraphQL] Add custom queries

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | TODO

Possibly related: #1700 #473 #2309 #1870

Currently partially implements custom queries on existing resources, but the goal would be at least full custom query and mutation support on existing resources, possibly even for "new" resources, custom types etc. (maybe in another PR)

**Usage example**

`User.php` (Resource class)
```php
/**
 * @ApiResource(graphql={
 *     "query",
 *     "delete",
 *     "update",
 *     "create",
 *     "current"={
 *          "item"=CurrentUserResolver::class,
 *          "description"="Get currently logged in user"
 *     },
 *     "active"={
 *          "collection"=ActiveUsersResolver::class,
 *          "description"="Get all active users"
 *     }
 * })
 */
```

`CurrentUserResolver.php`
```php
class CurrentUserResolver implements QueryResolverInterface
{
    private $tokenStorage;
    private $normalizer;

    public function __construct(TokenStorageInterface $tokenStorage, NormalizerInterface $normalizer)
    {
        $this->tokenStorage = $tokenStorage;
        $this->normalizer = $normalizer;
    }

    public function __invoke($source, $args, $context, ResolveInfo $info)
    {
        $token = $this->tokenStorage->getToken();

        if (!$token) return null;

        return $this->normalizer->normalize($token->getUser(), ItemNormalizer::FORMAT);
    }
}
```

`ActiveUsersResolver.php`
```php
class ActiveUsersResolver implements QueryResolverInterface
{
    private $repository;
    private $normalizer;

    public function __construct(UserRepository $repository, NormalizerInterface $normalizer)
    {
        $this->repository = $repository;
        $this->normalizer = $normalizer;
    }

    public function __invoke($source, $args, $context, ResolveInfo $info)
    {
        $collection = $this->repository->findActive();

        $offset = 0;
        $data = ['totalCount' => 0.0, 'edges' => [], 'pageInfo' => ['endCursor' => null, 'hasNextPage' => false]];

        foreach ($collection as $index => $object) {
            $data['edges'][$index] = [
                'node' => $this->normalizer->normalize($object, ItemNormalizer::FORMAT),
                'cursor' => base64_encode((string) ($index + $offset)),
            ];
        }

        return $data;
    }
}
```


lukasluecke:
@alanpoulain Any early input from you, while I continue working on this? (e.g. the configuration structure, scope of this feature, ..)

alanpoulain:
I will have a look this weekend. @raoulclais is working on the custom mutations, he will probably make a PR soon.

lukasluecke:
@alanpoulain Thank you

@raoulclais Is there anywhere I can check out your current progress on the custom mutations?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlResolverPass.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Injects GraphQL resolvers.
+ *
+ * @internal
+ *
+ * @author Lukas Lücke <lukas@luecke.me>
+ */
+final class GraphQlResolverPass implements CompilerPassInterface
```
alanpoulain: `GraphQlQueryResolverPass` instead?

src/GraphQl/Resolver/ResolverInterface.php
```diff
@@ -0,0 +1,35 @@
...
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * A function retrieving an item to resolve a GraphQL query.
+ *
+ * @experimental
+ *
+ * @author Lukas Lücke <lukas@luecke.me>
+ */
+interface ResolverInterface
```
alanpoulain: `QueryResolverInterface` instead?

alanpoulain:
I think it should be better to have something like this:
```php

/**
 * @ApiResource(graphql={
 *     "query",
 *     "delete",
 *     "update",
 *     "create",
 *    "myQuery"={
 *        "query"=MyQuery::class
 *    }
 * })
 */
```
Like the PR for the custom mutations: https://github.com/api-platform/core/pull/2447
You should be able to identify them with the `query` key.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -172,6 +183,38 @@ private function getQueryFields(string $resourceClass, ResourceMetadata $resourc
         return $queryFields;
     }
 
+    /**
+     * Gets additional custom query fields of the schema.
+     */
+    private function getCustomQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata): array
```
alanpoulain: I think it should be possible to reuse the `getQueryFields` method.

lukasluecke:
@alanpoulain Thank you for your feedback. I have incorporated your suggestions and updated the resulting code sample in the PR description.

I'm not really sure about the configuration still. We need more than just a `query` field, as suggested by you. For now I have chosen `item` and `collection` fields, which serve both as the resolver class, and also as an indicator for what kind of query it is.

But if we go further, e.g custom input-types or result types for collections I'm not sure this will hold up.

alanpoulain:
IMO it should not be an issue in the future.
We would have something like this:
```php
/**
 * @ApiResource(graphql={
 *     "query",
 *     "delete",
 *     "update",
 *     "create",
 *     "current"={
 *          "item"=CurrentUserResolver::class,
 *          "description"="Get currently logged in user",
 *          "input"={"type"=MyType::class, "description"="Custom description input", args=[]}
 *     },
 *     "active"={
 *          "collection"=ActiveUsersResolver::class,
 *          "description"="Get all active users"
 *     }
 * })
 */
```
WDYT?

I think we need to add the "query" term anyway. My suggestion: `item_query` and `collection_query`. Are you OK with this?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlQueryResolverPass.php
```diff
@@ -0,0 +1,50 @@
...
+ * @internal
+ *
+ * @author Lukas Lücke <lukas@luecke.me>
+ */
+final class GraphQlQueryResolverPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws RuntimeException
```
alanpoulain: Can be removed.

lukasluecke:
LGTM. I‘ll make the changes sometime next week, probably towards the end (currently on vacation) 👍🏻

SCIF:
Guys, don't you think that custom operation controller/invoked class should return the type of resource/iterable collection of this resource objects like it's done for REST?

However, the resource term doesn't fit very good for graphql and it's more about operations and types. What do you think about idea to introduce `ApiOperation` annotation providing easy way to handle custom operations? Something like:

```
/**
 * @ApiOperation(iri="somecustomname", arguments={…)
 */
final class AddSubscription
{
    public function __constructor(…)
    {
        …
    }

    public function __invoke(…): Subscription
    {
        …
    } 
}
```

Returning type could easily detected by analysing `__invoke()` method.

My apologies in case if I'm wrong. it's just a suggestion.

lukasluecke:
@alanpoulain I have updated the config keys as agreed, and think this could be merged now. Further improvements (e.g. custom inputs) will follow in separate PRs.

(Sorry for the delay, had the flu after my vacation 🙄)

lukasluecke:
@SCIF Thanks fo your input and suggestion. I am not opposed to supporting annotations for this, but think it could be added later on as an additional / alternative way of configuring this behaviour.

src/Bridge/Symfony/Bundle/ApiPlatformBundle.php
```diff
@@ -38,6 +39,7 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new DataProviderPass());
         $container->addCompilerPass(new AnnotationFilterPass());
         $container->addCompilerPass(new FilterPass());
+        $container->addCompilerPass(new GraphQlQueryResolverPass());
```
alanpoulain: Maybe move it with the `GraphQlTypePass`?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlQueryResolverPass.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $resolvers = [];
+        foreach ($container->findTaggedServiceIds('api_platform.graphql.query_resolver', true) as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                if (!isset($tag['id'])) {
```
alanpoulain:
```php
$types[$tag['id'] ?? $serviceId] = new Reference($serviceId);
```
Like the `GraphQlTypePass`.

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
```diff
@@ -65,9 +65,14 @@
             <argument type="service" id="api_platform.graphql.resolver.resource_field" />
             <argument type="service" id="api_platform.graphql.types_factory" />
             <argument type="service" id="api_platform.filter_locator" />
+            <argument type="service" id="api_platform.graphql.query_resolver_locator" />
             <argument>%api_platform.collection.pagination.enabled%</argument>
         </service>
 
+        <service id="api_platform.graphql.query_resolver_locator" class="Symfony\Component\DependencyInjection\ServiceLocator">
```
alanpoulain: Add it with the resolvers?

lukasluecke: Then it would be before the `$typesFactory` argument, which is not nullable right now. But the `$queryResolverLocator` is. Should I just make it required? (what was the consesus on the null / required discussion?)

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlQueryResolverPass.php
lukasluecke: I originally based it on the FilterPass, which used the "old" syntax.

src/GraphQl/Resolver/QueryResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * A function retrieving an item to resolve a GraphQL query.
+ * Should return the normalized item or collection.
+ *
+ * @experimental
+ *
+ * @author Lukas Lücke <lukas@luecke.me>
+ */
+interface QueryResolverInterface
+{
+    public function __invoke($source, $args, $context, ResolveInfo $info);
```
alanpoulain: Maybe add:
```php
    /**
     * @return mixed The queried item
     */
```

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
alanpoulain: That's not what I meant :slightly_smiling_face:
I was just saying the service definition should be with the other resolver service definitions (below `<!-- Resolvers -->`).
But to answer your question, since we decided that the `typesFactory` is required, there is already a breaking change. Therefore I think you can make it required and move it before the `typesFactory`.

lukasluecke: Ah I see it now, makes sense. I'll make both changes 👍

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -86,12 +88,28 @@ public function getSchema(): Schema
             $graphqlConfiguration = $resourceMetadata->getGraphql() ?? [];
             foreach ($graphqlConfiguration as $operationName => $value) {
                 if ('query' === $operationName) {
-                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata);
+                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata, $operationName, ['args' => ['id' => ['type' => GraphQLType::id()]]], true);
```
alanpoulain: Why `true` for the `$collection` parameter?

alanpoulain: Maybe an empty array will be better IMO.

lukasluecke: Because that is the default behaviour. `query` generates both an item and a collection query (using the default input object).

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -156,19 +174,26 @@ private function getNodeQueryField(): array
     /**
      * Gets the query fields of the schema.
      */
-    private function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata): array
+    private function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata, string $operationName = 'query', $item, $collection): array
```
alanpoulain:
```suggestion
    private function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata, string $operationName, $itemConfiguration, $collectionConfiguration): array
```
I think you should add PHPDoc for the `$itemConfiguration` and `$collectionConfiguration` parameters like this:
```php
/**
 * @param array|bool $itemConfiguration false if not configured
 * @param array|bool $collectionConfiguration false if not configured
 */
```

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: Your choice. Would just need to modify the condition inside `getQueryFields` to explicitly check for `!== false` instead of "truthy" ;)

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -155,20 +173,30 @@ private function getNodeQueryField(): array
...
-        $deprecationReason = $resourceMetadata->getGraphqlAttribute('query', 'deprecation_reason', '', true);
+        $fieldName = lcfirst('query' === $operationName ? $shortName : $operationName.$shortName);
 
-        if ($fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, true, $resourceClass), $resourceClass)) {
-            $fieldConfiguration['args'] += ['id' => ['type' => GraphQLType::id()]];
-            $queryFields[lcfirst($shortName)] = $fieldConfiguration;
+        $deprecationReason = $resourceMetadata->getGraphqlAttribute($operationName, 'deprecation_reason', '', true);
+
+        if ($itemConfiguration && $fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, true, $resourceClass), $resourceClass)) {
+            if (\is_array($itemConfiguration)) {
```
alanpoulain: Should be not necessary if you use an empty array (see my previous comment).

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -155,20 +173,30 @@ private function getNodeQueryField(): array
...
+                $fieldConfiguration = array_merge($fieldConfiguration, $itemConfiguration);
+            }
+            $queryFields[$fieldName] = $fieldConfiguration;
         }
 
-        if ($fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, false, null, true, null, new Type(Type::BUILTIN_TYPE_OBJECT, false, $resourceClass)), $resourceClass)) {
-            $queryFields[lcfirst(Inflector::pluralize($shortName))] = $fieldConfiguration;
+        if ($collectionConfiguration && $fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, false, null, true, null, new Type(Type::BUILTIN_TYPE_OBJECT, false, $resourceClass)), $resourceClass)) {
+            if (\is_array($collectionConfiguration)) {
```
alanpoulain: Same.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -155,20 +173,30 @@ private function getNodeQueryField(): array
 
     /**
      * Gets the query fields of the schema.
+     *
+     * @param array|bool $itemConfiguration false if not configured
+     * @param array|bool $collectionConfiguration false if not configured
      */
-    private function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata): array
+    private function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata, string $operationName = 'query', $itemConfiguration, $collectionConfiguration): array
```
alanpoulain:
```suggestion
    private function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata, string $operationName, $itemConfiguration, $collectionConfiguration): array
```

src/GraphQl/Type/SchemaBuilder.php
alanpoulain: I think it would be better since it would also remove the `is_array` checks :wink:

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -86,12 +88,28 @@ public function getSchema(): Schema
...
             foreach ($graphqlConfiguration as $operationName => $value) {
                 if ('query' === $operationName) {
-                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata);
+                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata, $operationName, ['args' => ['id' => ['type' => GraphQLType::id()]]], true);
 
                     continue;
                 }
 
-                $mutationFields[$operationName.$resourceMetadata->getShortName()] = $this->getMutationFields($resourceClass, $resourceMetadata, $operationName);
+                if ($item = $resourceMetadata->getGraphqlAttribute($operationName, 'item_query')) {
```
alanpoulain:
```suggestion
                if ($itemConfiguration = $resourceMetadata->getGraphqlAttribute($operationName, 'item_query')) {
```

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -86,12 +88,28 @@ public function getSchema(): Schema
...
+                    $value['resolve'] = $this->queryResolverLocator->get($item);
+                    unset($value['item_query']);
+
+                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata, $operationName, $value, false);
+                    continue;
+                }
+
+                if ($collection = $resourceMetadata->getGraphqlAttribute($operationName, 'collection_query')) {
```
alanpoulain:
```suggestion
                if ($collectionConfiguration = $resourceMetadata->getGraphqlAttribute($operationName, 'collection_query')) {
```

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -86,12 +88,28 @@ public function getSchema(): Schema
...
                 if ('query' === $operationName) {
-                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata);
+                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata, $operationName, ['args' => ['id' => ['type' => GraphQLType::id()]]], true);
 
                     continue;
                 }
 
-                $mutationFields[$operationName.$resourceMetadata->getShortName()] = $this->getMutationFields($resourceClass, $resourceMetadata, $operationName);
+                if ($item = $resourceMetadata->getGraphqlAttribute($operationName, 'item_query')) {
+                    $value['resolve'] = $this->queryResolverLocator->get($item);
```
alanpoulain:
```suggestion
                    $itemConfiguration['resolve'] = $this->queryResolverLocator->get($item);
```

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -86,12 +88,28 @@ public function getSchema(): Schema
...
+                    unset($value['item_query']);
+
+                    $queryFields += $this->getQueryFields($resourceClass, $resourceMetadata, $operationName, $value, false);
+                    continue;
+                }
+
+                if ($collection = $resourceMetadata->getGraphqlAttribute($operationName, 'collection_query')) {
+                    $value['resolve'] = $this->queryResolverLocator->get($collection);
```
alanpoulain:
```suggestion
                    $collectionConfiguration['resolve'] = $this->queryResolverLocator->get($collection);
```

alanpoulain:
I think you need to add Behat tests and unit tests if possible to make sure it's working correctly.
Thank you for working on this, it will be a great improvement!

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -155,20 +173,30 @@ private function getNodeQueryField(): array
...
+        $fieldName = lcfirst('query' === $operationName ? $shortName : $operationName.$shortName);
 
-        if ($fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, true, $resourceClass), $resourceClass)) {
-            $fieldConfiguration['args'] += ['id' => ['type' => GraphQLType::id()]];
-            $queryFields[lcfirst($shortName)] = $fieldConfiguration;
+        $deprecationReason = $resourceMetadata->getGraphqlAttribute($operationName, 'deprecation_reason', '', true);
+
+        if ($itemConfiguration && $fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, true, $resourceClass), $resourceClass)) {
+            if (\is_array($itemConfiguration)) {
+                $fieldConfiguration = array_merge($fieldConfiguration, $itemConfiguration);
```
alanpoulain:
```suggestion
                $fieldConfiguration += $itemConfiguration;
```

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -155,20 +173,30 @@ private function getNodeQueryField(): array
...
+            }
+            $queryFields[$fieldName] = $fieldConfiguration;
         }
 
-        if ($fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, false, null, true, null, new Type(Type::BUILTIN_TYPE_OBJECT, false, $resourceClass)), $resourceClass)) {
-            $queryFields[lcfirst(Inflector::pluralize($shortName))] = $fieldConfiguration;
+        if ($collectionConfiguration && $fieldConfiguration = $this->getResourceFieldConfiguration($resourceClass, $resourceMetadata, null, $deprecationReason, new Type(Type::BUILTIN_TYPE_OBJECT, false, null, true, null, new Type(Type::BUILTIN_TYPE_OBJECT, false, $resourceClass)), $resourceClass)) {
+            if (\is_array($collectionConfiguration)) {
+                $fieldConfiguration = array_merge($fieldConfiguration, $collectionConfiguration);
```
alanpoulain:
```suggestion
                $fieldConfiguration += $collectionConfiguration;
```

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: This does not result in the same behaviour, in case of keys existing in both arrays it would take those of the `$fieldConfiguration` (the "generated" ones). I think it should be able to override them using `$itemConfiguration`, which is why I chose `array_merge`.

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: This does not result in the same behaviour, in case of keys existing in both arrays it would take those of the `$fieldConfiguration` (the "generated" ones). I think it should be able to override them using `$collectionConfiguration`, which is why I chose `array_merge`.

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: Makes sense 👍

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: `collection_query` is not the same as `$collectionConfiguration`

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: `item_query` is not the same as `$itemConfiguration`

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: `$value` is the whole configuration of the operation, whereas `$itemQuery` is only the FQDN of the resolver class.

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: `$value` is the whole configuration of the operation, whereas `$collectionQuery` is only the FQDN of the resolver class.

lukasluecke:
I will try to get some work done on tests soon, but the rest of the week will be quite busy for me, so not sure when I'll get to it. Thanks for your feedback so far! 🥇

SCIF:
Hi @lukasluecke , my message is not about resource/operation definition way but about DX. I tried to suggest a way simpler DX flow of implementation of custom operations. I think that normalization and composing an edge information should not have place in custom operation executor.

lukasluecke:
@alanpoulain I have added a simple behat test for `item_query`, do you think it's enough if I add a similar one for `collection_query` for now?

I am still not quite content with the configuration format (`item_query`, `collection_query`) and would prefer a cleaner approach, but I think that can be discussed later.

SCIF:
Hi @lukasluecke ,

It would be nice to have tests with parameters.

alanpoulain:
@SCIF It is planned in another PR :slightly_smiling_face: This is a first step.

lukasluecke:
Thank you very much @alanpoulain 🚀

SCIF:
Hi @lukasluecke , @alanpoulain , are you going to implement custom mutations in separate PR?

lukasluecke:
There is still #2447 for custom mutations, but I‘ll probably work on mutations and more options in a new PR (if it‘s not merged by the time I get to it).

SCIF:
@lukasluecke , @alanpoulain 
But guys, what do you think on idea to extract Normalizing stage out of custom operations implementation?  In REST it's out of custom operation action, which is very convenient. Moreover, it will bring mismatch of behavior between REST and GraphQL in case if you will leave current behaviour.

alanpoulain:
@lukasluecke I will finish the PR for custom mutations soon since the major part of the work has already been done.
I have also worked on a way to modify easily the type system.
@SCIF I don't understand what you mean, could you elaborate?

alanpoulain:
Great addition, thank you @lukasluecke!

SCIF:
@alanpoulain 
Current PR implements custom resolver as a way to execute/invoke custom operation. That means a developer MUST handle normalizing WITHIN every custom operation.

While framework's [REST implementation](https://api-platform.com/docs/core/operations/#recommended-method) handles data type/resource and the developer MUST NOT care on normalizing.

I've tried to express this in [comment above](https://github.com/api-platform/core/pull/2445#issuecomment-464431474).

SCIF:
I'm sure, that ability even to implement normalizing stage could be useful, while these are really not common cases.

alanpoulain:
OK thank you for the explanation. It could have been done this way if we reused the existing resolvers. But it would have meant less flexibility. I will try to think about it.

SCIF:
@alanpoulain ,
So, to be clear, it's much better to call this PR's feature as `custom resolvers` rather than `custom  operation` to distinct different DX.

PR #2447, opened 2019-01-11
[GraphQL] Add custom mutations

raoulclais:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Introduce a new `MutationResolverInterface` interface.

Create a service which implements this interface and declare your custom mutation on your entity:
```
@ApiResource(graphql={
    "create",
    "update",
    "increment"={"mutation"=IncrementMutation::class}
})
```

If `null` is returned from the resolver, the item will not be persisted.

### DIFF
--- a/features/bootstrap/DoctrineContext.php
+++ b/features/bootstrap/DoctrineContext.php
@@ -20,6 +20,7 @@ use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\Dummy as DummyDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyAggregateOffer as DummyAggregateOfferDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyCar as DummyCarDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyCarColor as DummyCarColorDocument;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyCustomMutation as DummyCustomMutationDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyCustomQuery as DummyCustomQueryDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyDate as DummyDateDocument;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyDtoCustom as DummyDtoCustomDocument;
@@ -61,6 +62,7 @@ use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyAggregateOffer;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCar;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCarColor;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCustomMutation;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyCustomQuery;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyDate;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyDtoCustom;
@@ -424,6 +426,21 @@ final class DoctrineContext implements Context
             $dummyCustomQuery = $this->buildDummyCustomQuery();
 
             $this->manager->persist($dummyCustomQuery);
+        }
+
+        $this->manager->flush();
+    }
+
+    /**
+     * @Given there are :nb dummyCustomMutation objects
+     */
+    public function thereAreDummyCustomMutationObjects(int $nb)
+    {
+        for ($i = 1; $i <= $nb; ++$i) {
+            $customMutationDummy = $this->buildDummyCustomMutation();
+            $customMutationDummy->setOperandA(3);
+
+            $this->manager->persist($customMutationDummy);
         }
 
         $this->manager->flush();
@@ -1353,6 +1370,14 @@ final class DoctrineContext implements Context
     private function buildDummyCustomQuery()
     {
         return $this->isOrm() ? new DummyCustomQuery() : new DummyCustomQueryDocument();
+    }
+
+    /**
+     * @return DummyCustomMutation|DummyCustomMutationDocument
+     */
+    private function buildDummyCustomMutation()
+    {
+        return $this->isOrm() ? new DummyCustomMutation() : new DummyCustomMutationDocument();
     }
 
     /**
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlMutationResolverPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlMutationResolverPass.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Injects GraphQL Mutation resolvers.
+ *
+ * @internal
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+final class GraphQlMutationResolverPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $mutations = [];
+        foreach ($container->findTaggedServiceIds('api_platform.graphql.mutation_resolver', true) as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                $mutations[$tag['id'] ?? $serviceId] = new Reference($serviceId);
+            }
+        }
+
+        $container->getDefinition('api_platform.graphql.mutation_resolver_locator')->addArgument($mutations);
+    }
+}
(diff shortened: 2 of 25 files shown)


src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -92,11 +95,31 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
                 case 'update':
+                default:
                     $context = null === $item ? ['resource_class' => $resourceClass] : ['resource_class' => $resourceClass, 'object_to_populate' => $item];
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
+
+                    if ('create' !== $operationName && 'update' !== $operationName) {
+                        $mutationId = $resourceMetadata->getGraphqlAttribute($operationName, 'mutation');
+                        if (!$this->mutationLocator->has($mutationId)) {
+                            @trigger_error(sprintf('The mutation %s does not exist', $mutationId));
```
dunglas: Why not throwing an exception instead?

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -92,11 +95,31 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
             switch ($operationName) {
+                case 'delete':
+                    if ($item) {
+                        $this->dataPersister->remove($item);
+                        $data['id'] = $args['input']['id'];
+                    } else {
+                        $data['id'] = null;
+                    }
+                    break;
                 case 'create':
```
dunglas: The two following lines can be removed then, right?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MutationPass.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Injects mutations.
+ *
+ * @internal
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+final class MutationPass implements CompilerPassInterface
```
alanpoulain: `GraphqlMutationPass` instead?

src/GraphQl/Mutation/MutationInterface.php
```diff
@@ -0,0 +1,28 @@
...
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * Interface MutationInterface.
+ *
+ * @experimental
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+interface MutationInterface
```
alanpoulain: `MutationResolverInterface` instead?

alanpoulain:
It should be great to have a custom input, but can be done later in another PR.

src/GraphQl/Mutation/MutationResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\GraphQl\Mutation;
+
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * Interface MutationInterface.
```
dunglas: Should be replaced by an explicit definition or removed

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -48,10 +49,11 @@
     private $dataPersister;
     private $normalizer;
     private $resourceMetadataFactory;
+    private $mutationLocator;
     private $resourceAccessChecker;
     private $validator;
 
-    public function __construct(IriConverterInterface $iriConverter, DataPersisterInterface $dataPersister, NormalizerInterface $normalizer, ResourceMetadataFactoryInterface $resourceMetadataFactory, ResourceAccessCheckerInterface $resourceAccessChecker = null, ValidatorInterface $validator = null)
+    public function __construct(IriConverterInterface $iriConverter, DataPersisterInterface $dataPersister, NormalizerInterface $normalizer, ResourceMetadataFactoryInterface $resourceMetadataFactory, ContainerInterface $mutationLocator, ResourceAccessCheckerInterface $resourceAccessChecker = null, ValidatorInterface $validator = null)
```
dunglas: This is a BC break. It's allowed for `@experimental` classes but it would be better to avoid it. Can we also made this argument optional? It doesn't have to be mandatory.

src/GraphQl/Mutation/MutationResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @experimental
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+interface MutationResolverInterface
+{
+    public function __invoke($item, array $input, ResolveInfo $info);
```
dunglas: Shouldn't the mutation return something? How is it take into account by the Schema generator?

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
alanpoulain: It can be made optional but do we really want to handle the "breaking" case (when it's null)?

features/bootstrap/DoctrineContext.php
```diff
@@ -1182,6 +1183,18 @@ public function thereIsAMaxDepthDummyWithLevelOfDescendants(int $level)
         $this->manager->flush();
     }
 
+    /**
+     * @Given there is a custom mutation dummy object
+     */
+    public function thereIsACustomMutationDummyObjects()
```
alanpoulain:
```suggestion
    public function thereIsACustomMutationDummyObject()
```

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -98,11 +102,33 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
+
+                    if ('create' !== $operationName && 'update' !== $operationName) {
+                        if (null === $this->mutationLocator) {
+                            throw new RuntimeException('MutationLocator service should not be null');
+                        }
+
+                        $mutationId = $resourceMetadata->getGraphqlAttribute($operationName, 'mutation');
+                        if (!$this->mutationLocator->has($mutationId)) {
```
alanpoulain:
```suggestion
                        if (!$this->mutationLocator || !$this->mutationLocator->has($mutationId)) {
```

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -98,11 +102,33 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
+                        $data['id'] = null;
+                    }
+                    break;
+                default:
                     $context = null === $item ? ['resource_class' => $resourceClass] : ['resource_class' => $resourceClass, 'object_to_populate' => $item];
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
+
+                    if ('create' !== $operationName && 'update' !== $operationName) {
+                        if (null === $this->mutationLocator) {
```
alanpoulain: I don't really like to throw an exception here because the service is null. I have added my suggestion below.

src/GraphQl/Mutation/MutationResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @experimental
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+interface MutationResolverInterface
+{
+    /**
+     * @return mixed The mutated item
+     */
+    public function __invoke($item, array $input, ResolveInfo $info);
```
alanpoulain:
```suggestion
    public function __invoke($item, array $context);
```
Maybe having a context is better to be independent of webonyx/graphql-php.
Then in context we have:
```php
['input' => [], 'resolve_info' => $info]
```

tests/Fixtures/TestBundle/Entity/CustomMutationDummy.php
```diff
@@ -0,0 +1,101 @@
...
+ *     "sum"={
+ *         "mutation"=SumMutation::class,
+ *         "normalization_context"={"groups"={"result"}},
+ *         "denormalization_context"={"groups"={"sum"}}
+ *     },
+ * })
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+class CustomMutationDummy
```
alanpoulain: Should be added also as document for MongoDB.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlMutationResolverPass.php
```diff
@@ -0,0 +1,50 @@
...
+ * @internal
+ *
+ * @author Raoul Clais <raoul.clais@gmail.com>
+ */
+final class GraphQlMutationResolverPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws RuntimeException
```
alanpoulain: Can be removed.

alanpoulain:
@SCIF, @lukasluecke I think this PR is good to go, could you take a look please?

SCIF:
It looks pretty same as custom query PR's. And complains are generally the same:
- `id` argument cannot be opt out
- GraphQl custom operations provides different DX comparing to REST custom operations

I'm pretty sure, that getting rid of `id` will not be done in foreseeable future (and definitely not in this PR). So, I suggest one more time to consider using `custom query resolvers` and `custom mutation resolvers` in documentation, rather than `custom operation`. I would highly appreciate creation of issue and starting of process designing/discussing proper way and proper DX of implementation of custom operations (ready to invest some time into development in case if we would arrange desired DX fast).

alanpoulain:
> `id` argument cannot be opt out
This is not right. You can with both custom mutation and query resolvers. I've added a Behat test in the custom query resolver PR and I will document it. Maybe I should add one here too.

> GraphQl custom operations provides different DX comparing to REST custom operations
I'm not sure to follow you. The documentation about custom operations for REST is here: https://api-platform.com/docs/core/operations/#creating-custom-operations-and-controllers
It says:
> API Platform automatically retrieves the appropriate PHP entity using the data provider then deserializes user data in it, and for `POST` and `PUT` requests updates the entity with data provided by the user.
This is exactly what this PR does for GraphQL.
I would like to see your proposal if it provides a better DX though.

lukasluecke:
@SCIF 

I think it's pretty clear that the merged PR was "custom queries" and this one is "custom mutations", and that should and will be reflected in the docs. Anything beyond that can be implemented at any point in the future - if you have any suggestions on that, maybe you can start by creating the issue, and we can discuss there?

I believe the id is not required, which should solve almost any concerns regarding it - yes, you can not remove the argument right now, but it does not really matter that much.

alanpoulain:
Thank you @raoulclais :tada:

Issue #2365, opened 2018-12-04
Support inclusion of resources from a path according to JSON:API spec


dematerializer:
Although [Inclusion of Related Resources](https://jsonapi.org/format/#fetching-includes) was partly implemented in #1949 as requested by #1934, unfortunately support for inclusion of resources from a path was not implemented. Would it be possible for this to be supported as well?

The following already works as expected (comments are included in the response):

```
GET /articles/1?include=comments HTTP/1.1
Accept: application/vnd.api+json
```

This doesn't work (neither comments nor authors are included in the response as they should be according to the spec):

```
GET /articles/1?include=comments.author HTTP/1.1
Accept: application/vnd.api+json
```

thomasdom:
The implementation of this enhancement would add a huge performance improvement for apps with a server using API Platform with JSON:API

tonymrakovcic:
Any ideas on this?
This would be a great improvement

simonihmig:
We (specifically @andreasschacht) have been working on this recently, and got it working in our app (with some overridden services). We just need to find the time to properly integrate the changes (mainly tests!) upstream here. I hope we can file a PR next week, or the week after!?

PR #2448, opened 2019-01-12
[JSONAPI] Add support for nested resource inclusion

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/issues/2365, https://github.com/api-platform/api-platform/issues/940
| Doc PR        | to be done

Here is a first attempt at supporting inclusion of nested resources.

I think it does not work well with [Sparse Fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets) but that's already an improvement.

@dematerializer, @RikudouSage it, would be really appreciated if you could try it and provide feedback

### DIFF
--- a/features/jsonapi/related-resouces-inclusion.feature
+++ b/features/jsonapi/related-resouces-inclusion.feature
@@ -662,3 +662,301 @@ Feature: JSON API Inclusion of Related Resources
         }]
     }
     """
+
+  @createSchema
+  Scenario: Request inclusion of nested resource on an item
+    Given there are 1 dummy objects with relatedDummy and its thirdLevel
+    When I send a "GET" request to "/dummies/1?include=relatedDummy.thirdLevel"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And print last JSON response
+    And the JSON should be valid according to the JSON API schema
+    And the JSON should be deep equal to:
+    """
+        {
+            "data": {
+                "id": "/dummies/1",
+                "type": "Dummy",
+                "attributes": {
+                    "description": null,
+                    "dummy": null,
+                    "dummyBoolean": null,
+                    "dummyDate": null,
+                    "dummyFloat": null,
+                    "dummyPrice": null,
+                    "jsonData": [],
+                    "arrayData": [],
+                    "name_converted": null,
+                    "_id": 1,
+                    "name": "Dummy #1",
+                    "alias": "Alias #0",
+                    "foo": null
+                },
+                "relationships": {
+                    "relatedDummy": {
+                        "data": {
+                            "type": "RelatedDummy",
+                            "id": "/related_dummies/1"
+                        }
+                    }
+                }
+            },
+            "included": [
+                {
+                  "id": "/third_levels/1",
+                  "type": "ThirdLevel",
+                  "attributes": {
+                      "_id": 1,
+                      "level": 3,
+                      "test": true
+                  }
+                },
+                {
+                    "id": "/related_dummies/1",
+                    "type": "RelatedDummy",
+                    "attributes": {
+                        "name": "RelatedDummy #1",
+                        "dummyDate": null,
+                        "dummyBoolean": null,
+                        "embeddedDummy": {
+                            "dummyName": null,
+                            "dummyBoolean": null,
+                            "dummyDate": null,
+                            "dummyFloat": null,
+                            "dummyPrice": null,
+                            "symfony": null
+                        },
+                        "_id": 1,
+                        "symfony": "symfony",
+                        "age": null
+                    },
+                    "relationships": {
+                        "thirdLevel": {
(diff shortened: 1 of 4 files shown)


RikudouSage:
The inclusion iself works great, but doesn't contain any relationship data so I don't know which base resource it belongs to if I run it on collection (my url was: `/claims?include=activities,activities.invoices`).

The relevant part of the response (second level resource):

```json
  {
      "id": "/api/v1/invoices/10",
      "type": "Invoice",
      "attributes": {
        "_id": 10,
        "vendor": "Vendor Ltd.",
        "number": "20190001",
        "amount": 1000,
        "created": "2019-01-14T16:17:18+01:00",
        "updated": "2019-01-14T16:17:18+01:00"
      }
    }
```

The first level include has relationship data correctly set:

```json
  {
      "id": "/api/v1/activities/9",
      "type": "Activity",
      "attributes": {
        "_id": 9,
        "name": "Activity no. 1",
        "errors": {
          "errors": [],
          "messages": []
        },
        "created": "2019-01-14T16:17:18+01:00",
        "updated": "2019-01-14T16:17:18+01:00"
      },
      "relationships": {
        "claim": {
          "data": {
            "id": "/api/v1/claims/9",
            "type": "Claim"
          }
        },
        "invoices": {
          "data": [
            {
              "id": "/api/v1/invoices/10",
              "type": "Invoice"
            }
          ]
        }
      }
    }
```

RikudouSage:
Also it works only for second-level properties, third etc. don't work.

antograssiot:
Thanks for your feedback, I'll rework this when I'll get a chance

jarrodirwin:
@antograssiot did this ever get re-worked and / or merged? I'm also in need of this functionality....

antograssiot:
Not yet unfortunately. I kind of forget about it. Do you want to give it a try @jarrodirwin ?

jarrodirwin:
So finally got around to testing this again. I had to re-roll the changes against v2.4.5 - 
[jsonapi-nested-inclusion-2.txt](https://github.com/api-platform/core/files/3395373/jsonapi-nested-inclusion-2.txt) (patch attached with .txt extension so it would be accepted... should probably be an updated pull request..).

From my testing it seems to work great... As @RikudouSage mentioned, the included (sub?) resources do not have any data on them to say what resource it is a relation of however that doesn't actually matter as the source resource does.

Sample request:
URL: `/test-items/4?include=testItemOptions,testItemOptions.referenceDatas,testItemOptions.referenceDatas.level&groups[]=list&groups[]=testitem-testitemoptions&groups[]=testitemoption-referencedatas`

This returns (snippet):
```json 
{
    "data": {
        "id": "/api/v2/test-items/4",
        "type": "TestItem",
        "attributes": {
            "_id": 4,
            "name": "Test Item 4"
        },
        "relationships": {
            "testItemOptions": {
                "data": [
                    {
                        "type": "TestItemOption",
                        "id": "/api/v2/test-item-options/14"
                    }
                ]
            }
        }
    },
    "included": [
        {
            "id": "/api/v2/levels/8",
            "type": "Level",
            "attributes": {
                "_id": 8,
                "name": "Level 8"
            }
        },
        {
            "id": "/api/v2/reference-datas/13",
            "type": "ReferenceData",
            "attributes": {
                "_id": 13,
                "name": "Reference Data 13"
            },
            "relationships": {
                "level": {
                    "data": {
                        "type": "Level",
                        "id": "/api/v2/levels/8"
                    }
                }
            }
        },
        {
            "id": "/api/v2/test-item-options/14",
            "type": "TestItemOption",
            "attributes": {
                "_id": 14,
                "name": "Test Item Option 14"
            },
            "relationships": {
                "referenceDatas": {
                    "data": [
                        {
                            "type": "ReferenceData",
                            "id": "/api/v2/reference-datas/13"
                        }
                    ]
                }
            }
        }
    ]
}
```

This is basically what I would be expecting... not sure what improvements are needed here?

RikudouSage:
@jarrodirwin That would be a problem in case you query for a collection and not a single item, you don't know which included resource belongs to which source resource.

jarrodirwin:
@RikudouSage This is true however does that matter? An included resource is the same for any source resource that includes it isn't it?

When reconstructing the relations (say in a front end) you are looping over the source then looking in the included array for the matching included resource. The source knows what relations it has and what should be included.

What am I missing here (as probably is something haha)? 

Are you thinking if there are 2 records returned in a collection call, both of which include a relationship to the same resource, that this included resource should be added to the included array twice with some identifier on it to say which source resource it is a relation of? Even though all the other data in the 2 included resources are identical? Or something else entirely?

Actually, what does JSONAPI spec say for this?

PR #2450, opened 2019-01-14
Allow to push relations to the client using HTTP/2

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/714

This new feature allows to use the web server to push relations of the current resources using HTTP/2. It's especially useful when used together with [the built-in HTTP cache invalidation system (based on Varnish)](https://api-platform.com/docs/core/performance/#enabling-the-built-in-http-cache-invalidation-system).

The usage is very easy:

```php
<?php

namespace App\Entity;

/**
 * @ApiResource
 */
class Book
{
    /**
     * @var Author $author
     * @ApiProperty(push=true)
     */
    public $author;
}
```

Then, the Symfony WebLink component will be used to generate the appropriate `Link` HTTP header that will be used by the web server to send the related resource(s) using a server push.

### DIFF
--- a/features/push_relations/push.feature
+++ b/features/push_relations/push.feature
@@ -1,0 +1,17 @@
+@sqlite
+Feature: Push relations using HTTP/2
+  In order to have a fast API
+  As an API software developer
+  I need to push relations using HTTP/2
+
+  @createSchema
+  Scenario: Push the relations of a collection of items
+    Given there are 2 dummy objects with relatedDummy
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/dummies"
+    Then the header "Link" should be equal to '</related_dummies/1>; rel="preload",</related_dummies/2>; rel="preload",<http://example.com/docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"'
+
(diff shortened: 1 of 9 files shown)


features/push_relations/push.feature
```diff
@@ -0,0 +1,17 @@
...
+  I need to push relations using HTTP/2
+
+  @createSchema
+  Scenario: Push the relations of a collection of items
+    Given there are 2 dummy objects with relatedDummy
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/dummies"
+    Then the header "Link" should be equal to '</related_dummies/1>; rel="preload",</related_dummies/2>; rel="preload",<http://example.com/docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"'
+
+  Scenario: Push the relations of a collection of items
```
alanpoulain:
```suggestion
  Scenario: Push the relations of an item
```

er1z:
Is it already documented with an example scenario?

PR #2453, opened 2019-01-16
Enhance the filter annotation

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

This PR allows to use the abstract definitions of filter services when using the filter annotation.
It also allows to use Elasticsearch filters with the filter annotation without issues. In fact, without this fix the passed name converter to the Elasticsearch filter constructor is `null` because of the autowiring.

WDYT?


dunglas:
Thanks @meyerbaptiste

PR #2454, opened 2019-01-17
WIP: [GraphQL] Pass input object to access control

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | ?
| New feature?  | ?
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This makes it possible to allow creation of new objects only if certain properties match provided rules.

### DIFF
--- a/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
@@ -92,7 +92,9 @@ final class ItemMutationResolverFactory implements ResolverFactoryInterface
             }
 
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
-            $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $item, $operationName);
+            if ('create' !== $operationName) {
+                $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $item, $operationName);
+            }
 
             $inputMetadata = $resourceMetadata->getAttribute('input', ['class' => $resourceClass]);
             if (null === $resourceClass = $inputMetadata['class'] ?? null) {
@@ -105,6 +107,11 @@ final class ItemMutationResolverFactory implements ResolverFactoryInterface
                     $context = null === $item ? ['resource_class' => $resourceClass] : ['resource_class' => $resourceClass, 'object_to_populate' => $item];
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
+
+                    if ('create' === $operationName) {
+                        $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $item, $operationName);
+                    }
+
                     $this->validate($item, $info, $resourceMetadata, $operationName);
                     $persistResult = $this->dataPersister->persist($item);
 


src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -92,14 +92,21 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             }
 
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
-            $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $item, $operationName);
+            if($operationName === 'update' || $operationName === 'delete') {
```
dunglas: Should be `'create' !== $operationName` (the security rule must be triggered for custom mutations).

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -92,14 +92,21 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
...
+            }
 
             switch ($operationName) {
                 case 'create':
                 case 'update':
                     $context = null === $item ? ['resource_class' => $resourceClass] : ['resource_class' => $resourceClass, 'object_to_populate' => $item];
                     $context += $resourceMetadata->getGraphqlAttribute($operationName, 'denormalization_context', [], true);
                     $item = $this->normalizer->denormalize($args['input'], $resourceClass, ItemNormalizer::FORMAT, $context);
+
+                    if($operationName === 'create') {
```
dunglas: Then it should be the same for `update` isn't it?

dunglas:
I wonder how this behaves in REST.  The behavior must be consistent.

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
lukasluecke: I‘m not certain, because for update `$item` is already populated with the existing resource, so it is checked in the first condition. It depends on what kind of behaviour you want to allow.

lukasluecke:
@dunglas I was thinking the same thing, but couldn‘t figure out how exactly it was working for REST. My assumption is, that for REST the subject is currently the classname of the resource. But I will try to dig a little deeper tomorrow :)

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
lukasluecke: Absolutely, good catch. Will update tomorrow :)

lukasluecke:
@dunglas I checked the way this behaves in REST. Seems that it also uses `null` right now (`EventListener/ReadListener.php:78`, `Security/EventListener/DenyAccessListener.php:71`).

I think it would make sense to change that as well, but I'm not sure what would be the best way to do it.

alanpoulain:
I've only rebased this branch against 2.4.

lukasluecke:
@alanpoulain @dunglas Could you maybe take a look at this again? I still don't see any better way to achieve this functionality.

I am trying to use the following entity, and allow only admins to create "root" entities (with no parent), and everyone else to create only "child" entities.

```php
/**
 * @ApiResource(
 *     mercure=true,
 *     graphql={
 *         "query",
 *         "create"={
 *             "access_control"="is_granted('create',object)" <-- `object` is null here!
 *         },
 *         "update"={
 *             "access_control"="is_granted('update',object)"
 *         },
 *         "delete"={
 *             "access_control"="is_granted('delete',object)"
 *         }
 *     }
 * )
 * @ORM\Entity
 */
class Dummy
{
    /**
     * @ORM\ManyToOne(targetEntity="App\Entity\Dummy", inversedBy="children")
     */
    private $parent;

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\Dummy", mappedBy="parent")
     */
    private $children;
}
```

```php
class DummyVoter extends Voter
{
    protected function supports($attribute, $subject)
    {
        return $subject instanceof Dummy; // <-- $subject is null here, so returns false on `create`!
    }

    protected function voteOnAttribute($attribute, $subject, TokenInterface $token)
    {
        $user = $token->getUser();

        /** @var Dummy $subject */
        switch ($attribute) {
            case 'create':
                if ($subject->getParent() !== null && [user is admin]) {
                    return true;
                }
                if ($subject->getParent() === null) {
                    return true;
                }
                return false;
            case self::UPDATE:
            case self::DELETE:
                [...]
        }
    }
}
```

I have also tried using validation constraints with expression language, which should be able to access the `parent` field, but ran into issues checking the additional access (e.g. user role) - and I'm not sure that it would be the right approach anyway..

lukasluecke:
Looks like #2814 should solve this, if not I'll create a new issue / PR.

PR #2465, opened 2019-01-20
Error if references are not stored correctly for lookups

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

To have an error if references are not stored as `id` or `ref`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,7 +195,7 @@ jobs:
           name: Run Behat tests
           command: |-
             mkdir -p build/logs/tmp build/cov
-            for f in $(find features -name '*.feature' -not -path 'features/main/exposed_state.feature' -not -path 'features/elasticsearch/*' | circleci tests split --split-by=timings); do
+            for f in $(find features -name '*.feature' -not -path 'features/main/exposed_state.feature' -not -path 'features/elasticsearch/*' -not -path 'features/mongodb/*' | circleci tests split --split-by=timings); do
               _f=${f//\//_}
               FEATURE="${_f}" phpdbg -qrr vendor/bin/behat --profile=coverage --suite=default --format=progress --out=std --format=junit --out=build/logs/tmp/"${_f}" "$f"
             done
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -18,7 +18,7 @@ default:
         - 'Behat\MinkExtension\Context\MinkContext'
         - 'Behatch\Context\RestContext'
       filters:
-        tags: '~@postgres&&~@elasticsearch'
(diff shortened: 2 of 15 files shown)


alanpoulain:
This PR is RFR. @alcaeus are you OK with this implementation?

alcaeus:
I can take a deeper look in a couple of days. At first glance, I wonder why the lookup isn’t done against the mapped field? If the lookup stage sees that you’re looking up a reference, it will deal with the reference type automatically and throw exceptions as necessary.

I’m also not sure if this covers the ref vs. id types correctly, but I need to pore over this for a bit longer.

alanpoulain:
I think you need more context to understand. In API Platform, we can use nested properties filters (https://api-platform.com/docs/core/filters/#filtering-on-nested-properties).
If the property is only nested at one level, all is good, the lookup can be done against the mapped field (that's how I did it before: https://github.com/alanpoulain/core/blob/0849be0dce78f29544d17db20458b41976b8843c/src/Bridge/Doctrine/MongoDbOdm/PropertyHelperTrait.php).
But if the filtered property is nested at level two or more, then it can't be done this way.
I've added test cases for this: https://github.com/api-platform/core/blob/a9bc50743e4d8d0a0e5a0591a8f9cc127969291c/features/doctrine/search_filter.feature#L573-L646

alanpoulain:
Could you look it @alcaeus (we plan to release a stable version soon)? :slightly_smiling_face:

alcaeus:
Sorry for the delay here but I didn't get to it during my vacation. Looking at it the lookup code, you'll want to replicate the logic in https://github.com/doctrine/mongodb-odm/blob/5d9b452c067a7d9eeb4d2555cfa168df76e6a99e/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/Lookup.php#L166..L199 to ensure you're covering inverse references with repository methods and different reference storage strategies as well.

alanpoulain:
That's right, except I didn't replicate the inverse references with repository methods exception... I should I suppose.
It would be much better to have not to replicate this behavior, but I cannot replace `$this->class` with what I need.

alcaeus:
> It would be much better to have not to replicate this behavior

I’ll work on supporting nested lookups in 2.x soon, would that help?

alanpoulain:
Of course! I'll gladly remove all this code then.

alanpoulain:
Has been merged for the stable release. Waiting for nested lookups in Doctrine MongoDB ODM to remove the unnecessary code.

Issue #2475, opened 2019-01-28
API link response header is wrong in OPTIONS request


vincentchalamon:
API link response header is wrong: http://localhost/docs.jsonld (request OPTIONS)
Bug revealed in https://github.com/api-platform/demo/pull/95, to investigate on current repository

PR #2477, opened 2019-01-28
Fix issue #2475

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2475
| License       | MIT


dunglas:
I'm unsure if we we should remove this `Link` header for `OPTIONS` or not. I prefer to fix the underlying bug (the host is not populated correctly) and let the user filter this header if we wants to using its web-server or reverse proxy config.

vincentchalamon:
Moving to https://github.com/nelmio/NelmioCorsBundle/issues/115

PR #2481, opened 2019-01-29
Allow configuring input and ouput class per operation

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This allows specifying the input and output class per operation.
It is [already possibly](https://github.com/api-platform/core/blob/15103682f8600ccbdca5d119827a076da02d4172/src/Bridge/Symfony/Routing/ApiLoader.php#L126) to specify them for subresources, so I think it was an oversight for operations, which is why i'm classifying this as a bug.

### DIFF
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -206,14 +206,25 @@ final class ApiLoader extends Loader
         $path = trim(trim($resourceMetadata->getAttribute('route_prefix', '')), '/');
         $path .= $this->operationPathResolver->resolveOperationPath($resourceShortName, $operation, $operationType, $operationName);
 
+        $inputClass = $resourceMetadata->getAttribute('input_class', $resourceClass);
+        $outputClass = $resourceMetadata->getAttribute('output_class', $resourceClass);
+
+        if (OperationType::ITEM === $operationType) {
+            $inputClass = $resourceMetadata->getItemOperationAttribute($operationName, 'input_class', $inputClass);
+            $outputClass = $resourceMetadata->getItemOperationAttribute($operationName, 'output_class', $outputClass);
+        } else {
+            $inputClass = $resourceMetadata->getCollectionOperationAttribute($operationName, 'input_class', $inputClass);
+            $outputClass = $resourceMetadata->getCollectionOperationAttribute($operationName, 'output_class', $outputClass);
+        }
+
         $route = new Route(
             $path,
             [
                 '_controller' => $controller,
                 '_format' => null,
                 '_api_resource_class' => $resourceClass,
-                '_api_input_class' => $resourceMetadata->getAttribute('input_class', $resourceClass),
-                '_api_output_class' => $resourceMetadata->getAttribute('output_class', $resourceClass),
+                '_api_input_class' => $inputClass,
+                '_api_output_class' => $outputClass,
                 sprintf('_api_%s_operation_name', $operationType) => $operationName,
             ] + ($operation['defaults'] ?? []),
             $operation['requirements'] ?? [],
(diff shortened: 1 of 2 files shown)


bendavies:
unrelated CI failure

dunglas:
I don't really understand why we need that. It should already work actually

bendavies:
I want to use a different DTO for creation and updating a resource. If you have a better idea I'm happy to hear it.

I couldn't get it to work at all per operation.

soyuka:
looks legit to me

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -206,14 +206,25 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
         $path = trim(trim($resourceMetadata->getAttribute('route_prefix', '')), '/');
         $path .= $this->operationPathResolver->resolveOperationPath($resourceShortName, $operation, $operationType, $operationName);
 
+        $inputClass = $resourceMetadata->getAttribute('input_class', $resourceClass);
+        $outputClass = $resourceMetadata->getAttribute('output_class', $resourceClass);
+
+        if (OperationType::ITEM === $operationType) {
+            $inputClass = $resourceMetadata->getItemOperationAttribute($operationName, 'input_class', $inputClass);
```
antograssiot:
```suggestion
            $inputClass = $resourceMetadata->getItemOperationAttribute($operationName, 'input_class', $resourceClass, true);
```

you could get rid of the lines 209-210 and use this syntax everywhere I think (inlining the specification of the default value and the attribute fallback)

dunglas:
I don't understand why we need a request attribute for that. As the operation is known at this point, we should just be able to lookup the attribute value from the operation metadata (using the metadata factory). Or am I missing something?

bendavies:
I'm not really sure, but this is where I traced to where input/output classes are set. Maybe @soyuka can comment.
the original PR is here https://github.com/api-platform/core/pull/2235

bendavies:
Can you try?

I have tried exactly that, it does not work. here is my actually attempt:

```php
/**
 * @ApiResource(
 *     normalizationContext={"groups": {"account:read"}},
 *     collectionOperations={
 *         "get": {"method": "GET", "path": "/accounts/"},
 *     },
 *     itemOperations={
 *         "get": {"output_class": "ThrowAnError", "method": "GET", "path": "/accounts/{id}", "requirements": {"id": "%uuid_regex%"}},
 *         "put": {"input_class": "ThrowAnError", "method": "PUT", "path": "/accounts/{id}", "requirements": {"id": "%uuid_regex%"}, "denormalization_context": {"groups": {"account:update"}}}
 *     }
 * )
 */
```

Which should throw an error, obviously

soyuka:
@dunglas we were not using the operation configuration but rather the resource-level one:

https://github.com/soyuka/core/blob/cb5421abdb19392ac225909f63cf8677583cedd3/src/Bridge/Symfony/Routing/ApiLoader.php#L215-L216 

Therefore, `/** @ApiResource(itemOperations={"get"={"output_class"="FooBar"}})*/` is not supported yet, `@ApiResource(input_class="")` is though.

src/Bridge/Symfony/Routing/ApiLoader.php
bendavies: nice, didn't notice that. thanks

dunglas:
The input class shouldn't be in the request attribute but in our own metadata We must change this behavior before tagging 2.4.

bendavies:
Closing due to #2483

PR #2482, opened 2019-01-31
Fix instantiating object with discriminator mapping

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

The bug here is:
if you are trying to update a resource that has discriminator mapping, using a URI that is mapped on the abstract class, the class discriminator mapping resolution may fail because when updating a resource, there is no need to send the discriminator in the request.
Instead we should prioritize `OBJECT_TO_POPULATE` as it is already known which resource we have, via the Read Listener.

### DIFF
--- a/features/main/crud_abstract.feature
+++ b/features/main/crud_abstract.feature
@@ -80,8 +80,8 @@ Feature: Create-Retrieve-Update-Delete on abstract resource
     """
 
   Scenario: Update a concrete resource
-      When I add "Content-Type" header equal to "application/ld+json"
-      And I send a "PUT" request to "/concrete_dummies/1" with body:
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/concrete_dummies/1" with body:
       """
       {
         "@id": "/concrete_dummies/1",
@@ -89,11 +89,11 @@ Feature: Create-Retrieve-Update-Delete on abstract resource
         "name": "A nice dummy"
       }
       """
-      Then the response status code should be 200
-      And the response should be in JSON
-      And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
-      And the header "Content-Location" should be equal to "/concrete_dummies/1"
-      And the JSON should be equal to:
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the header "Content-Location" should be equal to "/concrete_dummies/1"
+    And the JSON should be equal to:
       """
       {
         "@context": "/contexts/ConcreteDummy",
@@ -102,6 +102,32 @@ Feature: Create-Retrieve-Update-Delete on abstract resource
         "instance": "Become real",
         "id": 1,
         "name": "A nice dummy"
+      }
+      """
+
+  Scenario: Update a concrete resource using abstract resource uri
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/abstract_dummies/1" with body:
+      """
+      {
+        "@id": "/concrete_dummies/1",
+        "instance": "Become surreal",
+        "name": "A nicer dummy"
+      }
+      """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the header "Content-Location" should be equal to "/concrete_dummies/1"
+    And the JSON should be equal to:
+      """
+      {
+        "@context": "/contexts/ConcreteDummy",
+        "@id": "/concrete_dummies/1",
+        "@type": "ConcreteDummy",
+        "instance": "Become surreal",
+        "id": 1,
+        "name": "A nicer dummy"
       }
       """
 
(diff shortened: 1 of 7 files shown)


bendavies:
of course

bendavies:
@dunglas done

features/main/crud_abstract.feature
```diff
@@ -105,6 +105,31 @@ Feature: Create-Retrieve-Update-Delete on abstract resource
...
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/abstract_dummies/1" with body:
+      """
+      {
+        "@id": "/concrete_dummies/1",
+        "instance": "Become surreal",
+        "name": "A nicer dummy"
+      }
+      """
+    And the response should be in JSON
```
alanpoulain:
```suggestion
    Then the response should be in JSON
```

bendavies:
unrelated build failure

src/EventListener/WriteListener.php
```diff
@@ -62,12 +63,15 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
                 $event->setControllerResult($persistResult ?? $controllerResult);
 
-                // Controller result must be immutable for _api_write_item_iri
```
dunglas: This change should be reverted isn't it? At least we need to keep the comment.

src/EventListener/WriteListener.php
```diff
@@ -62,12 +63,15 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
...
-                // if it's class changed compared to the base class let's avoid calling the IriConverter
-                // especially that the Output class could be a DTO that's not referencing any route
-                if (null !== $this->iriConverter && (false !== $attributes['output_class'] ?? null) && $attributes['resource_class'] === ($class = \get_class($controllerResult)) && $class === \get_class($event->getControllerResult())) {
+                if (null === $this->iriConverter || false === $attributes['output_class']) {
+                    break;
+                }
+
+                try {
                     $request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($controllerResult));
+                } catch (InvalidArgumentException $e) {
```
dunglas: Throwing an exception on the hot path should be avoided

src/EventListener/WriteListener.php
bendavies: i don't know what that line means specifically, but I don't see why we should not generate and IRI for any controller result.

without this change, these lines will fail:
https://github.com/api-platform/core/pull/2482/files#diff-471e9b113802e057f7f37436181d4b55R121

https://github.com/api-platform/core/pull/2482/files#diff-471e9b113802e057f7f37436181d4b55R152

dunglas: Can you take a look @soyuka? I'm not confident at all in this part of the code base, we may introduce BC breaks...

bendavies: specifically, the `resource_class` here is the abstract class, because we are hitting the operation defined on the abstract class, but the controller result will be the concrete implementation.

bendavies: it sets the Content-Location in the RespondListener

dunglas: Or would be more solid IMP

bendavies: ok, looks not possible. in the `RespondListener`, the controller response is serialized (run after `SerializeListener`), so i have nothing to generate an IRI from. So, i think that's why it's done here in the `WriteListener`

src/EventListener/WriteListener.php
bendavies: it would be a DTO/Resource without an iri i guess? what do you suggest?

bendavies:
nothing to do here from my side as far as i know. let me know!

bendavies:
@dunglas @soyuka rebased. please play special attention to the changes in `WriteListener` to support abstract class hierarchy. 
Is there a better way?

src/EventListener/WriteListener.php
```diff
@@ -78,8 +75,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
                     $hasOutput = false !== $resourceMetadata->getOperationAttribute($attributes, 'output_class', null, true);
                 }
 
-                $class = \get_class($controllerResult);
-                if ($hasOutput && $attributes['resource_class'] === $class && $class === \get_class($event->getControllerResult())) {
+                if ($hasOutput) {
```
bendavies: how about 
```php
if ($hasOutput && ResourceClassResolver->isResourceClass(\get_class($event->getControllerResult())) {`
```

?

dunglas: We're changing this code

bendavies: @soyuka any comments here after your input/output dto changes?

bendavies:
rebased

src/EventListener/WriteListener.php
soyuka: I added a behat test for this particular case so if it doesn't break this looks okay!

bendavies:
@soyuka @dunglas i have no idea why that feature only fails on 7.3 with low deps...

bendavies:
FYI this fails on deps = low because that build runs on sqlite, and doesn't have this fix:
https://github.com/doctrine/dbal/pull/3141/files

bendavies:
suggest in another PR running deps=low on postgres, or something other than sqlite!

soyuka:
Thanks @bendavies !

bendavies:
Thanks!

PR #2489, opened 2019-02-03
Disable Messenger From Config

ragboyjr:
- Added ability to disable messenger from config
- Removed previous check if container has message_bus

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |


ragboyjr:
@dunglas CS fixed, squashed and pushed up.

dunglas:
Thanks @ragboyjr!

teohhanhui:
I don't think it's ideal that the user should get an error and have to manually disable the Symfony Messenger integration.

PR #2490, opened 2019-02-03
Fix Bug Preventing Denormalization of Input Classes with Different Fields

ragboyjr:
Fixes the bug where AbstractItemSerializer would use
the resource class instead of the input class to
determine allowed properties.

Enforced `$context['resource_class']` to be set to
current denormalized class to resolve issue. Figured this
should be the case since AbstractItemNormalizer only works
on resources.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |


ragboyjr:
@dunglas CS fixed. Squashed and pushed up.

dunglas:
Thanks @ragboyjr!

PR #2493, opened 2019-02-04
Doctrine ORM Paginator should provide possibility to get  Query object

desmax:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This pull request adds `getQuery` method to `ApiPlatform\Core\Bridge\Doctrine\Orm\AbstractPaginator`. 
Why: When making Translatable entity from Doctrine Extensions work I had to set hints for Query (https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/doc/translatable.md#using-orm-query-hint). The best way to do this from my point of view was to decorate pagination extension (`api_platform.doctrine.orm.query_extension.pagination`) and the only thing I was missing is a way to manipulate with a Query object. Otherwise I would have to rewrite whole Pagination Extension.

`Doctrine\ORM\Tools\Pagination\Paginator` exposes Query object so why its wrapper from Api Platform should not?

### DIFF
--- a/src/Bridge/Doctrine/Orm/Paginator.php
+++ b/src/Bridge/Doctrine/Orm/Paginator.php
@@ -14,13 +14,14 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
 
 use ApiPlatform\Core\DataProvider\PaginatorInterface;
+use Doctrine\ORM\Query;
 
 /**
  * Decorates the Doctrine ORM paginator.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class Paginator extends AbstractPaginator implements PaginatorInterface
+final class Paginator extends AbstractPaginator implements PaginatorInterface, QueryAwareInterface
 {
     /**
      * @var int
@@ -45,5 +46,13 @@ final class Paginator extends AbstractPaginator implements PaginatorInterface
     public function getTotalItems(): float
     {
         return (float) ($this->totalItems ?? $this->totalItems = \count($this->paginator));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getQuery(): Query
+    {
+        return $this->paginator->getQuery();
     }
 }
--- a/src/Bridge/Doctrine/Orm/QueryAwareInterface.php
+++ b/src/Bridge/Doctrine/Orm/QueryAwareInterface.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
+
+use Doctrine\ORM\Query;
+
+interface QueryAwareInterface
+{
+    /**
+     * Gets the Query object that will actually be executed.
+     *
+     * This should allow configuring options which could only be set on the Query
+     * object itself.
+     */
+    public function getQuery(): Query;
+}


src/Bridge/Doctrine/Orm/AbstractPaginator.php
```diff
@@ -41,6 +41,16 @@ public function __construct(DoctrinePaginator $paginator)
         $this->maxResults = $maxResults;
     }
 
+    /**
+     * Returns the query.
+     *
+     * @return Query
```
alanpoulain: Please use typehint instead of PHPDoc

desmax: @alanpoulain updated

soyuka:
> Each time you call getQuery from QueryBuiler it gives you new object. Thats why all the Hints get lost on the way to Paginator.

Exactly, usually I'm indeed rewriting the pagination extension. I'm not sure that it's a good idea to decorate it, for example if you don't use the pagination on the resource your decorator will not get used. 
I've tried to simplify the pagination so that we could re-use it easily but my attempts failed.
Anyway I'm not against adding this method but it may not be appropriate in some cases.

meyerbaptiste:
I just had the same use case as @desmax: change the query hint for the Doctrine paginator. This change could really be helpful and it would avoid some weird hacks...

src/Bridge/Doctrine/Orm/AbstractPaginator.php
```diff
@@ -41,6 +41,11 @@ public function __construct(DoctrinePaginator $paginator)
         $this->maxResults = $maxResults;
     }
 
+    public function getQuery(): Query
```
soyuka: This can actually be a good idea, although we need this in the interface we're implementing. To avoid a BC break you need to create a new interface `QueryAwarePartialPaginator` (that extends the current `PartialPaginatorInterface`).

desmax:
@soyuka I just added the interface as requested.
I also based my branch on 2.4, is this correct?

src/DataProvider/QueryAwarePartialPaginatorInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataProvider;
```
teohhanhui: This file must be moved.

```suggestion
namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
```

desmax: @teohhanhui Updated

desmax:
updated

src/Bridge/Doctrine/Orm/QueryAwarePartialPaginatorInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Orm;
+
+use ApiPlatform\Core\DataProvider\PartialPaginatorInterface;
+use Doctrine\ORM\Query;
+
+interface QueryAwarePartialPaginatorInterface extends PartialPaginatorInterface
```
alanpoulain: Why `Partial` if it is used in the `Paginator`?

desmax: the name was suggested by @soyuka. Not sure here.

alanpoulain: I think it's because the method was on the `AbstractPaginator` before. @soyuka do you confirm `QueryAwarePaginatorInterface` would be better?

teohhanhui: I think the `Partial` part of the name is redundant, even though I understand why @soyuka suggested that.

teohhanhui: It should not extend `PartialPaginatorInterface`.

src/Bridge/Doctrine/Orm/QueryAwarePartialPaginatorInterface.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function getQuery(): Query;
```
teohhanhui:
```suggestion
    /**
     * Gets the Query object that will actually be executed.
     *
     * This is necessary as some options could only be set on the Query object itself.
     */
    public function getQuery(): Query;
```

src/Bridge/Doctrine/Orm/QueryAwarePartialPaginatorInterface.php
teohhanhui: Perhaps we could just make it `QueryAwareInterface`. It could be used in other places as necessary, not tied to the `Paginator`.

soyuka: I like QueryAwarePartialPaginatorInterface if you indeed extend the AbstractPaginator, at least we know what we talk about by looking at the name no?

teohhanhui: The ability to get the `Query` object is not tied to the `Paginator` though... It's just a general interface that gives access to an underlying `Query` object.

desmax: @soyuka @teohhanhui @alanpoulain so what do you decide about this?

teohhanhui: Let's go with a generic `QueryAwareInterface`.

src/Bridge/Doctrine/Orm/QueryAwarePartialPaginatorInterface.php
teohhanhui: @desmax Please add the documentation as I've provided.

teohhanhui:
Anyway, this needs to target `master`, not `2.4`.

src/Bridge/Doctrine/Orm/QueryAwarePartialPaginatorInterface.php
desmax: @alanpoulain updated, though I don't understand why last two words have to be on a new line.

teohhanhui: It's just a hard line break at 80 characters, for readability.

teohhanhui: Sorry, I think you've misunderstood something. The PHPDoc should be on the interface, not the implementation.

desmax: @teohhanhui all updated, based it on master

teohhanhui:
We'd have to merge `2.4` into `master` first to fix the build. Please bear with us.

teohhanhui:
@desmax You can rebase again and it should be good.

soyuka:
Nice, thank you @desmax !

Issue #2488, opened 2019-02-03
Input/Output Class Semantics


ragboyjr:
I'd like to just discuss the overall feature of input/output classes and how they relate to API Platform as a whole. I feel like there are a few blind spots that should be considered and documented accordingly.

### Should Input Classes always be ApiP Resources?

I think it should be a hard requirement that Input classes should be marked as Resources since it can cause issues/inconsistencies if not.

1. **Only Resources can use the AbstractItemNormalizer**

    If input classes aren't resources, you lose the ability to denormalize other relationship objects from IRI string values.

    e.g.

    ```php
    // example input class
    class CreateBookRequest {
        private $name;
        private $author;
        public function __construct(string $name, Author $author) {
            //
        }
    }
    ```

    Input Data
    ```json
    {
        "name": "Book 1",
        "author": "/authors/1"
    }
    ```

    This ONLY works if the input class is also a resource because the AbstractItemNormalizer adds this functionality and it only supports ApiP Resources.
2. Only Resources can use `messenger=true` flag
3. If it's a resource, then we can use the input class resource shortname when building documentation instead of the md5 of its class name. See #2487 

### Are Input Classes Useful outside of Messenger Integration?

Input classes make CQRS a lot easier to achieve within ApiP. You can easily create custom write services while keeping ApiP's defaults for reading which is a best of both world's type of situation. With messenger, you can create a service/command handler that handles the request dto and interact with the domain and create the necessary resource.

This type of service wouldn't be coupled to the framework and usable throughout your system. However, to do the same without Messenger, you'd need to create a Data Persister which feels very framework specific and can be especially odd when you don't ever need to implement the `remove` method of the data persister.

Considering that you can use messenger as a command bus and don't need any external deps like redis/amqp for it to work, it seems like messenger and input classes should be the defacto way for using input classes.

soyuka:
We shouldn't tight input classes with messenger just in case. 

Just getting back on the names: 

> Yes but they are still prefixed by the main resource short name. So you’d have to use two same names input in different namespaces for the same main resource to see this conflict

It's a common thing, for example a `Category` resource that may be used in different namespaces.

ragboyjr:
Regarding names, if input classes are resources, then we can modify the documentation serializer to just use the short name. So you could disambiguate the example you brought up by changing the shortname.

Re messenger w input classes, what if there was an option or something to where you could a create an ApiResource with default config. So, maybe a new annotation like `@ApiDTO` which defaults to something like the following:

[...]

PR #2495, opened 2019-02-05
Improve Input/Output with normalizers

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2488 
| License       | MIT
| Doc PR        | todo

Input/Output through normalizers.
Outputs can be used in collections.
Functional tests through behat with input/output specifications by operation and on the resource.
Tests with relations on DTO classes, the Json-ld normalizer now works with any class (it doesn't have to be a ResourceClass anymore).
When a class is not a resource class but gets denormalized, its ResourceMetadata defaults to a shortname based on the DTO classname. 
When an IRI is needed, the [blank node identifier](https://json-ld.org/spec/ED/json-ld/20180215/#identifying-blank-nodes) is set to : `_:md5(serialize($object))`.


tests/Fixtures/TestBundle/Serializer/Denormalizer/RecoverPasswordDenormalizer.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsDenormalization($data, $type, $format = null, array $context = []): bool
+    {
+        return RecoverPasswordInput::class === ($context['input_class'] ?? null);
+    }
+}
```
soyuka: This example will be used in the documentation. @ragboyjr maybe that you have remarks about this?

ragboyjr: Is part of the new requirements of input/output classes are that we need to create our own normalizers/denormalizers?

soyuka: Yes, I'm working on a graph to explain how things works but it's definitely better then using persisters and providers.

soyuka:
I broke a lot of tests, I'll fix them when/if we validate the changes.

tests/Fixtures/TestBundle/Serializer/Denormalizer/RecoverPasswordDenormalizer.php
ragboyjr: Interesting, I can't speak for everyone, but my use case with input/output classes would be that I'm creating a custom handler for a specific endpoint. So instead of allowing ApiP to use their defaults, I'm intercepting it.

With messenger enabled, this is incredibly elegant, you set your input class, it gets routed to messenger which then routes to a service tagged as a message handler which is responsible for handling the input dto, creating the resource, persist it, then return it.

Without messenger, I feel like the next best solution is you'd set the input class, create a custom controller for the specific operation, and set api_persist=false since your controller will handle it.

With these new changes, that flow gets a lot more complicated since we'd know need to setup a new normalizer/denormalizer to bring our array data into the input dto, and then route it into the system. I'm not sure business logic should be put into data persisters/providers or serializers.

My goal would be that input/output classes get automatically normalized/denormalized according to the custom services created via messenger or via a custom controller.

ragboyjr: So my understanding with api_persist=false was that it disables the write listener. Why would input_class=false disable that? Also `input_class=null` which is the default means use resource_class for input, `input_class=SomeClass` means use that for input, `input_class=false` means disable the write listener?

soyuka: It's a shortcut introduced by @dunglas in https://github.com/api-platform/core/commit/e10d658cb25c00b0f7172b5f9c74b47993ec181c#diff-0170cc702a42034d94c5d83ccadad1deR81. It totally makes sense when you say "output_class=false" it means that you don't want any output, the response will therefore be empty. With `input_class=false` it says "don't read my data" (skips the deserialization).

ragboyjr: Gotcha, that makes sense, I think you meant `output_class=class` is the same as `api_persist=false` not `input_class=false`.

soyuka: Yes sorry!

ragboyjr:
@soyuka are you still moving forward with the normalizer/denormalizer design for handling input/output classes?

soyuka:
Yes it's almost RFR waiting for tests to pass!

soyuka:
Yes it's almost RFR waiting for tests to pass!

ragboyjr:
@soyuka Can you provide an example of you foresee the new system working with normalizers? It looks as if this PR expects us to put business logic in normalizers and treat them more like factories or something, was just a bit concerned on that point.

soyuka:
I've put my example in a separated commit for you: https://github.com/api-platform/core/pull/2495/commits/ae5ab60589efdbd8bdc1984da926eb90c1286e1e

I want to use these for the docs after.

ragboyjr:
Thanks @soyuka 

Going off of the examples you provided in that commit, I don't think using normalizers to utilize input/output classes is the best way to integrate these features for a few reasons:

1. Normalizers are library code responsible for transforming objects into arrays and vice versa. Extending that to also perform business logic seems like it would break Single Responsibility Principal i.e. we are both normalizing an array of objects AND performing critical business logic.
2. Normalizing is meant to be more of "pure" process meaning no IO or side affects should occur during it. Creating a normalizer that performs side affects could definitely cause issues related to the Liskov Substitution Principle i.e. expecting a normalization to be pure operation that can be done over and over may no longer work when embedding business logic inside of the process.
3. In the context of implementing CQRS (which is the main reason IMHO input/output classes are used), creating a normalizer or data persister feels like the wrong place to be making these hook points for writing commands into your domain model. It's not very self evident that normalizing an array into a CreateBookRequest (input class) would actually perform business logic for persisting into the system. One would expect to just return the CreateBookRequest object.
  
   Additionally, when using normalizers/data persisters, In both cases you need to create a class which implements unused methods in order to take control of the write model. Message handlers or a custom controller action would provide a nice explicit API that can be used outside of the context of ApiP and better the fit the mentality for writing commands into system.

If I may offer an alternative solution (I'd also be very willing to code up this feature and submit a PR), I suggest we should focus on streamlining the configuration for using custom controller actions and/or messenger handlers for the standard ways of utilizing input/output classes per resource operations.

Message Handlers, like Custom Controllers are very simple to make and autowire with SF out of the box. Controllers and Message Handlers can easily wrap a domain or application service, and for dx, we'd even be able to expose application services as Controllers or Message Handlers.

The moment a user wants to use a custom input class, they are already relinquishing the automation from API Platform in favor of an explicit hand written write model (while not sacrificing features with read model). So utilizing a custom controller action or message handler would better fit with expectations on implementing custom write model.

soyuka:
> Normalizers are library code responsible for transforming objects into arrays and vice versa. 

Not necessarily, IMO we can extend that (de)normalization principle to something wider, take something in input and give back something else on output. We're already doing this with identifiers for example.

Actually, let me illustrate the more complex case we're facing today with input/output classes. Say we have a `PUT` request, where the developer wants:

1. A custom input
2. A custom output

We draw multiple diagrams with @dunglas and we ended up with the following:

api-platform-put-i-o

From that on, `GET`, `POST` operations can be easily deduced and are easy to set up. 

> It's not very self evident that normalizing an array into a CreateBookRequest (input class) would actually perform business logic for persisting into the system. One would expect to just return the CreateBookRequest object.

There is no persisting logic actually, and indeed the normalization would just return the Input or Output class regarding which side we're on. Or maybe I missed something.

About controller's and messenger handler's, they could be a good alternative but in my above example they can still be added to the stack and Api Platform doesn't depend on them. We need to think a bit outside the symfony's box here because we also don't want to be to tight to symfony, for example Api Platform needs to work with laravel at some point (joins the https://github.com/api-platform/core/pull/2506 event simplification proposal).

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -257,8 +257,8 @@ private function getHydraOperation(string $resourceClass, ResourceMetadata $reso
         }
 
         $shortName = $resourceMetadata->getShortName();
-        $inputClass = $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, 'input_class', null, true);
-        $outputClass = $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, 'output_class', null, true);
+        $inputClass = $resourceMetadata->getTypedOperationIOClass($operationType, $operationName, 'input', false);
```
dunglas: We could maybe use constants for `input` / `output`?

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -42,9 +42,9 @@ final class ItemNormalizer extends AbstractItemNormalizer
     private $componentsCache = [];
     private $resourceMetadataFactory;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [])
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [], bool $allowDtoClass = false)
```
dunglas:
```suggestion
    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [], bool $allowIOClass = false)
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -119,7 +137,15 @@ public function normalize($object, $format = null, array $context = [])
      */
     public function supportsDenormalization($data, $type, $format = null)
     {
-        return $this->localCache[$type] ?? $this->localCache[$type] = $this->resourceClassResolver->isResourceClass($type);
+        if ('elasticsearch' === $format) {
```
dunglas: It's weird isn't it?

soyuka: change context

dunglas:
The problem with our approach is that we breaks the signature of the Symfony normalizer: `@return array|string|int|float|bool` (https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Normalizer/NormalizerInterface.php#L33).

What we could maybe do is:

1. Create new `InputToResourceTransformerInterface` and `ResourceToOutputTransformerInterface`
2. Directly in `ItemNormalizer`, if `output` or `input` is defined, call the transformer


I think it will also address @ragboyjr concerns and it's slight change. WDYT?

ragboyjr:
OK, @soyuka thank you for the flow chart, it helps me understand a lot more where you guys are coming from.

First, @dunglas using transformers to go from DTO -> resource after the denormalization step makes a LOT more sense IMHO. Users wouldn't have to make denormalizers, they could just make transformers, and those could probably setup as simply static methods on the resources to provide that transformation as well.

Regrading the flow chart, I think understand why we have the disconnect. From my limited perspective, input classes served as an escape hatch from ApiP for supporting CQRS, the idea being, if you want to create your own service to handle persistence of a resource from an input class, you could. 

Whereas It looks like you guys are trying to automate a lot of that process with the idea of transformers.

The way I envisioned is like the following: 

screen shot 2019-02-11 at 8 53 57 am

The parts in yellow showing the custom code a user would write.

ApiP already has the concept of readable/writeable properties which allow for different schemas to be created vs read. I'm not sure how much use input/output transformers would get since this PR doesn't seem to address the use case of making CQRS integration more seamless for the end user.

src/DataTransformer/ChainInputDataTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataTransformer;
+
+/**
+ * Transforms an Input to a Resource object.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class ChainInputDataTransformer implements DataTransformerInterface
```
soyuka: I also have a version with a local cache.

features/graphql/mutation.feature
```diff
@@ -323,29 +323,23 @@ Feature: GraphQL mutation support
     {
       "errors": [
         {
-          "message": "Field \"foo\" is not defined by type createDummyDtoNoInputInput.",
```
soyuka: @alanpoulain I think that this is better, see also the related change: https://github.com/api-platform/core/blob/e299566b5c4ee8c1c2791e50c5204146632435e9/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php#L97-L99

WDYT?

src/Serializer/AbstractItemNormalizer.php
ismail1432:
```php
 if (ApiPlatform\Core\Bridge\Elasticsearch\Serializer\ItemNormalizer::FORMAT === $format)
```

features/graphql/mutation.feature
dunglas: Then we should update the GraphQL Schema accordingly, and it will become impossible to trigger this code path.

src/DataTransformer/ChainInputDataTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+            if ($transformer->supportsTransformation($object, $context)) {
+                return $transformer->transform($object, $context);
+            }
+        }
+
+        return $object;
+    }
+
+    /**
+     * Creates a serialization context from a Request.
```
dunglas: Should changed I guess :P

src/DataTransformer/ChainOutputDataTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Transforms an Output to a Resource object.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class ChainOutputDataTransformer implements DataTransformerInterface
+{
+    private $transformers;
+
+    public function __construct(\Traversable $transformers)
```
dunglas: `iterable`?

src/DataTransformer/ChainOutputDataTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $this->transformers = $transformers;
+    }
+
+    /**
+     * Creates a serialization context from a Request.
+     */
+    public function transform($object, array $context = [])
```
dunglas:
```suggestion
    public function transform(object $object, array $context = [])
```

src/DataTransformer/DataTransformerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ * Transforms a DTO or an Anonymous class to a Resource object.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+interface DataTransformerInterface
+{
+    /**
+     * Creates a serialization context from a Request.
+     */
+    public function transform($object, array $context = []);
```
dunglas:
```suggestion
    public function transform($object, array $context = []): object;
```

src/DataTransformer/DataTransformerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Transforms a DTO or an Anonymous class to a Resource object.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+interface DataTransformerInterface
+{
+    /**
+     * Creates a serialization context from a Request.
```
dunglas: To update

src/EventListener/DeserializeListener.php
```diff
@@ -78,7 +82,7 @@ public function onKernelRequest(GetResponseEvent $event)
         }
 
         $context = $this->serializerContextBuilder->createFromRequest($request, false, $attributes);
-        if (false === $context['input_class']) {
+        if (isset($context['input']) && null === $context['input']['class']) {
```
dunglas: I fear that you have to use `array_key_exists` here. Here if input is defined nut not input.class, it will fail.

src/EventListener/DeserializeListener.php
```diff
@@ -94,12 +98,15 @@ public function onKernelRequest(GetResponseEvent $event)
...
-        $request->attributes->set(
-            'data',
-            $this->serializer->deserialize(
-                $requestContent, $context['input_class'], $format, $context
-            )
+        $data = $this->serializer->deserialize(
+            $requestContent, $context['input']['class'] ?? $context['resource_class'], $format, $context
         );
+
+        if (null !== $this->dataTransformer && $this->dataTransformer->supportsTransformation($data, $context)) {
```
dunglas: Maybe could we store if an input class is defined, and skip the call to `supportsTransformation` if not (because this call is on the hot path).

src/EventListener/SerializeListener.php
```diff
@@ -84,6 +82,10 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
         $request->attributes->set('_api_normalization_context', $context);
 
+        if (null !== $this->dataTransformer && $this->dataTransformer->supportsTransformation($controllerResult, $context)) {
```
dunglas: same here

src/EventListener/DeserializeListener.php
dunglas: By the way, shouldn't this call be done in the normalizer instead of here?

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -93,8 +93,9 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
 
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
             $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $item, $operationName);
-            if (false === $resourceClass = $resourceMetadata->getAttribute('input_class', $resourceClass)) {
-                return null;
+
+            if (null === $resourceClass = $resourceMetadata->getAttribute('input', ['class' => $resourceClass])['class']) {
+                throw Error::createLocatedError('Input is disabled.', $info->fieldNodes, $info->path);
```
dunglas: Shouldn't the GraphQL Schema be updated to not allow to trigger such code path?

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -381,10 +381,13 @@ private function convertType(Type $type, bool $input = false, string $mutationNa
...
-                if (false !== $dtoClass = $resourceMetadata->getAttribute($input ? 'input_class' : 'output_class', $resourceClass)) {
-                    $resourceClass = $dtoClass;
+                $ioClass = $resourceMetadata->getAttribute($input ? 'input' : 'output');
+
+                if (isset($ioClass['class']) && null !== $ioClass['class']) {
+                    $resourceClass = $ioClass['class'];
                 }
-                $graphqlType = $this->getResourceObjectType(false === $dtoClass ? null : $resourceClass, $resourceMetadata, $input, $mutationName, $depth);
+
+                $graphqlType = $this->getResourceObjectType(isset($ioClass['class']) && null === $ioClass['class'] ? null : $resourceClass, $resourceMetadata, $input, $mutationName, $depth);
```
dunglas: Same here, you'll have to use `array_key_exists` to make the code more robust

src/EventListener/WriteListener.php
```diff
@@ -75,7 +75,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
                 $hasOutput = true;
                 if (null !== $this->resourceMetadataFactory) {
                     $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
-                    $hasOutput = false !== $resourceMetadata->getOperationAttribute($attributes, 'output_class', null, true);
+                    $hasOutput = null !== $resourceMetadata->getOperationAttribute($attributes, 'output', ['class' => $attributes['resource_class']], true)['class'];
```
dunglas: The `class` key may not be defined

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -42,9 +42,9 @@ final class ItemNormalizer extends AbstractItemNormalizer
     private $componentsCache = [];
     private $resourceMetadataFactory;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [])
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [], bool $allowAnonymousClass = false)
```
dunglas: `allowUnmappedClass`?

src/JsonLd/AnonymousContextBuilderInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+interface AnonymousContextBuilderInterface extends ContextBuilderInterface
+{
+    /**
+     * Creates a JSON-LD context based on the given object.
+     * Usually this is used with an Input or Output DTO object.
+     */
+    public function getAnonymousResourceContext($object, array $context, int $referenceType = UrlGeneratorInterface::ABS_PATH): array;
```
dunglas:
```suggestion
    public function getAnonymousResourceContext(object $object, array $context, int $referenceType = UrlGeneratorInterface::ABS_PATH): array;
```

src/DataTransformer/ChainOutputDataTransformer.php
soyuka: Type declaration 'iterable' must be unqualified

src/DataTransformer/ChainOutputDataTransformer.php
soyuka: 7.2 only for object typing, I wanted to type it, also the return type

dunglas: Ok! At least at the PHPDoc then.

src/EventListener/WriteListener.php
soyuka: nope because if it's `null` it defaults to `$attributes['resource_class']`

kiler129:
Quick question about this PR: does it just solve input/output classes or DTOs in general? In other words let's say you have:

```php
/**
 * @ApiResource
 */
class Foo {
    /**
      * @var DTOWhichIsNotAResourceClass
      */
    public $bar;
}
```

do you still have to define `DTOWhichIsNotAResourceClass` as a fake resource without any operations?

dunglas:
@kiler129 it solves both!

soyuka:
> @kiler129 it solves both!

Added a test to be sure and yes this works now :).

src/EventListener/SerializeListener.php
```diff
@@ -34,11 +35,13 @@ final class SerializeListener
...
     private $serializer;
     private $serializerContextBuilder;
+    private $dataTransformer;
 
-    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder)
+    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, DataTransformerInterface $dataTransformer = null)
     {
         $this->serializer = $serializer;
         $this->serializerContextBuilder = $serializerContextBuilder;
+        $this->dataTransformer = $dataTransformer;
```
dunglas: Unused

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -418,12 +415,20 @@ private function getResourceObjectType(?string $resourceClass, ResourceMetadata
             return $this->graphqlTypes[$shortName];
         }
 
+        $ioMetadata = $resourceMetadata->getAttribute($input ? 'input' : 'output');
+
+        if (null !== $ioMetadata && \array_key_exists('class', $ioMetadata)) {
+            if (null !== $ioMetadata['class']) {
```
dunglas: Can be merged with the previous line

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -37,16 +39,15 @@
...
 {
+    use ClassInfoTrait;
+
     const FORMAT = 'jsonapi';
 
     private $componentsCache = [];
-    private $resourceMetadataFactory;
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [])
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory, array $defaultContext = [], bool $allowUnmappedClass = false, DataTransformerInterface $dataTransformer = null)
```
dunglas: `$allowUnmappedClass` should be the latest parameter to create a better upgrade path

soyuka:
I'll add some more unit tests on `dataTransformer` to improve coverage but we're close to merging this :).

soyuka:
Related documentation reviewed: https://github.com/api-platform/docs/pull/731

features/bootstrap/DoctrineContext.php
```diff
@@ -1182,6 +1184,36 @@ public function thereIsAMaxDepthDummyWithLevelOfDescendants(int $level)
         $this->manager->flush();
     }
 
+    /**
+     * @Given there is a DummyCustomDto
+     */
+    public function thereIsADummyCustomDto()
```
alanpoulain: Is this step really useful since you can use the second one?

features/main/input_output.feature
```diff
@@ -0,0 +1,212 @@
+Feature: DTO input and output
+  In order to use an hypermedia API
```
alanpoulain:
```suggestion
  In order to use a hypermedia API
```

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -51,7 +52,23 @@ public function normalize($object, $format = null, array $context = [])
...
+        if (null !== $outputClass && null !== $this->dataTransformer && $this->dataTransformer->supportsTransformation($object, $outputClass, $context)) {
+            $object = $this->dataTransformer->transform($object, $outputClass, $context);
+        }
+
+        try {
+            $resourceClass = $this->resourceClassResolver->getResourceClass($object, $context['resource_class'] ?? null, true);
+        } catch (InvalidArgumentException $e) {
+            $context = $this->initContext(\get_class($object), $context);
+            $rawData = parent::normalize($object, $format, $context);
+            if (!\is_array($rawData)) {
```
meyerbaptiste: Useless? You return `$rawData` in all cases.

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -74,7 +80,17 @@ public function normalize($object, $format = null, array $context = [])
         }
 
         // Get and populate item type
-        $resourceClass = $this->resourceClassResolver->getResourceClass($object, $context['resource_class'] ?? null, true);
+        try {
+            $resourceClass = $this->resourceClassResolver->getResourceClass($object, $context['resource_class'] ?? null, true);
+        } catch (InvalidArgumentException $e) {
+            $rawData = parent::normalize($object, $format, $context);
+            if (!\is_array($rawData)) {
```
meyerbaptiste: Useless?

src/JsonLd/ContextBuilder.php
```diff
@@ -87,9 +87,45 @@ public function getEntrypointContext(int $referenceType = UrlGeneratorInterface:
...
+
+        return $this->urlGenerator->generate('api_jsonld_context', ['shortName' => $resourceMetadata->getShortName()], $referenceType);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAnonymousResourceContext($object, array $context, int $referenceType = UrlGeneratorInterface::ABS_PATH): array
+    {
+        $id = $context['iri'] ?? '_:'.spl_object_id($object);
```
meyerbaptiste: Using `spl_object_id()` drops support for PHP 7.1. Otherwise you have to add `symfony/polyfill-php72` as dependency...

features/main/input_output.feature
```diff
@@ -0,0 +1,212 @@
+Feature: DTO input and output
+  In order to use a hypermedia API
+  As a client software developer
+  I need to be able to use DTOs on my resources as Input or Output objects.
+
+  @createSchema
+  Scenario: Create a resource with a custom Input.
```
alanpoulain:
```suggestion
  Scenario: Create a resource with a custom Input
```

features/main/input_output.feature
```diff
@@ -0,0 +1,212 @@
...
+        "bat": {
+          "@type": "@id"
+        }
+      },
+      "@type": "OutputDto",
+      "baz": 1,
+      "bat": "test"
+    }
+    """
+    Then I add "Content-Type" header equal to "application/ld+json"
```
alanpoulain:
```suggestion
    When I add "Content-Type" header equal to "application/ld+json"
```

features/main/input_output.feature
```diff
@@ -0,0 +1,212 @@
...
+        }
+      },
+      "@type": "OutputDto",
+      "baz": 2,
+      "bat": "test"
+    }
+    """
+
+  @createSchema
+  @!mongodb
```
alanpoulain: Why? FOSUser should be working with MongoDB.

features/main/input_output.feature
```diff
@@ -0,0 +1,212 @@
...
+    And I send a "POST" request to "/users" with body:
+    """
+    {
+      "username": "soyuka",
+      "plainPassword": "a real password",
+      "email": "soyuka@example.com"
+    }
+    """
+    Then the response status code should be 201
+    Then I add "Content-Type" header equal to "application/ld+json"
```
alanpoulain:
```suggestion
    When I add "Content-Type" header equal to "application/ld+json"
```

tests/Metadata/Resource/Factory/InputOutputResourceMetadataFactoryTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $factory = new InputOutputResourceMetadataFactory($decorated);
+        $this->assertSame($expected, $factory->create('Foo')->getAttributes()['input']);
+    }
+
+    public function getAttributes(): array
+    {
+        return [
+            // no input class defined
+            [[], null],
```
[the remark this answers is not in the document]
soyuka: it's not necessary an error, more that the metadata will be defined like this.

features/main/input_output.feature
soyuka: Because it's hard to make this test work on both. Small example, I declare a "fake" relation to `Dummy` in `RecoverPasswordOutput` with a doc block:

```
    /**
     * @var \ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy
     */
```

I can't specify `Dummy` and `DummyDocument` here, the property metadata factory will take the first type available. This means that the result will not be normalized the same way. 

If I wanted to make this test work on mongodb I have to duplicate everything, there's no point because the only thing that changes is the data provider and yes it definitely works.

soyuka:
Merging this, please don't hesitate to leave more comments we can still come back to them before a stable release!

Devristo:
@soyuka thanks for your work on this PR, it looks like it was quite a challenge!

Even though I tried to follow the discussion with @ragboyjr and @dunglas above. It's still not clear how an input_class DTO can be handled by the messenger persister.

Let's say I have a `Todo` resource. I would like the GET, DELETE handled by the API-platform. However for POSTing a new `Todo` I would like to use a `CreateTodoDto` and the Symfony Messenger integration instead. For example to do this asynchronously.

I assumed I could set the `POST` input class to `CreateTodoDto` (is not a resource) and add `messenger=true` to the `Todo` resource. However it seems that now the `AbstractItemNormalizer` enforces me to deserialize the request body to a `Todo` resource instead? 

Before this MR was merged I was able to get this to work, albeit a minor workaround which I no longer can recall.

In my understanding after this MR I need to create two resources to work around this. A `CreateTodoDto` resource which is 'persisted' by the messenger and the normal `Todo` resource. By default this means different URLs, however this URL can be altered to lets say `POST /todos`.

This feels dirty to me, am I missing the correct approach?

soyuka:
In fact nothing changed in how resources are persisted. 

Say we leave the messenger on the side for now. When you send a different representation of a class (eg use an `input` class) through a `POST` it does:

```
=> POST /todos `{ Input }`
=> Input transformed to Todo
=> Persist Todo
```

When you add the `messenger` flag, instead of using doctrine to persist it'll just send the resource through the messenger instead:

```
=> POST /todos `{ Input }`
=> Input transformed to Todo
=> Handle Todo message
```

Can you show me the difference between your input and your resource?

Devristo:
Thank you for your reply @soyka.

In my 'CQRS' scenario I would like to handle an instance of the input class as a message. Such that a `PUT {INPUT}` and `POST {DIFFERENT_INPUT}` end up in two separate messages and can be handled accordingly.

Now I can set different input classes for both PUT and POST, however in the end the message handler will always receive a `Todo`. At this point I no longer know have the input data I need for a PUT or a POST. Basically I cannot do CQRS like this, since I lost the domain event.

For now I have created separate `CreateTodo` and `UpdateTodo` resources, such that I can handle / persist these as separate messages. But in my opinion conceptually they should be operations of the same `Todo` resource. 

I think the approach @ragboyjr would make much more sense in CQRS-setup. Although I am fully aware that the scope of this project is much wider, it would be nice if there was a proper configuration for it.

soyuka:
I've found a good solution, it just needs a quick fix see https://github.com/api-platform/core/pull/2590.
Take a look at the following gist: https://gist.github.com/soyuka/312526207073dc0edb6add81d15f6249

Devristo:
This seems to work! I could even alter DataTransformer to handle all input-classes of a specific interface. Then I would not have to write one for each specific message.

```php
<?php
namespace App\DataTransformer;

use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
use App\Contract\DomainMessage;

class DomainMessageTransformer implements DataTransformerInterface {
    public function transform($object, string $to, array $context = [])
    {
        return $object;
    }

    public function supportsTransformation($object, string $to, array $context = []): bool
    {
        if (\is_object($object)) {
            return false;
        }

        $className = $context['input']['class'] ?? null;

        if (is_null($className)) {
            return false;
        }

        return is_subclass_of($className, DomainMessage::class);
    }
}
```

soyuka:
Exactly!

kiler129:
It seems (correct me if I'm wrong) this PR introduced a dangerous assumption - `AbstractItemNormalizer` now "owns" serialization of anything in the application, even if it's not an API context at all. 

Was this change intentional that now `ItemNormalizer` is used for **every** serialization? Before it was only used for resource classes, but now it checks `allowUnmappedClass` (which is effectively `true` with immediate deprecation of `false`) and claims serialization of anything.
Also since it has a priority higher than default object normalizer (`-923` vs `-1000`) it will always be checked first.

teohhanhui:
@kiler129 Don't worry. We've changed it so that it only handles API resources (but also non-resources reachable from a resource).

PR #2506, opened 2019-02-10
[Proposal] Internal events

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | ?
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Alternative to https://github.com/api-platform/core/pull/2329.

The aim is the same as the other PR but also to decouple the event system from Symfony as much as possible (to replace it easily with Laravel for instance).

The listeners in `ApiPlatform\Core\EventListener` listen to our internal events instead of the Symfony ones.

The Symfony events are listened by the `EventDispatcher` in the `Bridge` and dispatched as our events. It implies that the `symfony/event-dispatcher` dependency is now mandatory.

It will also simplify the part about priorities (https://api-platform.com/docs/core/events/#the-event-system) since the user has now to only listen to our events.

For existing applications, it should not change anything since listening to Symfony events (with priority) is equivalent to listening to our internal ones. For the listeners should we add deprecations about using Symfony events?

WDYT? @api-platform/core-team?


src/Events.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace ApiPlatform\Core;
+
+/**
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class Events
+{
+    public const PRE_READ = 'api_platform.pre_read';
+    public const READ = 'api_platform.read';
```
alanpoulain: This event will not be used by GraphQL and is internal. Should we expose it elsewhere? Should we call it something else (`api_platform.rest.read`)?

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
```diff
@@ -0,0 +1,47 @@
...
+        $this->eventName = $eventName;
+        $this->eventClass = $eventClass;
+        $this->dispatcher = $dispatcher;
+    }
+
+    public function dispatch(Event $event): void
+    {
+        $internalEvent = null;
+
+        switch ($event) {
```
vincentchalamon: Why creating a switch for just 1 case, instead of an if?

alanpoulain: It's an example for the read event. If the proposal is validated, there will be more cases (I would say two).

src/EventListener/ReadListener.php
```diff
@@ -52,9 +51,9 @@ public function __construct(CollectionDataProviderInterface $collectionDataProvi
      *
      * @throws NotFoundHttpException
      */
-    public function onKernelRequest(GetResponseEvent $event)
+    public function handleEvent(EventInterface $event)
```
vincentchalamon: This is a BC break, as you change a public method definition (for example, if any project override this service using service decoration)

alanpoulain: Yes sure. Maybe I will add a deprecation. It's too related to Symfony right now IMO.

src/Events.php
vincentchalamon: WDYT about keeping this event name, and dispatch a `api_platform.query` event for GraphQL?

vincentchalamon: I don't understand how the `api_platform.read` event could be only internal and not listened by any listener in a project

vincentchalamon:
I like the idea of this bridge of the EventDispatcher, it makes API Platform more flexible and compatible with other systems (such as Laravel as you said).
>For the listeners should we add deprecations?

Do you mean adding deprecations about listening to Symfony events in an API Platform project?

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
```diff
@@ -0,0 +1,47 @@
...
+            case $event instanceof GetResponseEvent:
+                $internalEvent = new $this->eventClass(null, ['request' => $event->getRequest()]);
```
Taluu: How about putting the original event in the internal event ?

src/Events.php
alanpoulain: For GraphQL, there will be no event like this dispatched (only pre and post). GraphQL is working differently. Using `query` does not work: there is a read step in the mutation resolver for example.
`api_platform.read` _could_ be listened, but it shouldn't since it is used by our `ReadListener`.

alanpoulain:
@vincentchalamon yes I mean that for the deprecations.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
alanpoulain: Not sure about this. I want to decouple from Symfony as much as possible. What is the use case?

ragboyjr:
@alanpoulain this is huge 👍 from me. One of the main things that confused me when first looking over API platform was that it felt a bit hackish if we wanted to extend specific API platform events because it was based on ensuring priority after api platform event listeners on SF events. 

This seems like it will not only make API Platform easier to fit on other popular frameworks, but will also  make extending API platform specific events easier for users.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
Taluu: The `GetResponseEvent` from Symfony has a response part which is not in the internal event, it could be useful to have it here.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        $this->dispatcher->dispatch($this->eventName, $internalEvent);
```
Taluu: if the internal event is null, how about dispatching the given event instead ?

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
alanpoulain: Yes but it's only to set the response and bypass everything else. I'm not sure we want this in API Platform, do we? If there are a "respose" (controller result) and a request (in `GetResponseForControllerResultEvent` for instance), I will add them both to the context.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
alanpoulain: Why would it be null? I'm thinking of throwing an exception instead (it would mean the service is not configured correctly).

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
Taluu: Yup, but then what if we get an exception event (`GetResponseForExceptionEvent` I think ?), and we want to get the thrown exception ? Same for all the other sub controllers.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
Taluu: As this service is kinda decorating the event dispatcher, if you want to dispatch other events, you will dispatch a `null` event instead. I think this can be a huge source of bugs or miscomprehension.

Unless this service is really for internal...

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
alanpoulain: I will add an `exception` field in the context then. IMO we want as much information as the Symfony event but we don't want to use the event itself since it's too tight to Symfony.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
alanpoulain: Yes I think it's really internal. If an user want to do something like this, it can do it itself. Maybe adding a `@internal` for the class would be a good idea.

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
Taluu: btw it should be `switch (true)` instead, the cases will not be validated otherwise if you test your cases with `instanceof`

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
Taluu: Then the case should be split accordingly depending on the event it handles (exception, controller result, ... and so on)

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class EventDispatcher
+{
+    private $eventName;
+    private $eventClass;
+    private $dispatcher;
+
+    public function __construct(string $eventName, string $eventClass, EventDispatcherInterface $dispatcher)
```
Taluu: so this dispatcher can handle only _one_ event ? you would have for X events (`read`, `write`, ... and so on) X declared services ? Wouldn't that be a mess ? :O

src/Bridge/Symfony/Bundle/EventListener/EventDispatcher.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $this->eventName = $eventName;
+        $this->eventClass = $eventClass;
```
Taluu: not having any check if the event class given is valid is a foot gun IMO

dunglas:
I'm wondering if we the latest addition in the library, we still need an event system. If we look at the diagram created with @soyuka, we already have extension point everywhere we need.

image

Maybe could we just remove the event system from the docs (it will become purely internal), and document how to hook custom logic during the execution flow:

1. custom data provider
2. custom deserializer
3. custom validator
4. custom data persister
5. custom serializer

It's simple, covers all use cases, and is compatible with Laravel, PSR-7 middlewares and so on.

alanpoulain:
@dunglas Why not. But we will need to keep the event listeners anyway, don't we? And it's better if they are decoupled from Symfony.

And I'm not sure if creating a custom persister for instance is a good idea for having a side-effect (send a mail for example). Are you sure it's easier than creating an event listener?

Also I don't think it's only documentation. We need a `ChainValidator` for example don't we?

dunglas:
> But we will need to keep the event listeners anyway, don't we?

We'll always need to plug on the Symfony kernel, anyway.

> Are you sure it's easier than creating an event listener?

I'm not sure, it's just some thought to start the discussion :)

> We need a `ChainValidator` for example don't we?

Not really, the user can create custom validation constraints by using the Symfony component, or implement the `ValidatorInterface` himself if he (or she) wants too.

alanpoulain:
> We'll always need to plug on the Symfony kernel, anyway.

Sure but it will be done in the bridge, and could be easily replaced with another listener. Not the case for the current event listeners because there are too tight to Symfony.

I think it's really easier to use events for side-effets than having to deal with custom services. I think both methods are complementary.

soyuka:
> And I'm not sure if creating a custom persister for instance is a good idea for having a side-effect (send a mail for example). Are you sure it's easier than creating an event listener?

To me this is a job for the messenger.

> I think it's really easier to use events for side-effets than having to deal with custom services. I think both methods are complementary.

Yes but using too many events can become a mess pretty fast. 

Updated schema with transformers:

api-platform-put-i-o

IMO we should try to deprecate the use of events as much as possible. Of course we'd have to keep symfony events but I'm not in favor of adding internal events like this. If we look at the above schema, the end developper should be able to hook using our concepts (persister, provider, transformer) for most of the use cases. 
If the use case is really specific, events may be a good solution but they should not be the developer first choice, which is what internal events tend to give.

alanpoulain:
> To me this is a job for the messenger.

All projects don't always use the `Messenger` component whereas events are pretty common.

> Of course we'd have to keep symfony events but I'm not in favor of adding internal events like this.

If we keep the Symfony events, it's always better to decouple our event listeners from Symfony. This is also the aim of this PR.

> If the use case is really specific, events may be a good solution but they should not be the developer first choice, which is what internal events tend to give.

Why? Listening to our internal events or listening to the Symfony events, I don't really see the difference. If a developer wants to do it, it will always find a way. We can outline in the documentation that this shouldn't be the first choice, but that's all.

jamesisaac:
> > And I'm not sure if creating a custom persister for instance is a good idea for having a side-effect (send a mail for example). Are you sure it's easier than creating an event listener?
>
> To me this is a job for the messenger.

This is over complicating things IMO.  I think it's great that with API Platform it's so simple to get a REST and/or GraphQL API up and running by simply defining the entities and a bit of configuration via annotations.  But unlike similar tools which also provide these capabilities (Prisma, PostGraphile, etc), you can *also very easily hook in custom business logic (e.g. side effects like sending email) through Symfony's events system*.  Deploying to something like Heroku is then just a few clicks away.

If I understand the Messenger component correctly, you're now talking about orchestrating multiple projects/services, replacing function calls with inter-process message queues.  Sure, there's a place for that (bigger projects with lots of devops resources), but for rapidly shipping MVPs, the single project "monolith" approach greatly simplifies things.  IMO, API Platform is currently the best in class for this use-case, and GraphQL events is one of the only gaps in offering full flexibility (as I experienced in #2167).

teohhanhui:
I'd suggest we stick to semantical events, in this case CRUD.

alanpoulain:
@teohhanhui I don't fully grasp what you say. My aim is to have our own events (to be decoupled from Symfony). And with them, users can add hooks if they want.

teohhanhui:
What I mean is, if we're going to have custom events, then those events should have well-defined semantics, not something like "add format" / "deserialize" which is merely an implementation detail. Having CRUD events would be good.

alanpoulain:
I'm not sure what you mean by CRUD events. We need a way for the user to modify things before or after our listeners. And we need these events for ourselves too.

soyuka:
He probably means that the user should be able to hook to:

```
PRE_CREATE
POST_CREATE
PRE_READ
POST_READ
PRE_UPDATE
POST_UPDATE
PRE_DELETE
POST_DELETE
```

This is independent from the implementation, and by the way I've reviewed my thoughts about having internal events and I think it might be good to detach these from symfony.

alanpoulain:
Adding these events later should be really easy with this implementation (just more events to dispatch).
But they can't replace the already existing events.

src/Event/EventInterface.php
```diff
@@ -0,0 +1,28 @@
...
+namespace ApiPlatform\Core\Event;
+
+/**
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+interface EventInterface
+{
+    public function getData();
+
+    public function setData($data): void;
```
dunglas: If we introduce our own event system, I would like to use immutability (as in our metadata system).

src/EventListener/ReadListener.php
```diff
@@ -51,10 +51,33 @@ public function __construct(CollectionDataProviderInterface $collectionDataProvi
...
+
+    /**
+     * Calls the data provider and sets the data attribute.
+     *
+     * @throws NotFoundHttpException
+     */
+    public function handleEvent(/*EventInterface */$event)
+    {
+        if ($event instanceof EventInterface) {
+            $request = $event->getContext()['request'];
```
dunglas: This is a hard coupling to Symfony. We should let the user injecting `RequestStack` instead of populating this key.

src/Events.php
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class Events
+{
+    public const QUERY_PARAMETER_VALIDATE = 'api_platform.query_parameter_validate';
+    public const FORMAT_ADD = 'api_platform.format_add';
+
+    public const PRE_READ = 'api_platform.pre_read';
+    public const READ = 'api_platform.read';
```
dunglas: Couldn't we only have the PRE ad POST events?

alanpoulain: I was thinking that `PRE` and `POST` events were for the user whereas the other ones were for our internal use only.

ragboyjr:
@jamesisaac Regarding using messenger, it's just used as a command bus in here and doesn't necessarily need to be hooked up to an actual async sending system.

teohhanhui:
> We need a way for the user to modify things before or after our listeners. And we need these events for ourselves too.

These listeners are supposed to be the integration with the Symfony framework. We should extract out all logic from the listeners, instead of going the other way round. So I'm strongly -1 for having implementation-specific events. It will tie our hands in the future.

alanpoulain:
Not sure about your position. For me, API Platform is _based_ on events and will always relies on an event system. No need to extract the logic from the listeners, we only need them to be decorrelated from the Symfony framework.

> It will tie our hands in the future.

I don't see why. Our hands are tied _today_ because we are too coupled to Symfony.

teohhanhui:
The way to decouple from Symfony, is to extract all logic out from listeners - "thin listeners". Same thing that we do for controllers - "thin controllers". Then we can integrate with different frameworks by reusing the services that we have. Listeners are just glue code.

alanpoulain:
You can consider this PR as a first step to do this then? First we decouple, then we extract the logic?

teohhanhui:
No, this is completely going in the wrong direction. I've explained in detail above. So I have no more to add.

alanpoulain:
I'm not convinced then and I think this PR is worth finishing. And with it GraphQL will have an event system.

teohhanhui:
I'm not against having our own events. In fact, it'll be wonderful. If the events have well-defined semantics and decoupled from implementation, then I have no objections.

I gave the example of CRUD events, as those are the most obvious (and probably already will provide enough flexibility anyway). But we could come up with other well-defined events. Taking our existing listeners implementation and creating events from those is the opposite of that.

alanpoulain:
You're probably right but I don't want to refactor the planet with this PR. Couldn't it be done in another PR once this one is merged?

Goutte:
> I'm not against having our own events. In fact, it'll be wonderful. If the events have well-defined semantics and decoupled from implementation, then I have no objections.

It would be wonderful to have CRUD events.

This thread was my lifeline for information after many google queries ; could someone be awesome and link to a current discussion about the subject, if any?  @alanpoulain

jamesisaac:
@Goutte https://github.com/api-platform/core/pull/2978 seems to be the latest iteration on this subject

PR #2492, opened 2019-02-03
[GraphQL] Custom types

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes?
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Add a `TypeInterface` to register automatically custom types in GraphQL.
The `IterableType` has been modified to be used this way.
A `TypesFactory` service has been added, along with its interface to allow to register custom types any way you want.
Should we add a deprecation (see my comment below)?

PR #2510, opened 2019-02-12
GraphQL: Prevent an error when the type factory is null

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Fixes a bug introduced by #2492.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.5.0 beta 1
+
+* GraphQL: Add support for custom types
+
 ## 2.4.0 beta 1
 
 * MongoDB: full support
--- a/src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
@@ -63,9 +63,9 @@
             <argument type="service" id="api_platform.graphql.resolver.factory.item_mutation" />
             <argument type="service" id="api_platform.graphql.resolver.item" />
             <argument type="service" id="api_platform.graphql.resolver.resource_field" />
+            <argument type="service" id="api_platform.graphql.types_factory" />
             <argument type="service" id="api_platform.filter_locator" />
             <argument>%api_platform.collection.pagination.enabled%</argument>
-            <argument type="service" id="api_platform.graphql.types_factory" />
         </service>
 
         <!-- Action -->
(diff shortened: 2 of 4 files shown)


dunglas:
I don't understand. Either the constructor property isn't nullable, or we must add this test. Anyway there currently is a bug (and it's weird that PHPStan doesn't catch it).
I'm for allowing to pass null for this prop, because it makes it easier to reuse this class without Symfony DI.

soyuka:
Also add to the graphql.xml definition `on-invalid=null`

alanpoulain:
It's because I forgot to remove `= null`. I don't like to keep the previous code (it's not simply a `if`, GraphQL will not work if you don't set the `iterable` type). It means we had to add a deprecated. Same as here: https://github.com/api-platform/core/pull/2447#discussion_r248777048
IMO it's too much hassle for an experimental feature.

dunglas:
> Also add to the graphql.xml definition `on-invalid=null`

Why? This case cannot happen.

> IMO it's too much hassle for an experimental feature.

It makes harder to use API Platform as a standalone lib. I don't get why we need to deprecate passing null, not everybody needs to add custom types. But I'm ok to remove the `=  null` as long a we fix the bug

alanpoulain:
> I don't get why we need to deprecate passing null, not everybody needs to add custom types.
Sure, but we use the type factory for our needs too: to add the `Iterable` type (needed here: https://github.com/api-platform/core/blob/2bb6f950b4fe3568e5c87d3bbe56fbf87f1901da/src/GraphQl/Type/SchemaBuilder.php#L363).

alanpoulain:
If you prefer, we can use the old behavior (https://github.com/api-platform/core/blob/22d350d95d592da7b3ee5117ac93e9f4e64b261f/src/GraphQl/Type/SchemaBuilder.php#L78) and only use the type factory for adding custom and non mandatory types.

soyuka:
Just remove the `= null` in the constructor then? This class is experimental anyway no?

dunglas:
Yes I'll update the pr

dunglas:
As pointed (privately) by @meyerbaptiste, it's not the fault of Symfony.
The catch should be moved here: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Routing/Router.php#L83

soyuka:
tests must be updated

PR #2513, opened 2019-02-12
Throw exception if no item found when denormalizing a relation using plain identifiers

gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes?
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

Scenario: This is enabled: `allow_plain_identifiers: true`, the request is sent in `json`, and a non existing plain identifier is sent in a relationship field. 

Instead of returning a `400` error code, a `200` code is returned.

That's because `denormalizeRelation` returns an empty object instead of throwing `InvalidArgumentException`.

If this PR makes sense I'll add a test for it too.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -393,7 +393,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             // repeat the code so that IRIs keep working with the json format
             if (true === $this->allowPlainIdentifiers && $this->itemDataProvider) {
                 try {
-                    return $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true]);
+                    if (null !== $item = $this->itemDataProvider->getItem($className, $value, null, $context + ['fetch_data' => true])) {
+                        return $item;
+                    }
                 } catch (ItemNotFoundException $e) {
                     throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
                 } catch (InvalidArgumentException $e) {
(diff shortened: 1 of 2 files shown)


soyuka:
Hi ! Can you please add a functional test (behat) that illustrates the bug? Also, because it's a bug fix please target the 2.4 branch!

gonzalovilaseca:
Sure, will update as soon as I have some time!

gonzalovilaseca:
@soyuka Is there any docker config I can use to run tests etc locally on this repo?

soyuka:
https://github.com/api-platform/core/blob/master/CONTRIBUTING.md#tests you don't need docker if you have phpunit, behat, php-cs-fixer, phpstan :D

gonzalovilaseca:
@soyuka Can't run composer install as I don't have mongo extension...building my own docker stack..
It's not related to #2530, that's another issue :-)

soyuka:
> @soyuka Can't run composer install as I don't have mongo extension...building my own docker stack..

Yeah we're trying to solve this issue... You could run the standard edition that has a docker compose file but I've never tried to run core tests from there.

soyuka:
Hi @gonzalovilaseca can you check https://github.com/api-platform/core/pull/2536 it looks like it fixes your issue. Thanks!

gonzalovilaseca:
Hi @soyuka It does, should we close this PR then?

soyuka:
I think that his patch is good indeed! Thanks for contributing anyway!

PR #2519, opened 2019-02-13
bugfix: throw InvalidArgument on matching something that looks like a…

soyuka:
…n url

I know that backporting is only for security fixes but this is a really old issue, maybe we can make an exception?

### DIFF
--- a/src/Bridge/Symfony/Routing/Router.php
+++ b/src/Bridge/Symfony/Routing/Router.php
@@ -14,7 +14,9 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Symfony\Routing;
 
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
+use Symfony\Component\HttpFoundation\Exception\RequestExceptionInterface;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\RequestContext;
 use Symfony\Component\Routing\RouterInterface;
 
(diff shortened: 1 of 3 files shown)


src/Bridge/Symfony/Routing/Router.php
```diff
@@ -72,7 +74,12 @@ public function match($pathInfo)
         $pathInfo = str_replace($baseContext->getBaseUrl(), '', $pathInfo);
 
         $request = Request::create($pathInfo, 'GET', [], [], [], ['HTTP_HOST' => $baseContext->getHost()]);
-        $context = (new RequestContext())->fromRequest($request);
+        try {
+            $context = (new RequestContext())->fromRequest($request);
+        } catch (RequestExceptionInterface $e) {
+            throw new RouteNotFoundException('Invalid request context.');
```
meyerbaptiste: Throwing a `RouteNotFoundException` is weird. According to the PHPDoc, it should be `NoConfigurationException` or `ResourceNotFoundException` or `MethodNotAllowedException`.

soyuka: discussed with @dunglas and we accepted this one because it's the only possibility so that it's catched correctly after. We may change the phpdoc to reflect this though.

FYI this has been merged to 2.4 but several people asked for a backport.

meyerbaptiste: `Symfony\Component\Routing\Exception\ResourceNotFoundException` implements the `Symfony\Component\Routing\Exception\ExceptionInterface`. So it's okay: the exception will be caught by the `IriConverter::getItemFromIri()` method!

PR #2523, opened 2019-02-15
fix #2342 add metadata aware name converter if available

### LINKED ISSUE #2342: The new MetadataAwareNameConverter is not used on Api-Platform
Formerly https://github.com/api-platform/core/issues/2332#issuecomment-440668720

The name converted set on framework bundle is not passed directly on the normalizer but on the metadata aware name converter: https://github.com/symfony/symfony/pull/28505/files#diff-0e793081ceb720201745c982a568903fL1366

But, in api-platform, you still copy the "old" method for your normalizers: https://github.com/api-platform/core/blob/d67848110778af711c5f54226fe1234ec5668230/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L77-L79

IMHO, the `name_converter` option should be deprecated (or is it used for something else?) and the `serializer.name_converter.metadata_aware` should be passed if exists.

The current workaround of this issue is quite simple, just add this line on the `api_platform` configuration:

```yaml
api_platform:
    name_converter: serializer.name_converter.metadata_aware
```

WDYT?

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2342
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/735

Make use of the new AdvancedNameConverterInterface

### DIFF
--- a/src/Bridge/Elasticsearch/Serializer/NameConverter/InnerFieldsNameConverter.php
+++ b/src/Bridge/Elasticsearch/Serializer/NameConverter/InnerFieldsNameConverter.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Elasticsearch\Serializer\NameConverter;
 
+use Symfony\Component\Serializer\NameConverter\AdvancedNameConverterInterface;
 use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
@@ -23,7 +24,7 @@ use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
  *
  * @author Baptiste Meyer <baptiste.meyer@gmail.com>
  */
-final class InnerFieldsNameConverter implements NameConverterInterface
+final class InnerFieldsNameConverter implements AdvancedNameConverterInterface
 {
     private $decorated;
 
@@ -35,25 +36,25 @@ final class InnerFieldsNameConverter implements NameConverterInterface
     /**
      * {@inheritdoc}
      */
-    public function normalize($propertyName)
+    public function normalize($propertyName, string $class = null, string $format = null, $context = [])
     {
-        return $this->convertInnerFields($propertyName, true);
+        return $this->convertInnerFields($propertyName, true, $class, $format, $context);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function denormalize($propertyName)
+    public function denormalize($propertyName, string $class = null, string $format = null, $context = [])
     {
-        return $this->convertInnerFields($propertyName, false);
+        return $this->convertInnerFields($propertyName, false, $class, $format, $context);
     }
 
-    private function convertInnerFields(string $propertyName, bool $normalization): string
+    private function convertInnerFields(string $propertyName, bool $normalization, string $class = null, string $format = null, $context = []): string
     {
         $convertedProperties = [];
 
         foreach (explode('.', $propertyName) as $decomposedProperty) {
-            $convertedProperties[] = $this->decorated->{$normalization ? 'normalize' : 'denormalize'}($decomposedProperty);
+            $convertedProperties[] = $this->decorated->{$normalization ? 'normalize' : 'denormalize'}($decomposedProperty, $class, $format, $context);
         }
 
         return implode('.', $convertedProperties);
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -79,10 +79,6 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         if (!isset($propertyInfoConfig['enabled'])) {
             $container->prependExtensionConfig('framework', ['property_info' => ['enabled' => true]]);
         }
-
-        if (isset($serializerConfig['name_converter'])) {
-            $container->prependExtensionConfig('api_platform', ['name_converter' => $serializerConfig['name_converter']]);
-        }
     }
 
     /**
--- a/src/Hal/Serializer/ItemNormalizer.php
+++ b/src/Hal/Serializer/ItemNormalizer.php
@@ -181,7 +181,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
 
             $relationName = $relation['name'];
             if ($this->nameConverter) {
-                $relationName = $this->nameConverter->normalize($relationName);
+                $relationName = $this->nameConverter->normalize($relationName, $class, $format, $context);
             }
 
             if ('one' === $relation['cardinality']) {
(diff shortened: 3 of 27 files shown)


bendavies:
hah wow. i just needed this, and you fixed it today!

bendavies:
hey @soyuka! i had a look, and this doesn't fix the issue fully.
in many places we are doing:
```php
$this->nameConverted->normalize($propertyName)
```
but we should be doing:
```php
$this->nameConverted->normalize($propertyName, $class, $format, $context)
```

WDYT?

soyuka:
Sure thing but this means I have to check that the nameConverter respects this interface!

bendavies:
I'm not sure you do.
The first params of the 
methods on `AdvancedNameConverterInterface` and `NameConverterInterface` are the same, so you can just call with all 4 params regardless. That's why core symfony does i think.

bendavies:
example:
https://github.com/symfony/symfony/blob/98693beca4845be9ce4ce5484d2c7cc20f376801/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L439

soyuka:
@bendavies should be good now!

bendavies:
LGTM.

Maybe the docs are out of date now?

https://api-platform.com/docs/core/serialization/#name-conversion
Overriding that means you can't make use of the metadata aware name converter.

Maybe the docs could be updated to this, so api platform's metadata aware name converter will use it as a fallback?

```yaml
framework:
    serializer:
        name_converter: 'Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter'
```

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -181,7 +181,7 @@ private function populateRelation(array $data, $object, string $format = null, a
 
             $relationName = $relation['name'];
             if ($this->nameConverter) {
-                $relationName = $this->nameConverter->normalize($relationName);
+                $relationName = $this->nameConverter->normalize($relationName, $class, $format);
```
bendavies: the 4th parameter `$context` is available here, (and probably in other places). why not pass it?

soyuka: I'm really not sure it makes sense, I thought about it but I don't know what informations the context brings here that can be useful.

soyuka:
We should deprecate configuring the name converter at api platform's level imo

src/Hal/Serializer/ItemNormalizer.php
bendavies: i guess if a user has implemented an `AdvancedNameConverterInterface` which uses the context (groups)?

bendavies:
yeah, something needs to happen!

soyuka:
Deprecated, updated the docs and added the context where available :).

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -195,6 +200,7 @@ private function handleConfig(ContainerBuilder $container, array $config, array
         $container->setAlias('api_platform.path_segment_name_generator', $config['path_segment_name_generator']);
 
         if ($config['name_converter']) {
+            @trigger_error('Configuring "name_converter" within the "api_platform" node is deprecated and will be removed in 3.0. Please use the "serializer.name_converter" node instead.', E_USER_DEPRECATED);
```
bendavies: https://symfony.com/doc/current/components/config/definition.html#deprecating-the-option

bendavies:
https://github.com/api-platform/docs/pull/734

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
soyuka: whoops yes indeed thanks

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -75,7 +75,11 @@ public function getConfigTreeBuilder()
                     ->setDeprecated('The use of the `default_operation_path_resolver` has been deprecated in 2.1 and will be removed in 3.0. Use `path_segment_name_generator` instead.')
                     ->info('Specify the default operation path resolver to use for generating resources operations path.')
                 ->end()
-                ->scalarNode('name_converter')->defaultNull()->info('Specify a name converter to use.')->end()
+                ->scalarNode('name_converter')
+                    ->defaultNull()
+                    ->setDeprecated('Configuring "name_converter" within the "api_platform" node is deprecated and will be removed in 3.0. Please use the "serializer.name_converter" node instead.')
+                    ->info('Specify a name converter to use.')
```
soyuka: On a second though I don't think we should deprecate this because it introduces a hard link with symfony.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -80,6 +81,10 @@ public function prepend(ContainerBuilder $container)
             $container->prependExtensionConfig('framework', ['property_info' => ['enabled' => true]]);
         }
 
+        if (class_exists(MetadataAwareNameConverter::class) && $container->hasDefinition('serializer.name_converter.metadata_aware')) {
```
meyerbaptiste: `$container->hasDefinition('serializer.name_converter.metadata_aware')` will always return `false` here.

src/JsonApi/Serializer/ReservedAttributeNameConverter.php
```diff
@@ -13,14 +13,15 @@
...
+use Symfony\Component\Serializer\NameConverter\AdvancedNameConverterInterface;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
 /**
  * Reserved attribute name converter.
  *
  * @author Baptiste Meyer <baptiste.meyer@gmail.com>
  */
-final class ReservedAttributeNameConverter implements NameConverterInterface
+final class ReservedAttributeNameConverter implements AdvancedNameConverterInterface
```
meyerbaptiste: What about  `InnerFieldsNameConverter`?

soyuka: Not sure to get what you want me to change :p.

src/Bridge/Elasticsearch/Serializer/NameConverter/InnerFieldsNameConverter.php
```diff
@@ -35,25 +36,25 @@ public function __construct(?NameConverterInterface $decorated = null)
     /**
      * {@inheritdoc}
      */
-    public function normalize($propertyName)
+    public function normalize($propertyName, string $class = null, string $format = null, $context = [])
     {
         return $this->convertInnerFields($propertyName, true);
```
meyerbaptiste: You have to pass the other arguments!

soyuka:
@bendavies could you check this? I want to merge it :)

bendavies:
LGTM 👍

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
teohhanhui: I don't know if that makes sense. We're already in a Symfony bundle in this case. :laughing:

When using our bundle, we have a hard dependency on `symfony/framework-bundle`, no? Otherwise everything falls apart.

Issue #2525, opened 2019-02-16
[GraphQL] Disable collection operations doesn't work


SCIF:
```
/**
 * @ApiResource(
 *     graphql={"query"},
 *     collectionOperations={}
 * )
 */
class User
```

In this case `UsersConnection` still created by `SchemaBuilder`. I assume it's a bug.

kiler129:
I think this will cause BC. Collections are slightly different than relations in GraphQL IMHO.

SCIF:
@kiler129 , what do you mean under relations?

Actual schema looks like:

```
user(id: ID): User
users(first: Int, after: String): UserConnection
```

PR is intended to drop `users` from it.

lukasluecke:
I'm not sure that mixing the "collectionOperations" (which afaik only applies to REST-API atm) and the GraphQL-API is a good idea. I believe that's also what @kiler129 was trying to convey.

It would probably make more sense to modify the `graphql` array of the API-resource config (e.g. split `query` into `query_item` and `query_collection`).

There are several other features that do not apply to the GraphQL-API right now (e.g. @ApiProperty (required, writeable, ..) - see https://github.com/api-platform/api-platform/issues/872

SCIF:
Thanks @lukasluecke for your input. However, I cannot see any point to duplicate settings under the `graphql` branch which is not autocompleted and validated properly. Moreover, at this date some `@ApiResource` attributes are used, for instance `inputClass`. I think it would be better to split `ApiResource` to `ApiRest` and `ApiGraphql` notations in future, to reflect specifics better but for now it sounds stupid for me to not honour not BC-breaking way of managing endpoints.

alanpoulain:
@SCIF Having `@ApiResource` for both GraphQL and REST has advantages and drawbacks.

The main advantage is to be able to share configuration between GraphQL and REST.

The main drawback is to have some keys only applied for REST or GraphQL. It's the case for `collectionOperations`.

I think the best way to solve this would be to have a `rest` key like the `graphql` one with only the specific REST configuration keys, but it would be a big BC break.

In the current state, I think this issue is not valid.

PR #2526, opened 2019-02-16
#2525: honour disabling "get" for `collectioOperations` for GraphQL

SCIF:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2525
| License       | MIT
| Doc PR        | no


SCIF:
To avoid mixing of discussion, I've answered in #2525.

alanpoulain:
As @lukasluecke said, `collectionOperations` is only for REST.
Maybe having `query_item` and `query_collection` would be a good idea.
There is also an issue with the `access_control` in the `query` key: the `object` variable should not exist in case of a collection.

PR #2530, opened 2019-02-18
Do not allow extra fields when denormalizing a resource in 

gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | yes?
| Deprecations? | no
| Tests pass?   | waiting for build...
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

When sending random payload on a `PATCH` request, the http result code is 200 because normalization doesn't fail as the extra fields setting is `true` by default.


flug:
@gonzalovilaseca Up is this pull request still valid?

PR #2532, opened 2019-02-19
Enable the pagination via cursor (no page-based pagination)

soyuka:
Add tests for the SoMany collection

Uses the PropertyAccessor service and don't break the BC

Allow to use multiple fields

Fix cs

To avoid breaking @sroze branch I'm creating a new PR with his work. Original PR is https://github.com/api-platform/core/pull/1915 . I only rebased, squashed and fixed CS.

### DIFF
--- a/features/hydra/collection.feature
+++ b/features/hydra/collection.feature
@@ -437,3 +437,109 @@ Feature: Collections support
     When I send a "GET" request to "/dummies?itemsPerPage=0&page=2"
     Then the response status code should be 400
     And the JSON node "hydra:description" should be equal to "Page should not be greater than 1 if limit is equal to 0"
+
+  Scenario: Cursor-based pagination with an empty collection
+    When I send a "GET" request to "/so_manies"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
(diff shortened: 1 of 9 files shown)


features/hydra/collection.feature
```diff
@@ -437,3 +437,99 @@ Feature: Collections support
...
+          }
+        }
+      }
+    }
+    """
+
+  @createSchema
+  Scenario: Cursor-based pagination with items
+    Given there are 10 of these so many objects
+    When I send a "GET" request to "/so_manies?order[id]=desc&id[gt]=10"
```
antograssiot: Also this relies "only" on the order and range filters for so I don't see why having 20 items would result in getting id 13, 12, 11 and not 20, 19, 18.
It would require `order[id]=desc&id[gt]=10&id[lt]=1' ?

soyuka:
Doc pr : https://github.com/api-platform/docs/pull/738

antograssiot:
@soyuka We might still be missing something here that we were probably not doing back when the initial PR was open. I think we should add a `$paginationViaCursor` property in the `ApiRessource` class and update the class docblock by adding ` @Attribute("paginationViaCursor", type="array"),`

norkunas:
Why this was not released in 2.4 ? :slightly_smiling_face:

soyuka:
Because it did not land in 2.4 but in master, it's an experimental feature.

norkunas:
Forgot about that lower branches are merged to master..

PR #2536, opened 2019-02-20
fix error case when denormalizing relations with plain identifiers.

bendavies:
fix error case when denormalizing relations with plain identifiers. null is returned, no exceptions are thrown.

Not sure which branch to merge into. let me know!

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |


soyuka:
thanks @bendavies !

PR #2547, opened 2019-02-21
Denormalize the input only if it is supported

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | todo update with use case

We still have an issue while denormalizing and using messenger for example, a same resource might be denormalized into an input although it's not one.

For example:

The resource is `RecoverPaswordResource` and has only an item operation where the output is disabled and the input is `RecoverPasswordInput`.
 
1) Request `PUT {RecoverPasswordInput} /recover_password`
2) The `AbstractItemNormalizer` denormalizes the json representation of `{RecoverPasswordInput}` in a `RecoverPasswordInput`
3) The `DataTransformer` transforms this `RecoverPasswordInput` in a `RecoverPasswordResource`
4) Messenger is used, we send the `RecoverPasswordResource`
5) The handler receives a `{RecoverPasswordResource}` json representation and tries to denormalize it
6) Because it has an `input`, the `AbstractItemNormalizer` tries to denormalize it as a `RecoverPasswordInput` which is wrong, it's a `{RecoverPasswordResource}`.

This fixes step 6), indeed if the resource to transform isn't a representation of the input, we need to return a falsy value in the `supportsTransformation`.


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -165,9 +165,8 @@ public function denormalize($data, $class, $format = null, array $context = [])
         $inputClass = $this->getInputClass($class, $context);
 
         if (null !== $inputClass) {
-            $data = parent::denormalize($data, $inputClass, $format, ['resource_class' => $inputClass] + $context);
-
             if (null !== $this->dataTransformer && $this->dataTransformer->supportsTransformation($data, $class, $context)) {
+                $data = parent::denormalize($data, $inputClass, $format, ['resource_class' => $inputClass] + $context);
```
GregoireHebert: I think you should merge the two `if` conditions into one.

soyuka:
Don't hesitate to comment, I merged because many people (thanks <3) are using the github branch and this fixes a critical issue on input/output :). Nothing to change on your side though.
It simplifies the code in the normalizers (less repetition) and adds some tests.

PR #2558, opened 2019-02-26
fix graphql operation xsd

GregoireHebert:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | does it?
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/tests/Metadata/schema/XmlSchemaTest.php
+++ b/tests/Metadata/schema/XmlSchemaTest.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


alanpoulain:
Shouldn't we add some tests to validate the XSD against an XML file?

tests/Metadata/schema/XmlSchemaTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+        try {
+            XmlUtils::loadFile($fixtures.'invalid.xml', $schema);
+            $this->fail();
+        } catch (\InvalidArgumentException $e) {
+            $this->assertContains('ERROR 1845', $e->getMessage());
+        }
+
+        // should not throw an exception
+        XmlUtils::loadFile($fixtures.'valid.xml', $schema);
```
alanpoulain: Maybe use `assertInstanceOf`?

alanpoulain:
Thank you @GregoireHebert!

Issue #2563, opened 2019-02-27
There is no PropertyInfo extractor supporting the class


raresserban:
Hello,
After upgrade from Api Platform 2.4 beta.1 to beta.2, I get the following error

```
There is no PropertyInfo extractor supporting the class \"Automotive\\SSMConnectorBundle\\DTO\\CredentialsDTO\"
```

Here is the class in question:

```php
/**
 * @method string getSessionId()
 * @method self setSessionId(?string $sessionId)
 * @method string getToken()
 * @method self setToken(?string $token)
 */
class CredentialsDTO implements \JsonSerializable
{
/** @var array */
    protected $objectData = [];

    /**
     * {@inheritdoc}
     */
    public function __call($name, $arguments)
    {
        if ('get' === substr($name, 0, 3)) {
            return $this->objectData[$this->getProperty($name)] ?? null;
        } elseif ('set' === substr($name, 0, 3) && array_key_exists(0, $arguments)) {
            $this->objectData[$this->getProperty($name)] = $arguments[0];

            return $this;
        }
    }

    /**
     * {@inheritdoc}
     */
    public function __isset($name)
    {
        return array_key_exists($name, $this->objectData);
    }

    /**
     * From \JsonSerializable interface.
     *
     * Specify data which should be serialized to JSON
     * @link https://php.net/manual/en/jsonserializable.jsonserialize.php
     * @return mixed data which can be serialized by <b>json_encode</b>,
     * which is a value of any type other than a resource.
     */
    public function jsonSerialize()
    {
        return $this->objectData;
    }

    /**
     * Return the property from a function name.
     *
     * @param $name
     *
     * @return string
     */
    private function getProperty($name)
    {
        return lcfirst(substr($name, 3));
    }
}
```

I use magic_call to true of the property_access component in order to have dynamic properties. This worked perfectly fine in beta.1 but no longer works in beta.2...
I am not sure what the problem is...

soyuka:
We now try to denormalize DTOs just as they were resources. This means we are indeed using the PropertyInfo component to gather properties and types for a given DTO. 

Your use case is really hackish and indeed the property info component doesn't find any properties and fails. The proper way of fixing this would be to register a property info extractor that handles your special case (but you don't know the properties ahead?).

Can you give me the error stack trace please? Thanks!

raresserban:
This DTO is not even related to api platform, it has no ApiResource annotation so I am not sure why it is affected.
[...]

PR #2566, opened 2019-02-27
Use the ItemNormalizer in the correct context

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a will prevent some
| License       | MIT
| Doc PR        | na

This PR prevents the use of ApiPlatform's ItemNormalizer when not in the correct context. 
This means that:
- if we're using Symfony without any supported formats and outside ApiPlatform context, we're using symfony normalizers
- if it's a hypermedia format (`jsonld`, `hal`, `jsonapi`) we're using ApiPlatform's normalizers even if we're inside symfony
- if it's another format but we're using ApiPlatform we're still using our ItemNormalizers

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -15,12 +15,16 @@ namespace ApiPlatform\Core\Serializer;
 
 use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
-use ApiPlatform\Core\Bridge\Elasticsearch\Serializer\ItemNormalizer;
+use ApiPlatform\Core\Bridge\Elasticsearch\Serializer\ItemNormalizer as ElasticsearchItemNormalizer;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\InvalidValueException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
+use ApiPlatform\Core\GraphQl\Serializer\ItemNormalizer as GraphQlItemNormalizer;
+use ApiPlatform\Core\Hal\Serializer\ItemNormalizer as HalItemNormalizer;
+use ApiPlatform\Core\JsonApi\Serializer\ItemNormalizer as JsonApiItemNormalizer;
+use ApiPlatform\Core\JsonLd\Serializer\ItemNormalizer as JsonLdItemNormalizer;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
@@ -36,6 +40,7 @@ use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface;
 use Symfony\Component\Serializer\NameConverter\AdvancedNameConverterInterface;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
+use Symfony\Component\Serializer\Normalizer\ContextAwareNormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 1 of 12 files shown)


src/Serializer/ItemNormalizer.php
```diff
@@ -98,4 +104,12 @@ private function updateObjectToPopulate(array $data, array &$context)
             $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(sprintf('%s/%s', $this->iriConverter->getIriFromResourceClass($context['resource_class']), $data[$identifier]), $context + ['fetch_data' => true]);
         }
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsNormalization($data, $format = null)
+    {
+        return (($context['use_api_platform'] ?? false) || \in_array($format, self::FORMATS, true)) && parent::supportsNormalization($data, $format);
```
dunglas: Can you introduce a const for `use_api_platform`?

alanpoulain:
Could you add a description to explain why you needed to do this, for future reference?

src/Serializer/ItemNormalizer.php
```diff
@@ -98,4 +104,12 @@ private function updateObjectToPopulate(array $data, array &$context)
             $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri(sprintf('%s/%s', $this->iriConverter->getIriFromResourceClass($context['resource_class']), $data[$identifier]), $context + ['fetch_data' => true]);
         }
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsNormalization($data, $format = null)
```
dunglas: `$context` isn't defined, you must implements the `ContextAwareInterface`

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -144,7 +144,7 @@ public function normalize($object, $format = null, array $context = [])
      */
     public function supportsDenormalization($data, $type, $format = null)
     {
-        if (ItemNormalizer::FORMAT === $format) {
+        if (ElasticsearchItemNormalizer::FORMAT === $format) {
```
dunglas: Why this change?

soyuka: Because it's easier to read, it can be confusing if you read `ItemNormalizer`

soyuka:
Currently failing because the `$context` never reaches the `supportsNormalization`, maybe because of https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Serializer.php#L240

dunglas:
@alanpoulain to prevent errors like this one: https://github.com/api-platform/core/issues/2563

A better strategy would be to use another instance of the `Serializer` class for API Platform than the Symfony built-in one. But it's not doable as is, we need to introduce some kind of factories to wire a fully featured Serializer in Symfony first. This context option is a quick fix.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -93,8 +101,12 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
     /**
      * {@inheritdoc}
      */
-    public function supportsNormalization($data, $format = null)
+    public function supportsNormalization($data, $format = null, array $context = [])
     {
+        if (!(($context[static::USE_API_PLATFORM] ?? false) || \in_array($format, self::FORMATS, true))) {
```
antograssiot: just asking, but given that this method now relies on the context to answer, can we still systematically return `true` in the `hasCacheableSupportsMethod` ?

teohhanhui: We can't. And that's one of the reasons why this fix cannot work...

soyuka: how should we fix it ? :P

teohhanhui: Revert 😝

teohhanhui: Sometimes we gotta acknowledge we've made a wrong turn...

dunglas: Just return `false` in `hasCacheableSupportsMethod`.
@teohhanhui revert what? All our new input/output system? It doesn't look like a good idea. It's a very good move, and a long awaited one.

The proper fix for this specific problem is to use our own instance of the serializer instead of the global one provided by Symfony. To do so, we need to patch SF first, to provide convenient factories to wire and configure a serializer (basically, `FrameworkExtension` must be revamped). It will also benefit to Doctrine JSON ODM, and many other libraries.
But it's a hard work, and we need a quick fix first, such as this one. We'll improve later.

teohhanhui:
Uhh... Do you want me to stop working on the alternative solution?

PR #2578, opened 2019-03-04
Publish mercure updates on flush

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

When persisting from a messenger Handler, mercure updates where not published because `kernel.terminate` is not used in that context. This will use doctrine `postFlush` instead to publish mercure updates.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_mercure_publisher.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_mercure_publisher.xml
@@ -17,7 +17,7 @@
             <argument type="service" id="mercure.hub.default.publisher" />
 
             <tag name="doctrine.event_listener" event="onFlush" />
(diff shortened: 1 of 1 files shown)


bendavies:
in what context is `kernel.terminate` not fired? this will cause delays to the client and was infact changed from doctrine postflush to kernel terminate in the original PR https://github.com/api-platform/core/pull/2282

dunglas:
@bendavies when using a Messenger consumer, this event is triggered only when the command running the consumers terminates (as it's a demon, it's not "real-time" at all anymore :)).

bendavies:
Ah right. I'm amazed the consumer doesn't provide an event to hook into here.

dunglas:
Such events could be added soon: https://github.com/symfony/symfony/pull/30557

Issue #2355, opened 2018-11-30
[RFC] Using decoration over inheritance for Normalizers


joelwurtz:
Actually all Item Normalizers (HAL / JSON-LD / Json Api) extends the AbstractItemNormalizer to add their context and informations.

I'm actually working on a library that change the way of serializing / deserializing object by using generated code (https://github.com/janephp/automapper) insted of the PropertyAccess Component. This change allows significant performance boost 

 see https://blackfire.io/profiles/compare/c6fa2145-c3aa-46f9-a211-538c6dc576a5/graph or https://travis-ci.org/php-serializers/ivory-serializer-benchmark/builds/460744766?utm_source=github_status&utm_medium=notification

The example try to serilalize 10000 objects, that's an use case that we have where we need to return a lots of data and where we cannot use cache

However for correct integration with API Platform i need to rewrite all the existing normalizers since class are final and use inheritance.

I belive by using decoration (injecting a normalizer that does the item normalization / denormalization instead of calling the parent class) this would be far easier to handle this use case. I'm willing to work on the PR if that's something you want to support.

dunglas:
It's something we definitely want, and that is currently being improved directly in Symfony. @fbourigault has already provided some patches in this way (some are included in SF 4.2), but they are still some work to do. As usual, all help on this will be very welcome.

joelwurtz:
For decoration ? Or code generation ?

dunglas:
Decoration.

dunglas:
Composition actually.

joelwurtz:
I began to do some works here, but i'm a bit loss with all the things that Normalizers do on API Platform, some of the ObjectNormalizer features of Symfony are overrided here, and it's hard to know why.

Is there some kinds of docs that explain the whole Normalizer architecture of API Platform ? Or Features that it needs in order to work correctly ?

soyuka:
It's really internal so no docs here, only code but I'm available to help you. Note for the sake of simplicity I'm skipping Elasticsearch and Graphql.

There are two kind of normalizers in ApiPlatform:
- `CollectionNormalizer`
- `ItemNormalizer`

You'll be more interested by the later as it is an extension of the `ObjectNormalizer`.
Each normalizer is declined in every supported formats:
- For collections: `JsonApi\Serializer\CollectionNormalizer`, `Hydra\Serializer\CollectionNormalizer`, `Hal\Serializer\CollectionNormalizer`
[...]

PR #2579, opened 2019-03-04
Use composition for Normalizers

joelwurtz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #2355 
| License       | MIT
| Doc PR        | /

This is a beginning for #2355 to use composition over inheritance in the various normalizer. Goal is to be decoupled from the ObjectNormalizer so someone can still use JSON LD / Hydra / Json API with a custom normalizer or a different implementation.

There is a lot of questions that i wonder, that's why it's far from being complete but it shows where i want to go.

Some questions i have: 

 * Should we have a new class instead ? (easier for depreciation)
 * Does every normalizer should use composition (like i began to do) or should we add a generic decoupling instead (not sure if it will be easier)
 * Some behavior may be not available in the sub normalizer (like object to populate) is there a way to enforce this or should we just tell the user to take care of this with documentation ?

### DIFF
--- a/src/Serializer/AddResourceClassNormalizer.php
+++ b/src/Serializer/AddResourceClassNormalizer.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace ApiPlatform\Core\Serializer;
+
+use ApiPlatform\Core\Api\IriConverterInterface;
+use ApiPlatform\Core\Bridge\Elasticsearch\Serializer\ItemNormalizer as ElasticsearchItemNormalizer;
+use ApiPlatform\Core\Api\ResourceClassResolverInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Util\ClassInfoTrait;
+use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+/**
+ * Add resource class to the context
+ */
+final class AddResourceClassNormalizer implements NormalizerInterface, DenormalizerInterface, CacheableSupportsMethodInterface
+{
+    use ContextTrait;
+    use ClassInfoTrait;
+
+    /** @var DenormalizerInterface|NormalizerInterface */
+    private $normalizer;
+
+    private $resourceClassResolver;
+
+    private $iriConverter;
+
+    public function __construct(NormalizerInterface $normalizer, ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter)
+    {
+        if (!$normalizer instanceof DenormalizerInterface) {
+            throw new InvalidArgumentException('Normalizer should be an instance of '.DenormalizerInterface::class);
+        }
+
+        $this->normalizer = $normalizer;
+        $this->resourceClassResolver = $resourceClassResolver;
+        $this->iriConverter = $iriConverter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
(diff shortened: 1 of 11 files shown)


joelwurtz:
Also began to split the AbstractItemNormalizer, as you can see i extract the context initlization into its own normalizer and also the data transformation into its own normalizer.

dunglas:
Let's see where it brings us. For now I'm not sure to fully understand the changes, but the final goal is 100% legit, and we'll improve the design (next step, or parallel step, would be to the same thing in the Symfony Serializer)

src/Serializer/AddResourceClassNormalizer.php
```diff
@@ -0,0 +1,132 @@
...
+
+    private $resourceClassResolver;
+
+    private $localCache = [];
+
+    private $allowUnmappedClass;
+
+    private $iriConverter;
+
+    public function __construct(NormalizerInterface $normalizer, ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, bool $allowUnmappedClass = false)
```
dunglas: You can remove `allowUnmappedClass` in a new class. It has been added only for BC in the current class. Here we must always support unmapped classes.

joelwurtz:
> For now I'm not sure to fully understand the changes

Me neither xD, there is a lot of code that is duplicated from ObjectNormalizer and actually trying to see the little differences betweens (like iri convertion / relation normalizatiing and others stuffs)

I think a may have a way to completly remove the AbstractItemNormalizer, but this may require new methods in some interfaces.

I will try to do a first shot by creating a new way of normalization (withtout the AbstractItemNormalizer) without touching the existing implementation so we can spot / benchmark / tests behaviors in parallel.

dunglas:
Maybe would it be easier to start refactoring the Symfony component first, then use its new extensions point that need to be created anyway. WDYT?

joelwurtz:
I prefer to refactor it here first, since api platform normalizers are heavily based on symfony serializer system and features, having a look first at the blockers and usage when extending it to the maximum will give us a better look on how to refactor better symfony.

Also i think we can something nice without touching symfony (but maybe i'm wrong).

joelwurtz:
So made a lot of revert, and begin to work in a new class, i will handle deprecation at the end if final design is accepted (mainly deprecate the old class to use the new one)

ATM there is a good exemple for the JsonLdItemNormalizer 

So overall design will look like this: Every feature is splitted into its own normalizer:

 * ResourceClassNormalizer handle the extraction of the resource class and check if the given object can be normalized / denormalized as a resource class
 * DataTransformerNormalizer handle data transformations using the DataTtransformerInterface
 * No more custom AbstractItemNormalizer, it directly use the ObjectNormalizer of symfony
* Each "format" normalizers (json ld / json api / ...) add the necessary metadata to the data (like before)

Each of this normalizer use a sub normalizer to handle data, so in fact the JsonLdItemNormalizer as a dependency like this:

ResourceClassNormalizer -> JsonLdItemNormalizer -> DataTransformerNormalizer -> ObjectNormalizer (Symfony)

Also i believe this will allow to remove all code that need to check about handling non resource class, as someone will be able to inject its own chain of normalizer and give necessary info into the context if needed.

teohhanhui:
@joelwurtz Just a heads up, I've changed the normalization of non-resource in https://github.com/api-platform/core/pull/2679

joelwurtz:
Thanks for the heads up, look better that way

teohhanhui:
Another heads up, denormalization of non-resources also uses composition now in #2797.

soyuka:
same reason I closed https://github.com/api-platform/core/issues/2355 we're now watching how the https://github.com/symfony/symfony/pull/51718 work progresses! thanks for the work!

PR #2569, opened 2019-03-01
Allow per-operation messenger configuration

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

Allow messenger attribute by operation.

PR #2583, opened 2019-03-06
Use ContextAware on data persister chains

soyuka:
Re-do https://github.com/api-platform/core/pull/2569 as I merged it too quickly

Updates the ChainDataPersisters with the context so that it gets propagated to the messenger data persister.


soyuka:
Thanks for the review guys.

PR #2586, opened 2019-03-06
Attribute to allow disabling elasticsearch

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | todo

If you want to use Elasticsearch on top of doctrine it's currently not possible (unless you override the Elasticsearch DataProvider). 
This adds an `elasticsearch` attribute that you can use to disable elasticsearch (it's enabled y default as soon as you have a mapping). 
For example:

```
 * @ApiResource(
 *   elasticsearch=false,
 *   collectionOperations={"get",
 *     "create"={"input"=BookmarkInput::class, "output"=false, "status"=202, "method"="POST", "messenger"=true},
 *     "search"={"elasticsearch"=true, "method"="GET", "path"="/bookmarks/search.{_format}"}
 *   }
 * )
```

The Elasticsearch data provider will only be used in the `/bookmarks/search` method.

### DIFF
--- a/src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
+++ b/src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
@@ -21,6 +21,8 @@ use ApiPlatform\Core\Bridge\Elasticsearch\Metadata\Document\Factory\DocumentMeta
 use ApiPlatform\Core\DataProvider\ContextAwareCollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\Pagination;
 use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
+use ApiPlatform\Core\Exception\ResourceClassNotFoundException;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Elasticsearch\Client;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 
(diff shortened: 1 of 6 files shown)


src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
```diff
@@ -70,7 +74,15 @@ public function supports(string $resourceClass, ?string $operationName = null, a
             return false;
         }
 
-        return true;
+        $isElasticsearchEnabled = true;
+
+        try {
```
meyerbaptiste: Maybe we should test that first?

src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
```diff
@@ -64,7 +67,15 @@ public function supports(string $resourceClass, ?string $operationName = null, a
             return false;
         }
 
-        return true;
+        $isElasticsearchEnabled = true;
```
meyerbaptiste: Same comment as above.

src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
```diff
@@ -38,20 +40,30 @@ final class ItemDataProvider implements ItemDataProviderInterface, RestrictedDat
...
      * {@inheritdoc}
      */
     public function supports(string $resourceClass, ?string $operationName = null, array $context = []): bool
     {
+        try {
+            $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
+            if (false === $resourceMetadata->getItemOperationAttribute($operationName, 'elasticsearch', true, true)) {
+                return false;
+            }
+        } catch (ResourceClassNotFoundException $e) {
```
meyerbaptiste: Then return `false`? 😛

PR #2590, opened 2019-03-08
Fix messenger persister

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | soon

When the context is available in the Messenger Data Persister, let's use it. This also gives us the ability to send an Input to the messenger (skip the transformation see https://github.com/api-platform/core/pull/2495 discussion).


dunglas:
Thanks @soyuka

Issue #2591, opened 2019-03-08
2.4beta is incompatible with FrameworkExtraBundle @template annotations


mmarton:
Hi!

with the 2.4beta I got "No encoder found for format "html". for controllers and actions that has nothing to do with api platform but are trying to use the Sension/FrameworkExtraBundle @template annotations.

https://symfony.com/doc/master/bundles/SensioFrameworkExtraBundle/annotations/view.html

<details><summary>Trace</summary>
<p>

```
Symfony\Component\Serializer\Exception\RuntimeException:
No encoder found for format "html".

  at vendor/symfony/serializer/Encoder/ChainEncoder.php:101
  at Symfony\Component\Serializer\Encoder\ChainEncoder->getEncoder('html', array())
     (vendor/symfony/serializer/Encoder/ChainEncoder.php:40)
  at Symfony\Component\Serializer\Encoder\ChainEncoder->encode(array(), 'html', array())
     (vendor/symfony/serializer/Serializer.php:301)
  at Symfony\Component\Serializer\Serializer->encode(array(), 'html')
     (vendor/api-platform/core/src/EventListener/SerializeListener.php:115)
  at ApiPlatform\Core\EventListener\SerializeListener->serializeRawData(object(GetResponseForControllerResultEvent), object(Request), array())
     (vendor/api-platform/core/src/EventListener/SerializeListener.php:57)
  at ApiPlatform\Core\EventListener\SerializeListener->onKernelView(object(GetResponseForControllerResultEvent), 'kernel.view', object(TraceableEventDispatcher))
     (vendor/symfony/event-dispatcher/Debug/WrappedListener.php:115)
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke(object(GetResponseForControllerResultEvent), 'kernel.view', object(TraceableEventDispatcher))
     (vendor/symfony/event-dispatcher/EventDispatcher.php:212)
  at Symfony\Component\EventDispatcher\EventDispatcher->doDispatch(array(object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener)), 'kernel.view', object(GetResponseForControllerResultEvent))
     (vendor/symfony/event-dispatcher/EventDispatcher.php:44)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch('kernel.view', object(GetResponseForControllerResultEvent))
[...]

PR #2592, opened 2019-03-08
Use separate instance of ItemNormalizer for handling non-resource

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2563, #2591, ...; alternative to #2566 (reverted here)
| License       | MIT
| Doc PR        | N/A

This should ensure the same caching of `supports*` methods as before for resource classes. For non-resource classes (handled by a separate instance), the `supports*` methods use the `$context`, thus not cacheable.

Not sure about the performance impact (positive or negative), but I think this is cleaner as we don't indiscriminately say yes to things we might not be able to handle (provided we're at a low enough priority). And there's no need for `use_api_platform` and/or special treatment of some formats.

Big thanks to @soyuka and @dunglas for the discussions and for bearing with me! :stuck_out_tongue_closed_eyes:


alanpoulain:
It seems it is a better way to do it indeed.

PR #2603, opened 2019-03-12
update for phpstan^0.11

soyuka:
I'm not sure if sometimes we should do an `is_array` check (for example in `CollectionNormalizer`) instead of doc-typing?


teohhanhui:
Hmm... The inline PHPDoc's are not in the correct format:

```
/** @var SomeType $someVar */
```

PR #2606, opened 2019-03-12
Install phpstan locally and use extensions

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Replaces #2603


composer.json
```diff
@@ -39,14 +39,19 @@
         "doctrine/mongodb-odm-bundle": "^4.0.0@beta",
         "doctrine/orm": "^2.6.3",
         "elasticsearch/elasticsearch": "^6.0",
-        "friendsofsymfony/user-bundle": "dev-fix-mongodb-mapping",
+        "friendsofsymfony/user-bundle": "dev-master",
```
alanpoulain: :sweat_smile:

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -41,6 +42,9 @@ public function normalize($object, $format = null, array $context = [])
     {
         $object = $this->transformOutput($object, $context);
         $data = parent::normalize($object, $format, $context);
+        if (!\is_array($data)) {
+            throw new UnexpectedValueException('Expected data to be an array');
```
alanpoulain: Add `@throws`?

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -224,12 +225,14 @@ private function populateRelation(array $data, $object, string $format = null, a
...
     /**
      * Gets the IRI of the given relation.
-     *
-     * @param array|string $rel
      */
     private function getRelationIri($rel): string
     {
-        return $rel['_links']['self']['href'] ?? $rel;
+        if (!(\is_array($rel) || \is_string($rel))) {
+            throw new UnexpectedValueException('Expected relation to be an IRI or array');
```
alanpoulain: `@throws`?

src/JsonApi/Serializer/CollectionNormalizer.php
```diff
@@ -82,9 +82,12 @@ protected function getItemsData($object, string $format = null, array $context =
 
         foreach ($object as $obj) {
             $item = $this->normalizer->normalize($obj, $format, $context);
+            if (!\is_array($item)) {
+                throw new UnexpectedValueException('Expected item to be an array');
+            }
 
             if (!isset($item['data'])) {
-                throw new InvalidArgumentException('The JSON API document must contain a "data" key.');
+                throw new UnexpectedValueException('The JSON API document must contain a "data" key.');
```
alanpoulain: Could it be a BC break?

composer.json
dunglas:
```suggestion
        "friendsofsymfony/user-bundle": "^2.2@dev",
```

src/GraphQl/Serializer/ItemNormalizer.php
teohhanhui: It's already in `NormalizerInterface`.

src/JsonApi/Serializer/CollectionNormalizer.php
teohhanhui: I'd consider it a bug fix, because we're not supposed to throw this type of exception from `NormalizerInterface::normalize` in the first place.

bendavies:
title says `Switch to phpstan-shim` but you are not (and i would advise against it)

teohhanhui:
@bendavies What do you mean? It does make the switch to phpstan-shim, which allows us to use extensions. That's very important as otherwise many problems go undetected.

bendavies:
@teohhanhui where do you use the shim? i see `"phpstan/phpstan": "^0.11.3",` in composer.json, not `phpstan/phpstan-shim`

teohhanhui:
@bendavies Oops. You're right. I didn't even realize that, but the intention was to use the shim. 😂😂😂

Yeah, I guess it's not actually necessary to use the shim.

bendavies:
FYI here is why i'd advise against the shim:
https://github.com/composer/composer/issues/7573

teohhanhui:
@bendavies That looks like a general Composer issue, but yeah, we have no need to use the shim, and could easily switch if / when the need arises.

bendavies:
yes, it's a composer issue, but one that manifests if you use the shim, which is very annoying :)

soyuka:
Thanks @teohhanhui !

PR #2608, opened 2019-03-12
Add an HTTP client dedicated to functional API testing

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

Currently, there are no very satisfactory solutions to write functional tests for web APIs built in PHP (API Platform and Symfony included):

* Mink/Behat/Behatch are BDD-oriented (which is completely fine, but doesn't fit with all projects/teams), are a bit complex to setup, have a high barrier to entry and are still not fully compatible with Symfony 4 (you need to rely on a dev version of Behat). They also lack of utilities, for database testing for instance.
* WebTestCase/BrowserKit are dedicated to webpage testing through web scraping (DOM Crawler, CSS selectors, simulation of browsers actions such as clicking or reloading a page...). Their API don't fit well with API testing. However, they benefit form the large ecosystem of PHPUnit, and give access to the numerous functional testing helpers provided by Symfony.
* External solutions that don't manipulate the Symfony Kernel (but are true HTTP clients) such as Postman or BlackFire Player require to setup a web server for the testing env and don't give access to the service container (for instance, to test if the database has been updated, or if a mail has been sent).

It's time to say hi to `ApiTestCase` and `Test\Client`! This new set of API testing utilities is built on top of the Symfony's `Client` test class, and implements the exact same interface than [the brand new Symfony HttpClient component](https://speakerdeck.com/fabpot/2-new-symfony-components-httpclient-and-mime).

As you'll see, it also plays very well with [Alice](https://github.com/nelmio/alice) (the fixture generator with an official Symfony recipe), that gets new power in the operation!

Let's see how it looks:

```php
namespace App\Tests;

use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase;
use Dunglas\UserBundle\Entity\ForgotPassword\Jti;
use Hautelook\AliceBundle\PhpUnit\RefreshDatabaseTrait;
use Lcobucci\JWT\Parser;

class UsersTest extends ApiTestCase
{
    use RefreshDatabaseTrait; // This new trait I added to HautelookAliceBundle will take care of refreshing the database content to put it in a known state between every tests

     // A simple test
    public function testCreateUser()
    {
        $client = static::createClient(); // $client implements the new Symfony's HttpClientInterface
        $response = $client->request('POST', '/api/users', ['json' => [
            'email' => 'dunglas+test@gmail.com',
            'plainPassword' => 'foo',
        ]]);
        $this->assertSame(201, $response->getStatusCode());

        $responseContent = $response->toArray();
        $this->assertStringStartsWith('/api/users/', $responseContent['@id']);

        $this->assertArraySubset([
            '@type' => 'http://schema.org/Person',
            'email' => 'dunglas+test@gmail.com',
        ], $responseContent);
        $this->assertArrayNotHasKey('password', $responseContent);
        $this->assertArrayNotHasKey('plainPassword', $responseContent);
    }

    // something more advanced, accessing the mails and the DB
    public function testChangePassword()
    {
        $client = static::createClient();
        $client->enableProfiler();
        $client->disableReboot();

        // Send the request
        $response = $client->request('POST', '/api/users/token_request', ['json' => ['username' => 'dunglas@gmail.com']]);
        $this->assertSame(202, $response->getStatusCode());

        $mailCollector = $client->getProfile()->getCollector('swiftmailer');

        // extract the JWT contained in the mail
        preg_match('#[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*#', $mailCollector->getMessages()[0]->getBody(), $matches);
        $jwt = $matches[0];
        $decodedToken = (new Parser())->parse($jwt);

        // Change the password using the provided token
        $response = $client->request('PUT', $decodedToken->getClaim('sub'), [
            'json' => ['plainPassword' => 'newPassword'],
            'auth_bearer' => $jwt,
        ]);
        $this->assertSame(200, $response->getStatusCode());

        // Checks that the user have been changed
        $user = self::$container->get('test.security.helper')->getUser();
        $encoder = self::$container->get('test.security.encoder_factory')->getEncoder($user);
        $this->assertTrue($encoder->isPasswordValid($user->getPassword(), 'newPassword', $user->getSalt()));

        // Checks that the JTI has been saved
        $this->assertNotNull(
            self::$container->get('test.doctrine')->getRepository(Jti::class)->find($decodedToken->getClaim('jti'))
        );
    }
}
```

The following fixtures are loaded automatically. The database is reseted to a known test between every test (regardless of it fails or not) using transactions for maximum performance.

```yaml
App\Entity\User:
  user_kevin:
    email: dunglas@gmail.com
    password: \$argon2i\$v=19\$m=1024,t=2,p=2\$TW9YajBjdGpUVUp4Uks1aA\$vwFQakIF9aGGoC5KjT+S3kuOYDtwtghzMuEZ/xP6FnM # maman
``` 

TODO:

* [x] Import tests
* [ ] Write docs
* [x] Add support for Basic auth
* [x] Wait for Symfony 4.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,6 +72,7 @@
         "symfony/finder": "^3.4 || ^4.0",
         "symfony/form": "^3.4 || ^4.0",
         "symfony/framework-bundle": "^4.3",
+        "symfony/http-client": "^4.3",
         "symfony/mercure-bundle": "*",
         "symfony/messenger": "^4.3",
         "symfony/phpunit-bridge": "^4.3.1",
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -35,6 +35,7 @@ use ApiPlatform\Core\GraphQl\Type\Definition\TypeInterface as GraphQlTypeInterfa
 use Doctrine\Common\Annotations\Annotation;
 use phpDocumentor\Reflection\DocBlockFactoryInterface;
 use Ramsey\Uuid\Uuid;
+use Symfony\Component\BrowserKit\AbstractBrowser;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\DirectoryResource;
@@ -120,6 +121,10 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             ->addTag('api_platform.subresource_data_provider');
         $container->registerForAutoconfiguration(FilterInterface::class)
             ->addTag('api_platform.filter');
+
+        if ($container->hasParameter('test.client.parameters') && class_exists(AbstractBrowser::class)) {
+            $loader->load('test.xml');
+        }
     }
 
     private function registerCommonConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader, array $formats, array $errorFormats): void
--- a/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
+++ b/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\Test;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
+
+/**
+ * Base class for functional API tests.
+ *
+ * @experimental
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+abstract class ApiTestCase extends KernelTestCase
+{
+    /**
+     * Creates a Client.
+     *
+     * @param array $options An array of options to pass to the createKernel method
+     */
+    protected static function createClient(array $options = []): Client
+    {
+        $kernel = static::bootKernel($options);
+
+        try {
+            /**
+             * @var Client
+             */
+            $client = $kernel->getContainer()->get('test.api_platform.client');
+        } catch (ServiceNotFoundException $e) {
+            throw new \LogicException('You cannot create the client used in functional tests if the BrowserKit component is not available. Try running "composer require symfony/browser-kit".');
+        }
+
+        return $client;
+    }
+}
--- a/src/Bridge/Symfony/Bundle/Test/Response.php
+++ b/src/Bridge/Symfony/Bundle/Test/Response.php
@@ -128,10 +128,6 @@ final class Response implements ResponseInterface
     /**
      * {@inheritdoc}
      */
-
-    /**
-     * {@inheritdoc}
-     */
     public function toArray(bool $throw = true): array
     {
         if ('' === $content = $this->getContent($throw)) {
--- a/src/Bridge/Symfony/Bundle/Test/Response.php
+++ b/src/Bridge/Symfony/Bundle/Test/Response.php
@@ -1,0 +1,191 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\Test;
+
+use Symfony\Component\BrowserKit\Response as BrowserKitResponse;
+use Symfony\Component\HttpClient\Exception\ClientException;
+use Symfony\Component\HttpClient\Exception\JsonException;
+use Symfony\Component\HttpClient\Exception\RedirectionException;
+use Symfony\Component\HttpClient\Exception\ServerException;
+use Symfony\Component\HttpClient\Exception\TransportException;
+use Symfony\Component\HttpFoundation\Response as HttpFoundationResponse;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+/**
+ * HTTP Response.
+ *
+ * @internal
+ *
+ * Partially copied from \Symfony\Component\HttpClient\Response\ResponseTrait
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+final class Response implements ResponseInterface
+{
+    private $httpFoundationResponse;
+    private $browserKitResponse;
+    private $headers;
+    private $info;
+    private $content;
+    private $jsonData;
+
+    public function __construct(HttpFoundationResponse $httpFoundationResponse, BrowserKitResponse $browserKitResponse, array $info)
+    {
+        $this->httpFoundationResponse = $httpFoundationResponse;
+        $this->browserKitResponse = $browserKitResponse;
+
+        $this->headers = $httpFoundationResponse->headers->all();
+
+        // Compute raw headers
+        $responseHeaders = [];
+        foreach ($this->headers as $key => $values) {
+            foreach ($values as $value) {
+                $responseHeaders[] = sprintf('%s: %s', $key, $value);
+            }
+        }
+
+        $this->content = $httpFoundationResponse->getContent();
+        $this->info = [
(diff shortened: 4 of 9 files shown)


composer.json
```diff
@@ -118,5 +119,7 @@
             "url": "https://github.com/alanpoulain/FOSUserBundle",
             "no-api": true
         }
-    ]
+    ],
+    "minimum-stability": "dev",
```
teohhanhui: These changes are unnecessary. Just use `^4.3@dev`

baudev:
It could be great adding the possibility of testing objects with the [Symfony Validator](https://symfony.com/doc/current/validation.html). 
I recently created a gist ([link](https://gist.github.com/baudev/1983635797aa4c920c51260a1ece58db)) concerning that.

It would allow testing an object with the following assertion: 
```php
$this->assertSymfonyValidate($objectToTest);
```

What do you think about implementing it in `ApiTestCase`?

src/Bridge/Symfony/Bundle/Test/Response.php
```diff
@@ -0,0 +1,173 @@
...
+        if ($type) {
+            return $this->info[$type] ?? null;
+        }
+
+        return $this->info;
+    }
+
+    private function checkStatusCode()
+    {
+        if (500 <= $this->info['http_code']) {
```
GregoireHebert: Eventually is there a performance gain, by not using constant here?

composer.json
dunglas: Currently, the are, or I'll have to add some extra deps to composer.json: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpClient/composer.json#L23-L24

src/Bridge/Symfony/Bundle/Test/Response.php
dunglas: I shamelessly copy/pasted this method from the Symfony `@internal` trait

dunglas:
@baudev I plan to add extra assertions in other PR (like a response against a JSON Schema), but the one you're talking about should be added directly to Symfony IMHO (it can benefit to non-API Platform projects as well).

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -147,6 +148,14 @@ public function load(array $configs, ContainerBuilder $container)
         $this->registerMessengerConfiguration($config, $loader);
         $this->registerElasticsearchConfiguration($container, $config, $loader);
         $this->registerDataTransformerConfiguration($container);
+
+        if ('test' === $container->getParameter('kernel.environment')) {
```
dunglas: Maybe could we find something more elegant.

vasilvestre: Would accessing to kernel and doing ->isDebug() more elegant ?

dunglas: debug != testing. What I have in mind is more something like that https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L207

src/Bridge/Symfony/Bundle/Test/Response.php
nicolas-grekas: and constants dont fit here: "500" is much more common knowledge than any local constant name

composer.json
teohhanhui: I think that should be preferable to using these root-only options.

teohhanhui:
Actually, is there any reason why this goes into API Platform instead of in Symfony? I can see that the implementation is completely decoupled from API Platform?

dunglas:
I just added a nice API to deal with API errors (validation errors for instance), and `Basic` authentication:

```php
<?php

namespace App\Tests;

use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase;
use Hautelook\AliceBundle\PhpUnit\RefreshDatabaseTrait;
use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;

class ProjectsTest extends ApiTestCase
{
    use RefreshDatabaseTrait;

    public function testNotAbleToCreateProjectWithoutOwner()
    {
        $this->expectException(ClientExceptionInterface::class);
        $this->expectExceptionMessage(<<<ERROR
users: This collection should contain 1 element or more.
users: The current logged in user must be part of the users owning this resource.
ERROR
);

        $client = static::createClient();
        $response = $client->request('POST', '/api/projects', [
            'auth' =>  'dunglas@gmail.com:maman',
            'body' => [
                'name' => 'My project',
            ],
        ]);
        $response->getContent();
    }
}
```

src/Bridge/Symfony/Bundle/Test/Response.php
```diff
@@ -0,0 +1,182 @@
...
+        return $this->info;
+    }
+
+    /**
+     * Checks the status, and try to extract message if appropriate.
+     */
+    private function checkStatusCode()
+    {
+        $message = 'An error '.$this->info['http_code'].' occured.';
+        if (isset($this->headers['content-type'][0]) && false !== preg_match('#^application/(?:.+\+)?json#', $this->headers['content-type'][0])) {
```
bendavies: what's going on here?
if the regex does not match the content type then it must be an error?
but doesn't `application/problem+json` match?

teohhanhui: IMO we should avoid doing naive regex matching of MIME types...

src/Bridge/Symfony/Bundle/Test/Response.php
```diff
@@ -0,0 +1,182 @@
...
+            if ($json = json_decode($this->content, true)) {
+                // Try to extract the error message from Hydra or RFC 7807 error structures
+                $message = $json['hydra:description'] ?? $json['hydra:title'] ?? $json['detail'] ?? $json['title'] ?? $message;
```
teohhanhui: :disappointed: I think we should only extract the correct error message based on the `Content-Type`.

bendavies:
i would prefer optional throwing (configurable?). in my tests, is manually check response codes and response content for errors.

src/Bridge/Symfony/Bundle/Test/Response.php
dunglas: @bendavies we just test if the body is json in case of error. If it's not, we do nothing.

src/Bridge/Symfony/Bundle/Test/Response.php
dunglas: It will add extra complexity for no benefits. If one of these key exist, and a faulty HTTP status code is set (the case here), this key will definitely contain an error, regardless of the MIME type.

dunglas:
@bendavies you can do it too, it's natively supported by HttpClient, just pass `true` to `$response->getContent()` and it will not throw, then you'll be able to do your custom assertions.

bendavies:
yeah thanks i should have remembered that...

src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
```diff
@@ -0,0 +1,47 @@
...
+abstract class ApiTestCase extends KernelTestCase
+{
+    /**
+     * Creates a Client.
+     *
+     * @param array $options An array of options to pass to the createKernel method
+     *
+     * @return Client A Client instance
+     */
+    protected static function createClient(array $options = [])
```
cyve: Hi ! What do you think about moving this method in a trait and rename it `createApiClient` ? So we can use it with any other/custom TestCase.

dunglas: I'm not sure it is worth it. This method just accesses to a service from the container. It can easoly be copy/pasted/adapted.

src/Bridge/Symfony/Bundle/Test/Response.php
```diff
@@ -0,0 +1,191 @@
...
+        if ($throw) {
+            $this->checkStatusCode();
+        }
+
+        return $this->headers;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
```
soyuka: comment to remove

tests/Bridge/Symfony/Bundle/Test/ClientTest.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function testAuthBasic($basic): void
+    {
+        $client = self::createClient();
+        $client->enableReboot();
+        $response = $client->request('GET', '/secured_dummies', ['auth_basic' => $basic]);
+        $this->assertSame(200, $response->getStatusCode());
+    }
+
+    public function basicProvider(): iterable
```
soyuka: authBasicProvider? :dancer:

Issue #2552, opened 2019-02-22
Suboptimal count query on entity list endpoint


luispabon:
Imagine a simple `post` table that contains a couple of columns in addition to a unique primary key `id` column.

When retrieving a list of posts, the following query is issued:

```sql
SELECT count(DISTINCT p0_.id) AS sclr_0 FROM post p0_
```

Where the distinct part is fairly expensive, but also redundant since the PK it's counting is already unique:

```
 explain analyze select count(distinct(id)) from post;
                                                  QUERY PLAN                                                   
---------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=1548.00..1548.01 rows=1 width=8) (actual time=6.833..6.833 rows=1 loops=1)
   ->  Seq Scan on post  (cost=0.00..1523.00 rows=10000 width=4) (actual time=0.012..3.407 rows=10000 loops=1)
 Planning Time: 0.115 ms
 Execution Time: 6.875 ms
(4 rows)
```

If we run the query however without the `DISTINCT`, which on a unique PK is redundant:

```
explain analyze select count(*) from post;
                                                  QUERY PLAN                                                   
---------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=1548.00..1548.01 rows=1 width=8) (actual time=5.029..5.030 rows=1 loops=1)
   ->  Seq Scan on post  (cost=0.00..1523.00 rows=10000 width=0) (actual time=0.013..3.523 rows=10000 loops=1)
 Planning Time: 0.152 ms
 Execution Time: 5.075 ms
(4 rows)
```
I've actually averaged around 10 samples into these values above as the explain info is the same, but varies only on the timings. The difference is pretty substantial, on the region of around 25%, and perhaps might be on something psql does post-query to DISTINCT values.

The table size is only 10K records, in postgres. 

I believe this query can be optimised in API platform in a scenario where the primary key is a single column, which I would wager is rather common.

soyuka:
this call comes from the doctrine paginator IIRC, try using the partial pagination? Although we shouldn't have to paginate on a post request?

bendavies:
the distinct is done incase there are joins on the count query IIRC

teohhanhui:
The `DISTINCT` is needed when there are joined relations, otherwise you will have the wrong value from `COUNT`. Perhaps we could set this query hint when there are no relations:

[...]

PR #2611, opened 2019-03-13
Do not use DISTINCT where there are no joins

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes ?
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/issues/2552
| License       | MIT
| Doc PR        | na

@teohhanhui would this work?

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -23,6 +23,7 @@ use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\QueryBuilder;
+use Doctrine\ORM\Tools\Pagination\CountWalker;
 use Doctrine\ORM\Tools\Pagination\Paginator as DoctrineOrmPaginator;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
@@ -143,7 +144,14 @@ final class PaginationExtension implements ContextAwareQueryResultCollectionExte
      */
     public function getResult(QueryBuilder $queryBuilder, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $doctrineOrmPaginator = new DoctrineOrmPaginator($queryBuilder, $this->useFetchJoinCollection($queryBuilder, $resourceClass, $operationName));
+        $query = $queryBuilder->getQuery();
+
+        // Only one alias, without joins, disable the DISTINCT on the COUNT
+        if (1 === \count($queryBuilder->getAllAliases())) {
+            $query->setHint(CountWalker::HINT_DISTINCT, false);
+        }
+
+        $doctrineOrmPaginator = new DoctrineOrmPaginator($query, $this->useFetchJoinCollection($queryBuilder, $resourceClass, $operationName));
         $doctrineOrmPaginator->setUseOutputWalkers($this->useOutputWalkers($queryBuilder));
 
         if (null === $this->requestStack) {
--- a/tests/Bridge/Doctrine/Orm/Extension/PaginationExtensionTest.php
+++ b/tests/Bridge/Doctrine/Orm/Extension/PaginationExtensionTest.php
@@ -27,6 +27,7 @@ use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
+use Doctrine\ORM\Tools\Pagination\CountWalker;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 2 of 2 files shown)


src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -143,7 +144,16 @@ public function supportsResult(string $resourceClass, string $operationName = nu
      */
     public function getResult(QueryBuilder $queryBuilder, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $doctrineOrmPaginator = new DoctrineOrmPaginator($queryBuilder, $this->useFetchJoinCollection($queryBuilder, $resourceClass, $operationName));
+        $parts = $queryBuilder->getDQLPart('join');
+        $rootAlias = $queryBuilder->getRootAliases();
+        $query = $queryBuilder->getQuery();
+
+        // Only one alias, without joins, disable the DISTINCT on the COUNT
+        if (!isset($rootAlias[1]) && !($parts[$rootAlias[0] ?? ''] ?? true)) {
```
bendavies: `1 === count($queryBuilder->getAllAliases())` ?

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -143,7 +144,16 @@ public function supportsResult(string $resourceClass, string $operationName = nu
      */
     public function getResult(QueryBuilder $queryBuilder, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $doctrineOrmPaginator = new DoctrineOrmPaginator($queryBuilder, $this->useFetchJoinCollection($queryBuilder, $resourceClass, $operationName));
+        $parts = $queryBuilder->getDQLPart('join');
+        $rootAlias = $queryBuilder->getRootAliases();
+        $query = $queryBuilder->getQuery();
+
+        // Only one alias, without joins, disable the DISTINCT on the COUNT
```
alanpoulain:
```suggestion
        // Only one alias, without join, disable the DISTINCT on the COUNT
```

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -143,7 +144,16 @@ public function supportsResult(string $resourceClass, string $operationName = nu
      */
     public function getResult(QueryBuilder $queryBuilder, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $doctrineOrmPaginator = new DoctrineOrmPaginator($queryBuilder, $this->useFetchJoinCollection($queryBuilder, $resourceClass, $operationName));
+        $parts = $queryBuilder->getDQLPart('join');
+        $rootAlias = $queryBuilder->getRootAliases();
```
alanpoulain:
```suggestion
        $rootAliases = $queryBuilder->getRootAliases();
```

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
bendavies: don't agree

teohhanhui: Fun English fact: Both "join" and "joins" are correct here! :see_no_evil:

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
soyuka: I tried to avoid count, what is better? :p

teohhanhui: Could it be possible that there are no root aliases? :P

So maybe

```php
if (1 >= \count($queryBuilder->getAllAliases())) {
```

teohhanhui: Why avoid count? :see_no_evil:

soyuka: perf :p. If we have more then 1 alias then we need distinct right?

teohhanhui: Sorry, I don't understand the perf concern. :laughing: At most we have a handful of aliases. It's `O(n)` but `n` is very small... (Actually, it seems like it might even be `O(1)` as PHP stores the number of elements.)

src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -143,7 +144,15 @@ public function supportsResult(string $resourceClass, string $operationName = nu
      */
     public function getResult(QueryBuilder $queryBuilder, string $resourceClass = null, string $operationName = null, array $context = [])
     {
-        $doctrineOrmPaginator = new DoctrineOrmPaginator($queryBuilder, $this->useFetchJoinCollection($queryBuilder, $resourceClass, $operationName));
+        $parts = $queryBuilder->getDQLPart('join');
```
teohhanhui: Could be removed.

tests/Bridge/Doctrine/Orm/Extension/PaginationExtensionTest.php
```diff
@@ -887,6 +888,35 @@ public function testGetResult()
         $this->assertInstanceOf(PaginatorInterface::class, $result);
     }
 
+    public function testGetResultWithoutDistinct()
+    {
+        $configuration = new Configuration();
+
+        $entityManagerProphecy = $this->prophesize(EntityManagerInterface::class);
+        $entityManagerProphecy->getConfiguration()->willReturn($configuration)->shouldBeCalled();
```
teohhanhui: Actually, `shouldBeCalled` should not be called, unless we actually need to make such an assertion.

teohhanhui: We need a PR to remove useless assertions. Lol...

soyuka: These are part of the test IMO there where missing and I added them for consistency with the legacy fn.

teohhanhui:
Thanks @soyuka! :tada:

bendavies:
having tried this, i think it's had to take advantage of this because of eager loading.

to improve, left joins could be removed from the count query, i think?

teohhanhui:
@bendavies But eager loading only kicks in if you use those relations in your serializer groups, no? So, `DISTINCT` is indeed necessary for such a query. (Or did you mean rewriting the query for `COUNT`?)

bendavies:
ah, it's because i have one property with `@ApiProperty(fetchEager=true)`

bendavies:
yes, i meant writing the query for count, as left joins don't affect the final count.

teohhanhui:
> yes, i meant writing the query for count, as left joins don't affect the final count.

You'll have to propose that in Doctrine ORM. ;)

bendavies:
there is an improvement to be made here.
the count query has the distinct removed, but the next query to get the list of IDs for list does not
image

soyuka:
Mhhh, I'm not sure that we have access to another query where we could set the hint?!

teohhanhui:
@bendavies Same response as above: https://github.com/api-platform/core/pull/2611#issuecomment-473392432

> You'll have to propose that in Doctrine ORM. ;)

er1z:
> > You'll have to propose that in Doctrine ORM. ;)

Good luck. :D

bendavies:
ah, that's not our query? oh well

meyerbaptiste:
No, That comes from Doctrine: https://github.com/doctrine/orm/blob/32efbd3edd54a6f0fc331b6a0bf394fcacb88444/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L144-L149

teohhanhui:
Hmm... But `fetchJoinCollection` should be set to `false` when there are no joins, right?

So, I think we could improve the logic here: https://github.com/api-platform/core/blob/049d5aaa1ef80c94fee40a2f1b99118264a6e416/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php#L278-L296

teohhanhui:
> By default the pagination extension does the following steps to compute the correct result:
>
>     - Perform a Count query using `DISTINCT` keyword.
>     - Perform a Limit Subquery with `DISTINCT` to find all ids of the entity in from on the current page.
>     - Perform a WHERE IN query to get all results for the current page.
>
> This behavior is only necessary if you actually fetch join a to-many collection. You can disable this behavior by setting the $fetchJoinCollection flag to false; in that case only 2 instead of the 3 queries described are executed. We hope to automate the detection for this in the future.

https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/pagination.html

I don't know how I've missed this important optimization... :see_no_evil:

teohhanhui:
Note to self: we probably have to take care of fetch `eager` associations.

Issue #2605, opened 2019-03-12
[GraphQL] update mutation doesn't return updated entity type


nsams:
An update mutation returns as payload all fields of the updated entity. But it doesn't return the entity itself, which prevents the client (apollo in my case) to update the cache.

Example of a simple api-platform mutation:

    input updateUserInput {
      id: ID!
      name: String
      clientMutationId: String!
    }

    type updateUserPayload {
      id: ID!
      name: String
      clientMutationId: String!
    }

    mutation UpdateUser($input: updateUserInput!) {
      updateUser(input: $input) {
        id
        name
        clientMutationId
      }
    }

As reference, 
https://facebook.github.io/relay/docs/en/graphql-server-specification.html#mutations does return a nested ship and fraction with their type

and https://facebook.github.io/relay/graphql/mutations.htm also has a nested status (it's type is not visible here though)

lukasluecke:
I think you are right, in that the Payload should instead have a nested field `user`, that resolves to the created entity (since we currently follow the Relay spec in all other regards).

What I am not sure about is how to handle relations / references in the Payload, as the linked example in the Relay specification does (returns both a `Faction` and a `Ship` as nested fields). Maybe just handle the "easy" case first, and think about the rest later?

Also this would be a BC change for everyone using the GraphQL-API, not sure how easy that will be (even though it's still experimental, it's quite a big change - maybe make it configurable?)

alanpoulain:
For the relations / references, we let the user decide by using the normalization context.

It would be a big BC break, yes but:
- we say we follow the Relay spec and this is not the case here;
- GraphQL support is still experimental to allow this kind of change.

I think we should do it, even for 2.4.

lukasluecke:
I'm working on the PR right now.

What should be returned for `delete` mutations? Stick with just the ID for now? I can't find anything about that in the Relay spec (or anywhere else).

alanpoulain:
I think sticking with the ID is alright.

dunglas:
> It would be a big BC break

It indeed is a bug fix to me, and this component is still marked as experimental, so BC breaks are allowed by our policy.

PR #2612, opened 2019-03-13
Return mutation result as entity object

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes (modified!)
| Fixed tickets | #2605 
| License       | MIT
| Doc PR        | -

See #2605 for details.


lukasluecke:
@alanpoulain Currently breaks custom `normalization_context` for specific mutations - see failing Behat test. 

Since we are now re-using the same GraphQL-type, I don't think we can add the fields dynamically. Any ideas?

lukasluecke:
@alanpoulain I have modified the failing test, but need some input on wether or not this should be the "correct" behaviour.

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -79,6 +79,10 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             $normalizationContext = $resourceMetadata->getGraphqlAttribute($operationName ?? '', 'normalization_context', [], true);
             $normalizationContext['attributes'] = $this->fieldsToAttributes($info);
 
+            if ('delete' !== $operationName && isset($normalizationContext['attributes'][lcfirst($resourceMetadata->getShortName())])) {
```
dunglas: you should store the result of `lcfirst($resourceMetadata->getShortName())` in a variable instead of computing it 3 times.

alanpoulain:
@lukasluecke I've done some modifications to handle delete like the other mutations (I've changed my mind about it sorry), to manage the normalization context correctly and I've fixed some latent issues. I will push these modifications to your branch today I hope, if you're okay with that.

lukasluecke:
@alanpoulain Sure, no problem at all 🙂

alanpoulain:
Done. The changes:
- new types to handle different normalization contexts: `PayloadData` (the object inside the output of a mutation) and `NestedPayload` (when there is a relation in a `PayloadData`).
- the input / output behavior has been changed (@soyuka if you can validate): when there is no output, it's not possible schema-wise to query any field; when there is no input it's not possible schema-wise to add field in the input (but you can give it an empty input with or without a `clientMutationId`, in this case there is a null response for the output object and no persist is done).
- the depth issue has been fixed (its value was strange before because it was inside a loop)

@lukasluecke please tell me if you are okay with the changes :slightly_smiling_face:

lukasluecke:
@alanpoulain I fear that adding these new wrapper types might lead back to the original issue, because the `__typename` of the returned objects will be different.

Apollo client, which is afaik the most popular JS client, would not "match" this for caching purposes
> By default, InMemoryCache will attempt to use the commonly found primary keys of id and _id for the unique identifier if they exist along with __typename on an object
https://www.apollographql.com/docs/react/advanced/caching.html

The Relay spec is not totally clear about this, but from my understanding it also expects the "real" type (and full object) to get returned, not just a "similar" object.

Which lead me to my earlier comment:
> I think it might make sense to not allow setting a normalization context on the mutations

I'll leave this decision up to you, but personally I'd prefer ease of use over the marginal flexibility.

alanpoulain:
You're right, this is really not clear. But in this part: https://www.apollographql.com/docs/react/advanced/caching.html#automatic-updates, this sentence seems to say the output of a mutation could be different from the queried data: `One nice way to take advantage of this property as much as possible is to make your mutation results have all of the data necessary to update the queries previously fetched.`.

lukasluecke:
That read’s different to me, it only says that you need to query all the fields that changed, in order to get them updated in the cache.

The problem is that the typename is part of the cache key in Apollo, which means that using different types will break it. That is also the reason why the issue currently exists (pre-PR), since all the fields of the object have always been there (including id).

alanpoulain:
I deleted my previous comment because it was wrong indeed. Maybe a solution would be to change the type only if the normalization context is different. WDYT?

lukasluecke:
I think that's a good idea! 👍 If you have time to make the changes, I'd appreciate it - otherwise I'll try to squeeze them in next week 🙂 

We should also `__typename` to the Behat test queries to make sure this is enforced.

alanpoulain:
Done :slightly_smiling_face:

alanpoulain:
Thank you @lukasluecke!

PR #2615, opened 2019-03-13
WIP: Fix output normalization

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | https://github.com/api-platform/api-platform/issues/1045
| License       | MIT
| Doc PR        | na

Needs work, just published it so that @phryneas can try.

### DIFF
--- a/src/GraphQl/Serializer/ItemNormalizer.php
+++ b/src/GraphQl/Serializer/ItemNormalizer.php
@@ -15,7 +15,10 @@ namespace ApiPlatform\Core\GraphQl\Serializer;
 
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Serializer\ItemNormalizer as BaseItemNormalizer;
+use ApiPlatform\Core\Util\ClassInfoTrait;
+use Symfony\Component\Serializer\Exception\LogicException;
 use Symfony\Component\Serializer\Exception\UnexpectedValueException;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
  * GraphQL normalizer.
@@ -24,6 +27,8 @@ use Symfony\Component\Serializer\Exception\UnexpectedValueException;
  */
 final class ItemNormalizer extends BaseItemNormalizer
 {
+    use ClassInfoTrait;
+
     const FORMAT = 'graphql';
     const ITEM_KEY = '#item';
 
@@ -42,11 +47,22 @@ final class ItemNormalizer extends BaseItemNormalizer
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        $object = $this->transformOutput($object, $context);
+        if (!$this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
+            if (!$this->serializer instanceof NormalizerInterface) {
+                throw new LogicException('Cannot normalize the transformed value because the injected serializer is not a normalizer');
+            }
+
+            $context['api_normalize'] = true;
+            $context['resource_class'] = $this->getObjectClass($transformed);
+
+            return $this->serializer->normalize($transformed, $format, $context);
+        }
+
         $data = parent::normalize($object, $format, $context);
         if (!\is_array($data)) {
             throw new UnexpectedValueException('Expected data to be an array');
         }
+
         $data[self::ITEM_KEY] = serialize($object); // calling serialize prevent weird normalization process done by Webonyx's GraphQL PHP
 
         return $data;
--- a/src/Hal/Serializer/ItemNormalizer.php
+++ b/src/Hal/Serializer/ItemNormalizer.php
@@ -17,8 +17,10 @@ use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Serializer\AbstractItemNormalizer;
 use ApiPlatform\Core\Serializer\ContextTrait;
 use ApiPlatform\Core\Util\ClassInfoTrait;
+use Symfony\Component\Serializer\Exception\LogicException;
 use Symfony\Component\Serializer\Exception\UnexpectedValueException;
 use Symfony\Component\Serializer\Mapping\AttributeMetadataInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
  * Converts between objects and array including HAL metadata.
@@ -48,11 +50,19 @@ final class ItemNormalizer extends AbstractItemNormalizer
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        if ($this->handleNonResource && ($context['api_normalize'] ?? false) || null !== $outputClass = $this->getOutputClass($this->getObjectClass($object), $context)) {
-            if (isset($outputClass)) {
-                $object = $this->transformOutput($object, $context);
+        if (!$this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
+            if (!$this->serializer instanceof NormalizerInterface) {
+                throw new LogicException('Cannot normalize the transformed value because the injected serializer is not a normalizer');
             }
 
+            $context['api_normalize'] = true;
+            $context['resource_class'] = $this->getObjectClass($transformed);
+
+            return $this->serializer->normalize($transformed, $format, $context);
+        }
+
+        if ($this->handleNonResource && $context['api_normalize'] ?? false) {
+            $object = $this->transformOutput($object, $context);
             $data = $this->initContext($this->getObjectClass($object), $context);
             $rawData = parent::normalize($object, $format, $context);
             if (!\is_array($rawData)) {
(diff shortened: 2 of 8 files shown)


src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -42,7 +44,16 @@ public function supportsNormalization($data, $format = null, array $context = []
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        $object = $this->transformOutput($object, $context);
+        $transformed = $this->transformOutput($object, $context);
+        // the object has been transformed, we need to find a new normalizer for it
+        if ($object !== $transformed) {
+            if (!$this->serializer instanceof NormalizerInterface) {
+                throw new LogicException('Cannot normalize transformed resource because the injected serializer is not a normalizer');
```
teohhanhui: I think it should be a `RuntimeException` instead.

phryneas:
Hi there, I've tried it, and it works so far.

But this reveals a follow-up-problem:

`id` is not handled as it would be in graphql.

Ususally, the normalized `id` property would then be something like `/api/languages/1`, but now it just returns `1`. Even worse, in this case I have `id` in the main Entity annotated with `@ApiProperty(identifier=false)` and `code` with `@ApiProperty(identifier=true)` so that I get `/api/languages/en` - this still works for querying those ids, but the output is still `1`.

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -48,11 +48,12 @@ public function supportsNormalization($data, $format = null, array $context = []
...
     {
-        if ($this->handleNonResource && ($context['api_normalize'] ?? false) || null !== $outputClass = $this->getOutputClass($this->getObjectClass($object), $context)) {
-            if (isset($outputClass)) {
-                $object = $this->transformOutput($object, $context);
-            }
+        if (false === $this->handleNonResource) {
+            $transformed = $this->transformOutput($object, $context);
+        }
 
+        if (($this->handleNonResource && $context['api_normalize'] ?? false) || (isset($transformed) && $transformed !== $object)) {
```
teohhanhui: Brackets around the second half of the condition are unnecessary.

teohhanhui: Also,

```php
($this->handleNonResource && $context['api_normalize'] ?? false)
```

should be kept as

```php
$this->handleNonResource && ($context['api_normalize'] ?? false)
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -116,7 +116,7 @@ public function hasCacheableSupportsMethod(): bool
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        if ($this->handleNonResource && ($context['api_normalize'] ?? false) || null !== $outputClass = $this->getOutputClass($this->getObjectClass($object), $context)) {
+        if ($this->handleNonResource) {
```
teohhanhui:
```php
if ($this->handleNonResource && ($context['api_normalize'] ?? false)) {
```

src/Serializer/ItemNormalizer.php
```diff
@@ -49,7 +49,13 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        return parent::normalize($this->transformOutput($object, $context), $format, $context);
+        if (false === $this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
```
teohhanhui:
```php
!$this->handleNonResource
```

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -62,11 +62,13 @@ public function supportsNormalization($data, $format = null, array $context = []
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        if ($this->handleNonResource && ($context['api_normalize'] ?? false) || null !== $outputClass = $this->getOutputClass($this->getObjectClass($object), $context)) {
-            if (isset($outputClass)) {
-                $object = $this->transformOutput($object, $context);
-            }
+        if (false === $this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
```
teohhanhui:
```php
!$this->handleNonResource
```

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -62,12 +62,10 @@ public function supportsNormalization($data, $format = null, array $context = []
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        if ($this->handleNonResource && ($context['api_normalize'] ?? false) || null !== $outputClass = $this->getOutputClass($this->getObjectClass($object), $context)) {
-            if (isset($outputClass)) {
-                $object = $this->transformOutput($object, $context);
-            }
+        if (false === $this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
```
teohhanhui:
```php
!$this->handleNonResource
```

src/Hal/Serializer/ItemNormalizer.php
```diff
@@ -48,18 +48,10 @@ public function supportsNormalization($data, $format = null, array $context = []
...
-            if (isset($outputClass)) {
-                $object = $this->transformOutput($object, $context);
-            }
-
-            $data = $this->initContext($this->getObjectClass($object), $context);
-            $rawData = parent::normalize($object, $format, $context);
-            if (!\is_array($rawData)) {
-                return $rawData;
-            }
+        if (false === $this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
```
teohhanhui:
```php
!$this->handleNonResource
```

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -42,11 +42,17 @@ public function supportsNormalization($data, $format = null, array $context = []
      */
     public function normalize($object, $format = null, array $context = [])
     {
-        $object = $this->transformOutput($object, $context);
+        if (false === $this->handleNonResource && $object !== $transformed = $this->transformOutput($object, $context)) {
```
teohhanhui:
```php
!$this->handleNonResource
```

src/Serializer/AbstractItemNormalizer.php
teohhanhui: Or perhaps we should do the check differently:

```php
if ($this->handleNonResource) {
    if (!($context['api_normalize'] ?? false)) {
        throw new UnexpectedValueException('"api_normalize" must be set to true in context to normalize non-resource');
    }

    return parent::normalize($object, $format, $context);
}
```

teohhanhui: I think the call to `initContext` is useless.

phryneas:
There seems to be a regression - with the current patch, I'm back to the old behavior.

Also, for the identifier problem: Should I open an extra ticket for that or is that something you can fix in the context of this PR?

src/Serializer/AbstractItemNormalizer.php
soyuka: no we need to change the resource_class

phryneas:
Oh, no, not a regression to old behaviour - it's a new one:

```
{
  "errors": [
    {
      "debugMessage": "No item route associated with the type \"App\\Entity\\Dto\\LanguageOutput\".",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 5,
          "column": 9
        }
      ],
      "path": [
        "french",
        "edges",
        0,
        "node",
        "id"
      ],
      "trace": [
        {
          "file": "/app/src/vendor/api-platform/core/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php",
          "line": 47,
          "call": "ApiPlatform\\Core\\Bridge\\Symfony\\Routing\\RouteNameResolver::getRouteName('App\\Entity\\Dto\\LanguageOutput', 'item', array(0))"
        },
        {
          "file": "/app/src/vendor/api-platform/core/src/Cache/CachedTrait.php",
          "line": 44,
          "call": "ApiPlatform\\Core\\Bridge\\Symfony\\Routing\\CachedRouteNameResolver::ApiPlatform\\Core\\Bridge\\Symfony\\Routing\\{closure}()"
        },
        {
          "file": "/app/src/vendor/api-platform/core/src/Bridge/Symfony/Routing/CachedRouteNameResolver.php",
          "line": 48,
          "call": "ApiPlatform\\Core\\Bridge\\Symfony\\Routing\\CachedRouteNameResolver::getCached('route_name_b8973be077493b412a51bb26a9bea702', instance of Closure)"
        },
[...]
```

I hope this helps - and thank your for your effort!

soyuka:
> There seems to be a regression - with the current patch, I'm back to the old behavior.
Also, for the identifier problem: Should I open an extra ticket for that or is that something you can fix in the context of this PR?

I know :p while fixing your issue yesterday I broke the rest, I need to find the correct solution and I'll let you know when it works. FYI I've reproduced your issue, am trying to fix it and I'll let you know when it works.

phryneas:
Oh, okay :) I'm going home for the day now, but I'll take my laptop with me - so if you have something you need tested in the evening, I hope I'll get to it :)

src/Serializer/AbstractItemNormalizer.php
teohhanhui: Uhh, because `AbstractObjectNormalizer::normalize` will call `getAttributeValue`. Gotcha!

teohhanhui: Hmm, I don't see where it'd be used... We're going to the parent and not coming back.

teohhanhui:
@soyuka Why was this merged? :x

PR #2622, opened 2019-03-18
Preserve manual join in FilterEagerLoadingExtension

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Given that the custom extension priority were changed, the FilterEagerLoadingExtension will need to be update to support "manual external" join in the querybuilder to avoid people in such scenario to have to define a negative priority -33 to -63 to be triggered after this extension but before pagination.

Without this fix, we could end up with a where condition using one or several aliases that haven't been joined previously


soyuka:
Nice catch! Thanks @antograssiot !

PR #2628, opened 2019-03-19
Add a helper to use Inputs as messages

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | wiip

### DIFF
--- a/src/Bridge/Symfony/Messenger/DataTransformer.php
+++ b/src/Bridge/Symfony/Messenger/DataTransformer.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Messenger;
+
+use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+
+/**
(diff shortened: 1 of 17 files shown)


src/Bridge/Symfony/Messenger/DataTransformer/InputMessageDataTransformer.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public function transform($object, string $to, array $context = [])
+    {
+        return $object;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsTransformation($object, string $to, array $context = []): bool
```
GregoireHebert: Not really related to this but I suggest changing `$object` with `$data` while it's still possible. It's a bit confusing.

dunglas:
I like the idea, but I'm not a fond of the interface. Couldn't we use a new value for the `messenger` attribute instead? `messeger="input"` for instance?

soyuka:
Done @dunglas wdyt?

src/Bridge/Symfony/Messenger/DataTransformer.php
```diff
@@ -0,0 +1,71 @@
...
+            return false;
+        }
+
+        $metadata = $this->resourceMetadataFactory->create($context['resource_class'] ?? $to);
+
+        if (!isset($context['operation_type'])) {
+            return 'input' === $metadata->getAttribute('messenger');
+        }
+
+        return 'input' === $this->resourceMetadataFactory->create($context['resource_class'] ?? $to)
```
dunglas:
```suggestion
        return 'input' === $metadata
```

dunglas:
looks good to me

Issue #2593, opened 2019-03-09
[Performance] questions and suggestions


bastnic:
Hello guys, 

Just wanted to share a few thoughts I have.
With a few big api platform I see some bad tendencies happening again and again that totally crashes the performance in both dev and prod env. In dev env it's quite difficult to live with.

1/ custom normalizers are rarely using the CacheableSupportsMethodInterface interface.

Normalization : we have a fully hydrated entity, it's easy to cache with some instanceof and an interface.

Denormalization : we have an array of yet unknown data and a string representing the class to hydrate. We use `class_implements` on the string classname to check if it's cacheable but it's not as nice.

If we are not able to use cache on denormalization, it's better to separate the normalizer and the denormalizer in two different classes so the normalizer can be cacheable even if the denormalizer is not.

It's really *really* important that the normalizer to be cacheable. 

On our application, with a critical route we have 36000 `getNormalizer` calls. 4 normalizers were not cachable, so it's 36000*4 `supportNormmosation` calls. Even for  simple stuff as native-* stuff.
Of course this route is cached in prod mode, but it's not in dev mode

suggestion : it's so critical that perhaps it should be visible somewhere in the profiler.

2/ IriConverter::getIriFromItem is costly when it's called.

What happens?
- on normalization, we arrive on `ApiPlatform\Core\Serializer\ItemNormalizer` so on `ApiPlatform\Core\Serializer\AbstractItemNormalizer::normalize`
- as `$context['resources']` seems to be always set, i calls `$resource = $context['iri'] ?? $this->iriConverter->getIriFromItem($object);`
- it does two things:
    - find the identifier (in our case "id" 100% on the time). This is the worst part. For each call, it goes straight to phpdocumentor on the file itself just to be able to extract "hey the identifier property is 'id'". 
    - generate the route via the router

As we have a react app with multiples http calls to the api backend, it creates a concurence on the file read. As we are on mac os x with an nfs and very poor access disk performance, every filesystem call cost us a lot.

There is multiple layer of caching that are all disabled in test/dev mode via the `ApiPlatformExtension::registerCacheConfiguration`. We can set cache back by enabling the `api_platform.metadata_cache` parameter to true in our app. But it enables the cache for every part of the stack. If we are in dev mode it's because we touch to the resource & serialization config a lot. But we NEVER change the identifier property. The cache removal is a very big on/off performance setback.

[...]

PR #2629, opened 2019-03-19
Do not disable cache in dev

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2593 (partially)
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/766

With the cache enabled, changing API resource metadata in `dev` still works without a manual `cache:clear`.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -436,17 +436,21 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
      */
     private function registerCacheConfiguration(ContainerBuilder $container)
     {
-        // Don't use system cache pool in dev
-        if ($container->hasParameter('api_platform.metadata_cache') ? $container->getParameter('api_platform.metadata_cache') : !$container->getParameter('kernel.debug')) {
+        if (!$container->hasParameter('api_platform.metadata_cache')) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


borNfreee:
Could you please explain what should be used instead? Or what are the defaults for prod/test environments?

It's not clear what the end user should do from the deprecation warning. Just remove and what is the alternative?

Thanks

teohhanhui:
Yes, just remove it. The cache is enabled. There should not be a need to disable it.

borNfreee:
> The cache is enabled.

The question is what is the adapter used for each env: `dev`, `test`, `prod`?

From what I have learned in previous `master`, the default cache adapter was `cache.system`, which is, in our case, is `APCu`. And we have to enable it explicitly, by setting `api_platform.metadata_cache` to `true`. In dev mode, it was overridden to `ArrayAdapter`.

What are the defaults now? Will api-platform use `cache.system` by default, even in `dev` mode? If so, how it will work without a "a manual `cache:clear`"?

Hope it will help other people that will have the same questions ;)

teohhanhui:
> What are the defaults now? Will api-platform use cache.system by default, even in dev mode?

Yes.

> If so, how it will work "without a manual cache:clear"?

We've always been using the `DirectoryResource` so the container is automatically rebuilt when any file in the directory is modified. However, perhaps adding a new file would require a manual `cache:clear`, just like for Symfony Serializer / Validator metadata. (Not sure about this point. I'll try to verify.)

soyuka:
Thanks @teohhanhui !

PR #2630, opened 2019-03-20
Various quality fixes

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/MongoDbOdm/Filter/BooleanFilter.php
+++ b/src/Bridge/Doctrine/MongoDbOdm/Filter/BooleanFilter.php
@@ -36,7 +36,7 @@ final class BooleanFilter extends AbstractFilter
 {
     use BooleanFilterTrait;
 
-    const DOCTRINE_BOOLEAN_TYPES = [
+    public const DOCTRINE_BOOLEAN_TYPES = [
         MongoDbType::BOOL => true,
         MongoDbType::BOOLEAN => true,
     ];
--- a/src/Metadata/Property/PropertyMetadata.php
+++ b/src/Metadata/Property/PropertyMetadata.php
@@ -55,10 +55,8 @@ final class PropertyMetadata
 
     /**
      * Gets type.
-     *
-     * @return Type|null
      */
-    public function getType()
+    public function getType(): ?Type
     {
         return $this->type;
     }
@@ -76,20 +74,16 @@ final class PropertyMetadata
 
     /**
      * Gets description.
-     *
-     * @return string|null
      */
-    public function getDescription()
+    public function getDescription(): ?string
     {
         return $this->description;
     }
 
     /**
      * Returns a new instance with the given description.
-     *
-     * @param string $description
      */
-    public function withDescription($description): self
+    public function withDescription(string $description): self
     {
         $metadata = clone $this;
         $metadata->description = $description;
@@ -99,10 +93,8 @@ final class PropertyMetadata
 
     /**
      * Is readable?
-     *
-     * @return bool|null
      */
-    public function isReadable()
+    public function isReadable(): ?bool
     {
         return $this->readable;
     }
@@ -120,10 +112,8 @@ final class PropertyMetadata
 
     /**
      * Is writable?
-     *
-     * @return bool|null
      */
-    public function isWritable()
+    public function isWritable(): ?bool
     {
         return $this->writable;
     }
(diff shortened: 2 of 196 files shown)


src/Bridge/Doctrine/MongoDbOdm/Filter/BooleanFilter.php
```diff
@@ -36,7 +36,7 @@ final class BooleanFilter extends AbstractFilter
 {
     use BooleanFilterTrait;
 
-    const DOCTRINE_BOOLEAN_TYPES = [
+    public const DOCTRINE_BOOLEAN_TYPES = [
```
soyuka: I'm -1 on this `public const` thing it's just noise IMO. Symfony is not using it for now?

src/Metadata/Property/PropertyMetadata.php
```diff
@@ -281,14 +255,32 @@ public function withAttributes(array $attributes): self
...
-     * @return string|null
+     * @deprecated since version 2.4, to be removed in 3.0.
      */
-    public function isChildInherited()
+    public function isChildInherited(): ?string
     {
-        return $this->childInherited;
+        @trigger_error(sprintf('The use of "%1$s::isChildInherited()" is deprecated since 2.4 and will be removed in 3.0. Use "%1$s::getChildInherited()" or "%1$s::hasChildInherited()" directly instead.', __CLASS__), E_USER_DEPRECATED);
+
+        return $this->getChildInherited();
```
soyuka: why the change?

meyerbaptiste: The name of the method and the return type are inconsistent... It's a nonsense.

src/Bridge/Doctrine/MongoDbOdm/Filter/BooleanFilter.php
meyerbaptiste: We should now pay attention to the visibility of our constants and define them explicitly. It's by default in the PHP-CS-Fixer ruleset `@PHP71Migration`.

teohhanhui: > It's by default in the PHP-CS-Fixer ruleset @PHP71Migration.

But we are already using both `@PHP71Migration` and `@PHP71Migration:risky` rulesets?

teohhanhui:
```
visibility_required [@PSR2, @Symfony, @PhpCsFixer, @PHP71Migration]

Visibility MUST be declared on all properties and methods; abstract and final MUST be declared before the visibility; static MUST be declared after the visibility.

Configuration options:

- elements (a subset of ['property', 'method', 'const']): the structural elements to fix (PHP >= 7.1 required for const); defaults to ['property', 'method']
```

So it's not enabled by default. If we're going to make this change, we must change that option.

soyuka: By default constants are public and it won't change, we should defined them if they're not public IMO.

https://secure.php.net/manual/en/migration71.new-features.php#migration71.new-features.class-constant-visibility

meyerbaptiste: Yep but because of the alphabetical order in the configuration, the `@Symfony` ruleset has the advantage and it uses the default configuration: `method` + `property` vs. `method` + `property` + `const` for the `@ PHP71Migration` ruleset.

meyerbaptiste: > By default constants are public and it won't change

By default method are public but you define a visibility explicitly even if it is public.

teohhanhui: Okay, let's add the correct configuration for this rule then.

teohhanhui: > Yep but because of the alphabetical order in our configuration, the @Symfony ruleset has the advantage

That is indeed unexpected, but might as well since I don't think we would ever find a "correct" order. :laughing:

teohhanhui:
Anyway, I suspect this PR should target `2.4` first, to reduce future merge conflicts.

meyerbaptiste:
Base branch changed.

src/Bridge/Doctrine/MongoDbOdm/Filter/BooleanFilter.php
meyerbaptiste: I just updated the `.php_cs.dist` file.

dunglas:
Thanks @meyerbaptiste

PR #2631, opened 2019-03-20
json-ld: correct embeded context

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na 
| License       | MIT
| Doc PR        | na

This fixes the JSON-LD embeded context when using outputs.

### DIFF
--- a/tests/JsonLd/ContextBuilderTest.php
+++ b/tests/JsonLd/ContextBuilderTest.php
@@ -23,6 +23,7 @@ use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use ApiPlatform\Core\Metadata\Resource\ResourceNameCollection;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Dto\OutputDto;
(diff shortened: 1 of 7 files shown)


tests/JsonLd/ContextBuilderTest.php
```diff
@@ -161,7 +162,7 @@ public function testAnonymousResourceContextWithIri()
             '@context' => [
                 '@vocab' => '#',
                 'hydra' => 'http://www.w3.org/ns/hydra/core#',
-                'dummyPropertyA' => '/dummies/dummyPropertyA',
+                'dummyPropertyA' => 'Dummy/dummyPropertyA',
             ],
             '@id' => '/dummies',
             '@type' => 'Dummy',
```
teohhanhui: I think there should be another test with `$context['api_resource']`?

tests/JsonLd/ContextBuilderTest.php
```diff
@@ -141,32 +142,56 @@ public function testAnonymousResourceContext()
             '@context' => [
                 '@vocab' => '#',
                 'hydra' => 'http://www.w3.org/ns/hydra/core#',
-                'dummyPropertyA' => $iri.'/dummyPropertyA',
+                'dummyPropertyA' => 'Dummy/dummyPropertyA',
             ],
             '@id' => $iri,
+            '@type' => 'Dummy',
```
teohhanhui: This test might be confusing because `Dummy` is indeed annotated with `@ApiResource`? :stuck_out_tongue_closed_eyes:

PR #2633, opened 2019-03-20
Create a resource with a falsy input

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

Adds a test to write on a resource with an `input=false`. Should we throw an error if the user sent data and only accept empty data?

### DIFF
--- a/src/EventListener/SerializeListener.php
+++ b/src/EventListener/SerializeListener.php
@@ -61,7 +61,14 @@ final class SerializeListener
 
         $context = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
 
-        if (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class']) {
+        if (
+            (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class'])
+            ||
+            (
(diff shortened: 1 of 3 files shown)


src/EventListener/SerializeListener.php
```diff
@@ -61,7 +61,9 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
 
         $context = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
 
-        if (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class']) {
+        if ((isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class']) ||
+        (null === $controllerResult && isset($context['input']) && \array_key_exists('class', $context['input']) && null === $context['input']['class'])
+        ) {
```
bendavies: this is if is sort of unreadable now.
can it be improved?

soyuka: I've done my best see updated code :p.

PR #2639, opened 2019-03-22
Don't use fetchJoinCollection on Doctrine Paginator when not needed

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes?
| New feature?  | no?
| BC breaks?    | no?
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/2611#issuecomment-474909777, https://github.com/api-platform/core/pull/2611#issuecomment-474910906
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/771

Refactor `QueryChecker` etc.

I've removed the flawed handling of "join by class", as we simply have not enough information to do the right thing there. The main caveat (as was before) is that our checking covers the "80%":

- join by association
- no complex expression in `SELECT` / `ORDER BY` clauses

but will be confused in the remaining "20%".

For example:

🗹 Supported:

```dql
SELECT d, a_1
FROM Dummy d
LEFT JOIN d.relatedDummies a_1
ORDER BY a_1.name
```

🗷 Not supported:

```dql
SELECT d, a_1
FROM Dummy d
LEFT JOIN RelatedDummy a_1 WITH 1 = 1
ORDER BY a_1.name
```

🗷 Not supported:

```dql
SELECT d, LENGTH(a_1.name) AS HIDDEN score
FROM Dummy d
LEFT JOIN d.relatedDummies a_1
ORDER BY score
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,6 @@
         "jangregor/phpstan-prophecy": "^0.3",
         "justinrainbow/json-schema": "^5.0",
         "nelmio/api-doc-bundle": "^2.13.3",
-        "php-mock/php-mock-phpunit": "^2.0",
         "phpdocumentor/reflection-docblock": "^3.0 || ^4.0",
         "phpdocumentor/type-resolver": "^0.3 || ^0.4",
         "phpspec/prophecy": "^1.8",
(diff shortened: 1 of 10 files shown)


teohhanhui:
Since we have no way to cover all use cases, and "we don't know when we don't know" (so we can't even fall back to safe defaults, unless we remove the whole guessing altogether), it's important that we let the user configure these in the resource / operation attributes (done in this PR).

PR #2640, opened 2019-03-22
Remove a BC break introduced by #2243

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a


meyerbaptiste:
Huum, there is also a BC break in the constructor and more difficult to handle because of the last argument `$properties`. I think the easiest way is to use a setter instead of the constructor for the `$existsParameterName` argument.  WDYT @api-platform/core-team?
Not so easy to deal with a non-final class associated to an abstract service definition when we want to inject a new argument in the constructor...

dunglas:
What’s the problem with the constructor? We should mark these class @final.

teohhanhui:
We cannot change a non-final class to final. That would be a BC break.

teohhanhui:
We can add the `@final` annotation, but we should also fix the BC break in the constructor.

teohhanhui:
This should target `2.4` as it's a bug fix.

antograssiot:
The regression is only in master @theohhanhui

meyerbaptiste:
I just added the BC for the constructor @api-platform/core-team.

PR #2642, opened 2019-03-24
Fix Nested Input DTOs Resource Not Found Error

ragboyjr:
- Fixes issue where trying to use nested dto's would
  use the wrong resource class due to the input class
  being passed into the context for the child dto and
  api platform would report a Resource not found error
  accordingly
- Separated input from resource class to ensure that
  resource_class is only set if the class is ACTUALLy
  a resource class

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/EventListener/DeserializeListener.php
+++ b/src/EventListener/DeserializeListener.php
@@ -97,7 +97,7 @@ final class DeserializeListener
         $request->attributes->set(
             'data',
             $this->serializer->deserialize(
-                $requestContent, $context['resource_class'], $format, $context
+                $requestContent, $context['input']['class'] ?? $context['resource_class'], $format, $context + ['api_denormalize' => true]
             )
         );
     }
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -175,7 +175,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer implement
 
         if (null !== $inputClass && null !== $dataTransformer = $this->getDataTransformer($data, $context['resource_class'], $context)) {
             $data = $dataTransformer->transform(
-                parent::denormalize($data, $inputClass, $format, ['input' => ['class' => null]] + $context), // ensure we don't transformData for nested relations
+                parent::denormalize($data, $inputClass, $format, ['input' => null] + $context), // ensure we don't transformData for nested relations
                 $class,
                 $context
             );
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -170,13 +170,12 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer implement
      */
     public function denormalize($data, $class, $format = null, array $context = [])
     {
-        $context['api_denormalize'] = true;
-        $context['resource_class'] = $class;
+        $context['input_class'] = $class;
         $inputClass = $this->getInputClass($class, $context);
 
-        if (null !== $inputClass && null !== $dataTransformer = $this->getDataTransformer($data, $class, $context)) {
+        if (null !== $inputClass && null !== $dataTransformer = $this->getDataTransformer($data, $context['resource_class'], $context)) {
             $data = $dataTransformer->transform(
-                parent::denormalize($data, $inputClass, $format, ['resource_class' => $inputClass] + $context),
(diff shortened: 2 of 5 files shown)


src/EventListener/DeserializeListener.php
```diff
@@ -97,7 +97,7 @@ public function onKernelRequest(GetResponseEvent $event): void
         $request->attributes->set(
             'data',
             $this->serializer->deserialize(
-                $requestContent, $context['resource_class'], $format, $context
+                $requestContent, $context['input']['class'] ?? $context['resource_class'], $format, $context + ['api_denormalize' => true]
```
soyuka: Not sure that we should do this here, normally the denormalizer checks if the given class has an input in which case it'll let the denormalization be handled by another normalizer.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -170,13 +170,12 @@ public function supportsDenormalization($data, $type, $format = null, array $con
...
-        $context['api_denormalize'] = true;
-        $context['resource_class'] = $class;
+        $context['input_class'] = $class;
         $inputClass = $this->getInputClass($class, $context);
 
-        if (null !== $inputClass && null !== $dataTransformer = $this->getDataTransformer($data, $class, $context)) {
+        if (null !== $inputClass && null !== $dataTransformer = $this->getDataTransformer($data, $context['resource_class'], $context)) {
             $data = $dataTransformer->transform(
-                parent::denormalize($data, $inputClass, $format, ['resource_class' => $inputClass] + $context),
+                parent::denormalize($data, $inputClass, $format, ['input' => ['class' => null]] + $context), // ensure we don't transformData for nested relations
```
soyuka: I see, can't we just remove the input metadata from the context here?

ragboyjr: y, that would work fine too I think.

src/EventListener/DeserializeListener.php
ragboyjr: The basic idea is that we are respecting the contract of the Serializer deserialize. The class we pass into it to deserialize is the class we get out which keeps the AbstractItemNormalizer simpler.

An alternative to this is we update the serializerContextBuilder to add a new parameter like `$context['input_class']` and also set 'api_denormalize` to true so that this calling is a lot easier.

src/Serializer/AbstractItemNormalizer.php
ragboyjr: Done.

src/EventListener/DeserializeListener.php
soyuka: > An alternative to this is we update the serializerContextBuilder to add a new parameter like $context['input_class'] and also set 'api_denormalize` to true so that this calling is a lot easier.

Let's see what @teohhanhui thinks about that. I'm afraid that this kind of changes will break some of the current use cases.

teohhanhui: > The basic idea is that we are respecting the contract of the Serializer deserialize. The class we pass into it to deserialize is the class we get out which keeps the AbstractItemNormalizer simpler.

I like the idea. But yeah, probably it could be done in `SerializerContextBuilder` if it makes things cleaner.

teohhanhui: But what's the reason of having both `$context['input']['class']` and `$context['input_class']`? Can't we just `unset($context['input'])` instead?

ragboyjr: @teohhanhui naming could probably be updated, `input_class` is more like `target_class` or `class_to_denormalize`. `['input']['class']` would refer to the actual input class configured for a resource, and `resource_class` would be the actual resource class.

The problem is that the code uses `$context['resource_class']` throughout the codebase.

In the AbstractItemNormalizer, it uses it as either the current class being denormalized or normalized, this is needed because in some of the protected methods that it overrides don't have access to the current class being denormalized/normalized, so using the value from context allows apip to do what it needs to do.

The other way `$context['resource_class']` is as the ACTUAL resource class to get resource metadata from.

Before input/output classes, this was fine because AbstractItemNormalizer only worked on resource classes, so the distinction between those two uses cases wasn't as evident and had no need to change.

But now that we support non-resource classes in AbstractItemNormalizer, we need to keep these fields separate because it breaks the semantics of `resource_class` and how it's used throughout the codebase.

PR #2643, opened 2019-03-25
[2.4] Allow autowiring of class extending SearchFilter

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

Remove a BC break introduced in #2630

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
@@ -41,7 +41,7 @@ class SearchFilter extends AbstractContextAwareFilter implements SearchFilterInt
     /**
      * @param RequestStack|null $requestStack No prefix to prevent autowiring of this deprecated property
      */
-    public function __construct(ManagerRegistry $managerRegistry, $requestStack, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null)
+    public function __construct(ManagerRegistry $managerRegistry, $requestStack = null, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null)
     {
         parent::__construct($managerRegistry, $requestStack, $logger, $properties);
 


soyuka:
https://github.com/api-platform/core/pull/2640 related? @meyerbaptiste can we merge this?

meyerbaptiste:
No. Did my change break the autowiring @antograssiot?

soyuka:
I'm asking because it has been changed in the cs fixes you did in: https://github.com/api-platform/core/commit/97fa954dabd4d5d49f3dd92171ae6c677a97f50b#diff-77ec54980605d6062780fcd93ba31c64

antograssiot:
@soyuka No #2640 is not related and on master only.
@meyerbaptiste yes it does break autowiring because the argument ha sno typehint (and it's normal as it is deprecated) but there's no default value anymore. Anyway it is a BC break IMO, it is not compliant with Symfony backward compatibility promise to remove default value of an argument in a public method

meyerbaptiste:
> it is not compliant with Symfony backward compatibility promise to remove default value of an argument in a public method

But a default value here is a non-sense because the following argument is not optional (`IriConverterInterface $iriConverter`)...

antograssiot:
I know/understand that it is a "non sense" to have this optional value before mandatory arguments but we can't change the argument order either. I don't have a better proposal

```
Cannot autowire service "Foo": argument "$requestStack" of method "ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\SearchFilter::__construct()" has no type-hint, you should configure its value explicitly.
```

meyerbaptiste:
Ahah, yes I agree with that and your patch regarding the autowiring. Maybe the DependencyInjection component should inject `null` in this case, but that's another debate and it's not ours. 😄

meyerbaptiste:
My bad, thx @antograssiot!

teohhanhui:
Now it feels like we're writing code for the autowiring lol!

teohhanhui:
Do we have a way (`_bind` or something) to let the autowiring know what to inject, without doing this weird hack in our code?

PR #2646, opened 2019-03-25
Ensure unsuccessful responses are not cached

bpolaszek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Following our conversation on Slack, this PR aims at avoiding Api-Platform to generate cache headers on 4xx / 5xx responses.


dunglas:
Thanks @bpolaszek, good catch!

PR #2647, opened 2019-03-25
GraphQl: parse input&messenger metadata

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

We forgot to take graphql mutations into consideration when retrieving the metadata. This now works correctly when performing a `createBookmark` mutation on my resource:

```
/**
 * @ApiResource(
 *      graphql={"create"={"input"=BookmarkInput::class, "messenger"="input"}}
 * )
 */
class Bookmark {}
``` 

It'll forward `BookmarkInput` to the messenger via the mutation :fire: :fire:  :fire:  . 

Note: for naming consistency I took the liberty of changing `graphql_operation_name` to `graphql_mutation_name` in our internal `$context` because else it's confusing.

### DIFF
--- a/features/main/input_output.feature
+++ b/features/main/input_output.feature
@@ -287,3 +287,29 @@ Feature: DTO input and output
       "data": 123
     }
     """
+
+  @!mongodb
+  Scenario: Use messenger with graphql and an input where the handler gives a synchronous result
+    When I send the following GraphQL request:
+    """
+    mutation {
+      createMessengerWithInput(input: {var: "test"}) {
+        messengerWithInput { id, name }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the JSON should be equal to:
+    """
+    {
+      "data": {
+        "createMessengerWithInput": {
+          "messengerWithInput": {
+              "id": "/messenger_with_inputs/1",
+              "name": "test"
+          }
+        }
+      }
+    }
+    """
--- a/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
@@ -94,19 +94,27 @@ final class ItemMutationResolverFactory implements ResolverFactoryInterface
 
             $this->canAccess($this->resourceAccessChecker, $resourceMetadata, $resourceClass, $info, $item, $operationName);
 
-            $inputMetadata = $resourceMetadata->getAttribute('input', ['class' => $resourceClass]);
-            if (null === $resourceClass = $inputMetadata['class'] ?? null) {
-                return $data;
+            $inputMetadata = $resourceMetadata->getGraphqlAttribute($operationName, 'input', null, true);
+            $inputClass = null;
+            if (\is_array($inputMetadata) && \array_key_exists('class', $inputMetadata)) {
+                if (null === $inputMetadata['class']) {
+                    return $data;
+                }
+
+                $inputClass = $inputMetadata['class'];
             }
 
             switch ($operationName) {
                 case 'create':
                 case 'update':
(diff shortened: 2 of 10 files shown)


features/main/input_output.feature
```diff
@@ -287,3 +287,30 @@ Feature: DTO input and output
       "data": 123
     }
     """
+
+  @!mongodb
```
alanpoulain: Why?

features/main/input_output.feature
```diff
@@ -287,3 +287,30 @@ Feature: DTO input and output
...
+  Scenario: Use messenger with graphql and an input where the handler gives a synchronous result
+    When I send the following GraphQL request:
+    """
+    mutation {
+      createMessengerWithInput(input: {var: "test"}) {
+        clientMutationId, messengerWithInput { id, name }
```
alanpoulain: Line break?

features/main/input_output.feature
```diff
@@ -287,3 +287,30 @@ Feature: DTO input and output
...
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the JSON should be equal to:
+    """
+    {
+      "data": {
+        "createMessengerWithInput": {
+          "clientMutationId": null,
```
alanpoulain: You can remove it.

src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
```diff
@@ -65,9 +65,9 @@ public function __construct(IriConverterInterface $iriConverter, DataPersisterIn
         $this->validator = $validator;
     }
 
-    public function __invoke(string $resourceClass = null, string $rootClass = null, string $operationName = null): callable
```
dunglas: I would revert this change. This is triggered for queries to, and queries aren't mutations. Mutations + Queries = operations.

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -419,7 +419,7 @@ private function getResourceObjectType(?string $resourceClass, ResourceMetadata
             return $this->graphqlTypes[$shortName];
         }
 
-        $ioMetadata = $resourceMetadata->getAttribute($input ? 'input' : 'output');
+        $ioMetadata = null === $mutationName ? $resourceMetadata->getAttribute($input ? 'input' : 'output') : $resourceMetadata->getGraphqlAttribute($mutationName, $input ? 'input' : 'output', null, true);
```
alanpoulain: Can't we have a specific output for a query?

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -492,7 +492,7 @@ private function getResourceObjectTypeFields(?string $resourceClass, ResourceMet
 
         if (null !== $resourceClass) {
             foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $property) {
-                $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $property, ['graphql_operation_name' => $mutationName ?? 'query']);
+                $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $property, ['graphql_mutation_name' => $mutationName ?? 'query']);
```
alanpoulain: It seems strange to use `graphql_mutation_name` when it's a query. And there will be custom queries too.

features/main/input_output.feature
soyuka: Because I don't want to create a document for this and also because this entity isn't even tight to a database persister so there's no point of running it with mongodb.

alanpoulain: Yes but I think it would be better to have it working with MongoDB anyway. This tag is only here when the functionality is not compatible with MongoDB.

Issue #2644, opened 2019-03-25
Cache not updated when entity annotation is changed


alexandresalome:
This morning, I started a new project with API Platform v2.4, and almost from scratch, I had to clear the cache to get my API documentation updated.

### Steps to reproduce

Create a new Symfony application api-platform-issue `symfony new --full api-platform-issue`.

Go to this new project (`cd api-platform-issue`).

Install API platform by using `composer req api`.

Install embedded web server `composer req server` and start it `bin/console server:run`.

Open API doc: http://localhost:8000/api

Create an entity Pizza:

```php
<?php

namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ApiResource
 * @ORM\Entity
 */
class Pizza
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     *
     * @var int
     */
    private $id;

    public function getId(): ?int
    {
        return $this->id;
    }
}
```

Refresh your API doc, you don't see any change.

Delete your cache (`rm -rf var/cache`).

API is updated.

### Further steps

If you delete the annotation and update cache, you'll see no change.

### Environment details

Bug reproduced:

- With a docker environment (PHP 7.3)
- With my local PHP setup (PHP 7.2)

dunglas:
Probably related to https://github.com/api-platform/core/pull/2629... @teohhanhui can you take a look? Maybe should we revert this patch until it is properly tested and "edge cases" (this one is a basic case) are handled properly.

soyuka:
This is more related to #2593 and it's not easy to take a decision that will not have an impact.

dunglas:
I prefer to favor DX than performance in dev env.

teohhanhui:
Symfony uses cache in `dev` environment out of the box. We should do the same. If we require more manual `cache:clear`, then that's something we should fix.

teohhanhui:
Okay, I was mistaken. Symfony does not use cache in `dev` environment out of the box:

Validator: https://github.com/symfony/symfony/blob/v4.2.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1134-L1136

Serializer: https://github.com/symfony/symfony/blob/v4.2.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1376-L1388

[...]

PR #2648, opened 2019-03-25
Revert "Do not disable cache in dev"

teohhanhui:
Reverts api-platform/core#2629

See discussion in #2644


soyuka:
I'm :+1: to keep this patch though (so :-1: on the revert :p)

teohhanhui:
Hmm... See https://github.com/api-platform/core/issues/2644#issuecomment-476193238

soyuka:
Yeah I saw the reasoning behind #2644 :). let's favor DX then but I still like this parameter being gone :P

dunglas:
It would be nice to improve perf in dev, but it must not introduce DX regressions. It's probably possible to have a better dev cache strategy, in another PR.

AristideHerveMbassi:
Please reopen because I have the same problem

PR #2655, opened 2019-03-26
[GraphQL] Manage custom queries by using the internal resolvers

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Following @SCIF comment: https://github.com/api-platform/core/pull/2445#issuecomment-476005587

The custom queries are now always using the internal resolvers in order to avoid security issues and to let API Platform handle the normalization, pagination, etc.

@lukasluecke WDYT of it?


lukasluecke:
Makes sense 👍

src/GraphQl/Resolver/Factory/ItemResolverFactory.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function __invoke(string $resourceClass = null, string $rootClass = null, string $operationName = null): callable
+    {
+        return function ($source, $args, $context, ResolveInfo $info) use ($operationName) {
+            // Data already fetched and normalized (field or nested resource)
+            if (isset($source[$info->fieldName])) {
+                return $source[$info->fieldName];
+            }
+
+            if (!isset($args['id'])) {
```
SCIF: I think this check required only for bundled resolver. Initially, one of the reasons of introduction of custom operations was  an idea to create endpoints/resources without compulsory arguments. Probably `iri` should be grabbed from custom operation configuration in case of unavailability of `id`.

lukasluecke: I agree. I think moving normalization to the factory makes sense, but the retrieval of items should be handled by the resolvers.

`iri` should not be needed either, and the `$item` parameter should be removed from `QueryItemResolverInterface`. Maybe we can pass the `iriConverter` and `baseNormalizationContext`, or some callable preconfigured with those, to the resolver - so you can actually load items inside of it?

alanpoulain: I think there will be three ways to add custom operations in the future:
- custom queries (this one). `id` arg is mandatory, the custom resolver receives the item/collection, it does some things, it returns it.
- custom mutations: an `input` is mandatory, the custom resolver receives the denormalized item, it does some things, it returns it and it will be persisted afterwards.
- custom operations: nothing is mandatory for the args, the resolver does anything it wants, nothing is done if it returns something (it needs to do the normalization itself). I'm not sure about this one. And IMO it can be strange to define it on a resource since you can do anything you want.
WDYT?

lukasluecke: My example (and currently implemented use-case) for custom queries was a "currentUser" query (returns logged in user based on token storage). Having an `id` arg does not really make sense here, but using the resource normalization does.

alanpoulain: OK. I will do the modifications.

lukasluecke: Thank you! 🙂🚀

alanpoulain:
Should be good now :slightly_smiling_face:

lukasluecke:
Looks good to me 🙂

src/GraphQl/Resolver/Factory/ItemResolverFactory.php
SCIF: What's the use-case for such custom queries (having `id`) then? It could be done by specifying `@ApiResource({grapqhl={"query"}})`. My comment mentioned in the starting message of PR is about resources free from `id`.

alanpoulain: Custom queries having an `id` argument could be useful to modify the received item or to add side effects to your query.
WIth what I've done, you have an `id` argument by default and you receive the item by default but you can easily change this behavior.

lukasluecke: I believe having the optional `id` argument should be fine for now 👍 If you don't need / use it, just ignore the item argument inside your resolver.

SCIF:
Hi @alanpoulain , @lukasluecke , correct me if I'm wrong please. This PR brakes custom operations at all.
Does anybody use custom operations using latest master? It's just broken :( Specifying of valid `id` is required, otherwise `IriConverter::getItemFromIri()` throws NOT `ItemNotFoundException`, but `InvalidArgumentException`.
I really disappointed since invested a lot of time into dropping my own Schema and Resolver and investigating how to use out of box ones. And they are not usable atm :(

lukasluecke:
@SCIF Take a look at #2738 for a quick patch, which I‘m using for my custom mutations. Didn‘t get around to adding tests yet, so it will probably get merged next week - but it work‘s fine for me in production.

SCIF:
Hi @lukasluecke , your patch is aimed for SchemaBuilder while I'm talking about `ItemResolverFactory` bug.

alanpoulain:
There is a working Behat test for this: https://github.com/api-platform/core/blob/bc0806be383a6fc6820a1230bbddd3ea29542ac7/features/graphql/query.feature#L287-L308
As you can see there is no `id` involved in this query.
Make sure the args are set like this in the entity: https://github.com/api-platform/core/blob/bc0806be383a6fc6820a1230bbddd3ea29542ac7/tests/Fixtures/TestBundle/Entity/DummyCustomQuery.php#L28-L31
This will be documented once I've done a required PR (possibility to manage types in args).

SCIF:
Hi @alanpoulain , my apogolies, looks like `"args"={}` is required and I missed that.

SCIF:
Don't know why tests are not crashed but I found [these changes](https://github.com/api-platform/core/commit/2bb4db1583521be72144d1da7da9cbe0b09808a2#diff-3e45e8e6ce08f8a7cbbbda1601d5ff29L121) breaking process of normalizing of id-less resolvers. Will do some more ingestigation later if nobody has an idea why iri is compulsory for now (and why tests are not broken).
@teohhanhui , any clue?

SCIF:
Heh. I found why tests are not crashed — because of ORM nature of resources. Resources not mapped to ORM Entites/Documents crash APIP with message of inability to generate IRI.

Issue #1096, opened 2017-05-04
Filter Only The Serializable Fields


ad3n:
Api Platform is the best API Framework Ever is used. It's great framework.

Recently, when we activate the filter feature without specific fields defined, for example
```    bisnis.filter.order_filter:
        parent: 'api_platform.doctrine.orm.order_filter'
        arguments: [ ~ ]
        tags:
            - { name: 'api_platform.filter', id: 'order.filter' }```

The Api Platform (see Api Doc) processing all fields based on doctrine metadata even the fields is not displayed. Is good choice to filter only the fields that displayed by serialization not all fields.

Reference: https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php#L84

PR #2676, opened 2019-04-01
Fix: don't filter unreadable properties

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #1096
| License       | MIT
| Doc PR        | n/a

* [x] Doctrine ORM
* [x] Doctrine ODM
* [x] Elasticsearch  
* [ ] Tests

### DIFF
--- a/src/Metadata/Property/PropertyMetadataFactoryOptionsTrait.php
+++ b/src/Metadata/Property/PropertyMetadataFactoryOptionsTrait.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Property;
+
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
+
+trait PropertyMetadataFactoryOptionsTrait
+{
+    /**
+     * Gets a valid context for property metadata factories.
+     *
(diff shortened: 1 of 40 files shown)


src/Metadata/Property/PropertyMetadataFactoryOptionsTrait.php
```diff
@@ -0,0 +1,43 @@
...
+
+        if (isset($context[AbstractNormalizer::GROUPS])) {
+            $options['serializer_groups'] = $context[AbstractNormalizer::GROUPS];
+        }
+
+        if (isset($context['collection_operation_name'])) {
+            $options['collection_operation_name'] = $context['collection_operation_name'];
+        }
+
+        if (isset($context['item_operation_name'])) {
```
teohhanhui: It's could only be one of collection operation or item operation, so I think we could use `elseif` here?

meyerbaptiste: Yes, it was just a copy/paste from `AbstractItemNormalizer::getFactoryOptions()` but I can change it.

teohhanhui: I think we should also rename this method to something more descriptive: `getPropertyMetadataFactoryOptions`? But of course rename it in the `use` statement of `AbstractItemNormalizer` for BC.

meyerbaptiste:
Todo: https://github.com/api-platform/core/pull/3503

VincentLanglet:
Hi, the 2.x version is not maintained anymore ; could you update your PR to target 4.2 instead ?
Thanks

PR #2679, opened 2019-04-02
Normalize non-resource objects in a standalone normalizer

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | only for a new change introduced in 2.4
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/1085
| License       | MIT
| Doc PR        | N/A

Note: Includes a minor BC break because we undo a new change introduced in 2.4. I think it's okay because `$handleNonResource` was internal anyway. The only people affected would be those who have extended `AbstractItemNormalizer` and explicitly pass that optional parameter (which, actually, still wouldn't result in an error anyway).

TODO:
- [x] Add regression test for https://github.com/api-platform/api-platform/issues/1085
- [x] Add tests for all formats
- [ ] ~Handle circular reference across resource / non-resource?~ Don't care for now


features/main/input_output.feature
```diff
@@ -179,7 +179,11 @@ Feature: DTO input and output
       },
       "@type": "User",
       "@id": "/users/1",
-      "dummy": "/dummies/1"
+      "dummy": {
```
teohhanhui: This should probably be handled better?

soyuka: It'd be better to have `@id` as a string no?

bastnic:
seems cleaner (and better performance, I hope)

toriqo:
when using `application/vnd.api+json` or `application/hal` non-resources are properly serialized and returned based on contexts

soyuka:
@teohhanhui my test case also covers @toriqo issue, should I duplicate it to make it run with other formats?

teohhanhui:
@soyuka I'll take care of that, since I've been working on something since yesterday.

.travis.yml
```diff
@@ -68,13 +68,13 @@ script:
     fi
   - tests/Fixtures/app/console cache:clear
   - if [[ $APP_ENV = 'postgres' ]]; then
-      vendor/bin/behat --suite=postgres --format=progress;
+      vendor/bin/behat --suite=postgres --format=progress --no-interaction;
```
soyuka: why?

bastnic:
> I like it! Really good to have removed this `handleNonResource` parameter and I feel that the code is a step forward to simplify composition! Good job!

+100

.travis.yml
teohhanhui: Undefined step `Then this is not supported`, and because `behat` CLI can't tell correctly when it's in a non-interactive context. :see_no_evil:

soyuka:
Thanks @teohhanhui

PR #2685, opened 2019-04-04
Metadata: Attributes default value should be null not an empty array

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

In every other metadata factory, we check if attributes is `null`. The AnnotationResourceMetadataFactory is one of the first to run and sets the attributes to an empty array. This then blocks other metadata factories, and they will not set the `attributes` value because it's not null.


teohhanhui:
I think the bug is real, but the fix is wrong. The default value for `attributes` property should be changed to `null` in the annotation.

PR #2688, opened 2019-04-04
@type

toriqo:
return the type so a client subscribed to more than one topics will know which resource type was deleted without having to parse the @id

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT


toriqo:
ping @dunglas @soyuka @antograssiot

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -150,6 +150,7 @@ private function storeEntityToPublish($entity, string $property): void
             $this->deletedEntities[(object) [
                 'id' => $this->iriConverter->getIriFromItem($entity),
                 'iri' => $this->iriConverter->getIriFromItem($entity, UrlGeneratorInterface::ABS_URL),
+                'type' => $this->resourceMetadataFactory->create($this->getObjectClass($entity))->getShortName(),
```
antograssiot: `$this->getObjectClass($entity)` is already known as `$resourceClass` line 123
And the metadata were also already created line 128 and can be re-used I guess

toriqo: good catch :)

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -150,6 +152,7 @@ private function storeEntityToPublish($entity, string $property): void
             $this->deletedEntities[(object) [
                 'id' => $this->iriConverter->getIriFromItem($entity),
                 'iri' => $this->iriConverter->getIriFromItem($entity, UrlGeneratorInterface::ABS_URL),
+                'type' => $metadata->getShortName(),
```
dunglas: The IRI must be used if defined: https://github.com/api-platform/core/blob/master/src/JsonLd/Serializer/ItemNormalizer.php#L89 (could use the coalesce operator now, by the way).

soyuka:
needs to target master

norkunas:
This could be useful.. @toriqo are you up to finish this?

bpolaszek:
Just saying: the only way, client side, to know that a Mercure update is a removal instead of an update, is that you only have the `@id` property in the payload, right?

If so, why not considering this as a BC break?

(I 100% agree with adding the `@type`, that's not my point)

divine:
Hello,

I do really hope this will be a priority before the next major release. It's a needed feature that has been delayed for a few years.

@dunglas @soyuka does it needs only rebasing or something is left? Should I try to take care of it?

Thanks!

dunglas:
This just needs a rebase, and a decision about `@id`. I also fear that it's a BC break. I would suggest not setting it, or at least using a configuration option such as `include_id` under the `mercure` key.

bpolaszek:
> This just needs a rebase, and a decision about `@id`. I also fear that it's a BC break. I would suggest not setting it, or at least using a configuration option such as `include_id` under the `mercure` key.

You mean `@type` and `include_type`, right?

dunglas:
Indeed.

bpolaszek:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

Let's add a coin in the juke box - wasn't it just a minor change?

alanpoulain:
I've picked this one. WDYT @bpolaszek?

alanpoulain:
Let's merge it then. It will be available in 3.1. Thanks everyone.

PR #2691, opened 2019-04-04
GraphQl: Use compiler pass only when enabled

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na

Compiler passes ran with no graphql which was producing errors.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlMutationResolverPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlMutationResolverPass.php
@@ -31,6 +31,10 @@ final class GraphQlMutationResolverPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
+        if (!$container->getParameter('api_platform.graphql.enabled')) {
+            return;
+        }
+
         $mutations = [];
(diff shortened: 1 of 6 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/GraphQlMutationResolverPass.php
```diff
@@ -31,6 +31,10 @@ final class GraphQlMutationResolverPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
+        if (!$container->getParameter('api_platform.graphql.enabled')) {
```
vincentchalamon: You should move this check in [ApiPlatformBundle::build](https://github.com/api-platform/core/blob/be932e901402000adf54fa38e0b741f36474be79/src/Bridge/Symfony/Bundle/ApiPlatformBundle.php#L48), so the CompilerPass won't be executed (same with other GraphQl CompilerPasses)

soyuka: thanks @vincentchalamon I updated the PR, I did the same thing for elasticsearch.

teohhanhui:
@soyuka I think you merged the wrong commit. :laughing:

soyuka:
Yes parameters are not ready there.

Issue #2692, opened 2019-04-05
WriteListener should use $persistResult if available?


Tjeerd:
Hi, I'm getting the `"Unable to generate an IRI for the item of type.."` error when using a custom DataPersister after sending a POST request. Reason being that the `WriteListener` is using the `$controllerResult` instead of `$persistResult` returned by the DataPersister.

In my specific case the "id" of the resource is generated in the DataPersister, the `$controllerResult` is not aware of this ID and therefore the `iriConverter` is unable to generate a valid IRI. In my case it would make sense to use the `$persistResult`, if available, to determine the IRI of the persisted resource. I might be missing other usecases though?

mikelerch:
In the `WriteListener`, the current logic is
```
if ($hasOutput) {
    $request->attributes->set('_api_write_item_iri', $this->iriConverter>getIriFromItem($controllerResult));
 }
```
As described above, the `$controllerResult` contains an object which does not have an `ID` yet. This is set afterwards in the `DataPerister`.

The `WriteListener` does update the controller result correctly with the object returned by the `DataPersister`.
```
$event->setControllerResult($persistResult ?? $controllerResult);
```

So when generating the IRI, the new controller result should be used instead.

```
$request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($event->getControllerResult()));
```

Tjeerd:
Fixed by #2693

slavo2:
Small note: As the fix is not yet release, one might need to use 
` "api-platform/core": "2.4.x-dev"` as composer requirement constraint.

PR #2693, opened 2019-04-05
Make sure to always use the persistResult for follow-up actions

Tjeerd:
Not using the persistResult could cause issues in generating the IRI for
new resources because the ID was not set on the controllerResult but
only on the persistResult for example.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | not sure?
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2692 
| License       | MIT
| Doc PR        | not relevant?

### DIFF
--- a/tests/EventListener/WriteListenerTest.php
+++ b/tests/EventListener/WriteListenerTest.php
@@ -95,6 +95,7 @@ class WriteListenerTest extends TestCase
 
     /**
      * @see https://github.com/api-platform/core/issues/1799
+     * @see https://github.com/api-platform/core/issues/2692
      */
     public function testOnKernelViewWithControllerResultAndPersistWithImmutableResource()
     {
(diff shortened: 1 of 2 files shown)


soyuka:
May you add a regression test with your example?

Tjeerd:
> May you add a regression test with your example?

I feel this already is, sort of, covered by the test I updated, but I will try to add a new test.

Tjeerd:
Ok, I've updated the test to have an assertion to test for any regression issues of the previous behaviour.

tests/EventListener/WriteListenerTest.php
```diff
@@ -108,7 +108,8 @@ public function testOnKernelViewWithControllerResultAndPersistWithImmutableResou
         $dataPersisterProphecy->supports($dummy, Argument::type('array'))->willReturn(true)->shouldBeCalled();
 
         $iriConverterProphecy = $this->prophesize(IriConverterInterface::class);
-        $iriConverterProphecy->getIriFromItem($dummy)->willReturn('/dummy/1')->shouldBeCalled();
+        $iriConverterProphecy->getIriFromItem($dummy)->shouldNotBeCalled();
+        $iriConverterProphecy->getIriFromItem($dummy2)->willReturn('/dummy/1')->shouldBeCalled();
```
teohhanhui: Do an assert on the `Request` attributes instead. :smile:

Tjeerd: I'm sorry, I don't follow, the above assertion checks whether or not the right call is done on the IriConverter. Since the same Iri is returned the current assertion on the request is still valid.

soyuka: Yes, what @teohhanhui means is that it'd be nice to check that `$request->attributes->set('_api_write_item_iri` has the expected IRI (https://github.com/api-platform/core/blob/3f1f23fcdabd695f2d679300f41f79c8e4d66ee8/src/EventListener/WriteListener.php#L81) because this was the root issue

Tjeerd: Yes, I agree, but that's already asserted :) That didn't actually change because Prophecy is returning the same IRI but only for another Dummy (`dummy2`) resource.

teohhanhui: `$dummy` and `$dummy2` should have different IRI for an effective test.

teohhanhui: @Tjeerd I'll take care of the test. :smile:

teohhanhui:
Thanks @Tjeerd! :tada:

PR #2697, opened 2019-04-06
fix (cache): prevent avoidable purge

GregoireHebert:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ~
| License       | MIT
| Doc PR        | ~

#EUFOSSA

Since we can get rid of get operations, we should not trigger the cache for these.
Moreover we should not call purge method if there is nothing to purge.


soyuka:
THANKS @GregoireHebert

PR #2054, opened 2018-06-28
Check that strict types work as expected

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Demonstrate a failing test case.

/cc @dunglas

PR #2699, opened 2019-04-06
Added an additional test for strict types validation

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/2054
| License       | MIT
| Doc PR        |

Apparently this has been fixed in the meantime. I still added the test and fixed it (`gettype()` returns `null` uppercase).

### DIFF
--- a/features/security/strong_typing.feature
+++ b/features/security/strong_typing.feature
@@ -56,7 +56,7 @@ Feature: Handle properly invalid data submitted to the API
     And the JSON node "@context" should be equal to "/contexts/Error"
     And the JSON node "@type" should be equal to "hydra:Error"
     And the JSON node "hydra:title" should be equal to "An error occurred"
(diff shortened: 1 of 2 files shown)


Toflar:
Done at `EUFOSSA Hackathon`. Needs the label :)

soyuka:
Thanks @Toflar

Toflar:
> Do you have an idea when it has been done?

I tried to track it down and it looks like it has actually never failed 😄 Here's the original build: https://travis-ci.org/api-platform/core/jobs/397805821
As you can see, the only thing that failed was that the `null` is not equal to `NULL` but the validation was actually correct 😄

dunglas:
Some context about the initial PR: Han told me that strict types weren’t working with API Platform. I was sure that it was, so he opened a PR with a failing test 😂

PR #2705, opened 2019-04-07
Test input messenger with MongoDB

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Input messenger was not tested with MongoDB.

### DIFF
--- a/features/graphql/input_output.feature
+++ b/features/graphql/input_output.feature
@@ -1,10 +1,10 @@
 Feature: GraphQL DTO input and output
-  In order to use a hypermedia API
+  In order to use the GraphQL API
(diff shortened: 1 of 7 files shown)


soyuka:
Maybe we could update CONTRIBUTING.md with instructions about adding tests (duplicate entities) on mongodb and how to run them `APP_ENV=mongodb`.

Also, as ODM annotations aren't the same as ORM ones, maybe we could add matches between the two?

PR #751, opened 2016-09-12
Makes url generation strategy default value configurable

maechler:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | https://github.com/api-platform/api-platform/issues/128 |
| License | MIT |

PR #2708, opened 2019-04-07
WIP - Makes url generation strategy default value configurable

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | to be done

Superseed #751 
This is a draft for discussion.
I'll open a doc PR if we agree on this before merging this.

### DIFF
--- a/src/Api/UrlGeneratorInterface.php
+++ b/src/Api/UrlGeneratorInterface.php
@@ -34,7 +34,7 @@ interface UrlGeneratorInterface
     /**
      * Allow to generate url using the globally configured strategy.
      */
-    public const DEFAULT = -1;
+    public const DEFAULT_STRATEGY = -1;
 
     /**
      * Generates an absolute URL, e.g. "http://example.com/dir/file".
--- a/src/Api/UrlGeneratorInterface.php
+++ b/src/Api/UrlGeneratorInterface.php
@@ -32,6 +32,11 @@ namespace ApiPlatform\Core\Api;
 interface UrlGeneratorInterface
 {
     /**
+     * Allow to generate url using the globally configured strategy.
+     */
+    public const DEFAULT = -1;
+
+    /**
      * Generates an absolute URL, e.g. "http://example.com/dir/file".
      */
     public const ABS_URL = 0;
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection;
 
+use ApiPlatform\Core\Api\UrlGeneratorInterface;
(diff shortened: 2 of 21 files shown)


src/Api/UrlGeneratorInterface.php
```diff
@@ -31,6 +31,11 @@
  */
 interface UrlGeneratorInterface
 {
+    /**
+     * Allow to generate url using the globally configured strategy.
+     */
+    public const DEFAULT = -1;
```
dunglas: Maybe would it be more consistent zither the rest of the API to use null instead? Not really sure, just asking.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -250,7 +251,10 @@ public function getConfigTreeBuilder()
                 ->arrayNode('messenger')
                     ->{interface_exists(MessageBusInterface::class) ? 'canBeDisabled' : 'canBeEnabled'}()
                 ->end()
-
+                ->integerNode('routing_strategy_type')
```
dunglas: url_generation_strategy?

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -126,16 +128,16 @@ public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface
             ), $e->getCode(), $e);
         }
 
-        return $this->getItemIriFromResourceClass($resourceClass, $identifiers, $referenceType);
+        return $this->getItemIriFromResourceClass($resourceClass, $identifiers, UrlGeneratorInterface::DEFAULT === $referenceType ? $this->routerStrategyType : $referenceType);
```
dunglas: By using null, you’ll also be able to simplify these check apps by using the coalesce operator.

src/Api/UrlGeneratorInterface.php
antograssiot: That would work and simplify the the conditional statements indeed as pointed in your other comment.
the only downside is that I'll have to silence PHPStan errors because the [IriConverterInterface](https://github.com/api-platform/core/blob/90f422faa56beb3afa51aea75053c5a9333886dc/src/Api/IriConverterInterface.php#L45) and the [ContextBuilderInterface](https://github.com/api-platform/core/blob/90f422faa56beb3afa51aea75053c5a9333886dc/src/JsonLd/ContextBuilderInterface.php) expects int and will be passed `null`

Do you think it is worth it ?

soyuka: And we can't use `0` I'd be to keep `-1`!

teohhanhui: We must respect the contract of the interface.

teohhanhui: Anyway, `UrlGeneratorInterface::DEFAULT` is not a good name for a constant. I'd say it should be `DEFAULT_STRATEGY`?

antograssiot:
superseeded by #3084

PR #2709, opened 2019-04-07
Use the metadata aware name converter when available fix #2677

### LINKED ISSUE #2677: MetadataAwareNameConverter is not used unless serializer.name_converter is set in SF.
Hey guys,

Not sure if this is the right place but the added code in `MetadataAwareNameConverterPass` ( https://github.com/api-platform/core/pull/2562 ) seems wrong to me.

Let me explain:

```
if (1 >= \count($definition->getArguments()) || null === $definition->getArgument(1)) {
       return;
 }
```
This checks for the presence of a second argument and that it's not null. The second argument of `MetadataAwareNameConverter` is the `$fallbackConverter` but the thing is, this `$fallbackConverter` is only set if you define `serializer.name_converter` in `framework.yaml`.

What this mean is, API Platform will not auto-use the `MetadataAwareNameConverter` if you don't define `serializer.name_converter`.
I was expecting it to use it regardless and lost a lot of time finding out why.

Also, I wanted to use both the `MetadataAwareNameConverter` and snake_case but if I follow the docs and set `name_converter: 'Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter'` in `api_platform.yaml` then I lose the `@SerializedName` behavior.

I think that ApiPlatform should **always** use `MetadataAwareNameConverter` (if available, of course) and use the value of `api_platform.name_converter` to set $fallbackConverter (or set it as the main NameConverter if `MetadataAwareNameConverter` is not available).

This behavior is kinda redundant with the default symfony one with `serializer.name_converter` BUT it allows API-Platform to stay free of a dependency with Symfony (which is what you guys wanted, as per https://github.com/api-platform/core/pull/2523#discussion_r257995961 ).

Cheers.

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2677
| License       | MIT
| Doc PR        | na

ping @ogirardkaliop @teohhanhui

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPass.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
 use ApiPlatform\Core\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * Injects the metadata aware name converter if available.
@@ -33,14 +34,20 @@ final class MetadataAwareNameConverterPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        if ($container->hasAlias('api_platform.name_converter') || !$container->hasDefinition('serializer.name_converter.metadata_aware')) {
+        if (!$container->hasDefinition('serializer.name_converter.metadata_aware')) {
             return;
         }
 
         $definition = $container->getDefinition('serializer.name_converter.metadata_aware');
+        $num = \count($definition->getArguments());
 
-        if (1 >= \count($definition->getArguments()) || null === $definition->getArgument(1)) {
-            return;
+        if ($container->hasAlias('api_platform.name_converter')) {
+            $nameConverter = new Reference((string) $container->getAlias('api_platform.name_converter'));
+            if (1 === $num) {
+                $definition->addArgument($nameConverter);
+            } elseif (1 < $num && null === $definition->getArgument(1)) {
+                $definition->setArgument(1, $nameConverter);
+            }
         }
 
         $container->setAlias('api_platform.name_converter', 'serializer.name_converter.metadata_aware');
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "symfony/http-kernel": "^3.4 || ^4.0",
         "symfony/property-access": "^3.4 || ^4.0",
         "symfony/property-info": "^3.4 || ^4.0",
-        "symfony/serializer": "^4.2",
+        "symfony/serializer": "^4.2.6",
         "symfony/web-link": "^4.1",
         "willdurand/negotiation": "^2.0.3"
     },
--- a/tests/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPassTest.php
+++ b/tests/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPassTest.php
@@ -15,6 +15,8 @@ namespace ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\DependencyInjection\Compi
 
 use ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler\MetadataAwareNameConverterPass;
 use PHPUnit\Framework\TestCase;
+use Prophecy\Argument;
+use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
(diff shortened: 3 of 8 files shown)


soyuka:
We have an issue now because of the following line https://github.com/api-platform/core/blob/master/src/Serializer/AbstractItemNormalizer.php#L322-L324 (ping @dunglas )

With this new `MetadataAwareNameConverter`, the attribute gets converted from `0` to `"0"` and then it fails. This is due to the fact that the signature of `denormalizeFallback` is:

```
    private function denormalizeFallback(string $propertyName, string $class = null, string $format = null, array $context = []): string
```

which transforms my `0` to `"0"`.

End of story, and of the EU-FOSSA Hackathon is that we need to fix this in symfony.

ogirardkaliop:
Thanks for the effort. Did you open an issue with Symfony to get this resolved ?

soyuka:
@ogirardkaliop not yet it was really at the end of the Hackathon that I figured that out haha, I'm going to open a PR in symfony asap!

src/Bridge/Symfony/Bundle/Resources/config/elasticsearch.xml
```diff
@@ -45,7 +45,6 @@
         <service id="ApiPlatform\Core\Bridge\Elasticsearch\Api\IdentifierExtractorInterface" alias="api_platform.elasticsearch.identifier_extractor" />
 
         <service id="api_platform.elasticsearch.name_converter.inner_fields" class="ApiPlatform\Core\Bridge\Elasticsearch\Serializer\NameConverter\InnerFieldsNameConverter" public="false">
-            <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
```
soyuka: @meyerbaptiste if I don't do this the default name converter doesn't convert query filters from Camel to snake and it fails. Do you have another solution in mind?

ogirardkaliop: You set the default "fallbackNameConverter" of the MetadataNameConverter to be "api_platform.name_converter" by default.

If people choose CamelCaseToSnakeCase then it will work fine.

soyuka: Yes it's now working properly as stated in #2677:

If MetadataAwareNameConverter it's used
If `api_platform.name_converter` exists it's used as fallback
If symfony's `name_converter` is configured it'll be used as fallback
If none is given, the `MetadataAwareNameConverter` will not do any conversion

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -117,6 +117,10 @@ public function load(array $configs, ContainerBuilder $container): void
             ->addTag('api_platform.subresource_data_provider');
         $container->registerForAutoconfiguration(FilterInterface::class)
             ->addTag('api_platform.filter');
+
+        if ($config['name_converter']) {
```
teohhanhui: Should be moved into `registerCommonConfiguration` method.

teohhanhui: Oh, in fact the code is already there.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPass.php
```diff
@@ -33,14 +34,20 @@ final class MetadataAwareNameConverterPass implements CompilerPassInterface
...
             return;
         }
 
         $definition = $container->getDefinition('serializer.name_converter.metadata_aware');
-
-        if (1 >= \count($definition->getArguments()) || null === $definition->getArgument(1)) {
-            return;
+        $num = \count($definition->getArguments());
+
+        if ($container->hasAlias('api_platform.name_converter')) {
```
teohhanhui:
```suggestion
        if ($container->has('api_platform.name_converter')) {
```

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPass.php
```diff
@@ -33,14 +34,20 @@ final class MetadataAwareNameConverterPass implements CompilerPassInterface
...
+            $nameConverter = new Reference((string) $container->getAlias('api_platform.name_converter'));
```
teohhanhui:
```suggestion
            $nameConverter = new Reference('api_platform.name_converter');
```

(Not sure if possible to create a `Reference` using an alias id.)

teohhanhui:
See also: https://github.com/api-platform/core/pull/2523#discussion_r277050064

:stuck_out_tongue_closed_eyes:

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPass.php
soyuka: I have a circular dependency if I do this.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/MetadataAwareNameConverterPass.php
soyuka: we only care if it's an alias. If we want to setup a name converter, the user should setup the symfony's name converter within the serializer configuration.

Issue #1646, opened 2018-01-12
Deserializer before security


vincentchalamon:
I have an ApiResource with an access control to `ROLE_EMPLOYEE`. When I send anonymously a request (with an iri of a non-existing object) to create an object, I get a 400, but I should receive a 403.

Entity:
```php
/**
 * @ApiResource(attributes={
 *     "access_control"="is_granted('ROLE_EMPLOYEE')"
 * })
 * …
 */
class Shipping
{
    /**
     * @var null|PsOrder
     * …
     */
    private $order;
}
```

Request:
```json
{
    "order": "/ps_orders/1"
}
```

Response: `Item not found for "/ps_orders/1".`

Maybe it's just a question of priority with DenyAccessListener that should be higher than DeserializeListener.

@api-platform/core-team What status code would you expect in this case: 400 or 401/403?

norkunas:
Current priority allows to deny based on deserialized data so imho this works as expected.

vincentchalamon:
I've discussed that point with @dunglas, and yep this works as expected, as the `access_control` requires an object, so it needs to be deserialized first. But I still find it weird that I get a 400 instead of a 401/403

weaverryan:
#2779 was merged, which allows you to use the *original* data in your `access_control`.

That's great! However, I still think the original problem from @vincentchalamon is valid:

```
 * @ApiResource(attributes={
 *     "access_control"="is_granted('ROLE_EMPLOYEE')"
 * })
```

With this setup, my validation rules are executed and the client (who should not have access to this/these endpoints at all) gains the ability to see the validation rules for it.

dunglas:
Indeed. We could do the security check after deserialization, but before validation. WDYT?

antograssiot:
but that's already the case @dunglas 
https://github.com/api-platform/core/blob/125dec90d05dc383e9214cd4968428323d0a1136/src/Bridge/Symfony/Bundle/Resources/config/security.xml#L24

https://github.com/api-platform/core/blob/2f941fc3c0dc9dd257da3b34a6bb3f5c21c4230c/src/Bridge/Symfony/Bundle/Resources/config/validator.xml#L23

dunglas:
So I don’t understand how this issue appears 😄. I’ll have to create a reproducer.

soyuka:
Mhhh so this isn't fixed :sweat_smile: ?

vincentchalamon:
[...]

PR #2710, opened 2019-04-07
Apply DenyAccessListener before deserialization

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1646
| License       | MIT
| Doc PR        | 

I agree with @vincentchalamon that the priority is not the expected one.
The access control must be checked against the data read from the DB in the first place.
We have been discussing this during the EU-FOSSA Hackathon and couldn't really imagine a scenario where this would not be the expected behavior

ping @dunglas

### DIFF
--- a/features/authorization/deny.feature
+++ b/features/authorization/deny.feature
@@ -58,15 +58,39 @@ Feature: Authorization checking
     """
     Then the response status code should be 201
 
-  Scenario: An user cannot retrieve an item he doesn't own
+  Scenario: A user cannot retrieve an item he doesn't own
     When I add "Accept" header equal to "application/ld+json"
     And I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
     And I send a "GET" request to "/secured_dummies/1"
     Then the response status code should be 403
     And the response should be in JSON
 
-  Scenario: An user can retrieve an item he owns
+  Scenario: A user can retrieve an item he owns
     When I add "Accept" header equal to "application/ld+json"
     And I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
     And I send a "GET" request to "/secured_dummies/2"
+    Then the response status code should be 200
+
+  Scenario: A user can't assign him an item he doesn't own
+    When I add "Accept" header equal to "application/ld+json"
+    And I add "Content-Type" header equal to "application/ld+json"
+    And I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
+    And I send a "PUT" request to "/secured_dummies/2" with body:
+    """
+    {
+        "owner": "kitten"
+    }
+    """
+    Then the response status code should be 403
+
+  Scenario: A user can update an item he owns and transfer it
+    When I add "Accept" header equal to "application/ld+json"
+    And I add "Content-Type" header equal to "application/ld+json"
+    And I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    And I send a "PUT" request to "/secured_dummies/2" with body:
+    """
+    {
+        "owner": "vincent"
+    }
+    """
     Then the response status code should be 200
--- a/src/Bridge/Symfony/Bundle/Resources/config/security.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/security.xml
@@ -21,7 +21,7 @@
             <argument type="service" id="api_platform.security.resource_access_checker" />
 
             <!-- This listener must be executed only when the current object is available -->
-            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="1" />
+            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="3" />
         </service>
 
         <service id="api_platform.security.expression_language_provider" class="ApiPlatform\Core\Security\Core\Authorization\ExpressionLanguageProvider" public="false">
--- a/src/Security/EventListener/DenyAccessListener.php
+++ b/src/Security/EventListener/DenyAccessListener.php
@@ -50,8 +50,6 @@ final class DenyAccessListener
     }
 
     /**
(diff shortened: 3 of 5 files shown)


features/authorization/deny.feature
```diff
@@ -70,3 +70,27 @@ Feature: Authorization checking
     And I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
     And I send a "GET" request to "/secured_dummies/2"
     Then the response status code should be 200
+
+  Scenario: An user can't assign him an item he doesn't own
```
vincentchalamon:
```suggestion
  Scenario: A user can't assign himself on an item he doesn't own
```

features/authorization/deny.feature
```diff
@@ -70,3 +70,27 @@ Feature: Authorization checking
...
+    And I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
+    And I send a "PUT" request to "/secured_dummies/2" with body:
+    """
+    {
+        "owner": "kitten"
+    }
+    """
+    Then the response status code should be 403
+
+  Scenario: An user can update an item he owns and transfer it
```
vincentchalamon:
```suggestion
  Scenario: A user can update an item he owns and transfer it
```

soyuka:
Changing a priority is not a BC break.

vincentchalamon:
But as the behavior is changed by changing this priority, I think it's a BC break, as users will get a 403 instead of a 400 in some cases now.

soyuka:
> But as the behavior is changed by changing this priority, I think it's a BC break, as users will get a 403 instead of a 400 in some cases now.

But because the previous behavior was wrong it's a bug fix :p.

norkunas:
This will break my security voters :expressionless:

antograssiot:
@norkunas Can you please give us more insight on your use case ?

norkunas:
I am checking in security voters if a user can make an action based if he is a member of a company and has permissions for it.
```php
class CompanyUpgradeRequest {
  public $company;
}
```
```php
protected function voteOnAttribute($attribute, $subject, TokenInterface $token): bool {
    switch ($attribute) {
        case self::CREATE:
            return $this->companyPermissionChecker->hasPermission($token->getUser(), $subject->company, 'upgrade');
    }
}
```
But probably I'll rewrite these later with a validator constraint. Maybe I'm not only who depends on this now :thinking:

dunglas:
Indeed using a validation constraint for that would be cleaner. It’s not a code BC break, but it’s still a BC break. It needs at least to be documented in the changelog.

teohhanhui:
I think this is just shifting the problem instead of solving it. The application likely needs both before AND after states to be able to make an authorization decision, i.e. we need to know what the user is trying to do before we can say whether it should be allowed or not.

vincentchalamon:
>i.e. we need to know what the user is trying to do before we can say whether it should be allowed or not.

If you base on the request content, it should be validation instead of security. Otherwise, could you please provide an example to explain your point?

teohhanhui:
Well, I can only speak from my personal experience. I've been trying to disprove my point that validation cannot be separated from authorization. So far, I still don't see how.

1. Is this change valid? That depends on whether the user is authorized to perform this change.
2. Is the user authorized to perform this change? It depends on what change the user is trying to perform. (Is it a valid change given the roles the user have?)

vincentchalamon:
>It depends on whether the user is authorized to perform this change.

There are serialization groups for that. For example, if you want to restrict a field to specific users roles, you can decorate the ContextBuilder and dynamically change the serialization context (as explained [here](https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-dynamically)).

teohhanhui:
No, it's not something straightforward like that. Authorization can be as complex as the business logic requires. Same for validation. It's the interdependence between the two that's troublesome, and requires access to both "before" and "after" states.

teohhanhui:
Here is an example from my old API Platform-based (probably 2.1?) project: https://gist.github.com/teohhanhui/994ab0d0bab439302ef96ac080568a8f

vincentchalamon:
So this is a specific use case that can be handled by a customer listener/subscriber?

teohhanhui:
If `DenyAccessListener` is only meant to cover the most basic use cases, then sure. Anything non trivial will require custom listeners.

lukasluecke:
Maybe relevant: #2454 (I wanted to check permissions to create an item, based on its passed properties - e.g. only allow some role to create "root" items in a hierarchy of related objects)

soyuka:
Might be that we should have 2 listeners? One pre-normalization for authentification guards, one after for object related guards? But wait isn't the second one part of the item validation process? 
I don't think that there is a perfect solution for this use case, to move forward should we just document the solution with a custom listener?

antograssiot:
Another listener is indeed a solution. Deprecating this listener and replacing it by an event subscriber is also an option. WDYT?

teohhanhui:
> But wait isn't the second one part of the item validation process?

So, a validator that uses the `AuthrizationChecker`? It works for some cases, but not if you need both before / after states. At least, that's what I remember as the lesson I've learnt.

dunglas:
I tend to agree here. We should check on the existing data, and let the validator check the new one.
Also, we could add a new request attribute (`old-data`?) containing a `clone` of the initial data?

antograssiot:
> we could add a new request attribute (old-data?) containing a clone of the initial data, and make it available in expression language
It is also a good option indeed. Maybe harder to discover and it might be easy to fall in the trap of checking "again" against the wrong data.
The `old-data` could become a `old_object` variable in ExpressionLanguage, or maybe a `read-data` and `read_object` variable is a better naming @dunglas ?

dunglas:
I like `original_data`. It allows to check exactly what we want.

antograssiot:
closing in favor of #2779

PR #2712, opened 2019-04-07
Convert HTTP method to uppercase

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no, but prevent bug in client application
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |


soyuka:
Could you add a test?

composer.json
```diff
@@ -17,6 +17,7 @@
         "doctrine/inflector": "^1.0",
         "psr/cache": "^1.0",
         "psr/container": "^1.0",
+        "symfony/config": "^2.0 || ^3.0 || ^4.0",
```
teohhanhui:
```suggestion
        "symfony/config": "^3.4 || ^4.0",
```

I don't think we should support old unmaintained versions. It's just needlessly increasing our maintenance surface.

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -123,6 +124,13 @@ private function normalize(bool $collection, ResourceMetadata $resourceMetadata,
                 $supported ? $operation['method'] = $upperOperationName : $operation['route_name'] = $operationName;
             }
 
+            if (strtoupper($operation['method']) !== $operation['method']) {
+                $message = sprintf('The method "%s" of operation "%s" should be in uppercase.', $operation['method'], $operationName);
+
+                // See DelegatingLoader::load to understand why we use a LoaderLoadException
```
teohhanhui: Sorry, I think it's very bad for our factory to throw this exception. It's completely unexpected and will catch any consumer of this code off-guard. We should find a better way if we want to do this.

teohhanhui:
Can't we just convert the HTTP method to uppercase, since that's what they should be in anyway?

lyrixx:
> Can't we just convert the HTTP method to uppercase, since that's what they should be in anyway?

If you think this is a good approach, let's do it. I will update the PR

lyrixx:
PR Updated

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -123,6 +123,8 @@ private function normalize(bool $collection, ResourceMetadata $resourceMetadata,
                 $supported ? $operation['method'] = $upperOperationName : $operation['route_name'] = $operationName;
             }
 
+            $operation['method'] = strtoupper($operation['method']);
```
teohhanhui: This does not seem correct to me. `$operation['method']` might not be set.

Anyway, this part of the code could probably use some refactoring. `$upperOperationName` is... weird. :see_no_evil:

teohhanhui:
I'll merge it in #2734 as @lyrixx does not allow pushing to his branch. :see_no_evil:

lyrixx:
> I'll merge it in #2734 as @lyrixx does not allow pushing to his branch.

Oups, sorry. This was not intended

Issue #2683, opened 2019-04-03
[RFC] Dealing with plain PHP Api Resource


joelwurtz:
We got an app where we would expose a referential for our users through our API (using APIP)

We have something like this:

```php
/**
 * @ApiResource
 */
abstract class FooReferential {
    abstract public function getId(): string;
    abstract public function getName(): string; 

   ...
}

final class FooReferentialItem1 extends FooReferential {
    ...
}

final class FooReferentialItem2 extends FooReferential {
    ...
}
```

Then we got a DataProvider where we inject all of this class and which allows to APIP to correctly get a collection.

Getting a collection works nicely, however other use cases does not work:

 * Getting a single item seems to work weirdly: adding an `@id` / `@context` on some properties
 * When having another Resource that use one of those item, class resolver get the final class (the item) and not the parent class which is the correct resource class. So this fails to correctly convert the item to an iri item.
 * Same as before when trying to post an entity with a reference to a referential item

I'm wondering if this is an use case that is supported or that you want to support: we want to expose a referential through an API where the data are in plain PHP (no Doctrine / Mongodb / ...)

IF that the case will do some PRS to try to support this is an better way in the future.

soyuka:
>When having another Resource that use one of those item, class resolver get the final class (the item) and not the parent class which is the correct resource class. So this fails to correctly convert the item to an iri item.

You expect the resource class to be `FooReferential` not the implementation itself? This should indeed work, if not it's a bug. May also be related to https://github.com/api-platform/core/issues/2609

joelwurtz:
> You expect the resource class to be FooReferential not the implementation itself?

Yes, Actually IriConverter resource class resolver do a `get_class` without checking is this class is a resource (i'm on 2.3 FYI, don't know if this change something)

Nek-:
@joelwurtz fixing this one will probably solve your problems or at least a part of them: https://github.com/api-platform/core/issues/2609 (currently working on).

teohhanhui:
Can you verify that #2797 fixes the problem? :smile:

Issue #2609, opened 2019-03-13
Identifier is ignored for interface (and maybe inherited resources)


Nek-:
Related to #2086 for some context.
Consider the following:

```yaml
# Example extracted from ApiPlatform test suite
# tests/Fixtures/TestBundle/Resources/config/api_resources.yml
resources:
    ApiPlatform\Core\Tests\Fixtures\TestBundle\OtherResources\ResourceInterface:
        itemOperations:
            get:
                method: 'GET'

        collectionOperations:
            get:
                method: 'GET'

        # Add this legit config
        properties:
            foo:
                identifier: true
```

Output will be **without id** while it's actually possible to retrieve the id since it is specified.

There are 2 problems leading to this conclusion:

1. The ResourceClassResolver does not check inheritance. A quick fix (idk if it work for any situation ATM) would be this:
```diff
diff --git a/src/Api/ResourceClassResolver.php b/src/Api/ResourceClassResolver.php
index f75f5822..a0a45722 100644
--- a/src/Api/ResourceClassResolver.php
+++ b/src/Api/ResourceClassResolver.php
@@ -50,6 +50,7 @@ final class ResourceClassResolver implements ResourceClassResolverInterface
         if (
             null === $type
             || ((!$strict || $resourceClass === $type) && $isResourceClass = $this->isResourceClass($type))
+            || is_subclass_of($value, $resourceClass)
         ) {
             return $resourceClass;
         }
```

2. The `IriConverter` uses only the actual class of the item to match ressource instead of also checking inheritance

There's maybe a bunch of other issues I'm not aware of, but fixing this should fix the main part of it.

**This is probably related to another issue** The following test does not check the id and guess what, it's actually wrong !
```gherkin
# features/main/table_inheritance.feature:294
   Scenario: Get the parent interface collection
     When I send a "GET" request to "/resource_interfaces"
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
     And the JSON should be valid according to this schema:
     """
     {
       "type": "object",
       "properties": {
         "hydra:member": {
           "type": "array",
           "items": {
             "type": "object",
             "properties": {
               "@type": {
                 "type": "string",
                 "pattern": "^ResourceInterface$"
               },
               "@id": {
                 "type": "string",
                 "pattern": "^_:"
               },
               "foo": {
                 "type": "string",
                 "required": "true"
               }
             }
           },
           "minItems": 1
         }
[...]

PR #2714, opened 2019-04-07
Resource resolver for inheritance

Nek-:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2609 / #2683 / #2759
| License       | MIT
| Doc PR        | N/A

1. The resource class resolver now works well with inheritance and is able to resolve resources using the ApiPlatform resource list (more clever work, was needed considering `isResource` is doing it!)
2. This change adds a new feature: inherited items from API resources now are working as an API, a new behat test proves it
3. some parts of ApiPlatform were not compatible because it does not uses the resource resolver but "get class" to get the resource, theses parts are also fixed

Notice: the cache from the resource class resolver is removed here. It was introduced in 2015 and was probably a great improvement, but since 2016 resources are not re-generated on each call to the factory but we call instead of a cache factory. This cache was just adding complexity so I removed it.

### DIFF
--- a/features/main/table_inheritance.feature
+++ b/features/main/table_inheritance.feature
@@ -307,6 +307,16 @@ Feature: Table inheritance
            "items": {
              "type": "object",
              "properties": {
+               "@type": {
+                 "type": "string",
+                 "pattern": "^ResourceInterface$",
+                 "required": "true"
+               },
+               "@id": {
+                 "type": "string",
+                 "pattern": "^/resource_interfaces/",
+                 "required": "true"
+               },
                "foo": {
                  "type": "string",
                  "required": "true"
@@ -337,6 +347,16 @@ Feature: Table inheritance
         "context": {
           "type": "string",
           "pattern": "ResourceInterface$"
+        },
+        "@id": {
+          "type": "string",
+          "pattern": "^/resource_interfaces",
+          "required": "true"
+        },
+        "@type": {
+          "type": "string",
+          "pattern": "^ResourceInterface$",
+          "required": "true"
         },
         "foo": {
           "type": "string",
--- a/features/main/table_inheritance.feature
+++ b/features/main/table_inheritance.feature
@@ -446,3 +446,114 @@ Feature: Table inheritance
       }
     }
     """
+
+  @!mongodb
+  Scenario: Generate iri from parent resource
+    Given there are 3 sites with internal owner
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/sites"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@type": {
+                "type": "string",
+                "pattern": "^Site$",
+                "required": "true"
+              },
+              "@id": {
+                "type": "string",
+                "pattern": "^/sites/\\d+$",
+                "required": "true"
+              },
+              "id": {
+                "type": "integer",
+                "required": "true"
+              },
+              "title": {
+                "type": "string",
+                "required": "true"
+              },
+              "description": {
+                "type": "string",
+                "required": "true"
+              },
+              "owner": {
+                "type": "string",
+                "pattern": "^/custom_users/\\d+$",
+                "required": "true"
+              }
+            }
+          },
+          "minItems": 3,
+          "maxItems": 3,
+          "required": "true"
+        }
+      }
+    }
+    """
+
+  @!mongodb
+  @createSchema
+  Scenario: Generate iri from current resource even if parent class is a resource
+    Given there are 3 sites with external owner
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/sites"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@type": {
+                "type": "string",
+                "pattern": "^Site$",
+                "required": "true"
+              },
+              "@id": {
+                "type": "string",
+                "pattern": "^/sites/\\d+$",
+                "required": "true"
+              },
+              "id": {
+                "type": "integer",
+                "required": "true"
+              },
+              "title": {
+                "type": "string",
+                "required": "true"
+              },
+              "description": {
+                "type": "string",
+                "required": "true"
+              },
+              "owner": {
+                "type": "string",
+                "pattern": "^/external_users/\\d+$",
+                "required": "true"
+              }
+            }
+          },
+          "minItems": 3,
+          "maxItems": 3,
+          "required": "true"
+        }
+      }
+    }
+    """
--- a/features/main/table_inheritance.feature
+++ b/features/main/table_inheritance.feature
@@ -20,15 +20,18 @@ Feature: Table inheritance
       "properties": {
         "@type": {
           "type": "string",
-          "pattern": "^DummyTableInheritanceChild$"
+          "pattern": "^DummyTableInheritanceChild$",
+          "required": "true"
         },
         "@context": {
           "type": "string",
-          "pattern": "^/contexts/DummyTableInheritanceChild$"
+          "pattern": "^/contexts/DummyTableInheritanceChild$",
+          "required": "true"
         },
         "@id": {
           "type": "string",
-          "pattern": "^/dummy_table_inheritance_children/1$"
+          "pattern": "^/dummy_table_inheritance_children/1$",
+          "required": "true"
         },
         "name": {
           "type": "string",
@@ -61,7 +64,8 @@ Feature: Table inheritance
             "properties": {
               "@type": {
                 "type": "string",
-                "pattern": "^DummyTableInheritanceChild$"
+                "pattern": "^DummyTableInheritanceChild$",
+                "required": "true"
               },
               "name": {
                 "type": "string",
@@ -80,7 +84,40 @@ Feature: Table inheritance
     }
     """
 
-  @createSchema
+  Scenario: Some children not api resources are created in the app
+    When some dummy table inheritance data but not api resource child are created
+    And I send a "GET" request to "/dummy_table_inheritances"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@type": {
+                "type": "string",
+                "pattern": "^DummyTableInheritance(Child)?$",
+                "required": "true"
+              },
+              "name": {
+                "type": "string",
+                "required": "true"
+              }
+            }
+          },
+          "minItems": 1
+        }
+      },
+      "required": ["hydra:member"]
+    }
+    """
+
   Scenario: Create a table inherited resource
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "POST" request to "/dummy_table_inheritance_children" with body:
@@ -97,15 +134,18 @@ Feature: Table inheritance
       "properties": {
         "@type": {
           "type": "string",
-          "pattern": "^DummyTableInheritanceChild$"
+          "pattern": "^DummyTableInheritanceChild$",
+          "required": "true"
         },
         "@context": {
           "type": "string",
-          "pattern": "^/contexts/DummyTableInheritanceChild$"
+          "pattern": "^/contexts/DummyTableInheritanceChild$",
+          "required": "true"
         },
         "@id": {
           "type": "string",
-          "pattern": "^/dummy_table_inheritance_children/1$"
+          "pattern": "^/dummy_table_inheritance_children/3$",
+          "required": "true"
         },
         "name": {
           "type": "string",
@@ -136,15 +176,18 @@ Feature: Table inheritance
       "properties": {
         "@type": {
           "type": "string",
-          "pattern": "^DummyTableInheritanceDifferentChild$"
+          "pattern": "^DummyTableInheritanceDifferentChild$",
+          "required": "true"
         },
         "@context": {
           "type": "string",
-          "pattern": "^/contexts/DummyTableInheritanceDifferentChild$"
+          "pattern": "^/contexts/DummyTableInheritanceDifferentChild$",
+          "required": "true"
         },
         "@id": {
           "type": "string",
-          "pattern": "^/dummy_table_inheritance_different_children/2$"
+          "pattern": "^/dummy_table_inheritance_different_children/4$",
+          "required": "true"
         },
         "name": {
           "type": "string",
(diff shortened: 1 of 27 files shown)


tests/Fixtures/TestBundle/Entity/DummyTableInheritanceNotApiResourceChild.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ */
+class DummyTableInheritanceNotApiResourceChild extends DummyTableInheritance
```
soyuka: `NonApiResourceTableInheritedDummyChild` would be better? @teohhanhui do you have a name in mind? :p

tests/Fixtures/TestBundle/Entity/DummyTableInheritanceNotApiResourceChild.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    /**
+     * Considering this class is not an ApiResource, what should be render is only parent fields.
```
soyuka: Because this class is not tagged as ApiResource, only the parent fields should be in the result.


I don't think that this comment belong here though.

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -112,10 +112,21 @@ public function getItemFromIri(string $iri, array $context = [])
...
     {
         $resourceClass = $this->getObjectClass($item);
+
+        return $this->getIriFromItemWithResource($item, $resourceClass, $referenceType);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getIriFromItemWithResource($item, string $resourceClass, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
```
soyuka: WDYT about getIriFromItemWithContext? Where we'd add a $context['resource_class']? If we have other use cases in the future at least we won't have to change this interface again.

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -40,7 +40,7 @@
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class IriConverter implements IriConverterInterface
+final class IriConverter implements ResourceIriConverterInterface
```
soyuka: `ContextAwareIriConverterInterface` if we follow my next comment.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -121,7 +121,10 @@ private function reset(): void
     private function storeEntityToPublish($entity, string $property): void
     {
         $resourceClass = $this->getObjectClass($entity);
-        if (!$this->resourceClassResolver->isResourceClass($resourceClass)) {
+
+        if ($this->resourceClassResolver->isResourceClass($resourceClass)) {
+            $resourceClass = $this->resourceClassResolver->getResourceClass($entity);
+        } else {
             return;
         }
```
soyuka:
```
        if (!$this->resourceClassResolver->isResourceClass($resourceClass)) {
             return;
        }

        $resourceClass = $this->resourceClassResolver->getResourceClass($entity);
```

src/Api/ResourceClassResolverInterface.php
```diff
@@ -25,12 +25,20 @@ interface ResourceClassResolverInterface
     /**
      * Guesses the associated resource.
      *
+     * @param mixed  $value         Object you're playing with
+     * @param string $resourceClass Resource class it is supposed to be (could be parent class for instance)
+     * @param bool   $strict        value must be type of resource class given or it will return type
+     *
```
soyuka: For next reviewers I'm to keep these comment for future developers because it's hard to guess.

src/Api/ResourceClassResolver.php
```diff
@@ -74,16 +89,28 @@ public function getResourceClass($value, string $resourceClass = null, bool $str
...
+            if ($type === $resourceClass || is_subclass_of($type, $resourceClass)) {
+                return true;
+            }
         }
 
+        return false;
+    }
+
+    /**
+     * Same of isResourceClass but more strict: it ignores inheritance.
```
alanpoulain:
```suggestion
     * Same of isResourceClass but stricter: it ignores inheritance.
```

features/main/table_inheritance.feature
```diff
@@ -80,7 +84,40 @@ Feature: Table inheritance
...
+      "type": "object",
+      "properties": {
+        "hydra:member": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "@type": {
+                "type": "string",
+                "pattern": "^DummyTableInheritance(Child)?$",
```
alanpoulain: We're not really testing there is at least one child, right?

tests/Fixtures/TestBundle/Entity/DummyTableInheritanceNotApiResourceChild.php
Nek-: Well, starting by "DummyTableInheritance" seems to have been a convention for similar objects.

features/main/table_inheritance.feature
alanpoulain: I don't understand why the dependency between scenarios is an issue here.
You're right about the first part though. But maybe we can avoid this for this scenario?

soyuka: > I don't understand why the dependency between scenarios is an issue here.

It's always an issue when you want to run 1 failing test case without running the whole suite. tests should respect the FIRST principle.

alanpoulain: Sure, I agree with this. But I don't why the link with what I said.

soyuka: Yes sorry you're right. IIUC you're saying we should either have `^DummyTableInheritanceChild$` or `^DummyTableInheritance$` ? 
IIRC there may be both types? If not we definitely need to pick one.

Nek-: @alanpoulain oh, it's possible to be sure the item is here at least once? How?

src/Bridge/Symfony/Routing/IriConverter.php
Nek-: I don't see any use case soon. I think it's confusing considering the main use case is to pass it from another context (serialization context). This seems good to me.

Also, I hate config-arrays. If you want it I'm in favor of creating a config object.

Nek-:
Appveyor is a liar. I'm pretty sure this is also cool on Windows! è_é

src/Api/ResourceClassResolver.php
Nek-: Change done.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
Nek-: Done.

src/Bridge/Symfony/Routing/IriConverter.php
Nek-: See my previous response.

tests/Fixtures/TestBundle/Entity/DummyTableInheritanceNotApiResourceChild.php
Nek-: Removed.

features/main/table_inheritance.feature
```diff
@@ -350,3 +446,118 @@ Feature: Table inheritance
...
+    }
+    """
+
+  @!mongodb
+  @createSchema
+  Scenario: Generate iri from current resource even if parent class is a resource
+    Given there are 3 sites with external owner
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/sites"
+    And print last JSON response
```
antograssiot:
```suggestion
```

features/main/table_inheritance.feature
antograssiot: you might be able use tuple validation maybe ? (that's only 3 items... https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation)

features/main/table_inheritance.feature
```diff
@@ -80,7 +84,40 @@ Feature: Table inheritance
...
+              "@type": {
+                "type": "string",
+                "pattern": "^DummyTableInheritance(Child)?$",
+                "required": "true"
+              },
+              "name": {
+                "type": "string",
+                "required": "true"
+              }
+            }
```
antograssiot:
```suggestion
            },
            "additionalProperties": false
```

and the same can be added in other tests to be more accurate

features/main/table_inheritance.feature
```diff
@@ -350,3 +446,118 @@ Feature: Table inheritance
       }
     }
     """
+
+  @!mongodb
+  @createSchema
+  @dropSchema
```
antograssiot:
```suggestion
```

This annotation doesn't exists anymore if I'm not mistaken

features/main/table_inheritance.feature
```diff
@@ -350,3 +446,118 @@ Feature: Table inheritance
...
+  @createSchema
+  Scenario: Generate iri from current resource even if parent class is a resource
+    Given there are 3 sites with external owner
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/sites"
+    And print last JSON response
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And print last JSON response
```
antograssiot:
```suggestion
```

tests/Serializer/AbstractItemNormalizerTest.php
```diff
@@ -174,6 +175,8 @@ public function testNormalize()
 
         $resourceClassResolverProphecy = $this->prophesize(ResourceClassResolverInterface::class);
         $resourceClassResolverProphecy->getResourceClass($dummy, null, true)->willReturn(Dummy::class)->shouldBeCalled();
+        //$resourceClassResolverProphecy->getResourceClass($relatedDummy, RelatedDummy::class, true)->willReturn(RelatedDummy::class)->shouldBeCalled();
```
antograssiot: leftover ?

tests/Api/ResourceClassResolverTest.php
```diff
@@ -170,17 +171,67 @@ public function testGetResourceClassWithChildResource()
...
+
+    public function testItResolveParentResourceClassOfChildrenOfResourceClasses()
+    {
+        $dummy = new DummyTableInheritanceNotApiResourceChild();
+        $resourceNameCollectionFactoryProphecy = $this->prophesize(ResourceNameCollectionFactoryInterface::class);
+        $resourceNameCollectionFactoryProphecy->create()->willReturn(new ResourceNameCollection([DummyTableInheritance::class]))->shouldBeCalled();
+
+        $resourceClassResolver = new ResourceClassResolver($resourceNameCollectionFactoryProphecy->reveal());
+//        $this->assertTrue($resourceClassResolver->isResourceClass(DummyTableInheritanceNotApiResourceChild::class));
+//        $this->assertEquals(DummyTableInheritance::class, $resourceClassResolver->getResourceClass($dummy));
```
antograssiot: leftover ?

features/main/table_inheritance.feature
Nek-: I did that ? 😇

tests/Api/ResourceClassResolverTest.php
Nek-: God. To be removed.

src/Api/ResourceClassResolverInterface.php
teohhanhui: The comments need to be rewritten as they don't make much sense now.

src/Api/ResourceClassResolver.php
```diff
@@ -74,16 +89,28 @@ public function getResourceClass($value, string $resourceClass = null, bool $str
      */
     public function isResourceClass(string $type): bool
     {
-        if (isset($this->localIsResourceClassCache[$type])) {
-            return $this->localIsResourceClassCache[$type];
+        foreach ($this->resourceNameCollectionFactory->create() as $resourceClass) {
+            if ($type === $resourceClass || is_subclass_of($type, $resourceClass)) {
```
teohhanhui:
```suggestion
            if (is_a($type, $resourceClass, true)) {
```

src/Api/ResourceIriConverterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace ApiPlatform\Core\Api;
+
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\RuntimeException;
+
+/**
+ * Converts item and resources to IRI and vice versa.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
teohhanhui: Wrong author.

src/Bridge/Symfony/Routing/IriConverter.php
teohhanhui: Why not just keep the same interface but add an optional argument that's commented out?

teohhanhui:
About the Behat test suite, I could only say most of the tests that use JSON Schema are wrong. I started fixing some of them in https://github.com/api-platform/core/pull/2723 but I'm afraid it's been abandoned.

src/Bridge/Symfony/Routing/IriConverter.php
soyuka: Why not, especially that it doesn't really make sense to **not** add the `resourceClass` to the `getIriFromItem`.

teohhanhui:
I'm sure there are many places where this would still fail, but here's one that I've encountered:

https://github.com/api-platform/core/blob/v2.4.2/src/EventListener/WriteListener.php#L80

teohhanhui:
Okay, so I think this is not ideal. Why don't we inject ResourceResolver into IriConverter? That way we can keep the old interface, and the caller does not need to care about which resource class to pass into IriConverter, as indeed that should not be their job. The IriConverter should know how to do its job.

teohhanhui:
See for example https://github.com/api-platform/core/blob/v2.4.2/src/EventListener/WriteListener.php#L80

Are we supposed to the resource class resolving there? It makes no sense to me. As a caller, I have a resource item, and I want to get the IRI for it. That's all.

soyuka:
Sounds like a good alternative, wdyt @Nek- ?

teohhanhui:
But perhaps the IriConverter would need the attributes.

Nek-:
@teohhanhui I will give it a try as soon as I have the time for it. I don't really think it will work considering it will have a huge impact on the whole app. (actually, I was thinking about not deprecating the existing method).

src/Api/IriConverterInterface.php
```diff
@@ -21,6 +21,8 @@
  * Converts item and resources to IRI and vice versa.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @deprecated in favor of ResourceIriConverterInterface to be removed in ApiPlatform 3.0
```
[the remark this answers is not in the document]
Nek-: I think your concern is legit. We could add a new parameter to the existing method. It would lose a bit of its sense. But really I think you're right about that fact.

teohhanhui:
I'm working on an alternative approach at https://github.com/api-platform/core/pull/2797

Nek-:
@teohhanhui feel free to take tests

teohhanhui:
@Nek- Yes, I'll take the Behat tests. :smile:

features/main/table_inheritance.feature
teohhanhui: We will use multiple `"contains"`: https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.4.6

teohhanhui:
Replaced by #2797

Thanks @Nek- for your work! :smile:

Issue #2502, opened 2019-01-30
Swagger HAL example is plain json


martijnhartlief:
When the only available format is hal+json, the docs shows the example in plain json. Maybe this is how it's supposed to work, but I think it's quite confusing since the actual output is completely different than the example.

soyuka:
#2415

Issue #2415, opened 2018-12-22
OpenAPI v3: generate specific JSON Schemas for JSON-LD, JSON:API...


dunglas:
Unlike OpenAPI v2, OpenAPI v3 allows to specify different JSON Schema per MIME type. It would now be possible to document the format-specific reserved keys such as `@id` and `@type` in JSON-LD, and the format specific structures such as `hydra:member` and the errors using OpenAPI.

It would be a nice addition.

MyDigitalLife:
@dunglas Could you specify a bit how this can be achieved? If so I might try and add some schema's for json:api when I have some free time.

dunglas:
Hi @MyDigitalLife the `response` object can now contain a specific JSON Schema per `content`: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#response-object

The idea would be to modify https://github.com/api-platform/core/blob/master/src/Swagger/Serializer/DocumentationNormalizer.php#L180 to generate a specific JSON Schema for the JSON-LD response that includes the definition for `@id`, `@type` `hydra:member` and so on. Later, we would do the same for other formats having specific structures (JSON:API and HAL).

MyDigitalLife:
I just did some digging and found a way to do this. Took some searching but i can now define a format for a MimeType and when selecting a new you it switches. You can see my work in progress here: https://github.com/api-platform/core/compare/master...MyDigitalLife:mime-type-components-schema

I will be updating it so that the other operations also use the new formats and will most likely will move the generation of the definition schema to its own class so new format can easily be added. 

@dunglas Got any comments on what I'm doing so far?

dunglas:
@MyDigitalLife sorry I hadn't reviewed your work yet, but I think that it's very important to have this supported. Would you mind opening a PR directly so it will be easier to review.

DrummerKH:
Hello, could you answer, please, when this feature will be released?

dunglas:
@DrummerKH when someone will finish the open PR or contribute a new one.

soyuka:
still relevant

PR #2715, opened 2019-04-07
Mime type components schema

MyDigitalLife:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2415 #2502
| License       | MIT
| Doc PR        | 

This is a PR for being able to have different body for mime types.

After these tests are fixed there should be a different body schema created for every mime type. The basics of this can be laid down in this pull request and adding the correct bodies can be done later as long as there is a default body that falls back to how things work now.

Any input in how this can be improved are welcome. 

- [x] Extract  JSON Schema formatting to its own classes
- [x] Load schema formatters using tags: https://symfony.com/doc/current/service_container/tags.html
- [x] Improve formatters to comply with at least 2 different standards (other can be added later on or in a other pull requests)
- [x] Fix tests


src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -604,14 +707,37 @@ private function getDefinitionSchema(bool $v3, string $resourceClass, ResourceMe
         }
 
         $options = isset($serializerContext[AbstractNormalizer::GROUPS]) ? ['serializer_groups' => $serializerContext[AbstractNormalizer::GROUPS]] : [];
+
+        if ($v3) {
+            if ('application/vnd.api+json' === $mimeType) {
```
dunglas: Cab we extract the code generating the JSON Schema corresponding to every supported formats in specific classes. It will make this class less complex (it is already to big and complex), and will allow users to register their on JSON Schema generators for their custom formats.

MyDigitalLife: Yeah, that's on my todo list. I'll add a todo list to the so we know what needs to happen.

MyDigitalLife: I added some functionality to do this, its not finished still needs to get the formatters from the config and create more functions for more complex setups. Its a start, just let me know if its going in the correct direction.

MyDigitalLife:
Alright, looks like this is mostly done. I would love some feedback. There are a few loose ends to tie up (php-cs-fixer and a few few test that still fail), but I would love some feedback. Mostly about the following points:

- Fix a one to many relationship in the JsonApiFormatter, there aren't any current tests that have this relationship in for the docs. I would love some help writing this one so I can finish the check when a relationship is a one to many or a many to one. ATM I don't have the knowledge to mock the resources I need for this. 
- Fix the broken tests. There are 3 tests that re broken and I have no idea why, you can find those [here](https://circleci.com/gh/api-platform/core/11412?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I would love some help with fixing these issues. 
- I think I'm using some design pattern name wrong, I'm calling the class i use to get the correct formatter a Factory but I'm not use that correct. 
- General improvements I can make to the code or maybe some extra tests i can write. 

cc @dunglas @teohhanhui 

Looking forward to your feedback and to get the code to the point it can be merged and start writing more formatters.

antograssiot:
The failing tests are because you change the path of references in component's schemas.

In [features/openapi/docs.feature](https://github.com/api-platform/core/blob/bc0806be383a6fc6820a1230bbddd3ea29542ac7/features/openapi/docs.feature#L19) we are checking in the generated OpenAPI for an existing class being defined using [SwaggerContext ::getClassInfo()](https://github.com/api-platform/core/blob/bc0806be383a6fc6820a1230bbddd3ea29542ac7/features/bootstrap/SwaggerContext.php#L178)

For OpenAPI v3, classes were expected in `#/components/schemas/classname` but now there are stored in `#/components/schemas/mimetype/classname` so you need to update this function

MyDigitalLife:
Thanks for the feedback.

I still need to write a test and implement a one to many relationship. I would love some help with this as I haven't figured out how to mock that in the correct tests.

src/Swagger/SchemaFormatter/JsonApiSchemaFormatter.php
```diff
@@ -0,0 +1,108 @@
...
+            ],
+        ];
+    }
+
+    public function setProperty(
+        \ArrayObject $definitionSchema,
+        $normalizedPropertyName,
+        \ArrayObject $property,
+        PropertyMetadata $propertyMetadata
+    ): void {
```
soyuka: We usually inline these.

src/Swagger/SchemaFormatter/JsonApiSchemaFormatter.php
```diff
@@ -0,0 +1,108 @@
...
+            ];
+            // @todo: Fix one to many statement.
+//            if (false) {
+//                $data = [
+//                    'type' => 'object',
+//                    'properties' => [
+//                        $data,
+//                    ],
+//                ];
+//            }
```
soyuka: We may leave the todo comment but not the commented code

src/Swagger/SchemaFormatter/JsonApiSchemaFormatter.php
```diff
@@ -0,0 +1,108 @@
...
+                                        'type' => 'string',
+                                    ],
+                                ],
+                            ],
+                            'data' => $data,
+                        ],
+                    ],
+                ],
+            ];
+        } else {
```
soyuka: return early to avoid the else?

src/Swagger/SchemaFormatter/JsonSchemaFormatter.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return [];
+    }
+
+    public function setProperty(
+        \ArrayObject $definitionSchema,
+        $normalizedPropertyName,
+        \ArrayObject $property,
+        PropertyMetadata $propertyMetadata
+    ): void {
```
soyuka: inline this?

src/Swagger/SchemaFormatter/SchemaFormatterProvider.php
```diff
@@ -0,0 +1,43 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Swagger\SchemaFormatter;
+
+use ApiPlatform\Core\Exception\FormatterNotFoundException;
+
+class SchemaFormatterProvider
+{
+    /** @var SchemaFormatterInterface[] */
```
dunglas: Move it to the contructors's doc block

src/Swagger/SchemaFormatter/SchemaFormatterProvider.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $this->formatters = $formatters;
+    }
+
+    /**
+     * @param string $mimeType
+     *
+     * @return SchemaFormatterInterface
+     */
+    public function getFormatter($mimeType)
```
dunglas:
```suggestion
    public function getFormatter(string $mimeType): SchemaFormatterInterface
```

src/Swagger/SchemaFormatter/SchemaFormatterProvider.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /** @var SchemaFormatterInterface[] */
+    private $formatters;
+
+    public function __construct(/* iterable */ $formatters)
+    {
+        $this->formatters = $formatters;
+    }
+
+    /**
```
dunglas: DocBlock to be removed, use typehints instead

src/Swagger/SchemaFormatter/SchemaFormatterProvider.php
```diff
@@ -0,0 +1,43 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Swagger\SchemaFormatter;
+
+use ApiPlatform\Core\Exception\FormatterNotFoundException;
+
+class SchemaFormatterProvider
```
dunglas: For consistency with the rest of the code base (see DataProvider / DataPersister for instance), you should rename this class `ChainSchemaFormatter` and make it an implementation of `SchemaFormatterInterface`. Then only inject `SchemaFormatterInterface` in the normalizer.

Also:
```suggestion
final class SchemaFormatterProvider
```

src/Swagger/SchemaFormatter/SchemaFormatterProvider.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function getFormatter($mimeType)
+    {
+        foreach ($this->formatters as $formatter) {
+            if ($formatter->supports($mimeType)) {
+                return $formatter;
+            }
+        }
+
+        throw new FormatterNotFoundException(sprintf('Formatter for mimetype "%s" not supported', $mimeType));
```
dunglas:
```suggestion
        throw new FormatterNotFoundException(sprintf('No formatter supporting the "%s" MIME type is available.', $mimeType));
```

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -87,11 +88,12 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
...
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceClassResolverInterface $resourceClassResolver, OperationMethodResolverInterface $operationMethodResolver, OperationPathResolverInterface $operationPathResolver, UrlGeneratorInterface $urlGenerator = null, $filterLocator = null, NameConverterInterface $nameConverter = null, bool $oauthEnabled = false, string $oauthType = '', string $oauthFlow = '', string $oauthTokenUrl = '', string $oauthAuthorizationUrl = '', array $oauthScopes = [], array $apiKeys = [], SubresourceOperationFactoryInterface $subresourceOperationFactory = null, bool $paginationEnabled = true, string $paginationPageParameterName = 'page', bool $clientItemsPerPage = false, string $itemsPerPageParameterName = 'itemsPerPage', OperationAwareFormatsProviderInterface $formatsProvider = null, bool $paginationClientEnabled = false, string $paginationClientEnabledParameterName = 'pagination', array $defaultContext = [], SchemaFormatterProvider $schemaFormatterFactory = null)
```
dunglas: `SchemaFormatterInterface $schemaFormatter` instead of `SchemaFormatterProvider` (see my previous comment)

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -116,14 +118,14 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
         $this->paginationEnabled = $paginationEnabled;
         $this->paginationPageParameterName = $paginationPageParameterName;
         $this->apiKeys = $apiKeys;
-        $this->subresourceOperationFactory = $subresourceOperationFactory;
```
dunglas: Should be reverted I guess

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -241,9 +243,17 @@ private function updateGetOperation(bool $v3, \ArrayObject $pathOperation, array
         $serializerContext = $this->getSerializerContext($operationType, false, $resourceMetadata, $operationName);
 
         $responseDefinitionKey = false;
+        $responseDefinitionKeys = [];
         $outputMetadata = $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, 'output', ['class' => $resourceClass], true);
         if (null !== $outputClass = $outputMetadata['class'] ?? null) {
-            $responseDefinitionKey = $this->getDefinition($v3, $definitions, $resourceMetadata, $resourceClass, $outputClass, $serializerContext);
+            if ($v3) {
+                foreach ($mimeTypes as $mimeType) {
+                    $responseDefinitionKeys[$mimeType] = $this->getDefinition($v3, $definitions, $resourceMetadata, $resourceClass,
```
dunglas: What about:

* Renaming `SchemaFormatterInterface` in `DefinititionNormalizerInterface`
* Adding only a simple single method to this interface such as `SchemaFormatterInterface::getDefinition(string $resourceClass, string $operationType, string $operationName = null, string $inputOrOutput, array $serializerContext = [], string $mimeType): array` (other arguments such as outputClass and  can be generated using the factory services)

It would simplify the code a lot

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -256,16 +266,21 @@ private function updateGetOperation(bool $v3, \ArrayObject $pathOperation, array
...
-            if ($responseDefinitionKey) {
-                if ($v3) {
-                    $successResponse['content'] = array_fill_keys($mimeTypes, [
-                        'schema' => [
-                            'type' => 'array',
-                            'items' => ['$ref' => sprintf('#/components/schemas/%s', $responseDefinitionKey)],
-                        ],
-                    ]);
-                } else {
+            if ($v3) {
```
dunglas: Is this change necessary? It generates the same output AFAIU

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -286,10 +301,19 @@ private function updateGetOperation(bool $v3, \ArrayObject $pathOperation, array
         $pathOperation = $this->addItemOperationParameters($v3, $pathOperation);
 
         $successResponse = ['description' => sprintf('%s resource response', $resourceShortName)];
-        if ($responseDefinitionKey) {
-            if ($v3) {
-                $successResponse['content'] = array_fill_keys($mimeTypes, ['schema' => ['$ref' => sprintf('#/components/schemas/%s', $responseDefinitionKey)]]);
-            } else {
+
+        if ($v3) {
```
dunglas: Same here

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -408,7 +446,16 @@ private function getSubresourceResponse(bool $v3, $mimeTypes, bool $collection,
             ];
 
             if ($v3) {
-                $okResponse['content'] = array_fill_keys($mimeTypes, ['schema' => ['$ref' => sprintf('#/components/schemas/%s', $definitionKey)]]);
+                foreach ($mimeTypes as $mimeType) {
```
dunglas: Same here. Try to reduce the diff size as much as possible to ease the review process

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -601,14 +714,30 @@ private function getDefinitionSchema(bool $v3, string $resourceClass, ResourceMe
         }
 
         $options = isset($serializerContext[AbstractNormalizer::GROUPS]) ? ['serializer_groups' => $serializerContext[AbstractNormalizer::GROUPS]] : [];
+
+        $schemaFormatter = null;
+        if ($v3) {
+            try {
+                $schemaFormatter = $this->schemaFormatterFactory->getFormatter($mimeType);
+            } catch (\Exception $e) {
```
dunglas: The default formatter's `supports` method should just always return `true` instead of thowing. It will allow to remove this hack.

src/Swagger/SchemaFormatter/JsonSchemaFormatter.php
```diff
@@ -0,0 +1,38 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Swagger\SchemaFormatter;
+
+use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
+
+class JsonSchemaFormatter implements SchemaFormatterInterface
```
dunglas: Should be renamed `DefaultSchemaFormatter`, cause it can it's the default one, also used for XML and so on.

src/Swagger/SchemaFormatter/JsonSchemaFormatter.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function supports(string $mimeType): bool
+    {
+        return 'application/json' === $mimeType;
```
dunglas: Should always return `true`, and should be registered with a low priority.

src/Swagger/Serializer/DocumentationNormalizer.php
MyDigitalLife: Actually this line is duplicate, check a few lines up. I just removed the duplicate line.

src/Swagger/Serializer/DocumentationNormalizer.php
Simperfit: this has been done

Issue #2461, opened 2019-01-18
2.4.0 beta is not compatible with symfony 4.2


lyrixx:
I was too excited to test 2.4.0 so I updated it, but it failed:

```
In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:
                                                                                                                 
  The service "api_platform.messenger.data_persister" has a dependency on a non-existent service "message_bus".           
```

We are using
```
symfony/symfony                      v4.2.1           The Symfony PHP framework
```

(yes, no flex, because it's an "old" application, but we get new releases anyway)

pcox64:
I fixed this in an upgraded Symfony 3 application by `composer require symfony/messenger` and adding the following under `framework` to `config.yml`:

```
messenger:
  default_bus: ~
```

soyuka:
This may need a fix on symfony's side no?

lyrixx:
IMHO, no. either you make the messenger component mandatory, or you use it only if the service is here.

soyuka:
I'll fix this asap

lyrixx:
Thanks

soyuka:
Mhh it's weird, we do have a check and load the `messenger.xml` only when the interface exists:

https://github.com/api-platform/core/blob/2.4/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L154-L156

lyrixx:
Class exisit is different from service exisit

ragboyjr:
@soyuka @lyrixx `messenger_bus` is a private service. When running off of the latest api-platform 2.4, this fails to load the messenger component, since `$container->has('message_bus')` is always false.

Checking how symfony's FrameworkBundle works, they don't ever check if a service exists, instead they do an interface exists check (like what we currently have and what was there previously), and in some cases, they check a config value to see if the module should be enabled or not.

@lyrixx correct me if i'm wrong, but if if you were getting that error, that means that you had symfony/messenger component installed but due to some configuration, the message bus service didn't exist.

I think we should revert the `$container->has('messenger.message_bus')` since it doesn't work, and possibly see about adding a config option to enable/disable messenger integration entirely since messenger isn't a stable component.

ragboyjr:
Y, re-reading @lyrixx's original post, it makes sense what's going on. You have the entire SF framework loaded which would cause the messenger class to exist, but you did NOT have messenger services loaded.

[...]

PR #2721, opened 2019-04-09
Refactor the bundle configuration and extension

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/issues/2461#issuecomment-478976338
| License       | MIT
| Doc PR        | N/A

Also make sure we disable the Symfony Messenger integration by default if `symfony/symfony` is installed.


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -401,31 +348,26 @@ private function registerJsonProblemConfiguration(array $errorFormats, XmlFileLo
...
         $loader->load('graphql.xml');
     }
 
-    /**
-     * Registers configuration for integration with third-party bundles.
-     *
-     * @param string[] $bundles
-     */
-    private function registerBundlesConfiguration(array $bundles, array $config, XmlFileLoader $loader)
+    private function registerLegacyBundlesConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader): void
```
teohhanhui: Yes, I consider `FOSUserBundle` as legacy. :wink:

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -581,32 +512,32 @@ private function registerDataCollectorConfiguration(ContainerBuilder $container,
...
             return;
         }
 
         $loader->load('messenger.xml');
     }
 
-    private function registerElasticsearchConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader)
+    private function registerElasticsearchConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader): void
     {
-        $enabled = $config['elasticsearch']['enabled'] && class_exists(Client::class);
```
teohhanhui: I removed this check because `elasticsearch` support needs to be explicitly enabled, so it's more useful for the user to get an exception in this case, instead of silently disabling support.

teohhanhui: Maybe better if we just throw an exception here directly. WDYT @meyerbaptiste?

teohhanhui: Changed to throw exception in `Configuration` instead.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -235,25 +290,56 @@ public function getConfigTreeBuilder()
...
+            ->end();
+    }
 
+    private function addMessengerSection(ArrayNodeDefinition $rootNode): void
+    {
+        $rootNode
+            ->children()
                 ->arrayNode('messenger')
-                    ->{interface_exists(MessageBusInterface::class) ? 'canBeDisabled' : 'canBeEnabled'}()
+                    ->{!class_exists(FullStack::class) && interface_exists(MessageBusInterface::class) ? 'canBeDisabled' : 'canBeEnabled'}()
```
soyuka: Not sure to get why you check for the `FullStack` class, messenger would work without it no?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -235,25 +290,56 @@ public function getConfigTreeBuilder()
...
 
+    private function addElasticsearchSection(ArrayNodeDefinition $rootNode): void
+    {
+        $rootNode
+            ->children()
                 ->arrayNode('elasticsearch')
                     ->canBeEnabled()
                     ->addDefaultsIfNotSet()
                     ->children()
+                        ->booleanNode('enabled')
```
soyuka: IIUC you removed calls like these in favor of a dynamic ` canBe(Dis|En)abled` except this one?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
teohhanhui: It's consistent with what's done in the FrameworkBundle, so that we use the correct default.

https://github.com/symfony/symfony/blob/v4.2.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L1003

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
teohhanhui: I'm not sure if this call is required since I'm redefining(?) a child node. I'll remove it if it makes no difference.

teohhanhui: I mean the `defaultTrue` / `defaultFalse`. We will always need the `booleanNode('enabled')` calls, since otherwise we can't get to the child node (for adding `validate`).

teohhanhui: Seems like there's nothing we can do about this.

soyuka:
Thanks @teohhanhui !

lyrixx:
Hello. Thanks for your work on this patch. Could you do a new release so that I can test it easily? Thanks

teohhanhui:
Yeah, we should probably tag a new version soon...

PR #2376, opened 2018-12-12
WIP: Allow configuration of extra info fields for API documentation

jpjoao:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | TBC 

The OpenAPI documentation specifies an info block (https://swagger.io/specification/#infoObject) when generating the JSON description of the API. Currently, API platform don't allow the correct configuration of such block.

This PR aims to allow such configuration by introducing new configurations and an improved way of dealing with it.

Target is that a configuration like this:
```yaml
api_platform:
  title: 'My API'
  description: >
    I have a very fancy API and I am proud of it
  # Not yet available on API-Platform:
  contact:
    name: "Proud API team"
    url: "https://very.proud.api"
    email: "tream@very.proud.api"
  # Not yet available on API-Platform:
  license:
    name: "Apache 2.0"
    url: "https://www.apache.org/licenses/LICENSE-2.0.html"
  version: '0.0.1'
 # Not yet available on API-Platform:
  termsOfService: 'http://example.com/terms/'
```

Would generate a swagger doc json whith this:

```JSON
{
 ... 
    "info": {
        "title": "My API",
        "version": "0.0.1",
        "contact": {
            "name": "Proud API team",
            "url": "https:\/\/very.proud.api",
            "email": "tream@very.proud.api"
        },
        "description": "I have a very fancy API and I am proud of it\n",
        "license": {
            "name": "Apache 2.0",
            "url": "https:\/\/www.apache.org\/licenses\/LICENSE-2.0.html"
        },
        "termsOfService": "http:\/\/example.com\/terms\/"
    }
...
}
```
Such information could also appear on the documentaion page (or maybe even have configuraiton to enable or disable it to be displayed).

The current state of this PR is "PoC" (Proof of Concept) and allows the configuration of "contact" only (License and terms of service are still pending) and was done with the only purpose of getting it done.

The idea is to learn from the PoC and make a better encapsulation of the info block and its content.

Feedback is very welcome and appreciated.

PR for documentation will be sent when the code is more mature.

PR #2727, opened 2019-04-11
Add full information into OpenApi info

jewome62:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2376
| License       | MIT
| Doc PR        | not yet

### DIFF
--- a/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
+++ b/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
@@ -105,7 +105,12 @@ final class SwaggerUiAction
             $this->formats = $this->formatsProvider->getFormatsFromAttributes(RequestAttributesExtractor::extractAttributes($request));
         }
 
-        $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $this->title, $this->description, $this->version, $this->formats);
+        $info = [
+            'title' => $this->title,
+            'description' => $this->description,
+            'version' => $this->version,
+        ];
+        $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $info, $this->formats);
 
         return new Response($this->twig->render('@ApiPlatform/SwaggerUi/index.html.twig', $this->getContext($request, $documentation)));
     }
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -272,8 +272,26 @@ final class Configuration implements ConfigurationInterface
                         ->end()
                     ->end()
                 ->end()
-
-            ->end();
+                ->arrayNode('open_api')
+                    ->canBeEnabled()
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('contact')
+                            ->children()
+                                ->scalarNode('name')->defaultValue('API Platform contact')->info('Name of the maintainer of the API')->end()
+                                ->scalarNode('url')->defaultValue('https://api-platform.com')->info('URL of the maintainer of the API')->end()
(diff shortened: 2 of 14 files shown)


src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -105,7 +105,12 @@ public function __invoke(Request $request)
             $this->formats = $this->formatsProvider->getFormatsFromAttributes(RequestAttributesExtractor::extractAttributes($request));
         }
 
-        $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $this->title, $this->description, $this->version, $this->formats);
+        $info = [
+            'title' => $this->title,
+            'description' => $this->description,
+            'version' => $this->version,
+        ];
+        $documentation = new Documentation($this->resourceNameCollectionFactory->create(), $info, $this->formats);
```
teohhanhui: We cannot change the constructor arguments like this. It's a BC break.

src/Documentation/Documentation.php
```diff
@@ -23,17 +23,13 @@
...
 {
     private $resourceNameCollection;
-    private $title;
-    private $description;
-    private $version;
+    private $info;
     private $mimeTypes = [];
 
-    public function __construct(ResourceNameCollection $resourceNameCollection, string $title = '', string $description = '', string $version = '', array $formats = [])
+    public function __construct(ResourceNameCollection $resourceNameCollection, array $info = [], array $formats = [])
```
teohhanhui: We cannot change the constructor arguments like this. It's a BC break.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -272,8 +272,26 @@ public function getConfigTreeBuilder()
                         ->end()
                     ->end()
                 ->end()
-
-            ->end();
+                ->arrayNode('open_api')
```
teohhanhui:
```suggestion
                ->arrayNode('openapi')
```

teohhanhui: Also, there should be a child node "info" under the "openapi" node.

jewome62: or call it openapi_info ?
because this config is not make for overide other node

teohhanhui: I disagree. We want a proper structure for the configuration. Not a bunch of top-level keys scattered without grouping.

teohhanhui: `enable_swagger` should be deprecated, and changed to this instead:

```yaml
openapi:
    enabled: true
```

src/Documentation/Documentation.php
```diff
@@ -23,17 +23,13 @@
...
     {
         $this->resourceNameCollection = $resourceNameCollection;
```
azjezz, line 31: As mentioned before, this would be a BC-break, what I suggest is doing this instead:

```suggestion
    public function __construct(ResourceNameCollection $resourceNameCollection, $titleOrInfo = '', string $description = '', string $version = '', array $formats = [])
    {
        $this->resourceNameCollection = $resourceNameCollection;
        if (\is_array($titleOrInfo)) {
          $this->info = $titleOrInfo;
        } else {
            // trigger a deprecation here : 'passing non-array as the third argument is deprecated...'
            $this->info = [
              'title' => $titleOrInfo,
              'description' => $description,
              'version' => $version,
            ];
        }
```

PR #2728, opened 2019-04-11
Change Messenger DataPersister to a high priority

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/1100
| License       | MIT
| Doc PR        | N/A

If the user needs a custom DataPersister that kicks in when `messenger` property is set to `true`, then they could use a higher priority, but that should be an edge case.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/messenger.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/messenger.xml
@@ -11,7 +11,7 @@
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
             <argument type="service" id="api_platform.message_bus" />
 
-            <tag name="api_platform.data_persister" priority="-900" />
+            <tag name="api_platform.data_persister" priority="1000" />
         </service>
 
         <service id="api_platform.messenger.data_transformer" class="ApiPlatform\Core\Bridge\Symfony\Messenger\DataTransformer" public="false">


soyuka:
I tent to disagree, if the user adds a data persister it'll be used before the messenger one and that's expected.

teohhanhui:
No, it's actually unexpected. You shouldn't need a custom DataPersister when using the Messenger integration. And adding a custom DataPersister shouldn't break that integration.

soyuka:
> You shouldn't need a custom DataPersister when using the Messenger integration. 

Exaclty so why add one?

> And adding a custom DataPersister shouldn't break that integration.

It doesn't, user code has a bigger priority than our code. If it supports the resource without checking we're in a messenger propagation case we can't do anything about it.

teohhanhui:
> If it supports the resource without checking we're in a messenger propagation case we can't do anything about it.

The case I'm making is that custom DataPersisters shouldn't need to care about not breaking the Messenger integration. Only if you need to do something custom about the Messenger integration, then set the priority to be higher than 1000. Otherwise everything works as expected.

soyuka:
The thing that bother's me is that you should have one persister per resource. For example if Dummy is persisted by MyDataProvider, I don't see how it would be persisted my the MessengerDataProvider. If you go through the Messenger, always use the Messenger as a data persister. If that's not the case, you're in the case where your Input is different and you can use the `input` option.
You could also use messenger in a synchronous matter and to use different buses we'd need to check https://github.com/api-platform/core/issues/2729.

If a developer has a data provider that uses a storage for every resources (replaces doctrine etc.) he'd need to use a priority lower then the messenger persister.

teohhanhui:
> If a developer has a data provider that uses a storage for every resources (replaces doctrine etc.) he'd need to use a priority lower then the messenger persister.

I think you meant data persister, but I still don't understand that sentence.

soyuka:
Did I convinced you last time we spoke about that? I'd like to close :smile:

teohhanhui:
I think let's wait to hear from @knp13 in https://github.com/api-platform/api-platform/issues/1100

soyuka:
Closing after exposing my view point on the subject, if someone really has a use case I'm all ears.

PR #2732, opened 2019-04-13
Update normalizer service priority

### LINKED ISSUE #1018: [question] Use Callbacks to Serialize Properties
Thank you to Api Platform for your great work !

Is it possible to use the [Serialize Properties Callback fonctionnality](https://symfony.com/doc/current/components/serializer.html#using-callbacks-to-serialize-properties-with-object-instances) in order to transform one field ?

In my example, I've got a Symfony Service which transform RTF string to HTML and I'd like with a callback to transform multiple RTF field to HTML.

Thank you !

JarJak:
This change fixes https://github.com/api-platform/api-platform/issues/1018.

Maybe it's not the best solution, but it can be a clue where to look for a bugfix or solution.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | https://github.com/api-platform/api-platform/issues/1018

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -104,7 +104,7 @@
             <argument>false</argument>
 
             <!-- Run after serializer.normalizer.data_uri but before serializer.normalizer.object -->
-            <tag name="serializer.normalizer" priority="-923" />
+            <tag name="serializer.normalizer" priority="-900" />
         </service>
 
         <!-- Resources Operations path resolver -->


src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -104,7 +104,7 @@
             <argument>false</argument>
 
             <!-- Run after serializer.normalizer.data_uri but before serializer.normalizer.object -->
-            <tag name="serializer.normalizer" priority="-923" />
+            <tag name="serializer.normalizer" priority="-900" />
```
teohhanhui: Do you know which normalizer was being called instead?

JarJak: No. How can I check it?

JarJak: It breaks with priority -901

teohhanhui: I think it must be related to a normalizer from another package you're using.

teohhanhui: Oh, okay, I understand now. Your resource class implements JsonSerializable?

JarJak: Yes it does!

teohhanhui: What do you expect to happen in this case? If your class implements JsonSerializable, we should respect your desire to customize the output, no?

JarJak: I expect output to be the same as it it was before 2.4.0-beta2. I use \JsonSerializable in other places of my code, not to customize the API output.

teohhanhui: Let me rephrase that:

I know what you expect in this case. But how do we know what the user expects when they implement JsonSerializable? I think the current priority is more correct.

JarJak: I think that user at least expects that doing an API request for a single item won't return a collection to him.

JarJak: Maybe we can assume that if dev defines serialization groups and puts them into de/normalization_context, then he wants to use those groups instead of jsonSerialize()?

teohhanhui: We can't do that. The lower priority means our normalizer is not called if you implement JsonSerializable. That to me is expected behavior. It's not expected for a resource class to implement JsonSerializable.

As for why it returns a collection, I'm not really sure without taking a deeper look. We check for iterable in the CollectionNormalizer.

JarJak: > It's not expected for a resource class to implement JsonSerializable.

This should be mentioned in the docs.

soyuka:
Shouldn't we support `JsonSerializable`?

teohhanhui:
> Shouldn't we support JsonSerializable?

It's already supported by [serializer.normalizer.json_serializable](https://github.com/symfony/symfony/blob/v4.2.5/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.xml#L54-L57)

soyuka:
Thanks @teohhanhui  it's `-900` we need to know what normalizer got executed instead.

teohhanhui:
Yeah, https://github.com/api-platform/core/pull/2732#discussion_r275135011 is still a concern.

teohhanhui:
Okay, so the reason why it gets converted into a collection is because of this call here (together with our `CollectionNormalizer`): https://github.com/symfony/symfony/blob/v4.3.2/src/Symfony/Component/Serializer/Normalizer/JsonSerializableNormalizer.php#L41

teohhanhui:
Replaced by #2921

Issue #2736, opened 2019-04-16
Custom GraphQL mutations without ID not possible


lukasluecke:
SchemaBuilder contains the following code in `getResourceObjectTypeFields`: 
```
$idField = ['type' => GraphQLType::nonNull(GraphQLType::id())];

[...]

if (!$input || 'create' !== $mutationName) {
    $fields['id'] = $idField;
}
```

This makes it impossible to create custom mutations without an ID field (i.e. custom create mutations).

Even when using a custom denormalization context for the mutation, the ID field remains.

lukasluecke:
@alanpoulain Any suggestions / ideas?

alanpoulain:
You mean an ID field in the input type? Yes maybe it should be changed to an allow list instead of a block list.

lukasluecke:
Yes, I want a mutation that does not have an ID field in the input type.

You mean changing `'create' !== $mutationName` to (something equivalent to) `in_array($mutationName, ['update','delete'])`? That would certainly solve my issue.

alanpoulain:
Yes I mean that :slightly_smiling_face:

PR #2738, opened 2019-04-16
[GraphQL] Make id field optional for custom mutations

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2736
| License       | MIT
| Doc PR        | -

All other mutations (except `create`) get an optional id field to allow different kinds of custom mutations.

### DIFF
--- a/features/graphql/mutation.feature
+++ b/features/graphql/mutation.feature
@@ -459,6 +459,24 @@ Feature: GraphQL mutation support
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "data.sumDummyCustomMutation.dummyCustomMutation.result" should be equal to "8"
 
+  Scenario: Execute a custom mutation without ID
+    When I send the following GraphQL request:
+    """
+    mutation {
+      sumCreateDummyCustomMutation(input: {operandA: 4, operandB: 5}) {
+        dummyCustomMutation {
+          id
+          result
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.sumCreateDummyCustomMutation.dummyCustomMutation.id" should be equal to "/dummy_custom_mutations/2"
+    And the JSON node "data.sumCreateDummyCustomMutation.dummyCustomMutation.result" should be equal to "9"
+
   Scenario: Execute a not persisted custom mutation
     When I send the following GraphQL request:
     """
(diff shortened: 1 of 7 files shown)


lukasluecke:
@alanpoulain What do you think? (test failure seems to be unrelated)

Probably should be in 2.4 and not master, but I'll leave that up to you 🙈

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -516,8 +516,10 @@ private function getResourceObjectTypeFields(?string $resourceClass, ResourceMet
             return $fields;
         }
 
-        if (!$input || 'create' !== $mutationName) {
+        if (!$input || \in_array($mutationName, ['update', 'delete'], true)) {
             $fields['id'] = $idField;
+        } elseif ('create' !== $mutationName) {
+            $fields['id'] = ['type' => GraphQLType::id()];
```
alanpoulain: I don't understand the aim of this `elseif`.

alanpoulain:
I think we should have a Behat test for this feature.

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: The idea is to still generate the `id` field for all mutations except `create`, but making it optional. This will allow custom mutations to mutate the specified resource, if provided - but also allows custom mutations that do not need an existing resource.

alanpoulain: OK I see. Could you make it an `$optionalIdField` variable to be clearer?

lukasluecke:
I will try to add the test later today 🙂

src/GraphQl/Type/SchemaBuilder.php
lukasluecke: Makes sense. I will make the change together with the test later today 🙂

lukasluecke:
@alanpoulain I added a simple test for this change 🙂

lukasluecke:
@alanpoulain @dunglas I rebased this onto the newest changes again, could we maybe get this merged?

The tests seem to be failing for some unrelated reason.

alanpoulain:
I'm not sure about this one. Now you can define your own arguments for your custom mutation.
By default the ID is required but you can remove it if you want.

lukasluecke:
I don't think that the current possibilities for custom arguments are really "usable" for this, as I'd have to re-define all arguments just to get rid of the required ID. However if you think that the way forward would be better custom arguments, I'll try to work on that instead (might take a little longer though) 🙂

kacperjurak:
Hello, is it possible to merge this change? I really need overwrite create mutation in my project.

PR #2744, opened 2019-04-18
Fix HTTP status code when output class is set to null

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Also fix Behat test for when input class is set to `null`.

### DIFF
--- a/src/EventListener/SerializeListener.php
+++ b/src/EventListener/SerializeListener.php
@@ -61,15 +61,8 @@ final class SerializeListener
 
         $context = $this->serializerContextBuilder->createFromRequest($request, true, $attributes);
 
-        if (
-            (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class'])
-            ||
-            (
-                null === $controllerResult && isset($context['input']) && \array_key_exists('class', $context['input']) &&
-                null === $context['input']['class']
-            )
-        ) {
-            $event->setControllerResult('');
+        if (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class']) {
+            $event->setControllerResult(null);
 
             return;
         }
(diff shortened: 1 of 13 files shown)


src/EventListener/SerializeListener.php
```diff
@@ -61,15 +61,8 @@ public function onKernelView(GetResponseForControllerResultEvent $event): void
...
-        if (
-            (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class'])
-            ||
-            (
-                null === $controllerResult && isset($context['input']) && \array_key_exists('class', $context['input']) &&
-                null === $context['input']['class']
-            )
-        ) {
-            $event->setControllerResult('');
+        if (isset($context['output']) && \array_key_exists('class', $context['output']) && null === $context['output']['class']) {
```
antograssiot: maybe 

```suggestion
        if (\array_key_exists('class', $context['output'] ?? []) && null === $context['output']['class']) {
```

even if it is not directly linked to your change

src/Metadata/Resource/Factory/InputOutputResourceMetadataFactory.php
```diff
@@ -64,6 +64,15 @@ private function getTransformedOperations(array $operations, array $resourceAttr
 
             $operation['input'] = isset($operation['input']) ? $this->transformInputOutput($operation['input']) : $resourceAttributes['input'];
             $operation['output'] = isset($operation['output']) ? $this->transformInputOutput($operation['output']) : $resourceAttributes['output'];
+
+            if (
+                !isset($operation['status'])
+                && isset($operation['output'])
+                && \array_key_exists('class', $operation['output'])
```
antograssiot:
```suggestion
                && \array_key_exists('class', $operation['output'] ?? [])
```

and drop the isset before ?

soyuka: I'd be to keep isset, it's readable and doesn't harm no?

src/EventListener/SerializeListener.php
soyuka: I'd be to keep isset, it's readable and doesn't harm no?

src/Metadata/Resource/Factory/InputOutputResourceMetadataFactory.php
antograssiot: ok

soyuka:
Thanks @teohhanhui !

PR #2745, opened 2019-04-19
 Fix bug in WriteListener.php

slavo2:
Alternative to https://github.com/api-platform/core/pull/2283

When calling POST on a non ORM resource, tha call failed with 400: "Unable to generate an IRI for the item of type..."
on api-platform/core/src/Bridge/Symfony/Routing/IriConverter.php line 131

This is supposed to fix it, as it tries to get the IRI from inserted resource, not from original data (where it does not exist in a POST).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | I don't know, I did not run tests. 
| Fixed tickets | 
| License       | MIT


soyuka:
There was a recent fix on this file: https://github.com/api-platform/core/commit/431cf6f8dc9cb8bd2fdc08ba3017c8255638cb92 it has not been released yet though, is it related?

slavo2:
Yes, @soyuka thanks for pointing that.

PR #2749, opened 2019-04-19
Allow changing the location of Swagger UI

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/935
| License       | MIT
| Doc PR        | 

This PR re-enable [what's described in the doc](https://api-platform.com/docs/core/swagger/#changing-the-location-of-swagger-ui)

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -320,13 +320,15 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
         $loader->load('json_schema.xml');
         $loader->load('swagger.xml');
+        $loader->load('swagger-ui.xml');
 
-        if ($config['enable_swagger_ui'] || $config['enable_re_doc']) {
-            $loader->load('swagger-ui.xml');
-            $container->setParameter('api_platform.enable_swagger_ui', $config['enable_swagger_ui']);
-            $container->setParameter('api_platform.enable_re_doc', $config['enable_re_doc']);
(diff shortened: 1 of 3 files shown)


teohhanhui:
I don't think this is the right way to fix it.

teohhanhui:
We should provide a way to customize the path instead.

teohhanhui:
I don't find any route that uses `SwaggerUiAction`, so it's actually enabled by `SwaggerUiListener`, right?

teohhanhui:
I imagine we could add a configuration such as:

```yaml
api_platform:
    openapi:
        swagger_ui:
            enabled: true
            path: /docs
```

and deprecate `enable_swagger_ui`.

antograssiot:
Indeed the listener  is handling it by forcing the controller of `api_entrypoint`. I too think it can be improved but wanted first to restore the old/documented behavior

teohhanhui:
Note to self: This problem was not properly considered in the [JWT docs](https://github.com/api-platform/docs/blob/2.4/core/jwt.md).

kl3sk:
I'll just test this branch as explain on https://github.com/api-platform/core/issues/2753 and confirm that work for me

marcoreni:
Is this still broken? It should at least be specified inside the docs...

soyuka:
@teohhanhui is something preventing us to merge this?

antograssiot:
updated, rebased and targeting 2.5 now

soyuka:
merged in https://github.com/api-platform/core/pull/3241 thanks @antograssiot

PR #2751, opened 2019-04-22
Add support for name converter in filters

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2090 
| License       | MIT
| Doc PR        | 

Alternative to #2377, which I think is more complete and more accurate as some filters were not properly covered.

- [x] Boolean Filter
- [x] Date Filter
- [x] Exists Filter
- [x] Numeric Filter
- [x] Order Filter
- [x] Property Filter
- [x] Range Filter
- [x] Search Filter

The Group Filter is not supposed to be updated I think

I've added Behat tests against the schema as we do usually but also plain JSON content. I've made sure that the schemas are actually really validating what I'm looking for but at the end it is way more verbose, and I'm not sure it will be easier to maintain. I'll of course leave only a single version of the test, let me know what's your prefered one.

I've been checking and it works correctly with Swagger UI/ReDoc

TODO
- [ ] Add tests to check that it is working with GraphQL ⚠️ 

Thanks for your feedback

cc @bendavies @soullivaneuh


src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
```diff
@@ -56,6 +56,7 @@
             <argument type="service" id="api_platform.property_accessor" />
             <argument type="service" id="logger" on-invalid="ignore" />
             <argument key="$identifiersExtractor" type="service" id="api_platform.identifiers_extractor.cached" on-invalid="ignore" />
+            <argument key="$nameConverter" type="service" id="api_platform.name_converter" on-invalid="ignore" />
```
bendavies: i'm not sure about this introduction of named arguments, mixing with old style, but i see a precedent has been set already.

bendavies:
awesome work!

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
soyuka: We checked this and it's fine indeed!

src/Serializer/Filter/PropertyFilter.php
```diff
@@ -129,18 +134,42 @@ private function getProperties(array $properties, array $whitelist = null): arra
...
+    {
+        if (null === $this->nameConverter || !$properties) {
+            return $properties;
+        }
+
+        $result = [];
+        foreach ($properties as $key => $value) {
+            if (is_numeric($key)) {
+                $result[] = $this->denormalizePropertyName($value);
+                continue;
```
soyuka: Why a condition for this? should be fine if it's through the same path as below no?

antograssiot: the signature of the method expects a string `private function denormalizePropertyName(string $property): string` and I wanted to keep as close as possible of `getProperties()` so it remains unchainged when those values are merge using `array_merge_recursive` or the when the diff with he whitelist is done

src/Serializer/Filter/PropertyFilter.php
```diff
@@ -129,18 +134,42 @@ private function getProperties(array $properties, array $whitelist = null): arra
 
         foreach ($properties as $key => $value) {
             if (is_numeric($key)) {
-                if (\in_array($value, $whitelist, true)) {
-                    $result[] = $value;
+                if (\in_array($propertyName = $this->denormalizePropertyName($value), $whitelist, true)) {
```
antograssiot: note for self, need to check here that `$value` is a string if not done before so it would not blow with strict typing

src/Serializer/Filter/PropertyFilter.php
```diff
@@ -129,18 +134,42 @@ private function getProperties(array $properties, array $whitelist = null): arra
...
+    private function denormalizeProperties(array $properties): array
+    {
+        if (null === $this->nameConverter || !$properties) {
+            return $properties;
+        }
+
+        $result = [];
+        foreach ($properties as $key => $value) {
+            if (is_numeric($key)) {
+                $result[] = $this->denormalizePropertyName($value);
```
antograssiot: note for self, need to check here that `$value` is a string if not done before so it would not blow with strict typing

src/Serializer/Filter/PropertyFilter.php
soyuka: fair enough! Although, the name converter signature isn't typed so `denormalizePropertyName` could also accept anything

src/Serializer/Filter/PropertyFilter.php
soyuka: wouldn't it be easier to simplify the method `denormalizePropertyName` so that it accept anything? It'd be compliant with the name converter denormalize method and avoid these checks.

antograssiot: i'll do that ! the code will be simpler :+1:

antograssiot: updated, thanks for pointing that out

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
bendavies: i know it works, but it's just a little strange to mix the styles of argument definition :)

antograssiot:
Do not merge that yet, I've just remembered that there are no tests for GraphQL yet.
Are name converters already supported @alanpoulain for property names ?

soyuka:
@antograssiot could we merge (pls rebase :D)  this and do the graphql support in another pr?

antograssiot:
Sire I'll rebase and merge tonight or tomorrow. GraphQl support is already in another PR anyway

antograssiot:
GraphQL support will land with #2765

Issue #1282, opened 2017-07-24
Empty body on custom operations


soyuka:
1. Keep 400 status code on standard operations when an empty body is sent (not valid json)
2. Allow empty body on custom operations (POST/PUT/DELETE)

Ref: #1274  #782 and #805. #1786

dunglas:
It looks weird to me to not have a different behavior for custom and built in operations.

What about this one: allow empty requests for a all operations IF Content-Type is not set?

soyuka:
Only thing I know for sure is that sending an empty body on our operations should return a `400` status code because it's invalid (i.e. empty string is not valid json). 

Now, there are users who may want to use empty contents (on POST/PUT operations). Forcing them to not use a `Content-Type` to be able to do so looks bad for DX no?

gorghoa:
At least, it should be consistent between `PUT` and `POST` (not the case right know I reckon)

gorghoa:
> Now, there are users who may want to use empty contents (on POST/PUT operations). Forcing them to not use a Content-Type to be able to do so looks bad for DX no?

I have to admit that in my case, it rings a bell :+1:

gorghoa:
> What about this one: allow empty requests for a all operations IF Content-Type is not set?

@dunglas, does it even work regarding content-negotiation?

dunglas:
I'm 👎 to allow empty bodies if the content-type is set to JSON or JSON-LD. An empty string is not a valid JSON document (you can try with https://jsonlint.com/) but the client is advertising that the payload is valid JSON. To me it should generate a 400.

Validating the content-type is an OWASP best practice.

sroze:
> I'm 👎 to allow empty bodies if the content-type is set to JSON or JSON-LD. 

Agree. Though I would 👍 very very much without a `Content-Type`.

gorghoa:
Wokay! I have to verify to be sure that PUT currently allow empty body while content-type set, but if so, shouldn't it be the same for `PUT` then? (BC breaking…)

dunglas:
I don't think we allow it for PUT neither.

gorghoa:
> I don't think we allow it for PUT neither.

@dunglas , I was assuming PUT was allowed to skip deserialization of empty content based on https://github.com/api-platform/core/blob/master/src/EventListener/DeserializeListener.php#L55

I may very well be missing something though, I am not very comfortable with the codebase yet :)

gorghoa:
[...]

Issue #2731, opened 2019-04-13
Cannot validate values of type "NULL" automatically


norkunas:
When calling a collection post operation and if forgot to include a body then the api should handle this gracefully and return a 400 error instead of 500.

Currently it tries to validate "null" because controller result is null and symfony validator obviously doesn't like it.

Response from api:
```
{  
   "@context":"\/contexts\/Error",
   "@type":"hydra:Error",
   "hydra:title":"An error occurred",
   "hydra:description":"Cannot validate values of type \u0022NULL\u0022 automatically. Please provide a constraint."
}

```

soyuka:
Duplicate: https://github.com/api-platform/core/issues/1282

teohhanhui:
See https://github.com/api-platform/core/pull/2757

jsawicki-internetcompany:
I had the same problem, and solved it by disabling the validation with `validate=false` in the custom operation in entity. 

```
"my_custom_operation"={
 *              "input"=MyCustomInput::class,
 *              "method"="POST",
 *              "path"="/an_entity/something",
 *              "controller"=MyCustomAction::class,
 *              "validate"=false,
 *              "normalization_context"={
 *                  "groups"={
 *                      "groupA:read",
 *                  }
 *              },
 *              "denormalization_context"={
 *                  "groups"={
 *                      "groupB:update",
 *                  }
 *              },
 *              "validation_groups"={
 *                  "Default",
 *                  "groupB:update",
 *              }
 *          },
```

ismaellastlevel:
I had the same problem, because if forgot to return the data after managing it from controller. So I just had to put a return.
from:
`public function postComment($data){
        $this->commentHandler->handleAdding($data);
    }`
to:
`public function postComment($data){
        return $this->commentHandler->handleAdding($data);
    }`

PR #2757, opened 2019-04-23
Fix handling of empty request content by allowing to disable listeners per operation

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes?
| New feature?  | kinda?
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1282, #2731
| License       | MIT
| Doc PR        | TODO

The toggleable listeners are:
- ReadListener ("read")
- DeserializeListener ("deserialize")
- ValidateListener ("validate")
- WriteListener ("write")
- SerializeListener ("serialize")

Basically, we fix the handling of empty request content by **removing** explicit handling of empty request content (which was wrong, resulting in weird bugs such as #2731), but allowing the user to be in control instead.

TODO:

- [x] Fix tests
- [x] Add tests
- [ ] GraphQL support

### DIFF
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -91,6 +91,12 @@ Feature: Create-Retrieve-Update-Delete
     }
     """
 
+  Scenario: Create a resource with empty body
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/dummies"
+    Then the response status code should be 400
+    And the JSON node "hydra:description" should be equal to "Syntax error"
+
   Scenario: Get a not found exception
     When I send a "GET" request to "/dummies/42"
     Then the response status code should be 404
@@ -526,42 +532,8 @@ Feature: Create-Retrieve-Update-Delete
   Scenario: Update a resource with empty body
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "PUT" request to "/dummies/1"
-    Then the response status code should be 200
-    And the response should be in JSON
-    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
-    And the header "Content-Location" should be equal to "/dummies/1"
-    And the JSON should be equal to:
-    """
-    {
-      "@context": "/contexts/Dummy",
-      "@id": "/dummies/1",
-      "@type": "Dummy",
-      "description": null,
-      "dummy": null,
-      "dummyBoolean": null,
-      "dummyDate": "2018-12-01T13:12:00+00:00",
-      "dummyFloat": null,
-      "dummyPrice": null,
-      "relatedDummy": null,
-      "relatedDummies": [],
-      "jsonData": [
-        {
-          "key": "value1"
-        },
-        {
-          "key": "value2"
-        }
-      ],
-      "arrayData": [],
-      "name_converted": null,
-      "relatedOwnedDummy": null,
-      "relatedOwningDummy": null,
-      "id": 1,
-      "name": "A nice dummy",
-      "alias": null,
-      "foo": null
-    }
-    """
+    Then the response status code should be 400
+    And the JSON node "hydra:description" should be equal to "Syntax error"
 
   Scenario: Delete a resource
     When I send a "DELETE" request to "/dummies/1"
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -154,6 +154,7 @@
             <argument type="service" id="api_platform.subresource_data_provider" />
             <argument type="service" id="api_platform.serializer.context_builder" />
             <argument type="service" id="api_platform.identifier.converter" />
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
 
             <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="4" />
         </service>
@@ -161,7 +162,7 @@
         <service id="api_platform.listener.view.write" class="ApiPlatform\Core\EventListener\WriteListener">
             <argument type="service" id="api_platform.data_persister" />
             <argument type="service" id="api_platform.iri_converter" on-invalid="null" />
-            <argument type="service" id="api_platform.metadata.resource.metadata_factory" on-invalid="null" />
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
 
             <tag name="kernel.event_listener" event="kernel.view" method="onKernelView" priority="32" />
         </service>
@@ -170,6 +171,7 @@
             <argument type="service" id="api_platform.serializer" />
             <argument type="service" id="api_platform.serializer.context_builder" />
             <argument type="service" id="api_platform.formats_provider" />
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
 
             <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest" priority="2" />
         </service>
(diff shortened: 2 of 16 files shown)


soyuka:
Thanks to this you'd have the ability to remove some listeners (`@ApiResource(attributes={"read"=false})`) ?
What are the use cases? Why isn't it possible to send an empty body anymore?

teohhanhui:
> Thanks to this you'd have the ability to remove some listeners (@ApiResource(attributes={"read"=false})) ?

Yes.

> Why isn't it possible to send an empty body anymore?

I don't get what you mean. It's possible to send an empty body if you set `"deserialize"=false`

teohhanhui:
Well, we fix #1282 by doing the right thing out of the box, and allowing the user to disable specific listeners so that they're able to send empty body, as necessary.

teohhanhui:
But I'll add tests to show that.

soyuka:
> I don't get what you mean. It's possible to send an empty body if you set "deserialize"=false (and possibly "validate"=false, depending on the use case).

You removed a functional test, in a way a promise that this use case works :p. 

> Or in other words, no weird attempts by us to try to read the user's mind. see_no_evil

Couldn't we fix this by implementing the following table instead?

| content-type | body content | result | notes |
|-----------------|-----------------|---------|--------|
| not set         |  empty          |  process the request | |
| not set          | with data      | 406 | throw not acceptable: content-type is mandatory |
| set               |  empty          | may throw 400 if deserializer fails | an empty string is valid text but invalid json |
| set               | with data       | process the request ||

teohhanhui:
> You removed a functional test, in a way a promise that this use case works :p.

Because it's wrong.

> Couldn't we fix this by implementing the following table instead?

No. The interpretation depends on the content-type. We don't have enough information to make that decision. It's better to leave it up to the user. But we do the right thing by default, by leaving that up to the Serializer.

features/main/crud.feature
```diff
@@ -523,46 +523,6 @@ Feature: Create-Retrieve-Update-Delete
     }
     """
 
-  Scenario: Update a resource with empty body
```
dunglas: It's a BC break

src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -154,6 +154,7 @@
             <argument type="service" id="api_platform.subresource_data_provider" />
             <argument type="service" id="api_platform.serializer.context_builder" />
             <argument type="service" id="api_platform.identifier.converter" />
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" on-invalid="null" />
```
dunglas: `on-invalid="null"` is useless, isn't it?

src/Metadata/Resource/ToggleableOperationAttributeTrait.php
```diff
@@ -0,0 +1,35 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Resource;
+
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+
+trait ToggleableOperationAttributeTrait
```
dunglas:
```suggestion
/**
 * @internal
 */
trait ToggleableOperationAttributeTrait
```

dunglas:
Also, couldn't this replace the `_api_receive`, `_api_respond` and `_api_persist` request attributes?

features/main/crud.feature
teohhanhui: Yes, but it's a bug fix. The previous behavior was wrong. We can't fix this bug without... fixing it.

src/Bridge/Symfony/Bundle/Resources/config/api.xml
teohhanhui: Hmm... Indeed, `api_platform.metadata.resource.metadata_factory` must exist.

teohhanhui:
> Also, couldn't this replace the _api_receive, _api_respond and _api_persist request attributes?

Yes, but we'd need to add the same for:

- AddFormatListener ("negotiate")
- RespondListener ("send")

WDYT @dunglas?

teohhanhui:
The semantics would be of interest to the built-in events implementation. /cc @alanpoulain

teohhanhui:
> Also, couldn't this replace the _api_receive, _api_respond and _api_persist request attributes?

@dunglas Another difficulty is that `_api_respond` is also used in `SwaggerUiListener` and `ExceptionListener`. I'm not sure what we should do about that.

teohhanhui:
Could someone help implement the GraphQL support?

@alanpoulain? :smile:

teohhanhui:
@dunglas Could GraphQL support be added in a separate PR (in `master`)? My primary aim here is to fix a bug. It's only as a result of that that we end up introducing a new feature. :laughing:

features/main/crud.feature
```diff
@@ -91,6 +91,12 @@ Feature: Create-Retrieve-Update-Delete
     }
     """
 
+  Scenario: Create a resource with empty body
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "POST" request to "/dummies"
+    Then the response status code should be 400
+    And the JSON node "hydra:description" should be equal to 'Syntax error'
```
dunglas:
```suggestion
    And the JSON node "hydra:description" should be equal to "Syntax error"
```

features/main/crud.feature
```diff
@@ -526,42 +532,8 @@ Feature: Create-Retrieve-Update-Delete
...
-      "relatedOwnedDummy": null,
-      "relatedOwningDummy": null,
-      "id": 1,
-      "name": "A nice dummy",
-      "alias": null,
-      "foo": null
-    }
-    """
+    Then the response status code should be 400
+    And the JSON node "hydra:description" should be equal to 'Syntax error'
```
dunglas:
```suggestion
    And the JSON node "hydra:description" should be equal to "Syntax error"
```

soyuka:
This needs to be documented asap

borNfreee:
> This needs to be documented asap

Here I'm crying again. Sorry - we are trying to be up-to-date with new stable releases in quite a big project.

Our tests failing for 2.4.3, and I spent quite a bit of time looking to the changelog, commits between [2.4.2 and 2.4.3](https://github.com/api-platform/core/compare/v2.4.2...v2.4.3) and searching in 3 repos something similar to this issue to find that this feature is not documented (I'm not even saying that this is a BC break in patch version).

Could you please revise the way you guys deploy new features? If it's released, *please* make sure it's documented. Please make sure there is a note in the changelog.

dunglas:
@borNfreee I agree, I missed it wasn't targeting master. Maybe should we revert this patch and merge it only in master. Also, it introduces an inconsistency with GraphQL. WYDT @teohhanhui? Can I revert this patch safely?

borNfreee:
If you guys decide to revert it, please tag a `2.4.4` if that is possible ASAP, since now everyone who wants to update a patch version will have a problem

teohhanhui:
It's a bug fix. I don't see why it should be reverted.

But we should (I should) add the documentation.

dunglas:
It breaks existing applications that are just upgrading to a new patch version, it’s not acceptable.

teohhanhui:
It's only for custom operations which used an empty request body. Our previous hack introduced other bugs, such as throwing an unexpected and cryptic exception when the request body is empty.

There's no way to maintain compatibility with buggy behavior. So by definition a bug fix must break what depended on the buggy behavior.

teohhanhui:
I know this is overused, but...



On a more serious note, fixing bugs inevitably breaks apps that were depending on the buggy behavior. We could try to minimize that impact, but it should not dictate how we should / should not fix a bug.

dunglas:
We could at least agree to not merge bug fixes introducing BC breaks in patch releases. They should at least be in minor releases. This one is a true BC break, not just an edge case... because the old behavior was covered by a test.

It's too late to revert now, but could you add a note in the CHANGELOG explaining how to fix this BC break please?

teohhanhui:
All bug fixes are BC breaks. We have many test cases that are wrong. It's an unavoidable problem.

> could you add a note in the CHANGELOG

I'm on it.

teohhanhui:
See https://github.com/api-platform/core/pull/2815

Issue #2742, opened 2019-04-18
Ability to modify the vary header


smatyas:
In my use-case I have to implement an authenticated and cached API. Most of the endpoints' results should be cached per user, so the default setting like this works fine:

```yaml
vary: ['Content-Type', 'Authorization', 'Accept-Language']
```

But there are some endpoints that serve common, shared resources, so they shouldn't vary by `Authorization`.

The question is: is there any simple way to modify the `vary` header per operation currently? I know I can implement a listener to modify the response in these cases, but I was looking for something like the [Setting Custom HTTP Cache Headers](https://api-platform.com/docs/core/performance/#setting-custom-http-cache-headers) feature, but for the `vary` header.

If there is no such feature yet, I could help implementing something like what was done in #2288 if you find it useful.

teohhanhui:
Sounds good to me. It should be noted that the `Vary` header can be set multiple times, and that's perfectly valid.

soyuka:
It's not supported yet. Might be a good enhancement indeed.

teohhanhui:
Added in #2758

PR #2758, opened 2019-04-23
Ability to modify the Vary header

smatyas:
Besides of setting the default value of the `vary` header, it would be useful if one could also set/modify it by resource.

Real life use case:
I have to implement an authenticated and cached API. Most of the endpoints' results should be cached per user, so the default setting like this works fine:

```yaml
vary: ['Content-Type', 'Authorization', 'Accept-Language']
```

But there are some endpoints that serve common, shared resources, so they shouldn't vary by Authorization.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2742 
| License       | MIT
| Doc PR        | api-platform/docs#820

### DIFF
--- a/src/HttpCache/EventListener/AddHeadersListener.php
+++ b/src/HttpCache/EventListener/AddHeadersListener.php
@@ -70,9 +70,10 @@ final class AddHeadersListener
             $response->setMaxAge($maxAge);
         }
 
-        $defaultVary = $resourceCacheHeaders['vary'] ?? $this->vary;
-        if (null !== $defaultVary) {
-            $response->setVary(array_diff($defaultVary, $response->getVary()), false);
+        if (isset($resourceCacheHeaders['vary'])) {
+            $response->setVary($resourceCacheHeaders['vary']);
+        } elseif (null !== $this->vary) {
+            $response->setVary(array_diff($this->vary, $response->getVary()), false);
         }
 
         if (null !== ($sharedMaxAge = $resourceCacheHeaders['shared_max_age'] ?? $this->sharedMaxAge) && !$response->headers->hasCacheControlDirective('s-maxage')) {
--- a/src/HttpCache/EventListener/AddHeadersListener.php
+++ b/src/HttpCache/EventListener/AddHeadersListener.php
@@ -70,8 +70,9 @@ final class AddHeadersListener
             $response->setMaxAge($maxAge);
         }
 
-        if (null !== $this->vary) {
-            $response->setVary(array_diff($this->vary, $response->getVary()), false);
+        $defaultVary = $resourceCacheHeaders['vary'] ?? $this->vary;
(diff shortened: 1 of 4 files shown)


smatyas:
There are a couple of open questions here:

1. I used the same `cache_headers` attribute of the `ApiResource` class that is already can be used for setting the `max_age` and `shared_max_age`. I consider this to be a cache header as well, but let me know if you think differently.

2. The current solution solves the original problem only partially, and this is the main reason of the WIP status. Currently it can be used **only to add new vary headers** to the configured default ones, but **it cannot be used** in the other way, **to remove some of the defaults** .

It could be used to solve the original problem, but considering the following situation, it would be a lot of "custom setting" on the resources, which wouldn't make it really easier:

- having 50 different resources, and only 2 of them should be shared
- setting the default to ```vary: ['Content-Type', 'Accept-Language']```
- in this implementation I would have to configure 48 resources to add the ```vary: ['Authorization']``` as well

So we have a couple of options here:
- make the setting on the resource always **replace** the default one
- add an additional parameter/attribute that could drive the replace false/true behavior
- create different attributes inside the `cache_headers` i.e. `vary_add`, `vary_remove`, `vary_set`
- create another level of attribute in the `cache_headers`, stg. like this: ```attributes={"cache_headers"={"vary"={"add"={"Additional-Vary"}, "remove"={"Vary-To-Remove-1", "Vary-To-Remove-2"}}}```
- creating "cache groups" with different default settings
- ...

I bet there are a lot more reasonable choices here, any of them could be implemented. The question is which one should we choose and implement? It should be just intuitive, easy to use and seamlessly fit into the current naming conventions and structures. 

What's your opinion?

soyuka:
> make the setting on the resource always replace the default one

This looks good to me. I'm :-1: on the replace/add/remove things. If you have a case where you need more complex header handling, you could always override the listener.

dunglas:
Thanks @smatyas!

src/HttpCache/EventListener/AddHeadersListener.php
```diff
@@ -70,7 +70,9 @@ public function onKernelResponse(FilterResponseEvent $event): void
             $response->setMaxAge($maxAge);
         }
 
-        if (null !== $this->vary) {
+        if (isset($resourceCacheHeaders['vary'])) {
+            $response->setVary($resourceCacheHeaders['vary']);
```
teohhanhui: This might result in duplicate `Vary` headers. Which is why 2 lines below we use `array_diff`. We should refactor this code a little to something like:

```php
$vary = $resourceCacheHeaders['vary'] ?? $this->vary;
if (null !== $vary) {
    $response->setVary(array_diff($vary, $response->getVary()), false);
}
```

PR #2765, opened 2019-04-27
Add support for name converter in GraphQL

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | ?
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

That will also make filtering possible with #2751 

I could add unit tests if you feel it is needed and the idea/implementation is OK

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
@@ -48,6 +48,7 @@
             <argument type="service" id="api_platform.collection_data_provider" />
             <argument type="service" id="api_platform.subresource_data_provider" />
             <argument type="service" id="api_platform.graphql.serializer.context_builder" />
+            <argument>%api_platform.graphql.nesting_separator%</argument>
         </service>
 
         <service id="api_platform.graphql.resolver.stage.security" class="ApiPlatform\Core\GraphQl\Resolver\Stage\SecurityStage" public="false">
@@ -140,6 +141,8 @@
             <argument type="service" id="api_platform.graphql.resolver.factory.item_mutation" />
             <argument type="service" id="api_platform.filter_locator" />
             <argument type="service" id="api_platform.pagination" />
+            <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
+            <argument>%api_platform.graphql.nesting_separator%</argument>
         </service>
 
         <service id="api_platform.graphql.fields_builder_locator" class="Symfony\Component\DependencyInjection\ServiceLocator" public="false">
@@ -216,6 +219,7 @@
 
         <service id="api_platform.graphql.serializer.context_builder" class="ApiPlatform\Core\GraphQl\Serializer\SerializerContextBuilder" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
+            <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
         </service>
         <service id="ApiPlatform\Core\GraphQl\Serializer\SerializerContextBuilderInterface" alias="api_platform.graphql.serializer.context_builder" />
 
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -29,6 +29,7 @@ use GraphQL\Type\Definition\WrappingType;
 use Psr\Container\ContainerInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidTypeException;
 use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
 /**
  * Builds the GraphQL fields.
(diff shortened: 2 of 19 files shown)


alanpoulain:
Nice! For the unit tests I'm refactoring the schema builder, it will be a lot more easier to add them then. I hope to do the PR this weekend.

dunglas:
Can you rebase @antograssiot? It would be nice to have it in the next patch release.

antograssiot:
Yeah I'll finish this but it won't land in the patch release @dunglas, the support of name converters in filters have landed in master (#2751) and that's also the target here.
I just need to add a unit test and I'll push

antograssiot:
rebased and added on top of the recent refactoring.
ping @dunglas @alanpoulain

antograssiot:
This PHPStan error appear often recently on master, I'm really not sure why but it is definitely not related to this PR I think 🤔

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
```diff
@@ -15,6 +15,7 @@
             <argument type="service" id="serializer" />
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
             <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="ignore" />
+            <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
```
alanpoulain: I think we could add it before and remove the `on-invalid`. There is no BC policy for GraphQL and we have already done a lot of BC breaks in master anyway.

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
```diff
@@ -26,6 +27,7 @@
             <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="null" />
             <argument type="service" id="request_stack" />
             <argument>%api_platform.collection.pagination.enabled%</argument>
+            <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
```
alanpoulain: Same for all the services.

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
antograssiot: Actually there's not always a name_converter in every app, that's why it is nullable @alanpoulain, not really for BC reason bit I might be wrong...

alanpoulain: OK. Could you just change its position then (before a parameter for instance)?

antograssiot: Well optional parameters are always supposed to be the last ones, and parameter are not optional unless I'm wrong @alanpoulain ?

alanpoulain: Because you are adding `on-invalid="ignore"`, it will be replaced by `null` if the service doesn't exist.
IMO we have to use an optional parameter at the last and with a default value only for BC policy: we don't want a decorator to break for instance.

antograssiot:
I guess this one is ready.

I've added support for nested filtering even when using name converters like done in #2897

To do so I had to create a new configuration value because the way we support nested properties (using the underscore (_) syntax instead of the dot (.) syntax ) is problematic when the property name contains underscore(s). To avoid breaking BC and to allow users that need this feature, I've introduced the possibility to choose the separator that should be used using a new configuration option.

I don't think that it is worth deprecating using the actual value (`_`) which i the default configuration value as I think that users using such name converters and graphql are not a lot (at least for now), that's really an edge case I guess. Of course if you agree, I'll open a doc PR to explain all of this.
ping @api-platform/core-team

dunglas:
I suggest to I lu support the dot. Underscore is weird, and was not intended imho. Let’s not complexify the code, BC breaks are allowed for GraphQL.

antograssiot:
> I suggest to support the dot. 

@dunglas, as far as I remember the dot is not allowed in GraphQl queries, that's why underscores where used in the first place, this PR didn't changed that. ping @alanpoulain 
If it is now allowed, then it will be easy to support it, we just need to drop the usage of `parse_str()` in the `FieldsBuilder`

alanpoulain:
Yes the dot is not allowed in names: https://github.com/graphql/graphql-spec/blob/master/spec/Section%202%20--%20Language.md#names

src/GraphQl/Resolver/FieldsToAttributesTrait.php
```diff
@@ -34,15 +34,17 @@ private function fieldsToAttributes(ResolveInfo $info): array
...
+                continue;
             }
+            $denormalizedFields[$this->denormalizePropertyName($key)] = \is_array($fields[$key]) ? $this->replaceIdKeys($fields[$key]) : $value;
         }
 
-        return $fields;
+        return $denormalizedFields;
     }
+
+    abstract protected function denormalizePropertyName($property);
```
alanpoulain: Why not be stricter?

```suggestion
    abstract protected function denormalizePropertyName(string $property): string;
```

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -329,7 +337,7 @@ private function convertFilterArgsToTypes(array $args): array
         foreach ($args as $key => $value) {
             if (strpos($key, '.')) {
                 // Declare relations/nested fields in a GraphQL compatible syntax.
-                $args[str_replace('.', '_', $key)] = $value;
+                $args[str_replace('.', '__', $key)] = $value;
```
alanpoulain: Why not using the nesting operator twice here?

antograssiot: That's a leftover, I'll update

src/GraphQl/Resolver/FieldsToAttributesTrait.php
antograssiot: The idea was both to respect the underlying Symfony NameConverterInterface and to be in line with what has been done in ORM filters ([see original discussion](https://github.com/api-platform/core/pull/2751#discussion_r277631577))
I'm 100% ok to update if those args are not worth it, let me know

alanpoulain: Personally, I would prefer a stricter approach. I think Symfony will change its interface soon too.

Simperfit: @alanpoulain is right the interfaces will be changed in 5.0

alanpoulain:
Thanks a lot @antograssiot! It would be great if you could do the doc PR too :slightly_smiling_face:

Issue #2752, opened 2019-04-22
Unable to generate an IRI for resource with custom route define


ghanbari:
I do define this DTO:

```
* @ApiResource(
 *     collectionOperations={},
 *     itemOperations={
 *          "get"={
 *              "controller"=GetDomainStatisticsAction::class,
 *              "path"="domains/{domain}/statistics",
 *              "defaults"={"_api_receive"=false},
 *          },
 *     }
 * )
 *
 * @ApiFilter(DateFilter::class, properties={"date"})
 * @ApiFilter(PropertyFilter::class, arguments={"parameterName": "properties", "overrideDefaultProperties": false})
 */
final class DomainStatistics
{
    /**
     * @var int
     *
     * @ApiProperty(identifier=true)
     */
    public $domain;
```

but i get this error:

```
Unable to generate an IRI for the item of type "App\Api\Dto\DomainStatistics"

Symfony\Component\Routing\Exception\MissingMandatoryParametersException:
Some mandatory parameters are missing ("domain") to generate a URL for route "api_domain_statistics_get_item".

  at vendor/symfony/routing/Generator/UrlGenerator.php:151
  at Symfony\Component\Routing\Generator\UrlGenerator->doGenerate(array('domain' => 0), array('_controller' => 'App\\Controller\\GetDomainStatisticsAction', '_format' => null, '_api_resource_class' => 'App\\Api\\Dto\\DomainStatistics', '_api_item_operation_name' => 'get', '_api_receive' => false), array(), array(array('text', '/statistics'), array('variable', '/', '[^/]++', 'domain', true), array('text', '/api/domains')), array('id' => '10'), 'api_domain_statistics_get_item', 1, array(), array())
     (var/cache/dev/srcApp_KernelDevDebugContainerUrlGenerator.php:142)
  at srcApp_KernelDevDebugContainerUrlGenerator->generate('api_domain_statistics_get_item', array('id' => '10'), 1)
     (vendor/symfony/routing/Router.php:240)
  at Symfony\Component\Routing\Router->generate('api_domain_statistics_get_item', array('id' => '10'), 1)
     (vendor/api-platform/core/src/Bridge/Symfony/Routing/Router.php:101)
  at ApiPlatform\Core\Bridge\Symfony\Routing\Router->generate('api_domain_statistics_get_item', array('id' => '10'), 1)
     (vendor/api-platform/core/src/Bridge/Symfony/Routing/IriConverter.php:124)
[...]

PR #2766, opened 2019-04-27
use current route params for generating IRI

ghanbari:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #2752
| License       | MIT

we can define custom path (`domains/{domainId}/reports/{date}`) for custom controllers.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -61,6 +61,7 @@
             <argument type="service" id="api_platform.identifiers_extractor.cached" />
             <argument type="service" id="api_platform.subresource_data_provider" on-invalid="ignore" />
             <argument type="service" id="api_platform.identifier.converter" on-invalid="ignore" />
+            <argument type="service" id="request_stack" />
         </service>
         <service id="ApiPlatform\Core\Api\IriConverterInterface" alias="api_platform.iri_converter" />
 
--- a/src/Bridge/Symfony/Routing/IriConverter.php
+++ b/src/Bridge/Symfony/Routing/IriConverter.php
@@ -30,6 +30,7 @@ use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Util\AttributesExtractor;
 use ApiPlatform\Core\Util\ClassInfoTrait;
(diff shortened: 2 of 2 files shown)


src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -61,6 +61,7 @@
             <argument type="service" id="api_platform.identifiers_extractor.cached" />
             <argument type="service" id="api_platform.subresource_data_provider" on-invalid="ignore" />
             <argument type="service" id="api_platform.identifier.converter" on-invalid="ignore" />
+            <argument type="service" id="request_stack" />
```
teohhanhui: I don't think we want to have a dependency on the `RequestStack` here. We're aiming for support of other frameworks.

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -120,8 +123,20 @@ public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface
 
         try {
             $identifiers = $this->generateIdentifiersUrl($this->identifiersExtractor->getIdentifiersFromItem($item), $resourceClass);
+            $currentRouteParams = $this->request->attributes->get('_route_params');
+            $pathVariables = $route = $this->router->getRouteCollection()->get($routeName)->compile()->getPathVariables();
```
teohhanhui: This is very bad for performance.

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -120,8 +123,20 @@ public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface
...
+            if ($resourceClass === $this->request->attributes->get('_api_resource_class')) {
```
teohhanhui: You cannot assume that the current route is the same as the item route we got, just because the resouce class is the same.

ghanbari: I know, we need check resource class & operation, i assume get operation as master item operation that put & delete have equal path and route params

src/Bridge/Symfony/Routing/IriConverter.php
ghanbari: Yes you are right, is compiled route not cached?

soyuka:
Couldn't you resolve this by decorating the IriConverter in your project instead? I'm not sure that we would be able to merge this as it's a hot spot.

PR #2772, opened 2019-05-01
[GraphQL] Services Refactoring

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (experimental)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

A big improvement of the GraphQL system with more services having consistent responsibilities.
With this refactoring, the system will be more maintainable, well tested and customizable :rocket:

Featuring:
- a new `TypeConverter` service to manage types. It can be decorated to manage custom types. An example is done in the tests.
- a `TypesContainer` service containing all the loaded types and respecting the PSR-11.
- all unit tests rewritten to cover the majority of cases for each service.

In order to do so, it was mandatory to break the circular dependencies of the services induced by the recursive way the system is working.
That's why a `fields_builder_locator` is used in the `TypeBuilder` service (used to construct an object type because it needs the corresponding fields).

@dunglas @lukasluecke I would like your opinion on this PR if you are willing to.

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -1,0 +1,371 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\GraphQl\Type;
+
+use ApiPlatform\Core\GraphQl\Resolver\Factory\ResolverFactoryInterface;
+use ApiPlatform\Core\GraphQl\Type\Definition\TypeInterface;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use Doctrine\Common\Inflector\Inflector;
+use GraphQL\Type\Definition\InputObjectType;
+use GraphQL\Type\Definition\Type as GraphQLType;
+use GraphQL\Type\Definition\WrappingType;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidTypeException;
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * Builds the GraphQL fields.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class FieldsBuilder implements FieldsBuilderInterface
+{
+    private $propertyNameCollectionFactory;
+    private $propertyMetadataFactory;
+    private $resourceMetadataFactory;
+    private $typesContainer;
+    private $typeBuilder;
+    private $typeConverter;
+    private $itemResolverFactory;
+    private $collectionResolverFactory;
+    private $itemMutationResolverFactory;
+    private $filterLocator;
+    private $paginationEnabled;
+
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, TypesContainerInterface $typesContainer, TypeBuilderInterface $typeBuilder, TypeConverterInterface $typeConverter, ResolverFactoryInterface $itemResolverFactory, ResolverFactoryInterface $collectionResolverFactory, ResolverFactoryInterface $itemMutationResolverFactory, ContainerInterface $filterLocator, bool $paginationEnabled)
+    {
+        $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
+        $this->propertyMetadataFactory = $propertyMetadataFactory;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->typesContainer = $typesContainer;
+        $this->typeBuilder = $typeBuilder;
+        $this->typeConverter = $typeConverter;
+        $this->itemResolverFactory = $itemResolverFactory;
+        $this->collectionResolverFactory = $collectionResolverFactory;
+        $this->itemMutationResolverFactory = $itemMutationResolverFactory;
+        $this->filterLocator = $filterLocator;
+        $this->paginationEnabled = $paginationEnabled;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getNodeQueryFields(): array
+    {
+        return [
+            'type' => $this->typeBuilder->getNodeInterface(),
+            'args' => [
+                'id' => ['type' => GraphQLType::nonNull(GraphQLType::id())],
+            ],
+            'resolve' => ($this->itemResolverFactory)(),
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getQueryFields(string $resourceClass, ResourceMetadata $resourceMetadata, string $queryName, $itemConfiguration, $collectionConfiguration): array
+    {
+        $queryFields = [];
+        $shortName = $resourceMetadata->getShortName();
+        $fieldName = lcfirst('query' === $queryName ? $shortName : $queryName.$shortName);
+
+        $deprecationReason = $resourceMetadata->getGraphqlAttribute($queryName, 'deprecation_reason', '', true);
+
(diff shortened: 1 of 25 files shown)


lukasluecke:
Thank you for your work @alanpoulain! So far it looks promising, will give you a detailed opinion by Friday 🙂

lukasluecke:
I thinks it's a good first step to facilitate future changes and improve maintainability.

I'll use this code as the base for my other GraphQL issues now, so would be nice if this could be merged in time 😉

src/GraphQl/Type/TypesContainer.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $this->graphqlTypes[$id] = $type;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function get($id): GraphQLType
+    {
+        if (!$this->has($id)) {
```
soyuka:
```suggestion
        if ($this->has($id)) {
            return $this->graphqlTypes[$id];
        }
        
        throw ...
```

tests/Fixtures/TestBundle/GraphQl/Type/Definition/DateTimeType.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function serialize($value)
+    {
+        // Already serialized.
+        if (\is_string($value)) {
+            // Should be better in a custom normalizer.
+            return (new \DateTime($value))->format('Y-m-d');
```
soyuka: Definitely, shouldn't you inject the DateTimeNormalizer?

tests/GraphQl/Type/TypesContainerTest.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->assertSame(['test' => $type], $this->typesContainer->all());
+    }
+
+    public function testHas(): void
+    {
+        $type = $this->prophesize(GraphQLType::class)->reveal();
+
+        $this->typesContainer->set('test', $type);
+
+        $this->assertTrue($this->typesContainer->has('test'));
```
soyuka: Just aesthetics but we could remove empty lines in these functions.

tests/Fixtures/TestBundle/GraphQl/Type/Definition/DateTimeType.php
alanpoulain: Why? The comment is here to warn the user that using a custom normalizer is a better idea that using a custom type.
But I need to test the custom type to. I you have a better example for using a custom type, you can tell me because this one is not really great.

soyuka: Oh I understand now thanks. I'd have used UUIDs as custom types for example but it's not a big deal.

alanpoulain: Is there something to manage with UUID? Because IMO it's just a string doesn't it? Maybe just the validation?

soyuka: no you could transform it into a Ramsey\Uuid object?

alanpoulain: Hmm I see. I could try.

src/GraphQl/Type/TypesContainerInterface.php
```diff
@@ -0,0 +1,62 @@
...
+use Psr\Container\ContainerInterface;
+
+/**
+ * Interface implemented to contain the GraphQL types.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+interface TypesContainerInterface extends ContainerInterface
```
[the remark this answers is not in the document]
alanpoulain: I was inspired by what Symfony has done for its `ContainerInterface`: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/ContainerInterface.php
I think it's good to follow the PSR to be able to reuse another existing implementation but having stricter methods is also important to make correct assumptions in the code.

tests/Fixtures/TestBundle/GraphQl/Type/Definition/DateTimeType.php
alanpoulain: It seems there are other issues with Ramsey UUID (`IdentifierConverter` is never called) so I'm keeping this test for now.

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+            }
+
+            $graphqlType = $this->typesContainer->get($graphqlType);
+        }
+
+        if ($this->typeBuilder->isCollection($type)) {
+            return $this->paginationEnabled && !$input ? $this->typeBuilder->getResourcePaginatedCollectionType($graphqlType) : GraphQLType::listOf($graphqlType);
+        }
+
+        return $type->isNullable() || (null !== $mutationName && 'update' === $mutationName) ? $graphqlType : GraphQLType::nonNull($graphqlType);
```
oleg-brizy, line 369: Why not just `return $graphqlType;`?

I did a fork because of this strange condition (to remove it).

oleg-brizy, line 369: Because of this condition I can't customize GraphQLlType for class/entity properties.

Even if I decorate TypeConverter ([called above at line 351](https://github.com/alanpoulain/core/blob/6e13e6628126027cfe3bbc60b3d743669ccafcd1/src/GraphQl/Type/FieldsBuilder.php#L351)) and return exactly the type I need, it is forced to `nonNull` by this condition (even if I don't need `nonNull`) and I can't do nothing about it (because it's inside a private method).

The only solution was to fork API Platform and [replace this condition with just `return $graphqlType;`](https://github.com/bagrinsergiu/api-platform-core/commit/17ad2ad1c83fe3a1e22226f5041a8d0429dcbf77). Also this condition can't be removed because it would introduce breaking changes for existing projects _(all auto generated entity property types will be [`String` instead of `String!`](https://github.com/alanpoulain/core/blob/6e13e6628126027cfe3bbc60b3d743669ccafcd1/tests/GraphQl/Type/FieldsBuilderTest.php#L352-L361))_.

I would like this condition to be moved in a public method which is part of interface, so I can decorate this service and change method behavior to `return $graphqlType;`.

alanpoulain: If your property type is nullable (for instance with typehint or PHPDoc), the type is not wrapped. Which is your use case?

oleg-brizy: 1. I created an annotation where I specify explicitly my type.
2. I created a compiler pass to read all classes, extract the annotated properties and cache the result as a service argument.
3. I decorated TypeConverter to return the annotated explicit class property type by using the cache from previous step.
    ```php
    <?php
    
    declare(strict_types=1);
    
    namespace App\Type\GraphQL;
    
    use ApiPlatform\Core\GraphQl\Type\TypeConverterInterface;
    use App\Service\GraphQL\GraphQLTypeAnnotationCache;
    use GraphQL\Type\Definition\Type as GraphQLType;
    
    class TypeConverter implements TypeConverterInterface
    {
        private TypeConverterInterface $decorated;
        private GraphQLTypeAnnotationCache $annotationCache;
    
        public function __construct(
            TypeConverterInterface $decorated,
            GraphQLTypeAnnotationCache $annotationCache
        ) {
            $this->annotationCache = $annotationCache;
        }
    
        public function convertType(
            Type $type,
            bool $input,
            ?string $queryName,
            ?string $mutationName,
            ?string $subscriptionName,
            string $resourceClass,
            string $rootResource,
            ?string $property,
            int $depth
        ) {
            // required to be called before annotation cache getter
            //   because entity types must be defined before use
            $defaultType = $this->decorated->convertType(
                $type,
                $input,
                $queryName,
                $mutationName,
                $subscriptionName,
                $rootResource,
                $rootResource,
                $property,
                $depth
            );
    
            if ($property && ($customTypeName = $this->annotationCache->getGraphQLType($rootResource, $property))) {
                return $this->resolveType($customTypeName);
            }
    
            return $defaultType;
        }

        public function resolveType(string $type): ?GraphQLType
        {
            return $this->decorated->resolveType($type);
        }
    }
    ```

4. But this condition messes/changes the type I returned from TypeConverter and I always get `nonNull`.

----

Examples of annotation:

```php
// in Entity
// CmsApplicationStatus is enum

/**
 * @var string
 * @GraphQLType("CmsApplicationStatus")
 */
public $status;
```

```php
// in Dto
// CollectionTypeSettingsInput is custom type

/**
 * @var array
 * @GraphQLType("CollectionTypeSettingsInput")
 */
public $settings = [];
```

alanpoulain: This is really an edge case. Why not doing this:
```php
// in Entity
// CmsApplicationStatus is enum

/**
 * @var null|string
 * @GraphQLType("CmsApplicationStatus")
 */
public $status;
```

oleg-brizy: Because it's a hack - I have to keep in mind that `@var` affects `@GraphQLType` and they must be used together. I also have cases where I don't have `@var` on property.

Ok, if it's an edge case then I will continue using my fork. Thanks

PR #2779, opened 2019-05-07
Add the original data to attributes and allow to access it in ExpressionLanguage

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1646
| License       | MIT
| Doc PR        | TODO

Alternative approach to #2710 as discussed there.

The good thing about this one is that it is totally BC.
The disadvantage is that it might be harder to discover even if covered by documentation and users will not be "protected" out of the box.

### DIFF
--- a/src/EventListener/ReadListener.php
+++ b/src/EventListener/ReadListener.php
@@ -107,5 +107,6 @@ final class ReadListener
         }
 
         $request->attributes->set('data', $data);
+        $request->attributes->set('previous_data', \is_object($data) ? clone $data : $data);
     }
 }
--- a/tests/Fixtures/TestBundle/Document/SecuredDummy.php
+++ b/tests/Fixtures/TestBundle/Document/SecuredDummy.php
@@ -30,7 +30,8 @@ use Symfony\Component\Validator\Constraints as Assert;
  *         "post"={"access_control"="has_role('ROLE_ADMIN')"}
  *     },
  *     itemOperations={
- *         "get"={"access_control"="has_role('ROLE_USER') and object.getOwner() == user"}
+ *         "get"={"access_control"="has_role('ROLE_USER') and object.getOwner() == user"},
+ *         "put"={"access_control"="has_role('ROLE_USER') and previous_object.getOwner() == user"},
  *     },
  *     graphql={
  *         "query"={},
--- a/features/authorization/deny.feature
+++ b/features/authorization/deny.feature
@@ -58,15 +58,39 @@ Feature: Authorization checking
     """
     Then the response status code should be 201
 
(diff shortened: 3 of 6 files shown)


src/EventListener/ReadListener.php
```diff
@@ -107,5 +107,6 @@ public function onKernelRequest(GetResponseEvent $event): void
         }
 
         $request->attributes->set('data', $data);
+        $request->attributes->set('originalData', is_object($data) ? clone $data : $data);
```
dunglas: Should be `origina_data` according to Symfony convention IIRC.

tests/Fixtures/TestBundle/Document/SecuredDummy.php
```diff
@@ -30,7 +30,8 @@
  *         "post"={"access_control"="has_role('ROLE_ADMIN')"}
  *     },
  *     itemOperations={
- *         "get"={"access_control"="has_role('ROLE_USER') and object.getOwner() == user"}
+ *         "get"={"access_control"="has_role('ROLE_USER') and object.getOwner() == user"},
+ *         "put"={"access_control"="has_role('ROLE_USER') and original_object.getOwner() == user"},
```
dunglas: May I suggest to test both?

```suggestion
 *         "put"={"access_control"="has_role('ROLE_USER') and object.getOwner() == user and  original_object.getOwner() == user"},
```

dunglas:
Great one! I'm just a bit unsure about the naming. React use `prevState`, what about `previous_data`?

tests/Fixtures/TestBundle/Document/SecuredDummy.php
antograssiot: Testing both isn't ok @dunglas as we are trying to make sure that only the actual owner can update the owner to become someone else. This someone shouldn't be be able to assign the object to himself

antograssiot: Well I can check owner for the "get" operation if you prefer, they are the same in this case...

teohhanhui:
@antograssiot What happened with your last force push? Looks like it's using the wrong branch. This should target `master`.

antograssiot:
@teohhanhui it already targets master and the diff looks good to me 🤔
What does looks wrong to you ?

teohhanhui:
@antograssiot Oh, never mind. I was confused by the force push diff. :laughing:

teohhanhui:
Thanks @antograssiot! :tada:

antograssiot:
I'll send a doc PR

PR #2780, opened 2019-05-08
Make function protected

bugsfunny:
Can we pass function to protected for inheritance capacity

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/Bridge/Doctrine/Common/Filter/RangeFilterTrait.php
+++ b/src/Bridge/Doctrine/Common/Filter/RangeFilterTrait.php
@@ -76,7 +76,7 @@ trait RangeFilterTrait
         ];
     }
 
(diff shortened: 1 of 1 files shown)


dunglas:
Hi, and thanks for contributing.

As it's in a trait, you can already change the visibility of the method when importing it in your own class: https://www.php.net/manual/en/language.oop5.traits.php#example-217

Most classes of the projects are `final` and methods are `private` because opening them to inheritance make the maintenance way harder, and evolutions difficult. The recommended way to change the behaviors of classes provided by API Platform is to use composition instead of inheritance (for instance using the decorator pattern).

PR #2784, opened 2019-05-10
Compatiblity with Symfony 4.3

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a


features/jsonld/context.feature
```diff
@@ -41,7 +41,6 @@ Feature: JSON-LD contexts generation
             "jsonData": "Dummy/jsonData",
             "arrayData": "Dummy/arrayData",
             "nameConverted": "Dummy/nameConverted",
-            "id": "Dummy/id",
```
dunglas: Why do we need to change this @soyuka? It looks like a BC break.

alanpoulain: I've done this change. It's because of you :stuck_out_tongue:
https://github.com/symfony/symfony/pull/31486

alanpoulain: That's also why I've added what you've done for MongoDB too: https://github.com/api-platform/core/pull/2784/files#diff-c5fed7f9ec0eb28b0039798760fac3ac

dunglas: Got it!

dunglas:
Thank you very much @soyuka and @alanpoulain for the help!

composer.json
```diff
@@ -71,10 +71,10 @@
         "symfony/expression-language": "^3.4 || ^4.0",
         "symfony/finder": "^3.4 || ^4.0",
         "symfony/form": "^3.4 || ^4.0",
-        "symfony/framework-bundle": "^4.2",
+        "symfony/framework-bundle": "^4.3",
         "symfony/mercure-bundle": "*",
-        "symfony/messenger": "^4.2",
-        "symfony/phpunit-bridge": "^4.3",
+        "symfony/messenger": "^4.3",
```
teohhanhui: Is it a good idea to bump the dependencies in a patch version? It means that our (potential) future security fixes will not reach the users who are stuck on `<=2.4.4`

teohhanhui: What this essentially means:

You must upgrade Symfony Messenger to continue to receive security fixes for 2.4

src/Serializer/ItemNormalizer.php
```diff
@@ -30,8 +30,6 @@
 /**
  * Generic item normalizer.
  *
- * @final
```
teohhanhui: I don't think this is right. The GraphQL ItemNormalizer should stop extending this class instead.

tests/Fixtures/app/AppKernel.php
```diff
@@ -96,10 +97,11 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
 
         $loader->load(__DIR__."/config/config_{$this->getEnvironment()}.yml");
 
+        $alg = class_exists(SodiumPasswordEncoder::class) && SodiumPasswordEncoder::isSupported() ? 'auto' : 'bcrypt';
```
teohhanhui: Uhh... Symfony really screwed this one up. :x

teohhanhui:
It seems like this is not addressed? https://github.com/api-platform/core/issues/2856

composer.json
alanpoulain: It's not for a patch version but a minor one. I think it's OK because this component is experimental in Symfony anyway.

teohhanhui: Uhh... Didn't notice this was targeting `master` 🙈

soyuka:
I'm not sure to understand how #2856 is an issue here...

teohhanhui:
@soyuka No, it's not an issue here. Just that I'd consider #2856 essential for Symfony 4.3 compatibility. Because it's not doing the right thing now.

tests/Fixtures/app/AppKernel.php
maxhelias: It's not bcrypt if sodium isn't supported but the native encoder. I do not know if it's important here.
The bcrypt encoder is now deprecated

dunglas: It's only for old versions so I think we don't care (this code will go be thrown away at some point).

soyuka:
True although I'm not sure to get what we need to do on our side.

bastnic:
Why only master and not 2.4 branch ? My tests are full of `  59403x: Method "ApiPlatform\Core\Serializer\ItemNormalizer::createChildContext()" will have a third "?string $format" argument in version 5.0; not defining it is deprecated since Symfony 4.3.`

dunglas:
Because Messenger. Messenger 4.3 introduces some BC breaks, that in cascade force us to add some BC breaks too. Also, to use Messenger 4.3 we had to bump the version of several other Symfony components. We then decided that it wasn't a good idea to introduce such changes in a patch release.
Anyway, we're working hard on finishing 2.5, a beta should be available soon.

teohhanhui:
We could certainly backport the other changes that are not related to Symfony Messenger.

dunglas:
Maybe some of them yes! Feel free to open a PR!

src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -88,6 +88,6 @@ public function persist($data, array $context = [])
      */
     public function remove($data, array $context = [])
     {
-        $this->messageBus->dispatch(new Envelope($data, new RemoveStamp()));
+        $this->messageBus->dispatch(new Envelope($data, [new RemoveStamp()]));
```
teohhanhui:
```php
$this->messageBus->dispatch(
    (new Envelope($data))
        ->with(new RemoveStamp())
);
```

teohhanhui: I'm not sure why it's still possible to pass in the stamps through the constructor...

features/graphql/mutation.feature
```diff
@@ -324,7 +324,7 @@ Feature: GraphQL mutation support
     When I send the following GraphQL request:
     """
     mutation {
-      createDummy(input: {_id: 12, name: "", foo: [], clientMutationId: "myId"}) {
```
teohhanhui: @dunglas @alanpoulain Do you know what's going on here?

PR #2785, opened 2019-05-10
Allow to set HTTP STATUS CODE in GetResponseForControllerResultEvent dynamically

Arkowsky:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT
| Doc PR        | todo

This feature allow to set status from controller event. Some resource can return 200 or 201 code.

### DIFF
--- a/src/EventListener/RespondListener.php
+++ b/src/EventListener/RespondListener.php
@@ -79,6 +79,10 @@ final class RespondListener
             }
 
             $status = $resourceMetadata->getOperationAttribute($attributes, 'status');
+
+            if ('request_attributes_status' === $status) {
+                $status = $request->attributes->get($status);
+            }
         }
 
         $event->setResponse(new Response(
--- a/tests/EventListener/RespondListenerTest.php
+++ b/tests/EventListener/RespondListenerTest.php
@@ -181,6 +181,30 @@ class RespondListenerTest extends TestCase
         $this->assertSame(Response::HTTP_ACCEPTED, $event->getResponse()->getStatusCode());
     }
 
+    public function testSetCustomStatusDynamicallyFromResultEvent()
+    {
+        $kernelProphecy = $this->prophesize(HttpKernelInterface::class);
+
+        $request = new Request([], [], ['_api_resource_class' => Dummy::class, '_api_item_operation_name' => 'get', '_api_respond' => true]);
+
+        $event = new GetResponseForControllerResultEvent(
+            $kernelProphecy->reveal(),
+            $request,
+            HttpKernelInterface::MASTER_REQUEST,
+            'bar'
+        );
+        $eventRequest = $event->getRequest();
+        $eventRequest->attributes->set('request_attributes_status', Response::HTTP_CREATED);
+
+        $resourceMetadataFactoryProphecy = $this->prophesize(ResourceMetadataFactoryInterface::class);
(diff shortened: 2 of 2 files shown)


dunglas:
You can already set the status code with an operation attribute as illustrated here: https://api-platform.com/docs/core/messenger/

Arkowsky:
Thanks @dunglas from quick reply but this is about dynamically set status from your event.
Below draft from api_platform/docs I'm preparing:

### Customize the HTTP status code to send

#### Set HTTP status code statically

You can customize status code for successful operation when you set for selected operation `status: 200`.
This could be useful when you use POST for different action than create resource.

```yaml
# api/config/api_platform/resources.yaml
App\Entity\Book:
    itemOperations:
        post:
            ...
            status: 200 # customize the HTTP status code to send
```

#### Set HTTP status code dynamically

Sometimes it's necessary to set status code dynamically from your Event:

```php
class CustomPostSubscriber implements EventSubscriberInterface
{
    public function write(GetResponseForControllerResultEvent $event)
    {
        // action code ...
        
        if ($condition1) {
            $request->attributes->set('status', 201);
        } else {
            $request->attributes->set('status', 200);
        }
    }

    public static function getSubscribedEvents()
    {
        return [
            KernelEvents::VIEW => [
                'write',
                EventPriorities::PRE_WRITE,
            ]
        ];
    }
}

```
To allow set that from your event you have to set status as `'request_attributes_status'` as below:

```yaml
# api/config/api_platform/resources.yaml
App\Entity\Book:
    itemOperations:
        post:
            ...
            status: 'request_attributes_status' # customize the HTTP status code to send
```

Arkowsky:
What do you think @dunglas ? Does it make sense?

teohhanhui:
We should maybe think of a generic way of overriding any operation / resource metadata, not individual implementations for specific attributes scattered everywhere.

teohhanhui:
Anyway, as I've discussed yesterday with @dunglas, I think it's preferable that we make everything stateless as much as possible.

Arkowsky:
OK, I understand that this is not solution to merge with master in your opinion. To solve that issue I can override response in my EventPriorities::PRE_WRITE event but this cause that I have to implement toJson method in my entity - this is not elegant solution too.
Have you any idea to solve it?

Arkowsky:
At the moment I replace response in my write event using my trait:
```
<?php


namespace App\Infrastructure\ApiPlatform;


use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
use ApiPlatform\Core\Util\RequestAttributesExtractor;
use Symfony\Component\HttpFoundation\Response;

trait BuildResponseWithStatusCodeTrait
{
    /**
     * @var SerializerInterface
     */
    private $serializer;
    /**
     * @var SerializerContextBuilderInterface
     */
    private $contextBuilder;

    protected function setSerializer(SerializerInterface $serializer)
    {
    }

    protected function setContextBuilder(SerializerContextBuilderInterface $contextBuilder)
    {
        $this->contextBuilder = $contextBuilder;
    }

    protected function buildResponseWithHttpStatusCode(Request $request, $object, $httpStatusCode = 200)
    {
        $attributes = RequestAttributesExtractor::extractAttributes($request);
        $context = $this->contextBuilder->createFromRequest($request, true, $attributes);
        $serializedResponse = $this->serializer->serialize($object, $request->getRequestFormat(), $context);

        return new Response(
            $serializedResponse,
            $httpStatusCode
        );
    }
}
```

Do you think is good idea to make pull request with such trait and documentation for that?

PR #2703, opened 2019-04-06
Removed hard dependencies on doctrine odm to run the tests

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2484
| License       | MIT
| Doc PR        |

PR #2787, opened 2019-05-12
[DX] Dockerfile and docker-compose

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

I'm using Docker for testing API Platform and I think it could be interesting to add it to API Platform to improve DX.

The image comes with MongoDB and XDebug.

It's complementary to https://github.com/api-platform/core/pull/2703 because even if the tests are not run with MongoDB by default, if some tests related to MongoDB fail in Travis, you might need to run them locally.

WDYT?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+version: '3'
+
+services:
+    php:
+        build:
+            context: ./docker
+        ports:
+            - "27017:27017"
+        environment:
+            XDEBUG_CONFIG: remote_host=REPLACE_WITH_YOUR_BRIDGE_IP
+            PHP_IDE_CONFIG: "serverName=docker"
+        volumes:
+            - .:/srv/php:rw,cached
+            - /opt:/opt
+        working_dir: /srv/php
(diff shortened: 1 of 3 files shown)


teohhanhui:
I just use the Docker setup from api-platform/api-platform, and do (in the `php` container):

`pecl install mongodb && docker-php-ext-enable mongodb`

teohhanhui:
> it could be interesting to add it to API Platform to improve DX

We should keep `Dockerfile`s in a separate repo if we want to have custom Docker images for running tests / builds. But there should be no `docker-compose.yml`.

Having them in this repo will create more confusion.

alanpoulain:
> Having them in this repo will create more confusion.

I was thinking this too, that's why it's a proposal. I don't know, maybe if it's correctly explained it will not cause confusion.

dunglas:
I'm -1 too. Our skeleton and https://github.com/symfony/flex/pull/128 should be enough.

PR #2791, opened 2019-05-14
Fix joinRelations with multiple associations.

xPolo360:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Hello,
I propose my first contribution to the project.

I found that the final SQL query was not correct in the case where an entity had several doctrinal relationships and was asked to retrieve some properties with the [PropertyFilter](https://api-platform.com/docs/core/filters/#property-filter).

Example:
```php
class Example {
     public $property;
     public $relation1;
     public $relation2;
}
class Relation {
     public $property1;
     public $property2;
}
```
uri: `/api/examples?properties[relation1][]=property1&properties[relation2][]=property2`

In this case the relation1 has been added but relation2 is ignored.

After searching, you can see that the bug is coming from: `src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php` line 169 or the context is overwritten.

I propose my modification to fix the bug.

Sorry I'm not good at unit testing so if someone can help me write it I'm interested.

### DIFF
--- a/features/doctrine/eager_loading.feature
+++ b/features/doctrine/eager_loading.feature
@@ -8,7 +8,7 @@ Feature: Eager Loading
   Scenario: Eager loading for a relation
     Given there is a RelatedDummy with 2 friends
     When I send a "GET" request to "/related_dummies/1"
-    And the response status code should be 200
+    Then the response status code should be 200
     And the DQL should be equal to:
     """
     SELECT o, thirdLevel_a1, relatedToDummyFriend_a2, dummyFriend_a3
@@ -42,7 +42,7 @@ Feature: Eager Loading
   Scenario: Eager loading for a relation and a search filter
     Given there is a RelatedDummy with 2 friends
     When I send a "GET" request to "/related_dummies?relatedToDummyFriend.dummyFriend=2"
-    And the response status code should be 200
+    Then the response status code should be 200
     And the DQL should be equal to:
     """
     SELECT o, thirdLevel_a4, relatedToDummyFriend_a1, dummyFriend_a5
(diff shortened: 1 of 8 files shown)


xPolo360:
Hello,

I come back to this pullRequest because I just noticed that it does not affect the number sql request but also the Order filter If have relation ...ToMany.

I have create an example [Demo ](https://github.com/xPolo360/DemoJoinRelations)to show you.
Clone my repository and enter make for install and launch.

Test 1:
`curl -X GET -H 'Accept: application/ld+json' -i 'http://localhost:8000/api/examples?properties[]=property1&properties[relation1][]=relationProperty2&properties[relation2][]=relationProperty1'`

OK but  32 Database Queries 

Test 2:
`curl -X GET -H 'Accept: application/ld+json' -i 'http://localhost:8000/api/examples?order[relation2.relationProperty1]=asc&properties[]=property1&properties[relation1][]=relationProperty2&properties[relation2][]=relationProperty1'`

NOK Status code 500
'Cannot select distinct identifiers from query with LIMIT and ORDER BY on a column from a fetch joined to-many association. Use output walkers.'

Test 3:
`curl -X GET -H 'Accept: application/ld+json' -i 'http://localhost:8000/api/examples?order[relation2.relationProperty1]=asc&properties[relation2][]=relationProperty1'`

OK

xPolo360:
Hi  @dunglas, is possible to valid this pull request ? because it is a problem in my production app

soyuka:
The fix looks correct, can you add your test (from your demo) to our behat test suite? See Tests/Fixtures and [features/filters](https://github.com/api-platform/core/blob/master/features/filter/property_filter.feature). Also, bug fixes should now target the 2.5 branch.

xPolo360:
I have change target to 2.5 but for test it is possible to help me because I'm not very experienced.
Learn Beahat and PhpUnit is on my todo list for 2020

soyuka:
We can guide you to set it up :). Copy one of the tests from the file I linked above, write your spec to reproduce the `Test 3` of your first post, add some entities to the fixtures to match your example properties and you're all set!
We're not in a hurry it's open source :).

alanpoulain:
Thank you @xPolo360.

Issue #2792, opened 2019-05-14
Wrong join condition alias replacement in FilterEagerLoadingExtension


tezvi:
This has been already reported and mentioned here #2763, #2221 and #1310. However it seems there is still a problem that results in bad DB performance on my side (using Oracle 12).

The FilterEagerLoadingExtension extension replaces all existing join aliases with new ones, but it does not replace old aliases in complex / custom WITH JOIN conditions.

Given an example with the following SQL query:

```
SELECT a1.*
FROM b AS b1
JOIN c AS c1 ON b1.id = c1.b_id AND c1.foo IN (1,2,3)
```
would result in something like this:
```
SELECT a1.*
FROM b AS b1
JOIN c AS c1 ON b1.id = c1.b_id AND c1.foo IN (1,2,3)
WHERE a1.id IN (
  SELECT a2.id
  FROM b AS b2
  JOIN c AS c2 ON b2.id = c2.b_id AND c1.foo IN (1,2,3)
)
```

The IN clause subquery still refers to old alias "**c1**". This does not impact query results but it impacts query performance. When the real-life query defines multiple JOINS this drastically reduces query performance from 50 ms to 100 seconds in my case.

https://github.com/api-platform/core/blob/796f613b438b135249339f5f20bccbabe987736d/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php#L150-L155
Notice that old alias **$joinPart->getAlias()** is added to **$aliases** array after the search/replace has been made so it will become available only for further joins, not the current one. Also the new alias should be somehow replaced in **$condition** variable before invoking **QueryBuilderHelper::addJoinOnce()**

teohhanhui:
@tezvi #2763 has not been resolved. It's still an open issue. I don't see any PRs for it yet either.

teohhanhui:
But your issue seems to be https://github.com/api-platform/core/pull/2221 :smile:

It's not related to #2763

tezvi:
Yeah it refers to the same code chunk, but you're right. #2221 is what basically gives me problems 😀

tezvi:
I could try to provide PR but would have to change QueryBuilderHelper::addJoinOnce() call and generate newAlias before adding join.
Do you think this would be a good approach?
It probably won't cover more complex WITH join conditions.

PR #2221, opened 2018-09-30
Fixed join condition replacement with new aliases.

Siregacvek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

FilterEagerLoadingExtension didn't work properly with join conditions. As a result, the reference to the old alias remained in the condition.

PR #2793, opened 2019-05-14
[2.4] Fixed join condition replacement.

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2792
| License       | MIT
| Doc PR        | 

This PR aims to finish #2221 by adding a non-regression test.
I didn't push on the original fork as it was targeting master and I think this needs to target 2.4 as a bugfix.

I've kept your commit @Siregacvek
ping @beleneglorion, please confirm that it works in your case too

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/FilterEagerLoadingExtension.php
@@ -149,10 +149,11 @@ final class FilterEagerLoadingExtension implements ContextAwareQueryCollectionEx
             }
             $alias = substr($joinString, 0, $pos);
             $association = substr($joinString, $pos + 1);
-            $condition = str_replace($aliases, $replacements, $joinPart->getCondition());
-            $newAlias = QueryBuilderHelper::addJoinOnce($queryBuilderClone, $queryNameGenerator, $alias, $association, $joinPart->getJoinType(), $joinPart->getConditionType(), $condition, $originAlias);
+            $newAlias = $queryNameGenerator->generateJoinAlias($association);
             $aliases[] = "{$joinPart->getAlias()}.";
             $replacements[] = "$newAlias.";
+            $condition = str_replace($aliases, $replacements, $joinPart->getCondition());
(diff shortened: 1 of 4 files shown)


beleneglorion:
@antograssiot yes Siregacvek commit works for me, i have tested it on my app using a fork.

tezvi:
@antograssiot nice, fast reaction on #2792 :)
Do you know approximately when this would be available on next release as I need it badly :)

soyuka:
Nice job, tyvm @antograssiot !

betd-sthibault:
Hi, 
i just think about a case not managed  in the correction. if the join already Exist in the QueryBuilderClone,

```php 
public static function addJoinOnce(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $association, string $joinType = null, string $conditionType = null, string $condition = null, string $originAlias = null, string $newAlias = null): string
    {
        $join = self::getExistingJoin($queryBuilder, $alias, $association, $originAlias);
       // if we path in this condition 
        if (null !== $join) {
            return $join->getAlias(); // 
        }
```
the newAlias si not present in the Join Part and then it should not be replaced in "where" part or other "join" condition.

perhaps add something like this 
```php
$createdAlias = QueryBuilderHelper::addJoinOnce($queryBuilderClone, $queryNameGenerator, $alias, $association, $joinPart->getJoinType(), $joinPart->getConditionType(), $condition, $originAlias, $newAlias);
if ($newAlias !== $createdAlias ) {
 array_pop($aliases);
 array_pop($replacements);
}
```

soyuka:
@betd-sthibault thanks for your input, is this related to https://github.com/api-platform/core/pull/2791 ? Would you be able to propose a PR ?

tezvi:
@betd-sthibault good point! Also I'd like to add that \ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryBuilderHelper::getExistingJoin does not check if custom JOIN condition expressions differ. Theoretically a query can contain multiple outer joins on same association but with different ON conditions. The current implementation could reuse wrong join aliases.

PR #2794, opened 2019-05-15
Allow nullable relationships to be POST or PUT with null value

Growiel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not sure
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Hello,

Preface: I'm using `allow_identifiers` for my relationships instead of IRIs (legacy requirement, can't change it).

I have an entity that has an optional ("nullable") relation but when I tried to POST a new entity with said relation set to `null`, I was getting an error:

`"Invalid identifier \"\", \"id\" was not found."`

Turns out that when the `AbstractItemNormalizer` tries to get the relationship value, because the value is "null", it's unable to find the matching item in the database.

Upon further investigation, the `nullable` option of the `@JoinColumn` annotation is ignored when the `PropertyMetadata` is built.

My PR adds the required code in `DoctrineOrmPropertyMetadataFactory` to mark a relationship as nullable if it is.

Note: A relationship will be considered nullable only if all the JoinColumns are nullable=true.

After this code is implemented it's now possible to POST something like:

`{"mandatory_relation":1,"nullable_relation":null}` and have API Platform correctly save a new Entity without the nullable_relation set.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php
+++ b/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php
@@ -82,10 +82,10 @@ final class DoctrineOrmPropertyMetadataFactory implements PropertyMetadataFactor
                 if (!isset($associationMapping['joinColumns'])) {
                     continue;
                 }
-                
+
                 $isNullable = true;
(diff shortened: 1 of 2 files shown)


Growiel:
I'll reopen this after I spend some time fixing the tests.

PR #2795, opened 2019-05-15
Allow nullable relationships to be POST or PUT with null value

Growiel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Hello,

Preface: I'm using allow_identifiers for my relationships instead of IRIs (legacy requirement, can't change it).

I have an entity that has an optional ("nullable") relation but when I tried to POST a new entity with said relation set to null, I was getting an error:

"Invalid identifier \"\", \"id\" was not found."

Turns out that when the AbstractItemNormalizer tries to get the relationship value, because the value is "null", it's unable to find the matching item in the database.

Upon further investigation, the nullable option of the @JoinColumn annotation is ignored when the PropertyMetadata is built.

My PR adds the required code in DoctrineOrmPropertyMetadataFactory to mark a relationship as nullable if it is.

Note: A relationship will be considered nullable only if all the JoinColumns are nullable=true.

After this code is implemented it's now possible to POST something like:

{"mandatory_relation":1,"nullable_relation":null} and have API Platform correctly save a new Entity without the nullable_relation set.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php
+++ b/src/Bridge/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Symfony\Component\PropertyInfo\Type;
 
 /**
  * Use Doctrine metadata to populate the identifier property.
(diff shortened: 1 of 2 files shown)


Growiel:
I'm unable to run the Behat tests locally (error: Driver not found) so there's no way I can fix them.

Could you look at them or point me on the right direction to be able to run (and fix) them ?

dunglas:
Shouldn't this case be handled directly in Symfony PropertyInfo (in this class https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/PropertyInfo/DoctrineExtractor.php)?

Growiel:
It should and seeing the code it should already be the case.

I have another custom code to disable Doctrine's type checking (I have a field that can be sent as either an integer or a string, but Doctrine only allows one type and as far as I know there's no way to fix that other than disable it completely) that might be causing the issue here.

I'll look around and report back.

Issue #2759, opened 2019-04-24
"No item route associated with the type" exception using class inheritance


vincentchalamon:
With class inheritance with `ApiResource` annotation on the parent (abstract) class, I get a `No item route associated with the type...` exception on the serialization of a related object through the IriConverter > RouteNameResolver.

PR in progress...

Nek-:
https://github.com/api-platform/core/pull/2714 fix it already isn't it?

vincentchalamon:
@Nek- I though it too, but your PR is about ApiResource inheritance, right? In my case, it's only about routing.

vincentchalamon:
But I agree it's linked

Nek-:
@vincentchalamon AFAIK if the problem is fixed at the resource resolver level, it is also in routing, isn't it?

vincentchalamon:
@Nek- I'm not sure about that, cause the RouteNameResolver compares strictly the resource class with the route resource class.

https://github.com/api-platform/core/blob/df7087816254085741820cd1513aeb5fdaaf13a1/src/Bridge/Symfony/Routing/RouteNameResolver.php#L53

Nek-:
@vincentchalamon 
I meant, by this change: https://github.com/api-platform/core/pull/2714/files#diff-2ec5bcaf896c1cc215d3c801a173c5f2R50

vincentchalamon:
@Nek- I don't see the link with the RouteNameResolver? Or maybe it's the missing point in your PR: to call the ResourceClassResolver from the IriConverter before any call to the RouteNameConverter to be sure to handle the right resourceClass (i.e.: AbstractUser instead of InternalUser in my PR behat tests). WDYT?

Nek-:
You're adding resource guess in the route name resolver because it's broken for your case. My PR fixes your case but unfortunately, there's still a bad usage in the AbstractItemNoramlizer after my commits. I took your tests and managed to make them pass with simple modifications using my modifications on the ResourceClassResolver.

I really think managing resource in only one place is the best.

Please see and comment if needed : https://github.com/api-platform/core/pull/2714/commits/6dcaee0df2a924e906b581fdc43015ab19005cb4

teohhanhui:
Should be fixed by #2797. Please verify.

PR #2797, opened 2019-05-15
Fix ResourceClassResolver handling of inheritance 

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2609, #2683, #2759
| License       | MIT
| Doc PR        | N/A

Alternative to https://github.com/api-platform/core/pull/2714

TODO:
- [x] Add tests from https://github.com/api-platform/core/pull/2714

Notable changes:
- `ResourceClassResolver` returns the most specific resource class for an object.
- Use `ResourceClassResolver` where we didn't before.
- Denormalization of non-resources no longer handled by our `ItemNormalizer`.

### DIFF
--- a/src/Api/IdentifiersExtractor.php
+++ b/src/Api/IdentifiersExtractor.php
@@ -16,7 +16,7 @@ namespace ApiPlatform\Core\Api;
 use ApiPlatform\Core\Exception\RuntimeException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
-use ApiPlatform\Core\Util\ClassInfoTrait;
+use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 
@@ -27,12 +27,11 @@ use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
  */
 final class IdentifiersExtractor implements IdentifiersExtractorInterface
 {
-    use ClassInfoTrait;
+    use ResourceClassInfoTrait;
 
     private $propertyNameCollectionFactory;
     private $propertyMetadataFactory;
     private $propertyAccessor;
-    private $resourceClassResolver;
 
     public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, PropertyAccessorInterface $propertyAccessor = null, ResourceClassResolverInterface $resourceClassResolver = null)
     {
@@ -67,7 +66,8 @@ final class IdentifiersExtractor implements IdentifiersExtractorInterface
     public function getIdentifiersFromItem($item): array
     {
         $identifiers = [];
-        $resourceClass = $this->getObjectClass($item);
+        $resourceClass = $this->getResourceClass($item, true);
+
         foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $propertyName) {
             $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
             $identifier = $propertyMetadata->isIdentifier();
@@ -81,9 +81,7 @@ final class IdentifiersExtractor implements IdentifiersExtractorInterface
                 continue;
             }
 
-            $relatedResourceClass = $this->getObjectClass($identifier);
-
-            if (null !== $this->resourceClassResolver && !$this->resourceClassResolver->isResourceClass($relatedResourceClass)) {
+            if (null === $relatedResourceClass = $this->getResourceClass($identifier)) {
                 continue;
             }
 
--- a/src/Api/ResourceClassResolver.php
+++ b/src/Api/ResourceClassResolver.php
@@ -40,33 +40,45 @@ final class ResourceClassResolver implements ResourceClassResolverInterface
      */
     public function getResourceClass($value, string $resourceClass = null, bool $strict = false): string
     {
-        $type = \is_object($value) && !$value instanceof \Traversable ? $this->getObjectClass($value) : $resourceClass;
-        $resourceClass = $resourceClass ?? $type;
-
-        if (null === $resourceClass) {
-            throw new InvalidArgumentException(sprintf('No resource class found.'));
+        if ($strict && null === $resourceClass) {
+            throw new InvalidArgumentException('Strict checking is only possible when resource class is specified.');
         }
 
-        if (
-            null === $type
-            || ((!$strict || $resourceClass === $type) && $isResourceClass = $this->isResourceClass($type))
-        ) {
+        $actualClass = \is_object($value) && !$value instanceof \Traversable ? $this->getObjectClass($value) : null;
+
+        if (null === $actualClass && null === $resourceClass) {
+            throw new InvalidArgumentException('Resource type could not be determined. Resource class must be specified.');
+        }
+
+        if (null !== $resourceClass && !$this->isResourceClass($resourceClass)) {
+            throw new InvalidArgumentException(sprintf('Specified class "%s" is not a resource class.', $resourceClass));
+        }
+
+        if (null === $actualClass) {
             return $resourceClass;
         }
 
-        // The Resource is an interface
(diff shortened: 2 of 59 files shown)


teohhanhui:
Hmm... 1 test fails on PostgreSQL: https://travis-ci.org/api-platform/core/jobs/533956026#L770-L777

I'm not sure what is going on. :laughing:

teohhanhui:
Fixed failing test on postgres.

src/Api/IdentifiersExtractor.php
```diff
@@ -67,7 +67,8 @@ public function getIdentifiersFromResourceClass(string $resourceClass): array
     public function getIdentifiersFromItem($item): array
     {
         $identifiers = [];
-        $resourceClass = $this->getObjectClass($item);
+        $resourceClass = null !== $this->resourceClassResolver ? $this->resourceClassResolver->getResourceClass($item) : $this->getObjectClass($item);
```
dunglas:
```suggestion
        $resourceClass = null === $this->resourceClassResolver ? $this->getObjectClass($item) : $this->resourceClassResolver->getResourceClass($item);
```

src/Api/ResourceClassResolver.php
```diff
@@ -40,33 +40,43 @@ public function __construct(ResourceNameCollectionFactoryInterface $resourceName
...
-            throw new InvalidArgumentException(sprintf('The given object\'s resource is the interface "%s", finding a class is not possible.', $resourceClass));
+        if ($strict && !($typesMatch = is_a($actualClass, $resourceClass, true))) {
+            throw new InvalidArgumentException(sprintf('Object of type "%s" does not match "%s" resource class.', $actualClass, $resourceClass));
+        }
+
+        $mostSpecificResourceClass = null;
+
+        foreach ($this->resourceNameCollectionFactory->create() as $resourceClassName) {
+            if (is_a($actualClass, $resourceClassName, true)) {
+                if (null === $mostSpecificResourceClass || is_subclass_of($resourceClassName, $mostSpecificResourceClass, true)) {
```
dunglas: could be merged with the previous condition to reduce the complexity

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -120,10 +120,11 @@ private function reset(): void
      */
     private function storeEntityToPublish($entity, string $property): void
     {
-        $resourceClass = $this->getObjectClass($entity);
-        if (!$this->resourceClassResolver->isResourceClass($resourceClass)) {
+        $class = $this->getObjectClass($entity);
+        if (!$this->resourceClassResolver->isResourceClass($class)) {
             return;
         }
+        $resourceClass = $this->resourceClassResolver->getResourceClass($entity);
```
dunglas: Shouldn't we add this call directly in the trait, to prevent further similar issues?

src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -115,7 +118,7 @@ public function getItemFromIri(string $iri, array $context = [])
      */
     public function getIriFromItem($item, int $referenceType = UrlGeneratorInterface::ABS_PATH): string
     {
-        $resourceClass = $this->getObjectClass($item);
+        $resourceClass = null !== $this->resourceClassResolver ? $this->resourceClassResolver->getResourceClass($item) : $this->getObjectClass($item);
```
dunglas:
```suggestion
        $resourceClass = null === $this->resourceClassResolver ? $this->getObjectClass($item) : $this->resourceClassResolver->getResourceClass($item);
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -171,19 +171,54 @@ public function supportsDenormalization($data, $type, $format = null, array $con
...
-            return $dataTransformer->transform(
-                parent::denormalize($data, $inputClass, $format, ['resource_class' => $inputClass] + $context),
-                $class,
-                $context
-            );
+        if (null !== ($inputClass = $this->getInputClass($resourceClass, $context)) && null !== ($dataTransformer = $this->getDataTransformer($data, $resourceClass, $context))) {
+            $dataTransformerContext = $context;
+
+            unset($context['input']);
+            unset($context['resource_class']);
```
dunglas: Could be merged with the previous call.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -171,19 +171,54 @@ public function supportsDenormalization($data, $type, $format = null, array $con
...
+            } catch (InvalidArgumentException $e) {
+                if (!$supportsPlainIdentifiers) {
+                    throw new UnexpectedValueException(sprintf('Invalid IRI "%s".', $data), $e->getCode(), $e);
+                }
+            }
         }
 
-        return parent::denormalize($data, $class, $format, $context);
+        if (!\is_array($data)) {
+            if ($supportsPlainIdentifiers) {
```
dunglas:
```suggestion
            if (!$supportsPlainIdentifiers) {
                throw new UnexpectedValueException(sprintf('Expected IRI or document for resource "%s", "%s" given.', $resourceClass, \gettype($data)));
            }
```

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -413,47 +409,49 @@ protected function denormalizeCollection(string $attribute, PropertyMetadata $pr
...
+                if (!$supportsPlainIdentifiers) {
                     throw $e;
                 }
             }
         }
 
         if (!\is_array($value)) {
-            // repeat the code so that IRIs keep working with the json format
-            if (true === $this->allowPlainIdentifiers && $this->itemDataProvider) {
+            if ($supportsPlainIdentifiers) {
```
dunglas: Same here, this test can be transformed in guard clause

src/Api/IdentifiersExtractor.php
teohhanhui: This feels weird to me. Having the resource class resolver should be the "nominal" case, so it's more logical that it comes first. Someone reading the code doesn't have to go to the "exceptional" branch to see it.

src/Api/ResourceClassResolver.php
teohhanhui: I've opted to change it to this instead:

```php
foreach ($this->resourceNameCollectionFactory->create() as $resourceClassName) {
    if (!is_a($actualClass, $resourceClassName, true)) {
        continue;
    }

    if (null === $mostSpecificResourceClass || is_subclass_of($resourceClassName, $mostSpecificResourceClass, true)) {
        $mostSpecificResourceClass = $resourceClassName;
    }
}
```

src/Bridge/Symfony/Routing/IriConverter.php
teohhanhui: Same as above. I don't see what's the benefit of reserving this. It makes the code harder to follow.

src/Api/IdentifiersExtractor.php
dunglas: negative condition are hard to read (at least for me). It's considered a bad practice by many quality checking tools: https://pmd.sourceforge.io/pmd-5.0.3/rules/java/design.html#ConfusingTernary

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
teohhanhui: I'm not sure it'd be a good fit. Do you have any suggestions? There are cases where we want the exception to be thrown if it's not a resource class, but in places like this one, we just want to do nothing.

src/Api/IdentifiersExtractor.php
teohhanhui: I disagree that it's "hard to read". That seems entirely subjective. For me it's easier to read in this case, because what it says is: "If we have the resource class resolver, let's do the right thing, otherwise we use a fallback."

teohhanhui: If you'd like, I could change `null !== $this->resourceClassResolver` to simply `$this->resourceClassResolver`

src/Serializer/AbstractItemNormalizer.php
teohhanhui: But it'd be harder to read... :disappointed:

teohhanhui: Looks like there's absolutely no difference in terms of opcodes:

https://3v4l.org/GREph/vld#output

vs

https://3v4l.org/3kNUI/vld#output

PR #2799, opened 2019-05-16
Fixing bug where servers would be added and set to an empty string in OpenAPI

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | not needed

Hi!

I noticed that the `servers` key on OpenAPI always look like this:

Screen Shot 2019-05-16 at 10 06 49 AM

The `url` originally comes from `$request->getBasePath()` in `DocumentationAction`. The problem is that if I try to pass the URL to the documentation straight to Postman (which supports OpenAPI 3), I see this error:

> Error while importing Open API 3.0: The OpenAPI spec should have a valid URL in each servers object

In fact, `servers` defaults to `/` in OpenAPI if not specified. There was just a bit of missing logic to skip setting it if the base path is an empty string. This fixes the import into Postman.

Cheers!

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -698,7 +698,7 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
 
         if ($v3) {
             $docs = ['openapi' => self::OPENAPI_VERSION];
-            if ('/' !== $baseUrl) {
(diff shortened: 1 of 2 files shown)


soyuka:
thanks @weaverryan !

Issue #2803, opened 2019-05-16
Broken behavior processing operations with defined 'output' class


Perf:
Hello,
an exception is thrown when processing request by `WriteListener.php`.
At line 81 there is a call to get Iri from object of output class. But, obviously, it's not always possible, as the output class could be another object rather than ApiResource.
```php
if ($hasOutput) {
    $request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($controllerResult));
}
```
Exception:
> No item route associated with the type "Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationSuccessResponse" at /srv/api/vendor/api-platform/core/src/Bridge/Symfony/Routing/RouteNameResolver.php line 62

Here is a sample configuration of a custom operation:
```yaml
    collectionOperations:
        confirmation:
            method: 'POST'
            path: '/users/confirmation'
            access_control: 'is_anonymous()'
            messenger: 'input'
            input: 'App\Entity\User\Input\UserConfirm'
            output: 'Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationSuccessResponse'
```

**Some additional notes:**
There was a related issue #2550 that was fixed with #2628.
It worked for a while... and seems to be was 'broken' with: https://github.com/api-platform/core/commit/433ce07877655f245f164df36f1a249a7faf2abb#diff-3bc4e0ec1a3570f4deb961f3652800a6
So, now `$hasOutput` gets `true` at line 77, because now `$controllerResult` is an instance of `$outputMetadata['class']`
https://github.com/api-platform/core/blob/125dec90d05dc383e9214cd4968428323d0a1136/src/EventListener/WriteListener.php#L77
which it totally correct... but this leads to improper call of 
https://github.com/api-platform/core/blob/125dec90d05dc383e9214cd4968428323d0a1136/src/EventListener/WriteListener.php#L81

_Imho, at line 80 it's necessary to add additional check, if object in `$controllerResult` is an "ApiResource" object._

dunglas:
@Perf good catch. A cleanest patch would be to pass the resource object related to the DTO to be able to populate this attribute. Would you want to work on a patch?

Perf:
hey @dunglas,
yep, I could try to develop it. But need to clarify a few details.

- Why do we need to get an item Iri for a case when we respond with an object that is not (completely) related to any of Api Resources?

What we have in current config:
[...]

PR #2810, opened 2019-05-20
set _api_write_item_iri only if result object is a subclass of a resource class

Perf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #2803 
| License       | MIT
| Doc PR        | -

Issue happens when `output` class for a specific operation is not related to a class that represents api resource.
`WriteListener` tries to set `_api_write_item_iri` from a result object (defined in `output` attribute of an operation) but fails as it's not possible.

### DIFF
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -91,7 +91,7 @@ final class WriteListener
                 }
 
                 if ($hasOutput) {
-                    $request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromItem($controllerResult));
+                    $request->attributes->set('_api_write_item_iri', $this->iriConverter->getIriFromResourceClass($attributes['resource_class']));
                 }
             break;
             case 'DELETE':
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -84,13 +84,22 @@ final class WriteListener
                 }
 
                 $hasOutput = true;
+                $getIriFromItem = false;
+                $getIriFromResourceClass = false;
                 if (null !== $this->resourceMetadataFactory) {
                     $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
                     $outputMetadata = $resourceMetadata->getOperationAttribute($attributes, 'output', ['class' => $attributes['resource_class']], true);
                     $hasOutput = \array_key_exists('class', $outputMetadata) && null !== $outputMetadata['class'] && $controllerResult instanceof $outputMetadata['class'];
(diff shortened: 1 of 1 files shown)


borNfreee:
Could you please use descriptive commit messages and update the title of this issue so it can be searched/indexed properly in the future? Thanks! :)

teohhanhui:
The fix is wrong. `getIriFromResourceClass` gives the collection IRI, not the item IRI.

I think we should just check whether it's a resource item, and skip setting the attribute if it's not. Because we should not make assumptions that won't hold true.

teohhanhui:
If you don't mind, maybe we could do this after #2797 is merged? I think it'll be easier.

Perf:
> Could you please use descriptive commit messages and update the title of this issue so it can be searched/indexed properly in the future? Thanks! :)
Hi @borNfreee,
thanks! Title and description updated.

Perf:
> The fix is wrong. `getIriFromResourceClass` gives the collection IRI, not the item IRI.
> 
> I think we should just check whether it's a resource item, and skip setting the attribute if it's not. Because we should not make assumptions that won't hold true.

Hi @teohhanhui, 
true, I updated PR, no it will set `_api_write_item_iri` only if result object is a subclass of a resource class.

> If you don't mind, maybe we could do this after #2797 is merged? I think it'll be easier.
Probably, you know better, I guess ;)

olivermack:
Wouldn't it be less restrictive to check if the `$controllerResult` is just any kind of known `ApiResource`? With the current state of the PR we'd not be able to link to any other resource type other than the one that has been requested.

E.g. when we dispatch a request for a DTO or any other unidentifiable resource and the DataPersister returns any other applicable ApiResource than given in `$attributes['resource_class']` the link may be especially handy - because a client simply could not even assume the result-link.

```php
// ... SNIP
if ($hasOutput) {
    try {
    } catch (ResourceClassNotFoundException $e) {
        // swallow
    }
}
```

Unfortunately the IriConverter bubbles only a simple `ApiPlatform\Core\Exception\InvalidArgumentException` instead of a (possibly) more suitable `ApiPlatform\Core\Exception\ResourceClassNotFoundException`.

karser:
Looks like this issue is related https://github.com/api-platform/core/issues/2860

teohhanhui:
See https://github.com/api-platform/core/issues/2860#issuecomment-502706606

I think this is not a bug.

teohhanhui:
Replaced by #2910

PR #2811, opened 2019-05-20
Backport original_data access control variable to 2.4 as a bug fix

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | api-platform/doc#814

Hi friends!

#2779 was coded as a feature and therefore was merged into master. I think that's a mistake, as it's related to security. It's currently not possible to do something like this:

```
"access_control"="is_granted('ROLE_USER') and object.getOwner() == user",
```

Without potentially exposing your API to unwanted access: a user could change the `owner` to be themselves. For the reason, even though this *is* technically a feature, I think it's critical enough for security (and simple) that it should be backported.

Thanks!

### DIFF
--- a/features/authorization/deny.feature
+++ b/features/authorization/deny.feature
@@ -58,15 +58,39 @@ Feature: Authorization checking
     """
     Then the response status code should be 201
 
-  Scenario: An user cannot retrieve an item he doesn't own
+  Scenario: A user cannot retrieve an item he doesn't own
     When I add "Accept" header equal to "application/ld+json"
     And I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
     And I send a "GET" request to "/secured_dummies/1"
     Then the response status code should be 403
     And the response should be in JSON
 
-  Scenario: An user can retrieve an item he owns
+  Scenario: A user can retrieve an item he owns
     When I add "Accept" header equal to "application/ld+json"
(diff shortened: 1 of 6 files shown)


teohhanhui:
It's possible to do with a custom listener, so I don't think it's a security issue. Being able to access the original object is a new feature.

weaverryan:
Very true! It's a subjective judgment for sure - definitely a new feature on a technical level, but one that, without it, makes it more difficult to setup authorization correctly. On a practical level, if it's merged into 2.4, it'll make it into our SymfonyCasts tutorial. If not, I'll still mention it, but won't be able to show it.

teohhanhui:
I have no objections. :smile:

dunglas:
Normally, I would have preferred to merge this patch in master, but I agree that having this feature in SymfonyCast is worth it. Let's merge.

dunglas:
However the Behat failure looks related.

teohhanhui:
@dunglas No, the Behat failure is not related. It's a known failure that happens sometimes.

/cc @alanpoulain :see_no_evil:

weaverryan:
Thank you :) :) :)

dunglas:
Thanks @weaverryan

PR #2817, opened 2019-05-21
Fix BC break in AbstractItemNormalizer introduced in v2.4

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Implementing additional interfaces (`ContextAwareNormalizerInterface` and `ContextAwareDenormalizerInterface`) in an abstract class is a BC break.

### DIFF
--- a/src/GraphQl/Serializer/ItemNormalizer.php
+++ b/src/GraphQl/Serializer/ItemNormalizer.php
@@ -33,9 +33,9 @@ final class ItemNormalizer extends BaseItemNormalizer
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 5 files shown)


dunglas:
You could remove the `implements` but use `func_get_args` to get and pass the context. So you keep the feature while fixing the BC break.

teohhanhui:
We don't need those interfaces. It was added earlier but refactored away (separate `ObjectNormalizer`).

dunglas:
Got it, let's merge.

PR #2822, opened 2019-05-25
Fix serialization when using interface as resource

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Superseeds #2797

TODO:

- [x] Add tests


src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -52,13 +55,14 @@ public function create(string $resourceClass, string $property, array $options =
 
         try {
             [$normalizationGroups, $denormalizationGroups] = $this->getEffectiveSerializerGroups($options, $resourceClass);
-
-            $propertyMetadata = $this->transformReadWrite($propertyMetadata, $resourceClass, $property, $normalizationGroups, $denormalizationGroups);
-            $propertyMetadata = $this->transformLinkStatus($propertyMetadata, $normalizationGroups, $denormalizationGroups);
         } catch (ResourceClassNotFoundException $e) {
-            // No need to check link status if related class is not a resource
+            // hack to allow non-resource classes (actually they should not be supported)
```
teohhanhui: If this hack is removed, all hell breaks loose, because we're using the `PropertyMetadataFactory` for input/output classes (DTOs). :disappointed:

dunglas: Yes we done it on purpose. It's a good move actually, it allows to retrieve the shape of any PHP structure, regardless if it is mapped with `@ApiProperty` or not.

teohhanhui: Okay, I'll update the comment. 😎

teohhanhui: But actually, I can see many problems with using the property metadata system like this. It's simply not designed for non-resources, as you can see here. It'd give the wrong results, which is arguably worse than not using it at all.

teohhanhui: @dunglas I hope this is okay for now: https://github.com/api-platform/core/compare/075222cd357ad018bc517e954ab46b5b4e9e90b3..731b45f5554d7ec832a33ed59fd1d1fac2230373

composer.json
```diff
@@ -39,16 +39,17 @@
...
+        "jangregor/phpstan-prophecy": "^0.4",
         "justinrainbow/json-schema": "^5.0",
         "nelmio/api-doc-bundle": "^2.13.3",
         "phpdocumentor/reflection-docblock": "^3.0 || ^4.0",
         "phpdocumentor/type-resolver": "^0.3 || ^0.4",
         "phpspec/prophecy": "^1.8",
-        "phpstan/phpstan": "^0.11.3",
-        "phpstan/phpstan-doctrine": "^0.11.2",
+        "phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan": "^0.11 <0.11.8",
```
teohhanhui: https://github.com/phpstan/phpstan/issues/2148

features/bootstrap/DoctrineContext.php
```diff
@@ -1227,6 +1232,108 @@ public function thereAreNbDummyDtoCustom($nb)
...
+        $this->manager->clear();
+    }
+
+    /**
+     * @Given there are :nb sites with internal owner
+     */
+    public function thereAreSitesWithInternalOwner(int $nb)
+    {
+        for ($i = 1; $i <= $nb; ++$i) {
+            $internalUser = new \ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\InternalUser();
```
alanpoulain: Why not use?

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -459,7 +459,7 @@ private function getProperty(PropertyMetadata $propertyMetadata, string $propert
     {
         $propertyData = [
             '@id' => $propertyMetadata->getIri() ?? "#$shortName/$propertyName",
-            '@type' => $propertyMetadata->isReadableLink() ? 'rdf:Property' : 'hydra:Link',
+            '@type' => false === $propertyMetadata->isReadableLink() ? 'hydra:Link' : 'rdf:Property',
```
alanpoulain: Why?

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -29,12 +30,14 @@ final class SerializerPropertyMetadataFactory implements PropertyMetadataFactory
...
     private $decorated;
+    private $resourceClassResolver;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerClassMetadataFactoryInterface $serializerClassMetadataFactory, PropertyMetadataFactoryInterface $decorated)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerClassMetadataFactoryInterface $serializerClassMetadataFactory, PropertyMetadataFactoryInterface $decorated, ResourceClassResolverInterface $resourceClassResolver = null)
     {
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->serializerClassMetadataFactory = $serializerClassMetadataFactory;
         $this->decorated = $decorated;
+        $this->resourceClassResolver = $resourceClassResolver;
```
alanpoulain: Deprecated if `null`?

features/bootstrap/DoctrineContext.php
teohhanhui: Didn't notice this as this commit was taken from the other PR. But I wonder why PHP-CS-Fixer doesn't catch this... :thinking:

src/Hydra/Serializer/DocumentationNormalizer.php
teohhanhui: Because `readableLink` could be `null`. But not sure what's the correct default / right thing to do here...

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
teohhanhui: Hmm... We need to do this in many places. Will do this in a follow-up PR?

teohhanhui:
Build failures should be fixed by https://github.com/api-platform/core/pull/2841

We shall re-trigger the build after that's merged.

teohhanhui:
Build still failing, will be fixed by https://github.com/api-platform/core/pull/2843

src/Hydra/Serializer/DocumentationNormalizer.php
teohhanhui: Oh yeah, we should not have `hydra:Link` when `readableLink` is `null`, so this is correct.

soyuka:
nice :D @Nek- will be happy with this :D

bendavies:
this has caused a regression for me and I can't seem to fix
```
{"type":"https:\/\/tools.ietf.org\/html\/rfc2616#section-10","title":"An error occurred","detail":"Object of type \u0022ClassA\u0022 does not match \u0022ClassB\u0022 resource class."}
```

I have an operation which returns a different class to resources the operation is defined on.

teohhanhui:
@bendavies Can you open an issue?

bendavies:
@teohhanhui done.

PR #2840, opened 2019-06-04
Repair a bug where properties didn't get array type (range) in documentation.

Nykilor:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Before and after the change in f.i. api platform admin:
Before (if you would try to edit, there would be just an error see https://github.com/api-platform/admin/issues/179) :
image
After:
image

The used entity file:  

			<?php

			namespace App\Entity;

			/**
			 * ApiResource, config in config/api_platform/Domain.yaml
			 *
			 * @ORM\Entity(repositoryClass="App\Repository\DomainRepository")
			 */
			class Domain
			{
				/**
				 * @ORM\Id()
				 * @ORM\GeneratedValue()
				 * @ORM\Column(type="integer")
				 */
				private $id;

				/**
				 * @ORM\Column(type="string", length=255)
				 */
				private $prefix;

				/**
				 * @ORM\Column(type="json")
				 */
				private $hosts = [];

				/**
				 * @ORM\Column(type="string", length=255)
				 */
				private $base_dn;

				/**
				 * @ORM\Column(type="string", length=255)
				 * @Assert\Choice({"ActiveDirectory", "FreeIPA", "OpenLDAP"})
				 */
				private $connection_schema = "ActiveDirectory";

				/**
				 * @ORM\Column(type="integer")
				 */
				private $port = 389;

				/**
				 * @ORM\Column(type="boolean")
				 */
				private $use_ssl = 0;

				/**
				 * @ORM\Column(type="boolean")
				 */
				private $use_tls = 0;

				/**
				 * @ORM\Column(type="integer")
				 * @Assert\Range(
				 *  min = 2,
				 *  max = 3
				 * )
				 */
				private $version = 3;

				/**
				 * @ORM\Column(type="integer")
				 */
				private $timeout = 5;

				/**
				 * @ORM\Column(type="json")
				 */
				private $custom = [];

				public function getId(): ?int
				{
					return $this->id;
				}

				public function getPrefix(): ?string
				{
					return $this->prefix;
				}

				public function setPrefix(?string $prefix): self
				{
					$this->prefix = $prefix;

					return $this;
				}

				public function getHosts(): ?array
				{
					return $this->hosts;
				}

				public function setHosts(array $hosts): self
				{
					$this->hosts = $hosts;

					return $this;
				}

				public function getBaseDn(): ?string
				{
					return $this->base_dn;
				}

				public function setBaseDn(string $base_dn): self
				{
					$this->base_dn = $base_dn;

					return $this;
				}

				public function getConnectionSchema(): ?string
				{
					return $this->connection_schema;
				}

				public function setConnectionSchema(string $connection_schema): self
				{
					$this->connection_schema = $connection_schema;

					return $this;
				}

				public function getPort(): ?int
				{
					return $this->port;
				}

				public function setPort(int $port): self
				{
					$this->port = $port;

					return $this;
				}

				public function getUseSsl(): ?bool
				{
					return $this->use_ssl;
				}

				public function setUseSsl(bool $use_ssl): self
				{
					$this->use_ssl = $use_ssl;

					return $this;
				}

				public function getUseTls(): ?bool
				{
					return $this->use_tls;
				}

				public function setUseTls(bool $use_tls): self
				{
					$this->use_tls = $use_tls;

					return $this;
				}

				public function getVersion(): ?int
				{
					return $this->version;
				}

				public function setVersion(int $version): self
				{
					$this->version = $version;

					return $this;
				}

				public function getTimeout(): ?int
				{
					return $this->timeout;
				}

				public function setTimeout(int $timeout): self
				{
					$this->timeout = $timeout;

					return $this;
				}

				public function getCustom(): ?array
				{
					return $this->custom;
				}

				public function setCustom(?array $custom): self
				{
					$this->custom = $custom;

					return $this;
				}
			}

### DIFF
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -350,6 +350,8 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
                 return 'xmls:decimal';
             case Type::BUILTIN_TYPE_BOOL:
                 return 'xmls:boolean';
+            case Type::BUILTIN_TYPE_ARRAY:
+                return 'xmls:array';
             case Type::BUILTIN_TYPE_OBJECT:
                 if (null === $className = $type->getClassName()) {
                     return null;


Nykilor:
I have taken as reference the code of Api Platform Admin https://github.com/api-platform/admin/blob/master/src/inputFactory.js#L76-84, I was thinking about just making it "array" without the "xmls:" prefix but then the Api Platform Admin would have to be adjusted too.

dunglas:
Actually, it's not possible to represent a "schema-less" such as a JSON document data structure using XML Schema types. I'm not sure how we could handle this gracefully, but this is definitely not the way to go. I'll try to dive into this at some point.

alanpoulain:
Same as https://github.com/api-platform/core/pull/1676.
Closing.
The need is here though.

PR #2841, opened 2019-06-04
Fix build with Symfony 4.3

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -530,12 +530,10 @@ jobs:
 
 workflows:
   version: 2
-  lint:
+  lint-and-coverage:
     jobs:
       - php-cs-fixer
       - phpstan
-  test-with-coverage:
-    jobs:
       - phpunit-coverage
       - behat-coverage
       - phpunit-mongodb-coverage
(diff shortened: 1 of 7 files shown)


dunglas:
Shouldn't we merge #2784 in 2.4 instead?

teohhanhui:
@dunglas But without the Symfony Messenger changes? As even for experimental code we should not do BC break in a patch version, only in minor version.

teohhanhui:
Or are we able to keep BC on our end?

dunglas:
The BC introduced by Symfony in Messenger is annoying, I'm not really sure of what to do.

teohhanhui:
Maybe we could merge this first. It's harmless. If we decide to do more (e.g. deprecation fixes), we can do that later?

dunglas:
I agree

PR #2843, opened 2019-06-05
Don't try to mock UnitOfWork as it's final

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Also re-add mongodb packages to `require-dev` as we want developers to be able to run `phpstan` successfully. This means they must have `ext-mongodb`, but it's easy to satisfy this requirement...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,20 +83,12 @@ commands:
       - run:
           name: Disable Xdebug PHP extension
           command: sudo rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
-  install-doctrine-mongodb-odm:
-    steps:
-      - run:
-          name: Install Doctrine MongoDB ODM
-          command: |
-            composer require --prefer-dist --no-progress --no-suggest --no-update --ansi \
-              doctrine/mongodb-odm:^2.0@beta \
-              doctrine/mongodb-odm-bundle:^4.0@beta \
   install-mongodb-php-extension:
     steps:
       - run:
           name: Install mongodb PHP extension
           command: |
-            sudo pecl install mongodb
+            sudo pecl install mongodb-1.5.4
             sudo docker-php-ext-enable mongodb
   merge-code-coverage-reports:
     parameters:
@@ -274,17 +266,18 @@ jobs:
     working_directory: ~/api-platform/core
     steps:
       - checkout
-      - restore-composer-cache
-      - restore-php-cs-fixer-cache
+      - install-mongodb-php-extension
       - disable-xdebug-php-extension
       - disable-php-memory-limit
       - update-composer
+      - restore-composer-cache
       - run:
           name: Install PHP-CS-Fixer
           command: |
             composer global require --prefer-dist --no-progress --no-suggest --ansi \
               friendsofphp/php-cs-fixer:^2.14
       - save-composer-cache
+      - restore-php-cs-fixer-cache
       - run:
           name: Run PHP-CS-Fixer
           command: |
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -83,20 +83,12 @@ commands:
...
-            composer require --prefer-dist --no-progress --no-suggest --no-update --ansi \
-              doctrine/mongodb-odm:^2.0@beta \
-              doctrine/mongodb-odm-bundle:^4.0@beta \
   install-mongodb-php-extension:
     steps:
       - run:
           name: Install mongodb PHP extension
           command: |
-            sudo pecl install mongodb
+            sudo pecl install mongodb-1.5.4
```
alanpoulain: Why fixing a version?

teohhanhui: It fails otherwise, as you know pecl is very picky about this. Recently an alpha version has been released, so it stopped working. :laughing:

teohhanhui: It was only working before this because the last release happened to be the stable version.

src/Test/DoctrineMongoDbOdmTestCase.php
```diff
@@ -0,0 +1,148 @@
...
+use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver;
+use Doctrine\ODM\MongoDB\UnitOfWork;
+use MongoDB\Client;
+use MongoDB\Model\DatabaseInfo;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @see https://github.com/doctrine/mongodb-odm/blob/2.0.0-RC1/tests/Doctrine/ODM/MongoDB/Tests/BaseTest.php
+ */
+abstract class DoctrineMongoDbOdmTestCase extends TestCase
```
alanpoulain: Is there a difference with https://github.com/doctrine/mongodb-odm/blob/2.0.0-RC1/tests/Doctrine/ODM/MongoDB/Tests/BaseTest.php?
If not why not using it directly? If there is, shouldn't we document which one?

teohhanhui: We cannot use it directly, as it's `export-ignore`: https://github.com/doctrine/mongodb-odm/blob/a50624c38534a3f238353c83c12fb268eee61281/.gitattributes#L7

teohhanhui: It's exactly the same, except for the two mock filters which I've removed, as we don't need those and they introduce dependencies: https://github.com/doctrine/mongodb-odm/blob/2.0.0-RC1/tests/Doctrine/ODM/MongoDB/Tests/BaseTest.php#L80-L81

teohhanhui: Maybe we only need this: https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.0.0-RC1/Tests/TestCase.php

I'll try.

.circleci/config.yml
alanpoulain: OK that's too bad :disappointed:

src/Test/DoctrineMongoDbOdmTestCase.php
alanpoulain: There's more or less already something like this in API Platform: https://github.com/api-platform/core/blob/master/src/Test/DoctrineMongoDbOdmSetup.php

teohhanhui: Okay... :laughing: Let's try to use it then.

teohhanhui:
Need to fix this: https://github.com/doctrine/DoctrineMongoDBBundle/issues/557

composer.json
```diff
@@ -33,8 +33,11 @@
         "behat/symfony2-extension": "^2.1.1",
         "behatch/contexts": "3.1.0",
         "doctrine/annotations": "^1.2",
+        "doctrine/data-fixtures": "^1.2.2",
```
teohhanhui: Temporary workaround for https://github.com/doctrine/DoctrineMongoDBBundle/issues/557

dunglas:
Thanks @teohhanhui!

PR #2849, opened 2019-06-10
Remove unnecessary not null check

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Based on static analysis this can never return null, so no need to check.


teohhanhui:
Thanks @norkunas! :tada:

PR #2859, opened 2019-06-13
Fixed denormalization of a constructor argument which is a collection of non-resources

karser:
Hi Folks!

After upgrading to 2.4.4 (from 2.4.2) I started getting an error.
The error wording is: `The property path constructor needs a string or an instance of \"Symfony\\Component\\PropertyAccess\\PropertyPath\". Got: \"integer\" in symfony/serializer/Normalizer/AbstractObjectNormalizer.php, line 302`.

I fixed this bug by adding '[]' to the class of the collection argument at [this line](https://github.com/api-platform/core/blob/v2.4.4/src/Serializer/AbstractItemNormalizer.php#L666).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | PR adds the test case
| License       | MIT

### DIFF
--- a/features/serializer/deserialize_objects_using_constructor.feature
+++ b/features/serializer/deserialize_objects_using_constructor.feature
@@ -10,7 +10,12 @@ Feature: Resource with constructor deserializable
     """
     {
       "foo": "hello",
-      "bar": "world"
+      "bar": "world",
+      "items": [
+        {
+          "foo": "bar"
+        }
+      ]
     }
     """
     Then the response status code should be 201
@@ -25,7 +30,11 @@ Feature: Resource with constructor deserializable
       "id": 1,
       "foo": "hello",
       "bar": "world",
+      "items": [
+        {
+          "foo": "bar"
+        }
+      ],
       "baz": null
     }
     """
-
--- a/tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
+++ b/tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
 
 use ApiPlatform\Core\Annotation\ApiResource;
+use ApiPlatform\Core\Tests\Fixtures\DummyObjectWithoutConstructor;
 use Doctrine\ORM\Mapping as ORM;
 use Symfony\Component\Serializer\Annotation\Groups;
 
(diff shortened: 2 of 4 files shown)


soyuka:
Mhhh shouldn't this case go through the `normalizeCollection` few lines before?

karser:
The [few lines before case](https://github.com/api-platform/core/blob/v2.4.4/src/Serializer/AbstractItemNormalizer.php#L655) is for resource collection. It checks for `isResourceClass`. The [case that I fixed](https://github.com/api-platform/core/blob/v2.4.4/src/Serializer/AbstractItemNormalizer.php#L666) with `[]` is apparently, for non-resource collection. 
The `createAttributeValue` method was added in commit `Fix ResourceClassResolver handling of inheritance`. @teohhanhui Is this PR correct?

teohhanhui:
Yes, you're right. The `[]` suffix is important for `ArrayDenormalizer`. I'll check soon.

features/serializer/deserialize_objects_using_constructor.feature
```diff
@@ -10,7 +10,10 @@ Feature: Resource with constructor deserializable
     """
     {
       "foo": "hello",
-      "bar": "world"
+      "bar": "world",
+      "items": [
+        {"foo": "bar"}
```
teohhanhui: Please format correctly.

features/serializer/deserialize_objects_using_constructor.feature
```diff
@@ -25,6 +28,7 @@ Feature: Resource with constructor deserializable
       "id": 1,
       "foo": "hello",
       "bar": "world",
+      "items": [{"foo": "bar"}],
```
teohhanhui: Please format correctly.

tests/Fixtures/TestBundle/Entity/DummyEntityWithConstructor.php
```diff
@@ -64,10 +65,19 @@ class DummyEntityWithConstructor
      */
     private $baz;
 
-    public function __construct(string $foo, string $bar)
+    /**
+     * @var DummyObjectWithoutConstructor[]
```
teohhanhui: Doctrine ORM mapping is missing.

teohhanhui:
@karser You need to to do the same for the corresponding class mapped as MongoDB Document.

karser:
@teohhanhui done. MongoDB documents reuse behat features, don't they?

teohhanhui:
Hmm... Not sure why the same test for mongodb is still failing.

teohhanhui:
Yikes. Looks like we're expecting the same order for the JSON keys, which is nonsensical.

karser:
Not fully following and it's not evident what you changed. I added `$items` in the corresponding places in the entity and document. If test passes for the entity, the order of json keys in the output shouldn't be different for the document.

teohhanhui:
It's because you added the `getItems` method in a different order in the document compared to the entity, so.... :see_no_evil:

karser:
I see now, ugh!

karser:
Just wondering, since the entities and documents must be 100% identical, wouldn't it make sense to use the same classes and move the orm and odm annotations to xml files?

teohhanhui:
Thanks @karser! :tada: 🚀

Issue #2855, opened 2019-06-12
Regression where operation returns resource class different from input resource class


bendavies:
Introduced here:
https://github.com/api-platform/core/pull/2822

Example operation:
```php
/**
 * @ApiResource(
 *     itemOperations={
 *         "get": {
 *             "method": "GET",
 *             "path": "/payments/{id}",
 *         },
 *         "void": {
 *             "method": "POST",
 *             "path": "/payments/{id}/void",
 *             "controller": PaymentVoid::class,
 *             "deserialize": false
 *         }
 *     }
 * )
 */
class Payment { ... 
```

`PaymentVoid` controller returns a `Void`, which is also a resource.
This worked on `2.4.3`.

Now errors:
```
{"type":"https:\/\/tools.ietf.org\/html\/rfc2616#section-10","title":"An error occurred","detail":"Object of type \u0022Void\u0022 does not match \u0022Payment\u0022 resource class."}
```

teohhanhui:
Do you have a stack trace?

bendavies:
@teohhanhui i'll DM you

bastnic:
FYI I have the same error but in a slightly different context. 

parent have multiple childs (relation OneToMany), child have have one parent (relation ManyToOne).

each serializaion group (parent:list / child:list) relate each other exect the relation field that the child.parent is not in serialization group parent:list to avoid recursion.

a call to /parents/ give all the parent with there relative multiple childs.

but a call to /childs/ crash `"Object of type \"...\Parent" does not match \"Child\" resource class."`. 

stacktrace : 
- src/Api/ResourceClassResolver.php L63
- src/Serializer/AbstractItemNormalizer.php L127 arguments : (object : proxy parent 1, string : class child, boolean true)

PR #2861, opened 2019-06-14
Allow custom operations to return a different class than the expected resource class

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2855, api-platform/api-platform#1172
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/features/main/custom_operation.feature
+++ b/features/main/custom_operation.feature
@@ -2,10 +2,13 @@ Feature: Custom operation
   As a client software developer
   I need to be able to create custom operations
 
+  Background:
+    Given I add "Accept" header equal to "application/ld+json"
+    And I add "Content-Type" header equal to "application/ld+json"
+
   @createSchema
   Scenario: Custom normalization operation
     When I send a "POST" request to "/custom/denormalization"
-    And I add "Content-Type" header equal to "application/ld+json"
     Then the JSON should be equal to:
     """
     {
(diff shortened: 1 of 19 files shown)


bendavies:
this obviously begs the question, what is `$strict` even for?

tests/Fixtures/TestBundle/Document/Payment.php
```diff
@@ -0,0 +1,82 @@
...
+     * @var int|null
+     *
+     * @ODM\Id(strategy="INCREMENT", type="integer")
+     */
+    private $id;
+
+    /**
+     * @var string
+     */
+    private $amount;
```
teohhanhui: Not mapped as Doctrine MongoDB ODM does not have a decimal type. I'm not sure what's the right thing to do here... If I use `type="float"`, the type validation fails.

teohhanhui: https://github.com/doctrine/mongodb-odm/issues/1835

features/main/custom_operation.feature
```diff
@@ -2,10 +2,13 @@ Feature: Custom operation
   As a client software developer
   I need to be able to create custom operations
 
+  Background:
```
soyuka: oh this is nice, does this allow to add steps to every scenario?

teohhanhui: Yeah.

teohhanhui:
> Is the strict parameter still used?

No. But I don't think we should deprecate it either? There are tests that use it so it's at least covered.

soyuka:
> No. But I don't think we should deprecate it either? There are tests that use it so it's at least covered.

Hmm yeah leave it like this, at first I thought that this class was `@internal`.

PR #2865, opened 2019-06-17
send Varnish BAN requests with smaller headers

jocel1:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

The idea of this PR is to split the iris in several smaller chunks to avoid the "400 Request Header Or Cookie Too Large" on Varnish when the iris contains a large amount of elements.
CHUNK_SIZE is currently hardcoded, but if this could be controlled as a YAML config parameter, it would be even better.


soyuka:
Is there a limit on varnish? How did you come up with the value `100`? 

Nice improvement!

bastnic:
FYI, there is a chunk made in FosHTTPCache https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/Varnish.php#L84 and there is méthod dedicated to calculate how much tags fits in the http header : https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/HttpProxyClient.php#L137

jocel1:
@soyuka it's a really conservative value, but default values in Varnish are not that big (you have by default http_req_hdr_len=8k and http_req_size=32k in Varnish, but even with http_req_size=64k & http_req_hdr_len=4096k I was still experiencing 400 Bad Request issues).

jocel1:
@bastnic thanks for the tips, this looks great!

dunglas:
👍 on my side! Copying (and crediting) the method from FOSHttpCache looks like a good idea. Could you also fix the PHPStan issue please?

jocel1:
@dunglas changes done + adding a yaml config variable to set the max_header_length supported by the server

jocel1:
Hi @dunglas , any plan to merge this one ? :)

src/HttpCache/VarnishPurger.php
```diff
@@ -24,14 +24,43 @@
...
         $this->clients = $clients;
+        $this->maxHeaderLength = $maxHeaderLength;
+    }
+
+    /**
+     * Calculate how many tags fit into the header.
+     *
+     * This assumes that the tags are separated by one character.
+     *
+     * From https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/HttpProxyClient.php#L137
```
teohhanhui: Please reference a specific commit id, otherwise this comment will be outdated real soon.

jocel1: right, done!

Cocray:
> Hi @dunglas , any plan to merge this one ? :)

Hello there! Should we have a release date for this feature? It's something we're waiting with excitement :)

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -375,6 +375,10 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
                                     ->prototype('scalar')->end()
                                     ->info('URLs of the Varnish servers to purge using cache tags when a resource is updated.')
                                 ->end()
+                                ->integerNode('max_header_length')
+                                    ->defaultValue(7500)
```
teohhanhui: How is this default derived?

src/HttpCache/VarnishPurger.php
```diff
@@ -24,14 +24,43 @@
...
 {
+    private $maxHeaderLength;
     private $clients;
 
     /**
      * @param ClientInterface[] $clients
+     * @param int               $maxHeaderLength
      */
-    public function __construct(array $clients)
+    public function __construct(array $clients, $maxHeaderLength = 7500)
```
teohhanhui:
```suggestion
    public function __construct(array $clients, int $maxHeaderLength = 7500)
```

src/HttpCache/VarnishPurger.php
```diff
@@ -24,14 +24,43 @@
  */
 final class VarnishPurger implements PurgerInterface
 {
+    private $maxHeaderLength;
     private $clients;
 
     /**
      * @param ClientInterface[] $clients
+     * @param int               $maxHeaderLength
```
teohhanhui: Remove useless phpdoc (use typehint instead).

src/HttpCache/VarnishPurger.php
```diff
@@ -24,14 +24,43 @@
  */
 final class VarnishPurger implements PurgerInterface
 {
+    private $maxHeaderLength;
```
teohhanhui: This should be moved after `$clients`.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
jocel1: Default Apache max headers size is 8K so the idea is to let some room for other headers, hence the 7500 value

src/HttpCache/VarnishPurger.php
jocel1: ok for the typehint, but keeping the phpdoc avoid warning from IDE like phpstorm

jocel1: phpdoc removed anyway :)

jocel1:
Hi there! Any update on this one?

src/HttpCache/VarnishPurger.php
```diff
@@ -25,13 +25,41 @@
...
+     * This assumes that the tags are separated by one character.
+     *
+     * From https://github.com/FriendsOfSymfony/FOSHttpCache/blob/2.8.0/src/ProxyClient/HttpProxyClient.php#L137
+     *
+     * @param string[] $escapedTags
+     * @param string   $glue        The concatenation string to use
+     *
+     * @return int Number of tags per tag invalidation request
+     */
+    private function determineTagsPerHeader($escapedTags, $glue)
```
teohhanhui:
```suggestion
    private function determineTagsPerHeader(array $escapedTags, string $glue): int
```

teohhanhui:
phpstan failure not related.

teohhanhui:
Thanks @jocel1! :tada: :rocket:

Issue #615, opened 2016-07-13
[Bug] Abstract doctrine inheritance collection properties


soyuka:
Is there a reason why collection properties on an abstract entity doesn't have it's child entities properties? I also tried to manage those with groups but they are not taken into consideration. 
For example [this test](https://github.com/api-platform/core/blob/master/features/crud_abstract.feature#L45) will not have the `instance` property, even though it is fetched from doctrine because it's a `SINGLE_TABLE` inheritance.

On an item, the behavior is different. See [this test](https://github.com/api-platform/core/blob/master/features/crud_abstract.feature#L29) where child properties are normalized well.

soyuka:
I think that it has to do with the fact that the serializer only sees properties belonging to the parent (ie the abstract class).

soyuka:
I narrowed down the issue to the NameCollectionFactories not having inherited properties.

@dunglas, I just saw that the `AnnotationPropertyNameCollectionFactory` is not used anymore. Also, you wrote that `PropertyInfoPropertyNameCollectionFactory` is not a decorator in the class comments. I agree that annotation for properties is a bit too much work to configure but getting the reflection class methods seems a good idea. Can you explain your choice? 

On one hand, `AnnotationPropertyNameCollectionFactory` had a reflection which would be able to get the parent properties (if I'm not mistaken `public object ReflectionClass::getParentClass ( void )`).
On the other hand, as this issue is doctrine-related I was going to add a `DoctrineOrmPropertyNameCollectionFactory` that checks for inheritance on the class metadata. On second thoughts I need your opinion (ping @teohhanhui) because properties inheritance should also be effective on non-doctrine resources, don't you think?
So, if that's the case, should the inherited properties be added in the `PropertyInfoPropertyNameCollectionFactory`? If yes, any ideas on how this could work seamlessly? 

Thanks!

dunglas:
Currently the system doesn't support inheritance at all. It would be nice to be able to support it but indeed it should be done in a generic way (it should work for all loaders).

soyuka:
Ok, my thoughts were correct.

> What do you think about adding a new ParentPropertyFactory class that will scan all parents classes, load their properties metadata and add them to the current Value Object if it doesn't already exist?
That way it would add inheritance support for all existing loaders.

[...]

PR #2869, opened 2019-06-19
fix InheritedPropertyMetadataFactory to build correctly the metadata

alessandroniciforo:
The InheritedPropertyMetadataFactory is supposed to build the metadata of all the inherited properties of a class from its parents. This was not working because of a mistake on the `is_subclass_of` check.
The related test was always passing because of a wrong mock. I fixed that, too.

There are some related issues that have been fixed in the past but that were probably still actual now, such as #1071 and #615

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | not needed

### DIFF
--- a/src/Metadata/Property/Factory/InheritedPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/InheritedPropertyMetadataFactory.php
@@ -44,7 +44,7 @@ final class InheritedPropertyMetadataFactory implements PropertyMetadataFactoryI
                 continue;
             }
 
-            if (is_subclass_of($knownResourceClass, $resourceClass)) {
+            if (is_subclass_of($resourceClass, $knownResourceClass)) {
                 $propertyMetadata = $this->create($knownResourceClass, $property, $options);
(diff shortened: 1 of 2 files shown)


soyuka:
This is a really hot path haha, I'm not sure that it's the correct fix, maybe that https://github.com/api-platform/core/pull/2861 is related?

teohhanhui:
@soyuka It's not related to #2861.

Yes, I've noticed this class before, but never understood what it's supposed to do / how it works lol...

soyuka:
It fixes metadata for classes that inherit from some other class.

fri0z:
When can I expect this fix to be released?

polc:
Just tested this patch with my current project and it seems to fix my issues with single table inheritances :)

teohhanhui:
This is obsolete after #3273.

Issue #2860, opened 2019-06-13
[2.4.3] No resource class found for object of type when using DTO as output


karser:
Hi Folks!

I'm using api-platform for handling custom use cases:
```
App\UseCase\SampleUseCaseRequest:
    collectionOperations:
        post:
            path: '/sample-path'
            messenger: true
            output: 'App\UseCase\SampleUseCaseResponse'
```

This configuration worked well with v2.4.2. After upgrading to 2.4.3 (and 2.4.4 as well) I started getting this issue:
```
{"@context":"\/contexts\/Error","@type":"hydra:Error","hydra:title":"An error occurred","hydra:description":"No resource class found for object of type \u0022App\\UseCase\\SampleUseCaseResponse\u0022.","trace":[{"namespace":"","short_class":"","class":"","type":"","function":"","file":"vendor\/api-platform\/core\/src\/Api\/ResourceClassResolver.php","line":54,"args":[]},{"namespace":"ApiPlatform\\Core\\Api","short_class":"ResourceClassResolver","class":"ApiPlatform\\Core\\Api\\ResourceClassResolver","type":"-\u003E","function":"getResourceClass","file":"vendor\/api-platform\/core\/src\/Util\/ResourceClassInfoTrait.php","line":52,"args":[["object","App\\UseCase\\SampleUseCaseResponse"]]},{"namespace":"ApiPlatform\\Core\\Bridge\\Symfony\\Routing","short_class":"IriConverter","class":"ApiPlatform\\Core\\Bridge\\Symfony\\Routing\\IriConverter","type":"-\u003E","function":"getResourceClass","file":"vendor\/api-platform\/core\/src\/Bridge\/Symfony\/Routing\/IriConverter.php","line":120,"args":[["object","App\\UseCase\\SampleUseCaseResponse"],["boolean",true]]},{"namespace":"ApiPlatform\\Core\\Bridge\\Symfony\\Routing","short_class":"IriConverter","class":"ApiPlatform\\Core\\Bridge\\Symfony\\Routing\\IriConverter","type":"-\u003E","function":"getIriFromItem","file":"src\/ApiPlatform\/ParamsAwareIriConverter.php","line":62
```

So `IriConverter::getIriFromItem` calls `ResourceClassResolver::getResourceClass` which throws `InvalidArgumentException` if the output is not an api resource.
![image](https://user-images.githubusercontent.com/1675033/59468881-e48a6300-8e3b-11e9-9c9b-dfcbd6c299f0.png)

[...]

Issue #2147, opened 2018-08-06
[Swagger] [Question] [Potential PR] nested object that are not ressources 


jcotineauCoffreo:
Hello everyone, i am opening this issue as a question and an hypothetic PR.
 
I've faced an issue while using api-platform, quite the same as #285 which is closed now.
to fix it i changed the behavior of the DocumentNormalizer and the SerializerPropertyMetadataFactory so that nested object that are not resource are still displayed in the swagger document; but i am not sure if this changes in behavior still match the philosophy of api-platform.

So here come the example : 
I have a TimeRecord, WHICH IS a resource 
And two other classes NbHours and Extras that ARE NOT resources 

```php
/**
 * Class TimeRecord.
 * @ApiResource
 */
class TimeRecord
{
    /**
     * @var string
     */
    public $id;

    /**
     * @var string
     */
    public $date;

    /**
     * @var string
     */
    public $comment;

    /**
     * @var \App\Entity\NbHours
     */
    public $nbHours;

    /**
     * @var \App\Entity\Extras
     */
    public $extras;

    /**
     * TimeRecord constructor.
     */
    public function __construct()
    {
    }
}

/**
 * Class NbHours.
 */
class NbHours
{
    /**
     * @var int
     */
    public $day = 0;

    /**
     * @var int
     */
    public $night = 0;

    /**
     * NbHours constructor.
     */
    public function __construct()
    {
    }
}
```
With the default configuration my TimeRecord model with be displayed like this :
![timerecordmodel](https://user-images.githubusercontent.com/42067143/43723368-c042d19c-9997-11e8-8afc-e45c42b94d62.png)

With the modifications I've done it will display this : 
![timerercordmodelfixed](https://user-images.githubusercontent.com/42067143/43723460-01f64920-9998-11e8-8a78-e1f86be7e323.png)

Same example with the json returned from operations 
Before the modifications : 
![jsonreturned](https://user-images.githubusercontent.com/42067143/43723979-40719ad2-9999-11e8-9d25-4b4d37f7dd3e.png)

After the modifications : 
![jsonreturnedfixed](https://user-images.githubusercontent.com/42067143/43723988-44a8e9d4-9999-11e8-9b78-7ad85c629861.png)

I've seen this issue many times and it seemed great to me to have a complete model of an object based on the serialization groups and property types even for non-resource object.

I am a trainee and i have not a lot of knowledge, neither in api-platform nor swagger nor symfony, but i'll be very happy to contribute to this project.

[...]

PR #2878, opened 2019-06-25
Allow messenger handler to return non-resource output

karser:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2860 #2810
| License       | MIT

Hi folks!

Let's imagine ResetPasswordRequest / ResetPasswordRequestHandler case from the api-platform docs (https://api-platform.com/docs/core/messenger/).

Before api-platform v2.4.3 it was possible to return the output dto object (ResetPasswordResponse) directly from the messenger handler. It was very convenient, and that's what dto's are for - return a custom response without exposing an entity.
As of v2.4.3 it started to complain that ResetPasswordResponse must be a resource, so this PR allows returning the non-resource output.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -164,6 +164,7 @@
             <argument type="service" id="api_platform.data_persister" />
             <argument type="service" id="api_platform.iri_converter" on-invalid="null" />
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
+            <argument type="service" id="api_platform.resource_class_resolver" />
 
             <tag name="kernel.event_listener" event="kernel.view" method="onKernelView" priority="32" />
         </service>
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\EventListener;
 
 use ApiPlatform\Core\Api\IriConverterInterface;
+use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\DataPersister\DataPersisterInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ToggleableOperationAttributeTrait;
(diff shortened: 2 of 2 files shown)


teohhanhui:
I've already explained to you on Slack why this is wrong. But let me say it again: it's the job of the DataTransformer to transform a resource to the output DTO. If you want to return a non-resource, just don't use the `output` option.

karser:
Sorry for bothering you again with this case but I'm not seeing an alternative in the approach you suggest. Moreover, it was working in 2.4.2. Why a bugfix update breaks the backwards compatibility?

> If you want to return a non-resource, just remove the output option and it should work.

the output model is auto-documented. If I remove the output, the documentation will be missing.

> it's the job of the DataTransformer to transform a resource to the output DTO

the code would be too scattered if ResetPasswordRequestHandler returned a resource and ResetPasswordRequestDataTransformer returned an output object. Also, the resource doesn't have to be convertable to the output dto.

My points are:
- returning the output was described in the [first version of dto docs](https://github.com/api-platform/docs/pull/369/files) added by @lyrixx. The messenger+output is supposed to be a successor of that approach.
- output is amazingly useful - it solves the problem of custom methods and adds the output to swagger doc

I'm wondering what is @dunglas and @soyuka opinion on this case.

soyuka:
> I've already explained to you on Slack why this is wrong. But let me say it again: it's the job of the DataTransformer to transform a resource to the output DTO. If you want to return a non-resource, just remove the output option and it should work.

Wait we'ŕe in the context of a MessageHandler, isn't it a bit different? Could the MessageHandler act as a data transformer? Shouldn't we support non-resources anyway (to me we definitely should)?

> the code would be too scattered if ResetPasswordRequestHandler returned a resource and ResetPasswordRequestDataTransformer returned an output object. 

To me this makes sense and removes some useless boilerplate code.

I don't really see anything wrong with the PR here, it just simplifies this use case no?

karser:
Hey guys! What should I do to get this PR merged? Submit it again? Add some
tests?

teohhanhui:
@soyuka @karser

> > If you want to return a non-resource, just remove the output option and it should work.
>
> the output model is auto-documented. If I remove the output, the documentation will be missing.

This is the actual issue. We should allow documenting any classes.

Otherwise it's already correct. Just don't use `output` class if you don't want to use a DataTransformer. It's unnecessary.

karser:
@teohhanhui  You are right, the documentation - that's what I use the output for.

>We should allow documenting any classes

how do you imaging that? a special option or using swagger_context?

Tweaking `swagger_context` won't work, looks like the DTO class has to be added to swagger definitions.

```
App\UseCase\ForgotPasswordRequest:
    collectionOperations:
        post:
            path: '/users/forgot-password-request'
            messenger: true
#            output: 'App\UseCase\ForgotPasswordResponse'
            status: 200
            swagger_context:
                responses:
                    200:
                        description: 'Response'
                        schema:
                            type: 'App\UseCase\ForgotPasswordResponse'
#                            $ref doesn't seem to be working as well
```

image

teohhanhui:
See https://github.com/api-platform/core/issues/2147

PR #2890, opened 2019-06-28
Varnish: switch from Guzzle to HttpClient

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

It will allow to benefit from the upcoming integration in the Symfony profiler. It also removes some non-Symfony dependencies.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -41,11 +41,13 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\DirectoryResource;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\HttpClient\ScopingHttpClient;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Component\Yaml\Yaml;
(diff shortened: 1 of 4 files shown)


teohhanhui:
How do we handle the BC breaks? Also, is there a reason why we're not using PSR-18 interfaces?

dunglas:
Actually there is no code BC breaks because the signature of the only method we use is the same for Guzzle and HttpClient.
However there is a potential config BC break that I’m not sure how to fix.

We’ll not use PSR-18 because it requires one more dependency and because it has limitations that may hurt us at some point (it doesn’t allow to handle timeouts correctly for instance).

teohhanhui:
> We’ll not use PSR-18 because it requires one more dependency and because it has limitations that may hurt us at some point (it doesn’t allow to handle timeouts correctly for instance).

I think interoperability is really important, especially if we want API Platform to support more than the Symfony framework. We should use PSRs wherever possible.

dunglas:
- We have symfony/contracts for interop
- We need a powerful HttpClient (for other use cases), and the PSR isn’t good enough for our cases, the Symfony interface is better
- Even Guzzle doesn’t implement PSR-18
- Both Laravel, Doctrine and Symfony have leave the PHP-FIG. API Platform has never been part of it. The time where the FIG was useful is unfortunately finished.

I’m not against implementing PSRs that are relevant for us (we’ll maybe use the event dispatcher one for our new event system for instance), but in the specific case of HttpClient, the Symfony one is better (no extra deps, more flexible) and for interop people can just use the interface provided by Symfony Contracts.

teohhanhui:
> We need a powerful HttpClient (for other use cases), and the PSR isn’t good enough for our cases, the Symfony interface is better

They're not mutually exclusive. Since I believe `symfony/http-client` will be an optional dependency anyway, the user can choose if they want to install `symfony/http-client` and use the API Platform features that require it, or just use any `psr/http-client` implementation for the features that don't.

It's a principle that we should follow especially if we want to make API Platform modular. We should not require an extended interface when we only need the base one.

dunglas:
And another reason, more pragmatic: depending of PSR-18 would force the user to install 2 packages by hand to get Varnish support (and any other feature requiring an HttpClient), while currently it’s `composer req http-client`.

dunglas:
Symfony HttpClientInterface doesn’t extend the PSR-18 one. It’s an alternative proposal. So it changes nothing regarding modularity.

dunglas:
And the PSR-18 compatibility layer adds a tiny overhead for both client and the response classes (it uses decorators).

teohhanhui:
Hmm... Anyway, I believe we have a strong reason to not make the switch yet. The Symfony HttpClient component is still experimental. For Symfony Messenger integration, that's necessarily the case, but not here.

dunglas:
I agree. It’s also why the the PR is in draft. And it will be easier to target 3.0

soyuka:
https://github.com/api-platform/core/pull/4695

PR #2894, opened 2019-07-01
Allow to extract blank values from composite identifier

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

This modification allow to correctly fetch item with composite identifiers where some fields might be blank (not null) but are still needed to determine the correct record.

Currently it would fail as the `ApiPlatform\Core\Identifier\IdentifierConverter` would throw an `ApiPlatform\Core\Exception\InvalidIdentifierException` [here](https://github.com/api-platform/core/blob/d46ac32d404e0858c2d2dd13966fb1a5a52f6719/src/Identifier/IdentifierConverter.php#L54-L65) resulting in a `Symfony\Component\HttpKernel\Exception\NotFoundHttpException` in the `ApiPlatform\Core\EventListener\ReadLister`, because of an invalid identifier configuration` [here](https://github.com/api-platform/core/blob/d46ac32d404e0858c2d2dd13966fb1a5a52f6719/src/EventListener/ReadListener.php#L109-L111)

### DIFF
--- a/src/Identifier/CompositeIdentifierParser.php
+++ b/src/Identifier/CompositeIdentifierParser.php
@@ -33,7 +33,7 @@ final class CompositeIdentifierParser
     {
         $matches = [];
         $identifiers = [];
-        $num = preg_match_all('/(\w+)=(?<=\w=)(.+?)(?=;\w+=)|(\w+)=([^;]+);?$/', $identifier, $matches, PREG_SET_ORDER);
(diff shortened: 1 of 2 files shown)


soyuka:
Needs a rebase somehow (feel free to merge after @antograssiot !)

antograssiot:
I didn't target the correct branch, this should land in 2.4 IMO, do you still agree @dunglas @soyuka

dunglas:
LGTM! Thanks @antograssiot

Issue #759, opened 2016-09-20
JSON PATCH support


dkarlovi:
#147 suggested adding a PATCH support for partials updates, it was closed as PUT already does that.

How about adding a [JSON PATCH](http://jsonpatch.com/) support? Difference here is you're able to do operations WITH entities, not only ON entites, quoted from [Why PATCH is Good for Your HTTP API](https://www.mnot.net/blog/2012/09/05/patch)

> PATCH is atomic, which means that if you need to do some complex changes to the state of a resource, you can do it with confidence. It also means that you can create synthetic resources and PATCH them if you need to orchestrate changes to the state of several resources.

dunglas:
It would definitely be a great improvement to have support for JSON PATCH.

dunglas:
On the other end, [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) will be easier to implement, and easier to use client-side.

teohhanhui:
http://erosb.github.io/post/json-patch-vs-merge-patch/

asimonf:
@dunglas given that the content-type of both methods is different, couldn't both be implemented? I understand that Merge Patch has limitations, but for many cases, it's simplicity could be great for certain applications.

What I mean is, I don't see why we should have to pick one or the other. They are just different formats and both could be supported. If that's ok with you, I might have a go at trying to implement merge patch, time permitting.

dunglas:
Yew we can support both. It should be possible to support JSON MERGE PATCH with just a few modifications to the current system.

teohhanhui:
Implementing JSON Patch will force us to improve our design. And it will allow us to deprecate the current partial PUT semantics (which is wrong). :smile:

asimonf:
@teohhanhui I don't see why both (JSON MERGE and JSON PATCH) can't be implemented. I don't disagree with you and I'm sure if a PR appears for JSON PATCH, @dunglas would be willing to merge assuming it passes review. In my use-case, the simplicity of JSON MERGE trumps the flexibility of JSON PATCH.

teohhanhui:
I'm okay with having both. :)

teohhanhui:
Also, I think we can deprecate the partial PUT once we have JSON Patch.
Because it already covers the existing semantics.

StephenOTT:
anyone actively working on this?

dunglas:
AFAIK, no. But we should probably consider using JSON Merge Patch instead: https://tools.ietf.org/html/rfc7386

[...]

PR #2895, opened 2019-07-01
Add support for PATCH, refactor formats detection

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #759 (partially)
| License       | MIT
| Doc PR        | n/a

Add proper support for partial modifications using the `PATCH` HTTP method. The first supported format will be [JSON Merge Patch (RFC 7386)](https://tools.ietf.org/html/rfc7386), while JSON:API Patch will keep working. Supporting [JSON Patch (RFC 6902)](https://tools.ietf.org/html/rfc6902) isn't a a goal as this format is not very popular in the wild, but could be done by leveraging the new infrastructure I'm building.

This PR also deprecates `FormatsProviderInterface` and `OperationMethodResolverInterface` and all there implementation in favor of populating the appropriate metadata. This approach as several benefits:

* simplification of the code base
* better performance because we move some run time logic to compile time
* better flexibility

TODO:

* [x] Fix remaining deprecations
* [x] Set Serializer's `skip_null_values` to true when appropriate (always when PATCH is enabled?)
* [x] Use patch formats in docs
* [ ] Find a way to block partial `PUT` (I've no clue of how to that for now, but actually it could be some documentation telling that there is no guarantees than doing that will work).


src/Metadata/Resource/Factory/FormatsResourceMetadataFactory.php
```diff
@@ -0,0 +1,52 @@
...
+     * Creates a resource metadata.
+     * PATCH formats are normalized in the OperationResourceMetadataFactory class.
+     *
+     * @see OperationResourceMetadataFactory
+     *
+     * @throws ResourceClassNotFoundException
+     */
+    public function create(string $resourceClass): ResourceMetadata
+    {
+        $resourceMetadata = $this->decorated->create($resourceClass);
```
antograssiot: to get rid of the `FormatProvider`, we will need the method `private FormatProvider::getOperationFormats()` to be moved and keep I think.
That's what provide shortcut allowing to only specify format name if the mimes type is defined in the configuration for example

dkarlovi:
> Supporting JSON Patch (RFC 6902) isn't a a goal as this format is not very popular in the wild

As [noted before](https://github.com/api-platform/core/issues/759#issuecomment-422010608), 

> It solves the partial collection management problem well and IMO it should be kept as a future option.

So having it not be "popular" doesn't mean it might not become so if well supported here.

Not saying you should implement it in this PR, of course.

dunglas:
> So having it not be "popular" doesn't mean it might not become so if well supported here.

I'll be glad to merge support for it if someone contribute it (it's why the new system is designed to support several patch formats)! To clarify: it's not on **my** todo list (yet)!

src/EventListener/DeserializeListener.php
```diff
@@ -39,28 +38,27 @@ final class DeserializeListener
...
+    private $formats;
     private $formatsProvider;
-    private $formatMatcher;
 
     /**
-     * @throws InvalidArgumentException
+     * @param ResourceMetadataFactoryInterface|FormatsProviderInterface|array $resourceMetadataFactory
      */
-    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, /* FormatsProviderInterface */$formatsProvider, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
+    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, $resourceMetadataFactory, ResourceMetadataFactoryInterface $legacyResourceMetadataFactory = null)
```
bendavies: how is this not a BC break?

dunglas: Why would it be? It's even covered by a test.

bendavies: correct me if i'm wrong, but `$formatsProvider` was the 3rd parameter, which was deprecated in 2.3 (line 54 below), but `$formatsProvider` has been removed. can't you only remove it in 3.0?

bendavies: ah i see now, you've just lost the deprecation warning for usage or `FormatsProviderInterface`? it needs to remain?

bendavies: it was 2.3, but now the warning says 2.5

bendavies:
`Find a way to block partial PUT` how would you prevent a BC break here?
partial PUT is how we currently do PATCH of course. PUT with denormalization groups.

src/Serializer/SerializerContextBuilder.php
```diff
@@ -89,6 +91,14 @@ public function createFromRequest(Request $request, bool $normalization, array $
 
         unset($context[DocumentationNormalizer::SWAGGER_DEFINITION_NAME]);
 
+        if (
```
bendavies: does this mean it will not be possible to null a value during a PATCH?
Is this standard PATCH semantics?

bendavies: from my reading, this goes against the json merge patch spec?

dunglas: Actually, it’s the only way to comply with the JSON Merge Patch RFC:

> Null values in the merge patch are given special meaning to indicate the removal of existing values in the target.

>  This design means that merge patch documents are suitable for describing modifications to JSON documents that primarily use objects for their structure and do not make use of explicit null values.  The merge patch format is not appropriate for all JSON syntaxes.

So we need to ignore null values when serializing. Sending patches contains NULL will have the effect to delete the key from the JSON representation of this resource.

bendavies: yep, my bad. i thought this was for deserialization.

jewome62:
Do you think implement the Header Accept-Patch ?
https://tools.ietf.org/html/rfc5789#section-3.1

That's just the new `patchFormat` but this is only on OPTIONS request.
So that's managed by `nelmio/cors-bundle` ?

dunglas:
yes it should be implemented in API Platform, but in a follow up PR (and maybe not in 2.5).

src/EventListener/DeserializeListener.php
dunglas: Because the deprecation has changed.

dunglas:
> how would you prevent a BC break here? partial PUT is how we currently do PATCH of course. PUT with denormalization groups.

It would be opt-in in 2.x, and opt-out in 3.x. But currently I struggle to find a god way to implement this: https://twitter.com/dunglas/status/1148578034787782656

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -87,9 +87,15 @@ public function load(array $configs, ContainerBuilder $container): void
         $config = $this->processConfiguration($configuration, $configs);
 
         $formats = $this->getFormats($config['formats']);
+        $patchFormats = $this->getFormats($config['patch_formats']);
         $errorFormats = $this->getFormats($config['error_formats']);
 
-        $this->registerCommonConfiguration($container, $config, $loader, $formats, $errorFormats);
+        // Backward Compatibility layer
+        if (isset($formats['jsonapi']) && !isset($patchFormats['jsonapi'])) {
```
teohhanhui: I really think we should deprecate specifying the format keys, and just use the mime type directly.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -163,6 +163,7 @@ public function getConfigTreeBuilder()
             'json' => ['mime_types' => ['application/json']], // Swagger support
             'html' => ['mime_types' => ['text/html']], // Swagger UI support
         ]);
+        $this->addFormatSection($rootNode, 'patch_formats', []);
```
teohhanhui: This doesn't seem like a good default?

src/EventListener/AddFormatListener.php
```diff
@@ -31,26 +31,28 @@
...
-    private $mimeTypes;
     private $formatsProvider;
     private $formatMatcher;
 
     /**
-     * @throws InvalidArgumentException
+     * @param ResourceMetadataFactoryInterface|FormatsProviderInterface|array $resourceMetadataFactory
      */
-    public function __construct(Negotiator $negotiator, /* FormatsProviderInterface */ $formatsProvider)
+    public function __construct(Negotiator $negotiator, $resourceMetadataFactory, array $formats = [])
```
teohhanhui: I think we need to do something like this in so many places. Otherwise we'll end up with lots of holes in our arguments in 3.0 :laughing:

src/EventListener/AddFormatListener.php
```diff
@@ -63,33 +65,54 @@ public function __construct(Negotiator $negotiator, /* FormatsProviderInterface
     public function onKernelRequest(GetResponseEvent $event): void
     {
         $request = $event->getRequest();
-        if (!($request->attributes->has('_api_resource_class') || $request->attributes->getBoolean('_api_respond', false) || $request->attributes->getBoolean('_graphql', false))) {
+        if (
+                !($request->attributes->has('_api_resource_class')
```
teohhanhui: This is indented too much.

src/EventListener/AddFormatListener.php
```diff
@@ -130,37 +153,29 @@ private function addRequestFormats(Request $request, array $formats): void
     }
 
     /**
-     * Populates the $mimeTypes property.
+     * Retries the flattened list of MIME types.
```
teohhanhui: Retries -> Retrieves

src/EventListener/DeserializeListener.php
teohhanhui: We should remove the typehint for `$legacyResourceMetadataFactory`, right? And check that in the constructor body?

PR #2897, opened 2019-07-02
Fix nested filtering with name converters

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2893
| License       | MIT
| Doc PR        | 

This will ensure that nested properties will pass through name converters by parts.

PR #2906, opened 2019-07-04
Use workaround for a phpstan error

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Use a workaround for a phpstan error ignored in https://github.com/api-platform/core/pull/2897

See https://github.com/phpstan/phpstan/issues/1905#issuecomment-464388880 for some context. But in short:

phpstan does not infer property type from constructor argument type, and it's not planned to be supported

I believe this would be a problem for any project that follows the current Symfony CS of not adding `@var` on a property when the matching argument is already type-hinted in the constructor. We've been told that IDEs support that just fine, but we will not have relief in phpstan until we can upgrade to PHP 7.4


teohhanhui:
Travis failures seem unrelated. Is the build broken on `master`, @dunglas?

dunglas:
@teohhanhui it looks like the `lowest` build is indeed broken. I'm on it.

dunglas:
A rebase and it should be green!

PR #2884, opened 2019-06-26
Allow data provider to throw ItemNotFoundException

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Converted to `NotFoundHttpException` in `ReadListener`

TODO:

- [ ] Add tests

PR #2917, opened 2019-07-08
Simplify "interface as resource" test case

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

Actually needed in #2884, but to ease maintenance, this commit is cherry-picked for `2.4`.


teohhanhui:
AppVeyor failure due to symfony/symfony#32395

tests/Fixtures/TestBundle/DataProvider/TaxonItemDataProvider.php
```diff
@@ -41,6 +44,8 @@ public function supports(string $resourceClass, string $operationName = null, ar
      */
     public function getItem(string $resourceClass, $id, string $operationName = null, array $context = [])
     {
-        return $this->managerRegistry->getRepository(Taxon::class)->find($id);
+        return $this->managerRegistry->getRepository($this->orm ? Taxon::class : TaxonDocument::class)->findOneBy([
```
alanpoulain: Maybe we can have a `get_class` for the manager registry instead of `$orm`?

teohhanhui: The problem is, we cannot tell whether we're using the ORM or ODM, as it's the same interface as `$resourceClass` for both.

teohhanhui: Oh, I get what you mean now. But that seems hacky.

teohhanhui: And in fact, both are https://github.com/symfony/symfony/blob/v4.3.2/src/Symfony/Bridge/Doctrine/ManagerRegistry.php so it doesn't help.

Issue #2896, opened 2019-07-01
Fix remaining Symfony deprecations


dunglas:
I figured out that many Symfony deprecations aren't fixed yet. They are just hidden by our default PHPUnit Bridge configuration.

To display these deprecations, you can run `SYMFONY_DEPRECATIONS_HELPER=max[total]=0 vendor/bin/phpunit`.

Pull Requests welcome!

zairigimad:
Hello @dunglas , I think the deprecations are part of FosUserBundle.

teohhanhui:
> SYMFONY_DEPRECATIONS_HELPER=max[total]=0

we already use that in one of our Travis jobs, right?

PR #2920, opened 2019-07-08
Fix deprecation notices in 2.4

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2896
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\EventListener;
 
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\EventListener\ExceptionListener as BaseExceptionListener;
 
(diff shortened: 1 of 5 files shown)


dunglas:
Thanks @meyerbaptiste

src/EventListener/ExceptionListener.php
```diff
@@ -23,8 +24,15 @@
...
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-final class ExceptionListener extends BaseExceptionListener
+final class ExceptionListener
 {
+    private $exceptionListener;
+
+    public function __construct($controller, LoggerInterface $logger = null, $debug = false, string $charset = null, $fileLinkFormat = null)
+    {
+        $this->exceptionListener = new BaseExceptionListener($controller, $logger, $debug, $charset, $fileLinkFormat);
```
teohhanhui: @meyerbaptiste In which case would the constructor have these arguments? I can't find anything in Symfony.

teohhanhui: Oh, looks like it was reverted: https://github.com/symfony/symfony/pull/31868

So we don't need to handle that.

PR #2922, opened 2019-07-09
Add error code for JsonApi errors.

gries:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Hi,
I added the possibility to specify an error code for JsonApi Errors as specified in the JsonApi Spec.
https://jsonapi.org/format/#errors

In short when throwing an Exception like:
```
throw new MyError('my bad error', 1234);
```
the Normalizer now uses the Exception Code like so:
```
{
  "title": "..",
  "description": "...",
  "code": 1234
}
```
It's a minor change so I'm not sure if master is the correct base branch.

### DIFF
--- a/src/Exception/ErrorCodeSerializableInterface.php
+++ b/src/Exception/ErrorCodeSerializableInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * An exception which has a serializable application-specific error code.
+ */
+interface ErrorCodeSerializableInterface
+{
+    /**
+     * Gets the application-specific error code.
+     */
+    public static function getErrorCode(): string;
+}
--- a/src/JsonApi/Serializer/ErrorNormalizer.php
+++ b/src/JsonApi/Serializer/ErrorNormalizer.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\JsonApi\Serializer;
 
-use ApiPlatform\Core\Exception\ErrorCodeSerializableInterface;
 use ApiPlatform\Core\Problem\Serializer\ErrorNormalizerTrait;
 use Symfony\Component\Debug\Exception\FlattenException as LegacyFlattenException;
 use Symfony\Component\ErrorHandler\Exception\FlattenException;
--- a/src/JsonApi/Serializer/ErrorNormalizer.php
+++ b/src/JsonApi/Serializer/ErrorNormalizer.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\JsonApi\Serializer;
 
+use ApiPlatform\Core\Exception\ErrorCodeSerializableInterface;
 use ApiPlatform\Core\Problem\Serializer\ErrorNormalizerTrait;
 use Symfony\Component\Debug\Exception\FlattenException as LegacyFlattenException;
 use Symfony\Component\ErrorHandler\Exception\FlattenException;
@@ -42,12 +43,16 @@ final class ErrorNormalizer implements NormalizerInterface, CacheableSupportsMet
         $this->defaultContext = array_merge($this->defaultContext, $defaultContext);
     }
 
-    public function normalize($object, $format = null, array $context = [])
+    public function normalize($object, string $format = null, array $context = [])
     {
         $data = [
             'title' => $context[self::TITLE] ?? $this->defaultContext[self::TITLE],
             'description' => $this->getErrorMessage($object, $context, $this->debug),
         ];
+
+        if (null !== $errorCode = $this->getErrorCode($object)) {
+            $data['code'] = $errorCode;
+        }
 
         if ($this->debug && null !== $trace = $object->getTrace()) {
             $data['trace'] = $trace;
--- a/src/JsonApi/Serializer/ErrorNormalizer.php
+++ b/src/JsonApi/Serializer/ErrorNormalizer.php
@@ -43,7 +43,7 @@ final class ErrorNormalizer implements NormalizerInterface, CacheableSupportsMet
         $this->defaultContext = array_merge($this->defaultContext, $defaultContext);
     }
 
-    public function normalize($object, string $format = null, array $context = [])
+    public function normalize($object, $format = null, array $context = [])
     {
         $data = [
             'title' => $context[self::TITLE] ?? $this->defaultContext[self::TITLE],
--- a/src/Problem/Serializer/ErrorNormalizerTrait.php
+++ b/src/Problem/Serializer/ErrorNormalizerTrait.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Problem\Serializer;
 
+use ApiPlatform\Core\Exception\ErrorCodeSerializableInterface;
 use Symfony\Component\Debug\Exception\FlattenException as LegacyFlattenException;
 use Symfony\Component\ErrorHandler\Exception\FlattenException;
 use Symfony\Component\HttpFoundation\Response;
@@ -35,5 +36,19 @@ trait ErrorNormalizerTrait
         }
 
         return $message;
+    }
+
+    private function getErrorCode($object): ?string
+    {
+        $exceptionClass = get_class($object);
+        if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
+            $exceptionClass = $object->getClass();
+        }
+
+        if (is_a($exceptionClass, ErrorCodeSerializableInterface::class, true)) {
+            return $exceptionClass::getErrorCode();
+        }
+
+        return null;
     }
 }
--- a/src/Problem/Serializer/ErrorNormalizerTrait.php
+++ b/src/Problem/Serializer/ErrorNormalizerTrait.php
@@ -40,9 +40,10 @@ trait ErrorNormalizerTrait
 
     private function getErrorCode($object): ?string
     {
-        $exceptionClass = get_class($object);
         if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
             $exceptionClass = $object->getClass();
+        } else {
+            $exceptionClass = get_class($object);
         }
 
         if (is_a($exceptionClass, ErrorCodeSerializableInterface::class, true)) {
--- a/src/Problem/Serializer/ErrorNormalizerTrait.php
+++ b/src/Problem/Serializer/ErrorNormalizerTrait.php
@@ -43,7 +43,7 @@ trait ErrorNormalizerTrait
         if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
             $exceptionClass = $object->getClass();
         } else {
-            $exceptionClass = get_class($object);
+            $exceptionClass = \get_class($object);
         }
 
         if (is_a($exceptionClass, ErrorCodeSerializableInterface::class, true)) {
--- a/tests/JsonApi/Serializer/ErrorNormalizerTest.php
+++ b/tests/JsonApi/Serializer/ErrorNormalizerTest.php
@@ -16,7 +16,7 @@ namespace ApiPlatform\Core\Tests\JsonApi\Serializer;
 use ApiPlatform\Core\JsonApi\Serializer\ErrorNormalizer;
 use ApiPlatform\Core\Tests\Mock\Exception\ErrorCodeSerializable;
 use PHPUnit\Framework\TestCase;
-use Symfony\Component\ErrorHandler\Exception\FlattenException;
+use Symfony\Component\Debug\Exception\FlattenException;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
--- a/tests/JsonApi/Serializer/ErrorNormalizerTest.php
+++ b/tests/JsonApi/Serializer/ErrorNormalizerTest.php
@@ -14,8 +14,9 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Tests\JsonApi\Serializer;
 
 use ApiPlatform\Core\JsonApi\Serializer\ErrorNormalizer;
+use ApiPlatform\Core\Tests\Mock\Exception\ErrorCodeSerializable;
 use PHPUnit\Framework\TestCase;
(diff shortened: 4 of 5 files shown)


src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -52,6 +52,10 @@ public function normalize($object, $format = null, array $context = [])
             $data['trace'] = $trace;
         }
 
+        if (0 !== $code = $object->getCode()) {
```
dunglas: Could you move this block before the one setting the `trace` key?

src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -52,6 +52,10 @@ public function normalize($object, $format = null, array $context = [])
...
+            $data['code'] = $code;
```
dunglas: According to the JSON:API spec, the code must be casted as a string.

dunglas:
The target branch is good, as it's a new feature!

gries:
I made the requested changes rebased and squashed hope it's fine now.

gries:
any update on this?

soyuka:
There are some concerns on whether this is a good idea or not (exposing exception codes).

teohhanhui:
The JSON:API says it should be an "application-specific" error code. We cannot assume that all exceptions thrown are controlled by / originate from the application. I don't think it's a good idea to map the exception code to error code indiscriminately.

I'm -1 unless we introduce a mapping or some other way of doing that conversion.

gries:
Currently the exception code is only set of its not the default code which is 0.
So somewhere the application must have explicitly set the exception code.

If you the idea of an exception code to error code mapping sounds interesting so if you want I can add that if you point me to the right config place where I should put it.

teohhanhui:
> Currently the exception code is only set of its not the default code which is 0.
> So somewhere the application must have explicitly set the exception code.

I think that's a dangerous assumption. The exception code could be set by another library, not the app. So we may be exposing some exception code that makes no sense to the client.

But the main problem is that the error code could be a string. There is no reason to just take the exception code. So yeah, let's introduce a mapping or an interface like this:

```php
interface ErrorCodeExceptionInterface extends \Throwable
{
    public static function getErrorCode(): string;
}
```

(Static method because the error code should not change between different instances of the class.)

gries:
Hi,
I have added an implementation that uses an `ErrorCodeExceptionInterface` please have a look, I'm not fully happy with the way it is implemented in the normalizer but I saw no other way as we are dealing with FlattenExceptions.

gries:
just a bump to see if something if missing or can this be merged?

src/JsonApi/Serializer/ErrorNormalizer.php
gries: this is outdated and no longer needed as the ExceptionInterface makes sure getCode returns a string.

src/JsonApi/Serializer/ErrorNormalizer.php
gries: this is outdated and no longer needed as the ExceptionInterface makes sure getCode returns a string.

teohhanhui:
Closing as I've explained why this is not the right approach.

teohhanhui:
Sorry, I see you've updated to a different approach.

src/Exception/ErrorCodeExceptionInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * Interface ErrorCodeExceptionInterface.
+ */
+interface ErrorCodeExceptionInterface extends \Throwable
```
teohhanhui:
```suggestion
interface ErrorCodeAvailableInterface
```

Not sure about this name but it should not extend `\Throwable`. This interface should be an optional interface which custom exception classes choose to implement in addition to the usual.

src/Exception/ErrorCodeExceptionInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * Interface ErrorCodeExceptionInterface.
```
teohhanhui: We need a descriptive summary here.

tests/Mock/Exception/MockErrorCodeException.php
```diff
@@ -0,0 +1,27 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Mock\Exception;
+
+use ApiPlatform\Core\Exception\ErrorCodeExceptionInterface;
+
+class MockErrorCodeException extends \Exception implements ErrorCodeExceptionInterface
```
teohhanhui: The `Mock` part of the name is completely redundant. It's already in the namespace.

src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -48,6 +49,12 @@ public function normalize($object, $format = null, array $context = [])
             'description' => $this->getErrorMessage($object, $context, $this->debug),
         ];
 
+        if ($object instanceof FlattenException) {
```
teohhanhui: This is wrong. We should also support `\Exception`s other than `FlattenException`.

src/Exception/ErrorCodeExceptionInterface.php
gries: I hope this suffices.

src/JsonApi/Serializer/ErrorNormalizer.php
gries: I have added a handling for both FlattenException + "normal" Exception, 
I also added a testcase for both.

tests/Mock/Exception/MockErrorCodeException.php
gries: removed

src/Exception/ErrorCodeExceptionInterface.php
gries: updated the name and removed the extends.

src/Exception/ErrorCodeSerializableInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+/**
+ * Interface ErrorCodeExceptionInterface.
+ *
+ * An Exception that supports a custom error-code to be used during normalization.
+ */
```
teohhanhui:
```suggestion
/**
 * An exception which has a serializable custom error code.
 */
```

src/Exception/ErrorCodeSerializableInterface.php
```diff
@@ -0,0 +1,27 @@
...
+interface ErrorCodeSerializableInterface
+{
+    /**
+     * Application specific error code used when normalizing an Exception.
```
teohhanhui:
```suggestion
     * Gets the application-specific error code.
```

src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -49,6 +50,15 @@ public function normalize($object, $format = null, array $context = [])
             'description' => $this->getErrorMessage($object, $context, $this->debug),
         ];
 
+        $exceptionClass = get_class($object);
+        if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
+            $exceptionClass = $object->getClass();
+        }
+
+        if (is_subclass_of($exceptionClass, ErrorCodeSerializableInterface::class)) {
```
teohhanhui:
```suggestion
        if (is_a($exceptionClass, ErrorCodeSerializableInterface::class, true)) {
```

src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -49,6 +50,15 @@ public function normalize($object, $format = null, array $context = [])
...
+            $data['code'] = \call_user_func([$exceptionClass, 'getErrorCode']);
```
teohhanhui:
```suggestion
            $data['code'] = $exceptionClass::getErrorCode();
```

tests/JsonApi/Serializer/ErrorNormalizerTest.php
```diff
@@ -61,6 +62,42 @@ public function testNormalize($status, $originalMessage, $debug)
         $this->assertEquals($expected, $normalizer->normalize($exception, ErrorNormalizer::FORMAT, ['statusCode' => $status]));
     }
 
+    public function testNormalizeAnExceptionWithCustomErrorCode()
```
teohhanhui:
```suggestion
    public function testNormalizeAnExceptionWithCustomErrorCode(): void
```

tests/JsonApi/Serializer/ErrorNormalizerTest.php
```diff
@@ -61,6 +62,42 @@ public function testNormalize($status, $originalMessage, $debug)
...
+        $expected = [
+            'title' => 'An error occurred',
+            'description' => 'my-message',
+            'code' => ErrorCodeSerializable::getErrorCode(),
+        ];
+
+        $this->assertEquals($expected, $normalizer->normalize($exception, ErrorNormalizer::FORMAT, ['statusCode' => $status]));
+    }
+
+    public function testNormalizeAFlattenExceptionWithCustomErrorCode()
```
teohhanhui:
```suggestion
    public function testNormalizeAFlattenExceptionWithCustomErrorCode(): void
```

tests/Mock/Exception/ErrorCodeSerializable.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace ApiPlatform\Core\Tests\Mock\Exception;
+
+use ApiPlatform\Core\Exception\ErrorCodeSerializableInterface;
+
+class ErrorCodeSerializable extends \Exception implements ErrorCodeSerializableInterface
+{
+    /**
+     * Application specific error code used when normalizing an Exception.
+     */
```
teohhanhui:
```suggestion
    /**
     * {@inheritdoc}
     */
```

teohhanhui:
Only minor changes. LGTM! :+1:

gries:
I have applied your suggestions thanks for the help :)

src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -49,6 +50,15 @@ public function normalize($object, $format = null, array $context = [])
             'description' => $this->getErrorMessage($object, $context, $this->debug),
         ];
 
+        $exceptionClass = get_class($object);
+        if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
+            $exceptionClass = $object->getClass();
+        }
+
+        if (is_a($exceptionClass, ErrorCodeSerializableInterface::class, true)) {
+            $data['code'] = $exceptionClass::getErrorCode();
```
teohhanhui: I understand you're only interested in JSON:API, but this got me wondering, we should do something similar for other error formats, no?

teohhanhui: WDYT about extracting this as `getErrorCode` method in `ErrorNormalizerTrait`?

gries: sure if there are similar fields other formats.
in hydra I only find the "@id" field that can be used to reference an error but I think that's not what this is for. (http://www.hydra-cg.com/spec/latest/core/#description-of-http-status-codes-and-errors)
Maybe GraphQL has got something but I never used it before.

teohhanhui: For Hydra, `@id` seems like the way to go. But of course it should be a valid IRI.

gries: I could try to implement that but I have no idea how to generate a valid IRI for an error code,
that would imply that there are error resources with details that can be added etc. right?

teohhanhui: Hmm, yeah, I think something like that needs more thought. For now, let's just extract this method into the trait. :smile:

gries: something like:
```
ErrorNormalizerTrait
{
  function getErrorCode(\Exception $exception) : string
}
?

Comment (teohhanhui) [reply src/JsonApi/Serializer/ErrorNormalizer.php]: There is already an `ErrorNormalizerTrait`: https://github.com/api-platform/core/blob/master/src/Problem/Serializer/ErrorNormalizerTrait.php

Sure, that method signature looks okay to me:

```php
private function getErrorCode(\Exception $exception): string
```

gries: I have updated the code now to use the trait.
I stuck with $object instead of \Exception $exception because FlattenExceptions dont seem to extend Exceptions.
I also made the return value nullable.

src/Problem/Serializer/ErrorNormalizerTrait.php
```diff
@@ -36,4 +37,18 @@ private function getErrorMessage($object, array $context, bool $debug = false):
 
         return $message;
     }
+
+    private function getErrorCode($object): ?string
+    {
+        $exceptionClass = get_class($object);
+        if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
+            $exceptionClass = $object->getClass();
+        }
```
teohhanhui:
```suggestion
        if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
            $exceptionClass = $object->getClass();
        } else {
            $exceptionClass = get_class($object);
        }
```

src/JsonApi/Serializer/ErrorNormalizer.php
```diff
@@ -42,13 +43,17 @@ public function __construct(bool $debug = false, array $defaultContext = [])
         $this->defaultContext = array_merge($this->defaultContext, $defaultContext);
     }
 
-    public function normalize($object, $format = null, array $context = [])
+    public function normalize($object, string $format = null, array $context = [])
```
teohhanhui: Please do not change this.

teohhanhui: It's the cause of the job failure.

gries: ok I had symfony5 installed which has a changed the format.

gries:
@teohhanhui I fixed errors caused by my PR I think the other problems are not related to my changes.

teohhanhui:
Thanks @gries! :tada:

Issue #2856, opened 2019-06-12
symfony/messenger v4.3 wraps any exception into HandlerFailedException


karser:
Hi folks!

Before symfony/messenger v4.3, the exceptions were transparent - you got what you threw. So that you could specify them in the exception_to_status map and receive meaningful statuses.
As of v4.3, symfony/messenger [started wrapping any exception into HandlerFailedException](https://github.com/symfony/messenger/blob/master/CHANGELOG.md), so api-platform always returns 500.
Here is a similar issue described in SO https://stackoverflow.com/questions/55558350/custom-exception-from-messenger-handler

So the idea is to recover previous exception and match to exception_to_status map.
```
if ($exception instanceof HandlerFailedException) {
    $exception = $exception->getPrevious();
}
```
What do you think of this approach? Does it make sense or I'm missing something?

dunglas:
It looks a bit hacky to me... but I don't have another solution in mind. Do you have some thoughts about that @sroze?

karser:
I see, v2.4.4 doesn't support it yet, it has `"conflict": {"symfony/messenger": ">=4.3"}`. So I must have tested v2.4.3 + messenger v4.3.

PR #2930, opened 2019-07-11
Unwrap exception thrown in handler for Symfony Messenger 4.3

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes?
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2856
| License       | MIT
| Doc PR        | N/A

As suggested by @karser

Also, since coverage jobs were failing, due to a `phpdbg` bug (see https://github.com/symfony/symfony/issues/32521):

Allow phpunit 8 and use `pcov` for coverage jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,13 @@ commands:
           command: |
             sudo pecl install mongodb-1.5.4
             sudo docker-php-ext-enable mongodb
+  install-pcov-php-extension:
+    steps:
+      - run:
+          name: Install pcov PHP extension
+          command: |
+            sudo pecl install pcov-1.0.6
+            sudo docker-php-ext-enable pcov
   merge-code-coverage-reports:
     parameters:
       dir:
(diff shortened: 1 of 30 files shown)


teohhanhui:
CircleCI is fucked for some reason... :laughing:

teohhanhui:
I'm wondering if perhaps the `ExceptionListener` is not the right place to unwrap the exception. Maybe we should add a try-catch here instead?

https://github.com/api-platform/core/blob/a62a077f69332ccd8e525cd96e7a43e522b60f79/src/Bridge/Symfony/Messenger/DataPersister.php#L78

https://github.com/api-platform/core/blob/a62a077f69332ccd8e525cd96e7a43e522b60f79/src/Bridge/Symfony/Messenger/DataPersister.php#L91

teohhanhui:
Hmm... We still have a problem in the case of nested message handlers, where a message handler dispatches another message, such as that mentioned in https://symfony.com/doc/current/messenger/message-recorder.html

> A message handler can dispatch new messages during execution, to either the same or a different bus (if the application has multiple buses).

When a nested message handler fails, we end up with a `HandlerFailedException` wrapped in another `HandlerFailedException`.

teohhanhui:
> I'm wondering if perhaps the ExceptionListener is not the right place to unwrap the exception.

We've encountered another problem: https://github.com/api-platform/core/blob/6fe4593182cb1878630d4acb34f70d6077581c23/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php

It doesn't sound right to me that each `kernel.exception` listener needs to unwrap this.

@sroze Do you have any thoughts on this?

Issue #2931, opened 2019-07-12
Entity inheritance normalises based on parent class


jmwri:
Tested against 2.4.4 and 2.4.5.

It seems that the normalisation process will only use the properties defined the the parent class if you use inheritance within entities.

Example entities:

Contact.php
```
/**
 * Class Contact
 * @ApiResource()
 */
abstract class Contact
{
    /**
     * @ApiProperty(
     *     description="ID of contact",
     *     identifier=true,
     * )
     * @Assert\Type(type="string")
     * @Assert\Length(min="36", max="36")
     * @var ?string
     */
    protected $id;

    /**
     * @ApiProperty(
     *     description="Type of contact"
     * )
     * @Assert\NotBlank()
     * @var string
     */
    protected $type;

    /**
     * Contact constructor.
     * @param string $type
     * @param string|null $id
     */
    public function __construct(
        string $type,
        ?string $id = null
    ) {
        $this->setType($type);
        $this->setId($id);
    }

    /**
     * @return string
     */
    public function getType(): string
    {
        return $this->type;
    }

    /**
     * @param string $type
     * @return Contact
     */
    protected function setType(string $type): Contact
    {
        $this->type = $type;
        return $this;
    }

    /**
     * @return mixed
     */
    public function getId(): ?string
    {
        return $this->id;
    }

    /**
     * @param string $id
     * @return Contact
     */
    public function setId($id): Contact
    {
        $this->id = $id;
        return $this;
    }
}
```

ContactA.php
```
/**
 * Class ContactA
 */
class ContactA extends Contact
{
    /**
     * @ApiProperty(
     *     description="Property only existing on contact type A"
     * )
     * @var ?string
     */
    protected $propertyA;

    /**
     * ImplementationContact constructor.
     * @param string|null $id
     * @param string|null $propertyA
     */
    public function __construct(
        ?string $id = null,
        ?string $propertyA = null
    ) {
        parent::__construct('a', $id);
        $this->setPropertyA($propertyA);
    }

    /**
     * @return string|null
     */
    public function getPropertyA(): ?string
    {
        return $this->propertyA;
    }

    /**
     * @param string $propertyA
     * @return ContactA
     */
    public function setPropertyA(?string $propertyA): ContactA
    {
        $this->propertyA = $propertyA;
        return $this;
    }
}
```

ContactB.php
```
/**
 * Class ContactB
 */
class ContactB extends Contact
{
    /**
     * @ApiProperty(
[...]

PR #2932, opened 2019-07-12
get properties of class passed to geteAllowedAttributes rather than f…

jmwri:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2931 
| License       | MIT
| Doc PR        | 

I've documented the issue in #2931.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -322,11 +322,12 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     protected function getAllowedAttributes($classOrObject, array $context, $attributesAsString = false)
     {
         $options = $this->getFactoryOptions($context);
-        $propertyNames = $this->propertyNameCollectionFactory->create($context['resource_class'], $options);
+        $class = \is_object($classOrObject) ? $this->getObjectClass($classOrObject) : $classOrObject;
+        $propertyNames = $this->propertyNameCollectionFactory->create($class, $options);
 
         $allowedAttributes = [];
         foreach ($propertyNames as $propertyName) {
-            $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $propertyName, $options);
+            $propertyMetadata = $this->propertyMetadataFactory->create($class, $propertyName, $options);
 
             if (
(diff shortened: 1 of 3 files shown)


jmwri:
@dunglas CI tests are failing due to memory issues. I assume this is a general issue and nothing to do with this PR?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -322,11 +322,12 @@ protected function extractAttributes($object, $format = null, array $context = [
     protected function getAllowedAttributes($classOrObject, array $context, $attributesAsString = false)
     {
         $options = $this->getFactoryOptions($context);
-        $propertyNames = $this->propertyNameCollectionFactory->create($context['resource_class'], $options);
+        $class = \is_object($classOrObject) ? \get_class($classOrObject) : $classOrObject;
```
soyuka: it should use `$this->getObjectClass`

jmwri:
Hi @soyuka,

I've updated the PR as per your comment :)

soyuka:
can you try to rebase / make tests pass?

teohhanhui:
Sorry, this does not make sense.

If you're asking for a `ResourceInterface`, and the `@type` in the response is `ResourceInterface`, then you're only supposed to get what's exposed by the `ResourceInterface` API resource. Otherwise we're leaking information and that's dangerous.

jmwri:
I've replied in the related issue. Is this an issue of implementation? Would it make more sense to check if the `$class == $context['resource_class'] || is_subclass_of($class, $context['resource_class']`?

teohhanhui:
@jmwri It's not a problem with your implementation. The current behaviour seems correct to me, and this PR actually will reintroduce an information leak. So the change is undesirable.

parsingphase:
@teohhanhui Under what cases would you create a subclass of a displayable entity and not expect that subclass to be displayed? I appreciate the concern if this was a more general leak but I'd have to assume that users of the API would expect inheritors of entities to work this way.

teohhanhui:
You can have any number of subclasses, and they shouldn't leak information unless you explicitly say that those subclasses should be exposed, by marking them as API resource classes. See discussion in https://github.com/api-platform/core/issues/2931#issuecomment-513138882

PR #2944, opened 2019-07-20
[GraphQL] Disable some steps by setting operation attributes

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        | Will be added in https://github.com/api-platform/docs/pull/830

Since the introduction of custom queries and mutations, a needed feature was missing: the possibility to disable some "steps" the main resolver is doing.

For instance, if we want to create a custom mutation and letting it persist the mutated item as it wants, it was not possible before (`null` could be returned but the returned data was also `null`).

Now it can be done by disabling the `write` step in the operation attributes:
```php
<?php

namespace App\Model;

use App\Resolver\MyMutationResolver;

/**
 * @ApiResource(graphql={
 *     "noWrite"={
 *         "mutation"=MyMutationResolver::class,
 *         "write"=false
 *     }
 * })
 */
```
The REST part has already the possibility to do so (see https://api-platform.com/docs/core/events/#built-in-event-listeners).

However this PR also shows that the resolvers are getting bigger and bigger and a refactor is required.
I think adopting an architecture based on events could be a solution (the resolver only sends the events and the current code will be divided into the listeners) or we can introduce new services and interfaces to do so. WDYT @api-platform/core-team?

Fortunately, I think the added `if` clauses can help to break the different steps the resolver is doing.

### DIFF
--- a/features/graphql/collection.feature
+++ b/features/graphql/collection.feature
@@ -518,6 +518,35 @@ Feature: GraphQL collection support
     """
 
   @createSchema
+  Scenario: Custom collection query with read and serialize set to false
+    Given there are 2 dummyCustomQuery objects
+    When I send the following GraphQL request:
+    """
+    {
+      testCollectionNoReadAndSerializeDummyCustomQueries {
+        edges {
+          node {
+            message
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON should be equal to:
+    """
+    {
+      "data": {
+        "testCollectionNoReadAndSerializeDummyCustomQueries": {
+          "edges": []
+        }
+      }
+    }
+    """
+
+  @createSchema
   Scenario: Custom collection query with custom arguments
     Given there are 2 dummyCustomQuery objects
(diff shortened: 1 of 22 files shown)


lukasluecke:
I think using events would be awesome!

Maybe we could also work on moving the schema generation to the same approach?

This would allow for things like:
- Adding a "dynamic" field to a GraphQL resource which doesn't exist on the API resource without having to use a DTO or something (field resolvers)
- Modifying a single argument of a GraphQL query, e.g. making it required, changing it's type (ID <> String) without having to specify all of them again

alanpoulain:
Sure we should find a way to provide these possibilities as well.
I'm not sure using the events is necessarily the answer. It could also be extension points, for instance by using tagged services.

dunglas:
Decorators, as everywhere else in the API Platform code base, should do the trick right?

alanpoulain:
Yes it could be an answer. It would mean introducing one service per step.

dunglas:
Indeed, it would be more SOLID code anyway, so let's do this!

PR #2949, opened 2019-07-24
Fix Elasticsearch order filter on related entity property

bastnic:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I use Elasticsearch integration and I want to sort on properties like `user.id` where `user` is a related entity and `id` is an integer. The serialization and the ES mapping allow the order to operate.

Current implementation try to check dotted properties after searching if the property exists, period. So the whole code seems to be for nothing. 

I'm not sure if the real fix should be that `hasProperty` return true even for dotted property. But currently this patch seems to be ok for me, so Hi!

### DIFF
--- a/src/Bridge/Elasticsearch/DataProvider/Filter/AbstractFilter.php
+++ b/src/Bridge/Elasticsearch/DataProvider/Filter/AbstractFilter.php
@@ -81,13 +81,13 @@ abstract class AbstractFilter implements FilterInterface
     protected function getMetadata(string $resourceClass, string $property): array
     {
         $noop = [null, null, null, null];
+        $properties = explode('.', $property);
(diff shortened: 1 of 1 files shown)


bastnic:
Humm it works for one to many relationship, but not for many to many as APIP make the asumption that colletions of objects are **nested**.

In my case, many to many are not mapped as nested as nested is quite costly in performance and we don't need them to be treated as differents objects.

soyuka:
> Humm it works for one to many relationship, but not for many to many 

We can still merge this imo, could you add some functional test?

PR #2951, opened 2019-07-25
[GraphQL] Add graphiql-explorer

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Add https://github.com/OneGraph/graphiql-explorer to the GraphiQL view.

I needed to make some tweaks to graphiql-explorer in order to make it work: https://github.com/OneGraph/graphiql-explorer/pull/22

We should probably wait for them to merge this PR. If we bump the JS dependencies, it will not work anymore.

graphiql-explorer


lukasluecke:
Another possibility would be to do it the way they do it in their "self-hosted" example (https://github.com/OneGraph/graphiql-explorer-example/blob/master/src/App.js) - just storing the schema in a state variable, which the fetcher updates. Just in case they don't merge the PR 😉

alanpoulain:
@lukasluecke Yes I know but the issue here is that you need to use the `graphql` package if you do it that way. And there is no minified version of it: you have to use it in a Webpack context, which is not our case (it would require some work to introduce it).

Issue #2796, opened 2019-05-15
How to disable REST completely?


danaki:
In my project I use GraphQL only and need no to configure REST operations. How do I disable REST and leave GraphQL support only?

teohhanhui:
That makes me sad. :cry:

But on a more serious note, it's not currently possible, because GraphQL uses the generic `application/json` MIME type.

dunglas:
You can register only the GraphQL route, instead of all API Platform routes. Replace the content of `config/routes/api_platform.yml` by:

```
api_platform_graphql:
    resource: '@ApiPlatformBundle/Resources/config/routing/graphql.xml'
```

It should do the trick.

dunglas:
Ideally, we should add a flag to disable REST, and not register the kernel listeners in this case. In the meantime the previous solution should be good enough.

lukasluecke:
@dunglas Your suggestion with the routing modifications sadly doesn't work.

The main issue is that GraphQL still uses the routes for IRI generation, not sure if that can easily be solved by still registering the routes, but not attaching the listeners?

teohhanhui:
@lukasluecke You could try manually overriding some of the built-in event listeners with the same service IDs. But we need to identify which listeners are used by GraphQL, and which are not.

teohhanhui:
Hmm... Actually I think an easier way would be to just do:

```php
<?php
// api/src/Entity/Book.php

use ApiPlatform\Core\Annotation\ApiResource;
use App\Controller\PublishBookAction;

/**
 * @ApiResource(
 *     graphql={
 *         ...
 *     },
 *     collectionOperations={
 *     },
 *     itemOperations={
 *     },
 * )
 */
class Book
{
```

lukasluecke:
Related to #2954 (since this should be the only thing that's currently still required from the REST operations, if you want GraphQL to work)

lukasluecke:
@dunglas Was this not already resolved? (#3079 and api-platform/docs#886)
Maybe we should close #2954 to prevent confusion.

dunglas:
Good catch! To entirely fix this one we should a config option to easily disable everything that can be disabled from the REST subsystem.

lukasluecke:
@dunglas I'll try to get something in for #2946 soonish, which should also make it possible to globally define these operation settings (and then have a doc entry with the exact settings to use) - do you think that would be enough?

dunglas:
Could also be related to https://github.com/api-platform/core/issues/3099 right?

[...]

PR #2954, opened 2019-07-26
WIP: Ability to remove the item operation

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | lots? https://github.com/api-platform/core/issues/2796
| License       | MIT
| Doc PR        | na

This is a proposal to make resources work without a defined GET item operation. When it fails we should fallback to a blank IRI instead of resulting in a 400 error.

Note that I think that we should make this behavior configurable. If we decide that it should be the default behavior maybe a configuration to keep the old behavior?
About the implementation:
- I'd like to put this further down the stack so that we can use the CachedRouteResolver with these (performance issue)
- I thought about adding a new `getSafeIriFromItem` that never throws (means to add a new interface that extends the current one)

This definitely breaks some current behavior but I think that it also makes ApiPlatform less strict regarding this item operation. The only solution for now is to decorate the IriConverter with our own where it patches these kind of Resources. Also, by using a blank iri we enforce a good practice regarding JSON-LD.

### DIFF
--- a/features/bootstrap/DoctrineContext.php
+++ b/features/bootstrap/DoctrineContext.php
@@ -78,6 +78,8 @@ use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\ConvertedOwner;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\ConvertedRelated;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\ConvertedString;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Customer;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DisableItemOperation;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DisableItemOperation as DisableItemOperationDocument;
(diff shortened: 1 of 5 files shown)


teohhanhui:
For the record, the correct term is "blank node identifier", not "blank IRI" :stuck_out_tongue:

But yeah, this seems like the right way forward.

lukasluecke:
@soyuka Any update on this? I think it would be really great to easily be able to completely disable the REST API if you're only interested in GraphQL. Leaving it enabled forces me to duplicate security related configuration at the moment 😕

soyuka:
> @soyuka Any update on this? I think it would be really great to easily be able to completely disable the REST API if you're only interested in GraphQL. Leaving it enabled forces me to duplicate security related configuration at the moment confused

I need to continue working on this, I can make a priority out of this next week!

lukasluecke:
@soyuka Thank you very much, that would be awesome 🎉

teohhanhui:
@soyuka I think we should close this in favour of (an implementation of) #3051.

soyuka:
Superseeded by #3079 although blank node identifiers may still be an interesting feature.

PR #2956, opened 2019-07-27
[GraphQL] Add possibility to change GraphiQL path and default IDE for the entrypoint

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2501
| License       | MIT
| Doc PR        | api-platform/doc#851

The location of GraphiQL is now easily changeable.
This PR also lays the ground for adding another IDE (https://github.com/prisma/graphql-playground).


src/Bridge/Symfony/Bundle/Resources/views/Graphiql/index.html.twig
```diff
@@ -11,6 +11,10 @@
 <body>
 <div id="graphiql">Loading...</div>
 
+<script>
+    var entrypoint = '{{ entrypoint }}';
```
teohhanhui: Is it really a good idea to use an unnamespaced global variable like this?

alanpoulain: I don't think it's really harmful, but how could it be done differently?

teohhanhui: https://github.com/api-platform/core/blob/9aba73225cbeca44bce34c8250fe0c2494cbfb6f/src/Bridge/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig#L13-L14

teohhanhui: In your case it'd just be `type="text/plain"`, of course.

alanpoulain: Done thanks.

Issue #2969, opened 2019-08-07
[GraphQL] Incorrect (parent) filters in nested collections


lukasluecke:
The arguments (filters) for nested collections are based on the parent resource configuration, not on the collection's resource configuration.

lukasluecke:
I will create a PR with the fix soon 🙂

luca-simonetti:
I thought that was the expected behaviour. I see the light at the end of the tunnel now!
Do you have an ETA for this to be merged in master? Can I help somehow to make things quicker?

alanpoulain:
Fixed in master and 2.4.
@luca-simonetti either wait for a release or use the dev version.

PR #2970, opened 2019-08-07
Use correct resource configuration for filter args of nested collection

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #2969 
| License       | MIT
| Doc PR        | N/A

See issue for details.

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\GraphQl\Type;
 
+use ApiPlatform\Core\Exception\ResourceClassNotFoundException;
 use ApiPlatform\Core\GraphQl\Resolver\Factory\ResolverFactoryInterface;
 use ApiPlatform\Core\GraphQl\Type\Definition\TypeInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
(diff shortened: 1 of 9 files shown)


src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -253,15 +254,15 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
...
                     if (null === $this->filterLocator || !$this->filterLocator->has($filterId)) {
                         continue;
                     }
 
-                    foreach ($this->filterLocator->get($filterId)->getDescription($resourceClass) as $key => $value) {
+                    foreach ($this->filterLocator->get($filterId)->getDescription($className) as $key => $value) {
                         $nullable = isset($value['required']) ? !$value['required'] : true;
                         $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
-                        $graphqlFilterType = $this->convertType($filterType, false, $queryName, $mutationName, $resourceClass, $property, $depth);
+                        $graphqlFilterType = $this->convertType($filterType, false, $queryName, $mutationName, $className, $property, $depth);
```
lukasluecke: @alanpoulain This parameter seems to not be used inside the `convertType` method?

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -230,8 +230,9 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
                 $className = $this->typeBuilder->isCollection($type) && ($collectionValueType = $type->getCollectionValueType()) ? $collectionValueType->getClassName() : $type->getClassName();
             }
 
-            $args = [];
```
alanpoulain: Why the removal?

alanpoulain:
Could you explicit the issue with an example? And maybe a Behat test?

src/GraphQl/Type/FieldsBuilder.php
lukasluecke: Missed/deleted that part copying the changes from my local version 🙈 should be fixed now 🙂

lukasluecke:
The issue should be pretty clear, filters on nested collections were always using the "root" resource's filters as args, instead of the actual collection resource.

**Example**
Group entity with `name` filter, User entity with `type` filter would generate the following schema previously:
```
groups(
        name: String
    ): GroupConnection
[...]

type Group implements Node {
    id: ID!
    name: String!
    users(
        name: String
    ): UserConnection
    _id: String!
}
```

But the `users` field inside of `Group` should have a `type` argument, not `name`.

I probably won't have time for a Behat test for this in the near future, but feel free to add one 🙂

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -232,6 +232,8 @@ private function getResourceFieldConfiguration(string $resourceClass, ResourceMe
 
             $args = [];
             if (!$input && null === $mutationName && !$isStandardGraphqlType && $this->typeBuilder->isCollection($type)) {
+                $propertyResourceMetadata = $this->resourceMetadataFactory->create($className);
```
alanpoulain: Why property? Maybe we could use `$resourceMetadata` and rename the argument in `$parentResourceMetadata`?

src/GraphQl/Type/FieldsBuilder.php
alanpoulain: Not used but is useful if you create your own TypeConverter. You have an example in the tests.

src/GraphQl/Type/FieldsBuilder.php
lukasluecke: Because the method parameter for `getResourceFieldConfiguration` is `property`, and that's what this resourceMetadata belongs to - but your naming would also be fine for me 👍

alanpoulain:
OK I understand the issue now. Thanks for the explanation. I think having a non-regression test is important for this kind of issue.

lukasluecke:
I agree, but I'll be on vacation soon and need to finish up some other stuff - not sure when I'll get to it. But until then, for anyone that has the same issue, at least there is a patch available 🙂

src/GraphQl/Type/FieldsBuilder.php
alanpoulain: But this method is not used necessarily with a property :slightly_smiling_face: It's also used at the top-level.

lukasluecke: You are right, maybe `$collectionResourceMetadata` would be better?

alanpoulain: In fact we should even remove the `$resourceMetadata` argument in the method.
And I would move this initialization above the condition like this:
```php
$resourceMetadata = null;
if (!empty($className)) {
    $resourceMetadata = $this->resourceMetadataFactory->create($className);
}
```

lukasluecke: Looks good to me 👍 - also catches the other usage I've overlooked (in `mergeFilterArgs`) 🙈

alanpoulain:
Thank you @lukasluecke :slightly_smiling_face:
The code is even clearer now.

lukasluecke:
Thank you for finishing it 🙂

Issue #2891, opened 2019-06-28
More documentation info can be infered from filters and extensions 


tigitz:
Hello,

I was trying to go the "design first" route for a simple api by coding some resources without wiring them at first and then generating the swagger/open api spec from it.

I found myself adding more and more stuff about filters in my custom swagger documentation normalizer that I thought could already be inferred.

Here's a short preview:

**OrderFilter:** the allowed values are `desc` and `asc`, so it could generates the following in addition:
```json
{
  "parameters": [
    {
      "type": "string",
      "enum": [
        "asc",
        "desc"
      ]
    }
  ]
}
```

Resulting in a nice select in swagger ui:

![image](https://user-images.githubusercontent.com/1524501/60364592-e0874500-99e6-11e9-9424-f2efb57b21a2.png)

**Items per page:**  When enabled, api-platform knows about the default value, the maximum and minimum value per resource and collection operation.

**BooleanFilter & ExistsFilter:** enum of `true`,`false`,`1`, `0`

**GroupFilter**: It seems technically possible to infer the allowed serialization groups and therefore producing the right enum documentation

**PropertyFilter**: It seems technically possible to infer the exposed properties and therefore producing the right enum documentation

I might have overstepped some edge cases that make my assumptions wrong though. Let me know if it's the case.

Anyway, if these suggestions are deemed valid and interesting enough, I could work on them if needed.

dunglas:
But +1 on my side to support add support for this!

PR #2971, opened 2019-08-07
Improve order filter generated API doc

fredericbarthelet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2891 
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Bridge/Doctrine/Common/Filter/OrderFilterTrait.php
+++ b/src/Bridge/Doctrine/Common/Filter/OrderFilterTrait.php
@@ -52,6 +52,13 @@ trait OrderFilterTrait
                 'property' => $propertyName,
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        strtolower(OrderFilterInterface::DIRECTION_ASC),
+                        strtolower(OrderFilterInterface::DIRECTION_DESC),
+                    ],
+                ],
             ];
         }
 
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -690,6 +690,10 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
                 $type = \in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string'];
                 $v3 ? $parameter['schema'] = $type : $parameter += $type;
 
+                if ($v3 && isset($data['schema'])) {
+                    $parameter['schema'] = $data['schema'];
+                }
+
                 if ('array' === ($type['type'] ?? '')) {
                     $deepObject = \in_array($data['type'], [Type::BUILTIN_TYPE_ARRAY, Type::BUILTIN_TYPE_OBJECT], true);
 
--- a/tests/Bridge/Doctrine/Common/Filter/OrderFilterTestTrait.php
+++ b/tests/Bridge/Doctrine/Common/Filter/OrderFilterTestTrait.php
@@ -28,11 +28,25 @@ trait OrderFilterTestTrait
                 'property' => 'id',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
             'order[name]' => [
                 'property' => 'name',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
         ], $filter->getDescription($this->resourceClass));
     }
--- a/tests/Bridge/Doctrine/MongoDbOdm/Filter/OrderFilterTest.php
+++ b/tests/Bridge/Doctrine/MongoDbOdm/Filter/OrderFilterTest.php
@@ -39,81 +39,193 @@ class OrderFilterTest extends DoctrineMongoDbOdmFilterTestCase
                 'property' => 'id',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
             'order[name]' => [
                 'property' => 'name',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
             'order[alias]' => [
                 'property' => 'alias',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
             'order[description]' => [
                 'property' => 'description',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
             'order[dummy]' => [
                 'property' => 'dummy',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
             'order[dummyDate]' => [
                 'property' => 'dummyDate',
                 'type' => 'string',
                 'required' => false,
+                'schema' => [
+                    'type' => 'string',
+                    'enum' => [
+                        'asc',
+                        'desc',
+                    ],
+                ],
             ],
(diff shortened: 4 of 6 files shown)


fredericbarthelet:
Yep
image
image

src/Bridge/Doctrine/Common/Filter/OrderFilterTrait.php
```diff
@@ -52,6 +52,10 @@ public function getDescription(string $resourceClass): array
                 'property' => $propertyName,
                 'type' => 'string',
                 'required' => false,
+                'enum' => [
```
teohhanhui: Not sure about this... :x

src/Bridge/Doctrine/Common/Filter/OrderFilterTrait.php
```diff
@@ -52,6 +52,10 @@ public function getDescription(string $resourceClass): array
...
+                    strtolower(OrderFilterInterface::DIRECTION_ASC),
```
teohhanhui: Actually the order filter parameter could also be used without a value, e.g. `GET /books?order[datePublished]`

It'd use the default order configured for that property. (I don't like it, but it's something that's supported.)

See Behat test scenario here: https://github.com/api-platform/core/blob/cfa8f34d0d5779e0d7e4805ae0d1062afd699aab/features/doctrine/order_filter.feature#L244

alanpoulain: It can be documented like this but could be used without value too, doesn't it?

teohhanhui: Enum usually means you must use one of the specified values in the list. Haha...

teohhanhui: So, there is `allowEmptyValue` in OpenAPI v3. We could probably add that for other filters too, where applicable. But perhaps in another PR.

alanpoulain: So this PR could be merged as is.

teohhanhui: @alanpoulain See the other comment.

src/Bridge/Doctrine/Common/Filter/OrderFilterTrait.php
alanpoulain: If we do this, I don't think the UI would understand it.

teohhanhui: No, this has nothing to do with the UI. I'm talking about what we return from `getDescription`. It's the job of the `DocumentationNormalizer` to convert that into Swagger / OpenAPI docs, for example.

teohhanhui: And in fact, in Swagger / OpenAPI, we're indeed doing:

```yaml
schema:
    enum:
```

fredericbarthelet: In the swagger v3 we do :
```
schema:
   type: string
   enum: ...
```
so i felt like getDescription should not return nested arrays as it would mean that type is as well misplaced. Or should `type` be part of nested array indexed `schema` returned by getDescription ?

teohhanhui: @fredericbarthelet I think the confusion comes from treating `getDescription` as if it's meant to map to Swagger / OpenAPI directly. It's not.

We are returning some metadata about a filter, and it's the job of the respective `DocumentationNormalizer` to convert that into what's appropriate for their format.

teohhanhui: What we can do in the Swagger / OpenAPI `DocumentationNormalizer` though, is to take the `schema` value from the filter's `getDescription`, if it exists. Otherwise, we generate the `schema` value like the existing code does.

It means, from `OrderFilter`, we have to return:

```php
[
    'property' => $propertyName,
    'type' => 'string',
    'required' => false,
    'schema' => [
        'type' => 'string',
        'enum' => [
            strtolower(OrderFilterInterface::DIRECTION_ASC),
            strtolower(OrderFilterInterface::DIRECTION_DESC),
        ],
    ],
]
```

WDYT?

fredericbarthelet: you don't mind the double `'type' => 'string'` declaration ? Even if `schema` declaration would then take over `type` declaration, you still have to declare both as `type` is required for filter documentation computation.

Or should I modify `DocumentationNormalizer` so that `type` is no more required and have this output from `getDescription` : 

```
    'property' => $propertyName,
    'required' => false,
    'schema' => [
        'type' => 'string',
        'enum' => [
            strtolower(OrderFilterInterface::DIRECTION_ASC),
            strtolower(OrderFilterInterface::DIRECTION_DESC),
        ],
    ],
]
```

teohhanhui: I don't mind the duplication. You should be able to provide the `type` without providing `schema`. Also, we cannot change this because of BC.

fredericbarthelet: OK, i'll update my PR accordingly. Will let you know when it's done :)

fredericbarthelet: @teohhanhui changes made. Let me know if I got it right this time ;)

dunglas:
@teohhanhui can we merge this one?

teohhanhui:
Test is missing for the Swagger `DocumentationNormalizer`.

fredericbarthelet:
@dunglas @teohhanhui test added in `DocumentationNormalizerV3Test`

teohhanhui:
Thanks @fredericbarthelet! :tada:

src/Bridge/Doctrine/Common/Filter/OrderFilterTrait.php
```diff
@@ -52,6 +52,13 @@ public function getDescription(string $resourceClass): array
                 'property' => $propertyName,
                 'type' => 'string',
```
dunglas: Actually, I think that we should deprecate `type` and `required`, in favor of `schema` (that is a JSON Schema). Using a JSON Schema to describe the filter is smart, and we should only support that. WDYT @teohhanhui?

teohhanhui: I think that'll just make a lot of code more complicated...

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -690,6 +690,10 @@ private function getFiltersParameters(bool $v3, string $resourceClass, string $o
                 $type = \in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string'];
                 $v3 ? $parameter['schema'] = $type : $parameter += $type;
 
+                if ($v3 && isset($data['schema'])) {
```
dunglas: Shouldn't we do this test before calling `$this->jsonSchemaTypeFactory->getType`, in order to not call it when it's useless (when `data['schema']` exists)? Also, are the provided schema incompatible with Swagger v2?

teohhanhui: We already do useless work like this in many other places in this class. (At least it was the case before. Not sure after your refactoring.)

I guess it's okay in this case as it keeps the code as simple as possible.

> Also, are the provided schema incompatible with Swagger v2?

There was no support for schema in parameters.

dunglas: > We already do useless work like this in many other places in this class. (At least it was the case before. Not sure after your refactoring.)

I tried to improve it as much as possible :) 

> I guess it's okay in this case as it keeps the code as simple as possible.

It doesn't increase the complexity that much:

```php
                if ($v3 && isset($data['schema'])) {
                    $parameter['schema'] = $data['schema'];
                } else {
                 $type = \in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string'];
                $v3 ? $parameter['schema'] = $type : $parameter += $type;
              }
```

(it could probably be more simplified btw)

>It has no support for schema in parameters.

It is supported: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7 (`If in is "body":`).

dunglas: And because it is supported, the code can be simplified a lot:

```php
$type = $data['schema'] ?? \in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string'];
$v3 ? $parameter['schema'] = $type : $parameter += $type;
```

teohhanhui: No, it's not supported. That's for the request body. We're talking about filters here.

PR #2972, opened 2019-08-07
Fix compatibility with symfony/error-renderer

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Opens the door to error-renderer while keeping compat with debug, on 2.4. 
Cannot be tested since error-renderer conflicts with http-kernel <4.4 and http-kernel ^3.4|^4.0 is supported here, so `--prefer-lowest` ends up being non-installable.
Support for symfony/debug should probably be deprecated in a next minor, as soon as symfony/* is bumped to `^4.4`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,6 +163,11 @@ commands:
       - run:
           name: Update project dependencies
           command: composer update --prefer-dist --no-progress --no-suggest --ansi
+  update-project-dependencies-high:
+    steps:
+      - run:
+          name: Update project dependencies (high deps)
+          command: composer config minimum-stability dev && composer require symfony/error-renderer:^4.4 --no-update && composer update --prefer-dist --no-progress --no-suggest --ansi
   wait-for-elasticsearch:
     steps:
       - wait-for-service:
@@ -255,7 +260,7 @@ jobs:
       - disable-php-memory-limit
       - update-composer
       - restore-composer-cache
-      - update-project-dependencies
+      - update-project-dependencies-high
       - save-composer-cache
       - clear-test-app-cache
       - restore-phpstan-cache
--- a/src/Action/ExceptionAction.php
+++ b/src/Action/ExceptionAction.php
@@ -14,7 +14,8 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Action;
 
 use ApiPlatform\Core\Util\ErrorFormatGuesser;
-use Symfony\Component\Debug\Exception\FlattenException;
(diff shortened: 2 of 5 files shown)


alanpoulain:
PHPStan error seems related.

teohhanhui:
This should target `master`, no? Symfony 4.4 is not even in beta yet... So I don't see why it should be supported in 2.4 :laughing:

chalasr:
Well, symfony/debug is deprecated as of 4.4 and api-platform/core supports `^3.4|^4.0`.
When symfony 4.4 will be released, people will start getting deprecations about the debug being deprecated. At this stage, they'll probably follow the upgrade path which is to replace their `symfony/debug` dependency by `symfony/error-renderer`. 
Once done, if they use api-platform/core 2.4, they will get their api error handling broken because of the lack of support for symfony/error-renderer there, and have some hard debugging time.

Unless api-platform/core v2.4 forbids symfony/* >=4.4, I think supporting symfony/error-renderer makes sense. Am I missing something?

Usually, we see this kind of PRs opened in hurry after the first beta of a new symfony minor version comes out (see similar compatibility fixes on doctrine bundles for instance). 
Just anticipating here :)

And of course, we need to deprecate support for symfony/debug on master.

dunglas:
The thing is that we plan to release API Platform 3.0 at the same time than Symfony 5. And v3 will probably support only Symfony 5 as the goal is to simplify the code base as much as possible.

Maybe it’s time to open a 3.0 branch?

teohhanhui:
My argument is that by the time the first beta of Symfony 4.4 comes out, I hope we've already released 2.5 😂 That's why it doesn't belong in 2.4, especially because Symfony 4.4 might still be in flux.

teohhanhui:
> Maybe it’s time to open a 3.0 branch?

Is the plan to release 2.5 and 3.0 at the same time? Or we don't mind that it'll be 2.6? Because 2.5 has been a long time in the making... I think many people are waiting for a release.

chalasr:
Got it, 2.5 is coming and you want 2.4 to be gone asap :) I will rebase this one on master, thanks for the hint.
Creating the 3.0 branch seems fine to me, not sure about releasing 2.5 and 3.0 at the same time.

dunglas:
I’ll release 2.5 as soon as possible (I’ll be back from vacations on Monday). I’m not sure if it will be a 2.6 release or not.

PR #2974, opened 2019-08-08
Making assertArraySubset() compatible with phpunit v6

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | not needed

Hi!

I'm playing around with the unreleased new test files. I'm using them with `simple-phpunit`, which on my system installed phpunit 6.5 (which happens via the phpunit.xml.dist file via the recipe https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/4.1/phpunit.xml.dist#L15).

Anyways, using version 6.5 causes an error because the `assertArraySubset()` method DOES exist in that version of phpunit, but the signatures don't match. Removing the type-hints, which not ideal, fixes the problem with (I think) minimal pain.

Thanks!

### DIFF
--- a/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
+++ b/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
@@ -88,7 +88,7 @@ trait ApiTestAssertionsTrait
      * @throws \SebastianBergmann\RecursionContext\InvalidArgumentException
      * @throws \Exception
      */
-    public static function assertArraySubset($subset, $array, bool $checkForObjectIdentity = false, string $message = ''): void
+    public static function assertArraySubset($subset, $array, $checkForObjectIdentity = false, $message = '')
     {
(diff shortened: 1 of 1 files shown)


dunglas:
Hi Ryan,

And thanks for the contrib. We only plan to support the latest stable version of PHPUnit. It’s a pain to maintain compatibility with older versions...

weaverryan:
It is! Fine with me. The problem we're going to have is that many Symfony devs will have the 6.5. version (due to the recipe), and the error (unfortunately) is really poor. Easiest (imperfect) solution would be to clearly state in the docs for the new testing tools that 7 or 8 is required and that you need to update your `phpunit.xml.dist` file for this.

teohhanhui:
@ryanweaver The recipe must be updated as PHPUnit 6 is already EOL: https://phpunit.de/supported-versions.html

PR #2959, opened 2019-07-28
[GraphQL] Resolver Stages

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1904
| License       | MIT
| Doc PR        |

This PR introduces **Stages** for the Resolvers.

They are the counterpart of the EventListeners in REST.

Each Stage can be decorated in order to have more control in the Resolver flow.

The Stages are:
- ReadStage
- DeserializeStage
- ValidateStage
- DenyAccessStage
- WriteStage
- SerializeStage

This PR also introduce a SerializerContextBuilder in order to be able to control the serialization in the Resolvers.

PR #2978, opened 2019-08-09
[Proposal] Add Stages

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

As I've done with GraphQL (https://github.com/api-platform/core/pull/2959), I would like to do the same for the REST part.

The aim is to create **Stage** services.

They will be responsible for all the stages (or steps) API Platform is doing to resolve a request.

They will be independent of the request and will be used by the event listeners.

A lot of benefits:
- instead of using events and having issues with priorities (https://api-platform.com/docs/core/events/), the services could be decorated instead;
- since they will be independent of events and the request, they could be used in different contexts (I'm thinking of Laravel for instance).

Right now I've only done the read stage. Behat tests are passing.

WDYT @api-platform/core-team?

### DIFF
--- a/src/EventListener/ReadListener.php
+++ b/src/EventListener/ReadListener.php
@@ -15,14 +15,13 @@ namespace ApiPlatform\Core\EventListener;
 
 use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
-use ApiPlatform\Core\DataProvider\OperationDataProviderTrait;
 use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
-use ApiPlatform\Core\Exception\InvalidIdentifierException;
-use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Exception\NotFoundException;
 use ApiPlatform\Core\Identifier\IdentifierConverterInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
-use ApiPlatform\Core\Metadata\Resource\ToggleableOperationAttributeTrait;
 use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use ApiPlatform\Core\Stage\ReadStage;
+use ApiPlatform\Core\Stage\ReadStageInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use ApiPlatform\Core\Util\RequestParser;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
@@ -35,21 +34,29 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
  */
 final class ReadListener
 {
-    use OperationDataProviderTrait;
-    use ToggleableOperationAttributeTrait;
-
     public const OPERATION_ATTRIBUTE_KEY = 'read';
 
+    private $readStage;
     private $serializerContextBuilder;
 
-    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
+    public function __construct(/*CollectionDataProviderInterface */$readStage, /*ItemDataProviderInterface */$serializerContextBuilder/*, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory = null*/)
     {
-        $this->collectionDataProvider = $collectionDataProvider;
-        $this->itemDataProvider = $itemDataProvider;
-        $this->subresourceDataProvider = $subresourceDataProvider;
+        $this->readStage = $readStage;
         $this->serializerContextBuilder = $serializerContextBuilder;
-        $this->identifierConverter = $identifierConverter;
-        $this->resourceMetadataFactory = $resourceMetadataFactory;
+
+        if (\func_num_args() > 2) {
+            @trigger_error(sprintf('Not injecting only "%s" and "%s" is deprecated since API Platform 2.5 and will not be possible anymore in API Platform 3', ReadStageInterface::class, SerializerContextBuilderInterface::class), E_USER_DEPRECATED);
+
+            $collectionDataProvider = $readStage;
+            $itemDataProvider = $serializerContextBuilder;
+            $subresourceDataProvider = func_get_arg(2);
+            $serializerContextBuilder = func_get_arg(3);
+            $identifierConverter = func_get_arg(4);
+            $resourceMetadataFactory = func_get_arg(5);
+
+            $this->readStage = new ReadStage($collectionDataProvider, $itemDataProvider, $subresourceDataProvider, $identifierConverter, $resourceMetadataFactory);
+            $this->serializerContextBuilder = $serializerContextBuilder;
+        }
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Stage/ReadStage.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->itemDataProvider = $itemDataProvider;
+        $this->subresourceDataProvider = $subresourceDataProvider;
+        $this->identifierConverter = $identifierConverter;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(array $attributes, array $parameters, ?array $filters, string $method, array $normalizationContext)
```
dunglas: Instead of this long list of unstructured and untyped parameters, shouldn't we introduce a new `ApiRequest` or `ApiQuery`/`ApiCommand` class(es), that we'll construct from the (Symfony/Laravel/PSR-7) HTTP request, or from the GraphQL query, or from any other source of request regarless of the protocol. It will allow to have a more strict typing, and will improve the evolvability of API Platform (in case we need to add more data over the time).

Basically, the responsibility of the Symfony Event Listener, or of the Laravel middleware, or of the GraphQL resolver would be to create this `ApiRequest` instance.

src/Stage/ReadStage.php
```diff
@@ -0,0 +1,95 @@
...
+            $context[IdentifierConverterInterface::HAS_IDENTIFIER_CONVERTER] = true;
+        }
+
+        try {
+            $identifiers = $this->extractIdentifiers($parameters, $attributes);
+
+            if (isset($attributes['item_operation_name'])) {
+                $data = $this->getItemData($identifiers, $attributes, $context);
+            } elseif (isset($attributes['subresource_operation_name'])) {
+                // Legacy
```
dunglas: Why legacy? It's a valid use case to not inject a `SubResourceDataPprovider` (especially if you use the library directly, without the Symfony bridge).

src/Stage/ReadStageInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Stage;
+
+/**
+ * Retrieves data from the applicable data provider and sets it as a request parameter called data.
```
dunglas: Outdated

src/EventListener/ReadListener.php
```diff
@@ -60,61 +57,33 @@ public function __construct(CollectionDataProviderInterface $collectionDataProvi
...
+        }
+
+        if ($normalizationContext) {
+            // Builtin data providers are able to use the serialization context to automatically add join clauses
+            $request->attributes->set('_api_normalization_context', $normalizationContext);
         }
 
         $request->attributes->set('data', $data);
-        $request->attributes->set('previous_data', \is_object($data) ? clone $data : $data);
+        $request->attributes->set('previous_data', \is_object($data) && (new \ReflectionClass(\get_class($data)))->isCloneable() ? clone $data : $data);
```
dunglas: This should be backported on 2.4 as a bugfix.

src/EventListener/ReadListener.php
```diff
@@ -35,21 +34,19 @@
...
-    use OperationDataProviderTrait;
-    use ToggleableOperationAttributeTrait;
-
     public const OPERATION_ATTRIBUTE_KEY = 'read';
 
     private $serializerContextBuilder;
+    private $readStage;
 
-    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
+    public function __construct(CollectionDataProviderInterface $collectionDataProvider, ItemDataProviderInterface $itemDataProvider, SubresourceDataProviderInterface $subresourceDataProvider = null, SerializerContextBuilderInterface $serializerContextBuilder = null, IdentifierConverterInterface $identifierConverter = null, ResourceMetadataFactoryInterface $resourceMetadataFactory = null, ReadStageInterface $readStage = null)
```
dunglas: Shouldn't the listener depend only of this new service? And this new service be the first parameter, and passing all other services been deprecated?

src/Stage/ReadStageInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+interface ReadStageInterface
+{
+    public const OPERATION_ATTRIBUTE_KEY = 'read';
+
+    /**
+     * @return object|iterable|null
+     */
+    public function apply(array $attributes, array $parameters, ?array $filters, string $method, array $normalizationContext);
```
dunglas: `apply` isn't very descriptive (I'm not a fond of "stage" too btw :D). Can't we more explicit names?

src/Stage/ReadStage.php
```diff
@@ -0,0 +1,95 @@
...
+use ApiPlatform\Core\Identifier\IdentifierConverterInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ToggleableOperationAttributeTrait;
+
+/**
+ * Retrieves data from the applicable data provider and sets it as a request parameter called data.
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class ReadStage implements ReadStageInterface
```
dunglas: Why isn't this class also used by the GraphQL resolver?

src/Stage/ReadStageInterface.php
```diff
@@ -0,0 +1,29 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Stage;
+
+/**
+ * Retrieves data from the applicable data provider and sets it as a request parameter called data.
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+interface ReadStageInterface
```
dunglas: I wonder if this new concept is really necessary. Does this logic belong to the DataProvider? Or should we create a `ApiRequestFactory` class instead (see my previous comment?). I'm not sure yet, I need to think about it, but I've the feeling that we're introducing unnecessary indirection layers.

src/Stage/ReadStage.php
```diff
@@ -0,0 +1,95 @@
...
+use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use ApiPlatform\Core\Exception\NotFoundException;
+use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Identifier\IdentifierConverterInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ToggleableOperationAttributeTrait;
+
+/**
+ * Retrieves data from the applicable data provider and sets it as a request parameter called data.
```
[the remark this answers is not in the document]
alanpoulain: You think? It's basically what this stage does (excepted the set part which I can remove).

src/Stage/ReadStage.php
alanpoulain: I've always wondered if it would be interesting to introduce this kind of class instead of having a lot of parameters and a `context` array with more things inside :sweat_smile:
I'm :+1: for it.

src/Stage/ReadStage.php
alanpoulain: Not my code here but I can remove the comment. The `SubresourceData` is mandatory in the GraphQL read stage.

src/Stage/ReadStageInterface.php
alanpoulain: See above.

src/EventListener/ReadListener.php
alanpoulain: This is not really a bugfix. It's because now this method is *always* called whereas before it was not called if the operation was a collection.

src/EventListener/ReadListener.php
alanpoulain: You're right.

src/Stage/ReadStageInterface.php
alanpoulain: Which one do you have in mind? Something like `read`?
I know you would say that :smile: Naming is really hard and subjective. "Stage" was interesting because it was not used and its meaning was alright: it's a "step in a process".

src/Stage/ReadStage.php
alanpoulain: I would like to merge both too but sadly they are very different. I don't think it will be doable easily.

src/Stage/ReadStageInterface.php
alanpoulain: I think their responsibility is clear:
- allow or not to apply the stage
- extract, collect and normalize the right parameters (mainly extract the identifiers and build the (de)normalization context)
- call the good "solver"
We *could* add the second point inside an `ApiRequestFactory`. But IMO it will not necessarily be a good thing. It will have a lot of conditional clauses and could quickly become a monster.
And I don't know where we could put the other parts. Maybe having a `ChainDataProvider` and modify the `ChainDataPersister` to choose the right method (`persist` or `remove` to call based on the `ApiRequest`)?

alanpoulain: Also IMHO, it's easier for the user to understand and decorate this kind of service than to use the data provider / persister and the serializer.

src/EventListener/ReadListener.php
dunglas: An issue has been reported for non-cloneable objects IIRC

src/Stage/ReadStageInterface.php
dunglas: I’m really not sure about the last point! It’s more concepts to learn with no clear responsibilities.

src/Stage/ReadStageInterface.php
dunglas: Stage sounds like “util” or “service” to me: too generic and not very descriptive

src/Stage/ReadStage.php
dunglas: But then it doesn’t fix the issue of having good extensions points working both for REST and GraphQL. Which is one of the goals.

alanpoulain: Why not having different ones? The need to use both REST and GraphQL is not common. It would just mean having two different service names to use.

src/EventListener/ReadListener.php
alanpoulain: In this case, without this change, the tests wouldn't pass.

src/Stage/ReadStageInterface.php
alanpoulain: I guess it depends of the developer and their way to think.
But a lot of developers are using the events already to add more logic so the need is there IMO.

src/Stage/ReadStageInterface.php
alanpoulain: But it is something kind of generic. It's a step or an operation. How would you call the `read`, `serialize`, `validate`, etc. things?

bendavies: Make it a callable? `__invoke`?

alanpoulain: Good idea! WDYT @dunglas?

dunglas:
I need to find some time to dive deeper into this. I'm still not convinced that we need this new layer of complexity. I also don't like the fact that this new class isn't used by the GraphQL subsystem.

src/Stage/ReadStage.php
Goutte: To me it's very common, implementing both.  That's actually what drove me to API Platform in the first place, I thought it would provide a(nother) way to DRY my Controllers and Resolvers.  If you feel it's too much work for this MR, that's of course your prerogative, but please don't sacrifice forever the goal of unification.  Also, `Stage` feels okay ; it fits, and I can't find anything I deem more suitable.

alanpoulain: We could introduce stage for both REST and GraphQL, but the service will be an empty shell with practically no code IMO. It will be just there to be decorated.

PR #2990, opened 2019-08-19
Prevent cloning generator when instanciating previous_data

Nitneuk2:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |  https://github.com/api-platform/docs/pull/858

Add a clone trait with a method `clone` that returns a clone of the given variable if cloneable.
Inspired by @alanpoulain from #2978

### DIFF
--- a/src/Util/CloneTrait.php
+++ b/src/Util/CloneTrait.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Util;
+
+/**
+ * Clones given data if cloneable.
+ *
+ * @internal
+ *
+ * @author Quentin Barloy <quentin.barloy@gmail.com>
+ */
+trait CloneTrait
+{
+    public function clone($data)
+    {
+        if (!\is_object($data)) {
+            return $data;
+        }
+
+        try {
+            return (new \ReflectionClass($data))->isCloneable() ? clone $data : null;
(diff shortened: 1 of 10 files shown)


src/Util/CloneTrait.php
```diff
@@ -0,0 +1,33 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Util;
+
+/**
+ * Clone given data if cloneable.
```
dunglas:
```suggestion
 * Clones given data if cloneable.
```

src/Util/CloneTrait.php
```diff
@@ -0,0 +1,33 @@
...
+ * @internal
+ *
+ * @author Quentin Barloy <quentin.barloy@gmail.com>
+ */
+trait CloneTrait
+{
+    public function clone($data): bool
+    {
+        try {
+            return \is_object($data) && (new \ReflectionClass(\get_class($data)))->isCloneable() ? clone $data : $data;
```
dunglas: We should return `null` instead of the updated data if not cloneable (same in the exception).

src/Util/CloneTrait.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Clone given data if cloneable.
+ *
+ * @internal
+ *
+ * @author Quentin Barloy <quentin.barloy@gmail.com>
+ */
+trait CloneTrait
+{
+    public function clone($data): bool
```
teohhanhui: Wrong return type.

dunglas:
Thanks @Nitneuk2

alanpoulain:
Shouldn't the `ItemResolver` use the trait too?

dunglas:
good catch

PR #2992, opened 2019-08-19
Deprecate access_control attribute, add security + security_post_denormalize attributes

### LINKED ISSUE #1646: Deserializer before security

vincentchalamon:
| Q             | A
| ------------- | ---
| Waiting for PR?      | api-platform/core#2990
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1646, api-platform/api-platform#1218
| License       | MIT
| Doc PR        | api-platform/docs#859

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -25,6 +25,10 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
  * @Attributes(
  *     @Attribute("accessControl", type="string"),
  *     @Attribute("accessControlMessage", type="string"),
+ *     @Attribute("security", type="string"),
+ *     @Attribute("securityMessage", type="string"),
+ *     @Attribute("securityPostDenormalize", type="string"),
+ *     @Attribute("securityPostDenormalizeMessage", type="string"),
  *     @Attribute("attributes", type="array"),
  *     @Attribute("cacheHeaders", type="array"),
  *     @Attribute("collectionOperations", type="array"),
@@ -108,14 +112,28 @@ final class ApiResource
      *
      * @var string
      */
-    private $accessControl;
+    private $security;
 
     /**
      * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
      *
      * @var string
      */
-    private $accessControlMessage;
+    private $securityMessage;
+
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var string
+     */
+    private $securityPostDenormalize;
+
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var string
+     */
+    private $securityPostDenormalizeMessage;
 
     /**
      * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
--- a/src/Security/EventListener/DenyAccessListener.php
+++ b/src/Security/EventListener/DenyAccessListener.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\Security\ExpressionLanguage;
 use ApiPlatform\Core\Security\ResourceAccessChecker;
 use ApiPlatform\Core\Security\ResourceAccessCheckerInterface;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\Security\Core\Authentication\AuthenticationTrustResolverInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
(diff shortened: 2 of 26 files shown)


src/EventListener/ReadListener.php
```diff
@@ -115,6 +115,6 @@ public function onKernelRequest(GetResponseEvent $event): void
         }
 
         $request->attributes->set('data', $data);
-        $request->attributes->set('previous_data', \is_object($data) ? clone $data : $data);
+        $request->attributes->set('previous_data', \is_object($data) && (new \ReflectionClass(\get_class($data)))->isCloneable() ? clone $data : $data);
```
dunglas: This will have to be updated when #2990 will be merged

src/GraphQl/Resolver/ResourceAccessCheckerTrait.php
```diff
@@ -30,17 +30,24 @@ trait ResourceAccessCheckerTrait
...
+    public function canAccess(?ResourceAccessCheckerInterface $resourceAccessChecker, ResourceMetadata $resourceMetadata, string $resourceClass, ResolveInfo $info, $extraVariables = [], string $operationName = null, string $attribute = 'security'): void
     {
         if (null === $resourceAccessChecker) {
             return;
         }
 
-        $isGranted = $resourceMetadata->getGraphqlAttribute($operationName ?? '', 'access_control', null, true);
+        $isGranted = $resourceMetadata->getGraphqlAttribute($operationName ?? '', $attribute, null, true);
+        if (null === $isGranted) {
+            $isGranted = $resourceMetadata->getGraphqlAttribute($operationName ?? '', 'access_control', null, true);
```
dunglas: As this class is experimental, you could also just make the `$attribute` mandatory, and trigger the deprecation only if its value is `access_control`

src/Security/EventListener/DenyAccessListener.php
```diff
@@ -56,27 +57,53 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
...
             return;
         }
 
         $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
 
-        $isGranted = $resourceMetadata->getOperationAttribute($attributes, 'access_control', null, true);
+        $isGranted = $resourceMetadata->getOperationAttribute($attributes, $attribute, null, true);
+        if (null === $isGranted) {
+            // Backward compatibility
+            $isGranted = $resourceMetadata->getOperationAttribute($attributes, 'access_control', null, true);
```
dunglas: It's possible to pass `access_control` in `onKernelRequest`, then you don't need this check anymore.

vincentchalamon: Can't, otherwise I couldn't trigger the deprecation

teohhanhui:
If we're really doing this, I'd suggest:

```php
"access_control"={
    "precondition"="...",
    "postcondition"="...",
},
```

dunglas:
Sub-attributes are too verbose, I prefer this approach (see my comment in the related issue).

teohhanhui:
"security" and "late_security" are really confusing names.

If you don't want nested attributes, then perhaps:

* access_control_precondition
* access_control_postcondition

src/Annotation/ApiResource.php
```diff
@@ -117,6 +121,34 @@ final class ApiResource
      */
     private $accessControlMessage;
```
dunglas: `$accessControl` and `$accessControlMessage` should be deleted, so the autocompletion will not be triggered anymore in PHPStorm.

vincentchalamon: Shouldn't I tag them as deprecated instead?

dunglas:
Awesome! Thanks @vincentchalamon

c-lambert:
Hello,
i'm not sure to understand the fix applied.
I use API Platform 2.6.4 with security annotation on itemOperation. But when i call this endpoint with unauthorized user, API Platform responds 404 when entity doesn't existe, and 403 when entity exist.

`security` attribute is not `early` by default ?

norkunas:
`security` is executed only after reading data, so you could have dynamic security by object data, if you want to not expose that item exist you could use security acl for that

c-lambert:
> `security` is executed only after reading data, so you could have dynamic security by object data, if you want to not expose that item exist you could use security acl for that

Ok thank you, i'll do that.
I was hopping that we can bypass this behavior with a parameter.

Issue #2982, opened 2019-08-12
Nest testing tools: Undefined method prepareRequest()


weaverryan:
Hi friends!

The new testing [Client](https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Test/Client.php) class implements `Symfony\Contracts\HttpClient\HttpClientInterface`, which is cool! But, it also implicitly relies on a concrete method on `HttpClient` itself - the `prepareRequest()` method: https://github.com/api-platform/core/blob/92f655c84e9d03f38ee2986906fba11ce529bc2f/src/Bridge/Symfony/Bundle/Test/Client.php#L79

If you don't have http-client installed, you'll get a cryptic error about calling an undefined function.

**Edit** Oddly, I get an "undefined method" on the first run... and then a "Trait 'Symfony\Component\HttpClient\HttpClientTrait' not found" on the second... which makes more sense...

That function specifically lives in [HttpClientTrait](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/HttpClient/HttpClientTrait.php)

In other words, as it's written now, `symfony/http-client` is a hard dependency on the new `Client` class. We could make this more user-friendly by checking for the presence of that trait before using it. That's not possible inside of `Client` itself (you can't conditionally `use` a trait), so... we could do it inside of `ApiTestCase::createClient()` instead... which is how a user would "usually" create the `Client`.

... or we could require `symfony/http-client` as a dependency of api-platform/core... though afaik, it's only needed in this one spot currently...

Thoughts?

teohhanhui:
But why would the `test.api_platform.client` service exist, if the instantiated class `ApiPlatform\Core\Bridge\Symfony\Bundle\Test\Client` has incomplete definition? It sounds like something that should be fixed in Symfony DI?

dunglas:
I agree that we should throw an error if the deps is missing!

dunglas:
@teohhanhui it's not a bug in SF: we always create this service in test mode: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L131-L133

We could indeed not create it if the HttpClient component is not installed, but I think that @weaverryan's proposal is more user friendly (and it has no performance drawback as this service is only registered in test mode).

teohhanhui:
What I mean is, if we try to instantiate our `Client` class, it should already fail, as @weaverryan has pointed out:

```
[...]

PR #2994, opened 2019-08-19
Fix : Throw a better message if http-client is not installed

Zowac:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | Yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2982 
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -46,6 +46,7 @@ use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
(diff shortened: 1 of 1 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -129,6 +130,10 @@ public function load(array $configs, ContainerBuilder $container): void
             ->addTag('api_platform.filter');
 
         if ($container->hasParameter('test.client.parameters') && class_exists(AbstractBrowser::class)) {
+            if (!class_exists(HttpClient::class)){
+                throw new RuntimeException('Run "composer require symfony/http-client" to use the test client"');
```
dunglas:
```suggestion
                throw new RuntimeException('Run "composer require --dev symfony/http-client" to use the test client"');
```

Can you also rebase and run PHP CS Fixer please?

Zowac: Done :)

dunglas:
Thanks @Zowac!

dunglas:
Actually, I've probably be too fast to merge. It would be better to throw here: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php#L44

Currently the exception will be triggered for every Symfony application not having the HttpClient installed, even if it doesn't use our test client.

PR #2996, opened 2019-08-19
add command to generate json schema

jockos:
Add a command to generate a JSON Schema for a resource.
Can be used for the whole resource or by operation

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

### DIFF
--- a/src/JsonSchema/Command/JsonSchemaGenerateCommand.php
+++ b/src/JsonSchema/Command/JsonSchemaGenerateCommand.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\JsonSchema\Command;
+
+use ApiPlatform\Core\Api\OperationType;
+use ApiPlatform\Core\JsonSchema\SchemaFactoryInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Exception\InvalidOptionException;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Generates a resource JSON Schema.
+ *
+ * @author Jacques Lefebvre <jacques@les-tilleuls.coop>
+ */
+final class JsonSchemaGenerateCommand extends Command
+{
+    private $schemaFactory;
+
+    private $formats;
+
+    public function __construct(SchemaFactoryInterface $schemaFactory, array $formats)
+    {
+        $this->schemaFactory = $schemaFactory;
+        $this->formats = $formats;
+
+        parent::__construct();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('api:json-schema:generate')
+            ->setDescription('Generates the JSON Schema for a resource operation.')
+            ->addArgument('resource', InputArgument::REQUIRED, 'The Fully Qualified Class Name (FQCN) of the resource')
+            ->addOption('itemOperation', null, InputOption::VALUE_OPTIONAL, 'The item operation')
+            ->addOption('collectionOperation', null, InputOption::VALUE_OPTIONAL, 'The collection operation')
+            ->addOption('format', null, InputOption::VALUE_OPTIONAL, 'The response format', array_key_first($this->formats))
+            ->addOption('output', null, InputOption::VALUE_NONE, 'Use this option to get the output version')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $resource = $input->getArgument('resource');
+        $itemOperation = $input->getOption('itemOperation');
+        $collectionOperation = $input->getOption('collectionOperation');
+        $format = $input->getOption('format');
+        $outputType = $input->getOption('output');
+
+        if (!isset($this->formats[$format])) {
+            throw new InvalidOptionException(\sprintf('The response format "%s" is not supported. Supported formats are : %s.', $format, implode(', ', array_keys($this->formats))));
+        }
+
+        $operationType = null;
+        $operationName = null;
+
+        if ($itemOperation && $collectionOperation) {
+            throw new InvalidOptionException('You can only use one of "--itemOperation" and "--collectionOperation" options at the same time.');
+        }
+
+        if (null !== $itemOperation || null !== $collectionOperation) {
+            $operationType = $itemOperation ? OperationType::ITEM : OperationType::COLLECTION;
+            $operationName = $itemOperation ?? $collectionOperation;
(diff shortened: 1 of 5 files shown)


src/JsonSchema/Command/JsonSchemaGenerateCommand.php
```diff
@@ -0,0 +1,90 @@
...
+        if (null !== $itemOperation || null !== $collectionOperation) {
+            $operationType = $itemOperation ? OperationType::ITEM : OperationType::COLLECTION;
+            $operationName = $itemOperation ?? $collectionOperation;
+        }
+
+        $schema = $this->schemaFactory->buildSchema($resource, $format,  $outputType , $operationType, $operationName);
+
+        if (!$schema->isDefined()) {
+            $io->error(\sprintf('There is no %s defined for the operation "%s" of the resource "%s".',  $outputType ? 'outputs': 'inputs', $operationName, $resource));
+            return;
```
dunglas: `return 1` so the command will have a return code different than 0, and will be detected as failed in shell scripts.

src/JsonSchema/SchemaFactory.php
```diff
@@ -20,6 +20,7 @@
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer;
+use Symfony\Component\HttpFoundation\Request;
```
dunglas: We don't want this part of the code to depend of HttpFoundation. Use the name of the methods as strings directly.

tests/JsonSchema/Command/JsonSchemaGenerateCommandTest.php
```diff
@@ -0,0 +1,70 @@
...
+    public function testExecuteWithoutOption()
+    {
+        $this->tester->run(['command' => 'api:json-schema:generate', 'resource' => Dummy::class]);
+
+        $this->assertJson($this->tester->getDisplay());
+    }
+
+    public function testExecuteWithItemOperationGet()
+    {
+        $this->tester->run(['command' => 'api:json-schema:generate', 'resource' => Dummy::class, '--itemOperation' => 'get']);
```
dunglas: If case we're running the MongoDB build, you must test with the Dummy document instead of the Dummy entity.

ping @alanpoulain, we need to do something about this, it annoys almost all contributors, and is very difficult to debug.

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
```diff
@@ -0,0 +1,98 @@
...
+    protected function configure()
+    {
+        $this
+            ->setName('api:json-schema:generate')
+            ->setDescription('Generates the JSON Schema for a resource operation.')
+            ->addArgument('resource', InputArgument::REQUIRED, 'The Fully Qualified Class Name (FQCN) of the resource')
+            ->addOption('itemOperation', null, InputOption::VALUE_OPTIONAL, 'The item operation')
+            ->addOption('collectionOperation', null, InputOption::VALUE_OPTIONAL, 'The collection operation')
+            ->addOption('format', null, InputOption::VALUE_OPTIONAL, 'The response format', array_key_first($this->formats))
+            ->addOption('output', null, InputOption::VALUE_NONE, 'Use this option to get the output version');
```
antograssiot: in every other command provided, `output` is used to specify a file path to write to a file.
Can we rename this one ?

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
```diff
@@ -0,0 +1,98 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Generates a resource JSON Schema.
+ *
+ * @author Jacques Lefebvre <jacques@les-tilleuls.coop>
+ */
+final class JsonSchemaGenerateCommand extends Command
```
antograssiot: Shouldn't this be located in in `src/Bridge/Symfony/Bundle/Command` like the other commands ?

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
dunglas: Do you have a suggestion? `request`/`response`?

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
dunglas: We discussed this with @jockos IRL, and actually I think that it's the other commands that should be outside the bridge. They don't depend of frameworkbundle, only of the plain old PHP library, so they shouldn't be in the bridge.

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
teohhanhui: Just use `--type=input|output`?

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('api:json-schema:generate')
+            ->setDescription('Generates the JSON Schema for a resource operation.')
+            ->addArgument('resource', InputArgument::REQUIRED, 'The Fully Qualified Class Name (FQCN) of the resource')
+            ->addOption('itemOperation', null, InputOption::VALUE_OPTIONAL, 'The item operation')
```
teohhanhui: All of the `InputOption::VALUE_OPTIONAL` should actually be `InputOption::VALUE_REQUIRED`.

From the documentation:

> InputOption::VALUE_REQUIRED
>
>    This value is required (e.g. --iterations=5 or -i5), the option itself is still optional

teohhanhui: Another thing, don't we usually use dashes in option names, not camelCase?

src/JsonSchema/Command/JsonSchemaGenerateCommand.php
```diff
@@ -0,0 +1,115 @@
...
+    protected function configure()
+    {
+        $this
+            ->setName('api:json-schema:generate')
+            ->setDescription('Generates the JSON Schema for a resource operation.')
+            ->addArgument('resource', InputArgument::REQUIRED, 'The Fully Qualified Class Name (FQCN) of the resource')
+            ->addOption('itemOperation', null, InputOption::VALUE_REQUIRED, 'The item operation')
+            ->addOption('collectionOperation', null, InputOption::VALUE_REQUIRED, 'The collection operation')
+            ->addOption('format', null, InputOption::VALUE_REQUIRED, 'The response format', (string) $this->formats[0])
+            ->addOption('type', null, InputOption::VALUE_REQUIRED, 'Use this option to set the type of output', 'input');
```
teohhanhui: Perhaps:

```
The type of schema to generate (input or output)
```

dunglas:
Thanks @jockos!

PR #3003, opened 2019-08-20
Add assertions for testing response against JSON Schema from API resource

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "friendsofsymfony/user-bundle": "^2.2@dev",
         "guzzlehttp/guzzle": "^6.0",
         "jangregor/phpstan-prophecy": "^0.4.2",
-        "justinrainbow/json-schema": "^5.0",
+        "justinrainbow/json-schema": "^5.2.1",
         "nelmio/api-doc-bundle": "^2.13.4",
         "phpdocumentor/reflection-docblock": "^3.0 || ^4.0",
         "phpdocumentor/type-resolver": "^0.3 || ^0.4",
--- a/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
+++ b/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
@@ -13,9 +13,13 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\Test;
 
+use ApiPlatform\Core\Api\OperationType;
 use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\Constraint\ArraySubset;
 use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\Constraint\MatchesJsonSchema;
+use ApiPlatform\Core\JsonSchema\Schema;
+use ApiPlatform\Core\JsonSchema\SchemaFactoryInterface;
 use PHPUnit\Framework\ExpectationFailedException;
+use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
@@ -28,9 +32,9 @@ trait ApiTestAssertionsTrait
     use BrowserKitAssertionsTrait;
 
     /**
-     * Asserts that the retrieved JSON contains has the specified subset.
+     * Asserts that the retrieved JSON contains the specified subset.
      *
-     * This method delegates to self::assertArraySubset().
+     * This method delegates to static::assertArraySubset().
      *
      * @param array|string $subset
      *
@@ -91,16 +95,32 @@ trait ApiTestAssertionsTrait
     public static function assertArraySubset($subset, $array, bool $checkForObjectIdentity = false, string $message = ''): void
     {
         $constraint = new ArraySubset($subset, $checkForObjectIdentity);
+
         static::assertThat($array, $constraint, $message);
     }
 
     /**
-     * @param array|string $jsonSchema
+     * @param object|array|string $jsonSchema
      */
     public static function assertMatchesJsonSchema($jsonSchema, ?int $checkMode = null, string $message = ''): void
     {
         $constraint = new MatchesJsonSchema($jsonSchema, $checkMode);
+
         static::assertThat(self::getHttpResponse()->toArray(false), $constraint, $message);
+    }
+
+    public static function assertMatchesResourceCollectionJsonSchema(string $resourceClass, ?string $operationName = null, string $format = 'jsonld'): void
+    {
+        $schema = self::getSchemaFactory()->buildSchema($resourceClass, $format, Schema::TYPE_OUTPUT, OperationType::COLLECTION, $operationName);
+
+        static::assertMatchesJsonSchema($schema);
+    }
+
+    public static function assertMatchesResourceItemJsonSchema(string $resourceClass, ?string $operationName = null, string $format = 'jsonld'): void
+    {
+        $schema = self::getSchemaFactory()->buildSchema($resourceClass, $format, Schema::TYPE_OUTPUT, OperationType::ITEM, $operationName);
+
+        static::assertMatchesJsonSchema($schema);
     }
 
     private static function getHttpClient(Client $newClient = null): ?Client
@@ -125,5 +145,17 @@ trait ApiTestAssertionsTrait
         }
 
         return $response;
+    }
+
+    private static function getSchemaFactory(): SchemaFactoryInterface
+    {
+        try {
+            /** @var SchemaFactoryInterface $schemaFactory */
+            $schemaFactory = static::$container->get('api_platform.json_schema.schema_factory');
+        } catch (ServiceNotFoundException $e) {
+            throw new \LogicException('You cannot use the resource JSON Schema assertions if the "api_platform.swagger.versions" config is null or empty.');
+        }
+
+        return $schemaFactory;
     }
 }
--- a/src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
+++ b/src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
@@ -25,16 +25,19 @@ use PHPUnit\Framework\Constraint\Constraint;
  */
 final class MatchesJsonSchema extends Constraint
 {
+    /**
+     * @var object|array
+     */
     private $schema;
     private $checkMode;
 
     /**
-     * @param array|string $schema
+     * @param object|array|string $schema
      */
     public function __construct($schema, ?int $checkMode = null)
     {
+        $this->schema = \is_string($schema) ? json_decode($schema) : $schema;
         $this->checkMode = $checkMode;
-        $this->schema = \is_array($schema) ? (object) $schema : json_decode($schema);
     }
 
     /**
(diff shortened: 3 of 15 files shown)


src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
```diff
@@ -126,4 +135,19 @@ private static function getHttpResponse(): ResponseInterface
...
+    private static function getSchemaFactory(): SchemaFactoryInterface
+    {
+        static $schemaFactory;
+
+        if (null !== $schemaFactory) {
+            return $schemaFactory;
+        }
+
+        if (!isset(static::$container) || !static::$container instanceof ContainerInterface || !static::$container->has('api_platform.json_schema.schema_factory')) {
+            static::fail();
```
dunglas: Shouldn't we throw an exception instead?

src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
```diff
@@ -95,14 +98,20 @@ public static function assertArraySubset($subset, $array, bool $checkForObjectId
...
+     * @param Schema|array|string $jsonSchema
      */
     public static function assertMatchesJsonSchema($jsonSchema, ?int $checkMode = null, string $message = ''): void
     {
         $constraint = new MatchesJsonSchema($jsonSchema, $checkMode);
+
         static::assertThat(self::getHttpResponse()->toArray(false), $constraint, $message);
     }
 
+    public static function assertMatchesResourceJsonSchema(string $resourceClass, string $format = 'json', bool $output = true, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): void
```
teohhanhui: This seems like a nightmarish assertion method. :see_no_evil:

Probably not a good idea. It'll be so unreadable.

dunglas: Could just be `string $resourceClass, string $format = 'json', bool $output = true, ?string $operationType = null, ?string $operationName = null` IMHO.

Also, we should rename `output` in `response`.

teohhanhui: Do you want to read in your test:

```php
$this->assertMatchesResourceJsonSchema(Book::class, 'json', true, 'collection', 'post');
```

I'm sure I don't. It's completely ridiculous as an assertion method.

soyuka: :+1: what is output used for? 

We should try to move some usual arguments in the assertion function name then, something like `assertCollectionMatchesResourceJsonSchema`? 
Does it make sense to have another format then `json` to compare against a json schema?
Not sure to get how it differs by operation name (`post`, `get`) for the comparison.

teohhanhui: > what is output used for?

It's actually to choose the type of the schema, i.e. input schema or output schema.

teohhanhui: @dunglas I think it's better if we can change `SchemaFactory::buildSchema` to take a `string $type` argument instead, with constants for input and output.

soyuka: Okay then we'll have 3 arguments:
- format
- operation
- type (Output/Input constants defaults to Output)

dunglas: @teohhanhui LGTM! Can you provide open a PR? Or even add a commit to this one?

tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
```diff
@@ -120,6 +121,18 @@ public function testAssertMatchesJsonSchema(): void
         $this->assertMatchesJsonSchema(json_decode($jsonSchema, true));
     }
 
+    public function testAssertMatchesResourceCollectionJsonSchema(): void
+    {
+        self::createClient()->request('GET', '/resource_interfaces');
+        $this->assertMatchesResourceCollectionJsonSchema(ResourceInterface::class);
```
teohhanhui: This test is failing because of https://github.com/justinrainbow/json-schema/issues/523

I'm not sure what we could do here.

composer.json
```diff
@@ -43,7 +43,7 @@
         "friendsofsymfony/user-bundle": "^2.2@dev",
         "guzzlehttp/guzzle": "^6.0",
         "jangregor/phpstan-prophecy": "^0.3",
-        "justinrainbow/json-schema": "^5.0",
+        "justinrainbow/json-schema": "^5.2",
```
teohhanhui: https://github.com/justinrainbow/json-schema/pull/378

teohhanhui:
The Travis failures for Elasticsearch are... weird. The same Behat tests pass on CircleCI.

src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
```diff
@@ -91,18 +95,34 @@ public static function assertJsonEquals($json, string $message = ''): void
     public static function assertArraySubset($subset, $array, bool $checkForObjectIdentity = false, string $message = ''): void
     {
         $constraint = new ArraySubset($subset, $checkForObjectIdentity);
+
         static::assertThat($array, $constraint, $message);
     }
 
     /**
-     * @param array|string $jsonSchema
+     * @param Schema|array|string $jsonSchema
```
dunglas: Could be `\ArrayObject` instead of `Schema` isn't it?

src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
```diff
@@ -126,4 +146,19 @@ private static function getHttpResponse(): ResponseInterface
 
         return $response;
     }
+
+    private static function getSchemaFactory(): SchemaFactoryInterface
+    {
+        static $schemaFactory;
```
dunglas: Wouldn't it be safe to always fetch this service from the container?

src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
```diff
@@ -29,12 +30,16 @@ final class MatchesJsonSchema extends Constraint
     private $checkMode;
 
     /**
-     * @param array|string $schema
+     * @param Schema|array|string $schema
      */
     public function __construct($schema, ?int $checkMode = null)
     {
+        /** @var array|string $schema */
+        $schema = $schema instanceof Schema ? json_encode($schema) : $schema;
```
dunglas:
```suggestion
        $schema = $schema instanceof \ArrayObject ? json_encode($schema) : $schema;
```

src/Hydra/JsonSchema/SchemaFactory.php
```diff
@@ -74,24 +74,29 @@ public function buildSchema(string $resourceClass, string $format = 'jsonld', bo
...
-                            'type' => 'integer',
-                            'minimum' => 1,
+                            'type' => 'string',
+                            'format' => 'iri-reference',
                         ],
                         'hydra:next' => [
-                            'type' => 'integer',
-                            'minimum' => 1,
+                            'type' => 'string',
+                            'format' => 'iri-reference',
```
dunglas: I wasn't aware of the existence of `iri-reference`! We should use it everywhere we know that an IRI will be generated!

src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
teohhanhui: Actually, do we really need to support this here?

src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
teohhanhui: Actually, what about not supporting this here at all?

src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
```diff
@@ -46,15 +47,16 @@ public function toString(): string
...
      */
     protected function matches($other): bool
     {
         if (!class_exists(Validator::class)) {
             throw new \RuntimeException('The "justinrainbow/json-schema" library must be installed to use "assertMatchesJsonSchema()". Try running "composer require --dev justinrainbow/json-schema".');
         }
 
-        $other = (object) $other;
+        $other = (string) json_encode($other);
+        $other = json_decode($other);
```
soyuka: mhh can we comment/explain why we need to do this? Also used in `additionalFailureDescription` mb put this in a priv commented method to avoid code duplication?

src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
```diff
@@ -80,4 +77,20 @@ protected function additionalFailureDescription($other): string
 
         return implode("\n", $errors);
     }
+
+    private function canonicalizeJson($data)
+    {
+        if (!(\is_array($data) || \is_object($data))) {
```
dunglas:
```suggestion
        if (!\is_array($data) && !\is_object($data)) {
```

For readability?

teohhanhui: Okay... Though to me `NOT (A OR B)` is clearer than `NOT A AND NOT B`.

src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
```diff
@@ -80,4 +77,20 @@ protected function additionalFailureDescription($other): string
 
         return implode("\n", $errors);
     }
+
+    private function canonicalizeJson($data)
```
teohhanhui: We support validating any JSON document against a JSON Schema now. :tada:

src/Bridge/Symfony/Bundle/Test/Constraint/MatchesJsonSchema.php
```diff
@@ -80,4 +83,27 @@ protected function additionalFailureDescription($other): string
 
         return implode("\n", $errors);
     }
+
+    /**
+     * Normalizes a JSON document.
+     *
+     * Specifically, we need to ensure that a JSON object is represented as a PHP object, not as an associative array.
+     */
+    private function normalizeJson($document)
```
teohhanhui: We support validating any JSON document against a JSON Schema now. :tada:

src/Hydra/JsonSchema/SchemaFactory.php
teohhanhui: @dunglas Do you have any in mind?

composer.json
```diff
@@ -43,7 +43,7 @@
         "friendsofsymfony/user-bundle": "^2.2@dev",
         "guzzlehttp/guzzle": "^6.0",
         "jangregor/phpstan-prophecy": "^0.4.2",
-        "justinrainbow/json-schema": "^5.0",
+        "justinrainbow/json-schema": "^5.2.1",
```
teohhanhui: https://github.com/justinrainbow/json-schema/pull/378
https://github.com/justinrainbow/json-schema/pull/409

teohhanhui:
Simplified the implementation a little.

src/Hydra/JsonSchema/SchemaFactory.php
soyuka: he means in the rest of the core, so def another pr!

soyuka:
thanks @meyerbaptiste @teohhanhui !

PR #3004, opened 2019-08-20
Fix docblock DataTransformerInterface::supportsTransformation

cacahouwete:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Update docblock for data transformer interface. Support method signature was wrong, object name is wrong too. It can be object or array depending if we normalize / denormalize.

### DIFF
--- a/src/DataTransformer/DataTransformerInterface.php
+++ b/src/DataTransformer/DataTransformerInterface.php
@@ -31,9 +31,9 @@ interface DataTransformerInterface
     public function transform($object, string $to, array $context = []);
 
     /**
-     * Checks whether the transformation is supported for a given object and context.
+     * Checks whether the transformation is supported for a given data and context.
      *
-     * @param object $object
+     * @param object|array $data object on normalize / array on denormalize
      */
-    public function supportsTransformation($object, string $to, array $context = []): bool;
+    public function supportsTransformation($data, string $to, array $context = []): bool;
 }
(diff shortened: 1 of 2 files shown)


dunglas:
Can you switch to 2.4 as target branch as it's a bug fix please?

teohhanhui:
I think this points to a deeper problem. The `DataTransformerInterface` does not differentiate the direction of transformation.

dunglas:
> I think this points to a deeper problem. The DataTransformerInterface does not differentiate the direction of transformation.

Indeed, but we must merge this PR anyway. @cacahouwete let me know when the rebase is done.

soyuka:
> I think this points to a deeper problem. The DataTransformerInterface does not differentiate the direction of transformation.

It has been done like this by design.

teohhanhui:
@soyuka I think it's bad design. Look at the code the users have to write. It's not clean.

soyuka:
> @soyuka I think it's bad design. Look at the code the users have to write. It's not clean.

In fact we wanted to implement the automapper and even make this part optional. We didn't wanted to add 2 interfaces for this.

cacahouwete:
@dunglas Rebase done.

dunglas:
Thanks @cacahouwete!

PR #3007, opened 2019-08-20
Improve serialization groups filter doc

fredericbarthelet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2891 
| License       | MIT
| Doc PR        | -

@dunglas @teohhanhui, I got inspired by existing `getDescription` function of FilterProperty to add `'enum'` information to generated documentation.
https://github.com/api-platform/core/blob/master/src/Serializer/Filter/PropertyFilter.php#L73-L105

However, I was wondering if I should not use your newly created `TypeFactory` to enrich generated array. I would do that by :
- creating a new `Schema` in `FilterProperty::getDescription` method with one unique `enum` property set to the withlisted serialization groups.
- return description array with a new `schema` key
- update https://github.com/api-platform/core/blob/master/src/Swagger/Serializer/DocumentationNormalizer.php#L690 to pass optional `schema` value found in parameter data as last `getType` argument
- update `TypeFactory::getType` method to use provided schema to improve returned type (not only for `BUILTIN_TYPE_OBJECT` as it is the case right now)

What do you think ? This could be ported as well on #2971, which is currently relying on a new `Schema` property defined without the `Schema::class` type.

#hackdayparis

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -706,11 +706,18 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
                     'required' => $data['required'],
                 ];
 
+                if (isset($data['description'])) {
+                    $parameter['description'] = $data['description'];
+                }
+
                 $type = \in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string'];
                 $v3 ? $parameter['schema'] = $type : $parameter += $type;
 
                 if ($v3 && isset($data['schema'])) {
                     $parameter['schema'] = $data['schema'];
+                }
+                if (!$v3 && isset($data['schema']['items'])) {
+                    $parameter['items'] = $data['schema']['items'];
                 }
 
                 if ('array' === ($type['type'] ?? '')) {
@@ -724,6 +731,7 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
                     }
                 }
 
+                /** @deprecated schema key should be used instead */
                 $key = $v3 ? 'openapi' : 'swagger';
                 if (isset($data[$key])) {
                     $parameter = $data[$key] + $parameter;
--- a/src/Api/FilterInterface.php
+++ b/src/Api/FilterInterface.php
@@ -29,20 +29,17 @@ interface FilterInterface
      *   - required: if this filter is required
      *   - strategy (optional): the used strategy
      *   - is_collection (optional): is this filter is collection
-     *   - swagger (optional): additional parameters for the path operation,
-     *     e.g. 'swagger' => [
-     *       'description' => 'My Description',
-     *       'name' => 'My Name',
-     *       'type' => 'integer',
+     *   - schema (optional): additional parameters related to schema description
+     *     e.g. 'schema' => [
+     *         'type' => 'string',
+     *         'enum' => [
+     *             'value1',
+     *             'value2',
+     *         ],
      *     ]
-     *   - openapi (optional): additional parameters for the path operation in the version 3 spec,
-     *     e.g. 'openapi' => [
-     *       'description' => 'My Description',
-     *       'name' => 'My Name',
-     *       'schema' => [
-     *          'type' => 'integer',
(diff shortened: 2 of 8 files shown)


teohhanhui:
The filter's description should not be tied to the serialization formats. But it might be a good idea to use a `Schema` object instead of an array.

dunglas:
I agree with @teohhanhui, the filter should be agnostic of the documentation format (Swagger/OpenAPI/Hydra etc). However, describing it with a JSON Schema looks OK to me. Formats relying on JSON Schema (such as OpenAPI/Swagger) can use it directly, and others (such as Hydra) will have to convert it. As JSON Schema is kind of a standard (an Internet Draft actually, but that should become a RFC at some point), using it looks better to me than creating another custom format of our own.

So the filter should only provide a `schema` key (instead of `swagger` and `openapi`).

dunglas:
Using a plain array (following the JSON Schema semantic) looks better to me than using a `Schema` instance (easier for the end user). We could support both, (it's easy to transform an array in a `Schema` instance in `DocumentationNormalizer`).

fredericbarthelet:
@teohhanhui @dunglas understood, no Schema use in getDescription return value.

As of now we than have 2 strategy to "enrich" generated plain array description, that will then be eventually used by the DocumentationNormalizer :
- using `swagger` and `openapi` key (like in this PR or for PropertyFilter)
- using `schema` key (since #2971 like for OrderFilter)

Which one do you prefer in order to ensure consistency in this array ?
Should I add a `Description` object to standardize en ensure `getDescription` returns are all consistent and avoid this discussion in the future ?

fredericbarthelet:
@teohhanhui @dunglas I made a small concept PR to show you what I mean by `Description` object : https://github.com/fredericbarthelet/core/pull/2/files

This would ensure :
- normalization of getDescription output (thus avoiding the above conversation). PHPdoc on the Filter interface did not prevent me from introducing additional keys (which is bad i guess ?)
- cleaner fetching of additional parameters based on current documentation version (swagger or openapi)

WDYT ?

teohhanhui:
Uhh... I didn't realize we have `swagger` and `openapi` keys in the filter's description. :disappointed:

That's bad and should be deprecated.

fredericbarthelet:
@teohhanhui, noted, I removed swagger and openapi keys in my `Description` proposal. It is OK then to normalize this output using an object like described in https://github.com/fredericbarthelet/core/pull/2/files ? If so, I will update this PR replacing all occurrences of swagger and openapi keys with new `Description` object.

teohhanhui:
No, sorry. My comment is about those keys in the existing implementation.

As for your suggestion of using a class, I don't think we'll do it. (Personally, I'd like to see everything being well-defined classes, but I understand it's not what we'd like to do in this project. That's why you see `array $context = []` a lot.)

fredericbarthelet:
@teohhanhui I updated `FilterInterface` PHPDoc to reflect decision not to use `swagger` and `openapi` keys. I added non-existing `schema`, `items` and `description` keys in the description. I updated my PR to match this "interface" and remove remaining swagger and openapi usage on `PropertyFilter`.

For my knowledge, what is the advantage of using plain arrays instead of object to describe "options" arguments ? Is something like `src/JsonSchema/Schema.php` a sort of compromise, allowing for requirement in construction of certain arguments, but leaving free the possibility to add properties not initially defined on the object ?

fredericbarthelet:
Hi @teohhanhui, I updated my PR with latest changes from master. Did you see my previous message :) ? Is there any remaining modification to be made on this PR ?

fredericbarthelet:
Hello @teohhanhui, sorry to insist, but could you let me know if this PR is fine for you now :) ?

teohhanhui:
Sorry, I've somehow missed this one. :bowing_man:

I'm going to take a look now.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -724,6 +731,7 @@ private function getFiltersParameters(bool $v3, string $resourceClass, string $o
                     }
                 }
 
+                /** @deprecated schema key should be used instead */
```
teohhanhui: We need to trigger deprecation errors instead. This phpdoc is unnecessary.

PR #3019, opened 2019-08-27
make oauth2 implicit workflow redirect work

WhiteRabbitDE:
This PR adds the missing oauth2 redirect html file and sets the oauth2RedirectUrl when creating the SwaggerUIBundle.

I would appreciate some input on how to pass the generated URL to the init-swagger.js in a more elegant way. 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Fixed tickets | -
| License       | MIT
| Doc PR        | -


teohhanhui:
Could you explain why this needs to be in API Platform?

WhiteRabbitDE:
> Could you explain why this needs to be in API Platform?

If you use enable Oauth in the config, the SwaggerUI gives you the possibility to authorize yourself and adds the bearer token to the requests so you can still test your request in the UI when they are secured.

At the momen, the implicit workflow (using a redirect from your provider back to the swaggerui) is not working because the html file is missing and the oauth2RedirectUrl is not set properly.

teohhanhui:
Yeah, this can target the `2.4` branch as it's a bug fix. Also, could you add a line to the `update-js.sh` script (similar to what you've done in #3021) so that this file is copied correctly when we update the JS deps?

Thanks!

src/Bridge/Symfony/Bundle/Resources/public/swagger-ui/oauth2-redirect.html
```diff
@@ -0,0 +1,67 @@
...
+        if (/code|token|error/.test(window.location.hash)) {
+            qp = window.location.hash.substring(1);
+        } else {
+            qp = location.search.substring(1);
+        }
+
+        arr = qp.split("&")
+        arr.forEach(function (v,i,_arr) { _arr[i] = '"' + v.replace('=', '":"') + '"';})
+        qp = qp ? JSON.parse('{' + arr.join() + '}',
+            function (key, value) {
```
teohhanhui: I noticed there are whitespace differences from https://github.com/swagger-api/swagger-ui/blob/7cae0d2bdf079fe2f5514f58920eb9e3c1260d14/dist/oauth2-redirect.html

Can you copy the file as it is? Thanks!

src/Bridge/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig
```diff
@@ -11,7 +11,8 @@
     {% endblock %}
 
     {# json_encode(65) is for JSON_UNESCAPED_SLASHES|JSON_HEX_TAG to avoid JS XSS #}
-    <script id="swagger-data" type="application/json">{{ swagger_data|json_encode(65)|raw }}</script>
+    <script id="swagger-data" type="application/json" data-oauth-redirect-url="{{ absolute_url(asset('bundles/apiplatform/swagger-ui/oauth2-redirect.html')) }}">{{ swagger_data|json_encode(65)|raw }}</script>
```
teohhanhui: `data-oauth-redirect-url` doesn't belong here. It should be part of `swagger_data`. :smile:

update-js.sh
```diff
@@ -8,10 +8,12 @@ dest=src/Bridge/Symfony/Bundle/Resources/public/swagger-ui/
 if [[ -d "$dest" ]]; then
   rm -Rf "$dest"
 fi
-mkdir -p "$dest"
+
```
teohhanhui: Remove this blank line?

WhiteRabbitDE:
I will rework this mess and create a new PR targeting 2.4 branch.

PR #3020, opened 2019-08-27
Improve error message when Filter class is not imported

ismail1432:
Using a Filter class without indicate the FQCN throw the message `The filter class "%s" does not implement FilterInterface` but most if the time it's just because the Filter class was not imported.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Annotation/ApiFilter.php
+++ b/src/Annotation/ApiFilter.php
@@ -58,7 +58,7 @@ final class ApiFilter
         }
 
         if (!is_a($options['value'], FilterInterface::class, true)) {
-            throw new InvalidArgumentException(sprintf('The filter class "%s" does not implement "%s".', $options['value'], FilterInterface::class));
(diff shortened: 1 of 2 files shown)


src/Annotation/ApiFilter.php
```diff
@@ -58,7 +58,7 @@ public function __construct($options = [])
         }
 
         if (!is_a($options['value'], FilterInterface::class, true)) {
-            throw new InvalidArgumentException(sprintf('The filter class "%s" does not implement "%s".', $options['value'], FilterInterface::class));
+            throw new InvalidArgumentException(sprintf('The filter class "%s" does not implement "%s". Did you forget a use statement ?', $options['value'], FilterInterface::class));
```
soyuka: isn't there a test to update on this?

ismail1432: I think you’re right ! I update test too

soyuka:
thanks @ismail1432

Issue #2963, opened 2019-08-01
RangeFilter throws 500 when trying to search with float by integer field


karser:
Hello, I encountered an issue where RangeFilter is searching with a float value by integer field.
I'm getting an error: SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: "9.5"

It turns out that you can't bind float to where clause in Postgres https://stackoverflow.com/questions/57314159/php-pdo-cant-bind-a-float-parameter-to-search-by-integer-field

But still, shouldn't RangeFilter cast the search value to the field type to avoid this error? In this case,  `float` to `integer`?

karser:
Here is what can be done in RangeFilter:

Solution 1: cast left side to float (Postgres-specific):
```
$stmt = $con->prepare('SELECT t.* FROM my_table t WHERE CAST(t.integer_field AS FLOAT) > ?');
$stmt->execute([9.5]);
```

Solution 2: bypass pdo parameters bindings (more generic though):
```
$v = '9.5'; //user input
$v = filter_var($v, FILTER_VALIDATE_FLOAT);
if ($v === false) {
    throw new InvalidArgumentException("Property {$property}: '{$v}' is not valid number");
}
$stmt = $con->prepare('SELECT t.* FROM my_table t WHERE t.integer_field > '.$v);
$stmt->execute();
```

Solution 3: truncate / round the search value and cast its type to `int` if the field type is integer.

What do you think?

soyuka:
I'd do number 3 :p. Number 1 is fine though.

soyuka:
closing as we now have query string validation right?

PR #3037, opened 2019-09-02
[RangeFilter] Cast float value to int if field is integer (to avoid 500)

karser:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2963
| License       | MIT

Hello, here is Solution 3 implemented as @soyuka [suggested](https://github.com/api-platform/core/issues/2963#issuecomment-517616913)

### DIFF
--- a/src/Bridge/Doctrine/Common/Filter/RangeFilterTrait.php
+++ b/src/Bridge/Doctrine/Common/Filter/RangeFilterTrait.php
@@ -96,7 +96,7 @@ trait RangeFilterTrait
     /**
      * Normalize the values array for between operator.
      */
-    private function normalizeBetweenValues(array $values): ?array
+    private function normalizeBetweenValues(array $values, ?string $fieldType): ?array
     {
         if (2 !== \count($values)) {
             $this->getLogger()->notice('Invalid filter ignored', [
@@ -114,7 +114,7 @@ trait RangeFilterTrait
             return null;
         }
 
-        return [$values[0] + 0, $values[1] + 0]; // coerce to the right types.
+        return [$this->castValue($values[0], $fieldType), $this->castValue($values[1], $fieldType)];
     }
 
     /**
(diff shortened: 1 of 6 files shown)


teohhanhui:
I don't think this is the correct solution. It's better to just ignore the filter if the value is not in the correct type. (Or throw an exception leading to HTTP 400, if that makes sense.)

karser:
@teohhanhui I performed upfront research and suggested [3 solutions](https://github.com/api-platform/core/issues/2963#issuecomment-517435386) in the original issue. You should have posted your thought there before it's implemented.

One core member recommends this approach and another discards it as soon as it is implemented. This is frustrating.

teohhanhui:
@karser I understand your frustration, but not all of us are able to immediately respond to everything.

Thanks for your patience as we find the best solution together.

soyuka:
We definitely need to reach a consensus... Your work here, even if we choose to log instead of convert can still be used and a few lines need to be changed. I won't see this as "unnecessary work" on the contrary it's a good step forward!

karser:
Sure, let's discuss this once again.

>We should just ignore the filter if the value is not in the correct type (and log it, as we already do in some cases).

This is rare case and no one will notice that the float filter value is incorrect and silently ignored. And this bug will quickly be shipped to production. Logging will not help because no one reads logs unless sentry/graylog/etc is used.

If casting float to integer is not an option, I would rather throw a client error (400 code).
Although I still think that casting when possible is better than silently ignoring.

dunglas:
Both ignoring and casting are fine to me. The sure thing is that a malformed query parameter must not allow to throw a 500 in production. Until we reach a consensus, I'm all in favor of merging this.

soyuka:
> If casting float to integer is not an option, I would rather throw a client error (400 code).

throwing a 400 is too strong especially that it'd be triggerable through a query parameter :|.

> Although I still think that casting when possible is better than silently ignoring.

I do agree and I don't see the harm, if the documented type isn't respected the API should still work and it's the responsibility of the client to use the correct type when requesting it. Silently ignoring isn't an option and we would definitely log something.

teohhanhui:
We already ignore invalid filters while logging the problem. If the user tries to filter, but it has no effect, they'll try to find out why. Silently rounding is more dangerous as it may lead to incorrect behaviour.

PR #3046, opened 2019-09-04
WIP: Call Doctrine QueryExtension on Subresource request

ottaviano:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (I think)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Hi,
What do you think about calling the ItemExtensions for the parent resource query on Subrequest operation?

It can be useful if the parent resource has an ItemExtension like:

```php
class EnabledExtension implements QueryItemExtensionInterface, QueryCollectionExtensionInterface
{
    public function applyToItem(...)
    {
        if (is_subclass_of($resourceClass, EnabledInterface::class)) {
            $queryBuilder->andWhere(
                sprintf('%s.isEnabled = 1', $queryBuilder->getRootAliases()[0])
            );
        }
    }
}
```

otherwise, I find no way to change the parent subquery...

### DIFF
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -114,6 +114,10 @@ final class SubresourceDataProvider implements SubresourceDataProviderInterface
             }
         }
 
+        foreach ($this->itemExtensions as $extension) {
+            $extension->applyToItem($queryBuilder, $queryNameGenerator, $resourceClass, $identifiers, $context['subresource_operation_name'], $context);
+        }
+
         $query = $queryBuilder->getQuery();
 
         return $context['collection'] ? $query->getResult() : $query->getOneOrNullResult();


teohhanhui:
Can you share a use case so that it's easier to understand this?

ottaviano:
yes, for example, I have two entities: `Article` and `Comment`
```php
/**
 * @ApiResource
 */
class Article 
{
  /**
   * @ORM\Column(type="boolean")
   */
  private $published;

  /**
   * @ApiSubresource
   * 
   * @ORM\OneToMany(...)
   */
  private $comments;
}
```

I add a QueryExtension which modifies `Article` query for adding `article.published = true` condition.

```php
class ArticlePublishedExtension implements QueryItemExtensionInterface, QueryCollectionExtensionInterface
{
    public function applyToItem(...)
    {
        if (Article::class === $resourceClass) {
            $queryBuilder->andWhere(
                sprintf('%s.published = true', $queryBuilder->getRootAliases()[0])
            );
        }
    }
}
```

The queryEtension is well called for these routes: 
```
/articles
/articles/{id}
```

but not for `Comment` subresource query `/articles/{id}/comments`.

`SubresourceDataProvider` will generate this kind of query: 
```sql
SELECT 
 comment.*
FROM comment
WHERE 
  comment.article_id IN (
      SELECT id FROM articles WHERE id = :article_id
  )
```
I suggest calling the queryExtension for Subresource queries too, so the generated query will be: 
```sql
SELECT 
 comment.*
FROM comment
WHERE 
  comment.article_id IN (
      SELECT id FROM articles WHERE id = :article_id AND published = true
  )
```

soyuka:
I think that this kinda makes sense but it may break too many current implementations because we will run the actual item extensions... Subresources definitely need some refactor to ease everything :crying_cat_face:

PR #3056, opened 2019-09-09
Expose iri subformat

bpolaszek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR adds the `iri-reference` subformat to OpenAPI for declared resources. Other PHP objects aren't affected by this extra information.

iri-reference

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
@@ -8,6 +8,7 @@
         <defaults public="false" />
 
         <service id="api_platform.json_schema.type_factory" class="ApiPlatform\Core\JsonSchema\TypeFactory">
+            <argument type="service" id="api_platform.metadata.resource.name_collection_factory"></argument>
             <call method="setSchemaFactory">
                 <argument type="service" id="api_platform.json_schema.schema_factory"/>
             </call>
--- a/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
@@ -8,7 +8,7 @@
         <defaults public="false" />
 
         <service id="api_platform.json_schema.type_factory" class="ApiPlatform\Core\JsonSchema\TypeFactory">
-            <argument type="service" id="api_platform.metadata.resource.name_collection_factory"></argument>
+            <argument type="service" id="api_platform.metadata.resource.name_collection_factory"/>
             <call method="setSchemaFactory">
                 <argument type="service" id="api_platform.json_schema.schema_factory"/>
             </call>
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -31,16 +31,18 @@ final class TypeFactory implements TypeFactoryInterface
     private $schemaFactory;
 
     /**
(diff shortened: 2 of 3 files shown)


bpolaszek:
FYI: Swagger validation passes for both versions 2 and 3, `format` being an arbitrary string with recommended values.

src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
```diff
@@ -8,6 +8,7 @@
         <defaults public="false" />
 
         <service id="api_platform.json_schema.type_factory" class="ApiPlatform\Core\JsonSchema\TypeFactory">
+            <argument type="service" id="api_platform.metadata.resource.name_collection_factory"></argument>
```
dunglas:
```suggestion
            <argument type="service" id="api_platform.metadata.resource.name_collection_factory" />
```

src/JsonSchema/TypeFactory.php
```diff
@@ -29,6 +30,19 @@ final class TypeFactory implements TypeFactoryInterface
      */
     private $schemaFactory;
 
+    /**
+     * @var array
+     */
+    private $resourceClasses;
+
+    /**
+     * TypeFactory constructor.
```
dunglas: Useless docblock, to remove

src/JsonSchema/TypeFactory.php
```diff
@@ -29,6 +30,19 @@ final class TypeFactory implements TypeFactoryInterface
...
+     */
+    public function __construct(ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory = null)
+    {
+        $this->resourceClasses = $resourceNameCollectionFactory ? iterator_to_array($resourceNameCollectionFactory->create()) : [];
```
dunglas: Can you delay the call to `iterator_to_array` by moving it outside of the constructor, in case the service is constructed but not used, for performance reasons.

tests/JsonSchema/TypeFactoryTest.php
```diff
@@ -28,7 +30,16 @@ class TypeFactoryTest extends TestCase
      */
     public function testGetType(array $schema, Type $type): void
     {
-        $typeFactory = new TypeFactory();
+        $typeFactory = new TypeFactory(new ExtractorResourceNameCollectionFactory(
```
dunglas: Can't you use `this->prophesize(ResourceNameCollectionFactory::class)` instead? It will make the test easier to read

src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
bpolaszek: good catch

src/JsonSchema/TypeFactory.php
bpolaszek: sure.

src/JsonSchema/TypeFactory.php
bpolaszek: alright :-)

tests/JsonSchema/TypeFactoryTest.php
bpolaszek: Done.

Issue #3067, opened 2019-09-12
[GraphQL] Data missing in "delete" Mutation result


lukasluecke:
I am trying to access a relation's data of the deleted object with the following query:
```graphql
mutation DeleteItem($id: ID!) {
    deleteItem(input: {id: $id}) {
        item {
            id
            relatedItem
            {
                id
            }
        }
    }
```

This query fails with the following error message:
> Cannot return null for non-nullable field Item.relatedItem

The problematic code seems to be in:
https://github.com/api-platform/core/blob/0354379e3f016909c60e5d89d90a4bccb42cbfbb/src/GraphQl/Resolver/Stage/SerializeStage.php#L80-L84
https://github.com/api-platform/core/blob/0354379e3f016909c60e5d89d90a4bccb42cbfbb/src/GraphQl/Resolver/Stage/WriteStage.php#L53-L57

This is not valid, since the type we set in the Schema for the field `item` inside `deleteItemPayload` is `Item`, so it needs to have all the (required) data.

PR #3068, opened 2019-09-12
[GraphQL] Fix "delete" Mutation result

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not yet
| Fixed tickets | #3067
| License       | MIT
| Doc PR        | N/A

See #3067 for details.

### DIFF
--- a/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ItemMutationResolverFactory.php
@@ -91,9 +91,10 @@ final class ItemMutationResolverFactory implements ResolverFactoryInterface
                         'previous_object' => $previousItem,
                     ],
                 ]);
-                $item = ($this->writeStage)($item, $resourceClass, $operationName, $resolverContext);
+                $itemBeforeDelete = ($this->serializeStage)($item, $resourceClass, $operationName, $resolverContext);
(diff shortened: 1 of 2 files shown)


alanpoulain:
The fields in the Schema needs to be modified too.
I'm not sure returning all the fields of the deleted item is a good idea.
It seems the convention nowadays is to return the deleted id and the related items only (examples: https://help.shopify.com/en/api/graphql-admin-api/reference/mutation/customerdelete, https://developer.github.com/v4/mutation/deleteprojectcard/). Graphcool has chosen to return all the fields (https://www.graph.cool/docs/reference/graphql-api/mutation-api-ol0yuoz6go#deleting-a-node) but I can't see in which case it would be useful.

lukasluecke:
Why do the fields need to be modified, if we return all data? It might not seem useful to you now, but is there a reason against it? I think this would be the easiest solution, with the lowest cognitive overhead - introducing even more different types just for this seems overkill to me.

You can always just request the `id` in the mutation result if you want to, but sometimes you might want more data. I don't think that we should take Shopify or Github as examples for this case, since these are their public APIs, but the GraphQL-API of API Platform is also used "internally", e.g. for SPAs, and should therefore allow the developer more flexibility imo.

Issue #3025, opened 2019-08-27
Operation ids when dumping swagger json


WhiteRabbitDE:
Hi everyone,

in issue #3022 i noticed that the Operation id is generated with lcfirst. When you use uppercase in the annotations you end Up with Something like gETResourceCollection. 

Imho it would be better to replace lcfirst with lower. Example in the mentioned issue #3022.

Wdyt ?

soyuka:
I'd recommend to use camelCase and not a mix of uppercase/lowercase. Anyway, you could always override the [DocumentationNormalizer](https://github.com/api-platform/core/blob/1c19bf56d45debbfa83fcf80d051e0a2fd540f36/src/Swagger/Serializer/DocumentationNormalizer.php#L254) and use your own naming conventions there? 

I see no harm in modifying this though, feel free to open a pull request !

soyuka:
Mhh after a second look I think that there's a confusion here between a method and an operation name. The operation name has absolutely no reason to be written in Uppercase and we definitely should not change that. The error in #3022 comes from a configuration error/confusion and is expected to happen imo.

Issue #3022, opened 2019-08-27
Swagger generates operation id's with small first letter.


Nykilor:
I have a problem, when I tried to write functional tests for my app I got an following error 

		PHPUnit 7.5.15 by Sebastian Bergmann and contributors.

		Testing Project Test Suite
		2019-08-27T14:16:24+02:00 [critical] Uncaught PHP Exception ApiPlatform\Core\Exception\RuntimeException: "The operation "GETQuizCollection" cannot be found in the Swagger specification." at D:\xampp\htdocs\quizz\quizz-api\vendor\api-platform\core\src\Bridge\Symfony\Bundle\Action\SwaggerUiAction.php line 180
		E                                                                   1 / 1 (100%)

		Time: 4.33 seconds, Memory: 44.00 MB

		There was 1 error:

		1) App\Tests\QuizResourceApiTest::testRetriveQuizzesList
		Undefined property: Symfony\Component\DomCrawler\Crawler::$headers

		D:\xampp\htdocs\quizz\quizz-api\tests\QuizResourceApiTest.php:19

		ERRORS!
		Tests: 1, Assertions: 0, Errors: 1.

So I was wondering why do I get this if that route should be available, so I ran the command api:openapi:export, and to my surprise it appears that all the operationId start with a small letter.

		{
			"swagger": "2.0",
			"basePath": "\/",
			"info": {
				"title": "",
				"version": "0.0.0"
			},
			"paths": {
				"\/api\/answers": {
					"post": {
						"tags": [
							"Answer"
						],
						"operationId": "pOSTAnswerCollection",
						"consumes": [
							"application\/ld+json",
							"application\/json",
							"text\/html"
						],
						"produces": [
							"application\/ld+json",
							"application\/json",
							"text\/html"
						],
						"summary": "Creates a Answer resource.",
						"responses": {
							"201": {
								"description": "Answer resource created",
								"schema": {
									"$ref": "#\/definitions\/Answer"
								}
							},
							"400": {
								"description": "Invalid input"
							},
							"404": {
								"description": "Resource not found"
							}
						},
						"parameters": [
							{
								"name": "answer",
								"in": "body",
								"description": "The new Answer resource",
								"schema": {
									"$ref": "#\/definitions\/Answer"
								}
							}
						]
					}
				},
				"\/api\/answers\/{id}": {
					"put": {
						"tags": [
							"Answer"
						],
						"operationId": "pUTAnswerItem",
						"consumes": [
							"application\/ld+json",
[...]

PR #3069, opened 2019-09-13
use strtolower for operationNames in DocumentationNormalizer

WhiteRabbitDE:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Fixed tickets | #3025, #3022
| License       | MIT
| Doc PR        | -

It is a pretty minor thing, still ended up in an unexpected result ( see #3022 )


teohhanhui:
Sorry, I'm not sure this is the right fix. Which part is using the `operationId` in the first place?

WhiteRabbitDE:
> Sorry, I'm not sure this is the right fix. Which part is using the `operationId` in the first place?
I hope i got your question right.

The operationId is used in the path item object and is used to uniquely identify the operation. As described in #3025 you could now end up in an operationId like "pOSTAnswerCollection" because only the first letter of "POST" is transformed to lowercase. This happens when you use all Uppercase letters in your Annotations  ` collectionOperations={"POST"},`

teohhanhui:
```
Uncaught PHP Exception ApiPlatform\Core\Exception\RuntimeException: "The operation "GETQuizCollection" cannot be found in the Swagger specification." at D:\xampp\htdocs\quizz\quizz-api\vendor\api-platform\core\src\Bridge\Symfony\Bundle\Action\SwaggerUiAction.php line 180
```

This is the actual bug. We should fix it here.

soyuka:
@teohhanhui the problem here is that the operation will end up being `pOST...` instead of `POST...`. Lowercasing the whole method will definitely do the trick and it's definitely better to have `post` then `pOST` no?

teohhanhui:
And `postCustom` will become `postcustom`. Not good. As we've both agreed upon, this is the wrong use case. Don't use UPPERCASE for operation names.

soyuka:
Mhh yeah you're right, and I just got the "thing we should fix":

```
Uncaught PHP Exception ApiPlatform\Core\Exception\RuntimeException: "The operation "GETQuizCollection" cannot be found in the Swagger specification." at D:\xampp\htdocs\quizz\quizz-api\vendor\api-platform\core\src\Bridge\Symfony\Bundle\Action\SwaggerUiAction.php line 180
```

It should look for `gETQuizCollection` :p

soyuka:
https://github.com/api-platform/core/issues/3025#issuecomment-532184300 sorry about the miscomprehension on my end I should've thought that a bit more.

PR #3073, opened 2019-09-16
Disable put and delete operations support for elastic item data provider

devmaslov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I enabled elasticsearch support for one entity and everything worked fine until I made a put request. New entity was about to be created instead of updating existing one. 

It appears that when elasticsearch support is turned on even on put and delete requests entities are loaded from elastic and denormalized from the payload - as a result doctrine see them as detached entities. This leads to a new entity create on `put` operation and 500 error on `delete`.

I think elasticsearch item data provider shouldn't support either `put` or `delete` operations. Or is there a better way to fix this issue?

### DIFF
--- a/src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
+++ b/src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
@@ -63,7 +63,7 @@ final class ItemDataProvider implements ItemDataProviderInterface, RestrictedDat
             }
 
             $method = $resourceMetadata->getItemOperationAttribute($operationName, 'method');
-            if (\in_array($method, ['PUT', 'PATCH', 'DELETE'])) {
+            if (\in_array($method, ['PUT', 'PATCH', 'DELETE'], true)) {
                 return false;
             }
         } catch (ResourceClassNotFoundException $e) {
--- a/src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
+++ b/src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
@@ -56,7 +56,7 @@ final class ItemDataProvider implements ItemDataProviderInterface, RestrictedDat
      */
     public function supports(string $resourceClass, ?string $operationName = null, array $context = []): bool
     {
-        if (in_array($operationName, ['put', 'delete'])) {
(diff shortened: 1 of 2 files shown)


dunglas:
PATCH should also be disabled

src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
```diff
@@ -56,6 +56,10 @@ public function __construct(Client $client, DocumentMetadataFactoryInterface $do
      */
     public function supports(string $resourceClass, ?string $operationName = null, array $context = []): bool
     {
+        if (\in_array($operationName, ['put', 'delete'], true)) {
```
soyuka: I don't think we can related on the operation name here as the route method can be different from the operation name no?

devmaslov:
> PATCH should also be disabled

Agree. Then probably it makes sense to support only "GET" operations, not just exclude "put", "patch", "delete".

src/Bridge/Elasticsearch/DataProvider/ItemDataProvider.php
devmaslov: I see your point. I can just rename the operation to "put2" via annotations for example and the fix won't work already. 

Option #2
Use item operation "method" attribute in the resource metadata. It's a bit better, but we can't rely on it either, since it's possible to omit it. You can just set "route_name" and there won't be "method" property.

Option #3
Another option would be to directly check current request method.

soyuka: Meh yeah you're right about option 2... @teohhanhui what would you do?

teohhanhui:
Can we just make sure that no default `PUT`, `DELETE`, and `PATCH` operations are generated when elasticsearch support is enabled? I don't think there is a general fix.

devmaslov:
I guess our best shot then to just check whether it's a `PUT`, `DELETE` or `PATCH` operation based on operation method. If it's a custom operation without "method" specified for it, elasticsearch support should be disabled via attributes directly, if it's necessary.

Updated the pull request.

Issue #3051, opened 2019-09-05
Add a new attribute for "placeholder" routes


dunglas:
Because in API Platform (and when following the web architecture in general), it's better to identify every resources using an IRI, it's what does API Platform by default. While not mandatory, it's also considered a best practice for an IRI to be "dereferencable". When requesting the IRI, a representation of the resource, or the associated documentation should be displayed.

However, sometimes, you need to use reference to resources that aren't publicly available. Because API Platform internally uses the Symfony router to map objects to URL (forth and back), a route must exist for any resource type exposed through the API (or, in PHP term, for any class marked with `@ApiResource`). A common problem with API Platform is that because of this the `GET` operation must always exist, but sometimes you don't want this route (used to generate IRIs) to return anything. In this case, also want to this route to be hidden from the OpenAPI documentation, because it is useless for humans and code generator. This kind of routes are just placeholders used to generate identifiers.

My proposal is to introduce a new operation attribute, called `iriOnly`. When set to true, the route will exist but will return a 404 not found status code, and no PHP code, nor SQL request will be triggered. Also, such operations will be automatically hidden from OpenAPI and Swagger documentations. Optionally, a HTTP status code could be associated with this attribute, such as `iriOnly=204`.

teohhanhui:
I think it's better to provide a URN in such cases. We avoid the whole mess altogether.

teohhanhui:
> Note
> 
> As noted in § 1.1 How to Read this Document, IRIs can often be confused with URLs (Uniform Resource Locators), the primary distinction is that a URL locates a resource on the web, an IRI identifies a resource. While it is a good practice for resource identifiers to be dereferenceable, sometimes this is not practical. In particular, note the [URN] scheme for Uniform Resource Names, such as UUID. An example UUID is urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6.

> Authors are strongly encouraged to avoid labeling properties using blank node identifiers, instead, consider one of the following mechanisms:
> 
> * a relative IRI, either relative to the document or the vocabulary (see § 4.1.2.1 Using the Document Base for the Default Vocabulary for a discussion on using the document base as part of the vocabulary mapping).
> * a URN such as urn:example:1, see [URN], or
> * a "Skolem IRI" as per Replacing Blank Nodes with IRIs of [RDF11-CONCEPTS].

https://www.w3.org/TR/json-ld11/

dunglas:
[...]

PR #3079, opened 2019-09-18
Add the ability to declare empty item operations

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3051
| License       | MIT
| Doc PR        | na

### DIFF
--- a/src/Action/NotFoundAction.php
+++ b/src/Action/NotFoundAction.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Action;
+
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * Not found action.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class NotFoundAction
+{
+    public function __invoke()
+    {
+        throw new NotFoundHttpException();
+    }
+}
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -230,7 +230,7 @@ final class SchemaFactory implements SchemaFactoryInterface
             $inputOrOutput = $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, ['class' => $resourceClass], true);
         }
 
-        if (null === ($inputOrOutput['class'] ?? null)) {
+        if (false === ($inputOrOutput['class'] ?? false)) {
             // input or output disabled
             return null;
         }
--- a/src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Metadata\Resource\Factory;
 
+use ApiPlatform\Core\Action\NotFoundAction;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 
 /**
@@ -81,6 +82,13 @@ final class OperationResourceMetadataFactory implements ResourceMetadataFactoryI
             $resourceMetadata = $this->normalize(false, $resourceClass, $resourceMetadata, $itemOperations);
         }
 
+        if ($this->needsDefaultGetOperation($resourceMetadata)) {
+            $resourceMetadata = $resourceMetadata->withItemOperations(array_merge(
+                $resourceMetadata->getItemOperations(),
+                ['get' => ['method' => 'GET', 'read' => false, 'output' => ['class' => false], 'controller' => NotFoundAction::class]])
+            );
+        }
+
         $graphql = $resourceMetadata->getGraphql();
         if (null === $graphql) {
             $resourceMetadata = $resourceMetadata->withGraphql(['item_query' => [], 'collection_query' => [], 'delete' => [], 'update' => [], 'create' => []]);
@@ -147,5 +155,18 @@ final class OperationResourceMetadataFactory implements ResourceMetadataFactoryI
         }
 
         return $resourceMetadata->withGraphql($operations);
+    }
+
+    private function needsDefaultGetOperation(ResourceMetadata $resourceMetadata): bool
+    {
+        $itemOperations = $resourceMetadata->getItemOperations();
+
+        foreach ($itemOperations as $itemOperation) {
+            if ('GET' === ($itemOperation['method'] ?? false)) {
+                return false;
+            }
+        }
+
+        return true;
     }
 }
--- a/features/main/operation.feature
+++ b/features/main/operation.feature
@@ -55,3 +55,11 @@ Feature: Operation support
         }
     }
     """
+
+  Scenario: Get the collection of a resource that doesn't have a defined item operation
+    When I send a "GET" request to "/disable_item_operations"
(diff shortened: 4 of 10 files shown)


src/Action/PlaceholderAction.php
```diff
@@ -25,7 +25,7 @@ final class PlaceholderAction
      *
      * @return object
      */
-    public function __invoke($data)
+    public function __invoke($data = null)
```
soyuka: Should I maybe do a new controller instead of changing this?

dunglas: Yes, a new one that directly throw a 404 (so we can simplify the metadata).

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -78,6 +78,10 @@ public function create(string $resourceClass): ResourceMetadata
 
             $resourceMetadata = $resourceMetadata->withItemOperations($this->createOperations($methods));
         } else {
+            if (!$itemOperations) {
```
dunglas: You must explicitly check for a get operation (only PUT could be defined for instance).

src/Action/PlaceholderAction.php
teohhanhui: I thought we need to do it in the item data provider?

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -78,6 +78,10 @@ public function create(string $resourceClass): ResourceMetadata
...
+                $itemOperations = ['placeholder' => ['method' => 'GET', 'read' => false, 'serialize' => false, 'status' => 404]];
```
teohhanhui: The operation name should be `get` instead of `placeholder`.

teohhanhui: Can't we use `'output' => false` here?

soyuka: I wanted to be able to differentiate the get operation but sure can do!

teohhanhui: > to be able to differentiate the get operation

I know. I'm against that. :laughing:

I think there should be nothing special about it. Just a good / safe default when the `get` item operation is missing.

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
soyuka: how can this be done without links to the Router? I can move the `method` guess from ApiLoader to metadata first, but when using a `route_name`, we won't be able to check the method without retrieving the route from the RouteCollection right?

teohhanhui: > when using a route_name, we won't be able to check the method without retrieving the route from the RouteCollection right?

Applying the 80/20 rule, I'd say let's do nothing when we encounter such edge cases?

src/Action/NotFoundAction.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Not found action.
+ *
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ */
+final class NotFoundAction
+{
+    public function __invoke()
+    {
+        return new Response(null, 404);
```
soyuka: Would you prefer to throw an `HttpException` here?

dunglas: Yes please!

dunglas: Yes, it will trigger the custom error page system etc

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -81,6 +82,13 @@ public function create(string $resourceClass): ResourceMetadata
             $resourceMetadata = $this->normalize(false, $resourceClass, $resourceMetadata, $itemOperations);
         }
 
+        if ($this->needsDefaultGetOperation($resourceMetadata)) {
+            $resourceMetadata = $resourceMetadata->withItemOperations(array_merge(
+                $resourceMetadata->getItemOperations(),
+                ['get' => ['method' => 'GET', 'read' => false, 'serialize' => false, 'output' => ['class' => false], 'controller' => NotFoundAction::class]])
```
dunglas: I think that `serialize` can be omitted, because the controller throw anyway.

dunglas:
Could you open a docs PR for that @soyuka?

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -81,6 +82,13 @@ public function create(string $resourceClass): ResourceMetadata
             $resourceMetadata = $this->normalize(false, $resourceClass, $resourceMetadata, $itemOperations);
         }
 
+        if ($this->needsDefaultGetOperation($resourceMetadata)) {
+            $resourceMetadata = $resourceMetadata->withItemOperations(array_merge(
+                $resourceMetadata->getItemOperations(),
+                ['get' => ['method' => 'GET', 'read' => false, 'output' => ['class' => false], 'controller' => NotFoundAction::class]])
```
teohhanhui: I believe you've made a mistake here, it's supposed to be `null`:

https://github.com/api-platform/core/blob/81d276c33c871ee45605222d474181a503cdf1bd/src/Metadata/Resource/Factory/InputOutputResourceMetadataFactory.php#L95-L97

src/JsonSchema/SchemaFactory.php
```diff
@@ -230,7 +230,7 @@ private function getMetadata(string $resourceClass, string $type = Schema::TYPE_
             $inputOrOutput = $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, ['class' => $resourceClass], true);
         }
 
-        if (null === ($inputOrOutput['class'] ?? null)) {
+        if (false === ($inputOrOutput['class'] ?? false)) {
```
teohhanhui: Should be reverted.

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -148,4 +156,17 @@ private function normalizeGraphQl(ResourceMetadata $resourceMetadata, array $ope
 
         return $resourceMetadata->withGraphql($operations);
     }
+
+    private function needsDefaultGetOperation(ResourceMetadata $resourceMetadata): bool
+    {
+        $itemOperations = $resourceMetadata->getItemOperations();
+
+        foreach ($itemOperations as $itemOperation) {
+            if ('GET' === ($itemOperation['method'] ?? false)) {
```
teohhanhui: Why not `?? null` instead?

soyuka: We could document the custom get operation with one that disables it! I'm fixing what you said!

teohhanhui: > We could document the custom get operation with one that disables it!

What do you mean?

Anyway, I've given it more thought and it's still unsafe anyway. We don't know that there's not already a `GET` item operation because of `route_name`. So I really think we must revert this.

soyuka: see api-platform/docs#886

soyuka: you said it yourself  it's not recommended to use `route_name`?

teohhanhui: Yes, but if there's an operation using `route_name` which is **the** `GET` item operation, we could never tell, and we would be creating this `get` item operation with `NotFoundAction`. And there's nothing the user could do to fix that.

Issue #3074, opened 2019-09-17
References to Resources in an External API


mab05k:
I have 2 API services. 

API 1 Resource with a reference to a resource in API 2
```json
{
  "id": 1,
  "api2Resource": "https://example.com/api/api_2_resource/1"
}
```

I want to be able to store the Id of the external resource `1` in my Database, but present it as an IRI (with or without the hostname and protocol) in the API interface.

Is this something this project is willing to support?

soyuka:
I think that this is a use case for subresources. But yes it is supported as IRIs are linked to the Router, if the route is present it should work as expected.

mab05k:
That doesn't sound quite right for what I'm looking for. The "API 2" (think 3rd party API) resource is in a completely different Service/Application, and "API 1" would not be exposing that route, or have it defined in its router.

soyuka:
mhh you can take a look at adding a custom normalizer for this or maybe the IriConverter

mab05k:
The pull request I submitted adds ability for the Absolute URL. I'm also taking a look at putting something together for you to review that could support the external resources like I suggested.

Although, maybe this is something I should just implement in my own application instead of attempting to add the feature into the project? @soyuka

soyuka:
> Although, maybe this is something I should just implement in my own application instead of attempting to add the feature into the project? @soyuka

How would you add this? I'm not sure that we could easily find a generative way of doing so :|. I'd try for my application first and check if something can come out from it but definitely not sure on how it could work for everyone.

mab05k:
I was able to get it working to generate this. (Note: `/external_dummies` is at `http://example.com/external_dummies`)
```
  Scenario: External Resources should support hostnames and routes separate from the API
    Given there are 1 externalResource objects
    And I add "Accept" header equal to "application/ld+json"
    And I add "Content-Type" header equal to "application/json"
    And I send a "GET" request to "/external_dummies"
    And the JSON should be equal to:
    """
    {
      "@context": "\/contexts\/ExternalDummy",
      "@id": "\/external_dummies",
      "@type": "hydra:Collection",
      "hydra:member": [
        {
          "@id": "\/external_dummies\/1",
          "@type": "ExternalDummy",
          "externalResource": "https:\/\/external-api.com\/external_resources\/1",
          "id": 1
        }
      ],
      "hydra:totalItems": 1
    }
    """
```
[...]

PR #3084, opened 2019-09-22
add ability to return absolute urls instead of relative IRIs

mab05k:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3074
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/893

Adds configuration ability to return Absolute URLs in API requests.

```yaml
api_platform:
  absolute_url: true # defaults to false
```
This configuration will generate IRIs with the absolute URL
```json
    {
      "@context": "http:\/\/example.com\/contexts\/AbsoluteUrlRelationDummy",
      "@id": "http:\/\/example.com\/absolute_url_relation_dummies\/2",
      "@type": "AbsoluteUrlRelationDummy",
      "absoluteUrlDummies": [],
      "id": 2
    }
```


src/Bridge/Symfony/Routing/IriConverter.php
```diff
@@ -168,12 +170,21 @@ public function getItemIriFromResourceClass(string $resourceClass, array $identi
...
         } catch (RoutingExceptionInterface $e) {
             throw new InvalidArgumentException(sprintf('Unable to generate an IRI for "%s".', $resourceClass), $e->getCode(), $e);
         }
     }
 
+    /**
+     * @param array $parameters
+     * @param int   $referenceType
+     */
+    private function generate(string $name, $parameters = [], $referenceType = UrlGeneratorInterface::ABS_PATH): string
```
dunglas: It doesn't look good to me. This override globally a context parameter. A better approach would be to pass `null` as the default value of `$referenceType`, and only if it's null set the good default depending of the global configuration.

Also, please inline this, there is no need to create a new method and it will hurt performance as it's on the hot path.

.circleci/config.yml
```diff
@@ -471,6 +471,48 @@ jobs:
           file: build/logs/clover.xml
       - save-npm-cache
 
+  behat-absolute-url-coverage:
```
dunglas: Is it really necessary to create a new environment to test this config option?

mab05k: I don't know :). Do you have any recommendations? I assumed I needed to because it is a "global" configuration. Is there a way I can load certain configuration options for specific tests?

src/Bridge/Symfony/Routing/IriConverter.php
mab05k: Thanks for this. I agree it was a bit heavy-handed.

.circleci/config.yml
soyuka: I think that because of behat we're a bit limited over what we can configure, this looks fine to me. @teohhanhui maybe you have improvement thoughts on this?

src/Bridge/Symfony/Routing/IriConverter.php
mab05k: @dunglas I think I got them all - added tests for `hal`, `jsonld` and `jsonapi` and verified it works for the entrypoints as well.

.circleci/config.yml
mab05k: Any idea why the new environment is failing? 
```
#!/bin/bash -eo pipefail
mkdir -p "$(dirname -- "build/logs/clover.xml")"
phpcov merge --clover "build/logs/clover.xml" "build/coverage"
```
is giving the error 
```
The "build/coverage" directory does not exist.
```
but this is reusing the same configuration as the rest of the circleci tests.

antograssiot:
I guess I should try to rebase https://github.com/api-platform/core/pull/2708, it was basically allowing the same or closing it as superseeded by this one

soyuka:
continuing the work in https://github.com/api-platform/core/pull/3198

PR #3093, opened 2019-09-26
Fix disabled operation metadata

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/3079
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/886

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -89,7 +89,8 @@
     },
     "conflict": {
         "doctrine/common": "<2.7",
-        "doctrine/mongodb-odm": "<2.0"
+        "doctrine/mongodb-odm": "<2.0",
+        "doctrine/mongodb-odm-bundle": "4.0.0-RC2"
     },
     "suggest": {
         "doctrine/mongodb-odm-bundle": "To support MongoDB. Only versions 4.0 and later are supported.",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -89,7 +89,8 @@
     },
     "conflict": {
         "doctrine/common": "<2.7",
-        "doctrine/mongodb-odm": "<2.0"
+        "doctrine/mongodb-odm": "<2.0",
+        "doctrine/mongodb-odm-bundle": "4.0.0-RC2"
```
teohhanhui: We shouldn't do this...

alanpoulain: We will forget it.

teohhanhui: And it only conflicts with a specific version so when RC3 is out it'll break again. :laughing:

alanpoulain: RC3 should resolve the issue so I don't think it will break.

teohhanhui: Oh, okay. If it's a bug in the bundle, then I'm kind of fine with this. We can remove this conflict when RC3 is released.

teohhanhui: Anyway, this commit doesn't belong here. :laughing:

soyuka: it's a cherry pick to make things green :)

PR #3113, opened 2019-09-27
Fix BC-break when using short-syntax notation for access_control

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

This BC-break as introduced in https://github.com/api-platform/core/pull/2992

When using the `accesControl` attribute directly in the ApiRessource annotation, the `AttributeHydratorTrait` would throw the following:
```
Unknown property "accessControl" on annotation "ApiPlatform\Core\Annotation  \ApiResource". 
```

https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/Annotation/AttributesHydratorTrait.php#L54-L56

### DIFF
--- a/src/Annotation/AttributesHydratorTrait.php
+++ b/src/Annotation/AttributesHydratorTrait.php
@@ -49,12 +49,12 @@ trait AttributesHydratorTrait
             unset($values['attributes']);
         }
 
-        if (array_key_exists('accessControl', $values)) {
+        if (\array_key_exists('accessControl', $values)) {
             $values['security'] = $values['accessControl'];
             @trigger_error('Attribute "accessControl" is deprecated in annotation since API Platform 2.5, prefer using "security" attribute instead', E_USER_DEPRECATED);
             unset($values['accessControl']);
         }
-        if (array_key_exists('accessControlMessage', $values)) {
+        if (\array_key_exists('accessControlMessage', $values)) {
             $values['securityMessage'] = $values['accessControlMessage'];
             @trigger_error('Attribute "accessControlMessage" is deprecated in annotation since API Platform 2.5, prefer using "securityMessage" attribute instead', E_USER_DEPRECATED);
             unset($values['accessControlMessage']);
--- a/src/Annotation/AttributesHydratorTrait.php
+++ b/src/Annotation/AttributesHydratorTrait.php
@@ -49,6 +49,17 @@ trait AttributesHydratorTrait
             unset($values['attributes']);
         }
 
+        if (array_key_exists('accessControl', $values)) {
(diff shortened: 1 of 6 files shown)


teohhanhui:
This must target `2.5` instead of `master`.

antograssiot:
another option would be to update directly the AttributeExtractorTrait with something like, it would allow to trigger proper deprecation, WDYT ?
```php

        if (array_key_exists('accessControl', $values)) {
            $values['security'] = $values['accessControl'];
            @trigger_error('Attribute "accessControl" is deprecated in annotation since API Platform 2.5, prefer using "security" attribute instead', E_USER_DEPRECATED);
            unset($values['accessControl']);
        }
        if (array_key_exists('accessControlMessage', $values)) {
            $values['securityMessage'] = $values['accessControlMessage'];
            @trigger_error('Attribute "accessControlMessage" is deprecated in annotation since API Platform 2.5, prefer using "securityMessage" attribute instead', E_USER_DEPRECATED);
            unset($values['accessControlMessage']);
        }
```

teohhanhui:
You should cherry pick your commit. We don't want the commits from `master` to go into `2.5`.

antograssiot:
> You should cherry pick your commit. We don't want the commits from master to go into 2.5.

I know it was ongoing ;)

antograssiot:
I've updated the test to add `@group legacy` to avoid the folllowing in PHPUNIT, but it is not directly related and I don't like it very much... is there any other way to silence them ?
```
Remaining self deprecation notices (4)

  2x: Attribute "accessControl" is deprecated in annotation since API Platform 2.5, prefer using "security" attribute instead
    1x in BooleanFilterTest::setUp from ApiPlatform\Core\Tests\Bridge\Doctrine\Orm\Filter
    1x in SwaggerCommandTest::testExecuteWithAliasVersion3 from ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Command

  2x: Attribute "accessControlMessage" is deprecated in annotation since API Platform 2.5, prefer using "securityMessage" attribute instead
    1x in BooleanFilterTest::setUp from ApiPlatform\Core\Tests\Bridge\Doctrine\Orm\Filter
    1x in SwaggerCommandTest::testExecuteWithAliasVersion3 from ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Command
```

soyuka:
thanks @antograssiot !

PR #3128, opened 2019-10-01
Bugfix: Notice: Undefined index: method thrown in SerializerContextBu…

mhbert:
Bugfix: Notice: Undefined index: method thrown in SerializerContextBuilder (since version 2.5.0)
- switch the order of the checks in the if statement to prevent checking for non-existent 'method' key in $operation check.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no tests written
| Fixed tickets | no ticket yet
| License       | MIT

Upgrading from 2.4.* to 2.5 caused the above error. If you simply switch the order of the checks everything works as expected.

### DIFF
--- a/src/Serializer/SerializerContextBuilder.php
+++ b/src/Serializer/SerializerContextBuilder.php
@@ -94,7 +94,7 @@ final class SerializerContextBuilder implements SerializerContextBuilderInterfac
         }
 
         foreach ($resourceMetadata->getItemOperations() as $operation) {
-            if ('PATCH' === $operation['method'] && \in_array('application/merge-patch+json', $operation['input_formats']['json'] ?? [], true)) {
+            if (\in_array('application/merge-patch+json', $operation['input_formats']['json'] ?? [], true) && 'PATCH' === $operation['method']) {
                 $context['skip_null_values'] = true;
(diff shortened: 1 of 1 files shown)


src/Serializer/SerializerContextBuilder.php
```diff
@@ -94,7 +94,7 @@ public function createFromRequest(Request $request, bool $normalization, array $
         }
 
         foreach ($resourceMetadata->getItemOperations() as $operation) {
-            if ('PATCH' === $operation['method'] && \in_array('application/merge-patch+json', $operation['input_formats']['json'] ?? [], true)) {
+            if (\in_array('application/merge-patch+json', $operation['input_formats']['json'] ?? [], true) && 'PATCH' === $operation['method']) {
```
dunglas:
```suggestion
            if ('PATCH' === $operation['method'] && \in_array('application/merge-patch+json', $operation['input_formats']['json'] ?? [], true)) {
```

Because the comparison is faster.

dunglas:
Thanks @mhbert

Issue #3045, opened 2019-09-04
Error message for invalid UUID on the URL


raphaelgarnica:
Hi.

Having an API Resource with the following identifier configuration (I got it from the demo), when I send a request like this: `curl -X GET "https://demo.api-platform.com/books/wrong-id" -H "accept: application/ld+json"` It fails because of the denormalization, returning a 404 error thrown by the `ReadListener`

```
 /**
  * @var UuidInterface
  *
  * @ORM\Column(type="uuid", unique=true)
  * @ORM\Id
  * @ORM\GeneratedValue(strategy="CUSTOM")
  * @ORM\CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidGenerator")
  */
 private $id;
```

So, the error handling works perfectly. But, the message response of the request is: "Not found, because of an invalid identifier configuration". My question: What is the best approach to customize this message? It isn't a configuration mismatch on that case.

teohhanhui:
The exception message is indeed misleading. It should say something like: "Not found, because of an invalid identifier format"

WDYT?

rgarnica:
Oh. I just realized I opened this with the wrong user.

Anyway. I agree. We could throw a different exception type from the `ApiPlatform\Core\Bridge\RamseyUuid\Identifier\Normalizer\UuidNormalizer::denormalize` when the parse failed.

teohhanhui:
This should be the correct exception: https://github.com/symfony/symfony/blob/v4.3.4/src/Symfony/Component/Serializer/Exception/NotNormalizableValueException.php

rgarnica:
Great. It's ok if I work on it?

teohhanhui:
Please go ahead. PR welcome!

soyuka:
Mb related https://github.com/api-platform/core/issues/2191

djpremier:
https://github.com/api-platform/api-platform/issues/1594#issue-663347361

soyuka:
we fixed this in #3132 afaik

PR #3132, opened 2019-10-01
Update thrown exception and clarify error message for UuidNormalizer.

Taminoful:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes 
| Fixed tickets | #3045 
| License       | MIT
| Doc PR        | not applicable

As discussed in issue #3045, the given error message was misleading.
I also updated the thrown exception according to the Exception
mentioned in the same issue. Of course I updated the related PHPUnit
test too.

### DIFF
--- a/src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
+++ b/src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
@@ -13,10 +13,10 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\RamseyUuid\Identifier\Normalizer;
 
-use ApiPlatform\Core\Exception\InvalidIdentifierException;
 use Ramsey\Uuid\Exception\InvalidUuidStringException;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 
 /**
@@ -28,20 +28,22 @@ final class UuidNormalizer implements DenormalizerInterface
 {
     /**
      * {@inheritdoc}
+     *
+     * @throws NotNormalizableValueException Occurs when the identifier format is invalid
      */
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         try {
             return Uuid::fromString($data);
         } catch (InvalidUuidStringException $e) {
-            throw new InvalidIdentifierException($e->getMessage(), $e->getCode(), $e);
+            throw new NotNormalizableValueException('Not found, because of an invalid identifier format', $e->getCode(), $e);
         }
     }
 
     /**
      * {@inheritdoc}
      */
-    public function supportsDenormalization($data, $type, $format = null)
+    public function supportsDenormalization($data, $type, $format = null): bool
     {
         return is_a($type, UuidInterface::class, true);
     }
--- a/src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
+++ b/src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
@@ -13,10 +13,10 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\RamseyUuid\Identifier\Normalizer;
 
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
 use Ramsey\Uuid\Exception\InvalidUuidStringException;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
-use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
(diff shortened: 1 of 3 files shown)


Taminoful:
Currently fixing a behat testcase I've overlooked.

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
```diff
@@ -28,20 +28,22 @@ final class UuidNormalizer implements DenormalizerInterface
...
+     *
+     * @throws NotNormalizableValueException Occurs when the identifier format is invalid
      */
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         try {
             return Uuid::fromString($data);
         } catch (InvalidUuidStringException $e) {
-            throw new InvalidIdentifierException($e->getMessage(), $e->getCode(), $e);
+            throw new NotNormalizableValueException('Not found, because of an invalid identifier format', 404, $e);
```
soyuka: This exception doesn't take an http status code as second argument. I'd still keep the `$e->getMessage()` error but that's my opinion.

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
```diff
@@ -28,20 +28,22 @@ final class UuidNormalizer implements DenormalizerInterface
...
         }
     }
 
     /**
      * {@inheritdoc}
      */
-    public function supportsDenormalization($data, $type, $format = null)
+    public function supportsDenormalization($data, $type, $format = null): bool
```
soyuka: you can't change the return type here it's not in the interface

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
Taminoful: Hey @soyuka, I initially added the http status code instead of `$e->getMessage()` since the Behat test (`features/main/uuid.feature`) demands that it throws an 404 error (which would be the correct code) but when I run the test suite it returns 400 instead and I don't know why. Any ideas?

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
Taminoful: I'll remove it with my next commit. Thank you! 😃

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
soyuka: Because you changed the exception... InvalidIdentifierException is the correct one imo.  I'm not sure we should change that...

See ReadListener (I think) where InvalidIdentifierException is switched to a 404.

Taminoful: Ah! Thank you! @teohhanhui mentioned this exception as "it should be the correct one". TBH sounds `InvalidIdentifierException` like the correct exception here when you take the new exception message into account.

teohhanhui: Yes, `NotNormalizableValueException` is the correct one. We should not throw our own exception class here, as this is a (de)normalizer.

See https://github.com/symfony/symfony/blob/v4.3.4/src/Symfony/Component/Serializer/Normalizer/DenormalizerInterface.php#L45

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
teohhanhui: It's safe to add the return type here. :smile:

src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php
soyuka: Then we need to catch it [here](https://github.com/api-platform/core/blob/8d625d00404958d90dd6f94f84ccec26be30dd72/src/EventListener/ReadListener.php#L112) and throw a notfound but I'm afraid that it'll interfer with what should be a 400.

At first when I initiated this class I didn't wanted to use the symfony interface because they're not used in the serializer but in the identifier normalization phase.

teohhanhui: You have a point. Anyway, this PR is not the right fix for #3045.

IMO, we don't need to change anything here. Just change the error message in the `ReadListener` so that it's not misleading:

https://github.com/api-platform/core/blob/81f819b3befe1f0bb726e1901f4c8c6e500bb60e/src/EventListener/ReadListener.php#L112

`Invalid identifier value or configuration.`

soyuka: exactly

soyuka:
Thanks @Taminoful !

Issue #3099, opened 2019-09-26
Allow to set global defaults for ApiResource attributes and other options


dunglas:
It would be nice to configure global defaults for all class marked with `@ApiResource` (or configured in YAML/XML etc).

It could be a new config section such as the following:

```yaml
api_platform:
  defaults:
    attributes: # ...
    item_operations: # ...
    collection_operations: # ...
    normalization_context: # ...
    graphql: # ...
    # ...

# All properties defined in https://github.com/api-platform/core/blob/master/src/Annotation/ApiResource.php should be configurable
```

For instance, it would allow to make the API globally read only:

```yaml
api_platform:
    defaults:
        itemOperations: ["get"]
        collectionOperations: ["get"]
```

Other example, enable Mercure for all resource:

```yaml
api_platform:
    defaults:
        mercure: true
```

Or always use Messenger:

```yaml
api_platform:
    defaults:
        messenger: true
```

Implementing this feature should be trivial:

* add the new config option in https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
* Add a new `ResourceMetadataFactory`, with a low priority, applying the defaults if not set

Most options are stored in metadata's attributes, some other are top level metadata properties (itemCollections and collectionOperations and graphql AFAIR). Basically, the content of `defaults` should be passed as-is to resources' attributes, except the special one previously mentioned.

Existing configuration option for resource attributes (pagination, cache headers...) should probably be deprecated for consistency.

teohhanhui:
```yaml
api_platform:
    defaults:
        item_perations: ["get"]
        collection_perations: ["get"]
```

I don't think we need this. It'll make things very confusing...

xavren:
Could be nice to configure defaults operations to protect by default all `@ApiResource` this could be very usefull i think

julienfalque:
Is there anything left to do to implement this after #3151?

soyuka:
Indeed this should be fixed by #3151

teohhanhui:
I remember @dunglas discovered an issue with `cache_headers`.

PR #3135, opened 2019-10-02
Feature: resource default config

bpolaszek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3099
| License       | MIT

This PR introduces a `defaults` configuration as an implementation of #3099.

**Dummy implementation**
As discussed offline with @alanpoulain and @dunglas, the implementation has been thought to be as dummy as possible, i.e. any configurable property or attribute in `@ApiResource` (and even future ones) can be configured on a default basis.

This means this configuration is perfectly correct for all resources:
```yaml
api_platform:
    defaults:
        attributes:
            mercure: true
            messenger: true
```

But this one too:
```yaml
api_platform:
    defaults:
        description: foo
```

Which means unconfigured `@ApiResource`s will have the same description. That doesn't make sense, but we assume it's up to the developer, not the implementation, to decide what makes sense and what doesn't.

**Short-hand operations configuration aren't supported yet**
This short-hand configuration:
```yaml
api_platform:
    defaults:
        itemOperations: ["get"]
        collectionOperations: ["get"]
```
is currently normalized in a private method from another `ResourceMetadataFactory` with a higher priority. What are your recommendations to standardize it?

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -117,7 +117,6 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $this->registerElasticsearchConfiguration($container, $config, $loader);
         $this->registerDataTransformerConfiguration($container);
         $this->registerSecurityConfiguration($container, $loader);
-        $this->registerDefaultsConfiguration($container, $config);
 
         $container->registerForAutoconfiguration(DataPersisterInterface::class)
             ->addTag('api_platform.data_persister');
@@ -192,6 +191,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         if ($config['name_converter']) {
             $container->setAlias('api_platform.name_converter', $config['name_converter']);
         }
+        $container->setParameter('api_platform.defaults', $config['defaults']);
     }
 
     private function registerMetadataConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader): void
@@ -602,10 +602,5 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         if (isset($bundles['SecurityBundle'])) {
             $loader->load('security.xml');
         }
-    }
-
-    private function registerDefaultsConfiguration(ContainerBuilder $container, array $config): void
-    {
-        $container->setParameter('api_platform.defaults', $config['defaults']);
     }
 }
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -117,6 +117,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $this->registerElasticsearchConfiguration($container, $config, $loader);
         $this->registerDataTransformerConfiguration($container);
         $this->registerSecurityConfiguration($container, $loader);
+        $this->registerDefaultsConfiguration($container, $config);
 
         $container->registerForAutoconfiguration(DataPersisterInterface::class)
             ->addTag('api_platform.data_persister');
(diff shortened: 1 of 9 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -602,4 +603,9 @@ private function registerSecurityConfiguration(ContainerBuilder $container, XmlF
             $loader->load('security.xml');
         }
     }
+
+    private function registerDefaultsConfiguration(ContainerBuilder $container, array $config): void
+    {
+        $container->setParameter('api_platform.defaults', $config['defaults']);
```
alanpoulain: Maybe it would better to move it to `registerCommonConfiguration`?

src/Metadata/Resource/Factory/DefaultConfigurationResourceMetadataFactory.php
```diff
@@ -0,0 +1,65 @@
...
+namespace ApiPlatform\Core\Metadata\Resource\Factory;
+
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+
+/**
+ * Creates a resource metadata from the default configuration.
+ *
+ * @author Beno!t POLASZEK <bpolaszek@gmail.com>
+ */
+final class DefaultConfigurationResourceMetadataFactory implements ResourceMetadataFactoryInterface
```
alanpoulain: Defaults? To be uniform?

src/Metadata/Resource/Factory/DefaultConfigurationResourceMetadataFactory.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(string $resourceClass): ResourceMetadata
+    {
+        $resourceMetadata = $this->decorated->create($resourceClass);
+        $attributes = $resourceMetadata->getAttributes();
+
+        $baseConfig = array_filter($this->options, function (string $key) {
```
alanpoulain: Maybe we should just add a continue in the `foreach` instead of having to use `array_filter`?

tests/Metadata/Resource/Factory/DefaultConfigurationResourceMetadataFactoryTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+use ApiPlatform\Core\Metadata\Resource\Factory\DefaultConfigurationResourceMetadataFactory;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
+use PHPUnit\Framework\TestCase;
+
+class DefaultConfigurationResourceMetadataFactoryTest extends TestCase
+{
+    public function testCreate()
```
alanpoulain:
```suggestion
    public function testCreate(): void
```

src/Metadata/Resource/Factory/DefaultConfigurationResourceMetadataFactory.php
bpolaszek: My bad, refactored code which didn't get simplified.

src/Metadata/Resource/Factory/DefaultConfigurationResourceMetadataFactory.php
bpolaszek: Alright

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
bpolaszek: As you like.

bpolaszek:
As discussed offline with @dunglas, several issues may be raised with this approach:

- Defaults should be flattened (no more `attributes` sublevel)
- snake_case should be the used
- Using a decorating service with a low priority may prevent it from having the right configuration, as some data is normalized before.
- Some properties should indeed be prevented from having defaults, e.g. shortName.

I'll open another PR to reflect those changes.

Issue #3136, opened 2019-10-02
[GraphQL] Relations need `item_query` and `collection_query` to be set


lukasluecke:
Right now setting `item_query` and `collection_query` on a Resource will add these queries both to the root `Query` type, and also as fields to all other Resources that have a relation to this one.

I would like to make certain queries only available through relations, without "polluting" the root Query.

Disabling / removing `item_query` and `collection_query` seem to also remove the respective relation fields.

alanpoulain:
The relations are resources (they use the `@ApiResource` annotation), right? It seems logical to me that the root queries are created. If the user doesn't want them, it can disable the operations.
It's the same as REST.

lukasluecke:
If I disable the operations, then the relations are also not available - that's the issue 🙈 Sorry if my description was not clear enough.

alanpoulain:
It's a bug then since it should be.

lukasluecke:
Okay, so the relations should always be created? And the `item_query` and `collection_query` should only affect the root queries?

I'll try to hunt down the issue then 🙂

lukasluecke:
@alanpoulain The problem seems to be here: https://github.com/api-platform/core/blob/e468d98d260e0cf49d63d088fd5678ecab1b8912/src/GraphQl/Type/TypeConverter.php#L106-L110

An empty `graphql` array should mean that the Resource is still available to be exposed, just not directly accessible. What do you think?

alanpoulain:
It can be tried like this, yes.

lukasluecke:
How about changing the condition to the following? Would that be the correct behaviour? I'm not sure which cases the previous fallback was supposed to handle 🙈 

```php
if (null === $resourceMetadata->getGraphql()) {
    return null;
}
```

alanpoulain:
Will be fixed in 2.5.6.

PR #3138, opened 2019-10-02
[GraphQL] If resource has no operations, skip only root query fields

lukasluecke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3136 
| License       | MIT
| Doc PR        | N/A

This way we still generate the types necessary to use these resources elsewhere, e.g. in relations.

### DIFF
--- a/src/GraphQl/Type/TypeConverter.php
+++ b/src/GraphQl/Type/TypeConverter.php
@@ -105,7 +105,7 @@ final class TypeConverter implements TypeConverterInterface
 
         try {
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
-            if ([] === ($resourceMetadata->getGraphql() ?? [])) {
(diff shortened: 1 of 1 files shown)


lukasluecke:
@alanpoulain I think this is the correct approach - it will skip everything if we're at root level, and continue as expected for everything else (i.e. relations).

alanpoulain:
And if we change it to:
```php
if (null === $resourceMetadata->getGraphql()) {
```
What would be the consequences?

lukasluecke:
@alanpoulain I tried that first, but then it just generated everything alyways. I didn't check much further, what exactly was happening, but it didn't seem like the correct way to go about it.

alanpoulain:
Replaced by https://github.com/api-platform/core/pull/3532.
I think it's normal that the type would be generated and could be introspected if available through a relation.
Thank you anyway @lukasluecke!

PR #3144, opened 2019-10-03
[MongoDB] Possibility to add execute options

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/918

Allow to use execute options (corresponding to the [aggregate command fields](https://docs.mongodb.com/manual/reference/command/aggregate/#command-fields) in MongoDB) like `allowDiskUse`.

To do so, modify the resource configuration like this:
```php
/**
 * @ApiResource(attributes={
 *     "doctrine_mongodb"={
 *       "execute_options"={
 *         "allowDiskUse"=true
 *       }
 *     }
 * })
 * @ODM\Document
 */
```

### DIFF
--- a/src/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
+++ b/src/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\Bridge\Doctrine\MongoDbOdm\Extension\AggregationResultColle
 use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
 use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Repository\DocumentRepository;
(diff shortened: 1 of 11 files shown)


src/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
```diff
@@ -32,14 +33,16 @@
...
 {
     private $managerRegistry;
+    private $resourceMetadataFactory;
     private $collectionExtensions;
 
     /**
      * @param AggregationCollectionExtensionInterface[] $collectionExtensions
      */
-    public function __construct(ManagerRegistry $managerRegistry, iterable $collectionExtensions = [])
+    public function __construct(ManagerRegistry $managerRegistry, ResourceMetadataFactoryInterface $resourceMetadataFactory, iterable $collectionExtensions = [])
```
dunglas: Could be made optional (for convenience when not using the bundle), and as last parameter (for backward compatibility, even if I know this class is experimental) isn't it?

alanpoulain: > for convenience when not using the bundle

I don't understand this sentence. If the MongoDB ODM Bundle is not used, this service is not instantiated, so it shouldn't be a problem.

For BC, I don't have a strong opinion but why the need of having the `experimental` annotation then?

soyuka: it is experimental, it could indeed be optional but I'm not sure there's added value to that.

PR #3149, opened 2019-10-04
Fixed dependency issue when SecurityBundle is not installed

reypm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/api-platform/issues/1265
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
@@ -53,12 +53,12 @@
 
         <service id="api_platform.graphql.resolver.stage.security" class="ApiPlatform\Core\GraphQl\Resolver\Stage\SecurityStage" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
-            <argument type="service" id="api_platform.security.resource_access_checker" />
+            <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="ignore" />
         </service>
 
         <service id="api_platform.graphql.resolver.stage.security_post_denormalize" class="ApiPlatform\Core\GraphQl\Resolver\Stage\SecurityPostDenormalizeStage" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
-            <argument type="service" id="api_platform.security.resource_access_checker" />
+            <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="ignore" />
         </service>
 
         <service id="api_platform.graphql.resolver.stage.serialize" class="ApiPlatform\Core\GraphQl\Resolver\Stage\SerializeStage" public="false">
--- a/src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
+++ b/src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
@@ -45,16 +45,16 @@ final class SecurityPostDenormalizeStage implements SecurityPostDenormalizeStage
 
         $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'security_post_denormalize', null, true);
 
-        if (null !== $isGranted && null === $this->resourceAccessChecker) {
-            throw new \LogicException('Cannot check security expression when SecurityBundle is not installed. Try running "composer require symfony/security-bundle".');
-        }
-
         if (null === $isGranted) {
             // Backward compatibility
             $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'access_control', null, true);
             if (null !== $isGranted) {
                 @trigger_error('Attribute "access_control" is deprecated since API Platform 2.5, prefer using "security" attribute instead', E_USER_DEPRECATED);
             }
+        }
+
+        if (null !== $isGranted && null === $this->resourceAccessChecker) {
+            throw new \LogicException('Cannot check security expression when SecurityBundle is not installed. Try running "composer require symfony/security-bundle".');
         }
 
         if (null === $isGranted || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
--- a/src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
+++ b/src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
@@ -30,7 +30,7 @@ final class SecurityPostDenormalizeStage implements SecurityPostDenormalizeStage
     private $resourceMetadataFactory;
     private $resourceAccessChecker;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, ResourceAccessCheckerInterface $resourceAccessChecker)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, ?ResourceAccessCheckerInterface $resourceAccessChecker)
     {
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->resourceAccessChecker = $resourceAccessChecker;
@@ -44,6 +44,11 @@ final class SecurityPostDenormalizeStage implements SecurityPostDenormalizeStage
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
 
         $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'security_post_denormalize', null, true);
+
+        if (null !== $isGranted && null === $this->resourceAccessChecker) {
+            throw new \LogicException('Cannot check security expression when SecurityBundle is not installed. Try running "composer require symfony/security-bundle".');
+        }
+
         if (null === $isGranted) {
             // Backward compatibility
             $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'access_control', null, true);
--- a/src/GraphQl/Resolver/Stage/SecurityStage.php
+++ b/src/GraphQl/Resolver/Stage/SecurityStage.php
@@ -30,7 +30,7 @@ final class SecurityStage implements SecurityStageInterface
     private $resourceMetadataFactory;
     private $resourceAccessChecker;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, ResourceAccessCheckerInterface $resourceAccessChecker)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, ?ResourceAccessCheckerInterface $resourceAccessChecker)
     {
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->resourceAccessChecker = $resourceAccessChecker;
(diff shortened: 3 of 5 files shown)


alanpoulain:
You need to modify the code inside the stages too to handle the case when `$resourceAccessChecker` is null.

reypm:
> You need to modify the code inside the stages too to handle the case when `$resourceAccessChecker` is null.

What do you mean? Not following you on the word `stages` 🤔

alanpoulain:
The services you are modifying are "stages" (`SecurityStage`, `SecurityPostDenormalizeStage`).

reypm:
> The services you are modifying are "stages" (`SecurityStage`, `SecurityPostDenormalizeStage`).

You mean add an extra check at for `null` check at `src/GraphQl/Resolver/Stage/SecurityStage.php`??

```
    if (null === $isGranted || null === $this->resourceAccessChecker || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
        return;
    }
```

alanpoulain:
Even before this condition but that's the idea. And make the variable in the constructor nullable.
Don't forget to add unit tests too!

teohhanhui:
@reypm

```php
    if (null === $isGranted || null === $this->resourceAccessChecker || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
        return;
    }
```

This is wrong. We must not silently allow access if the user has specified some condition (i.e. `$isGranted` is not `null`) but `$this->resourceAccessChecker` is `null`. We should fail in this case.

alanpoulain:
Oh yes you're right @teohhanhui

reypm:
> We should fail in this case.

@teohhanhui and by fail you mean throw the Error as in last line on that `__invoke` function? I am following you right you want this? 

```
/**  @var ResolveInfo $info */
$info = $context['info'];

$isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'security', null, true);
if (null === $this->resourceAccessChecker && $isGranted) {
    throw Error::createLocatedError($resourceMetadata->getGraphqlAttribute($operationName, 'security_message', 'Access Denied.'), $info->fieldNodes, $info->path);
}

if (null === $isGranted || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
    return;
}

throw Error::createLocatedError($resourceMetadata->getGraphqlAttribute($operationName, 'security_message', 'Access Denied.'), $info->fieldNodes, $info->path);
```

@alanpoulain yes, I will add the test once the desired solution is approved by you both :) 

Sorry for not being following you guys, first time "fixing" an api-platform issue and understanding it's internal.

teohhanhui:
> @teohhanhui and by fail you mean throw the Error as in last line on that __invoke function? I am following you right you want this?

I think we should do something like:

```php
throw new \LogicException('Cannot check security expression when SecurityBundle is not installed. Try running "composer require symfony/security-bundle".');
```

as this is a configuration issue.

Or do you always want to avoid throwing exceptions at all costs, @alanpoulain? (Why?)

alanpoulain:
We can throw a `LogicException`, it will be converted in an `Error` when https://github.com/api-platform/core/pull/3063 will be merged (and it's not an issue in the meantime if it's not the case).

reypm:
@teohhanhui / @alanpoulain better now? Did I miss something else?

src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
```diff
@@ -52,7 +57,7 @@ public function __invoke(string $resourceClass, string $operationName, array $co
             }
         }
 
-        if (null === $isGranted || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
+        if (null === $isGranted || null === $this->resourceAccessChecker || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
```
teohhanhui: This change is unnecessary. It's not possible at this point that `null !== $isGranted && null === $this->resourceAccessChecker` would evaluate to `true` (exactly what we've already checked above). (I'd also suggest changing the above condition to this form, as it reads better.)

reypm: Ups! That was my bad I miss that check in there, fixing it now.

src/Bridge/Symfony/Bundle/Resources/config/graphql.xml
```diff
@@ -53,12 +53,12 @@
 
         <service id="api_platform.graphql.resolver.stage.security" class="ApiPlatform\Core\GraphQl\Resolver\Stage\SecurityStage" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
-            <argument type="service" id="api_platform.security.resource_access_checker" />
+            <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="ignore" />
```
dunglas: Should be `null`, not `ignore` isn't it? Or calling the constructor will still fail.

reypm: Why test are passing with this configuration? 🤔

dunglas: Because the security component is installed on Travis.

teohhanhui: No, they behave the same way. The only difference is when inside a method call.

https://symfony.com/doc/current/service_container/optional_dependencies.html#ignoring-missing-dependencies

teohhanhui: I suggest for consistency's sake, we should not use `on-invalid="null"` except in the specific cases where we need that behaviour (i.e. a method call with `null` value).

src/GraphQl/Resolver/Stage/SecurityStage.php
```diff
@@ -44,6 +44,10 @@ public function __invoke(string $resourceClass, string $operationName, array $co
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
 
         $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'security', null, true);
+        if (null === $this->resourceAccessChecker && $isGranted) {
```
alanpoulain: Please add a new line before?

src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
```diff
@@ -44,6 +44,11 @@ public function __invoke(string $resourceClass, string $operationName, array $co
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
 
         $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'security_post_denormalize', null, true);
+
+        if (null === $this->resourceAccessChecker && $isGranted) {
+            throw new \LogicException('Cannot check security expression when SecurityBundle is not installed. Try running "composer require symfony/security-bundle".');
+        }
+
```
alanpoulain: I think we have an issue if `null === $this->resourceAccessChecker` and `$isGranted` is false (see the code below).

tests/GraphQl/Resolver/Stage/SecurityPostDenormalizeStageTest.php
```diff
@@ -115,4 +115,34 @@ public function testNotGranted(): void
...
     }
+
+    public function testNoSecurityBundleInstalled(): void
+    {
+        $operationName = 'item_query';
+        $resourceClass = 'myResource';
+        $isGranted = 'not_granted';
+        $extraVariables = ['extra' => false];
+        $resourceMetadata = (new ResourceMetadata())->withGraphql(
+            [
```
alanpoulain: You can follow the other CS.

tests/GraphQl/Resolver/Stage/SecurityStageTest.php
```diff
@@ -96,4 +96,18 @@ public function testNotGranted(): void
...
+        $this->securityStage = new SecurityStage($this->resourceMetadataFactoryProphecy->reveal(), null);
+
+        $operationName = 'item_query';
+        $resourceClass = 'myResource';
+        $isGranted = 'not_granted';
+        $extraVariables = ['extra' => false];
+        $resourceMetadata = (new ResourceMetadata())->withGraphql([$operationName => ['security' => $isGranted]]);
+        $this->resourceMetadataFactoryProphecy->create($resourceClass)->willReturn($resourceMetadata);
+        $this->expectException(\LogicException::class);
+        ($this->securityStage)($resourceClass, 'item_query', ['extra_variables' => $extraVariables]);
```
alanpoulain: Why the CS is so different (new lines, ...)?

src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
```diff
@@ -44,6 +44,11 @@ public function __invoke(string $resourceClass, string $operationName, array $co
         $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
 
         $isGranted = $resourceMetadata->getGraphqlAttribute($operationName, 'security_post_denormalize', null, true);
+
+        if (null === $this->resourceAccessChecker && $isGranted) {
```
teohhanhui:
```suggestion
        if (null !== $isGranted && null === $this->resourceAccessChecker) {
```

src/GraphQl/Resolver/Stage/SecurityStage.php
teohhanhui:
```suggestion
        if (null !== $isGranted && null === $this->resourceAccessChecker) {
```

tests/GraphQl/Resolver/Stage/SecurityStageTest.php
reypm: That was maybe my IDE setup for CS, I will fix it in the next commit :)

src/GraphQl/Resolver/Stage/SecurityStage.php
reypm: This change has been applied, thanks

tests/GraphQl/Resolver/Stage/SecurityPostDenormalizeStageTest.php
reypm: Done.

src/GraphQl/Resolver/Stage/SecurityPostDenormalizeStage.php
reypm: Fixed with @teohhanhui suggestions.

alanpoulain:
I've rebased your PR for 2.5 and made some little changes. Thank you @reypm!

PR #3151, opened 2019-10-04
Global resource defaults implementation

bpolaszek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3099
| License       | MIT

This PR introduces a `defaults` configuration as a new implementation of #3099 (#3135 was dropped).

This one doesn't rely on a low-priority `DefaultsResourceMetadataFactory` as suggested first (because low priority means late access to some normalization logic) and implements the _defaults_ pattern in `AnnotationResourceMetadataFactory` and `ExtractorResourceMetadataFactory`, making configuration more consistent, which means support for short-hand config like this:

```yaml
api_platform:
    defaults:
        itemOperations: ["get"]
        collectionOperations: ["get"]
        mercure: true
```
Default configuration is now flattened at the bundle configuration level, removing the `attributes` sublevel to improve DX and normalizing all keys to snake_case to follow Symfony standards (both camelCase and snake_case are actually supported to be less error-prone, attributes being mostly documented as annotations):
```yaml
api_platform:
    defaults:
        description:          ~
        iri:                  ~
        item_operations:      ~
        collection_operations: ~
        graphql:              ~
        security:             ~
        security_message:     ~
        security_post_denormalize: ~
        security_post_denormalize_message: ~
        cache_headers:        ~
        denormalization_context: ~
        deprecation_reason:   ~
        elasticsearch:        ~
        fetch_partial:        ~
        force_eager:          ~
        formats:              ~
        filters:              ~
        hydra_context:        ~
        mercure:              ~
        messenger:            ~
        normalization_context: ~
        order:                ~
        pagination_client_enabled: ~
        pagination_client_items_per_page: ~
        pagination_client_partial: ~
        pagination_enabled:   ~
        pagination_fetch_join_collection: ~
        pagination_items_per_page: ~
        maximum_items_per_page: ~
        pagination_maximum_items_per_page: ~
        pagination_partial:   ~
        pagination_via_cursor: ~
        route_prefix:         ~
        swagger_context:      ~
        validation_groups:    ~
        sunset:               ~
        input:                ~
        output:               ~
        openapi_context:      ~
```


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +192,42 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         if ($config['name_converter']) {
             $container->setAlias('api_platform.name_converter', $config['name_converter']);
         }
+        $container->setParameter('api_platform.defaults', $this->normalizeDefaults($config['defaults'] ?? []));
+    }
+
+    private function normalizeDefaults(array $defaults): array
+    {
+        $normalizedDefaults = ['attributes' => []];
+        $reflClass = new \ReflectionClass(ApiResource::class);
```
alanpoulain:
```suggestion
        $$reflectionClass = new \ReflectionClass(ApiResource::class);
```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +192,42 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
...
+
+        foreach ($defaults as $key => $value) {
```
alanpoulain: Please use more representative names than `key`/`value`.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +192,42 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
...
+            // camelCase allows us comparing to @ApiResource properties
+            $camelCased = lcfirst(
```
alanpoulain: Shouldn't we use a name converter instead?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +192,42 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
...
+                    '/(^|_|\.)+(.)/',
+                    static function ($match) {
+                        return ('.' === $match[1] ? '_' : '').strtoupper($match[2]);
+                    },
+                    $key
+                )
+            );
+
+            // snake_case is the reference for all attribute keys
+            $snakeCased = strtolower(preg_replace('/[A-Z]/', '_\\0', $key));
```
alanpoulain: Same.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +192,42 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
...
+
+            try {
+                $reflProperty = $reflClass->getProperty($camelCased);
```
alanpoulain: Better name?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -494,4 +497,25 @@ private function addFormatSection(ArrayNodeDefinition $rootNode, string $key, ar
...
+            'attributes',
+            'subresourceOperations',
+        ];
+
+        foreach ($properties as $property) {
+            $optionName = $property->getName();
+            if (\in_array($optionName, $ignoredOptions, true)) { // "attributes" is a sub-level and not a config option
+                continue;
+            }
+            $snakeCased = strtolower(preg_replace('/[A-Z]/', '_\\0', $optionName));
```
alanpoulain: We should use a name converter IMO.

src/Metadata/Resource/Factory/AnnotationResourceMetadataFactory.php
```diff
@@ -27,11 +27,16 @@ final class AnnotationResourceMetadataFactory implements ResourceMetadataFactory
 {
     private $reader;
     private $decorated;
+    /**
+     * @var array
```
alanpoulain: Needed?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
soyuka:
```suggestion
                $reflectionProperty = $reflectionClass->getProperty($camelCased);
```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +192,42 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
...
+                if ($reflProperty->isPublic()) {
```
soyuka:
```suggestion
                if ($reflectionProperty->isPublic()) {
```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +193,34 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         if ($config['name_converter']) {
             $container->setAlias('api_platform.name_converter', $config['name_converter']);
         }
+        $container->setParameter('api_platform.defaults', $this->normalizeDefaults($config['defaults'] ?? []));
+    }
+
+    private function normalizeDefaults(array $defaults): array
+    {
+        $normalizedDefaults = ['attributes' => []];
+        $reflectionClass = new \ReflectionClass(ApiResource::class);
```
dunglas: Maybe could we just hardcore the list of the 7 public properties instead of using the reflection and catching exceptions (which is known to be slow).

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -191,6 +193,34 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         if ($config['name_converter']) {
             $container->setAlias('api_platform.name_converter', $config['name_converter']);
         }
+        $container->setParameter('api_platform.defaults', $this->normalizeDefaults($config['defaults'] ?? []));
+    }
+
+    private function normalizeDefaults(array $defaults): array
```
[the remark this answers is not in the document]
bpolaszek: I didn't know that trick, but it looks like it is intended to fix XML's kebab-case. Currently we want camelCase / snake_case normalization (which means supporting both and outputting to snake_case). Any idea?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
bpolaszek: Since it's executed at compile time I thought it was no big deal (and I'm not a big fan of hardcoding things) but I'm not against it.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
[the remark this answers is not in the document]
dunglas: You can write your custom logic in `beforeNormalization`

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
bpolaszek: Wait, I actually need to fetch all properties from `ApiResource::class`, dynamically, which means Reflection is not an option, right? Or do you mean only refactoring the `$reflectionProperty->isPublic()` part?

dunglas: for instance you could add a `public static array` marked as `@internal` in `ApiResource` containing the list. It will also simplify the code.

dunglas:
Thanks @bpolaszek!

Issue #3082, opened 2019-09-21
Messenger component does not work with PUT, DELETE, GET


mxkh:
Hi! I want to use Message component with PUT\PATCH, DELETE and GET but as I understand the message component work only with POST method, am I right? For example

```
/**
 * @ApiResource(
 *     shortName="Club",
 *     routePrefix="crm",
 *     messenger="input",
 *     collectionOperations={
 *          "get",
 *          "post"={"path"="/clubs", "input"=CreateClubCommand::class},
 *     },
 *     itemOperations={
 *          "get"={"path"="/clubs/{id}", "input"=FindClubByNameQueryItem::class},
 *          "put"={"path"="/clubs/{id}", "input"=UpdateClubCommand::class}
 *     },
 *     normalizationContext={"groups"={"club_item", "club_list"}},
 *     denormalizationContext={"groups"={"club_write"}}
 * )
 */
final class ClubResource
```

In this case, I expect that all input DTO's (commands\queries) will be passed to their handlers as in POST case. Is it possible to do without custom code? How do you see it's better to use message component for other request methods e.g. GET, PUT, DELETE to implement CQS or CQRS?

soyuka:
The fact is that `input` data is given by the user when creating the request, because "read" routes are not denormalizing data they won't work with input classes, indeed they're queries. I don't really understand the need to use messenger for read routes, or anything that does not need asynchronous processing.

mxkh:
@soyuka Hi! Thank you for your response.
I tested today POST and PUT again and noticed that PUT works well with message component, I guess previously it was a cache problem and in my case PUT Command did not pass to the handler.

I agree with you at some point about GET operations, but what about DELETE. DELETE is a mutator and in my opinion, should work in the same way as other mutators (POST or PUT).

dunglas:
`DELETE` is already supported. However, it would be great to be able to detect in the handler what was the original HTTP verb.
My suggestion is to create a new `ContextStamp` containing the content of the `$context` parameter of `DataPersister::persist()` and `DataPersister::remove()`. As already done for remove: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Messenger/DataPersister.php#L94

It would allow to access the whole context (including the HTTP verb used), and I'm sure that it will be useful to many!

soyuka:
Isn't yet fixed see https://github.com/api-platform/core/pull/3157#issuecomment-543274766

toby-griffiths:
[...]

PR #3157, opened 2019-10-06
Context stamp

sergepavle:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3082 
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Messenger/ContextStamp.php
+++ b/src/Bridge/Symfony/Messenger/ContextStamp.php
@@ -24,4 +24,21 @@ use Symfony\Component\Messenger\Stamp\StampInterface;
  */
 final class ContextStamp implements StampInterface
 {
+    /**
+     * @var array
+     */
+    private $context;
+
+    public function __construct($context = [])
+    {
+        $this->context = $context;
+    }
+
+    /**
+     * Get the context related to a message.
+     */
+    public function getContext(): array
+    {
+        return $this->context;
+    }
 }
--- a/src/Bridge/Symfony/Messenger/ContextStamp.php
+++ b/src/Bridge/Symfony/Messenger/ContextStamp.php
@@ -24,12 +24,9 @@ use Symfony\Component\Messenger\Stamp\StampInterface;
  */
 final class ContextStamp implements StampInterface
 {
-    /**
-     * @var array
-     */
     private $context;
 
-    public function __construct($context = [])
+    public function __construct(array $context = [])
     {
         $this->context = $context;
     }
--- a/src/Bridge/Symfony/Messenger/ContextStamp.php
+++ b/src/Bridge/Symfony/Messenger/ContextStamp.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Messenger;
+
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
+/**
+ * An envelope stamp with context which related to a message.
+ *
+ * @experimental
+ *
+ * @author Sergii Pavlenko <sergii.pavlenko.v@gmail.com>
+ */
+final class ContextStamp implements StampInterface
+{
+}
--- a/tests/Bridge/Symfony/Messenger/ContextStampTest.php
+++ b/tests/Bridge/Symfony/Messenger/ContextStampTest.php
@@ -3,7 +3,7 @@
 /*
  * This file is part of the API Platform project.
  *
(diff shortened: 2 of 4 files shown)


src/Bridge/Symfony/Messenger/ContextStamp.php
```diff
@@ -0,0 +1,44 @@
...
+ * @author Sergii Pavlenko <sergii.pavlenko.v@gmail.com>
+ */
+final class ContextStamp implements StampInterface
+{
+    /**
+     * @var array
+     */
+    private $context;
+
+    public function __construct($context = [])
```
dunglas:
```suggestion
    public function __construct(array $context = [])
```

src/Bridge/Symfony/Messenger/ContextStamp.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @experimental
+ *
+ * @author Sergii Pavlenko <sergii.pavlenko.v@gmail.com>
+ */
+final class ContextStamp implements StampInterface
+{
+    /**
+     * @var array
+     */
```
dunglas: Should be removed (inferred from the constructor's typehint)

tests/Bridge/Symfony/Messenger/ContextStampTest.php
```diff
@@ -0,0 +1,35 @@
...
+class ContextStampTest extends TestCase
+{
+    public function testConstruct()
+    {
+        $this->assertInstanceOf(StampInterface::class, new ContextStamp());
+    }
+
+    public function testGetContext()
+    {
+        $context_stamp = new ContextStamp();
```
dunglas:
```suggestion
        $contextStamp = new ContextStamp();
```

(PSR)

tests/Bridge/Symfony/Messenger/ContextStampTest.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->assertIsArray($context_stamp->getContext());
```
dunglas:
```suggestion
        $this->assertIsArray($contextStamp->getContext());
```

tests/Bridge/Symfony/Messenger/DataPersisterTest.php
```diff
@@ -56,7 +57,9 @@ public function testPersist()
         $dummy = new Dummy();
 
         $messageBus = $this->prophesize(MessageBusInterface::class);
-        $messageBus->dispatch($dummy)->willReturn(new Envelope($dummy))->shouldBeCalled();
+        $messageBus->dispatch(Argument::that(function (Envelope $envelope) use ($dummy) {
+            return $dummy === $envelope->getMessage() && null !== $envelope->last(ContextStamp::class);
```
dunglas: Why is this change necessary?

src/Bridge/Symfony/Messenger/ContextStamp.php
sergepavle: fixed

src/Bridge/Symfony/Messenger/ContextStamp.php
sergepavle: fixed

tests/Bridge/Symfony/Messenger/ContextStampTest.php
sergepavle: done

tests/Bridge/Symfony/Messenger/ContextStampTest.php
sergepavle: Fixed

tests/Bridge/Symfony/Messenger/DataPersisterTest.php
sergepavle: This changes consistent with code from testRemove().
By this changes we make sure that have envelop with ContextStamp.

dunglas:
Thanks @sergepavle! Could you also open a docs PR? At least, adding an example of how to access the context using a middleware would be nice!

teohhanhui:
This seems to make sense at first glance, but actually it doesn't work. You're not supposed to receive an `Envelope` in your message handler, e.g.:

```php
<?php

declare(strict_types=1);

namespace App\MessageHandler\Api\Subscription;

use App\Model\Api\Subscription\PaymentDetails;
use Symfony\Component\Messenger\Envelope;
use Symfony\Component\Messenger\Handler\MessageHandlerInterface;

final class PaymentDetailsHandler implements MessageHandlerInterface
{
    // wtf?
    // public function __invoke(Envelope $envelope): void
    public function __invoke(PaymentDetails $paymentDetails): void
    {
        // ...
    }
}
```

dunglas:
You need to create a middleware. It’s why we must add docs!

teohhanhui:
Creating a middleware still would not help the stated use case at all. :smile:

entermix:
@dunglas, what about this use case? This is similar to what @teohhanhui writes about.

https://github.com/api-platform/core/issues/3082#issuecomment-758328003

Issue #3049, opened 2019-09-05
Creating custom filters without constuctor throw an error


dunglas:
It's not possible to create a filter without a constructor with and a `array $properties` parameters, because of how we wire the class passed as parameter of `@ApiFilter`, we should handle this case, and allow to create custom filters without constructor.

Also when using the default Dependency Injection configuration of Symfony 4, classes passed as parameters of `@ApiFilter` are registered two times, one time by Symfony's autowiring, one time by our compiler pass. If there are no constructors, it's useless and we should prevent this.

soyuka:
Code link: https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php

LouisPinsard:
Hello @dunglas @soyuka 

I would love to help on that matter.
I am not sure of what you mean, do you target to be able to write something like that ?
```
**
* @ApiFilter(CustomFilter::class)
*/
class DummyEntityFilterAnnotated
```

What is the use case for such a filter ?
Thanks !

soyuka:
Filters do change the query so it could add whatever :). 

Doing this should be active on every property of the class I guess. The issue here is more when you have a filter without constructor it still adds arguments (https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php#L80) and obviously fails.

LouisPinsard:
@soyuka Ok thanks I'll have a look at it this week

PR #3166, opened 2019-10-11
Fix support for custom filters without constructor in ApiFilter annotation

LouisPinsard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |  #3049 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
 
+use ApiPlatform\Core\Annotation\ApiFilter;
 use ApiPlatform\Core\Util\AnnotationFilterExtractorTrait;
 use ApiPlatform\Core\Util\ReflectionClassRecursiveIterator;
 use Doctrine\Common\Annotations\Reader;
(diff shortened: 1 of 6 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/AnnotationFilterPass.php
```diff
@@ -77,6 +78,9 @@ private function createFilterDefinitions(\ReflectionClass $reflectionClass, Cont
             $definition->setAutowired(true);
 
             foreach ($arguments as $key => $value) {
+                if (null === $reflectionConstructor = $reflectionClass->getConstructor()) {
```
teohhanhui: `$reflectionClass` could be `null` here. Also, this check should be done before the `foreach` loop.

tests/Fixtures/DummyFilterWithoutConstructor.php
```diff
@@ -0,0 +1,42 @@
...
+namespace ApiPlatform\Core\Tests\Fixtures;
+
+use ApiPlatform\Core\Api\FilterInterface;
+
+/**
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class DummyFilterWithoutConstructor implements FilterInterface
+{
+    /**
```
teohhanhui: Do not copy the phpdoc here. Just use:
```php
    /**
     * {@inheritdoc}
     */
```

LouisPinsard:
Hello @teohhanhui I modified the PR with your return if you can take a look !

soyuka:
thanks @teohhanhui @LouisPinsard !

PR #3167, opened 2019-10-11
DI: Improve swagger versions parameter (fix #3115)

### LINKED ISSUE #3115: [2.5] Problem with the new swagger version configuration
Dear guys, by default when swagger is not enabled, the version are set to empty

https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L59-L68

Therefore the `json_schema.xml` is not loaded and the `api_platform.swagger.versions` parameter is not set at all.

https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L315-L328

The thing is that it will throw in the ContainerBuilder `You have requested a non-existent service "api_platform.json_schema.schema_factory".` 

That because the service is decorated in `hydra.xml` :

https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/Bridge/Symfony/Bundle/Resources/config/hydra.xml#L81-L83

In the same way, the `api_platform.action.documentation` is requiring the `api_platform.swagger.versions` parameter (which hasn't been defined) leads to the following error : `The service "api_platform.action.documentation" has a dependency on a non-existent parameter "api_platform.swagger.versions". Did you mean this: "api_platform.version` 

https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/Bridge/Symfony/Bundle/Resources/config/api.xml#L245
The `on-invalid="null"` doesn't seems to have any impact...

# Proposed solution
Set the `versions` to the default array [2, 3] even when swagger is not enabled because obviously it is used anyway. 

I can do the PR off course but I just wanted to make sure/understand the reason to empty this values in every cases...

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3115
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -312,6 +312,8 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
      */
     private function registerSwaggerConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader): void
     {
+        $container->setParameter('api_platform.swagger.versions', $config['swagger']['versions']);
+
         if (empty($config['swagger']['versions'])) {
             return;
         }
@@ -325,7 +327,6 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             $container->setParameter('api_platform.enable_re_doc', $config['enable_re_doc']);
         }
 
-        $container->setParameter('api_platform.swagger.versions', $config['swagger']['versions']);
         $container->setParameter('api_platform.swagger.api_keys', $config['swagger']['api_keys']);
     }
 
@@ -346,6 +347,10 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
         $loader->load('jsonld.xml');
         $loader->load('hydra.xml');
+
+        if (!$container->has('api_platform.json_schema.schema_factory')) {
+            $container->removeDefinition('api_platform.hydra.json_schema.schema_factory');
+        }
 
         if (!$docEnabled) {
             $container->removeDefinition('api_platform.hydra.listener.response.add_link_header');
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
```diff
@@ -28,6 +28,11 @@
             <argument>%api_platform.formats%</argument>
             <tag name="console.command" />
         </service>
+
+        <service id="api_platform.json_schema.hydra.schema_factory" class="ApiPlatform\Core\Hydra\JsonSchema\SchemaFactory" decorates="api_platform.json_schema.schema_factory">
```
teohhanhui: As previously discussed, this is incorrect. Can't we instead do:

```php
if (!$container->has('api_platform.json_schema.schema_factory')) {
    $container->removeDefinition('api_platform.json_schema.hydra.schema_factory');
}
```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -313,6 +313,8 @@ private function registerOAuthConfiguration(ContainerBuilder $container, array $
     private function registerSwaggerConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader): void
     {
         if (empty($config['swagger']['versions'])) {
```
dunglas: `$container->setParameter('api_platform.swagger.versions', $config['swagger']['versions']);` could be added just before this line and removed everywhere else.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -344,6 +345,10 @@ private function registerJsonLdHydraConfiguration(ContainerBuilder $container, a
             return;
         }
 
+        if (!$container->has('api_platform.json_schema.schema_factory')) {
+            $container->removeDefinition('api_platform.hydra.json_schema.schema_factory');
```
teohhanhui: You cannot remove a definition before it's loaded (below)...

```php
$loader->load('hydra.xml');
```

PR #3182, opened 2019-10-16
Bugfix: Wrong parameter type

thomasumme:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

When ItemDataProviderInterface is used with custom Denormalizers, the id parameter can be pretty much everything.

I am really struggling here with phpstan.


teohhanhui:
> When ItemDataProviderInterface is used with custom Denormalizers, the id parameter can be pretty much everything.

I'm not sold on this. We're restricting the types in our interface. Your custom item data provider must adhere to these types for the `$id`.

thomasumme:
@teohhanhui 

My idea was based on [API Platforms documentation about Identifiers](https://api-platform.com/docs/core/identifiers/).  There it says 

> Although this identifier is usually a number, it can also be an UUID, a date, or the type of your choice.

Followed by an example of a Normalizer and a DataProvider using a UUID object.

How can I use the type of my choice if the Data Providers interface has these restrictions?

Or is there an error in the documentation?

antograssiot:
you could use [DenormalizedIdentifiersAwareItemDataProviderInterface](https://github.com/api-platform/core/blob/master/src/DataProvider/DenormalizedIdentifiersAwareItemDataProviderInterface.php#L21) and you'll always get an array

soyuka:
thanks for your interest in the project, please see @antograssiot answer which should be the correct fix!

teohhanhui:
I think there's a misunderstanding here. The `$id` you pass to `ItemDataProviderInterface::getItem` must be `array<int|string>|int|string`. It's then your responsibility in the data provider to denormalize it (e.g. into the `Uuid` type). See for example:

https://github.com/api-platform/core/blob/4d0b7776bcd3dc8d05f2b5ced1e244a44abf1724/src/Bridge/Doctrine/Orm/ItemDataProvider.php#L73-L75

teohhanhui:
Okay, my bad. It's a design problem. I guess we've kinda "fixed" that by `DenormalizedIdentifiersAwareItemDataProviderInterface` :laughing:

thomasumme:
Thank you. I was not aware of the existence of `DenormalizedIdentifiersAwareItemDataProviderInterface`

PR #3187, opened 2019-10-17
[2.5] Fix dashed path segment generation

### LINKED ISSUE #3188: DashPathSegmentNameGenerator not working properly
**API Platform version(s) affected**: 2.4.7

**Description**  
The mentioned class applies Doctrine Inflector's `pluralize` on the whole  resource's name while the method expects a single word, resulting in unexpected output.

For example: when you have an `OrderingPerson` entity added as a resource and use `DashPathSegmentNameGenerator`, the result is `ordering-oeople` instead of `ordering-people`

**How to reproduce**  
1. Create an `OrderingPerson` entity
2. Add it as API Platform resource
3. In `api_platform` configuration section add a `path_segment_name_generator: 'api_platform.path_segment_name_generator.dash'` entry
4. Navigate to endpoints listing, see the result

**Possible Solution**  
Refactor `ApiPlatform\Core\Operation\DashPathSegmentNameGenerator` to treat words from `$name` individually when using `Doctrine\Common\Inflector\Inflector::pluralize`

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fix #3188
| License       | MIT
| Doc PR        | 

This will correctly convert `orderingPerson` to `ordering-people`

### DIFF
--- a/tests/Operation/DashedPathSegmentNameGeneratorTest.php
+++ b/tests/Operation/DashedPathSegmentNameGeneratorTest.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Operation;
+
+use ApiPlatform\Core\Operation\DashPathSegmentNameGenerator;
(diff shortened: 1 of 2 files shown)


tests/Operation/DashedPathSegmentNameGeneratorTest.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
soyuka: we had no tests for this class? oO

teohhanhui:
Thanks @antograssiot! :tada:

PR #3191, opened 2019-10-18
Fix denormalization of basic property-types in XML and CSV

warslett:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/api-platform/issues/1299
| License       | MIT

This PR fixes an issue where by posting XML content throws errors when a field has a non string primitive type because the XML encoder can only normalize primitive values as strings. This means that post create and update requests fail. This PR adds some additional behaviour where by the AbstractItemNormalizer will cast valid strings as the correct type.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -9,6 +9,7 @@ default:
         - 'HttpHeaderContext'
         - 'GraphqlContext'
         - 'JsonContext'
+        - 'XmlContext'
         - 'HydraContext'
         - 'SwaggerContext'
         - 'HttpCacheContext'
--- a/features/bootstrap/XmlContext.php
+++ b/features/bootstrap/XmlContext.php
@@ -1,0 +1,8 @@
+<?php
+
+use Behatch\Context\XmlContext as BaseXmlContext;
+
+class XmlContext extends BaseXmlContext
+{
+
+}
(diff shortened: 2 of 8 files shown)


teohhanhui:
This is not the correct approach. The type cast feature should be added to Symfony Serializer. It should be controlled by a flag in the context.

warslett:
@teohhanhui that was my initial thinking aswell. The problem is that the AbstractObjectNormalizer handles this behaviour in the Serializer component and is extended by AbstractItemNormalizer in the API Platform. AbstractObjectNormalizer resolves types using an optional dependency called PropertyTypeExtractor but AbstractItemNormalizer overrides the constructor of AbstractObjectNormalizer and does not inject this dependency. Instead, AbstractItemNormalizer has a dependency on a different service PropertyMetadataFactory which can be used to determine types.

This means that AbstractItemNormalizer used with API Platform has a completely different mechanism for resolving types than AbstractObjectNormalizer when used with vanilla Serializer component. Therefor this problem needs to be solved separately in the Serializer component and the API Platform. This is not ideal but is due to the way inheritance has been used.

teohhanhui:
Yes, I get what you mean, but I think we should fix this in Symfony first. Then we can see how we'll fix it in API Platform. (Probably by calling the same methods from Symfony Serializer.)

alanpoulain:
Superseded by https://github.com/api-platform/core/pull/4145.

alanpoulain:
Thank you @warslett.

PR #3196, opened 2019-10-19
Update ApiGatewayNormalizer.php

### LINKED ISSUE #3193: AWS ApiGateway compatibility
**API Platform version(s) affected**:
`"api-platform/core": "^2.5",`

**Description**  
When I try to import the schema from `http://localhost/api/docs.json?api_gateway=true` to ApiGateway it errors.

**How to reproduce**  
Create an Entity which has a normalizationGroup, get the api_gateway "compatible" json spec and try to import it.

**Possible Solution**  
Entity name has colon in it, which API Gateway doesn't allow.

**Additional Context**  

Error Message from APIGateway importer:

```
Your API was not imported due to errors in the Swagger file.
Unable to create model for 'ConfirmEmailconfirmEmail:write': Model name must be alphanumeric: ConfirmEmailconfirmEmail:write
Unable to create model for 'EmailCheck:b7add851abebff79e75557aab315cdc2emailCheck:read': Model name must be alphanumeric: EmailCheck:b7add851abebff79e75557aab315cdc2emailCheck:read
Unable to create model for 'EmailCheckemailCheck:write': Model name must be alphanumeric: EmailCheckemailCheck:write
Unable to create model for 'GenerateTokengenerateToken:write': Model name must be alphanumeric: GenerateTokengenerateToken:write
Unable to create model for 'RefreshTokenrefreshToken:write': Model name must be alphanumeric: RefreshTokenrefreshToken:write
Unable to create model for 'Register:5b4fa5b4800265236a947e59c5cd3661register:read': Model name must be alphanumeric: Register:5b4fa5b4800265236a947e59c5cd3661register:read
Unable to create model for 'Registerregister:write': Model name must be alphanumeric: Registerregister:write
Unable to create model for 'Userregister:read': Model name must be alphanumeric: Userregister:read
Unable to create model for 'Useruser:read': Model name must be alphanumeric: Useruser:read
Unable to put method 'POST' on resource at path '/api/confirm_email': Invalid model identifier specified: null
Unable to put method 'POST' on resource at path '/api/email_check': Invalid model identifier specified: null
Unable to put method 'POST' on resource at path '/api/register': Invalid model identifier specified: null
Unable to put method 'PUT' on resource at path '/api/users/{id}/change_email': Invalid model identifier specified: null
Unable to put method 'PUT' on resource at path '/api/users/{id}/change_password': Invalid model identifier specified: null
Unable to put method 'POST' on resource at path '/token/generate': Invalid model identifier specified: null
Unable to put method 'POST' on resource at path '/token/refresh': Invalid model identifier specified: null
Additionally, these warnings were found:
API Key security definition 'apiKey' has unexpected name or location. Ignoring.
Unknown or invalid reference specified by '#/definitions/ConfirmEmailconfirmEmail:write'. Ignoring.
Unknown or invalid reference specified by '#/definitions/EmailCheckemailCheck:write'. Ignoring.
Unknown or invalid reference specified by '#/definitions/Registerregister:write'. Ignoring.
Unknown or invalid reference specified by '#/definitions/RequestPasswordReset:7f546cb601bbd9e0eb1a7cfa305ea7fb'. Ignoring.
Invalid format for model application/json for method POST
Unknown or invalid reference specified by '#/definitions/Useruser:read'. Ignoring.
Unable to create model for 200 response to method 'GET /api/users': Invalid model specified: Validation Result: warnings : [], errors : [Invalid model specified: Useruser:read]
Unknown or invalid reference specified by '#/definitions/Useruser:read'. Ignoring.
Invalid format for model application/json for method GET
Unknown or invalid reference specified by '#/definitions/ChangeEmail:5bab9375ca064bc028c7d4f29f2eaf2f'. Ignoring.
Unknown or invalid reference specified by '#/definitions/ChangePassword:b12382e7dd54bc0f641b9ded532416fc'. Ignoring.
Unknown or invalid reference specified by '#/definitions/GenerateTokengenerateToken:write'. Ignoring.
Unknown or invalid reference specified by '#/definitions/RefreshTokenrefreshToken:write'. Ignoring.
```

MLDMoritz:
Resolves #3193 an issue with Model names containing a colon and security definitions

| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3193 
| License       | MIT

### DIFF
--- a/src/Swagger/Serializer/ApiGatewayNormalizer.php
+++ b/src/Swagger/Serializer/ApiGatewayNormalizer.php
@@ -31,7 +31,9 @@ final class ApiGatewayNormalizer implements NormalizerInterface, CacheableSuppor
     public const API_GATEWAY = 'api_gateway';
 
     private $documentationNormalizer;
-    private $defaultContext = [self::API_GATEWAY => false];
+    private $defaultContext = [
+        self::API_GATEWAY => false,
+    ];
 
     public function __construct(NormalizerInterface $documentationNormalizer, $defaultContext = [])
     {
@@ -66,19 +68,19 @@ final class ApiGatewayNormalizer implements NormalizerInterface, CacheableSuppor
                         if (!preg_match('/^[a-zA-Z0-9._$-]+$/', $parameter['name'])) {
                             unset($data['paths'][$path][$operation]['parameters'][$key]);
                         }
-                        if (isset($parameter['schema']['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $parameter['schema']['$ref'])) {
-                            $data['paths'][$path][$operation]['parameters'][$key]['schema']['$ref'] = str_replace(['-', '_'], '', $parameter['schema']['$ref']);
+                        if (isset($parameter['schema']['$ref']) && $this->isLocalRef($parameter['schema']['$ref'])) {
+                            $data['paths'][$path][$operation]['parameters'][$key]['schema']['$ref'] = $this->normalizeRef($parameter['schema']['$ref']);
                         }
                     }
                     $data['paths'][$path][$operation]['parameters'] = array_values($data['paths'][$path][$operation]['parameters']);
                 }
                 if (isset($options['responses'])) {
                     foreach ($options['responses'] as $statusCode => $response) {
-                        if (isset($response['schema']['items']['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $response['schema']['items']['$ref'])) {
-                            $data['paths'][$path][$operation]['responses'][$statusCode]['schema']['items']['$ref'] = str_replace(['-', '_'], '', $response['schema']['items']['$ref']);
+                        if (isset($response['schema']['items']['$ref']) && $this->isLocalRef($response['schema']['items']['$ref'])) {
+                            $data['paths'][$path][$operation]['responses'][$statusCode]['schema']['items']['$ref'] = $this->normalizeRef($response['schema']['items']['$ref']);
                         }
-                        if (isset($response['schema']['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $response['schema']['$ref'])) {
-                            $data['paths'][$path][$operation]['responses'][$statusCode]['schema']['$ref'] = str_replace(['-', '_'], '', $response['schema']['$ref']);
+                        if (isset($response['schema']['$ref']) && $this->isLocalRef($response['schema']['$ref'])) {
+                            $data['paths'][$path][$operation]['responses'][$statusCode]['schema']['$ref'] = $this->normalizeRef($response['schema']['$ref']);
                         }
                     }
                 }
@@ -93,19 +95,19 @@ final class ApiGatewayNormalizer implements NormalizerInterface, CacheableSuppor
                 if (isset($propertyOptions['readOnly'])) {
                     unset($data['definitions'][$definition]['properties'][$property]['readOnly']);
                 }
-                if (isset($propertyOptions['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $propertyOptions['$ref'])) {
-                    $data['definitions'][$definition]['properties'][$property]['$ref'] = str_replace(['-', '_'], '', $propertyOptions['$ref']);
+                if (isset($propertyOptions['$ref']) && $this->isLocalRef($propertyOptions['$ref'])) {
+                    $data['definitions'][$definition]['properties'][$property]['$ref'] = $this->normalizeRef($propertyOptions['$ref']);
                 }
-                if (isset($propertyOptions['items']['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $propertyOptions['items']['$ref'])) {
-                    $data['definitions'][$definition]['properties'][$property]['items']['$ref'] = str_replace(['-', '_'], '', $propertyOptions['items']['$ref']);
+                if (isset($propertyOptions['items']['$ref']) && $this->isLocalRef($propertyOptions['items']['$ref'])) {
+                    $data['definitions'][$definition]['properties'][$property]['items']['$ref'] = $this->normalizeRef($propertyOptions['items']['$ref']);
                 }
             }
         }
 
         // $data['definitions'] is an instance of \ArrayObject
         foreach (array_keys($data['definitions']->getArrayCopy()) as $definition) {
-            if (!preg_match('/^[A-z]+$/', (string) $definition)) {
-                $data['definitions'][str_replace(['-', '_'], '', (string) $definition)] = $data['definitions'][$definition];
+            if (!preg_match('/^[0-9A-Za-z]+$/', (string) $definition)) {
+                $data['definitions'][preg_replace('/[^0-9A-Za-z]/', '', (string) $definition)] = $data['definitions'][$definition];
                 unset($data['definitions'][$definition]);
             }
         }
@@ -127,5 +129,21 @@ final class ApiGatewayNormalizer implements NormalizerInterface, CacheableSuppor
     public function hasCacheableSupportsMethod(): bool
     {
         return $this->documentationNormalizer instanceof CacheableSupportsMethodInterface && $this->documentationNormalizer->hasCacheableSupportsMethod();
+    }
+
+    private function isLocalRef(string $ref): bool
+    {
+        return '#/' === substr($ref, 0, 2);
+    }
+
+    private function normalizeRef(string $ref): string
+    {
+        $refParts = explode('/', $ref);
+
+        $schemaName = array_pop($refParts);
+        $schemaName = preg_replace('/[^0-9A-Za-z]/', '', $schemaName);
+        $refParts[] = $schemaName;
+
+        return implode('/', $refParts);
     }
 }
--- a/tests/Swagger/Serializer/ApiGatewayNormalizerTest.php
+++ b/tests/Swagger/Serializer/ApiGatewayNormalizerTest.php
@@ -13,220 +13,533 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Tests\Swagger\Serializer;
 
+use ApiPlatform\Core\Documentation\Documentation;
+use ApiPlatform\Core\Metadata\Resource\ResourceNameCollection;
 use ApiPlatform\Core\Swagger\Serializer\ApiGatewayNormalizer;
+use ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy;
 use PHPUnit\Framework\TestCase;
-use Prophecy\Prophecy\ObjectProphecy;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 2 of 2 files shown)


src/Swagger/Serializer/ApiGatewayNormalizer.php
```diff
@@ -67,18 +70,18 @@ public function normalize($object, $format = null, array $context = [])
                             unset($data['paths'][$path][$operation]['parameters'][$key]);
                         }
                         if (isset($parameter['schema']['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $parameter['schema']['$ref'])) {
-                            $data['paths'][$path][$operation]['parameters'][$key]['schema']['$ref'] = str_replace(['-', '_'], '', $parameter['schema']['$ref']);
+                            $data['paths'][$path][$operation]['parameters'][$key]['schema']['$ref'] = str_replace(['-', ':', '_'], '', $parameter['schema']['$ref']);
```
teohhanhui: We should not hardcode the characters to replace. Just use `preg_replace` with an appropriate pattern.

src/Swagger/Serializer/ApiGatewayNormalizer.php
```diff
@@ -59,6 +59,9 @@ public function normalize($object, $format = null, array $context = [])
             $data['basePath'] = '/';
         }
 
+        unset($data['security']);
```
teohhanhui: Why these changes?

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: I'm totally with you, but it wouldn't fit here as Model names can only contain alphanumeric characters. However, this affects the whole $ref string and not just the Model name.

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: Indeed, that shouldn't be there.

MLDMoritz: Indeed, that shouldn't be there.

src/Swagger/Serializer/ApiGatewayNormalizer.php
teohhanhui: What I mean is, we remove all non-alphanumeric characters using `preg_replace`. :smile:

MLDMoritz: I understand what you mean, but removing all non-alpha chars in the $ref string would strip out needed characters (like #, /). I actually would prefer to have a Normalizer for the Model name itself, rather than iterating through all ref strings (with the chance of missing some) (would probably fix DTO name generation with custom input too (the hash after the colon)).

src/Swagger/Serializer/ApiGatewayNormalizer.php
```diff
@@ -59,6 +59,8 @@ public function normalize($object, $format = null, array $context = [])
             $data['basePath'] = '/';
         }
 
+        unset($data['securityDefinitions']);
```
teohhanhui: Same question: why is this here?

src/Swagger/Serializer/ApiGatewayNormalizer.php
teohhanhui: Okay, I get what you mean now...

teohhanhui: But actually, how do we know that these characters aren't used in other parts of the `$ref`? After all, `$ref` can be a URI, right?

MLDMoritz: $ref can be an URI, right. I'm not quite sure what you mean, I think a good solution would be to provide a function that can modify the Model names which the normalizer uses.

MLDMoritz: I just saw #3198 - This could be a step in the right direction, what do you think?

teohhanhui: If `$ref` could be any URI, and not just the ones we generate, then it's unsafe to perform any kind of search and replace. :smile:

On the other hand, I don't see any relevance between #3198 and this issue.

MLDMoritz: I see the problem, however in it's current version the ApiGatewayNormalizer is non-working, so this commit would at least allow users to use this feature (which is also referenced in the docs).

teohhanhui: Can we check if the `$ref` value starts with `#/` before doing this replacement? :smile:

src/Swagger/Serializer/ApiGatewayNormalizer.php
```diff
@@ -94,18 +96,18 @@ public function normalize($object, $format = null, array $context = [])
...
+                    $data['definitions'][$definition]['properties'][$property]['items']['$ref'] = str_replace(['-', ':', '_'], '', $propertyOptions['items']['$ref']);
                 }
             }
         }
 
         // $data['definitions'] is an instance of \ArrayObject
         foreach (array_keys($data['definitions']->getArrayCopy()) as $definition) {
             if (!preg_match('/^[A-z]+$/', (string) $definition)) {
-                $data['definitions'][str_replace(['-', '_'], '', (string) $definition)] = $data['definitions'][$definition];
+                $data['definitions'][str_replace(['-', ':', '_'], '', (string) $definition)] = $data['definitions'][$definition];
```
teohhanhui: In this case, it's safe to use `preg_replace` to get rid of all non-alphanumeric characters, yes?

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: The way securityDefinitions is described is not compatible with AWS. `API Key security definition 'apiKey' has unexpected name or location. Ignoring.`

teohhanhui: @MLDMoritz Can you explain this change please? :smile:

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: On first sight, probably yes.

teohhanhui: This is the model name, right? So API Gateway does not allow any non-alphanumeric character here.

MLDMoritz: `$data['definitions'][preg_replace('/[^\da-z]/i', '', (string) $definition)] = $data['definitions'][$definition];` this works.

teohhanhui: `a-z` is wrong. It should be `A-z`

teohhanhui: Oh, `/i` flag, my bad.

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: I could do a check if it starts with #/, then I would strip everything until the last / and pass the model name to a custom function. Sounds good?

src/Swagger/Serializer/ApiGatewayNormalizer.php
```diff
@@ -94,18 +96,18 @@ public function normalize($object, $format = null, array $context = [])
...
                 if (isset($propertyOptions['items']['$ref']) && !preg_match('/^#\/definitions\/[A-z]+$/', $propertyOptions['items']['$ref'])) {
-                    $data['definitions'][$definition]['properties'][$property]['items']['$ref'] = str_replace(['-', '_'], '', $propertyOptions['items']['$ref']);
+                    $data['definitions'][$definition]['properties'][$property]['items']['$ref'] = str_replace(['-', ':', '_'], '', $propertyOptions['items']['$ref']);
                 }
             }
         }
 
         // $data['definitions'] is an instance of \ArrayObject
         foreach (array_keys($data['definitions']->getArrayCopy()) as $definition) {
             if (!preg_match('/^[A-z]+$/', (string) $definition)) {
```
teohhanhui: This pattern should also be modified to match: `/^[\dA-z]+$/`

src/Swagger/Serializer/ApiGatewayNormalizer.php
```diff
@@ -94,18 +96,18 @@ public function normalize($object, $format = null, array $context = [])
...
-                $data['definitions'][str_replace(['-', '_'], '', (string) $definition)] = $data['definitions'][$definition];
+                $data['definitions'][preg_replace('/[^\da-z]/i', '', (string) $definition)] = $data['definitions'][$definition];
```
teohhanhui: Anyway, let's just use: `/[^\dA-z]/`

src/Swagger/Serializer/ApiGatewayNormalizer.php
teohhanhui: It's just a warning, so let's revert this change since it's not a blocker.

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: $ref starts with #/ which would need the check you mentioned above.

teohhanhui: This is not a `$ref`. This is the definition / model name.

src/Swagger/Serializer/ApiGatewayNormalizer.php
teohhanhui: Never mind, I've just noticed we already have a `preg_match` above. We just need to update the pattern to allow alphanumeric for the definition name, instead of only alpha.

teohhanhui: Since we're already doing a `preg_match`, we can get the matches from there! So, no separate function.

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: AWS is not completely compliant to swagger spec, the securityDefinition must be extended to this to be imported without errors:

```"securityDefinitions": {
        "apiKey": {
            "type": "apiKey",
            "in": "header",
            "description": "Value for the Authorization header",
            "name": "Authorization",
            "x-amazon-apigateway-authtype": "awsSigv4"
        }
    },```

I could just hardcode that, but it won't do anything in the aws console because they implement custom authorizers, so I stripped it, because AWS doesn't make use of this at all.

src/Swagger/Serializer/ApiGatewayNormalizer.php
MLDMoritz: A-z represents char code 65 to 122.

image

teohhanhui: Yeah, my bad, it should be `A-Za-z`.

MLDMoritz:
Codecoverage is affected because of the new method I think, however, I won't write a test for the private function.

teohhanhui:
@MLDMoritz I've cleaned up the code and rewritten the tests. Can you confirm that this works correctly for you?

teohhanhui:
Thanks @MLDMoritz! :tada:

PR #3197, opened 2019-10-20
Move behat tests to CI workflow on GitHub Actions

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Move the Behat tests to the CI workflow on GitHub Actions, so that we can move away from Travis CI completely.

TODO:

- [x] Fix failing Behat test related to GraphQL (previously flaky on CircleCI, but always failing on GitHub Actions :thinking:) (@alanpoulain?)
- [ ] ~Move Swagger / OpenAPI tests to CircleCI?~ Looks like this will complicate things too much...
- [ ] ~See if we can speed up the jobs on CircleCI~ CircleCI doesn't support `tmpfs` for the `docker` executor :disappointed:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
 
 env:
   COMPOSER_ALLOW_SUPERUSER: '1' # https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
+  COMPOSER_INSTALLER_COMMIT: fb22b78362d31c0d2bf516d1f8cdfd2745caa431
+  EXT_MONGODB_VERSION: '1.5.5'
+  LEGACY: '0'
   SYMFONY_REQUIRE: ^3.4 || ^4.0
 
 jobs:
@@ -14,6 +17,8 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: php:${{ matrix.php }}-alpine
+      options: >-
+        --tmpfs /tmp:exec
     strategy:
       matrix:
         php:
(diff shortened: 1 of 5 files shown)


features/graphql/mutation.feature
```diff
@@ -31,14 +31,353 @@ Feature: GraphQL mutation support
...
+            "__type": {
+              "type": "object",
+              "required": [
+                "fields"
+              ],
+              "properties": {
+                "fields": {
+                  "type": "array",
+                  "minItems": 1,
+                  "items": {
```
teohhanhui: This is not ideal, but `justinrainbow/json-schema` doesn't support JSON Schema draft-06 or newer... :disappointed:

.github/workflows/ci.yml
```diff
@@ -14,6 +17,8 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: php:${{ matrix.php }}-alpine
+      options: >-
+        --tmpfs /tmp:exec
```
bendavies: what does this speed up?

teohhanhui: I think the main thing that made the Behat tests really slow was slow disk I/O for sqlite. I'm not sure why...

bendavies: nice

.github/workflows/ci.yml
```diff
@@ -30,37 +35,108 @@ jobs:
...
       - name: Clear test app cache
-        run: tests/Fixtures/app/console cache:clear --ansi
+        run: |
+          mkdir -p /tmp/api-platform/core/var
+          ln -s /tmp/api-platform/core/var tests/Fixtures/app/var
+          tests/Fixtures/app/console cache:clear --ansi
       - name: Run PHPUnit tests
         run: vendor/bin/phpunit --colors=always
 
+  behat:
```
bendavies: remember with github actions that there is no need for all these jobs to go in one file.
you could also split them out to make a bit more readable?

`ci-behat.yml`
```
  behat:
    name: "Behat (PHP ${{ matrix.php }}) (${{ matrix.dependencies }} dependencies)"

    runs-on: ubuntu-latest

    strategy:
      matrix:
        php:
          - 7.1
          - 7.2
          - 7.3
          - 7.4.rc

        dependencies:
          - lowest
          - highest

...
```

teohhanhui: I disagree. It's the same CI workflow.

bendavies: yeah true, the UI would be annoying then with N different workflows.

bendavies: could still combine the behat jobs (and others) with a matrix?
or maybe the `if` exclusions would become hard to manage then

teohhanhui: It'll be ugly. Better to respect SRP. :laughing: (No "phpunit-or-behat" job, for example.)

soyuka:
thanks @teohhanhui nice job!

PR #3198, opened 2019-10-20
Add possibility to specify a url generation strategy

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #3074
| License       | MIT
| Doc PR        | not done yet, waiting approval

I took all the commits of the excellent work of @mab05k in #3084 (and rebasing on master) and added a commit on top of it to propose a similar approach to #2708

I'm still not sure it is better, or if we need to support this instead of the simpler boolean option but at least we can choose.

### DIFF
--- a/features/jsonld/absolute_url.feature
+++ b/features/jsonld/absolute_url.feature
@@ -1,0 +1,83 @@
+Feature: IRI should contain Absolute URL
+  In order to add detail to IRIs
+  Include the absolute url
+
+  @createSchema
+  Scenario: I should be able to GET a collection of Objects with Absolute Urls
+    Given there are 1 absoluteUrlDummy objects with a related absoluteUrlRelationDummy
+    And I add "Accept" header equal to "application/ld+json"
+    And I send a "GET" request to "/absolute_url_dummies"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "http://example.com/contexts/AbsoluteUrlDummy",
+      "@id": "http://example.com/absolute_url_dummies",
+      "@type": "hydra:Collection",
+      "hydra:member": [
+        {
+          "@id": "http://example.com/absolute_url_dummies/1",
+          "@type": "AbsoluteUrlDummy",
+          "absoluteUrlRelationDummy": "http://example.com/absolute_url_relation_dummies/1",
+          "id": 1
+        }
+      ],
+      "hydra:totalItems": 1
+    }
+
+    """
+
+  Scenario: I should be able to POST an object using an Absolute Url
+    Given I add "Accept" header equal to "application/ld+json"
+    And I add "Content-Type" header equal to "application/json"
+    And I send a "POST" request to "/absolute_url_relation_dummies" with body:
+    """
+    {
+      "absolute_url_dummies": "http://example.com/absolute_url_dummies/1"
+    }
+    """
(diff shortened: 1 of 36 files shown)


.circleci/config.yml
```diff
@@ -489,6 +489,48 @@ jobs:
           file: build/logs/clover.xml
       - save-npm-cache
 
+  behat-absolute-url-coverage:
```
teohhanhui: Why do we need a separate job for this?

teohhanhui: If we do this for every feature we'll have a combinatorial explosion.

mab05k: I added that because it is a configuration that changes the output of all API responses (IRIs). From `/api/resource/1` to `http://example.com/api/resource/1`. If you have any other ideas on how I could set a configuration for a particular test, I'd be happy to implement.

teohhanhui: We only need a test for it, not a separate job. :smile:

teohhanhui: I see you've duplicated a part of our test suite. I think that's overkill. :see_no_evil:

.travis.yml
```diff
@@ -249,4 +249,45 @@ jobs:
         - *validate-openapi-v3-json
         - *validate-openapi-v3-yaml
 
+    - php: '7.3'
+      env: APP_ENV=absolute_url
```
teohhanhui: Still the same problem as https://github.com/api-platform/core/pull/3198#discussion_r336795993

We should not duplicate (parts of) our test suite just for one feature, or require extra CI jobs for it.

antograssiot: What do you propose @teohhanhui, I don't see other way of testing this in behat

soyuka: We need to have an attribute in the ResourceMetadata that changes this parameter resource-wise, then add a test on a new resource

dunglas: I agree, using an attribute, and the new "defaults" mechanism looks better!

antograssiot: I've started to update yesterday and it is working but there is one issue with this approach that I'm not sure how to tackle. How to update the Entrypoint in every format and the @context for errors in JSON-LD. Those are not tight to resource. Any idea ?

antograssiot: would you use the new default mechanism to still set a parameter in the container for those cases ?

antograssiot: updated, switched to resource attribute @soyuka  @dunglas 
no more suite duplication @teohhanhui

features/jsonld/absolute_url.feature
```diff
@@ -0,0 +1,83 @@
...
+    """
+    {
+      "absolute_url_dummies": "http://example.com/absolute_url_dummies/1"
+    }
+    """
+    Then the response status code should be 201
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "http:\/\/example.com\/contexts\/AbsoluteUrlRelationDummy",
```
soyuka: Do we really have to strip slashes in features?

antograssiot: Not at all they were in the initial commits and didn't noticed, updated :wink:

dunglas:
Thank you very much @mab05k @antograssiot

PR #3201, opened 2019-10-22
fix #3189 - JSONAPI pagination ignored when using filter parameter

### LINKED ISSUE #3189: JSONAPI pagination ignored when using filter parameter
**API Platform version(s) affected**: 2.5.0 (or older)

**Description**  
When requesting a resource using the headers `Accept: application/vnd.api+json` and applying both page[page] parameter and a filter[property] parameter. The request attribute `_api_filters` gets cleared in the `TransformFilteringParametersListener`. Resulting that the `page` value on this attribute is removed and pagination is ignored.
The reason why this happens, is because the `TransformPaginationParametersListener` gets executed before the `TransformFilteringParametersListener`. While the `TransformFilteringParametersListener` has the following piece of code:

```PHP
$request->attributes->set('_api_filters', $filters);
```

**How to reproduce**  
Do a request like below:
```
curl -v 'https://demo.api-platform.com/books?page[page]=5&filter[title]=re' -H 'Accept: application/vnd.api+json' -H 'Host: demo.api-platform.com'
```

Now you'll notice the meta in the response indicates we received `currentPage: 1`

When doing the same request without filter:
```
curl -v 'https://demo.api-platform.com/books?page[page]=5' -H 'Accept: application/vnd.api+json' -H 'Host: demo.api-platform.com'
```
The metadata indicates `currentPage: 5`.

**Possible Solution**  
1. Apply a stricter execution order for the listeners, the filter listener should execute before the pagination listener (not sure what side effects get introduced here).
2. In the `TransformFilteringParametersListener` we extend the `_api_filters` request attribute instead of overwrite.

Since we are not aware of the side effects we have applied solution 2 locally, which seem to work fine for now.

wcoppens:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  |no 
| BC breaks?    | no
| Deprecations? |no
| Tickets       | fixes #3189 
| License       | MIT

This PR fixes issue #3189, when using the jsonapi protocol pagination was overwritten when applying additional filters.

### DIFF
--- a/features/jsonapi/filtering.feature
+++ b/features/jsonapi/filtering.feature
@@ -23,6 +23,14 @@ Feature: JSON API filter handling
     And the JSON should be valid according to the JSON API schema
     And the JSON node "data" should have 0 elements
 
+  Scenario: Apply filters and pagination at the same time
+    When I send a "GET" request to "/dummies?filter[name]=foo&page[page]=2"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the JSON should be valid according to the JSON API schema
+    Then the JSON node "meta.currentPage" should be a number
+    Then the JSON node "meta.currentPage" should be equal to "2"
+
   Scenario: Apply property filter based on the 'fields'
     Given there are 2 dummy property objects
     When I send a "GET" request to "/dummy_properties?fields[DummyProperty]=id,foo,bar"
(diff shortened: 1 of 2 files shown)


soyuka:
can you rebase please?

dunglas:
Thanks @wcoppens!

PR #3218, opened 2019-10-28
Link to local XSD

BackEndTea:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/a
| License       | MIT
| Doc PR        | N/a

This will allow IDEs to provide hints/ autocompletion

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,7 +2,7 @@
 
 <!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.6/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
          backupGlobals="false"
          bootstrap="tests/Fixtures/app/bootstrap.php"
          colors="true"


dunglas:
Thanks @BackEndTea

BackEndTea:
Thanks @dunglas

teohhanhui:
I think this change should be reverted...

BackEndTea:
Does this change break anything for you?

teohhanhui:
No, but see: https://www.jetbrains.com/help/phpstorm/referencing-xml-schemas-and-dtds.html

This change seems unnecessary as the IDE can indeed work with remote XSD files.

But I see this is following https://github.com/symfony/recipes/pull/652

For a project that probably makes sense, but I'm not sure about a library... (What are the pros and cons?)

BackEndTea:
One of the benefits is that there is no extra user interaction needed to get any kind of auto completion/ hints. Users would have to trigger that request themselves in order to get it.

It is also possible that the link is referring to a version of the XSD that does not match the current one. Either we have an old link, or we link to a 'latest' version, which may be too new. There is also the danger that the link no longer works.

Because we use the xsd phpunit provides us with, we always have the correct version, we don't need to make any additional HTTP requests, and we can even use its information if we are offline.

teohhanhui:
Gotcha! We should cherry-pick this for `2.5` actually. :smile:

PR #3224, opened 2019-10-30
Use IriConverter without output with new attribute

FabienSalles:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Allow to have content-location without output if output is set to false and set_content_location (the new attribute) is set to true
 #hackdayparis

### DIFF
--- a/src/EventListener/WriteListener.php
+++ b/src/EventListener/WriteListener.php
@@ -87,6 +87,7 @@ final class WriteListener
                 }
 
                 $hasOutput = true;
+                $resourceMetadata = null;
                 if ($this->resourceMetadataFactory instanceof ResourceMetadataFactoryInterface) {
                     $resourceMetadata = $this->resourceMetadataFactory->create($attributes['resource_class']);
                     $outputMetadata = $resourceMetadata->getOperationAttribute($attributes, 'output', [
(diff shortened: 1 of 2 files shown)


dunglas:
This attribute must also be added to https://github.com/api-platform/core/blob/master/src/Annotation/ApiResource.php#L69 for the Developer Experience. Sorry I forgot about that yesterday :(. Could you add it please?

soyuka:
@dunglas isn't this a fix though?

dunglas:
No it's a new feature.

FabienSalles:
> This attribute must also be added to https://github.com/api-platform/core/blob/master/src/Annotation/ApiResource.php#L69 for the Developer Experience. Sorry I forgot about that yesterday :(. Could you add it please?

it is only for documentation purpose ?

Currently, without any modification I can use the attribute for all operations like this : 

```
/** 
 * @ApiResource(
 * attributes={"set_content_location"=true},
 * output=false
 * )
 */
``` 

Or for a specific operation like this : 
```
/**
 * @ApiResource(
 * collectionOperations={
 * "post"={"set_content_location"=true}
 * },
 * output=false
 * )
```

If I declare the attribute as annotation, the declaration could change and it will be logic (I think) to use the attribute like this : 
```
/** 
 * @ApiResource(
 * set_content_location=true,
 * output=false
 * )
 */
```
But I didn't find a way to retrieve the value with this declaration and the method `getOperationAttribute` used in the `WriteListener` class

FabienSalles:
Another observation : every attributes annotated are in camelCase so if I have to add it and use it differently, I should convert it to camelcase ?

And, given that the content-location will be set only for write actions by the `WriteListener`, I would have preferred changed `set` to `write` like `write_content_location` or `writeContentLocation`

alanpoulain:
Should we update this PR @dunglas? It would great to have a use case (a PR in the docs would be nice).

PR #773, opened 2016-09-28
test DoctrineOrmPropertyMetadataFactory

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Made this because:
1. there was none
2. I needed to test something

PR #3227, opened 2019-10-31
Implement ApiProperty security attribute

fredericbarthelet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | [fixes #773](https://github.com/api-platform/api-platform/issues/773)
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1000

Implement new `security` attribute on ApiProperty annotations.
Allows built-in support for property removal from resource using acl expression language.

This initial PR is a POC to explain my intention :
- adding new ApiProperty attribute `security`
- injecting `ResourceAccessChecker` within `AbstractItemNormalizer`
- supercharge `isAllowedAttribute` method inherited from symfony `AbstractNormalizer` service
- early return based on acl resolution if resolved to false

WDYT ?


soyuka:
I find this interesting, @dunglas is this what you had in mind?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -55,13 +56,14 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
...
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, ResourceAccessCheckerInterface $resourceAccessChecker, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, ItemDataProviderInterface $itemDataProvider = null, bool $allowPlainIdentifiers = false, array $defaultContext = [], iterable $dataTransformers = [], ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
```
antograssiot: Additional arguments must be added at the end with and with a default value or it is a BC break

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -345,6 +348,21 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
+    protected function isAllowedAttribute($classOrObject, $attribute, $format = null, array $context = [])
+    {
+        $options = $this->getFactoryOptions($context);
+        $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $attribute, $options);
+        $security = $propertyMetadata->getAttribute('security');
+        if ($security && !$this->resourceAccessChecker->isGranted($attribute, $security)) {
+            return false;
+        }
+
+        return parent::isAllowedAttribute($classOrObject, $attribute, $format, $context);
```
teohhanhui: The parent method should be called first. No need to do extra work when we can return the value directly if it's already `false`. We only need to check the against the `security` attribute when the parent method returns `true`, since we're taking a blacklist approach here.

tests/Fixtures/TestBundle/Entity/Dummy.php
```diff
@@ -55,7 +55,7 @@ class Dummy
      *
      * @ORM\Column
      * @Assert\NotBlank
-     * @ApiProperty(iri="http://schema.org/name")
+     * @ApiProperty(iri="http://schema.org/name", security="is_granted('ROLE_USER')")
```
teohhanhui: Please don't make this change. I suspect this will affect many existing tests.

teohhanhui:
Just to provide some background: @fredericbarthelet discussed this with me during the Hackday in Paris.

Thanks for following up on this!

tests/Fixtures/TestBundle/Entity/Dummy.php
```diff
@@ -67,6 +67,14 @@ class Dummy
      */
     private $alias;
 
+    /**
+     * @var string The dummy secret property, only readable/writable by specific users
+     *
+     * @ORM\Column(nullable=true)
+     * @ApiProperty(security="is_granted('ROLE_ADMIN')")
+     */
+    private $adminOnlyProperty;
```
antograssiot: the same changes must be done in `tests/Fixtures/TestBundle/Document/Dummy.php` so the mongoDB suite will pass too

antograssiot: Just a suggestion, but using another Entity/Document will avoid to have to modify filters tests

Maybe this could be added to the SecuredDummy instead, it will have less side impacts and will be at the same place than the security attribute from ApiResource

fredericbarthelet: @antograssiot that what I noticed as well when I started implementing behat tests. I will move  it to `SecuredDummy` instead :)

fredericbarthelet:
@soyuka, @antograssiot, @teohhanhui  I updated my PR to take into account your comments.
I encountered an issue when integrated security property on graphQL API.
When a user makes a graphQL request with a field he cannot access based on this new security property, ie :

*Resource*
```
    /**
     * @ApiProperty(security="is_granted('ROLE_ADMIN')")
     */
    private $adminOnlyProperty = '';
```

*Request*
```
    {
      securedDummy(id: "/secured_dummies/2") {
        owner
        title
        adminOnlyProperty
      }
    }
```

I get a 200 HTTP answer with an `Internal Server Error` message because `Cannot return null for non-nullable field SecuredDummy.adminOnlyProperty`. IMHO, the ideal feedback message should be similar to the one returned when requesting non-existing properties, aka `Cannot query field \"adminOnlyProperty\" on type \"SecuredDummy\"`.

My idea would be to inject as well `ResourceAccessCheckerInterface` service in `src/GraphQl/Type/FieldsBuilder.php` and somehow modify built fields on graphQL allowed operations to remove those removed by the same `isAllowedAttribute` method of the `ItemNormalizer`. Do you agree with such solution ?

alanpoulain:
@fredericbarthelet I'm not sure it's a good idea to check the security inside the FieldsBuilder. The GraphQL system has two steps:
- building the schema
- resolving a query

The built schema should not be dynamic and dependent on the query. That's why you can remove fields for instance by using non-dynamic normalization context but not in your case IMO.

I'm not sure it's the best way, but my proposal would be to make the field nullable (or to throw an exception to force the user to make it nullable?) if it has a security attribute and set it to null if the user doesn't have the rights to see the value of this field.

teohhanhui:
> my proposal would be to make the field nullable (or to throw an exception to force the user to make it nullable?) if it has a security attribute and set it to null if the user doesn't have the rights to see the value of this field

That does not make sense... :x

alanpoulain:
From what I read in the GraphQL examples, it seems to be the current practice: https://graphql.org/learn/authorization/

teohhanhui:
I don't see the relevance of the link you've shared. I'm not advocating doing it at the GraphQL layer. Making fields nullable is propagating this authorization logic to the GraphQL layer, which is exactly what they've warned against.

alanpoulain:
In most GraphQL systems, the schema cannot change dynamically depending on the context (very often it's a static file). So you have to make sure the fields can be nullable if you want some users to not see the corresponding value.

That's also why in the type system, the types are nullable _by default_.

If you take the GitHub API (https://developer.github.com/v4), 95% of the fields are nullable.

alanpoulain:
See also: https://github.com/graphql/graphql-js/issues/113

alanpoulain:
GraphQL Ruby seems to be the only one doing it. I'm not sure we should too. https://graphql-ruby.org/authorization/visibility

teohhanhui:
I'm obviously not familiar with GraphQL, but to me, it seems like if you say some field is nullable, then you'd expect to get null value for it, but that's not the case here. It's just that the field is simply not available to you. Returning null would be wrong!

alanpoulain:
GraphQL has been thought by Facebook like this. Please read the linked issue I sent before.

teohhanhui:
Yes, I've read the linked issue. It seems like GitHub is using that feature provided by `graphql-ruby` too: https://github.com/graphql/graphql-js/issues/113#issuecomment-289442685

Also: https://github.com/graphql/graphql-js/issues/113#issuecomment-447665909

alanpoulain:
Anyway I don't think it's a good idea to do it like this here.
GraphQL Ruby is using an *explicit* annotation and GraphQL Java too.
If we want to do it, it should be explicit too IMO.

fredericbarthelet:
As of today, @teohhanhui @alanpoulain  I shall then remove any implementation of property mapping based on security attribute in graphQL part ? I would then only make the use of this property available for rest endpoints, correct ?

teohhanhui:
That sounds like a good strategy, yes. Let's get this feature for REST first.

alanpoulain:
I think it will be good to be coherent with GraphQL too...
For me there are two choices:
- force the field to be nullable in the GraphQL schema if it has the `security` attribute
- throw an exception if the field is not nullable if it has a `security` attribute

The @api-platform/core-team should make a choice and not just leave the problem because it's GraphQL.

alanpoulain:
I've added an issue for the `visibility` attribute too: https://github.com/api-platform/core/issues/3279.

teohhanhui:
I agree that it should be supported for GraphQL too, but it needs more work before that's possible. We don't need to block this PR because of that.

The nullable proposal is not feasible, per discussion above.

alanpoulain:
The nullable property is totally feasible.
If you refer to:
> Making fields nullable is propagating this authorization logic to the GraphQL layer, which is exactly what they've warned against.

I think you misunderstood what they say. First of all, they are talking about the resolver part of the GraphQL system. We are talking about the building part here. Secondly, it's only a recommendation to avoid duplication. Here we can do this with no issue like this. And finally the "problem" would be the same for the `visibility` attribute.

Now that the GraphQL system is introduced and is fully working, we should make sure new features are working for it too.

teohhanhui:
https://github.com/api-platform/core/pull/3227#issuecomment-555033943

alanpoulain:
I've already responded to this comment.
Yes you cannot distinguish between a "true" null value and a null value because it's not available for you.
But that's how GraphQL has been thought. Even if you don't agree with this (which I can understand), it's working like this today in most GraphQL systems. We are not here to question the GraphQL specification.
And if the user has a problem with it, the `visibility` attribute could solve it.
That's how GraphQL Ruby is doing it for instance: https://graphql-ruby.org/authorization/authorization#handling-unauthorized-objects

teohhanhui:
So yes, if you insist that GraphQL support needs to be implemented in this PR (which I disagree with), then we're now blocked by #3279

alanpoulain:
Blocked? No we're not. This `visibility` attribute can be done in a second time (or even never).
I've proposed two solutions above to solve the GraphQL part. Both have advantages and drawbacks but we just need to make a choice.

teohhanhui:
Do you have examples of other GraphQL implementations which return null on unauthorized access? I'm not convinced that it's a feasible alternative.

alanpoulain:
Sure...
GraphQLite: https://graphqlite.thecodingmachine.io/docs/authentication_authorization#constant-schemas
GraphQL Java: https://github.com/graphql-java/graphql-java/issues/290#issuecomment-271115728
Graphene: http://docs.graphene-python.org/projects/django/en/latest/authorization/#filtering-id-based-node-access

I'm returning your question: please find a GraphQL system which *doesn't* work like this.

theofidry:
Can also confirm it’s the case for OnyxGraphQL as well. Note that it’s the
same on error: but in both cases you may have an error entry besides the
data one in the returned payload

alanpoulain:
Some GraphQL systems have not implemented the visibility feature yet.
@theofidry having an error entry could be a third solution. I don't really like it because it's not really an error for me but we can discuss it.

theofidry:
It is, because you requested access to a field you don’t have access to.
That said I would recommend to stick to the standards and not going for a
new exotic solution

theofidry:
Note however that I’m no GraphQL expert. I’m merely using one PHP
implementation and was pointing out how it behaves as a reference

fredericbarthelet:
I updated my PR, removing the implementation I had for GraphQL part for now.

@alanpoulain I would be happy to help implement the desired strategy on graphQL in a seperate PR related to the newly opened issue #3279 if that's OK for you :)

Any remaining comments on rest part on this one @teohhanhui ?

GregoireHebert:
I like this, could you also add tests for writing operations?

chalasr:
Finished in #3503, can be closed.

soyuka:
thanks @fredericbarthelet

dunglas:
For the record, I'm also in favor of returning `null` instead of throwing an error when a property cannot be accessed because of a security rule in the GraphQL subsystem.

PR #3229, opened 2019-10-31
Split iri converter interface

### LINKED ISSUE #1664: Consider splitting the IriConverter in 2 classes

ArnoudThibaut:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #1664  
| License       | MIT
| Doc PR        | NA

I have replace the usage of the IriConverterInterface by a more specific one when possible on classes tagged as experimental but some of them are quite old (PurgeHttpCacheListener for exemple) so I don't know if this classes are still experimental.

### DIFF
--- a/src/Api/IriConverterInterface.php
+++ b/src/Api/IriConverterInterface.php
@@ -13,55 +13,11 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Api;
 
-use ApiPlatform\Core\Exception\InvalidArgumentException;
-use ApiPlatform\Core\Exception\ItemNotFoundException;
-use ApiPlatform\Core\Exception\RuntimeException;
-
 /**
  * Converts item and resources to IRI and vice versa.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-interface IriConverterInterface
+interface IriConverterInterface extends ItemFromIriConverterInterface, IriFromItemConverterInterface
 {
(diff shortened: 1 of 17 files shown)


dunglas:
You can replace usage of this interface everywhere by the way, it will not break anything as the old interface still has the type of the new ones.

teohhanhui:
What is the rationale for this change? While I like the idea of splitting up the interface, I think we need better names.

lyrixx:
@teohhanhui https://github.com/api-platform/core/issues/1664

Issue #2031, opened 2018-06-20
Too many normalize iterations


SergeyMilegov:
Just imagine.
I have 50 custom normalizers.
I have api action which returns me 1000 translations (key => value).
The Api Platform try to normalize every value.
So we have 50000 useless iterations. What for?
Make ability to disable normalization for scalar attributes please.

soyuka:
Can't you create your own CollectionNormalizer that override ours? Then you apply the patch from https://github.com/api-platform/core/pull/2029 and you're good to go? :)

SergeyMilegov:
How does it help another people?

ymarillet:
The problem is potentially `I have api action which returns me 1000 translations (key => value).`, can't you filter at this level ?

teohhanhui:
The 1000 translations are not attributes on the object, right? If the key-value pairs are in a PHP `array`, it should just be 1 call. Or ideally in a [Map](https://secure.php.net/manual/en/class.ds-map.php), and we should make sure that it's well supported...

SergeyMilegov:
>If the key-value pairs are in a PHP array, it should just be 1 call.
Yes it is array with (key-value). And api platform tries to normalize every value.

teohhanhui:
Okay, that sounds bad.

PR #3232, opened 2019-11-01
Skip the serializer for null and scalar values on normalize

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no (but see note below)
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/core/issues/2593#issuecomment-476881738, fixes https://github.com/api-platform/core/issues/2031
| License       | MIT
| Doc PR        | N/A

Notable changes:

- The serializer is no longer called for `null` and scalar values during normalization. This is consistent with the behaviour in Symfony Serializer's `AbstractObjectNormalizer`: https://github.com/symfony/symfony/blob/v4.3.6/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L194-L196
- No more "child inherited" nonsense.

Potential BC breaks:

- If anyone has been using custom normalizers for `null` and scalar values, those custom normalizers will no longer be called. I think that is really an edge case, while the performance win of this change would be felt by the majority of our users.

TODO:

- [x] Rebase onto `master`
- [ ] Blackfire profile


tests/Serializer/AbstractItemNormalizerTest.php
```diff
@@ -835,8 +835,6 @@ public function testChildInheritedProperty()
 
         $serializerProphecy = $this->prophesize(SerializerInterface::class);
         $serializerProphecy->willImplement(NormalizerInterface::class);
-        $serializerProphecy->normalize('foo', null, Argument::type('array'))->willReturn('foo')->shouldBeCalled();
-        $serializerProphecy->normalize(null, null, Argument::type('array'))->willReturn(null)->shouldBeCalled();
```
soyuka: but now this may break some stuff no? :p

teohhanhui: See OP:

> Potential BC breaks:
> 
> - If anyone has been using custom normalizers for null and scalar values, those custom normalizers will no longer be called. I think that is really an edge case, while the performance win of this change would be felt by the majority of our users.

dunglas:
Could you provide a Blackfire profile to see if it is worth it?

dunglas:
Also, it must target master, not 2.5. BC breaks shouldn't be accepted even in master... but they definitely can't be shipped with patch releases.

teohhanhui:
I'd say it's a bug that we were even calling the serializer for normalizing `null` and scalar values, but sure, I'll target `master` if that's what you want...

Taluu:
I may have had just a quick look, but what if you want to remove a relationship (e.g a "nullable relation", or "nullable value" actually ?

(Disclaimer, again : Just had a quick look)

teohhanhui:
@Taluu:
> what if you want to remove a relationship?

Not relevant, as this change only affects normalization, not denormalization.

Taluu:
Oh, then what if you want to expose a `null` to say "there is no relation" here ? At least when I'm consuming an api, I'm pretty used to have this kind of "value" rather than having to check if the key is set. 

I may be wrong though.

teohhanhui:
> Oh, then what if you want to expose a null to say "there is no relation" here ? At least when I'm consuming an api, I'm pretty used to have this kind of "value" rather than having to check if the key is set.

It'll remain the same as before. This PR does not change any of that. The only potential BC break is if someone had custom normalizers for null / scalar values. Very highly unlikely.

Taluu:
Had one for denormalization so I guess that yes, shouldn't be a bother. 😄

src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
```diff
@@ -49,11 +49,6 @@
             <argument type="service" id="api_platform.property_info" />
         </service>
 
-        <service id="api_platform.metadata.property.name_collection_factory.inherited" class="ApiPlatform\Core\Metadata\Property\Factory\InheritedPropertyNameCollectionFactory" decorates="api_platform.metadata.property.name_collection_factory" decoration-priority="10" public="false">
-            <argument type="service" id="api_platform.metadata.resource.name_collection_factory" />
-            <argument type="service" id="api_platform.metadata.property.name_collection_factory.inherited.inner" />
-        </service>
```
Taluu: shouldn't these services be marked as deprecated instead ?

teohhanhui: This service was only decorating another service. It was never intended to be used directly.

soyuka:
Why did you close?

teohhanhui:
Continued in #3248 (targeting `2.5`) using a safer approach. Will open another PR targeting `master` after this, for the removal of "child inherited" metadata.

We will not skip the serializer for `null` and scalar property values.

PR #3234, opened 2019-11-03
Fix parent fields serialization when using inheritance

### LINKED ISSUE #3176: Parent model list child(ren) field(s) on parent resource calls
**API Platform version(s) affected**: 2.4.7

**Description**  
Given the following Model structure:
```
/**
 * @ApiResource()
 */
class Human
{
    /**
     * @var string
     * @ApiProperty(identifier=true)
     */
    public $id;

    /**
     * @var int
     */
    public $age;
}
```
and
```
/**
 * @ApiResource()
 */
class Person extends Human
{
    /**
     * @var string
     */
    public $name;
}
```

I except to have the property `name` on the Resource `Person` and not in the `Human` one
But I get the `name` property also on `/humans` calls.

**How to reproduce**  
Create one resource inheriting from an other.
Call an endpoint on the parent one and see the child's field being there.

**Possible Solution**   
The problem seems to come from https://github.com/api-platform/core/blob/c4f0ef14709ff85fecb36075501e78e247ea2993/src/Metadata/Property/Factory/InheritedPropertyMetadataFactory.php#L47-L51

When I'm doing the call `GET /humans` I see that `$resourceClass` is `Human` and have `$knownResourceClass` being `Person`, so `is_subclass_of` return true, and it adds the child property to the parent.
Doing `if (is_subclass_of($resourceClass, $knownResourceClass))` instead resolves the problem.

But before doing a PR I wanted to know if this behavior is correct, because ... ? 
Otherwise I would be happy to do the PR.

dFayet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3176 
| License       | MIT

 I know the case described in the issue is really possible with doctrine, but if you use a separate ApiResource class it might be a valid use case.

### DIFF
--- a/src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php
+++ b/src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php
@@ -51,7 +51,7 @@ final class InheritedPropertyNameCollectionFactory implements PropertyNameCollec
                 continue;
             }
 
-            if (is_subclass_of($knownResourceClass, $resourceClass)) {
+            if (is_subclass_of($resourceClass, $knownResourceClass)) {
                 foreach ($this->create($knownResourceClass) as $propertyName) {
                     $propertyNames[$propertyName] = $propertyName;
                 }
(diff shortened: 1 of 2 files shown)


soyuka:
bug fix should target 2.5

dFayet:
Seems to be no more relevant. See https://github.com/api-platform/core/issues/3176#issuecomment-552176832

PR #3235, opened 2019-11-03
Add Api-Platform's version in debug bar

dFayet:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

After a small discussion with @dunglas it appears it could be nice to display the Api-Platform's version in the Symfony's debug-bar.

Two solutions were talked about, adding a constant (:-1: because it needs to be maintained), or using ocramius/package-versions (this package is required-dev by symfony/orm).

This feature was proposed to Symfony but it was [refused](https://github.com/symfony/symfony/pull/34221).

So here we are :-)

Output:
Capture d’écran de 2019-11-03 18-35-05
Capture d’écran de 2019-11-03 18-34-55

Note:
I was not sure about adding tests or not, as the outpu depends of a final class, that might exists or not

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -96,6 +96,7 @@
         "doctrine/mongodb-odm-bundle": "To support MongoDB. Only versions 4.0 and later are supported.",
         "elasticsearch/elasticsearch": "To support Elasticsearch.",
         "guzzlehttp/guzzle": "To use the HTTP cache invalidation system.",
+        "ocramius/package-versions": "To display the API Platform's version in the debug bar.",
         "phpdocumentor/reflection-docblock": "To support extracting metadata from PHPDoc.",
         "psr/cache-implementation": "To use metadata caching.",
         "ramsey/uuid": "To support Ramsey's UUID identifiers.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.6.0
 
+* Display the API Platform's version in the debug-bar
 * MongoDB: Possibility to add execute options (aggregate command fields) for a resource, like `allowDiskUse` (#3144)
 * MongoDB: Mercure support (#3290)
 * GraphQL: Subscription support with Mercure (#3321)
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -97,6 +97,7 @@
         "elasticsearch/elasticsearch": "To support Elasticsearch.",
         "friendsofsymfony/user-bundle": "To use the FOSUserBundle bridge.",
         "guzzlehttp/guzzle": "To use the HTTP cache invalidation system.",
+        "ocramius/package-versions": "To display the Api-Platform's version in the debug bar.",
```
dunglas: `API Platform` (same everywhere)

antograssiot:
It migh be enough to add it in the small tooltip opened on mouseover to save space in the toolbar ?

dFayet:
> Could you also add a test please?

I've added a basic test, but as the data class is final + with a private construct I cannot mock it :(

alanpoulain:
Thank you @dFayet!

composer.json
```diff
@@ -96,6 +96,7 @@
         "doctrine/mongodb-odm-bundle": "To support MongoDB. Only versions 4.0 and later are supported.",
         "elasticsearch/elasticsearch": "To support Elasticsearch.",
         "guzzlehttp/guzzle": "To use the HTTP cache invalidation system.",
+        "ocramius/package-versions": "To display the API Platform's version in the debug bar.",
```
maxhelias, line 99: @alanpoulain wouldn't it be more interesting to suggest https://github.com/composer/package-versions-deprecated for composer 1 & 2 compatibility?

alanpoulain: Not sure about this. If the user doesn't use PHP 7.4 or Composer 2, it will find this package anyway, don't you think?
And `ocramius/package-versions` is the original package, it makes sense to me to add it to the suggest list.

maxhelias: Indeed 👍

PR #3246, opened 2019-11-12
fix compat with Symfony 5

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -89,7 +89,7 @@ parameters:
 			path: %currentWorkingDirectory%/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
 		-
 			message: '#Service "api_platform.iri_converter" is private\.#'
-			path:  %currentWorkingDirectory%/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
+			path: %currentWorkingDirectory%/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
 
 		# Expected, due to optional interfaces
 		- '#Method ApiPlatform\\Core\\Bridge\\Doctrine\\Orm\\Extension\\QueryCollectionExtensionInterface::applyToCollection\(\) invoked with 5 parameters, 3-4 required\.#'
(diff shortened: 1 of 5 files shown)


phpstan.neon.dist
```diff
@@ -117,3 +117,5 @@ parameters:
 		-
 			message: '#Call to method get(Class|Headers|StatusCode)\(\) on an unknown class Symfony\\Component\\ErrorRenderer\\Exception\\FlattenException\.#'
 			path: %currentWorkingDirectory%/src/Action/ExceptionAction.php
+
+		- '"#Symfony\\Component\\HttpKernel\\EventListener\\ErrorListener#'
```
dunglas:
```suggestion
		- '#Symfony\\Component\\HttpKernel\\EventListener\\ErrorListener#'
```

my bad

teohhanhui: I think this should be:

1. moved to the bottom-most "forward compatibility" section
2. match the entire phpstan error message
3. scoped to the relevant path(s)

teohhanhui:
This should target `2.5` branch, right?

nicolas-grekas:
Now targetting branch 2.5

alanpoulain:
Errors seem related.

teohhanhui:
Test mocking needs to handle both cases.

soyuka:
rebased + fixed by @flug waiting for CI

phpstan.neon.dist
```diff
@@ -114,5 +114,10 @@ parameters:
 		-
 			message: '#Parameter \$exception of method ApiPlatform\\Core\\Action\\ExceptionAction::__invoke\(\) has invalid typehint type Symfony\\Component\\ErrorHandler\\Exception\\FlattenException\.#'
 			path: %currentWorkingDirectory%/src/Action/ExceptionAction.php
+
```
teohhanhui: Remove blank line.

phpstan.neon.dist
```diff
@@ -114,5 +114,10 @@ parameters:
...
 		- '#Call to method get(Class|Headers|StatusCode)\(\) on an unknown class Symfony\\Component\\ErrorHandler\\Exception\\FlattenException\.#'
 		- '#Class Symfony\\Component\\ErrorHandler\\Exception\\FlattenException not found\.#'
+		- '#Symfony\\Component\\HttpKernel\\EventListener\\ErrorListener#'
```
teohhanhui: I really don't like this one. It's too generic. Can we make it specific please?

phpstan.neon.dist
```diff
@@ -114,5 +114,10 @@ parameters:
...
+		-
+			message: "#Call to function method_exists\\(\\) with 'Symfony\\\\\\\\Component.+' and 'getException' will always evaluate to false\\.#"
```
teohhanhui: The over-escaping is scary. :see_no_evil:

soyuka: I've copy pasted from above.

teohhanhui:
@soyuka You've messed up the indentation in `phpstan.neon.dist` :laughing:

phpstan.neon.dist
```diff
@@ -87,7 +87,7 @@ parameters:
 			path: %currentWorkingDirectory%/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
 		-
 			message: '#Service "api_platform.iri_converter" is private\.#'
-			path:  %currentWorkingDirectory%/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
+			path:	%currentWorkingDirectory%/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
```
teohhanhui: Whitespace error. :laughing:

teohhanhui:
Looks like we still need to add more conditional mocking.

src/Action/ExceptionAction.php
```diff
@@ -50,6 +52,10 @@ public function __construct(SerializerInterface $serializer, array $errorFormats
      */
     public function __invoke($exception, Request $request): Response
     {
+        if ($exception instanceof \Throwable) {
+            $exception = FlattenException::createFromThrowable($exception);
+        }
```
soyuka: We have an issue here but I'm not sure where to fix it (it should definitely be handled elsewhere). 
In case of a AccessDeniedException we receive an AccessDeniedHttpException here instead of a FlattenException. Something is not right.

teohhanhui: @chalasr Thoughts?

chalasr: We started to make `FlattenException`  go away in https://github.com/symfony/symfony/pull/34309, some `\Throwable` aren't flattened anymore at this stage. Handling any `Throwable` seems right, that's what the symfony's `ErrorController` does (which should make this one obsolete btw, we may want to deprecate it on master).

teohhanhui: Sorry, I don't understand what's the connection between https://github.com/symfony/symfony/pull/34309 and this issue.

The problem here is that we're expecting `FlattenException` but getting `AccessDeniedHttpException` instead...

chalasr: Bad reference sorry. The new ErrorListener does not flatten it anymore unless the $exception argument has a FlattenException type declaration, see https://github.com/symfony/symfony/pull/34312/files#r344517703

chalasr: See https://github.com/symfony/symfony/pull/34411

teohhanhui: Thanks! I hope it'll be green now.

soyuka:
Thanks @nicolas-grekas @chalasr @teohhanhui @flug

PR #3247, opened 2019-11-12
Fix ApiResource annotation : 'int' should be 'bool'

ThomasLassalle:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

`$paginationPartial` in `src/Annotation/ApiResource.php` is declared as a boolean in the attributes annotation, but is declared as an int in the class.


alanpoulain:
Thanks @ThomasLassalle.

PR #3256, opened 2019-11-14
Sort mapping resources

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes-ish
| New feature?  | no-ish
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR makes the loading of mapping resources idempotent by sorting files.

Given two mapping files for two different resources that define the following routes:

* resource FooBar (collection): `/foo/bar`
* resource Foo (item): `/foo/{id}`

Functional tests prove that both routes work as expected. Though, once the tests are executed on some CI, they randomly fail. After a long debugging session, I noticed that routes can only work if the files are loaded in the same order as above because otherwise the second route will always match (because Symfony stops matching as soon as a matching route is found). The only way I was able to reproduce the issue was by forcing the order of files (as in this PR). I assume that the context of the CI had some specific effect on the ordering.

The proper fix was to add a pattern to the `id` parameter: `/foo/{id<(?!bar)[^/]+>}`. This guarantees that both routes work as expected, whatever the order they are defined in. But if the order in which files are loaded is always the same, debugging becomes easier.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -323,7 +323,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
         foreach ($paths as $path) {
             if (is_dir($path)) {
-                foreach (Finder::create()->followLinks()->files()->in($path)->name('/\.(xml|ya?ml)$/') as $file) {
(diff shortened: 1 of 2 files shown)


dunglas:
LGTM

alanpoulain (approved):

alanpoulain:
Thanks @julienfalque.

julienfalque:
Thank you @alanpoulain @dunglas.

PR #3260, opened 2019-11-14
Cache ext-mongodb build directory

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

The `./configure` and `make` were taking quite a long time.

Thanks @bendavies for your exploration in #3253

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ env:
   COMPOSER_ALLOW_SUPERUSER: '1' # https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
   COMPOSER_INSTALLER_COMMIT: fb22b78362d31c0d2bf516d1f8cdfd2745caa431
   COVERAGE: '0'
-  EXT_MONGODB_VERSION: '1.5.5'
+  EXT_MONGODB_VERSION: '1.6.0'
   EXT_PCOV_VERSION: '1.0.6'
   LEGACY: '0'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -40,17 +41,47 @@ jobs:
         run: |
           apk add \
             unzip \
+      - name: Cache mongodb PHP extension build
+        if: (!startsWith(matrix.php, '7.1'))
+        uses: actions/cache@v1
+        with:
+          path: /var/tmp/build/ext-mongodb-${{ env.EXT_MONGODB_VERSION }}
+          key: ext-mongodb-${{ env.EXT_MONGODB_VERSION }}-php${{ matrix.php }}-${{ github.sha }}
```
bendavies: is `${{ github.sha }}` really needed here?
we only added it for composer to generate uniqueness so the cache is additive. doesn't' seem neccesary here

teohhanhui: We want to save the cache if anything has changed. This is the only way.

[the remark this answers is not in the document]
teohhanhui: You're right, but this will be hard to achieve here... It'll require another step before this. :x

I think we can live with this for now. Keep things simple.

soyuka:
thanks @teohhanhui @bendavies

PR #3263, opened 2019-11-15
feat: add payload controller value resolver

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | -
| Doc PR        | -

When creating a custom controller, the argument which is passed the deserialized payload of the request must be named `$data`. This can be very frustrating, especially when you don't know this requirement and you don't understand why your controller is not working properly.

This PR introduces a controller value resolver that allows to use any argument name, as long as its type declaration matches the resource class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * DataProvider: Add `TraversablePaginator` (#3783)
 * JSON:API: Support inclusion of resources from path (#3288)
 * Swagger UI: Add `swagger_ui_extra_configuration` to Swagger / OpenAPI configuration (#3731)
+* Allow controller argument with a name different from `$data` thanks to an argument resolver (#3263)
 * GraphQL: Support `ApiProperty` security (#4143)
 * GraphQL: **BC** Fix security on association collection properties. The collection resource `item_query` security is no longer used. `ApiProperty` security can now be used to secure collection (or any other) properties. (#4143)
 * Deprecate `allow_plain_identifiers` option (#4167)
--- a/src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
+++ b/src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\ArgumentResolver;
+
+use ApiPlatform\Core\EventListener\ToggleableDeserializationTrait;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+final class PayloadArgumentResolver implements ArgumentValueResolverInterface
+{
+    use ToggleableDeserializationTrait;
+
+    private $serializationContextBuilder;
+
+    public function __construct(
+        ResourceMetadataFactoryInterface $resourceMetadataFactory,
+        SerializerContextBuilderInterface $serializationContextBuilder
+    ) {
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->serializationContextBuilder = $serializationContextBuilder;
+    }
+
+    public function supports(Request $request, ArgumentMetadata $argument): bool
+    {
+        if ($argument->isVariadic()) {
+            return false;
+        }
+
+        $class = $argument->getType();
+
+        if (null === $class) {
+            return false;
+        }
+
+        if (null === $request->attributes->get('data')) {
+            return false;
+        }
+
+        $inputClass = $this->getExpectedInputClass($request);
+
+        if (null === $inputClass) {
+            return false;
+        }
+
+        return $inputClass === $class || is_subclass_of($inputClass, $class);
+    }
+
+    public function resolve(Request $request, ArgumentMetadata $argument): \Generator
+    {
+        yield $request->attributes->get('data');
+    }
+
+    private function getExpectedInputClass(Request $request): ?string
+    {
+        $attributes = RequestAttributesExtractor::extractAttributes($request);
+
+        if (!$this->isRequestToDeserialize($request, $attributes)) {
+            return null;
+        }
+
+        $context = $this->serializationContextBuilder->createFromRequest($request, false, $attributes);
+
+        return $context['input'] ?? $context['resource_class'];
+    }
+}
--- a/src/Bridge/Symfony/Bundle/Resources/config/argument_resolver.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/argument_resolver.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="api_platform.argument_resolver.payload" class="ApiPlatform\Core\Bridge\Symfony\Bundle\ArgumentResolver\PayloadArgumentResolver" public="false">
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory" />
+            <argument type="service" id="api_platform.serializer.context_builder" />
+
+            <tag name="controller.argument_value_resolver" />
+        </service>
+    </services>
+
+</container>
--- a/tests/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolverTest.php
+++ b/tests/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolverTest.php
@@ -1,0 +1,308 @@
+<?php
+
+/*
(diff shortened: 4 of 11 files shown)


dunglas:
I like it!

ro0NL:
from #3268 , we do

```php
<?php

use ApiPlatform\Core\Validator\ValidatorInterface;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
use Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterInterface;

class ApiInputConverter implements ParamConverterInterface
{
    /**
     * @var SerializerInterface
     */
    private $serializer;

    /**
     * @var ValidatorInterface
     */
    private $validator;

    public function __construct(SerializerInterface $serializer, ValidatorInterface $validator) {
        $this->validator = $validator;
    }

    public function apply(Request $request, ParamConverter $configuration)
    {
        $attribute = $configuration->getName();

        if ($request->attributes->has($attribute)) {
            return false;
        }

        $input = $this->serializer->deserialize($request->getContent(), $configuration->getClass(), $request->getRequestFormat());

        $this->validator->validate($input);

        $request->attributes->set($attribute, $input);

        return true;
    }

    public function supports(ParamConverter $configuration)
    {
        $options = $configuration->getOptions();

        if ($options['api_input_dto'] ?? false) {
            if (null === $configuration->getClass()) {
                throw new \LogicException('Parameter "'.$configuration->getName().'" must have a class type hint set to deserialize.');
            }

            return true;
        }

        return false;
    }
}
```

dunglas:
Do we really need the dependency to SensioExtraFramework? Symfony has some native Param Converter without this dependency for instance.

ro0NL:
AFAIK it's simply matter of setting a request attribute with deserialized/validated payload. Then we can map it to controller arguments, if needed.

To use different attributes names, i think implementing `ArgumentValueResolverInterface` will do indeed :+1:

julienfalque:
Right, I forgot about `ArgumentValueResolverInterface`, I'll rework my PR.

julienfalque:
PR reworked.

I'd like to add some functional tests. Does it make sense to use Behat for that? I mean: what this PR introduces is more of a technical improvement than a real user feature.

julienfalque:
Added some integration tests.

src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
```diff
@@ -0,0 +1,87 @@
...
+
+    private function getExpectedInputClass(Request $request): ?string
+    {
+        $attributes = RequestAttributesExtractor::extractAttributes($request);
+
+        if (!$this->isRequestToDeserialize($request, $attributes)) {
+            return null;
+        }
+
+        $context = $this->serializationContextBuilder->createFromRequest($request, false, $attributes);
```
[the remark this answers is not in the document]
julienfalque: Mmh indeed, `resource_class` is available in `$attributes` directly, but I'm not sure about `input`: the context builder seems to have some custom logic to retrieve it from operation configuration.

src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
```diff
@@ -0,0 +1,87 @@
...
+
+use ApiPlatform\Core\EventListener\ToggleableDeserializationTrait;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+class PayloadArgumentResolver implements ArgumentValueResolverInterface
```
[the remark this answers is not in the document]
julienfalque: Updated.

src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
```diff
@@ -0,0 +1,87 @@
...
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+final class PayloadArgumentResolver implements ArgumentValueResolverInterface
+{
+    use ToggleableDeserializationTrait;
+
+    /**
+     * @var SerializerContextBuilderInterface
+     */
```
alanpoulain, line 30: It's not needed.

julienfalque: Removed.

src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        return $inputClass === $class || is_subclass_of($inputClass, $class);
+    }
+
+    public function resolve(Request $request, ArgumentMetadata $argument): \Generator
+    {
+        if (!$this->supports($request, $argument)) {
+            throw new \InvalidArgumentException('Given request and argument not supported.');
+        }
```
alanpoulain, line 67: This check shouldn't be necessary, isn't it?

src/Bridge/Symfony/Bundle/Resources/config/argument_resolver.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="api_platform.payload_argument_resolver" class="ApiPlatform\Core\Bridge\Symfony\Bundle\ArgumentResolver\PayloadArgumentResolver" public="false">
```
alanpoulain, line 8: Isn't it better in the case we add other argument resolvers?
```suggestion
        <service id="api_platform.argument_resolver.payload" class="ApiPlatform\Core\Bridge\Symfony\Bundle\ArgumentResolver\PayloadArgumentResolver" public="false">
```

tests/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolverTest.php
```diff
@@ -0,0 +1,343 @@
...
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
+use Prophecy\Argument;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+/**
+ * @covers \ApiPlatform\Core\Bridge\Symfony\Bundle\ArgumentResolver\PayloadArgumentResolver
```
alanpoulain, line 26: We don't use the `@covers` annotation.

tests/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolverTest.php
```diff
@@ -0,0 +1,343 @@
...
+    public function provideUnsupportedArguments(): iterable
+    {
+        yield 'argument without type declaration' => [$this->createArgumentMetadata()];
+        yield 'variadic argument' => [$this->createArgumentMetadata(ResourceImplementation::class, true)];
+    }
+
+    /**
+     * @dataProvider provideIntegrationCases
+     */
+    public function testIntegration(Request $request, callable $controller, array $expectedArguments): void
```
alanpoulain, line 234: Shouldn't we add the integration tests in another file? If possible we prefer to use Behat for this kind of tests too (but maybe it would be harder to test it and these tests seem OK to maintain).

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * MongoDB: `date_immutable` support (#3940)
 * DataProvider: Add `TraversablePaginator` (#3783)
 * Swagger UI: Add `swagger_ui_extra_configuration` to Swagger / OpenAPI configuration (#3731)
+* Allow controller argument with a name different from `$data` thanks to an argument resolver
```
alanpoulain, line 12:
```suggestion
* Allow controller argument with a name different from `$data` thanks to an argument resolver (#3263)
```

tests/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolverTest.php
julienfalque: Removed.

src/Bridge/Symfony/Bundle/Resources/config/argument_resolver.xml
julienfalque: Updated.

src/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
julienfalque: Removed.

tests/Bridge/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolverTest.php
julienfalque: > If possible we prefer to use Behat

I think it would be much harder to have reliable Behat tests and would require more work to write and maintain. Not worth it IMO.

> Shouldn't we add the integration tests in another file

IMO it's fine to keep it here: all tests in this file cover the same subject. But if you prefer, I can extract it to it own file.

alanpoulain:
@julienfalque LGTM. Could you provide a documentation?

alanpoulain:
Thank you @julienfalque.

julienfalque:
Thank you @alanpoulain @dunglas. Sorry I didn't update the documentation, I totally forgot about your request.

PR #3262, opened 2019-11-14
[WIP] Add RouterListener to handle request on preflight responses

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | https://github.com/nelmio/NelmioCorsBundle/issues/115
| License       | MIT
| PR related | https://github.com/nelmio/NelmioCorsBundle/pull/140

Please see https://github.com/nelmio/NelmioCorsBundle/pull/125 comments about why this fix should be done here instead of NelmioCorsBundle

PR #3265, opened 2019-11-15
Prevent adding Link headers for CORS preflight requets

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixes #3262
| License       | MIT
| Doc PR        | n/a

Preflight requests cannot be accessed in JS, therefore settings the `Link` headers for Hydra and Mercure is useless for these requests Not setting them prevent the issues with invalid URLs being generated. This an alternative fix to #3262.

### DIFF
--- a/src/Hydra/EventListener/AddLinkHeaderListener.php
+++ b/src/Hydra/EventListener/AddLinkHeaderListener.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\Hydra\EventListener;
 
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
 use ApiPlatform\Core\JsonLd\ContextBuilder;
+use ApiPlatform\Core\Util\CorsTrait;
 use Fig\Link\GenericLinkProvider;
 use Fig\Link\Link;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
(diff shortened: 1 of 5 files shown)


teohhanhui:
I think this approach is risky. During my exploration from many years ago, I came upon this conclusion: https://github.com/nelmio/NelmioCorsBundle/pull/54#issuecomment-162838133

teohhanhui:
Really, `NelmioCorsBundle` is breaking all `OPTIONS` requests, it should be fixed there...

dunglas:
But is there a better short term solution as NelmioCors won't change soon? It doesn't hurt for these 2 specific headers to not be generated for preflight requests (actually, it will even help reducing servers' load). It's not a general solution, but for Mercure and Hydra it's good enough IMO.

teohhanhui:
There are consequences for caching. We must add `Vary: Access-Control-Request-Method` now.

teohhanhui:
For a short term solution, I'd rather we set the `RequestContext`, instead of varying the response based on this header.

dunglas:
I don't want to introduce such complexity (a new listener that we'll have to maintain BC for) just to fix this issue. And NelmioCors is already returning a different response for Preflight requests and non-preflight requests, so the potential cache issue already exists.

My proposal is to:

* "temporary" merge this as a quick fix
* remove those checks if/when NelmioCors is fixed, or if Symfony includes native CORS support (it would be the best option IMHO)

teohhanhui:
Thanks @dunglas! :tada:

PR #3273, opened 2019-11-18
Remove nonsensical "child inherited" metadata

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes?
| New feature?  | no
| BC breaks?    | no?
| Deprecations? | yes
| Tickets       | not sure
| License       | MIT
| Doc PR        | N/A

From #3232. Targeting `master` because it's considered risky to go into a patch release.


teohhanhui:
phpstan error will be fixed by https://github.com/api-platform/core/pull/3274

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -503,6 +512,7 @@ protected function getAttributeValue($object, $attribute, $format = null, array
         $context['api_attribute'] = $attribute;
         $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $attribute, $this->getFactoryOptions($context));
 
+        // BC to be removed in 3.0
```
soyuka: what should be removed?

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -48,8 +48,7 @@ public function create(string $resourceClass, string $property, array $options =
     {
         $propertyMetadata = $this->decorated->create($resourceClass, $property, $options);
 
-        // in case of a property inherited (in a child class), we need it's properties
-        // to be mapped against serialization groups instead of the parent ones.
+        // BC to be removed in 3.0
         if (null !== ($childResourceClass = $propertyMetadata->getChildInherited())) {
```
soyuka: can you specify that `childInherited` is deprecated which is why it needs to be removed?

src/Serializer/AbstractItemNormalizer.php
teohhanhui: The whole try-catch except for the getValue.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
teohhanhui: Hmm... We don't do this elsewhere. I think it's enough to know that this block should be removed when preparing 3. The code should not even be used under normal circumstances.

soyuka:
can you rebase, let's merge this :)

teohhanhui:
Rebased. :smile:

teohhanhui:
phpstan failure seems unrelated... Looks like it's caused by the release of Symfony 5.0 :laughing:

PR #3275, opened 2019-11-18
Improve Vulcain support

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

Allows to generate IRIs instead of embedding documents for Hydra collections.

TODO:

* [x] update the JSON-LD context accordingly


teohhanhui:
So is it expected to be used like this for now?

```yaml
api_platform:
    defaults:
        normalization_context:
            iri_only: true
```

dunglas:
Exactly, until https://github.com/symfony/symfony/pull/28927 is ready.

PR #3276, opened 2019-11-19
Remove http_cache.invalidation deprecation and fix some remaining deprec

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -52,7 +52,7 @@ final class OrderExtension implements ContextAwareQueryCollectionExtensionInterf
         $identifiers = $classMetaData->getIdentifier();
         if (null !== $this->resourceMetadataFactory) {
             $defaultOrder = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('order');
-            if (null !== $defaultOrder) {
+            if (\is_array($defaultOrder)) {
                 foreach ($defaultOrder as $field => $order) {
                     if (\is_int($field)) {
                         // Default direction
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -169,24 +169,24 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.collection.exists_parameter_name', $config['collection']['exists_parameter_name']);
         $container->setParameter('api_platform.collection.order', $config['collection']['order']);
         $container->setParameter('api_platform.collection.order_parameter_name', $config['collection']['order_parameter_name']);
-        $container->setParameter('api_platform.collection.pagination.enabled', $this->isConfigEnabled($container, $config['collection']['pagination']));
-        $container->setParameter('api_platform.collection.pagination.partial', $config['collection']['pagination']['partial']);
-        $container->setParameter('api_platform.collection.pagination.client_enabled', $config['collection']['pagination']['client_enabled']);
-        $container->setParameter('api_platform.collection.pagination.client_items_per_page', $config['collection']['pagination']['client_items_per_page']);
-        $container->setParameter('api_platform.collection.pagination.client_partial', $config['collection']['pagination']['client_partial']);
-        $container->setParameter('api_platform.collection.pagination.items_per_page', $config['collection']['pagination']['items_per_page']);
-        $container->setParameter('api_platform.collection.pagination.maximum_items_per_page', $config['collection']['pagination']['maximum_items_per_page']);
-        $container->setParameter('api_platform.collection.pagination.page_parameter_name', $config['collection']['pagination']['page_parameter_name']);
-        $container->setParameter('api_platform.collection.pagination.enabled_parameter_name', $config['collection']['pagination']['enabled_parameter_name']);
-        $container->setParameter('api_platform.collection.pagination.items_per_page_parameter_name', $config['collection']['pagination']['items_per_page_parameter_name']);
-        $container->setParameter('api_platform.collection.pagination.partial_parameter_name', $config['collection']['pagination']['partial_parameter_name']);
-        $container->setParameter('api_platform.collection.pagination', $config['collection']['pagination']);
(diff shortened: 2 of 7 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -366,7 +366,6 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
                         ->end()
                         ->booleanNode('public')->defaultNull()->info('To make all responses public by default.')->end()
                         ->arrayNode('invalidation')
-                            ->setDeprecated('The use of the `http_cache.invalidation` has been deprecated in 2.6 and will be removed in 3.0. Use `defaults.cache_headers.invalidation` instead.')
```
teohhanhui: Why?

dunglas: Because AFAIU it’s not an attribute, so setting it in defaults is useless.

tests/Fixtures/app/config/config_common.yml
```diff
@@ -58,25 +58,25 @@ api_platform:
...
         ApiPlatform\Core\Exception\FilterValidationException: !php/const Symfony\Component\HttpFoundation\Response::HTTP_BAD_REQUEST
     http_cache:
         invalidation:
             enabled: true
-        max_age:        60
-        shared_max_age: 3600
-        vary:           ['Accept', 'Cookie']
-        public:         true
+    defaults:
+        pagination_client_enabled:            true
```
teohhanhui: We should stop doing the weird indentation thing...

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -169,24 +169,24 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         $container->setParameter('api_platform.collection.exists_parameter_name', $config['collection']['exists_parameter_name']);
         $container->setParameter('api_platform.collection.order', $config['collection']['order']);
```
teohhanhui: This needs to be changed, no?

tests/Fixtures/app/config/config_common.yml
```diff
@@ -58,25 +58,25 @@ api_platform:
...
+        pagination_client_items_per_page:     true
+        pagination_client_partial:            true
+        pagination_items_per_page:            3
+        order_parameter_name:          'order'
```
teohhanhui: There's no such key supported in the defaults.

src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
```diff
@@ -73,6 +74,9 @@ public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGenerator
 
                 return;
             }
+            if (\is_string($defaultOrder)) {
+                $this->order = $defaultOrder;
```
teohhanhui: This looks wrong to me.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
soyuka: Indeed, not sure yet if we should deprecate these parameters?

tests/Fixtures/app/config/config_common.yml
soyuka: :+1: in fact the defaults are meant to be used on the `@ApiResource` (IIUC) and parameter names are supposed to be globally defined ?

src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
soyuka: agreed

tests/Fixtures/app/config/config_common.yml
teohhanhui: Parameter names are customizable per-filter, so indeed they have no place in the resource / operation attributes, I think...

soyuka:
Note that https://github.com/api-platform/core/pull/3472 will improve this

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -167,26 +167,26 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         $container->setParameter('api_platform.eager_loading.fetch_partial', $config['eager_loading']['fetch_partial']);
         $container->setParameter('api_platform.eager_loading.force_eager', $config['eager_loading']['force_eager']);
         $container->setParameter('api_platform.collection.exists_parameter_name', $config['collection']['exists_parameter_name']);
-        $container->setParameter('api_platform.collection.order', $config['collection']['order']);
+        $container->setParameter('api_platform.collection.order', $config['defaults']['order'] ?? $config['collection']['order']);
```
guilliamxavier, line 170: This change was wrong! `defaults.order` is supposed to be the global resource default for https://api-platform.com/docs/core/default-order/#overriding-default-order i.e. an array like `{"foo":"ASC"}` or `["foo", "bar"]`, not to be confused with `collection.order` that is a string `"ASC"` or `"DESC"`!

I first wrote this:

```yaml
api_platform:
    collection:
        order: DESC
    defaults:
        order: [createdAt, id]
```

but got this error:

> Argument 1 passed to ApiPlatform\\Core\\Bridge\\Doctrine\\Orm\\Extension\\OrderExtension::__construct() must be of the type string or null, array given

Fortunately I could work around by using the longer:

```yaml
api_platform:
    collection:
        order: DESC
    defaults:
        attributes:
            order: [createdAt, id]
```

but still...

Can this change be reverted as a bug fix or would it be a BC break now?

alanpoulain: IMO it's more a bugfix since this change is wrong.

guilliamxavier: Thanks, I opened #4170

PR #3283, opened 2019-11-21
added a new feature to scan folders where entities are located

flug:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -179,6 +179,7 @@ final class Configuration implements ConfigurationInterface
                         ->arrayNode('paths')
                             ->prototype('scalar')->end()
                         ->end()
+                        ->booleanNode('auto_mapping')->defaultFalse()->end()
                     ->end()
                 ->end()
                 ->arrayNode('resource_class_directories')
--- a/src/Bridge/Symfony/Bundle/Resources/config/metadata/directory.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/metadata/directory.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services">
+    <services>
+        <service id="api_platform.metadata.resource.name_collection_factory.directory" decorates="api_platform.metadata.resource.name_collection_factory"
+                 class="ApiPlatform\Core\Metadata\Resource\Factory\DirectoryResourceNameCollectionFactory" public="false">
+            <argument>%api_platform.resource_class_directories%</argument>
+            <argument type="service" id="api_platform.metadata.resource.name_collection_factory.directory.inner" />
+        </service>
+
+        <service id="api_platform.metadata.resource.metadata_factory.directory"
+                 decorates="api_platform.metadata.resource.metadata_factory" decoration-priority="40"
+                 class="ApiPlatform\Core\Metadata\Resource\Factory\DirectoryResourceMetadataFactory" public="false">
+            <argument>%api_platform.defaults%</argument>
+            <argument type="service" id="api_platform.metadata.resource.metadata_factory.directory.inner"/>
+        </service>
+    </services>
+</container>
--- a/src/Bridge/Symfony/Bundle/Resources/config/metadata/directory.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/metadata/directory.xml
@@ -9,7 +9,7 @@
         </service>
 
         <service id="api_platform.metadata.resource.metadata_factory.directory"
-                 decorates="api_platform.metadata.resource.metadata_factory" decoration-priority="40"
+                 decorates="api_platform.metadata.resource.metadata_factory" decoration-priority="5"
                  class="ApiPlatform\Core\Metadata\Resource\Factory\DirectoryResourceMetadataFactory" public="false">
             <argument>%api_platform.defaults%</argument>
             <argument type="service" id="api_platform.metadata.resource.metadata_factory.directory.inner"/>
--- a/src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
@@ -14,6 +14,7 @@
         <service id="ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface" alias="api_platform.metadata.resource.name_collection_factory" />
 
         <!-- Resource metadata -->
+
         <service id="api_platform.metadata.resource.metadata_factory.input_output" class="ApiPlatform\Core\Metadata\Resource\Factory\InputOutputResourceMetadataFactory" decorates="api_platform.metadata.resource.metadata_factory" decoration-priority="30" public="false">
             <argument type="service" id="api_platform.metadata.resource.metadata_factory.input_output.inner" />
         </service>
--- a/src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
@@ -38,7 +38,16 @@ final class DirectoryResourceMetadataFactory implements ResourceMetadataFactoryI
             } catch (ResourceClassNotFoundException $e) {
             }
         }
-        $resource = [] + $this->defaults;
+        $resource = [];
+        $resource['shortName'] = $this->defaults['shortName'] ?? (new \ReflectionClass($resourceClass))->getShortName();
+        $resource['description'] = $this->defaults['description'] ?? '';
+        $resource['iri'] = $this->defaults['iri'] ?? null;
+        $resource['itemOperations'] = $this->defaults['itemOperations'] ?? [];
+        $resource['collectionOperations'] = $this->defaults['collectionOperations'] ?? [];
+        $resource['subresourceOperations'] = $this->defaults['collectionOperations'] ?? [];
+        $resource['graphql'] = $this->defaults['graphql'] ?? null;
+        $resource['attributes'] = $this->defaults['attributes'] ?? [];
+
         return $this->update($parentResourceMetadata ?: new ResourceMetadata(), $resource);
     }
 
--- a/src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
@@ -38,23 +38,12 @@ final class DirectoryResourceMetadataFactory implements ResourceMetadataFactoryI
             } catch (ResourceClassNotFoundException $e) {
             }
         }
-        $resource['shortName'] = (new \ReflectionClass($resourceClass))->getShortName();
+        $resourceMetadata = $parentResourceMetadata ?: new ResourceMetadata();
 
-        return $this->update($parentResourceMetadata ?: new ResourceMetadata(), $resource);
-    }
-
-    /**
-     * Creates a new instance of metadata if the property is not already set.
-     */
-    private function update(ResourceMetadata $resourceMetadata, array $metadata): ResourceMetadata
-    {
-        foreach (['shortName'] as $property) {
-            if (null === $metadata[$property] || null !== $resourceMetadata->{'get'.ucfirst($property)}()) {
-                continue;
-            }
-            $resourceMetadata = $resourceMetadata->{'with'.ucfirst($property)}($metadata[$property]);
+        if (null !== $resourceMetadata->getShortname()) {
+            return $resourceMetadata;
         }
 
-        return $resourceMetadata;
+        return $resourceMetadata->withShortname((new \ReflectionClass($resourceClass))->getShortName());
     }
 }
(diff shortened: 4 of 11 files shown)


teohhanhui:
What is this for? We already support loading from directories, be it XML, YAML or annotations.

flug:
@soyuka wanted me to bring this feature to see with him and @dunglas

dunglas:
@teohhanhui Basically with this and `defaults`, you don't need the any annotations to expose a simple API, just pure PHP classes.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -177,8 +177,9 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->arrayNode('paths')
-                            ->prototype('scalar')->end()
-                        ->end()
+                        ->prototype('scalar')->end()
+                    ->end()
```
alanpoulain: Bad indentation.

src/Bridge/Symfony/Bundle/Resources/config/metadata/directory.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services">
+    <services>
+        <service id="api_platform.metadata.resource.name_collection_factory.directory" decorates="api_platform.metadata.resource.name_collection_factory"
+                 class="ApiPlatform\Core\Metadata\Resource\Factory\DirectoryResourceNameCollectionFactory" public="false">
+           <argument>%api_platform.resource_class_directories%</argument>
```
alanpoulain:
```suggestion
            <argument>%api_platform.resource_class_directories%</argument>
```

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Extractor;
+
+use ApiPlatform\Core\Util\ReflectionClassRecursiveIterator;
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\ORM\Mapping as ORM;
+use Psr\Container\ContainerInterface;
+
+class ClassExtractor extends AbstractExtractor
```
alanpoulain: final?

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+
+    public function __construct(array $paths, ContainerInterface $container = null, Reader $reader)
+    {
+        $this->reader = $reader;
+        parent::__construct($paths, $container);
+    }
+
+    protected function extractPath(string $path)
+    {
+        try {
```
alanpoulain: ?

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+{
+    private $reader;
+
+    public function __construct(array $paths, ContainerInterface $container = null, Reader $reader)
+    {
+        $this->reader = $reader;
+        parent::__construct($paths, $container);
+    }
+
+    protected function extractPath(string $path)
```
alanpoulain:
```suggestion
    protected function extractPath(string $path): void
```

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        try {
+        } catch (\InvalidArgumentException $e) {
+            throw new \InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+        }
+        if ('' === $path) {
+            return;
+        }
+        /**
+         * @var string
```
alanpoulain: To remove.

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+                'attributes' => null,
+                'identifier' => $this->isIdentifier($property),
+                'subresource' => null,
+            ];
+        }
+
+        return $properties;
+    }
+
+    private function extractDocBlock(\ReflectionProperty $reflectionProperty)
```
alanpoulain: Missing typehint.

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+        if (false === $reflectionProperty->getDocComment()) {
+            return [];
+        }
+        $pattern = "#(@[a-zA-Z]+\s*[a-zA-Z0-9, ()_].*)#";
+        preg_match_all($pattern, $reflectionProperty->getDocComment(), $matches, PREG_PATTERN_ORDER);
+
+        return $matches;
+    }
+
+    private function getVarDoc(\ReflectionProperty $reflectionProperty)
```
alanpoulain:
```suggestion
    private function getVarDoc(\ReflectionProperty $reflectionProperty): ?string
```

teohhanhui:
> you don't need the any annotations to expose a simple API, just pure PHP classes.

Shouldn't it use Symfony PropertyInfo then?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -177,8 +177,9 @@ public function getConfigTreeBuilder()
...
+                        ->booleanNode('scan_dir')->defaultFalse()->end()
```
teohhanhui: I don't think we should introduce a new config option. Instead, we should reuse `api_platform.mapping.paths`. Having so many similar options is confusing. We can chain this new extractor with the lowest priority. If necessary, we could provide a toggle to enable/disable this feature.

flug: You just want to detect the file? I agree with the option that it's potentially too much. After that if you want I can detect if there is the presence of @ApiResource and automatically disable this option. No ?

flug:
@teohhanhui  would you prefer that I use symfony PropertyInfo?

src/Metadata/Extractor/ClassExtractor.php
```diff
@@ -0,0 +1,136 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Metadata\Extractor;
+
+use ApiPlatform\Core\Util\ReflectionClassRecursiveIterator;
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\ORM\Mapping as ORM;
+use Psr\Container\ContainerInterface;
+
+final class ClassExtractor extends AbstractExtractor
```
soyuka: We shouldn't need this, discussed internally.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
teohhanhui: It's possible to use different extractors together, e.g. using annotations together with YAML / XML. That should not be a problem. We do not need to detect the `@ApiResource` annotation. What's important is to make sure that we **do not override** any metadata provided by the user.

src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
```diff
@@ -0,0 +1,56 @@
...
+
+use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
+use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
+
+final class DirectoryResourceMetadataFactory implements ResourceMetadataFactoryInterface
+{
+    private $propertyInfo;
+    private $defaults;
+
+    public function __construct(array $defaults = [], PropertyInfoExtractorInterface $propertyInfo)
```
soyuka: should follow the decorator pattern

src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
```diff
@@ -0,0 +1,56 @@
...
+    ): ResourceMetadata {
+        $properties = array_map(function (\ReflectionProperty $reflectionProperty) {
+            return $reflectionProperty->getName();
+        }, $reflectionClass->getProperties());
+        $attributes = (null === $properties && [] === $this->defaults['attributes']) ? null :
+            (array)
+            $properties + $this->defaults['attributes'];
+
+        return new ResourceMetadata(
+            $reflectionClass->getShortName(),
```
soyuka: Isn't that the responsibility of https://github.com/api-platform/core/blob/2f941fc3c0dc9dd257da3b34a6bb3f5c21c4230c/src/Metadata/Resource/Factory/ShortNameResourceMetadataFactory.php#L23 ?

src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return $this->createMetadata(new \ReflectionClass($resourceClass));
+    }
+
+    private function createMetadata(
+        \ReflectionClass $reflectionClass
+    ): ResourceMetadata {
+        $properties = array_map(function (\ReflectionProperty $reflectionProperty) {
+            return $reflectionProperty->getName();
+        }, $reflectionClass->getProperties());
```
soyuka: properties should be done in a PropertyNameCollectionMetadataFactory, I think it's already present with the PropertyInfo

tests/Metadata/Resource/Factory/DirectoryResourceFilterMetadataFactoryTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace ApiPlatform\Core\Tests\Metadata\Resource\Factory;
+
+use ApiPlatform\Core\Metadata\Resource\Factory\DirectoryResourceNameCollectionFactory;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceNameCollection;
+use ApiPlatform\Core\Util\ReflectionClassRecursiveIterator;
+use PHPUnit\Framework\TestCase;
+
+class DirectoryResourceFilterMetadataFactoryTest extends TestCase
```
soyuka: DirectoryResourceNameCollectionFactoryTest

src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
```diff
@@ -0,0 +1,77 @@
...
+    public function create(string $resourceClass): ResourceMetadata
+    {
+        $parentResourceMetadata = null;
+        if (null !== $this->decorated) {
+            try {
+                $parentResourceMetadata = $this->decorated->create($resourceClass);
+            } catch (ResourceClassNotFoundException $e) {
+            }
+        }
+        $resource['shortName'] = $resource['shortName'] ?? $this->defaults['shortName'] ?? null;
```
soyuka: I don't get where this array is instantiated, could be simplified maybe:
```
$resource = [] + $this->defaults;
```

src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
```diff
@@ -0,0 +1,60 @@
...
+
+        return $this->update($parentResourceMetadata ?: new ResourceMetadata(), $resource);
+    }
+
+    /**
+     * Creates a new instance of metadata if the property is not already set.
+     */
+    private function update(ResourceMetadata $resourceMetadata, array $metadata): ResourceMetadata
+    {
+        foreach (['shortName'] as $property) {
```
soyuka: do we really need this if there's one property? Let's simplify the code.

src/Metadata/Resource/Factory/DirectoryResourceMetadataFactory.php
```diff
@@ -0,0 +1,54 @@
...
+        $parentResourceMetadata = null;
+        if ($this->decorated) {
+            try {
+                $parentResourceMetadata = $this->decorated->create($resourceClass);
+            } catch (ResourceClassNotFoundException $e) {
+            }
+        }
+        $resource['shortName'] = (new \ReflectionClass($resourceClass))->getShortName();
+
+        return $this->update($parentResourceMetadata ?: new ResourceMetadata(), $resource);
```
soyuka:
```
$resourceMetadata = $parentResourceMetadata ?: new ResourceMetadata();

if (null !== $resourceMetadata->getShortname()) {
    return $resourceMetadata;
}

return $resourceMetadata->withShortname((new \ReflectionClass($resourceClass))->getShortName());
```

`update` method isn't required, as isn't the array

src/Bridge/Symfony/Bundle/Resources/config/metadata/metadata.xml
```diff
@@ -14,6 +14,7 @@
         <service id="ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface" alias="api_platform.metadata.resource.name_collection_factory" />
 
         <!-- Resource metadata -->
+
```
soyuka: empty line can be removed

PR #3285, opened 2019-11-22
Fix symfony 4.4 compat

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes
| Tickets       | https://github.com/symfony/symfony/issues/34509
| License       | MIT
| Doc PR        | 

If is not enough to check for the existence of `Symfony\Component\HttpKernel\EventListener\ErrorListener` to know if we can use it or not on symfony 4.4.

It may have been [removed from the container](https://github.com/symfony/symfony/blob/bfae515d521a421b413cb39e8fae727ffb4ca28b/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/ExceptionListenerPass.php#L37).

The new ErrorListener is functionally different to the deprecated one, as the new one [flattens exceptions with the KernelEvents::CONTROLLER_ARGUMENTS event](https://github.com/symfony/symfony/blob/bfae515d521a421b413cb39e8fae727ffb4ca28b/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php#L90-L106), and not when the [attributes are added](https://github.com/symfony/symfony/blob/bfae515d521a421b413cb39e8fae727ffb4ca28b/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php#L140).

If `ErrorListener` is removed from the container, the exception will never be converted to a flattened exception, causing the wrong Exception type to be passed to the [ExceptionAction](https://github.com/api-platform/core/blob/f4dfa2e86a9300e15c65a7c5acc9dab7975c6421/src/Action/ExceptionAction.php#L53).

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -17,7 +17,7 @@ use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 use Symfony\Component\HttpKernel\EventListener\ErrorListener;
-use Symfony\Component\HttpKernel\EventListener\ExceptionListener as BaseExceptionListener;
+use Symfony\Component\HttpKernel\EventListener\ExceptionListener as LegacyExceptionListener;
 
 /**
  * Handles requests errors.
@@ -34,7 +34,7 @@ final class ExceptionListener
         if (class_exists(ErrorListener::class)) {
             $this->exceptionListener = new ErrorListener($controller, $logger, $debug);
         } else {
-            $this->exceptionListener = new BaseExceptionListener($controller, $logger, $debug);
+            $this->exceptionListener = new LegacyExceptionListener($controller, $logger, $debug);
         }
     }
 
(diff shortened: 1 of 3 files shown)


bendavies:
btw, uses can fix this themselves by not using a deprecated twig configuration:
https://github.com/symfony/symfony/pull/34519

src/EventListener/ExceptionListener.php
```diff
@@ -29,12 +29,12 @@ final class ExceptionListener
 {
     private $exceptionListener;
 
-    public function __construct($controller, LoggerInterface $logger = null, $debug = false)
+    public function __construct($controller, LoggerInterface $logger = null, $debug = false, ErrorListener $errorListener = null)
     {
-        if (class_exists(ErrorListener::class)) {
+        if (null !== $errorListener) {
```
dunglas: Could you invert the test (test for `=== null`)

bendavies:
I have no idea why that `no deprecations` build is failing

alanpoulain:
It's not required, don't worry.

dunglas:
It's because FOSUser isn't compatible with Symfony 4.4 yet.

dunglas:
Thanks @bendavies

PR #3290, opened 2019-11-29
[MongoDB] Mercure support

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | ye
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Not really a new feature nor a bug fix. Mercure was not supported with MongoDB. This PR fixes it.
It seems it's also the case for the `PurgeHttpCacheListener`.


src/Bridge/Doctrine/Orm/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public function onFlush(EventArgs $eventArgs): void
+    {
+        if (!$eventArgs instanceof OnFlushEventArgs) {
+            return;
+        }
+
+        $uow = $eventArgs->getEntityManager()->getUnitOfWork();
+
+        foreach ($uow->getScheduledEntityInsertions() as $entity) {
```
dunglas: If the only difference between ODM and ORM is the method name, maybe could we merge all classes and do something like:

```php
$methodName = eventArgs instanceof OdmOnFlushEventArgs ? getScheduledDocumentInsertions : getScheduledEntityInsertions;

foreach ($uow->$methodName() //...
```

WDYT?
```

Comment (alanpoulain) [reply src/Bridge/Doctrine/Orm/EventListener/PublishMercureUpdatesListener.php]: I've no strong opinion about this. Calling a method dynamically is often considered a bad practice. And having a separate class allows to customize things easily without adding conditional statements.
But on the other hand it multiplies the number of classes and adds a complexity by having an inherited abstract class.

**Comment by alanpoulain** [comment]
You think it should target master? I thought we could consider it like a bug fix since we didn't say that Mercure was only for ORM.

**Comment by dunglas** [comment]
It's definitely a new feature :)
Regarding the dynamic call vs having 3 classes, I have no strong opinions either. I feel like having only one class will make the maintenance easier and the evolvability better, especially if we add new features at some points, such as JSON Patch support (partial updates).

**Comment by alanpoulain** [comment]
It targets master now. The code is merged inside one class too. I've added another commit in case we want to go back to the previous version (I think I prefer it).

**Comment by alanpoulain** [comment]
@dunglas are you OK with the current implementation?

File: src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_mercure_publisher.xml [hunk 2]
```diff
@@ -8,7 +8,7 @@
 
         <!-- Event listener -->
 
-        <service id="api_platform.doctrine.listener.mercure.publish" class="ApiPlatform\Core\Bridge\Doctrine\EventListener\PublishMercureUpdatesListener">
+        <service id="api_platform.doctrine.orm.listener.mercure.publish" class="ApiPlatform\Core\Bridge\Doctrine\EventListener\PublishMercureUpdatesListener">
```
dunglas: It's a BC break isn't it? We should at least add a (deprecated) alias to the old service name.

alanpoulain: Even if it's experimental?

chalasr: Please, yes. The class is marked as experimental but not the definition, and I know that Arte rely on this for their doctrine-less implementation.

dunglas: And it doesn't hurt. It's just one line to add, that will be dropped in 3.0.

PR #3298, opened 2019-12-03
fix key value relationship and embedded

flug:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no 
| New feature?  | yes/no 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixes #... 
| License       | MIT


flug:
okay I just did a test with this pr https://github.com/api-platform/core/pull/3283, well that actually fixes the problem, but "$readableLink" is still null

teohhanhui:
#3283 is not related to this bug. We do **not** want to turn these non-resource classes into resources.

teohhanhui:
See #3309

PR #3300, opened 2019-12-05
Throw exception when trying to (de)normalize an unreadable/unwritable property

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes-ish
| New feature?  | no-ish
| BC breaks?    | no-ish
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Currently, if the serialization mapping defines a property that is not readable/writable, API Platform silently ignores it when normalizing/denormalizing it. This PR changes this behavior to throw an exception instead to help spotting the configuration issue.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -325,13 +325,29 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         foreach ($propertyNames as $propertyName) {
             $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $propertyName, $options);
 
-            if (
-                $this->isAllowedAttribute($classOrObject, $propertyName, null, $context) &&
-                (
-                    isset($context['api_normalize']) && $propertyMetadata->isReadable() ||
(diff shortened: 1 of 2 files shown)


julienfalque:
Actually this is the wrong fix and it breaks everything. Maybe it should be fixed in `Symfony\Component\PropertyInfo\Extractor\SerializerExtractor` but I'm not sure it's even possible.

fmata:
@julienfalque to not break everything may be could you add a config such as : 

```yaml
api_platform:
    throw_on_invalid_property: true # defaults to false
```

WDYT ?

julienfalque:
Throwing an exception like I tried to do introduced a lot of regressions I didn't expect (see [PHPUnit](https://github.com/api-platform/core/runs/335412773) and [Behat](https://github.com/api-platform/core/runs/335412700)). Having an option to enable this behavior won't solve this issue. Also I think this behavior should be the default to help spotting invalid mappings.

Now I would say the issue is that `Symfony\Component\PropertyInfo\Extractor\SerializerExtractor` returns all mapped fields found in config files without checking whether they actually match a property (or some getter). But fixing it would require using Reflection API and maybe share a lot of logic with `Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor`. Not sure it's worth the hassle.

PR #3301, opened 2019-12-05
Fix serialization of related DTOs

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

When serializing a resource that contains a related resource that is configured to always use a DTO, using serialization groups on the DTO doesn't work unless the group is also defined for the related resource it represents:
```yaml
# api_resources.yaml
resources:
    Foo:
        # ...
    Bar:
        # ...
        output: BarDto
```
```yaml
# serialization.yaml
Foo:
    attributes:
        bar:
            groups: [foo]
BarDto:
    attributes:
        id:
            groups: [foo]

# this should not be needed
Bar:
    attributes:
        anyProperty:
            groups: [foo]
```

### DIFF
--- a/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\Exception\ResourceClassNotFoundException;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Serializer\InputOutputMetadataTrait;
 use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface as SerializerClassMetadataFactoryInterface;
 
@@ -28,6 +29,7 @@ use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface a
  */
 final class SerializerPropertyMetadataFactory implements PropertyMetadataFactoryInterface
 {
+    use InputOutputMetadataTrait;
     use ResourceClassInfoTrait;
 
     private $serializerClassMetadataFactory;
@@ -124,13 +126,15 @@ final class SerializerPropertyMetadataFactory implements PropertyMetadataFactory
             $relatedClass = $this->resourceClassResolver->getResourceClass(null, $relatedClass);
         }
 
-        $relatedGroups = $this->getClassSerializerGroups($relatedClass);
-
         if (null === $propertyMetadata->isReadableLink()) {
+            $relatedGroups = $this->getClassSerializerGroups($this->getOutputClass($relatedClass) ?? $relatedClass);
+
             $propertyMetadata = $propertyMetadata->withReadableLink(null !== $normalizationGroups && !empty(array_intersect($normalizationGroups, $relatedGroups)));
         }
 
         if (null === $propertyMetadata->isWritableLink()) {
+            $relatedGroups = $this->getClassSerializerGroups($this->getInputClass($relatedClass) ?? $relatedClass);
+
             $propertyMetadata = $propertyMetadata->withWritableLink(null !== $denormalizationGroups && !empty(array_intersect($denormalizationGroups, $relatedGroups)));
         }
 
--- a/src/Metadata/Resource/ToggleableOperationAttributeTrait.php
+++ b/src/Metadata/Resource/ToggleableOperationAttributeTrait.php
@@ -23,7 +23,7 @@ trait ToggleableOperationAttributeTrait
     /**
      * @var ResourceMetadataFactoryInterface|null
      */
-    private $resourceMetadataFactory;
+    protected $resourceMetadataFactory;
 
     private function isOperationAttributeDisabled(array $attributes, string $attribute, bool $default = false, bool $resourceFallback = true): bool
     {
--- a/src/Util/ResourceClassInfoTrait.php
+++ b/src/Util/ResourceClassInfoTrait.php
@@ -34,7 +34,7 @@ trait ResourceClassInfoTrait
     /**
      * @var ResourceMetadataFactoryInterface|null
      */
-    private $resourceMetadataFactory;
+    protected $resourceMetadataFactory;
 
     /**
      * Gets the resource class of the given object.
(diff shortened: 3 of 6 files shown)


src/Util/ResourceClassInfoTrait.php
```diff
@@ -34,7 +34,7 @@ trait ResourceClassInfoTrait
     /**
      * @var ResourceMetadataFactoryInterface|null
      */
-    private $resourceMetadataFactory;
+    protected $resourceMetadataFactory;
```
soyuka: why?

src/Metadata/Resource/ToggleableOperationAttributeTrait.php
```diff
@@ -23,7 +23,7 @@ trait ToggleableOperationAttributeTrait
     /**
      * @var ResourceMetadataFactoryInterface|null
      */
-    private $resourceMetadataFactory;
+    protected $resourceMetadataFactory;
```
soyuka: why?

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -124,12 +126,16 @@ private function transformLinkStatus(PropertyMetadata $propertyMetadata, array $
...
-        $relatedGroups = $this->getClassSerializerGroups($relatedClass);
+        $outputClass = $this->getOutputClass($relatedClass) ?? $relatedClass;
+        $relatedGroups = $this->getClassSerializerGroups($outputClass);
 
         if (null === $propertyMetadata->isReadableLink()) {
             $propertyMetadata = $propertyMetadata->withReadableLink(null !== $normalizationGroups && !empty(array_intersect($normalizationGroups, $relatedGroups)));
         }
 
+        $inputClass = $this->getInputClass($relatedClass) ?? $relatedClass;
+        $relatedGroups = $this->getClassSerializerGroups($inputClass);
```
soyuka: are `$outputClass` and `$inputClass` used? Not sure we need a variable for this?

src/Util/ResourceClassInfoTrait.php
julienfalque: Because of property conflicts with trait composition: sometimes the property was `protected`, sometimes `private`. PHP doesn't allow that and does not provide a way to resolve conflicts like it does for methods.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
julienfalque: I inlined the variables.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -124,12 +126,18 @@ private function transformLinkStatus(PropertyMetadata $propertyMetadata, array $
             $relatedClass = $this->resourceClassResolver->getResourceClass(null, $relatedClass);
         }
 
-        $relatedGroups = $this->getClassSerializerGroups($relatedClass);
+        $relatedGroups = $this->getClassSerializerGroups(
```
teohhanhui: CS: Don't split across multiple lines.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -124,12 +126,18 @@ private function transformLinkStatus(PropertyMetadata $propertyMetadata, array $
...
+            $this->getOutputClass($relatedClass) ?? $relatedClass
+        );
 
         if (null === $propertyMetadata->isReadableLink()) {
             $propertyMetadata = $propertyMetadata->withReadableLink(null !== $normalizationGroups && !empty(array_intersect($normalizationGroups, $relatedGroups)));
         }
 
+        $relatedGroups = $this->getClassSerializerGroups(
```
teohhanhui: CS: Don't split across multiple lines.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -29,6 +30,7 @@
 final class SerializerPropertyMetadataFactory implements PropertyMetadataFactoryInterface
 {
     use ResourceClassInfoTrait;
+    use InputOutputMetadataTrait;
```
teohhanhui: CS: Sort alphabetically.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -124,12 +126,18 @@ private function transformLinkStatus(PropertyMetadata $propertyMetadata, array $
             $relatedClass = $this->resourceClassResolver->getResourceClass(null, $relatedClass);
         }
 
-        $relatedGroups = $this->getClassSerializerGroups($relatedClass);
+        $relatedGroups = $this->getClassSerializerGroups(
+            $this->getOutputClass($relatedClass) ?? $relatedClass
```
teohhanhui: Hmm... There is a problem here. `getOutputClass` / `getInputClass` do not respect overrides being set at the operation level. @soyuka Is there a reason why we're not using `getOperationAttribute` instead?

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
julienfalque: Done.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
julienfalque: Done.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
julienfalque: Done.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
julienfalque: Do we really care about operations of the related class here? The operation that matters is the one that led to seriliazing the root object, not the embedded one (`$relatedClass`), which means we only care about whether the embedded object has a global input/output representation, am I right?

teohhanhui:
Okay, re-reading the OP:

> related resource that is configured to always use a DTO

It seems to me that this PR is invalid. An input/output class is only used at the top level. So this is working as intended and not a bug.

teohhanhui:
> unless the group is also defined for the related resource it represents

This is not supposed to work either. We should throw an exception if we're expecting `Bar` but got `BarDto`.

julienfalque:
So output DTOs cannot be applied globally? Is there an alternative?

julienfalque:
Answering my own question: using a custom serialization normalizer is probably the way to go to change the representation of a resource globally.

If so, doesn't it mean that their is still something to fix? Currently API Platform is kind of sitting on the fence when serialiazing a nested resource:
* it always uses the global `output` attribute if defined instead of the resource class;
* but it always ignores the serialization groups for the output class (which is why I needed groups on both the resource and the output class in my example above)

This PR aimed to change the second point, but if this is not valid, shouldn't the first point be changed then?

julienfalque:
@teohhanhui Sorry to ping you but what do you think about my previous comment?

PR #3304, opened 2019-12-09
PublishMercureUpdatesListener unable to add custom updates to publish

gomcodoctor:
I wanted to add custom publish based on different logic ( not for all updates ), there was no way to extend PublishMercureUpdatesListener so I created this commit, following changes are done.

1. Entities to publish is moved to different service.
2. Service ( 1 ) is added as dependency  to PublishMercureUpdatesListener

| Q             | A
| ------------- | ---
| Bug fix?      |  No
| New feature?  | yes
| BC breaks?    | no
| License       | MIT


teohhanhui:
You can use your own Doctrine event listener / subscriber. This is not designed to be extensible.

gomcodoctor:
Yes we can use own listener but it required 90% code copying from this listener

Issue #3116, opened 2019-09-28
GraphQl: interface feature implementation


CvekCoding:
Q | A
-- | --
Bug fix? | no
New feature? | yes
BC breaks? | probably
Deprecations? | no
Tests pass? | no
Fixed tickets |  
License | MIT
Doc PR |  

I started the work on implementation of interfaces. The feature is in draft version now and works this way:
1. I've added 2 properties in `ApiResource`: bool `interface` (false by default) and array `implements`.
2. If some resource's `interface` property is `true`, then I create new type as an instance of `InterfaceType` and add postfix `Interface`. E.g. `DocumentInterface`.
3. If some resource has property `implements` which is array of FQN resources, then I validate them and add here https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/GraphQl/Type/TypeBuilder.php#L124

This works fine except one thing. See here: https://github.com/api-platform/core/blob/9799193466b0059fd82254798e9a06f32d8ef2f4/src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php#L54-L58

What happens here: any parent resource receives all the properties of his **descendants**. May be this is good until we have interfaces, but this is definitely wrong when we start using Interfaces. `InterfaceType` must contain only his own (and his ancestor's) properties but not his descendants.

See example:
```
/**
 * @ApiResource(interface=true)
 */
class Parent
{
 public parentProperty;
}

/**
 * @ApiResource(implements={"Parent"})
 */
class Child extends Parent
{
 public childProperty;
}
```

`InheritedPropertyNameCollectionFactory` makes `Parent` to have `childProperty`. Which is wrong if we want to use interfaces and query common and specific fields this way:
```
{
  parents {
    parentProperty
    ... on Child {
      childProperty
    }
  }
}
```
To solve the issue I can develop work around (e.g. implement `PropertyNameCollectionFactoryInterface` and use it for resources with new feature, but for all other resources leave the current PropertyFactory to not create BC).

But may be I'm wrong and you can advise me the better way how to implement `interface` feature to avoid the problem explained above?

alanpoulain:
I wonder if it's a good idea to use a property of `ApiResource`. Why not using a PHP interface? They are already managed by API Platform. And you would not have the problem you describe.

CvekCoding:
[...]

PR #3306, opened 2019-12-09
GraphQl: interface implementation

CvekCoding:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #3116
| License       | MIT
| Doc PR        | 

Added interface feature to `Graphql`.

How to use:
```
/**
 * @ApiResource(isInterface=true)
*/
abstract class AA {}

/**
 * @ApiResource(isInterface=true)
*/
abstract class AB {}

/**
 * @ApiResource(
 *     implements={AA::class},
 * )
*/
class A extends AA {}

/**
 * @ApiResource(
 *     implements={AA::class, AB::class},
 * )
*/
class C {}
```

In a scheme you will see that:
- `AAInterface` has two implementations: `A` and `C`;
- `ABInterface` has one implementations: `C`;
- `A` implements one interface `AAInterface`;
- `C` implements two interfaces `AAInterface` and `ABInterface`.

I'm far from thought that my implementation is ready to be merged. Feature was developed in haste because I needed it in my project, but I didnt have time to think about the best architecture.

I'm ready to answer the questions.

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -117,6 +117,16 @@ final class ApiResource
     public $subresourceOperations;
 
     /**
+     * @var bool
+     */
+    public $isInterface;
+
+    /**
+     * @var array
+     */
+    public $implements;
+
+    /**
      * @see https://api-platform.com/docs/core/performance/#setting-custom-http-cache-headers
      * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
      *
--- a/src/GraphQl/Type/SchemaBuilder.php
+++ b/src/GraphQl/Type/SchemaBuilder.php
@@ -72,6 +72,10 @@ final class SchemaBuilder implements SchemaBuilderInterface
                     continue;
                 }
 
+                if ($resourceMetadata->isInterface()) {
+                    continue;
+                }
+
                 if ($resourceMetadata->getGraphqlAttribute($operationName, 'item_query')) {
                     $queryFields += $this->fieldsBuilder->getItemQueryFields($resourceClass, $resourceMetadata, $operationName, $value);
 
--- a/src/GraphQl/Type/TypeBuilder.php
+++ b/src/GraphQl/Type/TypeBuilder.php
@@ -72,7 +72,7 @@ final class TypeBuilder implements TypeBuilderInterface
             $shortName .= self::INTERFACE_POSTFIX;
         }
 
-        if (('item_query' === $queryName || 'collection_query' === $queryName)
+        if (!$resourceMetadata->isInterface() && ('item_query' === $queryName || 'collection_query' === $queryName)
             && $resourceMetadata->getGraphqlAttribute('item_query', 'normalization_context', [], true) !== $resourceMetadata->getGraphqlAttribute('collection_query', 'normalization_context', [], true)) {
             if ('item_query' === $queryName) {
                 $shortName .= self::ITEM_POSTFIX;
@@ -213,8 +213,8 @@ final class TypeBuilder implements TypeBuilderInterface
         }
 
         $wrapData = !$wrapped && null !== $mutationName && !$input && $depth < 1;
-        $interfaces = ($interface = $resourceMetadata->getImplements())
-            ? $this->getInterfaceTypes($interface)
+        $interfaceTypes = ($interfaces = $resourceMetadata->getImplements())
+            ? $this->getInterfaceTypes($interfaces, $queryName === 'collection_query')
             : [];
 
         $configuration = [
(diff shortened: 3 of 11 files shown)


lukasluecke:
@CvekCoding I think this should target `master`, and maybe have a more descriptive title 🙂 

@alanpoulain Do you think you could take a look at this? I think it would be a big step forward for our GraphQL-support 🚀

src/GraphQl/Type/SchemaBuilder.php
```diff
@@ -72,6 +72,10 @@ public function getSchema(): Schema
                     continue;
                 }
 
+                if ($resourceMetadata->isInterface()) {
+                    continue;
+                }
+
```
lukasluecke: I am not sure skipping the operations for interfaces is the correct approach. And if we do, we should at least trigger warnings or something if they are defined anyway.

src/Annotation/ApiResource.php
```diff
@@ -116,6 +116,16 @@ final class ApiResource
...
+    /**
+     * @var bool
+     */
+    public $isInterface;
+
+    /**
+     * @var array
+     */
+    public $implements;
+
```
lukasluecke: This should probably be under the `graphql`-key, as this is only supported for the GraphQL format / endpoints.

src/GraphQl/Type/TypeBuilder.php
```diff
@@ -226,4 +205,141 @@ public function isCollection(Type $type): bool
...
+            'resolveType' => function ($value, $context, $info) {
+                if (!isset($value[ItemNormalizer::ITEM_RESOURCE_CLASS_KEY])) {
+                    throw new \UnexpectedValueException('Resource class was not passed. Interface type can not be used.');
+                }
+
+                $shortName = (new \ReflectionClass($value[ItemNormalizer::ITEM_RESOURCE_CLASS_KEY]))->getShortName();
+
+                if (!$this->typesContainer->has($shortName)) {
+                    $shortName .= self::ITEM_POSTFIX;
+                    if (!$this->typesContainer->has($shortName)) {
```
lukasluecke: This should probably not be used as a fallback, but instead implement the proper checks (i.e. use `Item` postfix, if the serialization context differs between `Item` and `Collection` queries)

src/GraphQl/Type/SchemaBuilder.php
CvekCoding: Weak place, agree.
I did this because in my project I dont need mutations in interfaces. And I use custom operations very few.
Additionally I think I had some issues with mutations in interface types, but dont remember details now :(

We can safely move this conditions to include custom query operations here https://github.com/api-platform/core/blob/a1570c836e13d42c9a065f49847e2887410c3c16/src/GraphQl/Type/SchemaBuilder.php#L90

src/Annotation/ApiResource.php
CvekCoding: @lukasluecke can you explain `graphql-key` - what does it mean? Prefix for property name?

lukasluecke: @CvekCoding There is already a configuration key `graphql`, under which we configure the operations etc. - I think it would make sense to include it inside of that, instead of at the root level.

lukasluecke:
@alanpoulain What do you think about this? I've been using it in my project for some time now and it's working quite well.

I think we should address both the name resolution handling (based on serialization context) and the resource config layout (under `graphql` key), and also simplify this PR so it can be merged with #3321 better (i.e. improve the `buildResourceObjectType` extraction).

Handling the other operations for interface resources could be addressed in a follow-up PR I think.

alanpoulain:
I plan to take some time to take a look at it once subscriptions are merged.

ollietb:
It would be great to see this merged!

lukasluecke:
@alanpoulain Any chance to get this revisited? I'd be happy to help and contribute, but would be good to have your input first 🙂

ollietb:
@lukasluecke I can't speak for @alanpoulain but I would imagine this PR will at least need to have tests for it to be considered for merging

lukasluecke:
@ollietb Of course, but he said he would take a look - not merge it 🙂 this should just serve as an inspiration for a "proper" implementation in my mind, we would first need to define what behavior we want to have, before writing any tests.

ollietb:
It would be good to see this PR finished, my front-end developers are always complaining about missing features in the GraphQL spec, like interfaces and enums. @alanpoulain would you be able to look at this PR and advise @lukasluecke if he is on the right path before he writes some tests and submits the completed PR?

PR #3307, opened 2019-12-10
Improve DX when missing filter class import

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no?
| Deprecations? | no
| Tickets       | - 
| License       | MIT
| Doc PR        | -

Improve DX to get more helpful message when missing import for filter class.

Currently we get:
`The filter class "SearchFilter" does not implement "ApiPlatform\Core\Api\FilterInterface". Did you forget a use statement?`

After this we'd get:

`The filter class "SearchFilter" does not implement "ApiPlatform\Core\Api\FilterInterface" in "ApiPlatform\Core\Tests\Fixtures\DummyEntityInvalidFilter". Did you forget a use statement?`

But this changes behavior a little bit, because now when constructing ApiFilter annotation with invalid class it won't throw directly.

### DIFF
--- a/src/Annotation/ApiFilter.php
+++ b/src/Annotation/ApiFilter.php
@@ -57,10 +57,6 @@ final class ApiFilter
             throw new InvalidArgumentException('This annotation needs a value representing the filter class.');
         }
 
-        if (!is_a($options['value'], FilterInterface::class, true)) {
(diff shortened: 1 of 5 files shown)


soyuka:
I see, I agree that the message is clearer. I'd be :+1: to merge this as a new feature, may you check why the tests are failing though?

PR #1128, opened 2017-05-18
specify left join for filter associations if there was a previous left join

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/api-platform/core/pull/789#issuecomment-302324794
| License       | MIT
| Doc PR        | n/a

see https://github.com/api-platform/core/pull/789#issuecomment-296218102

PR #3308, opened 2019-12-10
Fix logic of protection against adding same join multiple times

klamparski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Hello,

In project which I'm involved we found problems with filtering. Our custom full text filter is filtering by nested properties of associations thus it is trying to add LEFT JOIN.
It works correctly but not in all cases. If before our full text filter any other filter (or extension) is applied and it adds INNER JOIN for association, then our LEFT JOIN is ignored because `QueryBuilderHelper::getExistingJoin` does not take into account the type of the JOIN and simply returns true if join for this association already exists.

Such behaviour makes filters not reliable, because they can give different results depending on apply ordering.

Moreover in `QueryBuilderHelper::addJoinOnce` there was also condition which makes all joins type LEFT if there exist already any LEFT join in the query. It also makes filters not reliable. This PR also removes that condition.


klamparski:
I see now that my changes make test `\ApiPlatform\Core\Tests\Bridge\Doctrine\Orm\Filter\SearchFilterTest::doTestJoinLeft` fails. Sorry for that.

Is that really expected that if any LEFT JOIN in query exists then all next joins from filters will be also LEFT? What is the reason for that?

soyuka:
> Is that really expected that if any LEFT JOIN in query exists then all next joins from filters will be also LEFT? What is the reason for that?

Because this usually comes from the fact that the relation is nullable. If you're adding an `INNER JOIN` for example, you won't get the same results as expected as it'll "filter out" every null value.

teohhanhui:
> Is that really expected that if any LEFT JOIN in query exists then all next joins from filters will be also LEFT? What is the reason for that?

I think that's correct. Doing an `INNER JOIN` after a `LEFT JOIN` would give incorrect results, as the previous assumption would no longer be correct. This fix was done a long time ago in https://github.com/api-platform/core/pull/1128

alanpoulain:
It seems this fix is not relevant. I'm closing this PR.
Feel free to open a new PR with another fix.

PR #3309, opened 2019-12-11
Fix JSON Schema generation for non-resource class

### LINKED ISSUE #1787: Swagger Documentation of Doctrine Embeddables
Hi,

Not a bug probably but I just can't wrap my head around this one. I've read like 50 issues about embedded objects but nothing seems to solve my issue.

I have an `@ORM\Embeddable` that's used quite a lot in other real resources. Serialization/deserialization seem to work just fine. But the swagger documentation is giving me headache.

I can of course do something like:

    /*
     * @ORM\Embedded(class="App\Entity\Id")
     * @ApiProperty(
     *  attributes={
     *   "swagger_context"={
     *     "type"="object",
     *     "properties"= {
     *        "type" = "int",
     *        "code" = "string",
     *     },
     *     "example"={"type":1, "code":"1234567-1"}
     *   }
     *  }
     * )
     *
     * /
    private $embeddedThing;
But that seems really counterproductive when I have multiple users for the embeddable. I just wouldn't like to maintain it like this.

Couple of solution possibilities comes to mind. The optimal solution would be to pull the embedded entity as a swagger schema automatically and referencing it in the parent.

I could live with something like this:

     * @ApiProperty(
     *  attributes={
     *   "swagger_context"={
     *     "$ref"="#/definitions/Id"
     *   }
     *  }
     * )
But I just can't seem to figure out how to make API Platform create that model for my Id embeddable.

Any ideas how to approach this? Can I somehow force the model creation from Embeddables? Any workaround other than maintaining the embeddable documentation manually?

### LINKED ISSUE #2818: How to generate docs for subresource of input DTO?
I want to receive the following json:

```
{
   "flat": "string",
   "object": {
      "key": "string"
   }
}
```

I created `DtoInput` class

```
   public $flat;
   /**
    * @var ObjectModel
    */
   public $object;
   public function setFlat(string $flat) { ... }
   public function setObject(ObjectModel $flat) { ... }
```

and class `ObjectModel` 

```
   public $key;
   public function setKey(string $key) { ... }
```

but on link `/api` i can see model

{
  "flat": "string",
  "object": "string",
}

I want to see

```
{
   "flat": "string",
   "object": {
      "key": "string"
   }
}
```

I was tried to use @Groups from `Symfony\Component\Serializer\Annotation\Groups` and `denormalizationContext` but without success. I think problem can be connected with DTO because of without DTO I configured `denormalizationContext` correctly.

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #1787, fixes #2818, replaces #3298
| License       | MIT
| Doc PR        | N/A

TODO:

- [ ] ~Fix tests~ There were no broken tests(!)
- [x] Add tests
- [ ] More tests perhaps?

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml
@@ -8,6 +8,7 @@
         <defaults public="false" />
 
         <service id="api_platform.json_schema.type_factory" class="ApiPlatform\Core\JsonSchema\TypeFactory">
+            <argument type="service" id="api_platform.resource_class_resolver" />
             <call method="setSchemaFactory">
                 <argument type="service" id="api_platform.json_schema.schema_factory"/>
             </call>
@@ -20,6 +21,7 @@
             <argument type="service" id="api_platform.metadata.property.name_collection_factory" />
             <argument type="service" id="api_platform.metadata.property.metadata_factory" />
             <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
+            <argument type="service" id="api_platform.resource_class_resolver" />
         </service>
         <service id="ApiPlatform\Core\JsonSchema\SchemaFactoryInterface" alias="api_platform.json_schema.schema_factory" />
 
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -18,7 +18,7 @@ use ApiPlatform\Core\JsonSchema\SchemaFactory as BaseSchemaFactory;
 use ApiPlatform\Core\JsonSchema\SchemaFactoryInterface;
 
 /**
- * Generates the JSON Schema corresponding to a Hydra document.
+ * Decorator factory which adds Hydra properties to the JSON Schema document.
  *
  * @experimental
  *
@@ -47,9 +47,9 @@ final class SchemaFactory implements SchemaFactoryInterface
     /**
      * {@inheritdoc}
      */
-    public function buildSchema(string $resourceClass, string $format = 'jsonld', string $type = Schema::TYPE_OUTPUT, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
+    public function buildSchema(string $className, string $format = 'jsonld', string $type = Schema::TYPE_OUTPUT, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
     {
-        $schema = $this->schemaFactory->buildSchema($resourceClass, $format, $type, $operationType, $operationName, $schema, $serializerContext, $forceCollection);
+        $schema = $this->schemaFactory->buildSchema($className, $format, $type, $operationType, $operationName, $schema, $serializerContext, $forceCollection);
         if ('jsonld' !== $format) {
             return $schema;
         }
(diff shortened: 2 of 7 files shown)


teohhanhui:
AppVeyor failure seems like an AppVeyor bug. We should consider moving the Windows jobs to GitHub Actions.

julienfalque:
Does this fix #2818?

teohhanhui:
Yes, it'll fix #2818

soyuka:
Nice one, will you add some tests then? I guess you picked some of the work of @flug ?

soyuka:
Can you rebase?

teohhanhui:
Rebased. Sorry for the delay!

fgamess:
Hi, 

I think that this PR does not fully solve the issue. When [Using Objects As Relations Inside Resources](https://api-platform.com/docs/core/dto/#using-objects-as-relations-inside-resources) which means the object is not a resource, we have indeed the doc generated for those object but the **examples** values set in the **properties** of the **Resource*** are not rendered. We can see **string** as the value for all the fields of the **non-resource object**. When configuring the object as a **resource** then this does not happen. 

**Possible solutions:**

- Support the creation of Swagger custom [components](https://swagger.io/docs/specification/components/). It could eventually override the **generated** schemas for the ** non-resource object** in the Swagger UI.
- Fill the **example values** in the **generated** schemas for the ** non-resource object** with the examples from the properties of the **Resource**. If I have a **resource MyResource** related to a **non-resource MyNonResource** it should be possible to describe the properties of **MyResource** with examples, including the **myNonResource** property and then see all the example values displayed in both **MyResource** and **MyNonResource** schemas. Also, the schemas generated for the serialization groups



@teohhanhui @dunglas @soyuka @flug

teohhanhui:
@fgamess Thanks for the very detailed report, but could you please open a new issue? Thanks.

fgamess:
@teohhanhui Shall I just copy paste the message above and create an issue from it?

fgamess:
@teohhanhui the issue has been created [3385](https://github.com/api-platform/core/issues/3385)

mab05k:
This broke me :( 
I have a resource that is using Traits, it no longer resolves namespaces from these traits properly.

```php
namespace App;
use SomeTrait;
class Transaction
{
    use SomeTrait;
}
```
```php
namespace App\Traits;
use App\Data\Price;
trait SomeTrait
{
    /**
     * @var Price
    */
    public $price;
}
```
With the above I get ReflectionException with error messages like
`class App\Transaction\Price not found` instead of properly resolving the `Price` namespace to `App\Data\Price`

Any ideas on workarounds?

fgamess:
@mab05k could you try 

```php
namespace App;

use App\Traits\SomeTrait;

class Transaction
{
    use SomeTrait;
}
```
Seems that inside Transaction class, you put a wrong use statement for SomeTrait. `use App\SomeTrait` instead of `App\Traits\SomeTraits`. 

Also, I don't think your issue is related to the issue here but tell me if what I suggested does fix your problem.

teohhanhui:
@mab05k Your issue seems unrelated. Please ask support questions on the Symfony Slack channel. Thanks!

mab05k:
@teohhanhui thanks.  One other note I didn't add in my last message: I did verify that there was no issue with v2.5.3. this problem arose when upgrading to v2.5.4. Both using symfony 4.4.4.

teohhanhui:
Feel free to open a new issue if you think you've found a bug. :smile:

teohhanhui:
@mab05k Looks like your issue is the same as #3349. Unfortunately the bug is upstream and not in API Platform.

mab05k:
Thanks for taking a look! I haven't gotten a chance to look further into it.

PR #3311, opened 2019-12-17
Check `pagination_use_output_walkers` attribute for graphQl operation

### LINKED ISSUE #3305: Bug: use output walkers in Doctrine bridge
**API Platform version(s) affected**: 2.5

**Description**  
Wrong recognition whether the Doctrine ORM Paginator should use output walkers.
Previously (in APIP2.4) it worked fine, but after recent code changes it's broken.

**How to reproduce**  
In my custom filter I use the following code:
```
$queryBuilder->addSelect(sprintf('CASE WHEN %s.%s = FALSE THEN 0 ELSE 1 END AS HIDDEN %s', $alias, 'showOnGrid', $nullRankHiddenField));
```
It's necessary to add this expression into `OrderBy` clause.

In `QueryChecker::hasOrderByOnFetchJoinedToManyAssociation()` there is a following logic:
```
            foreach (QueryBuilderHelper::traverseJoins($orderByAlias, $queryBuilder, $managerRegistry) as $alias => [$metadata, $association]) {
                if ($inToManyContext && \in_array($alias, $selectAliases, true)) {
                    return true;
                }

                if (null !== $association && $metadata->isCollectionValuedAssociation($association)) {
                    $inToManyContext = true;
                }
            }
```
As you can see, you check `join` name against `selectAlias`, but `selectAlias` is smth like "case ..." and the code above will never return `true`.

**Possible Solution**  
As this moment I reverted back the previous version of this method.
**Additional Context**

CvekCoding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3305
| License       | MIT
| Doc PR        |

In a case if we use graphql, we have to check attribute `pagination_use_output_walkers` in graphQl operations too.

I feel that this part can be improved - we have to know the operation type - is it `Rest` or `Graphql`, because I think this is not good to check this attribute in all possible sources blindly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.6.4
 
 * Doctrine: Fix purging HTTP cache for unreadable relations (#3441)
+* GraphQL: Manage `pagination_use_output_walkers` and `pagination_fetch_join_collection` for operations (#3311)
 * Swagger UI: Remove Google fonts (#4112)
 
 ## 2.6.3
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -164,8 +164,8 @@ final class PaginationExtension implements ContextAwareQueryResultCollectionExte
             $query->setHint(CountWalker::HINT_DISTINCT, false);
         }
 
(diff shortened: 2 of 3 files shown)


alanpoulain:
Thanks @CvekCoding.

PR #3312, opened 2019-12-18
[OpenAPI] Add nullable property description

kayneth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

While trying to generate a client with [Jane](https://github.com/janephp/janephp), I encountered a missing piece of OpenAPI 3.0 for [nullable properties](https://swagger.io/docs/specification/data-models/data-types/#null). It's useful in order to generate models with nullable getters.

As stated on `src/JsonSchema/SchemaFactory.php:173`: "externalDocs is an OpenAPI specific extension, but JSON Schema allows additional keys, so we always add it" for the property schema. 
Then we could also add the nullable feature. 

Another take, would be to considered each components schema non `required` properties as optional. Therefore, I should make a PR to Jane in order to use non-required properties instead of nullable properties to generate models. But [this is not a desired behaviour on their side](https://github.com/janephp/janephp/issues/100#issuecomment-562044344).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * GraphQL: Allow to format GraphQL errors based on exceptions (#3063)
 * GraphQL: Add page-based pagination (#3175)
 * OpenAPI: Add PHP default values to the documentation (#2386)
+* OpenAPI: Add nullable schema components properties
  
 ## 2.5.3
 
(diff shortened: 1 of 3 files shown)


kayneth:
Ok, done. Thank you

alanpoulain:
Done by https://github.com/api-platform/core/pull/3402.

Issue #3206, opened 2019-10-24
Mercure: Publish updates to collections (addition/removals)


Richard87:
**API Platform version(s) affected**: 2.5.0

**Description**  
API-Platform doesnt publish updates when a collection is updated/changed

**How to reproduce**  
Add or remove a entity that belons in a collection

**Possible Solution**  
Add these lines to `PublishMercureUpdateListener`:
```php
/** @var PersistentCollection $collection */
foreach ($uow->getScheduledCollectionDeletions() as $collection) {
    $this->storeEntityToPublish($collection->getOwner(), 'updatedEntities');
}

/** @var PersistentCollection $collection */
foreach ($uow->getScheduledCollectionUpdates() as $collection) {
    $this->storeEntityToPublish($collection->getOwner(), 'updatedEntities');
}
``` 

**Additional Context**  
If a new Collection (owner) is created, this will publish duplicate events, the "item created", and "item updated" (for collections)

Original issue: https://github.com/symfony/mercure/issues/13

My use case is a React app, with a Group component, that uses the list if items, and passes on the item id to the child... I a new Child is added to a group, it's not beeing pushed to the client since the Group is not updated even if its collection is updated....

lukasluecke:
I think it makes sense to add this - the issue with duplicate events shouldn't matter, since the subscriber can just choose to ignore certain topics / events in these cases.

Richard87:
It's not a big issue, but it would be nice to avoid it...

Instead of having 3 different change-lists, there should be 2, 'regular-changes', and deleted changes...
Process the deleted changes first, and if there is a update in 'regular-changes', remove the entity from that list, then process the remaining changes (I feel it should make the code cleaner too, but I'll think about it...)

Richard87:
A new problem have come up... Mercure is spammed with updates because a collection is updated. But that collection is not in use in front-end, and is not serialized by API-Platform...

Which solution would be best?

1. Filter the collection updates based on context and groups on the property?
2. Create a new `@mercure` annotation for the collection property, to signal API-Platform to push updates on it
3. Add a `groups` property on the existing `@mercure` annotation to specify which serialization fields should be watched for changes.

I prefer 2, a new `@mercure` annotation for the collection that should be watched, but a `groups` property could also work nicely, and be easy to start using...

Richard87:
After thinking about it, and noticing Mercure already uses the class `normalization_context`, I think using that to check the changes is optimal.

[...]

PR #3321, opened 2019-12-20
[GraphQL] Mercure Subscription Support

### LINKED ISSUE #3124: Add Mercure support for GraphQL subscriptions
Currently, API Platform supports Mercure for REST (https://api-platform.com/docs/core/mercure/), but not for GraphQL.
It could be done as described here: https://github.com/dunglas/mercure#how-to-use-mercure-with-graphql

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3124 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1034

[GraphQL Subscriptions](https://github.com/graphql/graphql-spec/blob/master/rfcs/Subscriptions.md) in API Platform will use [Mercure](https://github.com/dunglas/mercure) as the underlying protocol (for the first version) with this PR.

Only the **update** subscription is available. Maybe the delete subscription would be useful too but it could be done later.

A subscription can be done like this:
```graphql
subscription {
  updateDummyMercureSubscribe(input: {id: "/dummy_mercures/1", clientSubscriptionId: "myId"}) {
    dummyMercure {
      id
      name
      relatedDummy {
        name
      }
    }
    mercureUrl
    clientSubscriptionId
  }
}
```

In the JSON response (in `mercureUrl` field), you will receive an URL like this:
`https://demo.mercure.rocks/hub?topic=http://example.com/subscriptions/fc806a60cca4e93544d555912ce8b549`

This URL is used by the Mercure protocol to subscribe to updates by using [Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events):
```js
const eventSource = new EventSource(mercureUrl);

// The callback will be called every time an update is published
eventSource.onmessage = e => console.log(e); // do something with the payload
```

To enable subscriptions for a resource, follow the [related Mercure documentation](https://api-platform.com/docs/core/mercure). It mainly consists of adding this annotation:
```php

/**
 * @ApiResource(mercure=true)
 */
class Book
{
    // ...
}
```

You will only receive the update if the payload you have given when subscribing has changed.
To do so, a [Symfony cache](https://symfony.com/doc/current/cache.html) is used to store the subscriptions and their payloads:
`api_platform.graphql.cache.subscription`.
The best way to store the subscriptions is to use an adapter like Redis.

An Apollo transport like this one: https://github.com/CodeCommission/subscriptions-transport-sse could be written later in order to make it work seamlessly with Apollo.

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -19,12 +19,15 @@ use ApiPlatform\Core\Api\UrlGeneratorInterface;
 use ApiPlatform\Core\Bridge\Symfony\Messenger\DispatchTrait;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Core\GraphQl\Subscription\MercureSubscriptionIriGeneratorInterface as GraphQlMercureSubscriptionIriGeneratorInterface;
+use ApiPlatform\Core\GraphQl\Subscription\SubscriptionManagerInterface as GraphQlSubscriptionManagerInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Doctrine\Common\EventArgs;
 use Doctrine\ODM\MongoDB\Event\OnFlushEventArgs as MongoDbOdmOnFlushEventArgs;
 use Doctrine\ORM\Event\OnFlushEventArgs as OrmOnFlushEventArgs;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\Mercure\Update;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Serializer\SerializerInterface;
@@ -49,11 +52,13 @@ final class PublishMercureUpdatesListener
     private $updatedObjects;
     private $deletedObjects;
     private $formats;
+    private $graphQlSubscriptionManager;
+    private $graphQlMercureSubscriptionIriGenerator;
 
     /**
      * @param array<string, string[]|string> $formats
      */
-    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ExpressionLanguage $expressionLanguage = null)
+    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
     {
         if (null === $messageBus && null === $publisher) {
             throw new InvalidArgumentException('A message bus or a publisher must be provided.');
@@ -67,6 +72,8 @@ final class PublishMercureUpdatesListener
         $this->messageBus = $messageBus;
         $this->publisher = $publisher;
         $this->expressionLanguage = $expressionLanguage ?? class_exists(ExpressionLanguage::class) ? new ExpressionLanguage() : null;
+        $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
+        $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
         $this->reset();
     }
 
@@ -106,15 +113,15 @@ final class PublishMercureUpdatesListener
     {
         try {
             foreach ($this->createdObjects as $object) {
-                $this->publishUpdate($object, $this->createdObjects[$object]);
+                $this->publishUpdate($object, $this->createdObjects[$object], 'create');
             }
 
             foreach ($this->updatedObjects as $object) {
-                $this->publishUpdate($object, $this->updatedObjects[$object]);
+                $this->publishUpdate($object, $this->updatedObjects[$object], 'update');
             }
 
             foreach ($this->deletedObjects as $object) {
-                $this->publishUpdate($object, $this->deletedObjects[$object]);
+                $this->publishUpdate($object, $this->deletedObjects[$object], 'delete');
             }
         } finally {
             $this->reset();
@@ -173,7 +180,7 @@ final class PublishMercureUpdatesListener
     /**
      * @param object $object
      */
-    private function publishUpdate($object, array $targets): void
+    private function publishUpdate($object, array $targets, string $type): void
     {
         if ($object instanceof \stdClass) {
             // By convention, if the object has been deleted, we send only its IRI.
@@ -190,7 +197,35 @@ final class PublishMercureUpdatesListener
             $data = $this->serializer->serialize($object, key($this->formats), $context);
         }
 
-        $update = new Update($iri, $data, $targets);
-        $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+        $updates = array_merge([new Update($iri, $data, $targets)], $this->getGraphQlSubscriptionUpdates($object, $targets, $type));
+
+        foreach ($updates as $update) {
+            $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+        }
+    }
+
+    /**
+     * @param object $object
+     *
+     * @return Update[]
+     */
+    private function getGraphQlSubscriptionUpdates($object, array $targets, string $type): array
+    {
+        if ('update' !== $type || !$this->graphQlSubscriptionManager || !$this->graphQlMercureSubscriptionIriGenerator) {
+            return [];
+        }
+
+        $payloads = $this->graphQlSubscriptionManager->getPushPayloads($object);
+
+        $updates = [];
+        foreach ($payloads as [$subscriptionId, $data]) {
+            $updates[] = new Update(
+                $this->graphQlMercureSubscriptionIriGenerator->generateTopicIri($subscriptionId),
+                (string) (new JsonResponse($data))->getContent(),
+                $targets
+            );
+        }
+
+        return $updates;
     }
 }
--- a/src/GraphQl/Resolver/Factory/ItemSubscriptionResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ItemSubscriptionResolverFactory.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\GraphQl\Resolver\Factory;
+
+use ApiPlatform\Core\GraphQl\Resolver\Stage\ReadStageInterface;
+use ApiPlatform\Core\GraphQl\Resolver\Stage\SecurityStageInterface;
+use ApiPlatform\Core\GraphQl\Resolver\Stage\SerializeStageInterface;
+use ApiPlatform\Core\GraphQl\Subscription\MercureSubscriptionIriGeneratorInterface;
+use ApiPlatform\Core\GraphQl\Subscription\SubscriptionManagerInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Util\ClassInfoTrait;
+use ApiPlatform\Core\Util\CloneTrait;
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * Creates a function resolving a GraphQL subscription of an item.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class ItemSubscriptionResolverFactory implements ResolverFactoryInterface
+{
+    use ClassInfoTrait;
+    use CloneTrait;
+
+    private $readStage;
+    private $securityStage;
+    private $serializeStage;
+    private $resourceMetadataFactory;
+    private $subscriptionManager;
+    private $mercureSubscriptionIriGenerator;
+
+    public function __construct(ReadStageInterface $readStage, SecurityStageInterface $securityStage, SerializeStageInterface $serializeStage, ResourceMetadataFactoryInterface $resourceMetadataFactory, SubscriptionManagerInterface $subscriptionManager, ?MercureSubscriptionIriGeneratorInterface $mercureSubscriptionIriGenerator)
+    {
+        $this->readStage = $readStage;
+        $this->securityStage = $securityStage;
+        $this->serializeStage = $serializeStage;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
+        $this->subscriptionManager = $subscriptionManager;
+        $this->mercureSubscriptionIriGenerator = $mercureSubscriptionIriGenerator;
+    }
+
+    public function __invoke(?string $resourceClass = null, ?string $rootClass = null, ?string $operationName = null): callable
+    {
+        return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operationName) {
+            if (null === $resourceClass || null === $operationName) {
+                return null;
+            }
+
+            $resolverContext = ['source' => $source, 'args' => $args, 'info' => $info, 'is_collection' => false, 'is_mutation' => false, 'is_subscription' => true];
+
+            $item = ($this->readStage)($resourceClass, $rootClass, $operationName, $resolverContext);
+            if (null !== $item && !\is_object($item)) {
+                throw new \LogicException('Item from read stage should be a nullable object.');
+            }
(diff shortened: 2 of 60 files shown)


alanpoulain:
@lukasluecke @mahmoodbazdar @CvekCoding if you could have a look at this PR?

lukasluecke:
@alanpoulain Thanks for your effort! I will check it out in the next days 🙂

CvekCoding:
@alanpoulain wow, great. I'll check this on NY vacation (at least I'll try:)).

Great thanks for this update - sounds great. As far as I know it's a first graphql subscription implementation in php world.

lukasluecke:
@alanpoulain Looks good so far - I'll see if I can find the time this week to try it out (i.e. build a working client-side integration utilizing this).

src/GraphQl/Subscription/SubscriptionManager.php
```diff
@@ -0,0 +1,124 @@
...
+        $iri = $this->iriConverter->getIriFromItem($object);
+        $subscriptions = $this->getSubscriptionsFromIri($iri);
+
+        $resourceClass = $this->getObjectClass($object);
+
+        $payloads = [];
+        foreach ($subscriptions as [$subscriptionId, $subscriptionFields, $subscriptionResult]) {
+            $resolverContext = ['fields' => $subscriptionFields, 'is_collection' => false, 'is_mutation' => false, 'is_subscription' => true];
+
+            $data = ($this->serializeStage)($object, $resourceClass, 'update', $resolverContext);
```
lukasluecke: @alanpoulain This means that the subscription data will get serialized using the `update` context, right? Should there be a way to override it for the subscription payload specifically?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -190,7 +196,35 @@ private function publishUpdate($object, array $targets): void
...
+            return [];
+        }
+
+        $payloads = $this->graphQlSubscriptionManager->getPushPayloads($object);
+
+        $updates = [];
+        foreach ($payloads as [$subscriptionId, $data]) {
+            $updates[] = new Update(
+                $this->graphQlMercureSubscriptionUriGenerator->generateMercureTopicUrl($subscriptionId),
+                $this->serializer->serialize($data, key($this->formats)),
```
lukasluecke: @alanpoulain Wouldn't this serialize the messages using the same format (defaults to JSON-LD I believe?) as the normal mercure updates? I think it would make sense to use GraphQL here, no?

src/GraphQl/Subscription/SubscriptionManager.php
alanpoulain: I think it would be a bad idea to have a `subscription` context. Subscriptions are here to send the updated data like you would receive it if you made a mutation and read the received data.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
alanpoulain: The normalized data is sent as JSON like `graphql-php` would do. Maybe I should use `json` instead of the key of formats, WDYT?

alanpoulain: I've changed it to mimic how the endpoint is doing it: https://github.com/api-platform/core/blob/9298c4502b56f3bff7ed91ad05ea71ca216b58cf/src/GraphQl/Action/EntrypointAction.php#L85.

src/GraphQl/Subscription/SubscriptionManager.php
lukasluecke: You might be right 😉  what about the case where there is no `update` mutation on the resource? (It can still be changed from other parts of the API, right?)

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
lukasluecke: LGTM 👍

src/GraphQl/Subscription/MercureSubscriptionUriGenerator.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function generateMercureTopicIri(string $subscriptionId): string
+    {
+        return "/subscriptions/$subscriptionId";
+    }
+
+    public function generateMercureTopicUrl(string $subscriptionId): string
+    {
+        if (('' === $scheme = $this->requestContext->getScheme()) || ('' === $host = $this->requestContext->getHost())) {
+            throw new \RuntimeException('Cannot generate topic URL from request context.');
```
dunglas: I suggest to use a default value if the context isn't available, such as `https://api-platform.com/subsriptions/`. IRIs don't have to exist.

src/GraphQl/Subscription/MercureSubscriptionUriGeneratorInterface.php
```diff
@@ -0,0 +1,30 @@
...
+namespace ApiPlatform\Core\GraphQl\Subscription;
+
+/**
+ * Generates Mercure-related URIs from a subscription ID.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+interface MercureSubscriptionUriGeneratorInterface
```
dunglas: `MercureSubscriptionIriGeneratorInterface` for consistency (we use IRI everywhere).

src/GraphQl/Subscription/MercureSubscriptionUriGeneratorInterface.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    public function generateMercureTopicIri(string $subscriptionId): string;
+
+    public function generateMercureTopicUrl(string $subscriptionId): string;
+
+    public function generateMercureUrl(string $subscriptionId): string;
```
dunglas: I suggest to simplify this interface and to provide only a single method `getTopic(string $subscriptionId)`

src/GraphQl/Subscription/SubscriptionIdentifierGenerator.php
```diff
@@ -0,0 +1,29 @@
...
+namespace ApiPlatform\Core\GraphQl\Subscription;
+
+/**
+ * Generates an identifier used to identify a subscription.
+ *
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class SubscriptionIdentifierGenerator implements SubscriptionIdentifierGeneratorInterface
```
dunglas: Are this class and the related really necessary? Cannot we merge it with the topic generator, and use the topic as ID everywhere?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -49,11 +52,13 @@ final class PublishMercureUpdatesListener
...
     private $deletedObjects;
     private $formats;
+    private $graphQlSubscriptionManager;
+    private $graphQlMercureSubscriptionUriGenerator;
 
     /**
      * @param array<string, string[]|string> $formats
      */
-    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ExpressionLanguage $expressionLanguage = null)
+    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionUriGeneratorInterface $graphQlMercureSubscriptionUriGenerator = null, ExpressionLanguage $expressionLanguage = null)
```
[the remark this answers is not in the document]
alanpoulain: The expression language is not defined in the configuration so it would mean doing dirty tricks.
Since it's experimental, I thought it would be better to not prevent the BC.

src/GraphQl/Subscription/SubscriptionManager.php
alanpoulain: Would you see the `item_query` context instead then?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -190,7 +197,35 @@ private function publishUpdate($object, array $targets): void
...
+            $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+        }
+    }
+
+    /**
+     * @param object $object
+     *
+     * @return Update[]
+     */
+    private function getGraphQlSubscriptionUpdates($object, array $targets, string $type): array
```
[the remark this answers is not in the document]
alanpoulain: It would mean introducing methods to get the created/updated/deleted items and to duplicate the code in `postFlush` and a little from `publishUpdate`. Is it worth it?

src/GraphQl/Resolver/Factory/ItemSubscriptionResolverFactory.php
```diff
@@ -0,0 +1,90 @@
...
+            ]);
+
+            $result = ($this->serializeStage)($item, $resourceClass, $operationName, $resolverContext);
+
+            $subscriptionId = $this->subscriptionManager->retrieveSubscriptionId($resolverContext, $result);
+
+            $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
+
+            if ($subscriptionId && $resourceMetadata->getAttribute('mercure', false)) {
+                if (!$this->mercureSubscriptionUriGenerator) {
```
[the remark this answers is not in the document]
alanpoulain: I've thought the subscription system to be independent of Mercure. Here, we need to see if the resource has a `mercure` attribute and, if it's the case, throw the exception if the bundle is not required. We cannot do this in the constructor.

src/GraphQl/Subscription/MercureSubscriptionUriGenerator.php
```diff
@@ -0,0 +1,56 @@
...
+ * @experimental
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+final class MercureSubscriptionUriGenerator implements MercureSubscriptionUriGeneratorInterface
+{
+    private $requestContext;
+    private $hub;
+
+    public function __construct(RequestContext $requestContext, string $hub)
```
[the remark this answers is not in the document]
alanpoulain: It would be nice but these parameters are not dynamically updated whereas it's the case for the service.

src/GraphQl/Subscription/MercureSubscriptionUriGenerator.php
alanpoulain: OK, it's an edge case anyway.

src/GraphQl/Subscription/MercureSubscriptionUriGeneratorInterface.php
alanpoulain: I can remove the `generateMercureTopicIri` but I need two methods: one for getting the topic IRI and one for getting the Mercure URL (with the topic in the query).

src/GraphQl/Subscription/SubscriptionIdentifierGenerator.php
alanpoulain: This class is necessary because I use a random string for the subscriptions. I can remove it if I use a hash instead. Maybe using a SHA-256 hash would be a better idea like we discussed since Apollo is doing it for its persisted queries.
Also I cannot add it to the generator since it's not related to Mercure.

lukasluecke: Using a hash sounds like a good idea!

alanpoulain: Done. I think keeping the service is a good idea since it allows to change how the identifier is generated if needed.

src/GraphQl/Subscription/MercureSubscriptionUriGenerator.php
dunglas: Actually, maybe should we always use this default, and let the user pass another namespace if required, instead of relying on the router.

src/GraphQl/Subscription/SubscriptionManager.php
lukasluecke: @alanpoulain Sorry for the late reply - I think trying `update` first and falling back to `item_query` would be good, wdyt?

lukasluecke:
@alanpoulain I have now implemented this in my project with a matching frontend client, and it works great! Thank you very much for your work! 🏆 

I do have some questions / improvement requests for the future:
Currently the subscriptions are bound only to the update of the main resource (the one that generates the subscription operation, based on `mercure=true`), right?
Is there any way we could update this also based on the selection set, so inner resource updates would also trigger new data (would something like #3206 help)?
Could we add custom Subscriptions, similar to the Queries and Mutations, later on? Not sure how exactly their signature would look like, but that could be an interesting feature - either mix multiple resources and their updates in one subscription, or even integrate "outside" sources, e.g. custom Messenger messages.

alanpoulain:
Thank you for your feedback @lukasluecke!
For your first question, yes https://github.com/api-platform/core/issues/3206 would help to do this.
Custom subscriptions could be a nice addition later on :slightly_smiling_face:

lukasluecke:
@alanpoulain Great! After some testing it would probably be good to get #3206 (or something comparable) in together with this, otherwise things might get confusing 🙈 Could you maybe take a look at this?

alanpoulain:
Can't it be done later?

lukasluecke:
Of course, it's not strictly required / related to the changes directly 🙂 I'll try to look into it in the next days 👍

src/GraphQl/Subscription/SubscriptionManager.php
alanpoulain: Actually, if you don't have an `update` operation, the subscription fields are not generated, so it makes sense to me to keep it like this.

plamenh:
Hey guys, is this now widely available? Doesn't seem to work on 2.5.4 (I get 'Internal Error: Subscription type is not defined the Schema' with relay). Is this my fault or it's in this version? Thanks.

alanpoulain:
Not yet. In 2.6.

lukasluecke:
@alanpoulain Quick question - is there an easy solution to make Mercure use the GraphQL serializer & format for the updates it sends? Would make it a lot easier to properly integrate clientside, instead of dealing with the JSONLD and the differences.

alanpoulain:
It shouldn't use JSON-LD by default but the GraphQL format. See this test: https://github.com/api-platform/core/blob/2c87089c4330cd2f571f023fbcc36b5edd2b4dbb/features/graphql/subscription.feature#L196-L207

lukasluecke:
@alanpoulain Sorry, I was a little confused when posting. It's not JSONLD, but it's also not what usually is available in a GraphQL response.

I'm using a normalized cache in the frontend, and in order to match the subscription results I need the `__typename` and `id` to match - however in Mercure updates the `__typename` is not available at all, and the `id` is just the numeric ID, not the IRI as it should be. Is this just a misconception on my side?

I fixed it temporarily in `SubscriptionManager` with the following lines (not nice, but it works 😅):
```php
$rootObject = array_keys($data)[0];
$data[$rootObject]['__typename'] = $this->resourceMetadataFactory->create($resourceClass)->getShortName();
$data[$rootObject]['id'] = $iri;
```

alanpoulain:
I understand for the `__typename`, because it is given by `graphql-php` and not the serialize stage.

For the ID and not the IRI, I don't remember why it's like this (and I even wrote it in the tests), but it seems an issue to me indeed. Maybe the context is not given correctly to the stage.

SlimenTN:
hey @alanpoulain thanks for your great work to integrate graphql subscription
but I'm using angular with apiplatform and when I try to create a subscription I get this error 
```javascript
Schema is not configured for subscriptions.
```
I want to get notification whenever a new one is created so I have a `Notification` entity with mercure activated `mercure=true`, and this is my Angular side code:
```javascript
ngOnInit(){
    this.apollo.subscribe({
      query: gql`
        subscription{
          newNotification{
            title
            content
          }
        }`,
      fetchPolicy: "no-cache"
    })
    .subscribe(res => console.log(res));
  }
```
why it doesn't know the subscription schema ?
same thing happens if I run it in the Qraphiql IDE that comes with ApiPlatform:
image

lukasluecke:
You need to use a custom link that knows how to handle the mercure subscription. I can give you an example of what I'm using tomorrow, but it's for urql, not Apollo.

SlimenTN:
@lukasluecke ok thanks for your response
this custom link is it mentioned anywhere in the documentation? because there isn't a lot mentioned about the subscription in apiplatform's documentation!

xleliberty:
Hi, @lukasluecke @SlimenTN, i'm trying using apollo client with a splitLink to connect to api-platform subscriptions, but i didn't succeeded and didn't find any good exemple code. 

Have some of you found a "good" way to implement this ? 
@lukasluecke would you mind sharing your urql code ? 

Thanks
Xavier

SlimenTN:
Hi @xleliberty actually still waiting for @lukasluecke's example, because as I mentioned there isn't a lot mentioned about the subscription in apiplatform's documentation.
So I hope he can answer us soon 😄

xleliberty:
Ok Thanks Slimen, 

You're right, i feel too like we are missing a bit of informations about subscription consumption on the JS side. Once we have more info bits, i'll try to add a little doc PR for this. 

@alanpoulain , maybe you have some hints about js consumption ?

Thanks

lukasluecke:
@SlimenTN @xleliberty Sorry, totally forgot about this! You can see my example at https://gist.github.com/lukasluecke/0c8e2c12a955198926779a8691874c03.

xleliberty:
Hi @lukasluecke , thanks for your input !

lermontex:
@SlimenTN, @xleliberty with Apollo, you can use it like a normal Query, something like this:

```
    const { result } = useQuery(
      gql`
        subscription yourSubscriptionName($id: ID!) {
          updateTaskSubscribe(input: { id: $id }) {
            dummyMercure {
              id
              name
              relatedDummy {
                name
              }
            }
            mercureUrl
          }
        }
      `,
      () => ({
        id: "/dummy_mercures/1",
      })
    );
```

PR #3333, opened 2020-01-01
[OpenApi] [Oauth] Fix render oauth when using OpenApi v3

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

Following the documentation for Oauth in OAS3 (https://swagger.io/docs/specification/authentication/oauth2/), flows now need to be an array.

This fix turn this
image
in this
image

I kept the current behaviour for swagger v2.

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -639,7 +639,10 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
 
             $securityDefinitions['oauth'] = [
                 'type' => $this->oauthType,
-                'description' => 'OAuth client_credentials Grant',
+                'description' => sprintf(
+                    'OAuth 2.0 %s Grant',
+                    strtolower(preg_replace('/[A-Z]/', ' \\0', lcfirst($this->oauthFlow)))
+                ),
             ];
 
             if ($v3) {
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -631,14 +631,25 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
         $security = [];
 
         if ($this->oauthEnabled) {
-            $securityDefinitions['oauth'] = [
-                'type' => $this->oauthType,
-                'description' => 'OAuth client_credentials Grant',
-                'flow' => $this->oauthFlow,
+            $oauthAttributes = [
                 'tokenUrl' => $this->oauthTokenUrl,
                 'authorizationUrl' => $this->oauthAuthorizationUrl,
                 'scopes' => $this->oauthScopes,
             ];
+
+            $securityDefinitions['oauth'] = [
+                'type' => $this->oauthType,
+                'description' => 'OAuth client_credentials Grant',
+            ];
+
+            if ($v3) {
+                $securityDefinitions['oauth']['flows'] = [
+                    $this->oauthFlow => $oauthAttributes,
+                ];
+            } else {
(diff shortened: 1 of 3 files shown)


teohhanhui:
https://demo.api-platform.com/docs?spec_version=3 seems to work. Is it a bug in Swagger UI (not respecting this part of the OAS3 spec) that has recently been fixed?

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -631,15 +631,26 @@ private function computeDoc(bool $v3, Documentation $documentation, \ArrayObject
...
             ];
 
+            $securityDefinitions['oauth'] = [
+                'type' => $this->oauthType,
+                'description' => 'OAuth client_credentials Grant',
+            ];
+
+            if ($v3) {
+                $securityDefinitions['oauth']['flows'] = [
+                    $this->oauthFlow => $oauthAttributes,
```
teohhanhui: This seems wrong.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -631,15 +631,26 @@ private function computeDoc(bool $v3, Documentation $documentation, \ArrayObject
...
-                'flow' => $this->oauthFlow,
+            $oauthAttributes = [
                 'tokenUrl' => $this->oauthTokenUrl,
                 'authorizationUrl' => $this->oauthAuthorizationUrl,
                 'scopes' => $this->oauthScopes,
             ];
 
+            $securityDefinitions['oauth'] = [
+                'type' => $this->oauthType,
+                'description' => 'OAuth client_credentials Grant',
```
teohhanhui: Not sure why this is hardcoded. Can you fix this as well?

Jibbarth: @teohhanhui IMO, as this description is not yet configurable, we should change it in an other PR on master, because change this should be a new feature ? WDYT ?

teohhanhui: It doesn't need to be configurable, but we should generate the correct description based on the flow, no?

Jibbarth: Oh sure :sweat_smile: 

Something like this ?

```suggestion
                'description' => 'This API uses OAuth 2 with the ' . $this->oauthFlow . ' grant flow.',
```

teohhanhui:
```suggestion
                'description' => sprintf('OAuth %s Grant', $this->oauthFlow),
```

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -631,15 +631,29 @@ private function computeDoc(bool $v3, Documentation $documentation, \ArrayObject
...
                 'tokenUrl' => $this->oauthTokenUrl,
                 'authorizationUrl' => $this->oauthAuthorizationUrl,
                 'scopes' => $this->oauthScopes,
             ];
 
+            $securityDefinitions['oauth'] = [
+                'type' => $this->oauthType,
+                'description' => sprintf(
+                    'OAuth 2.0 %s Grant',
+                    strtolower(preg_replace('/[A-Z]/', ' \\0', lcfirst($this->oauthFlow)))
```
Jibbarth: Heavily inspired by [CamelCaseToSnakeCaseNameConverter](https://github.com/symfony/serializer/blob/master/NameConverter/CamelCaseToSnakeCaseNameConverter.php#L40)

soyuka:
Thanks @Jibbarth !

PR #3338, opened 2020-01-09
Wrong sorting of arguments

sfeni:
The service definition in https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Resources/config/swagger.xml differs from constructor, so the YAML exporter is not recognizing the swaggerVersions at the moment. And only if defaultContext is the last one we can change it. Please change the file. Thanks

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | no
| License       | MIT

The service definition in https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Resources/config/swagger.xml differs from constructor, so the YAML exporter is not recognizing the swaggerVersions at the moment. And only if defaultContext is the last one we can change it.


soyuka:
Looks like this breaks the test suite, also if it's a fix please target the 2.5 branch.

soyuka:
fixed in https://github.com/api-platform/core/pull/3443

PR #3342, opened 2020-01-11
#3316: local cache on AbstractItemNormalizer::getAllowedAttributes()

### LINKED ISSUE #3316: [performance] Add local cache on ResourceClassResolver::getResourceClass()
**API Platform version(s) affected**: 4.5.* 

**Description**  
Follow up of this closed ticket: https://github.com/api-platform/core/issues/1068, I find a way to optimize a lot `ResourceClassResolver::getResourceClass()`.

**How to reproduce**  
Just go to a normal apip route. My case is violent because I've a lot of api resources and the response returns a lot of entities.

But there is room for improvment!

**Possible Solution**  
`ResourceClassResolver::getResourceClass()` is responsible of no less than 13% in prod mode, because foreach entity, it will go through ALL possible resource, without any break in the loop (seems logic) nor any cache. And so, `is_a` is called 63630 times.

Just adding a local cache clear everything:

```php
        if (isset($this->localMostSpecificResourceClassCache[$targetClass])) {
            return $this->localMostSpecificResourceClassCache[$targetClass];
        }

        foreach ($this->resourceNameCollectionFactory->create() as $resourceClassName) {
            if (!is_a($targetClass, $resourceClassName, true)) {
                continue;
            }

            if (null === $mostSpecificResourceClass || is_subclass_of($resourceClassName, $mostSpecificResourceClass)) {
                $mostSpecificResourceClass = $resourceClassName;
            }
        }

        if (null === $mostSpecificResourceClass) {
            throw new \LogicException('Unexpected execution flow.');
        }

        $this->localMostSpecificResourceClassCache[$targetClass] = $mostSpecificResourceClass;
```

Before : 

image

Diff :

image

bastnic:
| Q             | A
| ------------- | ---
| Bug fix?      | no (performance)
| New feature?  | hmm (performance)
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3316 
| License       | MIT
| Doc PR        | 

The code of getAllowedAttributes is quite heavy and relate on a lot of micro cache, despite being a medium hot path.

Just add a local cache on it, but this is a draft. The cache key only relates on `$options` that I don't think is enough. Relates on `$context` isn't possible because of `$context['cache_Key']`. Should I just remove it and pass ĉontext ?

image

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -50,6 +50,7 @@ use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
  */
 abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 {
+    use CacheKeyTrait;
     use ClassInfoTrait;
     use ContextTrait;
     use InputOutputMetadataTrait;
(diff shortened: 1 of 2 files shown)


alanpoulain:
WDYT of using the `getCacheKey` method like I modified @bastnic?

alanpoulain:
I took the changes from Symfony (it was copied from it).

alanpoulain:
Is there a test we could make to see the performance improvement with Blackfire?

PR #3346, opened 2020-01-16
Allow to autoconfigure validation groups generators

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | api-platform/docs#1023

When having many validation groups generators for resources, you need to make every generator as a public service, and this doesn't work with `_instanceof` configuration where you could add a custom marker and make those services public, because symfony doesn't detect that those services are used and removes from container.

This adds `ApiPlatform\Core\Validator\ValidationGroupsGeneratorInterface` so apps can leverage the autoconfiguration + autowiring of those generators. Generators are collected and registered in a custom service locator, but consumers won't need anymore any configuration for them if will implement the interface. For not breaking backwards compatibility left the full service container as a fallback.

TODO:
- [x] Add more tests if needed?
- [x] Make docs PR

### DIFF
--- a/src/Bridge/Symfony/Validator/Validator.php
+++ b/src/Bridge/Symfony/Validator/Validator.php
@@ -50,6 +50,10 @@ class Validator implements ValidatorInterface
                 ($service = $this->container->get($validationGroups)) &&
                 \is_callable($service)
             ) {
+                if (!$service instanceof ValidationGroupsGeneratorInterface) {
+                    @trigger_error(sprintf('Using a public validation groups generator service not implementing "%s" is deprecated since 2.6 and will be removed in 3.0.', ValidationGroupsGeneratorInterface::class), E_USER_DEPRECATED);
+                }
+
                 $validationGroups = $service($data);
             } elseif (\is_callable($validationGroups)) {
                 $validationGroups = $validationGroups($data);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * GraphQL: Allow to format GraphQL errors based on exceptions (#3063)
 * GraphQL: Add page-based pagination (#3175)
 * OpenAPI: Add PHP default values to the documentation (#2386)
+* Deprecate using a validation groups generator service not implementing `ApiPlatform\Core\Bridge\Symfony\Validator\ValidationGroupsGeneratorInterface` (#3346)
 
 ## 2.5.x-dev
 
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -24,6 +24,7 @@ use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\QueryItemExtensionInterface;
 use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\AbstractContextAwareFilter as DoctrineOrmAbstractContextAwareFilter;
 use ApiPlatform\Core\Bridge\Elasticsearch\DataProvider\Extension\RequestBodySearchCollectionExtensionInterface;
 use ApiPlatform\Core\Bridge\Symfony\Validator\Metadata\Property\Restriction\PropertySchemaRestrictionMetadataInterface;
+use ApiPlatform\Core\Bridge\Symfony\Validator\ValidationGroupsGeneratorInterface;
 use ApiPlatform\Core\DataPersister\DataPersisterInterface;
 use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
@@ -554,8 +555,12 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
     {
         if (interface_exists(ValidatorInterface::class)) {
             $loader->load('validator.xml');
+
+            $container->registerForAutoconfiguration(ValidationGroupsGeneratorInterface::class)
+                ->addTag('api_platform.validation_groups_generator')
+                ->setPublic(true); // this line should be removed in 3.0
             $container->registerForAutoconfiguration(PropertySchemaRestrictionMetadataInterface::class)
-                      ->addTag('api_platform.metadata.property_schema_restriction');
+                ->addTag('api_platform.metadata.property_schema_restriction');
         }
 
         if (!$config['validator']) {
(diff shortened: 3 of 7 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorFilterPass.php
```diff
@@ -0,0 +1,42 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Registers validation groups generators.
+ *
+ * @internal
+ *
+ * @author Tomas Norkūnas <dunglas@gmail.com>
```
soyuka: wrong email or intended?

norkunas: copy paste :d

norkunas:
Also updated return typehint from generator interface not to be strict and allow return GroupSequence for consistency.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorFilterPass.php
```diff
@@ -0,0 +1,42 @@
...
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Registers validation groups generators.
+ *
+ * @internal
+ *
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ */
+final class ValidationGroupsGeneratorFilterPass implements CompilerPassInterface
```
dunglas: Can't this class be removed in favor of `!tagged`? https://symfony.com/doc/current/service_container/tags.html#reference-tagged-services

src/Bridge/Symfony/Validator/Validator.php
```diff
@@ -51,6 +53,14 @@ public function validate($data, array $context = [])
                 \is_callable($service)
             ) {
                 $validationGroups = $service($data);
+            } elseif (
```
dunglas: Maybe should we deprecate using public services not implementing the interface.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorFilterPass.php
norkunas: I'm not sure, because `ServiceLocator` holds the index to the service so this compiler pass maintains the indexes, otherwise we should use `index_by` which was introduced in Sf 4.3 and then the generators should define the method to get the index. Prove me wrong!

norkunas: Also service locator throws this when trying to inject !tagged:
`Invalid definition for service "app_service_locator": an array of references is expected as first argument when the "container.service_locator" tag is set.`

norkunas:
ping :)

src/Bridge/Symfony/Validator/Validator.php
```diff
@@ -30,11 +31,13 @@ class Validator implements ValidatorInterface
 {
     private $validator;
     private $container;
+    private $serviceContainer;
 
-    public function __construct(SymfonyValidatorInterface $validator, ContainerInterface $container = null)
+    public function __construct(SymfonyValidatorInterface $validator, ContainerInterface $container = null, ContainerInterface $serviceContainer = null)
```
teohhanhui: No need to introduce a new argument here. We can just switch from the full container to the service locator in 3.0, since we trigger a deprecation error for services not implementing the interface.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorFilterPass.php
teohhanhui: See https://github.com/api-platform/core/pull/3346/files#r383869931

Let's only do this in API Platform 3.0 - just stick to the full container for now.

src/Bridge/Symfony/Validator/Validator.php
norkunas: You mean just mark the generators as public to be available in the full container to avoid another container until 3.0?

teohhanhui: Uhh... That's not ideal indeed, but yeah, they'd have to be public.

norkunas:
Updated, but I don't understand what's wrong with the tests. How a single definition can affect another definitions :thinking:

teohhanhui:
```
Error: Call to a member function setPublic() on null
```

Looks like there's a missing mock?

norkunas:
> ```
> Error: Call to a member function setPublic() on null
> ```
> 
> Looks like there's a missing mock?

Which mock?

teohhanhui:
> Which mock?

The `ChildDefinition` prophecy must return the same instance on `addTag(...)`, e.g. https://github.com/phpspec/prophecy/pull/190#issuecomment-149793777

norkunas:
> > Which mock?
> 
> The `ChildDefinition` prophecy must return the same instance on `addTag(...)`, e.g. [phpspec/prophecy#190 (comment)](https://github.com/phpspec/prophecy/pull/190#issuecomment-149793777)
> 
> My suggestion is to add:
> 
> ```
> $this->childDefinitionProphecy->addTag(Argument::type('string'))->willReturn($this->childDefinitionProphecy);
> ```
> 
> in `getBaseContainerBuilderProphecy` method.

Didn't help :man_facepalming: :grin:

teohhanhui:
Let me fix this...

src/Bridge/Symfony/Validator/Validator.php
```diff
@@ -50,6 +50,10 @@ public function validate($data, array $context = [])
                 ($service = $this->container->get($validationGroups)) &&
                 \is_callable($service)
             ) {
+                if (!$service instanceof ValidationGroupsGeneratorInterface) {
```
dunglas: Is this deprecation mandatory?

teohhanhui: Yes, so that we can inject a service locator in 3.0

norkunas: @dunglas https://github.com/api-platform/core/pull/3346#discussion_r369280721

teohhanhui:
Thanks @norkunas! :tada: :rocket:

Issue #3349, opened 2020-01-20
ReflectionException: Class not found


ivansrdic:
**API Platform version(s) affected**: 2.5.4 (from commit https://github.com/api-platform/core/commit/baf054234eff3b9276b6783efb825bff7099d5c9)

**Description** / **How to reproduce**  
Loading the api docs causes several "class not found" ReflectionExceptions.
![Screenshot from 2020-01-20 13-53-51](https://user-images.githubusercontent.com/7668873/72727966-86212200-3b8c-11ea-967f-d9cdf07f7d35.png)
Possibly related to https://github.com/api-platform/core/issues/3344

**Possible Solution**  
Seems to be related to `src/JsonSchema/TypeFactory:102`. If the check in changed from and to or, everything works (working example below)
```
        if ($this->isResourceClass($className) || true !== $readableLink) {
            return [
                'type' => 'string',
                'format' => 'iri-reference',
            ];
        }
```

Unfortunately I don't know the internals enough to know if this breaks anything.

soyuka:
Definitely related. Could you contribute your solution?

soyuka:
btw it looks like this class really doesn't exist, you have a bad import.

ivansrdic:
I've checked, import is correct. The error message is gone if in the related trait I remove the `use` for the EventDispatcherInterface statement and replace the class references with the FQN. But the error appears with all used traits that have properties that are used in ApiResource entities. I guess this means that the class resolver has a bug somewhere.

Example of when it throws an exception
```
<?php

namespace App\DependencyInjection;

use \Symfony\Component\EventDispatcher\EventDispatcherInterface;

trait EventDispatcherAwareTrait
{

    /**
     * @var EventDispatcherInterface
     */
    protected $eventDispatcher;

    /**
     * @inheritDoc
     */
    public function setEventDispatcher(EventDispatcherInterface $eventDispatcher)
    {
        $this->eventDispatcher = $eventDispatcher;
    }
}
```

No exception after this change
```
<?php

namespace App\DependencyInjection;

trait EventDispatcherAwareTrait
{

    /**
     * @var \Symfony\Component\EventDispatcher\EventDispatcherInterface
     */
    protected $eventDispatcher;

    /**
     * @inheritDoc
     */
    public function setEventDispatcher(\Symfony\Component\EventDispatcher\EventDispatcherInterface $eventDispatcher)
    {
        $this->eventDispatcher = $eventDispatcher;
    }
}
```

[...]

Issue #3343, opened 2020-01-13
There is no PropertyInfo extractor supporting the class "DateInterval".


cacahouwete:
**API Platform version(s) affected**: 2.5.4

**Description**  
When I upgrade api-platform/core 2.5.3 to 2.5.4 I have the following problem on api docs: "There is no PropertyInfo extractor supporting the class "DateInterval"."

**How to reproduce**  
I think we just have to put a property with a DateInterval in an exposed entity and it will reproduce the problem. 

**Possible Solution**  
No idea

**Additional Context**  

```
ApiPlatform\Core\Exception\RuntimeException:
There is no PropertyInfo extractor supporting the class "DateInterval".

  at vendor/api-platform/core/src/Bridge/Symfony/PropertyInfo/Metadata/Property/PropertyInfoPropertyNameCollectionFactory.php:46
  at ApiPlatform\Core\Bridge\Symfony\PropertyInfo\Metadata\Property\PropertyInfoPropertyNameCollectionFactory->create('DateInterval', array())
     (vendor/api-platform/core/src/Metadata/Property/Factory/InheritedPropertyNameCollectionFactory.php:44)
  at ApiPlatform\Core\Metadata\Property\Factory\InheritedPropertyNameCollectionFactory->create('DateInterval', array())
     (vendor/api-platform/core/src/Metadata/Property/Factory/ExtractorPropertyNameCollectionFactory.php:50)
  at ApiPlatform\Core\Metadata\Property\Factory\ExtractorPropertyNameCollectionFactory->create('DateInterval', array())
     (vendor/api-platform/core/src/Metadata/Property/Factory/ExtractorPropertyNameCollectionFactory.php:50)
  at ApiPlatform\Core\Metadata\Property\Factory\ExtractorPropertyNameCollectionFactory->create('DateInterval', array())
     (vendor/api-platform/core/src/Metadata/Property/Factory/CachedPropertyNameCollectionFactory.php:47)
  at ApiPlatform\Core\Metadata\Property\Factory\CachedPropertyNameCollectionFactory->ApiPlatform\Core\Metadata\Property\Factory\{closure}()
     (vendor/api-platform/core/src/Cache/CachedTrait.php:44)
  at ApiPlatform\Core\Metadata\Property\Factory\CachedPropertyNameCollectionFactory->getCached('property_name_collection_3bf8de46e9d114c9746a5da4376c201b', object(Closure))
[...]

PR #3350, opened 2020-01-20
Fix class not found schema factory (do not merge)

### LINKED ISSUE #3344: Class elf does not exist
**API Platform version(s) affected**: 2.5.4 (2.5.3 is not affected)

**Description** / **How to reproduce**  
When using this property in an ApiResource

```php
    /**
     * @var self
     */
    public $foo;
```

.. I get the error "Class elf does not exist" on api docs.

**Additional Context**  

Screenshot 2020-01-15 14 52 23

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixed #3344 #3349 #3343
| License       | MIT
| Doc PR        | 

I'm not really sure that we should merge this, it skips wrong classes but there is definitely something more about https://github.com/api-platform/core/issues/3344 that I didn't found yet (class name is `elf` should be `self` and mb that we should just use a reference to the class itself instead of just leaving this error behind).

Attempt to reproduce linked issues: https://github.com/soyuka/self-referencing-metadata-api-platform

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -69,7 +69,7 @@ final class SchemaFactory implements SchemaFactoryInterface
     public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
     {
         $schema = $schema ?? new Schema();
-        if (null === $metadata = $this->getMetadata($className, $type, $operationType, $operationName, $serializerContext)) {
+        if (!class_exists($className) || null === $metadata = $this->getMetadata($className, $type, $operationType, $operationName, $serializerContext)) {
             return $schema;
         }
         [$resourceMetadata, $serializerContext, $inputOrOutputClass] = $metadata;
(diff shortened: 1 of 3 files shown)


src/JsonSchema/SchemaFactory.php
```diff
@@ -69,7 +69,7 @@ public function addDistinctFormat(string $format): void
     public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
     {
         $schema = $schema ?? new Schema();
-        if (null === $metadata = $this->getMetadata($className, $type, $operationType, $operationName, $serializerContext)) {
+        if (!class_exists($className) || null === $metadata = $this->getMetadata($className, $type, $operationType, $operationName, $serializerContext)) {
```
soyuka: I don't think that it's a good idea to hide the fact that a non-existing class went through here.

teohhanhui: How do we end up with a bad class? Isn't this because of corruption in `phpdocumentor/reflection-docblock` and `symfony/property-info`?

soyuka: Yes but I first tried to fix it in Api Platform before I could trace it back to the problem in phpdocumentor used by property-info...

PR #3351, opened 2020-01-20
Add date interval to type factory

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | na
| License       | MIT

Add the DateInterval class as duration in json schemas

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -87,6 +87,12 @@ final class TypeFactory implements TypeFactoryInterface
                 'format' => 'date-time',
             ];
         }
+        if (is_a($className, \DateInterval::class, true)) {
+            return [
+                'type' => 'string',
+                'format' => 'duration',
+            ];
+        }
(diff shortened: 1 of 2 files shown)


dunglas:
Isn't this a new feature?

teohhanhui:
I think this is a bug fix because we were missing the special handling for `\DateInterval`, which led to #3343 (even though the behaviour observed there was probably caused by #3349).

src/JsonSchema/TypeFactory.php
```diff
@@ -93,6 +93,12 @@ private function getClassType(?string $className, string $format = 'json', ?bool
                 'format' => 'uuid',
             ];
         }
+        if (is_a($className, \DateInterval::class, true)) {
```
teohhanhui: I think we could move this block one item up.

teohhanhui:
Needs rebase.

teohhanhui:
Thanks @soyuka! :tada:

PR #3356, opened 2020-01-22
Fix first entry is beign skipped when after...

### LINKED ISSUE #3355: GraphQL Pagination skips first element, when there is parameter `after` set and it's left as empty or set as zero.

**Description**  
GraphQL Pagination skips first element, when there is parameter `after` set and it's left as empty or set as zero.

**How to reproduce**  
```graphql
query listAll {
  products(name: "", after: "") {
    pageInfo {
      hasNextPage,
      hasPreviousPage,
    }
    totalCount,
    edges {
      cursor,
      node {
        id,
        name,
        ean,
        price,
        expiring, 
      }
    }
  }
}
```

or

```graphql
query listAll {
  products(name: "", after: "MA==") {
    pageInfo {
      hasNextPage,
      hasPreviousPage,
    }
    totalCount,
    edges {
      cursor,
      node {
        id,
        name,
        ean,
        price,
        expiring, 
      }
    }
  }
}
```

**Possible Solution**  
If parameter empty, return also first (`MA==`) entry also. Do not return one "after" it. If it is set to 0 (`MA==`), return 1 (`MQ==`) as first entry.
Posting a PR soon.

NikoGrano:
Resolves #3355

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3355 
| License       | MIT
| Doc PR        | None

Fixes issue where graphql parameter `after` is set, but empty string and return skips first value of collection.

### DIFF
--- a/src/GraphQl/Resolver/Stage/SerializeStage.php
+++ b/src/GraphQl/Resolver/Stage/SerializeStage.php
@@ -125,7 +125,7 @@ final class SerializeStage implements SerializeStageInterface
         $totalItems = $collection->getTotalItems();
         $nbPageItems = $collection->count();
         if (isset($args['after'])) {
-            $after = base64_decode($args['after'], true);
+            $after = '' === $args['after'] ? -1 : base64_decode($args['after'], true);
             if (false === $after) {
                 throw Error::createLocatedError(sprintf('Cursor %s is invalid', $args['after']), $info->fieldNodes, $info->path);
             }


NikoGrano:
Force push should fix CodeStyle issues and keep everything in one commit.

NikoGrano:
Closed based on discussion #3355. Will be opening new PR with different approach.

Issue #3361, opened 2020-01-23
Fix TypeFactory getClassType


Raulnet:
**API Platform version(s) affected**: x.y.z

**Description**  
in class 'ApiPlatform\Core\JsonSchema\TypeFactory'
the method getClassType could return an error if the argument '$classname' is not a resource and if the argument '$readablelink' is null;

teohhanhui:
Can you give an example?

teohhanhui:
ping @Raulnet

> could return an error

Could you provide more details please?

PR #3362, opened 2020-01-23
fix getClassType with no readableLink

Raulnet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | [issue 3361](https://github.com/api-platform/core/issues/3361)
| License       | MIT

Fix condition

Additionally:
 - Always add tests and ensure they pass.
 - Never break backward compatibility.
 - Bug fixes should be based against the current stable version branch.
 - Features and deprecations must be submitted against master branch.
 - Legacy code removals go to the master branch.
 - Update CHANGELOG.md file.
-->

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -99,11 +99,15 @@ final class TypeFactory implements TypeFactoryInterface
             return ['type' => 'string'];
         }
 
-        if ($this->isResourceClass($className) && true !== $readableLink) {
-            return [
-                'type' => 'string',
-                'format' => 'iri-reference',
-            ];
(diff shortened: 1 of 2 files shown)


dunglas:
Thanks @Raulnet!

teohhanhui:
I don't think this is the right fix. The logic seems wrong to me.

teohhanhui:
ping @dunglas @Raulnet

soyuka:
I tend to agree it's "too simple to be true" ^^

teohhanhui:
> "too simple to be true"

Well, I can confirm that it's wrong based on the condition.

PR #3366, opened 2020-01-28
allow to use output walkers from subresource operations

blanselle:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

pagination_use_output_walkers options is not used in case of subresourceOperations.


src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -353,6 +353,10 @@ private function shouldDoctrinePaginatorUseOutputWalkers(QueryBuilder $queryBuil
             if (null !== $useOutputWalkers = $resourceMetadata->getCollectionOperationAttribute($operationName, 'pagination_use_output_walkers', null, true)) {
```
soyuka: We should check if it's a subresource above and then do the right call accordingly

Issue #2782, opened 2019-05-09
Field incorrectly shows up as hydra @context but is not included in @vocab


weaverryan:
Hi!

Apologies if I'm misunderstanding something :); Suppose you have this situation:

```php
/**
 * @ApiResource(
 *     normalizationContext={"groups"={"cheese_listing:read"}},
 *     denormalizationContext={"groups"={"cheese_listing:write"}
 * )
 * @ORM\Entity()
 */
class CheeseListing
{
    // ...

    private $isPublished;

    public function getIsPublished(): ?bool
    {
        return $this->isPublished;
    }

    public function setIsPublished(bool $isPublished): self
    {
        $this->isPublished = $isPublished;
    }
}
```

This has an `isPublished` field, which is *not* included in the normalization or denormalization context. This field is simply not part of the API in any way. However, if you go to: https://localhost:8000/api/contexts/cheese_listings, it *is* present:

```json
{
  "@context": {
    "@vocab": "https://localhost:8000/api/docs.jsonld#",
    "hydra": "http://www.w3.org/ns/hydra/core#",
    "isPublished": "cheese_listings/isPublished"
  }
}
```

I believe the `isPublished` field should *not* be there. And, to further my point, if you go to https://localhost:8000/api/docs.jsonld to look at the vocab, there is *no* `hydra:property` for `cheese_listings/isPublished`. In other words, the `isPublished` property in the hydra `@context` is referring to a property that doesn't exist anywhere.

**Note**: If I'm correct that the `isPublished` property should not be added to the `cheese_listings` context, things get more interesting if you add a filter referencing the field:

```
* @ApiFilter(BooleanFilter::class, properties={"isPublished"})
```

What should happen here? This adds a `hydra:search` to the collection resource like this:

```json
 "hydra:search": {
    "@type": "hydra:IriTemplate",
    "hydra:template": "/api/cheeses{?isPublished}",
    "hydra:variableRepresentation": "BasicRepresentation",
    "hydra:mapping": [
      {
        "@type": "IriTemplateMapping",
        "variable": "isPublished",
        "property": "isPublished",
        "required": false
      }
    ]
  }
```

Should the presence of this filter *now* cause the property to be defined in https://localhost:8000/api/docs.jsonld - because it's referenced under the `hydra:mapping`?

Thanks!

soyuka:
[...]

PR #3367, opened 2020-01-28
Use serializer context for Hydra and OpenAPI documentation

### LINKED ISSUE #1835: [RFC] Use custom validation groups to build swagger doc
```
I have a bunch of dynamic role based normalization/denormalization groups. However, the swagger always seems to generate based on the default groups (or at least static ones in the entity files). This creates problems both with the swagger docs not being accurate for what you can do and/or make using the react admin difficult.

For example, I don't allow the username to be changed after sign-up (by the user). This is a dynamic de-normalization group that only allows that field when signing up (the admin can change it anytime). It works great, but it means username doesn't appear as a field option in the swagger document anywhere even in the POST method for the collection. (it does appear as a read-only field because I made a static group for public read only).

My suggestion is you allow the user to specify normalization/denormalization groups for document generation (overriding the functional ones). That way I can specify all the groups that should be documented and it would use those instead of the functional ones. (that way I can specify dynamic groups that should be documented but are restricted by user dynamically).

Is this doable? Thoughts? Ask any questions. It seems like changing the context's used to build the swagger shouldn't be all that difficult.

--Derrek
```

from @dleute https://github.com/api-platform/api-platform/issues/518

### LINKED ISSUE #2649: Adding a dynamic serialization group on a field will not update the entity model
Could be related to #1948.

When adding a denormalization serialization group dynamically on a field with the method of the [documentation](https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-dynamically) it will not show up in the POST / PUT models in the Swagger UI. Furthermore, when using the Admin component you can not create or edit the field. However, when directly calling the API it will serialize the field correctly.

Screenshot from 2019-03-25 16-39-14

I created a simple reproducer in this [repository](https://github.com/nielsvandermolen/reproducer-apip-serialization-groups). 

Is there something I am missing or is this a bug?

### LINKED ISSUE #3146: Serialization context is ignored when building JsonSchema
Unless I am wrong, when an `ApiResource` field can be serialized within different context, the context is ignored - so the second time a schema is generated, the previous definition is taken into account.

For instance, in
```
/** 
   ...
   @Groups({"g1", "g2"})
*/
public $myfield;
```

The schema definition will be corrected for the first serialization group used, but not for the second.

### LINKED ISSUE #3365: Allow custom serialization groups when generating hydra documentation
**Description**  
While working with API Platform, sometimes it happened to me that I had resources with some properties being serialized only with dynamic serialization groups (i.e. serialization groups that are set at request-time, for example depending on the user's role). 

Example resource:

```php
/**
 * @ApiResource(
 *     normalizationContext={"groups":{"customer:read"}},
 *     denormalizationContext={"groups":{"customer:write"}},
 * )
 */
class Customer {
    /**
     * @Groups({"customer:read","customer:write"})
     */
    private $property;

    /**
     * @Groups({"customer:admin:read","customer:admin:write"})
     */
    private $otherProperty;
}
```

In this example, `otherProperty` would be serialized only if an admin was making the request.

This works perfectly at the serialization level, however the problem I have is with the Hydra documentation (and consequently, the clients we can generate using the client generator), as all the properties that do not have the "default" serialization groups (specified in the `@ApiResource` annotation) are not exposed in the generated Hydra documentation. Because of that, I always have to add fields to the generated interfaces (I'm using the typescript generator and another custom generator for Angular I wrote) to "keep up" with these new fields with non default serialization groups.

**Proposal**: It would be great if we could specify additional serialization groups in the querystring of the `/api/docs.jsonld` endpoint. To allow more flexibility, the endpoint should accept multiple regular expressions that are then matched against all the serialization groups used in the resource (as extracted from Symfony's `Symfony\Component\Serializer\Mapping\ClassMetadataInterface`). Example: `/api/docs.jsonld?serializerGroups[]=.*:admin:.*&serializerGroups[]=.*otherRegex*`
This would make the DocumentationNormalizer use all serialization groups that match the given regexes, in addition to the default ones, and thus exposing any additional field that used only that serialization groups.

If the proposal seems good for you, I think I might be able to implement it (at least for the Hydra documentation, as I only had a look at that at the moment)

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tickets       | fixes #3365, fixes #2649, fixes #1835, fixes #3146, fixes https://github.com/api-platform/api-platform/issues/1072, fixes https://github.com/api-platform/api-platform/issues/866, https://github.com/api-platform/core/issues/2782
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1041

Alternative of https://github.com/api-platform/core/pull/2260.

A long-standing bug: the Hydra and OpenAPI documentation were not taking into account the dynamic serialization groups set by the serializer context builder.

For instance, with the Admin, there was no difference between a connected user and an anonymous one, if the connected user had access to more fields of the resources.

I had to introduce two changes in order to make it work:
- a new property factory option: `deserializer_groups`. It's needed by the `SerializerPropertyMetadataFactory` to make the difference between serialization and deserialization in order to fill correctly the property metadata (readable / writable).
- a special operation name for the `SerializerContextBuilder`: `resource_operation_name`. It's needed to make sure the serializer context builder uses the attributes of the resource (and not of the item or the collection).

This behavior is **not** enabled by default, a `enable_serialization_context_doc` flag in the configuration has to be set to `true` to enable it.

The serializer context builder has been replaced by a serializer context provider, not depending of the request.

@desmax maybe you are interesting by this PR.


dunglas:
Nice! This option should be opt-in (and not enabled by default). It is often useful to discover all the capabilities of the API. For instance, it allows the (smart) client to expose all available features to the end user, including the one requiring to be authenticated, or another role. The typical use case is to display a login form when a 401 is returned at runtime.

dunglas:
> Sadly, I had to use some workaround in order to use the serializer context builder.

I suggest to deprecate both the REST and the GraphQL interfaces, and to replace them by an unified one for both REST and GraphQL (it will just be a copy of the existing GraphQL interface). WDYT?

alanpoulain:
It's not the case today. Only the public readable properties are exposed. This PR doesn't modify this behavior (I didn't change the tests). It only adds more information when the user is connected. I don't think it should be opt-in, because it's the users' wanted behavior (see the number of issues!).

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -64,6 +69,12 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
         $this->urlGenerator = $urlGenerator;
         $this->subresourceOperationFactory = $subresourceOperationFactory;
         $this->nameConverter = $nameConverter;
+        if (!$serializerContextBuilder) {
+            @trigger_error(sprintf('Not passing an instance of %s to %s() is deprecated since version 2.5 and will be removed in 3.0.', SerializerContextBuilderInterface::class, __METHOD__), E_USER_DEPRECATED);
```
dunglas: Is this deprecation mandatory? Havin as few mandatory deps as possible is always better, it allows to reuse the class easily when installing API Platform as a standalone library.

alanpoulain:
> I suggest to deprecate both the REST and the GraphQL interfaces, and to replace them by an unified one for both REST and GraphQL (it will just be a copy of the existing GraphQL interface). WDYT?

Yes, I think it would be nice to do that. There is a lot of changes to do though, and I don't think it would be good to do this in this PR.
The workaround is easily removable later (and it will be done naturally since there will be no more `$request` parameter).

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -64,6 +69,12 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
...
+            $this->serializerContextBuilder = new SerializerContextBuilder($this->resourceMetadataFactory);
```
dunglas: So,

```suggestion
            $this->serializerContextBuilder = $serializerContextBuilder ?? new SerializerContextBuilder($this->resourceMetadataFactory);
```

src/JsonSchema/SchemaFactory.php
```diff
@@ -35,22 +38,30 @@
...
+    public function __construct(TypeFactoryInterface $typeFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null, SerializerContextBuilderInterface $serializerContextBuilder = null)
     {
         $this->typeFactory = $typeFactory;
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->propertyNameCollectionFactory = $propertyNameCollectionFactory;
         $this->propertyMetadataFactory = $propertyMetadataFactory;
         $this->nameConverter = $nameConverter;
         $this->resourceClassResolver = $resourceClassResolver;
+        if (!$serializerContextBuilder) {
+            @trigger_error(sprintf('Not passing an instance of %s to %s() is deprecated since version 2.5 and will be removed in 3.0.', SerializerContextBuilderInterface::class, __METHOD__), E_USER_DEPRECATED);
```
dunglas: Same here.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -155,6 +155,12 @@ private function getEffectiveSerializerGroups(array $options, string $resourceCl
         if (isset($options['serializer_groups'])) {
             $groups = (array) $options['serializer_groups'];
 
+            if (isset($options['deserializer_groups'])) {
```
dunglas: Couldn't `deserializer_groups` be set while `serializer_groups` aren't? To cover this case, shouldn't this test be outside ` if (isset($options['serializer_groups']))`?

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -155,6 +155,12 @@ private function getEffectiveSerializerGroups(array $options, string $resourceCl
...
+                $deserializerGroups = (array) $options['deserializer_groups'];
```
dunglas: `serialization_groups` and `deserialization_groups` or `normalization_groups` and `denormalization_context` (as `(de)normalization_context` that will already use will probably be more consistent; but it's just some nitpicking, I don't think deprecating the existing option is worth it.

src/Serializer/PropertyFactoryOptionsTrait.php
```diff
@@ -0,0 +1,57 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Serializer;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
+
+/**
+ * Manipulate the property factory options.
```
dunglas:
```suggestion
 * Manipulates the property factory options.
```

src/Serializer/PropertyFactoryOptionsTrait.php
```diff
@@ -0,0 +1,57 @@
...
+ *
```
dunglas:
```suggestion
 * @internal
```

src/Serializer/PropertyFactoryOptionsTrait.php
```diff
@@ -0,0 +1,57 @@
...
+ * @author @author Alan Poulain <contact@alanpoulain.eu>
+ */
+trait PropertyFactoryOptionsTrait
+{
+    /**
+     * Gets the options for a property factory.
+     */
+    private function getPropertyFactoryOptions(string $resourceClass, bool $splitDeserialization = false): array
+    {
+        $serializationContext = $this->serializerContextBuilder->createFromRequest(Request::createFromGlobals(), true, [
```
dunglas: This isn't reliable: super globals aren't available in CLI, and the documentation is available in CLI. We should use a new interface not depending of `Request` as suggested in the comments.

Also, the result of `Request::createFromGlobals()` should be stored in a variable to prevent calling the method t two times.

src/Serializer/SerializerContextBuilder.php
```diff
@@ -52,12 +52,19 @@ public function createFromRequest(Request $request, bool $normalization, array $
         } elseif (isset($attributes['item_operation_name'])) {
             $operationKey = 'item_operation_name';
             $operationType = OperationType::ITEM;
+        } elseif (isset($attributes['resource_operation_name'])) {
```
dunglas: `docs_operation_name`? `metadata_operation_name`? `global_operation_name`?

src/Serializer/SerializerContextBuilder.php
```diff
@@ -52,12 +52,19 @@ public function createFromRequest(Request $request, bool $normalization, array $
...
+            $operationKey = 'resource_operation_name';
+            $operationType = 'resource';
```
dunglas: Then we should introduce a constant in `OperationType` as for other types.

src/Serializer/SerializerContextBuilder.php
```diff
@@ -52,12 +52,19 @@ public function createFromRequest(Request $request, bool $normalization, array $
...
         } else {
             $operationKey = 'subresource_operation_name';
             $operationType = OperationType::SUBRESOURCE;
         }
 
-        $context = $resourceMetadata->getTypedOperationAttribute($operationType, $attributes[$operationKey], $key, [], true);
+        if ('resource' !== $operationType) {
```
dunglas: Invert the test for readability (start with `===`)

src/Serializer/SerializerContextBuilder.php
```diff
@@ -52,12 +52,19 @@ public function createFromRequest(Request $request, bool $normalization, array $
...
+            $context = $resourceMetadata->getTypedOperationAttribute($operationType, $attributes[$operationKey], $key, [], true);
```
dunglas: It would be better to add support the new type in `getTypedOperationAttribute()`.

src/Serializer/PropertyFactoryOptionsTrait.php
alanpoulain: Actually, we don't really care of the data given by the request. It's only a workaround to make it work.

dunglas:
@alanpoulain we don't speak about the same thing. It's intended to filter properties that are internal and must never be available publicly. However, here we're speaking of another topic: dynamic filtering depending of the context (currently logged in user, type of client, the current date...). In most case, the documentation should document **all** the capabilities of the API. Even the ones available only in some contexts, for some users. Then HTTP provides mechanism to negotiate a runtime how to trigger this feature (for instance, a 403 status code will be returned if the user hasn't sufficient credentials to trigger this feature).

> It only adds more information when the user is connected.

Yes, but this valuable information is very valuable for smart clients (what capabilities are available for logged in users for instance). I'm not sure if this feature should be enabled or disabled by default. I fear that enabling it by default will cause some BC breaks.

For Hydra at least, not available properties shouldn't be hidden from the spec, because it's in the spec that these properties MAY NOT be returned according to the current context (user permissions etc).
For OpenAPI, we can maybe use the `security requirements` object (I'm not sure, I need to study this more in depth): https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#securityRequirementObject

dunglas:
> Yes, I think it would be nice to do that. There is a lot of changes to do though, and I don't think it would be good to do this in this PR.

I agree that it should be done in another PR, but then this PR must be done and merged before this one. If we don't do it this way, We'll have to add a BC layer everywhere in this new code to support both the deprecated and the new interface. Also, see my comments, the current implementation is partially broken (in CLI) and doesn't respect the context separation (`Request` shouldn't be used outside of the controller layer).

src/Serializer/PropertyFactoryOptionsTrait.php
dunglas: Yes but it's ugly... and can be avoided.

src/Serializer/SerializerContextBuilder.php
alanpoulain: I think `resource` is clearer. In documentation, we call the corresponding attributes the resource ones (for instance: https://api-platform.com/docs/core/content-negotiation/#configuring-formats-for-a-specific-resource-or-operation => `can also be configured at resource and operation level`).
Since the attributes are directly below the `ApiResource` annotation, I think it makes sense.

src/Serializer/SerializerContextBuilder.php
alanpoulain: It's not really an operation, that's why I don't think it's a good idea to add it to `OperationType`.
Moreover, I can't add it to the `TYPES`, since there is no service corresponding to it.

alanpoulain: I've added it anyway but I find it strange.

src/Serializer/SerializerContextBuilder.php
alanpoulain: Like the previous comment, I'm not sure it's a good idea to do this, since it's a "special" operation (for instance the resource fallback parameter doesn't make sense).
I've changed it anyway.

src/Serializer/SerializerFilterContextProvider.php
```diff
@@ -0,0 +1,67 @@
...
+            return $serializerContext;
+        }
+
+        foreach ($resourceFilters as $filterId) {
+            if ($this->filterLocator->has($filterId)) {
+                $filter = $this->filterLocator->get($filterId);
+                if ($filter instanceof SerializerContextFilterInterface) {
+                    $filter->applyToSerializerContext($resourceClass, $operationName, $normalization, $context, $serializerContext);
+                } elseif ($filter instanceof FilterInterface) {
+                    throw new RuntimeException(sprintf('The filter "%s" implements the "%s" interface but "%s" is only compatible with filters implementing the "%s" interface.', \get_class($filter), FilterInterface::class, __CLASS__, SerializerContextFilterInterface::class));
```
alanpoulain: It would be hard to keep a strict BC policy here, because we would need to get the request.
However, this exception should be thrown only if the user has created a custom serializer filter or has decorated an existing one (group or property).

alanpoulain:
@teohhanhui is against this PR, because of what @dunglas said, and would like the documentation to have *all* the available fields. The data for a field may not be returned if the user is not allowed to see it.

However I see some issues with this:
- it would be a big BC to make the previously hidden fields in the documentation publicly shown. If we make this feature opt-in, I don't think a lot of users would use it;
- some clients (like our admin) use the documentation to generate the interface. Doing this would mean to change how they render the interface: they would need to use the documentation and to change the interface depending of the received data (if the field is not present, then the column should not be displayed). It would be more complex.

The approach of this PR is to solve the problem in a practical way. A lot of users don't understand why it behaves like this, and that's why I tried to fix it.

I would like the opinion of the @api-platform/core-team. What should we do? I don't think letting the situation as-is is a good idea, we should make a choice to resolve this issue.

teohhanhui:
> it would be a big BC to make the previously hidden fields in the documentation publicly shown

The user would have to do something to make this happen anyway, due to the dynamic nature of it.

alanpoulain:
After some thinking, I think the best outcome for API Platform is the following:
- we keep this PR as is with the `enable_serialization_context_doc` configuration flag,
- in another PR, we add a new `enable_expose_all_properties` configuration flag. When this flag is enabled, all the properties are exposed, even the hidden ones.
- since this flag will probably not be used by users, we make it the default in the recipe.
- we deprecate not setting the configuration flag by saying, it will be `true` by default in API Platform 3.

With this path, the final user can do what it wants. By default, all the properties will be exposed, and the context will not change the documentation. But if the user wants the documentation to change depending of the context and not all the properties to be publicly exposed, it can. This will allow to handle it step by step in the admin too.

WDYT @api-platform/core-team?

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
```diff
@@ -40,25 +43,28 @@
...
     {
-        if (null !== $this->requestStack) {
+        if (null !== $requestStack) {
             @trigger_error(sprintf('Passing an instance of "%s" is deprecated since version 2.2 and will be removed in 3.0. Use the data provider\'s context instead.', RequestStack::class), E_USER_DEPRECATED);
         }
-        if (null !== $this->serializerContextBuilder) {
-            @trigger_error(sprintf('Passing an instance of "%s" is deprecated since version 2.2 and will be removed in 3.0. Use the data provider\'s context instead.', SerializerContextBuilderInterface::class), E_USER_DEPRECATED);
+        if (null !== $serializerContextBuilder) {
+            @trigger_error(sprintf('Passing an instance of "%s" or "%s" is deprecated since version 2.2 and will be removed in 3.0. Use the data provider\'s context instead.', SerializerContextBuilderInterface::class, SerializerContextFactoryInterface::class), E_USER_DEPRECATED);
```
dunglas: We don't need to add support for the new `SerializerContextFactoryInterface` as passing a context factory is deprecated anyway and shouldn't be done. I think you can remove the changes in this file.

src/EventListener/DeserializeListener.php
```diff
@@ -32,22 +34,24 @@
...
-     * @param ResourceMetadataFactoryInterface|FormatsProviderInterface|array $resourceMetadataFactory
+     * @param SerializerContextBuilderInterface|SerializerContextFactoryInterface $serializerContextBuilder
+     * @param ResourceMetadataFactoryInterface|FormatsProviderInterface|array     $resourceMetadataFactory
      */
-    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, $resourceMetadataFactory, ResourceMetadataFactoryInterface $legacyResourceMetadataFactory = null)
+    public function __construct(SerializerInterface $serializer, /* SerializerContextFactoryInterface $serializerContextFactory */$serializerContextBuilder, $resourceMetadataFactory, ResourceMetadataFactoryInterface $legacyResourceMetadataFactory = null)
     {
         $this->serializer = $serializer;
-        $this->serializerContextBuilder = $serializerContextBuilder;
+        $this->serializerContextFactory = $serializerContextBuilder;
```
dunglas: Missing deprecation?

src/EventListener/SerializeListener.php
```diff
@@ -35,17 +37,21 @@
...
 
-    public function __construct(SerializerInterface $serializer, SerializerContextBuilderInterface $serializerContextBuilder, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
+    /**
+     * @param SerializerContextBuilderInterface|SerializerContextFactoryInterface $serializerContextBuilder
+     */
+    public function __construct(SerializerInterface $serializer, /* SerializerContextFactoryInterface $serializerContextFactory */$serializerContextBuilder, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
     {
         $this->serializer = $serializer;
-        $this->serializerContextBuilder = $serializerContextBuilder;
+        $this->serializerContextFactory = $serializerContextBuilder;
```
dunglas: Missing deprecation?

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -49,8 +52,9 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
...
     private $subresourceOperationFactory;
     private $nameConverter;
+    private $serializerContextFactory;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceClassResolverInterface $resourceClassResolver, OperationMethodResolverInterface $operationMethodResolver = null, UrlGeneratorInterface $urlGenerator, SubresourceOperationFactoryInterface $subresourceOperationFactory = null, NameConverterInterface $nameConverter = null)
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceClassResolverInterface $resourceClassResolver, OperationMethodResolverInterface $operationMethodResolver = null, UrlGeneratorInterface $urlGenerator, SubresourceOperationFactoryInterface $subresourceOperationFactory = null, NameConverterInterface $nameConverter = null, ?SerializerContextFactoryInterface $serializerContextFactory = null)
```
dunglas:
```suggestion
    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, ResourceClassResolverInterface $resourceClassResolver, OperationMethodResolverInterface $operationMethodResolver = null, UrlGeneratorInterface $urlGenerator, SubresourceOperationFactoryInterface $subresourceOperationFactory = null, NameConverterInterface $nameConverter = null, SerializerContextFactoryInterface $serializerContextFactory = null)
```

src/JsonSchema/SchemaFactory.php
```diff
@@ -35,22 +36,25 @@
...
 
     private $typeFactory;
     private $propertyNameCollectionFactory;
     private $propertyMetadataFactory;
     private $nameConverter;
     private $distinctFormats = [];
+    private $serializerContextFactory;
 
-    public function __construct(TypeFactoryInterface $typeFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null)
+    public function __construct(TypeFactoryInterface $typeFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null, ?SerializerContextFactoryInterface $serializerContextFactory = null)
```
dunglas:
```suggestion
    public function __construct(TypeFactoryInterface $typeFactory, ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null, SerializerContextFactoryInterface $serializerContextFactory = null)
```

src/Serializer/ContextTrait.php
```diff
@@ -30,4 +35,34 @@ private function initContext(string $resourceClass, array $context): array
             'resource_class' => $resourceClass,
         ]);
     }
+
+    private function addRequestContext(Request $request, array $context): array
+    {
+        return array_merge($context, [
```
dunglas: Why  do we need to add so much data to the context? We're creating an indirect coupling to HTTP here, and the whole point of this change is to decouple from HTTP.

dunglas: Actually, HTTP-related data must not leak outside of `ReadListener`. Filters are already extracted in this class, there are no needs for these new keys.

src/Serializer/Filter/GroupFilter.php
```diff
@@ -60,6 +64,32 @@ public function apply(Request $request, bool $normalization, array $attributes,
         $context[AbstractNormalizer::GROUPS] = $groups;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function applyToSerializerContext(string $resourceClass, string $operationName, bool $normalization, array $context, array &$serializerContext): void
+    {
+        if (\array_key_exists($this->parameterName, $commonAttribute = ($context['request_attributes']['_api_filters'] ?? []))) {
```
dunglas: Filters should already be available in `$context`.

teohhanhui:
> new `enable_expose_all_properties` configuration flag. When this flag is enabled, all the properties are exposed, even the hidden ones.

How do you propose doing that? We cannot know the set of all possible properties as the serializer context / serializer groups can be dynamic. This doesn't seem technically feasible to me.

src/Api/OperationType.php
```diff
@@ -18,5 +18,6 @@ final class OperationType
     public const ITEM = 'item';
     public const COLLECTION = 'collection';
     public const SUBRESOURCE = 'subresource';
+    public const RESOURCE = 'resource';
```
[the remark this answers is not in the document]
alanpoulain: "Metadata" is even weirder IMO since the operation types are used to get specific properties in the resource metadata.
I see it more like a level of the metadata: at the `resource` level, at the `item` (operation) level.
Maybe it's not the best name, but I don't have another one.

src/EventListener/DeserializeListener.php
alanpoulain: Since the service and the method (`createFromRequest`) are deprecated, I didn't thought it was necessary.

src/Hydra/Serializer/DocumentationNormalizer.php
alanpoulain: I made it nullable (and not just with a default `null` value, even if PHP make it nullable in this case) because a `null` value can be given, even if the BC was not handled.

src/Serializer/ContextTrait.php
alanpoulain: Please see the `GroupFilter` and the `PropertyFilter`. These filters are very coupled to the request data.
The `SerializerContextFactory` needs also a lot of request data in order to build the serializer context (related to `api_allow_update`, to CSV, ...).
It also adds the `request_uri` key used in some serializers.
I don't know how you would like to change all this.

src/Serializer/Filter/GroupFilter.php
alanpoulain: I can check this :+1:

src/Bridge/Doctrine/Orm/Extension/EagerLoadingExtension.php
alanpoulain: They were some bugs with the deprecations though (that's why the `$this` has been removed).

src/Api/OperationType.php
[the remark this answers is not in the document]
teohhanhui: Hmm... Isn't it akin to the HTTP `HEAD` method?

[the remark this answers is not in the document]
alanpoulain: Hmm not really? It's mainly used by the `getTypedOperationAttribute` of the `ResourceMetadata` to retrieve the attributes of the resource, and not the attributes of a specific operation.

src/Serializer/Filter/GroupFilter.php
alanpoulain: Seems OK to do it. There is not related Behat tests though.

src/Api/OperationType.php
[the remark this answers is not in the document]
teohhanhui: Okay, then that's very hackish. An "operation type" that's not actually an operation type. :see_no_evil:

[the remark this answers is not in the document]
alanpoulain: Kind of, yeah... If you have a better idea, I'm taking it!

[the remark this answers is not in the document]
dunglas: It definitely must not be resource. It's misleading.

src/Hydra/Serializer/DocumentationNormalizer.php
dunglas: But it's not consistent with what we currently do in API Platform (and Symfony). When we'll change this, we'll change it everywhere in one batch using PHP CS Fixer.

src/Serializer/ContextTrait.php
dunglas: The `request_stack` should be injected directly in those specific filters. But copying the full HTTP request in the context isn't a good idea.

src/Hydra/Serializer/DocumentationNormalizer.php
alanpoulain: OK then. Too bad.

src/Serializer/ContextTrait.php
alanpoulain: I can do it for the filters, but not for the other cases.

alanpoulain: Also, I think a direct coupling with the request stack is not really a good idea, especially if we want to be able to manage other frameworks than Symfony.

src/Api/OperationType.php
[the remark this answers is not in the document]
alanpoulain: I don't understand why it's misleading. We set the attributes at the *resource* level. I don't think we ever used another term to describe it.
We don't say at the "global" level. Or at the "metadata" level. Nor at the "root" level. Even it's not a perfect fit, because by essence it's not really an "operation" (and we can ask ourselves what a "subresource" operation is...), I don't think there is a better alternative.

src/Serializer/ContextTrait.php
alanpoulain: Another thing I thought: a lot of users **need** to have the request context for creating their own context factory. You can take this question in SO as an example: https://stackoverflow.com/questions/60340261/how-to-set-normalization-groups-based-on-users-role-and-request-method/60346166#60346166

teohhanhui:
Looking at #3146, I don't think this is the right fix for that.

alanpoulain:
@teohhanhui maybe not this PR? It isn't exactly the same bug at the others, but maybe this PR fixes it anyway with some luck.

PR #3368, opened 2020-01-29
Ensure that `api:openapi:export` produces JSON with unescaped forward slashes

Ocramius:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

This ensures that the generated OpenAPI schema JSON dump does not contain forward-slashes, such as `"#\/foo\/bar\/MyModel"`, which confuses tooling relying on direct string match in referenced models (most notably, the IntelliJ OpenAPI/Swagger tooling)

### DIFF
--- a/tests/Bridge/Symfony/Bundle/Command/SwaggerCommandUnitTest.php
+++ b/tests/Bridge/Symfony/Bundle/Command/SwaggerCommandUnitTest.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Command;
+
+use ApiPlatform\Core\Bridge\Symfony\Bundle\Command\SwaggerCommand;
+use ApiPlatform\Core\Documentation\Documentation;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceNameCollection;
+use PHPUnit\Framework\MockObject\MockObject;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+class SwaggerCommandUnitTest extends KernelTestCase
+{
+    /** @var MockObject&NormalizerInterface */
+    private $normalizer;
+
+    /** @var ResourceNameCollectionFactoryInterface&MockObject */
+    private $resources;
+
+    /** @var SwaggerCommand */
+    private $command;
+
+    protected function setUp(): void
+    {
+        $this->normalizer = $this->createMock(NormalizerInterface::class);
+        $this->resources = $this->createMock(ResourceNameCollectionFactoryInterface::class);
+        $this->command = new SwaggerCommand(
+            $this->normalizer,
+            $this->resources,
+            'My API',
+            'I told you already: it is my API',
+            'one-zero-zero'
+        );
+
+        $this->resources->method('create')
+            ->willReturn(new ResourceNameCollection());
+    }
+
+    public function testDocumentationJsonDoesNotUseEscapedSlashes(): void
+    {
+        $this->normalizer->method('normalize')
+            ->with(self::isInstanceOf(Documentation::class))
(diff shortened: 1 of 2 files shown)


tests/Bridge/Symfony/Bundle/Command/SwaggerCommandUnitTest.php
```diff
@@ -0,0 +1,69 @@
...
+use ApiPlatform\Core\Documentation\Documentation;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ResourceNameCollection;
+use PHPUnit\Framework\MockObject\MockObject;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+class SwaggerCommandUnitTest extends KernelTestCase
```
Ocramius: Note: a unit test was much simpler and efficient to write - the current integration test is not something I could work with anyway, since I don't run MongoDB on any system I maintain (including my dev environment) by choice

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandUnitTest.php
```diff
@@ -0,0 +1,69 @@
...
+            'My API',
+            'I told you already: it is my API',
+            'one-zero-zero'
+        );
+
+        $this->resources->method('create')
+            ->willReturn(new ResourceNameCollection());
+    }
+
+    public function testDocumentationJsonDoesNotUseEscapedSlashes()
```
alanpoulain:
```suggestion
    public function testDocumentationJsonDoesNotUseEscapedSlashes(): void
```

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandUnitTest.php
alanpoulain: Actually there was a an attempt to remove the hard dev dependency to MongoDB (https://github.com/api-platform/core/pull/2703), but it has been added again by https://github.com/api-platform/core/pull/2843 (an issue with PHPStan it seems).

By default though, the Behat tests are run without MongoDB.

Ocramius: I hope that can be achieved again at some point: for now, I'm happy with adding a unit test, and providing more unit tests in future, shall I encounter isolated issues, like this one.

alanpoulain:
Couldn't this be a BC if the client has been relying on the escaped slashes?

Ocramius:
Doubt that, but there is indeed a change in dumped definitions. Forward slashes do not impose an escaping sequence risk.

dunglas:
Thanks @Ocramius!

tests/Bridge/Symfony/Bundle/Command/SwaggerCommandUnitTest.php
teohhanhui: > PHP extensions are no longer required to be loaded - thanks to Roave/BetterReflection and PhpStorm stubs!

We should be able to do this once we upgrade to phpstan `^0.12.4`! :smile:

teohhanhui: Unfortunately, we still can't remove the MongoDB packages, because we'd get errors such as:

```
 ------ -------------------------------------------------------------------------------------------- 
  Line   src/Bridge/Doctrine/Common/Util/IdentifierManagerTrait.php (in context of anonymous class)  
 ------ -------------------------------------------------------------------------------------------- 
  47     Class Doctrine\ODM\MongoDB\DocumentManager not found.                                       
  86     Call to static method hasType() on an unknown class Doctrine\ODM\MongoDB\Types\Type.        
  87     Call to static method getType() on an unknown class Doctrine\ODM\MongoDB\Types\Type.        
 ------ --------------------------------------------------------------------------------------------
```

:disappointed:

teohhanhui: @ondrejmirtes Do you perhaps have any suggestions on how this could be resolved? I see there are MongoDB ODM stubs in https://github.com/phpstan/phpstan-doctrine/tree/master/stubs

Is there a way to tell phpstan that these types "exist" purely based on the stubs?

teohhanhui: See #3431 for an attempt.

ondrejmirtes: > Is there a way to tell phpstan that these types "exist" purely based on the stubs?

You'd have to make the class available for the autoloader. See: https://phpstan.org/user-guide/autoloading

PR #3375, opened 2020-01-31
Fix a bug preventing to serialize validator's payload

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n:a
| License       | MIT
| Doc PR        |n/a

The current behavior when dealing with validation error payload (severity etc) isn't in sync with what is expected and documented: https://api-platform.com/docs/core/validation/#error-levels-and-payload-serialization

This PR fixes this problem.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -94,7 +94,7 @@ final class Configuration implements ConfigurationInterface
                 ->arrayNode('validator')
                     ->addDefaultsIfNotSet()
                     ->children()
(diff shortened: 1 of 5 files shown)


teohhanhui:
Hmm... I'm not familiar with this, but using an empty array to mean "all fields" is so confusing...

teohhanhui:
I think we should fix the documentation instead. `null` should mean all fields, and the empty array (used as default value) should mean none. WDYT @dunglas?

dunglas:
@teohhanhui it would be a BC break

dunglas:
@teohhanhui you're right it's a better solution. Fixed.

PR #3376, opened 2020-02-03
Add possibility to hydrate input before denormalize

cacahouwete:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Add an interface to give the possibility to prehydrate input for datatransformer to make it possible to do partial update using input dto.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -38,6 +38,7 @@ use Symfony\Component\Serializer\Exception\UnexpectedValueException;
 use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface;
 use Symfony\Component\Serializer\NameConverter\AdvancedNameConverterInterface;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
 use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
@@ -191,14 +192,11 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                 throw new LogicException('Cannot denormalize the input because the injected serializer is not a denormalizer');
             }
             if ($dataTransformer instanceof PreHydrateInputInterface) {
-                $context[static::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
+                $context[AbstractNormalizer::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
             }
             $denormalizedInput = $this->serializer->denormalize($data, $inputClass, $format, $context);
             if (!\is_object($denormalizedInput)) {
                 throw new \UnexpectedValueException('Expected denormalized input to be an object.');
-            }
-            if ($dataTransformer instanceof PreHydrateInputInterface) {
-                $context[static::OBJECT_TO_POPULATE] = $objectToPopulate;
             }
 
             return $dataTransformer->transform($denormalizedInput, $resourceClass, $dataTransformerContext);
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -38,7 +38,6 @@ use Symfony\Component\Serializer\Exception\UnexpectedValueException;
 use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface;
 use Symfony\Component\Serializer\NameConverter\AdvancedNameConverterInterface;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
-use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
 use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
@@ -192,7 +191,8 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                 throw new LogicException('Cannot denormalize the input because the injected serializer is not a denormalizer');
             }
             if ($dataTransformer instanceof PreHydrateInputInterface) {
-                $context[AbstractNormalizer::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
+                $context[AbstractObjectNormalizer::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
+                $context[AbstractObjectNormalizer::DEEP_OBJECT_TO_POPULATE] = true;
             }
             $denormalizedInput = $this->serializer->denormalize($data, $inputClass, $format, $context);
             if (!\is_object($denormalizedInput)) {
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -189,9 +189,18 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             if (!$this->serializer instanceof DenormalizerInterface) {
                 throw new LogicException('Cannot denormalize the input because the injected serializer is not a denormalizer');
             }
+            if ($dataTransformer instanceof PreHydrateInputInterface) {
+                $context[static::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
+            }
             $denormalizedInput = $this->serializer->denormalize($data, $inputClass, $format, $context);
             if (!\is_object($denormalizedInput)) {
                 throw new \UnexpectedValueException('Expected denormalized input to be an object.');
+            }
+            if ($dataTransformer instanceof PreHydrateInputInterface) {
+                unset($context[static::OBJECT_TO_POPULATE]);
+            }
+            if (null !== $objectToPopulate) {
+                $context[static::OBJECT_TO_POPULATE] = $objectToPopulate;
             }
 
             return $dataTransformer->transform($denormalizedInput, $resourceClass, $dataTransformerContext);
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -17,6 +17,7 @@ use ApiPlatform\Core\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
+use ApiPlatform\Core\DataTransformer\PreHydrateInputInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\InvalidValueException;
 use ApiPlatform\Core\Exception\ItemNotFoundException;
@@ -197,9 +198,6 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                 throw new \UnexpectedValueException('Expected denormalized input to be an object.');
             }
             if ($dataTransformer instanceof PreHydrateInputInterface) {
-                unset($context[static::OBJECT_TO_POPULATE]);
-            }
-            if (null !== $objectToPopulate) {
                 $context[static::OBJECT_TO_POPULATE] = $objectToPopulate;
             }
 
--- a/src/Serializer/PreHydrateInputInterface.php
+++ b/src/Serializer/PreHydrateInputInterface.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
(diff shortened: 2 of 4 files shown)


src/Serializer/PreHydrateInputInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Serializer;
```
soyuka: should be in the DataTransformer namespace

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -186,10 +186,19 @@ public function denormalize($data, $class, $format = null, array $context = [])
...
+                $context[static::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
+            }
             $denormalizedInput = $this->serializer->denormalize($data, $inputClass, $format, $context);
             if (!\is_object($denormalizedInput)) {
                 throw new \UnexpectedValueException('Expected denormalized input to be an object.');
             }
+            if ($dataTransformer instanceof PreHydrateInputInterface) {
+                unset($context[static::OBJECT_TO_POPULATE]);
+            }
+            if (null !== $objectToPopulate) {
```
soyuka: this condition looks dangerous and should be merged with the previous `$dataTransformer instanceof PreHydrateInputInterface` no?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -186,10 +186,19 @@ public function denormalize($data, $class, $format = null, array $context = [])
             if (!$this->serializer instanceof DenormalizerInterface) {
                 throw new LogicException('Cannot denormalize the input because the injected serializer is not a denormalizer');
             }
+            if ($dataTransformer instanceof PreHydrateInputInterface) {
+                $context[static::OBJECT_TO_POPULATE] = $dataTransformer->createInput($inputClass, $context);
```
soyuka: I'm not sure to understand what you're trying to do here, this doesn't look to go the way we meant data transformers to work, could you explain your use case?

norkunas:
This is a nice feature because when you allow to relation be null or a object/resource class you can't detect if api consumer set the `null` explicitly or it was not provided at all on the "put" operation.
Borrowed this as a custom normalizer :)

src/Serializer/AbstractItemNormalizer.php
cacahouwete: I set this key value by the object we want to populate by the symfony serializer (https://symfony.com/doc/current/components/serializer.html#deserializing-in-an-existing-object)

src/Serializer/PreHydrateInputInterface.php
cacahouwete: Done

src/Serializer/AbstractItemNormalizer.php
cacahouwete: Done

soyuka:
Can you guys give me a functional example ? I'd like some tests / docs to illustrate a use case where this helps.

cacahouwete:
For example in my case, I have an API with inputDto to validate data and hydrate my entities. The problem is I need to put all data in the payload to update one field in the entity (a full payload for a partial update). If we could prehydrate the inputDto with the targeted entity ($context['object_to_populate']) we could do partial update with partial payload: we could send only the fields we want to update and keep the integrity of the input to validate it correctly. At the moment, if we want to do partial update we have to use the entity directly (and don't use inputDto).

To illustrate it I have theses two class:
```php
<?php
final class DummyForAdditionalFieldsInput
{
    /** @Assert\NotBlank() */
    public string $dummyName;
    /** @Assert\Length(min=10) */
    public ?string $dummyDescription = null;

    public function getDummyName(): string
    {
        return $this->dummyName;
    }

    public function getDummyDescription(): ?string
    {
        return $this->dummyDescription;
    }
}

class DummyForAdditionalFields
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;
    /** @ORM\Column */
    private $name;
    /** @ORM\Column */
    private $slug;
    /** @ORM\Column(nullable=true) */
    private $description;

    public function __construct(string $name, string $slug)
    {
        $this->name = $name;
        $this->slug = $slug;
    }

    public function getId(): ?int
    {
        return $this->id;
    }

    {
        return $this->name;
    }

    public function setName(string $name): void
    {
        $this->name = $name;
    }

    public function getSlug(): string
    {
        return $this->slug;
    }

    public function getDescription(): ?string
    {
        return $this->description;
    }

    public function setDescription(?string $description): void
    {
        $this->description = $description;
    }
}
```

If I want update entity DummyForAdditionalFields by an inputDTO actualy I have to send :
```json
{
    "dummyName": "Dummy Name",
    "dummyDescription": "New Dummy Description"
}
```
with following data in DummyForAdditionalFields :
| field | value |
| ----- | -------- |
| name | Dummy Name |
| description | Dummy Description |

In this case I send all payload to just modify description. If we can prehydrate InputDTO then dto will have name and description set before deserialize and then final possible payload to change only fields we want change will be:
```json
{
    "dummyDescription": "New Dummy Description"
}
```

teohhanhui:
I'm not sure I understand.

`$context['object_to_populate']` is set by `ReadListener`.

cacahouwete:
Yes this data is the entity we want change in transform of dataTransformer but when you call denormalize with $context[AbstractNormalizer::OBJECT_TO_POPULATE] = new MyEntity()
and $inputCass = MyEntityInput
symfony denormalizer will see that object to populate class !== $inputClass then he will ignore object to populate and create a new $inputClass().

If we can have the possibility to create and hydrate this object_to_populate before denormalize we could give him the good object and then hydrat it partialy

teohhanhui:
Okay, I understand what you're trying to achieve now, and in my opinion it should not be done. An "input class" should only be used in very limited scenarios, as a form of shortcut (e.g. `PasswordResetRequest`). Usually, the resource class should be used.

norkunas:
> Okay, I understand what you're trying to achieve now, and in my opinion it should not be done. An "input class" should only be used in very limited scenarios, as a form of shortcut (e.g. a `PasswordResetRequest` input class for a `User` resource). Usually, the resource class should be used.

So you suggest to allow entities to be in invalid state :slightly_smiling_face:

teohhanhui:
@norkunas:
> So you suggest to allow entities to be in invalid state slightly_smiling_face

Can you elaborate? The resource class can be separate from the entity class.

norkunas:
> @norkunas:
> 
> > So you suggest to allow entities to be in invalid state slightly_smiling_face
> 
> Can you elaborate? The resource class can be separate from the entity class. The resource class is what defines the "shape" of your API resource. So, if you find that your entity / database representation is not the same as what you want to expose through your REST API, use a separate resource class, **not** input/output classes.

To separate @ApiResource configuration from entities that would be much work. Until input classes was introduced I think most apip users used to allow entities to be in invalid state to work with symfony validation, but the partial updates did work. Later when it was introduced this was a good way to escape from invalid states, but partial updates ability was lost, so in my opinion this is just a missing piece.

cacahouwete:
For me we should never pass by entity directly because setter, ..., should guaranteed the model. For example:
```php
<?php
class Address {
    private float $latitude
    private float $longitude
    public function __construct(float $latitude, float $longitude)
    {
        $this->setLatitude($latitude);
        $this->setLongitude($longitude);
    }

    public static function latitudeIsValid(float $latitude): bool
    {
        return self::MIN_LATITUDE <= $latitude && self::MAX_LATITUDE >= $latitude;
    }

    public static function longitudeIsValid(float $longitude): bool
    {
        return self::MIN_LONGITUDE <= $longitude && self::MAX_LONGITUDE >= $longitude;
    }

    public function getLatitude(): float
    {
        return $this->latitude;
    }

    public function getLongitude(): float
    {
        return $this->longitude;
    }

    /** @throws UnexpectedValueException */
    public function setLatitude(float $latitude): void
    {
        if (!self::latitudeIsValid($latitude)) {
            throw new UnexpectedValueException('Latitude '.$latitude.' is out of bounds.');
        }
        $this->latitude = $latitude;
    }

    /** @throws UnexpectedValueException */
    public function setLongitude(float $longitude): void
    {
        if (!self::longitudeIsValid($longitude)) {
            throw new UnexpectedValueException('Longitude '.$longitude.' is out of bounds.');
        }
        $this->longitude = $longitude;
    }
}
```

If I want instanciate or change any value in this object It will trow an error if it's not valid and it's will no possibility to persist it in database.
If I have an inputDto it's to put some data and validate the integrity. then `class AddressInput` will have some public attributes with some constraint like not blank, greater than, ... . if theses constraints are not valid it will never pass to the next step (put valid value in the model).

teohhanhui:
I wouldn't recommend abusing input/output classes like that, and certainly wouldn't like to see more features that would make it easier for users to use it in such a way. But it's just my personal opinion. :smile:

teohhanhui:
The problem of the resource being in an invalid state could be solved by having 2-step validation (or "pre-validation").

I did a quick and dirty hack here before: https://gist.github.com/teohhanhui/0cde3f7c12b6d0debba44d76aa6db4d9

norkunas:
I understand you, all fine, if it wont make to the core, i've already have
a custom normalizer for this. But one small thing - shape of db and my
resource has no difference, it's just that i can't know if consumer has
included property in payload or not. For sure this can be solved with
__set/__get but that sucks.

2020-02-19, tr 16:21, Teoh Han Hui <notifications@github.com> rašė:

soyuka:
Actually we're trying to understand the need more precisely. It can be that this is an interesting missing feature, and we think that using Inputs isn't the correct way of managing that. If I follow, you want to validate the deserialized Input?

norkunas:
> Actually we're trying to understand the need more precisely. It can be that this is an interesting missing feature, and we think that using Inputs isn't the correct way of managing that. If I follow, you want to validate the deserialized Input?

It's more about partial payload than validation. Simple example:
```
class User {}
/** @ApiResource */
class Employee {
    public $manager; // User or null
    // .. other properties
}
```
When you don't use input class, and send partial payload the `$manager` is left intact.

When you use input class which defines the `$manager` field and if it won't be included in payload then the `$manager` in the input class will be null, so in this case data transformer would replace the Employee manager with null, instead of leaving it as it is.

Data transformer could check if it's not null and then set it, but then it won't be possible to set the manager to null.

norkunas:
I'm not 100% sure if the input prehydration is the way to go, but currently I don't see the better solution.

soyuka:
I think I get it. With your example:

```
class Input {
  public $manager;
  public $name;
}

/** @ApiResource(input=Input::class) */
class Employee {
    public $manager;
    public $name;
}
```

If I send `curl -H "Content-Type: application/json" -d '{"name":"xyz"}' http://localhost/api`

The input will be `Input({manager: null, name: 'xyz`})` and it'd end up with `Employee({manager: null, name: 'xyz'})`.

What you would want is to partially update `Employee` with the non-null fields from `Input` so that:

If I send `curl -H "Content-Type: application/json" -d '{"name":"xyz"}' http://localhost/api`

The input will be `Input({manager: null, name: 'xyz`})` and it'd end up with `Employee({manager: left intact, name: 'xyz})`.

Correct?

norkunas:
That's right.

soyuka:
I don't get why it can't be done just using a simple data transformer though:

```
<?php
namespace App\DataTransformer;

use ApiPlatform\Core\DataTransformer\DataTransformerInterface;

final class InputDataTransformer implements DataTransformerInterface
{
    public function transform($data, string $to, array $context = [])
    {
        $existingEmployee = $context[AbstractItemNormalizer::OBJECT_TO_POPULATE];
        if (null === $data->manager) {
            $existingEmployee->manager = $data->manager;
        }
        return $existingEmployee;
    }

    public function supportsTransformation($data, string $to, array $context = []): bool {}
}
```

Also, API Platform supports `PATCH` which should work like this without any input hassle.

norkunas:
> I don't get why it can't be done just using a simple data transformer though:
> 
> ```
> <?php
> namespace App\DataTransformer;
> 
> use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
> use ApiPlatform\Core\Serializer\AbstractItemNormalizer;
> 
> final class InputDataTransformer implements DataTransformerInterface
> {
>     public function transform($data, string $to, array $context = [])
>     {
>         $existingEmployee = $context[AbstractItemNormalizer::OBJECT_TO_POPULATE];
>         if (null === $data->manager) {
>             $existingEmployee->manager = $data->manager;
>         }
>         return $existingEmployee;
>     }
> 
>     public function supportsTransformation($data, string $to, array $context = []): bool {}
> }
> ```

Because the case `curl -H "Content-Type: application/json" -d '{"manager":null}' http://localhost/api` to unset the manager would not work.

cacahouwete:
For me there are a difference to say 
`curl -H "Content-Type: application/json" -d '{"manager":null, "name": "new value"}'`
and
`curl -H "Content-Type: application/json" -d '{"name": "new value"}'`

The first one, we explicitly say that manager have to be null, we want set this attribute as null.
The second exemple , we say we want change name , but manager stay with his current value.

teohhanhui:
The input/output class feature is supposed to enable shortcuts where you need a different representation than the resource. If you're trying to separate resource and entity, do **not** use input/output classes for that. If you're trying to prevent the resource from getting into an invalid state, do **not** use input/output classes for that.

PR #3381, opened 2020-02-06
IsRequired should return true even if not writable

quentinus95:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #3389
| License       | MIT
| Doc PR        | 

While I understand what is the logic under the code I'm removing, I believe this check is wrong because it assumes the property belongs to a schema describing a model for a write operation.

In that case, we *could* say the field is not required because it is not writable.

However, when describing a model for a read operation (e.g. using a group in the `normalizationContext`), we *do* want read-only fields to be marked as required. Otherwise tooling such as code generators will assume the field is nullable.

This is especially happening for all identifiers when auto-generated in the backend. They are always read-only but they are also always required (the client should not assume that the id field may be null).

### DIFF
--- a/tests/Metadata/Property/PropertyMetadataTest.php
+++ b/tests/Metadata/Property/PropertyMetadataTest.php
@@ -84,14 +84,14 @@ class PropertyMetadataTest extends TestCase
         $this->assertTrue($newMetadata->isInitializable());
     }
 
-    public function testShouldReturnRequiredFalseWhenRequiredTrueIsSetButMaskedByWritableFalse()
+    public function testShouldReturnRequiredTrueEvenIfWritableFalseSoClientCanAssumeValueCannotBeUndefinedOnReadOperation()
     {
         $metadata = new PropertyMetadata();
 
         $metadata = $metadata->withRequired(true);
         $metadata = $metadata->withWritable(false);
 
-        $this->assertFalse($metadata->isRequired());
+        $this->assertTrue($metadata->isRequired());
     }
 
     public function testShouldReturnPreviouslySetRequiredTrueWhenWritableFalseUnmasked()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.5.5
 
 * GraphQL: Do not allow empty cursor values on `before` or `after`
+* OpenAPI: Fix required fields by keeping them required even if they are not writable to allow valid read types generation on client side (#3381)
 
 ## 2.5.4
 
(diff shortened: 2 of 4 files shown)


quentinus95:
Also I'm facing the same issue when I have a virtual property using a getter:

```
    /*     
     * @ApiProperty(
     *     required=true
     * )
     */
    public function getSomething()
    {
    }
```

This won't set the property to required due to the 4 lines of code I'm removing.

tests/Metadata/Property/PropertyMetadataTest.php
```diff
@@ -84,14 +84,14 @@ public function testValueObject()
         $this->assertTrue($newMetadata->isInitializable());
     }
 
-    public function testShouldReturnRequiredFalseWhenRequiredTrueIsSetButMaskedByWritableFalse()
+    public function testShouldReturnRequiredTrueEvenIfWritableFalseSoClientCanAssumeValueCannotBeUndefinedOnReadOperation()
```
quentinus95: maybe we should remove / rephrase this test

quentinus95:
Failing tests are unrelated.

soyuka:
I'm unsure about this, what you say makes sense, wdyt @api-platform/core-team ?

quentinus95:
@soyuka is there anything I can do (more explanations, implementation change, ...) to get some feedback and know if this branch can/will be merged in the future?

Thanks!

alanpoulain:
Isn't the issue the assumption of nullability if the property is not required?
This PR is also related: https://github.com/api-platform/core/pull/3518
We only have the specification of JSON Schema to understand it (https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-6.5.3):
> An object instance is valid against this keyword if every item in the array is the name of a property in the instance.

It only says that the property may not appear, not that it is nullable.

But I think you are right to say that if a property is always normalized, then it should probably be required.

mvlad:
This bug also leads to the wrong documentation when constructor property promotion is used.

For example:
```php
<?php

namespace App\Customer\ApiResource;

use Symfony\Component\Validator\Constraints\NotBlank;

#[ApiResource(
    collectionOperations: ['post'],
    itemOperations: [],
)]
class Book
{
    public function __construct(
        #[NotBlank]
        #[ApiProperty(identifier: true)]
        protected string $name,
    ) {
    }

    {
        return $this->name;
    }
}
```

Here the `name` property is required and ApiPlatform can set it. But in the documentation, this property is not marked as "required" because it's not writable.

roman-eonx:
It's been a while since this PR was created but the issue is still actual. Especially with properties set from constructors in our use case. @soyuka what is required to continue working on this issue? Please let us know if you want a fresh PR for the current version to be created.

PR #3396, opened 2020-02-18
Improve subresources DB query performance by removing a subquery

clemherreman:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (kinda) 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | none
| License       | MIT
| Doc PR        | N/A

## Abstract 

This PR try to change DQL queries generated to fetch subresource, to improve performance on large tables.

## Before

The generated SQL to query a subresource uses a `WHERE ... IN ( <subquery> )`, in order to be recursive when subresources also have subresources (introduced by #1608).

The subquery, on large table (at work 4M+ lines), takes a significant amount of time (2-5s), even with the proper SQL indexes (at least on MySQL).

```SQL
SELECT
    DISTINCT c0_.idpurchase AS idpurchase_0,
    c0_.idpurchase AS idpurchase_1
FROM purchase c0_
LEFT JOIN customer c1_ ON c0_.idcustomer = c1_.idcustomer
-- Some other LEFT JOINs
WHERE c0_.idcustomer IN 
(  -- here please notice the usage of a subquery
    SELECT c16_.idcustomer
    FROM customer c16_
    WHERE c16_.idcustomer = :identifier
)
```

## After

This PRs tries to optimize for the common case, having only one level of subresources, by replacing the subquery via a direct `WHERE <identifier> = <identifier-alias>` **only on the last recursion level**. This means it shouldn't break subresources that have subresources, while still improving the query for the nominal cases.

```SQL
SELECT
    DISTINCT c0_.idpurchase AS idpurchase_0,
    c0_.idpurchase AS idpurchase_1
FROM purchase c0_
LEFT JOIN customer c1_ ON c0_.idcustomer = c1_.idcustomer
-- Some other LEFT JOINs
WHERE c0_.idcustomer = :identifier

```

**This significantly reduces the SQL query time, going from 2.45s to 15ms on my e-commerce app.**

## BC & missing things.

I broke the unit tests, mainly the mocking, but also some tests fails as they compare the previously generated query, with the new one (which makes sense). All the Behat tests are green locally though.

I feel out of my depth here, and if someone would help me writing those, ensuring no BC is done, is would be great.

On my side, I run our rather large behat test suite against my fork, and not tests complained about wrong results, so I am fairly confident I haven't introduced some BC.

# TODO

- [x] Fix Behat tests
- [x] Fix PHPUnit tests

### DIFF
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -206,7 +206,7 @@ final class SubresourceDataProvider implements SubresourceDataProviderInterface
             $topQueryBuilder->setParameter($placeholder, $value, (string) $classMetadata->getTypeOfField($key));
 
             // Optimization: add where clause for identifiers, but not via a WHERE ... IN ( ...subquery... ).
-            // Instead we use a direct identifier equality clause, to speed thing up when dealing with large tables.
+            // Instead we use a direct identifier equality clause, to speed things up when dealing with large tables.
             // We may do so if there is no more recursion levels from here, and if relation allows it.
             $association = $classMetadata->hasAssociation($previousAssociationProperty) ? $classMetadata->getAssociationMapping($previousAssociationProperty) : [];
             $oneToOneBidirectional = isset($association['inversedBy']) && ClassMetadataInfo::ONE_TO_ONE === $association['type'];
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -205,7 +205,7 @@ final class SubresourceDataProvider implements SubresourceDataProviderInterface
             $placeholder = $queryNameGenerator->generateParameterName($key);
             $topQueryBuilder->setParameter($placeholder, $value, (string) $classMetadata->getTypeOfField($key));
 
-            // Optimizaion: add where clause for identifiers, but not via a WHERE ... IN ( ...subquery... ).
+            // Optimization: add where clause for identifiers, but not via a WHERE ... IN ( ...subquery... ).
             // Instead we use a direct identifier equality clause, to speed thing up when dealing with large tables.
             // We may do so if there is no more recursion levels from here, and if relation allows it.
             $association = $classMetadata->hasAssociation($previousAssociationProperty) ? $classMetadata->getAssociationMapping($previousAssociationProperty) : [];
--- a/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
+++ b/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
@@ -198,41 +198,34 @@ final class SubresourceDataProvider implements SubresourceDataProviderInterface
                 ->from($identifierResourceClass, $alias);
         }
 
-        $lastIdentifer = 1 === $remainingIdentifiers;
-        $association = $classMetadata->hasAssociation($previousAssociationProperty) ? $classMetadata->getAssociationMapping($previousAssociationProperty) : [];
-        $optimizable = $lastIdentifer && (
-                (isset($association['inversedBy']) && ClassMetadataInfo::ONE_TO_ONE === $association['type'])
-                || (isset($association['mappedBy']) && ClassMetadataInfo::ONE_TO_MANY === $association['type'])
-            );
+        $isLeaf = 1 === $remainingIdentifiers;
 
         // Add where clause for identifiers
         foreach ($normalizedIdentifiers as $key => $value) {
             $placeholder = $queryNameGenerator->generateParameterName($key);
-            if ($optimizable) {
-                // Add where clause for identifiers, but not via a WHERE ... IN ( ...subquery... ). Instead we use
-                // a direct identifier equality clause, to speed thing up when dealing with large tables.
-                // We may do so as there is no more recursion levels from here.
-
-                if (ClassMetadataInfo::ONE_TO_ONE === $association['type']) {
-                    $joinAlias = $queryNameGenerator->generateJoinAlias($association['inversedBy']);
-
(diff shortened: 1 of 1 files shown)


src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -142,7 +142,9 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
         $classMetadata = $manager->getClassMetadata($identifierResourceClass);
 
         if (!$classMetadata instanceof ClassMetadataInfo) {
-            throw new RuntimeException("The class metadata for $identifierResourceClass must be an instance of ClassMetadataInfo.");
+            throw new RuntimeException(
+                "The class metadata for $identifierResourceClass must be an instance of ClassMetadataInfo."
```
dkarlovi: Code style change? I'd revert it.

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -198,16 +200,42 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
                 ->from($identifierResourceClass, $alias);
         }
 
+        $lastIdentifer = $remainingIdentifiers === 1;
+        $association = $classMetadata->hasAssociation($previousAssociationProperty) ? $classMetadata->getAssociationMapping($previousAssociationProperty) : [];
+        $optimizable = $lastIdentifer && (
```
dkarlovi: Better explain *why* it's optimizable and make that the variable name here.

soyuka:
> All the Behat tests are green locally though.

That's good, I've covered the edge cases of this in the behat test suite. The queries you show give the same results, therefore no BC break will outcome of this change. You can change the broken phpunit tests by fixing the relevant queries.

teohhanhui:
Thanks for doing this! :tada:

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
clemherreman: Reverted. I guess it sneaked out because PHPStorm was too enthusiastic.

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
clemherreman: I added a long comment about the optimization, when using `$optimizable`. I could not find a variable name that was both concise and explanatory though.

Is that ok with you?

clemherreman:
Fixed the PHPUnit tests, and rebased against latest `2.5`.

If the comment I added after @dkarlovi feedback is enough, I'd say this is ready to be merged.

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
teohhanhui: `$needsSubquery`? (So it'd have the opposite value.)

teohhanhui: Or `$isLeaf` to make it merely descriptive and decoupled from the implementation.

clemherreman: I reworked the way the code is written, to have more readable code branches and variable name. WDYT?

teohhanhui: `$eligibleOneToOneOptimization` -> `$oneToOneBidirectional`
`$eligibleOneToManyOptimization` -> `$oneToManyBidirectional`

Same idea... The names can be merely descriptive.

clemherreman: Done :)

teohhanhui:
> Behat tests are still failing, but because of either ccode coverage generation failing (?) or call to Dummy::hasRole() failing in the serialization. Both errors seems unrelated to that pull request, isn't it?

Yeah, those are unrelated. Please ignore those failures. :smile:

clemherreman:
@teohhanhui Behat tests already covers the correctness of the query results (thanks to @alanpoulain)

Is there a point in checking the generated DQL itself? I doubt so. This is a perf improvement on the query level.

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -198,11 +198,30 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
...
 
+        $isLeaf = 1 === $remainingIdentifiers;
+
         // Add where clause for identifiers
         foreach ($normalizedIdentifiers as $key => $value) {
             $placeholder = $queryNameGenerator->generateParameterName($key);
-            $qb->andWhere("$alias.$key = :$placeholder");
             $topQueryBuilder->setParameter($placeholder, $value, (string) $classMetadata->getTypeOfField($key));
+
+            // Optimizaion: add where clause for identifiers, but not via a WHERE ... IN ( ...subquery... ).
```
alanpoulain:
```suggestion
            // Optimization: add where clause for identifiers, but not via a WHERE ... IN ( ...subquery... ).
```

src/Bridge/Doctrine/Orm/SubresourceDataProvider.php
```diff
@@ -198,11 +198,30 @@ private function buildQuery(array $identifiers, array $context, QueryNameGenerat
...
+            // Instead we use a direct identifier equality clause, to speed thing up when dealing with large tables.
```
teohhanhui:
```suggestion
            // Instead we use a direct identifier equality clause, to speed things up when dealing with large tables.
```

teohhanhui:
Thanks @clemherreman! :tada:

mxmp210:
Please have a look at [#1542](https://github.com/api-platform/api-platform/issues/1542), As it breaks custom identifiers, feel free to review.

PR #3397, opened 2020-02-18
Revert "fix getClassType with no readableLink"

teohhanhui:
Reverts api-platform/core#3362

It was an invalid fix.


soyuka:
:+1: Can we keep the unit test and try to fix it properly?

teohhanhui:
Let's do that in another PR. The original issue was very lacking in details anyway.

PR #3402, opened 2020-02-19
Corrected schema generation for `array<string, T>`, `object`, `?array`, `?object` and `?type`

Ocramius:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no - please check `Type::BUILTIN_TYPE_OBJECT` changes though
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

The previous version of the `TypeFactory` generated following **WRONG** definitions:

 * `null|T[]` as `{"type": array, "items": {"type": "T"}}`
 * `?T[]` as `{"type": array, "items": {"type": "T"}}`
 * `array<string, T>` as `{"type": array, "items": {"type": "T"}}`
 * `object` without explicit schema definition as `{"type": "string"}`
 * `?T` as `{"type": T}`

The new definitions instead do fix this by mapping:

 * `array<string, T>` as `{"type": "object", "additionalProperties": {"type": "T"}}`
 * `array<string, ?T>` as `{"type": object, "additionalProperties": {"type": ["T", "null"]}}`
 * `null|array<string, T>` as `{"type": ["object", "null"], "additionalProperties": {"type": "T"}}`
 * `array<int, T>` as `{"type": "array", "items": {"type": "T"}}` (not very precise, but list support is not yet in symfony)
 * `object` without explicit schema definition as `{"type": "object"}`
 * `?T[]` as `{"type": "array", "items": {"type": ["T", "null"]}}`
 * `null|T[]` as `{"type": ["array", "null"], "items": {"type": "T"}}`

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -258,19 +258,19 @@ final class SchemaFactory implements SchemaFactoryInterface
 
         return [
             $resourceMetadata,
-            $this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName, $serializerContext),
+            $serializerContext ?? $this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName),
             $inputOrOutput['class'],
         ];
     }
 
-    private function getSerializerContext(ResourceMetadata $resourceMetadata, string $type = Schema::TYPE_OUTPUT, ?string $operationType, ?string $operationName, ?array $previousSerializerContext): array
+    private function getSerializerContext(ResourceMetadata $resourceMetadata, string $type = Schema::TYPE_OUTPUT, ?string $operationType, ?string $operationName): array
     {
         $attribute = Schema::TYPE_OUTPUT === $type ? 'normalization_context' : 'denormalization_context';
 
         if (null === $operationType || null === $operationName) {
-            return array_merge($resourceMetadata->getAttribute($attribute, []), (array) $previousSerializerContext);
+            return $resourceMetadata->getAttribute($attribute, []);
         }
 
-        return array_merge($resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, [], true), (array) $previousSerializerContext);
+        return $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, [], true);
     }
 }
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -258,19 +258,19 @@ final class SchemaFactory implements SchemaFactoryInterface
 
         return [
             $resourceMetadata,
-            $serializerContext ?? $this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName),
+            $this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName, $serializerContext),
             $inputOrOutput['class'],
         ];
     }
 
-    private function getSerializerContext(ResourceMetadata $resourceMetadata, string $type = Schema::TYPE_OUTPUT, ?string $operationType, ?string $operationName): array
+    private function getSerializerContext(ResourceMetadata $resourceMetadata, string $type = Schema::TYPE_OUTPUT, ?string $operationType, ?string $operationName, ?array $previousSerializerContext): array
     {
         $attribute = Schema::TYPE_OUTPUT === $type ? 'normalization_context' : 'denormalization_context';
 
         if (null === $operationType || null === $operationName) {
-            return $resourceMetadata->getAttribute($attribute, []);
+            return array_merge($resourceMetadata->getAttribute($attribute, []), (array) $previousSerializerContext);
         }
 
-        return $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, [], true);
+        return array_merge($resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, [], true), (array) $previousSerializerContext);
     }
 }
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -156,10 +156,6 @@ final class TypeFactory implements TypeFactoryInterface
      * @param array<string, mixed> $jsonSchema
      *
      * @return array<string, mixed>
-     *
-     * @psalm-param array{type=: string|list<string>} $jsonSchema
-     *
-     * @psalm-return array{type=: string|list<string>, $ref=: string}
      */
     private function addNullabilityToTypeDefinition(array $jsonSchema, Type $type): array
     {
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -64,8 +64,7 @@ final class TypeFactory implements TypeFactoryInterface
     {
         if ($type->isCollection()) {
             $keyType = $type->getCollectionKeyType();
-            $subType = $type->getCollectionValueType()
-                ?? new Type($type->getBuiltinType(), false, $type->getClassName(), false);
+            $subType = $type->getCollectionValueType() ?? new Type($type->getBuiltinType(), false, $type->getClassName(), false);
 
             if (null !== $keyType && Type::BUILTIN_TYPE_STRING === $keyType->getBuiltinType()) {
                 return $this->addNullabilityToTypeDefinition(
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -17,6 +17,9 @@ use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Ramsey\Uuid\UuidInterface;
 use Symfony\Component\PropertyInfo\Type;
+use function array_merge;
+use function array_unique;
+use function array_values;
 
 /**
  * {@inheritdoc}
@@ -50,14 +53,37 @@ final class TypeFactory implements TypeFactoryInterface
     public function getType(Type $type, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, Schema $schema = null): array
     {
         if ($type->isCollection()) {
-            $subType = new Type($type->getBuiltinType(), $type->isNullable(), $type->getClassName(), false);
+            $keyType = $type->getCollectionKeyType();
+            $subType = $type->getCollectionValueType()
+                ?? new Type($type->getBuiltinType(), false, $type->getClassName(), false);
 
-            return [
-                'type' => 'array',
-                'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
-            ];
+            if (null !== $keyType && Type::BUILTIN_TYPE_STRING === $keyType->getBuiltinType()) {
+                return $this->addNullabilityToTypeDefinition(
+                    [
+                        'type' => 'object',
+                        'additionalProperties' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
+                    ],
+                    $type
+                );
+            }
+
+            return $this->addNullabilityToTypeDefinition(
+                [
+                    'type' => 'array',
+                    'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
+                ],
+                $type
+            );
         }
 
+        return $this->addNullabilityToTypeDefinition(
+            $this->makeBasicType($type, $format, $readableLink, $serializerContext, $schema),
+            $type
+        );
+    }
+
+    private function makeBasicType(Type $type, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, Schema $schema = null): array
+    {
         switch ($type->getBuiltinType()) {
             case Type::BUILTIN_TYPE_INT:
                 return ['type' => 'integer'];
@@ -78,7 +104,7 @@ final class TypeFactory implements TypeFactoryInterface
     private function getClassType(?string $className, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, ?Schema $schema = null): array
     {
         if (null === $className) {
-            return ['type' => 'string'];
+            return ['type' => 'object'];
         }
 
         if (is_a($className, \DateTimeInterface::class, true)) {
@@ -102,7 +128,7 @@ final class TypeFactory implements TypeFactoryInterface
 
         // Skip if $schema is null (filters only support basic types)
         if (null === $schema) {
-            return ['type' => 'string'];
+            return ['type' => 'object'];
         }
 
         if ($this->isResourceClass($className) && true !== $readableLink) {
@@ -124,5 +150,44 @@ final class TypeFactory implements TypeFactoryInterface
         $subSchema = $this->schemaFactory->buildSchema($className, $format, Schema::TYPE_OUTPUT, null, null, $subSchema, $serializerContext);
 
         return ['$ref' => $subSchema['$ref']];
+    }
+
+    /**
+     * @param array<string, mixed> $jsonSchema
+     *
+     * @return array<string, mixed>
+     *
+     * @psalm-param array{type=: string|list<string>} $jsonSchema
+     *
+     * @psalm-return array{type=: string|list<string>, $ref=: string}
+     */
+    private function addNullabilityToTypeDefinition(array $jsonSchema, Type $type): array
+    {
+        if (!$type->isNullable()) {
+            return $jsonSchema;
+        }
+
+        if (!\array_key_exists('type', $jsonSchema)) {
+            return [
+                'oneOf' => [
+                    ['type' => 'null'],
+                    $jsonSchema,
+                ],
+            ];
+        }
+
+        return array_merge($jsonSchema, ['type' => $this->addNullToTypes((array) $jsonSchema['type'])]);
+    }
+
+    /**
+     * @param string[] $types
+     *
+     * @return string[]
+     *
+     * @psalm-param list<string> $types
+     */
+    private function addNullToTypes(array $types): array
+    {
+        return array_values(array_unique(array_merge($types, ['null'])));
     }
 }
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -180,11 +180,13 @@ final class TypeFactory implements TypeFactoryInterface
             return $jsonSchema;
         }
 
-        return [
-            'oneOf' => [
-                ['type' => 'null'],
-                $jsonSchema,
-            ],
-        ];
+        if (\array_key_exists('$ref', $jsonSchema)) {
+            return [
+                'nullable' => true,
+                'anyOf' => [$jsonSchema],
+            ];
+        }
+
+        return array_merge($jsonSchema, ['nullable' => true]);
     }
 }
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -30,19 +30,6 @@ final class TypeFactory implements TypeFactoryInterface
     use ResourceClassInfoTrait;
 
     /**
-     * This constant is to be provided as serializer context key to conditionally enable types to be generated in
-     * a format that is compatible with OpenAPI specifications **PREVIOUS** to 3.0.
-     *
-     * Without this flag being set, the generated format will only be compatible with Swagger 3.0 or newer.
-     *
-     * Once support for OpenAPI < 3.0 is gone, this constant **WILL BE REMOVED**
-     *
-     * @internal Once support for OpenAPI < 3.0 is gone, this constant **WILL BE REMOVED** - do not rely on
-     *           it in downstream projects!
-     */
-    public const CONTEXT_SERIALIZATION_FORMAT_OPENAPI_PRE_V3_0 = self::class.'::CONTEXT_SERIALIZATION_FORMAT_OPENAPI_PRE_V3_0';
-
-    /**
      * @var SchemaFactoryInterface|null
      */
     private $schemaFactory;
@@ -67,31 +54,19 @@ final class TypeFactory implements TypeFactoryInterface
             $subType = $type->getCollectionValueType() ?? new Type($type->getBuiltinType(), false, $type->getClassName(), false);
 
             if (null !== $keyType && Type::BUILTIN_TYPE_STRING === $keyType->getBuiltinType()) {
-                return $this->addNullabilityToTypeDefinition(
-                    [
-                        'type' => 'object',
-                        'additionalProperties' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
-                    ],
-                    $type,
-                    (array) $serializerContext
-                );
+                return $this->addNullabilityToTypeDefinition([
+                    'type' => 'object',
+                    'additionalProperties' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
+                ], $type, $schema);
             }
 
-            return $this->addNullabilityToTypeDefinition(
-                [
-                    'type' => 'array',
-                    'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
-                ],
-                $type,
-                (array) $serializerContext
-            );
+            return $this->addNullabilityToTypeDefinition([
+                'type' => 'array',
+                'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
+            ], $type, $schema);
         }
 
-        return $this->addNullabilityToTypeDefinition(
-            $this->makeBasicType($type, $format, $readableLink, $serializerContext, $schema),
-            $type,
-            (array) $serializerContext
-        );
+        return $this->addNullabilityToTypeDefinition($this->makeBasicType($type, $format, $readableLink, $serializerContext, $schema), $type, $schema);
     }
 
     private function makeBasicType(Type $type, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, Schema $schema = null): array
@@ -169,9 +144,9 @@ final class TypeFactory implements TypeFactoryInterface
      *
      * @return array<string, mixed>
      */
-    private function addNullabilityToTypeDefinition(array $jsonSchema, Type $type, array $serializerContext): array
+    private function addNullabilityToTypeDefinition(array $jsonSchema, Type $type, ?Schema $schema): array
     {
-        if (\array_key_exists(self::CONTEXT_SERIALIZATION_FORMAT_OPENAPI_PRE_V3_0, $serializerContext)) {
+        if ($schema && Schema::VERSION_SWAGGER === $schema->getVersion()) {
             return $jsonSchema;
         }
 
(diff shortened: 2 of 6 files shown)


tests/JsonSchema/TypeFactoryTest.php
```diff
@@ -35,13 +35,89 @@ public function testGetType(array $schema, Type $type): void
...
+        yield [['type' => ['boolean', 'null']], new Type(Type::BUILTIN_TYPE_BOOL, true)];
+        yield [['type' => 'string'], new Type(Type::BUILTIN_TYPE_STRING)];
+        yield [['type' => ['string', 'null']], new Type(Type::BUILTIN_TYPE_STRING, true)];
+        yield [['type' => 'object'], new Type(Type::BUILTIN_TYPE_OBJECT)];
+        yield [['type' => ['object', 'null']], new Type(Type::BUILTIN_TYPE_OBJECT, true)];
         yield [['type' => 'string', 'format' => 'date-time'], new Type(Type::BUILTIN_TYPE_OBJECT, false, \DateTimeImmutable::class)];
+        yield [['type' => ['string', 'null'], 'format' => 'date-time'], new Type(Type::BUILTIN_TYPE_OBJECT, true, \DateTimeImmutable::class)];
         yield [['type' => 'string', 'format' => 'duration'], new Type(Type::BUILTIN_TYPE_OBJECT, false, \DateInterval::class)];
-        yield [['type' => 'string'], new Type(Type::BUILTIN_TYPE_OBJECT, false, Dummy::class)];
+        yield [['type' => 'object'], new Type(Type::BUILTIN_TYPE_OBJECT, false, Dummy::class)];
```
Ocramius: Please mind this change: I don't understand why `object` was being mapped as `string` previously, but I suppose the idea was to map UUID-alike single-field value types? That is indeed wrong, unless I'm missing a basic use-case.

tests/JsonSchema/TypeFactoryTest.php
```diff
@@ -35,13 +35,89 @@ public function testGetType(array $schema, Type $type): void
...
         yield [['type' => 'integer'], new Type(Type::BUILTIN_TYPE_INT)];
+        yield [['type' => ['integer', 'null']], new Type(Type::BUILTIN_TYPE_INT, true)];
         yield [['type' => 'number'], new Type(Type::BUILTIN_TYPE_FLOAT)];
+        yield [['type' => ['number', 'null']], new Type(Type::BUILTIN_TYPE_FLOAT, true)];
         yield [['type' => 'boolean'], new Type(Type::BUILTIN_TYPE_BOOL)];
-        yield [['type' => 'string'], new Type(Type::BUILTIN_TYPE_OBJECT)];
+        yield [['type' => ['boolean', 'null']], new Type(Type::BUILTIN_TYPE_BOOL, true)];
+        yield [['type' => 'string'], new Type(Type::BUILTIN_TYPE_STRING)];
+        yield [['type' => ['string', 'null']], new Type(Type::BUILTIN_TYPE_STRING, true)];
+        yield [['type' => 'object'], new Type(Type::BUILTIN_TYPE_OBJECT)];
```
Ocramius: Please mind this change: I don't understand why `object` was being mapped as `string` previously, but I suppose the idea was to map UUID-alike single-field value types? That is indeed wrong, unless I'm missing a basic use-case.

src/JsonSchema/TypeFactory.php
```diff
@@ -17,6 +17,12 @@
 use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Ramsey\Uuid\UuidInterface;
 use Symfony\Component\PropertyInfo\Type;
+use function array_key_exists;
+use function array_merge;
+use function array_unique;
+use function array_values;
+use function is_array;
+use function is_string;
```
alanpoulain: Why?

src/JsonSchema/TypeFactory.php
```diff
@@ -50,14 +56,37 @@ public function setSchemaFactory(SchemaFactoryInterface $schemaFactory): void
...
     {
         if ($type->isCollection()) {
-            $subType = new Type($type->getBuiltinType(), $type->isNullable(), $type->getClassName(), false);
-
-            return [
-                'type' => 'array',
-                'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
-            ];
+            $keyType = $type->getCollectionKeyType();
+            $subType = $type->getCollectionValueType()
```
alanpoulain: I think we prefer to have it in one line.

src/JsonSchema/TypeFactory.php
Ocramius: Mostly something I do by default: not sure if there's a coding standard for this project, didn't check build failures on that front yet.

Overall, relying on global function fallback is more trouble.

src/JsonSchema/TypeFactory.php
```diff
@@ -125,4 +154,43 @@ private function getClassType(?string $className, string $format = 'json', ?bool
...
+     * @psalm-return array{type=: string|list<string>, $ref=: string}
+     */
+    private function addNullabilityToTypeDefinition(array $jsonSchema, Type $type) : array
+    {
+        if (! $type->isNullable()) {
+            return $jsonSchema;
+        }
+
+        if (! array_key_exists('type', $jsonSchema)) {
+            return [
```
zerkms: How would this returned literal type check the `@psalm-return array{type=: string|list<string>, $ref=: string}` if it does not have a `type` and `$ref` keys?

src/JsonSchema/TypeFactory.php
Ocramius: I work on 3440x1440, and this was already quite too wide :D 

I don't mind reverting, but please focus review on non-cs issues first.

src/JsonSchema/TypeFactory.php
Ocramius: Good point: will inspect if any other more precise type is possible, otherwise remove.

Ocramius: 899f1e3ce1c329db01c2971548b0538a44bb5ce3

Ocramius:
Hmm, can't run the failing test cases locally:

```
8) ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ApiTestCaseTest::testFindIriBy
ReflectionException: Class "MongoDB\Driver\Monitoring\CommandSubscriber" not found while loading "Doctrine\Bundle\MongoDBBundle\APM\PSRCommandLogger".

/home/ocramius/Documents/api-platform/core/vendor/symfony/config/Resource/ClassExistenceResource.php:74
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/ContainerBuilder.php:353
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/PriorityTaggedServiceTrait.php:64
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/ResolveTaggedIteratorArgumentPass.php:34
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/AbstractRecursivePass.php:82
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/ResolveTaggedIteratorArgumentPass.php:31
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/AbstractRecursivePass.php:91
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/ResolveTaggedIteratorArgumentPass.php:31
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/AbstractRecursivePass.php:82
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/ResolveTaggedIteratorArgumentPass.php:31
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/AbstractRecursivePass.php:46
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/Compiler/Compiler.php:94
/home/ocramius/Documents/api-platform/core/vendor/symfony/dependency-injection/ContainerBuilder.php:762
/home/ocramius/Documents/api-platform/core/vendor/symfony/http-kernel/Kernel.php:626
/home/ocramius/Documents/api-platform/core/vendor/symfony/http-kernel/Kernel.php:136
/home/ocramius/Documents/api-platform/core/vendor/symfony/framework-bundle/Test/KernelTestCase.php:80
/home/ocramius/Documents/api-platform/core/tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php:163
```

Ocramius:
Similarly to the above, cannot really dump the generated OpenAPI schema to check what validation doesn't work (fails schema validation at https://travis-ci.org/api-platform/core/jobs/652720111)

Ocramius:
Did some local testing against a real world project: seems to work correctly (thus far), but the type definitions aren't necessarily very elaborate (especially `$ref` nullability missing).

If anyone can give me a copy of the generated schema in the failing builds, lemme know: will fix up if it's broken by this patch, or will fix the patch if the generated schema is indeed pants-over-head broken :+1:

Ocramius:
Ok, seems like a misconception (coming from swagger-world). The official docs state: https://swagger.io/docs/specification/data-models/data-types/

 > type takes a single value. type as a list is not valid in OpenAPI (even though it is valid in JSON Schema): 

I'll have to fix that :-P

Ocramius:
Right, seems like the generated schema is now valid for OpenAPI, but Swagger will obviously be "special", since `oneOf` doesn't seem to properly work.

OpenAPI 2.0 doesn't seem to have the `'null'` type at all: what to do?

@dunglas can we chop the head off 2.x? Would need to re-target the patch.

tests/JsonSchema/TypeFactoryTest.php
dunglas: Not it is a limitation: until very recently, API Platform where only able to deal with object mapped with the @ApiResource annotation. And because it’s an hypermedia framework, when such object are encountered, they are replaced by their IRI during the serialization. (An IRI is just a string with a specific format in JSON Schema).

Ocramius: So if I pass in an `stdClass` it would be replaced by a dummy string at runtime also now?

dunglas:
Unfortunately many tools in the OpenAPI ecosystem aren’t supporting v3 yet. We plan to remove support for v2 in API Platform (and we hope most tools will support OpenAPI v3 at this point).

Ocramius:
> Unfortunately many tools in the OpenAPI ecosystem aren’t supporting v3 yet. We plan to remove support for v2 in API Platform (and we hope most tools will support OpenAPI v3 at this point).

Supposedly, we'll need different service configurations for when the `v2` and `v3` schemas are replaced then, to allow `TypeFactory` to be swapped out?

We'd then inject both services into the command, and decide which to run depending on the flag: would that work as an approach? Problem with it is that I don't have a stack where to test that properly.

tests/JsonSchema/TypeFactoryTest.php
dunglas: No, it should work now (since the last minor version), this code path has probably been forgotten. To be tested to be sure :)

Ocramius: So is the diff correct, or does it need adjustments here?

Mostly trying to avoid an unintentional BC break, if this is supposed to stay as `"string"` :+1:

dunglas:
Yes that sounds good to me. Regarding the stack, I’ll check tomorrow morning (sorry I only have my phone with me right now).

Ocramius:
Looking at `ApiPlatform\Core\JsonSchema\TypeFactory::getType()`, without introducing a BC break or too much complexity right now, I could add an `'openapi-v3.0'` key (optional) that switches the entire behavior on/off conditionally.

Ugly as heck, but then it would be easy to switch over to the new behavior once v2 is gone.

This API is only used in 2 locations, so the addition shouldn't be too problematic.

What is unclear to me is why the validator still fails in CI (without `2.0` flags)

tests/JsonSchema/TypeFactoryTest.php
soyuka: It's because an object is normalized as a string identifier ie `/users/1` (I guess?).

src/JsonSchema/TypeFactory.php
teohhanhui: There's word wrapping. Just saying...

src/JsonSchema/TypeFactory.php
```diff
@@ -50,14 +63,40 @@ public function setSchemaFactory(SchemaFactoryInterface $schemaFactory): void
...
-                'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
-            ];
+            $keyType = $type->getCollectionKeyType();
+            $subType = $type->getCollectionValueType()
+                ?? new Type($type->getBuiltinType(), false, $type->getClassName(), false);
+
+            if (null !== $keyType && Type::BUILTIN_TYPE_STRING === $keyType->getBuiltinType()) {
+                return $this->addNullabilityToTypeDefinition(
+                    [
+                        'type' => 'object',
```
teohhanhui: Can't it be an actual collection of strings, instead of IRIs? (If I understand the code correctly.)

teohhanhui: Don't we need to check the value type too?

teohhanhui:
> ?T[]

What does this mean? Is this syntax defined somewhere?

Ocramius:
> > ?T[]
 > 
 > What does this mean? Is this syntax defined somewhere?

Mostly psalm/phpstan. `T` stands for a type variable. This is generally unbeknownst to most of the PHP community, but very common in contexts where better type systems exist (Scala, Java, Kotlin, Haskell, C++, etc.)

src/JsonSchema/TypeFactory.php
Ocramius: If it has string keys, it is not an `array` in JSON-Schema. `array` and `map` are very clear concepts in other languages: not in PHP (we do everything in one shot, and that's the cause of a lot of confusion)

Ocramius:
```

004 Scenario: Get collection by dummyBoolean false             # features/doctrine/boolean_filter.feature:129
      When I send a "GET" request to "/dummies?dummyBoolean=0" # features/doctrine/boolean_filter.feature:130
        Type error: Too few arguments to function ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy::hasRole(), 0 passed in /tmp/api-platform/core/vendor/symfony/property-access/PropertyAccessor.php on line 382 and exactly 1 expected (Behat\Testwork\Call\Exception\FatalThrowableError)

```

No clue what's going on here: is it already broken in `2.5`? Seems like unrelated to schema definitions, unless schema definitions are also re-used internally (at runtime)

src/JsonSchema/TypeFactory.php
```diff
@@ -102,7 +141,7 @@ private function getClassType(?string $className, string $format = 'json', ?bool
 
         // Skip if $schema is null (filters only support basic types)
         if (null === $schema) {
-            return ['type' => 'string'];
+            return ['type' => 'object'];
```
dunglas: I fear it will break our support for filters (query parameters): we use JSON Schema to describe the expected format of parameters, and in an URL context it must be string by default.

tests/JsonSchema/TypeFactoryTest.php
dunglas: It looks good to me

dunglas:
@Ocramius the failure you reported is probably because of a new bug in Symfony, @teohhanhui is investigating it.

Ocramius:
More misunderstanding of the spec: `["type" => "null"]` is not valid OpenAPI v3. Instead, `"nullable": true` is to be used, but special handling is needed for `$ref` composition

src/JsonSchema/TypeFactory.php
teohhanhui: Uhh, yes. That flew way past my head. :laughing:

teohhanhui:
> @teohhanhui is investigating it

I better start investigating...

Ocramius:
Note: `{"type": ["integer", "null"]}` (first iteration of this patch) is valid according to OpenAPI **3.1** spec: https://github.com/OAI/OpenAPI-Specification/pull/1977

For **3.0**, we will have to use `nullable: true`.

src/JsonSchema/TypeFactory.php
Ocramius: I can revert this, since most of the embedded payload items are going to be `$ref` anyway, and `stdClass` can be mapped explicitly (instead of using `object`)

teohhanhui:
And it looks like OpenAPI 3.1 is still not released yet, though it'd be nice to be forward-compatible (`nullable` becoming deprecated).

src/JsonSchema/TypeFactory.php
Ocramius: I'll revert this, and open a new issue about it.

Ocramius: Reverted, deferred to https://github.com/api-platform/core/issues/3403

tests/JsonSchema/TypeFactoryTest.php
Ocramius: Reverted: deferred to https://github.com/api-platform/core/issues/3403

Ocramius:
Ok, besides 7.3 builds (that seem to fail due to `hasRole()` calls), this should be all green.

src/JsonSchema/TypeFactory.php
```diff
@@ -27,6 +27,19 @@
...
+     * a format that is compatible with OpenAPI specifications **PREVIOUS** to 3.0.
+     *
+     * Without this flag being set, the generated format will only be compatible with Swagger 3.0 or newer.
+     *
+     * Once support for OpenAPI < 3.0 is gone, this constant **WILL BE REMOVED**
+     *
+     * @internal Once support for OpenAPI < 3.0 is gone, this constant **WILL BE REMOVED** - do not rely on
+     *           it in downstream projects!
+     */
+    public const CONTEXT_SERIALIZATION_FORMAT_OPENAPI_PRE_V3_0 = self::class.'::CONTEXT_SERIALIZATION_FORMAT_OPENAPI_PRE_V3_0';
```
teohhanhui: CS: PSR-12: `use` statements for traits must be at the top of the class, so this must be moved below.

src/JsonSchema/TypeFactory.php
```diff
@@ -50,14 +63,40 @@ public function setSchemaFactory(SchemaFactoryInterface $schemaFactory): void
...
     {
         if ($type->isCollection()) {
-            $subType = new Type($type->getBuiltinType(), $type->isNullable(), $type->getClassName(), false);
-
-            return [
-                'type' => 'array',
-                'items' => $this->getType($subType, $format, $readableLink, $serializerContext, $schema),
-            ];
+            $keyType = $type->getCollectionKeyType();
+            $subType = $type->getCollectionValueType()
```
teohhanhui: CS: keep in one line

src/JsonSchema/SchemaFactory.php
```diff
@@ -258,19 +258,19 @@ private function getMetadata(string $className, string $type = Schema::TYPE_OUTP
...
     {
         $attribute = Schema::TYPE_OUTPUT === $type ? 'normalization_context' : 'denormalization_context';
 
         if (null === $operationType || null === $operationName) {
-            return $resourceMetadata->getAttribute($attribute, []);
+            return array_merge($resourceMetadata->getAttribute($attribute, []), (array) $previousSerializerContext);
         }
 
-        return $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, [], true);
+        return array_merge($resourceMetadata->getTypedOperationAttribute($operationType, $operationName, $attribute, [], true), (array) $previousSerializerContext);
```
teohhanhui: Is the intention only to preserve `CONTEXT_SERIALIZATION_FORMAT_OPENAPI_PRE_V3_0`? If so, perhaps we'll have less complications just explicitly doing that outside of this function.

src/JsonSchema/TypeFactory.php
Ocramius: There's no `use` here: mis-placed comment?

src/JsonSchema/TypeFactory.php
Ocramius: Disagree: stated above in https://github.com/api-platform/core/pull/3402#discussion_r381574437

src/JsonSchema/SchemaFactory.php
Ocramius: This code generates a "default serializer context": if we push the context out of here, we add conditionals around all of it when only that one key is defined.

Instead, here we make it less messy, by always producing an `array` (be it already filled or not), and having all code around this detail ignore the specifics.

After all, this function is specifically called `getSerializerContext`: it is its job to prepare the serializer context for operation.

src/JsonSchema/TypeFactory.php
teohhanhui: @alanpoulain asked for the same thing in the previous comment. I think we'd like to keep our CS.

src/JsonSchema/TypeFactory.php
teohhanhui: `use` statement is at line 43 below. The `const` has to be moved below it.

teohhanhui: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-12-extended-coding-style-guide.md#42-using-traits

Ocramius: Ah, I see: I forget about traits, since I pretty much banned them overall in almost all what I do :+1:

Ocramius: dd017d27bb6406e91fcc3c5f4478338bd5a0572c

src/JsonSchema/TypeFactory.php
Ocramius: If the custom CS is really the blocker for this patch: edaac898c8c741a4b425acaaa85da86bf8dd94ec

src/JsonSchema/SchemaFactory.php
teohhanhui: Hmm... Indeed it's not the previous serializer context, but the provided serializer context, right?

Before: Use the provided serializer context, otherwise get the default from the resource metadata.
After: Always get the serializer context from the resource metadata, but merge it into the provided serializer context.

This change seems wrong to me, as it removes the possibility of overriding the serializer context (instead of getting it from the resource metadata), which looks to me to be the intention of the original design.

Ocramius: The original code provided its own serializer context *if* none was given.

Note that the added flag is not supposed to be given from the outside: we already have all the "schema version" flags for that.

teohhanhui: Consider this:

```php
$schemaFactory->buildSchema(..., $myCustomSerializerContext, ...)
```

Before this change, the provided serializer context would get used, overriding the serializer context defined in the resource metadata.

But after this change, that's no longer the case, as the provided serializer context merely gets merged into the serializer context defined in the resource metadata.

Ocramius: No, that's exactly why the `$previousSerializerContext` is the last argument here: its values need to be preserved.

teohhanhui: I think we're talking past each other here. Sorry... Let me try to rephrase:

The provided serializer context is supposed to override the serializer context defined in the resource metadata (which is used by default, if no serializer context is passed in).

teohhanhui: Sure, the `array_merge` keeps the keys from the passed in serializer context, but it'd still be a problem if the serializer context from the resource metadata has additional keys.

Ocramius: From what I saw thus far, it only did so when none was provided.

Do you have a test case that fails with this diff, and passes with the `2.5` state? I think that would highlight the issue in more detail.

Ocramius: > it'd still be a problem if the serializer context from the resource metadata has additional keys.

Does it make sense to override keys that are user-defined here (at all)? What would the keys be, for example?

teohhanhui: Example:

Serializer context in resource metadata:

```php
[
    'a' => 1,
    'b' => 2,
]
```

Serializer context passed in to `SchemaFactoryInterface::buildSchema`:

```php
[
    'a' => 1,
]
```

Expected (before this change) - passed in serializer context overrides what's defined in the resource metadata:

```php
[
    'a' => 1,
]
```

Actual (after this change) - passed in serializer context gets merged into what's defined in resource metadata:

```php
[
    'a' => 1,
    'b' => 2,
]
```

Ocramius: Hmm, at this point I won't get to fix (or inspect) that, because I'm off for a few weeks starting tomorrow: would you be OK if we documented this into a new issue?

Ocramius: BTW, what kind of issue could arise by inheriting the serializer context, instead of explicitly overriding it?

teohhanhui: About the potential impact, I don't know. There are so many things that could be in the serializer context (from Symfony itself, from API Platform, and of course any custom keys from the user).

Ocramius: Yep, `array` is indeed a squishy API. Will investigate then :)

teohhanhui: Resolved in https://github.com/Ocramius/core/commit/9d17089d1c75d65122bdd092c25dce00d2bf0763

teohhanhui: Resolved in https://github.com/Ocramius/core/commit/9d17089d1c75d65122bdd092c25dce00d2bf0763

teohhanhui:
Thanks @Ocramius! :tada: :rocket:

teohhanhui:
Houston, we have a problem.

Swagger UI does not support `oneOf` / `anyOf` / `allOf`: https://github.com/swagger-api/swagger-ui/issues/3803

This causes the property to be omitted from the example entirely.

Ocramius:
Is this pertinent to v3 or v2? For v2, we disabled everything, so the swagger UI should still work.

teohhanhui:
See the linked issues. Unfortunately, it seems like our use of `anyOf` has not gained us anything (does not fulfill the goal of allowing `null` value for a relation), but has caused us to lose something (unintentionally breaking Swagger UI - though it's really their fault at this point).

teohhanhui:
We can use the `NullValue` trick from https://github.com/OAI/OpenAPI-Specification/issues/1368#issuecomment-580103688

teohhanhui:
@Ocramius would it fit your use case if we only add `null` type for OpenAPI >= 3.1? I think it might be the sweet spot.

Ocramius:
@teohhanhui is there an issue tracking that? I would kinda isolate it to a smaller scope.

ahmed-bhs:
Hello there, is it normal, that the schema generated based on annotation `@ORM\Column(type="array")` could override the schema generated based on the PHPDoc ?

Example:

```php
/**
 * @var int[]
 * @ORM\Column(type="array")
 */
private $productIds = [];

```
Going to generate :

```
"productIds": {
        "description": "...",
        "type": "array",
        "items": {
            "type": "string"
        }
    },
```
Instead of:

```
"productIds": {
        "description": "...",
        "type": "array",
        "items": {
            "type": "integer"
        }
    },
``` 

Cheers!

Issue #3379, opened 2020-02-05
Bug introduced between  v2.4.6 => v2.5.4


tristanbes:
**API Platform version(s) affected**: somewhere between  after v2.4.6 and  v2.5.4

**Description**  
When I try to get a collection of users, in v2.4.6:
```
curl -X GET "http://cosmos.vm/api/users" -H "accept: application/ld+json"
```

Response:
```json
{
  "@context": "/api/contexts/User",
  "@id": "/api/users",
  "@type": "hydra:Collection",
  "hydra:member": [
    {
      "@id": "/api/users/1",
      "@type": "User",
      "id": 1,
      "email": "user1@example.com",
      "roles": [
        "ROLE_TECHPRO_ADMIN",
        "ROLE_USER"
      ],
      "civility": "CIVILITY_MISTER",
      "firstName": "Lily",
      "lastName": "Dibbert",
      "mobile": "+33600000000",
      "phone": "+33400000000",
      "profession": "PROFESSION_DIRECTOR",
      "info": "Un texte d'info plutot court.",
      "avatar": {
        "@id": "/api/media/1",
        "@type": "http://schema.org/MediaObject",
        "title": "avatar 1",
        "filename": "sample_pic.jpeg",
        "mimeType": "image/png",
        "size": 76134831,
        "contentUrl": "http://cosmos.vm/media/cache/resolve/original/fixtures/sample_pic.jpeg"
      },
      "addressIsSameAsSociety": true,
      "address": {
        "@id": "/api/addresses/1",
        "id": 1,
        "street": "30 Rue Sergent Michel Berthet",
        "additional": null,
        "postalCode": "69009",
        "town": "Lyon",
        "country": "France"
      },
      "events": [],
      "createdAt": "2019-03-01T14:00:00+01:00",
      "updatedAt": "2019-04-01T15:55:00+02:00",
      "primaryRole": "ROLE_TECHPRO_OFFICES",
      "sectors": null
    },
.......
```

**How to reproduce**  
Upgrade to v2.5.4

```
curl -X GET "http://cosmos.vm/api/users" -H "accept: application/ld+json"
```

**Response:**
```json
{
  "@context": "/api/contexts/Error",
  "@type": "hydra:Error",
  "hydra:title": "An error occurred",
  "hydra:description": "Unexpected non-iterable value for to-many relation.",
  "trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
[...]

Issue #3278, opened 2019-11-20
Unexpected non-iterable value for to-many relation


pesseyjulien:
Hi,

Updated to the latest version (2.5.2), and now I'm getting the following error :

```
Symfony\Component\Serializer\Exception\UnexpectedValueException: Unexpected non-iterable value for to-many relation.
#9 /vendor/api-platform/core/src/Serializer/AbstractItemNormalizer.php(526): ApiPlatform\Core\Serializer\AbstractItemNormalizer::getAttributeValue
#8 /vendor/symfony/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php(181): Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer::normalize
#7 /vendor/api-platform/core/src/Serializer/AbstractItemNormalizer.php(151): ApiPlatform\Core\Serializer\AbstractItemNormalizer::normalize
#6 /vendor/symfony/symfony/src/Symfony/Component/Serializer/Serializer.php(152): Symfony\Component\Serializer\Serializer::normalize
#5 /vendor/symfony/symfony/src/Symfony/Component/Serializer/Serializer.php(125): Symfony\Component\Serializer\Serializer::serialize
#4 /src/AppBundle/Controller/JobController.php(90): AppBundle\Controller\JobController::createJobAction
#3 /vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(151): Symfony\Component\HttpKernel\HttpKernel::handleRaw
#2 /vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(68): Symfony\Component\HttpKernel\HttpKernel::handle
#1 /vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(198): Symfony\Component\HttpKernel\Kernel::handle
#0 /public/index.php(24): null
```

Any idea why ?

teohhanhui:
Could you share what this value is? Might be a regression introduced in #3248

pesseyjulien:
No idea, here is the issue : https://sentry.io/share/issue/0062c90e3b27452ab8fb207489ee6d04/

teohhanhui:
Maybe some `dump` calls would help in debugging this.

pesseyjulien:
For now I have downgraded to the previous version, I will try some calls later and I will let you know. thanks !

im-jackhansard:
[...]

PR #3412, opened 2020-02-25
Case where Item attribute might be a Nullable to-many relationship

shehi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Case where Item attribute might be a Nullable to-many relationship. Otherwise this IF condition blocks the flow of the execution where it shouldn't.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -522,7 +522,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             ($className = $collectionValueType->getClassName()) &&
             $this->resourceClassResolver->isResourceClass($className)
         ) {
-            if (!$type->isNullable() && null === $attributeValue) {
+            if (null === $attributeValue && !$type->isNullable()) {
                 throw new UnexpectedValueException('Unexpected null value for non-nullable to-many relation.');
             }
             if (null !== $attributeValue && !is_iterable($attributeValue)) {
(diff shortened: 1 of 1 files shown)


shehi:
Behat failing, couldn't find those failing scenarios...

Few lines below the line I changed, you guys have put a proper check:
```
if (!\is_object($attributeValue) && null !== $attributeValue) {
    throw new UnexpectedValueException('Unexpected non-object value for to-one relation.');
}
```

However for some reason the same check against nullability doesn't exist. With that, one must put empty arrays while sending resource/dto request to API, which is quite ugly way of doing it.

shehi:
@teohhanhui , please review the overall changes I introduced and see if there is something about them that doesn't make sense. If everything looks OK, please direct me to proper Behat feature(s) to fix these failing tests.

teohhanhui:
As I've explained before, a "nullable to-many relationship" does not exist. You can have 0 items in the collection, but the relation itself cannot have a `null` value.

shehi:
I don't get it @teohhanhui : It is not a mandatory relation. Rule may exist, but actual references might not. As such, you don't send any data for that relation and skip them. EmbedMany or ReferenceMany relations can easily work like this. If that is the case, why would one send empty array for those fields in Api everytime they want to do something: it is a nuisance. One could just omit those fields altogether. This `IF` check here doesn't allow that. Now I have to send that relation field as an empty array every time I send an API post/put request.

shehi:
@dunglas , can we have your input on this one, because I believe @teohhanhui or me are misunderstanding something here.

teohhanhui:
> I don't know what you mean by "as I've explained before" - when did we have this convo?

Sorry, it was not with you. I remembered wrongly. But this is a duplicate of #3278 and #3379 (probably other issues as well).

> to-Many relations don't need to be a mandatory relation

That makes no sense. There is either a relation, or there is not. Each relationship has a [cardinality](https://en.wikipedia.org/wiki/Cardinality_(data_modeling)), but there's no optional relationship in the manner that you've described.

shehi:
> That makes no sense. There is either a relation, or there is not. Each relationship has a [cardinality](https://en.wikipedia.org/wiki/Cardinality_(data_modeling)), but there's no optional relationship in the manner that you've described.

After some thinking I actually discovered the cause of problem: you guys explain it wrongly.

In ORM entities and ODM documents, and in ApiResources based on those we just instantiate empty relation fields with empty collection. ```$this->fieldA = ArrayCollection();```. But in other types of ApiResources (mostly DTOs) we need to achieve the same result by instantiating the field to empty array: ```protected $fieldA = [];```. That is the cause of confusion.

Ok thanks. :D

PR #3418, opened 2020-02-27
feat: CookieJar on Client

Grafikart:
| Q             | A
| ------------- | ---
| Bug fix?      | yno
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Added a new method on Client to access the cookie jar to test api endpoint that uses cookie as a mechanism for authentication.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Test/Client.php
+++ b/src/Bridge/Symfony/Bundle/Test/Client.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\Test;
 
 use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Component\BrowserKit\CookieJar;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpClient\HttpClientTrait;
 use Symfony\Component\HttpKernel\KernelInterface;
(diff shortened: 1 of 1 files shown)


Grafikart:
I don't know why some tests are failing :(

soyuka:
we're working on it, it's not related to your change!

src/Bridge/Symfony/Bundle/Test/Client.php
```diff
@@ -166,6 +167,16 @@ public function getContainer(): ?ContainerInterface
         return $this->kernelBrowser->getContainer();
     }
 
+    /**
```
teohhanhui: Please remove useless phpdoc.

src/Bridge/Symfony/Bundle/Test/Client.php
```diff
@@ -166,6 +167,16 @@ public function getContainer(): ?ContainerInterface
...
+     * Returns the CookieJar instance.
```
teohhanhui: The summary should be more descriptive.

src/Bridge/Symfony/Bundle/Test/Client.php
```diff
@@ -166,6 +167,16 @@ public function getContainer(): ?ContainerInterface
...
+     *
+     * @return CookieJar A CookieJar instance
```
teohhanhui: Remove this `@return` as it doesn't add any useful information.

teohhanhui:
I don't think this is the right approach.

> The HTTP client provided by this component is stateless but handling cookies requires a stateful storage (because responses can update cookies and they must be used for subsequent requests). That's why this component doesn't handle cookies automatically.
> 
> https://symfony.com/doc/current/components/http_client.html#cookies

The same rationale should apply here. The `CookieJar` class should be used outside of the `Client` class. The `Client` class is indeed stateful in regards to the kernel, a necessary(?) evil.

Grafikart:
I copied the comment (phpdoc) from the Symfony\Component\BrowserKit\Browserkit so I'm not sure I can describe it better :(
You are right for the return it's useless and I remove it.

dunglas:
Thanks @Grafikart!

PR #3425, opened 2020-03-03
Pass $options to PropertyInfoMetadataFactory

### LINKED ISSUE #2868: Non-mapped properties and non-mapped nested classes are not shown in swagger output since update to v2.4
As the title suggests, non-mapped properties are not shown in swagger output anymore.
There are different scenarios. Sometimes one entity has a non-mapped property, some entities have other non-mapped entities with naturally also non-mapped properties. They all have in commen, that they don't appear in swagger anymore. They also have propper getters and setters.
Serialization/Dezerialization of them are working like before.

### LINKED ISSUE #3146: Serialization context is ignored when building JsonSchema

### LINKED ISSUE #3424: Wrong Swagger types when creating relations on a single POST
**API Platform version(s) affected**: 2.5.1

**Description**  
When allowing to create relations using `denormalization_context`, the generated Swagger documentation is a list of strings instead of an union of strings and objects.

**How to reproduce**  
It's easy to reproduce the issue on the demo. We have by default the following types (which are already wrong) on the Book entity:
```json
{
  //...
  "reviews": [
      {
        "body": "string",
        "rating": 0,
        "author": "string",
        "publicationDate": "2020-03-02T10:31:24.895Z"
      }
  ]
}
```

but the expected types should be an array of unions (`oneOf`) of the above type with strings (in the case IRIs are passed).

Now, for the most annoying bug:

In Review, we have the following annotations on `body`:

```php
    /**
     * @var string The actual body of the review
     *
     * @ORM\Column(type="text", nullable=true)
     * @Groups({"review:read", "book:read"})
     * @ApiProperty(iri="http://schema.org/reviewBody")
     */
    public $body;
```

If you remove the `book:read` group, then the generated swagger type showed above is turned into:
```json
{
  //...
  "reviews": ["string"]
}
```

Which does not make any sense. I have removed here an annotation related to the `normalizationContext` and it changes the types which are related to the `denormalization_context`. This change should not have impacted the expected type on POST.

**Possible Solution**  
Generated types should be an union type of `string` (to allow the user to passe an IRI) and the body matching the denormalization configuration. In the case of the demo, the change shown above should not have impacted the JSON types.

**Additional Context**
Before the change (good but should be an union):
image

After the change (string only despite allowing creating the relation directly):
image

remoteclient:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #2868, fixes #3146, fixes #3424
| License       | MIT
| Doc PR        | N/A

I had an issue where nested properties / classes didn't show up on OpenApi / Swagger docs.
With this PR the issue seems to be solved as the PropertyMetadataFactory gets the options. Thanks to @teohhanhui for the help on this.

TODO:

- [x] Fix tests (@teohhanhui)
- [x] Add tests (@teohhanhui)

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -133,7 +133,7 @@ final class SchemaFactory implements SchemaFactoryInterface
 
         $options = isset($serializerContext[AbstractNormalizer::GROUPS]) ? ['serializer_groups' => (array) $serializerContext[AbstractNormalizer::GROUPS]] : [];
         foreach ($this->propertyNameCollectionFactory->create($inputOrOutputClass, $options) as $propertyName) {
-            $propertyMetadata = $this->propertyMetadataFactory->create($inputOrOutputClass, $propertyName);
+            $propertyMetadata = $this->propertyMetadataFactory->create($inputOrOutputClass, $propertyName, $options);
             if (!$propertyMetadata->isReadable() && !$propertyMetadata->isWritable()) {
                 continue;
             }
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -131,7 +131,7 @@ final class SchemaFactory implements SchemaFactoryInterface
             $definition['externalDocs'] = ['url' => $iri];
         }
 
-        $options = isset($serializerContext[AbstractNormalizer::GROUPS]) ? ['serializer_groups' => (array) $serializerContext[AbstractNormalizer::GROUPS]] : [];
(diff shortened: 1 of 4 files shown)


teohhanhui:
Thanks @remoteclient! :tada: :rocket:

PR #3427, opened 2020-03-05
Only save /tmp/phpstan/resultCache.php to cache

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

The result cache is a huge win.

CircleCI takes a very long time saving /tmp/phpstan/cache for some reason,
and this cache doesn't affect the run time much anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,12 +162,12 @@ commands:
       - save_cache:
           name: Save PHPStan cache
           paths:
-            - /tmp/phpstan/cache
+            - /tmp/phpstan/resultCache.php
           key: phpstan-{{ .Branch }}-{{ .BuildNum }}
       - save_cache:
           name: Save PHPStan cache
           paths:
-            - /tmp/phpstan/cache
+            - /tmp/phpstan/resultCache.php
           key: phpstan-{{ .Revision }}-{{ .BuildNum }}
   update-project-dependencies:
     steps:


teohhanhui:
Somehow the result cache is not being used. Investigating...

teohhanhui:
/cc @ondrejmirtes perhaps you could lend a hand here... :pray:

Here you can see that `/tmp/phpstan/resultCache.php` is restored (and the archive size of 36 KiB matches what I have locally):
https://app.circleci.com/jobs/github/api-platform/core/22032/parallel-runs/0/steps/0-113

But here phpstan is taking 14s, so it doesn't seem like the result cache is being used:
https://app.circleci.com/jobs/github/api-platform/core/22032/parallel-runs/0/steps/0-114

Is there some kind of file modification checking that breaks with a CI workflow?

ondrejmirtes:
Yes, in some CI systems the file modification time will not be preserved across runs. I plan to switch to file hashes instead but I need to inspect performance implications of that.

ondrejmirtes:
Right now I verified the result cache works in Jenkins and TeamCity, but I'm not sure about GitLab CI and similar Docker-based systems.

teohhanhui:
md5 should be very fast, even though it's prone to collisions... But good enough for this kind of use case. :laughing:

teohhanhui:
Merging as is to avoid the CircleCI issue with saving `/tmp/phpstan/cache`.

PR #3431, opened 2020-03-06
Add a stub package for ext-mongodb

### LINKED ISSUE #3236: Move some dev-deps on suggest
On the last clone I made just to check for things, I couldn't run `composer install` because I don't have mongo (or its ext) installed.

If you want to contribute to API Platform, you shouldn't force the dev to install these dependencies (or worse, go through `--ignore-platform-reqs` which is something that should never be used, and is sure to make the project go :boom: ). These should be "suggested" instead. So that if the contributor want to contribute to the odm / orm bridge (for example), these should be installed on his part.

The dev dependency should really be limited to just things such as phpstan, phpunit and dev tools used.

This kinda related to #3122. If you don't limit, don't ask the contributor to use something else.

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3236, fixes https://github.com/api-platform/core/pull/3368#discussion_r372307596
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,5 @@
 version: 2.1
 
-.constants:
-  - &ext-mongodb-version '1.6.1'
-
 commands:
   clear-test-app-cache:
     steps:
@@ -19,11 +16,6 @@ commands:
       - run:
           name: Disable xdebug PHP extension
           command: sudo rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
-  install-mongodb-php-extension:
-    steps:
-      - install-php-extension-from-pecl:
-          extension_name: mongodb
-          extension_version: *ext-mongodb-version
   install-php-extension-from-pecl:
     parameters:
       extension_name:
@@ -64,16 +56,6 @@ commands:
             - composer-{{ .Revision }}
             - composer-{{ .Branch }}
             - composer-
-  restore-mongodb-php-extension-build-cache:
-    parameters:
-      php_version:
-        description: PHP version (major.minor).
-        type: string
-    steps:
-      - restore-php-extension-build-cache:
-          extension_name: mongodb
-          extension_version: *ext-mongodb-version
-          php_version: << parameters.php_version >>
   restore-php-cs-fixer-cache:
     steps:
       - restore_cache:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -231,11 +203,6 @@ jobs:
     working_directory: ~/api-platform/core
     steps:
       - checkout
-      - restore-mongodb-php-extension-build-cache:
-          php_version: << parameters.php_version >>
-      - install-mongodb-php-extension
```
teohhanhui: Rationale: Let PHPStan use our stubs so that we know when it's broken (time to sync changes from https://github.com/JetBrains/phpstorm-stubs/blob/master/mongodb/mongodb.php)... :pray:

dunglas:
Awesome

teohhanhui:
Do you have any concerns about these stub files being in the repository?

composer.json
```diff
@@ -87,7 +87,8 @@
         "symfony/web-profiler-bundle": "^4.2 || ^5.0",
         "symfony/yaml": "^3.4 || ^4.0 || ^5.0",
         "twig/twig": "^1.42.3 || ^2.12",
-        "webonyx/graphql-php": ">=0.13.1 <1.0"
+        "webonyx/graphql-php": ">=0.13.1 <1.0",
+        "zz-api-platform-dev/stubs-mongodb": "@dev"
```
soyuka: what's with the zz-api-platform-dev ? :p

soyuka:
Indeed I'm not a huge fan of having this in this repository, can't we install this from a distant package via composer?

composer.json
teohhanhui: So that it's sorted at the end. 🤯

teohhanhui:
I'm not sure, but we might run into https://github.com/composer/composer/issues/5030

teohhanhui:
@soyuka I've extracted the stubs to https://github.com/teohhanhui/stubs-mongodb

teohhanhui:
Travis CI needed a bit of hand holding. Hopefully future jobs will be fine...

soyuka:
Awesome job thanks @teohhanhui it'll help a lot on contributing to api-platform :D

PR #3435, opened 2020-03-09
Support HTTP "From" header when using Gedmo Blameable extension

### LINKED ISSUE #3383: Support HTTP "From" header when using Gedmo Blameable extension
**Description**  
There are situations when a user making a request does not necessarily need to be authenticated with an API. We use this functionality in our internal APIs to track the users that have made CRUD requests. I think this functionality should be upstream rather than in a separate github extension repository or maintained locally within a private repository.

**Example**  
Similar to the FOSUserBundle integration, this integration would be controlled by a flag for another Bridge. By default, it would accept this "From" header on any POST or PATCH request. This would be added to the Swagger documentation for these endpoints automatically.

In addition two other config values should be available:
* An array of regex matches on URLs to disable this in the documentation for any specific endpoint
* A boolean config that allows this to be disabled on any endpoint

inferont:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3383 
| License       | MIT

**Description**
There are situations when a user making a request does not necessarily need to be authenticated with an API. We use this functionality in our internal APIs to track the users that have made CRUD requests. I think this functionality should be upstream rather than in a separate github extension repository or maintained locally within a private repository.

**Example**
Similar to the FOSUserBundle integration, this integration would be controlled by a flag for another Bridge. By default, it would accept this "From" header on any POST or PATCH request. This would be added to the Swagger documentation for these endpoints automatically.

**Testing Notes**
1. Add the following to the api-platform.yaml config
`doctrine_gedmo_extensions:
        from_header_enabled: true
        from_header_unmatch: ['/_action/']`
2. Make sure the gedmo blameable extension is installed 
3. Add the blameable annotation to a field in an entity
4. Submit an update with a "From" header
5. See the "From" user being written to the entity

**Unit tests will be added to the PR after I get a thumbs up from an API Platform developer**


inferont:
@soyuka Would you please review this? You flagged the issue as an enhancement.

src/Bridge/GedmoDoctrine/EventListener/FromListener.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\GedmoDoctrine\EventListener;
```
teohhanhui:
```suggestion
namespace ApiPlatform\Core\Bridge\Gedmo\DoctrineExtensions\EventListener;
```

src/Bridge/GedmoDoctrine/Model/FromUser.php
```diff
@@ -0,0 +1,46 @@
...
+namespace ApiPlatform\Core\Bridge\GedmoDoctrine\Model;
+
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * A model for holding the email address sent in a "From" header request.
+ *
+ * @author Ryan Jefferson <ryanhjefferson@gmail.com>
+ */
+class FromUser
```
teohhanhui: IMO this class should be removed. Just set the email address directly.

src/Bridge/GedmoDoctrine/Swagger/SwaggerFromDecorator.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\GedmoDoctrine\Swagger;
```
teohhanhui:
```suggestion
namespace ApiPlatform\Core\Bridge\Gedmo\DoctrineExtensions\Serializer;
```

src/Bridge/GedmoDoctrine/Swagger/SwaggerFromDecorator.php
```diff
@@ -0,0 +1,69 @@
...
+
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+/**
+ * Class SwaggerFromDecorator.
+ *
+ * @author Ryan Jefferson <ryanhjefferson@gmail.com>
+ */
+final class SwaggerFromDecorator implements NormalizerInterface
```
teohhanhui:
```suggestion
final class SwaggerDocumentationNormalizer implements NormalizerInterface
```

teohhanhui:
There is a bigger problem. This behaviour is too specific to your use case. I doubt this is generally useful for the typical API Platform user.

teohhanhui:
Typically, one would expect to set an instance of `Symfony\Component\Security\Core\User\UserInterface`. While I understand that's not your use case, I'm not convinced that it's actually worth having this code in API Platform to support a very specific use case.

inferont:
@teohhanhui 
> There is a bigger problem. This behaviour is too specific to your use case. I doubt this is generally useful for the typical API Platform user.

I agree that this is not generally useful for the current typical API Platform user. A lot of other code in this repository isn't generally useful either though? As an example, the FOS User Bridge is recommend against in the documentation itself, yet it is still there.

I think that there is an opportune area of development for API Platform with microservices or backend to backend APIs. Implementing this "From" header support allows us to support the basics of authentication on one backend and forward it directly to an API without creating a weird second authentication mechanism for an internal API. It also has other uses 

REST APIs attempt to closely follow the HTTP standard.  This "From" header has been a part of the HTTP 1.1 spec for more than two decades. I am attempting to support this standard header within API Platform in a way that would be used by most Symfony applications.
- https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/From

teohhanhui:
> As an example, the FOS User Bridge is recommend against in the documentation itself, yet it is still there.

It's legacy code. We're going to deprecate it.

> Implementing this "From" header support allows us to support the basics of authentication on one backend and forward it directly to an API without creating a weird second authentication mechanism for an internal API.

It's important to note:
> You shouldn't use the `From` header for access control or authentication.
>
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/From

So please don't use this other than for tracking.

> This "From" header has been a part of the HTTP 1.1 spec for more than two decades. I am attempting to support this standard header within API Platform in a way that would be used by most Symfony applications.

I agree on the first sentence, but disagree on the second sentence. How you use `BlameableListener::setUserValue` is very specific to your application. A typical Symfony / API Platform application would expect an instance of `Symfony\Component\Security\Core\User\UserInterface` (as I've mentioned above).

inferont:
@teohhanhui 
> It's legacy code. We're going to deprecate it.
When it's deprecated, are you going to move it somewhere else? I see a few hundred projects on github relying on it. Maybe having another repository for these less common bridges like the doctrine extensions project would be useful? I have a few ideas for other bridges that would be useful.  If there's an API Platform bridges project I could contribute this and other stuff there instead.

> So please don't use this other than for tracking.
That's exactly the use case for this. This From header is only intended for tracking after something else performed authentication or when authentication is not needed.

> I agree on the first sentence, but disagree on the second sentence. How you use BlameableListener::setUserValue is very specific to your application. A typical Symfony / API Platform application would expect an instance of Symfony\Component\Security\Core\User\UserInterface (as I've mentioned above).

I think there's degrees of support for users here:
1. Tracking users requests: From header is sufficient and is the HTTP standard
2. Authenticating users on API
3. Authenticating users and validating requests for each resource

So, what do you think? Should I finish this PR up with your code suggestions?

soyuka:
> This From header is only intended for tracking after something else performed authentication or when authentication is not needed.

But this value then ends up in queries via blameable extension. I don't think that this header should be used for authentication. 

IMO this is way too specific, requires the user to be in a case he use Blameable + this From header which won't be most cases. I'm :-1:  for having this in core.

PR #3436, opened 2020-03-09
Add stateless ApiResource attribute

mtarld:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | TODO

In Symfony 5.1, it will be possible to declare routes as stateless (https://github.com/symfony/symfony/pull/35732)
For stateless routes, kernel will either throw an exception (debug enabled) or log a warning (debug disabled) on session usage.

As API Platform shouldn't use session (according to REST architecture constraints), this PR proposes to add an ApiResource attribute in order to declare related routes as stateless: `stateless`

`false` will be the default in order to prevent BC break (and we could set it at `true` in the next major)

One question still remains. API Plateform could be on top of older version of Symfony (4.x, 5.0.x).
How to handle it ?
- Throw an exception saying that this feature isn't supported for that Symfony's version
- Log a warning saying that this feature isn't supported for that Symfony's version
- Do nothing (which is dangerous IMO)

WDYT ?

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -63,6 +63,7 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
  *     @Attribute("securityPostDenormalize", type="string"),
  *     @Attribute("securityPostDenormalizeMessage", type="string"),
  *     @Attribute("shortName", type="string"),
+ *     @Attribute("stateless", type="bool"),
  *     @Attribute("subresourceOperations", type="array"),
  *     @Attribute("sunset", type="string"),
  *     @Attribute("swaggerContext", type="array"),
@@ -118,6 +119,7 @@ final class ApiResource
         'paginationPartial',
         'paginationViaCursor',
         'routePrefix',
+        'stateless',
         'sunset',
         'swaggerContext',
         'urlGenerationStrategy',
@@ -413,6 +415,13 @@ final class ApiResource
      * @var string
      */
     private $securityPostDenormalizeMessage;
+
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var bool
+     */
+    private $stateless;
 
     /**
      * @see https://api-platform.com/docs/core/deprecations/#setting-the-sunset-http-header-to-indicate-when-a-resource-or-an-operation-will-be-removed
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -257,6 +257,11 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             }
         }
 
+        if (!\array_key_exists('stateless', $defaults)) {
+            @trigger_error('Not setting the "api_platform.defaults.stateless" configuration is deprecated since API Platform 2.6 and it will default to `true` in 3.0. You can override this at the operation level if you have stateful operations (highly not recommended).', E_USER_DEPRECATED);
+            $normalizedDefaults['attributes']['stateless'] = false;
+        }
+
         return $normalizedDefaults;
     }
 
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -126,6 +126,7 @@ final class ApiLoader extends Loader
                     [
                         '_controller' => $controller,
                         '_format' => null,
+                        '_stateless' => $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless'),
                         '_api_resource_class' => $operation['resource_class'],
                         '_api_subresource_operation_name' => $operation['route_name'],
                         '_api_subresource_context' => [
@@ -229,6 +230,7 @@ final class ApiLoader extends Loader
             [
                 '_controller' => $controller,
                 '_format' => null,
+                '_stateless' => $operation['stateless'],
                 '_api_resource_class' => $resourceClass,
                 sprintf('_api_%s_operation_name', $operationType) => $operationName,
             ] + ($operation['defaults'] ?? []),
--- a/src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
@@ -59,7 +59,7 @@ final class OperationResourceMetadataFactory implements ResourceMetadataFactoryI
 
         $collectionOperations = $resourceMetadata->getCollectionOperations();
         if (null === $collectionOperations) {
-            $resourceMetadata = $resourceMetadata->withCollectionOperations($this->createOperations($isAbstract ? ['GET'] : ['GET', 'POST']));
+            $resourceMetadata = $resourceMetadata->withCollectionOperations($this->createOperations($isAbstract ? ['GET'] : ['GET', 'POST'], $resourceMetadata));
         } else {
             $resourceMetadata = $this->normalize(true, $resourceClass, $resourceMetadata, $collectionOperations);
         }
@@ -76,7 +76,7 @@ final class OperationResourceMetadataFactory implements ResourceMetadataFactoryI
                 }
             }
 
-            $resourceMetadata = $resourceMetadata->withItemOperations($this->createOperations($methods));
+            $resourceMetadata = $resourceMetadata->withItemOperations($this->createOperations($methods, $resourceMetadata));
         } else {
             $resourceMetadata = $this->normalize(false, $resourceClass, $resourceMetadata, $itemOperations);
         }
@@ -91,11 +91,11 @@ final class OperationResourceMetadataFactory implements ResourceMetadataFactoryI
         return $resourceMetadata;
     }
 
-    private function createOperations(array $methods): array
+    private function createOperations(array $methods, ResourceMetadata $resourceMetadata): array
     {
         $operations = [];
         foreach ($methods as $method) {
-            $operations[strtolower($method)] = ['method' => $method];
+            $operations[strtolower($method)] = ['method' => $method, 'stateless' => $resourceMetadata->getAttribute('stateless')];
         }
 
         return $operations;
@@ -130,6 +130,8 @@ final class OperationResourceMetadataFactory implements ResourceMetadataFactoryI
             if (isset($operation['method'])) {
                 $operation['method'] = strtoupper($operation['method']);
             }
+
+            $operation['stateless'] = $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless');
 
             $newOperations[$operationName] = $operation;
         }
--- a/src/Operation/Factory/SubresourceOperationFactory.php
+++ b/src/Operation/Factory/SubresourceOperationFactory.php
@@ -26,7 +26,7 @@ final class SubresourceOperationFactory implements SubresourceOperationFactoryIn
 {
     public const SUBRESOURCE_SUFFIX = '_subresource';
     public const FORMAT_SUFFIX = '.{_format}';
-    public const ROUTE_OPTIONS = ['defaults' => [], 'requirements' => [], 'options' => [], 'host' => '', 'schemes' => [], 'condition' => '', 'controller' => null];
+    public const ROUTE_OPTIONS = ['defaults' => [], 'requirements' => [], 'options' => [], 'host' => '', 'schemes' => [], 'condition' => '', 'controller' => null, 'stateless' => null];
 
     private $resourceMetadataFactory;
     private $propertyNameCollectionFactory;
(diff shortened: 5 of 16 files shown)


teohhanhui:
I think the default cannot be `false`, but `null`. This allows us to trigger a deprecation error when it's `null`, so as to force the user to set it to `true` (or `false`). And then we can default to `true` in 3.0

WDYT @dunglas?

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -134,7 +134,7 @@ public function load($data, $type = null): RouteCollection
                             'collection' => $operation['collection'],
                             'operationId' => $operationId,
                         ],
-                    ] + ($operation['defaults'] ?? []),
+                    ] + ($operation['defaults'] ?? []) + ($resourceMetadata->getStateless() ? ['_stateless' => true] : []),
```
teohhanhui: This doesn't seem correct to me. We wouldn't want to have special cases like this.

teohhanhui:
> API Platform could be on top of older version of Symfony (4.x, 5.0.x).
> How to handle it ?

We could detect the version of `symfony/http-kernel` by checking if `Symfony\Component\HttpKernel\Exception\UnexpectedSessionUsageException` exists. I don't think doing nothing is dangerous, as it's just an extra safety net that didn't exist before, so we use it if it exists, otherwise nothing has changed.

mtarld:
> We could detect the version of `symfony/http-kernel` by checking if `Symfony\Component\HttpKernel\Exception\UnexpectedSessionUsageException` exists. I don't think doing nothing is dangerous, as it's just an extra safety net that didn't exist before, so we use it if it exists, otherwise nothing has changed.

We could as well use `version_compare` on `Kernel::VERSION` ?

Furthermore, what I meant is just telling the developer that he is using `stateless` but nothing will occur (as it's not handled by Symfony)

src/Bridge/Symfony/Routing/ApiLoader.php
mtarld: Yes I get it, but how do you think we should handle it ?

teohhanhui:
> We could as well use version_compare on Kernel::VERSION ?

Uhh, yeah, I forgot this is the HttpKernel we're talking about. Lol...

> telling the developer that he is using stateless but nothing will occur (as it's not handled by Symfony)

Personally, I think it's fine to just do nothing. It should be documented that this feature requires Symfony 5.1

src/Annotation/ApiResource.php
```diff
@@ -165,6 +167,13 @@ final class ApiResource
      */
     public $subresourceOperations;
 
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var bool
+     */
+    public $stateless;
```
teohhanhui: Sort alphabetically. `T` comes before `U`...

src/Annotation/ApiResource.php
```diff
@@ -158,6 +160,13 @@ final class ApiResource
      */
     public $shortName;
 
+    /**
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var bool|null
```
teohhanhui:
```suggestion
     * @var bool
```

It cannot be set to `null` here. `null` is only used as the default value for the attribute.

src/Bridge/Symfony/Routing/ApiLoader.php
mtarld: As discussed together, it should be put in resource metadata attributes instead

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -134,7 +136,7 @@ public function load($data, $type = null): RouteCollection
                             'collection' => $operation['collection'],
                             'operationId' => $operationId,
                         ],
-                    ] + ($operation['defaults'] ?? []),
+                    ] + ($operation['defaults'] ?? []) + $stateless,
```
teohhanhui:
```suggestion
                    ] + ($operation['defaults'] ?? []),
```

Revert this and add after line 130:

```php
                       '_stateless' => $operation['stateless'],
```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -216,6 +216,10 @@ private function normalizeDefaults(array $defaults): array
             }
         }
 
+        if (!\array_key_exists('stateless', $defaults)) {
+            @trigger_error("The node 'api_platform.defaults.stateless' should be specified", E_USER_DEPRECATED);
```
teohhanhui:
```suggestion
        if (true !== $defaults['stateless'] ?? null) {
            @trigger_error('Not setting the "api_platform.defaults.stateless" configuration option to `true` is deprecated since API Platform 2.6 and it will default to `true` in 3.0. You can override this at the operation level if you have stateful operations (highly not recommended).', E_USER_DEPRECATED);
```

mtarld: Yes, much better!
But I may recommend to write:
> Not setting the "api_platform.defaults.stateless" configuration option to `true` is deprecated since API Platform 2.6 and it will default to `true` in 3.0. You can override this **at the resource or the operation level** if you have stateful operations (highly not recommended).

instead.
WDYT ?

teohhanhui: While technically correct, I don't want to bring attention to that because what's a stateful resource?

mtarld: Yes you're right

mtarld:
Ok, using the current approach, stateless priority will be handled as the following:
1. Operation stateless attribute in `attributes`
2. Operation stateless attribute
3. Resource stateless attribute in `attributes`
4. Resource stateless attribute
5. Global defaults

Same goes for subresources.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -216,6 +216,10 @@ private function normalizeDefaults(array $defaults): array
             }
         }
 
+        if (true !== ($defaults['stateless'] ?? null)) {
+            @trigger_error('Not setting the "api_platform.defaults.stateless" configuration option to `true` is deprecated since API Platform 2.6 and it will default to `true` in 3.0. You can override this at the operation level if you have stateful operations (highly not recommended).', E_USER_DEPRECATED);
```
dunglas, line 220: It shouldn’t be forced to be true IMHO. Some (most ?) authorization mechanisms supported by Symfony require to use the session, and - even if using stateless mechanisms is a best practice - this must continue to be supported. It should be possible to disable the stateless mode (but it’s ok to have it enabled by default).

mtarld: I totally agree with you, I'll remove it then.

teohhanhui: @mtarld We must still trigger a deprecation error, however not forcing it to be set to `true`.

mtarld: Ok! Which message do you want to trigger then?

mtarld: Ok, then the new message will be:
`Not setting the "api_platform.defaults.stateless" configuration is deprecated since API Platform 2.6 and it will default to `true` in 3.0. You can override this at the operation level if you have stateful operations (highly not recommended).`

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -121,11 +121,14 @@ public function load($data, $type = null): RouteCollection
...
                 }
 
+                $operationStateless = $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless', false);
+
                 $routeCollection->add($operation['route_name'], new Route(
                     $operation['path'],
                     [
                         '_controller' => $controller,
                         '_format' => null,
+                        '_stateless' => $operationStateless,
```
teohhanhui, line 131:
```suggestion
                        '_stateless' => $operation['stateless'],
```

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -224,11 +227,14 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
...
         $path .= $this->operationPathResolver->resolveOperationPath($resourceShortName, $operation, $operationType, $operationName);
 
+        $operationStateless = $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless', false);
+
         $route = new Route(
             $path,
             [
                 '_controller' => $controller,
                 '_format' => null,
+                '_stateless' => $operationStateless,
```
teohhanhui, line 237:
```suggestion
                '_stateless' => $operation['stateless'],
```

src/Metadata/Resource/Factory/OperationResourceMetadataFactory.php
```diff
@@ -131,6 +131,10 @@ private function normalize(bool $collection, string $resourceClass, ResourceMeta
                 $operation['method'] = strtoupper($operation['method']);
             }
 
+            if ($stateless = ($operation['attributes']['stateless'] ?? $operation['stateless'] ?? null)) {
+                $operation['stateless'] = $stateless;
+            }
```
teohhanhui, line 136:
```suggestion
            $operation['stateless'] ?? $operation['stateless'] = $resourceMetadata->getAttribute('stateless');
```

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -121,11 +121,14 @@ public function load($data, $type = null): RouteCollection
                     }
                 }
 
+                $operationStateless = $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless', false);
+
```
teohhanhui, line 125:
```suggestion
```

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -224,11 +227,14 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
         $path = trim(trim($resourceMetadata->getAttribute('route_prefix', '')), '/');
         $path .= $this->operationPathResolver->resolveOperationPath($resourceShortName, $operation, $operationType, $operationName);
 
+        $operationStateless = $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless', false);
+
```
teohhanhui, line 231:
```suggestion
```

src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -142,13 +142,19 @@ private function computeSubresourceOperations(string $resourceClass, array &$tre
                 if (!\in_array($rootShortname, $operation['shortNames'], true)) {
                     $operation['shortNames'][] = $rootShortname;
                 }
+
```
teohhanhui, line 145: Please add `stateless` to `ROUTE_OPTIONS` and revert all other changes in this file.

teohhanhui:
I think there are some fundamental problems with the current implementation... This needs work.

teohhanhui:
And sorry for the earlier confusion, there's no explicit `attributes` key under each operation. The attributes are just directly specified.

mtarld:
No worries! I should have been aware of that. The new implementation is much cleaner now, thanks.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -216,6 +216,11 @@ private function normalizeDefaults(array $defaults): array
             }
         }
 
+        if (!\array_key_exists('stateless', $defaults)) {
+            @trigger_error('Not setting the "api_platform.defaults.stateless" configuration is deprecated since API Platform 2.6 and it will default to `true` in 3.0. You can override this at the operation level if you have stateful operations (highly not recommended).', E_USER_DEPRECATED);
```
dunglas, line 220: Why just not default to `false` instead of forcing the user to set this key explicitly?

mtarld: How would you notify that the configuration will default to `true` in `3.0` then?

mtarld: And by the way, it falls back to false already.

WDYT @dunglas ?

dunglas: LGTM!

PR #3448, opened 2020-03-13
Fixed an error where an exception was thrown if the messenger input data has an id field

karser:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

How to reproduce:
```
#api_resources.yaml:
App\Dto\DummyWithIdField:
    attributes:
        messenger: true
        output: false
    collectionOperations:
        post:
            path: '/path/to/dummy'
            status: 202
    itemOperations: {}
```

```
namespace App\Dto;

class DummyWithIdField {
    /** @var string|null */
    public $id;
}
```

```
curl --location --request POST 'https://api.domain.com/path/to/dummy' \
--header 'Content-Type: application/json' \
--data-raw '{"id":"ev_16CHO7Rt6vyOp8ng"}
'
```

The error in response:
```
{
    "@type": "hydra:Error",
    "hydra:description": "Update is not allowed for this operation.",
    "trace": [
        {
            "namespace": "",
            "short_class": "",
            "class": "",
            "type": "",
            "function": "",
            "file": "vendor/api-platform/core/src/Serializer/ItemNormalizer.php",
            "line": 59,
            "args": []
        },
```
The entire stacktrace https://gist.github.com/karser/9425657bff25da4026591e6af6e227fb

### DIFF
--- a/src/JsonApi/Serializer/ItemNormalizer.php
+++ b/src/JsonApi/Serializer/ItemNormalizer.php
@@ -132,7 +132,10 @@ final class ItemNormalizer extends AbstractItemNormalizer
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (!isset($context[self::OBJECT_TO_POPULATE]) && isset($data['data']['id'])) {
+        if (!isset($context[self::OBJECT_TO_POPULATE])
+            && isset($data['data']['id'])
+            && !$this->getResourceMetadataAttribute($class, 'messenger', false)
+        ) {
             if (true !== ($context['api_allow_update'] ?? true)) {
                 throw new NotNormalizableValueException('Update is not allowed for this operation.');
             }
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -108,7 +108,10 @@ final class ItemNormalizer extends AbstractItemNormalizer
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['@id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
+        if (isset($data['@id'])
+            && !isset($context[self::OBJECT_TO_POPULATE])
+            && !$this->getResourceMetadataAttribute($class, 'messenger', false)
+        ) {
             if (true !== ($context['api_allow_update'] ?? true)) {
                 throw new NotNormalizableValueException('Update is not allowed for this operation.');
             }
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -51,7 +51,10 @@ class ItemNormalizer extends AbstractItemNormalizer
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
+        if (isset($data['id'])
+            && !isset($context[self::OBJECT_TO_POPULATE])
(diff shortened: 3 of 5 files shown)


src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -132,7 +132,10 @@ public function supportsDenormalization($data, $type, $format = null): bool
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (!isset($context[self::OBJECT_TO_POPULATE]) && isset($data['data']['id'])) {
+        if (!isset($context[self::OBJECT_TO_POPULATE])
+            && isset($data['data']['id'])
+            && (null === $this->resourceMetadataFactory || !$this->resourceMetadataFactory->create($class)->getAttribute('messenger', false))
```
soyuka:
```
$this->resourceMetadataFactory->create($class)
```
may be null

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -106,7 +106,10 @@ public function supportsDenormalization($data, $type, $format = null): bool
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['@id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
+        if (isset($data['@id'])
+            && !isset($context[self::OBJECT_TO_POPULATE])
+            && (null === $this->resourceMetadataFactory || !$this->resourceMetadataFactory->create($class)->getAttribute('messenger', false))
```
soyuka:
```
$this->resourceMetadataFactory->create($class)
```
may be null

src/Serializer/ItemNormalizer.php
```diff
@@ -51,7 +51,10 @@ public function __construct(PropertyNameCollectionFactoryInterface $propertyName
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
+        if (isset($data['id'])
+            && !isset($context[self::OBJECT_TO_POPULATE])
+            && (null === $this->resourceMetadataFactory || !$this->resourceMetadataFactory->create($class)->getAttribute('messenger', false))
```
soyuka:
```
$this->resourceMetadataFactory->create($class)
```
may be null

src/JsonLd/Serializer/ItemNormalizer.php
karser: `(true || x)` will never evaluate x

src/JsonApi/Serializer/ItemNormalizer.php
karser: `(true || x)` will never evaluate x

soyuka: Not `resourceMetadataFactory` I mean the result of calling `create` can be null https://3v4l.org/18tE1 ?

karser: Sorry, got it

karser: Fixed it

soyuka:
can you rebase on 2.5 ?

karser:
I rebased against 2.5

teohhanhui:
I'm sorry, but this doesn't seem like the right fix at all.

teohhanhui:
If there is a bug, we should address the core issue. This workaround is wrong.

soyuka:
IMO the `isset($data['data']['id'])` check is already a wrong assumption, if we can avoid messenger to go through this check it simplifies some cases. I already hit that bug but can't remember what I did to fix it.

PR #3451, opened 2020-03-13
Fix error pagination

webant-ltd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a 
| License       | MIT
| Doc PR        | n/a

I integrate a schemathesis tool for automatically testing api based on the openapi standard

When setting a large pag value, for example 9223372036854775807, a 500 error is caused

"Return value of ApiPlatform \\ Core \\ DataProvider \\ Pagination :: getOffset () must be of the type integer, float returned"

Added a check for the type of the return value in the getOffset method on Pagination class,

 if the value is outside the scope of int64 (https://www.php.net/manual/en/language.types.integer.php) throw an error "Large page and limit values"

### DIFF
--- a/src/DataProvider/Pagination.php
+++ b/src/DataProvider/Pagination.php
@@ -90,9 +90,9 @@ final class Pagination
         if ($graphql && null !== ($last = $this->getParameterFromContext($context, 'last'))) {
             return ($offset = ($context['count'] ?? 0) - $last) < 0 ? 0 : $offset;
         }
-        
+
         $offset = ($this->getPage($context) - 1) * $limit;
-        
+
         if (!\is_int($offset)) {
             throw new InvalidArgumentException('Page parameter is too large.');
         }
--- a/src/DataProvider/Pagination.php
+++ b/src/DataProvider/Pagination.php
@@ -90,8 +90,14 @@ final class Pagination
         if ($graphql && null !== ($last = $this->getParameterFromContext($context, 'last'))) {
             return ($offset = ($context['count'] ?? 0) - $last) < 0 ? 0 : $offset;
         }
+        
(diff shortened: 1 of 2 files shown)


src/DataProvider/Pagination.php
```diff
@@ -91,6 +91,10 @@ public function getOffset(string $resourceClass = null, string $operationName =
             return ($offset = ($context['count'] ?? 0) - $last) < 0 ? 0 : $offset;
         }
 
+        if (!\is_int(($this->getPage($context) - 1) * $limit)) {
+            throw new InvalidArgumentException('Large page and limit values');
```
soyuka:
```suggestion
            throw new InvalidArgumentException('Page parameter is too large.');
```

webant-ltd:
@soyuka I add test

src/DataProvider/Pagination.php
```diff
@@ -91,6 +91,10 @@ public function getOffset(string $resourceClass = null, string $operationName =
             return ($offset = ($context['count'] ?? 0) - $last) < 0 ? 0 : $offset;
         }
 
+        if (!\is_int(($this->getPage($context) - 1) * $limit)) {
```
alanpoulain, line 94: Could you add a variable to avoid duplication?

webant-ltd: Done

alanpoulain:
A unit test too would be even better, but I think it's fine.

soyuka:
Please run php-cs-fixer

soyuka:
Thanks @webant-ltd @vladimirDaron !

PR #3454, opened 2020-03-16
feat: improves iri_only implementation

PierreThibaudeau:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | no
| License       | MIT

adds an "iri_only" parameter to the "collectionOperations" annotation. allows the display of resources in IRI list form

### DIFF
--- a/features/jsonld/iri_only.feature
+++ b/features/jsonld/iri_only.feature
@@ -1,9 +1,9 @@
-Feature: JSON-LD using IriOnly parameter
+Feature: JSON-LD using iri_only parameter
   In order to improve Vulcain support
-  As a vulcain user and as a developer
-  I should be able to only get an IRI list when I ask a resource to do so.
+  As a Vulcain user and as a developer
+  I should be able to only get an IRI list when I ask a resource.
 
-  Scenario: Retrieve Dummy's resource context with IriOnly
+  Scenario: Retrieve Dummy's resource context with iri_only
     When I send a "GET" request to "/contexts/IriOnlyDummy"
     Then the response status code should be 200
     And the response should be in JSON
@@ -22,7 +22,7 @@ Feature: JSON-LD using IriOnly parameter
       """
 
   @createSchema
-  Scenario: Retrieve Dummies with an embedded IriOnly context
+  Scenario: Retrieve Dummies with iri_only and jsonld_embed_context
     Given there are 3 iriOnlyDummies
     When I send a "GET" request to "/iri_only_dummies"
     Then the response status code should be 200
--- a/features/jsonld/iri_only.feature
+++ b/features/jsonld/iri_only.feature
@@ -1,0 +1,50 @@
+Feature: JSON-LD using IriOnly parameter
+  In order to improve Vulcain support
+  As a vulcain user and as a developer
+  I should be able to only get an IRI list when I ask a resource to do so.
+
+  Scenario: Retrieve Dummy's resource context with IriOnly
+    When I send a "GET" request to "/contexts/IriOnlyDummy"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+      """
+      {
+          "@context": {
+              "@vocab": "http://example.com/docs.jsonld#",
+              "hydra": "http://www.w3.org/ns/hydra/core#",
+              "hydra:member": {
+                  "@type": "@id"
+              }
+          }
+      }
+      """
+
+  @createSchema
+  Scenario: Retrieve Dummies with an embedded IriOnly context
+    Given there are 3 iriOnlyDummies
+    When I send a "GET" request to "/iri_only_dummies"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+      """
+      {
+          "@context": {
+              "@vocab": "http://example.com/docs.jsonld#",
+              "hydra": "http://www.w3.org/ns/hydra/core#",
+              "hydra:member": {
+                  "@type": "@id"
+              }
+          },
+          "@id": "/iri_only_dummies",
+          "@type": "hydra:Collection",
+          "hydra:member": [
+              "/iri_only_dummies/1",
+              "/iri_only_dummies/2",
+              "/iri_only_dummies/3"
+          ],
+          "hydra:totalItems": 3
+      }
+      """
--- a/src/Hydra/Serializer/CollectionNormalizer.php
+++ b/src/Hydra/Serializer/CollectionNormalizer.php
@@ -39,21 +39,16 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
     use NormalizerAwareTrait;
 
     public const FORMAT = 'jsonld';
-    public const IRI_ONLY = 'iri_only';
 
     private $contextBuilder;
     private $resourceClassResolver;
     private $iriConverter;
-    private $defaultContext = [
-        self::IRI_ONLY => false,
-    ];
 
-    public function __construct(ContextBuilderInterface $contextBuilder, ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, array $defaultContext = [])
+    public function __construct(ContextBuilderInterface $contextBuilder, ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter)
     {
         $this->contextBuilder = $contextBuilder;
         $this->resourceClassResolver = $resourceClassResolver;
         $this->iriConverter = $iriConverter;
-        $this->defaultContext = array_merge($this->defaultContext, $defaultContext);
     }
 
     /**
@@ -86,11 +81,10 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
         }
 
         $data['@type'] = 'hydra:Collection';
-
         $data['hydra:member'] = [];
-        $iriOnly = $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY];
+        $iriOnly = $context['iri_only'] ?? false;
         foreach ($object as $obj) {
-            $data['hydra:member'][] = $iriOnly ? ['@id' => $this->iriConverter->getIriFromItem($obj)] : $this->normalizer->normalize($obj, $format, $context);
+            $data['hydra:member'][] = $iriOnly ? $this->iriConverter->getIriFromItem($obj) : $this->normalizer->normalize($obj, $format, $context);
         }
 
         if ($object instanceof PaginatorInterface) {
--- a/src/Hydra/Serializer/CollectionNormalizer.php
+++ b/src/Hydra/Serializer/CollectionNormalizer.php
@@ -39,16 +39,21 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
     use NormalizerAwareTrait;
 
     public const FORMAT = 'jsonld';
+    public const IRI_ONLY = 'iri_only';
 
     private $contextBuilder;
     private $resourceClassResolver;
     private $iriConverter;
+    private $defaultContext = [
+        self::IRI_ONLY => false,
+    ];
 
-    public function __construct(ContextBuilderInterface $contextBuilder, ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter)
+    public function __construct(ContextBuilderInterface $contextBuilder, ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, array $defaultContext = [])
     {
         $this->contextBuilder = $contextBuilder;
         $this->resourceClassResolver = $resourceClassResolver;
         $this->iriConverter = $iriConverter;
+        $this->defaultContext = array_merge($this->defaultContext, $defaultContext);
     }
 
     /**
@@ -82,7 +87,7 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
 
         $data['@type'] = 'hydra:Collection';
         $data['hydra:member'] = [];
-        $iriOnly = $context['iri_only'] ?? false;
+        $iriOnly = $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY];
         foreach ($object as $obj) {
             $data['hydra:member'][] = $iriOnly ? $this->iriConverter->getIriFromItem($obj) : $this->normalizer->normalize($obj, $format, $context);
         }
--- a/src/JsonLd/ContextBuilder.php
+++ b/src/JsonLd/ContextBuilder.php
@@ -95,6 +95,13 @@ final class ContextBuilder implements AnonymousContextBuilderInterface
             return [];
         }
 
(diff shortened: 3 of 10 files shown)


src/Hydra/Serializer/CollectionNormalizer.php
```diff
@@ -86,11 +86,19 @@ public function normalize($object, $format = null, array $context = [])
...
         $data['@type'] = 'hydra:Collection';
-
         $data['hydra:member'] = [];
-        $iriOnly = $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY];
-        foreach ($object as $obj) {
-            $data['hydra:member'][] = $iriOnly ? ['@id' => $this->iriConverter->getIriFromItem($obj)] : $this->normalizer->normalize($obj, $format, $context);
+
+        // IriOnly's adaptations
+        if (true === $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY]) {
+            $data['@context']['hydra:member'] = ['@type' => '@id'];
```
soyuka: this should be the responsability of the JsonLd Context Builder

src/Serializer/SerializerContextBuilder.php
```diff
@@ -72,11 +73,16 @@ public function createFromRequest(Request $request, bool $normalization, array $
...
         $context['resource_class'] = $attributes['resource_class'];
+        $context[CollectionNormalizer::IRI_ONLY] = $resourceMetadata->getTypedOperationAttribute($operationType, $attributes[$operationKey], CollectionNormalizer::IRI_ONLY, null, true);
         $context['input'] = $resourceMetadata->getTypedOperationAttribute($operationType, $attributes[$operationKey], 'input', null, true);
         $context['output'] = $resourceMetadata->getTypedOperationAttribute($operationType, $attributes[$operationKey], 'output', null, true);
         $context['request_uri'] = $request->getRequestUri();
         $context['uri'] = $request->getUri();
 
+        if (true === $context[CollectionNormalizer::IRI_ONLY]) {
+            $context['jsonld_embed_context'] = true;
+        }
```
soyuka: we should'nt do this, it should be the user's choice to embed or not.

features/jsonld/iri_only.feature
```diff
@@ -0,0 +1,50 @@
...
+              "hydra": "http://www.w3.org/ns/hydra/core#",
+              "hydra:member": {
+                  "@type": "@id"
+              }
+          }
+      }
+      """
+
+  @createSchema
+  Scenario: Retrieve Dummy with an IriOnly context
```
soyuka: > Retrieve Dummy with an embed IriOnly context

soyuka:
```suggestion
  Scenario: Retrieve Dummy with an embeded IriOnly context
```

src/Hydra/Serializer/CollectionNormalizer.php
```diff
@@ -89,7 +89,7 @@ public function normalize($object, $format = null, array $context = [])
         $data['hydra:member'] = [];
 
         // IriOnly's adaptations
-        if (true === $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY]) {
+        if (isset($context[self::IRI_ONLY]) && true === $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY]) {
```
soyuka: `isset` is useless here as the [null coalesce operator `??` ](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce) does it already. Maybe we can use:
```
true === $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY] ?? false
```

tests/JsonLd/ContextBuilderTest.php
```diff
@@ -67,6 +67,27 @@ public function testResourceContext()
...
+
+        $expected = [
+            '@vocab' => '#',
+            'hydra' => 'http://www.w3.org/ns/hydra/core#',
+            'hydra:member' => [
+                '@type' => '@id'
+            ],
+        ];
+
+        dump($contextBuilder->getResourceContext($this->entityClass));
```
soyuka: remove this

PierreThibaudeau: already did in https://github.com/api-platform/core/pull/3454/commits/280ae045dc424b81750bb5e0aab346f807fb253b#diff-acf61da8ac68c0eec26247b10f067e27

src/Hydra/Serializer/CollectionNormalizer.php
PierreThibaudeau: I can't pass the unit tests if I do that. I got an "Undefined index: iri_only".

teohhanhui:
Sorry, I don't understand what this PR is trying to achieve. It's already possible to set `iri_only` in the normalization context

PierreThibaudeau:
The purpose is to get this: http://tinyurl.com/rzzdjm9 instead of that: http://tinyurl.com/rrg34e3. Kevin Dunglas asked me to do that to enhance support of mercure.

soyuka:
This just improves the iri_only feature by:
- returning an array of iris (`/book/1` instead of `{"@id": "/book/1"}`)
- fixing the context accordingly

This work is almost done and we worked on it with @dunglas. It'll improve Vulcain support in which the selector would become `/hydra:member/*` instead of `/hydra:member/*/@id/*` currently.

teohhanhui:
Hmm... I see. I think we should reuse the same `iri_only` option in the normalization context.

soyuka:
iri_only is only available on master

teohhanhui:
Okay, that makes it easier then. We can fix its behaviour without introducing this `iriOnly` attribute (which is badly named and doesn't belong at this level).

But if the idea is to replace it with something better, then I'd say it belongs at the collection operation level only, like the OP in this PR says, not per the current implementation of this PR.

PierreThibaudeau:
It is both now, like a generic attribute.

teohhanhui:
Taking a step back, I think there's more value to be had with an attribute such as `iriOnly` if it means "always normalize **all** relations as IRI, never embed", and not just for collection items. And the naming would make more sense too...

teohhanhui:
I'm taking another step back. It seems to me that `iri_only` would indeed make a lot of sense as a serialization "mode" to instruct the normalizer to never embed (resource) relations.

Ideally, an API resource or its operations should never need to be marked as such, but it should be a mode that can be triggered by some mechanism which would cause this `iri_only` option be passed in while calling `normalize`.

So I feel we're going more in the wrong direction with this PR and should rethink this.

PierreThibaudeau:
I'm not sure I understand what you mean (given my level on API P. Atm).
But you can use this feature without the embedded context shown in the jsonld playground. 
I added a ContextAction condition to allow different context link in case the resource's context is not the same as the operation's context. Is it what you meant ?

dunglas:
I don't understand what you mean @teohhanhui. This is not directly related to Vulcain. Having an option to **not** embed resources (which is what a pure REST API should prefer) doesn't have to be dynamic. Actually, in a perfect world, supporting resources embedding shouldn't even be supported as it's just a performance hack on top of REST.

I'm not against having a dynamic way to trigger this option (like for pagination), but it can be done in the future. This feature looks good enough as is for me.

src/Hydra/Serializer/CollectionNormalizer.php
```diff
@@ -86,11 +86,16 @@ public function normalize($object, $format = null, array $context = [])
         }
 
         $data['@type'] = 'hydra:Collection';
-
         $data['hydra:member'] = [];
-        $iriOnly = $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY];
-        foreach ($object as $obj) {
-            $data['hydra:member'][] = $iriOnly ? ['@id' => $this->iriConverter->getIriFromItem($obj)] : $this->normalizer->normalize($obj, $format, $context);
+
+        if (isset($context[self::IRI_ONLY]) && true === $context[self::IRI_ONLY] ?? $this->defaultContext[self::IRI_ONLY]) {
```
dunglas:
```suggestion
$data['hydra:member'][] = $iriOnly ? [$this->iriConverter->getIriFromItem($obj) : $this->normalizer->normalize($obj, $format, $context);
```

This should be enough.

src/JsonLd/Action/ContextAction.php
```diff
@@ -48,22 +49,46 @@ public function __construct(ContextBuilderInterface $contextBuilder, ResourceNam
      */
     public function __invoke(string $shortName): array
     {
-        if ('Entrypoint' === $shortName) {
+        $aPath = 1 === \count(explode('/', $shortName))
```
dunglas: All this logic looks too complicated. IMO we should just use the router to generate the appropriate URL somewhere. I'll take a deeper look ASAP.

src/JsonLd/ContextBuilder.php
```diff
@@ -14,6 +14,7 @@
 namespace ApiPlatform\Core\JsonLd;
 
 use ApiPlatform\Core\Api\UrlGeneratorInterface;
+use ApiPlatform\Core\Hydra\Serializer\CollectionNormalizer;
```
dunglas: The normalizer can depend of this class, but not the other way. You should move the main constant here IMHO.

dunglas: Actually, we usually don't use constants for attributes, so just use the raw string.

src/JsonLd/ContextBuilderInterface.php
```diff
@@ -47,8 +47,18 @@ public function getEntrypointContext(int $referenceType = UrlGeneratorInterface:
      */
     public function getResourceContext(string $resourceClass, int $referenceType = UrlGeneratorInterface::ABS_PATH): array;
 
+    /**
+     * Builds the JSON-LD context for the given operation's type & name.
+     */
+    public function getOperationContext(string $resourceClass, string $operationType, string $operationName, int $referenceType = UrlGeneratorInterface::ABS_PATH): array;
```
dunglas: This is a BC break. It should be done in another interface extending this one.

src/JsonLd/Serializer/JsonLdContextTrait.php
```diff
@@ -13,6 +13,7 @@
 
 namespace ApiPlatform\Core\JsonLd\Serializer;
 
+use ApiPlatform\Core\Hydra\Serializer\CollectionNormalizer;
```
dunglas: Same here.

src/JsonLd/Serializer/JsonLdContextTrait.php
```diff
@@ -33,17 +34,18 @@ private function addJsonLdContext(ContextBuilderInterface $contextBuilder, strin
...
         }
-
         $context['jsonld_has_context'] = true;
 
-        if (isset($context['jsonld_embed_context'])) {
-            $data['@context'] = $contextBuilder->getResourceContext($resourceClass);
-
-            return $data;
+        if (isset($context[CollectionNormalizer::IRI_ONLY]) && isset($context['operation_type']) && isset($context[$context['operation_type'].'_operation_name'])) {
+            $data['@context'] = isset($context['jsonld_embed_context']) && true === $context['jsonld_embed_context']
```
dunglas: Use `??` to simplify the code.

src/Hydra/Serializer/CollectionNormalizer.php
PierreThibaudeau: I thought it was faster to check once before the foreach rather than inside each loop.

dunglas:
This can be simplified a lot. Generating different contexts per operation is useless: https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22%40vocab%22%3A%22http%3A%2F%2Fschema.org%2F%22%2C%22hydra%3Amember%22%3A%7B%22%40type%22%3A%22%40id%22%7D%7D%2C%22%40type%22%3A%22Person%22%2C%22name%22%3A%22Jane%20Doe%22%2C%22jobTitle%22%3A%22Professor%22%2C%22telephone%22%3A%22(425)%20123-4567%22%2C%22url%22%3A%22http%3A%2F%2Fwww.janedoe.com%22%7D&frame=%7B%7D&context=%7B%7D

Just add the following in all generated context if the `iri_attribute` is set, and done!

```json
    "hydra:member": {
      "@type": "@id"
    }
```

teohhanhui:
> Actually, in a perfect world, supporting resources embedding shouldn't even be supported as it's just a performance hack on top of REST.

I completely disagree on this. Embedding is an intergral part of hypermedia and is not "just a performance hack". In fact, the fact that JSON-LD (and by extension, Hydra) supports [embedding](https://www.w3.org/TR/json-ld/#dfn-embedding) is a strong argument against your point. Each resource is uniquely identified by an IRI, so it's flexibility, not a hack.

dunglas:
@teohhanhui for JSON-LD and hypermedia in general, there are a lot of use cases for compound documents. For REST APIs however, it's considered a Best Practice (especially with HTTP/2) to have granular endpoints: https://medium.com/apis-you-wont-hate/lets-stop-building-apis-around-a-network-hack-9a68f7e83dd2

Anyway we're going far away from the topic.

> Taking a step back, I think there's more value to be had with an attribute such as iriOnly if it means "always normalize all relations as IRI, never embed" (since embedding is counter-productive for Vulcain), and not just for collection items. And the naming would make more sense too (still weird, but at least it starts to make sense)...

That was exactly my goal when I introduced this attribute. I'm not very happy with the name either. I would be glad to find a better name.

Regarding the "dynamic" mechanism. We could introduce that later, has we've done for client-side pagination for instance. We already have all the infrastructure to do that if needed (but I'm not sure that we need it, most users - for simplicity, cache etc - will choose between a Vulcain-like API, or one relying on embedded documents, usually not both at the same time).

src/Hydra/Serializer/CollectionNormalizer.php
```diff
@@ -39,21 +39,16 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
     use NormalizerAwareTrait;
 
     public const FORMAT = 'jsonld';
-    public const IRI_ONLY = 'iri_only';
```
dunglas, line 42: Can you keep this constant? There are no benefits to remove it.

src/Hydra/Serializer/CollectionNormalizer.php
```diff
@@ -39,21 +39,16 @@ final class CollectionNormalizer implements NormalizerInterface, NormalizerAware
...
 
     private $contextBuilder;
     private $resourceClassResolver;
     private $iriConverter;
-    private $defaultContext = [
```
dunglas, line 47: I suggest to keep the context, but default to the value of the attribute. It will allow to make it dynamic mechanism in the future as suggested by @teohhanhui.

Richard87:
For an outside perspective, this is *almost* exactly what i need, but on a property level.

I have a Tree view, with Parent and Child beeing the same component, so modifying normalization_groups for the operation doesn't work.

I would like to add a `@ApiProperty(iri_only=true)` to my "@ORM\OneToMany` property, and only have a list of ID's returned.

Or use `@MaxDepth(0)`, or `@MaxDepth(1, iri_only=true)` or similar...

alanpoulain (approved):

alanpoulain:
Thank you @PierreThibaudeau!

PR #3455, opened 2020-03-17
[GraphQL] add support for Symfony Serializer's `@SerializedName` metadata

MartkCz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | https://github.com/api-platform/api-platform/issues/1205
| License       | MIT

inspired by https://github.com/api-platform/core/blob/master/src/JsonSchema/SchemaFactory.php#L141

but idk if it's good idea pass 3rd and 4th argument too

### DIFF
--- a/src/GraphQl/Serializer/Exception/ValidationExceptionNormalizer.php
+++ b/src/GraphQl/Serializer/Exception/ValidationExceptionNormalizer.php
@@ -17,7 +17,6 @@ use ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException;
 use GraphQL\Error\Error;
 use GraphQL\Error\FormattedError;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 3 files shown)


src/GraphQl/Serializer/Exception/ValidationExceptionNormalizer.php
```diff
@@ -61,4 +69,24 @@ public function supportsNormalization($data, $format = null): bool
     {
         return $data instanceof Error && $data->getPrevious() instanceof ValidationException;
     }
+
+    private function normalizeValidationException(ValidationException $exception): string
```
alanpoulain: IMO this code is too specific and should be reverted.

alanpoulain:
Thanks. Could you add a Behat test to make sure it works as expected?

PR #3456, opened 2020-03-18
Remove location header on 2xx responses except for 201 and 202

deguif:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | ??
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        |

Currently when making a POST request that returns a 200 status code with `write` enabled (by default), the location header is returned that make clients like Insomnia consider it as 302.

That's because the `WriteListener` adds the request attribute `_api_write_item_iri` that is then used in the `RespondListener` to set the `Location` header.

Location header for 2xx status code is only valid for 201 and 202 at my knowledge.

From [Wikipedia](https://en.wikipedia.org/wiki/HTTP_location) (not the best reference for this)
```markdown
The HTTP Location header field is returned in responses from an HTTP server under two circumstances:

1) To ask a web browser to load a different web page (URL redirection). In this circumstance, the Location header should be sent with an HTTP status code of 3xx. It is passed as part of the response by a web server when the requested URI has:
  - Moved temporarily;
  - Moved permanently; or
  - Processed a request, e.g. a POSTed form, and is providing the result of that request at a different URI
2) To provide information about the location of a newly created resource. In this circumstance, the Location header should be sent with an HTTP status code of 201 or 202.[1]
```

### DIFF
--- a/src/EventListener/RespondListener.php
+++ b/src/EventListener/RespondListener.php
@@ -83,9 +83,15 @@ final class RespondListener
             $status = $resourceMetadata->getOperationAttribute($attributes, 'status');
         }
 
+        $status = $status ?? self::METHOD_TO_CODE[$request->getMethod()] ?? Response::HTTP_OK;
+
+        if ($request->isMethod('POST') && !\in_array($status, [Response::HTTP_CREATED, Response::HTTP_ACCEPTED], true)) {
+            unset($headers['Location']);
+        }
+
         $event->setResponse(new Response(
             $controllerResult,
-            $status ?? self::METHOD_TO_CODE[$request->getMethod()] ?? Response::HTTP_OK,
+            $status,
             $headers
         ));
     }
--- a/tests/EventListener/RespondListenerTest.php
+++ b/tests/EventListener/RespondListenerTest.php
@@ -78,6 +78,7 @@ class RespondListenerTest extends TestCase
         $this->assertEquals('Accept', $response->headers->get('Vary'));
         $this->assertEquals('nosniff', $response->headers->get('X-Content-Type-Options'));
         $this->assertEquals('deny', $response->headers->get('X-Frame-Options'));
+        $this->assertNull($response->headers->get('Location'));
     }
 
     public function testCreate201Response()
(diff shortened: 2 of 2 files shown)


src/EventListener/RespondListener.php
```diff
@@ -83,9 +83,15 @@ public function onKernelView(ViewEvent $event): void
             $status = $resourceMetadata->getOperationAttribute($attributes, 'status');
         }
 
+        $status = $status ?? self::METHOD_TO_CODE[$request->getMethod()] ?? Response::HTTP_OK;
+
+        if ($request->isMethod('POST') && !\in_array($status, [Response::HTTP_CREATED, Response::HTTP_ACCEPTED], true)) {
```
deguif: Should we check for `3xx` status code too?

tests/EventListener/RespondListenerTest.php
```diff
@@ -78,6 +78,7 @@ public function testCreate200Response()
         $this->assertEquals('Accept', $response->headers->get('Vary'));
         $this->assertEquals('nosniff', $response->headers->get('X-Content-Type-Options'));
         $this->assertEquals('deny', $response->headers->get('X-Frame-Options'));
+        $this->assertNull($response->headers->get('Location'));
```
deguif: That doesn't test the current change as this a `GET` request. Should I create a `testCreate200PostResponse` method or something like that?

teohhanhui:
Looking through RFC 7231, it seems to me that the semantics of the `Location` header is only defined for 201 and 3xx responses.

But I think the correct solution would be to allow the user to suppress these response headers (`Location`, `Content-Location`, or both), instead of checking the response status code.

soyuka:
If it's in the RFC it's not an assumption. I'd like to merge this as a bug fix. I definitely agree that these added headers should be allowed to be removed by the user.

teohhanhui:
I'm not sure about that. On one hand, it might be considered a "bug" (actually not a bug, it's just undefined semantics), but on the other hand, the real bug is here:

> the location header is returned that make clients like Insomnia consider it as 302.

soyuka:
> Those clients are disrespecting the RFC.

Definitely agree on that... @deguif did you open a bug on insomnia?

fmata:
Oop's I did not see this before pushing #3497 which is a little bit different.

soyuka:
Let's follow this in  #3497

Issue #3461, opened 2020-03-24
AbstractItemNormalizer check on Traversable for support results in false negative


renaudclosset:
**API Platform version(s) affected**: v2.5.4

**Description**  
The abstract ItemNormalizer (`ApiPlatform\Core\Serializer\AbstractItemNormalizer`) checks if the passed data is an object or an instance of `\Traversable` for normalization support.

If an object is passed that implements the `\Traversable` interface, it is handled as a collection.
I presume this was added to ease the normalization of (Doctrine) Collection objects.

**How to reproduce**  
Create an object implementing the `\Traversable` interface with a custom DataTransformer. The DataTransformer will never be called.

**Possible Solution**  
Don't check on `\Traversable` but rely on the fact if this class is a Resource.

```php
public function supportsNormalization($data, $format = null)
{
  if (!\is_object($data)) {
    return false;
  }
  return $this->resourceClassResolver->isResourceClass($this->getObjectClass($data));
}
```

teohhanhui:
Basically, we treat `iterable` as a collection. It's not specific to Doctrine `Collection`.

Your proposed alternative is not feasible.

teohhanhui:
Hmm... Even if we make the change as you've suggested, you'd still run into an issue here:

https://github.com/api-platform/core/blob/48fc5bb75e90d7ccef4ba540630a85bd12016f44/src/Api/ResourceClassResolver.php#L48

May we recommend not implementing `\Traversable` in your resource class?

renaudclosset:
As far as i could tell the `getResourceClass` is never called in this scope, but it is based on the `isResourceClass` method.
The `isResourceClass` looks directly into the `ResourceClassResolver::$localIsResourceClassCache` list, which in turn is fed by the class' `resourceNameCollectionFactory`, which lists all classes configured as Resources.

This issue is actually the question "_Are we dealing with an Item or a Collection?_"
If we base this on the implementation `\Traversable` in a class, this could results if false negatives. But looking a the Resource list that was explicitly defined by the user seems to be a more robust approach.

If some Resource are automatically added, without explicit user configuration, we could fallback to the `\Traversable` implementation check. no?

renaudclosset:
Hey,

I created a pull request that should deal with this issue without any BC breaks.
Could that be reviewed/discussed?

Thanks

teohhanhui:
[...]

PR #3463, opened 2020-03-24
…Do not treat classes implementing \Traversable as collection

renaudclosset:
Do not treat classes implementing `\Traversable` as collection if defined as resource

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #3461
| License       | MIT
| Doc PR        | -

Classes implementing `\Traversable` were always interpreted as Collection by the `ResourceClassResolver `service, based on the fact they implemented the `\Traversable` interface.

Classes that implement this interface will now only be treated as a Collection if they are not explicitly defined as a Resource in the configuration.


soyuka:
cs-fixes unrelated thanks @renaudclosset

PR #3466, opened 2020-03-27
SubresourceOperations customisation DX enhancement

### LINKED ISSUE #3458: Subresource customisation DX is poor and confusing.
**Description**  
The need arose in my project to start using sub resources.
Here's where I had trouble.
I needed to:
1. specify a custom path for the sub resource
1. specify custom serialization groups in the normalization context.

Great, I thought, both of these features are supported: [1](https://api-platform.com/docs/core/subresources/#using-custom-paths), [2](https://api-platform.com/docs/core/subresources/#using-serialization-groups)

However, after failing to get these working together for about 30 minutes, I realized I had read the docs incorrectly.
If we take the example from the docs, this is what I was trying:

```php
<?php
// api/src/Entity/Question.php

/**
 * ...
 * @ApiResource(
 *      subresourceOperations={
 *          "api_questions_answer_get_subresource"={
 *              "method"="GET",
 *              "path"="/questions/{id}/all-answers",
  *             "normalization_context"={"groups"={"foobar"}}
 *          },
 *      },
 * )
 */
class Question
{
    /**
     * @ORM\OneToOne(targetEntity="Answer", inversedBy="question")
     * @ORM\JoinColumn(referencedColumnName="id", unique=true)
     * @ApiSubresource
     */
    public $answer;
}
```

Looks fine right?
Well actually it's wrong.

Two issues,
1. The `subresourceOperations` name is wrong. [PR Raised.](https://github.com/api-platform/docs/pull/1071) ~~`~~api_questions_`~~`answer_get_subresource`. 
1. To to override the normalization_context, the `subresourceOperations` options must actually be on the `Answer` Entity, not `Question`. This IS actually in the docs, but i missed it because it's not really obvious or explained.

So the correct config is this:

```php
<?php
// api/src/Entity/Question.php

/**
 * ...
 * @ApiResource(
 *      subresourceOperations={
 *          "answer_get_subresource"={
 *              "method"="GET",
 *              "path"="/questions/{id}/all-answers"
 *          },
 *      },
 * )
 */
class Question
{
    /**
     * @ORM\OneToOne(targetEntity="Answer", inversedBy="question")
     * @ORM\JoinColumn(referencedColumnName="id", unique=true)
     * @ApiSubresource
     */
    public $answer;
}
```

```php
<?php
// api/src/Entity/Answer.php

namespace App\Entity;

/**
 * @ApiResource(subresourceOperations={
 *     "api_questions_answer_get_subresource"={
 *         "normalization_context"={"groups"={"foobar"}}
 *     }
 * })
 */
class Answer
{
}
```

This works, but here is the kicker:

For this to work the operation name must be
1. On `Question`: `answer_get_subresource`. This is the `operation_name`.
2. On `Answer`: `api_questions_answer_get_subresource`. This is the `route_name`.

This took me far too long to work out and seems crazy from a DX point of view.
I'm not sure if this is a bug or a design constraint. Can it be improved?

The optimum for DX in my opinion would be if the the config was done as my first example, which doesn't work, with it all sitting in the `Question` Entity.

Thanks

PierreThibaudeau:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3458
| License       | MIT
| Doc PR        | api-platform/docs#1084

Changes the way to define serialization groups for subresources.
Changes the nomenclature to override subresource operations. 

Example:

| OLD             | NEW
| ------------- | ---
| api_questions_answers_get_subresource    |answers_get
| api_questions_answers_question_get_subresource |answers_question_get


soyuka:
TODO:
- [ ] deprecate old operation names for now (they should still work)
- [ ] add behat test for serialization groups
- [ ] ask @bendavies to try this :smile:

GregoireHebert:
IMO this should still be prefixed by the associated resource.
I did not run any test, but AFAIK If you have multiple resources having the same resource as subresource, can't you end-up with a conflict?

vincentchalamon:
Is this PR still relevant, considering subresources are deprecated on 2.7 and removed on 3.0?

PR #3468, opened 2020-03-29
[GraphQL] Fix filter arguments order

### LINKED ISSUE #2808: [GraphQL] OrderFilter does not consider the order of the properties to filter on
When a resource is defined with an `OrderFilter` on multiple properties, the GRAPHQL endpoint does not keep the sorting of the properties (to filter on).

In the following example
```
/**
 * @ApiResource()
 */
class Dummy
{
   /** @var string */
   private $firstname;

   /** @var string */
   private $laststname;

   // ...
}
```
and the api-p filter service defintion :
```
dummy.order_filter:
    parent: api_platform.doctrine.orm.order_filter
    arguments:
       properties: 
          firstname: ~
          lastname: ~
```
Calling the REST endpoint `/dummies?order[lastname]=ASC&order[firstname]=ASC` returns the  expected results ; the `OrderFilter` is called and the `$context[filters][order]` argument is populated with the properties to filter on in the expected order (1st the `lastname` and 2nde the `firstname`)
```
context => [ 
  filters => [
    order => [
      lastname => ASC,
      firstname => ASC
    ]
  ]
]
```
So the SQL query is build keeping the properties sorting : 
`SELECT * FROM dummies ORDER BY lastname ASC, firstname ASC`

The issue is on the GRAPHQL endpoint that keeps the default order (defined in the service definition). 
Calling the syntax `dummies(order: {"lastname": "ASC", "firstname": "ASC"}) { ... }`, the SQL query is build keeping the default order :
`SELECT * FROM dummies ORDER BY firstname ASC, lastname ASC`
In this case, the `OrderFilter` is called and the `$context[filters][order]` argument is : 
```
context => [ 
  filters => [
    order => [
      firstname => ASC
      lastname => ASC,
    ]
  ]
]
```

ArnoudThibaut:
The InputObjectType doesn't preserve the order of the arguments passed to it and it's the order declared in the service declaration that is used. This can lead to bug when the order of the arguments matter, like with the OrderFilter for example.

The solution is to change the type of the filter argument from InputObjectType to a list of InputObjectType because the order of a list is preserved.

The "old" syntax `order: {foo: asc, bar: desc}` is still "working" (but it's buggy) and the following syntax should be used instead `order: [{foo: 'asc'}, {bar: 'desc'}]`

In the first commit there is a falling behat test that use the current syntax to order a collection across 2 fields, and the order is first apply one the field `name` before the field `description`.

| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | not sur,the build-in filters are still working but I don't know if it can break custom filters
| Deprecations? | not yet, but I think we should find a way to depreciate the "old" syntax on the build-in filter that are buggy because of the order problem
| Tickets       | fixes #2808
| License       | MIT
| Doc PR        | TODO, as the old syntax is still usabled maybe we should add doc on how to use the filters with graphql

Ps: the unit tests are failing but I want to be sure on how preserve the right order and if we should apply the changes for all filter or just for the build-in one because I'm not sure if it can broke something :/

### DIFF
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -30,7 +30,7 @@ Feature: Collections filtering
     When I send the following GraphQL request:
     """
     {
-      dummies(exists: {relatedDummy: true}) {
+      dummies(exists: [{relatedDummy: true}]) {
         edges {
           node {
             id
@@ -52,7 +52,7 @@ Feature: Collections filtering
     When I send the following GraphQL request:
     """
     {
-      dummies(dummyDate: {after: "2015-04-02"}) {
+      dummies(dummyDate: [{after: "2015-04-02"}]) {
         edges {
           node {
             id
@@ -204,7 +204,7 @@ Feature: Collections filtering
     When I send the following GraphQL request:
     """
     {
-      dummies(order: {relatedDummy__name: "DESC"}) {
+      dummies(order: [{relatedDummy__name: "DESC"}]) {
         edges {
           node {
             name
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -228,7 +228,7 @@ Feature: Collections filtering
     When I send the following GraphQL request:
     """
     {
-      dummies(order: {description: "ASC", name: "ASC"}) {
+      dummies(order: [{description: "ASC"}, {name: "ASC"}]) {
         edges {
           node {
             id
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -223,6 +223,31 @@ Feature: Collections filtering
     And the JSON node "data.dummies.edges[1].node.name" should be equal to "Dummy #1"
 
   @createSchema
+  Scenario: Retrieve a collection ordered correctly given the order of the argument
+    Given there are dummies with similar properties
+    When I send the following GraphQL request:
+    """
+    {
+      dummies(order: {description: "ASC", name: "ASC"}) {
+        edges {
+          node {
+            id
+            name
+            description
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.dummies.edges[0].node.name" should be equal to "baz"
+    And the JSON node "data.dummies.edges[0].node.description" should be equal to "bar"
+    And the JSON node "data.dummies.edges[1].node.name" should be equal to "foo"
+    And the JSON node "data.dummies.edges[1].node.description" should be equal to "bar"
+
+
+  @createSchema
   Scenario: Retrieve a collection filtered using the related search filter with two values and exact strategy
     Given there are 3 dummy objects with relatedDummy
     When  I send the following GraphQL request:
--- a/features/graphql/filters.feature
+++ b/features/graphql/filters.feature
@@ -246,7 +246,6 @@ Feature: Collections filtering
     And the JSON node "data.dummies.edges[1].node.name" should be equal to "foo"
     And the JSON node "data.dummies.edges[1].node.description" should be equal to "bar"
 
-
   @createSchema
   Scenario: Retrieve a collection filtered using the related search filter with two values and exact strategy
     Given there are 3 dummy objects with relatedDummy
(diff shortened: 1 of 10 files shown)


src/GraphQl/Resolver/Stage/ReadStage.php
```diff
@@ -172,4 +177,19 @@ private function getSubresource(string $rootClass, array $rootResolvedFields, st
             'collection' => true,
         ], $operationName);
     }
+
+    private function isNumericIndexArrayOfArray(array $array): bool
```
alanpoulain: I think it should be elsewhere, like in `Util/ArrayTrait` for instance.
Also, it should probably be split into two methods like `isSequentialArray` and `containsOnly` (maybe a more general method by using `gettype`).

features/graphql/filters.feature
```diff
@@ -222,6 +222,31 @@ Feature: Collections filtering
...
+    }
+    """
+    Then the response status code should be 200
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.dummies.edges[0].node.name" should be equal to "baz"
+    And the JSON node "data.dummies.edges[0].node.description" should be equal to "bar"
+    And the JSON node "data.dummies.edges[1].node.name" should be equal to "foo"
+    And the JSON node "data.dummies.edges[1].node.description" should be equal to "bar"
+
+
```
alanpoulain: One line to remove (sorry :stuck_out_tongue:).

src/GraphQl/Resolver/Stage/ReadStage.php
```diff
@@ -140,9 +140,14 @@ private function getNormalizedFilters(array $args): array
 
         foreach ($filters as $name => $value) {
             if (\is_array($value)) {
-                if (strpos($name, '_list')) {
+                if (\is_string($name) && strpos($name, '_list')) {
                     $name = substr($name, 0, \strlen($name) - \strlen('_list'));
                 }
+
+                // If it's just a simple array of string values, for example when using the _list syntax we don't want to decompose the array because it's will return multiple string and not multipe array that we can merge
```
alanpoulain: I think you should invert the sense of this sentence to tell the developer when this `if` is applied.
Like:
If the value contains arrays, we need to merge them for the filters to understand this syntax, proper to GraphQL to preserve the order of the arguments.

alanpoulain:
IMHO I think all the filters should have the same syntax if they are used in the same way (like the `exists` filter).

src/GraphQl/Resolver/Stage/ReadStage.php
```diff
@@ -131,9 +133,14 @@ private function getNormalizedFilters(array $args): array
 
         foreach ($filters as $name => $value) {
             if (\is_array($value)) {
-                if (strpos($name, '_list')) {
+                if (\is_string($name) && strpos($name, '_list')) {
```
alanpoulain, line 136: Why this check?

src/Util/ArrayTrait.php
```diff
@@ -0,0 +1,42 @@
...
+
+trait ArrayTrait
+{
+    public function isNumericIndexArrayOfArray(array $array): bool
+    {
+        if (!$this->isSequentialArray($array)) {
+            return false;
+        }
+
+        return $this->arrayContainsOnly($array, \gettype([]));
```
alanpoulain, line 24: Why not:
```suggestion
        return $this->arrayContainsOnly($array, 'array');
```

src/Util/ArrayTrait.php
```diff
@@ -0,0 +1,42 @@
...
+        if ([] === $array) {
+            return false;
+        }
+
+        return array_keys($array) === range(0, \count($array) - 1);
+    }
+
+    public function arrayContainsOnly(array $array, string $type): bool
+    {
+        return $array === array_filter($array, static function ($item) use ($type) {
```
alanpoulain, line 38:
```suggestion
        return $array === array_filter($array, static function ($item) use ($type): bool {
```

src/Util/ArrayTrait.php
```diff
@@ -0,0 +1,42 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Util;
+
+trait ArrayTrait
+{
+    public function isNumericIndexArrayOfArray(array $array): bool
```
alanpoulain, line 18: To be coherent:
```suggestion
    public function isSequentialArrayOfArrays(array $array): bool
```

alanpoulain:
Thank you @ArnoudThibaut!

PR #3470, opened 2020-03-29
Use SchemaBuilderInterface in GraphQlExportCommand

xterr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

The reason for this change is to allow decorating the `SchemaBuilder`.
Currently, `SchemaBuilder` cannot be decorated because in the constructor of `GraphQlExportCommand` the interface is not used.


alanpoulain:
Thank you @xterr.

PR #3471, opened 2020-03-30
Add default validation groups

pierre-H:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | no
| License       | MIT
| Doc PR        | not yet

Add an option for default validation groups. It's useful when a bundle has it's own entities, with it's own validations rules with groups and you want to easily take them into account.

### DIFF
--- a/src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
+++ b/src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
@@ -82,7 +82,7 @@ final class ValidateListener
         } elseif (\is_callable($validationGroups)) {
             $validationGroups = $validationGroups($data);
         }
-        
+
         if ($this->container instanceof SymfonyContainerInterface && empty($validationGroups)) {
             $validationGroups = $this->container->getParameter('api_platform.validator.default_groups');
         }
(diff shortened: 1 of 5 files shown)


src/Bridge/Symfony/Validator/EventListener/ValidateListener.php
```diff
@@ -82,6 +83,10 @@ public function onKernelView(ViewEvent $event): void
             $validationGroups = $validationGroups($data);
         }
 
+        if ($this->container instanceof SymfonyContainerInterface && empty($validationGroups)) {
```
norkunas: But this won't work anymore in 3.0 :man_facepalming: :smile: Container will be replaced with a locator.

norkunas: You are adding a feature to deprecated class :guardsman:

pierre-H: Oh you're right ! I didn't see that. I will fix that.

dunglas:
The master version now has a new feature to define default for any exiting attributes (including validation groups). You can now achieve what you want like this:

```yaml
api_platform:
    defaults:
        validation_groups: [foo]
```

Thanks for your contribution! Closing as already supported.

PR #3476, opened 2020-04-03
fix (serializer): returns a 400 HTTP status code instead of 500 on an input type mismatch

GregoireHebert:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | ~

When a developer sets a non nullable typed property on a resource 
```php
<?php

namespace App\Entity;

/**
 * @ApiResource
 * @ORM\Entity
 */
class Greeting
{
    /**
     * @var int The entity Id
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @var string A nice person
     *
     * @ORM\Column
     * @Assert\NotBlank
     */
    public $name = '';

    public function getId(): int
    {
        return $this->id;
    }
}
```

And one sends a post request with a null value
```cli
curl --location --request POST 'https://localhost:8443/greetings' \
--header 'Content-Type: application/json' \
--data-raw '{
	"name": null
}'
```

The response received comes with a 500 (Internal Server Error) http status code.
```
{
    "@type": "hydra:Error",
    "hydra:description": "The type of the \"name\" attribute must be \"string\", \"NULL\" given.",
    "trace": [ ...
    ]
}
```

But it's not a server error, the configuration is intended.
If the developer wanted to allow null to be used, one would have put the `name` property nullable. Which is not the case. The problem comes from the request itself.
It should have been a 400 (Bad Request) that target the api consuming developer.

And then, The 2 developers could argue that they should have let the property being nullable, but it's not a technical error, it's a design error. Something not related to API Platform itself.

:warning: *Warning*
API consumers relying on the response status to be a 500 will fail.

### DIFF
--- a/src/Exception/InvalidRequestException.php
+++ b/src/Exception/InvalidRequestException.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Exception;
+
+use Symfony\Component\HttpFoundation\Exception\RequestExceptionInterface;
+
+/**
+ * Invalid Request exception.
+ *
+ * @author Grégoire Hébert <contact@gheb.dev>
+ */
+class InvalidRequestException extends InvalidArgumentException implements RequestExceptionInterface
+{
+}
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -18,6 +18,7 @@ use ApiPlatform\Core\Api\ResourceClassResolverInterface;
 use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
 use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
 use ApiPlatform\Core\Exception\InvalidArgumentException;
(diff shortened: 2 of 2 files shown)


GregoireHebert:
Scrutinizer failure seems not to be related to this PR

soyuka:
Please add a behat test

GregoireHebert:
After additional checking when writing the behat test,
I realised that it's already the default behaviour.
But the cases I was encountering did not take in account this doc page:
https://api-platform.com/docs/core/errors/#converting-php-exceptions-to-http-errors

`# The 4 following handlers are registered by default, keep those lines to prevent unexpected side effects`

in 
```yaml
api_platform:
    # ...
    exception_to_status:
        # The 4 following handlers are registered by default, keep those lines to prevent unexpected side effects
        Symfony\Component\Serializer\Exception\ExceptionInterface: 400 # Use a raw status code (recommended)
        ApiPlatform\Core\Exception\InvalidArgumentException: !php/const Symfony\Component\HttpFoundation\Response::HTTP_BAD_REQUEST
        ApiPlatform\Core\Exception\FilterValidationException: 400
        Doctrine\ORM\OptimisticLockException: 409

        # Custom mapping
        App\Exception\ProductNotFoundException: 404 # Here is the handler for our custom exception
```

see https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L474

PR #3478, opened 2020-04-03
Bug Fix: DataTransformer returning same class as original should include JSONLD Data

### LINKED ISSUE #3474: Using Data Transformer which returns same class == missing extended data in JSONLD
API Platform version(s) affected: 2.5.4

Description
I want to trigger some DataTransformers on resources but instead of using a different class, I simply want to add some data to some of the properties on the existing class's properties. So to trigger the DataTransformer I set the output class of the resource to it's own class name again. This then triggers the Data Transformers to be detected and executed.

The issue is that when this happens the additional data expected in JSONLD vs JSON is missing. The @id, @context etc. This happens whether there is a transformer that is supported or not.

How to reproduce
Set an output class on a API Platform resource and request it as jsonld

Possible Solution
Unfortunately I haven't been able to trace where this is happening yet but I will update tomorrow once I continue to investigate why this happens.

Additional Context
I think this is enough information but I'll be happy to elaborate on any points here or admit my mistakes if I am going about this the wrong way. Thank you in advance.

silverbackdan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3474
| License       | MIT

**I'm sorry but I'm very bad with tests so have not written one and do not know where it'd be best placed.**

In `ApiPlatform\Core\DataTransformer\DataTransformerInterface` there is a comment that states that DTOs should be allowed to return the same original object if no transformation is done. This resulted in missing LD data. (No @id, @context etc.).
```php
     /**
     * Transforms the given object to something else, usually another object.
     * This must return the original object if no transformation has been done.
     *
     * @param object $object
     *
     * @return object
     */
    public function transform($object, string $to, array $context = []);
```
This update checks if the output class is the same as the original, and if so populated the extended metadata in the JsonLd\ItemNormalizer as it would not be added using the JsonLd\ObjectNormalizer


silverbackdan:
Sorry will re-do patch done on the 2.5 branch.

PR #3479, opened 2020-04-03
Bug Fix: DataTransformer returning same class as original should include JSONLD Data

### LINKED ISSUE #3474: Using Data Transformer which returns same class == missing extended data in JSONLD

silverbackdan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3474
| License       | MIT

In `ApiPlatform\Core\DataTransformer\DataTransformerInterface` there is a comment that states that DTOs should be allowed to return the same original object if no transformation is done. This resulted in missing LD data. (No @id, @context etc.).
```php
     /**
     * Transforms the given object to something else, usually another object.
     * This must return the original object if no transformation has been done.
     *
     * @param object $object
     *
     * @return object
     */
    public function transform($object, string $to, array $context = []);
```
This update checks if the output class is the same as the original, and if so populated the extended metadata in the JsonLd\ItemNormalizer as it would not be added using the JsonLd\ObjectNormalizer


silverbackdan:
Scrutinizer does not appear to be related to this PR.

Behat test added.

If this shouldn't be allowed perhaps it's worth throwing some kind of exception, but it does seem a pretty useful way of using this feature for me and it doesn't seem to cost much to allow this feature to be used in this way.

silverbackdan:
I've just noticed this won't fix if the DTO returns the original object but the output class was set to something else which is what the comment in the interface describes should be allowed. That's because the checks to see if the object is different happens in the parent class to the JsonLd normalizer... Not sure how to address that though. Comments welcome.

silverbackdan:
The latest commit handles both scenarios by using a context key. As the original resource will use the item normalizer that supports API Platform resources, this context key allows us to prevent an infinite loop.

silverbackdan:
All test failures appear to be related to the 2.5 branch and not this PR.

soyuka:
@teohhanhui  do you have thoughts on this? I think it's good, just not fond of the context to prevent recursion.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -637,9 +645,8 @@ protected function getDataTransformer($data, string $to, array $context = []): ?
      * For a given resource, it returns an output representation if any
      * If not, the resource is returned.
      */
-    protected function transformOutput($object, array $context = [])
+    protected function transformOutput($object, string $outputClass, array $context = [])
```
silverbackdan, line 648: I'm thinking that to prevent BC change here perhaps it is better to add the parameter at the end with a default to null and to reinstate the call to `getOutputMethod` if it is null. I removed to prevent a duplicate call but failed to think about others who may be extending this class as it is not marked `@internal`

soyuka: :+1: indeed

tests/Fixtures/TestBundle/DataTransformer/OutputDtoSameClassTransformer.php
```diff
@@ -0,0 +1,56 @@
...
+
+use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyDtoOutputFallbackToSameClass as DummyDtoOutputFallbackToSameClassDocument;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\DummyDtoOutputSameClass as DummyDtoOutputSameClassDocument;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Dto\OutputDtoDummy;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyDtoOutputFallbackToSameClass;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\DummyDtoOutputSameClass;
+
+/**
+ * OutputDtoUnmodifiedDataTransformer.
```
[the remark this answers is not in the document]
silverbackdan: Thank you, I've applied these changes and rebased.

src/Serializer/AbstractItemNormalizer.php
silverbackdan: Thanks!

silverbackdan:
Hi, I did make these changes but I think I did something wrong with rebasing and I'm not sure how to notify that the changes requested are done ... so this is me doing that.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -51,6 +51,8 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     use ContextTrait;
     use InputOutputMetadataTrait;
 
+    public const IS_TRANSFORMED_TO_SAME_CLASS_CONTEXT_KEY = 'is_transformed_to_same_class';
```
vincentchalamon, line 54:
```suggestion
    public const IS_TRANSFORMED_TO_SAME_CLASS = 'is_transformed_to_same_class';
```

tests/Fixtures/TestBundle/Document/DummyDtoOutputFallbackToSameClass.php
```diff
@@ -0,0 +1,55 @@
...
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Document;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Dto\OutputDtoDummy;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+/**
+ * Dummy InputOutput.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
vincentchalamon, line 23:
```suggestion
 * @author Daniel West <daniel@silverback.is>
```

tests/Fixtures/TestBundle/Entity/DummyDtoOutputFallbackToSameClass.php
```diff
@@ -0,0 +1,57 @@
...
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Dto\OutputDtoDummy;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Dummy InputOutput.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
vincentchalamon, line 23:
```suggestion
 * @author Daniel West <daniel@silverback.is>
```

tests/Fixtures/TestBundle/Document/DummyDtoOutputSameClass.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Document;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+
+/**
+ * Dummy InputOutput.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
vincentchalamon, line 22:
```suggestion
 * @author Daniel West <daniel@silverback.is>
```

tests/Fixtures/TestBundle/Entity/DummyDtoOutputSameClass.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Dummy InputOutput.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
vincentchalamon, line 22:
```suggestion
 * @author Daniel West <daniel@silverback.is>
```

src/Serializer/AbstractItemNormalizer.php
silverbackdan: Applied changes

tests/Fixtures/TestBundle/Document/DummyDtoOutputFallbackToSameClass.php
silverbackdan: Applied changes

tests/Fixtures/TestBundle/Entity/DummyDtoOutputFallbackToSameClass.php
silverbackdan: Applied changes

tests/Fixtures/TestBundle/Document/DummyDtoOutputSameClass.php
silverbackdan: Applied changes

tests/Fixtures/TestBundle/Entity/DummyDtoOutputSameClass.php
silverbackdan: Applied changes

silverbackdan:
Thank you @vincentchalamon for the review - all changes applied now. I'll wait for core team's opinion.

soyuka:
I like this, thanks @vincentchalamon @silverbackdan

vincentchalamon:
@silverbackdan Could you please update your branch from the 2.5 remote branch?

silverbackdan:
Thanks all! @vincentchalamon - I've just pressed the button to update from 2.5 remote - seems to have been able to merge without conflict.

soyuka:
thanks @silverbackdan

PR #3484, opened 2020-04-09
add type factory decorator for elao enum types

### LINKED ISSUE #3432: [SwaggerUI] Unable to overwrite context with elao enum types

**Description**  

I'm using elao/enum in my project. When generating the SwaggerUI, the Enums are generated as Definitions for Read and Write operation. Unforunatly those definition is empty ( Links to #/definitions/MediaEnum-read ) . 
I wanted to overwrite the swagger_context to help the user to understand what values they have to input.
e.g.
`* @ApiProperty( attributes={"openapi_context"={
     *     "type"="string",
     *     "enum"={"video/mp4", "video/mpeg"},
     *     "example"="video/mpeg"}})
     *`
The field type is mediaEnum in this example which is an instance of Elao\Enum\Bridge\Doctrine\DBAL\Types\AbstractEnumType.

The problem is that the overwritten openapi_context is not used for this field. It works for simple fields like the "description" field which is just  type="string"

I know there are some Issues/PRs open and merged regarding the swagger ui. Does one of them solve this problem ? If not, im more than willing to work on a solution if anyone could point me into the right direction.

WhiteRabbitDE:
| Q             | A
| ------------- | ---
| Bug fix?      |  yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3432
| License       | MIT

I was not sure about the comparison with "'Elao\Enum\EnumInterface", i would prefer EnumInterface::class but i think adding elao as a dependency is overhead. 

@teohhanhui suggested to move all custom classes like ElaoEnum and RamseyUuid into decorator. If this approach is approved i could do the same for Ramsey.

### DIFF
--- a/src/JsonSchema/Type/ElaoEnumType.php
+++ b/src/JsonSchema/Type/ElaoEnumType.php
@@ -1,0 +1,52 @@
+<?php
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\JsonSchema\Type;
+
+use ApiPlatform\Core\JsonSchema\Schema;
+use ApiPlatform\Core\JsonSchema\TypeFactoryInterface;
+use Symfony\Component\PropertyInfo\Type;
+
+class ElaoEnumType implements TypeFactoryInterface
+{
+    /**
+     * @var TypeFactoryInterface
+     */
+    private $decorated;
+
+    /**
+     * @param TypeFactoryInterface $decorated
+     */
+    public function __construct(TypeFactoryInterface $decorated)
+    {
+        $this->decorated = $decorated;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getType(Type $type, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, Schema $schema = null): array
+    {
+        if (is_a($enumClass = $type->getClassName(), 'Elao\Enum\EnumInterface', true)) {
+
+            $values = $enumClass::values();
+
+            return [
+                'type' => 'string',
+                'enum' => $values,
+                'example' => reset($values),
+            ];
+        }
+
+        return $this->decorated->getType($type, $format, $readableLink, $serializerContext, $schema);
+    }
+}
--- a/src/JsonSchema/Type/ElaoEnumType.php
+++ b/src/JsonSchema/Type/ElaoEnumType.php
@@ -44,7 +44,7 @@ final class ElaoEnumType implements TypeFactoryInterface
         return [
             'type' => 'string',
             'enum' => $values,
-            'example' => reset($values),
+            'example' => $values[0],
         ];
     }
 }
--- a/src/JsonSchema/Type/ElaoEnumType.php
+++ b/src/JsonSchema/Type/ElaoEnumType.php
@@ -23,9 +23,6 @@ class ElaoEnumType implements TypeFactoryInterface
      */
     private $decorated;
 
-    /**
-     * @param TypeFactoryInterface $decorated
-     */
     public function __construct(TypeFactoryInterface $decorated)
     {
         $this->decorated = $decorated;
--- a/src/JsonSchema/Type/ElaoEnumType.php
+++ b/src/JsonSchema/Type/ElaoEnumType.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the API Platform project.
  *
@@ -14,6 +15,7 @@ namespace ApiPlatform\Core\JsonSchema\Type;
 
 use ApiPlatform\Core\JsonSchema\Schema;
 use ApiPlatform\Core\JsonSchema\TypeFactoryInterface;
+use Elao\Enum\EnumInterface;
 use Symfony\Component\PropertyInfo\Type;
 
 class ElaoEnumType implements TypeFactoryInterface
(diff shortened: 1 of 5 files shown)


src/JsonSchema/Type/ElaoEnumType.php
```diff
@@ -0,0 +1,49 @@
...
+            $values = $enumClass::values();
+
+            return [
+                'type' => 'string',
+                'enum' => $values,
+                'example' => reset($values),
+            ];
+        }
+
+        return $this->decorated->getType($type, $format, $readableLink, $serializerContext, $schema);
```
soyuka, line 47:
```suggestion
        if (!is_a($enumClass = $type->getClassName(), EnumInterface::class, true)) {
            return $this->decorated->getType($type, $format, $readableLink, $serializerContext, $schema);
        }
        
        $values = $enumClass::values();

        return [
           'type' => 'string',
           'enum' => $values,
           'example' => reset($values),
        ];      
```

Don't forget to import EnumInterface

WhiteRabbitDE: I did not use the EnumInterface::class because elao is not required in the composer.json. Doesnt that cause an Error ?

src/JsonSchema/Type/ElaoEnumType.php
```diff
@@ -0,0 +1,49 @@
+<?php
+/*
```
azjezz, line 2:
```suggestion
<?php

/*
```

src/JsonSchema/Type/ElaoEnumType.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getType(Type $type, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, Schema $schema = null): array
+    {
+        if (is_a($enumClass = $type->getClassName(), 'Elao\Enum\EnumInterface', true)) {
+
+            $values = $enumClass::values();
```
azjezz, line 38:
```suggestion
        if (is_a($enumClass = $type->getClassName(), 'Elao\Enum\EnumInterface', true)) {
            $values = $enumClass::values();
```

src/JsonSchema/Type/ElaoEnumType.php
azjezz: @WhiteRabbitDE no.

WhiteRabbitDE: Done.

src/JsonSchema/Type/ElaoEnumType.php
WhiteRabbitDE: Done.

WhiteRabbitDE:
PHPStan has an error now since the class is unknown. Should I add the ignore to the phpstan config ?

  Line   src/JsonSchema/Type/ElaoEnumType.php      
  38     Class Elao\Enum\EnumInterface not found.

teohhanhui:
You should add the elao package to `require-dev` (make sure it's sorted alphabetically, if you're adding manually).

src/JsonSchema/Type/ElaoEnumType.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\JsonSchema\Type;
+
+use ApiPlatform\Core\JsonSchema\Schema;
+use ApiPlatform\Core\JsonSchema\TypeFactoryInterface;
+use Elao\Enum\EnumInterface;
+use Symfony\Component\PropertyInfo\Type;
+
+class ElaoEnumType implements TypeFactoryInterface
```
soyuka, line 21:
```suggestion
final class ElaoEnumType implements TypeFactoryInterface
```

src/JsonSchema/Type/ElaoEnumType.php
```diff
@@ -0,0 +1,50 @@
...
+        if (!is_a($enumClass = $type->getClassName(), EnumInterface::class, true)) {
+            return $this->decorated->getType($type, $format, $readableLink, $serializerContext, $schema);
+        }
+
+        $values = $enumClass::values();
+
+        return [
+            'type' => 'string',
+            'enum' => $values,
+            'example' => reset($values),
```
soyuka, line 47: I'm not sure to get why the `reset` here, it should work without it right?

soyuka:
:+1: for the patch, to me it's a new feature though and should target master! I'll merge 2.5 onto master later today.

src/JsonSchema/Type/ElaoEnumType.php
WhiteRabbitDE: Because $values is an array and example takes a string afaik. Its just to get the first value as example.

soyuka: Indeed, or just ` $values[0]` if that works.

WhiteRabbitDE: I thought it was not number-indexed but i checked in my project and it is. It changed it to $values[0].

dunglas:
I would prefer to not have this in core. It's yet another (dev) dependency to a third party library, and it's a pain to maintain. IMHO it belongs in a third party package or directly in the Elao lib. Also Other parts of API Platform (Schema Generator) use @mnapoli's php-enum, if we really want to support an enum package in Core, I would prefer to support Mathieu's one for consistency with the rest of the code base

TBH, I don't think that supporting a 3rd party package is good idea, we should probably write a Symfony polyfill for `SplEnum` instead.

WhiteRabbitDE:
@dunglas For me it is fine if we dont merge that and i understand the reasons. 

But imho we should add docs for people encountering this problem because atm it is not possible to overwrite the (swagger/openapi)_context for the enums. And, at least for me, it was not obvious that i cant do that and need a decorator.

dunglas:
For sure! If you find some time to document this, it will definitely be very helpful.

Regarding the Elao Enum integration, it would be nice to publish it as bundle (and we can add a section about this bundle in the documentation).

Issue #3299, opened 2019-12-05
Add support for file uploads for test client


nicodemuz:
**Description**  
Add support for file uploads for test client. Current client does not seem to support this.

teohhanhui:
Please see the relevant docs section here: https://symfony.com/doc/current/components/http_client.html#uploading-data

> To submit a form with file uploads, it is your responsibility to encode the body according to the multipart/form-data content-type. The Symfony Mime component makes it a few lines of code:
> 
> ```php
> use Symfony\Component\Mime\Part\DataPart;
> use Symfony\Component\Mime\Part\Multipart\FormDataPart;
> 
> $formFields = [
>     'regular_field' => 'some value',
>     'file_field' => DataPart::fromPath('/path/to/uploaded/file'),
> ];
> $formData = new FormDataPart($formFields);
> $client->request('POST', 'https://...', [
>     'headers' => $formData->getPreparedHeaders()->toArray(),
>     'body' => $formData->bodyToIterable(),
> ]);
> ```

dunglas:
That being said, it could be nice to add a link to this Symfony docs entry in our own docs.

teohhanhui:
Sorry, I was mistaken on this one. Our test `Client` indeed does not support file uploads.

teohhanhui:
See

https://github.com/api-platform/core/blob/215709aafd73a5daa491c977b76f152a3aa2aaed/src/Bridge/Symfony/Bundle/Test/Client.php#L124

We're passing an empty array for the `$files` argument (4th argument).

bettinz:
Hello, is there any option to test file upload with Api-Platform? I'm using Vich and Flysystem with your guide. Thanks

MrNicodemuz:
I have managed to make it work using the following code:
```php
<?php

namespace App\Tests\Controller\Api;

use App\Entity\Media;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\HttpFoundation\File\UploadedFile;
use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
use App\Repository\UserRepository;

/**
 * @group functional
 * @group api
 */
class MediaApiTest extends WebTestCase
{
    /**
     * Create a client with a default Authorization header.
     *
     * @param string $username
     * @param string $password
     *
     * @return \Symfony\Bundle\FrameworkBundle\Client
     */
    protected function createAuthenticatedApiClient($username = 'test')
    {
[...]

PR #3486, opened 2020-04-09
Add files to test client

CoalaJoe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #3299
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/Test/Client.php
+++ b/src/Bridge/Symfony/Bundle/Test/Client.php
@@ -84,7 +84,7 @@ final class Client implements HttpClientInterface
      *
      * @return Response
      */
-    public function request(string $method, string $url, array $options = [], array $files = []): ResponseInterface
+    public function request(string $method, string $url, array $options = []): ResponseInterface
     {
         $basic = $options['auth_basic'] ?? null;
         [$url, $options] = self::prepareRequest($method, $url, $options, $this->defaultOptions);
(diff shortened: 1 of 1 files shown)


src/Bridge/Symfony/Bundle/Test/Client.php
```diff
@@ -84,7 +84,7 @@ public function setDefaultOptions(array $defaultOptions): void
      *
      * @return Response
      */
-    public function request(string $method, string $url, array $options = []): ResponseInterface
+    public function request(string $method, string $url, array $options = [], array $files = []): ResponseInterface
```
soyuka, line 87: We can't change the method sign here as it should respect this interface: https://github.com/symfony/contracts/blob/master/HttpClient/HttpClientInterface.php#L84. Files aren't in the option list, we could maybe support this via the `extra` key?

CoalaJoe: Updated to your suggestion

src/Bridge/Symfony/Bundle/Test/Client.php
```diff
@@ -121,7 +122,8 @@ public function request(string $method, string $url, array $options = []): Respo
             'url' => $resolvedUrl,
             'primary_port' => 'http:' === $url['scheme'] ? 80 : 443,
         ];
-        $this->kernelBrowser->request($method, $resolvedUrl, [], [], $server, $options['body'] ?? null);
+        $extra = $options['extra'] ?? [];
+        $this->kernelBrowser->request($method, $resolvedUrl, [], $extra['files'] ?? [], $server, $options['body'] ?? null);
```
soyuka, line 126: you can use `$options['extra']['files'] ?? []` directly here

CoalaJoe: I've updated it based on your suggestion

soyuka:
Thanks @CoalaJoe we could maybe add some documentation to that in api-platform/docs!

dunglas:
As reported on Slack, this should be backported in 2.5 as it's a bug fix.

PR #3488, opened 2020-04-15
Improve the RangeFilter DQL query in case the "between" values are equals

epourail:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Improve the behavior of the RangeFilter filter in case the values of the between operator are equals
In such case, the DQL query is defined with the equals operation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.5.x-dev
 
 * GraphQL: Do not allow empty cursor values on `before` or `after`
+* Filter: Improve the RangeFilter query in case the values are equals using the between operator
 
 ## 2.5.4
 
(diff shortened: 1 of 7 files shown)


soyuka:
I consider this as a bug fix, could you please target 2.5? Thanks!

epourail:
@soyuka Done (updated to target 2.5)

bendavies:
`BETWEEN` is inclusive of boundaries so what does this fix?

epourail:
It will just improve the generated query syntax to use a single-comparison operation instead of a two-ways comparison (greater than and less than).

soyuka:
looks like php-cs-fixer needs to be run, make sure to rebase against the latest 2.5 though :)

soyuka:
thanks @epourail

PR #3497, opened 2020-04-16
Prevent header Location on POST when status code is not 201 or 30x

fmata:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

When a POST request issues a response with status code 200 setted by API Platform, the client receives 302 status code because of header Location added by RespondListener.

In this case, Safari follows the redirect and the client performs a GET operation on the URL in Location header.

According [PHP doc](https://www.php.net/manual/en/function.header) :

> The second special case is the "Location:" header. Not only does it send this header back to the browser, but it also returns a REDIRECT (302) status code to the browser unless the 201 or a 3xx status code has already been set.

PHP seems overwrite the original status code to 302 when it's not 201 or 30x and Location header exists.

If I force a status code to 201, it works as expected.

### DIFF
--- a/tests/EventListener/RespondListenerTest.php
+++ b/tests/EventListener/RespondListenerTest.php
@@ -80,6 +80,38 @@ class RespondListenerTest extends TestCase
         $this->assertEquals('deny', $response->headers->get('X-Frame-Options'));
     }
 
+    public function testPostCreate200Response()
+    {
+        $kernelProphecy = $this->prophesize(HttpKernelInterface::class);
+
+        $request = new Request([], [], ['_api_respond' => true, '_api_write_item_iri' => '/dummy_entities/1', '_api_item_operation_name' => 'post']);
+        $request->setMethod('POST');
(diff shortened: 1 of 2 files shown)


soyuka:
Definitely the right fix + respects RFC 7231

May you add a non-regression unit test please? Maybe we can still add this to 2.5.5 (hoping to release in the following days).

tests/EventListener/RespondListenerTest.php
```diff
@@ -80,6 +80,38 @@ public function testCreate200Response()
         $this->assertEquals('deny', $response->headers->get('X-Frame-Options'));
     }
 
+    public function testPostCreate200Response()
```
julienfalque, line 83: Maybe worth adding tests for codes between 300 and 400?

soyuka:
thanks @fmata

PR #3502, opened 2020-04-18
Add PHPUnit sqlite dependency in composer

symfonyaml:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no

## Problem
When I run the phpunit tests `vendor/bin/phpunit` without sqllite extension installed,
Then some tests failed like `ApiPlatformProfilerPanelTest`

## Solution
Add PHPUnit sqlite dependency in composer
```json
"require-dev": {
    "ext-sqlite3": "*"
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
         "willdurand/negotiation": "^2.0.3"
     },
     "require-dev": {
+        "ext-sqlite3": "*",
         "behat/behat": "^3.1",
         "behat/mink": "^1.7",
         "behat/mink-browserkit-driver": "^1.3.1",


teohhanhui:
> The SQLite3 extension is enabled by default as of PHP 5.3.0.
>
> https://www.php.net/manual/en/sqlite3.installation.php

I don't see the point of listing every default extension.

symfonyaml:
@teohhanhui I totally agree we don't want to list all the extensions. I just forgot to mention that I did this PR because when you run the tests without SQLite, **nowhere it says that the sqlite3 extension is missing**. I had to debug the tests for a long time to realise that doctrine was using SQLite.

So if we can avoid new developers like me to spend the same time digging n this issue, and tell directly we need SQLite, that would be some time savings.

PR #3518, opened 2020-04-23
Adjust null handling in exported JSON Schema / OpenAPI

Korbeil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

https://github.com/api-platform/core/pull/3402/ was a really good start to have a better null handling in exported schema.

But actually we should use required attribute as a condition for nullability.
A non-required property should be nullable.

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -228,7 +228,7 @@ final class SchemaFactory implements SchemaFactoryInterface
                 $className = $valueType->getClassName();
             }
 
-            $valueSchema = $this->typeFactory->getType(new Type($builtinType, $type->isNullable(), $className, $isCollection), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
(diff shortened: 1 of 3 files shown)


src/JsonSchema/SchemaFactory.php
```diff
@@ -214,7 +214,7 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
                 $className = $valueType->getClassName();
             }
 
-            $valueSchema = $this->typeFactory->getType(new Type($builtinType, $type->isNullable(), $className, $isCollection), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
+            $valueSchema = $this->typeFactory->getType(new Type($builtinType, !$propertyMetadata->isRequired() || $type->isNullable(), $className, $isCollection), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
```
Korbeil, line 217: Being nullable is not enought here, we also need to check if the property is required.
If it's not required, it means this property should be nullable.

alanpoulain:
LGTM, shouldn't we merge it @soyuka?

alanpoulain:
Should it be really the case though?
If a property is not required but has a default value, it shouldn't be nullable, isn't it?

PR #3519, opened 2020-04-23
Allow defining exception_to_status per operation

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR allows introduces the `exception_to_status` attribute in operations and resources mapping.

The use case is when some domain exception should be transformed into a specific HTTP status on some API endpoints, but not on others.

For example: a method in some entity throws a `NotPossible` exception when called through an API endpoint. You want that endpoint to return a `400 Bad Request`. But if the same entity method throws that exception when called from unrelated internal code, maybe that's actually an error in the code and should result in a `500 Internal Server Error`.

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -70,6 +70,7 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
  *     @Attribute("swaggerContext", type="array"),
  *     @Attribute("urlGenerationStrategy", type="int"),
  *     @Attribute("validationGroups", type="mixed"),
+ *     @Attribute("exceptionToStatus", type="array"),
  * )
  */
 #[\Attribute(\Attribute::TARGET_CLASS)]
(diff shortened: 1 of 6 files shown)


alanpoulain:
Thank you for your PR!
Why not indeed.
Could you open a PR for the documentation too?

julienfalque:
Sure, see https://github.com/api-platform/docs/pull/1309.

julienfalque:
@alanpoulain I updated `ApiResource` according to https://github.com/api-platform/docs/pull/1309#discussion_r587638510. As far as I understand, the root `exception_to_status` will be copied to `$attributes` so everything should work but I'm not actually sure. Do you mind having a look?

src/Annotation/ApiResource.php
```diff
@@ -219,7 +221,8 @@ public function __construct(
         ?array $swaggerContext = null,
         ?array $validationGroups = null,
         ?int $urlGenerationStrategy = null,
-        ?bool $compositeIdentifier = null
+        ?bool $compositeIdentifier = null,
+        array $exceptionToStatus = []
```
alanpoulain, line 225:
```suggestion
        ?array $exceptionToStatus = null
```

alanpoulain:
Yes, it will be OK like this!

src/Annotation/ApiResource.php
julienfalque: Updated.

alanpoulain:
Thank you @julienfalque.

julienfalque:
Thank you @alanpoulain.

Issue #3487, opened 2020-04-14
FilterEagerLoadingExtension duplicates joins added by EagerLoadingExtension


julienfalque:
**API Platform version(s) affected**: 2.5.4

**Description**  
On a collection operation with filtering, `FilterEagerLoadingExtension` transforms the query into a wrapped subquery: `... WHERE o.id IN( <original query> )`. When doing so, it duplicates joins from the original query and keeps them on the wrapping query in the result.

Though some of those joins might have been added by `EagerLoadingExtension`. Duplicating them is not necessary as they don't change the result (the retrieved resources will be the same). They should appear in the wrapping query, but not in the wrapped one. With many extra joins, this can lead to performance penalty. In my case I'm not sure yet whether the performance penalty was a database issue (e.g. missing index) but preventing joins duplication fixed it.

**How to reproduce**  
Use a filter on a collection operation of a resource that has many eager loaded relations.

**Possible Solution**  
To prevent the duplication, I changed the `priority` attribute of the `api_platform.doctrine.orm.query_extension.collection` tag of `EagerLoadingExtension`: I set it to `-18` instead of `-8` so it runs *after* `FilterEagerLoadingExtension`. This did not break any test on the project but I don't know all the implications of such change. Should the tag be updated in `api-platform/core`? If so, maybe `FilterEagerLoadingExtension` could be merged into `EagerLoadingExtension`?

soyuka:
Maybe related: https://github.com/api-platform/core/issues/2908

IIRC the order is important because FilterEagerLoadingExtension should be run after filter which runs after eagerloadingextension or something like that :p. We covered the use case with a behat test so if you try running them it should fail somewhere. The solution proposed in the related issue may be a cleaner one?

julienfalque:
As far as I understand, #2908 is a slighly different issue: it is about not adding eager loading joins when those joins already exist in the query. In my case, there was no joins in the original query.

I'll have a look at the Behat test you mentionned.

julienfalque:
So I tried running Behat tests with the change described above and all tests pass on both `2.5` and `master` branches. Shall we give my proposal a try?

soyuka:
Why not indeed!

julienfalque:
#3525 has been merged.

PR #3525, opened 2020-04-24
Fix duplicated eager loading joins

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #3487
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
@@ -131,7 +131,8 @@
             <argument type="service" id="serializer.mapping.class_metadata_factory" />
 
             <tag name="api_platform.doctrine.orm.query_extension.item" priority="-8" />
-            <tag name="api_platform.doctrine.orm.query_extension.collection" priority="-8" />
+            <!-- After filter_eager_loading -->
+            <tag name="api_platform.doctrine.orm.query_extension.collection" priority="-18" />
         </service>
         <service id="ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\EagerLoadingExtension" alias="api_platform.doctrine.orm.query_extension.eager_loading" />
 
--- a/features/doctrine/eager_loading.feature
+++ b/features/doctrine/eager_loading.feature
@@ -1,0 +1,80 @@
+@!mongodb
+Feature: Eager Loading
+  In order to have better performance
+  As a client software developer
+  The eager loading should be enabled
+
+  @createSchema
+  Scenario: Eager loading for a relation
+    Given there is a RelatedDummy with 2 friends
+    When I send a "GET" request to "/related_dummies/1"
+    And the response status code should be 200
+    And the DQL should be equal to:
+    """
+    SELECT o, thirdLevel_a1, relatedToDummyFriend_a2, dummyFriend_a3
+    FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy o
+        LEFT JOIN o.thirdLevel thirdLevel_a1
+        LEFT JOIN o.relatedToDummyFriend relatedToDummyFriend_a2
+        LEFT JOIN relatedToDummyFriend_a2.dummyFriend dummyFriend_a3
+    WHERE o.id = :id_id
+    """
+
+  Scenario: Eager loading for the search filter
+    Given there is a dummy object with a fourth level relation
+    When I send a "GET" request to "/dummies?relatedDummy.thirdLevel.level=3"
+    Then the response status code should be 200
+    And the DQL should be equal to:
+    """
+    SELECT o
+    FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o
+        INNER JOIN o.relatedDummy relatedDummy_a1
+        INNER JOIN relatedDummy_a1.thirdLevel thirdLevel_a2
+    WHERE o IN(
+            SELECT o_a3
+            FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o_a3
+                INNER JOIN o_a3.relatedDummy relatedDummy_a4
+                INNER JOIN relatedDummy_a4.thirdLevel thirdLevel_a5
+            WHERE thirdLevel_a5.level = :level_p1
+        )
+    ORDER BY o.id ASC
+    """
+
+  Scenario: Eager loading for a relation and a search filter
+    Given there is a RelatedDummy with 2 friends
+    When I send a "GET" request to "/related_dummies?relatedToDummyFriend.dummyFriend=2"
+    And the response status code should be 200
+    And the DQL should be equal to:
+    """
+    SELECT o, thirdLevel_a4, relatedToDummyFriend_a1, dummyFriend_a5
+    FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy o
+        INNER JOIN o.relatedToDummyFriend relatedToDummyFriend_a1
+        LEFT JOIN o.thirdLevel thirdLevel_a4
+        INNER JOIN relatedToDummyFriend_a1.dummyFriend dummyFriend_a5
+    WHERE o IN(
+            SELECT o_a2
+            FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy o_a2
+                INNER JOIN o_a2.relatedToDummyFriend relatedToDummyFriend_a3
+            WHERE relatedToDummyFriend_a3.dummyFriend = :dummyFriend_p1
+        )
+    ORDER BY o.id ASC
+    """
+
+  Scenario: Eager loading for a relation with complex sub-query filter
+    Given there is a RelatedDummy with 2 friends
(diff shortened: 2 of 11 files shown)


soyuka:
This looks good according to the tests, do you think that there's a way to add a non-regression test?

julienfalque:
I was not able to reproduce the issue I mentioned in #3487 yet, I need a more complex filter I think. For now, changing the priority results in:

```diff
 SELECT o, relatedDummy_a1
 FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o
-    LEFT JOIN o.relatedDummy relatedDummy_a1
-    LEFT JOIN relatedDummy_a1.thirdLevel thirdLevel_a2
+    INNER JOIN o.relatedDummy relatedDummy_a1
+    INNER JOIN relatedDummy_a1.thirdLevel thirdLevel_a2
 WHERE o IN(
         SELECT o_a3
         FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o_a3
-            LEFT JOIN o_a3.relatedDummy relatedDummy_a4
-            LEFT JOIN relatedDummy_a4.thirdLevel thirdLevel_a5
+            INNER JOIN o_a3.relatedDummy relatedDummy_a4
+            INNER JOIN relatedDummy_a4.thirdLevel thirdLevel_a5
         WHERE thirdLevel_a5.level = :level_p1
     )
 ORDER BY o.id ASC
```
which looks ok. What do you think about the current approach?

julienfalque:
Ok I reproduced the issue (see method [`testWithComplexFilterAndEagerLoading()`](https://github.com/api-platform/core/pull/3525/files#diff-7a879088cbd9fce2bb1551e247ef471eR120)). Without the changes the resulting DQL query would be:
```
SELECT o, relatedDummy_a1
FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o
    LEFT JOIN o.relatedDummy relatedDummy_a1
WHERE o IN(
        SELECT o_a7
        FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o_a7
            LEFT JOIN o_a7.relatedDummy relatedDummy_a8
        WHERE o_a7.id IN (
                SELECT dummy_a2.id
                FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy dummy_a2
                    INNER JOIN dummy_a2.related_dummy related_dummy_a3
                        WITH related_dummy_a3.name = :name_p1
                        AND related_dummy_a3.dummyDate >= :dummy_date_p2
                    INNER JOIN dummy_a2.related_dummies related_dummies_a4
                        WITH related_dummies_a4.name = :name_p1
                        AND related_dummies_a4.dummyDate >= :dummy_date_p2
                    INNER JOIN dummy_a2.related_owned_dummy related_owned_dummy_a5
                        WITH related_owned_dummy_a5.name <= :name_p1
                    INNER JOIN dummy_a2.related_owning_dummy related_owning_dummy_a6
                        WITH related_owning_dummy_a6.name <= :name_p1
            )
    )
ORDER BY o.id ASC
```
The nested query in the first `WHERE IN (...)` is the same as the outer query. With the fix, this duplicate is removed:
```
SELECT o, relatedDummy_a6
FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy o
    LEFT JOIN o.relatedDummy relatedDummy_a6
WHERE o.id IN (
        SELECT dummy_a1.id
        FROM ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy dummy_a1
            INNER JOIN dummy_a1.related_dummy related_dummy_a2
                WITH related_dummy_a2.name = :name_p1
                AND related_dummy_a2.dummyDate >= :dummy_date_p2
            INNER JOIN dummy_a1.related_dummies related_dummies_a3
                WITH related_dummies_a3.name = :name_p1
                AND related_dummies_a3.dummyDate >= :dummy_date_p2
            INNER JOIN dummy_a1.related_owned_dummy related_owned_dummy_a4
                WITH related_owned_dummy_a4.name <= :name_p1
            INNER JOIN dummy_a1.related_owning_dummy related_owning_dummy_a5
                WITH related_owning_dummy_a5.name <= :name_p1
    )
ORDER BY o.id ASC
```

julienfalque:
Behat failures seem unrelated.

alanpoulain:
Hello @julienfalque. Sorry for the time we took to review your PR.
I've replaced your integration tests with Behat ones, to make it easier to maintain.
However I haven't reproduced the `testWithComplexFilterAndEagerLoading`.
Could you explain to me what this test was doing and if we should keep it?

julienfalque:
@alanpoulain Thank you for tackling this. To be honest I don't remember exactly. I think this specific test case was the only one that actually reproduced #3487 (see https://github.com/api-platform/core/pull/3525#issuecomment-622501021) so it should be kept. I'll have a closer look as soon as I can.

alanpoulain:
OK I've added it again then.

alanpoulain:
@soyuka could you take a look at it?

julienfalque:
I cannot approve my own PR but your changes look :+1:, thanks!

alanpoulain:
Thank you :slightly_smiling_face:

julienfalque:
Thank you @alanpoulain @soyuka.

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml
```diff
@@ -131,7 +131,8 @@
             <argument type="service" id="serializer.mapping.class_metadata_factory" />
 
             <tag name="api_platform.doctrine.orm.query_extension.item" priority="-8" />
-            <tag name="api_platform.doctrine.orm.query_extension.collection" priority="-8" />
+            <!-- After filter_eager_loading -->
+            <tag name="api_platform.doctrine.orm.query_extension.collection" priority="-18" />
```
zolex, line 135: this priority change breaks filters that rely on existing joins from the eager loading extension because now the eager loading runs after the filters.

zolex: exmaple filter that now breaks.

```
protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
    {
        if ('searchTerm' !== $property || !in_array($resourceClass, [User::class, Employment::class]))
            return;

        $alias = $rootAlias = $queryBuilder->getRootAliases()[0];
        if (User::class !== $resourceClass) {
            $joins = $queryBuilder->getDQLPart("join");
            if (array_key_exists($rootAlias, $joins)) {
                /** @var \Doctrine\ORM\Query\Expr\Join $join */
                foreach ($joins[$rootAlias] as $join) {
                    if (sprintf('%s.user', $rootAlias) === $join->getJoin())
                        $alias = $join->getAlias();
                }
            }
        }

        $searchParam = $queryNameGenerator->generateParameterName('search');

        $conditions = [];
        $conditions[] = sprintf('%1$s.firstname LIKE :%2$s', $alias, $searchParam);
        $conditions[] = sprintf('%1$s.lastname LIKE :%2$s', $alias, $searchParam);
        $conditions[] = sprintf('CONCAT(%1$s.firstname, \' \', %1$s.lastname) LIKE :%2$s', $alias, $searchParam);
        $conditions[] = sprintf('%1$s.email LIKE :%2$s', $alias, $searchParam);

        $queryBuilder->andWhere(implode(' OR ', $conditions))
            ->setParameter($searchParam, $value . "%");
    }
```

zolex: we could workaround this issue by adding the join explicitly in the filter, but now we join the relation two times, one by the filter and another by the eager loading extension.

alanpoulain: > we could workaround this issue by adding the join explicitly in the filter, but now we join the relation two times, one by the filter and another by the eager loading extension.

If the extension works correctly, the existing join is reused then.

@julienfalque do you think it's possible to solve the issue without changing the priority?

julienfalque: I'll have to check again, but as I remember this was required to prevent duplicating joins as described in #3487.

julienfalque: Just had a look: changing this priority is required to get `EagerLoadingExtension` applied after `FilterEagerLoadingExtension` to fix #3487. I don't think `FilterEagerLoadingExtension`'s priority can be changed (it has to be applied after all filters).

> we could workaround this issue by adding the join explicitly in the filter, but now we join the relation two times, one by the filter and another by the eager loading extension.

If I remember correctly, the changes in this PR make `EagerLoadingExtension` able to rely on existing joins when possible instead of always adding its own. @zolex Are you sure you have duplicated joins if you add one in your filter?

zolex: I just assumed it because in earlier Versions it happened. I will check the resulting queries tomorrow and let you know.

PR #3526, opened 2020-04-24
Improve PaginationExtension declaration

ilicmsreten:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Currently [PaginationExtension::applyToCollection](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php#L119) method has implementation defined in [ContextAwareQueryCollectionExtensionInterface](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Extension/ContextAwareQueryCollectionExtensionInterface.php#L29).

### DIFF
--- a/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
@@ -34,7 +34,7 @@ use Symfony\Component\HttpFoundation\RequestStack;
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Samuel ROZE <samuel.roze@gmail.com>
  */
-final class PaginationExtension implements ContextAwareQueryResultCollectionExtensionInterface
+final class PaginationExtension implements ContextAwareQueryResultCollectionExtensionInterface, ContextAwareQueryCollectionExtensionInterface
 {
     private $managerRegistry;
     private $requestStack;


src/Bridge/Doctrine/Orm/Extension/PaginationExtension.php
```diff
@@ -34,7 +34,7 @@
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Samuel ROZE <samuel.roze@gmail.com>
  */
-final class PaginationExtension implements ContextAwareQueryResultCollectionExtensionInterface
+final class PaginationExtension implements ContextAwareQueryResultCollectionExtensionInterface, ContextAwareQueryCollectionExtensionInterface
```
soyuka, line 37: Should target 2.5 as a bug fix

soyuka, line 37: Thing is ContextAwareQueryResultCollectionExtensionInterface extends QueryCollectionExtensionInterface and changing the interface would go against the BC promise see https://symfony.com/doc/current/contributing/code/bc.html#changing-interfaces (at least I think). We can't really do this for now :/

PR #3521, opened 2020-04-24
Search filter introduces #3331 back with deprecations

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes
| Tickets       | introduces #3331 back with deprecations 
| License       | MIT
| Doc PR        | na

Note: Merge 2.5 onto master before merging this.

PR #3541, opened 2020-05-01
SearchFilter: Refactor code to reduce code duplication and makes it easier to add new strategies to classes inheriting from SearchFilter class

jpierront:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

All query manipulations are now located in the same method to ease the creation on classes inherited from SearchFilter.

Example:
```
class ExtendedSearchFilter extends SearchFilter
{
    public const STRATEGY_OTHER = 'other';

    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, string $alias, string $field, $fieldType, $values, bool $caseSensitive, string $valueParameter)
    {
        if (self::STRATEGY_OTHER === $strategy) {
            // ...

            return;
        }

        parent::addWhereByStrategy($strategy, $queryBuilder, $alias, $field, $fieldType, $values, $caseSensitive, $valueParameter);
    }
}
```

We will be able to create easily a "not_int" strategy or more complexe cases.

### DIFF
--- a/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
@@ -87,12 +87,17 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
             [$matchField, $field, $associations] = $this->addLookupsForNestedProperty($property, $aggregationBuilder, $resourceClass);
         }
 
-        /**
-         * @var MongoDBClassMetadata
-         */
-        $metadata = $this->getNestedMetadata($resourceClass, $associations);
-
         $caseSensitive = true;
+        $strategy = $this->properties[$property] ?? self::STRATEGY_EXACT;
+
+        // prefixing the strategy with i makes it case insensitive
+        if (0 === strpos($strategy, 'i')) {
+            $strategy = substr($strategy, 1);
+            $caseSensitive = false;
+        }
+
+        /** @var MongoDBClassMetadata */
+        $metadata = $this->getNestedMetadata($resourceClass, $associations);
 
         if ($metadata->hasField($field) && !$metadata->hasAssociation($field)) {
             if ('id' === $field) {
@@ -107,23 +112,9 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
                 return;
             }
 
-            $strategy = $this->properties[$property] ?? self::STRATEGY_EXACT;
+            $this->addEqualityMatchStrategy($strategy, $aggregationBuilder, $field, $matchField, $values, $caseSensitive, $metadata);
 
-            // prefixing the strategy with i makes it case insensitive
-            if (0 === strpos($strategy, 'i')) {
-                $strategy = substr($strategy, 1);
-                $caseSensitive = false;
-            }
-
-            $inValues = [];
-            foreach ($values as $inValue) {
-                $inValues[] = $this->addEqualityMatchStrategy($strategy, $field, $inValue, $caseSensitive, $metadata);
-            }
-
-            $aggregationBuilder
-                ->match()
-                ->field($matchField)
-                ->in($inValues);
+            return;
         }
 
         // metadata doesn't have the field, nor an association on the field
@@ -132,7 +123,6 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
         }
 
         $values = array_map([$this, 'getIdFromValue'], $values);
-        $associationFieldIdentifier = 'id';
         $doctrineTypeField = $this->getDoctrineFieldType($property, $resourceClass);
 
         if (null !== $this->identifiersExtractor) {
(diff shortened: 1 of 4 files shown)


jpierront:
I'm not sure of what I did for the MongoDbOdm part ;-)

alanpoulain:
Seems nice! Could you add a Behat test to validate its behavior? It will allow to make sure MongoDB works too.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -155,63 +137,67 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
+        $this->addWhereByStrategy($strategy, $queryBuilder, $associationAlias, $associationField, $doctrineTypeField, $values, $caseSensitive, $valueParameter);
     }
 
     /**
      * Adds where clause according to the strategy.
      *
      * @throws InvalidArgumentException If strategy does not exist
      */
-    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $fieldType, $value, bool $caseSensitive)
+    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, string $alias, string $field, $fieldType, array $values, bool $caseSensitive, string $valueParameter)
```
alanpoulain, line 157: It's a BC break.

jpierront:
Finally, by creating the behat tests, I realized that my code was not really doing what I expected.

* Doing a "partial" search on "relatedDumies.name" was already working before my PR.
* Doing a "partial" search on "relatedDumies" (So on the identifier) is not accepted by Doctrine ORM (It doesn't accept to make a "LIKE" on identifier in this case)

Finally I’ll just keep the refactoring simplifying the creation of new strategies in the inherited classes. It was what I needed initially ;-)

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -149,10 +141,25 @@ protected function filterProperty(string $property, $value, Builder $aggregation
...
 
+        $this->addEqualityMatchStrategy($strategy, $aggregationBuilder, $matchField, $doctrineTypeField, $values, $caseSensitive);
+    }
+
+    /**
+     * Adds where clause according to the strategy.
+     *
+     * @throws InvalidArgumentException If strategy does not exist
+     */
+    protected function addEqualityMatchStrategy(string $strategy, Builder $aggregationBuilder, string $matchField, $fieldType, $values, bool $caseSensitive)
```
alanpoulain, line 152: Since the class is final, it should be private.

jpierront: So we allow to extends the ORM version but not the ODM one ? :/

alanpoulain: Yes. Because the ORM version will also be a final class in the next major version (3.0).

jpierront:
OK, so my original intention will lose its meaning in V3. ;-)
My refactoring still seems interesting to centralize the code related to the strategy in the same place.

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -149,10 +141,25 @@ protected function filterProperty(string $property, $value, Builder $aggregation
             return;
         }
 
+        $this->addEqualityMatchStrategy($strategy, $aggregationBuilder, $matchField, $doctrineTypeField, $values, $caseSensitive);
+    }
+
+    /**
+     * Adds where clause according to the strategy.
```
alanpoulain, line 148:
```suggestion
     * Add equality match stage according to the strategy.
```

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -162,10 +169,8 @@ protected function filterProperty(string $property, $value, Builder $aggregation
      *
      * @return Regex|string
      */
-    private function addEqualityMatchStrategy(string $strategy, string $field, $value, bool $caseSensitive, ClassMetadata $metadata)
+    private function getEqualityMatchStrategy(string $strategy, $value, bool $caseSensitive, string $type)
```
alanpoulain, line 172:
```suggestion
    private function getEqualityMatchValue(string $strategy, $value, bool $caseSensitive, string $type)
```

alanpoulain: PHPDoc should be modified too.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -155,63 +136,69 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
+        $this->addWhereByStrategy($strategy, $queryBuilder, $queryNameGenerator, $associationAlias, $associationField, $doctrineTypeField, $values, $caseSensitive, $valueParameter);
     }
 
     /**
      * Adds where clause according to the strategy.
      *
      * @throws InvalidArgumentException If strategy does not exist
      */
-    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $fieldType, $value, bool $caseSensitive)
+    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $fieldType, $values, bool $caseSensitive, string $valueParameter = null)
```
alanpoulain, line 156: Even like this, it's a BC break. You should mimic what has been done here: https://github.com/api-platform/core/pull/3521.

alanpoulain:
Thanks @jpierront.

PR #3543, opened 2020-05-04
Allow a resource Cache-Control HTTP header to be private

renaudclosset:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

This change allows to mark a resource as "private" in the cache headers, to avoid having it being cached by shared caching services (Varnish, etc.)

**Example:**
User resources (`/api/user/xxx`) may contain sensitive information that cannot be controlled fine-grained enough with a Vary-header. So these resources should only be cached on the client.

**Configuration:**
```yaml
resources:
  App\Entity\User:
    attributes:
      cache_headers:
        public: false
        shared_max_age: 0 # <- this is ignored as not relevant on private resources
```

### DIFF
--- a/src/HttpCache/EventListener/AddHeadersListener.php
+++ b/src/HttpCache/EventListener/AddHeadersListener.php
@@ -76,13 +76,14 @@ final class AddHeadersListener
             $response->setVary(array_diff($this->vary, $response->getVary()), false);
         }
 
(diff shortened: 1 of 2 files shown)


dunglas:
Good catch, thanks for the fix!

PR #3546, opened 2020-05-05
Fix: Limit range of PHP versions

localheinz:
This PR

* [x] limits the range of PHP versions this package can be installed on

💁‍♂️ We don't know yet if this package will be compatible with PHP 9000, do we?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


dunglas:
PHP isn’t using semver and has a strong BC policy. I’m not sure if it is worth it.

PR #3548, opened 2020-05-06
[GraphQL] Allow `id` type in Filter::getDescription()

oleg-brizy:
```php
{
    return [
        'someProperty' => [
            'property' => 'someProperty',
            'type' => Type::nonNull(Type::id()), // <-- error
            'description' => 'MyClassName',
        ]
    ];
}
```

The error I got:

```
Argument 3 passed to Symfony\\Component\\PropertyInfo\\Type::__construct() must be of the type string or null, object given
```

P.S. This is a quick fix and I don't have to to investigate how to make a proper pull request with tests covering the added change (sorry).

| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -380,8 +380,12 @@ final class FieldsBuilder implements FieldsBuilderInterface
 
             foreach ($this->filterLocator->get($filterId)->getDescription($resourceClass) as $key => $value) {
                 $nullable = isset($value['required']) ? !$value['required'] : true;
-                $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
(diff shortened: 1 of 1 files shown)


alanpoulain:
Sorry I don't see why it would be necessary to allow this change.

oleg-brizy:
How to set parameter type on filter then?
I want to add a filter `{"type": "ID!"}` but currently it's possible only basic PHP types like `String`.

alanpoulain:
It's only to handle the `ID` type then? In this case, I would prefer an `id` string in the description of the filter and to convert it to `GraphQLType::id()` in the `FieldsBuilder` (and handling the nullable property).
Also it will not be merged without adding a unit test.

oleg-brizy:
Not only ID, any type.
Anyway I am using [this fork](https://github.com/bagrinsergiu/api-platform-core) in my project.

PR #3552, opened 2020-05-08
Remove URL encoding of identifier

### LINKED ISSUE #1575: Identifiers url-encoded twice

silverbackdan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       |  fixes #1575
| License       | MIT

It looks as though the url encoding was not necessary and caused duplicate encoding in any case. The example I have tested originally returned `%253A` in place of the colon, which is the percent sign of %3A re-encoding.

This fix removes all URL encoding, it doesn't appear necessary in the IRI converter.

Would this class as a BC change instead of a fix though? Is it better to drill down further and still url encode the IRI once?


silverbackdan:
Just to clarify that this should work well. The generated identifier is passed into `->router->generate` as a parameter. The Symfony router should handle url encoding any IDs as necessary.

I've clarified the use case a bit by including a character that should also be encoded in the `@id` key. Either way, it appears you can choose whether or not to encode any or the characters in this instance.

The Behat failing test appears to be unrelated. Please let me know if there's anything else though that I didn't spot.

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -73,6 +73,7 @@ public function normalize($object, $format = null, array $context = [])
 
         $resourceClass = $this->resourceClassResolver->getResourceClass($object, $context['resource_class'] ?? null);
         $context = $this->initContext($resourceClass, $context);
+
```
soyuka, line 76: remove newline : 
```suggestion
```

features/main/url_encoded_id.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Allowing resource identifiers with characters that should be URL encoded
+  In order to have a resource with an id with special characters
+  As a client software developer
+  I need to be able to set and retrieve these resources with the URL encoded ID
+
+  @createSchema
+  Scenario Outline: Get a resource whether or not the id is URL encoded
+    Given there is a UrlEncodedId resource
+    And I add "Content-Type" header equal to "application/ld+json"
+    When I send a "GET" request to "<url>"
```
soyuka, line 10: shouldn't this be `/url_encoded_ids/%encode:id` I don't really understand how `<url>` matches the entity

dunglas:
I also fear that this change will introduce a BC break. We should take a deeper look to the URI RFC. It contains the list of characters that should and shouldn’t be encoded in the path section.

dunglas:
For instance, Go has a `url.PathEscape()` as we’ll as a `url.QueryEscape()` function. They encode different classes of characters. AFAIU, we should use the equivalent of `PathEscape` here because chars such as `/` still have to be encoded.

features/main/url_encoded_id.feature
silverbackdan: It's from the scenario outline variable below. Line 21. So I'm testing many different ways of fetching the entity with and without encoding.

dunglas:
If it fixes a double encoding issue, it looks good to me. We need to double check that this change doesn’t introduce regressions.

features/main/url_encoded_id.feature
soyuka: oh thanks I wasn't aware it was possible :)

soyuka:
Would you be able to target master? It'll be safer to introduce this as a potentially breaking change :)

silverbackdan:
@dunglas I'm not sure if it'd count as a regression, but the `@id` would no longer encode characters such as colons in the JSON output. However, you can still request the resource with the character encoded if you wanted to. The tests I've added go through variations of encoding the characters in the request to make sure the same resource is still returned. The slash is still encoded.

@soyuka sure no problem. thank you :) and you can thank Vincent for teaching me about Scenario Outlines ;)

soyuka:
superseeded by #3611

PR #3560, opened 2020-05-15
Rewrite static access to Inflector as it has been deprecated

### LINKED ISSUE #2626: Doctrine\Common\Util\Inflector is deprecated
can Doctrine\Common\Util\Inflector be removed from 

Bridge/Symfony/Routing/RouteNameGenerator.php
GraphQI/Type/SchemaBuilder.php
Operation/DashPathSegmentNameGenerator.php
Operation/UnderscrorePathSegmentNameGenerator.php

jnugh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| BC breaks?    | `doctrine/inflector` raised to `^1.4 \|\| ^2.0` 
| Deprecations? |no
| Tickets       | fixes #2626
| License       | MIT
| Doc PR        | none

`doctrine/inflector` deprecated the use of `Doctrine\Common\Inflector\Inflector` when they released version 2.0 (where the class has been removed).
They suggest to use a factory pattern like this: https://github.com/doctrine/inflector/blob/master/docs/en/index.rst#usage

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
     "require": {
         "php": ">=7.1",
-        "doctrine/inflector": "^1.0",
+        "doctrine/inflector": "^1.4 || ^2.0",
         "fig/link-util": "^1.0",
         "psr/cache": "^1.0",
         "psr/container": "^1.0",
--- a/src/Annotation/AttributesHydratorTrait.php
+++ b/src/Annotation/AttributesHydratorTrait.php
@@ -14,7 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Annotation;
 
 use ApiPlatform\Core\Exception\InvalidArgumentException;
-use Doctrine\Common\Inflector\Inflector;
+use Doctrine\Inflector\InflectorFactory;
 
 /**
  * Hydrates attributes from annotation's parameters.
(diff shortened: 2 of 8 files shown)


alanpoulain:
Thank you for your PR.
However we still maintain PHP 7.1 for API Platform and Doctrine Inflector 1.4 doesn't allow it.
There are two choices:
- not supporting PHP 7.1 anymore (@api-platform/core-team)
- not bumping to 1.4 but handling the two API (legacy + new)

jnugh:
> not bumping to 1.4 but handling both API (legacy + new)

I could do that by detecting if `Doctrine\Inflector\InflectorFactory` exists and build an abstraction for this, if 7.1 wouldn't be an option.

derrabus:
This PR recreates the inflector everytime it's used while the library's own compat layer maintains an instance for subsequent calls. I haven't measured how expensive the new factory logic actually is but I don't this we should call it everytime we want to inflect something.

I had to fix this issue for one of my projects as well, so I c&p'ed my solution to #3564. Here I'm using a static facade to switch between old and new implementation.

jnugh:
The factory has an internal cache as well, so likely very little overhead. 
But as your PR could be adopted to work with both inflector versions more easily I'll close my PR :smile:

PR #3564, opened 2020-05-16
Handle deprecations from Doctrine Inflector

derrabus:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #2626
| License       | MIT

This is an alternative approach to #3560. Since Doctrine Inflector cannot be bumped to version 1.4 without dropping support for php 7.1, this PR introduces a façade that uses the new API if it's available and falls back to the 1.3 API otherwise.

This change removed a lot of messages from my application's deprecation log.

### DIFF
--- a/src/Bridge/Doctrine/Inflector/Inflector.php
+++ b/src/Bridge/Doctrine/Inflector/Inflector.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Doctrine\Inflector;
+
+use Doctrine\Common\Inflector\Inflector as LegacyInflector;
+use Doctrine\Inflector\Inflector as InflectorObject;
+use Doctrine\Inflector\InflectorFactory;
+
+/**
+ * Facade for Doctrine Inflector.
+ *
+ * This class allows us to maintain compatibility with Doctrine Inflector 1.3 and 2.0 at the same time.
+ *
+ * @internal
+ */
+final class Inflector
+{
+    /**
+     * @var InflectorObject
+     */
+    private static $instance;
+
+    private static function getInstance(): InflectorObject
+    {
+        return self::$instance
+            ?? self::$instance = InflectorFactory::create()->build();
+    }
+
+    /**
+     * @see LegacyInflector::tableize()
+     */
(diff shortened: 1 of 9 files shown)


alanpoulain:
Don't you think it would be nice to use an interface to change easily the implementation (for instance if we want to use Symfony String)?

derrabus:
> Don't you think it would be nice to use an interface to change easily the implementation (for instance if we want to use Symfony String)?

First of all, you could very well hide Symfony String behind the façade I'm sugesting here as well.

Yes, I think an interface would be nice. But in order to use such an interface effectively, the inflector shouldn't be called statically but injected via DI instead. That would be a larger change. I'd suggest wait with that change until php 7.1 support is dropped so we actually have access to Symfony String or the non-static Doctrine Inflector.

My goal was to provide a minimal solution, so a bugfix release could be shipped that does not trigger deprecation warnings anymore.

jnugh:
The factory has an internal cache as well, so likely very little overhead. 
But as your PR could be adopted to work with both inflector versions more easily I'll close my PR :smile:

src/Bridge/Doctrine/Inflector/Inflector.php
```diff
@@ -0,0 +1,55 @@
...
+    private static $instance;
+
+    private static function getInstance(): InflectorObject
+    {
+        return self::$instance
+            ?? self::$instance = InflectorFactory::create()->build();
+    }
+
+    /**
+     * Converts a word into the format for a Doctrine table name. Converts 'ModelName' to 'model_name'.
```
soyuka, line 41:
```suggestion
     * @see LegacyInflector::tableize()
```

src/Bridge/Doctrine/Inflector/Inflector.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public static function tableize(string $word): string
+    {
+        return class_exists(InflectorFactory::class) ? self::getInstance()->tableize($word) : LegacyInflector::tableize($word);
+    }
+
+    /**
+     * Returns a word in plural form.
```
soyuka, line 49:
```suggestion
     * @see LegacyInflector::pluralize()
```

soyuka:
> Don't you think it would be nice to use an interface to change easily the implementation (for instance if we want to use Symfony String)?

Definitely agree on this but this would be considered as a new feature as we'd need to change more code and add some dependency injection. We need to turn our tests to green and to keep on supporting php 7.1 for now.

derrabus:
> I like this solution for now, would you be able to rebase fix my comments and rebase?

Done.

dunglas:
Thank you very much @derrabus!

PR #3572, opened 2020-05-18
Remove PHP 7.1 from CI

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,7 +146,6 @@ jobs:
     strategy:
       matrix:
         php:
-          - '7.1'
           - '7.2'
           - '7.3'
           - '7.4'
@@ -164,7 +163,6 @@ jobs:
             bash \
             unzip \
       - name: Cache mongodb PHP extension build
-        if: (!startsWith(matrix.php, '7.1'))
         uses: actions/cache@v1
         with:
           path: /var/tmp/build/ext-mongodb-${{ env.EXT_MONGODB_VERSION }}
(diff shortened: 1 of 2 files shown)


alanpoulain:
@dunglas could you remove 7.1 from required status check please?

dunglas:
But why? We still support PHP 7.1 isn't it?

alanpoulain:
We still support it, but it causes test issues because of the new Doctrine Inflector (BC changes and PHP > 7.1).
Also, I think it would be a good thing to drop it for 2.6 (because this version is not maintained anymore and some libs don't support it like the Doctrine ones).

dunglas:
Can't we add a BC layer to support older versions of Doctrine Inflector? Or aren't these versions maintained? That being said, OK for me to trop support for PHP 7.1 in 2.6, but we should keep the tests for 2.5.

Btw, at some point I would like to switch from Doctrine Inflector to Symfony String).

alanpoulain:
Yes, we can add a BC layer (that's what it has been done here: https://github.com/api-platform/core/pull/3564).
However some tests cannot be kept (mainly because the plural of `taxon` was `taxons` before 1.4 and `taxa` after). That's why I wanted to remove PHP 7.1 from the tests but keep the compatibility anyway for API Platform 2.5. But not for 2.6.

flug:
Have you tried this? https://github.com/doctrine/inflector/issues/147
```
"conflict": {
        "doctrine/inflector": "^1.4"
    },
```

soyuka:
let's rebase this onto master, wdyt @alanpoulain ?

roukmoute:
> Doctrine Inflector to Symfony String

Why @dunglas?

dunglas:
To prevent this kind of issues.

Issue #3473, opened 2020-03-31
Exception when using Ramsey/Uuid ^4.0


WhiteRabbitDE:
**API Platform version(s) affected**: x.y.z

**Description**  
After updating to **Ramsey/Uuid ^4.0** i get an exception when querying an item with identifier. For example  api/books/3fa85f64-5717-4562-b3fc-2c963f66afa6

The exception:
`Argument 1 passed to Ramsey\Uuid\Uuid::fromString() must be of the type string, object given, called in ..\vendor\api-platform\core\src\Bridge\RamseyUuid\Identifier\Normalizer\UuidNormalizer.php on line 35`

$data contains an Uuid object instead of an string , the Uuid is my identifier.

It works fine with **Ramsey/Uuid ^3.9** !

**Possible Solution**  
Add a check if $data is a string or an instance of UuidInterface in the UuidNormalizer

soyuka:
You may not need this normalizer after all, could you trying removing it to see if everything works as expected?

WhiteRabbitDE:
> You may not need this normalizer after all, could you trying removing it to see if everything works as expected?

All my GET-Item Operations are working without the Normalizer. My POST test is failing now 

`Uncaught PHP Exception Symfony\Component\Serializer\Exception\MissingConstructorArgumentsException: "Cannot create an instance of "Ramsey\Uuid\Uuid" from serialized data because its constructor requires parameter "fields" to be present`

soyuka:
My guess is that the denormalizer can't use the Uuid class as-is. Would you be able to add a denormalizer for the Uuid type (see https://symfony.com/doc/current/serializer/custom_normalizer.html#creating-a-new-normalizer) to see if it fixes it.

WhiteRabbitDE:
It works if i adjust the UuidNormalizer with the check i provided as "Possible solution". 
I changed https://github.com/api-platform/core/blob/6cdf4e044dee9781b41907e59b326395f818e59c/src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php#L35 
to
` return ($data instanceof UuidInterface) ? $data : Uuid::fromString($data);`

That fixes the problem and should not impact older versions of ramsey. 

@soyuka Im not sure what denormalizer i should add. The existing one already supports the UuidInterface. Can you give me a hint ?

soyuka:
There is a difference between the Identifier normalizer and the denormalizer I suggested. In fact what bother's me is that we already have:

```
[...]

PR #3578, opened 2020-05-26
fix: Uuid identifier normalizer should support only strings

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | na
| License       | MIT
| Doc PR        | na

The supports method wasn't explicit enough and should not be called with something other then a string.


guilliamxavier:
Refs #3473

Issue #3542, opened 2020-05-04
[GraphQl] OneToMany without Doctrine


MartkCz:
**API Platform version(s) affected**: current dev-master

```php
/**
 * @ApiResource()
 */
final class Timeline
{

	/**
	 * @var PostUnion[]|PaginatorInterface
	 */
	public iterable $posts;

}
```

```php
/**
 * @ApiResource()
 */
final class PostUnion
{

	public string $type;

	public ?string $subject;

	public string $content;

}
```

I'm getting IterableConnection instead of PostUnionConnection.

The goal is:

```php
$timeline = new Timeline();
$timeline->posts = new ApiPlatform\Core\Bridge\Doctrine\Orm\Paginator($postPaginator);

return $timeline;
```

Any suggestion or it's bug?

mxmp210:
Use PostUnion[] | Collection as variable type.

You will also need to hydrate collectionand return it with ArrayPaginator instance from your data provider in order to work with GraphQL as simple arrays are not supported with graphql for pagination purposes.

alanpoulain:
Fixed by https://github.com/api-platform/core/pull/3591.

PR #3591, opened 2020-06-02
Better handling of relations without Doctrine.

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/core/issues/3542
| License       | MIT
| Doc PR        | N/A

When Doctrine is not used, the built-in type of a relation is an array
and not an object.


xfifix:
this solution didn't fix totally the problem i have always an array empty  in the result

PR #3594, opened 2020-06-03
[GraphQL] Fix subresource DTO

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/api-platform/issues/1528
| License       | MIT
| Doc PR        | N/A

Fix subresource DTO by using the real class for the identifiers.
The class is retrieved by using the `ITEM_RESOURCE_CLASS_KEY` set by the normalizer.


d3bug-git:
hello @alanpoulain thank you for the subresource DTO fix, but when id do my request i have void array in recipe.
image

d3bug-git:
hello @alanpoulain thank you for subresource DTO fix, however when i change the ApiProperty indentifier in RecipeCategory
image
and in Recipe
image
the result of the recipes is void in my query
image

.I see the problem come from https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php#L94
because in https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/SubresourceDataProvider.php#L117
the dump of $query in my case is 
image
but i think IDENTITY(o.recipeCategory) give the integer foreign key it is not uuid .
thank you in advance for helping me

PR #3599, opened 2020-06-15
#3577 moved content negotiation to before firewall

### LINKED ISSUE #3577: Content Negotiation should be performed before Authentication / FirewallListener
**API Platform version(s) affected**: 2.5.5

**Description**  
Currently, `AddFormatListener` (priority 7) is executed after `FirewallListener` (priority 8). See registered `kernel.request` event listeners. This means the Security system cannot rely on `Symfony\HttpFoundation\Request::getRequestFormat()` as it has not yet been initialized.

> When setting a response for the kernel.request event, the propagation is stopped. This means listeners with lower priority won't be executed.

**How to reproduce**  
What if we had a `Symfony\Component\Security\Http\EntryPoint\AuthenticationEntryPointInterface` which relied on the request format to decide if it should return a HTTP 302 (see https://tools.ietf.org/html/rfc6749#section-4.1) redirect a user to login. But also supports https://tools.ietf.org/html/rfc6750#section-3 - notify an API client they must authenticate with a Bearer token?

```php
    public function start(
        HttpFoundation\Request $request,
        AuthenticationException $authException = null
    ): HttpFoundation\Response {
        if ($request->getRequestFormat() === 'html') {
            /**
             * @see https://tools.ietf.org/html/rfc6749#section-4.1.1
             */
            $state = sha1(openssl_random_pseudo_bytes(1000));
            $stateKey = AuthorizationCodeAuthenticator::getStateKey($state);
            $this->cache->save($stateKey, $state, 300);

            $uri = $this->router->generate('auth_client_authorization_v2_authorize', [
                'response_type' => 'code',
                'client_id' => $this->oAuthClientId,
                'state' => $state,
            ], UrlGeneratorInterface::ABSOLUTE_URL);

            return new HttpFoundation\RedirectResponse($uri);
        } else {
            /**
             * @see https://tools.ietf.org/html/rfc6750#section-3
             */
            $data = [
                'message' => 'Authentication Required',
            ];
            $wwwAuthenticateHeader = sprintf(
                'Bearer realm="%s"',
                $this->realm
            );
            if (
                $request->headers->has('Authorization')
                && 'Bearer' === substr($request->headers->get('Authorization'), 0, 6)
            ) {
                $wwwAuthenticateHeader .= ', error="invalid_token"';
            }
            return new HttpFoundation\JsonResponse($data, HttpFoundation\Response::HTTP_UNAUTHORIZED, [
                'WWW-Authenticate' => $wwwAuthenticateHeader,
            ]);
        }
    }
```

**Possible Solution**  
Move `AddFormatListener` to a priority `> 8`

**Additional Context**  

https://github.com/api-platform/core/commit/3b3f10df67fd3c96960cd4f57150f6240c0bff14#r39354527

> but we must double check that it has no bad side effects in term of security before changing this.

The side effects of running `ApiPlatform\Core\EventListener\AddFormatListener` are:

* the `Request::$formats` are registered (Map from format names to mime types e.g. `'json' => ['application/json']`)
* the `Request::$format` is set (The format of the current request)
* A `NotAcceptableHttpException` could be thrown (I think this is a good thing, content negotiation is generally cheaper than authentication)

When priority is `< 8` the `Request::$format` is left `null` (unless manipulated outside of `api-platform/core` by another higher priority listener).

Since event dispatcher priorities are not strictly defined by Symfony, we should look to other projects for consensus on a good priority to use for `AddFormatListener`.

We should start a list of notable bundles that register `kernel.request` event listeners which might affect the chosen priority for `AddFormatListener`. Interoperability with other libraries is an important consideration.

Example order of event listeners in a standard symfony 5 project:

| Listener                                                                                       | Priority |
| ---------------------------------------------------------------------------------------------- | --------:|
| `Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure()`                |     2048 |
| `Symfony\Component\HttpKernel\EventListener\ValidateRequestListener::onKernelRequest()`        |      256 |
| `Symfony\Component\HttpKernel\EventListener\SessionListener::onKernelRequest()`                |      128 |
| `Symfony\Component\HttpKernel\EventListener\LocaleListener::setDefaultLocale()`                |      100 |
| `Symfony\Component\HttpKernel\EventListener\RouterListener::onKernelRequest()`                 |       32 |
| `Symfony\Component\HttpKernel\EventListener\LocaleListener::onKernelRequest()`                 |       16 |
| `Symfony\Component\HttpKernel\EventListener\LocaleAwareListener::onKernelRequest()`            |       15 |
| `Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener::configureLogoutUrlGenerator()` |        8 |
| `Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener::onKernelRequest()`             |        8 |
| `Sentry\SentryBundle\EventListener\RequestListener::onKernelRequest()`                         |        1 |
| `Sentry\SentryBundle\EventListener\SubRequestListener::onKernelRequest()`                      |        1 |

kralos:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3577
| License       | MIT
| Doc PR        | 

Moved content negotiation to before firewall to allow content aware security responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 ## 2.7.0
 
+* **BC**: Change `api_platform.listener.request.add_format` priority from 7 to 28 to execute it before firewall (priority 8) (#3599)
 * Doctrine: Better exception to find which resource is linked to an exception (#3965)
 * Doctrine: Allow mixed type value for date filter (notice if invalid) (#3870)
+* Doctrine: Add `nulls_always_first` and `nulls_always_last` to `nulls_comparison` in order filter (#4103)
 * MongoDB: `date_immutable` support (#3940)
 * DataProvider: Add `TraversablePaginator` (#3783)
 * Swagger UI: Add `swagger_ui_extra_configuration` to Swagger / OpenAPI configuration (#3731)
-* OrderFilter: Add `nulls_always_first` and `nulls_always_last` to `nulls_comparison` (#4103)
 
 ## 2.6.3
 
--- a/features/filter/filter_validation.feature
+++ b/features/filter/filter_validation.feature
@@ -1,5 +1,8 @@
 Feature: Validate filters based upon filter description
 
+  Background:
+    Given I add "Accept" header equal to "application/json"
+
   @createSchema
   Scenario: Required filter should not throw an error if set
     When I am on "/filter_validators?required=foo&required-allow-empty=&arrayRequired[foo]="
@@ -13,7 +16,7 @@ Feature: Validate filters based upon filter description
   Scenario: Required filter should throw an error if not set
     When I am on "/filter_validators"
     Then the response status code should be 400
-    Then the JSON node "detail" should match '/^Query parameter "required" is required\nQuery parameter "required-allow-empty" is required$/'
+    And the JSON node "detail" should match '/^Query parameter "required" is required\nQuery parameter "required-allow-empty" is required$/'
 
   Scenario: Required filter should not throw an error if set
     When I am on "/array_filter_validators?arrayRequired[]=foo&indexedArrayRequired[foo]=foo"
--- a/features/main/content_negotiation.feature
+++ b/features/main/content_negotiation.feature
@@ -22,7 +22,7 @@ Feature: Content Negotiation support
     <response><description/><dummy/><dummyBoolean/><dummyDate/><dummyFloat/><dummyPrice/><relatedDummy/><relatedDummies/><jsonData/><arrayData/><name_converted/><relatedOwnedDummy/><relatedOwningDummy/><id>1</id><name>XML!</name><alias/><foo/></response>
     """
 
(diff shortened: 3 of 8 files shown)


kralos:
CI appears to be failing the same on branch `2.5` (base)

I found a bug in `features/filter/filter_validation.feature`

This is effectively using the default format (`'application/ld+json'`) because it's the first format in `tests/Fixtures/app/config/config_common.yml` `api_platform.formats`

dunglas:
Shouldn't we explicitly set the `Accept` and header instead of changing the test?
Also could you add a specific test covering your change to prevent any future regression please?

alanpoulain:
Thanks @kralos. Hopefully it will be not a BC break too much.

PR #3611, opened 2020-06-18
Fix/identifier encode

### LINKED ISSUE #1575: Identifiers url-encoded twice

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       |  fixes #1575
| License       | MIT

It looks as though the url encoding was not necessary and caused duplicate encoding in any case. The example I have tested originally returned `%253A` in place of the colon, which is the percent sign of %3A re-encoding.

This fix removes all URL encoding, it doesn't appear necessary in the IRI converter.

Would this class as a BC change instead of a fix though? Is it better to drill down further and still url encode the IRI once?

From https://github.com/api-platform/core/pull/3552

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * OpenAPI: Add PHP default values to the documentation (#2386)
 * Deprecate using a validation groups generator service not implementing `ApiPlatform\Core\Bridge\Symfony\Validator\ValidationGroupsGeneratorInterface` (#3346)
 * Subresources: subresource resourceClass can now be defined as a container parameter in XML and Yaml definitions
-* Url Encoded IRIs: Fix IRI url encoding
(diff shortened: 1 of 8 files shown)


silverbackdan:
Thanks for taking over this PR @soyuka !

soyuka:
thanks @silverbackdan !

Kitton:
@soyuka do you think we can have a release with this fix included?

silverbackdan:
Version 2.6 when released will include this fix - for now you could use version `2.6.x-dev` and it'll include this update. See the change log here for other possible bc breaks at the time of installing though https://github.com/api-platform/core/blob/master/CHANGELOG.md

PR #2545, opened 2019-02-20
Fixed maxDepth handling for selfreferencing subresources

ghry5:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2533 
| License       | MIT
| Doc PR        |

Issue #2533, opened 2019-02-19
[Bug] @ApiSubresource maxDepth ignored for self-referencing subresources.


ghry5:
I have the following class:
```
class Person {
  //@ApiSubresource(maxDepth=1)
  private $address;
  //@ApiSubresource(maxDepth=1)
  private $boss;
  //@ApiSubresource(maxDepth=1)
  private $subordinates;
}
```
`$boss` and `$subordinates` are references to the same Person entity (bidirectional oneToMany) and `$address` references an Address entity. Due to setting maxDepth=1 on all three i would expect the following three subresources:

- /people/{id}/address
- /people/{id}/boss
- /people/{id}/subordinates

The actual subresources are:

- /people/{id}/address
- /people/{id}/boss
- /people/{id}/boss/address
- /people/{id}/boss/subordinates
- /people/{id}/boss/subordinates/address
- /people/{id}/subordinates
- /people/{id}/subordinates/{subordinates}/address
- /people/{id}/subordinates/{subordinates}/boss
- /people/{id}/subordinates/{subordinates}/boss/address

The culprit seems to be the following in `SubresourceOperationFactory`
```
            // Handle maxDepth
            if ($rootResourceClass === $resourceClass) {
                $maxDepth = $subresource->getMaxDepth();
                // reset depth when we return to rootResourceClass
                $depth = 0;
            }
            if (null !== $maxDepth && $depth >= $maxDepth) {
                break;
            }
            ...

            $this->computeSubresourceOperations($subresourceClass, $tree, $rootResourceClass, $operation, $visited + [$visiting => true], ++$depth, $maxDepth);
```

Since `$rootResourceClass` equals `$resourceClass` (as $boss and $subordinates refer to the same entity) the 'current' $depth will get reset every time.

I propose the following solution:
```
      if (null === $rootResourceClass) {
            $rootResourceClass = $resourceClass;
            // initialize depth to 0 on root level
            $depth=0;
        }
        ...

            // Handle maxDepth
            // get the minimum maxDepth from the rootMaxDepth and the maxDepth of the current to be visited Subresource
            $currentMaxDepth = array_diff([$maxDepth, $subresource->getMaxDepth],[null]));
            $currentMaxDepth = empty($currentMaxDepth) ? null : min($currentMaxDepth);

            if (null !== $currentMaxDepth && $depth >= $currentMaxDepth) {
                break;
            }
            ...

[...]

PR #3616, opened 2020-06-22
Changed depth and maxDepth handling for selfreferencing subresourceOp…

soyuka:
…erations.

Addresses: https://github.com/api-platform/core/issues/2533.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #2533 
| License       | MIT
| Doc PR        | na

this is a re-do of https://github.com/api-platform/core/pull/2545

### DIFF
--- a/src/Operation/Factory/SubresourceOperationFactory.php
+++ b/src/Operation/Factory/SubresourceOperationFactory.php
@@ -64,6 +64,7 @@ final class SubresourceOperationFactory implements SubresourceOperationFactoryIn
     {
         if (null === $rootResourceClass) {
             $rootResourceClass = $resourceClass;
+            $depth = 0;
         }
 
         foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $property) {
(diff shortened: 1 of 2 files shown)


src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -183,7 +178,11 @@ private function computeSubresourceOperations(string $resourceClass, array &$tre
 
             $tree[$operation['route_name']] = $operation;
 
-            $this->computeSubresourceOperations($subresourceClass, $tree, $rootResourceClass, $operation, $visited + [$visiting => true], ++$depth, $maxDepth);
+            // get the minimum maxDepth from the rootMaxDepth and the maxDepth of the to be visited Subresource
```
nightio, line 181: If I understand well, if in class A has (subresource B with maxDepth=5), and B has (subresource C with maxDepth=1), finally only 2 levels will be generated even C has another subresource (and so on)?
If so it would be good to add some more comment to clarify.

src/Operation/Factory/SubresourceOperationFactory.php
```diff
@@ -183,7 +178,11 @@ private function computeSubresourceOperations(string $resourceClass, array &$tre
...
+            $currentMaxDepth = array_diff([$maxDepth, $subresource->getMaxDepth()], [null]);
```
nightio, line 182: optionally array_filter([$maxDepth, $subresource->getMaxDepth()], 'is_int')

soyuka: Indeed, is way more performant thx

soyuka:
thanks @ghry5 and @rkopera

PR #3617, opened 2020-06-22
Add messenger=persist

### LINKED ISSUE #3371: [Proposal] Allow array in messenger attribute
**Description**

Using `messenger=true` attribute on a Doctrine entity disable the persist/remove of the object to respect the CQRS, it must be done asynchronously. In some projects, it's required to save the object synchronously before sending the object.

This cannot be done in an EventSubscriber according to [Extending API Platform](https://api-platform.com/docs/core/extending/) section, and must be done in a custom data persister. This behavior could be automatized in API Platform using a new value in the `messenger` attribute.

Current values are:
* `messenger=true`: dispatching a Resource through the Message Bus
* `messenger=input`: using Messenger with an Input Object

Additional values could be:
* `messenger="persist"`: saving an object and dispatching a Resource through the Message Bus
* `messenger=["persist", "input"]`: saving an Input Object and using Messenger with it

**Example**

```php
<?php

// api/src/Entity/Content.php

namespace App\Entity;

/**
 * @ORM\Entity
 * @ApiResource(messenger="persist")
 */
class Content
{
    // ...
}
```

```php
<?php

// api/src/Entity/Content.php

namespace App\Entity;

use App\Dto\ContentRequest;

/**
 * @ORM\Entity
 * @ApiResource(
 *     collectionOperations={
 *         "post"={"status"=202}
 *     },
 *     itemOperations={},
 *     messenger=["persist", "input"],
 *     input=ContentRequest::class,
 *     output=false
 * )
 */
class Content
{
    // ...
}
```

**Proposal**

In order to handle this new feature, the MessengerDataPersister should be updated to support the additional values: `save` or an array. If the attribute value matches `save` or contains it, then the ChainDataPersister must be called again with a context option to prevent an infinite loop. Otherwise, the MessengerDataPersister works as expected, following the CQRS pattern.
<br><br><br><br><br><br>
Sorry for the long description. Here is an API potato:

image

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3371
| License       | MIT
| Doc PR        | api-platform/docs#1132

- [x] Support for "persist" and array value on _messenger_ attribute
- [x] Update unit tests
- [x] Add functional tests
- [x] Add doc PR


src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -98,4 +86,33 @@ public function remove($data, array $context = [])
...
+                    false,
+                    true
+                );
+        }
+
+        if (isset($context['graphql_operation_name'])) {
+            return $resourceMetadata->getGraphqlAttribute($context['graphql_operation_name'], 'messenger', false, true);
+        }
+
+        return $resourceMetadata->getAttribute('messenger', false);
```
soyuka, line 116: I think we can safely return `false` here as we're already using the `resourceFallback` in the getAttributes methods above.

src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -98,4 +86,37 @@ public function remove($data, array $context = [])
                 ->with(new RemoveStamp())
         );
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function loop($data, array $context = []): bool
```
alanpoulain, line 93: Why "loop"?

src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -98,4 +86,37 @@ public function remove($data, array $context = [])
...
+        try {
+            $value = $this->getMessengerAttributeValue($this->resourceMetadataFactory->create($context['resource_class'] ?? $this->getObjectClass($data)), $context);
+        } catch (ResourceClassNotFoundException $exception) {
+            return false;
+        }
+
+        return 'persist' === $value || (\is_array($value) && \in_array('persist', $value, true));
+    }
+
+    private function getMessengerAttributeValue(ResourceMetadata $resourceMetadata, array $context = [])
```
alanpoulain, line 104: Missing typehint?

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
```diff
@@ -78,6 +87,10 @@ private function tracePersisters($data, array $context = [])
             }
         }
 
-        return $match;
+        foreach ($this->persisters as $persister) {
```
alanpoulain, line 90: Why not yielding in the previous loop?

tests/Fixtures/TestBundle/Entity/MessengerWithArray.php
```diff
@@ -0,0 +1,44 @@
...
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiProperty;
+use ApiPlatform\Core\Annotation\ApiResource;
+use ApiPlatform\Core\Tests\Fixtures\TestBundle\Dto\MessengerInput;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ApiResource(messenger={"persist", "input"}, input=MessengerInput::class)
+ * @ORM\Entity
```
alanpoulain, line 23: You need to introduce MongoDB documents too.

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
alanpoulain, line 90: I think the `persistersResponse` is not coherent anymore too.

src/Bridge/Symfony/Messenger/DataPersister.php
vincentchalamon: No cause the typehint could be boolean, string or array

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
vincentchalamon: That's why there are 2 loops in the `tracePersisters` method, to keep the `persistersResponse` coherent

src/Bridge/Symfony/Messenger/DataPersister.php
alanpoulain: Missing PHPDoc then :stuck_out_tongue:

src/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersister.php
alanpoulain: Seems weird to me. The `persistersResponse` is telling that only one persister matches whereas there are more than one.

vincentchalamon: The `tracePersisters` method originally looped on all persisters to store true/false/null value (depending if it supports or not), and returned the first matched persister. The `persistersResponse` should store an array of all persisters looped. For example, with 5 persisters but only the second supports: `[false, true, null, null, null]`

Now I need to handle multiple persisters (thanks to the LoopDataPersisterInterface), but I need to keep the same behavior for the `persistersResponse` array. So if the 2nd persister supports and loops, and the 4th supports (but not the 3rd), it should result something like: `[false, true, false, true, null]`

src/Bridge/Symfony/Messenger/DataPersister.php
```diff
@@ -98,4 +86,40 @@ public function remove($data, array $context = [])
...
+     */
+    public function loop($data, array $context = []): bool
+    {
+        try {
+            $value = $this->getMessengerAttributeValue($this->resourceMetadataFactory->create($context['resource_class'] ?? $this->getObjectClass($data)), $context);
+        } catch (ResourceClassNotFoundException $exception) {
+            return false;
+        }
+
+        return 'persist' === $value || (\is_array($value) && \in_array('persist', $value, true));
```
dunglas, line 101: Maybe could we use an array (`["persist"=>true]`) to allow passing more options later?

src/DataPersister/LoopDataPersisterInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\DataPersister;
+
+/**
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
+ */
+interface LoopDataPersisterInterface
```
dunglas, line 19: I’m not a fond of this name, can we found something more explicit such as “StopableDataPersister”, “TransientDataPersister” or “NonFinalDataPersister” or “NonLastDataPersister”?

Also, do we really need this new interface? Can’t we just transform the MessengerDataPersister into a decorator and call the decorated service is persist is set to true? It will simplify the code a lot.

tests/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersisterTest.php
```diff
@@ -87,6 +88,11 @@ public function persist($data)
                     public function remove($data)
                     {
                     }
+
+                    public function loop($data, array $context = []): bool
```
dunglas, line 92: `stop()` ?

src/Bridge/Symfony/Messenger/DataPersister.php
vincentchalamon: I don't like the idea of adding an option "_just in case we need it later_"

src/DataPersister/LoopDataPersisterInterface.php
vincentchalamon: The interface helps to continue to the next data provider from the ChainDataProvider, and prevents any direct relation between the ChainDataProvider and the Messenger DataProvider: https://github.com/api-platform/core/pull/3617/files#diff-4c062f54ee87035a3b6a2a1689356789R61

>I’m not a fond of this name, can we found something more explicit such as “StopableDataPersister”, “TransientDataPersister” or “NonFinalDataPersister” or “NonLastDataPersister”?

For code quality and to prevent any useless code complexity, I would prefer something "positive", instead of "stop", "non-...", etc.
I'll try "transient" as you suggested, ok for you too @soyuka @alanpoulain ?

src/Bridge/Symfony/Messenger/DataPersister.php
dunglas: We're already adding an option :) It's important to add it in a future-proof way or we'll have to deprecate it and maintain a BC layer at some point, and it will hurt users as well as us.

vincentchalamon: So should I support all those formats? `'persist'`, `['persist']`, `['persist' => true]`

dunglas: Only the array IMHO, to make the API simpler.

vincentchalamon: But that would be weird if I use it in addition to `input` data: `['persist' => true, 'input']`

soyuka:
Thanks @vincentchalamon !

PR #3626, opened 2020-06-30
PHPUnits `expectExceptionMessageRegExp` becomes deprecated in Phpunit 9

wRLSS:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes
| Tickets       | -//- 
| License       | MIT
| Doc PR        | -//-

PHPUnit `expectExceptionMessageRegExp` becomes deprecated in PHPUnit 9, we should promote `expectExceptionMessageMatches` instead.


dunglas:
It looks like it breaks our test suite. You should probably bump the minimum version of `symfony/phpunit-bridge` to be sure that symfony/symfony#36408 is install even when using `composer --prefer-lowest`.

dunglas:
Will be fixed by #3688

PR #3627, opened 2020-06-30
Allow guzzle 7 in require dev section

gmponos:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| License       | MIT

Allows guzzle 7 version in require-dev

Left guzzle 6 since from what I see there is a prefer-lowest in ci.


dunglas:
Could you rebase to target the 2.5 branch please? If no code change is required, it would be nice to run our test suite with Guzzle 6 and 7 even on the stable branche.

alanpoulain:
Already done.

PR #3629, opened 2020-07-01
Check change tracking deferred for ODM

Jean-Beru:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        | 

`DataPersister` checks change tracking policy before persisting it. It works only with ORM since ODM metadata class is not the same. The aim of this PR is to fix it.


soyuka:
I rebased your pull request but the phpunit test fail https://github.com/api-platform/core/pull/3629/checks?check_run_id=902984718 is related to this code

Jean-Beru:
Good catch. `ClassMetadata::isChangeTrackingDeferredExplicit` does not exist in this ODM version. So, I've just added a check in my test.

alanpoulain:
Thank you @Jean-Beru!

PR #3630, opened 2020-07-01
Adding an ErrorHandler to GraphQL

ollietb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

Errors thrown from the platform are caught by GraphQL and can't be logged or tracked in production environments, making it impossible to track down the cause of errors or use error analysers. This PR allows the errors from GraphQL to be handled by the developers by decorating the ErrorHandler. The errors can either logged (though a service like monolog) or sent to a 3rd party error analyser, like Sentry or NewRelic.

I realise it might conflict with PR #3063 but I wanted to try and get it into this branch as it seems like quite a big oversight that could seriously jeopardize the stability of live environments.


alanpoulain:
Thanks for this PR.
However I think it should target `master` and take into account https://github.com/api-platform/core/pull/3063 indeed. Even if it's a needed feature, it's still a new feature.

alanpoulain:
Superseded by https://github.com/api-platform/core/pull/3632.

PR #3635, opened 2020-07-08
Lazy console commands

vyacheslav-startsev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Type       | improvement

Since Symfony 3.4 console commands can be lazy thanks to a [static name property](https://github.com/symfony/symfony/pull/23887).
This change does not break BC because the BC layer is already [in the base command's constructor](https://github.com/symfony/symfony/pull/23887/files#diff-8e80595663798837db1b033187835535R75).


vyacheslav-startsev:
> Good catch? Can you just run `php-cs-fixer` please?

done

soyuka:
thanks @vyacheslav-startsev

PR #3641, opened 2020-07-17
Fix whitespace in namespace

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Fix whitespace in namespace token.
If the RFC at https://wiki.php.net/rfc/namespaced_names_as_token gets accepted, then this will cause a parse error in PHP 8.


alanpoulain:
Thanks!

PR #3664, opened 2020-08-05
Rework to improve and simplify identifiers management

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Tickets       | rework of https://github.com/api-platform/core/pull/2126
| License       | MIT
| Doc PR        | na

Basically my idea was to specify on the operation what it's identifiers are. This helps a lot with identifiers parsing etc. 

The code speaks for itself. I still need to find a solution for composite identifiers, for example we can deprecate them but keep an option:

```
@ApiResource(identifiedBy='id')
```

Would work and force a single identifier in the route path even if there were several.


nicodmf:
+++ 1 @soyuka can i help you ?

soyuka:
> Very important feature +++ 1 @soyuka can i help you ?

There are several issues with this:
- how do we keep today's composite identifiers (I have a solution for this) 
- `IdentifiersExtractorInterface` should not be added to the ApiLoader and IMO identifiers should be within the Metadata (this results in a recursive dependency injection) not yet a solution but I need to dig into it
- this improves subresources a lot, we're going to talk about this with core contributors by the end of the month 

I'll let you know when and if we're testing it would love feedbacks. Thanks for the help.

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -221,6 +224,10 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
             }
         }
 
+        if (!isset($operation['identifiedBy'])) {
+            $operation['identifiedBy'] = null !== $this->identifiersExtractor ? $this->identifiersExtractor->getIdentifiersFromResourceClass($resourceClass) : ['id'];
+        }
```
soyuka, line 229: this is wrong IMO and should be computed by the ResourceMetadataFactory, do you have an idea on how to use `PropertyMetadataFactory` inside a `ResourceMetadataFactory` @dunglas  ?

features/json/plain_identifier.feature
```diff
@@ -0,0 +1,85 @@
+# Feature: JSON relations with plain identifiers support
```
alanpoulain, line 1: Shouldn't it be removed?

features/main/custom_identifier.feature
```diff
@@ -12,6 +12,7 @@ Feature: Using custom identifier on resource
       "name": "My Dummy"
     }
     """
+    Then print last JSON response
```
alanpoulain, line 15: To remove.

features/main/operation.feature
```diff
@@ -63,3 +63,23 @@ Feature: Operation support
   Scenario: Get a 404 response for the disabled item operation
     When I send a "GET" request to "/disable_item_operations/1"
     Then the response status code should be 404
+
+  @createSchema
+  Scenario: Get a book by it's ISBN
```
alanpoulain, line 68:
```suggestion
  Scenario: Get a book by its ISBN
```

src/Bridge/Doctrine/MongoDbOdm/ItemDataProvider.php
```diff
@@ -64,19 +63,13 @@ public function supports(string $resourceClass, string $operationName = null, ar
      *
      * @throws RuntimeException
      */
-    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = [])
+    public function getItem(string $resourceClass, $identifiers, ?string $operationName = null, array $context = [])
```
alanpoulain, line 66: Not typehinted to avoid BC? Maybe add a typehint comment for 3.0.

src/Bridge/Symfony/Bundle/ApiPlatformBundle.php
```diff
@@ -49,5 +50,6 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new GraphQlQueryResolverPass());
         $container->addCompilerPass(new GraphQlMutationResolverPass());
         $container->addCompilerPass(new MetadataAwareNameConverterPass());
+        // $container->addCompilerPass(new IdentifiedByResourceMetadataFactoryPass());
```
alanpoulain, line 53: ?

src/Bridge/Doctrine/MongoDbOdm/ItemDataProvider.php
soyuka: Indeed I didn't touch the interfaces yet but I can update them, this needs to target 3.0 though.

alebedev80:
@soyuka when it will be released?

soyuka:
@alebedev80 https://github.com/api-platform/core/pull/3825#issuecomment-765254256

OskarStark:
Hi @soyuka,

sorry but my code is breaking (only in CI 😄 ) since 2.6 release. Unfortunately there isn't an UPGRADE guide, so I have to ask.

I get the following error:
```
45) App\Tests\Integration\Entity\BenutzerTest::canBeSerializedWithBenutzerUpdateSerializationGroup

ApiPlatform\Core\Exception\InvalidArgumentException: Unable to generate an IRI for "App\Entity\Benutzer".

Caused by
Symfony\Component\Routing\Exception\MissingMandatoryParametersException: Some mandatory parameters are missing ("id") to generate a URL for route "api_benutzers_get_item".
```

So I think I am on the right track with this PR/feature.

The object is the only one which does not use `$id` as identifier, but `$objectGuid`:
```php
<?php

namespace App\Entity;

/**
 * @API\ApiResource(
 *     collectionOperations={
 *         "get": {
 *             "method": "GET",
 *             "path": "/benutzer.{_format}",
 *             "normalization_context": {"groups": {"benutzer_collection_output"}},
 *         },
 *     },
 *     itemOperations={
 *         "get": {
 *             "method": "GET",
 *             "path": "/benutzer/{id}.{_format}",
 *             "requirements": {"id": "%uuid_format%"},
 *             "normalization_context": {"groups": {"benutzer_item_output"}},
 *         },
 *     },
 *     attributes={
 *         "order": {"anzeigename": "ASC"},
 *     },
 * )
 * @API\ApiFilter(BooleanFilter::class)
 * @ORM\Entity(repositoryClass="App\Repository\BenutzerRepository")
 * @ORM\Table(
 *     name="benutzer",
 *     indexes={
 *         @ORM\Index(name="benutzer_anzeigename_idx", columns={"anzeigename"})
 *     },
 *     uniqueConstraints={
 *         @ORM\UniqueConstraint(name="benutzer_kontoname", columns={"kontoname"}),
 *     },
 * )
 */
class Benutzer implements UserInterface
{
    /**
     * @API\ApiProperty(identifier=true, writable=false)
     * @ORM\Column(name="object_guid", type="guid")
     * @ORM\Id
     * @Serial\Groups({
     *     "benutzer_collection_output",
     *     "benutzer_item_output",
     *     "fall_item_output",
     * })
     * @Assert\Uuid
     *
     * @var string
     */
    private $objectGuid;
    
    // ...
    
}
```

I think I need a custom Siri converter from now on, am I right?

soyuka:
Mhh it should not break on this I think it's a bug. Do you really have to declare the `path`? In the meantime try:

```
  "path": "/benutzer/{objectGuid}.{_format}",
```

Or specify 

```
  "identifiers": {"id": {"objectGuid", Benutzer::class}}
```

to see if it works?

OskarStark:
Thanks for your feedback 👍 

Unfortunately this doesn't work.

I used your code snippets like this:
```diff
 *     itemOperations={
 *         "get": {
 *             "method": "GET",
- *             "path": "/benutzer/{id}.{_format}",
- *             "requirements": {"id": "%uuid_format%"}, 
+ *             "path": "/benutzer/{objectGuid}.{_format}",
+ *             "requirements": {"objectGuid": "%uuid_format%"},
 *             "normalization_context": {"groups": {"benutzer_item_output"}},
 *         },
 *     },
```

and (but only one at a time) like this:

```diff
 *     attributes={
+ *         "identifiers": {"id": {"objectGuid", Benutzer::class}},
 *         "order": {"anzeigename": "ASC"},
 *     },
```

Correct?

Where can I find some documentation for the "identifiers" ? 🤔

soyuka:
```diff
 *     itemOperations={
 *         "get": {
 *             "method": "GET",
-*             "path": "/benutzer/{id}.{_format}",
-*             "requirements": {"id": "%uuid_format%"}, 
+*             "path": "/benutzer/{objectGuid}.{_format}",
+*             "requirements": {"objectGuid": "%uuid_format%"},
 *             "normalization_context": {"groups": {"benutzer_item_output"}},
 *         },
 *     },
```

Is working.

Issue #3168, opened 2019-10-11
Varnish cache tags with a big number of entities 


bastnic:
On a biiig api platform website with lots and lots of nested entities, we need varnish to be fully working. 

I talked with @alanpoulain last night and we concurred that I don't have an usual setup but the experience must be shared as it can helps some people.
I already have fixed all my issues on my side, but cool if we can fix the official release.

There is multiples troubles: 
- too much entities means too much iris, iris are quite long so with more than one hundred iris it crashes because `Cache-Tags` headers are too long. More on that later.
- too much entities updated means too much iris to clean on VarnishPurger, it crashes and do not clean anything.
- we also had https://github.com/api-platform/core/pull/2905 but it's fixed now.

## Too much cache tags to add on the response

On the biggest response i found, the Cache-Tags header weight 100k chars. It's quite long.
I tried multiples approaches:
- increase (a lot) varnish headers (hi `http_resp_hdr_len`, `http_resp_size`): doesn't scale as much as I want
- chunk the headers, as it's allowed by http (hi `http_max_hdr`): it's seems to work at first but in fact when the purge happens, it only use the Cache tags of the first header line. Maybe a bug on my side
- on a collection, just strip all the tags of the items of that collection as when we do an operation on a resource, the iri of the collection is also given. It reduces a LOT! 
- BUT, it only removes the iris of the resource itself. My entities are quite nested so I also have as much iris as different subsresources. So much that I can almost said that this is a collection itself.

Brace yourself, awful code:

```php
// in AddTagsListener
$posibleCollections = [];

        // first, attempt to get all iri's prefix to check if collections emerged
        foreach ($resources as $resource) {
            if (preg_match('#(.*)/.*-.*-.*.*$#', $resource, $matches)) {
                if (!isset($posibleCollections[$matches[1]])) {
                    $posibleCollections[$matches[1]] = 0;
                }
                ++$posibleCollections[$matches[1]];
            }
        }

        // extract all collections (more than XX iris of the same type)
        $posibleCollections = array_filter($posibleCollections, function ($count, $collection) {
            return $count > self::MAX_IRIS_TO_BE_CONSIDERED_AS_COLLECTION; // magic number
        }, ARRAY_FILTER_USE_BOTH);

        // then remove the corresponding iris
        $resources = array_filter($resources, function ($item) use ($posibleCollections) {
            foreach ($posibleCollections as $collection => $count) {
                if (strpos($item, $collection) === 0) {
                    return false;
                }
            }

[...]

Issue #2926, opened 2019-07-11
is exists anyway to disable cache invalidation in runtime?


ghanbari:
Hi, I have some command that do heavy work, but for some reason i don't need to purge cache,
I want to disable cache invalidation for those commands, i removed `PurgeHttpCacheListener` listener from entityManager.

is exists anyway to disable cache invalidation by calling a service method?

teohhanhui:
Do you mean to disable the cache tags only for certain operations?

teohhanhui:
Looking at the code, this is not currently possible.

ghanbari:
> Do you mean to disable the cache tags only for certain operations?

No, i want to disable cache tags for certain resource or for all resource in certain console commands.



> Looking at the code, this is not currently possible.

I check code and didn't saw anything, maybe we have this feature in future?

teohhanhui:
I don't see the support for disabling cache tags for a certain resource either.

These should be easy to add. PRs welcome!

oussj:
Hello @teohhanhui 
Would love to help on this matter as I'm experiencing the same problem. 
First, which developer interface do you have in mind for this feature.
As I see it, I see three options : 
- an Annotation on a service that would disable the cache invalidation temporarly
`/**
 * @DisableCachePurge
 */
class MyMassiveImportCommand
{`
- an interface that would implement a enableCachePurge and disableCachePurge method 
`interface ControlCachePurgeInterface
{
    public function enableCachePurge();
    
    public function disableCachePurge();
}
`
Do you think one of these solutions would be apropriate ?

Thanks !

oussj:
Hello @teohhanhui 

What do you think about the solutions I suggested? Do you have a better idea to fix this?

Thanks!

oussj:
Hello @teohhanhui 

Were you able to have a look at the solution I suggested? Looking forward to have your feedbacks and other ideas!

Thanks!

teohhanhui:
@oussj Thanks for the suggestions, but I don't think they're the right ones for this case. :smile:

If you want to disable cache purging for a certain console command, that should already be possible by using e.g.

```yaml
api_platform:
    http_cache:
        invalidation:
            enabled: '%env(bool:CACHE_INVALIDATION_ENABLED)%'
```

    $ CACHE_INVALIDATION_ENABLED=0 bin/console app:my:command

oussj:
Ok thanks ! 
I'm looking into it !

bastnic:
[...]

PR #3665, opened 2020-08-06
[core] Enable/Disable AddTagsListener

guillaumedelre:
Add attribute cache_invalidation in ApiResource

| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

My customer uses varnish cache to maximize performance, but for some cases, I need to disable http cache invalidation.

**Proposal**: Enable/Disable AddTagsListener at runtime

It could be done **globally** with something like that:

```php
<?php
// api/src/Entity/Book.php

namespace App\Entity;

/**
 * ...
 * @ApiResource(
 *     cacheInvalidation=false,
 * ...
 * )
 */
class Book
{
    // ...
}
```

It could be disabled **at operation level** with something like that:

```php
<?php
// api/src/Entity/Book.php

namespace App\Entity;

/**
 * ...
 * @ApiResource(
 *     itemOperations={
 *         "get"={
 *             "cache_invalidation"=false,
 *             ...
 *         },
 *     },
 * ...
 * )
 */
class Book
{
    // ...
}
```

### DIFF
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -27,6 +27,7 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
  *     @Attribute("accessControlMessage", type="string"),
  *     @Attribute("attributes", type="array"),
  *     @Attribute("cacheHeaders", type="array"),
+ *     @Attribute("cacheInvalidation", type="bool"),
  *     @Attribute("collectionOperations", type="array"),
  *     @Attribute("denormalizationContext", type="array"),
  *     @Attribute("deprecationReason", type="string"),
@@ -87,6 +88,7 @@ final class ApiResource
         'securityPostDenormalize',
         'securityPostDenormalizeMessage',
         'cacheHeaders',
+        'cacheInvalidation',
         'collectionOperations',
         'denormalizationContext',
         'deprecationReason',
@@ -174,6 +176,14 @@ final class ApiResource
      * @var array
      */
     private $cacheHeaders;
+
+    /**
+     * @see https://api-platform.com/docs/core/performance/#enabling-disabling-http-cache-invalidation
+     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
+     *
+     * @var bool
+     */
+    private $cacheInvalidation;
 
     /**
      * @see https://api-platform.com/docs/core/serialization/#using-serialization-groups
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -178,9 +178,6 @@ final class ApiResource
     private $cacheHeaders;
 
     /**
-     * @see https://api-platform.com/docs/core/performance/#enabling-disabling-http-cache-invalidation
-     * @see https://github.com/Haehnchen/idea-php-annotation-plugin/issues/112
-     *
      * @var bool
      */
     private $cacheInvalidation;
--- a/src/HttpCache/EventListener/AddTagsListener.php
+++ b/src/HttpCache/EventListener/AddTagsListener.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\HttpCache\EventListener;
 
 use ApiPlatform\Core\Api\IriConverterInterface;
+use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Metadata\Resource\ToggleableOperationAttributeTrait;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
 
@@ -30,11 +32,17 @@ use Symfony\Component\HttpKernel\Event\ResponseEvent;
  */
 final class AddTagsListener
 {
-    private $iriConverter;
+    use ToggleableOperationAttributeTrait;
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public const OPERATION_ATTRIBUTE_KEY = 'cache_invalidation';
+
+    private $iriConverter;
+    private $resourceMetadataFactory;
+
+    public function __construct(IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory = null)
     {
         $this->iriConverter = $iriConverter;
+        $this->resourceMetadataFactory = $resourceMetadataFactory;
     }
 
     /**
@@ -49,6 +57,8 @@ final class AddTagsListener
             !$request->isMethodCacheable()
             || !$response->isCacheable()
             || (!$attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !$attributes['cache_invalidation']
+            || $this->isOperationAttributeDisabled($attributes, self::OPERATION_ATTRIBUTE_KEY)
         ) {
             return;
         }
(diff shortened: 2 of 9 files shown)


guillaumedelre:
This is my very first PR, I hope someone will helps me :smile:

soyuka:
Related issues: https://github.com/api-platform/core/issues/2926 https://github.com/api-platform/core/issues/3168 ping @bastnic

src/Annotation/ApiResource.php
```diff
@@ -175,6 +177,14 @@ final class ApiResource
      */
     private $cacheHeaders;
 
+    /**
+     * @see https://api-platform.com/docs/core/performance/#enabling-disabling-http-cache-invalidation
```
soyuka, line 181: not needed

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -49,6 +57,8 @@ public function onKernelResponse(ResponseEvent $event): void
             !$request->isMethodCacheable()
             || !$response->isCacheable()
             || (!$attributes = RequestAttributesExtractor::extractAttributes($request))
+            || !$attributes['cache_invalidation']
+            || $this->isOperationAttributeDisabled($attributes, self::OPERATION_ATTRIBUTE_KEY)
```
soyuka, line 61: I think that the default should be `true` to keep the currnet behavior?

src/Annotation/ApiResource.php
guillaumedelre: done

antograssiot:
Shouldn't this be done in PurgeHttpCacheListener listener also to avoid computing tags that needs to be purged ?
Also maybe `cacheTagsInvalidation` or something similar would be more accurate ?

guillaumedelre:
> Also maybe `cacheTagsInvalidation` or something similar would be more accurate ?
my next commit will change the naming for `cacheTagsInvalidation`.

> Shouldn't this be done in PurgeHttpCacheListener listener also to avoid computing tags that needs to be purged ?
Something like that could be acceptable in the PurgeHttpCacheListener ?
- inject the requestStack 
- use ToggleableOperationAttributeTrait
- use the private method `isCacheTagsInvalidationEnabled` in the binded events

```php
final class PurgeHttpCacheListener
{
    use ToggleableOperationAttributeTrait;

    private $purger;
    private $iriConverter;
    private $propertyAccessor;
    private $requestStack;
    private $tags = [];
even if I don't like this coupling
    public function __construct(PurgerInterface $purger, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, RequestStack $requestStack, PropertyAccessorInterface $propertyAccessor = null)
    {
        $this->purger = $purger;
        $this->requestStack = $requestStack;
        $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
    }

    private function isCacheTagsInvalidationEnabled(): bool
    {
        $request = $this->requestStack->getCurrentRequest();

        return (
            (!$attributes = RequestAttributesExtractor::extractAttributes($request))
            || !$attributes['cache_tags_invalidation']
            || $this->isOperationAttributeDisabled($attributes, 'cache_tags_invalidation', false)
        ) ? false : true;
    }

    /**
     * Collects tags from the previous and the current version of the updated entities to purge related documents.
     */
    public function preUpdate(PreUpdateEventArgs $eventArgs): void
    {
        if (!$this->isCacheTagsInvalidationEnabled()) {
            return;
        }
```

guillaumedelre:
@soyuka @antograssiot : 
Or in the `ApiPlatform\Core\EventListener\WriteListener`, we enrich the `$attributes` array with a new key that **allow us to disable** the listener `PurgeHttpCacheListener` in the `ApiPlatform\Core\Bridge\Doctrine\Common\DataPersister`


`ApiPlatform\Core\EventListener\WriteListener`: 
```php
...
        if (
            $controllerResult instanceof Response
            || $request->isMethodSafe()
            || !($attributes = RequestAttributesExtractor::extractAttributes($request))
            || !$attributes['persist']
            || $this->isOperationAttributeDisabled($attributes, self::OPERATION_ATTRIBUTE_KEY)
        ) {
            return;
        }

        $attributes['purge_http_cache_listener_disabled'] = $attributes['cache_tags_invalidation'] || $this->isOperationAttributeDisabled($attributes, 'cache_tags_invalidation', false);

        if (!$this->dataPersister->supports($controllerResult, $attributes)) {
            return;
        }
...
```

`ApiPlatform\Core\Bridge\Doctrine\Common\DataPersister`:
```php
    private $managerRegistry;
    private $purgeHttpCacheToggler;

    public function __construct(ManagerRegistry $managerRegistry, PurgeHttpCacheToggler $purgeHttpCacheToggler)
    {
        $this->managerRegistry = $managerRegistry;
        $this->purgeHttpCacheToggler = $purgeHttpCacheToggler;
    }
...

    /**
     * {@inheritdoc}
     */
    public function persist($data, array $context = [])
    {
        if (!$manager = $this->getManager($data)) {
            return $data;
        }

        if ($context['purge_http_cache_listener_disabled']) {
            $this->purgeHttpCacheToggler->disablePurgeHttpCacheListener();
        }

        if (!$manager->contains($data) || $this->isDeferredExplicit($manager, $data)) {
            $manager->persist($data);
        }

        $manager->flush();
        $manager->refresh($data);

        if ($context['purge_http_cache_listener_disabled']) {
            $this->purgeHttpCacheToggler->enablePurgeHttpCacheListener();
        }

        return $data;
    }

    /**
     * {@inheritdoc}
     */
    public function remove($data, array $context = [])
    {
        if (!$manager = $this->getManager($data)) {
            return;
        }

        if ($context['purge_http_cache_listener_disabled']) {
            $this->purgeHttpCacheToggler->disablePurgeHttpCacheListener();
        }

        $manager->remove($data);
        $manager->flush();

        if ($context['purge_http_cache_listener_disabled']) {
            $this->purgeHttpCacheToggler->enablePurgeHttpCacheListener();
        }
    }
...
```

et le nouveau service `ApiPlatform\Core\Bridge\Doctrine\Common\PurgeHttpCacheToggler`:
```php
<?php

declare(strict_types=1);

namespace ApiPlatform\Core\Bridge\Doctrine\Common;

use ApiPlatform\Core\Bridge\Doctrine\EventListener\PurgeHttpCacheListener;
use Doctrine\Common\EventManager;

final class PurgeHttpCacheToggler
{
    private $eventManager;

    public function __construct(EventManager $eventManager)
    {
        $this->eventManager = $eventManager;
    }

    private function getListener(): ?PurgeHttpCacheListener
    {
        foreach ($this->eventManager->getListeners() as $event => $listeners) {
            foreach ($listeners as $key => $listener) {
                if (!$listener instanceof PurgeHttpCacheListener) {
                    continue;
                }
                return $listener;
            }
        }

        return null;
    }

    public function disablePurgeHttpCacheListener():void
    {
        $this->eventManager->removeEventListener(['onFlush', 'preUpdate', 'postFlush'], $this->getListener());
    }

    public function enablePurgeHttpCacheListener(): void
    {
        $this->eventManager->addEventListener(['onFlush', 'preUpdate', 'postFlush'], $this->getListener());
    }
}
```

soyuka:
I'm not a fan of the second solution you proposed, I tend to prefer the first even though I don't like the coupling either. 
I think we need @dunglas input on this.

guillaumedelre:
hello @dunglas could youo give me inputs on this feature ? thank you

vasilvestre:
I would like to help here if possible ! @dunglas are you available for input ?

dunglas:
I need to take a deeper look. In my opinion the cleanest way would be to skip generating tags related to resources having invalidation disabled. However we must measure if computing metadata of all traversed resource types doesn't introduce an overhead (metadata should be cached, but we must double check to be sure).

Any help on implementing this patch and measuring its impact (using Blackfire for instance) will be very appreciated!

PR #3669, opened 2020-08-11
[core] Encode definition names to be compliant with JSON Pointer Specification

### LINKED ISSUE #1969: Swagger documentation: Could not resolve reference
When the resource short-name contain a slash the Swagger documentation fail to link the ref and fire error saying "Could not resolve reference ...".

It can be fix by converting every slash by something else (like a -) when generating the definition key in the function getDefinition.

I think this bug was fixed in last version of Swagger UI.

stchr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | fixes #1969
| License       | MIT

~~This PR currently only urlencodes the colon `:` separator in schema names as this currently seems the only illegal char besides the allowed `a-zA-Z0-9 _ . -` in OpenAPI 3.0.~~

This PR urlencodes the whole name and additionally replacing `/` and `~` according to [JSON Pointer Specification](https://tools.ietf.org/html/rfc6901).


src/JsonSchema/SchemaFactory.php
```diff
@@ -235,12 +235,12 @@ private function buildDefinitionName(string $className, string $format = 'json',
...
+            $prefix .= urlencode(':').$shortName;
         }
 
         if (isset($this->distinctFormats[$format])) {
             // JSON is the default, and so isn't included in the definition name
-            $prefix .= ':'.$format;
+            $prefix .= urlencode(':').$format;
         }
 
         if (isset($serializerContext[DocumentationNormalizer::SWAGGER_DEFINITION_NAME])) {
```
soyuka, line 246: I'd encode at the bottom of this function and follow https://github.com/api-platform/core/issues/1969#issuecomment-392033381 by:

- replacing `~` and `/` 
- urlencoding the final `$name`

soyuka:
behat tests failing

src/JsonSchema/SchemaFactory.php
stchr: Introduced new method `\ApiPlatform\Core\JsonSchema\SchemaFactory::encodeDefinitionName` which does this.
But need to urlencode first to not overwrite replacing for `~` and `/`.
And it comes out that only the refs should be encoded.

stchr:
> behat tests failing

@soyuka Done.

soyuka:
looks good, @dunglas could you check ?

stchr:
@soyuka @dunglas As it comes out only the refs are affected (and not the schema definition names in general) there should be no more BC break and we could merge this in the 2.5 branch.
What do you think?

dszwcz:
@soyuka @dunglas it would be great to merge it :)

src/JsonSchema/SchemaFactory.php
```diff
@@ -253,6 +254,14 @@ private function buildDefinitionName(string $className, string $format = 'json',
         return $name;
     }
 
+    private function encodeDefinitionName(string $name): string
+    {
+        $entities = ['%2F', '%7E'];
+        $replacements = ['~1', '~0'];
```
dunglas, line 260: Why is it necessary? IIRC only `~` and `/` must be encoded, not there URL encoded versions. Or am I missing something?

features/openapi/docs.feature
```diff
@@ -95,7 +95,7 @@ Feature: Documentation support
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters" should have 6 elements
 
     # Subcollection - check schema
-    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.responses.200.content.application/ld+json.schema.properties.hydra:member.items.$ref" should be equal to "#/components/schemas/RelatedToDummyFriend:jsonld-fakemanytomany"
+    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.responses.200.content.application/ld+json.schema.properties.hydra:member.items.$ref" should be equal to "#/components/schemas/RelatedToDummyFriend%3Ajsonld-fakemanytomany"
```
dunglas, line 98: If we introduce a BC break, shouldn't we also change to a URL-safe character as separator to make the names more human-readable?

dunglas:
Actually, as pointed out by @hkosova (https://github.com/api-platform/core/issues/1969#issuecomment-392033381), the generated identifiers are still invalid according to the OpenAPI spec: `All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.`. We must use safe characters according to this regex instead of escaping.

features/openapi/docs.feature
stchr: Currently there should be no BC break as only the refs are affected, not the definition names itself.
And you prefered urlencode in https://github.com/api-platform/core/issues/1969#issuecomment-391735278
But sure we can switch to safe chars, if you like. Which do you prefer? I guess `.` would be a good alternative to `:`?

src/JsonSchema/SchemaFactory.php
stchr: Please have a look to https://github.com/api-platform/core/issues/1969#issuecomment-392033381

stchr:
> Actually, as pointed out by @hkosova ([#1969 (comment)](https://github.com/api-platform/core/issues/1969#issuecomment-392033381)), the generated identifiers are still invalid according to the OpenAPI spec: `All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.`. We must use safe characters according to this regex instead of escaping.

@dunglas Good point, missed out OpenAPI 3.0 spec. Which character do you prefer from this regex? Would go with `.`?

dszwcz:
> > Actually, as pointed out by @hkosova ([#1969 (comment)](https://github.com/api-platform/core/issues/1969#issuecomment-392033381)), the generated identifiers are still invalid according to the OpenAPI spec: `All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.`. We must use safe characters according to this regex instead of escaping.
> 
> @dunglas Good point, missed out OpenAPI 3.0 spec. Which character do you prefer from this regex? Would go with `.`?

dot is ok for me

dunglas:
`.` looks good to me too.

stchr:
Messed up this PR 😞 , new and clean one is #3705

PR #3670, opened 2020-08-12
Fix missing use statement

stchr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

Just added a missing use statement (see line 31/32)


stchr:
Curious why tests from PHP 7.2 upwards are failing. I'm not aware any changes in PHP would cause this.

dunglas:
Thanks @stchr. I'll take care of the tests.

PR #3671, opened 2020-08-12
Fixing error for none database identifiers Api ids (with tests)

Arnaud-J:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | none 
| License       | MIT
| Doc PR        | none

Add tests for #3529 as asked by @soyuka

PR #3529, opened 2020-04-26
Fixing error for none database identifiers Api ids

mahmoodbazdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  |no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | none
| License       | MIT
| Doc PR        | none

There is an error on sub resources queries when Api platform identifier is not an identifier in database like when using slug but using id as database  identifier.

related to: https://github.com/api-platform/core/pull/3396

**Update:** I guess I should change the target to 2.5 because this error exist there also.

PR #3672, opened 2020-08-12
Patch fix none identifier join

soyuka:
#3529 and #3671


soyuka:
Mhh github CI is green but we had this `push` thing that's red. Let me re-open this pull-request.

Issue #3676, opened 2020-08-13
Subresource pagination does not work if primary resource pagination is disabled


nightio:
**API Platform version(s) affected**: 2.5.x (and probably all others)

**Description**  
Subresource pagination is not working if primary resource pagination is disabled

**How to reproduce**  
```php 
<?php


namespace App\Infrastructure\Entity;


use ApiPlatform\Core\Annotation\ApiResource;
use ApiPlatform\Core\Annotation\ApiSubresource;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * @ApiResource(
 *     attributes={
 *          "pagination_enabled"=false
 *     },
 *     itemOperations={"get"}
 * )
 */
class Site
{
    /**
     * @var string|null
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private ?string $id;

    /**
     * @var string
     *
     * @ORM\Column
     * @Assert\NotBlank
     */
    private string $name;

    /**
     * @var Item[]|array
     *
     * @ORM\OneToMany(targetEntity="Item", mappedBy="site")
     *
     * @ApiSubresource
     */
    private array $items = [];

    /**
     * Site constructor.
     * @param string|null $id
     * @param string $name
     */
    public function __construct(?string $id, string $name)
    {
        $this->id = $id;
        $this->name = $name;
    }

    /** Getters */
}
```

```php
<?php


namespace App\Infrastructure\Entity;


use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * @ApiResource(
 *     itemOperations={"get"},
 *     subresourceOperations={
 *          "items_get_subresource"= {
 *              "pagination_enabled"=true
 *          }
 *     }
 * )
 */
class Item
{
    /**
     * @var string|null
     *
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private ?string $id;

    /**
     * @var string
     *
     * @ORM\Column
     * @Assert\NotBlank
     */
    private string $name;

    /**
     * @var Site
     *
     * @ORM\ManyToOne(targetEntity="Site", inversedBy="items")
     */
    private Site $site;

    /**
     * Item constructor.
     * @param string|null $id
     * @param string $name
     * @param Site $site
     */
[...]

PR #3678, opened 2020-08-17
#3676 - fixed subresource pagination support if primary resource has pagination disabled

nightio:
| Q             | A
| ------------- | ---
| Bug fix?      | yes - it fixes wrong behavior
| New feature?  | no 
| BC breaks?    | possibly it can enable pagination (in documentation) in some specyfic cases
| Deprecations? |no
| Tickets       | #3676 
| License       | MIT
| Doc PR        | -

Now the decision of subresource pagination is based on subresource operation or if not set, on subresource resource. Previously was based on Primary resource get operation or Primary resource itself.


dunglas (approved):
Could you add a non-regression test please?

dunglas:
Thanks @nightio, a non-regression test will still be very appreciated as an other PR :)

Issue #3551, opened 2020-05-08
Subresource Operation Overrides Custom Subresource Operation in Swagger Documentation


mab05k:
Given I have the following configuration
```
# config/resources/question.yaml
resources:
  App\Entity\Question:
    properties:
      answers:
        subresource:
          collection: true
            resourceClass: App\Entity\Answer

  App\Entity\Answer:
    collectionOperations:
      subresource_post:
        method: POST
        path: /questions/{id}/answers
        defaults:
          _api_subresource_context:
            identifiers:
            - [ 'id', 'App\Entity\Question', true ]
```
I have a custom DataPersister for handling the POST.

Problem:
When the Swagger is generated in the `ApiPlatform\Core\Swagger\Serializer\DocumentationNormalizer`, the following snippet within the `normalize()` function `:174` is causing the problem
```
foreach ($this->subresourceOperationFactory->create($resourceClass) as $operationId => $subresourceOperation) {
  $paths[$this->getPath($subresourceOperation['shortNames'][0], $subresourceOperation['route_name'], $subresourceOperation, OperationType::SUBRESOURCE)] = $this->addSubresourceOperation($v3, $subresourceOperation, $definitions, $operationId, $resourceMetadata);
}
```

Prior to this, the `paths` ArrayCollection is created for my operation like this
```
[  '/questions/{id}/answers' => [ 'post' => post_operation...] ]
```

After the above snippet is run, instead of updating to this
```
[  '/questions/{id}/answers' => [ 'post' => post_operation..., 'get' => subresource_operation... ] ]
```
We get
```
[  '/questions/{id}/answers' => subresource_operation... ]
```

I think we'd need to somehow resolve what operation the subresource is and nest it properly in the `paths` array

```
  $paths[$this->getPath($subresourceOperation['shortNames'][0], $subresourceOperation['route_name'], $subresourceOperation, OperationType::SUBRESOURCE)][OPERATION METHOD HERE!!!] = $this->addSubresourceOperation($v3, $subresourceOperation, $definitions, $operationId, $resourceMetadata);
```

It doesn't look like there is currently any data available about the subresource at this point that indicates what operation type we're dealing with, so maybe default to "GET"?

PR #3685, opened 2020-08-26
[Swagger] Fix that subresource endpoints do not override custom endpoints

GSadee:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/core/issues/3551
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -192,7 +192,8 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
             }
 
             foreach ($this->subresourceOperationFactory->create($resourceClass) as $operationId => $subresourceOperation) {
-                $paths[$this->getPath($subresourceOperation['shortNames'][0], $subresourceOperation['route_name'], $subresourceOperation, OperationType::SUBRESOURCE)] = $this->addSubresourceOperation($v3, $subresourceOperation, $definitions, $operationId, $resourceMetadata);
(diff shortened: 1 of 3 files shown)


src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -470,7 +471,7 @@ private function addSubresourceOperation(bool $v3, array $subresourceOperation,
             $this->addPaginationParameters($v3, $resourceMetadata, $operationName, $pathOperation);
         }
 
-        return new \ArrayObject(['get' => $pathOperation]);
+        return $pathOperation;
```
dunglas, line 474: Shouldn't `ArrayObject` be kept?

```suggestion
        return new \ArrayObject($pathOperation);
```

dunglas (changes requested):

src/Swagger/Serializer/DocumentationNormalizer.php
GSadee: `$pathOperation` is already an `\ArrayObject`: https://github.com/api-platform/core/pull/3685/files#diff-32d63273c86a7e4c14c3fa6bb4131194R422

What is more, in my opinion, there was some inconsistency in `$paths`, because every subresource endpoint was added as `\ArrayObject` while other endpoints were a map `method => \ArrayObject`

Example from Sylius, before this change:
image

dunglas:
Thanks @GSadee

Issue #3177, opened 2019-10-15
subresourceOperations and security does not work as expected


lyrixx:
**Description**  
https://api-platform.com/docs/core/subresources/#access-control-of-subresources

In the documentation, `has_role('ROLE_AUTHENTICATED')` is used and there are no example about using the object.
It's because the `object` var is not the question (according to the doc), it's an array of answers.
So it's really not convenient to use it.

As we will use it in a subresource, it should be qustions IHMO (the resource) and not the answers (the sub resource**s**).

---

We can hack the system but this is so wrong:
```
"security"="object == [] or is_granted('ROLE_VIEW', object[0].getQuestion())"
```

julienfalque:
I'm facing the same issue: if one is not allowed to access the parent resource, they should not be able to see its subresource either. There should be a way to deny access based on the parent resource but this does not seem possible at the moment. As @lyrixx said, a hacky workaround is possible but is totally wrong.

Issue #1628, opened 2018-01-05
How to add new entry in a ManyToMany?


vincentchalamon:
I have a User entity linked with a bidirectional ManyToMany association to an Event entity: an event have users registered, and it's possible to get a user events through an ApiSubresource.

When a user wants to register to an event, I should update the event to add the user:
```
PUT /events/{id}
{
    "users": [
        "/users/1",
        "/users/2",
        …
        "/users/8888" <- the user who wants to register
    ]
}
```

This could also be done by adding the event to the user:
```
PUT /users/{id}
{
    "events": [
        "/events/1",
        "/events/2",
        …
        "/events/8888" <- the event the user wants to register
    ]
}
```

In the first example, I'll expose the users id, which could be a security issue. In the second example, I'll have to find all the user events id which is not optimized.
To optimize this process, I've created a custom entry-point `PUT /events/{id}/register` to register the current user to the event.

It would be awesome to allow to send a write request to my ApiSubresource to add an event without reminding all previous events, for instance:
```
PUT /users/12345/events
{
    "@id": "/events/67890"
}
```
And to remove an event from this ApiSubresource: `DELETE /users/12345/events/67890`

What do you think @api-platform/core-team ?

soyuka:
Writability on subresource is a wanted/needed/interesting feature indeed, it should not be too hard to implement now. Feel free to work on it if you have the time!

hassanjuniedi:
@vincentchalamon that's great, am in the middle of similar problem and am trying to create custom api entry but with no success , can you please post an example of how you created custom action for  **PUT /events/{id}/register** 
note: the entry am trying to accomplish is like this _/products/{id}/add-comment_ , i have to pass product id and comment object or comment id but inside the _invoke function how to get this data .

vincentchalamon:
Hi @hassanjuniedi, my case changed a little bit as I need a more complex object to register a user to an event. I've created a `Registration` entity, and am able to send a POST request to /registrations to register a user to an event.

In your case, I think you confuse the front routing and the api routing: you don't need to restrict the comment creation url under `/products`. IMO, the easiest way would be to send a POST request to `/comments`.

If you really need to send a POST request to `/products/{id}/comments`, you'll need to create a custom route:
```yaml
products_add_comment:
    path: /products/{id}/comments
    controller: api_platform.action.placeholder
    methods: [POST]
```

Then, add an itemOperation on your Product entity:
```php
/**
 * @ApiResource(attributes={...}, itemOperations={
[...]

PR #3689, opened 2020-08-27
Revamp ApiSubresource

### LINKED ISSUE #2706: [RFC] Subresource metadata
This issue merges multiple subresources issues because the root problem is the same.

While talking with core members at EU-FOSSA Hackathon we came to a solution. Best would be to refactor the `ApiSubresource` annotation so that it supports the same attributes as does the `ApiResource`. For example:

- routing options
- access_control
- normalization context

Thanks to this it'll be easier to declare these and we solve the issue (mentioned in https://github.com/api-platform/core/pull/1617#discussion_r15896391). The subresource metadata will not be in `ApiResource` anymore but directly inside the `ApiSubresource` annotation.

Todo list :

- [ ] Fix swagger and don't show the collection on the side that doesn't own the Subresource (https://github.com/api-platform/core/pull/1612)
- [ ] Fix the serialization context declaration (solves https://github.com/api-platform/core/pull/1617 and https://github.com/api-platform/core/issues/1616 and improves DX for https://github.com/api-platform/core/issues/1649)
- [ ]  Simplified the code for the documentation (ses https://github.com/api-platform/core/issues/1736)
- [ ] Subresource operations may override collection operations which should not happend (see https://github.com/api-platform/api-platform/issues/1022)

consider: https://github.com/api-platform/core/pull/3466  and work on https://github.com/api-platform/core/pull/3689
Maybe that this is a prerequisite for write support on subresources (see #2598 #2428)

ping @torreytsui @vincentchalamon please lmk if I forgot things
@lyrixx I'm closing related PRs we will definitely take back the work you did if we can, thanks for pushing this issues

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tickets       | fixes #2706, #1628, #3177 
| License       | MIT
| Doc PR        | **TODO**

- [ ] ADR document
- [ ] Add support for the new `path` attribute on ApiResource
- [ ] Allow setting ApiResource several times (annotation, xml and yaml formats)
- [ ] Patch the ApiLoader to use it
- [ ] Patch the IRI Converter to use it
- [ ] Patch the DataProviders to use it
- [ ] Patch the GraphQL subsystem to use it
- [ ] Patch the DataPersisters to use it
- [ ] Patch the DocumentationNormalizers to use it
- [ ] Add doc PR
- [ ] Deprecate the old ApiSubresource subsystem

Examples:
```php
/**
 * @ApiResource(path="/users")
 * @ApiResource(path="/companies/{companyId}/users", identifiers={"companyId": {Company::class, "id"})
 * @ApiResource(path="/companies/{companyId}/users")
 * @ApiResource(path="/countries/{countryId}/companies/{companyId}/users", identifiers={"companyId": {Company::class, "id"}, "countryId": {Country::class, "id"}})
 *
 * @ApiResource(path="/companies/{companyId}/administrators", identifiers={"companyId": {Company::class, "id"}, "??": {CompanyClass, "administrators"}})
 * @ApiResource(path="/companies/{companyId}/employees", identifiers={"companyId": {Company::class, "id"}, "??": {CompanyClass, "employees"}})
 */
```


soyuka:
I think we should base this branch over the #3664 now. 

The next thing to do is to "merge" the behavior of both `identifiedBy` and `identifiers` on the subresource's metadata. 
At the end, the following parts should be removed:

https://github.com/api-platform/core/blob/c187f365c1b8db2c569f032c4c9a6aa294050699/src/DataProvider/OperationDataProviderTrait.php#L79-L88

https://github.com/api-platform/core/blob/c187f365c1b8db2c569f032c4c9a6aa294050699/src/DataProvider/OperationDataProviderTrait.php#L103-L112.

jamesisaac:
Would this address #3177 ?  Haven't seen it mentioned specifically in the linked issues, but that's quite a big downside currently to Subresources vs normal ApiResources.

vincentchalamon:
@jamesisaac I think it'll fix this issue too, cause the subresources will be deprecated in favor of multiple ApiResources. So in final, the `security` option of ApiResource will be used.

PR #3690, opened 2020-08-27
fix-openapi-subresource

### LINKED ISSUE #3666: [OpenAPI] Subresource schema is incorrect - uses default normalizationContext, completely disregarding actual normalization context (includes minimal repro)
**API Platform version(s) affected**: 2.5.6 and master

**Description**  

API Platform has great handling of different normalization contexts per operation, able to handle complex combinations of serialization groups across many nested objects, and still produce a perfect set of OpenAPI schemas (which can then be used to e.g. generate client-side scaffolding).  However, this breaks down on Subresources (annotated with `@ApiSubresource`).

If I have a subresource with its own normalization context:

```
// Entity/Review.php

 * @ApiResource(
 *  normalizationContext={"groups"={"Review:Default"}},
 *  subresourceOperations={
 *      "api_books_reviews_get_subresource"={
 *          "normalization_context"={
 *              "groups"={
 *                  "Review:Subresource"
 *              },
 *          },
 *      },
 *  },
 * )
```

...then I would expect it (`Review:Subresource`) to be used for the OpenAPI docs.  But unfortunately it's completely ignored, and the OpenAPI docs just use the default normalization context (in this case `Review:Default`, but would just be Symfony's default if that wasn't specified), even if that's completely different from what the endpoint actually returns.

**How to reproduce**  

I've created a reproduction here:

https://github.com/jamesisaac/openapi-bugs

Once running, OpenAPI docs are visible at http://localhost:8000 or http://localhost:8000/docs.json

The `/books/{id}/reviews` endpoint returns the correct data when called (just the `id` and `rating`, as specified by `Review:Subresource`), but the OpenAPI docs are completely overpromising every field from `Review:Default`.

**Additional Context**  

- I've tried this on 2.5.6, and on master with #3407 merged, but the problem seems present on both.
- In this example, no schema for `Review:Subresource` is generated at all, so it's not just a case of just the wrong schema being assigned to the operation - makes it very difficult to work around as potentially the fields from `Review:Subresource` are never exposed anywhere.

cc @soyuka

GregoireHebert:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3666 and #1616
| License       | MIT
| Doc PR        | ~

### DIFF
--- a/tests/Operation/Factory/SubresourceOperationFactoryTest.php
+++ b/tests/Operation/Factory/SubresourceOperationFactoryTest.php
@@ -75,7 +75,7 @@ class SubresourceOperationFactoryTest extends TestCase
                 ],
                 'route_name' => 'api_dummy_entities_subresource_get_subresource',
                 'path' => '/dummy_entities/{id}/subresource.{_format}',
-                'operation_name' => 'subresource_get_subresource',
+                'operation_name' => 'api_dummy_entities_subresource_get_subresource',
             ] + SubresourceOperationFactory::ROUTE_OPTIONS,
             'api_dummy_entities_subresource_another_subresource_get_subresource' => [
                 'property' => 'anotherSubresource',
@@ -88,7 +88,7 @@ class SubresourceOperationFactoryTest extends TestCase
                 ],
                 'route_name' => 'api_dummy_entities_subresource_another_subresource_get_subresource',
                 'path' => '/dummy_entities/{id}/subresource/another_subresource.{_format}',
-                'operation_name' => 'subresource_another_subresource_get_subresource',
+                'operation_name' => 'api_dummy_entities_subresource_another_subresource_get_subresource',
             ] + SubresourceOperationFactory::ROUTE_OPTIONS,
             'api_dummy_entities_subresource_another_subresource_subcollections_get_subresource' => [
                 'property' => 'subcollection',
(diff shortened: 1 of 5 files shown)


vincentchalamon (approved):
LGTM

GregoireHebert:
I have no idea why the coverage is not satisfied...

tests/Operation/Factory/SubresourceOperationFactoryTest.php
```diff
@@ -75,7 +75,7 @@ public function testCreate()
                 ],
                 'route_name' => 'api_dummy_entities_subresource_get_subresource',
                 'path' => '/dummy_entities/{id}/subresource.{_format}',
-                'operation_name' => 'subresource_get_subresource',
+                'operation_name' => 'api_dummy_entities_subresource_get_subresource',
```
dunglas, line 78: isn't this a BC break?

jamesisaac:
Works perfectly in my testing (checked against project which caused me to open #3666 ).  Thanks a lot!  Looking forward to seeing this released.

tests/Operation/Factory/SubresourceOperationFactoryTest.php
GregoireHebert: Well, IMO yes and no. Yes because, even if generated, it could have been used. But no because I think it's a bug and that it should always have been like that. And no because subresources are experimental and that part is internal.

dunglas: It should at least be documented in the changelog as a BC break.

jamesisaac:
@GregoireHebert @soyuka Is there any chance this fix / something to this effect can be included in 2.6?

The bug outlined in #3666 is still present from what I can see.  However the situation gets a bit worse still, as now in 2.6 even if the subresource is using the default normalization context, that schema doesn't get output either if that normalization context isn't used in any item or collection operations, i.e. when using the pattern of:

```
 *     collectionOperations={},
 *     itemOperations={"get"={"controller"=NotFoundAction::class, "read"=false, "output"=false}},
``

(although that can be worked around by having the item GET endpoint available but blocked off by security instead).

Anyway, the state as it is now means when using subresources in non-trivial ways, the types are incorrectly exposed via OpenAPI.  So it creates lots of room for error in API consumers, and places where type schemas need to be determined manually, losing a big part of the benefit of OpenAPI.

soyuka:
This is a good patch but is definitely breaking things. I'll see if we can add this to the next release. 

Also related to: https://github.com/api-platform/core/issues/3458.

PR #3696, opened 2020-08-29
Checking the output option when calculating property metadata

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | this could change behavior for people relying on bug
| Deprecations? | no
| Tickets       | none
| License       | MIT
| Doc PR        | not needed

Hi!

Reproducer with instructions here: https://github.com/SymfonyCasts/api-platform/tree/embedded-output-iri-embedded-bug

It's super tricky and subtle. The tl;dr is that when the `readableLink` of `PropertyMetadata` is calculated for an embedded property, it uses the serializer groups from the target resource class when it *should* use the `output=` class (because that is what will actually be serialized).

I didn't include a test (yet) - I'd like some validation that this it the correct place/idea for this.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * OpenAPI: Do not set scheme to oauth2 when generating securitySchemes (#4073)
 * OpenAPI: Fix missing `$ref` when no `type` is used in context (#4076)
 * GraphQL: Fix "Resource class cannot be determined." error when a null iterable field is returned (#4092)
+* Metadata: Check the output class when calculating serializer groups (#3696)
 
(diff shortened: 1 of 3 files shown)


dunglas:
+1 on my side! Thanks for the patch.

soyuka:
:+1: about the patch, I think that this could go on 2.5

dunglas (approved):

dunglas:
Could you rebase against 2.5 please? It should fix the test.

weaverryan:
Test fails are legit - I need to check them!

PR #3697, opened 2020-08-31
Rewrite bad input denormalization error

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       |  na
| License       | MIT
| Doc PR        | na

Change the message error when an Input can not be denormalized. 

@weaverryan @dunglas

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -31,6 +31,7 @@ use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\Exception\LogicException;
 use Symfony\Component\Serializer\Exception\MissingConstructorArgumentsException;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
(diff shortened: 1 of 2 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -190,9 +191,14 @@ public function denormalize($data, $class, $format = null, array $context = [])
...
-            $denormalizedInput = $this->serializer->denormalize($data, $inputClass, $format, $context);
+            try {
+                $denormalizedInput = $this->serializer->denormalize($data, $inputClass, $format, $context);
+            } catch (NotNormalizableValueException $e) {
+                throw new UnexpectedValueException('The input data is misformatted.', $e->getCode(), $e);
+            }
+
             if (!\is_object($denormalizedInput)) {
-                throw new \UnexpectedValueException('Expected denormalized input to be an object.');
+                throw new UnexpectedValueException('Expected denormalized input to be an object.');
```
dunglas, line 201: Can we preserve at least the property name containing the bad data. It's already a public information.

dunglas (changes requested):

PR #3702, opened 2020-09-01
Fix/cov

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,8 @@ jobs:
             phpstan-
         continue-on-error: true
       - name: Run PHPStan analysis
+        env:
+          SYMFONY_PHPUNIT_VERSION: '9.2'
         run: vendor/bin/phpstan analyse --no-progress --no-interaction --ansi
 
   phpunit:
--- a/src/Bridge/Symfony/Bundle/Test/Constraint/ArraySubset.php
+++ b/src/Bridge/Symfony/Bundle/Test/Constraint/ArraySubset.php
@@ -18,7 +18,7 @@ use PHPUnit\SebastianBergmann\Comparator\ComparisonFailure;
 use SebastianBergmann\Comparator\ComparisonFailure as LegacyComparisonFailure;
 
 if (!class_exists(ComparisonFailure::class)) {
-    class_alias(LegacyComparisonFailure::class, ComparisonFailure::class);
+    class_alias(LegacyComparisonFailure::class, 'PHPUnit\SebastianBergmann\Comparator\ComparisonFailure');
 }
 
 /**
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           composer update --no-progress --no-suggest --ansi
       - name: Install PHPUnit
         env:
-          SYMFONY_PHPUNIT_VERSION: '9.2'
+          SYMFONY_PHPUNIT_VERSION: '8.3'
```
dunglas, line 123: It's better to run PHPStan with the latest versions of all deps.

src/Bridge/Symfony/Bundle/Test/Constraint/ArraySubset.php
```diff
@@ -13,94 +13,15 @@
 
 namespace ApiPlatform\Core\Bridge\Symfony\Bundle\Test\Constraint;
 
-use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\SebastianBergmann\Comparator\ComparisonFailure;
 use SebastianBergmann\Comparator\ComparisonFailure as LegacyComparisonFailure;
 
 if (!class_exists(ComparisonFailure::class)) {
-    class_alias(LegacyComparisonFailure::class, ComparisonFailure::class);
+    class_alias(LegacyComparisonFailure::class, 'PHPUnit\SebastianBergmann\Comparator\ComparisonFailure');
```
dunglas, line 20: You can keep using ::class even if the class doesn't exist.

src/Bridge/Symfony/Bundle/Test/Constraint/ArraySubsetV9.php
```diff
@@ -0,0 +1,42 @@
...
+ * subject array.
+ *
+ * Imported from dms/phpunit-arraysubset-asserts, because the original constraint has been deprecated.
+ *
+ * @copyright Sebastian Bergmann <sebastian@phpunit.de>
+ * @copyright Rafael Dohms <rdohms@gmail.com>
+ *
+ * @see https://github.com/sebastianbergmann/phpunit/issues/3494
+ */
+final class ArraySubsetV9 extends Constraint
```
dunglas, line 31: Mark these classes as internal maybe?

src/Bridge/Symfony/Bundle/Test/Constraint/ArraySubset.php
soyuka: phpstan does not agree

soyuka: When using `::class`, phpstan loads the class and fails because it doesn't find it.

dunglas:
Thanks @soyuka!

PR #3705, opened 2020-09-02
Fix definition names to be compliant with OpenAPI 3.0 and JSON Point Spec

### LINKED ISSUE #1969: Swagger documentation: Could not resolve reference

stchr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | fixes #1969
| License       | MIT

Replaces PR #3669 

From https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#componentsObject
`All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.`

Invalid characters will be replaced by `.`. (see https://github.com/api-platform/core/pull/3669#issuecomment-681941187)

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -255,7 +255,7 @@ final class SchemaFactory implements SchemaFactoryInterface
 
     private function encodeDefinitionName(string $name): string
     {
-        return preg_replace('/^[a-zA-Z0-9.\-_]+$./', '.', $name);
+        return preg_replace('/[^a-zA-Z0-9.\-_]/', '.', $name);
     }
 
     private function getMetadata(string $className, string $type = Schema::TYPE_OUTPUT, ?string $operationType, ?string $operationName, ?array $serializerContext): ?array
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -235,12 +235,12 @@ final class SchemaFactory implements SchemaFactoryInterface
         if (null !== $inputOrOutputClass && $className !== $inputOrOutputClass) {
             $parts = explode('\\', $inputOrOutputClass);
             $shortName = end($parts);
-            $prefix .= ':'.$shortName;
+            $prefix .= '.'.$shortName;
         }
 
         if (isset($this->distinctFormats[$format])) {
             // JSON is the default, and so isn't included in the definition name
-            $prefix .= ':'.$format;
+            $prefix .= '.'.$format;
         }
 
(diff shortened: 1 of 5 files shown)


dunglas (approved):
LGTM!

src/JsonSchema/SchemaFactory.php
```diff
@@ -250,7 +250,12 @@ private function buildDefinitionName(string $className, string $format = 'json',
             $name = $groups ? sprintf('%s-%s', $prefix, implode('_', $groups)) : $prefix;
         }
 
-        return $name;
+        return $this->encodeDefinitionName($name);
+    }
+
+    private function encodeDefinitionName(string $name): string
+    {
+        return preg_replace('/^[a-zA-Z0-9.\-_]+$./', '.', $name);
```
cziegenberg, line 258: Your commit didn't work for me (tested with a "shortName" set to "directory/sub-directory" - still got the error "Could not resolve reference: #/components/schemas/..."), and I think the problem is the regular expression in that line. I changed it to `preg_replace('/[^a-zA-Z0-9.\-_]+/', '.', $name);` and then it worked for me - but I don't know if this is correct.

stchr: @cziegenberg the correct regex should be `preg_replace('/[^a-zA-Z0-9.\-_]/', '.', $name)` in our case.
the additional `+` only says that the definition name must have 1 or more characters. But our use case is only to place INVALID characters.

stchr: @cziegenberg Should be fixed. Can you test?

@dunglas Please review.

cziegenberg:
@stchr I couldn't test it completely, because of other problems with the master branch, but the aforementioned error message doesn't occur anymore.

src/JsonSchema/SchemaFactory.php
cziegenberg: @stchr I couldn't test it completely, because of other problems with the master branch, but the aforementioned error message doesn't occur anymore.

stchr:
@dunglas Can be merged IMHO. Thanks!

dunglas (approved):

dunglas:
Thanks a lot @stchr!

PR #3708, opened 2020-09-04
Create client uses an existing kernel if one is present before the call

sqrt-1764:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

As discussed in api-platform/api-platform#1640 the current behavior of `ApiTestCase::createClient()` can lead to problems.
Currently createClient boots a new kernel upon every call. This leads to problems if a kernel was already booted and e.g. used to create entities with Doctrine. In that case `hautelook/alice-bundle::ResetDatabaseTrait()` does not work as expected any more.

So I created a workaround:
createClient() only boots a new kernel if `static::$kernel` isn't defined yet. Otherwise a kernel is already present and there is no need to boot a new one.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
+++ b/src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
@@ -41,19 +41,24 @@ abstract class ApiTestCase extends KernelTestCase
 
     /**
      * Creates a Client.
+     * In case there is an already booted kernel, that kernel is used and therefore 
+     * the $kernelOptions are ignored.
      *
-     * @param array $kernelOptions  Options to pass to the createKernel method
+     * @param array $kernelOptions  Options to pass to the createKernel method if
+     *                              no kernel is booted before this call
      * @param array $defaultOptions Default options for the requests
      */
     protected static function createClient(array $kernelOptions = [], array $defaultOptions = []): Client
     {
-        $kernel = static::bootKernel($kernelOptions);
+        if (static::$kernel === null) {
+            static::bootKernel($kernelOptions);
+        }
 
         try {
             /**
              * @var Client
              */
-            $client = $kernel->getContainer()->get('test.api_platform.client');
+            $client = static::$kernel->getContainer()->get('test.api_platform.client');
         } catch (ServiceNotFoundException $e) {
             if (!class_exists(AbstractBrowser::class) || !trait_exists(HttpClientTrait::class)) {
                 throw new \LogicException('You cannot create the client used in functional tests if the BrowserKit and HttpClient components are not available. Try running "composer require --dev symfony/browser-kit symfony/http-client".');
--- a/tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
+++ b/tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
@@ -24,6 +24,27 @@ use PHPUnit\Runner\Version;
 
 class ApiTestCaseTest extends ApiTestCase
 {
+    public function testCreateClientBootsKernelIfNotPresent(): void
+    {
+        $this->assertNull(static::$kernel, "Kernel not booted");
+
+        $this->createClient();
+
+        $this->assertNotNull(static::$kernel, "Kernel booted by createClient()");
+    }
+
+    public function testCreateClientUsesExistingKernelIfPresent(): void
+    {
+        $this->assertNull(static::$kernel, "Kernel not booted");
+
+        $initialKernel = self::bootKernel();
+        $this->assertNotNull(static::$kernel, "Kernel booted successfully");
+
+        $this->createClient();
+
+        $this->assertSame(static::$kernel, $initialKernel, "Kernel not changed by createClient");
+    }
+
     public function testAssertJsonContains(): void
     {
         if (version_compare(Version::id(), '8.0.0', '<')) {


tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
```diff
@@ -24,6 +24,27 @@
 
 class ApiTestCaseTest extends ApiTestCase
 {
+    public function testCreateClientBootsKernelIfNotPresent(): void
+    {
+        $this->assertNull(static::$kernel, "Kernel not booted");
```
sqrt-1764, line 29: Not sure what to do with this check failure.
I want to make sure, that the preconditions for my tests are fulfilled. If somebody later changes the code that a static::$kernel === null is no longer the default-value, my change would break. So I included this in the test to make it bullet-proof ...

src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
```diff
@@ -41,19 +41,24 @@ protected function tearDown(): void
...
      *
-     * @param array $kernelOptions  Options to pass to the createKernel method
+     * @param array $kernelOptions  Options to pass to the createKernel method if
+     *                              no kernel is booted before this call
      * @param array $defaultOptions Default options for the requests
      */
     protected static function createClient(array $kernelOptions = [], array $defaultOptions = []): Client
     {
-        $kernel = static::bootKernel($kernelOptions);
+        if (static::$kernel === null) {
```
dunglas, line 53: While reviewing I compared with the upstream `createClient` function from Symfony: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Test/WebTestCase.php#L40

I may be missing something but it looks like Symfony never reuse an existing kernel.

sqrt-1764: To be consistent with upstream is a strong argument, indeed. In any case that should be the final state.

Throwing an exception like upstream would have shown me my problem (although I would have to have read the code to understand, that createClient() also boots a kernel) - so it is a solution.

But in the other hand:
I am thinking of the following scenario:
A big test-class needs some test-entities. Maybe some variantions according to the individual tests. But that needs a kernel in any case. 
But also lets assume only a minority of the tests in that class need a client the rest only a kernel.

It depends how expensive the creation of the client is. If it is cheap - OK, we could always create a client to get a booted kernel (but that behaviour would then have to be documented better).
But if it is expensive to create a kernel it would be better to only boot the kernel when we do not need the client for the test. In that case we could not put the initialization of the entities in a `setUp()` method. We would have to call the correct variant of the initialization, one only with a kernel, one with a client, in every test.

My scenario might be not realistic as the way we usually structure our tests might imply that all tests of a class might need a client. But is it always the case?

But maybe the solution could be to also change upstream according to this solution here.

As far as I can see it would break no existing code. But it would open the option to create a kernel before the `createClient()` to do whatever has to be done before that call. (Which also would lead to a consitent behaviour between Symfony and Api-Platform ;-) )

sqrt-1764: [Followup] If we decide to go the route of my proposal, I would like to add an exception in case the kerneloptions-paramter is not empty and a kernel is reused.

tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
dunglas: Don't worry about this one, it's unrelated with your PR and is being fixed in #3702.

src/Bridge/Symfony/Bundle/Test/ApiTestCase.php
dunglas: I would prefer to keep consistent with the behavior of Symfony's `WebTestCase`. What I suggest is to throw the exception for now, and to open an issue on the Symfony bug tracker to propose your behavior. If it changes in SF, then we'll change it accordingly here.

`ApiTestCase` is designed to be as close as possible of `WebTestCase` and `PantherTestCase`.

WDYT?

sqrt-1764: Agreed. Adding that exception like in `WebTestCase` is an improvement worth a change until upstream might be changed in favour of my solution which then has to implemented also here.
Will change that accordingly.

(So my first PR jumps to another project 8-) )

sqrt-1764: @dunglas Unfortunately inserting the exception according to upstream breaks 7 existing tests:
```
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ApiTestCaseTest::testFindIriBy
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ClientTest::testRequest
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ClientTest::testCustomHeader
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ClientTest::testDefaultHeaders
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ClientTest::testAuthBasic with data set #0 ('dunglas:kevin')
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ClientTest::testAuthBasic with data set #1 (array('dunglas', 'kevin'))
ApiPlatform\Core\Tests\Bridge\Symfony\Bundle\Test\ClientTest::testComplexScenario
```
All these tests now throw the new LogicException because a kernel is booted before calling createClient.

IMHO changing those tests does not make much sense or might be difficult.

So suspending this PR and wait for the discussed change from upstream might be the way to go?

sqrt-1764:
F.Y.I.: Created issud symfony/symfony#38255 for the corresponding change in upstream.

soyuka:
I think that the behavior of creating a new Kernel on every test is correct. This goes with Alice using a transaction during each test to avoid side-effects. If you need the same data across tests I'd suggest to use the fixtures and load this data. 
You can always load some more fixtures in a specific test file using: 

```
/**
     * @before
     */
    public function setupSomeFixtures()
    {
        self::$kernel->getContainer()->get('doctrine')
    }
```

sqrt-1764:
@soyuka: Creating a fresh kernel for every test is not the question for me. A fresh kernel for every test is indeed the way to go.
My point is that creating the kernel in `ApiTestCase::createKernel()` is not the cleanest approach. It's the lazy approach that covers many cases - but not all.
There are cases where it makes sense to prepare test-data before creating the test-client. For this a kernel is needed.
For me this is the "natural" approach and when running into this problem I created a kernel myself, populated my data and then created the kernel. This did not work with the Alice-Bundle. My Data was not visible after createClient().

Currently I am able to rearrange my code so that the client is created first. But I doubt that this is always possible.
Preparing the similar PR for upstream (symfony/symfony#38255) which has a more secure implementation of createClient() I also found some tests that used a kernel before createClient(). There they explicitly killed that kernel before calling createKernel(). At the first glance this seems not to be necessary if createClient (re)uses an already created kernel - and hence my PR also there.

soyuka:
The correct fix is in https://github.com/api-platform/api-platform/issues/1668 can you confirm ?

sqrt-1764:
@soyuka Sorry for the delay - missed the notification mail and was busy on other topics ...

As far as I can see api-platform/api-platform#1668 addresses the same problem but makes a different proposal for a solution.

IMHO the problem starts with the fact that createClient() boots a fresh kernel in every case. Booting a fresh kernel might be convenient in most cases, but there are cases (like the case described in api-platform/api-platform#1668) where this behaviour interferes with the logic of the test-case.

As said before IMO the best solution would be, when createClient wuold reuse an already booted kernel instead of creating a fresh one. To keep the behaviour consistent with Symfony after discussion with @dunglas I created symfony/symfony#38255 to implement that fix in Symfony and then port it to Api-Platform afterwards.

Unfortunately I got no feedback there yet. Maybe shift the discussion there to get things going?

soyuka:
:+1: but what about `disableReboot`? Doesn't that fix your issue?

sqrt-1764:
The problem is, that this internal behaviour of the client is not documented. 
I was not aware of the method `disableReboot`, as it is nowhere documented in the docs I got hold of. And only when searching for this specific method I found some articles about it. The fact that client::request() also reboots the kernel also is not officially mentioned in the docs a user usually would read ...
And `disableReboot` only affects `Client::request()`, not `ApiTestCase::createClient()`

Thinking deeper about testing http-requests it seems to make sense to start with a fresh kernel upon every request in order to precisely mimic the situation for a real request. In a real request, one will usually also start with a fresh kernel. So far, so good.

But my initial point was that I created some configuration of the test-environment _before_ I created the client. I could reorder my statements to create the client first, but I could think of situations where this might not be feasible.

My proposal is to reuse an existing client if it exists prior to `createClient`. Currently `createClient` always creates a new kernel.

About `disableReboot()` I would have to think deeper:
The current behaviour is that it only reboots after the _first_ request. (If it also would reboot on the first request, the problems of api-platform/api-platform#1668 would also be present on the first request) In the sourcecode there is this property `hasPerformedRequest` ...

But anyway ... isn't this better to be discussed in symfony/symfony as the behaviour here is to be a copy of a behaviour there?

PR #3711, opened 2020-09-08
Bugfix | Don't try to set `null` values when merging metadata

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

Fatal error:
```bash
TypeError: Argument 1 passed to ApiPlatform\Core\Metadata\Resource\ResourceMetadata::withDescription() must be of the type string, null given, called in /var/www/vendor/api-platform/core/src/Metadata/Resource/Factory/AnnotationResourceMetadataFactory.php on line 120
```

This was cause when trying to merge two metadata, parent & child ones. By default `ApiPlatform\Core\Annotation\ApiResource` (as well as parent here: `ApiPlatform\Core\Metadata\Resource\ResourceMetadata`) has some of fields marked as default = `null`, when merging right now it's not validated that child has value set, thus due too type-hint it fails with `null` values.

 Dumps (parent + child):
```bash
^ ApiPlatform\Core\Metadata\Resource\ResourceMetadata^ {#9948
  -shortName: "Administrator"
  -description: null
  -iri: null
  -itemOperations: array:3 [
    "get" => []
    "put" => array:1 [
      "denormalization_context" => array:1 [
        "groups" => "admin_user:update"
      ]
    ]
    "delete" => []
  ]
  -collectionOperations: array:2 [
    "get" => []
    "post" => array:1 [
      "validation_groups" => array:2 [
        0 => "sylius"
        1 => "sylius_user_create"
      ]
    ]
  ]
  -subresourceOperations: null
  -graphql: null
  -attributes: array:4 [
    "route_prefix" => "admin"
    "normalization_context" => array:1 [
      "groups" => array:1 [
        0 => "admin_user:read"
      ]
    ]
    "denormalization_context" => array:1 [
      "groups" => array:1 [
        0 => "admin_user:create"
      ]
    ]
    "validation_groups" => "sylius"
  ]
}
```
```bash
ApiPlatform\Core\Annotation\ApiResource^ {#9922
  +collectionOperations: array:2 [
    0 => "get"
    "post" => array:2 [
      "security" => "is_granted('ROLE_ADMIN')"
      "security_message" => "Only admins can add doctors."
    ]
  ]
  +description: null
  +graphql: null
  +iri: null
  +itemOperations: array:2 [
    0 => "get"
    "patch" => array:1 [
[...]

PR #3724, opened 2020-09-15
Allow overwriting parent metadata resource informations

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes(?)
| New feature?  | yes(?)
| BC breaks?    | no(?)
| Deprecations? | no
| License       | MIT

This is both IMHO bugfix & potentially new feature with BC break (but IMO that BC break is good cause old approach was buggy).

Currently parent resource data is always being selected if set, while child is silently ignored, this changes the approach to allow extending resource change parent data. For example in Sylius predefined resources has `shortName`, but if you extend that model & add own `@ApiResource` it will ignore the setting you try to use **if** it was already set on the extended resource.

If you think it's correct, I will add tests soon to prove the problem & fix.

ps. not sure if this should land in master or be considered as bugfix & go on 2.5...

### DIFF
--- a/src/Metadata/Resource/Factory/AnnotationResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/AnnotationResourceMetadataFactory.php
@@ -107,8 +107,13 @@ final class AnnotationResourceMetadataFactory implements ResourceMetadataFactory
         $upperProperty = ucfirst($property);
         $getter = "get$upperProperty";
 
-        if (null !== $resourceMetadata->{$getter}()) {
-            return $resourceMetadata;
+        $currentValue = $resourceMetadata->{$getter}();
+        if (null !== $currentValue) {
+            if (null === $value) {
+                $value = $currentValue;
+            } elseif (is_array($currentValue)) {
+                $value = array_merge_recursive($currentValue, $value);
+            }
         }
 
         $wither = "with$upperProperty";
--- a/tests/Metadata/Resource/Factory/AnnotationResourceMetadataFactoryTest.php
+++ b/tests/Metadata/Resource/Factory/AnnotationResourceMetadataFactoryTest.php
@@ -34,7 +34,7 @@ class AnnotationResourceMetadataFactoryTest extends TestCase
     /**
      * @dataProvider getCreateDependencies
      */
-    public function testCreate($reader, $decorated, string $expectedShortName, string $expectedDescription)
+    public function testCreate($reader, $decorated, string $expectedShortName, ?string $expectedDescription)
     {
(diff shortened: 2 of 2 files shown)


src/Metadata/Resource/Factory/AnnotationResourceMetadataFactory.php
```diff
@@ -111,8 +111,13 @@ private function createWith(ResourceMetadata $resourceMetadata, string $property
...
         $getter = "get$upperProperty";
 
-        if (null !== $resourceMetadata->{$getter}()) {
-            return $resourceMetadata;
+        $currentValue = $resourceMetadata->{$getter}();
+        if (null !== $currentValue) {
+            if (null === $value) {
+                $value = $currentValue;
+            } elseif (is_array($currentValue)) {
+                $value = array_merge($currentValue, $value);
```
alanpoulain, line 119: Shouldn't it be recursive?

alanpoulain:
IMO it's more a bugfix. It hasn't been detected before because Sylius uses inheritance and extending heavily. Could be a BC break too though...

src/Metadata/Resource/Factory/AnnotationResourceMetadataFactory.php
stloyd: Didn't notice that need TBH, only one level on my Sylius based project. 

Dumps were exactly like in: https://github.com/api-platform/core/pull/3711

dunglas:
If it's in a potential BC break, please at least don't merge it in 2.5. It should also be documented.

stloyd:
> If it's in a potential BC break, please at least don't merge it in 2.5. It should also be documented.
@dunglas It can be BC break **only** in case if someone always had parent without any resource extending it, if right now he want's to extend it he must do **big copy&paste** from parent resource declaration & insert his/her own there to be able to replace parent resource data.

From my perspective this is bugfix cause from beginning it had _broken_ logic (parent class winning over child/extending one).

**But** if we're not sure how to proceed with this change, I'm totally okey going (as doing that right now already) against master/2.6-dev branch, locally I'm overloading the whole factory class to fix this bug & most likely will propose that to Sylius, cause developers using that project will be most affected ones.

Will try to document this change soon too.

Cheers!

lchrusciel:
Hey @stloyd, in case it will not land in 2.5, we can temporarily host it in Sylius, as internal. But of course, I would prefer not to. Still, thanks for solving this issue 👍

soyuka:
IMO this is a bug fix as the behavior described in the issue should be the default. I mean we do expect the Child resource to override the parent's one if null (please confirm my sentence). 

Can you target 2.5 ? I'd like a functional test to illustrate the fix if you can.

PR #3727, opened 2020-09-16
Fixing graphql error when requesting a null array

### LINKED ISSUE #3465: [GraphQl] Error in response when value is null

**Description**  
When a selected field is empty in database, the reponse of a GraphQl request contains the following error, but the result is still correct. The GraphQl query is correct and when I fill this data in DB, error disappears.

query : 
```
{
  item(id: "cc429bad-960a-4b1d-94bc-7188f8f7a8c0") {
    id
    name
    emails
  }
}
```

error : 
```
{
  "errors": [
    {
      "message": "Resource class cannot be determined.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ],
      "path": [
        "item",
        "emails"
      ]
    }
  ],
  "data": {
    "item": {
      "id": "cc429bad-960a-4b1d-94bc-7188f8f7a8c0",
      "name": "Item name",
      "emails": null
    }
  }
}
``` 
This error appears after apiplatform/core upgrade from 2.4.5 to 2.5.4.

moduleon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3465
| License       | MIT
| Doc PR        | api-platform/docs

This PR fixes graphql error thrown when requesting a property having a null value (notably array).

### DIFF
--- a/src/GraphQl/Resolver/Factory/ItemResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ItemResolverFactory.php
@@ -60,7 +60,7 @@ final class ItemResolverFactory implements ResolverFactoryInterface
     {
         return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operationName) {
             // Data already fetched and normalized (field or nested resource)
-            if (isset($source[$info->fieldName])) {
+            if (isset($source[$info->fieldName]) || (\is_array($source) && \array_key_exists($info->fieldName, $source) && null === $source[$info->fieldName] && null === $resourceClass)) {
                 return $source[$info->fieldName];
             }
 


alanpoulain:
Hello :slightly_smiling_face:

Thanks for this bugfix. It should target `2.5`, not `master`.

Could you also add:
- a Behat test to show the issue
- a non regression unit test

src/GraphQl/Resolver/Factory/ItemResolverFactory.php
```diff
@@ -59,7 +59,7 @@ public function __invoke(?string $resourceClass = null, ?string $rootClass = nul
     {
         return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operationName) {
             // Data already fetched and normalized (field or nested resource)
-            if (isset($source[$info->fieldName])) {
+            if (isset($source[$info->fieldName]) || (\is_array($source) && \array_key_exists($info->fieldName, $source) && null === $source[$info->fieldName] && null === $resourceClass)) {
```
alanpoulain, line 62: Wouldn't it better like this:
```php
null === $source[$info->fieldName] ?? false
```

Why checking if `$resourceClass` is `null`?

moduleon: Hi Alan,
If I'm not missing something, when resolver is used to fetch a property inside an entity, which is not a collection or sub entity, $resourceClass is null at this time. 
And everything in $source[$info->fieldName] at this point is the current value of the property requested.
So, if $resourceClass is null, and so is $source[$info->fieldName], then we should not go further and return what's inside $source[$info->fieldName] right away. 
If we go further, the resolver try to get an entity, but the property requested is not one, and we have an error telling `Resource class cannot be determined`.

alanpoulain: OK I think I understand the issue now, thanks for the explanation.
We should probably replace the `isset($source[$info->fieldName])` with a call to `array_key_exists` then (without checking if the `$resourceClass` is `null` since it would handle all the cases), don't you think?
But to be sure, we need to have a correct Behat test.

moduleon: Actually I'm not sure that we could erase the condition on $resourceClass. 
Because $source[$info->fieldName] can be null for a reason, and notably because we must fetch a sub entity or a collection in a new query (with specific filters, pagination or order). In that case $source[$info->fieldName] is null and $resourceClass is not.
But we can wipe out the first part of the condition I guess: isset($source[$info->fieldName]), which seems totally useless now.
I will try that and will write the Behat feature for reproducing the error.

alanpoulain:
Superseded by https://github.com/api-platform/core/pull/4092. Thank you @moduleon.

PR #3728, opened 2020-09-16
allow multiple topics mercure

jordan15:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes ...
| License       | MIT
| Doc PR        | api-platform/docs...

allow multiple topics mercure, the class Update check if $iri is String or Array and after cast as Array. So buildUpdate must not force type String for $iri parameter.

With this change we can set several topics.


dunglas (approved):
Good catch. Could you target the 2.5 branch as it's a bug fix please? It would be nice to add a test too.

divine:
This might be closed now as another PR was merged with the fix and test.

Thanks!

PR #3733, opened 2020-09-22
Add router loaders

### LINKED ISSUE #3539: Resource setting is very rough
**Description**  
Hello. Resource setting is very rough. 

`src/Bridge/Symfony/Routing/ApiLoader.php `

I suggest:
* Select the loading of routes from the list of files in a separate service. Perhaps **ApiFileLoader** for loading routes from a specific `file_path`.
* Create additional route loaders.
    * from bundles directory - _api_directory_

### ApiDirectoryLoader

Like Symfony Routing `Loader/DirectoryLoader.php`

To be able to load routes from a third-party bundle and specify their prefix for them.

**Example** 

### api_directory (type)

```yaml
#config/routes/api_platform.yaml
api_some_bundle:
    resource: '@BundleName/Entity/' # eg
    type: api_directory
    prefix: /some/api 
```

#### ApiDirectoryLoader
```php
    public function load($file, string $type = null)
    {
        $path = $this->locator->locate($file);

        $collection = new RouteCollection();
        $collection->addResource(new DirectoryResource($path));

        foreach (scandir($path) as $dir) {
            if ('.' !== $dir[0]) {
                $this->setCurrentDir($path);
                $subPath = $path.'/'.$dir;
                $subType = null;

                if (is_dir($subPath)) {
                    $subPath .= '/';
                    $subType = 'api_directory';
                }

                $subCollection = $this->import($subPath, $subType, false, $path);

                // TODO: Use a dedicated service 
                // to load routes from resource files marked @ApiPlatform annotation
                // It seems to work out just using a redirect to the loader `api_file` ApiFileLoader
                // just by specifying `$subType = 'api_file';` (and creating ApiFileLoader service)
                // -----------------------------------------------------------------------------------------
            }
        }

        return $collection;
    }
```

---

Transfer almost everything from `ApiLoader.php` to new `ApiFileLoader.php` - _api_file_
`ApiLoader.php` - leave only receiving a list of files from the config and address loading the route from a specific file in a loop for `ApiFileLoader.php` like `$this->import($filePath, 'api_file', false, $path);`

tebaly:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3539 
| License       | MIT

Hey. I have expanded the resource loading (routing) capabilities. As I could. At some point I needed it. Added options to the config. Described separately in thread #3539 
* **I cannot write tests**

**I merge this branch into my master long ago. I'll try to figure it out, maybe there were changes after.**


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -268,31 +280,48 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
...
-                $paths[] = "$dirname/Document";
-            }
-
-            foreach ($paths as $path) {
-                if ($container->fileExists($path, false)) {
-                    $bundlesResourcesPaths[] = $path;
-                }
+        }
+        if ($config['resources']['sources']['doctrine']) {
+            $paths[] = "$dirname/Entity";
```
soyuka, line 311: You can change this in doctrine settings

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -214,6 +215,54 @@ public function getConfigTreeBuilder()
         return $treeBuilder;
     }
 
+    private function addRoutingLoaderSection(ArrayNodeDefinition $rootNode)
+    {
+        $rootNode
+            ->children()
+                ->arrayNode('routing')
```
soyuka, line 222: this can be changed in symfony's settings

soyuka:
This looks overengineered. I don't understand the need behind it? To me most of the changes here can be overriden in doctrine or symfony's config. If you really need to remove some routes you can override the ApiLoader as I suggested at some point.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
tebaly: Are you sure you understand the essence of the changes?
https://github.com/api-platform/core/blob/5b2cfdcd8feea9b9ff96371df1a7eeccc72432f4/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L309
This is the code from the master. Why should I change the settings of the doctrine if I need to influence the settings of the Platform?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
tebaly: Are you sure you understand the essence of the changes?

I added this setting to manage Api Platform routing, not Symphony's global routing. Do you have any idea that your bundle can be used together and as an add-on? Why do I need to change the configured and running Symphony application when I need to change the settings of the Api Platform?

Issue #2632, opened 2019-03-20
ItemNormalizer::updateObjectToPopulate should use getItemIriFromResourceClass instead of getIriFromResourceClass


Inari-Whitebear:
Maybe I'm wrong with this, but
https://github.com/api-platform/core/blob/master/src/Serializer/ItemNormalizer.php#L98
seems to fetch the collection route and stick an identifier on. Shouldn't it instead use `getItemIriFromResourceClass` ( https://github.com/api-platform/core/blob/master/src/Api/IriConverterInterface.php#L59 ) to fetch the route for the item directly with the identifier?

dunglas:
Indeed (this code is probably older than the introduction of `getItemIriFromResourceClass`). Would you mind opening a pull request?

soyuka:
let's follow this in #3609

PR #3735, opened 2020-09-23
Get IRI for related objects from proper variables …

### LINKED ISSUE #3609: PATCHing a subresource returns a "No route matches %s" error
**API Platform version(s) affected**: v2.5.6 (and before)

**Description**  
When sending a PATCH request for a resource containing a subresource, the request fails with a "No route matches %s" error.
This is because the subresource is not found.

The culprit seems to be in the ItemNormalizer::updateObjectToPopulate.

https://github.com/api-platform/core/blob/master/src/Serializer/ItemNormalizer.php#L76

```php
$context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri((string) $data['id'], $context + ['fetch_data' => true]);
```

It should be `data['@id']` instead of `$data['id']` when calling `getItemFromIri`

**How to reproduce**  

Given a `Resource` entity with a `subresource` field of type `Subresource` entity on a OneToOne relationship.

```php
namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiSubresource;

/**
 * @ApiResource()
 */
class Resource
{
    /**
     * @ORM\OneToOne(targetEntity=Subresource::class, cascade={"persist", "remove"})
     */
     private $resource;

    public function getSubresource(): ?Subresource
    {
        return $this->subresource;
    }

    public function setSubresource(?Subresource $subresource): self
    {
        $this->subresource = $subresource;
        return $this;
    }
}
```

```php
namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiSubresource;

/**
 * @ApiResource()
 */
class Subresource
{
    /**
     * @ORM\Column(type="string", length=255, nullable=true)
     */
     private $someField;

    public function getSomeField(): ?string
    {
        return $this->someField;
    }

    public function setSomeField(?string $someField): self
    {
        $this->someField = $someField;
        return $this;
    }
}
```

Trying to send a PATCH request which will update the subresource (given that you already have created the resource + subresource first :D)

```json
{
  "@id": "/api/resource/1",
  "subresource": {
     "@id": "/api/subresource/1",
     "someField": "someValue"
  }
}
```

**Possible Solution**  

It should be `data['@id']` instead of `$data['id']` when calling `getItemFromIri`.
Tried it in local and the error goes away.

I am not confident of any side effect this "fix" could have as I am not a expert of the internals of Api Platform.

edit: typo

ntomka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3609 #2632 
| License       | MIT
| Doc PR        | -

This PR is based on the conversation in #3609 .

### DIFF
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -51,7 +51,7 @@ class ItemNormalizer extends AbstractItemNormalizer
     public function denormalize($data, $class, $format = null, array $context = [])
     {
         // Avoid issues with proxies if we populated the object
-        if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
+        if (isset($data['@id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
             if (isset($context['api_allow_update']) && true !== $context['api_allow_update']) {
                 throw new NotNormalizableValueException('Update is not allowed for this operation.');
             }
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -20,6 +20,7 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
+use ApiPlatform\Core\Security\ResourceAccessCheckerInterface;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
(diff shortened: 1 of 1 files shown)


dunglas:
The patch doesn't look good. It breaks the tests.

src/Serializer/ItemNormalizer.php
```diff
@@ -72,7 +72,7 @@ public function denormalize($data, $class, $format = null, array $context = [])
     private function updateObjectToPopulate(array $data, array &$context): void
     {
         try {
-            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri((string) $data['id'], $context + ['fetch_data' => true]);
+            $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getItemFromIri((string) $data['@id'], $context + ['fetch_data' => true]);
```
soyuka, line 75: this is wrong imo

soyuka:
This is not that easy, we'll work on this after fixing issues related to https://github.com/api-platform/core/issues/3609#issuecomment-723560213

PR #3736, opened 2020-09-24
Allow Validator to validate a custom constraint set

mtarld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | TODO

Currently, `src/Validator/ValidatorInterface.php` can't validate a custom set of constraints. It only allows validating a whole object (usually an ApiResource).

But sometimes, like in Symfony, we do need to validate only a custom set of constraints. Therefore we are forced to use Symfony's validator and do the following:
```php
// Symfony\Component\Validator\Validator\ValidatorInterface $this->validator

$errors = $this->validator->validate($product, [new UniqueProductReference()]);
if ($errors->count()) {
    throw new ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException($errors);
}
```

With this PR, I propose to add the custom constraint set validation in `ApiPlatform\Core\Validator\ValidatorInterface`, so that we can do this instead:
```php
// ApiPlatform\Core\Validator\ValidatorInterface $this->validator

$this->validator->validate($product, [], [new UniqueProductReference()]); // Could directly throw a ValidationException
```


src/Validator/ValidatorInterface.php
```diff
@@ -25,9 +26,9 @@ interface ValidatorInterface
     /**
      * Validates an item.
      *
-     * @param object $data
+     * @param Constraint[]|null $constraints
      *
      * @throws ValidationException
      */
-    public function validate($data, array $context = []);
+    public function validate($data, array $context = [], ?array $constraints = null);
```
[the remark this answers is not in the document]
mtarld: You're completely right, I'll handle the BC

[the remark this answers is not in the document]
dunglas: You can create an inherited interface to do add the extra parameter.

dunglas:
Instead of introducing our own interface, maybe should we add the `ValidatorInterface` to `symfony/contracts` and just use it? WDYT @nicolas-grekas?

nicolas-grekas:
In Symfony, there is already `ConstraintValidatorInterface` which has the same purpose, but with a different signature.

Before adding anything to contracts it could make sense to make both interfaces in SF and API converge somehow.

src/Validator/ValidatorInterface.php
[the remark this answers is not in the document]
mtarld: Good idea, I created a ConstraintValidatorInterface and deprecated the ValidatorInterface.

src/Validator/EventListener/ValidateListener.php
```diff
@@ -35,7 +35,7 @@ final class ValidateListener
     private $validator;
     private $resourceMetadataFactory;
 
-    public function __construct(ValidatorInterface $validator, ResourceMetadataFactoryInterface $resourceMetadataFactory)
+    public function __construct(ConstraintValidatorInterface $validator, ResourceMetadataFactoryInterface $resourceMetadataFactory)
```
soyuka, line 38: This is a backward compatibility break no?

src/Bridge/Symfony/Validator/Validator.php
```diff
@@ -26,7 +26,7 @@
  *
  * @final
  */
-class Validator implements ValidatorInterface
+class Validator implements ConstraintValidatorInterface
```
soyuka, line 29: We can't change an interface, the backward compatible solution is to add another interface that extends the first one. We did this with the ContextAwareDataProviderInterface

src/Validator/EventListener/ValidateListener.php
mtarld: The new `ConstraintValidatorInterface` is extending the actual `ValidatorInterface`. Therefore it's ok no?

src/Bridge/Symfony/Validator/Validator.php
mtarld: Same response as above. As `ConstraintValidatorInterface` is extending `ValidatorInterface`, it should not cause any BC issues. WDYT?

src/Validator/EventListener/ValidateListener.php
alanpoulain: No, @soyuka is right, this is a BC.
See for instance:
https://3v4l.org/H9fWU

norkunas:
Should the original `ValidatorInterface` be really deprecated? When you do not provide any constraints to the validator then there would be no changes to the code if it would not be deprecated and just when needed to provide constraints then the new interface could be used.

mtarld:
> Should the original `ValidatorInterface` be really deprecated? When you do not provide any constraints to the validator then there would be no changes to the code if it would not be deprecated and just when needed to provide constraints then the new interface could be used.

Indeed, as it's just extending the default behavior, deprecation isn't needed. Thanks!

src/Bridge/Symfony/Bundle/Resources/config/validator.xml
```diff
@@ -9,7 +9,7 @@
             <argument type="service" id="validator" />
             <argument type="service" id="service_container" />
         </service>
-        <service id="ApiPlatform\Core\Validator\ValidatorInterface" alias="api_platform.validator" />
+        <service id="ApiPlatform\Core\Validator\ConstraintValidatorInterface" alias="api_platform.validator" />
```
alanpoulain, line 12: This one is a BC break. If a user is autowiring the `ValidatorInterface`, it will not receive the `api_platform.validator` anymore.

mtarld: Yes, indeed. Sorry, I'm not at ease with BC handling...
May I add another alias instead?
```diff
  <service id="ApiPlatform\Core\Validator\ValidatorInterface" alias="api_platform.validator" />
+ <service id="ApiPlatform\Core\Validator\ConstraintValidatorInterface" alias="api_platform.validator" />
```

alanpoulain: Yes you can do that :slightly_smiling_face:

alanpoulain (approved):
LGTM

src/Validator/ConstraintValidatorInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Validator;
+
+use ApiPlatform\Core\Validator\Exception\ValidationException;
+use Symfony\Component\Validator\Constraint;
```
dunglas, line 17: This creates a constraint to the Symfony validator outside the Symfony bridge. This defeats the purpose of this abstraction and will cause issue to support other validator libraries (and manual validation).

Wouldn't it be better to just use the Symfony Validator directly for this kind of use cases?

dunglas (changes requested):

PR #3739, opened 2020-09-29
Fix expression language argument for PublishMercureUpdatesListener

matthew-banning:
Argument would always be ignored as null coalescing operator has a higher precedence than ternary operator

| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       |  
| License       | MIT
| Doc PR        |


dunglas (approved):

dunglas:
Thanks @matthew-banning

PR #3741, opened 2020-09-29
Use operation output class name instead Resource short name in operati...

### LINKED ISSUE #3740: Operation output class name should be used if set instead Resource short name in summary texts

**Description**  
If you create an operation for a resource that returns a different resource class (for example making a custom operation for a subresource), Swagger UI documentation shows the Resource class a returning type, even if you set a different output class in operation parameters. 

**How to reproduce**  
Create a Resource A and a Resource B. In resource A add this collection operation:

```
    get_subresource:
      method: GET
      route_name: my_route
      output:
        class: App\ResourceB
        name: ResourceB
      openapi_context:
        parameters:
          - { name: id, in: path, required: true, type: string }
```

**Possible Solution** 
If output is set take the class name from it in operation summary, take resourceShortName otherwise

maxtorete:
…ion summary

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3740 
| License       | MIT

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -323,9 +323,10 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
         }
 
         if (OperationType::COLLECTION === $operationType) {
-            $pathOperation['summary'] ?? $pathOperation['summary'] = sprintf('Retrieves the collection of %s resources.', $resourceShortName);
+            $outputResourseShortName = $resourceMetadata->getCollectionOperations()[$operationName]['output']['name'] ?? $resourceShortName;
+            $pathOperation['summary'] ?? $pathOperation['summary'] = sprintf('Retrieves the collection of %s resources.', $outputResourseShortName);
 
-            $successResponse = ['description' => sprintf('%s collection response', $resourceShortName)];
+            $successResponse = ['description' => sprintf('%s collection response', $outputResourseShortName)];
             [$successResponse] = $this->addSchemas($v3, $successResponse, $definitions, $resourceClass, $operationType, $operationName, $mimeTypes);
 
             $pathOperation['responses'] ?? $pathOperation['responses'] = [$successStatus => $successResponse];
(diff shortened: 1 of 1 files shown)


GregoireHebert:
I quite like the idea of allowing customisation (that's why you could change the shortname), but only for the Output? It seems weird to me. The Output should just be a different representation of the same resource, so should not have a different naming per se. But I might be wrong.

maxtorete:
> I quite like the idea of allowing customisation (that's why you could change the shortname), but only for the Output? It seems weird to me. The Output should just be a different representation of the same resource, so should not have a different naming per se. But I might be wrong.

Hi GregoireHebert, thanks for your reply!

What I'm trying to achieve is a subresource custom operation. Maybe there is a better option that I haven't found. For example, something like this `/myapp/workflow/{id}/statuses` will receive an Workflow object and respond with a collection of its related WorkflowStatus.

In the Workflow resource yaml I set the collection operation:
```
collectionOperations:
    get_workflow_status_collection:
      method: GET
      route_name: api_workflow_get_workflow_status_collection
      openapi_context:
        parameters:
          - { name: id, in: path, required: true, type: string, description: 'Workflow Resource identifier' }
```

And this is the route:
```
lince_company_api_workflow_get_workflow_status_collection:
  path: /api/lince_company/workflow/{id}/statuses
  methods: ['GET']
  defaults:
    _controller: App\Controller\WorkflowController::listWorkflowStatusSubresource
    _api_resource_class: App\Entity\WorkflowStatus
    _api_collection_operation_name: get_workflow_status_collection
```

The problem is that the _api_resource_class parameter in the route definition is ignored, so I changed the collection operation to this one:
```
collectionOperations:
    get_workflow_status_collection:
      method: GET
      route_name: api_workflow_get_workflow_status_collection
      output:
        class: App\Entity\WorkflowStatus
        name: WorkflowStatus
      openapi_context:
        parameters:
          - { name: id, in: path, required: true, type: string, description: 'Workflow Resource identifier' }
```

So now it works fine, even the SwaggerUI shows the right schema in the response sample value. But the only thing wrong is the Swagger description text for that Operation, as it says that it will return a collection of the parent resource, even if it shows the schema of the subresource in the example value. You can see it in action in this screenshot form the issue https://github.com/api-platform/core/issues/3740#issuecomment-701257245

Screenshot_2020-09-30 API Platform

If the output class schema is shown in the Example value schema, why its name is not show in the description?

GregoireHebert:
>  But the only thing wrong is the Swagger description text for that Operation, as it says that it will return a collection of the parent resource

You are absolutely right on this, and this one of the biggest problems on the current subResource implementation. It should be fixed thanks to https://github.com/api-platform/core/issues/2706 which won't take place before v2.7 / v3.

But in the meantime, maybe your solution could do in 2.6? IDK.

WDYT @dunglas @soyuka?

maxtorete:
I rebased to 2.5 from upstream, the failing checks look unrelated to my commit.

soyuka:
Can you rebase ? This is correct indeed and I'd like to merge it :).

maxtorete:
> Can you rebase ? This is correct indeed and I'd like to merge it :).

Rebased to upstream :-)

soyuka:
Thanks @maxtorete

maxtorete:
> Thanks @maxtorete

You are welcome!

PR #3742, opened 2020-10-01
fix elasticsearch properties when using GQL

jannes-io:
(duplicate of a previously created PR which started from master, this one starts from 2.5.7 so we can internally set composer to use my branch until it's merged or a different solution is proposed)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes api-platform/api-platform#1646
| License       | MIT

When calling any Elasticsearch filters through GraphQL the schema rejects snake_case property names and the query sent to Elasticsearch still contains CamelCase properties.

This bugfix solves that issue by normalizing properties using CamelCaseToSnakeCaseConverter from symfony/serializer in Elasticsearch filters.


src/Bridge/Elasticsearch/DataProvider/Filter/OrderFilter.php
```diff
@@ -70,11 +70,11 @@ public function apply(array $clauseBody, string $resourceClass, ?string $operati
             $order = ['order' => $direction];
 
             if (null !== $nestedPath = $this->getNestedFieldPath($resourceClass, $property)) {
-                $nestedPath = null === $this->nameConverter ? $nestedPath : $this->nameConverter->normalize($nestedPath, $resourceClass, null, $context);
```
soyuka, line 73: can't you use the CamelCaseToSnakeCaseNameConverter directly ? Or just provide a new NameConverter that does the same thing as you added in snakeCasePropertyPath ?

jannes-io: `snakeCasePropertyPath` uses the `CamelCaseToSnakeCaseNameConverter` but also takes care of nested paths like `someCamelCaseParent.someCamelCaseChild`, which also needs to be handled. And since this needs to be done in all search filters I thought adding it to the abstract filter was the way to go.

As for providing a NameConverter, where can we do that only on entity level? And only on the ES query (keep camelCasing in input/output)?

src/Bridge/Elasticsearch/DataProvider/Filter/AbstractFilter.php
```diff
@@ -142,4 +143,20 @@ protected function getMetadata(string $resourceClass, string $property): array
...
+
+    /**
+     * Convert CamelCase properties to snake_case.
+     *
+     * @example 'some_parent.some_child' === snakeCasePropertyPath('someParent.someChild');
+     */
+    public function snakeCasePropertyPath(string $property): string
+    {
+        $unnested = explode('.', $property);
+        $normalizer = new CamelCaseToSnakeCaseNameConverter($unnested);
```
soyuka, line 155: I think there's something wrong here as we still need to use `$this->nameConverter` here.

src/Bridge/Elasticsearch/DataProvider/Filter/OrderFilter.php
soyuka: see my other comment which is more appropriate for what I meant

src/Bridge/Elasticsearch/DataProvider/Filter/OrderFilter.php
```diff
@@ -70,11 +70,11 @@ public function apply(array $clauseBody, string $resourceClass, ?string $operati
...
+                $nestedPath = $this->snakeCasePropertyPath($nestedPath);
```
soyuka, line 73: property name need to be changed as this is not always a snakeCase path

jannes-io:
Hello @soyuka , thanks for the review! I've re-added the nameConverter in the AbstractFilter, additionally I've moved the property name conversions in Match and Term filters to the AbstractSearchFilter, since this is where they both extend from.

jannes-io:
Any update?? Been 2 months now

alanpoulain:
This PR isn't right: the inner fields name converter is used to convert the property name.
See:
https://github.com/api-platform/core/blob/3557329f15da2ae3d61cc328a28e45929697d89d/src/Bridge/Symfony/Bundle/Resources/config/elasticsearch.xml#L116
And:
https://github.com/api-platform/core/blob/3557329f15da2ae3d61cc328a28e45929697d89d/src/Bridge/Elasticsearch/Serializer/NameConverter/InnerFieldsNameConverter.php#L33
And also:
https://github.com/api-platform/core/blob/3557329f15da2ae3d61cc328a28e45929697d89d/src/Bridge/Elasticsearch/Serializer/NameConverter/InnerFieldsNameConverter.php#L52-L62

If there is an issue, a Behat test needs to show it.

Sorry we took so long to read again your PR. I'm closing it, but please open it again (or another) if you think there really is an issue, and if possible with a Behat test.

Maybe it's because you have changed the default name converter (https://api-platform.com/docs/core/serialization/#name-conversion)?

jannes-io:
@alanpoulain thanks for the feedback. We do not declare any name converter, we do use this:
`
api_platform:
  path_segment_name_generator: api_platform.path_segment_name_generator.dash
`
But no name converters other than that. We've been running my forked branch ever since it was made and are now also using it in production without any issues. Guess we'll just have to continue doing that then.

alanpoulain:
I think I see why. By default it seems that the `serializer.name_converter.metadata_aware` is used.
Could you try to set the name converter to the snake case one (`serializer.name_converter.camel_case_to_snake_case`) in API Platform configuration (https://api-platform.com/docs/core/serialization/#name-conversion) to see if it works.
Maybe we should always use the `CamelCaseToSnakeCaseNameConverter` in the `InnerFieldsNameConverter`.
@meyerbaptiste WDYT?

jannes-io:
@alanpoulain this has been on the backburner because we've been using this branch internally which fixed it for us, but as time goes on, we do want api platform updates and we can't keep on maintaining this branch.

We tried your suggestion by adding the name converter. It does fix the filters. As in, I can use "createdAt" in order for example, and it'll order the results properly.

However, it now requires our queries to be snake_cased, which doesn't make sense compared to non-elasticsearch entities.

For example:
query:
```
query getRegistrations($first: Int = 100, $order: RegistrationFilter_order) {
    registrations(first: $first, order: $order) {    
        edges {
            node {
                id     
                created_at 
                point_of_purchase
                user_info {
                    id
                }
            }
        }
    }
}
```
variables:
```
{
    "order": {
        "createdAt": "desc"
    }
}
```
expected query:
```
query getRegistrations($first: Int = 100, $order: RegistrationFilter_order) {
    registrations(first: $first, order: $order) {    
        edges {
            node {
                id     
                createdAt
                pointOfPurchase
                userInfo {
                    id
                }
            }
        }
    }
}
```

So to conclude, adding the nameconverter fixes filter variables, but breaks queries.

PR #3646, opened 2020-07-20
add php 8 rc to ci

flug:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no 
| New feature?  | yes/no 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixes #... 
| License       | MIT
| Doc PR        | api-platform/docs#...

PR #3745, opened 2020-10-05
Add support for PHP 8

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | replaces #3646
| License       | MIT
| Doc PR        | N/A

Add support for PHP 8.

This PR removes some deprecations (mostly defining a required argument after an optional argument). It also aliases the `Match` class from Doctrine Mongodb since `match` is now a reserved keyword in PHP 8 and causes parse errors.

All the tests pass if they are run individually. The biggest outstanding issue is with Doctrine. Currently, it fails with the error `Fatal error: Declaration of Doctrine\DBAL\Driver\PDOConnection::query() must be compatible with PDO::query(string $query, ?int $fetch_mode = null, mixed ...$fetch_mode_args)`.

This specific issue has been fixed in Doctrine 3.x (currently unreleased), but it's not possible to add it as a dev dependency currently as there are still a lot of other dependencies that don't allow Doctrine 3.x yet.


tests/Bridge/Symfony/Bundle/DataCollector/RequestDataCollectorTest.php
```diff
@@ -181,7 +181,12 @@ public function testWithResourceWithTraceables()
 
         $dataProvider = $dataCollector->getCollectionDataProviders();
         foreach ($dataProvider['responses'] as $class => $response) {
-            $this->assertStringStartsWith('class@anonymous', $class);
+            if (\PHP_VERSION_ID >= 80000) {
+                $this->assertStringStartsWith('ApiPlatform\\Core\\DataProvider\\CollectionDataProviderInterface@anonymous', $class);
```
dunglas, line 185: Maybe can we use the `::class` constant? Or just use `assertStringContainsString()` to avoid the condition totally.

tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
```diff
@@ -160,6 +160,10 @@ public function testAssertArraySubsetDoesNothingForValidScenario(): void
 
     public function testFindIriBy(): void
     {
+        if (\PHP_VERSION_ID >= 80000) {
```
dunglas, line 163: Why?

dunglas (commented):
This looks good! I fear it introduces some small BC breaks but I don't think that they can be prevented.

tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
pierredup: This causes the same error with Doctrine (incompatible method signatures). So, for now, I'm just skipping these tests on PHP 8, otherwise, the test suite fails with a fatal error and we can't see the rest of the test runs and if it passes.

stephanvierkant:
What's the status of this PR? I think the failing test is unrelated, isn't it?

dunglas:
Theoretically, this PR should be rebased against 2.5 as we consider supporting supporting new PHP versions as a bug fixes. However this PR introduces small (unavoidable) BC breaks, so merging it in a patch release doesn't look reasonable. WDYT @api-platform/core-team?

alanpoulain:
I think having this support for 2.6 is enough.

antograssiot:
:+1: for 2.6 and the BC breaks must be added in the release note/changelog (+ maybe a new UPGRADING.md ?)

dunglas:
The problem is that we still have no ETA for 2.6. If only this version supports PHP 8, we'll have to release 2.6 before or just after the release of PHP 8.

alanpoulain:
The release of PHP 8 is the 26th of November. It's a good ETA don't you think?

dunglas:
Many things aren't ready yet. Especially regarding Vulcain support in JS components.

alanpoulain:
Can't it be done later?

dunglas:
I'm not sure yet. Merging in 2.5 would be easier.

pierredup:
@dunglas Which changes are the BC breaks? Maybe there is a way around it, but I'm not sure which changes will break BC currently.

tests/Bridge/Symfony/Bundle/DataPersister/TraceableChainDataPersisterTest.php
```diff
@@ -32,6 +32,10 @@ public function testPersist($persister, $expected)
         $result = $dataPersister->getPersistersResponse();
         $this->assertCount(\count($expected), $result);
         $this->assertEmpty(array_filter($result, function ($key) {
+            if (\PHP_VERSION_ID >= 80000) {
```
dunglas, line 35: Why do we need this?

tests/Bridge/Symfony/Bundle/Test/ApiTestCaseTest.php
dunglas: Ok, can you at least add a `FIXME` or something?

tests/Bridge/Symfony/Bundle/Test/ClientTest.php
```diff
@@ -24,6 +24,10 @@ class ClientTest extends ApiTestCase
 {
     protected function setUp(): void
     {
+        if (\PHP_VERSION_ID >= 80000) {
+            $this->markTestSkipped();
```
dunglas, line 28: same here

dunglas:
@pierredup ok I double checked but we're (again) saved because most of our classes are final 😬!
So could you rebase against 2.5 then?

Thanks!

dotdevru:
doctrine/dbal released new version 2.12.x which supports php8, no bc/code updates needed only update doctrine/dbal to 2.12.x. So is there any ETA for php8? Also u need to update https://github.com/willdurand/Negotiation package which supports php8 only in mater branch (3.0-dev) for now.

dunglas:
@dotdevru we'll merge this PR as soon as possible.

dunglas:
@pierredup do you want me to rebase it against 2.5?

pierredup:
@dunglas I've rebased on 2.5 and removed the skipped tests since Doctrine DBAL now supports PHP 8.

dunglas (approved):

stephanvierkant:
Any estimate when this will be merged (and tagged)?

dunglas:
Thank you very much @pierredup! Great work.

@stephanvierkant we need to finish https://github.com/api-platform/core/pull/3794 before tagging a release.

PR #3752, opened 2020-10-12
Add previous_data to persisters context when available

### LINKED ISSUE #3751: [Discussion] Best practice for comparing changes in a DataPersister?
As the [docs say](https://api-platform.com/docs/core/extending/#extending-api-platform), the recommended way to add _"custom business logic and computations to trigger before or after persistence (ex: mail, call to an external API...)"_ is now to use DataPersisters, rather than the old method of EventSubscribers.

These seem to work well as a replacement in general, although one shortcoming I'm noticing is there doesn't seem to be a way to compare the current version of the resource (about to be persisted) with how it was originally before denormalization.  With an EventListener, I could have just stored the original property value in `KernelEvents::REQUEST`, and compared it in `KernelEvents::VIEW`.

Simple use-case example: when `User.approved` turns from `false` to `true` after a `PUT` request, sending out a "Your account has been approved" email.

There's Doctrine UnitOfWork's `getOriginalEntityData`, which could be used as follows:

```php
public function persist($data, array $context = [])
{
    if ($data instanceof User) {
        $original = $this->em->getUnitOfWork()->getOriginalEntityData($data);
        if (!($original['approved'] ?? null) && $data->approved) {
            // send email
        }
    }
}
```

However this feels a bit questionable to rely on, as it's not officially recommended anywhere, coupled to Doctrine, can't be statically analysed, etc.

Wondering if API Platform could provide a mechanism for this itself?  In an ideal world, something like:

```php
public function persist($data, array $context = [])
{
    $original = $context['original_resource'];
    if ($data instanceof User && $original instanceof User) {
        if (!$original->approved && $data->approved) {
            // send email
        }
    }
}
```

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3751
| License       | MIT
| Doc PR        | 

Expose the previous_data attribute into DataPersisters context.
ref https://github.com/api-platform/core/issues/3751#issuecomment-706753166

### DIFF
--- a/src/Util/AttributesExtractor.php
+++ b/src/Util/AttributesExtractor.php
@@ -53,6 +53,10 @@ final class AttributesExtractor
             }
         }
 
+        if ($previousObject = $attributes['previous_data'] ?? null) {
+            $result['previous_data'] = $previousObject;
+        }
+
         if (false === $hasRequestAttributeKey) {
             return [];
         }
--- a/tests/Util/RequestAttributesExtractorTest.php
+++ b/tests/Util/RequestAttributesExtractorTest.php
@@ -189,4 +189,22 @@ class RequestAttributesExtractorTest extends TestCase
     {
         $this->assertEmpty(RequestAttributesExtractor::extractAttributes(new Request([], [], ['_api_resource_class' => 'Foo'])));
     }
+
+    public function testExtractPreviousDataAttributes()
+    {
+        $object = new \stdClass();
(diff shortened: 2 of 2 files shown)


dunglas (approved):

soyuka (approved):

dunglas:
Thanks @antograssiot!

PR #3755, opened 2020-10-13
Escape search terms before using them in a Regex

### LINKED ISSUE #3754: Mongodb: search terms are not escaped before use in rexexp
**API Platform version(s) affected**: master

**Description**  
Searched strings are not properly escaped before being used in Mongodb regular expressions. 

**How to reproduce**  
I have an entry containing a string with parentheses (let's say it's `foo(bar)`)in Mongodb, and it appears that I have to escape that string in my HTTP query if I want to find that entry:

`http://my.host/my/entities?myField=foo(bar)` yields no results, but
`http://my.host/my/entities?myField=foo\(bar\)` yields all the results expected by performing the first query.

**Possible Solution**  
[SearchFilter::addEqualityMatchStrategy()](https://github.com/api-platform/core/blob/176cdf9cedd187de8ef172d1fbdaf669ccabd431/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php#L165) should escape strings before concatenating them to form a regular expression.

rimas-kudelis:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes: search terms no longer need or support explicit escaping, and consequently, regex special characters no longer behave like they're special.
| Deprecations? | no
| Tickets       | Fixes #3754
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
@@ -166,20 +166,24 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
     {
         $type = $metadata->getTypeOfField($field);
 
+        if (MongoDbType::STRING !== $type) {
+            return MongoDbType::getType($type)->convertToDatabaseValue($value);
+        }
+
+        $quotedValue = preg_quote($value);
+
         switch ($strategy) {
-            case MongoDbType::STRING !== $type:
-                return MongoDbType::getType($type)->convertToDatabaseValue($value);
             case null:
             case self::STRATEGY_EXACT:
-                return $caseSensitive ? $value : new Regex("^$value$", 'i');
+                return $caseSensitive ? $value : new Regex("^$quotedValue$", 'i');
             case self::STRATEGY_PARTIAL:
-                return new Regex($value, $caseSensitive ? '' : 'i');
+                return new Regex($quotedValue, $caseSensitive ? '' : 'i');
             case self::STRATEGY_START:
-                return new Regex("^$value", $caseSensitive ? '' : 'i');
+                return new Regex("^$quotedValue", $caseSensitive ? '' : 'i');
             case self::STRATEGY_END:
-                return new Regex("$value$", $caseSensitive ? '' : 'i');
+                return new Regex("$quotedValue$", $caseSensitive ? '' : 'i');
             case self::STRATEGY_WORD_START:
-                return new Regex("(^$value.*|.*\s$value.*)", $caseSensitive ? '' : 'i');
+                return new Regex("(^$quotedValue.*|.*\s$quotedValue.*)", $caseSensitive ? '' : 'i');
             default:
                 throw new InvalidArgumentException(sprintf('strategy %s does not exist.', $strategy));
         }
--- a/tests/Bridge/Doctrine/Common/Filter/SearchFilterTestTrait.php
+++ b/tests/Bridge/Doctrine/Common/Filter/SearchFilterTestTrait.php
@@ -200,6 +200,15 @@ trait SearchFilterTestTrait
                     'name' => 'exact',
                 ],
             ],
+            'exact (case insensitive, with special characters)' => [
+                [
+                    'id' => null,
+                    'name' => 'iexact',
+                ],
+                [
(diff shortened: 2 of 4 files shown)


src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -166,20 +166,24 @@ private function addEqualityMatchStrategy(string $strategy, string $field, $valu
     {
         $type = $metadata->getTypeOfField($field);
 
+        if (MongoDbType::STRING !== $type) {
+            return MongoDbType::getType($type)->convertToDatabaseValue($value);
+        }
```
alanpoulain, line 171: Why this check? Did you encounter another issue?

alanpoulain:
Thanks!
Could you add a unit test too?

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
rimas-kudelis: I just moved it upwards, away from the `switch` below.

rimas-kudelis:
> Thanks!
> Could you add a unit test too?

Erm... if only I could understand how those tests work, or at least install composer dependencies...
I've added a new element to `SearchFilterTest::provideApplyRestData()`, but I strongly doubt that will be enough, will it? :)

soyuka:
You need the mongodb php extension only if you want to test mongodb stuff. See https://github.com/api-platform/core/blob/master/CONTRIBUTING.md#tests for more informations on how to run tests.

rimas-kudelis:
I've added the test as requested.
Also added `phpspec/prophecy` to `require-dev`, because existing tests just failed without it.

composer.json
```diff
@@ -46,6 +46,7 @@
         "justinrainbow/json-schema": "^5.2.1",
         "phpdocumentor/reflection-docblock": "^3.0 || ^4.0 || ^5.1",
         "phpdocumentor/type-resolver": "^0.3 || ^0.4 || ^1.4",
+        "phpspec/prophecy": "^1.10",
```
soyuka, line 49: should not be needed

rimas-kudelis: Like I said, most of the existing PhpUnit tests were failing without this dependency, at least when I was testing MongoDB. They ran just fine after installing the package.
You can grep your `tests` directory for `use Prophecy` to get the picture.
If you still think I'm wrong, I can remove this requirement.

soyuka: I think that we fixed this by now as we're using Prophecy on other parts of the library.

rimas-kudelis: If you say so. :)

src/Bridge/Doctrine/MongoDbOdm/Filter/SearchFilter.php
```diff
@@ -166,20 +166,24 @@ private function addEqualityMatchStrategy(string $strategy, string $field, $valu
...
+
+        $quotedValue = preg_quote($value);
```
soyuka, line 173: Shouldn't we use the doctrine quote method? https://www.doctrine-project.org/projects/doctrine-dbal/en/2.12/reference/security.html#right-quoting-escaping-values

alanpoulain: Or maybe we should use parameters. We shouldn't have to escape the values, I think there are security issues here.

rimas-kudelis: We're not escaping SQL here. We're escaping regular expression characters. Thus I don't think the quoting link applies, or that parameters can be used in this case.

Ideally, the `Regex` class would have a static method to escape the necessary characters, but it doesn't seem to have a method like that.

Here's a [relevant Stack Overflow question](https://stackoverflow.com/questions/27962072/how-to-escape-special-characters-using-mongo-regex-in-php-format) from 5 years ago.

alanpoulain (approved):

rimas-kudelis:
Are we still waiting for something?

alanpoulain (approved):

alanpoulain:
Thank you @rimas-kudelis!

rimas-kudelis:
Thanks for merging! :+1:

PR #3715, opened 2020-09-10
Add Symfony Uid support

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3714
| License       | MIT
| Doc PR        | api-platform/docs#...

- Added support for denormalizing identifiers with symfony uuid/ulid;
- Added ulid/uuid support to json schema;

PR #3756, opened 2020-10-13
Try to fix master pipeline

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

-
We might want to skip the installation of symfony/uid for legacy tests completely. The rationale behind this is that we can't install this package when the symfony requirement is limited to `^3.4 | ^4.0` only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -765,9 +765,6 @@ jobs:
           composer remove --dev --no-progress --no-update --ansi \
             doctrine/mongodb-odm \
             doctrine/mongodb-odm-bundle \
-      - name: Require Symfony Uid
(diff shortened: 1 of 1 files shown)


dunglas:
But then we don't test the use of this component at all? Maybe should we add a new job just for that?

soyuka:
see also https://github.com/api-platform/core/pull/3715

soyuka (approved):

antograssiot:
Yes we do test it in other pipelines but on on the legacy suite that runs only on SF ^3.4|^4.0

soyuka:
thansk @antograssiot

PR #3757, opened 2020-10-13
Fix openApi when no data provided

Te4g:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Capture d’écran 2020-10-13 à 22 46 14

Found this bug on a brand new Symfony project while trying to check the Open Api documentation while no Entity was created yet with `"api-platform/core": "2.6.x-dev"`

### DIFF
--- a/src/OpenApi/Model/Paths.php
+++ b/src/OpenApi/Model/Paths.php
@@ -29,6 +29,6 @@ final class Paths
 
     public function getPaths(): array
     {
-        return $this->paths;
+        return $this->paths ?? [];
     }
 }


src/OpenApi/Model/Paths.php
```diff
@@ -27,7 +27,7 @@ public function getPath(string $path): ?PathItem
         return $this->paths[$path] ?? null;
     }
 
-    public function getPaths(): array
+    public function getPaths(): ?array
     {
         return $this->paths;
```
dunglas, line 32: Isn't it possible to initialize the property to an empty array instead?

soyuka:
Paths is required ([ref](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields)) but may be empty, please use an empty array as suggested.

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -56,7 +56,7 @@ private function recursiveClean($data): array
             }
 
             // Side effect of using getPaths(): Paths which itself contains the array
-            if ('paths' === $key) {
+            if ('paths' === $key && isset($data['paths']['paths'])) {
                 $value = $data['paths'] = $data['paths']['paths'];
```
soyuka, line 60:
```suggestion
                $value = $data['paths'] = $data['paths']['paths'] ?? [];
```

src/OpenApi/Model/Paths.php
```diff
@@ -27,7 +27,7 @@ public function getPath(string $path): ?PathItem
         return $this->paths[$path] ?? null;
     }
 
-    public function getPaths(): array
+    public function getPaths(): ?array
```
soyuka, line 30: revert this change

Te4g: 👍

src/OpenApi/Serializer/OpenApiNormalizer.php
Te4g: This is not working, we still have the array type error from the `getPaths` method, i just made a new proposal :)

src/OpenApi/Model/Paths.php
Te4g: Sure 👌

dunglas (approved):

soyuka:
Perfect thanks @Te4g

PR #3758, opened 2020-10-14
fix OpenApi operation servers/security

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | na
| License       | MIT
| Doc PR        | na

This fixes:
  ~~- ability to normalize OpenApi v3 using `/docs.json`~~ (will be in another patch #3759)
  - removes `servers` on operations when it's empty to fix the swagger-ui bug :
20201014_10h38m05s_grim
  - Allows to specify `security` as an empty array on an `openapi_context` to remove previously enabled security.

### DIFF
--- a/src/OpenApi/Serializer/OpenApiNormalizer.php
+++ b/src/OpenApi/Serializer/OpenApiNormalizer.php
@@ -24,7 +24,6 @@ use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface
 {
     public const FORMAT = 'json';
-    public const OPEN_API_PRESERVE_EMPTY_OBJECTS = 'open_api_preserve_empty_objects';
     private const EXTENSION_PROPERTIES_KEY = 'extensionProperties';
 
     private $decorated;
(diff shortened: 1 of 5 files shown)


src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -24,8 +24,10 @@
 final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface
 {
     public const FORMAT = 'json';
-    public const OPEN_API_PRESERVE_EMPTY_OBJECTS = 'open_api_preserve_empty_objects';
+    public const UNSET_VALUE = 'unset this value';
```
dunglas, line 27: I'm not sure to understand what this does.

soyuka: By default the `security`node on an Operation should not be added if it's not set. Still, if it's set as an empty value it should stay: 

> A declaration of which security mechanisms can be used for this operation. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. This definition overrides any declared top-level security. To remove a top-level security declaration, an empty array can be used.

dunglas: Couldn't we use the `null` value to distinguish from the empty array instead (and remove the they if it is null)?

dunglas (approved):

PR #3764, opened 2020-10-15
Configurable asset package

### LINKED ISSUE #3760: [Bridge/Symfony/Bundle] configurable asset package
**Description**  
Using the Symfony framework, it would be useful in some cases if we could configure the "asset package" used by the api platform.
In this way, we would have more control over the generated asset urls: https://symfony.com/doc/4.4/reference/configuration/framework.html#packages

**Example**  
There could be a new config option with the currently used `null` default value, so it would be backward-compatible.

```yaml
# app/config/packages/api_platform.yaml
api_platform:
    asset_package: 'custom_package'
```

The current templates should need only some minor modifications i.e.:

```diff
- <link rel="stylesheet" href="{{ asset('bundles/apiplatform/style.css') }}">
+ <link rel="stylesheet" href="{{ asset('bundles/apiplatform/style.css', asset_package) }}">
```

And of course that new parameter has to be set for the templates i.e. in `\ApiPlatform\Core\Bridge\Symfony\Bundle\Action\SwaggerUiAction::getContext`

smatyas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3760
| License       | MIT
| Doc PR        | api-platform/docs#1186

This PR introduces a new config parameter: `asset_package`

It enables the users to specify a custom asset package ([as in Symfony framework](https://symfony.com/doc/4.4/reference/configuration/framework.html#packages)) to be used by the Swagger UI template.
In this way, we have more fine-grained control over the asset url generations.

The new parameter is optional with a `null` default value. It results in the same `asset(...)` twig function calls as it was before.

----

Example usage of the new parameter:

```yaml
api_platform:
    asset_package: 'api_platform'
```

and the corresponding symfony framework setup example:

```yaml
# config/packages/framework.yaml
framework:
    # ...
    assets:
        base_path: '/custom_base_path' # the default
        packages:
            api_platform:
                base_path: '/'
```

### DIFF
--- a/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
+++ b/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
@@ -61,11 +61,12 @@ final class SwaggerUiAction
     private $graphQlPlaygroundEnabled;
     private $swaggerVersions;
     private $swaggerUiAction;
+    private $assetPackage;
 
     /**
      * @param int[] $swaggerVersions
      */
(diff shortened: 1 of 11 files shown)


smatyas:
The cs-fixer test seems to be invalid, the diff does not contain the failing change anymore.

soyuka (approved):

soyuka:
thanks @smatyas !

PR #3759, opened 2020-10-14
Fix OpenApi /docs?spec_version=3 not using OpenApiFactory

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | na
| License       | MIT
| Doc PR        | na

TBD:
  - add groups to definitions on v3
  - fix schema type missing on v2?

PR #3769, opened 2020-10-18
Fix: SF 5.1 deprecation notices

thbil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | ... 
| License       | MIT

Fix SF 5.1 deprecation notices (see https://github.com/api-platform/core/pull/3759/checks?check_run_id=1259046448#step:15:53).


soyuka:
The last deprecation is due to the node at https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_mercure_publisher.xml#L27 where we need to set: 

```
        <service id="api_platform.doctrine.listener.mercure.publish" alias="api_platform.doctrine.orm.listener.mercure.publish">
            <deprecated package="api-platform" version="2.6">Using "%alias_id%" service is deprecated since API Platform 2.6. Use "api_platform.doctrine.orm.listener.mercure.publish" instead.</deprecated>
        </service>
```

but if we do so the xml validation fails on other branches (sf < 5.1).

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/HandleDeprecatedPass.php
```diff
@@ -0,0 +1,26 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler;
+
+use Symfony\Component\Config\Definition\BaseNode;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+final class HandleDeprecatedPass implements CompilerPassInterface
```
soyuka, line 11: DeprecateMercurePublisherPass

+ comment on the class that it's to handle a deprecation on the `setDeprecated` method

features/bootstrap/HttpHeaderContext.php
```diff
@@ -25,24 +24,12 @@ public function __construct(Request $request)
...
-    /**
-     * Sets the default Accept HTTP header to null (workaround to artificially remove it).
+     * Sets the default Accept HTTP header to application/ld+json before each scenario.
      *
      * @BeforeScenario
      */
     public function removeAcceptHeaderBeforeScenario()
     {
-        $this->request->setHttpHeader('Accept', null);
+        $this->request->setHttpHeader('Accept', 'application/ld+json');
```
soyuka, line 33: as you removed almost every case in steps, I suggest we remove this method as well

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -1394,7 +1394,7 @@ private function getBaseContainerBuilderProphecy(array $doctrineIntegrationsToLo
             BooleanFilter::class => 'api_platform.doctrine.orm.boolean_filter',
             NumericFilter::class => 'api_platform.doctrine.orm.numeric_filter',
             ExistsFilter::class => 'api_platform.doctrine.orm.exists_filter',
-            'api_platform.doctrine.listener.mercure.publish' => 'api_platform.doctrine.orm.listener.mercure.publish',
+            //'api_platform.doctrine.listener.mercure.publish' => 'api_platform.doctrine.orm.listener.mercure.publish',
```
soyuka, line 1397: this should still be present

soyuka (commented):
almost there :+1:

features/bootstrap/HttpHeaderContext.php
thbil: Shouldn't we remove the entire class then ? (and remove all references to it in behat.yml.dist)

dunglas:
Could you rebase against 2.5 please @thbil?

soyuka:
Thanks @thbil merged in https://github.com/soyuka/core/pull/3 !

PR #3771, opened 2020-10-19
Check for the existence of a getter for the ID

toby-griffiths:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | tbc (will update if this PR is acceptable)

This is a draft proposal for discussion, before a do the work to update the tests & docs.

I had the requirement to use Value Objects from another package as resources in my API, however because the value object I was using was a wrapper around another class with the ID property, and simply had a getter for that, API didn't like it, as it checks for the existence of the property by that field name, rather than allowing for getters to access it as a virtual ID property.  This PR checks for the existence of a getter method (either `get{Property}()` or `{property}()`) and will use this if available, rather than requiring the actual property exists.

This may have further reaching consequences which I've not come across yet, so I'm open to having to expand on this change to support this, if it's acceptable to make this change to the codebase, so please can you provide feedback on whether you'd like this or not.

If acceptable, I will update the tests and documentation as well.

Still to do, if PR is wanted…

* [ ] Update tests
* [ ] Add PR to add details to docs
* [ ] Update CHANGELOG

### DIFF
--- a/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
@@ -17,6 +17,8 @@ use ApiPlatform\Core\Exception\PropertyNotFoundException;
 use ApiPlatform\Core\Metadata\Extractor\ExtractorInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Metadata\Property\SubresourceMetadata;
+use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
+use Symfony\Component\PropertyInfo\PropertyReadInfoExtractorInterface;
 
 /**
  * Creates properties's metadata using an extractor.
@@ -27,6 +29,7 @@ final class ExtractorPropertyMetadataFactory implements PropertyMetadataFactoryI
 {
     private $extractor;
     private $decorated;
+    private $propertyReadInfoExtractor;
 
     public function __construct(ExtractorInterface $extractor, PropertyMetadataFactoryInterface $decorated = null)
     {
@@ -148,8 +151,9 @@ final class ExtractorPropertyMetadataFactory implements PropertyMetadataFactoryI
         return new SubresourceMetadata($resourceClass, $isCollection, $maxDepth);
     }
 
-
     /**
+     * Determine if the identifier property is accessible.
+     *
      * @param string $resourceClass
      * @param string $property
      *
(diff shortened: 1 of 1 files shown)


GregoireHebert:
If I Understand Correctly, I think you've got a problem with the checks transformation in your new method. Checking directly that the interface exists seems wrong IMO.

This put aside, it's a good idea !
WDYT of using PropertyAccess component and check for Getters, Hassers/Issers, Magic __get() Method and Magic __call() Method?

toby-griffiths:
@GregoireHebert Thanks for the feedback.  i'm not sure what you mean by your comment…
> Checking directly that the interface exists seems wrong IMO.

Are you referring to [these lines](https://github.com/api-platform/core/pull/3771/files#diff-684960aecfe8ee5245e2129720efccb68220863fbf76b2847ed74b723f5e6776R160)…

```
        if (interface_exists($resourceClass)) {
            return false;
        }
```

If so, that's just taken what was there before my PR, and shifted it in to the check method.  This is just to ensure we're not attempting to handle an interface, I think.

As for the property accessor, that seems like a sensible approach.  For my use case I'm using accessors without the `get` prefix, so I initially thought that this wouldn't work, however I see from digging in to the PropertyAccessor component it's possible to define your own rules for read & write accessors, and it even appear to allow property name methods out the box as well, so glad you mentioned it.  You learn something new every day.  Thank you.

I'll see if I can update now with the PropertyAccessor & see when I get up to.

toby-griffiths:
On second thoughts… I don' think the PropertyAccessor will work as we're only given the resource class name, rather than the object, and the Property Accessor appears to only operate on instantiated objects.  That said, we could possibly use an instance of the `\Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor` which has a `isReadable()` method, or, so be more extensible, use whether `\Symfony\Component\PropertyInfo\PropertyReadInfoExtractorInterface::getReadInfo()` returns an instance of a `PropertyReadInfo` to to indicate whether the property is readable, allowing for custom `PropertyReadInfoExtractorInterface` instances to be injected if desired.

What do people think?

soyuka:
I'm really not sure to get the underlying use case. Could you provide a functional example? I think that there's a better solution.

toby-griffiths:
@soyuka My domain models sit outside my API Platform, and so I am using yaml to define the API Platform resources, however without this change I'm not able to.

My use case… I use a View Model to return what's to be displayed by the API, but internally in my Core domain, the view model is simply a wrapper on my read model, and so doesn't actually have an $id property directly.  It's returned from the wrapped read model…

```
<?php

class MyReadModel {
    
    private string $id;

    // …
    
    public function id(): string
    {
        return $this->id;
    }

    // …
}

class MyViewModel {
    private MyReadModel $wrapped;
    
    public function __constrct(MyReadModel $wrapped)
    {
    	$this->wrapped = $wrapped;
    }
    
    public function id(): string
    {
    	return $this->wrapped->id();
    }

    // …
}
```

I appreciate that this is probably a more unusual use case, however I think that there is still a benefit to not requiring the actual $id property to exist, as long as it can be accessed when needed.

soyuka:
Just use a `getId` in your ViewModel ? If the property name matches it's easy to fix on your code.

toby-griffiths:
@soyuka I can't use the getter do that.  The problem is the property does not exist on the `MyViewModel` that's being inspected.

[This original line](https://github.com/api-platform/core/pull/3771/files#diff-684960aecfe8ee5245e2129720efccb68220863fbf76b2847ed74b723f5e6776L54) is what's causing the problem, since it checks for the actual property rather than the ability to access the "virtual" property.  I don't want/shouldn't really have to have to add a property to a domain model purely for the purpose of getting it to work with the framework layer, right?

src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
```diff
@@ -48,10 +51,8 @@ public function create(string $resourceClass, string $property, array $options =
             }
         }
 
-        $isInterface = interface_exists($resourceClass);
-
         if (
-            !property_exists($resourceClass, $property) && !$isInterface ||
+            $this->doesIdentifierExist($resourceClass, $property) ||
```
soyuka, line 55: Can you use `$this->extractor->isReadable` instead?

toby-griffiths: Ah.  Possible.  I'll need to dig back into this, as it's obviously been a while.  I'll get back too you.

Sorry for the delay.

PR #3774, opened 2020-10-20
Handle binary UUID in SearchFilter

odoucet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/api-platform/issues/1139
| License       | MIT

When searching on a binary UUID field, current behaviour leads to search done on Doctrine with the string UUID and not the binary version of it.
This is because the field type must be provided to Doctrine each time we use QueryBuilder. Why ? Because Doctrine sees string on input, and does not know it needs conversion.

This PR reads `$metadata->getTypeOfField()` and force field type.

**WARNING**
Current PR passes the unit tests (so it does not break current behaviour), but does not implement two new needed tests (to make sure it fixes the issue mentioned) : 
- search on a single UUIDBinary
- search with multiple UUIDBinary values

*Test has been done on a local project so I know it works, but better to check this with a test :)*

I need some help adding custom type to SearchFilterTestTrait as `uuidBinary` is not recognized.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
@@ -151,40 +151,28 @@ class SearchFilter extends AbstractContextAwareFilter implements SearchFilterInt
             $associationField = $associationFieldIdentifier;
         }
 
-        /*
-         * If field type is string/float, Doctrine does not call convertToDatabaseValueSQL() because it
-         * does not know it needs conversion.
-         * This would lead to incorrect values for Ramsey\Uuid\Doctrine\UuidBinaryType for example.
-         * The only fix is to provide field type to doctrine ...
-         * It's easy if setParameter() sets only one value BUT impossible if multiple values are provided.
-         * The only way to do this is to rewrite the IN() statement to multiple values
-         * and map a global setParameters()
-         */
         $type = $metadata->getTypeOfField($associationField);
-        $nbValues = \count($values);
 
-        if (1 === $nbValues) {
+        if (1 === \count($values)) {
             $queryBuilder
                 ->andWhere($queryBuilder->expr()->eq($associationAlias.'.'.$associationField, ':'.$valueParameter))
                 ->setParameter($valueParameter, $values[0], $type);
-        } else {
-            // get current parameters, because QueryBuilder->setParameters() erase previous parameters set
-            $parameters = $queryBuilder->getParameters();
-            $inQuery = [];
 
-            // convertToDatabaseValue() can only convert one value at a time ... We can no longer pass an array of values, we should use multiple values
-
-            for ($i = 0; $i < $nbValues; ++$i) {
-                $inQuery[] = ':'.$valueParameter;
-                $parameters->add(new Parameter($valueParameter, $values[$i], $type));
-                $valueParameter = $queryNameGenerator->generateParameterName($associationField);
-            }
-
-            // we cannot use expr->in() here because it considers $inQuery parameters as strings.
-            $queryBuilder
-                ->andWhere($associationAlias.'.'.$associationField.' IN ('.implode(', ', $inQuery).')')
-                ->setParameters($parameters);
+            return;
         }
+
+        $parameters = $queryBuilder->getParameters();
+        $inQuery = [];
+
+        foreach ($values as $val) {
+            $inQuery[] = ':'.$valueParameter;
+            $parameters->add(new Parameter($valueParameter, $val, $type));
+            $valueParameter = $queryNameGenerator->generateParameterName($associationField);
+        }
+
+        $queryBuilder
+            ->andWhere($associationAlias.'.'.$associationField.' IN ('.implode(', ', $inQuery).')')
+            ->setParameters($parameters);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


alanpoulain:
Since it's a bugfix, it should target 2.5.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -184,37 +219,38 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
      *
      * @throws InvalidArgumentException If strategy does not exist
      */
-    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $value, bool $caseSensitive)
+    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $value, bool $caseSensitive, ClassMetadata $metadata)
```
alanpoulain, line 222: This is a BC break. You should use `func_get_arg` and add a deprecated message if the parameter is not given.

odoucet: it's a protected function that has only one call throughout the whole code in the same file. 
Are you thinking about people extending the class ?

alanpoulain: Yes, some users are extending this class since it's not final (yet). See also: https://symfony.com/doc/current/contributing/code/bc.html.

odoucet (commented):
I used a less generic parameter name ... Tell me if it's the right way to do.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -130,9 +133,42 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
+
+            // let's do a very simple query if we don't need the complex one (99% of time)
+            if (\in_array($metadata->getTypeOfField($field), ['string', 'integer'], true)) {
+                $queryBuilder
+                    ->andWhere(sprintf($wrapCase('%s.%s').' IN (:%s)', $alias, $field, $valueParameter))
+                    ->setParameter($valueParameter, $caseSensitive ? $values : array_map('strtolower', $values));
+            } else {
+                $inQuery = [];
+                $args = [];
+                $type = $metadata->getTypeOfField($field);
```
dunglas, line 156: You could save the result of this method in a variable to not call it two times.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -130,9 +133,42 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
+
+                for ($i = 0; $i < \count($values); ++$i) {
+                    $inQuery[] = ':filterproperty'.$i;
+                    $args[] = new Parameter('filterproperty'.$i, $caseSensitive ? $values[$i] : strtolower($values[$i]), $type);
+                }
+
+                // setParameters reset current parameters set, so we need to merge with getParameters()
+                $parameters = new ArrayCollection($args);
```
dunglas, line 164: Why don't you create an `ArrayCollection` directly instead of creating an intermediate array?
Actually, you could even just use the `ArrayCollection` returned by `$queryBuilder->getParameters()`.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -186,6 +222,13 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
      */
     protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $value, bool $caseSensitive)
     {
+        // check if we have metadata
+        if (($metadata = func_get_arg(7)) instanceof ClassMetadata) {
```
dunglas, line 226: you should add this new parameter as a comment in the method signature too

dunglas (changes requested):

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
odoucet: fixed (but I would like to make sure I handled it correctly)

soyuka:
Nice patch, please target 2.5

odoucet:
Updated to use QueryNameGenerator for variable names.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -139,27 +138,26 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
                 $inQuery = [];
-                $nbArgs = \count($values);
-
-                // QueryBuilder->setParameters() reset current parameters set, so we need to merge them all together
-                $parameters = $queryBuilder->getParameters();
-
-                for ($i = 0; $i < $nbArgs; ++$i) {
-                    $inQuery[] = ':filterproperty'.$i;
-                    $parameters->add(new Parameter('filterproperty'.$i, $caseSensitive ? $values[$i] : strtolower($values[$i]), $type));
+                $args = [];
```
alanpoulain, line 153: Why not always using this conditional branch instead of having an if / else?
It will:
- make the code easier to read (and maintain),
- prevent bugs to arise only in one conditional branch.
WDYT @soyuka @dunglas?

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -184,8 +207,15 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
      *
      * @throws InvalidArgumentException If strategy does not exist
      */
-    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $value, bool $caseSensitive)
+    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $value, bool $caseSensitive, ClassMetadata $metadata = null)
```
alanpoulain, line 210: You should do it like this:
```suggestion
    protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $alias, string $field, $value, bool $caseSensitive/*, ClassMetadata $metadata*/)
```

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
odoucet: This will require updating some tests expectations because when multiple values are given, they are set as one unique array parameter, and that will become several parameters. See https://github.com/api-platform/core/pull/3774/checks?check_run_id=1379433208#step:14:91

I'm currently blocked at adding UUID tests, I cant fully understand yet how tests are organized and how can I add custom types.
```
\Doctrine\DBAL\Types\Type::addType('uuid', \Ramsey\Uuid\Doctrine\UuidType::class);
```
is required to manipulate uuid types.

odoucet:
I've pushed an attempt to add test on UUID field, but the current test failed because `SearchFilterTest::testGetDescriptionDefaultFields()` failed on `$filter->getDescription` : it seems the `uuid` type is never loaded, despite adding it in `DoctrineOrmFilterTestCase:setUp()`.
I need help on this one... or I can update my branch with no additional test and it can be merged right away.

soyuka:
if you need more help @odoucet ping me on symfony's slack!

odoucet:
I'll update my PR with latest master branch in a few days.
I need help on how to provide test code for this (I did not understand how to create custom doctrine types in unit tests).

odoucet:
Hello,
I've updated my pull request to be up to date with latest master. 
This MR lacks specific unit tests because I was unable to add a binary doctrine type to tests. @soyuka if you can help, it would be great :) 
This MR can also be merged as-is, I know it works because I use binary type in a private project.

alanpoulain:
Hello @odoucet,
Adding PHPUnit tests (they are integration tests, not really unit ones) is not possible since the conversion is done at the execution.
I've added Behat tests instead.

alanpoulain:
Thank you @odoucet.

PR #3776, opened 2020-10-21
Slash baseUrl only from IRI beginning

### LINKED ISSUE #3775: Invalid IRI exception raised if your IRI contains the base url in the path

**Description**  
If your Resource IRI contains the base url in their path, e.g. /myapp/api/myapp_crm_bundle/my_resource/resource-uuid-string, any operation will fail as the router will slash any /myapp occurrence in the IRI, so it will try to match /api_crm_bundle/my_resource/resource-uuid-string, failing with a Invalid IRI exception.

The problem is this line in Router->match:

`$pathInfo = str_replace($baseContext->getBaseUrl(), '', $pathInfo);`

**How to reproduce**  
Create a resource with a custom path operation with the base Url in the path.

**Possible Solution**  
See pull request.

maxtorete:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3775
| License       | MIT
| Doc PR        | 

I made a comparison before slashing the first baseUrl from IRI as security check, maybe it can be removed. I used substr as it is faster than preg_replace.

### DIFF
--- a/src/Bridge/Symfony/Routing/Router.php
+++ b/src/Bridge/Symfony/Routing/Router.php
@@ -71,7 +71,10 @@ final class Router implements RouterInterface, UrlGeneratorInterface
     public function match($pathInfo)
     {
         $baseContext = $this->router->getContext();
-        $pathInfo = str_replace($baseContext->getBaseUrl(), '', $pathInfo);
+        $baseUrl = $baseContext->getBaseUrl();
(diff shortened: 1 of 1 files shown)


GregoireHebert (approved):
Seems good to me!

dunglas (approved):

dunglas:
Good catch, thanks!

PR #3784, opened 2020-10-24
[DEFAULT ORDER] #1246: Support default order for a specific custom operation

ahmed-bhs:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       |  #1321
| License       | MIT
| Doc PR        | api-platform/docs#1195

Currently, the `default-order` impact all the defined collection operations, so, it is not possible to have two different custom operations with different `default-order` configurations.

The idea is to support both, the definition of an order in a global way and also in a more specific way.

For example:
```yaml
resources:
    attributes:
        order: ["foo", "ASC"]
    App\Entity\Book:
        collectionOperations:
            get_desc:
               method: GET
               path: /books/desc
               order: ["name", "DESC"]
               normalization_context:
                    groups: ['group1']
            get_asc:
               method: GET
               path: /books/asc
               order: ["name", "ASC"]
               normalization_context:
                    groups: ['group2']
            get_another:
               method: GET
               path: /books/another
               normalization_context:
                    groups: ['group3']
```
* `order: ["name", "DESC"]` will be applied on *get_desc* operation.
* `order: ["name", "ASC"]` will be applied on *get_asc* operation.
* `order: order: ["foo", "ASC"]` will be applied on *get_another* operation.

This will allow to have a parent generic configuration and others more specific child configuration.

### DIFF
--- a/features/main/default_order.feature
+++ b/features/main/default_order.feature
@@ -117,3 +117,115 @@ Feature: Default order
       }
     }
     """
+
+  Scenario: Override custom order asc
+    When I send a "GET" request to "/custom_collection_asc_foos?itemsPerPage=10"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/Foo",
+      "@id": "/foos",
+      "@type": "hydra:Collection",
+      "hydra:member": [
+        {
+          "@id": "/foos/5",
+          "@type": "Foo",
+          "id": 5,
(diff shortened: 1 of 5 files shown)


ahmed-bhs:
@dunglas 🙋 Could please take a quick look and let me know what do you think about it!

soyuka (approved):
:+1: on the patch I just need to check with @dunglas if we can consider this a bug fix

dunglas:
To le it's good as a new feature! Thanks for working on this @ahmed-bhs.

soyuka:
Thanks @ahmed-bhs

PR #3787, opened 2020-10-25
Search Filter: Allow aliasing of properties, allow explicit strategy selection

### LINKED ISSUE #2310: Filter custom name
According to the documentation, I can set a custom name for an order filter, but not a classic search filter.

It would be great to be able to do this.

Example:

```php
/**
 * @var Os|null
 *
 * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Os", inversedBy="servers")
 * @ORM\JoinColumn(referencedColumnName="id", nullable=true)
 * @Groups({"get", "post:admin", "put:admin"})
 * @ApiFilter(SearchFilter::class, strategy="exact",
 *     properties={
 *         "os.name"
 *     })
 */
private $os;
```

This will give me the `os.name` filter. It is working, but not very user friendly nor useful for the API end user.

I would like to name this filter simply `os`.

It is not currently possible if I am right.

### LINKED ISSUE #2351: Search filter: configure at query level rather than definition level
Hey there,

Sharing a thought again. The current implementation of the SearchFilter forces to define one but only one specific search strategy for a property among exact, partial, start, end, word_start (and their insensitive variant).

Problem is we cannot have 2 or more of these strategies available for query. Thus it would be, imho, faily easy to achieve that with the same behaviour as the DateFilter.

The implementation could still define a strategy which could be the default strategy to apply.

Example with a SearchFilter configured on a prop named "city" with strategy "exact":
?city=amsterdam
?city[exact]=amsterdam //same effect as previous line
?city[start]=new
?city[end]=york
?city[partial]=ris
etc.

### LINKED ISSUE #3606: Configurable filter query string parameters
Filters use a single name as both the query string parameter and the underlying path to the filtered property. This ties the query string parameter to the underlying model: if the model changes, the parameter changes as well and its consumers must be updated. It also exposes internal details that the consumers might not have to be aware of.

I propose to allow setting different names for the query parameters by supporting passing an array of settings to filter properties configuration, e.g.:

**before**
```yaml
services:
  foo.search_filter:
    parent: api_platform.doctrine.orm.search_filter
    arguments:
      -
        foo.bar.baz: exact
```

**after**
```yaml
services:
  foo.search_filter:
    parent: api_platform.doctrine.orm.search_filter
    arguments:
      -
        baz:
          property: foo.bar.baz
          strategy: exact
```

luca-nardelli:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no (should we add them?)
| Tickets       | fixes #2310, fixes #3606, fixes #2351
| License       | MIT
| Doc PR        | TODO

**Note: This PR is still WIP, I've submitted it to discuss it while I'm working on it**

**TODO:**
- [ ] Documentation
- [ ] Extend changes to MongoDB SearchFilter

This PR adds 2 new features to the standard SearchFilter, while keeping backwards compatibilty.

## Feature 1: Property aliasing
The name used in the query param is now decoupled from the actual resource property name, this allows to define aliases for nested properties as well. I've used the syntax described here (https://github.com/api-platform/core/issues/3606)

Example:
```php

@ApiFilter(SearchFilter::class, properties: {
  "aliasedName" : {"property": "name", "defaultStrategy": "exact"},
  "aliasedNestedProperty" : {"property": "relatedDummy.name", "defaultStrategy": "exact"},
})
class Dummy() {
    /**
     * @ORM\Column
     */
    private $name;
    /**
     * @var RelatedDummy A related dummy
     * @ORM\ManyToOne(targetEntity="RelatedDummy")
     */
    public $relatedDummy;
}
```
If no `property` is defined, the filter assumes that the key in the properties array corresponds to the actual property name.
If no strategy is defined (`defaultStrategy` key), `exact` is assumed.
I used `defaultStrategy` instead of `strategy` because of feature2 (see below).

With this, users can do the following:

`GET /api/dummies?aliasedName=test`
This performs exact match with the property `name`

`GET /api/dummies?aliasedNestedProperty=test`
This performs exact match with the property `relatedDummy.name`

Old SearchFilter definitions in the form ```"<property>": "<strategy>"```  correspond to the following new definition 
```<property>: {"property": "<property>", "defaultStrategy": "<strategy>"}```

## Feature 2: Explicit strategy request
Using the same entity definition as before, developers can now do the following:

`GET /api/dummies?aliasedName=test`
This performs exact match with the property `name`

`GET /api/dummies?aliasedName[partial]=test`
This performs partial match with the property `name`

And so on for `start`, `end`, `word_start` and their case insensitive versions. I've used the syntax proposed in #2351 which seemed good to me.
If no strategy is provided explicitly, the `defaultStrategy` is used.

## Testing
I've added ~10 tests to test for both aliasing and explicit strategy selection. I've also kept all previous tests on the SearchFilter.

## Points of discussion

* Does this seem like a good improvement for the SearchFilter functionality? By looking at #2351 it seemed to me that some work could be done on the filter. If you think 
* Should we deprecate the old SearchFilter properties syntax?
* Should this be expanded to the ODM version of the Search Filter?
* I'm seeing some issues related to automated tests. More precisely, several tests are failing at `Require Symfony UID` with `Cannot update only a partial set of packages without a lock file present.` which seems unrelated to my changes, do you have any idea into how I could fix this? Thanks.

Let me know your thoughts!

### DIFF
--- a/src/Bridge/Doctrine/Common/Filter/SearchFilterInterface.php
+++ b/src/Bridge/Doctrine/Common/Filter/SearchFilterInterface.php
@@ -45,4 +45,33 @@ interface SearchFilterInterface
      * @var string Finds fields that are starting with the word
      */
     public const STRATEGY_WORD_START = 'word_start';
+
+    /**
+     * All avaliable strategies.
+     */
+    public const STRATEGIES = [
+        self::STRATEGY_EXACT,
+        self::STRATEGY_PARTIAL,
+        self::STRATEGY_START,
+        self::STRATEGY_END,
+        self::STRATEGY_WORD_START,
+        'i'.self::STRATEGY_EXACT,
+        'i'.self::STRATEGY_PARTIAL,
+        'i'.self::STRATEGY_START,
+        'i'.self::STRATEGY_END,
+        'i'.self::STRATEGY_WORD_START,
+    ];
+
+    public const STRATEGIES_MAP = [
+        self::STRATEGY_EXACT => true,
+        self::STRATEGY_PARTIAL => true,
+        self::STRATEGY_START => true,
+        self::STRATEGY_END => true,
+        self::STRATEGY_WORD_START => true,
+        'i'.self::STRATEGY_EXACT => true,
+        'i'.self::STRATEGY_PARTIAL => true,
+        'i'.self::STRATEGY_START => true,
+        'i'.self::STRATEGY_END => true,
+        'i'.self::STRATEGY_WORD_START => true,
+    ];
 }
--- a/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
@@ -32,9 +32,21 @@ use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 /**
  * Filter the collection by given properties.
  *
+ * Filter supports multiple strategies per property and also property aliasing. It can be configured as follows:
+ * properties = {
+ *   "<queryParamName>" : {
+ *     "property": "<entity property name, defaults to queryParamName>,
+ *     "defaultStrategy": "exact|partial|ipartial"
+ *   }
+ * }
+ *
+ * Filter can then be invoked as follows
+ * ?queryParamName[<strategy>]=<value>
+ * ?queryParamName[<strategy>][]=<value>
+ *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
-class SearchFilter extends AbstractContextAwareFilter implements SearchFilterInterface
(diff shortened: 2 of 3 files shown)


soyuka:
In my opinion this should be in your code, not in the vendors. You can easily declare a custom filter that matches your needs and adding this to the core means more complexity and more time to allow to related bugs etc. I'm not in favor of adding this to the core repository.

luca-nardelli:
Thanks for your answer! No problem in implementing this in my code, I saw there were some past discussions and before implementing it outside the api-platform/core I wanted to know if it could have been a useful addition. I'll close the PR then.

Votanar:
I have the same problem. More. I need to add a description to a search parameter and this solution with the next parameter (description) would be the solution for me.

But ok. I will write own custom filter.

PR #3789, opened 2020-10-25
Use serialized_name in Swagger UI Links section

### LINKED ISSUE #3788: Identifiers with custom serialized_name show incorrect name in SwaggerUI Links section

**Description**  
When you set a serialized_name to an identifier property, SwaggerUI shown the serialized_name in the schema, but in the SwaggerUI operations Links section the original property name is shown ignoring the custom [serialized_name](https://symfony.com/blog/new-in-symfony-4-2-simpler-custom-serialized-names).

**How to reproduce**  
Set a serialized_name to a identifier property in a Resource, then go to SwaggerUI and check Links section.

**Possible Solution**  
See pull request.

maxtorete:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3788 
| License       | MIT
| Doc PR        |

Show [custom serialized_name](https://symfony.com/blog/new-in-symfony-4-2-simpler-custom-serialized-names) if set from Resource attributes in Swagger UI Links section instead original Resource property.

### DIFF
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -793,6 +793,7 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
                 continue;
             }
 
+            $propertyName = $propertyMetadata->getSerializedName() ?? $propertyName;
             $linkObject['parameters'][$propertyName] = sprintf('$response.body#/%s', $propertyName);
             $identifiers[] = $propertyName;
         }
@@ -801,7 +802,7 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
             return [];
         }
         $linkObject['operationId'] = $operationId;
-        $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `%1$s` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
+        $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `id` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
 
         return $linkObject;
     }
--- a/src/Swagger/Serializer/DocumentationNormalizer.php
+++ b/src/Swagger/Serializer/DocumentationNormalizer.php
@@ -802,7 +802,7 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
             return [];
         }
         $linkObject['operationId'] = $operationId;
-        $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `id` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
+        $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `%1$s` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
 
         return $linkObject;
     }
--- a/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
@@ -61,6 +61,11 @@ final class SerializerPropertyMetadataFactory implements PropertyMetadataFactory
             return $propertyMetadata;
         }
 
+        $serializerMetadataAttributes = $this->serializerClassMetadataFactory->getMetadataFor($resourceClass)->getAttributesMetadata();
+        if (!empty($serializerMetadataAttributes[$property]) && null !== ($serializedName = $serializerMetadataAttributes[$property]->getSerializedName())) {
+            $propertyMetadata = $propertyMetadata->withSerializedName($serializedName);
+        }
+
         $propertyMetadata = $this->transformReadWrite($propertyMetadata, $resourceClass, $property, $normalizationGroups, $denormalizationGroups);
 
         return $this->transformLinkStatus($propertyMetadata, $normalizationGroups, $denormalizationGroups);
--- a/src/Metadata/Property/PropertyMetadata.php
+++ b/src/Metadata/Property/PropertyMetadata.php
@@ -35,8 +35,9 @@ final class PropertyMetadata
     private $attributes;
     private $subresource;
     private $initializable;
+    private $serializedName;
 
-    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null, bool $initializable = null)
+    public function __construct(Type $type = null, string $description = null, bool $readable = null, bool $writable = null, bool $readableLink = null, bool $writableLink = null, bool $required = null, bool $identifier = null, string $iri = null, $childInherited = null, array $attributes = null, SubresourceMetadata $subresource = null, bool $initializable = null, string $serializedName = null)
     {
         $this->type = $type;
         $this->description = $description;
@@ -51,6 +52,7 @@ final class PropertyMetadata
         $this->attributes = $attributes;
         $this->subresource = $subresource;
         $this->initializable = $initializable;
+        $this->serializedName = $serializedName;
     }
 
     /**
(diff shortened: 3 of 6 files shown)


src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -785,12 +785,14 @@ private function flattenMimeTypes(array $responseFormats): array
...
     {
         $linkObject = $identifiers = [];
+        $resourceMetadataAttributes = $this->resourceMetadataFactory->create($resourceClass)->getAttributes();
         foreach ($this->propertyNameCollectionFactory->create($resourceClass) as $propertyName) {
             $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $propertyName);
             if (!$propertyMetadata->isIdentifier()) {
                 continue;
             }
 
+            $propertyName = $resourceMetadataAttributes[$propertyName]['serialized_name'] ?? $propertyName;
```
dunglas, line 795: I don't understand why this attribute would exist. ResourceMetadata attributes have nothing to do with Serializer's metadata.

I don't think that it's the good way to fix this. A better way would be to copy the Serializer's `serialized_name` into account in the property metadata and to use it everywhere: https://github.com/api-platform/core/blob/master/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php

dunglas (changes requested):

maxtorete:
Thanks for your help @dunglas, could you check again if this is the right way please?

dunglas:
This looks good to me! Can you also add some tests please?

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -799,7 +800,7 @@ private function getLinkObject(string $resourceClass, string $operationId, strin
             return [];
         }
         $linkObject['operationId'] = $operationId;
-        $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `%1$s` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
+        $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `id` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
```
maxtorete, line 803: I hardcoded the `id` parameter name, as it is hardcoded in OperationPathResolver.php in line 52:

```
        if (OperationType::ITEM === $operationType) {
            $path .= '/{id}';
        }
```

So even if you display the right property name, the parameter in the path is always shown as {id}.

I think a new issue should be created to create the url identifier parameter name dynamically. It is a complex change as it could affect to all the Path Resolvers implemented.

maxtorete:
> This looks good to me! Can you also add some tests please?

@dunglas I added a test, but I found a problem in OperationPathResolver.php [explained in this comment ](https://github.com/api-platform/core/pull/3789#discussion_r512137755)

So I think we can user this pull request to fix the custom serializer name error, and then create another issue to fix the hardcoded {id} parameter in the paths. What do you think?

src/Swagger/Serializer/DocumentationNormalizer.php
dunglas: `{id}` is a URI template variable, so its name doesn't matter, does it?

maxtorete: Hi @dunglas ,

If you don't replace the second `%1$s` placeholder with the harcoded `id` string, and your identifier property is not called id, Swagger UI links section will display something like this if you don't set a custom path or route that operation: 

`The **code** value returned in the response can be used as the **code** parameter in GET /api/issues/{**id**}.`

Both placeholders will be replaced with the identifier property name. So the user is looking for a `code` parameter in the URL, but there isn't any, as it is always display as `/resource-path/{id}` in the URL.

But if you change the second placeholder to a fixed `id` string, anyone setting a custom path (for example `/api/issues/{code}`) or route name in the operation would have to use id as their Resource identifier parameter, or will display something like this:

`The **code** value returned in the response can be used as the **id** parameter in GET /api/issues/{**code**}.`

So if you call your property identifier id or you don't use a custom path or route name in the operation, the two options are fine. For most users, I think the first option will show the right combination, as it is aligned with the default API Plaform behaviour. I think there are more users changing their identifier property name than setting custom paths or routes.

You can see the current behaviour in this screenshot:
Screenshot_2020-10-29 API Platform

maxtorete:
@dunglas I took the URI parameter name in Swagger UI Links section from the URI itself, so it should match the identifier property name in almost all cases. I took the last parameter in the URI, so it will work on subresources too.

I also added another assertion to test serialized name loading from Serializer attributes.

dunglas (approved):

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -791,15 +791,19 @@ private function getLinkObject(string $resourceClass, string $operationId, strin
...
+            $propertyName = $propertyMetadata->getSerializedName() ?? $propertyName;
             $linkObject['parameters'][$propertyName] = sprintf('$response.body#/%s', $propertyName);
             $identifiers[] = $propertyName;
         }
 
         if (!$linkObject) {
             return [];
         }
+        preg_match_all('/{(.*)}/', $path, $uriParams);
+        $identifierUriParam = !empty($uriParams[1]) ? end($uriParams[1]) : 'id';
```
soyuka, line 803: I need to take a look at this as I think it's related to my work on identifiers. We'll definitely merge this at some point!

maxtorete: Hi @soyuka! Don't hesitate to ask me if you need some clarification or if you think any change is needed to avoid conflicts with your work on identifiers.

soyuka: Isn't there another way to get identifiers here ? I'm not fond of the regex to fetch params. Also, an url param is not always an identifier. 
I like the idea but maybe it'll be easier to add on master as a new feature once I can work on my backward compat on #3664

maxtorete: Hi @soyuka. 

You are right. This was a workaround so it should be fine on most cases (at least in all the cases the path is generated by API Platform, even if the identifier property has a custom serialized name), but you are totally right it is not a permanent fix and would fail in some edge cases with custom paths on non restful apis.

As we got a path string as parameter in getLinkObject function, I don't think we can find a real solution to this problem in 2.5, as any identifier url parameter should be done on a string guessing.

I think the best solution in this scenario, and until #3664 is merged and this issue can be revisited, is to omit the parameter name in the Links strings, changing it to ```The `%1$s` value returned in the response can be used as the Resource identifier parameter in `GET %3$s`.```

With that solution, we can merge this bugfix in 2.5 and search for a way to get the identifier parameter name in paths in master as a new feature.

soyuka: You can extract the identifiers by looping on `PropertyCollection` and finding using `isIdentifier` or using the IdentifiersExtractor class.

maxtorete: Getting the identifiers is only half the problem. In you use a custom path or set a route_name in the operation, the identifier parameter name in the URL could be different than the identifier property name on the Resource.

In getLinkObject we actually have an array of the identifiers, so I though to check if the $identifierUriParam extracted from the URI was named as the $identifiers[0] used in Links text. But it did't work, as the parameter don't have to be called as the identifier.

I'm not sure if we could guarantee that any parameter in the URL match to the resource identifier, as in Symfony routes don't exist any "real" mapping between parameters and entities/Resource properties, they are only a placeholder as @dunglas said.

Without any breaking change, I think the only way to not to fail in any case (even with a custom serialized name for the resource identifier or a custom path/route with an  URL parameter called different than the identifier property) is to omit the URL parameter name in the Links text.

soyuka: So let's just use `id` hard-coded. I'll fix this later properly when we'll have a way to specify identifiers.

I'm really not fond of the preg_match and if someone needs to change it there's always the possibility of overriding using a custom swagger normalizer. It'll introduce less complexity. Note that the code needs to be adapted to master where OpenAPI is not handled the same way but I'd like to use this serialized_name.

maxtorete: @soyuka so should I open another pull request adapting this to master? Or should we wait until you fix the identifiers problem?

In this pull request I will revert to the hardcoded `id` and will squash the commits in only one commit rebased to current 2.5.

soyuka: >  should I open another pull request adapting this to master? 

Sure! 

> In this pull request I will revert to the hardcoded id and will squash the commits in only one commit rebased to current 2.5.

You don't really need to squash we can do this when merging :), as you want

src/Swagger/Serializer/DocumentationNormalizer.php
soyuka:
```suggestion
         $linkObject['description'] = 1 === \count($identifiers) ? sprintf('The `%1$s` value returned in the response can be used as the `%1$s` parameter in `GET %2$s`.', $identifiers[0], $path) : sprintf('The values returned in the response can be used in `GET %s`.', $path);
```

PR #3790, opened 2020-10-25
Allow method/property to return false

### LINKED ISSUE #3681: Mercure Publish fails if expression returns bool false
**API Platform version(s) affected**: 2.5

**Description**  
Mercure Publish fails if an expression is used which could return false:
`The value of the "mercure" attribute of the "MyClass" resource class must be a boolean, an array of options or an expression 
returning this array, "boolean" given.'`

**How to reproduce**  
```
/**
  * @ApiResource(
  *    mercure="object.mercure()"
  * )
  */
class MyClass
{
[...]
    private function mercure(){
        return false;
    }
}
```

**Possible Solution**  
Maybe just move the if-clause in https://github.com/api-platform/core/blob/38d69e6fdeaa6d6d32d43b7467454acb17d11315/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php#L157 after the expression-evaluation in https://github.com/api-platform/core/blob/38d69e6fdeaa6d6d32d43b7467454acb17d11315/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php#L161

stephanvierkant:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3681
| License       | MIT

When using a method or property (like [this example](https://api-platform.com/docs/core/mercure/#dispatching-private-updates-authorized-mode) from the documentation) you can't return `false` right now.

Currently, this example:
```
<?php
// api/src/Entity/Book.php

namespace App\Entity;

/**
 * @ApiResource(mercure="object.mercureOptions()")
 */
class Book
{
    /**
     * @ORM\Column(type="string")
     */
    private $foo;

    public function mercureOptions()
    {
        if ($this->foo !== 'bar') {
            return false;
        }

        return ['private' => false];
    }

```

will give an error message:
> The value of the "mercure" attribute of the "Book" resource class must be a boolean, an array of options or an expression returning this array, "boolean" given.'

This 'must be boolean (..), "boolean" given' is very confusing.

With this PR, this example will work.

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -136,9 +136,6 @@ final class PublishMercureUpdatesListener
         }
 
         $options = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('mercure', false);
-        if (false === $options) {
-            return;
-        }
 
         if (\is_string($options)) {
             if (null === $this->expressionLanguage) {
@@ -146,6 +143,10 @@ final class PublishMercureUpdatesListener
             }
 
             $options = $this->expressionLanguage->evaluate($options, ['object' => $entity]);
+        }
+
+        if (false === $options) {
+            return;
         }
 
         if (true === $options) {


dunglas (approved):

dunglas:
Thanks @stephanvierkant

stephanvierkant:
@dunglas Any idea when a new (2.5) release would be tagged/published?

PR #3791, opened 2020-10-26
Fix tests

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This fixes the current failing tests

### DIFF
--- a/tests/Fixtures/app/bootstrap.php
+++ b/tests/Fixtures/app/bootstrap.php
@@ -22,6 +22,11 @@ if (!$phpunitInstalls) {
     die('PHPUnit is not installed. Please run ./vendor/bin/simple-phpunit to install it');
 }
 
+// Use the most recent version available
+$phpunitLoader = require end($phpunitInstalls).'/vendor/autoload.php';
+$phpunitLoader->unregister();
+$phpunitLoader->register();
+
 $loader = require __DIR__.'/../../../vendor/autoload.php';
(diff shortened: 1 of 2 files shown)


tests/Fixtures/app/bootstrap.php
```diff
@@ -23,7 +23,9 @@
 }
 
 // Use the most recent version available
-require end($phpunitInstalls).'/vendor/autoload.php';
+$phpunitLoader = require end($phpunitInstalls).'/vendor/autoload.php';
+$phpunitLoader->unregister();
+$phpunitLoader->register();
```
pierredup, line 28: This seems a bit hacky, but it's required in order to not prepend the PHPUnit autoloader, as it is registered before the normal autoloader. So this causes issues specifically with the `Composer\InstalledVersions` class, which is loaded from the `vendor/.phpunit/8.3/vendor` directory instead of the one in the standard `vendor` directory.
So here we need to unregister the autoloader and re-register it so that is it appended to the queue

dunglas (approved):
Thank you very much! I just rebased your PR, let's see if the tests are green, then merge.

dunglas:
Thanks @pierredup! It's very much appreciated.

PR #3797, opened 2020-10-27
Move configuration to php

Zowac:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| License       | MIT


dunglas:
This should target master as it's a new feature.

src/Bridge/Symfony/Bundle/Resources/config/data_provider.php
```diff
@@ -0,0 +1,40 @@
...
+use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
+use ApiPlatform\Core\DataProvider\ItemDataProviderInterface;
+use ApiPlatform\Core\DataProvider\Pagination;
+use ApiPlatform\Core\DataProvider\SubresourceDataProviderInterface;
+use Symfony\Component\DependencyInjection\ServiceLocator;
+
+return static function (ContainerConfigurator $container) {
+    $container->services()
+        ->set('api_platform.serializer_locator', ServiceLocator::class)
+            ->args([['serializer' => service('api_platform.serializer')]])
```
chalasr, line 26: You need to import that `service` function (see broken CI).
Symfony core does it by using `namespace Symfony\Component\DependencyInjection\Loader\Configurator` on top of each config file, but I'd rather add a `use function` statement  if it works without namespace.

Zowac: Change done, thank you :)

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -89,7 +89,7 @@ public function prepend(ContainerBuilder $container): void
      */
     public function load(array $configs, ContainerBuilder $container): void
     {
-        $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+        $phpFileLoader = new PhpFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
```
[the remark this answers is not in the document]
Zowac: revert done, thx

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.php
```diff
@@ -0,0 +1,132 @@
...
+        ->alias(OrderFilter::class, 'api_platform.doctrine.orm.order_filter')
+
+        ->set('api_platform.doctrine.orm.range_filter', RangeFilter::class)
+            ->abstract()
+            ->args([ref('doctrine'), null, ref('logger')->ignoreOnInvalid(), '$nameConverter' => ref('api_platform.name_converter')->ignoreOnInvalid()])
+        ->alias(RangeFilter::class, 'api_platform.doctrine.orm.range_filter')
+
+        ->set('api_platform.doctrine.orm.date_filter', DateFilter::class)
+            ->abstract()
+            ->args([ref('doctrine'), null, ref('logger')->ignoreOnInvalid(), '$nameConverter' => ref('api_platform.name_converter')->ignoreOnInvalid()])
```
chalasr, line 85: The `ref()` function has been renamed to `service()` in 5.1. We need to declare a `service()` proxy function that delegates to the right function internally depending on existence (and remove that proxy when support for symfony/dependency-injection 4.4 gets removed in the future).

src/Bridge/Symfony/Bundle/Resources/config/nelmio_api_doc.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
soyuka, line 1: you can ignore these in phpstan see what we did at https://github.com/api-platform/core/commit/bfca0eb9f064eff70ea570ba18c0326e6dd7291f#diff-6f19df6a6307a48db0940e6897591fb08776d2db8a6134737aa708defdb5c92dR19

Zowac: Done, thx :)

src/Bridge/Symfony/Bundle/Resources/function.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+if (!\function_exists(__NAMESPACE__.'\service')) {
```
chalasr, line 16:
```suggestion
// Forward compatibility layer for the "service()" method.
// To be removed along with support for symfony/dependency-injection <5.1.
if (!\function_exists(__NAMESPACE__.'\service')) {
```

chalasr (approved):
LGTM 👍

src/Bridge/Symfony/Bundle/Resources/function.php
Zowac: Comments added, thx :)

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.php
Zowac: Done, thx for you help

src/Bridge/Symfony/Bundle/Resources/function.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+// Forward compatibility layer for the "service()" method.
+// To be removed along with support for symfony/dependency-injection <5.1.
+
+if (!\function_exists(__NAMESPACE__.'\service')) {
+    eval('
```
dunglas, line 20: No way 😁

Either we bump the minimal version of dependency-injection to 5.1, or we keep XML for now and merge this PR in 3.0, but it's too ugly for a very limited benefit (if any) IMHO.

dunglas (changes requested):

src/Bridge/Symfony/Bundle/Resources/function.php
chalasr: Fair enough :) Let's bump Symfony to 5.1+ and merge this in 3.x then

PR #3806, opened 2020-10-30
[JSON Schema] Use validation groups for property metadata in JSON Schema factory

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This PR is part of a set of PR whose purpose is to make the assertions `assertMatchesResourceItemJsonSchema` and `assertMatchesResourceCollectionJsonSchema` work correctly.
The PR are:
- https://github.com/api-platform/core/pull/3803
- https://github.com/api-platform/core/pull/3804
- https://github.com/api-platform/core/pull/3806 (this one)
- https://github.com/api-platform/core/pull/3807

It uses the validation groups when creating the property metadata. The aim is to use correctly the `ValidatorPropertyMetadataFactory`, mainly to receive correctly the `required` attribute:
https://github.com/api-platform/core/blob/8e58a986970dc3abe1fa994f6849631ad77e09a1/src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php#L97-L99

PR #3804, opened 2020-10-30
[JSON Schema] Add `additionalProperties` (false) to the generated JSON schema

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This PR is part of a set of PR whose purpose is to make the assertions `assertMatchesResourceItemJsonSchema` and `assertMatchesResourceCollectionJsonSchema` work correctly.
The PR are:
- https://github.com/api-platform/core/pull/3803
- https://github.com/api-platform/core/pull/3804 (this one)
- https://github.com/api-platform/core/pull/3806
- https://github.com/api-platform/core/pull/3807

It add `additionalProperties` (to false) to the definitions in order to not validate a schema if the validated object has additional properties (it could be another resource in this case).

PR #3803, opened 2020-10-30
[JSON Schema] Add base Hydra properties to collection items definition

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This PR is part of a set of PR whose purpose is to make the assertions `assertMatchesResourceItemJsonSchema` and `assertMatchesResourceCollectionJsonSchema` work correctly.
The PR are:
- https://github.com/api-platform/core/pull/3803 (this one)
- https://github.com/api-platform/core/pull/3804
- https://github.com/api-platform/core/pull/3806
- https://github.com/api-platform/core/pull/3807

It adds the base Hydra properties (`@id`, `@type`) to collection items definition in the JSON Schema.

Before:
```json
{
   "$schema":"http:\/\/json-schema.org\/draft-07\/schema#",
   "type":"object",
   "definitions":{
      "Application:jsonld":{
         "type":"object",
         "description":"",
         "properties":{
            "id":{
               "readOnly":true,
               "type":"integer"
            },
            "name":{
               "type":"string"
            }
         }
      }
   },
   "properties":{
      "hydra:member":{
         "type":"array",
         "items":{
            "$ref":"#\/definitions\/Application:jsonld"
         }
      },
      "hydra:totalItems":{
         "type":"integer",
         "minimum":0
      },
      "hydra:view":{
         "type":"object",
         "properties":{
            "@id":{
               "type":"string",
               "format":"iri-reference"
            },
            "@type":{
               "type":"string"
            },
            "hydra:first":{
               "type":"string",
               "format":"iri-reference"
            },
            "hydra:last":{
               "type":"string",
               "format":"iri-reference"
            },
            "hydra:next":{
               "type":"string",
               "format":"iri-reference"
            }
         }
      },
      "hydra:search":{
         "type":"object",
         "properties":{
            "@type":{
               "type":"string"
            },
            "hydra:template":{
               "type":"string"
            },
            "hydra:variableRepresentation":{
               "type":"string"
            },
            "hydra:mapping":{
               "type":"array",
               "items":{
                  "type":"object",
                  "properties":{
                     "@type":{
                        "type":"string"
                     },
                     "variable":{
                        "type":"string"
                     },
                     "property":{
                        "type":"string"
                     },
                     "required":{
                        "type":"boolean"
                     }
                  }
               }
            }
         }
      }
   },
[...]

PR #3807, opened 2020-10-30
fix: use getArrayCopy for assertMatchesJsonSchema

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This PR is part of a set of PR whose purpose is to make the assertions `assertMatchesResourceItemJsonSchema` and `assertMatchesResourceCollectionJsonSchema` work correctly.
The PR are:
- https://github.com/api-platform/core/pull/3803
- https://github.com/api-platform/core/pull/3804
- https://github.com/api-platform/core/pull/3806
- https://github.com/api-platform/core/pull/3807 (this one)

Without doing this, the refs are not correctly expanded and the value is not correctly validated against the JSON Schema.
It's because of this line:
https://github.com/justinrainbow/json-schema/blob/fa4d2d3c1e40a222ded125b679891086c9a6c7d6/src/JsonSchema/SchemaStorage.php#L99

It seems that `property_exists` on an `ArrayObject` is not considering a key as a property.
See: https://3v4l.org/60Mbh

### DIFF
--- a/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
+++ b/src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
@@ -113,14 +113,14 @@ trait ApiTestAssertionsTrait
     {
         $schema = self::getSchemaFactory()->buildSchema($resourceClass, $format, Schema::TYPE_OUTPUT, OperationType::COLLECTION, $operationName);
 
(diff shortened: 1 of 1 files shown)


soyuka:
thanks @alanpoulain

PR #3808, opened 2020-10-31
[ValidationException] Allow customization of validation error status code

ahmed-bhs:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no 
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |  
| License       | MIT
| Doc PR |  api-platform/docs#1206

In general, I think it's a good practice to use *400 Bad Request* http status code when the request body is malformed (i.e. incorrect syntax in the payload or invalid parameters in the URL) and  use *422 Unprocessable Entity*  when the query is semantically incorrect.

According to the RFC https://tools.ietf.org/html/rfc4918#section-11.2 the code 422 seems to me to be much more appropriate and specific in a case of data validation failure.
Cause the server understands the content type of the request entity and the syntax of the request entity is correct but was unable to process the data because its semantically erroneous.

The idea, is to give the hand to configure the status code for the ValidationException under `exception_to_status` section, just like this :

```yaml
api_platform:
    exception_to_status:
        ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException: !php/const Symfony\Component\HttpFoundation\Response::HTTP_UNPROCESSABLE_ENTITY

```
Any way, just let me know if this requires to add some unit tests or some documentation, (if it'is an interesting idea).
Best regards,

### DIFF
--- a/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
+++ b/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
@@ -47,7 +47,7 @@ final class ValidationExceptionListener
             return;
         }
         $exceptionClass = \get_class($exception);
-        $statusCode = Response::HTTP_BAD_REQUEST;
+        $statusCode = Response::HTTP_UNPROCESSABLE_ENTITY;
 
         foreach ($this->exceptionToStatus as $class => $status) {
             if (is_a($exceptionClass, $class, true)) {
--- a/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
+++ b/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
@@ -28,11 +28,13 @@ final class ValidationExceptionListener
 {
     private $serializer;
     private $errorFormats;
+    private $exceptionToStatus;
 
-    public function __construct(SerializerInterface $serializer, array $errorFormats)
+    public function __construct(SerializerInterface $serializer, array $errorFormats, array $exceptionToStatus = [])
     {
         $this->serializer = $serializer;
         $this->errorFormats = $errorFormats;
+        $this->exceptionToStatus = $exceptionToStatus;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
```diff
@@ -44,12 +46,22 @@ public function onKernelException(ExceptionEvent $event): void
         if (!$exception instanceof ValidationException) {
             return;
         }
+        $exceptionClass = \get_class($exception);
+        $statusCode = Response::HTTP_BAD_REQUEST;
```
dunglas, line 50: Maybe should we default to `422` here to be as correct as possible by default?

dunglas (approved):

src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
ahmed-bhs: @dunglas  I appreciate your feedback, I definitely agree with you. Just are you ok with updating the http code for GraphQL ? https://github.com/api-platform/core/blob/master/src/GraphQl/Serializer/Exception/ValidationExceptionNormalizer.php#L42

ahmed-bhs: @dunglas  I appreciate your feedback, I definitely agree with you. Just, are you ok with also updating the http code for GraphQL ? https://github.com/api-platform/core/blob/master/src/GraphQl/Serializer/Exception/ValidationExceptionNormalizer.php#L42

dunglas: It looks good to me, wdyt @alanpoulain?

soyuka (commented):
:+1: for this, can you rebase ?

ahmed-bhs:
@soyuka rebase is done, but It seems like most of the ci builds are failing after merging #3764 PR!

soyuka:
I'll handle the rest I though I just fixed the missing tests in https://github.com/api-platform/core/pull/3818 :)

soyuka:
Thanks @ahmed-bhs !

silverbackdan:
Quick question: Should this not have been a BC break? People expecting 400 responses now get 422 on validation errors.

bpolaszek:
> Quick question: Should this not have been a BC break? People expecting 400 responses now get 422 on validation errors - may be worth marking as BC in the changelog?

It does, so it should :-)

odoucet:
Hello,
I confirm this is a BC, I needed to update all unit tests to match 422 instead of 400.

ahmed-bhs:
@odoucet you can keep the old http code if you dont want to update your unit tests, all you need is to update the exception status code:

```yaml
api_platform:
    exception_to_status:
        ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException: !php/const Symfony\Component\HttpFoundation\Response::HTTP_BAD_REQUEST
```

Issue #3809, opened 2020-11-02
/docs endpoint with extended json+ld does not include the version


silverbackdan:
**API Platform version(s) affected**: 2.5.7

**Description**  
When requesting the `docs` endpoint with extended jsonld formatting, the version is not in the output. With json it is in `info -> version` I see the `info` node is not really necessary as the title and description become hydra properties.

**How to reproduce**  
visit `/docs.jsonld` or set appropriate headers on the request

**Possible Solution**  
Tracking and will take a look at raising a PR for a solution to include the version.

**Additional Context**  
n/a

silverbackdan:
PR raised to include the API version in the docs output for ld+json. #3810

PR #3810, opened 2020-11-02
Add version to json+ld docs endpoint output

silverbackdan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Tickets       | #3809
| License       | MIT
| Doc PR        | n/a

I consider this a bug fix as the plain json output will provide the API version in the `info.version` node. This was missing in the extended ld+json output.

I've also updated tests. I'm happy to make adjustments to the target branch or the json node path as appropriate.

### DIFF
--- a/features/hydra/docs.feature
+++ b/features/hydra/docs.feature
@@ -35,6 +35,7 @@ Feature: Documentation support
     And the JSON node "hydra:description" should contain "This is a test API."
     And the JSON node "hydra:description" should contain "Made with love"
     And the JSON node "hydra:entrypoint" should be equal to "/"
+    And the JSON node "version" should be equal to "0.0.0"
     # Supported classes
     And the Hydra class "The API entrypoint" exists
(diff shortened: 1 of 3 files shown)


soyuka:
Hi, it's considered as a bad practice to version apis like this, the hydra spec recommends to use different adresses see http://www.hydra-cg.com/spec/latest/core/#api-versions. Note that this is supported with OpenAPI but still considered as a bad practice. 
If you really need to do so, you can always decorate the DocumentationNormalizer to add your own data.

silverbackdan:
Ah OK, thanks for clarifying @soyuka - I was looking to extend the API version in the openapi docs with the API Components Bundle so that it would also detail which version of the bundle is being used in brackets so it could be parsed by the front-end. I'll use the json endpoint for now but perhaps when we get working together on this project again we can look at a better way.

PR #3817, opened 2020-11-06
[JSON Schema] Manage correctly nullability

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Following https://github.com/api-platform/core/pull/3402 and the list of PR from https://github.com/api-platform/core/pull/3807.

In order to validate correctly nullable data against the generated JSON Schema (using `assertMatchesResourceItemJsonSchema` and `assertMatchesResourceCollectionJsonSchema`), the schema needs to be compliant with the specification for nullability:

```json
{"type": ["string", "null"]}
```

However OpenAPI < 3.1 doesn't understand this syntax (OpenAPI 3.1 will at least! https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0) and only understands:

```json
{"type": "string", "nullable": true}
```

That's why we need a condition when the schema is generated for JSON Schema only.

@Ocramius you may be interested.

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -161,6 +161,17 @@ final class TypeFactory implements TypeFactoryInterface
             ];
         }
 
+        if ($schema && Schema::VERSION_JSON_SCHEMA === $schema->getVersion()) {
+            return array_merge(
+                $jsonSchema,
+                [
+                    'type' => \is_array($jsonSchema['type'])
+                        ? array_merge($jsonSchema['type'], ['null'])
+                        : [$jsonSchema['type'], 'null'],
+                ]
+            );
+        }
+
         return array_merge($jsonSchema, ['nullable' => true]);
     }
 }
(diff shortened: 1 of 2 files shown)


Ocramius:
Overall OK for me - would be easier to just say that we generate OpenAPI 3.1 schema here though :shrug:

alanpoulain:
Not really because we are not in an OpenAPI context here (pure JSON Schema use). However we could also introduce a new key for OpenAPI 3.1, like you suggested in your PR. But maybe when it will be released, don't you think?

Ocramius:
Wait... OpenAPI 3.1 is still not out yet? :O

alanpoulain:
Sadly :cry: https://github.com/OAI/OpenAPI-Specification/releases

soyuka:
Anyways we're going to support only the latest openapi version, by that I mean that we're probably not going to handle multiple open api versions, too hard to maintain. 

@alanpoulain you should also try this on master as the OpenAPI code has been refactored there.

alanpoulain:
:thinking: The code is exactly the same on master, isn't it?
https://github.com/api-platform/core/blob/eae59dfb76027f705600906d57d0a2fd67d1532e/src/JsonSchema/TypeFactory.php#L172
I think it would be merged without conflicts.

soyuka:
The code for the schema yes the code for the open api generation no.

soyuka (approved):

PR #3770, opened 2020-10-18
Add support for @Ignore annotation

TNAJanssen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT
| Doc PR        | none

Issue #3370, opened 2020-01-30
Embeddables require an identifier


jannes-io:
**API Platform version(s) affected**: ^2.1

**Description**  
Doctrine embeddables require a scalar identifier for api-platform to recognize them. This makes no sense since the fields are placed directly into the parent's table.

**How to reproduce**  
ORM:
```
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping">
    <embeddable name="App\Component\Core\Model\DateRange">
        <field name="from" type="date"/>
        <field name="to" type="date"/>
    </embeddable>
</doctrine-mapping>
```
api-platform:
```
resources:
  App\Component\Core\Model\DateRange:
    attributes:
      denormalization_context:
        groups: ['date_range']
    collectionOperations:
      get:
        normalization_context:
          groups: ['date_range']
    itemOperations:
      get:
        normalization_context:
          groups: ['date_range']
```

Then create some entity that uses the embed.
```
<embedded name="visible" class="DateRange"/>
<embedded name="register" class="DateRange"/>
<embedded name="process" class="DateRange"/>
```

Now I can add a field "$id: int" with  just default value `0` and not assign it to any normalization group but this pollutes the table so it's not a viable workaround.

**Possible Solution**  
Remove the requirement for an id on sub-entities, since a sub-entity is not always a relationship.

**Additional Context**

PR #3820, opened 2020-11-08
Fix 3770

TheCrealm:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Original PR       | #3770 
| License       | MIT
| Doc PR        | none

This PR implements the @Ignore annotation. Original PR: #3370

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -353,11 +353,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
         $allowedAttributes = [];
         foreach ($propertyNames as $propertyName) {
-            if(
+            if (
                 null != $attributesMetadata && \array_key_exists($propertyName, $attributesMetadata) &&
                 method_exists($attributesMetadata[$propertyName], 'isIgnored') &&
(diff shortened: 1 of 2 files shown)


TheCrealm:
By fixing this, i discovered that the @ignore annotation was introduced with symfony 5.1.
Because of that, this will always fail for php7.1 unit tests as php7.2 is the minimum php version for this.

(*Happy first pr noises*)

soyuka:
May you add a unit test?

TheCrealm:
Is this ok? @soyuka 
i looked at the other tests to figure out how to do it. At the end its covered :)

soyuka (approved):
Good work, thanks for taking this over!

lucasgranberg:
We are discussing reverting this change at https://github.com/api-platform/api-platform/issues/1566 because this doesn't apply to schema generation so it would be better to do this with property metadata using `$propertyMetadata->withWritable(false)->withReadable(false)->isInitializable(false);`.
Adding a flag like `isIgnored` to the propertyMetadata could even be preferable to make all other extractors quit early.

What do people involved with this change think of that? The other option would be to do this separately for schema generation.

PR #3821, opened 2020-11-09
Parameter for pagination cannot be arrays

Oipnet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Today when you try to use an array for client pagination parameters the system return the default value

For example if in the configuration file you set items_per_page_parameter_name: 'pagination.perpage' it will find in the context the value pagination[perpage]

### DIFF
--- a/src/DataProvider/Pagination.php
+++ b/src/DataProvider/Pagination.php
@@ -265,6 +265,17 @@ final class Pagination
     private function getParameterFromContext(array $context, string $parameterName, $default = null)
     {
         $filters = $context['filters'] ?? [];
+        $parameterAsArray = explode('.', $parameterName);
+        if (count($parameterAsArray) > 1) {
+            foreach ($parameterAsArray as $key => $parameter) {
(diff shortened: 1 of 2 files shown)


soyuka:
Not sure to understand why you'd need to split the string, what is the use case behind it? Please add a test.

Oipnet:
I use for my frontend KTDatatable and I'm not able to change easily the pagination system.
For exemple it send me this query parameters :
pagination[page]: 1
pagination[pages]: 1
pagination[perpage]: 10
pagination[total]: 4

Today it's not possible to set this kind of parameter name.

I'am not sure split the string is the best solution but I don't see better

alanpoulain:
I don't really like the idea of having this feature, it makes it more complicated for not a lot of advantages.

soyuka:
I'm sorry but we won't take this patch as we feel it is too specific. Feel free to add this on your own code, you can also parse pagination parameters yourself.

PR #3837, opened 2020-11-19
OpenAPI - make all info model fields configurable

WhiteRabbitDE:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | Not yet

Enables the configuration of ToS, Contact and License in the SwaggerUI. 

I wasnt sure about the contact model. All properties there have a default value which is not needed imho, contact e.g. also works if you only set one field. I think it would be better to make them nullable so they are not serialized at all.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -106,6 +106,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $this->registerCommonConfiguration($container, $config, $loader, $formats, $patchFormats, $errorFormats);
         $this->registerMetadataConfiguration($container, $config, $loader);
         $this->registerOAuthConfiguration($container, $config);
+        $this->registerOpenApiConfiguration($container, $config);
         $this->registerSwaggerConfiguration($container, $config, $loader);
         $this->registerJsonApiConfiguration($formats, $loader);
         $this->registerJsonLdHydraConfiguration($container, $formats, $loader, $config['enable_docs']);
@@ -708,6 +709,16 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         if (isset($bundles['SecurityBundle'])) {
             $loader->load('security.xml');
         }
+    }
+
+    private function registerOpenApiConfiguration(ContainerBuilder $container, array $config): void
+    {
+        $container->setParameter('api_platform.openapi.termsOfService', $config['openapi']['termsOfService']);
+        $container->setParameter('api_platform.openapi.contact.name', $config['openapi']['contact']['name']);
+        $container->setParameter('api_platform.openapi.contact.url', $config['openapi']['contact']['url']);
+        $container->setParameter('api_platform.openapi.contact.email', $config['openapi']['contact']['email']);
+        $container->setParameter('api_platform.openapi.license.name', $config['openapi']['license']['name']);
+        $container->setParameter('api_platform.openapi.license.url', $config['openapi']['license']['url']);
     }
 
     private function buildDeprecationArgs(string $version, string $message): array
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -77,7 +77,9 @@ final class OpenApiFactory implements OpenApiFactoryInterface
     public function __invoke(array $context = []): OpenApi
     {
         $baseUrl = $context[self::BASE_URL] ?? '/';
-        $info = new Model\Info($this->openApiOptions->getTitle(), $this->openApiOptions->getVersion(), trim($this->openApiOptions->getDescription()));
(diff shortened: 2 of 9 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -178,6 +178,12 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         $container->setParameter('api_platform.title', $config['title']);
         $container->setParameter('api_platform.description', $config['description']);
         $container->setParameter('api_platform.version', $config['version']);
+        $container->setParameter('api_platform.termsOfService', $config['termsOfService']);
```
soyuka, line 181: Maybe that we should put this under an `openapi` node?

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -77,7 +77,9 @@ public function __construct(ResourceNameCollectionFactoryInterface $resourceName
     public function __invoke(array $context = []): OpenApi
     {
         $baseUrl = $context[self::BASE_URL] ?? '/';
-        $info = new Model\Info($this->openApiOptions->getTitle(), $this->openApiOptions->getVersion(), trim($this->openApiOptions->getDescription()));
+        $contact = new Model\Contact($this->openApiOptions->getContactName(), $this->openApiOptions->getContactUrl(), $this->openApiOptions->getContactEmail());
+        $license = null === $this->openApiOptions->getLicenseName() ? null : new Model\License($this->openApiOptions->getLicenseName(), $this->openApiOptions->getLicenseUrl());
+        $info = new Model\Info($this->openApiOptions->getTitle(), $this->openApiOptions->getVersion(), trim($this->openApiOptions->getDescription()), $this->openApiOptions->getTermsOfService(), $contact, $license);
```
soyuka, line 82: Why `trim`?

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -355,12 +357,12 @@ private function getFiltersParameters(ResourceMetadata $resourceMetadata, string
                     $data['openapi']['allowEmptyValue'] ?? true,
                     $schema,
                     'array' === $schema['type'] && \in_array($data['type'],
-                    [Type::BUILTIN_TYPE_ARRAY, Type::BUILTIN_TYPE_OBJECT], true) ? 'deepObject' : 'form',
+                        [Type::BUILTIN_TYPE_ARRAY, Type::BUILTIN_TYPE_OBJECT], true) ? 'deepObject' : 'form',
```
soyuka, line 360: Tab character was added by mistake?

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -355,12 +357,12 @@ private function getFiltersParameters(ResourceMetadata $resourceMetadata, string
...
                     'array' === $schema['type'],
                     $data['openapi']['allowReserved'] ?? false,
                     $data['openapi']['example'] ?? null,
                     isset($data['openapi']['examples']
-                ) ? new \ArrayObject($data['openapi']['examples']) : null);
+                    ) ? new \ArrayObject($data['openapi']['examples']) : null);
```
soyuka, line 365: Tab character was added by mistake?

soyuka:
indeed make them nullable!

src/OpenApi/Factory/OpenApiFactory.php
WhiteRabbitDE: The trim() was added by you. I didnt want to remove it.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
WhiteRabbitDE: I also had this idea but reverted it because i think if we add an openapi section we should also move title, description and version there to have it consistent.  wdyt ?

src/OpenApi/Factory/OpenApiFactory.php
WhiteRabbitDE: I checked that part and dont see why it would be needed. Shall i remove it within this PR ?

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
soyuka: Some are used by the Hydra documentation normalizer also. As these new ones are openapi specifics I'd add them under an open api node.

src/OpenApi/Factory/OpenApiFactory.php
soyuka: oh you're right I had an issue with a trailing newline, please keep that sorry for the comment.

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
WhiteRabbitDE: You are right. I moved only the new ones into an openapi section.

soyuka (approved):

soyuka:
Thanks @WhiteRabbitDE !

Issue #1856, opened 2018-04-12
VarnishPurger cannot handle alot of entities


Simperfit:
Based on this commit: https://github.com/Simperfit/api-perf/commit/4201cc28c4ddb7c493df31a095d8ad71b8fe1297#diff-c828d4117439a2216f86c406d522c79c

To reproduce it 

```
docker-compose pull
docker-compose up -d
docker-compose exec php bin/console d:f:l
```

Results in:
```

In RequestException.php line 113:
                                                                                   
  Client error: `BAN http://cache-proxy` resulted in a `400 Bad Request` response  
 
```

teohhanhui:
Would #1776 help? :smile:

BatsaxIV:
Any news on this issue?

dunglas:
As a workaround, you can do that: https://github.com/api-platform/demo/pull/24/commits/0bc68491cf2859538f6749218a227095ec0ec852#diff-628212751f227bfce20484adbd0d4191R1

Maybe can we enable the Varnish subsystem only for the `prod` env in the default skeleton too?

teohhanhui:
I'd again argue that there's a need for the developer to verify that the caching (especially the invalidation) by reverse proxy works as expected in the `dev` env. Surprises are bad.

teohhanhui:
And it's a bug we need to fix anyway.

dunglas:
The only way to "fix" it is to skip the listener in those specific batch cases. It's up the developper to purge Varnish using another way (probably a full purge) in those cases. But dev commands such as Doctrine Migrations ones must work.

teohhanhui:
A more efficient cache invalidation method should help. But yes, of course we could never guarantee that there will not be a timeout if a certain higher limit is still hit. That will be up to the developer to improvise. My point of disagreement is mainly on disabling it out of the box in the `dev` env.

dunglas:
We may at least disable it in CLI + dev to allow such commands to work.

dunglas:
Proposal: if in debug mode + cli : log a warning but disable it.

teohhanhui:
Perhaps also relevant: https://github.com/api-platform/core/pull/1286#issuecomment-318607817

Glideh:
I'm also having this issue trying to import data with Doctrine from a command (Which I need to do in production too).
Do we have a workaround ?
For now I've simply reduced my bulk size to avoid this.

jocel1:
[...]

PR #3843, opened 2020-11-21
Chunk headers to comply with Varnish max length

### LINKED ISSUE #3168: Varnish cache tags with a big number of entities

julienfalque:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3168 and #1856
| License       | MIT
| Doc PR        | -

This PR chunks the regex to send it in separate requests to avoid `400 Bad Request` responses from Varnish when the maximum header length is exceeded. The default limit is 8000 for now since it's Varnish's default value, but that could be easily made configurable in another iteration (maybe for next minor release instead of 2.5?).

### DIFF
--- a/src/HttpCache/VarnishPurger.php
+++ b/src/HttpCache/VarnishPurger.php
@@ -24,14 +24,18 @@ use GuzzleHttp\ClientInterface;
  */
 final class VarnishPurger implements PurgerInterface
 {
+    private const DEFAULT_VARNISH_MAX_HEADER_LENGTH = 8000;
+
     private $clients;
+    private $maxHeaderLength;
 
     /**
      * @param ClientInterface[] $clients
      */
-    public function __construct(array $clients)
+    public function __construct(array $clients, int $maxHeaderLength = self::DEFAULT_VARNISH_MAX_HEADER_LENGTH)
     {
         $this->clients = $clients;
+        $this->maxHeaderLength = $maxHeaderLength;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dunglas (approved):
LGTM

soyuka:
thansk @julienfalque !

soyuka:
:+1: to make this value configurable

julienfalque:
I'll work on that then :)

PR #3857, opened 2020-11-28
Add the violation code to the violation properties

theofidry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | potentially
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Unnecessary

When returning the list of violations, in some context it is better to rely on a violation code/identifier rather than the message itself. The Symfony Validator already provides such a feature (although may not be enforced by all custom constraints), so we can directly take advantage of it.

If you're fine with the idea I'll look into finishing this PR

### DIFF
--- a/features/hal/problem.feature
+++ b/features/hal/problem.feature
@@ -22,7 +22,8 @@ Feature: Error handling valid according to RFC 7807 (application/problem+json)
       "violations": [
         {
           "propertyPath": "name",
-          "message": "This value should not be blank."
+          "message": "This value should not be blank.",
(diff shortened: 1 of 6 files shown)


alanpoulain:
Seems fine to me :+1: I don't see why it could be a BC break though.

theofidry:
PR ready then :)

soyuka (approved):

alanpoulain (approved):

alanpoulain:
Should we merge it or we wait for the 2.6 release?

PR #3866, opened 2020-12-01
[Elasticsearch] inject denormalization context into `Paginator` to have access to serialization groups

hhamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | probably yes...
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | ~

This patch enables the Elasticsearch paginator to receive the serialization context so that documents can be denormalized to objects using the serialization groups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.7.0
+
+* Elasticsearch: The `Paginator` class constructor now receives the denormalization context to support denormalizing documents using serialization groups. This change may cause potential **BC** breaks for existing applications as denormalization was previously done without serialization groups.
+
 ## 2.6.0
 
 * Display the API Platform's version in the debug-bar
--- a/src/Bridge/Elasticsearch/DataProvider/Paginator.php
+++ b/src/Bridge/Elasticsearch/DataProvider/Paginator.php
@@ -33,14 +33,16 @@ final class Paginator implements \IteratorAggregate, PaginatorInterface
     private $limit;
     private $offset;
     private $cachedDenormalizedDocuments = [];
+    private $denormalizationContext = [];
 
-    public function __construct(DenormalizerInterface $denormalizer, array $documents, string $resourceClass, int $limit, int $offset)
+    public function __construct(DenormalizerInterface $denormalizer, array $documents, string $resourceClass, int $limit, int $offset, array $denormalizationContext = [])
     {
         $this->denormalizer = $denormalizer;
         $this->documents = $documents;
         $this->resourceClass = $resourceClass;
         $this->limit = $limit;
         $this->offset = $offset;
+        $this->denormalizationContext = $denormalizationContext;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.7.0
+
+* Elasticsearch: **BC** Paginator now receives the denormalization context to support denormalizing documents using serialization groups.
```
dunglas, line 5: Why is it a BC break? This change looks backward compatible (the `Paginator` class is `final` so adding an optional argument is allowed).

dunglas (changes requested):
Just a minor edit to do to the changelog, otherwise 👍!

tests/Bridge/Elasticsearch/DataProvider/CollectionDataProviderTest.php
```diff
@@ -92,6 +92,12 @@ public function testSupports()
 
     public function testGetCollection()
     {
+        $context = [
+            'groups' => ['custom'],
+        ];
+
+        $operationName = 'get';
```
soyuka, line 99: variables not needed?

tests/Bridge/Elasticsearch/DataProvider/CollectionDataProviderTest.php
```diff
@@ -92,6 +92,12 @@ public function testSupports()
...
+
```
soyuka, line 100:
```suggestion
```

src/Bridge/Elasticsearch/DataProvider/Paginator.php
```diff
@@ -102,6 +104,8 @@ public function getItemsPerPage(): float
      */
     public function getIterator(): \Traversable
     {
+        $denormalizationContext = array_merge([AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES => true], $this->denormalizationContext);
```
soyuka, line 107: is the variable really needed?

CHANGELOG.md
hhamon: I'm not totally sure but I think it may be a BC Break because for now the context is not given. This means, when documents are deserialized to objects, it's currently done without any serialization groups. Passing the context that includes groups now, it may affect some existing applications.

src/Bridge/Elasticsearch/DataProvider/Paginator.php
hhamon: Yes, it prevents performing the same `array_merge` statement inside the `foreach` loop.

CHANGELOG.md
dunglas: It makes sense. Maybe should we add this note on the changelog then.

hhamon: `CHANGELOG` has been updated

dunglas:
Thanks Hugo!

PR #3870, opened 2020-12-02
Allow mixed type value when apply to DateFilter.

### LINKED ISSUE #3858: ORM DateFilter throw fatal error when user submit unexpected query params.

**Description**  

Currently when user submit unexpected query param type like: `&date[after][][][]=1` fatal error will be throw because it's wrong type when pass it to `andWhere` method.

**How to reproduce**  

Submit query param `date[after][][][]=1`

**Possible Solution**  

Allow mixed type of `value` arg at [DateFilter#L131](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/DateFilter.php#L131) and change exception type to `\Throwable` at [DateFilter#L136](https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Filter/DateFilter.php#L136).

vuongxuongminh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3858 
| License       | MIT

### DIFF
--- a/src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
+++ b/src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
@@ -81,12 +81,9 @@ trait DateFilterTrait
         ];
     }
 
-    /**
-     * Normalize the value.
-     */
     private function normalizeValue($value, string $operator): ?string
     {
-        if (!\is_string($value)) {
+        if (false === \is_string($value)) {
             $this->getLogger()->notice('Invalid filter ignored', [
                 'exception' => new InvalidArgumentException(sprintf('Invalid value for "[%s]", expected string', $operator)),
             ]);
--- a/src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
+++ b/src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Doctrine\Common\Filter;
 
 use ApiPlatform\Core\Bridge\Doctrine\Common\PropertyHelperTrait;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
 
 /**
  * Trait for filtering the collection by date intervals.
@@ -78,5 +79,21 @@ trait DateFilterTrait
                 'required' => false,
             ],
         ];
+    }
+
+    /**
+     * Normalize the value.
+     */
+    private function normalizeValue($value, string $operator): ?string
+    {
+        if (!\is_string($value)) {
+            $this->getLogger()->notice('Invalid filter ignored', [
+                'exception' => new InvalidArgumentException(sprintf('Invalid value for "[%s]", expected string', $operator)),
+            ]);
+
+            return null;
+        }
+
+        return $value;
     }
 }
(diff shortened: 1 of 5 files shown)


src/Bridge/Doctrine/MongoDbOdm/Filter/DateFilter.php
```diff
@@ -107,11 +107,11 @@ protected function filterProperty(string $property, $values, Builder $aggregatio
     /**
      * Adds the match stage according to the chosen null management.
      */
-    private function addMatch(Builder $aggregationBuilder, string $field, string $operator, string $value, string $nullManagement = null): void
+    private function addMatch(Builder $aggregationBuilder, string $field, string $operator, $value, string $nullManagement = null): void
     {
         try {
             $value = new \DateTime($value);
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
```
dunglas, line 114: This should be reverted. A `\Throwable` should never be thrown. Instead, we should test that the $value is a string before passing it to the `DateTime` constructor.

dunglas (changes requested):

src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
```diff
@@ -79,4 +80,20 @@ protected function getFilterDescription(string $property, string $period): array
             ],
         ];
     }
+
+    /**
+     * Normalize the value.
+     */
+    private function normalizeValue($value, string $operator): ?string
+    {
+        if (!\is_string($value)) {
```
dunglas, line 89: Nitpicking but I would transform this test in a guard clause to simplify.

dunglas (approved):
LGTM!

dunglas (approved):

src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
```diff
@@ -79,4 +80,20 @@ protected function getFilterDescription(string $property, string $period): array
             ],
         ];
     }
+
+    /**
+     * Normalize the value.
+     */
+    private function normalizeValue($value, string $operator): ?string
```
drupol, line 87: I might be picky on this, but wouldn't it be better (semantically wise) to use early returns?

First exclude the bad scenario and then only return the happy scenario ?

```suggestion
    private function normalizeValue($value, string $operator): ?string
    {
            if (false === \is_string($value)) {
               $this->getLogger()->notice('Invalid filter ignored', [
                   'exception' => new InvalidArgumentException(sprintf('Invalid value for "[%s]", expected string', $operator)),
               ]);

               return null;
            }

            return $value;
    }
```

dunglas: Using a guard clause reduces the cyclomatic complexity and makes the code more readable (imho). I would keep this block as is.

drupol: But this is what I'm trying to say here :-)

First the guard (_the unhappy scenario_), then relevant returned `$value` (_the happy scenario_).

vuongxuongminh:
Is there anything needed to get this merged?

src/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
alanpoulain: I'm :+1: for @drupol's suggestion.

drupol: Let's not forget this thing :-)

drupol: Oh it's done already ! Great :-)

drupol (approved):

alanpoulain:
Thank you @vuongxuongminh.

drupol:
Yes, thanks !

PR #3871, opened 2020-12-02
Throw an error when no identifier is defined

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | kinda
| Deprecations? | no
| Tickets       | 
| License       | MIT

### DIFF
--- a/src/Api/IdentifiersExtractor.php
+++ b/src/Api/IdentifiersExtractor.php
@@ -57,6 +57,10 @@ final class IdentifiersExtractor implements IdentifiersExtractorInterface
             }
         }
 
+        if (!$identifiers) {
+            throw new RuntimeException(sprintf('No identifier defined "%s". You should add #[\ApiPlatform\Core\Annotation\ApiProperty(identifier: true)]" on the property identifying the resource."', $resourceClass));
+        }
+
         return $identifiers;
     }
 
--- a/tests/Api/IdentifiersExtractorTest.php
+++ b/tests/Api/IdentifiersExtractorTest.php
@@ -202,6 +202,20 @@ class IdentifiersExtractorTest extends TestCase
         $this->assertSame(['foo' => 'woot'], $identifiersExtractor->getIdentifiersFromItem($item));
     }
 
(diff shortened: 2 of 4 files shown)


src/Api/IdentifiersExtractor.php
```diff
@@ -57,6 +57,10 @@ public function getIdentifiersFromResourceClass(string $resourceClass): array
             }
         }
 
+        if (!$identifiers) {
+            throw new RuntimeException(sprintf('No identifier found in "%s"', $resourceClass));
```
dunglas, line 61:
```suggestion
            throw new RuntimeException(sprintf('No identifier defined "%s". You should add #[\ApiPlatform\Core\Annotation\ApiProperty(identifier: true)]" on the property identifying the resource.', $resourceClass));
```

dunglas (changes requested):

src/Api/IdentifiersExtractor.php
```diff
@@ -57,6 +57,10 @@ public function getIdentifiersFromResourceClass(string $resourceClass): array
             }
         }
 
+        if (!$identifiers) {
+            throw new RuntimeException(sprintf('No identifier defined "%s". You should add #[\ApiPlatform\Core\Annotation\ApiProperty(identifier: true)]" on the property identifying the resource."', $resourceClass));
```
maks-rafalko, line 61: What do you think about 

```diff
- No identifier defined "%s".
+ No identifier defined in "%s".
```

I mean, "in" is missed from my point of view

julienfalque:
This new exception seems to break POST/PUT endpoints that use a DTO without id as the resource. Is that intended? If so, what are the alternatives?

soyuka:
It should not break unless your DTO is a resource. Please open an issue and provide a reproducer I definitely want to check this!

maks-rafalko:
@soyuka there is already an issue https://github.com/api-platform/core/issues/3975

Please re-read it, author has edited description

PR #3873, opened 2020-12-02
fix: set stateless to false by default and remove the deprecation

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | n/a
| Tickets       | https://github.com/api-platform/core/pull/3436, https://github.com/api-platform/api-platform/pull/1703, https://github.com/symfony/recipes/pull/853
| License       | MIT
| Doc PR        | n/a

It's not possible to enable `stateless` by default because it conflicts with the default configuration of the Symfony firewall. Let's disable this option in the bundle, and enable it in the distribution.

### DIFF
--- a/tests/Fixtures/app/config/config_common.yml
+++ b/tests/Fixtures/app/config/config_common.yml
@@ -66,11 +66,6 @@ api_platform:
     collection:
         order_parameter_name:          'order'
         order:                         'ASC'
-        pagination:
-            client_enabled:            true
-            client_items_per_page:     true
-            client_partial:            true
-            items_per_page:            3
     exception_to_status:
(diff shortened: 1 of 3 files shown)


tests/Fixtures/app/config/config_common.yml
```diff
@@ -66,11 +66,6 @@ api_platform:
     collection:
         order_parameter_name:          'order'
         order:                         'ASC'
-        pagination:
```
dunglas, line 69: This isn't related, but prevent some deprecations to be triggered.

PR #3877, opened 2020-12-03
Make "hydra:mapping" item's property nullable

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

## Context

My entites have the following filter, to select properties. Parameter name have been overrode.
` * @ApiFilter(PropertyFilter::class, arguments={"parameterName": "fields"})`

And I got the following error in my tests, using `assertMatchesResourceCollectionJsonSchema` assertion

```
   │             ...
   │             3 => Array &9 (
   │                 '@type' => 'IriTemplateMapping'
   │                 'variable' => 'fields[]'
   │                 'property' => null
   │                 'required' => false
   │             )
   │             ...
   │     )
   │ ) matches the provided JSON Schema.
   │ hydra:search.hydra:mapping[3].property: NULL value found, but a string is required
```

### DIFF
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -121,7 +121,10 @@ final class SchemaFactory implements SchemaFactoryInterface
                                 'properties' => [
                                     '@type' => ['type' => 'string'],
                                     'variable' => ['type' => 'string'],
-                                    'property' => ['type' => 'string', 'nullable' => true],
+                                    'property' => [
+                                        'nullable' => true,
+                                        'anyOf' => [['type' => 'string'], ['type' => 'null']],
+                                    ],
                                     'required' => ['type' => 'boolean'],
                                 ],
                             ],
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -123,7 +123,7 @@ final class SchemaFactory implements SchemaFactoryInterface
                                     'variable' => ['type' => 'string'],
                                     'property' => [
                                         'nullable' => true,
-                                        'anyOf' => [['type' => 'string'], ['type' => 'null']],
+                                        'type' => ['string', 'null'],
                                     ],
                                     'required' => ['type' => 'boolean'],
                                 ],
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -74,15 +74,15 @@ final class SchemaFactory implements SchemaFactoryInterface
             $items = $schema['items'];
             unset($schema['items']);
 
+            $nullableStringDefinition = ['type' => 'string'];
+
             switch ($schema->getVersion()) {
                 case Schema::VERSION_JSON_SCHEMA:
-                    $mappingPropDefinition = ['type' => ['string', 'null']];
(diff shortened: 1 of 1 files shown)


GaryPEGEOT:
Turns out doesn't seems to support nullable properties: justinrainbow/json-schema#551

src/Hydra/JsonSchema/SchemaFactory.php
```diff
@@ -121,7 +121,10 @@ public function buildSchema(string $className, string $format = 'jsonld', string
                                 'properties' => [
                                     '@type' => ['type' => 'string'],
                                     'variable' => ['type' => 'string'],
-                                    'property' => ['type' => 'string'],
+                                    'property' => [
+                                        'nullable' => true,
```
alanpoulain, line 125: `'nullable' => true` is only for OpenAPI < 3.1.
I don't think we should add it here.

GaryPEGEOT: Ok. the change from string to `'type' => ['string', 'null'],` broke Behat pipelien, I'll update the tests as well if you're OK with the change ?

alanpoulain: Yes I think this change is necessary.

GaryPEGEOT: It seems that `'null'` is [valid type starting on Open-API 3.1](https://stackoverflow.com/questions/48111459/how-to-define-a-property-that-can-be-string-or-null-in-openapi-swagger#answer-48114322) and the schema generated by the `api:openapi:export` command is at "3.0.2", so it wont pass swagger-cli validation (which is normal). I could check the version to add `nullable` or `['string', 'null']` combo, but I don't think you have control over the generated minor version. Any idea on how to solve this ?

alanpoulain: It's more or less what's has been done here: https://github.com/api-platform/core/blob/f53bb64825950bf1da2b91e12673b465d3d5312b/src/JsonSchema/TypeFactory.php#L155-L184

GaryPEGEOT: Fixed ! I've kept nullable for `openapi` version, but if it's not necessary I'll move it to a ternary.

src/Hydra/JsonSchema/SchemaFactory.php
```diff
@@ -74,6 +74,17 @@ public function buildSchema(string $className, string $format = 'jsonld', string
...
 
+            switch ($schema->getVersion()) {
+                case Schema::VERSION_JSON_SCHEMA:
+                    $mappingPropDefinition = ['type' => ['string', 'null']];
+                    break;
+                case Schema::VERSION_OPENAPI:
+                    $mappingPropDefinition = ['type' => 'string', 'nullable' => true];
+                    break;
+                default:
+                    $mappingPropDefinition = ['type' => 'string'];
```
alanpoulain, line 85: Could you initialize the property before the switch and rename it `$nullableStringDefinition`? In case we want to reuse it elsewhere.

GaryPEGEOT: Fixed!

alanpoulain (approved):

alanpoulain:
Thank you @GaryPEGEOT.

PR #3885, opened 2020-12-05
Introduce `PaginatorFactoryInterface` interface

hhamon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1239

This MR introduces a new generic `PaginatorFactoryInterface` interface enabling for more flexibility for collection data providers to produce `PaginatorInterface` instance.

For example, this is mainly useful for the Elasticsearch implementation when applications need to embed more information that just the records to be paginated into the paginator. That's the case when using Elasticsearch `aggregations` (or faceting in other words) system. Having a factory enables to construct a `AggregatedPaginatorFactory` class that produces `AggregatedPaginator` instances. The later will contain the documents to be paginated but also the returned `aggregated` data from Elasticsearch. Finally, by implementing a custom normalizer (ie. `AggregatedCollectionNormalizer`) and checking the concrete type of the paginator, the normalizer can append the aggregated data alongside the collection of documents in the JSON response.

### DIFF
--- a/src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
+++ b/src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
@@ -20,7 +20,9 @@ use ApiPlatform\Core\Bridge\Elasticsearch\Exception\NonUniqueIdentifierException
 use ApiPlatform\Core\Bridge\Elasticsearch\Metadata\Document\Factory\DocumentMetadataFactoryInterface;
 use ApiPlatform\Core\DataProvider\ContextAwareCollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\Pagination;
+use ApiPlatform\Core\DataProvider\PaginatorFactoryInterface;
 use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
 use ApiPlatform\Core\Exception\ResourceClassNotFoundException;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use Elasticsearch\Client;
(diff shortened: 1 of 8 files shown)


src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
```diff
@@ -38,20 +38,20 @@ final class CollectionDataProvider implements ContextAwareCollectionDataProvider
...
-    private $denormalizer;
+    private $paginatorFactory;
     private $pagination;
     private $resourceMetadataFactory;
     private $collectionExtensions;
 
     /**
      * @param RequestBodySearchCollectionExtensionInterface[] $collectionExtensions
      */
-    public function __construct(Client $client, DocumentMetadataFactoryInterface $documentMetadataFactory, IdentifierExtractorInterface $identifierExtractor, DenormalizerInterface $denormalizer, Pagination $pagination, ResourceMetadataFactoryInterface $resourceMetadataFactory, iterable $collectionExtensions = [])
```
dunglas, line 49: It's a BC break. Can we avoid it by making the factory optional and just create the old service if not passed?

dunglas (changes requested):

src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
hhamon: The class is already final so I don't understand why it could be a BC break

hhamon: Requested changes have been made ;)

hhamon:
@dunglas ready for another review / merge ;)

src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
dunglas: The new argument must be the last one to prevent the break.

hhamon: The way I implemented the fix should be BC. I've simply remove the typehint on the `Denormalizer` argument to allow passing either a denormalizer or a paginator factory instance. When a denormalizer is given (aka legacy behavior), a deprecation warning message is raised to inform the user that passing a denormalizer will no longer be supported in APIP 3 in favor of passing a `PaginatorFactory` instance instead. When a denormalizer is given, it's wrapped into a paginator factory. So the code should be BC.

Anyway, I made another commit introducing a last optional argument for the paginator factory dependency.

dunglas (approved):

src/Bridge/Elasticsearch/DataProvider/CollectionDataProvider.php
```diff
@@ -111,13 +116,6 @@ public function getCollection(string $resourceClass, ?string $operationName = nu
...
 
-        return new Paginator(
-            $this->denormalizer,
-            $documents,
-            $resourceClass,
-            $limit,
-            $offset,
-            $context
-        );
+        return $this->paginatorFactory->createPaginator($documents, $limit, $offset, array_merge(['resourceClass' => $resourceClass], $context));
```
meyerbaptiste, line 119: You should use snake case and the union operator here:
```php
$context + ['resource_class' => $resourceClass]
```

src/Bridge/Elasticsearch/DataProvider/PaginatorFactory.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this->denormalizer = $denormalizer;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createPaginator($subject, int $limit, int $offset, array $context = []): PaginatorInterface
+    {
+        $resourceClass = $context['resourceClass'] ?? null;
```
meyerbaptiste, line 35: Is there a specific reason for passing the resource class in the context rather than in a parameter of the method?

meyerbaptiste (changes requested):
Does @api-platform/core-team think it would also be interesting to implement this new factory for the other providers now?

meyerbaptiste:
It would be nice to also add an example with the use of aggregations and this new factory in our documentation.

src/Bridge/Elasticsearch/DataProvider/PaginatorFactory.php
hhamon: Mostly saving one argument which doesn't trully make sense in the context of a paginator.

hhamon:
@meyerbaptiste I'll update the PR tomorrow ;)

hhamon:
@meyerbaptiste PR updated and PHPStan is failing but not because of me :)

hhamon:
@meyerbaptiste @dunglas Documentation PR has been submitted here: https://github.com/api-platform/docs/pull/1239

PR #3887, opened 2020-12-05
Fix for empty example and empty default value

### LINKED ISSUE #3882: Swagger UI always displays empty default and example keys
**API Platform version(s) affected**: 2.6.0 alpha, maybe 2.78

**Description**  

image

**How to reproduce**  

Use this entity: https://github.com/api-platform/api-platform/blob/master/api/src/Entity/Greeting.php

**Possible Solution**  

Check for empty values here: https://github.com/api-platform/core/blob/master/src/JsonSchema/SchemaFactory.php#L195-L201

Tayfun74:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3882
| License       | MIT

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -192,11 +192,11 @@ final class SchemaFactory implements SchemaFactoryInterface
             $propertySchema['externalDocs'] = ['url' => $iri];
         }
 
-        if (!isset($propertySchema['default']) && null !== ($default = $propertyMetadata->getDefault()) && !empty($default)) {
+        if (!isset($propertySchema['default']) && !empty($default = $propertyMetadata->getDefault())) {
             $propertySchema['default'] = $default;
         }
 
-        if (!isset($propertySchema['example']) && null !== ($example = $propertyMetadata->getExample()) && !empty($example)) {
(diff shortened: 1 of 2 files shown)


src/JsonSchema/SchemaFactory.php
```diff
@@ -192,11 +192,11 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
             $propertySchema['externalDocs'] = ['url' => $iri];
         }
 
-        if (!isset($propertySchema['default']) && null !== $default = $propertyMetadata->getDefault()) {
```
dunglas, line 195: You can remove the test for `null` values, it's already handled by `empty`.

dunglas (changes requested):
Could you also add a test please?

dunglas (approved):

dunglas:
Thank you @Tayfun74!

drupol:
@Tayfun74 I always try to never use `empty()` in my code because it's confusing.

I'm definitely sharing this point of view here: https://twitter.com/localheinz/status/1335638838279069696

PR #3888, opened 2020-12-05
Fix for additionalProp1 showing in example value

### LINKED ISSUE #3883: Swagger UI always  adds `"additionalProp1": {}` to the example value
**API Platform version(s) affected**: 2.6.0 alpha

**Description**  

image

**How to reproduce**  
Just install the distribution. It does that on any class mapped with API Resource.

**Possible Solution**  

We have no occurrences of `additionalProp1` in our code base, but they are occurences in Swagger UI's code base. It maybe caused by our generated JSON Schema (https://github.com/api-platform/core/blob/master/src/JsonSchema/SchemaFactory.php) or by a bug in Swagger UI.

Tayfun74:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3883
| License       | MIT

### DIFF
--- a/features/jsonld/context.feature
+++ b/features/jsonld/context.feature
@@ -80,7 +80,7 @@ Feature: JSON-LD contexts generation
               "hydra": "http://www.w3.org/ns/hydra/core#",
               "person": {
                   "@id": "http://example.com/id",
-                  "@var": "@id",
+                  "@type": "@id",
                   "foo": "bar"
               }
           }
--- a/features/jsonld/context.feature
+++ b/features/jsonld/context.feature
@@ -80,7 +80,7 @@ Feature: JSON-LD contexts generation
               "hydra": "http://www.w3.org/ns/hydra/core#",
               "person": {
                   "@id": "http://example.com/id",
-                  "@type": "@id",
+                  "@var": "@id",
                   "foo": "bar"
               }
           }
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -109,11 +109,18 @@ final class SchemaFactory implements SchemaFactoryInterface
             return $schema;
         }
 
-        $definition = new \ArrayObject(['type' => 'object', 'additionalProperties' => $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES] ?? false]);
+        $definition = new \ArrayObject(['type' => 'object']);
         $definitions[$definitionName] = $definition;
         if (null !== $resourceMetadata && null !== $description = $resourceMetadata->getDescription()) {
             $definition['description'] = $description;
         }
+
+        // additionalProperties are allowed by default, so it does not need to be set explicitly, unless allow_extra_attributes is false
+        // See https://json-schema.org/understanding-json-schema/reference/object.html#properties
+        if (isset($serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES]) && false === $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES]) {
+            $definition['additionalProperties'] = false;
+        }
+
         // see https://github.com/json-schema-org/json-schema-spec/pull/737
         if (
             Schema::VERSION_SWAGGER !== $version &&
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -109,7 +109,7 @@ final class SchemaFactory implements SchemaFactoryInterface
             return $schema;
         }
 
-        $definition = new \ArrayObject(['type' => 'object', 'additionalProperties' => $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES] ?? true]);
+        $definition = new \ArrayObject(['type' => 'object', 'additionalProperties' => $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES] ?? false]);
         $definitions[$definitionName] = $definition;
         if (null !== $resourceMetadata && null !== $description = $resourceMetadata->getDescription()) {
             $definition['description'] = $description;
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -109,6 +109,7 @@ final class SchemaFactory implements SchemaFactoryInterface
             return $schema;
         }
 
+        /** @var \ArrayObject<string, mixed> $definition */
         $definition = new \ArrayObject(['type' => 'object']);
         $definitions[$definitionName] = $definition;
         if (null !== $resourceMetadata && null !== $description = $resourceMetadata->getDescription()) {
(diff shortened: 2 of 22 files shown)


src/JsonSchema/SchemaFactory.php
```diff
@@ -109,7 +109,7 @@ public function buildSchema(string $className, string $format = 'json', string $
             return $schema;
         }
 
-        $definition = new \ArrayObject(['type' => 'object', 'additionalProperties' => $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES] ?? true]);
+        $definition = new \ArrayObject(['type' => 'object', 'additionalProperties' => $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES] ?? false]);
```
alanpoulain, line 112: I don't think this is the right fix. Allowing extra attributes should be `true` by default. We just want to remove the `additionalProperties=true` to the definition.

dunglas: Actually, do we really want to explicitly allow extra properties? It's technically allowed because of a limitation of the Symfony serializer (that has now been fixed, and we may disallow extra properties by default in API Platform 3), but it's more a side effects than something really intended isn't it? It looks good to me to not set this flag unless extra properties are explicitly allowed by the user.

Tayfun74: You mean not setting `additionalProperties` at all for the definition when allow_extra_attributes is not set in the $serializerContext?

dunglas: Yes! WDYT @alanpoulain @soyuka?

alanpoulain: OK so to be clear, if there is no `additionalProperties` in the definition, additional properties are **allowed** by default (see https://json-schema.org/understanding-json-schema/reference/object.html).
So I don't think we need to set it explicitly if `allow_extra_attributes` is set to `true`.
If there is no `allow_extra_attributes` in the serializer context, its value is considered `true` by default. So I think we should keep it this way for JSON Schema and don't explicitly set it too.
We *only* need to set it explicitly to `false` when `allow_extra_attributes` is set to `false`.
This is my opinion :slightly_smiling_face:

dunglas: I don't agree. It's not excepted to add extra properties by default to most API Platform documents and the additional property displayed in Swagger UI pollutes the example, degrading the user experience when using the sandbox (or just when reading the docs of the API).

dunglas: Si my proposal is to not set the key only when the context option is explicitly set to `true`.

alanpoulain: I don't follow you. The additional property is displayed in Swagger UI *because* `additionalProperties` is explicitly set to true.
If you follow my proposal, this problem doesn't exist anymore.

Tayfun74: Alan's proposal looks good to me

src/JsonSchema/SchemaFactory.php
```diff
@@ -109,11 +109,19 @@ public function buildSchema(string $className, string $format = 'json', string $
...
+        /** @var \ArrayObject<string, mixed> $definition */
+        $definition = new \ArrayObject(['type' => 'object']);
         $definitions[$definitionName] = $definition;
         if (null !== $resourceMetadata && null !== $description = $resourceMetadata->getDescription()) {
             $definition['description'] = $description;
         }
+
+        // additionalProperties are allowed by default, so it does not need to be set explicitly, unless allow_extra_attributes is false
+        // See https://json-schema.org/understanding-json-schema/reference/object.html#properties
+        if (isset($serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES]) && false === $serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES]) {
```
alanpoulain, line 121:
```suggestion
        if (false === ($serializerContext[AbstractNormalizer::ALLOW_EXTRA_ATTRIBUTES] ?? true)) {
```

alanpoulain:
Only the CS now :slightly_smiling_face:

features/jsonld/context.feature
```diff
@@ -80,7 +80,7 @@ Feature: JSON-LD contexts generation
               "hydra": "http://www.w3.org/ns/hydra/core#",
               "person": {
                   "@id": "http://example.com/id",
-                  "@type": "@id",
+                  "@var": "@id",
```
alanpoulain, line 83: Please revert this.

tests/Fixtures/TestBundle/Document/JsonldContextDummy.php
```diff
@@ -40,7 +40,7 @@ class JsonldContextDummy
      *     attributes={
      *         "jsonld_context"={
      *             "@id"="http://example.com/id",
-     *             "@type"="@id",
+     *             "@var"="@id",
```
alanpoulain, line 43: Please revert this.

tests/Fixtures/TestBundle/Document/RelationEmbedder.php
```diff
@@ -26,7 +26,7 @@
  *     attributes={
  *         "normalization_context"={"groups"={"barcelona"}},
  *         "denormalization_context"={"groups"={"chicago"}},
- *         "hydra_context"={"@type"="hydra:Operation", "hydra:title"="A custom operation", "returns"="xmls:string"}
+ *         "hydra_context"={"@var"="hydra:Operation", "hydra:title"="A custom operation", "returns"="xmls:string"}
```
alanpoulain, line 29: Please revert this.

tests/Fixtures/TestBundle/Entity/JsonldContextDummy.php
```diff
@@ -42,7 +42,7 @@ class JsonldContextDummy
      *     attributes={
      *         "jsonld_context"={
      *             "@id"="http://example.com/id",
-     *             "@type"="@id",
+     *             "@var"="@id",
```
alanpoulain, line 45: Please revert this.

tests/Fixtures/TestBundle/Entity/RelationEmbedder.php
```diff
@@ -26,7 +26,7 @@
  *     attributes={
  *         "normalization_context"={"groups"={"barcelona"}},
  *         "denormalization_context"={"groups"={"chicago"}},
- *         "hydra_context"={"@type"="hydra:Operation", "hydra:title"="A custom operation", "returns"="xmls:string"}
+ *         "hydra_context"={"@var"="hydra:Operation", "hydra:title"="A custom operation", "returns"="xmls:string"}
```
alanpoulain, line 29: Please revert this.

src/JsonSchema/SchemaFactory.php
dunglas: Ok, got it! Let's do this.

dunglas (approved):

dunglas:
Thanks @Tayfun74!

PR #3899, opened 2020-12-11
allow absence of token for sf authentication manager

cdaguerre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

Anonymous users don't exist anymore when using the Symfony Authentication manager (No token in token storage, which doesn't necessarily mean you're not behind a firewall)


dunglas (approved):

dunglas:
Thank you @cdaguerre!

PR #3886, opened 2020-12-05
Update php-cs-fixer configuration

drupol:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3884
| License       | MIT

PR #3903, opened 2020-12-14
Try to fix CI errors

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no (I hope)
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

About CS errors, I suspect a bug in PHP-CS-Fixer 2.17. I am waiting for their opinion: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5357.

~About PHPStan errors, I applied the fix from https://github.com/api-platform/core/pull/3886. I'm not sure if the new thrown exception might be a BC break.~

### DIFF
--- a/src/Util/Inflector.php
+++ b/src/Util/Inflector.php
@@ -27,7 +27,7 @@ use Doctrine\Inflector\InflectorFactory;
 final class Inflector
 {
     /**
-     * @var InflectorObject
+     * @var InflectorObject|null
      */
     private static $instance;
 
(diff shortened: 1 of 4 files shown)


src/Util/Inflector.php
```diff
@@ -42,14 +43,30 @@ private static function getInstance(): InflectorObject
...
-        return class_exists(InflectorFactory::class) ? self::getInstance()->tableize($word) : LegacyInflector::tableize($word);
+        if (class_exists(InflectorFactory::class)) {
+            return self::getInstance()->tableize($word);
+        }
+
+        if (class_exists(LegacyInflector::class)) {
+            return LegacyInflector::tableize($word);
+        }
+
+        throw new RuntimeException('Unable to find a proper Doctrine Inflector instance.');
```
dunglas, line 54: This can't happen because `doctrine/inflector` is a mandatory dependency.

src/Util/Inflector.php
```diff
@@ -42,14 +43,30 @@ private static function getInstance(): InflectorObject
      */
     public static function tableize(string $word): string
     {
-        return class_exists(InflectorFactory::class) ? self::getInstance()->tableize($word) : LegacyInflector::tableize($word);
```
dunglas, line 45: This looks like a PHPStan regression to me. What about just adding `// @phpstan-ignore-line`?

meyerbaptiste: I don't see it as a regression. As they say in the documentation: 
> PHPStan needs to be able to locate symbols (classes, functions, constants) used in the analyzed codebase.

So it does its job: the class does not exist, it reports the problem. Why not reverse the condition instead?
```php
return class_exists(LegacyInflector::class) ? LegacyInflector::tableize($word) : self::getInstance()->tableize($word);
```

dunglas: Yes reversing the condition is a better idea.

dunglas (approved):

dunglas:
Thanks @meyerbaptiste

PR #3915, opened 2020-12-18
ADR: Resource (re)-definition

soyuka:


alanpoulain:
It means these attributes would not be usable at all for GraphQL.

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,145 @@
...
+- GET /users
+- POST /users
+
+and three routes representing the `/users/{id}` resource:
+
+- GET /users/{id}
+- PUT /users/{id}
+- DELETE /users/{id}
+
+To keep our upgrade path, these would still be available on a single class:
```
dunglas, line 53:
```suggestion
To keep our upgrade path and for convenience, these would still be available on a single class:
```

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,145 @@
...
+<?php
+
+#[Resource]
+#[Get]
+class User {}
+```
+
+would declare:
+
+- GET /users/{id}
```
dunglas, line 74: In my opinion this should declare only `GET /users/{id}`, as it currently works, except we also have to explicitly declare collection operations (because currently it's not obvious that collection operations are automatically defined even if you define item operations).

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,145 @@
...
+<?php
+
+#[Resource]
+#[Get("/users")]
+class User {}
+```
+
+would declare:
+
+- GET /users/{id}
```
dunglas, line 100: It should declare only `GET /users` imho.

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,145 @@
...
+- PUT /users/{id}
+- DELETE /users/{id}
+- GET /users
+
+The `Resource` attributes would become optional and the only thing needed is to specify at least a verb and the IRI representing the Resource:
+
+```php
+<?php
+
+#[Get, "/users/{id}"]
```
dunglas, line 110:
```suggestion
#[Get("/users/{id}")]
```

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,145 @@
...
+#[Get, "/companies/{companyId}/users/{id}"]
```
dunglas, line 111:
```suggestion
#[Get("/companies/{companyId}/users/{id})]
```

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,145 @@
...
+
+#[Get("/companies/{companyId}/users/{id}", [identifiers=["companyId" => [Company::class, "id"], "id" => [User::class, "id"]]])]
+class User {
+  #[ApiProperty(identifier=true)] // TODO: rename Property ?
+  public $id;
+  public Company $company;
+}
+```
+
+Overriding serialization groups on a resource would require the `Resource` attribute:
```
dunglas, line 131: Maybe mention that it can be done directly on the `Get` attribute too?

dunglas (changes requested):

dunglas:
@alanpoulain, yes we'll have to introduce attributes for GraphQL too, such as `#[Query]` and `#[Mutation]`.

docs/adr/0002-resource-definition.md
```diff
@@ -0,0 +1,149 @@
...
+
+- GET /users/{id}
+
+For collections:
+
+```php
+<?php
+
+#[Resource]
+#[CGet]
```
stloyd, line 82: What about going a bit more descriptive way: `#[List]` ?

soyuka: List would only work for `get`, here collection would be used to declare routes without an identifier for example `POST /users` and `GET /users`.

stloyd: Hmmm I know it's not common, but I would personally do it like:
```php
#[List] <-- GET /users
#[Create] <-- POST /users

**Comment by Richard87** [comment]
As a user, I completely agree with @stloyd above.


I have been using API-Platform some time, and are migrating our legacy system to use more and more Api-Platform based operations/actions. We use a lot of custom operations/actions and it's frustrating with unclear documentation... Because of this I have been thinking a lot about what I would rather see... Also I would love to help with coding this up, but I have no idea where to start...

Also, maybe this is the completely wrong place to bring it up, if so, sorry!

- `#[Action] is the same as a custom operation (I feel like it sounds better :D) 
  - Defaults to a POST operation
  - Could be either messenger based, or connected to a custom controller
  - Messenger based would not include the object id in the url
  - When using a controller on a Class level:
    - the `receive=false`can be specified to not use ID in the url, or include the object in `__invoke`
    - Else the argument can be anything, and matching on type.
    - In case of ambiguity like `AddFriendController::__invoke(User $data, User $friend, OptionsDTO $options): User` 
      - Prefer User $user, but also allow User $data, if neither lets throw an error :) 
      - This would become:
         ```http
         POST /api/users/{id}/add_friend_controller 

        {
              "friend": "/api/users/123",
              "options": {...}
        }
         ```
        (You could use `name= "AddFriend"` or `path="/api/users/{id}/add_friend"` to change uri)
- I would love to use `#[List]` on methods for custom filtering
  - List always return a List of objects, but uses get Get method by default, and input as query parameters
  - If List method/controller specify a \Traversable output, query parameters are available, `output` would be inferred from PhpDocs/Signature (Maybe a MessageListDTO), or would be specified as `output=Message::class`.
  - The IRI of items returned in a list, should default to the "normal" iri, like `GET /api/users/{id}/messages` should preferably(by me at least) be `/api/messages/ {id}` if Message either has a `#[Resource]` or `#[Get] attribute`, or optionally `subresourceIri=true`(?) could change it to `/api/users/{id}/messages/{id}`?


I dream about a day where API-Platform is as easy to use as Symfony's auto-wiring dependency injection :D 

For a Class-level `#[Action]`, if no controller is specified, it must be a message, so `messenger=true` is not required and should be removed :) 
I want to always use input, and output, _except for when a controller or method is specified, where we can infer it from the signature._
So in the end, the `#[Action]` attribute on a Class could look like one of the following:
```php
#[Action(input=ResetPasswordMessage::class)] // No output
#[Action(controller=ResetPasswordController::class)] // input/output depending on `__invoke` signature
```
(It would be amazing in the future, if we could check if input was invokable, and if so drop the whole controller argument!)

```php

#[List(security="something"] // GET /api/users
#[Get] // GET /api/users/{id}
#[Create] // POST /api/users
#[Delete] // DELETE /api/users/{id}
#[Update] // PATCH/PUT /api/users/{id} (content type specify strategy)
#[Action(input=ResetPassword::class, output=User::class)] // POST /api/users/reset_password 
#[Action(security="is_granted('ROLE_ADMIN'),controller=Invite::class)] // POST /api/users/{id}/invite
#[List(name="important_messages",controller=FetchImportantMessages::class, output=Message::class)] // GET /api/users{id}/important_messages
class User {

    #[List] // GET /api/users/{id}/messages
    /** @var ArrayCollection<int, Message>|PersistentCollection */
    public $messages;

    #[List] // GET /api/users/{id}/message_senders?only_sent=true/false
    /** @var ArrayCollection<int, User> */
    public function getMessageSenders(bool $onlySent = false): ArrayCollection {
        return $this->messages->map(fn(Message $m)=> $m->getSender());
    }

    #[Action] // POST /api/users/{id}/send_message
    public function sendMessage(NewMessageDTO $dto): Message {
        $newMessage = Message::create($dto);
        $this->messages->add($newMessage);
        return $newMessage;
    }
} 

class FetchImportantMessages {
    __invoke(User $user, int $prioirity): Collection {...}
}
```

Sorry for writing so much, and thank you to everybody taking the time to read and listening!

soyuka:
Merging this feel free to open a new PR for edits.

PR #3921, opened 2020-12-28
fix: mercure multiple topic publishing

divine:
| Q             | A
| ------------- | ---
| Bug fix?      | yes  (2.6)
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

Currently, it's not possible to publish as multiple topics (arrays) via Mercure options because it's strictly checking for string type:

Encountered as `Argument 1 passed to ApiPlatform\Core\Bridge\Doctrine\EventListener\PublishMercureUpdatesListener::buildUpdate() must be of the type string, array given, called in /srv/api/vendor/api-platform/core/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php on line 22`

Related https://github.com/api-platform/core/pull/3728, however that PR went silent, so I went ahead  and added test.

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -255,6 +255,9 @@ final class PublishMercureUpdatesListener
         return $updates;
     }
 
+    /**
+     * @var string|string[] $iri
+     */
     private function buildUpdate($iri, string $data, array $options): Update
     {
         if (method_exists(Update::class, 'isPrivate')) {
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -255,7 +255,7 @@ final class PublishMercureUpdatesListener
         return $updates;
     }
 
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -255,7 +255,7 @@ private function getGraphQlSubscriptionUpdates($object, array $options, string $
         return $updates;
     }
 
-    private function buildUpdate(string $iri, string $data, array $options): Update
+    private function buildUpdate($iri, string $data, array $options): Update
```
dunglas, line 258:
```suggestion
    /**
     * @var string|string[] $iri
     */
    private function buildUpdate($iri, string $data, array $options): Update
```

dunglas (approved):
Thank you very much! Could you just document the accepted types?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
divine: Done. Thank you for your review!

dunglas:
Thanks @divine!

Issue #2400, opened 2018-12-19
[RFC] Filter composition


teohhanhui:
Instead of giving the client the ability to construct arbitrary queries / filtering (something better served by existing standards / specifications, e.g. OData and GraphQL), we could allow predefined composite filters in the API, each composed of one or more filter primitives.

For example:

`/users`
`q`: `match('username', 'ipartial') or match('email', 'ipartial') or match('firstName', 'iword_start') or match('lastName', 'iword_start')`

`/products`
`q`: `match('sku', 'iexact') or match('name', 'iword_start') or match('description', 'iword_start')`

This also supports the aliasing use case where you could easily define multiple filter parameters for the same property, e.g.:

`/users`
`email`: `match('email', 'iexact')`
`email_partial`: `match('email', 'ipartial')`

`/products`
`sku`: `match('sku', 'iexact')`
`sku_partial`: `match('sku', 'ipartial')`

Of course, this also forces us to have decoupled code between the declaration of filter parameters (currently the `$properties` argument) and the actual filtering (currently the `filterProperty` function). It'd be a good opportunity to get rid of inheritance of filter classes (mark as deprecated; to be removed in API Platform 3.0).

Related:
- https://github.com/api-platform/core/issues/1832
- https://github.com/api-platform/core/pull/2055

dunglas:
There are 2 other popular syntax:

* Postgrest syntax: http://postgrest.org/en/v5.1/api.html
* Loopback: https://loopback.io/doc/en/lb2/Where-filter.html

Both already have a lot of compatible client tools (including React Admin providers IIRC) and would be fine to me.

dunglas:
I actually like the loopback syntax because it looks very PHPish.

ragboyjr:
regarding query syntax, https://github.com/krakphp/aql is a solution I built that basically allows sql type expressions in the API, but it's parsed and validated and can contain certain constraints, it might be something of use here for allowing more complex querying/flexibility

teohhanhui:
I'm strongly -1 for giving too much control to the client (there's GraphQL for that lol, at least it's better than inventing yet another format / syntax).

Anyway, it's irrelevant to this RFC, which is about filter composition on the server side only.

ragboyjr:
ah, ok, cool beans, y, this would be awesome, part of me wonders if maybe just making the custom filter API easier would fix this. Like, if making filters was just more composable already, then we could easily have filter factories that can be used to create filters instead of building our own expression parser or something on the backend.

[...]

PR #3924, opened 2020-12-29
[WIP] [Feature] OrSearchFilter - a new filter to search for multiple fields

silverbackdan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | TBC (api-platform/docs#...)

This is a WIP to implement a search filter which will allow a user to query multiple fields of an entity using the syntax `or[field1,field2]=searchValue`

To do:
- Tests
- Docs PR
- RFC on mapping fields and values.

The RFC part may end up as a separate PR. I my application I wanted to query a computed field and then map this back to a database query. The computed field is `displayCurationStatus` and the mapped database field is `curationStatus`. Internally the project I was working on needed to store more status levels, however for the external application this is reduced to just a few. For example if `displayCurationStatus` is queried with the value `published`, we wanted to be returning both statuses `ready` and `published` from the `curationStatus` field of the resource. I don't think how I've adjusted this in my first attempt of this PR is flexible or configurable without lots of effort.

I wanted to submit now as I find limited time and shows that the work has been started and a place for discussion.

### DIFF
--- a/src/Bridge/Doctrine/Common/Filter/SearchFilterTrait.php
+++ b/src/Bridge/Doctrine/Common/Filter/SearchFilterTrait.php
@@ -167,4 +167,15 @@ trait SearchFilterTrait
 
         return true;
     }
+
+    /**
+     * This can be overwritten by a user extending a search filter to map computed properties and extend values that will be matched.
+     *
+     * @param string $property
+     * @param mixed $value
+     */
+    protected function mapCustomPropertyAndValue(string $property, $value): array
+    {
+        return [$property, $value];
+    }
 }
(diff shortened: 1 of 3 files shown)


alanpoulain:
Hello. Thank you for the PR but did you see https://github.com/api-platform/core/issues/2400? The aim would be to have a general filter not just and/or.

silverbackdan:
I had not seen that, sorry for my oversight - looks like it'll be great!

silverbackdan:
While I've closed this because it is an alternative method to the linked issue, the issue appears to be stale in regards to the really nice flexible approach due to Doctrine not being 'composable'. This may be a useful feature while the issues there are resolved (?)

PR #3926, opened 2020-12-30
Fix the ci

jockos:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Just fix the CI tests

PR #3927, opened 2020-12-31
OpenAPI: fix #3920 security scheme

### LINKED ISSUE #3920: [2.6-beta1] Issue with OpenApi Spec securitySchemes key
**API Platform version(s) affected**: 2.6.0-beta1

**Description**  

The changes in the OpenApi integration now generate incorrect `components.securitySchemes` docs.

**How to reproduce**  

Configure APIP as documented in the docs (https://api-platform.com/docs/core/jwt/#configuring-api-platform):

```yaml
    swagger:
        versions:
            - 3
        api_keys:
            apiKey:
                name: Authorization
                type: header
```

The spec generated will look something like this:

```yaml
             securitySchemes:
               apiKey:
                 type: apiKey
                 description: Value for the Authorization header parameter.
                 name: Authorization
                 in: header
                 scheme: bearer
```

Copy the generated OpenApi Spec and validate it on https://editor.swagger.io/.
It will tell you something along those lines:

```
Structural error at components.securitySchemes.apiKey
should NOT have additional properties
additionalProperty: scheme
```

**Possible Solution**  

Imho `scheme` needs to be omitted but it really depends on the `type` used so I'm not sure what the idea of @soyuka was here, hence the bug report :)
Also see https://swagger.io/docs/specification/authentication/#securitySchemes.

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixes #3920
| License       | MIT
| Doc PR        |  na.

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -446,7 +446,7 @@ final class OpenApiFactory implements OpenApiFactoryInterface
 
         foreach ($this->openApiOptions->getApiKeys() as $key => $apiKey) {
             $description = sprintf('Value for the %s %s parameter.', $apiKey['name'], $apiKey['type']);
-            $securitySchemes[$key] = new Model\SecurityScheme('apiKey', $description, $apiKey['name'], $apiKey['type'], 'bearer');
+            $securitySchemes[$key] = new Model\SecurityScheme('apiKey', $description, $apiKey['name'], $apiKey['type']);
         }
 
         return $securitySchemes;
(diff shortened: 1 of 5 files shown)


src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -256,7 +256,7 @@ private function getGraphQlSubscriptionUpdates($object, array $options, string $
     }
 
     /**
-     * @var string|string[] $iri
```
dunglas, line 259: To revert, it's fixed in #3926

dunglas (changes requested):

PR #3932, opened 2021-01-04
Restore composite identifier denormalization

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fi
| License       | MIT
| Doc PR        | 

Fixes a regression in 2.6 where composite identifiers are not denormalized correctly anymore
There was an existing test that was supposed to avoid such regression but it was marked as skipped during a refactoring unfortunately.
https://github.com/api-platform/core/blob/ed59e8df62a50e57eb2bbb3b8cdaefcfd241dcb5/tests/Identifier/IdentifierConverterTest.php#L33-L59

Not sure what would be the best place to test this now if needed, ping @soyuka

### DIFF
--- a/src/DataProvider/OperationDataProviderTrait.php
+++ b/src/DataProvider/OperationDataProviderTrait.php
@@ -106,7 +106,7 @@ trait OperationDataProviderTrait
                         throw new InvalidIdentifierException(sprintf('Expected %d identifiers, got %d', $identifiersNumber, $currentIdentifiersNumber));
                     }
 
-                    return $identifiers;
+                    return $this->identifierConverter->convert($identifiers, $attributes['resource_class']);
                 }
 
                 // TODO: Subresources tuple may have a third item representing if it is a "collection", this behavior will be removed in 3.0
--- a/tests/EventListener/ReadListenerTest.php
+++ b/tests/EventListener/ReadListenerTest.php
@@ -382,4 +382,27 @@ class ReadListenerTest extends TestCase
         $listener = new ReadListener($collectionDataProvider->reveal(), $itemDataProvider->reveal(), $subresourceDataProvider->reveal(), null, $identifierConverter->reveal());
         $listener->onKernelRequest($event->reveal());
     }
+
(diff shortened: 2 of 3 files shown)


src/DataProvider/OperationDataProviderTrait.php
```diff
@@ -106,7 +106,7 @@ private function extractIdentifiers(array $parameters, array $attributes)
                         throw new InvalidIdentifierException(sprintf('Expected %d identifiers, got %d', $identifiersNumber, $currentIdentifiersNumber));
                     }
 
-                    return $identifiers;
```
soyuka, line 109: This is the expected behavior now, the composite identifier parsing is done here: https://github.com/api-platform/core/blob/master/src/DataProvider/OperationDataProviderTrait.php#L102-L110 

I haven't skipped any functional test (behat), it seems weird to me that the behavior changed as it shouldn't. May you give me a reproducible case?

antograssiot: The code you mentionned is actually the one that I modified @soyuka, you might have read to fast

parsing is still ok but if your composite identifer is composed of an integer property and a string property, you still need to denormalize them.


Example would be
```php
@ApiResource
class Foo
{
    public int $identifer1;
    public string $identifier2;
}
```

when requesting `/foos/identifer1=1;identifier2=test`, without the patch you would get an identifier array like this
```php
[
    "identifer1" => "1",
    "identifer2" => "test",
]
```

where I expect 
```php
[
    "identifer1" => 1,
    "identifer2" => "test",
]
```

antograssiot: that's what was tested with integer and datetime line 57  and 58 of the test mentioned in the PR description

antograssiot: you can ping me on slack if you have time

soyuka:
Tyvm @antograssiot !

PR #3938, opened 2021-01-04
Fix the registration of the "test.api_platform.client" service

### LINKED ISSUE #3928: ApiTestCase Client creation fails with false bundle.php order
**API Platform version(s) affected**: 2.5.9

**Description**  
Creating of API Platform test client fails with `LogicException : You cannot create the client used in functional tests if the "framework.test" config is not set to true.`.

The` symfony/browser`-kit and `symfony/http-client` packages are installed (obviously ... because otherwise the Logic exception would be a different one). The `test` key in `config/packages/test/framework.yaml` is set to true.

`php bin/console -e dev debug:config framework` showed that the test key is set to true in the test environment (which use used by force from PHPUnit).

I found a "solution" but this is ridiculous to me. The problem is "solved" if the bundle.php file looks like this:
```php
<?php

return [
    Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true],
    ApiPlatform\Core\Bridge\Symfony\Bundle\ApiPlatformBundle::class => ['all' => true],
    ...
];
```
but not if it looks like this:
```php
<?php

return [
    ApiPlatform\Core\Bridge\Symfony\Bundle\ApiPlatformBundle::class => ['all' => true],
    Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true],
    ...
];
```

In the `ApiPlatformExtension.php` file is checked if a parameter is set in the container:
```php
        if ($container->hasParameter('test.client.parameters')) {
            $loader->load('test.xml');

            if (!class_exists(AbstractBrowser::class) || !trait_exists(HttpClientTrait::class)) {
                $container->removeDefinition('test.api_platform.client');
            }
        }
```
The `'test.client.parameters'` parameter seems to be only set by the symfony framework bundle an ONLY if this is loaded before the api-platform bundle (at least in my test environment). I honestly had no clue that the order of the bundles.php entries matters at all. As statet here [Sort classes on bundles.php #275](https://github.com/symfony/flex/issues/275), the order can matter (but should not) ... but that's not clearly communicated to my mind (symfony documentation improvement might be necessary as well).

If the order really was the only problem here, there should be some kind of exception thrown and message printed to hint to this because from `LogicException : You cannot create the client used in functional tests if the "framework.test" config is not set to true.` to "just resort the bundles, took me some hours ... frustrating hours.

**How to reproduce** 
The but should be reproducable by just writing a ApiTestCase test with the `bundle.php` file mentioned above.
But here are some configurations I am using right now:

composer.json:
<details>

```json
  {
      "require": {
          ...
          "symfony/asset": "5.2.*",
          "symfony/cache": "^5.2",
          "symfony/console": "^5.2",
          "symfony/dependency-injection": "^5.2",
          "symfony/doctrine-bridge": "^5.2",
          "symfony/dotenv": "^5.2",
          "symfony/error-handler": "^5.2",
          "symfony/event-dispatcher": "^5.2",
          "symfony/event-dispatcher-contracts": "^2.2",
          "symfony/flex": "^1.9",
          "symfony/framework-bundle": "^5.2",
          "symfony/http-client-contracts": "^2.3",
          "symfony/http-foundation": "^5.2",
          "symfony/http-kernel": "^5.2",
          "symfony/routing": "^5.2",
          "symfony/serializer": "5.2.*",
          "symfony/skeleton": "^5.2",
          "symfony/twig-bundle": "^5.2",
          "symfony/uid": "^5.2",
          "symfony/validator": "^5.2",
          "symfony/yaml": "^5.2",
          ...
      },
      "require-dev": {
          ...
          "symfony/browser-kit": "^5.2",
          "symfony/debug-bundle": "5.2.*",
          "symfony/http-client": "^5.2",
          "symfony/maker-bundle": "^1.26",
          "symfony/phpunit-bridge": "^5.2",
          "symfony/var-dumper": "5.2.*"
      },
      "config": {
          "optimize-autoloader": true,
          "preferred-install": {
              "*": "dist"
          },
          "sort-packages": true
      },
      "autoload": {
          "psr-4": {
              "": "src/"
          }
      },
      "autoload-dev": {
          "psr-4": {
              "Tests\\": "tests/"
          }
      },
  }
```
</details>

phpunit.xml.dist:
<details>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" backupGlobals="false" colors="true" bootstrap="tests/bootstrap.php">
  <coverage processUncoveredFiles="true">
    <include>
      <directory suffix=".php">src</directory>
    </include>
  </coverage>
  <php>
    <ini name="error_reporting" value="-1"/>
    <ini name="memory_limit" value="-1" />
    <server name="APP_ENV" value="test" force="true"/>
    <server name="KERNEL_CLASS" value="\Kernel"/>
    <server name="SHELL_VERBOSITY" value="-1"/>
    <server name="SYMFONY_PHPUNIT_REMOVE" value=""/>
    <server name="SYMFONY_PHPUNIT_VERSION" value="9.5"/>
  </php>
  <testsuites>
    <testsuite name="Project Test Suite">
      <directory>tests</directory>
    </testsuite>
  </testsuites>
  <listeners>
    <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
  </listeners>
</phpunit>
```

The `framework.yaml` files are the default ones which are also used by the api-platform demo project.

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3928 
| License       | MIT
| Doc PR        | n/a


src/Bridge/Symfony/Bundle/ApiPlatformBundle.php
```diff
@@ -51,5 +52,6 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new GraphQlMutationResolverPass());
         $container->addCompilerPass(new DeprecateMercurePublisherPass());
         $container->addCompilerPass(new MetadataAwareNameConverterPass());
+        $container->addCompilerPass(new TestPass());
```
alexislefebvre, line 55: I suggest to rename it to `TestClientPass`.

dunglas (approved):
Can you just rename `TestPass` in `TestClientPass` as suggested?

dunglas:
Thanks @meyerbaptiste!

Moh1267:
@dunglas @meyerbaptiste.
I have the 2.6 version and I still meet this error (LogicException: You cannot create the client used in functional tests if the "framework.test" config is not set to true.) Any idea about this please? 
thank you in advance.

PR #3939, opened 2021-01-04
Make sure both, description and summary are generated

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Up until now, `summary` has also been autogenerated (see e.g. https://github.com/api-platform/core/blob/2.5/src/Swagger/Serializer/DocumentationNormalizer.php#L327 or just search for `summary` there :)). It's also the one preferred by some JS tools like redoc for the navigation. I think there's nothing wrong in providing a good default for both, `description` as well as `summary`. You can always override them if you want to.


soyuka:
Thanks @Toflar !

PR #3940, opened 2021-01-04
Add mongodb date_immutable filter support

CheapHasz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Test pass? | kinda, (the tests I added pass, but I wasn't able to start the full mongob_phpunit unit test suite, it was failing on a Prophecy dependency. running only the tests/Bridge/Doctrine/MongoDbOdm/PropertyInfo/DoctrineExtractorTest.php does pass though)

DateFilter didn't work on date_immutable ODM/Field


alanpoulain:
Hello. Thanks for the PR. You should check the CS and PHPStan.
I don't remember why it has not been done before. It is a recent addition in MongoDB ODM?

CheapHasz:
I fixed the cs, which seems to have fixed the phpstan too.

As for the DateImmutableType, it seems to have been added to doctrine in 2019 https://github.com/doctrine/mongodb-odm/pull/2118, which I wouldn't exactly call recent

alanpoulain:
But released in the end of May 2020 :slightly_smiling_face: https://github.com/doctrine/mongodb-odm/releases/tag/2.1.0

alanpoulain:
Anyway, we should have a conflict with MongoDB ODM < 2.1.

CheapHasz:
Is that what makes the `PHP 7.4 lowest` tests fail ? Does it mean that the DateTimeImmutable support will be postponed until the MongoDB ODM requirement is bumped ?

alanpoulain:
That's right. If the MongoDB part was not experimental, we would have to introduce a non-breaking change, but since it's not the case, I think we can just bump the conflict.

CheapHasz:
I'm not sure to understand your reply, but let me know if I need to do anything more to help that fix get merged

alanpoulain:
I've added some commits to fix it. I think it's OK now. For the merge, we need to wait for the stable release of 2.6.

alanpoulain (approved):

CheapHasz:
Ok thanks

kralos:
@alanpoulain what's holding this one up? last from you was awaiting 2.6 stable however we're already up to 2.6.2

alanpoulain:
Thank you @CheapHasz.

kralos:
@alanpoulain I'm assuming main = `2.7`? are `MINOR` releases on api-platform time based? Should we add something more helpful explaining the schedule on https://api-platform.com/docs/extra/releases/#the-release-process

We don't really know what "when they are ready" means. Does it mean there is no schedule? Based on historical `MINOR` tags we're guessing it's roughly yearly?

With regards to this particular merge, we are now trying to decide if we should wait for `2.7` or revert our application to use `DateTime` / `date` instead of `DateTimeInterface` / `date_immutable`...

alanpoulain:
Yes, `main` will be the `2.7` version.

Yes, there is no schedule. It depends of the features we have, the time it needs to release a new version, the timing of the release (for instance with a new version of Symfony or PHP), etc.

I'm not sure what's the best choice in your case. The `2.6` version has just been released, so the `2.7` version will take time to be out. You can also use directly the `main` branch in the meantime.

PR #3943, opened 2021-01-05
[RFC] Encoding colons manually in identifiers

silverbackdan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | https://github.com/api-platform/core/issues
| License       | MIT
| Doc PR        | n/a

_There is another outstanding issue I've discovered which should probably be addressed on to of this one. If a resource's ID requirements are changed (e.g. to `.*`), when resources which relate to that entity are created, the reference to it does not inherit those requirements and therefore need to be added manually as I have done in `tests/Fixtures/TestBundle/Entity/RelatedToUrlEncodedId`_

This PR addresses an issue where if you are referencing another resource which has an ID containing a colon and what PHP will determine to be an invalid port number, the API will return 400 Invalid IRI.

In the issue there was talk of a few options:

1. *Identifier Denormalizer*: This would apply to outputting the ID to the user I believe. I tried this approach first, along with introducing an IdentifierNormalizer insteadd of denormalizing. It seemed very involved for the specific problem here.
2. *Prepending the host in the Router*: It seemed like it'd work well, but it broke other versions of this test which should allow the ID to be encoded as well if the user wants to to prevent BC breaks.

I also thought about doing `urlencode` on the iri - first decoding and then encoding it, but that would encode the slashes in the first part of the IRI as well. Currently this solution does allow the crazy scenario of there being slashes and dots in the identifier if required.

It seems a bit strange this solution of just replacing the colon, but I think the issue is very specific to this single character and how PHP's `parse_url` is processing the relative path.

All my tests pass at the moment, but as it's late, I haven't waited to see if the CI tests will all pass.

### DIFF
--- a/features/main/url_encoded_id.feature
+++ b/features/main/url_encoded_id.feature
@@ -3,26 +3,27 @@ Feature: Allowing resource identifiers with characters that should be URL encode
   As a client software developer
   I need to be able to set and retrieve these resources with the URL encoded ID
 
+  # Symfony\Component\Routing\Generator\UrlGenerator::generate uses rawurlencode so @id/iri will be encoded.
   @createSchema
   Scenario Outline: Get a resource whether or not the id is URL encoded
     Given there is a UrlEncodedId resource
     And I add "Content-Type" header equal to "application/ld+json"
-    When I send a "GET" request to "<url>"
+    When I send a "GET" request to "/url_encoded_ids/<id>"
     Then the response status code should be 200
     And the JSON should be equal to:
     """
     {
         "@context": "/contexts/UrlEncodedId",
-        "@id": "/url_encoded_ids/%25encode:001",
+        "@id": "/url_encoded_ids/%25./encode:001",
         "@type": "UrlEncodedId",
-        "id": "%encode:001"
+        "id": "%./encode:001"
     }
     """
     Examples:
-      | url                               |
-      | /url_encoded_ids/%encode:001      |
-      | /url_encoded_ids/%25encode%3A001  |
-      | /url_encoded_ids/%encode%3A001    |
+      | id                     |
+      | %25%2E%2Fencode%3A001  |
+      | %./encode%3A001        |
+      | %./encode:001          |
 
   @createSchema
   Scenario Outline: Add a resource relation which contains special URL characters
@@ -36,7 +37,7 @@ Feature: Allowing resource identifiers with characters that should be URL encode
     """
     Then the response status code should be 201
     Examples:
-      | id                   |
-      | %25encode%3A001      |
-      | %encode:001          |
-      | %encode%3A001         |
+      | id                     |
+      | %25%2E%2Fencode%3A001  |
+      | %./encode%3A001        |
+      | %./encode:001          |
--- a/features/main/url_encoded_id.feature
+++ b/features/main/url_encoded_id.feature
@@ -13,14 +13,30 @@ Feature: Allowing resource identifiers with characters that should be URL encode
     """
     {
         "@context": "/contexts/UrlEncodedId",
-        "@id": "/url_encoded_ids/%25encode:id",
+        "@id": "/url_encoded_ids/%25encode:001",
         "@type": "UrlEncodedId",
-        "id": "%encode:id"
+        "id": "%encode:001"
     }
     """
     Examples:
-      | url                              |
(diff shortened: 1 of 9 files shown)


src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -250,6 +252,7 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
             [$operation['method']],
             $operation['condition'] ?? ''
         );
+        $name = RouteNameGenerator::generate($operationName, $resourceShortName, $operationType);
```
soyuka, line 255: is this really needed?

tests/Fixtures/TestBundle/Document/RelatedToUrlEncodedId.php
```diff
@@ -0,0 +1,81 @@
...
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Document;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * Related To URL Encoded ID represent an association for encoding special characters
+ * TODO: We should not really need to have urlEncodedIdResource should we? Determine that the related resource has this requirement and then apply it as a requirement here would be preferable. Not sure about composite IDs though.
```
soyuka, line 22: Maybe if you use a `.` somewhere as symfony has a default regex for `.{_format}` IIRC

src/Bridge/Symfony/Routing/Router.php
```diff
@@ -78,7 +78,7 @@ public function match($pathInfo)
             $pathInfo = substr($pathInfo, \strlen($baseUrl));
         }
 
-        $request = Request::create($pathInfo, 'GET', [], [], [], ['HTTP_HOST' => $baseContext->getHost()]);
+        $request = Request::create(str_replace(':', '%3A', $pathInfo), 'GET', [], [], [], ['HTTP_HOST' => $baseContext->getHost()]);
```
soyuka, line 81: I really think we should use an identifier denormalizer

soyuka (commented):
Have you tried encoding in an identifier denormalizer? Wasn't that good enough?

src/Bridge/Symfony/Routing/ApiLoader.php
silverbackdan: No sorry, was there from previous workings related to the 'other outstanding issue' I mentioned - I forgot to remove it.

silverbackdan:
I'm pretty sure the IdentifierDenormalizer was happening after the router is trying to match so wouldn't be applied here. I'll do some more checks to be sure. Thanks for the comments.

tests/Fixtures/TestBundle/Document/RelatedToUrlEncodedId.php
silverbackdan: That's right, again part of the 'other issue' I mentioned. Just left this to do in there. But API Platform supports modifying the route parameter requirement on a resource so thought it should naturally cascade to resources that have a relation to it like here instead of this manual configuration.

alanpoulain:
As @dunglas mentioned, we should avoid using a replacement in the URL. Is there really no other way?

silverbackdan:
I think all the solutions I had break tests - I'm sure there will be I just haven't had time to review this again yet - so sorry!

features/main/url_encoded_id.feature
```diff
@@ -3,24 +3,41 @@ Feature: Allowing resource identifiers with characters that should be URL encode
...
     }
     """
     Examples:
-      | url                              |
-      | /url_encoded_ids/%encode:id      |
-      | /url_encoded_ids/%25encode%3Aid  |
-      | /url_encoded_ids/%25encode:id    |
-      | /url_encoded_ids/%encode%3Aid    |
+      | id                     |
+      | %25%2E%2Fencode%3A001  |
```
soyuka, line 24: @alanpoulain  this is the only identifier that feels right to me as this is the Resource identifier. It should not have non-urlencoded parts.

PR #3947, opened 2021-01-07
[Hydra] Use validation groups for Hydra documentation

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Same PR as https://github.com/api-platform/core/pull/3806, but for Hydra.

The `hydra:required` property was always set to `false` if validation groups were used for the property in the `NotBlank` constraint.
It's because the property metadata are filled with the `ValidatorPropertyMetadataFactory`:
https://github.com/api-platform/core/blob/8e58a986970dc3abe1fa994f6849631ad77e09a1/src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php#L97-L99

In Hydra, since the properties are decoupled from the operations, the code assumes that the "required" information is taken in the POST operation of the resource.

### DIFF
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -201,9 +201,13 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
         /** @var string[] $classes */
         $classes = array_keys($classes);
         $properties = [];
+        $options = [];
+        if ($validationGroups = $this->getValidationGroups($resourceMetadata)) {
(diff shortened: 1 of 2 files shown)


alanpoulain:
PHP 8 failure is not related.

alanpoulain:
WDYT @api-platform/core-team? For the part:
> In Hydra, since the properties are decoupled from the operations, the code assumes that the "required" information is taken in the POST operation of the resource.

PR #3956, opened 2021-01-13
fix: add prefix to subresources and operations

### LINKED ISSUE #3906: [OpenAPI] v2.6.0-alpha Subresources - paths are missing prefix in OpenAPI
**API Platform version(s) affected**: v2.6.0alpha

**Description**  
Generated documentation is not producing route paths including the configured route prefix.

`php bin/console debug:router` is generating correct paths, and is functioning as expected, so this problem appears isolated to the OpenAPI generation.

**How to reproduce**  
1. config/routes/api_platform.yaml
```
api_platform:
  resource: .
  type: api_platform
  prefix: /example
```

2. Create a resource, and subresource

3. Visit your local documentation. Subresources will not be prefixed with `/example`

**Possible Solution**  
The culprit looks to be within https://github.com/api-platform/core/blob/76ca5351337f9a817de6cfdb6f1b395f4dbd94ad/src/OpenApi/Factory/OpenApiFactory.php#L282

the `path` key appears to be present when the operation type is a subcollection, but not item/collection which is why this issue is limited.

 If the operationPathResolver was used, the route would contain the prefix.

GrimpEreau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3906
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -293,7 +293,7 @@ final class OpenApiFactory implements OpenApiFactoryInterface
      */
     private function getPath(string $resourceShortName, string $operationName, array $operation, string $operationType): string
     {
-        $path = $operation['path'] ?? $this->operationPathResolver->resolveOperationPath($resourceShortName, $operation, $operationType, $operationName);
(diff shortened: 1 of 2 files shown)


soyuka:
thanks @GrimpEreau !

PR #3957, opened 2021-01-14
Fix:  #3945

### LINKED ISSUE #3945: @ApiProperty() annotation is ignored for objects
**API Platform version(s) affected**: 2.5.9

**Description**  

`@ApiProperty()` annotation is ignored for objects

**How to reproduce**  

For a property that has a type of another class, we would like to override OpenAPI documentation [by adding `@ApiProperty` annotation](https://api-platform.com/docs/core/swagger/#using-the-openapi-and-swagger-contexts):

```php
/**
 * @ApiProperty(
 *     attributes={
 *         "openapi_context"={
 *             "type"="string",
 *             "enum"=AddressType::ALL,
 *             "example"=AddressType::MAILING
 *         }
 *     }
 * )
 */
private AddressType $addressType;
```

**In OpenAPI documentation request/response example:**

Expected Result:

```json
{
  "addressType": "mailing"
}
```

Actual Result:

```json
{
  "addressType": {}
}
```

**In generate OpenAPI/Swagger document:**

Expected Result:

```json
{
  "properties": {
    "addressType": {
      "type": "string",
      "enum": [
        "billing",
        "mailing"
      ],
      "example": "mailing"
    }
  }
}
```

Actual Result:

```json
{
  "properties": {
    "addressType": {
      "$ref": "#/components/schemas/AddressType-create_address",
      "type": "string",
      "enum": [
        "billing",
        "mailing"
      ],
      "example": "mailing"
    }
  }
}
```

(!) Note the `$ref` key added - this is an actual bug, because `$ref` key takes precedence and all other keys are ignored.

**Possible Solution**  

Solution if to NOT add `$ref` key for Property Schema when it already has other keys, here:

https://github.com/api-platform/core/blob/9e6ae813f7400681f8fb430af99f8b1ce587ec42/src/JsonSchema/SchemaFactory.php#L229

So, I would suggest the following diff:

```diff
- $propertySchema = new \ArrayObject($propertySchema + $valueSchema);
+ if (count($propertySchema) > 0 && array_key_exists('$ref', $valueSchema)) {
+    $propertySchema = new \ArrayObject($propertySchema);
+ } else {
+     $propertySchema = new \ArrayObject($propertySchema + $valueSchema);
+ }
```

Which means the following: if we have other keys in `$propertySchema`-  we don't add `$ref` from `$valueSchema` that will override everything.

For those, who already faced this issue, here is a workaround with decorating a service with this bug:

<details>

First, we need to create a decorator:

```php
<?php

declare(strict_types=1);

namespace App\JsonSchema;

use ApiPlatform\Core\JsonSchema\Schema;
use ApiPlatform\Core\JsonSchema\SchemaFactory as BaseSchemaFactory;
use ApiPlatform\Core\JsonSchema\SchemaFactoryInterface;
use function array_key_exists;
use function count;

class SchemaFactory implements SchemaFactoryInterface
{
    private BaseSchemaFactory $schemaFactory;

    public function __construct(BaseSchemaFactory $schemaFactory)
    {
        $this->schemaFactory = $schemaFactory;
    }

    public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
    {
        $schema = $this->schemaFactory->buildSchema(
            $className,
            $format,
            $type,
            $operationType,
            $operationName,
            $schema,
            $serializerContext,
            $forceCollection
        );

        // custom code to fix api-platform issue
        foreach ($schema->getDefinitions() as $definitionName => $definition) {
            if (array_key_exists('properties', $definition)) {
                foreach ($definition['properties'] as $propertyName => &$propertySchema) {
                    if (count($propertySchema) > 1 && array_key_exists('$ref', $propertySchema)) {
                        // remove incorrectly added $ref (schema) as we have custom definition - other keys
                        unset($propertySchema['$ref']);
                    }
                }
            }
        }

        return $schema;
    }

    public function addDistinctFormat(string $format): void
    {
        $this->schemaFactory->addDistinctFormat($format);
    }
}
```

and then register it in `services.yaml`:

```yaml
services:
    App\JsonSchema\SchemaFactory:
        decorates: 'api_platform.json_schema.schema_factory'
```
</details>

Zowac:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixes #3945  
| License       | MIT

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -226,7 +226,11 @@ final class SchemaFactory implements SchemaFactoryInterface
             $valueSchema = $this->typeFactory->getType(new Type($builtinType, $type->isNullable(), $className, $isCollection), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
         }
 
-        $propertySchema = new \ArrayObject($propertySchema + $valueSchema);
+        if (\count($propertySchema) > 0 && \array_key_exists('$ref', $valueSchema)) {
+            $propertySchema = new \ArrayObject($propertySchema);
+        } else {
+            $propertySchema = new \ArrayObject($propertySchema + $valueSchema);
+        }
         $schema->getDefinitions()[$definitionName]['properties'][$normalizedPropertyName] = $propertySchema;
     }
 
--- a/tests/Swagger/Serializer/DocumentationNormalizerV2Test.php
+++ b/tests/Swagger/Serializer/DocumentationNormalizerV2Test.php
@@ -2339,7 +2339,6 @@ class DocumentationNormalizerV2Test extends TestCase
                         ]),
                         'relatedDummy' => new \ArrayObject([
                             'description' => 'This is a related dummy \o/.',
-                            '$ref' => '#/definitions/'.$relatedDummyRef,
                         ]),
                     ],
                 ]),


soyuka (approved):

dunglas:
Thanks @Zowac

PR #3961, opened 2021-01-15
[OpenApi] Fix: add filters for subresources path

### LINKED ISSUE #3892: [OpenAPI] 2.6.0-alpha.1 Regression: Subresource operations show filters of the parent, not the child
**API Platform version(s) affected**: 2.6.0-alpha.1

**Description**  

Given the subresource route `/books/{id}/reviews`

- The **Book** is a single item resource, with the provided `id`
- A collection of **Review** entities is returned, which I may want to filter down.

On 2.5.x, the OpenAPI docs were showing the filters of the subresource on this route, correctly suggesting I can call `/books/123/reviews?stars=5`.

On 2.6.0-alpha.1 this behaviour has changed, and now it's showing the filters from the parent, incorrectly suggesting that I can call `/books/123/reviews?isbn=12345` (how can I filter down a single item which was already defined by its ID?), and not showing any filters from the child.

**How to reproduce**  

In my case, I just took a working 2.5 API, upgraded, and compared the `/docs.json?spec_version=3` output.

I will try and share a minimal repro for this soon.  But just wanted to get this behaviour noted for now.

**Additional Context**  

I tried to verify whether the parameters are respected when calling the endpoints, but unfortunately I'm getting 500 errors on any subresource endpoints.  That seems to be happening when it tries to serialize `hydra:search -> hydra:template` - at that point it chokes with `Serialization of 'Closure' is not allowed`.  Removing the filters didn't help.

For now I've reverted back to 2.5.  But when I have some time later I will try to dig into what's causing that one too.  I'm very keen to get onto 2.6 when it's ready - the new `identifiers` feature is great!

GrimpEreau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3892 
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/features/openapi/docs.feature
+++ b/features/openapi/docs.feature
@@ -87,7 +87,11 @@ Feature: Documentation support
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[4].required" should be false
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[4].schema.type" should be equal to "string"
 
-    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters" should have 5 elements
+    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[5].name" should be equal to "description"
+    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[5].in" should be equal to "query"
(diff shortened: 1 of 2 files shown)


soyuka (approved):

dunglas:
Thanks @GrimpEreau

PR #3503, opened 2020-04-20
Implement ApiProperty security attribute

GregoireHebert:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #773
| License       | MIT
| Doc PR        | api-platform/docs#1000

This PR continues https://github.com/api-platform/core/pull/3227

PR #3966, opened 2021-01-20
Also pass the object to the resource access checker in the serializer

Toflar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | sort of
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

In #3503 @GregoireHebert introduced support for the `security` attribute on an api property.
I found that this is very useful to me because I can get rid of my own normalizer. However, my permission system depends not only on roles but you can have different roles per organisation. This in turn means, it depends on the object.
The object itself is already passed on in e.g. the DenyAccessListener (https://github.com/api-platform/core/blob/master/src/Security/EventListener/DenyAccessListener.php#L100) so we should also have it here.

By introducing that, I was able to super conveniently restrict access to a certain property by registering my own custom expression language function. It now looks like this ❤️ : 

```php
    /**
     * @ApiProperty(security="user_has_role_for_organisation('ROLE_OWNER', object)")
     */
    private $stuff;
```

Regarding docs, I guess I could expand a bit on what Frédéric already wrote in https://github.com/api-platform/docs/pull/1000.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -398,7 +398,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $attribute, $options);
         $security = $propertyMetadata->getAttribute('security');
         if ($this->resourceAccessChecker && $security) {
-            return $this->resourceAccessChecker->isGranted($attribute, $security);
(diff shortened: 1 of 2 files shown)


alanpoulain (approved):
LGTM

chalasr (approved):

soyuka:
thanks @Toflar

PR #3967, opened 2021-01-21
fix: only display hydra:next when the item total is strictly greater …

jeremyteyssedre:
…than the number of items per page

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Tickets       | 
| License       | MIT
| Doc PR        | 

When the number of items per page is equal to the number of items present in the collection, the page found in the "hydra:next" field will return an empty collection. The "hydra:next" field should therefore only be displayed when the number of items present in the collection is strictly greater than the number of items per page.


dunglas:
Good catch, thanks @jeremyteyssedre!

clementtalleu:
When using [partial pagination](https://api-platform.com/docs/core/pagination/#partial-pagination-globally), this modification removes the "hydra:member" node from the api response.

PR #3968, opened 2021-01-21
Fix: allow to disable OpenAPI, Swagger UI

### LINKED ISSUE #3960: Not possible to disable OpenAPI, swagger UI

**Description**

Using the following configuration fails: 
  
```
api_platform:
    enable_swagger: false
    enable_swagger_ui: false
    enable_profiler: false
    enable_docs: false
    enable_entrypoint: false
    enable_re_doc: false
```

Indeed, it looks for services declared in 

```
        $loader->load(__DIR__.'/../../../vendor/api-platform/core/src/Bridge/Symfony/Bundle/Resources/config/openapi.xml');
        $loader->load(__DIR__.'/../../../vendor/api-platform/core/src/Bridge/Symfony/Bundle/Resources/config/json_schema.xml');
```

Some of the needed services must be loaded even when these flags are set to false (see Configuration.php).

GrimpEreau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3960
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -375,10 +375,6 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
     {
         $container->setParameter('api_platform.swagger.versions', $config['swagger']['versions']);
 
-        if (empty($config['swagger']['versions'])) {
-            return;
-        }
-
         $loader->load('json_schema.xml');
         $loader->load('swagger.xml');
         $loader->load('openapi.xml');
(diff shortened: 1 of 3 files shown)


alanpoulain:
Should we really load all the Swagger services all the time? I'm not sure it's the right fix.

soyuka:
We discussed this and this still requires changes:
  - schema factory must be loaded even when swagger is disabled
  - disabling swagger should disable swagger ui 
  - DocumentationAction has a dependency to openapi that needs the on-invalid=ignore flag

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -385,13 +385,18 @@ private function registerSwaggerConfiguration(ContainerBuilder $container, array
     {
         $container->setParameter('api_platform.swagger.versions', $config['swagger']['versions']);
 
-        if (empty($config['swagger']['versions'])) {
-            return;
+        if (!$config['enable_swagger'] && $config['enable_swagger_ui']) {
+            throw new RuntimeException('Unsupported configuration with enable_swagger: false and enable_swagger_ui: true');
```
soyuka, line 389:
```suggestion
            throw new RuntimeException('You can not enable the Swagger UI without enabling Swagger, fix this by enabling swagger via the configuration "enable_swagger: true".');
```

soyuka:
thanks @GrimpEreau !

PR #3985, opened 2021-01-26
Fix #3979 identifier chain not breaking after denormalizing

### LINKED ISSUE #3979: identifier denormalizer does not work anymore with custom needs.
**API Platform version(s) affected**: 2.6.x

**Description**  

This following code inspired by https://gist.github.com/soyuka/c25aa66728439eec56419a5a20295592
does not work anymore with 2.6.0 and 2.6.1.

```
<?php

declare(strict_types=1);

namespace App\Identifier;

use Monofony\Contracts\Core\Model\User\AppUserInterface;
use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
use Symfony\Component\Security\Core\Security;

final class AppUserIdentifierNormalizer implements DenormalizerInterface
{
    private $security;

    public function __construct(Security $security)
    {
        $this->security = $security;
    }

    /**
     * {@inheritdoc}
     *
     * @psalm-suppress InvalidReturnType
     * @psalm-suppress InvalidReturnStatement
     */
    public function denormalize($data, $type, $format = null, array $context = [])
    {
        $user = $this->security->getUser();

        if (null === $user || !$user instanceof AppUserInterface) {
            throw new AccessDeniedHttpException();
        }

        return (string) $user->getCustomer()->getId();
    }

    /**
     * {@inheritdoc}
     */
    {
        return 'me' === $data;
    }
}
```
**How to reproduce**  

**Possible Solution**  

**Additional Context**

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3979 
| License       | MIT
| Doc PR        | na

- lower priority of our identifier denormalizer for better DX
- break the chain of normalizers when an identifier transformer is found

### DIFF
--- a/src/Identifier/IdentifierConverter.php
+++ b/src/Identifier/IdentifierConverter.php
@@ -69,6 +69,7 @@ final class IdentifierConverter implements ContextAwareIdentifierConverterInterf
 
                 try {
                     $identifiers[$identifier] = $identifierTransformer->denormalize($value, $type);
+                    break;
                 } catch (InvalidIdentifierException $e) {
                     throw new InvalidIdentifierException(sprintf('Identifier "%s" could not be denormalized.', $identifier), $e->getCode(), $e);
                 }
--- a/src/Bridge/Symfony/Bundle/Resources/config/api.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/api.xml
@@ -275,11 +275,11 @@
         </service>
 
         <service id="api_platform.identifier.integer" class="ApiPlatform\Core\Identifier\Normalizer\IntegerDenormalizer" public="false">
-            <tag name="api_platform.identifier.denormalizer" />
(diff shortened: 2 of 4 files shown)


src/Identifier/IdentifierConverter.php
```diff
@@ -69,6 +69,7 @@ public function convert($data, string $class, array $context = []): array
 
                 try {
                     $identifiers[$identifier] = $identifierTransformer->denormalize($value, $type);
+                    break;
                 } catch (InvalidIdentifierException $e) {
                     throw new InvalidIdentifierException(sprintf('Identifier "%s" could not be denormalized.', $identifier), $e->getCode(), $e);
                 }
```
soyuka, line 75: add test

PR #3987, opened 2021-01-26
Allow mapping.paths override bundles configs

kayue:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Right now third party bundle's config is load after project's own resources definitions (ie `api_platform.mapping.paths`), and therefore vendor's config always override the project's own.

I suggest we change the load order to allow user to override 3rd party settings.


kayue:
@dunglas do you think we should have this feature?

alanpoulain:
I don't have a strong opinion on this one, but it's a BC.

alanpoulain:
Superseded by https://github.com/api-platform/core/pull/4465.

PR #3891, opened 2020-12-07
Fix tests with serialization groups

pmontoya:
…ceItemJsonSchema with serialization groups

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3881
| License       | MIT

The serialization context parameter added in src/Bridge/Symfony/Bundle/Test/ApiTestAssertionsTrait.php avoid extra_attributes but breaks serialization groups schema.

Serialization context is now merge in SchemaFactory

Issue #3896, opened 2020-12-09
Schema test wrong since v2.5.8: tested working on 2.5.7


FrankHouwelingCastor:
v2.5.8

**Description**  
The json schema in the test is different from the json schema in documentation and from the json schema of the API response.

**How to reproduce**  
![image](https://user-images.githubusercontent.com/69353572/101617741-4f0c6700-3a11-11eb-9181-7a7885c3d647.png)

```
static::createClient()->request('GET', '/studies');
        self::assertResponseIsSuccessful();
        self::assertResponseHeaderSame('content-type', 'application/ld+json; charset=utf-8');

        self::assertMatchesResourceCollectionJsonSchema(Study::class);
```

Will return:
hydra:member[0].roles[0].user: Object value found, but a string is required

If I var_dump the schema that is being generated by the ApiTestAssertionsTrait::assertMatchesResourceCollectionJsonSchema function it indeed expects a string. This is different from the output of the API docs, and also from what the API returns.

This functionality is broken because of the api-platform/core update (v2.5.7 => v2.5.8). Tested to be working before the update of these packages, with the same code.

madrzejewski:
I have the same issue here, all my tests passed on 2.5.7 and just after updating to 2.5.9 I have errors on the `assertMatchesResourceItemJsonSchema` test but the response seems fine to me.

```
Failed asserting that Array &0 (
    '@context' => '/api/contexts/Template'
    '@id' => '/api/templates/112237'
    '@type' => 'Template'
    'id' => 112237
    [....]
    'additionalsParams' => Array &3 (
        'purgeSecret' => '2c4eb408-f224-44b7-b2aa-69d2877d2bc9'
    )
) matches the provided JSON Schema.
additionalsParams: Object value found, but an array is required
```

alanpoulain:
See https://github.com/api-platform/core/issues/3918. The assertion was not working at all before.

madrzejewski:
@alanpoulain Thanks for your answer. Any ideas how I can correct such error ? I'm assuming I'm doing something wrong but I don't see anything weird. I'm new to Api Platform, I'm following the doc for the testing, I'm not familiar with any of this at this moment. 

I think the problem is in the entity, on the comment around the property ? 

```
    /**
     * @ORM\Column(type="array", nullable=true)
     * @ConstraintCustomJson(type="customHeaders")
     * @Groups({"template:read", "template:write", "template:item:patch"})
     */
```

alanpoulain:
[...]

PR #3993, opened 2021-01-28
SchemaFactory: fix #3881 merge serializer context

### LINKED ISSUE #3881: assertMatchesResourceCollectionJsonSchema and assertMatchesResourceItemJsonSchema breaks after adding ALLOW_EXTRA_ATTRIBUTES to false
**API Platform version(s) affected**: 2.5.8

**Description**  
[This commit](https://github.com/api-platform/core/commit/7595ac7c17eecd5be320026583c040c44c9ed851#diff-b0c03d04114819985843f011d3de75f6516e4c0c26ed9b9357c077338426eb6e) add the serialization context parameter but it breaks serialization groups schema.

I think this due to https://github.com/api-platform/core/blob/ab474b69c4e05318659f91e3cc55625b8c302735/src/JsonSchema/SchemaFactory.php#L283 : as `$serializerContext` is not null, `$this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName)` is not call so serialization groups aren't set in serialization context and assert is false.

**How to reproduce**  
Create an Entity with some properties not serialized and launch test on schema with one of `assertMatchesResourceCollectionJsonSchema` or `assertMatchesResourceItemJsonSchema` : the result will be false

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3881
| License       | MIT

Superseeds https://github.com/api-platform/core/pull/3891 by @pmontoya with a test.

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -292,7 +292,7 @@ final class SchemaFactory implements SchemaFactoryInterface
 
         return [
             $resourceMetadata,
-            array_merge($serializerContext ?? [], $this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName)),
+            array_merge($this->getSerializerContext($resourceMetadata, $type, $operationType, $operationName), $serializerContext ?? []),
             $this->getValidationGroups($resourceMetadata, $operationType, $operationName),
             $inputOrOutput['class'],
         ];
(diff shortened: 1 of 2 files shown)


alanpoulain:
Isn't it dangerous to start merging the contexts? I think it would be preferable to be able to generate context easily for the user.
See also: https://github.com/api-platform/core/issues/3896#issuecomment-763873357.

soyuka:
@pmontoya we did a core team meeting and decided to close this. We wish that user set this `ALLOW_EXTRA_ATTRIBUTES` at the global symfony level instead of in this context. Also, we opened https://github.com/api-platform/core/issues/4006 to follow the work that will be done at this level. Note that we will revert the `ALLOW_EXTRA_ATTRIBUTES` being set in the assertions.

PR #4012, opened 2021-02-02
Update resource metadata if new config is given

kayue:
If a resource is being defined twice, allow the later one to override the previous one.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

If a resource is being defined twice, allow the later one to override the previous one.

This is useful when resources is predefined by a 3rd party library, eg Sylius.

### DIFF
--- a/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
@@ -87,23 +87,23 @@ final class ExtractorResourceMetadataFactory implements ResourceMetadataFactoryI
     }
 
     /**
-     * Update resource metadata if new config is given
+     * Update resource metadata if new config is given.
      */
     private function update(ResourceMetadata $resourceMetadata, array $metadata): ResourceMetadata
     {
         foreach (['shortName', 'description', 'iri', 'itemOperations', 'collectionOperations', 'subresourceOperations', 'graphql', 'attributes'] as $property) {
-            $propertyData = $metadata[$property];
-            $parentPropertyData = $resourceMetadata->{'get'.ucfirst($property)}();
+            $propertyMetadata = $metadata[$property];
+            $parentPropertyMetadata = $resourceMetadata->{'get'.ucfirst($property)}();
 
-            if (null === $propertyData) {
+            if (null === $propertyMetadata) {
                 continue;
             }
 
-            if (is_array($parentPropertyData) && is_array($propertyData)) {
-                $propertyData = array_merge($parentPropertyData, $propertyData);
+            if (\is_array($parentPropertyMetadata) && \is_array($propertyMetadata)) {
+                $propertyMetadata = array_merge($parentPropertyMetadata, $propertyMetadata);
             }
 
-            $resourceMetadata = $resourceMetadata->{'with'.ucfirst($property)}($propertyData);
+            $resourceMetadata = $resourceMetadata->{'with'.ucfirst($property)}($propertyMetadata);
         }
 
         return $resourceMetadata;
--- a/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
+++ b/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
@@ -87,16 +87,23 @@ final class ExtractorResourceMetadataFactory implements ResourceMetadataFactoryI
     }
 
     /**
-     * Creates a new instance of metadata if the property is not already set.
+     * Update resource metadata if new config is given
      */
     private function update(ResourceMetadata $resourceMetadata, array $metadata): ResourceMetadata
     {
         foreach (['shortName', 'description', 'iri', 'itemOperations', 'collectionOperations', 'subresourceOperations', 'graphql', 'attributes'] as $property) {
-            if (null === $metadata[$property] || null !== $resourceMetadata->{'get'.ucfirst($property)}()) {
+            $propertyData = $metadata[$property];
+            $parentPropertyData = $resourceMetadata->{'get'.ucfirst($property)}();
+
+            if (null === $propertyData) {
                 continue;
             }
 
-            $resourceMetadata = $resourceMetadata->{'with'.ucfirst($property)}($metadata[$property]);
+            if (is_array($parentPropertyData) && is_array($propertyData)) {
+                $propertyData = array_merge($parentPropertyData, $propertyData);
+            }
+
+            $resourceMetadata = $resourceMetadata->{'with'.ucfirst($property)}($propertyData);
         }
 
         return $resourceMetadata;
--- a/tests/Fixtures/FileConfigurations/resources.xml
+++ b/tests/Fixtures/FileConfigurations/resources.xml
@@ -23,6 +23,7 @@
             <collectionOperation name="my_collection_op">
                 <attribute name="method">POST</attribute>
                 <attribute name="path">the/collection/path</attribute>
+                <attribute name="status">201</attribute>
             </collectionOperation>
         </collectionOperations>
         <subresourceOperations>
--- a/tests/Fixtures/FileConfigurations/resources.yml
+++ b/tests/Fixtures/FileConfigurations/resources.yml
@@ -12,6 +12,7 @@ resources:
             my_collection_op:
                 method: 'POST'
                 path: 'the/collection/path'
+                status: '201'
         subresourceOperations:
             my_collection_subresource:
                 path: 'the/subresource/path'
--- a/tests/Fixtures/FileConfigurations/resourcesoverride.xml
+++ b/tests/Fixtures/FileConfigurations/resourcesoverride.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<resources xmlns="https://api-platform.com/schema/metadata"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="https://api-platform.com/schema/metadata
+           https://api-platform.com/schema/metadata/metadata-2.0.xsd">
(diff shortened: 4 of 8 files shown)


src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
```diff
@@ -87,16 +87,23 @@ private function handleNotFound(?ResourceMetadata $parentPropertyMetadata, strin
...
+            $propertyData = $metadata[$property];
+            $parentPropertyData = $resourceMetadata->{'get'.ucfirst($property)}();
+
+            if (null === $propertyData) {
                 continue;
             }
 
-            $resourceMetadata = $resourceMetadata->{'with'.ucfirst($property)}($metadata[$property]);
+            if (is_array($parentPropertyData) && is_array($propertyData)) {
+                $propertyData = array_merge($parentPropertyData, $propertyData);
```
alanpoulain, line 103: Should we really merge them though? Which is the use case?

src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
```diff
@@ -87,16 +87,23 @@ private function handleNotFound(?ResourceMetadata $parentPropertyMetadata, strin
...
 
     /**
-     * Creates a new instance of metadata if the property is not already set.
+     * Update resource metadata if new config is given
      */
     private function update(ResourceMetadata $resourceMetadata, array $metadata): ResourceMetadata
     {
         foreach (['shortName', 'description', 'iri', 'itemOperations', 'collectionOperations', 'subresourceOperations', 'graphql', 'attributes'] as $property) {
-            if (null === $metadata[$property] || null !== $resourceMetadata->{'get'.ucfirst($property)}()) {
+            $propertyData = $metadata[$property];
```
alanpoulain, line 95: Should be named `$propertyMetadata`

src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
kayue: @alanpoulain Thank you for checking. I am using Sylius, a e-commerce software that utilise ApiPlatform, it [pre-defined  tons of APIs out of the boxes](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ApiBundle/Resources/config/api_resources). It is common that developer would like to change or extend its behaviour. 

Whether we should merge them, if I am not mistaken it is already the case for `resource.properties`.

https://github.com/api-platform/core/blob/41842c430e1eaf23b18d72a7cbc4744324347be4/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php#L110-L116

alanpoulain: Sure we merge the properties, but merging what's inside the properties seems to be dangerous.
Yes I know Sylius allow to extend its resources a lot, but maybe it would be better to redefine the properties when extending instead?

kayue: @alanpoulain This pull request aim to allow extending resource metadata such as `itemOperations`, `collectionOperations`, `subresourceOperations`, but resource properties mapping is not handled in this class.

Regarding safety, sorry I don't understand why merging config is danger. Symfony bundle configuration support merging too, I naturally assume api resources could do the same. Without proper merging, if one want to add a new operation to an existing resource, I will have to copy and paste all operations from Sylius' definition. 

Anyway, without this pull request even simple overriding is not possible, the new config is discarded due to line 95's null checking.

https://github.com/api-platform/core/blob/41842c430e1eaf23b18d72a7cbc4744324347be4/src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php#L94-L100

I am happy to improve the behaviour if you can tell me what is the ideal behaviour. 

Thanks.

alanpoulain: Merging configuration may be a danger because you also inherit security configuration or serialization context. I think it's better to be explicit in these cases to prevent unwanted configuration.
However I understand your needs, particularly in a Sylius context. I think Sylius needs to provide a decorator for this factory in its bridge with API Platform instead of having it directly here.

dunglas:
API Platform has always worked like that (replacing and not merging), and it's on purpose for reasons explained by @alanpoulain. We should find a way to allow an easier customization of the configuration provided by third parties such as Sylius, but we cannot change this behavior like that in a minor version, this could break and even introduce major security issues in existing projets (and don't think than merging is desirable in the context of API Platform, but we can discuss this for 3.0).

We have the same "issue" for defaults by the way (merging could've convenient and maybe expected, but it's inconsistent with the other parts of the framework).

kayue:
@dunglas @alanpoulain Understood. If merging is not acceptable, how about allow latter config to override previous one completely? Does the same security concern apply?

src/Metadata/Resource/Factory/ExtractorResourceMetadataFactory.php
kayue: @alanpoulain Understood. I have two suggestions:

- Throw warning when operation it is already defined, because silently ignoring new config is as danger as silently removing old config.
- Implement deep merge, make it behave like Symfony's `security.yaml` if this makes it any safer.

Kindly let me know your thought. Thanks!

dunglas:
This looks like a reasonable approach to me!

kayue:
Parent:
```yaml
resources:
    'ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy':
        description: 'Dummy resource'
        itemOperations:
            my_op_name:
                method: 'GET'
            my_other_op_name:
                method: 'POST'
        collectionOperations:
            my_collection_op:
                method: 'POST'
                path: 'the/collection/path'
                status: '201'
```

Child:
```yaml
resources:
    'ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy':
        description: New description
        itemOperations:
            my_op_name:
                method: 'GET'
                path: 'the/path'
        collectionOperations:
            my_collection_op:
                method: 'POST'
                status: '200'
```

alanpoulain:
I'm not sure we want to merge the operations like this either.
And it's a BC anyway.
Why not doing this is in Sylius, since most use cases are there?

kayue:
@alanpoulain I continued the work because @dunglas thought it is a reasonable approach. 

ApiPlatform aim to be the backbone of many projects, therefore I think it is a flexibility ApiPlatform should allow. Extendability is one of the best of Symfony ecosystem offer, and we should do the same here. 

Regarding your suggestion on decorate the `ExtractorResourceMetadataFactory` service, it would require copy and paste a large trunk of code, therefore I would like to avoid it in my project if possible.

I think we just need to decide whether this is an expected behaviour of ApiPlatform. In my opinion config overwrite is a common feature of Symfony bundles, I don't see why ApiPlatform is different. 

I understand this is a big BC break and I can understand this feature has to wait until the next major version. I just think this can be a useful feature for many projects out there.

To me I think the least ApiPlatform can do is to throw a warning when it decided to ignore developer's configuration.

dunglas:
Maybe can we just provide a new attribute `merge=true` to this behavior opt-in? It it can be added in `defaults` and we have best of both worlds. If it's proved convenient we could even enable this attribute by default in API Platform 3. WDYT?

alanpoulain:
Yes if we want this feature it should be configurable with an attribute to prevent BC (and to not wait for the next major).
Not sure about making it the default but we'll see!

kayue:
@dunglas @alanpoulain Should I can do that. Can you be a bit more specific on how you see this attribute works? Something like this?

```
resources:
    ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy:
        description: 'Dummy resource'
        attributes:
            merge: true
```

I assume this is an attribute defined in the child resource to allow developer to force merging?

(Where can I found the docs for `defaults` options? Couldn't found it on the official site.)

alanpoulain:
Shouldn't it be better at the root level? Like this:
```yaml
resources:
    ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\FileConfigDummy:
        description: 'Dummy resource'
        merge: true
```
@dunglas WDYT?
Having it in the child resource seems right to me.
I'm not sure there is a doc (it would be nice to have one). PR is here: https://github.com/api-platform/core/pull/3151

andriusvo:
@kayue Corrections should be done. I can give an example from real world with Sylius. F.e. we have `itemOperation` defined:

`
            <itemOperation name="shop_apply_coupon">
                <attribute name="method">PATCH</attribute>
                <attribute name="path">/shop/orders/{id}/apply-coupon</attribute>
                <attribute name="messenger">input</attribute>
                <attribute name="input">Sylius\Bundle\ApiBundle\Command\Cart\ApplyCouponToCart</attribute>
                <attribute name="denormalization_context">
                    <attribute name="groups">shop:cart:apply_coupon</attribute>
                </attribute>
                <attribute name="openapi_context">
                    <attribute name="summary">Applies coupon to cart</attribute>
                </attribute>
            </itemOperation>
`

And I want to add attribute `validation_groups` to this itemOperation. So, I expect to write the following code:

`
resources:
    App\Entity\Order\Order:
        itemOperations:
            shop_apply_coupon:
                validation_groups: [sylius, some_validation_group]
`

With function `array_merge`, only `validation_groups` attribute appear, and other attributes from `$parentPropertyMetadata` dissapear. In this case, maybe it's better to use function `array_replace_recursive` ?

kayue:
@andriusvo I don't know does Api Platform want this merging feature at all.

alanpoulain:
If this is opt-in, yes!

antonioperic:
It is not promoting our blog, 
but here is solution how we did it in company https://locastic.com/blog/clean-api-configuration-in-sylius/

PR #4015, opened 2021-02-02
Fix partial pagination which no longer returns the "hydra:next" property

meyerbaptiste:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/core/pull/3967#issuecomment-771677553
| License       | MIT
| Doc PR        | n/a

@jeremyteyssedre confused `$pageTotalItems` with the total number of items in the collection. `$pageTotalItems` is the number of items in the current page. As a reminder, with the partial pagination we cannot determine if it is the last page. Because of this, we will always have the `hydra:next` property, even on the last page. It is an "assumed bug".

cc @clementtalleu

### DIFF
--- a/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
+++ b/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
@@ -113,7 +113,7 @@ final class PartialCollectionViewNormalizer implements NormalizerInterface, Norm
                 $data['hydra:view']['hydra:previous'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $currentPage - 1.);
             }
 
-            if (null !== $lastPage && $currentPage < $lastPage || null === $lastPage && $pageTotalItems > $itemsPerPage) {
+            if (null !== $lastPage && $currentPage < $lastPage || null === $lastPage && $pageTotalItems >= $itemsPerPage) {
                 $data['hydra:view']['hydra:next'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $currentPage + 1.);
             }
         }
--- a/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
+++ b/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
@@ -62,7 +62,7 @@ final class PartialCollectionViewNormalizer implements NormalizerInterface, Norm
         }
 
         $currentPage = $lastPage = $itemsPerPage = $pageTotalItems = null;
-        if ($paginated = $object instanceof PartialPaginatorInterface) {
+        if ($paginated = ($object instanceof PartialPaginatorInterface)) {
             if ($object instanceof PaginatorInterface) {
                 $paginated = 1. !== $lastPage = $object->getLastPage();
             } else {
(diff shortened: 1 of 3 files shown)


src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -81,41 +81,20 @@ public function normalize($object, $format = null, array $context = [])
...
 
+        $cursorPaginationAttribute = null;
         $metadata = isset($context['resource_class']) && null !== $this->resourceMetadataFactory ? $this->resourceMetadataFactory->create($context['resource_class']) : null;
         $isPaginatedWithCursor = $paginated && null !== $metadata && null !== $cursorPaginationAttribute = $metadata->getCollectionOperationAttribute($context['collection_operation_name'] ?? $context['subresource_operation_name'], 'pagination_via_cursor', null, true);
 
-        $data['hydra:view'] = [
-            '@id' => IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $paginated && !$isPaginatedWithCursor ? $currentPage : null),
-            '@type' => 'hydra:PartialCollectionView',
-        ];
+        $data['hydra:view'] = ['@id' => null, '@type' => 'hydra:PartialCollectionView'];
```
soyuka, line 88: I don't think that `@id` should be null, maybe remove the key completely?

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -164,4 +143,41 @@ private function cursorPaginationFields(array $fields, int $direction, $object)
...
+            $data['hydra:view']['hydra:last'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $lastPage);
+        }
+
+        if (1. !== $currentPage) {
+            $data['hydra:view']['hydra:previous'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $currentPage - 1.);
+        }
+
+        if ((null !== $lastPage && $currentPage < $lastPage) || (null === $lastPage && null === $lastPage && $pageTotalItems >= $itemsPerPage)) {
+            $data['hydra:view']['hydra:next'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $currentPage + 1.);
+        }
```
soyuka, line 179: what about :
```
if ((null !== $lastPage && $currentPage >= $lastPage) || $pageTotalItems < $itemsPerPage) {
    return $data;
}

$data['hydra:view']['hydra:next'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], $this->pageParameterName, $currentPage + 1.);
```

 anyways `null === $lastPage && null === $lastPage` is there 2 times

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
meyerbaptiste: I agree, this is what I did but I finally changed to preserve the order of the properties.

src/Hydra/Serializer/PartialCollectionViewNormalizer.php
meyerbaptiste: It's a c/p but I'll take a look.

soyuka (approved):

soyuka:
thanks @meyerbaptiste !

PR #4016, opened 2021-02-02
Fix #4005 BC layer for openapi normalizer

### LINKED ISSUE #4005: OpenAPI BC Layer
- Add a flag `openapi_bc_layer` that uses the new normalizer + the OpenAPI Factory
- Inject the new normalizer in the old one (DocumentationNormalizer):
  - if the denormalized object is an OpenApi we use the new normalizer
  - if not trigger a deprecation + advice to use the OpenApiFactory

This allows users to still decorate the old service + adds a deprecation

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes
| Tickets       | fixes #4005 
| License       | MIT
| Doc PR        | na

Adds a BC layer to allow decorating the old normalizer with the new OpenApi object.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -483,6 +483,7 @@ final class Configuration implements ConfigurationInterface
                                 ->scalarNode('email')->defaultNull()->info('The email address of the contact person/organization. MUST be in the format of an email address.')->end()
                             ->end()
                         ->end()
+                        ->booleanNode('backward_compatibility')->defaultTrue()->info('Enable this to decorate the "api_platform.swagger.normalizer.documentation" instead of decorating the OpenAPI factory.')->end()
                         ->scalarNode('termsOfService')->defaultNull()->info('A URL to the Terms of Service for the API. MUST be in the format of a URL.')->end()
                         ->arrayNode('license')
                         ->addDefaultsIfNotSet()
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -483,7 +483,7 @@ final class Configuration implements ConfigurationInterface
                                 ->scalarNode('email')->defaultNull()->info('The email address of the contact person/organization. MUST be in the format of an email address.')->end()
                             ->end()
                         ->end()
-                        ->booleanNode('backward_compatibility')->defaultTrue()->info('Enable this to decorate the "api_platform.swagger.normalizer.documentation" instead of decorating the OpenAPI factory.')->end()
+                        ->booleanNode('backward_compatibility_layer')->defaultTrue()->info('Enable this to decorate the "api_platform.swagger.normalizer.documentation" instead of decorating the OpenAPI factory.')->end()
                         ->scalarNode('termsOfService')->defaultNull()->info('A URL to the Terms of Service for the API. MUST be in the format of a URL.')->end()
                         ->arrayNode('license')
                         ->addDefaultsIfNotSet()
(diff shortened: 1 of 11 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -483,6 +483,7 @@ private function addOpenApiSection(ArrayNodeDefinition $rootNode): void
                                 ->scalarNode('email')->defaultNull()->info('The email address of the contact person/organization. MUST be in the format of an email address.')->end()
                             ->end()
                         ->end()
+                        ->booleanNode('backward_compatibility')->defaultTrue()->info('Enable this to decorate the "api_platform.swagger.normalizer.documentation" instead of decorating the OpenAPI factory.')->end()
```
alanpoulain, line 486: Why not using `canBeEnabled`?

soyuka: because we don't need it?

alanpoulain: Oh yes it's because it's only a boolean value I see. `canBeEnabled` is used when there are other parameters too.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -171,13 +175,22 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
...
     }
 
     /**
      * {@inheritdoc}
      */
     public function normalize($object, $format = null, array $context = [])
     {
+        if ($object instanceof OpenApi) {
+            // uncomment this for 2.7
+            // @trigger_error('Using the swagger DocumentationNormalizer is deprecated in favor of decorating the OpenApiFactory', \E_USER_DEPRECATED);
```
dunglas, line 190: You can maybe uncomment this now. I know that we must not introduce deprecations in patch version, but this is more a bug fix than something else, it was missing in 2.6.0 and it currently doesn't work at all, so I can we can add it here.

src/Swagger/Serializer/DocumentationNormalizer.php
```diff
@@ -102,14 +103,17 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
...
... public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, $jsonSchemaFactory = null, $jsonSchemaTypeFactory = null, OperationPathResolverInterface $operationPathResolver = null, UrlGeneratorInterface $urlGenerator = null, $filterLocator = null, NameConverterInterface $nameConverter = null, bool $oauthEnabled = false, string $oauthType = '', string $oauthFlow = '', string $oauthTokenUrl = '', string $oauthAuthorizationUrl = '', array $oauthScopes = [], array $apiKeys = [], SubresourceOperationFactoryInterface $subresourceOperationFactory = null, bool $paginationEnabled = true, string $paginationPageParameterName = 'page', bool $clientItemsPerPage = false, string $itemsPerPageParameterName = 'itemsPerPage', $formats = [], bool $paginationClientEnabled = false, string $paginationClientEnabledParameterName = 'pagination', array $defaultContext = [], array $swaggerVersions = [2, 3], IdentifiersExtractorInterface $identifiersExtractor = null, bool $openApiBackwardCompatibility = true, NormalizerInterface $openApiNormalizer = null)
```
dunglas, line 116: Can't we just change the wiring in the extension instead of injecting `backward_compatibility` here?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
dunglas, line 486: backward_compatibility_layer?

dunglas (changes requested):

dunglas (approved):

PR #4019, opened 2021-02-03
Serialize openapi properly fix #3997

### LINKED ISSUE #3997: Symfony name converter breaking the OpenApi docs
**API Platform version(s) affected**: 2.6.1

**Description**  

If we use a Symfony name converter, the docs page is broken. The `Authorize` button on the top left which has been introduced in the version 2.6 is not shown. Also the `Request Body` section (between `Parameters` and `Responses`) under the endpoints are completely missing. So, you cannot also try out any endpoints as this section is completely missing

**How to reproduce**  

Configure Api Platform to use the Symfony's camel-to-snake case converter as documented [here](https://api-platform.com/docs/core/serialization/#name-conversion)
Activate the service:
```
# api/config/services.yaml
services:
    'Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter': ~
```
and then use it:
```
# api/config/packages/api_platform.yaml
api_platform:
    name_converter: 'Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter'
```
Now go and try some of your endpoints from the OpenApi (Swagger) docs. The `Request Body` section will no more show up. And the endpoints cannot be tested anymore.

**Possible Solution**  

Not yet looked into it but my assumption is, something is breaking when this name converter is run. Didn't try yet with custom name convertor.

**Additional Context**  

When trying with some API testing clients like [Postman](https://www.postman.com/), both the parameters `createdAt` as well as `created_at` are working.

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3997 
| License       | MIT
| Doc PR        | na

When adding a name converter some properties of OpenAPI were serialized without respecting the specification causing issues. This fixes it.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Resources/config/openapi.xml
+++ b/src/Bridge/Symfony/Bundle/Resources/config/openapi.xml
@@ -5,27 +5,24 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="api_platform.openapi.object_normalizer" class="Symfony\Component\Serializer\Normalizer\ObjectNormalizer" public="false">
-            <argument>null</argument>
-            <argument>null</argument>
-            <argument type="service" id="api_platform.property_accessor"/>
-            <argument type="service" id="api_platform.property_info"/>
-            <call method="setSerializer">
-                <argument type="service" id="api_platform.openapi.serializer"/>
-            </call>
-        </service>
-
-        <service id="api_platform.openapi.serializer" class="Symfony\Component\Serializer\Serializer" public="false">
-            <argument type="collection">
-                <argument type="service" id="api_platform.openapi.object_normalizer" />
-            </argument>
-            <argument type="collection">
-                <argument type="service" id="serializer.encoder.json" />
-            </argument>
-        </service>
-
         <service id="api_platform.openapi.normalizer" class="ApiPlatform\Core\OpenApi\Serializer\OpenApiNormalizer" public="false">
-            <argument type="service" id="api_platform.openapi.object_normalizer"/>
+            <argument type="service">
+                <service class="Symfony\Component\Serializer\Serializer">
+                    <argument type="collection">
+                        <argument type="service">
+                            <service class="Symfony\Component\Serializer\Normalizer\ObjectNormalizer">
+                                <argument>null</argument>
+                                <argument>null</argument>
+                                <argument type="service" id="api_platform.property_accessor"/>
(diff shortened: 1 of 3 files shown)


src/OpenApi/Model/Components.php
```diff
@@ -21,8 +23,10 @@ final class Components
     private $responses;
     private $parameters;
     private $examples;
+    /** @SerializedName("requestBodies") */
```
dunglas, line 26: You shouldn't rely on annotations, this creates a hard dependency to `doctrine/annotations` (this will not work if this lib isn't installed). You should probably use the XML format instead.

dunglas (changes requested):

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -26,59 +28,77 @@ final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsM
     public const FORMAT = 'json';
     private const EXTENSION_PROPERTIES_KEY = 'extensionProperties';
 
-    private $decorated;
+    private $propertyAccessor;
+    private $propertyInfo;
 
-    public function __construct(NormalizerInterface $decorated)
+    public function __construct(PropertyAccessorInterface $propertyAccessor, PropertyInfoExtractorInterface $propertyInfo)
```
dunglas, line 34: Isn't this a BC break?

dunglas (changes requested):
This looks good to me, but we must prevent this break (by removing the type hints and triggering a deprecation maybe?)

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -26,59 +28,77 @@ final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsM
...
     {
-        $context[AbstractObjectNormalizer::PRESERVE_EMPTY_OBJECTS] = true;
-        $context[AbstractObjectNormalizer::SKIP_NULL_VALUES] = true;
-
-        return $this->recursiveClean($this->decorated->normalize($object, $format, $context));
+        return $this->objectToArray($object);
     }
 
-    private function recursiveClean($data): array
+    private function objectToArray($object)
```
dunglas, line 48:
```suggestion
    private function objectToArray($object): array
```

?

soyuka: method renamed

dunglas (approved):

PR #4027, opened 2021-02-05
pattern must be compliant with ECMA 262 regex

### LINKED ISSUE #4022: OpenAPI: pattern for Regex Constraint should use htmlPattern
**API Platform version(s) affected**: 2.6.1

**Description**  
Given a pattern with regex delimiters for \Symfony\Component\Validator\Constraints\Regex, it cannot be used as http://dev.w3.org/html5/spec/single-page.html#the-pattern-attribute

**How to reproduce**  
with `'/^([-a-zA-Z0-9_])+$/'` is used in OpenAPI documentation for property:
```
...
      properties:
        property:
          pattern: '/^([-a-zA-Z0-9_])+$/
          description: '...'
          type: string
...
```
Correct would be (without delimiters):
```
...
      properties:
        property:
          pattern: '([-a-zA-Z0-9_])+'
          description: '...'
          type: string
...
```

**Possible Solution**  
In \ApiPlatform\Core\Bridge\Symfony\Validator\Metadata\Property\Restriction\PropertySchemaRegexRestriction::create(), instead of `$constraint->pattern` use `$constraint->getHtmlPattern()`

See https://github.com/symfony/validator/blob/367afc5864bfdb07508668dc0dfc06e56b6b89d5/Constraints/Regex.php#L99

**Additional Context**  
https://swagger.io/docs/specification/data-models/data-types/#pattern

stchr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #4022 
| License       | MIT

https://swagger.io/docs/specification/data-models/data-types/#pattern uses [ECMA 262 regex syntax](https://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1), so use getHtmlPattern() to get valid expression.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -50,9 +50,6 @@ parameters:
 		-
 			message: '#Call to an undefined method Doctrine\\Persistence\\ObjectManager::getConnection\(\)#'
 			path: src/Bridge/Doctrine/Common/Util/IdentifierManagerTrait.php
-		-
-			message: '#Call to an undefined method Symfony\\Component\\Validator\\Constraint::getHtmlPattern\(\)\.#'
-			path: src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
 		# https://github.com/willdurand/Negotiation/issues/89#issuecomment-513283286
 		-
 			message: '#Call to an undefined method Negotiation\\AcceptHeader::getType\(\)\.#'
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -50,6 +50,9 @@ parameters:
 		-
 			message: '#Call to an undefined method Doctrine\\Persistence\\ObjectManager::getConnection\(\)#'
 			path: src/Bridge/Doctrine/Common/Util/IdentifierManagerTrait.php
+		-
+			message: '#Call to an undefined method Symfony\\Component\\Validator\\Constraint::getHtmlPattern\(\)\.#'
+			path: src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
 		# https://github.com/willdurand/Negotiation/issues/89#issuecomment-513283286
 		-
 			message: '#Call to an undefined method Negotiation\\AcceptHeader::getType\(\)\.#'
--- a/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
@@ -29,7 +29,7 @@ class PropertySchemaRegexRestriction implements PropertySchemaRestrictionMetadat
      */
     public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
     {
-        return isset($constraint->pattern) ? ['pattern' => $constraint->pattern] : [];
+        return $constraint->getHtmlPattern() ? ['pattern' => $constraint->getHtmlPattern()] : [];
     }
 
     /**
--- a/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
@@ -29,7 +29,11 @@ class PropertySchemaRegexRestriction implements PropertySchemaRestrictionMetadat
      */
     public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
     {
-        return $constraint->getHtmlPattern() ? ['pattern' => $constraint->getHtmlPattern()] : [];
+        if ($constraint instanceof Regex && $constraint->getHtmlPattern()) {
+            return ['pattern' => $constraint->getHtmlPattern()];
+        }
+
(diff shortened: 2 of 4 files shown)


src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
```diff
@@ -29,7 +29,7 @@ class PropertySchemaRegexRestriction implements PropertySchemaRestrictionMetadat
      */
     public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
     {
-        return isset($constraint->pattern) ? ['pattern' => $constraint->pattern] : [];
+        return $constraint->getHtmlPattern() ? ['pattern' => $constraint->getHtmlPattern()] : [];
```
dunglas, line 32:
```suggestion
        return method_exists($constraint, 'getHtmlPattern') && $constraint->getHtmlPattern() ? ['pattern' => $constraint->getHtmlPattern()] : [];
```

(to prevent the PHPStan exception and bugs if the constraint doesn't implement this method).

dunglas (changes requested):

phpstan.neon.dist
```diff
@@ -50,6 +50,9 @@ parameters:
 		-
 			message: '#Call to an undefined method Doctrine\\Persistence\\ObjectManager::getConnection\(\)#'
 			path: src/Bridge/Doctrine/Common/Util/IdentifierManagerTrait.php
+		-
+			message: '#Call to an undefined method Symfony\\Component\\Validator\\Constraint::getHtmlPattern\(\)\.#'
```
soyuka, line 54: imo this is not a false positive, Contraint has not such a method only `Symfony\Component\Validator\Constraints\Regex` has it.

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
soyuka, line 32: this is valid only `if ($constraint instanceof Regex)`

soyuka (changes requested):

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
stchr: Sure, but that is the \ApiPlatform\Core\Bridge\Symfony\Validator\Metadata\Property\Restriction\PropertySchemaRegexRestriction::supports method for, isn't?
```
public function supports(Constraint $constraint, PropertyMetadata $propertyMetadata): bool
    {
        return $constraint instanceof Regex && $constraint->match;
    }
```

stchr: Anyway, instanceof check doesn't hurt, implemented.

phpstan.neon.dist
stchr: Reverted, now checking with instanceof Regex

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
stchr: Solved with suggestion from @soyuka, see https://github.com/api-platform/core/pull/4027#discussion_r571302294

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
```diff
@@ -29,7 +29,11 @@ class PropertySchemaRegexRestriction implements PropertySchemaRestrictionMetadat
      */
     public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
     {
-        return isset($constraint->pattern) ? ['pattern' => $constraint->pattern] : [];
+        if ($constraint instanceof Regex && $constraint->getHtmlPattern()) {
+            return ['pattern' => $constraint->getHtmlPattern()];
+        }
+
+        return [];
```
dunglas, line 36:
```suggestion
        return $constraint instanceof Regex && $constraint->getHtmlPattern() ? ['pattern' => $constraint->getHtmlPattern()] : [];
```

stchr: Okay, good for codecov ;-)

soyuka (approved):

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaRegexRestriction.php
dunglas: It's cheating but yes 😅

dunglas (approved):

dunglas (approved):

dunglas:
Thanks @stchr!

guilliamxavier:
Thanks for the fix but it is incomplete: for `@Assert\Regex(pattern="/^dummy$/")` we should generate a schema with `"pattern": "^dummy$"`, not `"dummy"`.

PR #4031, opened 2021-02-06
Properties need to be allowed by Symfony's serializer at the same time

kayue:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT

### The Problem

- Properties defined in Api Platform's resource cannot be hide by Symfony's serialization group
- Properties's visibility can be defined in both api resource and serializer config. This lead to confusion and unexpected behaviour. 

### The Solution

- Call `parent::getAllowedAttributes()` in `ApiPlatform\Core\Serializer\AbstractItemNormalizer`.
- But this will change the behaviour of ApiPlatform completely, this will force developer to well define their property in Serializer. Not sure is it something you want to merge easily.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -374,7 +374,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         // Gets allowed attributes using serialization groups
         $groupAllowedAttributes = parent::getAllowedAttributes($classOrObject, $context, $attributesAsString);
 
-        if (is_array($groupAllowedAttributes)) {
+        if (\is_array($groupAllowedAttributes)) {
             // Attributes need to be allowed by both API Resource and serialization groups at the same time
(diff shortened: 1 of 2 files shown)


alanpoulain:
Hello,
This is a BC, so it should be made configurable and target `main`.
Do you have a use case that shows the issue? Maybe the `@Ignore` annotation?
In this case it would be great to have a Behat scenario showing the issue.

kayue:
@alanpoulain Updated the branch and added a unit test.

In Sylius it uses both FosRestBundle and ApiPlatform, and they basically shared the same serializer. Therefore I want to be able to control property visibility using serialization group so the settings can be shared by both bundles. In fact this is what ApiPlatform recommended to use as well.

ApiPlatform's `AbstractItemNormalizer::getAllowedAttributes()` is extended from `\Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer`, but it doesn't call parent, therefore the behaviour is different from Symfony and it is unexpected.

PR #4039, opened 2021-02-08
update phpdoc ItemDataProviderInterface::getItem()

julienbornstein:
`ItemDataProviderInterface::getItem()` `$id` argument could also be an `object`

eg: an item with an identifier of type `Ramsey\Uuid\UuidInterface`

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/DataProvider/ItemDataProviderInterface.php
+++ b/src/DataProvider/ItemDataProviderInterface.php
@@ -25,7 +25,7 @@ interface ItemDataProviderInterface
     /**
      * Retrieves an item.
      *
-     * @param array|int|string $id
+     * @param array|int|object|string $id
      *
      * @throws ResourceClassNotSupportedException
      *


alanpoulain:
Hello.
No it shouldn't be an object. See: https://github.com/api-platform/core/blob/0d784fd1da03b1fe4def1222e928f73630110a3c/src/Bridge/Doctrine/Orm/ItemDataProvider.php#L73-L79.
Closing for now.
Feel free to reopen with an example / use case.

julienbornstein:
Hi, @alanpoulain Here's a use case. I've created a dummy project so you can check.

https://github.com/julienbornstein/api-uuid/blob/master/src/DataProvider/FooDataProvider.php#L16

alanpoulain:
Thank you for providing an example. I understand why now.

alanpoulain (approved):

alanpoulain:
Thanks @julienbornstein.

PR #4042, opened 2021-02-09
#4041 bugfix: FilterEagerLoadingExtension foreign identifier handling

### LINKED ISSUE #4041: FilterEagerLoadingExtension doesn't properly handle entities that contain foreign identifier
**API Platform version(s) affected**: 2.6.0

**Description**  
If the entity has a foreign identifier (for example `@Id` property is also a `@OneToOne` relation) - selectors in the generated DQL should be formatted with `IDENTITY(alias.identifier)` instead of just `alias` as **FilterEagerLoadingExtension** does now.

**How to reproduce**  
1. Create an entity that has a property that is identifier and a relation at the same time.
2. Expose GET colection endpoint (`@ApiResource`)
3. Add `@SearchFilter` to some another subproperty
4. Make a request with a subproperty filter

```
class Basket
{
    /**
     * @ORM\Id()
     * @ORM\OneToOne(targetEntity="Device")
     */
    private $device;
    
    /**
     * @ApiFilter(SearchFilter::class, properties={"user.email": "exact"})
     */
    private $user
}
```
Request will fail with the following error:
`[Semantical Error] line 0, col 76 near 'o IN(SELECT o_a2': Error: Invalid PathExpression. Must be a StateFieldPathExpression`

**Possible Solution**  
**FilterEagerLoadingExtension** should check if the entity contains foreign identifier & if it does - format selectors differently.

**Pull request with a proposed fix will be submitted shortly.**

Edvinas9:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #4041
| License       | MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.6.3
 
 * Mercure: Do not use data in options when deleting (#4056)
+* Doctrine: Support for foreign identifiers
 
 ## 2.6.2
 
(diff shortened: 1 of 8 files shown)


soyuka:
needs a test

alanpoulain:
Thanks @Edvinas9.

PR #4054, opened 2021-02-11
Mercure: deletion with custom method sends the full data

### LINKED ISSUE #3969: Mercure: deletion with custom method sends the full data

**Description**  
In case of DELETE call on an entity with `@ApiResource(mercure="object.mercureOptions()")`, if `mercureOptions()` forces 'data' value, the data is still sent to Mercure (and it should not) because entity is still filled when function is called.
There is no way in mercureOptions() to know if we are creating, updating or deleting the current entity.

**How to reproduce**  
Add an entity with : `@ApiResource(mercure="object.mercureOptions()")` and 
```
 public function mercureOptions() {
    return ['private' => true, 'data' => json_encode(["@id":"\/entity\/12", 'customField' => 'customData'])];
} 
```
then issue a DELETE call on this entity. Following message is emited: 
```
Symfony\Component\Mercure\Update {
   -data: "{"@id":"\/entity\/12", "customField":"customData"}"
   -private: true
   -id: null
   -type: null
   -retry: null
}
```
then frontend apps cannot see the deletion.

**Possible Solution**  
Unset data when we handle a deletion: 
```
diff --git a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
index 77d0057b..fb9f82d0 100644
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -192,6 +192,8 @@ final class PublishMercureUpdatesListener
         }
 
         if ('deletedObjects' === $property) {
+            // if it's a deletion, ignore 'data' provided
+            unset($options['data']);
             $this->deletedObjects[(object) [
                 'id' => $this->iriConverter->getIriFromItem($object),
                 'iri' => $this->iriConverter->getIriFromItem($object, UrlGeneratorInterface::ABS_URL),
```

It still works because 'data' is rewritten automatically by APIP and contains only "@id" field.

Tell me if I'm correct here or not, and I'll create a pull request accordingly.

odoucet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3969 
| License       | MIT

When using Mercure and when an entity is deleted, APIP evaluates how to retrieve entity properties.
If we use a custom method with `@ApiResource(mercure="object.mercureOptions()")` the data provided by this function is used, with no regard if we are posting, updating or deleting the entity.
If the custom function returns `data`, in case of a deletion, the data is still sent.

See bug report to fully understand what's wrong.

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -191,6 +191,8 @@ final class PublishMercureUpdatesListener
         }
 
         if ('deletedObjects' === $property) {
+            // if it's a deletion, ignore 'data' provided
+            unset($options['data']);
             $this->deletedObjects[(object) [
                 'id' => $this->iriConverter->getIriFromItem($object),
                 'iri' => $this->iriConverter->getIriFromItem($object, UrlGeneratorInterface::ABS_URL),


dunglas (approved):
LGTM

dunglas:
As it's a bug fix, could you target the 2.6 branch please?

odoucet:
sorry for dummy operations, I'm absolutely not at ease with branches :)
MR modified to target 2.6 branch

alanpoulain:
Shouldn't we have a test?

odoucet:
> Shouldn't we have a test?

I guess I should add some stuff in `PublishMercureUpdatesListenerTest` but I'm totally lost in APIP unit tests and don't understand how it works and what must be done :(

alanpoulain:
Hello @odoucet. I've done a PR with updated unit tests: https://github.com/api-platform/core/pull/4056.
The fix is also different: the `data` option is not used at all when deleting.
Could you confirm it fixes your issue?
I'm closing this PR but thank you!

PR #4056, opened 2021-02-12
fix(mercure): do not use data options when deleting

### LINKED ISSUE #3969: Mercure: deletion with custom method sends the full data

alanpoulain:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3969 
| License       | MIT
| Doc PR        | N/A

Supersedes https://github.com/api-platform/core/pull/4054.

> When using Mercure and when an entity is deleted, APIP evaluates how to retrieve entity properties.
> If we use a custom method with `@ApiResource(mercure="object.mercureOptions()")` the data provided by this function is used, with no regard if we are posting, updating or deleting the entity.
> If the custom function returns `data`, in case of a deletion, the data is still sent.
> 
> See bug report to fully understand what's wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.6.3
+
+* Mercure: Do not use data in options when deleting (#4056)
+
 ## 2.6.2
 
 * Validation: properties regex pattern is now compliant with ECMA 262 (#4027)
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -213,7 +213,7 @@ final class PublishMercureUpdatesListener
             // and I'm not a fond of this approach.
             $iri = $options['topics'] ?? $object->iri;
             /** @var string $data */
-            $data = $options['data'] ?? json_encode(['@id' => $object->id]);
+            $data = json_encode(['@id' => $object->id]);
         } else {
             $resourceClass = $this->getObjectClass($object);
             $context = $options['normalization_context'] ?? $this->resourceMetadataFactory->create($resourceClass)->getAttribute('normalization_context', []);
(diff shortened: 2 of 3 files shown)


odoucet:
Hello,

I can confirm this MR fixes the issue I have :)

alanpoulain:
Thanks for testing!

PR #4060, opened 2021-02-14
WIP: Don't always change DELETE response codes to 204

l-vo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

`RespondListener` convert a string returned by a controller into a response. To reach that, it tries to guess the status code. For a response tied to a DELETE request, the status code is always 204.

But according to [rfc 7231 section 4.3.5](https://tools.ietf.org/html/rfc7231#section-4.3.5), a response with a body can be returned when a DELETE request is received. So in this case, the status code should be 200 instead.

This PR tries to fix that.

### DIFF
--- a/src/EventListener/RespondListener.php
+++ b/src/EventListener/RespondListener.php
@@ -77,6 +77,10 @@ final class RespondListener
 
         $status = $status ?? self::METHOD_TO_CODE[$request->getMethod()] ?? Response::HTTP_OK;
 
+        if (Response::HTTP_NO_CONTENT === $status && $controllerResult) {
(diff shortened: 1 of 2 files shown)


l-vo:
Closed since it seems there is a real will to prevent to return a body for DELETE requests; for instance: 

https://github.com/api-platform/core/blob/39167de2b94ad9231bafb1cdd238c92dcc299661/src/EventListener/WriteListener.php#L110

So changing that probably imply deeper modifications.

Issue #4059, opened 2021-02-14
More options for elasticsearch config


CvekCoding:
**Description**  
I have a case when I need to tune some elastic client settings - my cloud provider require to use https (which is ok) but uses self-signed ca-certificate. Default behavior of elasticsearch client to skip this type of connection. So I need to have a way to tune this connection [this way](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/connceting.html#_self_signed_certificates).
I believe there are more cases when we need to have a way to fine tune elasticsearch client library. So I propose to add an `options` key to elastic config.
**Example**  
```
# src/config/packages/api_platform.yaml
api_platform:
    elasticsearch:
        hosts: ['%env(ELASTICSEARCH_HOST)%']
        options:
            sslVerification: '%env(resolve:ELASTICSEARCH_CA_ROOT)%'
```
We need only a few changes to implement it. I can create a merge request for this.
WDYT?

Xusifob:
+1 on this.

It would be nice to be able to pass a handler (such as https://github.com/jeskew/amazon-es-php) also, in order to use it in AWS for example.

meyerbaptiste:
You can already do it now with a compiler pass.

For example with a handler:

```php
// src/DependencyInjection/Compiler/ElasticsearchClientHandlerCompilerPass.php
namespace App\DependencyInjection\Compiler;

use Aws\ElasticsearchService\ElasticsearchPhpHandler;
use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;

final class ElasticsearchClientHandlerCompilerPass implements CompilerPassInterface
{
    public function process(ContainerBuilder $container): void
    {
        if (!$container->hasDefinition('api_platform.elasticsearch.client')) {
            return;
        }

        $clientDefinition = $container->getDefinition('api_platform.elasticsearch.client');

        [$factoryClass, $factoryMethod] = $clientDefinition->getFactory();

        if ('build' === $factoryMethod) {
            $clientDefinition->setFactory([$factoryClass, 'fromConfig']);
        }

        $clientConfiguration = $clientDefinition->getArguments()[0] ?? [];
        $clientConfiguration['Handler'] = new Reference(ElasticsearchPhpHandler::class);

        $clientDefinition->setArgument(0, $clientConfiguration);
    }
}
```
```php
// src/Kernel.php
namespace App;

[...]

PR #4100, opened 2021-02-27
Elastic options

CvekCoding:
Resolve issue #4059


alanpoulain:
Hello, thanks for the PR. The test failures are related, could you try to fix them?

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/ElasticsearchClientPass.php
```diff
@@ -40,6 +40,12 @@ public function process(ContainerBuilder $container)
             $clientConfiguration['hosts'] = $hosts;
         }
 
+        if ($options = $container->getParameter('api_platform.elasticsearch.options')) {
+            foreach ($options as $option => $value) {
+                $clientConfiguration[$option] = $value;
+            }
+        }
```
soyuka, line 47: I'm not a fan of using an array, what options are needed are there many options possible? IMO it's not specific enough.

CvekCoding: I'm not a great expert of Elastic client, but what I know is that all the options will be passed to `Builder` which has these properties: https://github.com/elastic/elasticsearch-php/blob/e0df11da101135310f91688f17d71fa4b00a2508/src/Elasticsearch/ClientBuilder.php#L105

At this moment I need only one option `sslVerification`.

CvekCoding:
> Hello, thanks for the PR. The test failures are related, could you try to fix them?

Sure, if my PR is acceptable for you.

src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/ElasticsearchClientPass.php
soyuka: I've taken a second look and this will be just fine as the [`fromConfig`](https://github.com/elastic/elasticsearch-php/blob/e0df11da101135310f91688f17d71fa4b00a2508/src/Elasticsearch/ClientBuilder.php#L192) method will do the checks.

PR #4115, opened 2021-03-05
OpenAPI invalid securityScheme with implicit flow (fix #4079)

### LINKED ISSUE #4079: [OpenApi] Invalid securityScheme when using implicit flow

**Description**  
When oauth is enabled in the api-platform configuration all urls (token/refresh/authorization) have a default value e.g.
```
"tokenUrl": "/oauth/v2/token",
```
This leads to an invalid swagger.json if one is using the implicit-flow for example because 
```  
components.securitySchemes.oauth.flows.implicit
should NOT have additional properties
additionalProperty: tokenUrl
```

**How to reproduce**  
Configure oauth like
``` 
 oauth:
        enabled: true
        clientId: my_client
        type: 'oauth2'
        flow: implicit
        authorizationUrl: 'http://my-custom-server/openid-connect/auth'
        scopes: []
``` 

**Possible Solution**  
I think there are some ways to fix this problem. 
1. remove the default values (the fields are nullable in the OAuthFlow class)
2. add some kind of documentation that you have to set them to null if not needed/allowed in your flow
3. add validation in the OAFactory to check if the generated flow is valid ( <-- thats maybe a bit overhead )

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4079 
| License       | MIT
| Doc PR        | done
This has to change default configurated values.


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -253,9 +253,9 @@ private function addOAuthSection(ArrayNodeDefinition $rootNode): void
                         ->scalarNode('clientSecret')->defaultValue('')->info('The oauth client secret.')->end()
                         ->scalarNode('type')->defaultValue('oauth2')->info('The oauth client secret.')->end()
                         ->scalarNode('flow')->defaultValue('application')->info('The oauth flow grant type.')->end()
-                        ->scalarNode('tokenUrl')->defaultValue('/oauth/v2/token')->info('The oauth token url.')->end()
-                        ->scalarNode('authorizationUrl')->defaultValue('/oauth/v2/auth')->info('The oauth authentication url.')->end()
-                        ->scalarNode('refreshUrl')->defaultValue('/oauth/v2/refresh')->info('The oauth refresh url.')->end()
+                        ->scalarNode('tokenUrl')->defaultValue('')->info('The oauth token url.')->end()
```
alanpoulain, line 256: Isn't this a BC break?

soyuka: I don't know as it's wrong like this.

Issue #4106, opened 2021-03-01
[OpenApi] "responses" no longer works in openapi context


jkuma:
**API Platform version(s) affected**: 2.6.2

**Description**  
Responses for a ressource operation can no longer been overriden using yaml configuration as described in https://swagger.io/docs/specification/describing-responses/

It was working in api-platform versions prior to 2.6.*

**How to reproduce**  
```yaml    
unblock:
  method: POST
  path: '/tours/{id}/unblock'
  input: Toto\Tata\UI\Http\Api\Tour\UnblockTourInvoicing\UnblockTourInvoicingModel
  output: false
  controller: Toto\Tata\UI\Http\Api\Tour\UnblockTourInvoicing\UnblockTourInvoicingController
  openapi_context:
    summary: Unblock a tour.
    description: Unblock tour invoicing.
    responses:
      202:
        description: The tour was unblocked.
        content:
          application/json:
          schema:
            $ref: '#/components/schemas/tour.UnblockTourInvoicingModel'
```

PR #4116, opened 2021-03-05
OpenAPI fixes

### LINKED ISSUE #4080: [OpenApi] Response - requestBody is not of type string

**Description**  
When i validate my generated swagger.json using validator.swagger.io  it is not valid because all write operations (post/patch/put) have an element "requestBody" with an empty array as value under the "links" node in responses.

Here is a snipped

``` 
      "patch": {
        "operationId": "patchNoteItem",
        "tags": [
          "Note"
        ],
        "responses": {
          "200": {
            "description": "Note resource updated",
            "content": {
              "application/ld+json": {
                "schema": {
                  "$ref": "#/components/schemas/Note.jsonld-read"
                }
              }
            },
            "links": {
              "GetNoteItem": {
                "operationId": "getNoteItem",
                "parameters": {
                  "uuid": "$response.body#/uuid"
                },
                "requestBody": [],  <--- THIS ONE
                "description": "The `uuid` value returned in the response can be used as the `uuid` parameter in `GET /api/v1/notes/{uuid}`."
              }
            }
          }, 
``` 

I'm not sure if it CAN be an array under this node but i think it has to be string if i understood https://swagger.io/docs/specification/links/ correctly.

I dont think it is used at the moment, its set to [] here --> https://github.com/api-platform/core/blob/b5805a17301db78f10bff92f937b24a2bb6049b3/src/OpenApi/Factory/OpenApiFactory.php#L358

**Possible Solution**  
The value is nullable in the Link model and that would be my suggestion. Set the default to null.

I can setup a PR but i would love to hear what @soyuka thinks since he set this default.

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4080 #4106
| License       | MIT
| Doc PR        | na

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 * OpenAPI: Fix error when schema is empty (#4051)
 * OpenAPI: Do not set scheme to oauth2 when generating securitySchemes (#4073)
 * OpenAPI: Fix missing `$ref` when no `type` is used in context (#4076)
+* OpenAPI: Fix `response` support via the `openapi_context` (#4116)
+* OpenAPI: Fix `Link->requestBody` default value (#4116)
 * GraphQL: Fix "Resource class cannot be determined." error when a null iterable field is returned (#4092)
 * Metadata: Check the output class when calculating serializer groups (#3696)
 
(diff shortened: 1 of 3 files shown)


alanpoulain (approved):

alanpoulain:
Thanks @soyuka.

nicodmf:
@soyuka @alanpoulain could we expect a tag for there bug fixes ? we are blocked by those.

alanpoulain:
@dunglas is responsible of the releases. I think a release will be done soon but we can't tell you exactly when.

PR #4117, opened 2021-03-05
Support psr/cache 2 and 3

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

The `psr/cache` interfaces have been updated with type declarations. This change should not make a difference for API platform because it only consumes cache implementations and does not provide its own cache implementations.

This PR also contains changes to a couple of test cases related to caching: Prophecy seems to be confused by the fact that the updated `CacheException` interface extends `Throwable`. Since the affected mocks can easily be replaced with a simple anonymous class, I did just that.


alanpoulain (approved):

dunglas (approved):

alanpoulain:
Thanks @derrabus!

PR #4122, opened 2021-03-08
Update OpenApi PathItem nullable operation test

d0niek:
Previous test didn't test `with*` methods because they return new
(cloned) instance of `$pathItem`. In assertions we test nulls from
`PathItem::_construct()` instead of from `with*` methods.


alanpoulain (approved):

alanpoulain:
Thank you @d0niek.

PR #4123, opened 2021-03-09
Do not throw for invalid IRI if type enforcement is disabled

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | na
| License       | MIT
| Doc PR        | tbd

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -228,20 +228,25 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                     throw new UnexpectedValueException($e->getMessage(), $e->getCode(), $e);
                 }
             } catch (InvalidArgumentException $e) {
-                if (!$supportsPlainIdentifiers) {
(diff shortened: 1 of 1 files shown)


soyuka:
We discussed this @norkunas and it feels wrong:
- we should never disable the `type enforcement` on api p
- a better approach is to finish the symfony's patch on the serializer that gives you the ability to collect every errors (there was a way to get the partially hydrated object)
- if the data is invalid, there's no need to do validation (eg if the IRI is wrong it's probably a software issue and should not happend)
- the @Assert\Type should not be used unless you have very specific use cases (multiple types) and it's here for historical reasons.

feel free to comment back as we can still discuss this!

norkunas:
I understand, but type enforcement is not enabled by default, it's an option for the users, so it feels wrong that it is ignored :)
Because for me there is no way to delegate the validation to validator :)

soyuka:
I would like to find a better solution we need to discuss this somewhere!

PR #4125, opened 2021-03-09
Allow serializing when property and method start from "is"

maks-rafalko:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Follow up for https://github.com/api-platform/core/pull/4064/files

In #4064, there was a fix for **documentation** generated on UI. But, unfortunately, when we use `public function isActive()` method and `$isActive` property, the value is not being serialized - it is not included in the response.

The reason is the same - `$options` array does not contain `'enable_getter_setter_extraction' => true` default value.

I have some issues with the tests, probably you can help me with understanding what's going on, will comment below.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -517,7 +517,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
      */
     protected function getFactoryOptions(array $context): array
     {
-        $options = [];
+        $options = [
+            /* @see https://github.com/symfony/symfony/blob/v5.1.0/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php */
+            'enable_getter_setter_extraction' => true,
+        ];
 
         if (isset($context[self::GROUPS])) {
             /* @see https://github.com/symfony/symfony/blob/v4.2.6/src/Symfony/Component/PropertyInfo/Extractor/SerializerExtractor.php */
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -384,7 +384,7 @@ jobs:
       - name: Clear test app cache
         run: tests/Fixtures/app/console cache:clear --ansi
       - name: Run Behat tests
-        run: vendor/bin/behat --out=std --format=progress --profile=default --no-interaction
+        run: vendor/bin/behat --out=std --format=progress --profile=default --no-interaction --tags '~@!lowest'
 
   postgresql:
     name: Behat (PHP ${{ matrix.php }}) (PostgreSQL)
--- a/features/serializer/boolean_properties.feature
+++ b/features/serializer/boolean_properties.feature
@@ -1,3 +1,4 @@
+@!lowest
 Feature: Serialization of boolean properties
   In order to use an hypermedia API
   As a client software developer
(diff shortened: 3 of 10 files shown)


features/main/dummy_boolean.feature
```diff
@@ -0,0 +1,30 @@
...
+    And the header "Location" should be equal to "/dummy_booleans/1"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/DummyBoolean",
+      "@id": "/dummy_booleans/1",
+      "@type": "DummyBoolean",
+      "id": 1,
+      "isDummyBoolean": true,
+      "dummyBoolean": true
```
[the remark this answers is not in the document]
alanpoulain: I think it's more a normalizer issue and it should be fixed in Symfony itself.
I saw it too when debugging the previous PR.

features/main/dummy_boolean.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Serialization of boolean properties
```
alanpoulain, line 1: Wouldn't it be better in `serializer`?

alanpoulain:
The unit tests need to be fixed too.

tests/Fixtures/TestBundle/Entity/DummyBoolean.php
```diff
@@ -38,11 +38,6 @@ class DummyBoolean
      */
     private $isDummyBoolean;
 
-    public function __construct(bool $isDummyBoolean)
-    {
-        $this->isDummyBoolean = $isDummyBoolean;
-    }
-
```
soyuka, line 45: we should avoid changing old tests

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -517,7 +517,10 @@ protected function denormalizeRelation(string $attributeName, PropertyMetadata $
      */
     protected function getFactoryOptions(array $context): array
     {
-        $options = [];
+        $options = [
+            /* @see https://github.com/symfony/symfony/blob/v5.1.0/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php */
+            'enable_getter_setter_extraction' => true,
```
soyuka, line 522: wondering if this should be opt-in, does this break existing tests?

alanpoulain: No it doesn't break existing tests. However it can be considered as a BC break.

alanpoulain:
I think this feature should probably target 2.7 and can be considered as a BC break.
For instance if you have an entity like this:
```php
<?php

/**
 * @ApiResource
 * @ORM\Entity
 */
class Test
{
    /**
     * @var int
     *
     * @ORM\Column(type="integer", nullable=true)
     * @ORM\Id
     */
    private $id;

    /**
     * @var bool
     *
     * @ORM\Column(type="boolean", nullable=true)
     */
    private $last;

    public function getId()
    {
        return $this->id;
    }

    public function last(): bool
    {
        return $this->last;
    }
}
```

Before it was not serialized and now it will be. WDYT @maks-rafalko @soyuka?

alanpoulain:
As @soyuka said we could make it opt-in too.
We should change what https://github.com/api-platform/core/pull/4064 is doing too to reflect what we do here.

maks-rafalko:
@stale 

It still the issue, however I don't have time/energy to complete it. If someone can help, that would be awesome.

PR #4134, opened 2021-03-12
SearchFilter: fix regression in 2.6.3

### LINKED ISSUE #4128: BC Break on SearchFilter with date as input
**API Platform version(s) affected**: x.y.z

**Description**  

If you use dates with the search filter, this was supported, it now triggers an exception. This is because previously the type was not always sent with the doctrine parameter, and doctrine was considering the date as string type (which is valid when you have a string like `2021-01-12` as input).

**How to reproduce**  

If you have the following filter configured on an entity, it will break:

```php
/**
 * ...
 * @ApiFilter(SearchFilter::class, properties={"something.id": "exact", "date": "exact"})
 */
```

**Possible Solution**  

Add support for dates, which is a special case that is not that simple because it requires to manage `date`, `datetime`, `datetime_immutable` and so on.

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4128 #4135 
| License       | MIT
| Doc PR        | na


src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -145,28 +143,30 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
         $associationAlias = $alias;
         $associationField = $field;
+        $valueParameter = $queryNameGenerator->generateParameterName($associationField);
         if ($metadata->isCollectionValuedAssociation($associationField)) {
             $associationAlias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $alias, $associationField);
             $associationField = $associationFieldIdentifier;
         }
 
-        $this->addWhereByStrategy($strategy, $queryBuilder, $queryNameGenerator, $associationAlias, $associationField, $values, $caseSensitive, $metadata);
+        if (1 === \count($values)) {
```
alanpoulain, line 152: It has been refactored, we shouldn't revert it.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -175,26 +175,18 @@ protected function addWhereByStrategy(string $strategy, QueryBuilder $queryBuild
         $valueParameter = ':'.$queryNameGenerator->generateParameterName($field);
         $aliasedField = sprintf('%s.%s', $alias, $field);
 
-        if (self::STRATEGY_EXACT === $strategy) {
-            if (1 === \count($values)) {
+        if (null == $strategy || self::STRATEGY_EXACT == $strategy) {
+            if (1 == \count($values)) {
```
alanpoulain, line 179: We shouldn't revert this part.

CHANGELOG.md
```diff
@@ -13,7 +13,6 @@
 * Security: Use a `NullToken` when using the new authenticator manager in the resource access checker (#4067)
 * Mercure: Do not use data in options when deleting (#4056)
 * Doctrine: Support for foreign identifiers (#4042)
-* Doctrine: Support for binary UUID in search filter (#3774)
```
alanpoulain, line 16: We should probably keep this line and add a new entry saying that it has been reverted.

src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
soyuka: I don't think so, 2 patches are on this file: 
https://github.com/api-platform/core/commit/9df5a5e3fe438cd265c20bd330f82bce8ef6e709 and https://github.com/api-platform/core/commit/414351b14a7a491d1c7ddfaaaa09254f49652b72 none touch this line ? :/

alanpoulain:
Thanks @soyuka.

odoucet:
I'll try to provide a new MR that fixes the original issue (search on UUID) and being compatible with dates.

alanpoulain:
@odoucet Not just with dates, it needs to be compatible with different kind of Doctrine types (enum, etc.).
I don't think there is a proper solution other than a whitelist (and we don't want to have it).
I think the only way is to have a custom specific search filter, only for the binary uuid type.
I'm not sure we would want it in the core but you can try :slightly_smiling_face:

masacc:
Hello @soyuka, this rollback creates a BC break : when searching for a string that is an integer, it causes this error with SQL Server :

```
SQLSTATE[22018]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Conversion failed when converting the varchar value '%' to data type int.
```

it works fine on 2.6.3

Issue #4124, opened 2021-03-09
[OpenApi] Explode attributes in parameters is not right normalized and cannot be overriden


tyx:
**API Platform version(s) affected**: 2.6.x

**Description**  

The new OpenApi implementation is really nice ! 

It just seems the `explode` attributes of `Model\Parameters` is not right normalized : 
https://github.com/api-platform/core/blob/4717bd5643d7761707229ed8705ae6824248f7e3/src/OpenApi/Model/Parameter.php#L101-L104

Replacing by  `getExplode` make it displaying in the swagger ui.

And moreover, it cannot be overriden

https://github.com/api-platform/core/blob/4717bd5643d7761707229ed8705ae6824248f7e3/src/OpenApi/Factory/OpenApiFactory.php#L383-L397

What about something like : 

```php 
$parameters[] = new Model\Parameter(
                    $name,
                    'query',
                    $data['description'] ?? '',
                    $data['required'] ?? false,
                    $data['openapi']['deprecated'] ?? false,
                    $data['openapi']['allowEmptyValue'] ?? true,
                    $schema,
                    'array' === $schema['type'] && \in_array($data['type'],
                    [Type::BUILTIN_TYPE_ARRAY, Type::BUILTIN_TYPE_OBJECT], true) ? 'deepObject' : 'form',
                    $data['openapi']['explode'] ?? ('array' === $schema['type']),
                    $data['openapi']['allowReserved'] ?? false,
                    $data['openapi']['example'] ?? null,
                    isset($data['openapi']['examples']
                ) ? new \ArrayObject($data['openapi']['examples']) : null);
```

**How to reproduce**  

Just drop a filter and watch the swagger-data in the swagger ui.

**Possible Solution**  

Not sure if both suggestion are the best. But if we think they are, I'm open to make a PR.

soyuka:
:+1: on the suggestion but you should reuse the `'array' === $schema['type'] && \in_array($data['type'], 
     [Type::BUILTIN_TYPE_ARRAY, Type::BUILTIN_TYPE_OBJECT], true)` test

vincentchalamon:
Fixed in 2.6.4

PR #4138, opened 2021-03-15
OpenAPI: Make sure we do not override defined parameters #4137

### LINKED ISSUE #4137: Overwriting parameters using annotation
**API Platform version(s) affected**: 2.6.2

**Description**  
When i try to overwrite parameters for an item operation using annotations, its added again in the swagger_docs.json instead of replacing the existing default one.

**How to reproduce**  
Here's how the ApiResource parameters annotation looks like:
```
 * @ApiResource(
 *     itemOperations={
 *         "get"={
 *             "openapi_context"={
 *                 "parameters"={
 *                     {
 *                         "name"="id",
 *                         "in"="path",
 *                         "required"=true,
 *                         "schema"={
 *                             "type"="string",
 *                             "format"="uuid"
 *                         }
 *                     }
 *                 }
 *             }
 *         }
 * )
 */
```

Here's how it looks after i generate the swagger_docs.json (parameters appear two times here):
```
                "parameters": [
                    {
                        "name": "id",
                        "in": "path",
                        "description": "",
                        "required": true,
                        "deprecated": false,
                        "schema": {
                            "type": "string",
                            "format": "uuid"
                        },
                        "style": "simple"
                    },
                    {
                        "name": "id",
                        "in": "path",
                        "description": "Resource identifier",
                        "required": true,
                        "deprecated": false,
                        "schema": {
                            "type": "string"
                        },
                        "style": "simple"
                    }
                ],
```

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4137 #4124
| License       | MIT
| Doc PR        | na

Improves performances by 60% https://blackfire.io/profiles/compare/d6d1c64b-ec39-4ac9-ac9f-630dac371da4/graph

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -175,19 +175,42 @@ final class OpenApiFactory implements OpenApiFactoryInterface
             // Set up parameters
             if (OperationType::ITEM === $operationType) {
                 foreach ($identifiers as $parameterName => $identifier) {
-                    $parameters[] = new Model\Parameter(\is_string($parameterName) ? $parameterName : $identifier, 'path', 'Resource identifier', true, false, false, ['type' => 'string']);
+                    $parameterName = \is_string($parameterName) ? $parameterName : $identifier;
+                    $parameter = new Model\Parameter($parameterName, 'path', 'Resource identifier', true, false, false, ['type' => 'string']);
+                    if ($this->hasParameter($parameter, $parameters)) {
+                        continue;
+                    }
+
+                    $parameters[] = $parameter;
                 }
                 $links[$operationId] = $this->getLink($resourceClass, $operationId, $path);
             } elseif (OperationType::COLLECTION === $operationType && 'GET' === $method) {
-                $parameters = array_merge($parameters, $this->getPaginationParameters($resourceMetadata, $operationName), $this->getFiltersParameters($resourceMetadata, $operationName, $resourceClass));
+                foreach (array_merge($this->getPaginationParameters($resourceMetadata, $operationName), $this->getFiltersParameters($resourceMetadata, $operationName, $resourceClass)) as $parameter) {
+                    if ($this->hasParameter($parameter, $parameters)) {
+                        continue;
+                    }
+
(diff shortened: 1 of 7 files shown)


src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -486,4 +511,25 @@ private function getSecuritySchemes(): array
 
         return $securitySchemes;
     }
+
+    private function appendSchemaDefinitions(\ArrayObject &$schemas, \ArrayObject $definitions): void
```
alanpoulain, line 515: Is the reference necessary?

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -486,4 +511,25 @@ private function getSecuritySchemes(): array
...
+    {
+        foreach ($definitions as $key => $value) {
+            $schemas[$key] = $value;
+        }
+    }
+
+    /**
+     * @var Model\Parameter[]
```
alanpoulain, line 523: Not needed?

src/OpenApi/Factory/OpenApiFactory.php
soyuka: I think it's more performant like this indeed

src/OpenApi/Factory/OpenApiFactory.php
soyuka: Yes or else I need to type `existingParameter` :)

alanpoulain (approved):

alanpoulain:
Thank you!

src/OpenApi/Model/Parameter.php
```diff
@@ -83,7 +83,7 @@ public function getDeprecated(): bool
         return $this->deprecated;
     }
 
-    public function canAllowEmptyValue(): bool
+    public function getAllowEmptyValue(): bool
```
guilliamxavier, line 86: BC break? Did you intend to duplicate (like the other ones below) rather than rename the existing?

PR #4139, opened 2021-03-16
Handle sequentially validator constraints when generating property metadata

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT
| Doc PR        | -

Currently property metadata from validator constraints (length/regex etc.) is only taken when these constraints are applied directly. Since Symfony 5.1 there is Sequentially constraint which allows to validate as a group sequence but much easier. But api platform doesn't take them because they're nested inside the Sequentially constraint.

~~I have updated to include all nested constraints from the Composite, but maybe that's too wide?
Also not sure about that maybe constraint flattening should be recursive?~~

### DIFF
--- a/src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php
@@ -30,6 +30,7 @@ use Symfony\Component\Validator\Constraints\Isbn;
 use Symfony\Component\Validator\Constraints\Issn;
 use Symfony\Component\Validator\Constraints\NotBlank;
 use Symfony\Component\Validator\Constraints\NotNull;
+use Symfony\Component\Validator\Constraints\Sequentially;
 use Symfony\Component\Validator\Constraints\Time;
 use Symfony\Component\Validator\Constraints\Url;
 use Symfony\Component\Validator\Constraints\Uuid;
@@ -170,11 +171,15 @@ final class ValidatorPropertyMetadataFactory implements PropertyMetadataFactoryI
             }
 
             foreach ($validatorPropertyMetadata->findConstraints($validationGroup) as $propertyConstraint) {
-                $constraints[] = $propertyConstraint;
+                if ($propertyConstraint instanceof Sequentially) {
+                    $constraints[] = $propertyConstraint->getNestedContraints();
+                } else {
+                    $constraints[] = [$propertyConstraint];
+                }
             }
         }
 
-        return $constraints;
+        return array_merge([], ...$constraints);
     }
 
     /**
--- a/tests/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactoryTest.php
+++ b/tests/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactoryTest.php
@@ -20,11 +20,13 @@ use ApiPlatform\Core\Bridge\Symfony\Validator\Metadata\Property\ValidatorPropert
 use ApiPlatform\Core\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use ApiPlatform\Core\Tests\Fixtures\DummyIriWithValidationEntity;
+use ApiPlatform\Core\Tests\Fixtures\DummySequentiallyValidatedEntity;
 use ApiPlatform\Core\Tests\Fixtures\DummyValidatedEntity;
 use ApiPlatform\Core\Tests\ProphecyTrait;
 use Doctrine\Common\Annotations\AnnotationReader;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Validator\Constraints\Sequentially;
 use Symfony\Component\Validator\Mapping\ClassMetadata;
 use Symfony\Component\Validator\Mapping\Factory\MetadataFactoryInterface;
 use Symfony\Component\Validator\Mapping\Loader\AnnotationLoader;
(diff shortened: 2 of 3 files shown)


src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php
```diff
@@ -170,11 +171,15 @@ private function getPropertyConstraints(
             }
 
             foreach ($validatorPropertyMetadata->findConstraints($validationGroup) as $propertyConstraint) {
-                $constraints[] = $propertyConstraint;
+                if ($propertyConstraint instanceof Composite) {
+                    $constraints[] = $propertyConstraint->getNestedContraints();
```
alanpoulain, line 175: Wouldn't be better?
```suggestion
                    $constraints = array_merge($constraints, $propertyConstraint->getNestedContraints());
                    continue;
```

And to add back the rest like before?

norkunas: https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#slow-array-function-used-in-loop

alanpoulain:
Hello,
Thanks for your PR, seems legit to me.
What do you mean by:
> I have updated to include all nested constraints from the Composite, but maybe that's too wide?

norkunas:
> Hello,
> Thanks for your PR, seems legit to me.
> What do you mean by:
> 
> > I have updated to include all nested constraints from the Composite, but maybe that's too wide?

For example there is an `AtLeastOneOf` constraint which is also Composite, but it has a different use case than Sequentially..

src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php
soyuka: :+1: never use array_merge in a loop

alanpoulain:
> For example there is an `AtLeastOneOf` constraint which is also Composite, but it has a different use case than Sequentially..

I see. It seems OK to me to use all the nested constraints of `AtLeastOneOf` too.

Shouldn't it be recursive instead of a loop? It seems to me that you can combine the composite constraints.

src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php
alanpoulain: OK!

norkunas:
> > For example there is an `AtLeastOneOf` constraint which is also Composite, but it has a different use case than Sequentially..
> 
> I see. It seems OK to me to use all the nested constraints of `AtLeastOneOf` too.
> 
> Shouldn't it be recursive instead of a loop? It seems to me that you can combine the composite constraints.

So you mean I should do recursive constraint resolving? :)

alanpoulain:
Yes, does it seem correct to you?

norkunas:
> Yes, does it seem correct to you?

Maybe.. added a second commit with flattening

src/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactory.php
```diff
@@ -190,4 +190,28 @@ private function isRequired(Constraint $constraint): bool
...
+     *
+     * @return Constraint[]
+     */
+    private function flattenConstraints(array $propertyConstraints): array
+    {
+        $constraints = [];
+
+        foreach ($propertyConstraints as $propertyConstraint) {
+            if ($propertyConstraint instanceof AtLeastOneOf) {
+                continue;
```
alanpoulain, line 205: Why continuing in this case? I think we should take the constraints of `AtLeastOneOf` too for the property metadata, don't you think?

norkunas: Hmm.. Not sure for real, because:
```
@Assert\AtLeastOneOf({
    @Assert\Length(min=1, max=2),
    @Assert\Length(min=5, max=6)
})
```
The last one would override the first one and then the schema would be wrong :thinking:

norkunas:
Also with the flattening I think this could go wrong, because the restrictions would be directly applied to the property, not to the array values
```
@Assert\Sequentially({
    @Assert\Type(type="array"),
    @Assert\All({
        @Assert\Length(min=1, max=32),
        @Assert\Regex(pattern="/^[a-z]$/")
    })
})
```

alanpoulain:
> Also with the flattening I think this could go wrong, because the restrictions would be directly applied to the property, not to the array values
> 
> ```
> @Assert\Sequentially({
>     @Assert\Type(type="array"),
>     @Assert\All({
>         @Assert\Length(min=1, max=32),
>         @Assert\Regex(pattern="/^[a-z]$/")
>     })
> })
> ```

You're right, I think we should blacklist (or maybe it would be preferable to whitelist the wanted constraints) `All` and `Collection`.

norkunas:
Maybe for now would be good to handle the `Sequentially` at the first level and in future if someone come up with another use cases then could make follow-up PR?

alanpoulain:
Yes, please do, sorry for the unwanted change.
I think the `AtLeastOneOf` constraint is manageable with https://json-schema.org/understanding-json-schema/reference/combining.html#oneof, but we can do it in another PR.

norkunas:
> Yes, please do, sorry for the unwanted change.
> I think the `AtLeastOneOf` constraint is manageable with https://json-schema.org/understanding-json-schema/reference/combining.html#oneof, but we can do it in another PR.

All good :) I agree, one problem at a time :)

tests/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactoryTest.php
```diff
@@ -335,4 +337,35 @@ public function testCreateWithPropertyFormatRestriction(): void
             $this->assertEquals($format, $schema['format']);
         }
     }
+
+    public function testCreateWithCompositeConstraint(): void
```
alanpoulain, line 341:
```suggestion
    public function testCreateWithSequentiallyConstraint(): void
```
To be clearer :slightly_smiling_face:

norkunas: Thanks, updated :)

alanpoulain:
I'm checking if we shouldn't merge this PR as a bugfix.

norkunas:
> I'm checking if we shouldn't merge this PR as a bugfix.

@soyuka said to submit as a feature :slightly_smiling_face:

chalasr:
While this looks like an enhancement (as it's adding support for a Symfony feature that is currently not supported), I think it qualifies as a bugfix as it is actually fixing compatibility with Symfony 5.1 (at least one way to use Symfony 5.1). 
Merging in 2.6 would probably be helpful for developers embracing Symfony 5.1 features and does not seem harmful in any way.

soyuka:
Thanks @norkunas !

norkunas:
Thanks for merging :)

PR #4142, opened 2021-03-18
upgrade to mercure-bundle:^0.3

azjezz:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no 
| New feature?  | yes
| Deprecations? | yes
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -68,7 +68,7 @@ final class PublishMercureUpdatesListener
 
     /**
      * @param array<string, string[]|string> $formats
-     * @param Mercure|callable $publisher
+     * @param Mercure|callable $mercure
      */
     public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, $mercure = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
     {
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -68,7 +68,7 @@ final class PublishMercureUpdatesListener
 
     /**
      * @param array<string, string[]|string> $formats
-     * @param Mercure|callable $mercure
+     * @param Mercure|callable               $mercure
      */
     public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, $mercure = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
     {
@@ -239,7 +239,7 @@ final class PublishMercureUpdatesListener
             if ($this->mercure instanceof Mercure) {
                 $this->mercure->getPublisher($options['hub'] ?? null)->publish($update);
             } else {
-                $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+                $this->messageBus ? $this->dispatch($update) : ($this->mercure)($update);
             }
         }
     }
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -26,6 +26,7 @@ use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Doctrine\Common\EventArgs;
 use Doctrine\ODM\MongoDB\Event\OnFlushEventArgs as MongoDbOdmOnFlushEventArgs;
 use Doctrine\ORM\Event\OnFlushEventArgs as OrmOnFlushEventArgs;
+use Symfony\Bundle\MercureBundle\Mercure;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\Mercure\Update;
@@ -51,11 +52,12 @@ final class PublishMercureUpdatesListener
         'type' => true,
         'retry' => true,
         'normalization_context' => true,
+        'hub' => true,
     ];
 
     private $iriConverter;
     private $serializer;
-    private $publisher;
+    private $mercure;
     private $expressionLanguage;
     private $createdObjects;
     private $updatedObjects;
@@ -66,11 +68,16 @@ final class PublishMercureUpdatesListener
 
     /**
      * @param array<string, string[]|string> $formats
+     * @param Mercure|callable $publisher
      */
-    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
+    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, $mercure = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
     {
-        if (null === $messageBus && null === $publisher) {
-            throw new InvalidArgumentException('A message bus or a publisher must be provided.');
+        if (null === $messageBus && null === $mercure) {
+            throw new InvalidArgumentException('A message bus or a mercure instance must be provided.');
+        }
+
+        if ($mercure && !$mercure instanceof Mercure) {
+            @trigger_error(sprintf('Passing a callable as the seventh argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Mercure::class), \E_USER_DEPRECATED);
         }
 
         $this->resourceClassResolver = $resourceClassResolver;
@@ -79,7 +86,7 @@ final class PublishMercureUpdatesListener
         $this->serializer = $serializer;
         $this->formats = $formats;
         $this->messageBus = $messageBus;
-        $this->publisher = $publisher;
+        $this->mercure = $mercure;
         $this->expressionLanguage = $expressionLanguage ?? (class_exists(ExpressionLanguage::class) ? new ExpressionLanguage() : null);
         $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
         $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
@@ -188,6 +195,10 @@ final class PublishMercureUpdatesListener
             if (!isset(self::ALLOWED_KEYS[$key])) {
                 throw new InvalidArgumentException(sprintf('The option "%s" set in the "mercure" attribute of the "%s" resource does not exist. Existing options: "%s"', $key, $resourceClass, implode('", "', self::ALLOWED_KEYS)));
             }
+
+            if ('hub' === $key && !$this->mercure instanceof Mercure) {
+                throw new InvalidArgumentException(sprintf('The option "hub" set in the "mercure" attribute of the "%s" resource is not supported. Please install mercure-bundle:^0.3 to enable it.', $resourceClass));
+            }
         }
 
         if ('deletedObjects' === $property) {
@@ -225,7 +236,11 @@ final class PublishMercureUpdatesListener
         $updates = array_merge([$this->buildUpdate($iri, $data, $options)], $this->getGraphQlSubscriptionUpdates($object, $options, $type));
 
         foreach ($updates as $update) {
-            $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+            if ($this->mercure instanceof Mercure) {
+                $this->mercure->getPublisher($options['hub'] ?? null)->publish($update);
+            } else {
+                $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+            }
         }
     }
 
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -77,7 +77,7 @@ final class PublishMercureUpdatesListener
         }
 
         if ($registry && !$registry instanceof HubRegistry) {
-            @trigger_error(sprintf('Passing a callable as the seventh argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Mercure::class), \E_USER_DEPRECATED);
+            @trigger_error(sprintf('Passing a callable as the seventh argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, HubRegistry::class), \E_USER_DEPRECATED);
         }
 
         $this->resourceClassResolver = $resourceClassResolver;
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -26,9 +26,9 @@ use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Doctrine\Common\EventArgs;
 use Doctrine\ODM\MongoDB\Event\OnFlushEventArgs as MongoDbOdmOnFlushEventArgs;
 use Doctrine\ORM\Event\OnFlushEventArgs as OrmOnFlushEventArgs;
-use Symfony\Bundle\MercureBundle\Mercure;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Mercure\HubRegistry;
 use Symfony\Component\Mercure\Update;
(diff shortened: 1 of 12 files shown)


azjezz:
waiting for https://github.com/symfony/mercure/pull/40

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -67,19 +68,23 @@ final class PublishMercureUpdatesListener
     /**
      * @param array<string, string[]|string> $formats
      */
-    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
-    {
-        if (null === $messageBus && null === $publisher) {
-            throw new InvalidArgumentException('A message bus or a publisher must be provided.');
-        }
-
+    public function __construct(
```
alanpoulain, line 71: This one is a BC break, isn't it?

azjezz: yes, not sure how we can avoid BC break here 🤔 

composer.json allows mercure-bundle: "*", but i think previous versions should conflict with 0.5

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -177,11 +182,7 @@ private function storeObjectToPublish($object, string $property): void
 
         foreach ($options as $key => $value) {
             if (0 === $key) {
-                if (method_exists(Update::class, 'isPrivate')) {
-                    throw new \InvalidArgumentException('Targets do not exist anymore since Mercure 0.10. Mark the update as private instead or downgrade the Mercure Component to version 0.3');
-                }
-
-                @trigger_error('Targets do not exist anymore since Mercure 0.10. Mark the update as private instead.', \E_USER_DEPRECATED);
```
alanpoulain, line 184: Why removing the deprecation?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -259,12 +260,6 @@ private function getGraphQlSubscriptionUpdates($object, array $options, string $
...
     private function buildUpdate($iri, string $data, array $options): Update
     {
-        if (method_exists(Update::class, 'isPrivate')) {
-            return new Update($iri, $data, $options['private'] ?? false, $options['id'] ?? null, $options['type'] ?? null, $options['retry'] ?? null);
-        }
-
-        // Mercure Component < 0.4.
-        /* @phpstan-ignore-next-line */
-        return new Update($iri, $data, $options);
+        return new Update($iri, $data, $options['private'] ?? false, $options['id'] ?? null, $options['type'] ?? null, $options['retry'] ?? null);
```
alanpoulain, line 263: Why removing the BC layer?

azjezz: this PR drops support for older mercure/mercure-bundle versions. do you think we should keep them? 🤔

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -262,7 +263,8 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
 
         if (!empty($config['resource_class_directories'])) {
             $container->setParameter('api_platform.resource_class_directories', array_merge(
-                $config['resource_class_directories'], $container->getParameter('api_platform.resource_class_directories')
+                $config['resource_class_directories'],
```
alanpoulain, line 266: We should avoid to change the code style on `main` (we can do it in `2.6`): conflicts will be harder to solve when merging `2.6` into `main`.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -225,7 +226,7 @@ private function publishUpdate($object, array $options, string $type): void
         $updates = array_merge([$this->buildUpdate($iri, $data, $options)], $this->getGraphQlSubscriptionUpdates($object, $options, $type));
 
         foreach ($updates as $update) {
-            $this->messageBus ? $this->dispatch($update) : ($this->publisher)($update);
+            $this->mercure->getPublisher($options['hub'] ?? null)->publish($update);
```
azjezz, line 229: @dunglas do we wanna keep support for message bus here?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
alanpoulain: We can add a conflict if necessary but it will not solve the issue of not adding BC break in this listener.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
alanpoulain: We can drop it if it's really necessary but we should add a conflict then.

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -26,15 +24,15 @@
...
-
     private $resourceMetadataFactory;
-    private $hub;
+    private $discovery;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, string $hub)
-    {
+    public function __construct(
+        ResourceMetadataFactoryInterface $resourceMetadataFactory,
+        Discovery $discovery
```
alanpoulain, line 32: This one is a BC break too.

src/GraphQl/Subscription/MercureSubscriptionIriGenerator.php
```diff
@@ -25,12 +26,12 @@
 final class MercureSubscriptionIriGenerator implements MercureSubscriptionIriGeneratorInterface
 {
     private $requestContext;
-    private $hub;
+    private $mercure;
 
-    public function __construct(RequestContext $requestContext, string $hub)
+    public function __construct(RequestContext $requestContext, Mercure $mercure)
```
alanpoulain, line 31: This one is a BC break but since the GraphQL system is experimental, it is allowed (but it would be better to avoid it).

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -43,26 +41,17 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
     public function onKernelResponse(ResponseEvent $event): void
     {
         $request = $event->getRequest();
-        // Prevent issues with NelmioCorsBundle
-        if ($this->isPreflightRequest($request)) {
```
alanpoulain, line 47: Why removing this?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
azjezz: if we are adding a BC layer, i don't think there's a need to add a conflict.

alanpoulain: Yes, not for this change anyway.

src/Mercure/EventListener/AddLinkHeaderListener.php
azjezz: Discovery service already checks for this.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
soyuka: good question, as the class is experimental I think that we can do pretty much what we want :)

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
soyuka: indeed this is the main issue, as API Platform is a library we don't have any `compose.lock` and therefore users may have issues installing api platform with an incompatible mercure-bundle

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
alanpoulain: Oh I didn't see that this listener was experimental. Sorry, you can indeed add a change like this here.

soyuka: IMO we need a BC layer

src/GraphQl/Resolver/Factory/ItemSubscriptionResolverFactory.php
```diff
@@ -77,11 +77,13 @@ public function __invoke(?string $resourceClass = null, ?string $rootClass = nul
 
             $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);
 
-            if ($subscriptionId && $resourceMetadata->getAttribute('mercure', false)) {
+            if ($subscriptionId && ($mercureConfigurations = $resourceMetadata->getAttribute('mercure', false))) {
```
alanpoulain, line 80: Why the plural?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
azjezz: the thing with messenger is that we are not gaining anything from using it, we are replacing an http request, but a request to rabbitmq, redis, mysql server .. etc. which would later result in an http request.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
azjezz: I think the best choice would be to add a BC layer, and drop it in the next major ( however, we should add a conflict with mercure-bundle > 0.5 for future releases, otherwise we would end up with a pile of BC layers to deal with ).

alanpoulain: I'm :+1: for adding a conflict with mercure-bundle < 0.6 for `2.7` (it needs to be documented in the CHANGELOG).

alanpoulain: The last version of the Mercure bundle is the 0.2.6, isn't it? I'm lost :sweat_smile:

azjezz: yea 😆 i meant to say 0.3, the last version of mercure component is 0.4, next ( the one compatible with mercure-bundle 0.3 ) is 0.5

alanpoulain: I think we should at least have a BC layer for the previous version (so in this case the `0.2` version of Mercure bundle) to allow a smooth upgrade path.

azjezz: make sense 👍

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
azjezz: reverted.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -66,11 +68,16 @@ final class PublishMercureUpdatesListener
 
     /**
      * @param array<string, string[]|string> $formats
+     * @param Mercure|callable $publisher
```
alanpoulain, line 71:
```suggestion
     * @param Mercure|callable $mercure
```

src/GraphQl/Subscription/MercureSubscriptionIriGenerator.php
```diff
@@ -25,12 +26,19 @@
...
+    private $mercure;
 
-    public function __construct(RequestContext $requestContext, string $hub)
+    /**
+     * @param Mercure|string $mercure
+     */
+    public function __construct(RequestContext $requestContext, $mercure)
     {
+        if (is_string($mercure)) {
+            @trigger_error(sprintf('Passing a string as the seconds argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Mercure::class), \E_USER_DEPRECATED);
```
alanpoulain, line 37:
```suggestion
            @trigger_error(sprintf('Passing a string as the second argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Mercure::class), \E_USER_DEPRECATED);
```

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -29,12 +30,19 @@ final class AddLinkHeaderListener
...
+    private $discovery;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, string $hub)
+    /**
+     * @param Discovery|string
+     */
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, $discovery)
     {
+        if (is_string($discovery)) {
+            @trigger_error(sprintf('Passing a string as the seconds argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Discovery::class), \E_USER_DEPRECATED);
```
alanpoulain, line 41:
```suggestion
            @trigger_error(sprintf('Passing a string as the second argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Discovery::class), \E_USER_DEPRECATED);
```

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -43,26 +51,32 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
...
-            false === $this->resourceMetadataFactory->create($resourceClass)->getAttribute('mercure', false)
+            false === ($mercure = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('mercure', false))
         ) {
             return;
         }
 
-        if (null === $linkProvider = $request->attributes->get('_links')) {
-            $request->attributes->set('_links', new GenericLinkProvider([$link]));
+        if (is_string($this->discovery)) {
+            $link = new Link('mercure', $this->hub);
```
azjezz, line 66:
```suggestion
            $link = new Link('mercure', $this->discovery);
```

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
azjezz: BC layer added 👍

azjezz:
test failure doesn't seem to be related 🤔

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -188,6 +195,10 @@ private function storeObjectToPublish($object, string $property): void
             if (!isset(self::ALLOWED_KEYS[$key])) {
                 throw new InvalidArgumentException(sprintf('The option "%s" set in the "mercure" attribute of the "%s" resource does not exist. Existing options: "%s"', $key, $resourceClass, implode('", "', self::ALLOWED_KEYS)));
             }
+
+            if ('hub' === $key && !$this->mercure instanceof Mercure) {
+                throw new InvalidArgumentException(sprintf('The option "hub" set in the "mercure" attribute of the "%s" resource is not supported. Please install mercure-bundle:^0.3 to enable it.', $resourceClass));
+            }
```
azjezz, line 201: not sure about the message here.

also, should we ignore the key instead?

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
dunglas: I propose to deprecate it (but we must provide an upgrade path).

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -26,6 +26,7 @@
 use Doctrine\Common\EventArgs;
 use Doctrine\ODM\MongoDB\Event\OnFlushEventArgs as MongoDbOdmOnFlushEventArgs;
 use Doctrine\ORM\Event\OnFlushEventArgs as OrmOnFlushEventArgs;
+use Symfony\Bundle\MercureBundle\Mercure;
```
dunglas, line 29: This bridge must be usable without the Symfony framework. It can depend on the Mercure Component, but not on MercureBundle.
Maybe should we move this class in the component?

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -17,6 +17,7 @@
 use ApiPlatform\Core\Util\CorsTrait;
 use Fig\Link\GenericLinkProvider;
 use Fig\Link\Link;
+use Symfony\Bundle\MercureBundle\Discovery;
```
dunglas, line 20: Same here, we cannot depend on the bundle.

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -29,12 +30,19 @@ final class AddLinkHeaderListener
...
-    private $hub;
+    private $discovery;
 
-    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, string $hub)
+    /**
+     * @param Discovery|string $discovery
+     */
+    public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFactory, $discovery)
     {
+        if (\is_string($discovery)) {
```
dunglas, line 40:
```suggestion
        if (!$discovery instanceof Discovery) {
```

dunglas (changes requested):

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
azjezz: how do you propose we deprecate this?

alanpoulain: Why not triggering a deprecation in an if?

azjezz: that means you will get the deprecation notice if you have messenger installed, so there's no way for you to get around the deprecation.

alanpoulain: Right. Then we would need to introduce a configuration parameter, like `enable_async_update` or something like this, make it `true` by default, trigger a deprecation if not set to `false`.
But maybe it's too much work for an experimental component :sweat_smile:
I let @dunglas decide but personally I'm OK for a BC break mentioned in the CHANGELOG.

azjezz:
@alanpoulain do you have any idea on the failure? seem weird 🤔

alanpoulain:
Probably because of the minimum stability. You should remove it.

azjezz:
> Probably because of the minimum stability. You should remove it.

done, but i think it would be better to test against it in the future.

alanpoulain:
It is :slightly_smiling_face::
https://github.com/api-platform/core/blob/9c348d2b4fdc55b55f202a91878e2f4dd8d6cf99/.github/workflows/ci.yml#L664

azjezz:
how can we avoid the deprecations trigged in the event listeners? should we move the `@trigger_error` to the event listener method instead of the constructor ( event listeners are constructed for every request, so the deprecation would be triggered for every test case. )

alanpoulain:
Superseded by https://github.com/api-platform/core/pull/4194. Thank you @azjezz.

Issue #3535, opened 2020-04-28
Access denied on toMany ApiSubresource collection


plamenh:
**API Platform version(s) affected**: 2.5.4
**Description**  
I get access denied for field 'units' when querying like this:

- Request

```
properties {
	    name
	    units {
	        name
	    }
	}
```

- Response

```
{
            "message": "Access Denied.",
            "extensions": {
                "category": "graphql"
            },
            "locations": [
                {
                    "line": 7,
                    "column": 6
                }
            ],
            "path": [
                "properties",
                0,
                "units"
            ]
        }
```

```
"properties": [
            {
                "name": "10-101",
                "units": null
            },
            {
                "name": "10-706",
                "units": null
            }
        ]
```

**How to reproduce**  
In Property.php:

```
/**
     * One Property has Many Units.
     *
     * @ORM\ManyToMany(targetEntity="Unit", inversedBy="properties")
     * @Expose
     * @MaxDepth(3)
     * Max depth of 3 needed to settle collections from property
     * @SkipWhenEmpty
     * @ApiSubresource
     * @Groups({"read"})
     */
    public $units;
```

In Unit.php:

```
/**
 *
 * @ORM\Table(name="unit")
 * @ORM\Entity
 * @Vich\Uploadable
 * @ExclusionPolicy("all")
 * @ApiResource(
 *      attributes={
 *          "security"="is_granted('ROLE_USER')",
 *          "pagination_enabled"=false
 *      },
 *      graphql={
 *         "item_query"={"security"="is_granted('view', object)"},
 *         "collection_query"={},
 *         "update"={"security"="is_granted('edit', object)"},
 *         "delete"={"security"="is_granted('ROLE_ADMIN')"},
 *         "create"={"security"="is_granted('ROLE_MANAGER') or is_granted('ROLE_CLIENT')"}
 *     }
 * )
 * @ApiFilter(ExistsFilter::class, properties={"properties"})
 */
```

**Possible Solution**  
No idea

**Additional Context**  
Querying for units separately produces results as expected:

- Request

```
units(exists: {properties: true} )  {
		name
	}
```

- Response

```
"data": {
        "units": [
            {
                "name": "10-101"
            },
            {
                "name": "10-706"
            }
        ]
    }
```

plamenh:
Anyone? This seems like a major issue.

DavidBennettUK:
[...]

Issue #3279, opened 2019-11-20
[GraphQL] Add a `visibily` attribute in `ApiProperty` to hide fields depending on the context


alanpoulain:
**Description**  
The PR https://github.com/api-platform/core/pull/3227 has added a `security` attribute in `ApiProperty` for not using a property if it is not allowed.

If the user wants to hide the field completely from the GraphQL schema, we could introduce a `visibility` attribute like GraphQL Ruby does (https://graphql-ruby.org/authorization/visibility).

For REST, it could have the same effect as the `security` attribute.

**Example**  
```php
class Offer
{
    /**
     * @var string The dummy secret property, only readable/writable by specific users
     *
     * @ApiProperty(visibility="is_granted('ROLE_ADMIN')")
     */
    private $adminOnlyProperty;
}
```

PR #4143, opened 2021-03-18
GraphQL association field security fixes + @ApiProperty security support

DavidBennettUK:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? |no
| Tickets       | https://github.com/api-platform/core/issues/4169
| License       | MIT
| Doc PR        | N/A

The security attribute was added to `ApiProperty` here: https://github.com/api-platform/core/pull/3503 which lets you add admin-only properties etc. (https://api-platform.com/docs/core/security/)

However, the GraphQL `ItemNormalizer` wasn't updated to inject the new `ResourceAccessCheckerInterface` dependency into the parent, and would fail silently. This PR adds this missing injection.

This should probably be a separate PR as it's a different issue that's been uncovered now this is fixed, but if you add security to a non-null GraphQL property then you will get a `Cannot return null for non-nullable field` error if trying to request a field you don't have access to. I'm not sure if it would be best if an access denied error was given instead, update the way the schema is built to detect that the field is nullable if security is present, or leave it to the user to make the field nullable.

### DIFF
--- a/features/graphql/authorization.feature
+++ b/features/graphql/authorization.feature
@@ -127,7 +127,7 @@ Feature: Authorization checking
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
-    And the JSON node "data.securedDummy.relatedDummies" should have 1 elements
+    And the JSON node "data.securedDummy.relatedDummies" should have 1 element
 
   Scenario: An admin can access a secured relation
     When I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
@@ -185,6 +185,122 @@ Feature: Authorization checking
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "data.securedDummy.relatedDummy" should be null
+
+  Scenario: A user can't access a secured relation resource directly
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
+    """
+    {
+      relatedSecuredDummy(id: "/related_secured_dummies/1") {
+        id
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "errors[0].extensions.status" should be equal to 403
+    And the JSON node "errors[0].extensions.category" should be equal to user
+    And the JSON node "errors[0].message" should be equal to "Access Denied."
+    And the JSON node "data.relatedSecuredDummy" should be null
+
+  Scenario: A user can't access a secured relation resource collection directly
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
+    """
+    {
+      relatedSecuredDummies {
+        edges {
+          node {
+            id
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "errors[0].extensions.status" should be equal to 403
+    And the JSON node "errors[0].extensions.category" should be equal to user
+    And the JSON node "errors[0].message" should be equal to "Access Denied."
+    And the JSON node "data.relatedSecuredDummies" should be null
+
+  Scenario: A user can access a secured collection relation
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
+    """
+    {
+      securedDummy(id: "/secured_dummies/1") {
+        relatedSecuredDummies {
+          edges {
+            node {
+              id
+            }
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedSecuredDummies" should have 1 element
+
+  Scenario: A user can access a secured relation
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
+    """
+    {
+      securedDummy(id: "/secured_dummies/1") {
+        relatedSecuredDummy {
+          id
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedSecuredDummy" should not be null
+
+  Scenario: A user can access a non-secured collection relation
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
+    """
+    {
+      securedDummy(id: "/secured_dummies/1") {
+        publicRelatedSecuredDummies {
+          edges {
+            node {
+              id
+            }
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.publicRelatedSecuredDummies" should have 1 element
+
+  Scenario: A user can access a non-secured relation
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
+    """
+    {
+      securedDummy(id: "/secured_dummies/1") {
+        publicRelatedSecuredDummy {
+          id
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.publicRelatedSecuredDummy" should not be null
 
   @createSchema
   Scenario: An admin can create a secured resource
--- a/features/graphql/authorization.feature
+++ b/features/graphql/authorization.feature
@@ -64,6 +64,7 @@ Feature: Authorization checking
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummies" should exist
     And the JSON node "data.securedDummies" should not be null
 
   Scenario: An anonymous user cannot retrieve a secured collection
@@ -148,6 +149,7 @@ Feature: Authorization checking
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedDummy" should exist
     And the JSON node "data.securedDummy.relatedDummy" should not be null
 
   @createSchema
@@ -266,6 +268,7 @@ Feature: Authorization checking
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedSecuredDummy" should exist
     And the JSON node "data.securedDummy.relatedSecuredDummy" should not be null
 
   Scenario: A user can access a non-secured collection relation
@@ -304,6 +307,7 @@ Feature: Authorization checking
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.publicRelatedSecuredDummy" should exist
     And the JSON node "data.securedDummy.publicRelatedSecuredDummy" should not be null
 
   @createSchema
@@ -395,6 +399,7 @@ Feature: Authorization checking
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.adminOnlyProperty" should exist
     And the JSON node "data.securedDummy.adminOnlyProperty" should not be null
 
   Scenario: A user can't see a secured admin-only property on an object they own
--- a/features/graphql/authorization.feature
+++ b/features/graphql/authorization.feature
@@ -419,6 +419,37 @@ Feature: Authorization checking
     And the header "Content-Type" should be equal to "application/json"
     And the JSON node "data.securedDummy.adminOnlyProperty" should be null
 
+  Scenario: A user can see a secured owner-only property on an object they own
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    And I send the following GraphQL request:
+    """
+    {
+      securedDummy(id: "/secured_dummies/2") {
+        ownerOnlyProperty
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.ownerOnlyProperty" should exist
+    And the JSON node "data.securedDummy.ownerOnlyProperty" should not be null
+
+  Scenario: An admin can't see a secured owner-only property on an object they don't own
+    When I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
+    And I send the following GraphQL request:
+    """
+    {
+      securedDummy(id: "/secured_dummies/2") {
+        ownerOnlyProperty
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.ownerOnlyProperty" should be null
+
   Scenario: A user can't assign to themself an item they doesn't own
     When I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
     And I send the following GraphQL request:
--- a/features/graphql/authorization.feature
+++ b/features/graphql/authorization.feature
@@ -107,6 +107,86 @@ Feature: Authorization checking
     And the JSON node "errors[0].message" should be equal to "Only admins can create a secured dummy."
 
   @createSchema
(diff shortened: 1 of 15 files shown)


alanpoulain:
Hello, thanks for the PR.

You can find a related discussion here: https://github.com/api-platform/core/pull/3227#issuecomment-556013267
And an issue for adding a `visibility` attribute here (complementary with the `security` one): https://github.com/api-platform/core/issues/3279

alanpoulain:
I think we cannot merge this PR without solving the underlying issue.

DavidBennettUK:
Thanks, I'd missed that discussion. I agree that the field should be marked as nullable if the security attribute is present. I've updated the PR with a suggestion for how this can be done, which seems to work well in my project - working how I'd expect. I'm not super familiar with the internals of this project, so this may not be the best way. I think the visibility attribute might be a separate thing, as I'd still want the property in the schema/docs for use with generating TypeScript types.

alanpoulain:
Yes the `visibility` attribute should be another PR indeed.
WDYT of throwing an exception if the field is not nullable? I think it will be less "magic", don't you think?

DavidBennettUK:
I agree, it's definitely a bit magic. It does make sense though, otherwise an invalid schema is generated if user forgets to make the field nullable. If there's a nice way to set a field's graph type as nullable I'd be for throwing an exception instead. However, unless I've missed something, I haven't seen a way to make a field's graph type nullable, without changing the Doctrine `@ORM\Column` to nullable (wouldn't want to change the database schema, as we shouldn't need to) - with the exception of writing a `TypeConverter`, which is a bit cumbersome. Do you know of a way?

Separately, I've encountered an issue trying to apply security to a Doctrine collection field. If the authorization fails, the collection field isn't present in the data returned by the graph [ItemNormalizer](https://github.com/api-platform/core/blob/2.6/src/GraphQl/Serializer/ItemNormalizer.php#L84) - which is correct. However, the collection field is still in the response due to the [CollectionResolver](https://github.com/api-platform/core/blob/2.6/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php#L90) being run, because it was set up by in [FieldsBuilder](https://github.com/api-platform/core/blob/2.6/src/GraphQl/Type/FieldsBuilder.php#L314) beforehand. This is getting a bit much for my knowledge of the project, so I'm not sure how to fix this. Any pointers?

I should probably open a separate issue for this, but I've noticed that the security stage in CollectionResolver uses the `item_query` security of the child type in the collection, with a `Paginator` as the object. I'm not sure if this is the correct behaviour, as `item_query` security would expect an instance of the entity, not a paginator. More [here](https://github.com/api-platform/core/issues/3535#issuecomment-723212145).

For the time being, is there another way to secure graph fields? I'm aware that it's possible to add dynamic serializer groups, but the issue there is that the dynamic fields aren't added to the generated documentation/schema - which means that using the schema for TypeScript type generation etc. won't work.

alanpoulain:
> Do you know of a way?

You're right, the Doctrine property type extractor has the highest priority. We should definitely solve this issue. I'm not sure how though so maybe the right way is like you did in this PR.

> Any pointers?

This one is tricky indeed. I think we can check if the field is not present in the `$source` array. If it's the case, it means that a security has been applied and we can return directly. WDYT?

> I should probably open a separate issue for this, but I've noticed that the security stage in CollectionResolver uses the `item_query` security of the child type in the collection, with a `Paginator` as the object. I'm not sure if this is the correct behaviour, as `item_query` security would expect an instance of the entity, not a paginator. More [here](https://github.com/api-platform/core/issues/3535#issuecomment-723212145).

Seems like a bug alright. Maybe we should always use `collection_query`. I need to think about it.

> For the time being, is there another way to secure graph fields?

Not for the moment :slightly_smiling_face:

DavidBennettUK:
Sorry for the late reply, I haven't had a chance to revisit this yet. I've opened up an issue (https://github.com/api-platform/core/issues/4169) regarding the security issues to see if this is worth pursuing.

DavidBennettUK:
@alanpoulain I've implemented fixes for the issues raised above (more detail in https://github.com/api-platform/core/issues/4169) + finished implementing `@ApiProperty` security using your suggestion 👍 It all seems to be working well now - security is running how I expect on relations, and fields (including relations) can now be secured. I've updated the PHPUnit tests and added a set of Behat tests to cover the usage scenarios.

Think this is ready to go if you think it looks good? Not sure if this counts as a breaking change or not - would you want me to target another branch?

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -498,7 +498,7 @@ private function convertType(Type $type, bool $input, ?string $queryName, ?strin
             return $this->pagination->isGraphQlEnabled($resourceClass, $operationName) && !$input ? $this->typeBuilder->getResourcePaginatedCollectionType($graphqlType, $resourceClass, $operationName) : GraphQLType::listOf($graphqlType);
         }
 
-        return !$graphqlType instanceof NullableType || $type->isNullable() || (null !== $mutationName && 'update' === $mutationName)
+        return !$graphqlType instanceof NullableType || $type->isNullable() || (null !== $mutationName && 'update' === $mutationName) || $forceNullable
```
alanpoulain, line 501: I would make it first since it's the most understandable and the less costly.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -59,7 +59,7 @@ public function __construct(ReadStageInterface $readStage, SecurityStageInterfac
     public function __invoke(?string $resourceClass = null, ?string $rootClass = null, ?string $operationName = null): callable
     {
         return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operationName) {
-            if (null === $resourceClass || null === $rootClass) {
+            if (null === $resourceClass || null === $rootClass || (null !== $source && !\array_key_exists($info->fieldName, $source))) {
```
alanpoulain, line 62: I think a comment would be helpful to understand what is going on here. Like:
If authorization has failed for a relation field, the field is not present in the source: null can be returned directly.

features/graphql/authorization.feature
```diff
@@ -106,6 +106,86 @@ Feature: Authorization checking
     And the JSON node "errors[0].extensions.category" should be equal to user
     And the JSON node "errors[0].message" should be equal to "Only admins can create a secured dummy."
 
+  @createSchema
+  Scenario: An admin can access a secured collection relation
+    Given there are 1 SecuredDummy objects owned by admin with related dummies
+    When I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
+    When I send the following GraphQL request:
```
alanpoulain, line 113:
```suggestion
    And I send the following GraphQL request:
```

features/graphql/authorization.feature
```diff
@@ -106,6 +106,86 @@ Feature: Authorization checking
...
+            }
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedDummies" should have 1 elements
```
alanpoulain, line 130:
```suggestion
    And the JSON node "data.securedDummy.relatedDummies" should have 1 element
```

features/graphql/authorization.feature
```diff
@@ -106,6 +106,86 @@ Feature: Authorization checking
...
+
+  Scenario: An admin can access a secured relation
+    When I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
+    When I send the following GraphQL request:
```
alanpoulain, line 134:
```suggestion
    And I send the following GraphQL request:
```

features/graphql/authorization.feature
```diff
@@ -106,6 +106,86 @@ Feature: Authorization checking
...
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedDummy" should not be null
+
+  @createSchema
+  Scenario: A user can't access a secured collection relation
+    Given there are 1 SecuredDummy objects owned by dunglas with related dummies
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
```
alanpoulain, line 153:
```suggestion
    And I send the following GraphQL request:
```

features/graphql/authorization.feature
```diff
@@ -106,6 +106,86 @@ Feature: Authorization checking
...
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON node "data.securedDummy.relatedDummies" should be null
+
+  Scenario: A user can't access a secured relation
+    When I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
+    When I send the following GraphQL request:
```
alanpoulain, line 174:
```suggestion
    And I send the following GraphQL request:
```

tests/Fixtures/TestBundle/Document/SecuredDummy.php
```diff
@@ -89,6 +90,27 @@ class SecuredDummy
      */
     private $owner;
 
+    /**
+     * @var ArrayCollection Several dummies
```
alanpoulain, line 94:
```suggestion
     * @var Collection<RelatedDummy>
```

tests/Fixtures/TestBundle/Entity/SecuredDummy.php
```diff
@@ -90,6 +91,27 @@ class SecuredDummy
      */
     private $owner;
 
+    /**
+     * @var ArrayCollection Several dummies
```
alanpoulain, line 95:
```suggestion
     * @var Collection<RelatedDummy>
```

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -87,17 +87,20 @@ public function __invoke(?string $resourceClass = null, ?string $rootClass = nul
...
-                ],
-            ]);
-            ($this->securityPostDenormalizeStage)($resourceClass, $operationName, $resolverContext + [
-                'extra_variables' => [
-                    'object' => $collection,
-                    'previous_object' => $this->clone($collection),
-                ],
-            ]);
+            // Only perform security stage on the top-level query
+            if (null === $source) {
```
alanpoulain, line 91: I think it's dangerous to do this. If the developer has added security in the related resource, they think it will be always applied in all contexts, don't you think?
Only applying the security at the top level means that if a user has access to a resource, it can access to secured resources by traversing the relations (if the relations have not been secured with your change of course).
I think the better approach is to change the operation name to be always `collection_query`.

tests/GraphQl/Resolver/Factory/CollectionResolverFactoryTest.php
```diff
@@ -74,7 +74,48 @@ public function testResolve(): void
...
+            ],
+        ])->shouldNotBeCalled();
+
+        $serializeStageData = ['serialized'];
+        $this->serializeStageProphecy->__invoke($readStageCollection, $resourceClass, $operationName, $resolverContext)->shouldBeCalled()->willReturn($serializeStageData);
+
+        $this->assertSame($serializeStageData, ($this->collectionResolverFactory)($resourceClass, $rootClass, $operationName)($source, $args, null, $info));
+    }
+
+    public function testResolveNullSource(): void
```
alanpoulain, line 113: We should add a test when the source is not null but the field name is not present as well.

alanpoulain (changes requested):
Since there are a lot changes, I think it would be safer to target `main` indeed.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
DavidBennettUK: I do see your point, and would certainly want to update the documentation to make it clearer if we do go this route.
That said, I see the `ApiResource` GraphQL `item_query`, `collection_query`, etc. as definining the actual GraphQL entrypoints i.e. the top-level queries.
Association collections are different collections to the main resource collection - they're filtered. The `collection_query` in `ApiResource` exposes the entire collection. I don't think it makes sense to use the same security on associations, as the requirements may be different. In my example (in https://github.com/api-platform/core/issues/4169) you would want admins to see the whole collection of documents, and not let normal users see them all, instead letting users only see their own documents via the `user.documents` association. Using the `collection_query` security on the association would mean only admins could ever see it.
You may also not want to have a `collection_query` exposing the whole collection as an endpoint, but still want to have associations. Relying on there being a `collection_query` present wouldn't work in this case.
Unless I've found a bug, in 2.6.3 at least, this PR makes collection security work the same way as single associations already do. `item_query` security from the association resource is not run on associations - meaning they are already exposed with no way to protect them. This is covered by the following two tests I've added:
- https://github.com/api-platform/core/pull/4143/files#diff-14861a046eb66cb65d4007864a6fbe8d08b64a6cd05a61f21d5cab03b7eb7525R193 - shows that it's not possible for a user to access an `item_query` secured resource directly
- https://github.com/api-platform/core/pull/4143/files#diff-14861a046eb66cb65d4007864a6fbe8d08b64a6cd05a61f21d5cab03b7eb7525R254 - shows that it's possible to access the `item_query` secured resource via a association. The `item_query` security is never run.

I hope that all makes sense!

alanpoulain:
I have changed the targeted branch to `main`, could you rebase your `2.6` branch on `main` (for future PR you should create a new branch)?
I have no right to push on your branch, you should allow maintainers to push on it, it makes things easier for us:
image

DavidBennettUK:
@alanpoulain thanks, I've rebased onto `main`. It looks like because I created the fork in an organisation, it's not possible to allow edits to maintainers except inviting them individually (https://github.com/isaacs/github/issues/1681) - I'll remember for next time. I've also added some entries to the changelog to cover the changes in this PR.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -249,6 +249,11 @@ public function denormalize($data, $class, $format = null, array $context = [])
             return $item;
         }
 
+        // Filter out input attributes that aren't allowed
+        $data = array_filter($data, function (string $attribute) use ($context) {
+            return $this->canAccessAttribute($context[self::OBJECT_TO_POPULATE] ?? null, $attribute, $context);
+        }, \ARRAY_FILTER_USE_KEY);
```
alanpoulain, line 255: Is this change necessary? Could you denormalize and update / create a secured field without the right to do so before?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -379,18 +384,28 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
     /**
      * {@inheritdoc}
      */
-    protected function isAllowedAttribute($classOrObject, $attribute, $format = null, array $context = [])
+    protected function isAllowedAttribute($classOrObject, string $attribute, string $format = null, array $context = [])
```
alanpoulain, line 387: You cannot change the typehint, it's a BC break.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -379,18 +384,28 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
             return false;
         }
 
+        return \is_object($classOrObject) ? $this->canAccessAttribute($classOrObject, $attribute, $context) : true;
+    }
+
+    /**
+     * Check if access to the attribute is granted.
+     *
+     * @param object|null $object
```
alanpoulain, line 399: It seems it can also be a `class-string`.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -379,18 +384,28 @@ protected function getAllowedAttributes($classOrObject, array $context, $attribu
...
      * {@inheritdoc}
      */
-    protected function isAllowedAttribute($classOrObject, $attribute, $format = null, array $context = [])
+    protected function isAllowedAttribute($classOrObject, string $attribute, string $format = null, array $context = [])
     {
         if (!parent::isAllowedAttribute($classOrObject, $attribute, $format, $context)) {
             return false;
         }
 
+        return \is_object($classOrObject) ? $this->canAccessAttribute($classOrObject, $attribute, $context) : true;
```
alanpoulain, line 393: Isn't it dangerous to do so? We could add a security without an object, isn't it?

alanpoulain:
For the change "Security: Fix `ApiProperty` security being passed a null object for updates.", please do it in another PR, it will be easier to discuss about it and I think it's not really related to this PR.

DavidBennettUK:
@alanpoulain that's fair, I did consider splitting it up, as it fixes an issue not just for GraphQL but all. I've removed it from this PR and will open another.

alanpoulain (approved):

alanpoulain:
This is some great work! Thank you very much @DavidBennettUK.
Can I ask you to write some documentation about it? I think it would be very helpful.

DavidBennettUK:
Thanks @alanpoulain - happy to be able to contribute to the project! I appreciate your thorough feedback and help. I'll get some documentation written up and put in a PR in the next few days - I spent a bit longer working on the other PR (https://github.com/api-platform/core/pull/4184) than I expected, but that's ready to be looked at now at least.

DavidBennettUK:
@alanpoulain Got the documention PR ready 👍  https://github.com/api-platform/docs/pull/1338

webda2l:
Thanks for this PR and #4184!

@alanpoulain It works great on `dev-main`, but any reason that it's to not already part of a 2.6 realease? If yes, any ETA to have this code on a stable (2.7 or 3.0)? Thank you!

alanpoulain:
It will be available for `2.7` but we don't have any ETA.

carlobeltrame:
Since this PR is relevant to security and there is no definite release date for 2.7, should we backport this to 2.6.x?

PR #4150, opened 2021-03-19
Add owners topic on publish Mercure updates listener

djpremier:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no 
| New feature?  | yes 
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT
| Doc PR        | api-platform/docs#1317 

Adds the `owners_topic` option to mercure to enable the solution of the access restriction problem, as exemplified in the mercure specification (https://mercure.rocks/spec#subscribers).

With this addition, you can switch to the option `owners_topic` a property or function that returns one or more objects, so that it is possible to publish alternative topics to filter the receipt, as the most common use and according to the example:
`https://example.com/users/foo/?topic=https%3A%2F%2Fexample.com%2Fbooks%2F1`

Example
```php
<?php
// api/src/Entity/Book.php

namespace App\Entity;

use App\Entity\User;

#[
    ApiResource(
        mercure: [
            "private" => true,
            "owners_topic" => "object.getOwner()",
        ],
    )
]
class Book
{
    private ?User $owner;

    public function getOwner(): ?User
    {
        return $this->owner;
    }
}
```

JWT Token:
```json
{
  "mercure": {
    "subscribe": [
      "https://example.com/users/foo/{?topic}"
    ],
  }
}
```

Subscribe topic
`https://example.com/books/{id}`

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -45,6 +45,7 @@ final class PublishMercureUpdatesListener
     use ResourceClassInfoTrait;
     private const ALLOWED_KEYS = [
         'topics' => true,
+        'owners_topic' => true,
         'data' => true,
         'private' => true,
         'id' => true,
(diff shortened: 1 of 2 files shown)


src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -222,7 +223,35 @@ private function publishUpdate($object, array $options, string $type): void
...
+            }
+
+            $topics = (array) $iri;
+
+            $owners = $this->expressionLanguage->evaluate($options['owners_topic'], ['object' => $object]);
+            $owners = empty($owners) || \is_array($owners) ? $owners : [$owners];
+
+            if (!empty($owners)) {
+                foreach ($owners as $owner) {
+                    $query_topics = '?';
```
alanpoulain, line 238:
```suggestion
                    $queryTopics = '?';
```

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -222,7 +223,35 @@ private function publishUpdate($object, array $options, string $type): void
...
+
+                    foreach ((array) $iri as $topic) {
+                        $query_topics .= \strlen($query_topics) > 1 ? '&' : '';
+                        $query_topics .= sprintf('topic=%s', urlencode($topic));
```
alanpoulain, line 242: Could you use `http_build_query` instead?

alanpoulain:
Thank you for your PR.
Could you provide some unit tests as well?

dunglas:
You can already to something like `#[ApiResource(mercure: ["'https://maintopic' ~ object.id", "'https://maintopic' ~ object.id ~ '?topic' ~ object.owner.id '"]`. Isn't it enough?

Maybe could we add `iri()` and `urlencode()` helper functions to the instance of ExpressionLanguage instead of this new key?

djpremier:
> You can already to something like `#[ApiResource(mercure: ["'https://maintopic' ~ object.id", "'https://maintopic' ~ object.id ~ '?topic' ~ object.owner.id '"]`. Isn't it enough?
> 
> Maybe could we add `iri()` and `urlencode()` helper functions to the instance of ExpressionLanguage instead of this new key?

Hello Dunglas, it would work in a certain way, but the system address (URL) is fixed, which may not be feasible.

Possibly we can change to inject the functions as you said, but at the moment I don't know how to do it, I will have to analyze it.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
djpremier: I was looking for something like this, but I made the change and it was not possible to use it. It turns out that the function receives an array of key / value, making it impossible to use the same key more than once. It is possible to put a prefix, leaving something like `topic_0`, `topic_1`, but I modified the subscribe to suit the pattern and it didn’t work, so I did `https://example.com/{?topic_[0-9]+}`.

So I don't know if Mercure is capable of interpreting the regex, or if it expects anything other than the syntax I know.

djpremier:
> You can already to something like `#[ApiResource(mercure: ["'https://maintopic' ~ object.id", "'https://maintopic' ~ object.id ~ '?topic' ~ object.owner.id '"]`. Isn't it enough?
> 
> Maybe could we add `iri()` and `urlencode()` helper functions to the instance of ExpressionLanguage instead of this new key?

What would you propose would be something like that, correct? I think it is more work for development, but if it is better, it can be done that way.

https://github.com/api-platform/core/pull/4153

PR #4151, opened 2021-03-21
Fix order filter with embedded and nulls comparison

### LINKED ISSUE #3440: [Filters] OrderFilter not working on Embedded object when using nulls_comparison

**Description**  

When trying to sort on a field of an `@ORM\Embedded` object, it doesn't work if I use the `nulls_comparison` of the `OrderFilter`.

Here is an example (for the full code, see the provided reproducer below):
```
/**
 * @ApiResource()
 * @ORM\Entity()
 *
 * @ApiFilter(OrderFilter::class, properties = {
 *     "bar.luckyNumber" : { "nulls_comparison" : OrderFilter::NULLS_SMALLEST },
 *     "bar.secondLuckyNumber"
 * })
 */
class Foo
```

Here, if `bar` is Embedded, the request to `/foos?order[bar.luckyNumber]=asc` will fail with the error 
> [Syntax Error] line 0, col 80: Error: Expected Doctrine\\ORM\\Query\\Lexer::T_FROM, got '.'

However, the request to `/foos?order[bar.secondLuckyNumber]=asc` will work as expected because it doesn't use the `nulls_comparison` option. It will use the default behavior of Postgres which is `NULL` considered as largest which is not what I want here.

**How to reproduce**  

For simplicity sake, I've created a reproducer repository which uses the `api-platform/api-platform` distribution.
Here it is : https://github.com/t-richard/order-filter-reproducer

I've also provided Doctrine migrations and Doctrine fixtures in the reproducer.

Once the project is up and the fixtures loaded, you can make an HTTP call to
- https://localhost:8443/foos?order[bar.luckyNumber]=asc which SHOULD NOT work
- https://localhost:8443/foos?order[bar.secondLuckyNumber]=asc which SHOULD work as expected (NULLs last)

**Possible Solution**  

I've not managed to find where the problem comes from, but I may be able to provide a PR if you guys can help me out to identify the problem.

Let me know if I can be of any help !

### LINKED ISSUE #4144: Comparing with Null Values doesn't work with embeddables
**API Platform version(s) affected**: 2.6.x-dev and 2.6.3

**Feature**

Order Filter (Sorting) > [Comparing with Null Values](https://api-platform.com/docs/core/filters/#comparing-with-null-values)

**Description**  
You cannot fine configure null value comparison strategy with embeddable.
✅ Adding an OrderFilter on an embeddable property does work as expected.
```
#[ApiFilter(OrderFilter::class, properties: [
    'embeddable.property'
])]
```
But trying to configure the option "nulls_comparaison", a QueryException exception is thrown. 🛑
```
#[ApiFilter(OrderFilter::class, properties: [
    'embeddable.property' => [
        'nulls_comparison' => OrderFilter::NULLS_LARGEST,
    ],
])]
```

**How to reproduce**  

[Repository to reproduce the issue](https://github.com/walva/api-platform-core-issue-4144)

Given you add a OrderFilter to an embeddable property (eg: embeddable. property)
Wen configuring the nulls comparison strategy (eg: OrderFilter::NULLS_LARGEST)
And you navigate to the endpoint using the order query parameter (eg: sort[embeddable.property]
Then the collection is sorted properly
Instead, it throws a **Doctrine\ORM\Query\QueryException**

**Possible Solution**  

**Additional Context**  

image

walva:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Tickets       | Fixes #4144 Fixes #3440
| License       | MIT
| Doc PR        | Feature already documented here: Order Filter (Sorting) > [Comparing with Null Values](https://api-platform.com/docs/core/filters/#comparing-with-null-values)

### Description

**Comparing with Null Values doesn't work with embeddable.**

Thanks @vsalvans who found the issue in #3440.

### Todo

- [X] Always add tests and ensure they pass. (PhpUnit)
- [ ] Always add tests and ensure they pass. (Behat)
- [ ] Update CHANGELOG.md file
- [ ] Add test for mondoDB ?

### Issue n°1: Is there a better way to add a test for this bug ?

I wrote a lot of PhpUnits Tests code. First, I started to test the correct behavior of OrderFilter regarding embaddables. I modified `\ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy` to add a `public $embeddedDummy;` property.
By doing this, I was able to add a new test scenario under `\ApiPlatform\Core\Tests\Bridge\Doctrine\Common\Filter\OrderFilterTestTrait::provideApplyTestArguments` : 
```
'embeddable_nulls_comparison' => [
    [
        'embeddedDummy.dummyName' => [
            'nulls_comparison' => 'nulls_largest',
        ],
    ],
    [
        'order' => [
            'embeddedDummy.dummyName' => 'asc',
        ],
    ],
],
```
Unfortunatly, adding a new property created several new entries in the API documentation which led me writing a lot of code for all other tests (eg: BooleanFilterTest, RangeFilterTest, etc...). Regarding the impacts in the tests (see above), I am not convinced it is the best way to go. Should I create more isolated tests specific to embeddable somewhere else ?

### Issue n°2: behat tests fails

While all PhpUnit tests succeed, the behats tests fail. The first failing scenario is creating a new Dummy resource using GraphQL. The reason is that to create a new Dummy entity, it now needs the missing part of  `public $embeddedDummy;` in the payload.
This encourages me thinking this is not the best way to go.

```
001 Scenario: Create an item with a subresource                                     # features/graphql/mutation.feature:428
      And the JSON node "data.createDummy.dummy.id" should be equal to "/dummies/2" # features/graphql/mutation.feature:451
        Failed to evaluate expression 'data.createDummy.dummy.id' (Exception)

```

### Issue n°3: I didn't write tests for mongoDB, but tests are green.

Am I missing something here ? Also, I never used mongoDB, and I'm not even sure the bug affect this database.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
@@ -105,7 +105,7 @@ class OrderFilter extends AbstractContextAwareFilter implements OrderFilterInter
         if (null !== $nullsComparison = $this->properties[$property]['nulls_comparison'] ?? null) {
             $nullsDirection = self::NULLS_DIRECTION_MAP[$nullsComparison][$direction];
 
-            $nullRankHiddenField = sprintf('_%s_%s_null_rank', $alias, $field);
+            $nullRankHiddenField = sprintf('_%s_%s_null_rank', $alias, str_replace('.', '_', $field));
 
             $queryBuilder->addSelect(sprintf('CASE WHEN %s.%s IS NULL THEN 0 ELSE 1 END AS HIDDEN %s', $alias, $field, $nullRankHiddenField));
             $queryBuilder->addOrderBy($nullRankHiddenField, $nullsDirection);
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           php-version: ${{ matrix.php }}
           extensions: intl, bcmath, curl, openssl, mbstring 
           ini-values: memory_limit=-1
-          tools: pecl, composer, php-cs-fixer
+          tools: pecl, composer, php-cs-fixer:2.18.4
           coverage: none
       - name: Run PHP-CS-Fixer fix
         run: php-cs-fixer fix --dry-run --diff --ansi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * JSON Schema: Manage Sequentially and AtLeastOneOf constraints when generating property metadata (#4139 and #4147)
 * Doctrine: Fix purging HTTP cache for unreadable relations (#3441)
 * Doctrine: Revert #3774 support for binary UUID in search filter (#4134)
+* Doctrine: Fix order filter when using embedded and nulls comparison (#4151)
 * GraphQL: Partial pagination support (#3223)
 * GraphQL: Manage `pagination_use_output_walkers` and `pagination_fetch_join_collection` for operations (#3311)
 * Swagger UI: Remove Google fonts (#4112)
--- a/src/Test/DoctrineMongoDbOdmFilterTestCase.php
+++ b/src/Test/DoctrineMongoDbOdmFilterTestCase.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\Test;
 
 use ApiPlatform\Core\Bridge\Doctrine\MongoDbOdm\Filter\FilterInterface;
 use ApiPlatform\Core\Tests\Fixtures\TestBundle\Document\Dummy;
+use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Repository\DocumentRepository;
 use Doctrine\Persistence\ManagerRegistry;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
@@ -26,6 +27,11 @@ use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
  */
 abstract class DoctrineMongoDbOdmFilterTestCase extends KernelTestCase
 {
+    /**
+     * @var DocumentManager
+     */
+    protected $manager;
+
     /**
      * @var ManagerRegistry
      */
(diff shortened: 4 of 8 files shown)


src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -106,12 +106,14 @@ protected function filterProperty(string $property, $direction, QueryBuilder $qu
...
 
             $nullRankHiddenField = sprintf('_%s_%s_null_rank', $alias, $field);
+            $nullRankHiddenField = sprintf('_%s_%s_null_rank', $alias, str_replace('.', '_', $field));
 
             $queryBuilder->addSelect(sprintf('CASE WHEN %s.%s IS NULL THEN 0 ELSE 1 END AS HIDDEN %s', $alias, $field, $nullRankHiddenField));
             $queryBuilder->addOrderBy($nullRankHiddenField, $nullsDirection);
         }
 
         $queryBuilder->addOrderBy(sprintf('%s.%s', $alias, $field), $direction);
+        // dump($queryBuilder->getQuery());
```
alanpoulain, line 116: Should be removed.

src/Bridge/Doctrine/Orm/Filter/OrderFilter.php
```diff
@@ -106,12 +106,14 @@ protected function filterProperty(string $property, $direction, QueryBuilder $qu
             $nullsDirection = self::NULLS_DIRECTION_MAP[$nullsComparison][$direction];
 
             $nullRankHiddenField = sprintf('_%s_%s_null_rank', $alias, $field);
+            $nullRankHiddenField = sprintf('_%s_%s_null_rank', $alias, str_replace('.', '_', $field));
```
alanpoulain, line 109: Please remove the previous line.

alanpoulain:
> Should I create more isolated tests specific to embeddable somewhere else ?

Yes I think it would be the best course of action. We can add an optional parameter to use another resource to do the tests in `DoctrineOrmFilterTestCase`.

For MongoDB, there is no null comparison, you don't need to change anything.

walva:
I force pushed another branch with a new approach based on previous feedback. WDYT ?
If the current version is better, what would be missing?

- [ ] Update CHANGELOG.md file

walva:
Also, I couldn't download PHP CS Fixer tool, website seems to be down at the moment.
(tools mentionned here : https://github.com/api-platform/core/blob/main/CONTRIBUTING.md )

alanpoulain:
I made some little fixes, I think it's good to merge :slightly_smiling_face:

alanpoulain (approved):

alanpoulain:
Thank you @walva.

walva:
Thank you too! Very instructive

vsalvans:
Thank you guys for all work done!!

t-richard:
Nice to see it was fixed :nerd_face: 

Thanks @walva for the work done here :muscle:

PR #4153, opened 2021-03-22
feat: Add iri and urlencode to expression language on Mercure

### LINKED ISSUE #4164: Api-PLatform & Mercure : custom topics
Hello,

My Users are divided in several user groups.
Most of my entities have a user group fiield refering to the user group they belong too.

when an entry is modified / suppressed I want the SSE message to be sent only to the users belonging to the entry usergroup.
I thought using custom topics for this purpose,

http://server/api/books/1/usergroup1
http://server/api/books/5/usergroup2

 but when specifying the topics option in Api-platform, it overides the original IRI based topic (and it seems to be impossible to get it back...)

In PublishMercureUpdatesListener.php, , publishUpdate function :
`$iri = $options['topics'] ?? $this->iriConverter->getIriFromItem($object, UrlGeneratorInterface::ABS_URL);`

It could also be pratical to specify several topics, including the original IRI based one :

```
/**
 * @ApiResource(
 *   mercure="object.getMercureOptions()",
 * )
  */
function getMercureOptions() {
return [ 
    'private' => false,
    "topics" => [
        originalIriBasedOne, 
        originalIriBasedOne/object.usergroup,
        fullCustomOne,
    ]
]}
```
Thank you,
 Alex

djpremier:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Bug fix?      | no 
| New feature?  | no 
| Deprecations? | no
| Tickets       | Fix #4164 
| License       | MIT
| Doc PR        | api-platform/docs/pull/1360 

Adds the `iri` (iriConverter) and `escape` (rawurlencode) functions to mercure allowing generate `topics` dynamically and satisfying the need to restrict the access of subscribers, according to the example described in the Mercure specification (https://mercure.rocks/spec#subscribers).

Example
```php
<?php
// api/src/Entity/Book.php

namespace App\Entity;

use App\Entity\User;

#[
    ApiResource(
        mercure: 'object.getMercureOptions()',
    )
]
class Book
{
    private ?User $owner;

    public function getMercureOptions(): array
    {
        # the '@=' prefix is required when using expressions for arguments in topics
        $topic1 = '@=iri(object)';
        $topic2 = '@=iri(object.getOwner()) ~ "/?topic=" ~ escape('.$topic1.')';
        $topic3 = '@=iri(object, '.UrlGeneratorInterface::ABS_PATH.')', # you can also change the reference type
        $topic4 = '@=iri(object, 1)', # you can also change the reference type
        $topic5 = 'https://example.com/books/1';

        return [
            'private' => true,
            'topics' => [$topic1, $topic2, $topic3, $topic4, $topic5],
        ];
    }

    public function getOwner(): ?User
    {
        return $this->owner;
    }
}
```

JWT Token:
```json
{
  "mercure": {
    "subscribe": [
      "https://example.com/users/foo/{?topic}"
    ],
  }
}
```

Subscribe topic
```https://example.com/books/{id}```

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -26,6 +26,7 @@ use ApiPlatform\Core\Util\ResourceClassInfoTrait;
 use Doctrine\Common\EventArgs;
 use Doctrine\ODM\MongoDB\Event\OnFlushEventArgs as MongoDbOdmOnFlushEventArgs;
 use Doctrine\ORM\Event\OnFlushEventArgs as OrmOnFlushEventArgs;
+use Symfony\Component\ExpressionLanguage\ExpressionFunction;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\Mercure\HubRegistry;
@@ -88,6 +89,17 @@ final class PublishMercureUpdatesListener
         $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
         $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
         $this->reset();
+
+        $rawurlencode = ExpressionFunction::fromPhp('rawurlencode', 'escape');
+        $this->expressionLanguage->addFunction($rawurlencode);
+
+        $this->expressionLanguage->addFunction(
+            new ExpressionFunction('iri', static function (string $apiResource, int $referenceType = UrlGeneratorInterface::ABS_URL): string {
+                return sprintf('iri(%s, %d)', $apiResource, $referenceType);
+            }, static function (array $arguments, $apiResource, int $referenceType = UrlGeneratorInterface::ABS_URL) use ($iriConverter): string {
+                return $iriConverter->getIriFromItem($apiResource, $referenceType);
+            })
+        );
     }
 
     /**
@@ -199,6 +211,29 @@ final class PublishMercureUpdatesListener
         }
 
         $options['enable_async_update'] = $options['enable_async_update'] ?? true;
+
+        if ($options['topics'] ?? false) {
+            $topics = [];
+            foreach ((array) $options['topics'] as $topic) {
+                if (!\is_string($topic)) {
+                    $topics[] = $topic;
+                    continue;
+                }
+
+                if (0 !== strpos($topic, '@=')) {
+                    $topics[] = $topic;
+                    continue;
+                }
+
+                if (null === $this->expressionLanguage) {
+                    throw new \LogicException('The "@=" expression syntax cannot be used without the Expression Language component. Try running "composer require symfony/expression-language".');
+                }
+
+                $topics[] = $this->expressionLanguage->evaluate(substr($topic, 2), ['object' => $object]);
+            }
+
+            $options['topics'] = $topics;
+        }
 
         if ('deletedObjects' === $property) {
             $this->deletedObjects[(object) [
--- a/tests/Bridge/Doctrine/EventListener/PublishMercureUpdatesListenerTest.php
+++ b/tests/Bridge/Doctrine/EventListener/PublishMercureUpdatesListenerTest.php
@@ -180,12 +180,15 @@ class PublishMercureUpdatesListenerTest extends TestCase
         $iriConverterProphecy->getIriFromItem($toDeleteExpressionLanguage, UrlGeneratorInterface::ABS_URL)->willReturn('http://example.com/dummy_friends/4')->shouldBeCalled();
         $iriConverterProphecy->getIriFromItem($toDeleteMercureOptions)->willReturn('/dummy_offers/5')->shouldBeCalled();
         $iriConverterProphecy->getIriFromItem($toDeleteMercureOptions, UrlGeneratorInterface::ABS_URL)->willReturn('http://example.com/dummy_offers/5')->shouldBeCalled();
+        $iriConverterProphecy->getIriFromItem($toDeleteMercureOptions, UrlGeneratorInterface::ABS_PATH)->willReturn('/dummy_offers/5')->shouldBeCalled();
+        $iriConverterProphecy->getIriFromItem($toDeleteMercureOptions, UrlGeneratorInterface::REL_PATH)->willReturn('./dummy_offers/5')->shouldBeCalled();
+        $iriConverterProphecy->getIriFromItem($toDeleteMercureOptions, UrlGeneratorInterface::NET_PATH)->willReturn('//example.com/dummy_offers/5')->shouldBeCalled();
 
         $resourceMetadataFactoryProphecy = $this->prophesize(ResourceMetadataFactoryInterface::class);
         $resourceMetadataFactoryProphecy->create(Dummy::class)->willReturn(new ResourceMetadata(null, null, null, null, null, ['mercure' => true, 'normalization_context' => ['groups' => ['foo', 'bar']]]));
         $resourceMetadataFactoryProphecy->create(DummyCar::class)->willReturn(new ResourceMetadata());
         $resourceMetadataFactoryProphecy->create(DummyFriend::class)->willReturn(new ResourceMetadata(null, null, null, null, null, ['mercure' => ['private' => true, 'retry' => 10]]));
-        $resourceMetadataFactoryProphecy->create(DummyOffer::class)->willReturn(new ResourceMetadata(null, null, null, null, null, ['mercure' => ['topics' => 'http://example.com/custom_topics/1', 'data' => 'mercure_custom_data', 'normalization_context' => ['groups' => ['baz']]]]));
+        $resourceMetadataFactoryProphecy->create(DummyOffer::class)->willReturn(new ResourceMetadata(null, null, null, null, null, ['mercure' => ['topics' => ['http://example.com/custom_topics/1', '@=iri(object)', '@=iri(object, '.UrlGeneratorInterface::ABS_URL.')', '@=iri(object, '.UrlGeneratorInterface::ABS_PATH.')', '@=iri(object, '.UrlGeneratorInterface::REL_PATH.')', '@=iri(object, '.UrlGeneratorInterface::NET_PATH.')', '@=iri(object) ~ "/?topic=" ~ escape(iri(object))', '@=iri(object) ~ "/?topic=" ~ escape(iri(object, '.UrlGeneratorInterface::ABS_PATH.'))'], 'data' => 'mercure_custom_data', 'normalization_context' => ['groups' => ['baz']]]]));
         $resourceMetadataFactoryProphecy->create(DummyMercure::class)->willReturn(new ResourceMetadata(null, null, null, null, null, ['mercure' => ['topics' => ['/dummies/1', '/users/3'], 'normalization_context' => ['groups' => ['baz']]]]));
 
         $serializerProphecy = $this->prophesize(SerializerInterface::class);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by dunglas** [comment]
Wouldn't be possible to do something like this?

```php
#[
    ApiResource(
        mercure: ['topics' => ['iri(object)', 'iri(object) ~ "?topic=" ~ urlencode(iri(object.getOwner()))']],
   )
]
```

(`urlencode()` is probably not necessary actually, to be tested).

djpremier:
> Wouldn't be possible to do something like this?
> 
> ```
> #[
>     ApiResource(
>         mercure: ['topics' => ['iri(object)', 'iri(object) ~ "?topic=" ~ urlencode(iri(object.getOwner()))']],
>    )
> ]
> ```
> 
> (`urlencode()` is probably not necessary actually, to be tested).

Yes, the absolute URL definition part is not mandatory, just a reinforcement of the access block.
Calling the function directly at the moment I still don't know how to do it, due to the fact that Expression's Parser directly understands that it is a function, so search for it in the array of functions that it knows about (change I made by adding the `urlencode`), however, as the function `iri` needs to start from the object instance, I still need to study if it is possible to do something about it.
Regarding `urlencode`, unfortunately the way Mercure is today does not work, because when the subscription request is made, the query parameter data is escaped, so the publisher needs to send it in the same format so that at Mercure it is possible validate the partner. Possibly, a change can be made at the Mercure so that it can be validated in both ways (raw and parsed), but I still don't know Go, and I can't help at this moment.

djpremier:
> > Wouldn't be possible to do something like this?
> > ```
> > #[
> >     ApiResource(
> >         mercure: ['topics' => ['iri(object)', 'iri(object) ~ "?topic=" ~ urlencode(iri(object.getOwner()))']],
> >    )
> > ]
> > ```
> > 
> > 
> > (`urlencode()` is probably not necessary actually, to be tested).
> 
> Yes, the absolute URL definition part is not mandatory, just a reinforcement of the access block.
> Calling the function directly at the moment I still don't know how to do it, due to the fact that Expression's Parser directly understands that it is a function, so search for it in the array of functions that it knows about (change I made by adding the `urlencode`), however, as the function `iri` needs to start from the object instance, I still need to study if it is possible to do something about it.
> Regarding `urlencode`, unfortunately the way Mercure is today does not work, because when the subscription request is made, the query parameter data is escaped, so the publisher needs to send it in the same format so that at Mercure it is possible validate the partner. Possibly, a change can be made at the Mercure so that it can be validated in both ways (raw and parsed), but I still don't know Go, and I can't help at this moment.

A serious problem that arose in this change concerns the treatment of the content, if you look at the test (I also executed it manually), the `evaluate` function of the` expression language` tries to convert the content of a raw URL, generating an error, so I believe that we will not be able to use this approach, or we will have to find a way to escape / differentiate raw information from parts that we want to convert.

TjorvenB:
@djpremier I was trying to do something similar in my own project. I tried different things but couldn't find a decent path forward without alter api platform itself. This seems to be almost there so I'm providing you with some info to help you along.

I second your serious problem with the treatment of content. But I think it could be solved adequately by adding a key that enables the topic evaluation. Something like bool evaluateTopics. In the service container I think they prefix expressions with '@=', which is an alternative that stays in line with the way symfony works.

As for your problem with the iri() function and starting from the object. I had this code in place.
```
$this->expressionLanguage->addFunction(
    new ExpressionFunction('iri', function ($apiResource) {
        return sprintf('iri(%1$s)', $apiResource);
    }, function ($arguments, $apiResource) use ($iriConverter) {
        return $iriConverter->getIriFromItem($apiResource, UrlGeneratorInterface::ABS_URL);
    })
);
```

But since you have control over the variables you could access the iriConverter via the arguments.

```
$this->expressionLanguage->addFunction(
    new ExpressionFunction('iri', function ($apiResource) {
        return sprintf('iri(%1$s)', $apiResource);
    }, function ($arguments, $apiResource) {
        return $arguments['iriConverter']->getIriFromItem($apiResource, UrlGeneratorInterface::ABS_URL);
    })
);
```

djpremier:
> @djpremier I was trying to do something similar in my own project. I tried different things but couldn't find a decent path forward without alter api platform itself. This seems to be almost there so I'm providing you with some info to help you along.
> 
> I second your serious problem with the treatment of content. But I think it could be solved adequately by adding a key that enables the topic evaluation. Something like bool evaluateTopics. In the service container I think they prefix expressions with '@=', which is an alternative that stays in line with the way symfony works.
> 
> As for your problem with the iri() function and starting from the object. I had this code in place.
> 
> ```
> $this->expressionLanguage->addFunction(
>     new ExpressionFunction('iri', function ($apiResource) {
>         return sprintf('iri(%1$s)', $apiResource);
>     }, function ($arguments, $apiResource) use ($iriConverter) {
>         return $iriConverter->getIriFromItem($apiResource, UrlGeneratorInterface::ABS_URL);
>     })
> );
> ```
> 
> But since you have control over the variables you could access the iriConverter via the arguments.
> 
> ```
> $this->expressionLanguage->addFunction(
>     new ExpressionFunction('iri', function ($apiResource) {
>         return sprintf('iri(%1$s)', $apiResource);
>     }, function ($arguments, $apiResource) {
>         return $arguments['iriConverter']->getIriFromItem($apiResource, UrlGeneratorInterface::ABS_URL);
>     })
> );
> ```

Thanks for the tip, I'll see what I can do.

I'm thinking of just putting it in the root, that is, as it is currently where we can pass the string in mercure and execute the expressions

djpremier:
@alanpoulain Can you review this?

I didn't want to be marking you, but I think the PR may have "fallen" into oblivion.

soyuka:
I like the work you did there! My only concern is about that prefix `=@` is this really necessary? 
Please add some tests as well thanks! 

May we close https://github.com/api-platform/core/pull/4150 ?

djpremier:
> I like the work you did there! My only concern is about that prefix `@=` is this really necessary?
> Please add some tests as well thanks!
> 
> May we close #4150 ?

Hello, yes because I need a way to know that the topic is an expression.  This need exists because if I pass the topic to the parser, and it is a url, this generates an error.

As for the tests, I tried something close to the real thing but I couldn't, I'm going to study a little about them and see what I can do.

TjorvenB:
> My only concern is about that prefix `@=` is this really necessary?

The "@=" is how symfony handles the same problem in the service container configuration. You need a way to distinguish regular strings from expression strings. This way you are consistent with the underlying framework.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -84,6 +85,17 @@ public function __construct(ResourceClassResolverInterface $resourceClassResolve
         $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
         $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
         $this->reset();
+
+        $urlencode = ExpressionFunction::fromPhp('urlencode');
+        $this->expressionLanguage->addFunction($urlencode);
+
+        $this->expressionLanguage->addFunction(
+            new ExpressionFunction('iri', function ($apiResource) {
```
alanpoulain, line 93:
```suggestion
            new ExpressionFunction('iri', static function (string $apiResource) {
```

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -84,6 +85,17 @@ public function __construct(ResourceClassResolverInterface $resourceClassResolve
...
+                return sprintf('iri(%1$s)', $apiResource);
+            }, function ($arguments, $apiResource) use ($iriConverter) {
```
alanpoulain, line 95:
```suggestion
            }, static function (array $arguments, $apiResource) use ($iriConverter): string {
```

alanpoulain:
Minor comments but LGTM :+1: As @soyuka said we need at least unit tests in order to merge it.

soyuka (approved):

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -84,6 +85,17 @@ public function __construct(ResourceClassResolverInterface $resourceClassResolve
         $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
         $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
         $this->reset();
+
+        $urlencode = ExpressionFunction::fromPhp('urlencode');
```
dunglas, line 89: We should use `rawurlencode`, `utlencode` is intended to be used to create request bodies (forms). https://www.php.net/manual/en/function.rawurlencode.php

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -84,6 +85,17 @@ public function __construct(ResourceClassResolverInterface $resourceClassResolve
...
+        $this->expressionLanguage->addFunction($urlencode);
+
+        $this->expressionLanguage->addFunction(
+            new ExpressionFunction('iri', static function (string $apiResource): string {
```
dunglas, line 93: I suggest to add the type of URL to generate (absolute, relative...) as an extra optional argument. Defaulting to absolute looks good to me.

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -190,6 +202,23 @@ private function storeObjectToPublish($object, string $property): void
...
+        if ($options['topics'] ?? false) {
+            $topics = [];
+            foreach ((array) $options['topics'] as $topic) {
+                if (\is_string($topic) && 0 === strpos($topic, '@=')) {
+                    if (null === $this->expressionLanguage) {
+                        throw new \LogicException('The "@=" expression syntax cannot be used without the Expression Language component. Try running "composer require symfony/expression-language".');
+                    }
+
+                    $topics[] = $this->expressionLanguage->evaluate(substr($topic, 2), ['object' => $object]);
+                } else {
```
dunglas, line 214: Nitpicking: use `continue` instead of the `else` to reduce the complexity (I would also invert the condition to transform it into a guard clause).

dunglas (changes requested):
Very useful addition, thanks!

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
djpremier: Do you think I leave it as "rawurlencode" or do an alias for "url"?

djpremier: Yes

djpremier:
> Very useful addition, thanks!

Done 🙂

dunglas (approved):

djpremier:
@alanpoulain sorry to bother you, but could you review?

dunglas:
Thanks @djpremier

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -88,6 +89,17 @@ public function __construct(ResourceClassResolverInterface $resourceClassResolve
         $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
         $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
         $this->reset();
+
+        $rawurlencode = ExpressionFunction::fromPhp('rawurlencode', 'escape');
+        $this->expressionLanguage->addFunction($rawurlencode);
```
norkunas, line 94: As I see expression language is optional so this code will fail without it

norkunas, line 94: Also with a real request I got: `Registering functions after calling evaluate(), compile() or parse() is not supported.`

djpremier: That's true, I'll fix it. Thanks.

djpremier: What would an actual request be? Can you provide an example?

alanpoulain: Yes it should throw an exception if the component is not installed indeed. Would you do it?

alanpoulain: This one is weird. I don't know where it should be put if not in the constructor. Maybe we need to use an expression provider (https://symfony.com/doc/current/components/expression_language/extending.html#using-expression-providers)?

djpremier: @alanpolain We have two alternatives here, I can do an if and not add the functions if there is no Expression, or make it mandatory, raising the exception. Which would we choose?

djpremier: I believe that the ```Expression Provider``` won't solve in the same way, but since I don't know why, I asked for the example to try to reproduce it, because when I did the addition it worked normally

alanpoulain: I think the best approach would be to not add the functions, and then throw an exception if the user uses them. WDYT?

djpremier: It can be, I'll create the PR

djpremier: https://github.com/api-platform/core/pull/4341

norkunas: > What would a real request? Can you provide an example?

Broswer request to the api :) will check if its reproduceable everytime.

norkunas: IMHO as @alanpoulain said it would be better to use expression provider as it is done for security part https://github.com/api-platform/core/blob/main/src/Security/Core/Authorization/ExpressionLanguageProvider.php

norkunas:
There is also another problem: topics are evaluated before postFlush so for the new entities @=iri(object) cannot be used if entity uses generated strategy for ids

PR #4162, opened 2021-03-25
Add support for generating property schema with Choice restriction

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Add a `PropertySchemaChoiceRestriction` to transform [Choice](https://symfony.com/doc/current/reference/constraints/Choice.html) validation constraint into these schemas:

`@Assert\Choice(choices={"choice1", "choice2"})`
```json
{
  "enum": ["choice1", "choice2"]
}
```
`@Assert\Choice(choices={"choice1", "choice2"}, multiple=true)`
```json
{
  "type": "array",
  "items": { "enum": ["choice1", "choice2"] }
}
```

`@Assert\Choice(choices={"choice1", "choice2", "choice3", "choice4"}, multiple=true, min=2, max=4)`
```json
{
  "type": "array",
  "items": { "enum": ["choice1", "choice2", "choice3", "choice4"] },
  "minItems": 2,
  "maxItems": 4
}
```

### DIFF
--- a/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Bridge\Symfony\Validator\Metadata\Property\Restriction;
+
+use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
+use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints\Choice;
+
+/**
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ */
+final class PropertySchemaChoiceRestriction implements PropertySchemaRestrictionMetadataInterface
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @param Choice $constraint
+     */
+    public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
+    {
+        $choices = [];
+
+        if (\is_callable($choices = $constraint->callback)) {
+            $choices = $choices();
+        } elseif (\is_array($constraint->choices)) {
+            $choices = $constraint->choices;
+        }
+
+        if (!$choices) {
+            return [];
+        }
+
+        $restriction = [];
+
+        if (!$constraint->multiple) {
+            $restriction['enum'] = $choices;
+
+            return $restriction;
+        }
+
+        $restriction['type'] = 'array';
+        $restriction['items'] = ['type' => Type::BUILTIN_TYPE_STRING === $propertyMetadata->getType()->getBuiltinType() ? 'string' : 'number', 'enum' => $choices];
+
+        if (null !== $constraint->min) {
+            $restriction['minItems'] = $constraint->min;
+        }
+
+        if (null !== $constraint->max) {
+            $restriction['maxItems'] = $constraint->max;
+        }
+
+        return $restriction;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(Constraint $constraint, PropertyMetadata $propertyMetadata): bool
+    {
+        return $constraint instanceof Choice && null !== ($type = $propertyMetadata->getType()) && \in_array($type->getBuiltinType(), [Type::BUILTIN_TYPE_STRING, Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_FLOAT], true);
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.7.0
 
+* JSON Schema: Add support for generating property schema with Choice restriction (#4162)
 * JSON Schema: Add support for generating property schema with Range restriction (#4158)
 * JSON Schema: Add support for generating property schema with Unique restriction (#4159)
 * **BC**: Change `api_platform.listener.request.add_format` priority from 7 to 28 to execute it before firewall (priority 8) (#3599)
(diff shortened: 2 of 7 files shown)


src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
```diff
@@ -0,0 +1,74 @@
...
+
+use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
+use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints\Choice;
+
+/**
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ */
+class PropertySchemaChoiceRestriction implements PropertySchemaRestrictionMetadataInterface
```
dunglas, line 24:
```suggestion
final class PropertySchemaChoiceRestriction implements PropertySchemaRestrictionMetadataInterface
```

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
+    {
+        if (!\in_array($builtinType = $propertyMetadata->getType()->getBuiltinType(), [Type::BUILTIN_TYPE_STRING, Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_FLOAT], true)) {
+            return [];
+        }
+
+        \assert($constraint instanceof Choice);
```
dunglas, line 35: We usally don't use `assert` in API Platform.

dunglas (changes requested):
Nice! This should also target `main`

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
```diff
@@ -0,0 +1,72 @@
...
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ */
+final class PropertySchemaChoiceRestriction implements PropertySchemaRestrictionMetadataInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function create(Constraint $constraint, PropertyMetadata $propertyMetadata): array
+    {
+        if (!\in_array($builtinType = $propertyMetadata->getType()->getBuiltinType(), [Type::BUILTIN_TYPE_STRING, Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_FLOAT], true)) {
```
alanpoulain, line 31: Shouldn't it be in the `supports` method?

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
```diff
@@ -0,0 +1,72 @@
...
+            return [];
+        }
+
+        if (\is_callable($choices = $constraint->callback)) {
+            $choices = $choices();
+        } elseif (\is_array($constraint->choices)) {
+            $choices = $constraint->choices;
+        } else {
+            $choices = [];
```
alanpoulain, line 40: You should avoid the `else` IMO (add a default value before).

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
```diff
@@ -0,0 +1,72 @@
...
+
+            if (null !== $constraint->min) {
+                $restriction['minItems'] = $constraint->min;
+            }
+
+            if (null !== $constraint->max) {
+                $restriction['maxItems'] = $constraint->max;
+            }
+        } else {
+            $restriction['enum'] = $choices;
```
alanpoulain, line 59: Would be better if it was before with an early return.

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
```diff
@@ -0,0 +1,72 @@
...
+            $choices = [];
+        }
+
+        if (!$choices) {
+            return [];
+        }
+
+        if ($constraint->multiple) {
+            $restriction['type'] = 'array';
+            $restriction['items'] = ['type' => Type::BUILTIN_TYPE_STRING !== $builtinType ? 'string' : 'number', 'enum' => $choices];
```
alanpoulain, line 49: Isn't the condition reversed?
```suggestion
            $restriction['items'] = ['type' => Type::BUILTIN_TYPE_STRING === $builtinType ? 'string' : 'number', 'enum' => $choices];
```

alanpoulain: I think the tests are not working as expected id it's the case.

norkunas: this is by accident, now locally tests was failing for me, will push in a second

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
norkunas: But then PHPStan complains:
```
  33     Access to an undefined property                                                                 
         Symfony\Component\Validator\Constraint::$callback.                                              
  35     Access to an undefined property                                                                 
         Symfony\Component\Validator\Constraint::$choices.                                               
  45     Access to an undefined property                                                                 
         Symfony\Component\Validator\Constraint::$multiple.                                              
  54     Access to an undefined property                                                                 
         Symfony\Component\Validator\Constraint::$min.                                                   
  58     Access to an undefined property                                                                 
         Symfony\Component\Validator\Constraint::$max.       
```

alanpoulain: You just need to add the right PHPDoc.

norkunas: okay, let's see how it goes

alanpoulain (approved):

alanpoulain:
Thank you very much!

bouffa:
I am trying to make this work with Api Platform Admin module with no luck. 
I inspected the docs.jsonld served by the api, and there is no enum specified. However, I can see the enum in the docs.json. Am I missing something?
Thanks

PR #4165, opened 2021-03-26
Add an option to disable query parameter validation

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1329

Query parameter validation can be costly if you are not using any required filters, so we should have an option to disable it.
If you think that this is a good addition, I'll add some tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.7.0
 
+* Validator: Add an option to disable query parameter validation (#4165)
 * JSON Schema: Add support for generating property schema with Choice restriction (#4162)
 * JSON Schema: Add support for generating property schema with Range restriction (#4158)
 * JSON Schema: Add support for generating property schema with Unique restriction (#4159)
--- a/src/Annotation/ApiResource.php
+++ b/src/Annotation/ApiResource.php
@@ -71,6 +71,7 @@ use ApiPlatform\Core\Exception\InvalidArgumentException;
  *     @Attribute("urlGenerationStrategy", type="int"),
  *     @Attribute("validationGroups", type="mixed"),
  *     @Attribute("exceptionToStatus", type="array"),
+ *     @Attribute("queryParameterValidationEnabled", type="bool")
  * )
  */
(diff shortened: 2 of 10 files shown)


alanpoulain:
Is it really? Maybe a Blackfire profile would help to see if it's the case.

norkunas:
> Is it really? Maybe a Blackfire profile would help to see if it's the case.

When filters depends on the dynamic structure which loads from external storages then yes.

alanpoulain:
> > Is it really? Maybe a Blackfire profile would help to see if it's the case.
> 
> When filters depends on the dynamic structure which loads from external storages then yes.

It's an edge case, isn't it? If it is, I think the best approach is to decorate the listener instead of having a specific and not used much configuration parameter.

norkunas:
> > > Is it really? Maybe a Blackfire profile would help to see if it's the case.
> > 
> > When filters depends on the dynamic structure which loads from external storages then yes.
> 
> It's an edge case, isn't it? If it is, I think the best approach is to decorate the listener instead of having a specific and not used much configuration parameter.

Why to run some validations when you don't use them at all? IIRC @soyuka in slack said that it's by default should not be enabled at all.

norkunas:
Also read/deserialize/serialize/validate can be disabled so why not query param validator?

antograssiot:
Didn't checked the PR but it is indeed costly and should be opt-in, totally 👍, we've discovered that on your project too and discussed it with @soyuka

alanpoulain:
OK! You can add some tests then @norkunas.

norkunas:
Added some tests.

alanpoulain (approved):
LGTM. Don't forget the CHANGELOG and the documentation.

norkunas:
Added some docs for this PR.

alanpoulain:
Nice work. Thank you @norkunas.

PR #4171, opened 2021-03-29
Make sure the order of nested order filters is preserved

jnugh:
Order filters are being written with underscores in GraphQL but are sent along with dots to access nested entities.

The previous solution appended the dot version to the order array. This leads to an issue where every order filter that accesses a nested property will always be added to the query after local properties.

Looking at the test case: The first order that WAS applied was `localField`. It should have been `some.field` which is now the case.

| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       |  I did not create any
| License       | MIT
| Doc PR        | -

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Swagger UI: Remove Google fonts (#4112)
 * Doctrine: Revert #3774 support for binary UUID in search filter (#4134) 
 * Do not override Vary headers already set in the Response 
+* GraphQL: Make sure the order of order filters is preserved if nested resources are used
 
 ## 2.6.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * Swagger UI: Remove Google fonts (#4112)
 * Doctrine: Revert #3774 support for binary UUID in search filter (#4134) 
 * Do not override Vary headers already set in the Response 
-* GraphQL: Make sure the order of order filters is preserved if nested resources are used
+* GraphQL: Make sure the order of order filters is preserved if nested resources are used (#4171)
 
 ## 2.6.3
 
--- a/src/GraphQl/Resolver/Stage/ReadStage.php
+++ b/src/GraphQl/Resolver/Stage/ReadStage.php
@@ -161,7 +161,7 @@ final class ReadStage implements ReadStageInterface
                 $filters =
                     \array_slice($filters, 0, $index + 1) +
                     [str_replace($this->nestingSeparator, '.', $name) => $value] +
-                    \array_slice($filters, $index);
+                    \array_slice($filters, $index + 1);
             }
         }
 
--- a/src/GraphQl/Resolver/Stage/ReadStage.php
+++ b/src/GraphQl/Resolver/Stage/ReadStage.php
@@ -157,7 +157,11 @@ final class ReadStage implements ReadStageInterface
 
             if (\is_string($name) && strpos($name, $this->nestingSeparator)) {
                 // Gives a chance to relations/nested fields.
-                $filters[str_replace($this->nestingSeparator, '.', $name)] = $value;
+                $index = array_search($name, array_keys($filters), true);
+                $filters =
+                    \array_slice($filters, 0, $index + 1) +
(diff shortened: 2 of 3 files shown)


src/GraphQl/Resolver/Stage/ReadStage.php
```diff
@@ -157,7 +157,11 @@ private function getNormalizedFilters(array $args): array
 
             if (\is_string($name) && strpos($name, $this->nestingSeparator)) {
                 // Gives a chance to relations/nested fields.
-                $filters[str_replace($this->nestingSeparator, '.', $name)] = $value;
+                $index = array_search($name, array_keys($filters), true);
+                $filters =
+                    \array_slice($filters, 0, $index + 1) +
+                    [str_replace($this->nestingSeparator, '.', $name) => $value] +
+                    \array_slice($filters, $index);
```
alanpoulain, line 164: Why starting at index?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Swagger UI: Remove Google fonts (#4112)
 * Doctrine: Revert #3774 support for binary UUID in search filter (#4134) 
 * Do not override Vary headers already set in the Response 
+* GraphQL: Make sure the order of order filters is preserved if nested resources are used
```
alanpoulain, line 16:
```suggestion
* GraphQL: Make sure the order of order filters is preserved if nested resources are used (#4171)
```

tests/GraphQl/Resolver/Stage/ReadStageTest.php
```diff
@@ -221,6 +222,45 @@ public function testApplyCollection(array $args, ?string $rootClass, ?array $sou
         $this->assertSame($expectedResult, $result);
     }
 
+    /**
+     * Test a regression where the order in which order fitlers were applied
+     * was wrong when the data had been ordered by a nested field.
+     */
+    public function testPreserveOrderOfOrderFitlersIfNested()
```
alanpoulain, line 229:
```suggestion
    public function testPreserveOrderOfOrderFiltersIfNested(): void
```

tests/GraphQl/Resolver/Stage/ReadStageTest.php
```diff
@@ -221,6 +222,45 @@ public function testApplyCollection(array $args, ?string $rootClass, ?array $sou
         $this->assertSame($expectedResult, $result);
     }
 
+    /**
+     * Test a regression where the order in which order fitlers were applied
+     * was wrong when the data had been ordered by a nested field.
+     */
```
alanpoulain, line 228: I think this comment is unnecessary.

alanpoulain:
Hello! A tricky one :slightly_smiling_face: Thanks for spotting it and fixing it.

src/GraphQl/Resolver/Stage/ReadStage.php
jnugh: It does not make a difference in the output as the duplicated key gets merged so I did not notice that this code adds it twice. Will add `+1` there as well.

jnugh:
It's probably not a super common use case to sort on a child entity first...
Thanks for the great work on api-platform! 😄

tests/GraphQl/Resolver/Stage/ReadStageTest.php
```diff
@@ -221,6 +222,41 @@ public function testApplyCollection(array $args, ?string $rootClass, ?array $sou
         $this->assertSame($expectedResult, $result);
     }
 
+    public function testPreserveOrderOfOrderFitlersIfNested(): void
```
alanpoulain, line 225:
```suggestion
    public function testPreserveOrderOfOrderFiltersIfNested(): void
```

tests/GraphQl/Resolver/Stage/ReadStageTest.php
jnugh: A classic `Fitler`. Missed that one, sorry. Applied 👍🏽

alanpoulain (approved):

alanpoulain:
Thank you @jnugh.

PR #3472, opened 2020-03-30
Fix behaviour of "order" attribute

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

/cc @soyuka

TODO:

- [ ] Add tests

PR #4178, opened 2021-03-31
Revert using defaults.order as collection.order

### LINKED ISSUE #4170: Conflict between defaults.order and collection.order configs
**API Platform version(s) affected**: 2.6.3

**Description**
The `defaults.order` config value, which would work as the global equivalent of https://api-platform.com/docs/core/default-order/#overriding-default-order (fields _array_), is also used in place of `collection.order` (direction _string_), which causes a TypeError.

**How to reproduce**
Config:
```yaml
api_platform:
    defaults:
        order: [createdAt, id]
```

Error (on collection GET request):
> Argument 1 passed to ApiPlatform\\Core\\Bridge\\Doctrine\\Orm\\Extension\\OrderExtension::__construct() must be of the type string or null, array given

**Possible Solution**
Revert the order-related changes of https://github.com/api-platform/core/commit/dc461fcc504e8b52e5134013f2f3c7d947325dd0 (note that, unlike the `collection.pagination.*` non-`*_parameter_name` keys, `collection.order` is _not_ deprecated, so I don't see the reason for wanting to move it to `defaults.order`, even more so since that key was already used after #3151)

**Additional Context**
Workaround for now: use the explicit `attributes` sublevel:
```yaml
api_platform:
    defaults:
        attributes:
            order: [createdAt, id]
```

guilliamxavier:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4170
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Bridge/Doctrine/MongoDbOdm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/MongoDbOdm/Extension/OrderExtension.php
@@ -58,7 +58,7 @@ final class OrderExtension implements AggregationCollectionExtensionInterface
             if (empty($defaultOrder)) {
                 $defaultOrder = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('order');
             }
-            if (\is_array($defaultOrder)) {
+            if (null !== $defaultOrder) {
                 foreach ($defaultOrder as $field => $order) {
                     if (\is_int($field)) {
                         // Default direction
--- a/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
+++ b/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php
@@ -56,7 +56,7 @@ final class OrderExtension implements ContextAwareQueryCollectionExtensionInterf
             if (empty($defaultOrder)) {
                 $defaultOrder = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('order');
             }
-            if (\is_array($defaultOrder)) {
+            if (null !== $defaultOrder) {
                 foreach ($defaultOrder as $field => $order) {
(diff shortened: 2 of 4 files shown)


src/Bridge/Doctrine/MongoDbOdm/Extension/OrderExtension.php
```diff
@@ -58,7 +58,7 @@ public function applyToCollection(Builder $aggregationBuilder, string $resourceC
             if (empty($defaultOrder)) {
                 $defaultOrder = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('order');
             }
-            if (\is_array($defaultOrder)) {
+            if (null !== $defaultOrder) {
```
alanpoulain, line 61: Isn't it better to keep `is_array`?

guilliamxavier: I don't know, because:
- I consistently reverted all the changes of 3492922c2733de934cf4798893d907755640869d
- With `null !==` (like it was originally) an incorrect config (e.g. a string value) gives you a Warning "Invalid argument supplied for foreach()" (then you fix your config :wink:), with `is_array()` it is silently ignored (until you notice your config isn't applied?)
- See also #3472

But if you think this part of the revert is too risky, I can drop it

alanpoulain: Not sure which one is the better :sweat_smile: Keep it like this, it's alright.

alanpoulain:
Thanks :slightly_smiling_face:

PR #4179, opened 2021-03-31
Do not overwrite variable with unknown value

guilliamxavier:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4162
| License       | MIT
| Doc PR        | 

PhpStorm was complaining on line 33:
> Unused local variable 'choices'. The value of the variable is overwritten immediately.

Also if you think about it, in the case of e.g. `@Assert\Choice(callback={"App\Entity\Foo", "getBats"})`, i.e.
- `$constraint->callback === ['App\Entity\Foo', 'getBats']` (_note: typo_, intended `'getBars'`)
- `$constraint->choices === null`
- `$constraint->multiple === false`

we would end up with an incorrect `"enum": ["App\\Entity\\Foo", "getBats"]`


alanpoulain:
Nice spotting, thanks.

PR #4180, opened 2021-04-01
Handle Compound validator constraints when generating property metadata

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

In #4139 I have missed that there is a new [Compound](https://symfony.com/doc/current/reference/constraints/Compound.html) constraint which helps to create reusable validations.

This adds support for Compound constraint, but only when used at the top level without nesting.

I think this factory should be refactored later to properly handle constraint nesting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.7.0
 
+* JSON Schema: Manage Compound constraint when generating property metadata (#4180)
 * Validator: Add an option to disable query parameter validation (#4165)
 * JSON Schema: Add support for generating property schema with Choice restriction (#4162)
 * JSON Schema: Add support for generating property schema with Range restriction (#4158)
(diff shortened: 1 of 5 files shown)


guilliamxavier:
Question @ team: #4139 was merged into `2.6` but this is currently targeting `main`, wouldn't https://github.com/api-platform/core/pull/4139#issuecomment-801307028 apply here too?

(PS: thanks @norkunas for all your PRs!)

alanpoulain:
@guilliamxavier it's because I first thought it was better to have it in `2.6` but after talking with @dunglas and @soyuka, we agreed it was dangerous to modify the generated OpenAPI schema in a patch release and that it could be considered as a BC break.
However we didn't choose to revert the merged PR, but for the following ones, they should target `main`, like this one.

guilliamxavier:
@alanpoulain: Thanks for explaining.
By the way then, why not revert the previous ones (#4139 and #4147) on `2.6` (but keep them on `main`), while 2.6.4 has not been released yet?

alanpoulain:
Ideally we should. But since the constraints are only available from Symfony 5.1, I think the risk is mitigated for the already merged PR.

alanpoulain (approved):

alanpoulain:
Thank you again @norkunas :slightly_smiling_face:

PR #4182, opened 2021-04-01
Add support for generating property schema with Collection restriction

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Add a `PropertySchemaCollectionRestriction` to transform [Collection](https://symfony.com/doc/current/reference/constraints/Collection.html) validation constraint into json schema.

### DIFF
--- a/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
@@ -18,7 +18,6 @@ use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints\Collection;
 use Symfony\Component\Validator\Constraints\Optional;
 use Symfony\Component\Validator\Constraints\Required;
-use Symfony\Component\Validator\Constraints\Type;
 
 /**
  * @author Tomas Norkūnas <norkunas.tom@gmail.com>
@@ -81,16 +80,10 @@ final class PropertySchemaCollectionRestriction implements PropertySchemaRestric
      */
     private function mergeConstraintRestrictions(Constraint $constraint, PropertyMetadata $propertyMetadata): array
     {
-        $propertyRestrictions = [['type' => 'string']];
+        $propertyRestrictions = [];
         $nestedConstraints = method_exists($constraint, 'getNestedContraints') ? $constraint->getNestedContraints() : $constraint->constraints;
 
         foreach ($nestedConstraints as $nestedConstraint) {
-            if ($nestedConstraint instanceof Type) {
-                /** @var string $nestedType */
-                $nestedType = $nestedConstraint->type;
-                $propertyRestrictions[0]['type'] = $this->getConstraintType($nestedType);
-            }
-
             foreach ($this->restrictionsMetadata as $restrictionMetadata) {
                 if ($restrictionMetadata->supports($nestedConstraint, $propertyMetadata) && !empty($nestedConstraintRestriction = $restrictionMetadata->create($nestedConstraint, $propertyMetadata))) {
                     $propertyRestrictions[] = $nestedConstraintRestriction;
@@ -99,30 +92,5 @@ final class PropertySchemaCollectionRestriction implements PropertySchemaRestric
         }
 
         return array_merge([], ...$propertyRestrictions);
-    }
-
-    private function getConstraintType(string $type): string
-    {
-        if (\in_array($type, ['bool', 'boolean'], true)) {
-            return 'boolean';
-        }
-
-        if (\in_array($type, ['int', 'integer', 'long'], true)) {
-            return 'integer';
-        }
-
-        if (\in_array($type, ['float', 'double', 'real', 'numeric'], true)) {
-            return 'number';
-        }
-
-        if (\in_array($type, ['array', 'iterable'], true)) {
-            return 'array';
-        }
-
-        if ('object' === $type) {
-            return 'object';
-        }
-
-        return 'string';
     }
 }
--- a/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
@@ -1,0 +1,128 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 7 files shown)


norkunas:
Not sure about the implementation, suggestions welcome :smiley:

tests/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestrictionTest.php
```diff
@@ -0,0 +1,121 @@
...
+                        new NotBlank(),
+                    ]),
+                    'email' => [
+                        new NotNull(),
+                        new Length(['min' => 2, 'max' => 255]),
+                        new Email(['mode' => Email::VALIDATION_MODE_LOOSE]),
+                    ],
+                    'phone' => new Optional([
+                        new \Symfony\Component\Validator\Constraints\Type(['type' => 'string']),
+                        new Regex(['pattern' => '/^[+]*[(]{0,1}[0-9]{1,4}[)]{0,1}[-\s\./0-9]*$/']),
```
guilliamxavier, line 94: Unrelated to the feature but: you must escape the delimiter (currently Warning "preg_match(): Unknown modifier '0'").
Also no need to escape the dot inside `[]`.

```suggestion
                        new Regex(['pattern' => '/^[+]*[(]{0,1}[0-9]{1,4}[)]{0,1}[-\s.\/0-9]*$/']),
```

Same inside `DummyCollectionValidatedEntity`

norkunas: thanks

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * @param Required|Optional $constraint
+     */
+    private function mergeConstraintRestrictions(Constraint $constraint, PropertyMetadata $propertyMetadata): array
+    {
+        $propertyRestrictions = [['type' => 'string']];
+
+        foreach ($constraint->getNestedContraints() as $nestedContraint) {
```
guilliamxavier, line 86: This causes an error with Symfony 4.4 because `Composite::getNestedContraints()` was only added in 5.1 (but I don't know how to work around, sorry 😖)

norkunas: Yes, I saw the errors, but looking at the source https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Validator/Constraints/Composite.php the method is there :thinking:

norkunas: Another option probably would be directly use `$constraints` property

guilliamxavier: Correction indeed: it was added in v3.4.44, v4.4.12, v5.1.4 and v5.2.0 (the "lowest" test uses symfony/validator v4.4.0)

norkunas: now green :)

norkunas:
phpstan failures are unrelated

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        $propertyRestrictions = [['type' => 'string']];
+        $nestedConstraints = method_exists($constraint, 'getNestedContraints') ? $constraint->getNestedContraints() : $constraint->constraints;
+
+        foreach ($nestedConstraints as $nestedConstraint) {
+            if ($nestedConstraint instanceof Type) {
+                /** @var string $nestedType */
+                $nestedType = $nestedConstraint->type;
+                $propertyRestrictions[0]['type'] = $this->getConstraintType($nestedType);
+            }
```
alanpoulain, line 92: Should we really add the type? We don't add it for a property, isn't it?

norkunas: trully not sure :)

norkunas: if the property metadata would have types for nested fields then we probably shouldn't, but when i declare a nested field which must be an integer i'd expect to that the int would be exposed

alanpoulain: I'm not sure to understand. Today, the `type` restriction is not added to the schema for a top-level property. Why would it be defined for a nested one?

norkunas: As you resist from this I've removed the type guessing :) I agree that this code should not be responsible for this, but then I think api-platform should ensure that it'd at least possible to do this:
```
/**
 * @var array{age:int, social:array{githubUsername:?string}}
 * @Assert\Collection(
 *     fields={
 *         "age"=..,
 *         "social"=@Assert\Collection({
 *            fields={"githubUsername"=..}
 *         })
 *     }
 * )
 */
```
and the types would be exposed from typedef

alanpoulain: I'm not saying you are wrong :slightly_smiling_face: It's just that we need to be coherent. We should add the type to the top-level properties beforehand if we want to add it to the nested ones.
I'm not sure why there is no type for the top-level properties. Anyway it's not the `Validator` responsibility to add the types, if we add it, it should be elsewhere.

norkunas: I think that for top-level properties types are fetched with property-info component, but not sure if it can provide nested ones

alanpoulain: But they are not added to the `schema`, aren't they?

norkunas: could be :)

alanpoulain (approved):

alanpoulain:
Thank you!

norkunas:
Thanks for merging!

PR #4184, opened 2021-04-01
feat: add `security_post_denormalize` in `ApiProperty`

DavidBennettUK:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

While implementing https://github.com/api-platform/core/pull/4143 I noticed that it's not possible to use `object` in the security expression of `ApiProperty` - not only for GraphQL, but for all.

It was already possible to use `object` for read operations - but doing so would cause unexpected exceptions to be thrown on updates/creates, as the `object` is actually a class string. For example, the security expression `object.getOwner() == user` would throw an exception when performing an update as `object` would be a class string - which the developer wouldn't expect. This exception would happen even if you weren't updating that specific field, as the Symfony `AbstractObjectNormalizer::denormalize()` calls `getAllowedAttributes` - which runs security on every single field, due to `canAccessAttribute` being overridden in `AbstractItemNormalizer`. 

There didn't seem to be a way to work around this in the expressions, as there isn't an obvious way to check if a variable is an object or string in the expression language.

This PR updates this so that `object` will always be the object or null - which can easily be checked for in the expression.

Also added is a `security_post_denormalize` which works similarly to those on `ApiResource`. This lets you be guaranteed that `object` isn't null and will also give you access to `previous_object` for more advanced checks.

This should hopefully not be a BC break unless developer workarounds were put in place possibly. For safety, and because it depends on https://github.com/api-platform/core/pull/4143, it is targeting main.


alanpoulain:
Related: https://github.com/api-platform/core/pull/3966
@Toflar, since you are the one who have created the PR, could you see if it's OK for you?

@DavidBennettUK I think @Toflar is able to distinguish between an object and a class string by using a custom expression language function (`user_has_role_for_organisation`).

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -249,7 +249,34 @@ public function denormalize($data, $class, $format = null, array $context = [])
             return $item;
         }
 
-        return parent::denormalize($data, $resourceClass, $format, $context);
+        $objectToPopulate = $context[self::OBJECT_TO_POPULATE] ?? null;
+        $isUpdate = null !== $objectToPopulate;
+
+        // Filter out input attributes that aren't allowed for updates
+        if ($isUpdate) {
+            $data = array_filter($data, function (string $attribute) use ($context, $objectToPopulate) {
```
alanpoulain, line 257: Why filtering the input? Was it possible to modify a property of a resource without the right to do so before?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -249,7 +249,34 @@ public function denormalize($data, $class, $format = null, array $context = [])
...
+        // Revert attributes that aren't allowed to be changed after a post-denormalize check
+        foreach (array_keys($data) as $attribute) {
+            if (!$this->canAccessAttributePostDenormalize($object, $previousObject, $attribute, $context)) {
+                if ($isUpdate) {
+                    $object[$attribute] = $previousObject[$attribute];
+                } else {
+                    $reflection = (new \ReflectionClass($object));
+                    $property = $reflection->getProperty($attribute);
+                    $property->setAccessible(true);
+                    $property->setValue($object, $reflection->getDefaultProperties()[$attribute] ?? null);
```
alanpoulain, line 274: Shouldn't you use the property accessor here?

Toflar:
In my case, I cannot check on the roles of the user directly because the same user has different roles for different organisations. So what I did is this:

```php
    /**
     * @ApiProperty(security="user_has_role_for_organisation('ROLE_OWNER', object)")
     */
    private BillingInfo $billingInfo;
```

And then a simple

```php
class AccessExpressionProvider implements ExpressionFunctionProviderInterface
{
    public function getFunctions(): array
    {
        return [
            new ExpressionFunction('user_has_role_for_organisation', function (string $role, Organisation $organisation) {
                return sprintf('$user->hasRoleForOrganisation(%s, %s)', $role, $organisation);
            }, function (array $variables, string $role, Organisation $organisation) {
                return $variables['user']->hasRoleForOrganisation($role, $organisation);
            }),
        ];
    }
}
```

This should still work as far as I can see.

alanpoulain:
That's weird. If I understand what @DavidBennettUK is saying, your function should fail when doing an update / create, because `$organisation` is then a string and not an `Organisation` object.

Toflar:
Everywhere I use this, it's only `GET` ;)

alanpoulain:
It makes sense then! It should fail if you try an update.

Toflar:
Might very well be but as this is not a requirement for me, I have no tests for that :)

alanpoulain:
No problem :slightly_smiling_face:

src/Serializer/AbstractItemNormalizer.php
DavidBennettUK: You're right that there was already a security pass - `AbstractObjectNormalizer::denormalize()` calls `AbstractItemNormalizer::getAllowedAttributes()` but only with the class string, so `object` can't be used. This addition does another pass over, calling the `security` expression with a populated `object` variable. Without this, `security` will always be called with a null object for updates, which may be unexpected. We could take this out and force use of `security_post_denormalize` if the developer wants an object, but I think an object would be expected for updates as it already exists.

src/Annotation/ApiProperty.php
```diff
@@ -136,6 +138,7 @@ public function __construct(
         ?array $openapiContext = null,
         ?bool $push = null,
         ?string $security = null,
+        ?string $securityPostDenormalize = null,
```
soyuka, line 141: should be the last value of the constructor no?

src/Serializer/AbstractItemNormalizer.php
soyuka: I'd be in favor of reducing the complexity in our code, enforcing the use of `security_post_denormalize`

DavidBennettUK:
@soyuka thanks for the feedback, I've made the suggested changes 👍

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -249,7 +249,22 @@ public function denormalize($data, $class, $format = null, array $context = [])
             return $item;
         }
 
-        return parent::denormalize($data, $resourceClass, $format, $context);
+        $previousObject = null !== $objectToPopulate ? clone $objectToPopulate : null;
+        $object = parent::denormalize($data, $resourceClass, $format, $context);
+
+        // Revert attributes that aren't allowed to be changed after a post-denormalize check
+        foreach (array_keys($data) as $attribute) {
```
alanpoulain, line 256: Was it possible to modify a property of a resource without the right to do so before?
If so, isn't it a bugfix for 2.6?

DavidBennettUK: Not that I'm aware of. This is only reverting the properties that fail the new `security_post_denormalize` check. Properties that fail the `security` check are still filtered from denormalization via `AbstractItemNormalizer::getAllowedAttributes()`

tests/Fixtures/TestBundle/Document/SecuredDummy.php
```diff
@@ -83,6 +83,17 @@ class SecuredDummy
      */
     private $adminOnlyProperty = '';
 
+    /**
+     * @var string Secret property, only readable/writable by owners
+     *
+     * @ODM\Field
+     * @ApiProperty(
+     *     security="object == null or object.getOwner() == user",
```
alanpoulain, line 91: Is it still useful?

tests/Fixtures/TestBundle/Entity/SecuredDummy.php
```diff
@@ -84,6 +84,17 @@ class SecuredDummy
      */
     private $adminOnlyProperty = '';
 
+    /**
+     * @var string Secret property, only readable/writable by owners
+     *
+     * @ORM\Column
+     * @ApiProperty(
+     *     security="object == null or object.getOwner() == user",
```
alanpoulain, line 92: Is it still useful?

alanpoulain:
LGTM. Don't forget the documentation entry.

tests/Fixtures/TestBundle/Entity/SecuredDummy.php
DavidBennettUK: As the `object` isn't available in creates/updates, in this scenario you'd need to pass the `security` check if the `object`is `null`, as otherwise it would fail this check (`object.getOwner()` would throw a null pointer exception) and never run `security_post_denormalize`.

DavidBennettUK:
Documentation PR https://github.com/api-platform/docs/pull/1349

tests/Fixtures/TestBundle/Entity/SecuredDummy.php
alanpoulain: But if there is already a `securityPostDenormalize`, why a `security` is needed?

alanpoulain: Found the answer myself: it's because it is used for the `get` operation too.

PR #4185, opened 2021-04-02
Add support for generating property schema format for Url and Hostname

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Add a `PropertySchemaUrlRestriction` to transform [Url](https://symfony.com/doc/current/reference/constraints/Url.html) validation constraint into json schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.7.0
 
+* JSON Schema: Add support for generating property schema format for Url and Hostname (#4185)
 * JSON Schema: Add support for generating property schema with Count restriction (#4186)
 * JSON Schema: Manage Compound constraint when generating property metadata (#4180)
 * Validator: Add an option to disable query parameter validation (#4165)
--- a/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaFormat.php
+++ b/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaFormat.php
@@ -16,8 +16,10 @@ namespace ApiPlatform\Core\Bridge\Symfony\Validator\Metadata\Property\Restrictio
 use ApiPlatform\Core\Metadata\Property\PropertyMetadata;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints\Email;
+use Symfony\Component\Validator\Constraints\Hostname;
 use Symfony\Component\Validator\Constraints\Ip;
 use Symfony\Component\Validator\Constraints\Ulid;
+use Symfony\Component\Validator\Constraints\Url;
 use Symfony\Component\Validator\Constraints\Uuid;
 
 /**
@@ -34,6 +36,14 @@ class PropertySchemaFormat implements PropertySchemaRestrictionMetadataInterface
     {
         if ($constraint instanceof Email) {
             return ['format' => 'email'];
+        }
+
+        if ($constraint instanceof Url) {
+            return ['format' => 'uri'];
+        }
+
+        if ($constraint instanceof Hostname) {
+            return ['format' => 'hostname'];
         }
 
         if ($constraint instanceof Uuid) {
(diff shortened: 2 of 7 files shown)


guilliamxavier:
Please remove me from the description :laughing: I'm already checking https://github.com/api-platform/core/pulse from time to time these days (also I'm not part of the team)

If it may help: the test failure for "lowest" is related to https://github.com/symfony/validator/commit/adab2122723e3f59cc091446999138ff83825ffb (introduced in v4.4.8)

But string manipulation and generating a huge pattern in the schema... is it really worth it? 🤔 Wouldn't it be enough to add something like this in the existing `PropertySchemaFormat`?

```diff
         if ($constraint instanceof Email) {
             return ['format' => 'email'];
         }
+
+        if ($constraint instanceof Url) {
+            return ['format' => 'uri'];
+        }
 
         if ($constraint instanceof Uuid) {
             return ['format' => 'uuid'];
         }
```

This would not be strictly equivalent (also ambiguity between "URL", "URI" and "URI reference"), but that's already true for `Email` (3 different validation modes)... Also `Url` is probably less used...

Anyway let's wait for a proper review from a team member :smile:

norkunas:
I just cc'ed because you were interested in these PR's :D

norkunas:
IMHO if backend validates based on some pattern then it should be exposed nevertheless..

norkunas:
Included also Hostname constraint as it's uri related

alanpoulain:
I agree with @guilliamxavier. Since the format is used, I don't think it's a good idea to have the regexp as well.
Since URI is a specification, there should be no difference between the "client" regular expression and the backend one.

norkunas:
But how would you distinguish then allowing only some other specific protocols?

alanpoulain:
You would not be able to, that's right. Personally, I think it's a little bit too much detailed though. And the implementation is very dependent of the pattern in `UrlValidator`. If something change, it will break.
But maybe I'm wrong on this one. @soyuka, do you have an opinion on this matter?

src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaUrlRestriction.php
```diff
@@ -0,0 +1,59 @@
...
+                $line = trim(substr($line, 0, $commentPos));
+            }
+
+            return $line;
+        }, explode("\n", $pattern));
+
+        $pattern = substr(implode('', $pattern), 1, -4); // strip options and delimiters
+        $pattern = sprintf($pattern, implode('|', $constraint->protocols));
+
+        return ['pattern' => $pattern];
```
soyuka, line 49: I'd suggest to use the available `uri` format on string hence:

```
[
  'type' => 'string',
  'format' => 'uri'
]
```

There are many `URL` formats out there and many corresponding formats on the json-schema specification (http://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3). Might be that Url is actually too broad here and it would need the user to specify the format himself?

Using `#[ApiProperty(schema: ['format' => 'uri'])]` should work (needs check).

norkunas: > I don't get the use of a pattern here, json-schema already validates uris. I'd suggest to use the available `uri` format on string hence:
> 
> ```
> [
>   'type' => 'string',
>   'format' => 'uri'
> ]
> ```
> 
> There are many `URL` formats out there and many corresponding formats on the json-schema specification (http://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3). Might be that Url is actually too broad here and it would need the user to specify the format himself?
> 
> Using `#[ApiProperty(schema: ['format' => 'uri'])]` should work (needs check).

I've added uri format to PropertySchemaFormat

soyuka: okay got it sorry I missed this part, you don't need the regex, it's the responsibility of the schema validator to validate an URI not ours in my opinion.

norkunas: Okay will remove that part

norkunas:
Removed the pattern part.

soyuka (approved):

soyuka:
Tyvm @norkunas !

PR #4187, opened 2021-04-02
Make twig dependency lazy

### LINKED ISSUE #1149: Drop Twig Dependency

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #1149 
| License       | MIT

`twig` is only a part of `require-dev`. But `api_platform.swagger.action.ui` and `api_platform.swagger_ui.action` are requiring it when swagger doc is enabled. As a result, theses services are requiring an unexisting service.

Here is a tiny reproducer:
```bash
$ composer create-project symfony/skeleton:^4.4 # Same goes for 5.x versions
$ cd skeleton
$ composer req api-platform/core:2.6 
[...]
Script cache:clear returned with error code 1
!!  
!!  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:
!!                                                                                 
!!    The service "api_platform.swagger.action.ui" has a dependency on a non-exis  
!!    tent service "twig". 
```

This fix approach is to make twig dependency lazy and throwing an exception asking the user to require twig bundle when calling swagger controllers.

### DIFF
--- a/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
+++ b/src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
@@ -67,7 +67,7 @@ final class SwaggerUiAction
      * @param int[]      $swaggerVersions
      * @param mixed|null $assetPackage
      */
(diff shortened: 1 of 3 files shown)


src/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -95,6 +95,10 @@ public function __construct(ResourceNameCollectionFactoryInterface $resourceName
         $this->swaggerUiAction = $swaggerUiAction;
         $this->assetPackage = $assetPackage;
 
+        if (null === $this->twig) {
+            throw new \RuntimeException('You cannot use the "twig" service since the Twig bundle is not installed. Try running "composer require symfony/twig-bundle".');
```
alanpoulain, line 99: WDYT of adding a more user-related exception?
'The documentation cannot be displayed since the Twig bundle is not installed. Try running "composer require symfony/twig-bundle".'

mtarld: This is much better indeed! Thanks

dunglas (approved):

alanpoulain:
Thank you @mtarld.

PR #3731, opened 2020-09-21
Add `swagger_ui_extra_configuration` to Swagger configuration

dvc:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #3730
| License       | MIT
| Doc PR        | api-platform/docs/pull/1168

PR #4193, opened 2021-04-07
FIX: ApiPlatform Extension extra configuration - merging arrays

camohob:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #...
| License       | MIT

In src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php:401 used ?? between `openapi` and `swagger` for `swagger_ui_extra_configuration`, but their value cannot be null - `[]` by default. So, `$config['swagger']['swagger_ui_extra_configuration']` can never be used.

I suggest using `array_merge` to solve this problem.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -398,7 +398,10 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.enable_swagger_ui', $config['enable_swagger_ui']);
         $container->setParameter('api_platform.enable_re_doc', $config['enable_re_doc']);
         $container->setParameter('api_platform.swagger.api_keys', $config['swagger']['api_keys']);
-        $container->setParameter('api_platform.swagger_ui.extra_configuration', array_merge($config['openapi']['swagger_ui_extra_configuration'], $config['swagger']['swagger_ui_extra_configuration']));
+        if ($config['openapi']['swagger_ui_extra_configuration'] && $config['swagger']['swagger_ui_extra_configuration']) {
+            throw new RuntimeException('You can not set "swagger_ui_extra_configuration" twice - in "openapi" and "swagger" section.');
+        }
+        $container->setParameter('api_platform.swagger_ui.extra_configuration', $config['openapi']['swagger_ui_extra_configuration'] ?: $config['swagger']['swagger_ui_extra_configuration']);
 
         if (true === $config['openapi']['backward_compatibility_layer']) {
             $container->getDefinition('api_platform.swagger.normalizer.documentation')->addArgument($container->getDefinition('api_platform.openapi.normalizer'));
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -398,7 +398,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.enable_swagger_ui', $config['enable_swagger_ui']);
         $container->setParameter('api_platform.enable_re_doc', $config['enable_re_doc']);
         $container->setParameter('api_platform.swagger.api_keys', $config['swagger']['api_keys']);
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -398,7 +398,7 @@ private function registerSwaggerConfiguration(ContainerBuilder $container, array
         $container->setParameter('api_platform.enable_swagger_ui', $config['enable_swagger_ui']);
         $container->setParameter('api_platform.enable_re_doc', $config['enable_re_doc']);
         $container->setParameter('api_platform.swagger.api_keys', $config['swagger']['api_keys']);
-        $container->setParameter('api_platform.swagger_ui.extra_configuration', $config['openapi']['swagger_ui_extra_configuration'] ?? $config['swagger']['swagger_ui_extra_configuration']);
+        $container->setParameter('api_platform.swagger_ui.extra_configuration', array_merge($config['openapi']['swagger_ui_extra_configuration'], $config['swagger']['swagger_ui_extra_configuration']));
```
soyuka, line 401:
```suggestion
         $container->setParameter('api_platform.swagger_ui.extra_configuration', $config['openapi']['swagger_ui_extra_configuration'] ?: $config['swagger']['swagger_ui_extra_configuration']);
```

Should do it if they're always defined. I don't like the confusion array_merge brings here.

camohob: Ok then i suggest adding an exception
```
        if ($config['openapi']['swagger_ui_extra_configuration'] && $config['swagger']['swagger_ui_extra_configuration']) {
            throw new RuntimeException('You can not set "swagger_ui_extra_configuration" twice - in "openapi" and "swagger" section.');
        }
```
before `$container->setParameter(...);`

guilliamxavier: Maybe we should also ask @alanpoulain what was the reason for [duplicating](https://github.com/api-platform/core/pull/3731/commits/288d0b297d6e4e94a931e87b170109bfe070addc) the `swagger_ui_extra_configuration` config in both `swagger` and `openapi`?

alanpoulain: It's mainly for backward compatibility reasons, since Swagger was the previous name of OpenAPI.
IMO it should be in the `openapi` section, but it makes also sense to have it in the `swagger` section since there is already the `api_keys` parameter.

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -859,6 +859,27 @@ static function (string $path): string {
         $this->extension->load($config, $containerBuilderProphecy->reveal());
     }
 
+    public function testExtraConfigurationMegrging()
```
alanpoulain, line 862:
```suggestion
    public function testExtraConfigurationMerging()
```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
guilliamxavier: But the `api_keys` and `versions` configs are in `swagger` *only*, not duplicated in `openapi`... so doesn't that (duplication) create unnecessary confusion?

alanpoulain: Yes you're probably right. Should we just add it in `swagger` then?

soyuka: > But the api_keys and versions configs are in swagger only, not duplicated in openapi... so doesn't that (duplication) create unnecessary confusion?

But this should be the case imo

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -859,6 +859,27 @@ static function (string $path): string {
         $this->extension->load($config, $containerBuilderProphecy->reveal());
     }
 
+    public function testExtraConfigurationMerging()
```
alanpoulain, line 862:
```suggestion
    public function testSwaggerUiExtraConfigurationMerging()
```

alanpoulain (approved):

alanpoulain:
Thank you @camohob.

PR #4194, opened 2021-04-07
Upgrade to mercure-bundle:^0.3

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| License       | MIT
| Doc PR        | N/A

This PR is based on https://github.com/api-platform/core/pull/4142 plus:
- Fix tests
- Move deprecations to methods instead of constructor
- Introduce a `enable_async_update` option in the `mercure` attribute with `true` as a default value.
- Deprecate that `true` value in order to progressively drop the async update feature

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -28,6 +28,7 @@ use Doctrine\ODM\MongoDB\Event\OnFlushEventArgs as MongoDbOdmOnFlushEventArgs;
 use Doctrine\ORM\Event\OnFlushEventArgs as OrmOnFlushEventArgs;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Mercure\HubRegistry;
 use Symfony\Component\Mercure\Update;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Serializer\SerializerInterface;
@@ -51,11 +52,13 @@ final class PublishMercureUpdatesListener
         'type' => true,
         'retry' => true,
         'normalization_context' => true,
+        'hub' => true,
+        'enable_async_update' => true,
     ];
 
     private $iriConverter;
     private $serializer;
-    private $publisher;
+    private $registry;
     private $expressionLanguage;
     private $createdObjects;
     private $updatedObjects;
@@ -66,11 +69,12 @@ final class PublishMercureUpdatesListener
 
     /**
      * @param array<string, string[]|string> $formats
+     * @param HubRegistry|callable           $registry
      */
-    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, callable $publisher = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
+    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, $registry = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
     {
-        if (null === $messageBus && null === $publisher) {
-            throw new InvalidArgumentException('A message bus or a publisher must be provided.');
+        if (null === $messageBus && null === $registry) {
+            throw new InvalidArgumentException('A message bus or a HubRegistry instance must be provided.');
         }
 
         $this->resourceClassResolver = $resourceClassResolver;
@@ -79,7 +83,7 @@ final class PublishMercureUpdatesListener
         $this->serializer = $serializer;
         $this->formats = $formats;
         $this->messageBus = $messageBus;
-        $this->publisher = $publisher;
+        $this->registry = $registry;
         $this->expressionLanguage = $expressionLanguage ?? (class_exists(ExpressionLanguage::class) ? new ExpressionLanguage() : null);
         $this->graphQlSubscriptionManager = $graphQlSubscriptionManager;
         $this->graphQlMercureSubscriptionIriGenerator = $graphQlMercureSubscriptionIriGenerator;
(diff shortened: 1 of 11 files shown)


src/GraphQl/Subscription/MercureSubscriptionIriGeneratorInterface.php
```diff
@@ -24,5 +24,5 @@ interface MercureSubscriptionIriGeneratorInterface
 {
     public function generateTopicIri(string $subscriptionId): string;
 
-    public function generateMercureUrl(string $subscriptionId): string;
+    public function generateMercureUrl(string $subscriptionId, ?string $hub = null): string;
```
soyuka, line 27: bc allowed because interface is experimental, we could even enforce this argument?

src/Mercure/EventListener/AddLinkHeaderListener.php
```diff
@@ -43,26 +47,34 @@ public function __construct(ResourceMetadataFactoryInterface $resourceMetadataFa
...
-            false === $this->resourceMetadataFactory->create($resourceClass)->getAttribute('mercure', false)
+            false === ($mercure = $this->resourceMetadataFactory->create($resourceClass)->getAttribute('mercure', false))
         ) {
             return;
         }
 
-        if (null === $linkProvider = $request->attributes->get('_links')) {
-            $request->attributes->set('_links', new GenericLinkProvider([$link]));
+        if (!$this->discovery instanceof Discovery) {
+            @trigger_error(sprintf('Passing a string as the second argument to "%s::__construct()" is deprecated, pass a "%s" instance instead.', __CLASS__, Discovery::class), \E_USER_DEPRECATED);
```
soyuka, line 62: wdyt about beeing more specific on this message mentioning that we switched from a string to a Discovery class?

tests/Bridge/Doctrine/EventListener/PublishMercureUpdatesListenerTest.php
```diff
@@ -319,6 +434,52 @@ public function testNoPublisher(): void
...
+
+        $uowProphecy = $this->prophesize(UnitOfWork::class);
+        $uowProphecy->getScheduledEntityInsertions()->willReturn([$toInsert])->shouldBeCalled();
+
+        $emProphecy = $this->prophesize(EntityManagerInterface::class);
+        $emProphecy->getUnitOfWork()->willReturn($uowProphecy->reveal())->shouldBeCalled();
+        $eventArgs = new OnFlushEventArgs($emProphecy->reveal());
+
+        $listener->onFlush($eventArgs);
+        //$listener->postFlush();
```
soyuka, line 476:
```suggestion

```

src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -263,7 +265,8 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
 
         if (!empty($config['resource_class_directories'])) {
             $container->setParameter('api_platform.resource_class_directories', array_merge(
-                $config['resource_class_directories'], $container->getParameter('api_platform.resource_class_directories')
+                $config['resource_class_directories'],
```
alanpoulain, line 268: This kind of changes should only be done in `2.6` to prevent conflicts.

src/GraphQl/Subscription/MercureSubscriptionIriGeneratorInterface.php
mtarld: If `$hub` is `null`, the registry will retrieve the default hub.
Therefore IMO this argument must be optional anyway.

src/Mercure/EventListener/AddLinkHeaderListener.php
mtarld: Yep sounds good! any suggestions?

soyuka: Using a string for the Mercure hub in the constructor of __CLASS__ is deprecated since 2.6 in favor of an instance of Discovery::class.

mtarld: It's perfect! Thanks

dunglas:
Shouldn't we merge this as a bug fix. I know that it's technically a new feature but I fear that it will be a pain for users if they cannot use the latest version of MercureBundle for a while.

mtarld:
> Shouldn't we merge this as a bug fix. I know that it's technically a new feature but I fear that it will be a pain for users if they cannot use the latest version of MercureBundle for a while.

+1, bumping deps while still supporting older ones may generally be considered as a bux fix IMO

mtarld:
May I target 2.6 instead then?

dunglas:
Yes, please!

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -66,11 +69,12 @@ final class PublishMercureUpdatesListener
...
+    public function __construct(ResourceClassResolverInterface $resourceClassResolver, IriConverterInterface $iriConverter, ResourceMetadataFactoryInterface $resourceMetadataFactory, SerializerInterface $serializer, array $formats, MessageBusInterface $messageBus = null, $registry = null, ?GraphQlSubscriptionManagerInterface $graphQlSubscriptionManager = null, ?GraphQlMercureSubscriptionIriGeneratorInterface $graphQlMercureSubscriptionIriGenerator = null, ExpressionLanguage $expressionLanguage = null)
     {
-        if (null === $messageBus && null === $publisher) {
-            throw new InvalidArgumentException('A message bus or a publisher must be provided.');
+        if (null === $messageBus && null === $registry) {
+            throw new InvalidArgumentException('A message bus or a HubRegistry instance must be provided.');
```
dunglas, line 77:
```suggestion
            throw new InvalidArgumentException('A message bus or a hub registry must be provided.');
```

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -188,6 +192,15 @@ private function storeObjectToPublish($object, string $property): void
...
             }
+
+            if ('hub' === $key && !$this->registry instanceof HubRegistry) {
+                throw new InvalidArgumentException(sprintf('The option "hub" set in the "mercure" attribute of the "%s" resource is not supported. Please install symfony/mercure:^0.5 to enable it.', $resourceClass));
+            }
+        }
+
+        $options['enable_async_update'] = $options['enable_async_update'] ?? true;
+        if ($options['enable_async_update']) {
+            @trigger_error(sprintf('Not setting "enable_async_update" option set in the "mercure" attribute of the "%s" resource to "false" is deprecated.', $resourceClass), \E_USER_DEPRECATED);
```
dunglas, line 203: We should only deprecate this in main, we try to not introduce deprecations in patch release except for security issues.

dunglas (approved):

src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
mtarld: May we add a TODO then?

mtarld: And the same goes for other deprecations I  can guess

mtarld: I didn't add a todo, I'll create a PR aiming to add these deprecations instead

alanpoulain (approved):
LGTM

chalasr (approved):

soyuka:
thanks everyone

PR #4200, opened 2021-04-09
Convert internal exception in UuidDenormalizer

### LINKED ISSUE #4189: Error 500 instead of 400 for invalid UUID in POST body
**API Platform version(s) affected**: 2.6.3, 2.6.4
- Symfony 4.4
- `ramsey/uuid` 3.9.3

**Description**  
(see title) when using the `UuidInterface` type declaration for automatic denormalization

**How to reproduce**  
On PHP 7.4
```php
<?php

namespace App\Entity;

/** @ApiResource */
class Foo
{
    public int $id = 0;
    public ?UuidInterface $uuid = null;
    public ?\DateTimeInterface $datetime = null;
}
```

Request: `POST /foos` with body (Content-Type: application/ld+json):
```json
{"uuid": "foo"}
```

Expected: 400 Bad Request
Actual: 500 Internal Server Error
(message: "Invalid UUID string: foo")

**Possible Solution**  
In [`ApiPlatform\Core\Bridge\RamseyUuid\Serializer\UuidDenormalizer`](https://github.com/api-platform/core/blob/2.6/src/Bridge/RamseyUuid/Serializer/UuidDenormalizer.php), wrap `return Uuid::fromString($data);` with a try-catch (like in [`ApiPlatform\Core\Bridge\RamseyUuid\Identifier\Normalizer\UuidNormalizer`](https://github.com/api-platform/core/blob/2.6/src/Bridge/RamseyUuid/Identifier/Normalizer/UuidNormalizer.php)) and convert the `Ramsey\Uuid\Exception\InvalidUuidStringException` to an appropriate exception (a subclass of `Symfony\Component\Serializer\Exception\ExceptionInterface`)?

**Additional Context**  
For this body:
```json
{"datetime": "foo"}
```
I get a 400 Bad Request, as expected
(message: "DateTimeImmutable::__construct(): Failed to parse time string (foo) at position 0 (f): The timezone could not be found in the database")

guilliamxavier:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4189
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Bridge/RamseyUuid/Serializer/UuidDenormalizer.php
+++ b/src/Bridge/RamseyUuid/Serializer/UuidDenormalizer.php
@@ -13,15 +13,21 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Bridge\RamseyUuid\Serializer;
 
+use Ramsey\Uuid\Exception\InvalidUuidStringException;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 
 final class UuidDenormalizer implements DenormalizerInterface
 {
     public function denormalize($data, $type, $format = null, array $context = [])
     {
-        return Uuid::fromString($data);
+        try {
+            return Uuid::fromString($data);
+        } catch (InvalidUuidStringException $e) {
+            throw new NotNormalizableValueException($e->getMessage(), $e->getCode(), $e);
+        }
     }
 
     public function supportsDenormalization($data, $type, $format = null)
--- a/tests/Bridge/RamseyUuid/Serializer/UuidDenormalizerTest.php
+++ b/tests/Bridge/RamseyUuid/Serializer/UuidDenormalizerTest.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Bridge\RamseyUuid\Serializer;
+
+use ApiPlatform\Core\Bridge\RamseyUuid\Serializer\UuidDenormalizer;
+use PHPUnit\Framework\TestCase;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
+
+class UuidDenormalizerTest extends TestCase
+{
+    public function testDenormalizeUuid(): void
+    {
+        $uuid = Uuid::uuid4();
+        $normalizer = new UuidDenormalizer();
+        self::assertTrue($normalizer->supportsDenormalization($uuid->toString(), Uuid::class));
+        self::assertEquals($uuid, $normalizer->denormalize($uuid->toString(), Uuid::class));
+    }
+
+    public function testNoSupportDenormalizeUuid(): void
+    {
+        $uuid = 'notanuuid';
+        $normalizer = new UuidDenormalizer();
+        self::assertFalse($normalizer->supportsDenormalization($uuid, ''));
+    }
+
+    public function testFailDenormalizeUuid(): void
+    {
+        $this->expectException(NotNormalizableValueException::class);
+
+        $uuid = 'notanuuid';
+        $normalizer = new UuidDenormalizer();
+        $normalizer->denormalize($uuid, Uuid::class);
+    }
+
+    public function testDoNotSupportNotString(): void
+    {
+        $uuid = Uuid::uuid4();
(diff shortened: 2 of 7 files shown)


guilliamxavier:
For the exception, I hesitate between `ApiPlatform\Core\Exception\DeserializationException` and `Symfony\Component\Serializer\Exception\NotNormalizableValueException` 🤔

Also I guess this will need new test(s), but do you already agree on the principle?

alanpoulain:
I don't think `NotNormalizableValueException` would lead to a 400 status code, isn't it (but it probably should)?

guilliamxavier:
Yes it would, like all implementations of `Symfony\Component\Serializer\Exception\ExceptionInterface`. But I thought it may be better to use an ApiPlatform exception

alanpoulain:
Oh yes you are right. Maybe using `NotNormalizableValueException` would be better then to be like the denormalizers in Symfony?

src/Bridge/RamseyUuid/Serializer/UuidDenormalizer.php
```diff
@@ -21,7 +23,11 @@ final class UuidDenormalizer implements DenormalizerInterface
 {
     public function denormalize($data, $type, $format = null, array $context = [])
     {
-        return Uuid::fromString($data);
```
soyuka, line 24: Warning here, these converters will not implement DenormalizerInterface anymore in 3.0 and will have their own interface as the one from the serializer does not fit our needs.

soyuka:
Are you looking for https://github.com/api-platform/core/blob/5cb0508dd86c15fcbf8ef8b9c3012a2c445f4d6f/src/Exception/InvalidIdentifierException.php ?

src/Bridge/RamseyUuid/Serializer/UuidDenormalizer.php
guilliamxavier: Hmm... okay, but does that impact on what I should do here?

guilliamxavier:
@alanpoulain: OK I will use the Symfony one after all, thanks.

@soyuka: Thanks but I thought that `InvalidIdentifierException` (already thrown in ApiPlatform\\Core\\Bridge\\RamseyUuid\\<i>Identifier\\Normalizer</i>\\UuidNormalizer) was for e.g. `GET` `/foos/invalid-uuid`, while this (in ApiPlatform\\Core\\Bridge\\RamseyUuid\\<i>Serializer</i>\\UuidDenormalizer) is for e.g. `POST` `/foos` `{"uuid":"invalid-uuid"}` (similar to the difference between ApiPlatform specific DateTime<i>Identifier</i>Denormalizer and Symfony generic DateTimeNormalizer)?

src/Bridge/RamseyUuid/Serializer/UuidDenormalizer.php
soyuka: Oh I thought this was about identifiers conversion mb

soyuka:
My bad I thought this was related to identifier conversion not normalization, I'd be :+1: to transform this exception indeed. NotNormalizableValueException looks like the way to go here.

guilliamxavier:
"PHPUnit (PHP 7.1)" [crashed](https://github.com/api-platform/core/pull/4200/checks?check_run_id=2324810956), it had [passed](https://github.com/api-platform/core/runs/2324689186) before my last [unrelated force-push](https://github.com/api-platform/core/compare/3383345a1d16714db302f5094e33a64cd2d9f9bd..c6478606dd64e507a0470fe9b27041ad75592310)... maybe someone can retry it?

tests/Bridge/RamseyUuid/Serializer/UuidDenormalizerTest.php
```diff
@@ -0,0 +1,54 @@
...
+namespace ApiPlatform\Core\Tests\Bridge\RamseyUuid\Serializer;
+
+use ApiPlatform\Core\Bridge\RamseyUuid\Serializer\UuidDenormalizer;
+use PHPUnit\Framework\TestCase;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
+
+class UuidDenormalizerTest extends TestCase
+{
+    public function testDenormalizeUuid()
```
alanpoulain, line 23: Don't hesitate to use `: void` and `self::assert` for the PHPUnit tests :wink:

tests/Bridge/RamseyUuid/Serializer/UuidDenormalizerTest.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->assertFalse($normalizer->supportsDenormalization($uuid, ''));
+    }
+
+    public function testFailDenormalizeUuid()
+    {
+        $this->expectException(NotNormalizableValueException::class);
+
+        $uuid = 'notanuuid';
+        $normalizer = new UuidDenormalizer();
+        $this->assertTrue($normalizer->supportsDenormalization($uuid, Uuid::class));
```
alanpoulain, line 44: You can remove this line.

alanpoulain (approved):

alanpoulain:
Thank you @guilliamxavier.

PR #4201, opened 2021-04-09
Fix #4133

### LINKED ISSUE #4133: JsonSchema doesn't handle nullable properties that are objects
**API Platform version(s) affected**: 2.6.3

**Description**  
Let's assume we have this two models:

```php
/**
 * @ApiResource
 */
class Cart
{
  protected string $id = 'an-identifier';

  protected ?ShippingAddress $shippingAddress = null;
}
```

```php
class ShippingAddress
{
  protected string $street = 'street';
}
```

**Note**: Only `Cart` is registered as Api Platform resource, `ShippingAddress` is only an embedded object.

Currently, the generated json schema looks like this:

```JSON
 {
    "$schema": "http://json-schema.org/draft-07/schema#",
    "$ref": "#/definitions/Cart.jsonld",
    "definitions": {
        "Cart.jsonld": {
            "type": "object",
            "properties": {
                "@context": {
                    "readOnly": true,
                    "type": "string"
                },
                "@id": {
                    "readOnly": true,
                    "type": "string"
                },
                "@type": {
                    "readOnly": true,
                    "type": "string"
                },
                "id": {
                    "readOnly": true,
                    "type": "string",
                    "format": "uuid"
                },
                "shippingAddress": {
                    "nullable": true,
                    "anyOf": [
                        {
                            "$ref": "#/definitions/ShippingAddress.jsonld"
                        }
                    ]
                }
            }
        },
        "ShippingAddress.jsonld": {
            "type": "object",
            "properties": {
                "@context": {
                    "readOnly": true,
                    "type": "string"
                },
                "@id": {
                    "readOnly": true,
                    "type": "string"
                },
                "@type": {
                    "readOnly": true,
                    "type": "string"
                },
                "street": {
                    "type": "string"
                }
            }
        }
    }
}
```

The `Cart.shippingAddress` part is not a valid json schema, it must be:

```
                "shippingAddress": {
                    "type": ["object", "null"],
                    "anyOf": [
                        {
                            "$ref": "#/definitions/ShippingAddress.jsonld"
                        }
                    ]
                }
```

**How to reproduce**  
Create the two models above and run `php bin/console api:json-schema:generate 'Cart'` to get the Json Schema.

**Additional Context**  
The unit test for
https://github.com/api-platform/core/blob/4717bd5643d7761707229ed8705ae6824248f7e3/src/JsonSchema/TypeFactory.php#L52
https://github.com/api-platform/core/blob/4717bd5643d7761707229ed8705ae6824248f7e3/tests/JsonSchema/TypeFactoryTest.php#L32-L36
is twisted because it doesn't test by providing the value of `$readableLink`, which is `true` in our case.

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4133
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -165,6 +165,7 @@ final class TypeFactory implements TypeFactoryInterface
         if (\array_key_exists('$ref', $jsonSchema)) {
             return [
                 'nullable' => true,
+                'type' => array_merge(isset($jsonSchema['type']) ? [$jsonSchema['type']] : ['object'], ['null']),
                 'anyOf' => [$jsonSchema],
             ];
         }
--- a/tests/Fixtures/TestBundle/Model/Cart.php
+++ b/tests/Fixtures/TestBundle/Model/Cart.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Tests\Fixtures\TestBundle\Model;
+
+use ApiPlatform\Core\Annotation\ApiProperty;
+use ApiPlatform\Core\Annotation\ApiResource;
+
+/**
+ * @ApiResource
+ *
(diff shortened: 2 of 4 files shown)


guilliamxavier:
Please see https://github.com/api-platform/core/issues/4133#issuecomment-816653794 for why the situation is a bit complex

PR #4205, opened 2021-04-11
Add $context to DataPersisterInterface for decorator on DataPersister

rsereir:
| Q             | A
| ------------- | ---
| Branch?       | next
| Bug fix?      | yes 
| New feature?  |no 
| Deprecations? |no
| Tickets       | On DataPersister we can't map $context argument to decorator methods because the prototype of interface is not complete.
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/DataPersister/DataPersisterInterface.php
+++ b/src/DataPersister/DataPersisterInterface.php
@@ -23,17 +23,17 @@ interface DataPersisterInterface
     /**
      * Is the data supported by the persister?
      */
-    public function supports($data): bool;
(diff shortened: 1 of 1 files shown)


antograssiot:
That would be a BC break and that's already provided by the ContextAwareDataPersisterInterface

dunglas:
Thanks for contributing @rsereir. As explained by @antograssiot, this case is already covered. This attribute will be part of the `DataPersisterInterface` interface in API Platform 3.

PR #4217, opened 2021-04-14
Fix graphql fetching with Elasticsearch

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fixes https://github.com/api-platform/api-platform/issues/1809  
| License       | MIT
| Doc PR        | na]

Elasticsearch doesn't work for subcollection as there are no SubresourceDataProvider. This patch adds one. 

Runs a GraphQL query like this:

```
query {
  products {
    collection {
      id, attributes { collection { id }}
    }
  }
}
```

Where you have a sub-collection with Elasticsearch.

### DIFF
--- a/features/elasticsearch/graphql.feature
+++ b/features/elasticsearch/graphql.feature
@@ -1,0 +1,99 @@
+Feature: GraphQL query support
+
+  @elasticsearch
+  Scenario: Execute a GraphQL query on an ElasticSearch model with SubResources
+    When I send the following GraphQL request:
+    """
+    query {
+      users {
+        edges {
+          node {
+            id,
+            gender,
+            tweets {
+              edges {
+                node {
+                  id,
+                  message,
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON should be equal to:
+    """
+    {
+      "data": {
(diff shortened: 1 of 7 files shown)


features/elasticsearch/graphql.feature
```diff
@@ -0,0 +1,100 @@
...
+                node {
+                  id,
+                  message,
+                }
+              }
+            }
+          }
+        }
+      }
+    }
```
soyuka, line 24: why is the query so complex? Couldn't you use colleciton instead of edges + node ?

```
query {
  esProducts {
    collection {
      id, attributes { collection { id }}
    }
  }
}
```

vincentchalamon: The `collection` doesn't seem available on the User existing model. I prefered using an existing models instead of creating new ones

soyuka: but collection is needed to go through the new DataProvider I think, does the test fail without the new provider?

vincentchalamon: I ran this test on 2.6 branch and the test was failing. It runs successfully on your branch

vincentchalamon (approved):

eradionov:
Seems it still does not work

alanpoulain:
Sadly it has been reverted with the new system metadata. We need to make it work again.

Issue #4216, opened 2021-04-14
Filtering on Nested Properties doesn't works with mongo since 2.6


bettinz:
Hello, I've updated Api-Platform to version 2.6.4 and since 2.6.x filter on nested properties doesn't work with doctrine and mongodb.
I've a user entity with email property and events with user.email as filter.
Other filters work as expected:
```
https://localhost/admin/events?page=1&itemsPerPage=50&mustSort=false&multiSort=false&user.email=user%40email.com&type=2
```
user.email is ignored
type is working

Can you help me? Thanks.

simocheccoli:
same issue here:
```
namespace App\Document;

use ApiPlatform\Core\Annotation\ApiFilter;
use ApiPlatform\Core\Annotation\ApiResource;
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
use ApiPlatform\Core\Bridge\Doctrine\MongoDbOdm\Filter\DateFilter;
use ApiPlatform\Core\Bridge\Doctrine\MongoDbOdm\Filter\SearchFilter;
use App\Document\User;

/**
 * Class Event
 * @package App\Entity
 *
 * @ApiResource(
 *      attributes={
 *          "doctrine_mongodb"={
 *              "execute_options"={
 *                  "allowDiskUse"=true
 *              }
 *          },
 *          "order"={"dateTime": "DESC"},
 *          "pagination_client_items_per_page"=true
 *      },
 *      collectionOperations={
 *          "get" = {
 *              "openapi_context"={
 *                  "summary" = "Retrieve the list of all Events",
 *               },
 *          }
 *      },
 *      itemOperations={
 *          "get" = {
 *              "openapi_context"={
 *                  "summary" = "Retrieve the informations about a single Event",
 *              },
 *          }
 *      }
 * )
 *
 * @ApiFilter(DateFilter::class, properties={"dateTime"})
 * @ApiFilter(SearchFilter::class , properties={"client": "exact", "type": "exact", "code": "exact", "subCode": "exact", "title": "partial", "description": "partial" ,"user.email": "exact", "body": "partial"})
 *
 * @ODM\Document(collection="event")
 */
class Event
{
    const TYPE_INFO = 'info';
    const TYPE_ERROR = 'error';

    /**
     * @ODM\Id(name="_id", type="string")
     */
    private $id;

    /**
     * @ODM\Field
     */
    private $type;

    /**
     * @ODM\Field
     */
    private $code;

    /**
[...]

Issue #4214, opened 2021-04-14
Filter\Validator\Require fails on nested property


stefanides:
**API Platform version(s) affected**: 2.6.0+

**Description**  
Hi, after upgrading v2.5.7 -> v2.6.4 I am facing the strange behaviour with required filter parameter on nested property. `QueryParameterValidator.php` (line 65) then returns `FilterValidationException: Query parameter "x.y" is required` even it is present in query and can work. 

Once the filter is set to be required i.e. in a custom filter, the `Filter\Validator\Require` class fails to test `requestHasQueryParameter()` when the parameter name contains dot (used for nested properties). This looks to be introduced in https://github.com/api-platform/core/pull/1723, where new `Require` class compares `$queryParameters` array in form "x.y" with result of `parse_str()` which converts dots to "x_y" (see e.g. https://stackoverflow.com/questions/68651/get-php-to-stop-replacing-characters-in-get-or-post-arrays) - this cannot match, then an exception is thrown. 

**How to reproduce**  

1. 
- Set Entity with nested property filter
- Create custom filter class, add `getDescription()` method with `'required' => true` like here https://api-platform.com/docs/core/filters/#creating-custom-filters
- Call API either directy or from ApiDoc
- `Filter\Validator\Require` complains on missing parameter?!

2.
- Set custom filter class with `'required' => false`
- `Filter\Validator\Require` test passes and filter works(!) 

**Possible Solution**  
Some workaround for `parse_str()` like here https://stackoverflow.com/questions/22539633/parse-string-containing-dots-in-php ?

Or compare output of `parse_str()` with  `$request->query->all()` which also has underscores instead of dots, as it is in v2.5.7 (https://github.com/api-platform/core/blob/37b5edf50cc05299b12cb8982be3efa035d726ce/src/Filter/QueryParameterValidateListener.php)

Or get underscores also in `$queryParameters`?

**Additional Context**  
`Require` class https://github.com/api-platform/core/pull/1723/files#diff-900c533383b3b17876e6a5c81e525632c8a492358f36f352b0db974e4d604a09

Test simple code
```
<?php
parse_str("x.y", $blabla);
var_dump($blabla);
```
which outputs
```
array(1) {
  ["x_y"]=>
  string(0) ""
}
```

alanpoulain:
Hello, I think the better approach would be to keep the dots if possible, like it's done in the `RequestParser`.
Could you make a PR to fix it?

stefanides:
[...]

PR #4218, opened 2021-04-14
fix: do not log ValidationException

lyrixx:
Without this patch, the exception is not catched, and it processed by
the Symfony ErrorListener and a log message is processed. This is kind
of useless.

With this patch, the exeption is not logged anymore and is processed is
any other validation exception.

### DIFF
--- a/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
+++ b/src/Bridge/Symfony/Validator/EventListener/ValidationExceptionListener.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Core\Bridge\Symfony\Validator\EventListener;
 
 use ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException;
+use ApiPlatform\Core\Exception\FilterValidationException;
 use ApiPlatform\Core\Util\ErrorFormatGuesser;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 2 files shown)


alanpoulain:
Probably related to https://github.com/api-platform/core/issues/4216 and https://github.com/api-platform/core/issues/4214, don't you think?

alanpoulain:
Could you add a unit test too please?

lyrixx:
> Probably related to #4216 and #4214, don't you think?

I'm not sure. But I may be wrong :)

alanpoulain:
I mean: since the filter validation only creates a log message, people are thinking that nested filtering is broken (https://github.com/api-platform/core/issues/4216) because they don't see the validation error (and the filter doesn't pass validation because of https://github.com/api-platform/core/issues/4214).

lyrixx:
I added a unit test ✅

dunglas (approved):

soyuka:
isn't this a bug fix?

alanpoulain:
It could be a bugfix in `2.6` IMO.

lyrixx:
What should I do? Should I rebase? just tell me :)

alanpoulain:
Yes it would be nice if you could rebase it.

lyrixx:
Done ✅

alanpoulain:
Sorry it wasn't clear, I meant rebase agaisnt `2.6`. I've done it.

alanpoulain:
Thank you @lyrixx.

PR #4221, opened 2021-04-15
Fix issue in Required filter validator with dot notation

### LINKED ISSUE #4214: Filter\Validator\Require fails on nested property

jdeniau:
Fixes #4214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.6.5
 
+* Filter validation: Fix issue in Required filter validator with dot notation (#4221)
 * OpenAPI: Fix notice/warning for `response` without `content` in the `openapi_context` (#4210)
 * Serializer: Convert internal error to HTTP 400 in Ramsey uuid denormalization from invalid body string (#4200)
 
--- a/src/Filter/Validator/Required.php
+++ b/src/Filter/Validator/Required.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Core\Filter\Validator;
 
+use ApiPlatform\Core\Util\RequestParser;
+
 final class Required implements ValidatorInterface
 {
     /**
@@ -47,8 +49,7 @@ final class Required implements ValidatorInterface
      */
     private function requestHasQueryParameter(array $queryParameters, string $name): bool
     {
-        $matches = [];
-        parse_str($name, $matches);
+        $matches = RequestParser::parseRequestParams($name);
         if (!$matches) {
             return false;
         }
@@ -76,8 +77,7 @@ final class Required implements ValidatorInterface
      */
     private function requestGetQueryParameter(array $queryParameters, string $name)
     {
-        $matches = [];
-        parse_str($name, $matches);
+        $matches = RequestParser::parseRequestParams($name);
         if (empty($matches)) {
             return null;
         }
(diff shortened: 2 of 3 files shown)


alanpoulain:
Thank you! Don't you think it's a bugfix? It should target 2.6, isn't it?

jdeniau:
@alanpoulain you are totally right 👍

norkunas:
In Symfony there is `HeaderUtils::parseQuery`, maybe apip could use this in `RequestParser::parseRequestParams` ?

guilliamxavier:
@norkunas: Symfony's `HeaderUtils::parseQuery` (which was actually partly inspired from API Platform's `RequestParser::parseRequestParams`) isn't available in Symfony < 5.2 (https://github.com/symfony/symfony/pull/37272)

norkunas:
I know, but maybe it could fallback to it if available and later could be dropped from apip itself

alanpoulain:
Yes it's possible to create a wrapper (and it's a good idea!), like we do for the Doctrine inflector: https://github.com/api-platform/core/blob/main/src/Util/Inflector.php.

alanpoulain:
Anyway it can be done in another PR. Let's merge this one first. Could you add a changelog entry @jdeniau?

jdeniau:
@alanpoulain done.

The CI fails for a reason that seems unrelated to my PR though. (some cache issue apparently)

alanpoulain (approved):

alanpoulain:
Thank you @jdeniau.

TheKassaK:
Hi, do you know when/how can I have this fix ? 

I'm on apiplatform 2.6.3

PR #4230, opened 2021-04-19
Add tests for Symfony Uuid

### LINKED ISSUE #3880: Symfony Uuid identifiers causes binary query and 404
**API Platform version(s) affected**: 2.6.0-alpha.1 with Symfony 5.2 on PHP 7.4

**Description**  
Using Symfony Uid as identifier of a resource and executing a request on an item given an existing uuid, returns a 404 error caused by a query with binary data instead of string representation of the uuid

**How to reproduce**  
resource example:

```
<?php
namespace App\Entity;
use Symfony\Bridge\Doctrine\IdGenerator\UuidV4Generator;
use Symfony\Component\Uid\Uuid;

/**
 * @ORM\Entity()
 * @ORM\Table(name="invoices")
 * @ApiResource(
 *     normalizationContext={"groups"={"invoices"}},
 *     collectionOperations={"get"},
 *     itemOperations={"get"}
 * )
 */
class Invoice
{
    /**
     * @ORM\Id
     * @ORM\Column(type="uuid", unique=true)
     * @ORM\CustomIdGenerator(class=UuidV4Generator::class)
     * @Groups({"invoices"})
     */
    private Uuid $id;

    /**
     * @ORM\Column(name="number", type="string")
     * @Groups({"invoices"})
     */
    private string $number;

    public function getId(): Uuid
    {
        return $this->id;
    }

    public function getNumber(): string
    {
        return $this->number;
    }
}
```

request example:
`http://localhost/invoices/b282f366-bcae-44cf-b09c-d0569b9cdf9d`

query produced:
```
SELECT i0_.id AS id_0, i0_.number AS number_1 FROM invoices i0_ WHERE i0_.id = 0xB282F366BCAE44CFB09CD0569B9CDF9D;
```

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes/no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #3880
| License       | MIT
| Doc PR        | N/A


features/main/uuid.feature
```diff
@@ -194,3 +194,12 @@ Feature: Using uuid identifier on resource
     Then the response status code should be 400
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+
+  @!mongodb
```
soyuka, line 198: should we also test this with mongo ?

vincentchalamon: No need to test it with mongo as symfony/uid uses doctrine/dbal which is only useful with doctrine/orm. Testing it with doctrine/odm will be irrelevant

.github/workflows/ci.yml
```diff
@@ -384,7 +388,7 @@ jobs:
       - name: Clear test app cache
         run: tests/Fixtures/app/console cache:clear --ansi
       - name: Run Behat tests
-        run: vendor/bin/behat --out=std --format=progress --profile=default --no-interaction
+        run: vendor/bin/behat --out=std --format=progress --profile=default --no-interaction --tags=!@symfony/uid
```
guilliamxavier, line 391:
```suggestion
        run: vendor/bin/behat --out=std --format=progress --profile=default --no-interaction --tags='~@symfony/uid'
```

tests/Fixtures/app/AppKernel.php
```diff
@@ -110,6 +111,10 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
 
         $loader->load(__DIR__."/config/config_{$this->getEnvironment()}.yml");
 
+        if (\PHP_VERSION_ID >= 72000 && class_exists(Uuid::class)) {
```
guilliamxavier, line 114:
```suggestion
        if (\PHP_VERSION_ID >= 70200 && class_exists(Uuid::class)) {
```
(not sure why you need it though)

soyuka:
thanks @vincentchalamon !

PR #4234, opened 2021-04-20
Implement HttpClientInterface::withOption(array $options)

Leprechaunz:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| License       | MIT


alanpoulain:
Could you explain why it would be needed / useful?

alanpoulain:
It's in the trait: https://github.com/symfony/symfony/blob/8361713367e2735fd9b63cd2b6661411a0339811/src/Symfony/Component/HttpClient/HttpClientTrait.php#L31-L37
Solving the deprecation should probably be done differently.

guilliamxavier:
(part of) the problem is that symfony/contracts 2.4 has been released while symfony/http-client 5.3 not yet (only beta) 😕

Leprechaunz:
@alanpoulain Good point, I think it's better to use trait when symfony/http-client 5.3 (as @guilliamxavier mentioned)

Closing this PR.

PR #4251, opened 2021-04-26
[GraphQL] Fix `FieldsBuilder` not always fully unwrapping a type before deciding if a resolver is needed.

DavidBennettUK:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes
| New feature?  | no
| Deprecations? |no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This fixes `FieldsBuilder` not using the recursive version of `getWrappedType`. Without this, if the type is wrapped in both `NonNull` and then `ListOfType`, then `ListOfType` is returned instead of the actual type of the list items - which I think was the original intention of this code, as it's trying to decide if the base type needs a special resolver. For example, a non-null list of non-null strings (`[String!]!`) will now correctly return `String` as the wrapped type with this fix. This means that simple arrays such as this, which are already deserialized by the Symfony deserializer and therefore do not need to use a special resolver, can now work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Serializer: Convert internal error to HTTP 400 in Ramsey uuid denormalization from invalid body string (#4200)
 * Symfony: Add tests with Symfony Uuid (#4230)
 * OpenAPI: Allow to set extensionProperties with YAML schema definition (#4228)
+* GraphQL: Fix `FieldsBuilder` not fully unwrapping nested types before deciding if a resolver is needed (#4251)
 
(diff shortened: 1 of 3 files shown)


alanpoulain:
Hello David, thanks for the PR.
Could you:
- target `2.6` since it's a bug fix,
- add a unit test?

DavidBennettUK:
@alanpoulain Thanks, all done 👍

alanpoulain (approved):

alanpoulain:
Thank you @DavidBennettUK.

PR #4253, opened 2021-04-27
test(mongodb): use MatchStage instead of Match (conflicts with php8)

soyuka:
Fixes a conflict with php8 on Match.


composer.json
```diff
@@ -33,7 +33,7 @@
         "doctrine/common": "^2.11 || ^3.0",
         "doctrine/data-fixtures": "^1.2.2",
         "doctrine/doctrine-bundle": "^1.12 || ^2.0",
-        "doctrine/mongodb-odm": "^2.1",
+        "doctrine/mongodb-odm": "^2.2",
```
alanpoulain, line 36: Shouldn't we change the conflict too?

PR #4258, opened 2021-04-29
Fixed the normalization of nested non-resources in AbstractItemNormalizer

webmozart:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Currently, the `createChildContext()` method isn't used when normalizing embedded objects that are not resources. Hence, for example, the `PropertiesFilter` won't work for these objects.

This PR fixes that by invoking `createChildContext()` also for nested non-resource objects.

### TODO

- [ ] Add test case


alanpoulain:
Hello. Thank you for the PR. It seems legit. Waiting for tests.

norkunas:
Would be nice to have fixed this as it's currently not possible to load only subset of fields..

webmozart:
Not sure yet when I will have time to write the test case. Does anyone want to take over?

norkunas:
I could take over but would need some guidance with behat as I'm not familiar with it :/

PR #4260, opened 2021-04-29
fix: avoid default order by duplication

### LINKED ISSUE #2690: How to override default ordering with an order filter
Hi everyone!

I've added default ordering to resource attributes (yaml). So far so good.
However, the default order is still applied when querying the collection with an explicit order filter (query param).

The query builder used by data provider orders by default and by order filter.

**Is there a way to ditch default order when order filter has been activated?**

**Also is it possible to define default order at operation level?**

felipyamorim:
| Q             | A
| ------------- | ---
| Branch?       | 2.6 
| Bug fix?      | yes 
| New feature?  | no 
| Deprecations? | no
| Tickets       | Fix #2690 
| License       | MIT
| Doc PR        |  

Avoid duplicate fields in "order by" when the field already exists in the filter order.


soyuka:
Not sure to get the fix, can you add a test?

felipyamorim:
Yes, i will try to create the tests.

PR #4263, opened 2021-04-30
Test patching 1:n relations

### LINKED ISSUE #1575: Identifiers url-encoded twice

mbrodala:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Bug fix?      | yes 
| New feature?  | no 
| Deprecations? | no
| Tickets       | Fix #1575 
| License       | MIT

### DIFF
--- a/features/main/patch.feature
+++ b/features/main/patch.feature
@@ -30,7 +30,7 @@ Feature: Sending PATCH requets
     Then the JSON node "name" should not exist
 
   @createSchema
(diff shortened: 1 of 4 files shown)


soyuka:
What's to be done here?

mbrodala:
The tests need to be finalized, especially wrt ODM integration. And then we need a fix for the failing tests of course. ;-)

Issue #4275, opened 2021-05-10
Custom exception handling has inconsistent behaviour between dev/test and prod environment


odoucet:
**API Platform version(s) affected**: 2.6.4 and lower (at least)

**Description**  
If you register a custom Exception in config with `api_platform.exception_to_status` and a status code >= 500 and < 600, trait `ApiPlatform\Core\Problem\Serializer\ErrorNormalizerTrait` has a very special handling that may prevent code from working in production environment (and it works on dev/test mode).

Problem is in getErrorMessage : 
```
// ApiPlatform\Core\Problem\Serializer/ErrorNormalizerTrait
private function getErrorMessage($object, array $context, bool $debug = false): string
    {
        $message = $object->getMessage();

        if ($debug) {
            return $message;
        }

        if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
            $statusCode = $context['statusCode'] ?? $object->getStatusCode();
            if ($statusCode >= 500 && $statusCode < 600) {
                $message = Response::$statusTexts[$statusCode];
            }
        }

        return $message;
    }
```
There is no check `Response::$statusTexts[$statusCode]` exists, and in this case, another error (TypeError) is raised and prevent code from working. 
I also do not understand why there is a special shortcut in debug mode.

**How to reproduce**  
in api_platform.yml : 
```
api_platform:
    exception_to_status:
        App\Exception\CustomException: 509
```
Note that I used a 509 HTTP code, that is not an official code.

File src\Exception\CustomException.php : 
```
<?php
namespace App\Exception;
class CustomException extends \Exception
{
}
```

**Possible Solution**  
There is two possible fixes and I need some feedback before creating a pull request : 
1) First, explain to me why there is a shortcut in debug mode. Why would we want a different message between envs ? This is really confusing.
2) then, we should check if `Response::$statusTexts[$statusCode]` exists and return a generic error message if not.

alanpoulain:
Hello Olivier.
For your question, the answer is really simple: it's because we don't want to show the message in production to the users, because it can leak important information.
Yes, it could be a good idea to do this. Do you want to work on this PR?

alanpoulain:
That's right. It's because 500+ errors are often unwanted errors whereas for instance 400+ errors (validation, ...) are thought to be shown to the end user.

odoucet:
I've open both MR.
A test would be great on #4277 but I really don't know how to handle this in APIP test framework.

PR #4277, opened 2021-05-10
fix: handle unofficial http status code with custom exceptions

odoucet:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       |  https://github.com/api-platform/core/issues/4275
| License       | MIT

When using custom Exception class and custom HTTP status code, a check was missing if the http status code really exists or is an unofficial ones.
This worked in devel mode because the verbosity mode does not replace the message but failed in production.

### DIFF
--- a/src/Problem/Serializer/ErrorNormalizerTrait.php
+++ b/src/Problem/Serializer/ErrorNormalizerTrait.php
@@ -31,7 +31,7 @@ trait ErrorNormalizerTrait
         if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
             $statusCode = $context['statusCode'] ?? $object->getStatusCode();
             if ($statusCode >= 500 && $statusCode < 600) {
-                $message = Response::$statusTexts[$statusCode];
+                $message = Response::$statusTexts[$statusCode] ?? Response::$statusTexts[Response::HTTP_INTERNAL_SERVER_ERROR];
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Problem/Serializer/ErrorNormalizerTrait.php
```diff
@@ -31,7 +31,12 @@ private function getErrorMessage($object, array $context, bool $debug = false):
         if ($object instanceof FlattenException || $object instanceof LegacyFlattenException) {
             $statusCode = $context['statusCode'] ?? $object->getStatusCode();
             if ($statusCode >= 500 && $statusCode < 600) {
-                $message = Response::$statusTexts[$statusCode];
```
alanpoulain, line 34: Why not:
```php
Response::$statusTexts[$statusCode] ?? Response::$statusTexts[500];
```

odoucet:
Code updated to be much simpler

soyuka:
needs a regression test, as it's a fix I think that it can target 2.6

odoucet:
I agree but I'm totally lost in the APIP testing framework.
Regression test is easy : 
1) emulate this block in api_platform.yml : 
```
api_platform:
    exception_to_status:
        App\Exception\CustomException: 509
```
2) add empty App\Exception\CustomException
3) throw this exception with debug=false

soyuka:
Add it there: https://github.com/api-platform/core/blob/main/tests/Fixtures/app/config/config_common.yml#L55 
Why the debug to false?

odoucet:
When debug is true, the original message is sent regardless of http status code.
To reproduce the issue, debug must be false.

alanpoulain:
We cannot set the debug flag to `false` in the Behat tests.
IMO a unit test is sufficient for this fix.

soyuka:
thanks @odoucet

PR #4282, opened 2021-05-18
fix(test): add withOptions to our HttpClient implementation

soyuka:


src/Bridge/Elasticsearch/DataProvider/Extension/AbstractFilterExtension.php
```diff
@@ -51,7 +51,7 @@ public function applyToCollection(array $requestBody, string $resourceClass, ?st
 
         foreach ($resourceFilters as $filterId) {
             if ($this->filterLocator->has($filterId) && is_a($filter = $this->filterLocator->get($filterId), $this->getFilterInterface(), true)) {
```
guilliamxavier, line 53: For PHPStan's error about `class-string<object>|object`, wouldn't this be better?

```suggestion
            if ($this->filterLocator->has($filterId) && is_a($filter = $this->filterLocator->get($filterId), $this->getFilterInterface())) {
```

(and revert the ` //@phpstan-ignore-line` below)

soyuka: :+1: thanks

Issue #4196, opened 2021-04-08
Filters do not filter on GraphQL requests


nspyke:
**API Platform version(s) affected**: 2.6.3

**Description**  
When adding a search or order filter to my GraphQL API Platform configuration, the query result returns all records, ignoring the parameters supplied in the GraphQL query.
I've also tried the exact filter. That doesn't work either, but this example does not show that filter in use.
These filters do however work as expected on a REST request.

**How to reproduce**  
My configuration;
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<resources xmlns="https://api-platform.com/schema/metadata"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="https://api-platform.com/schema/metadata
           https://api-platform.com/schema/metadata/metadata-2.0.xsd">
    <resource class="Core\Entity\Checklist\ChecklistItem" description="Checklist Item">
        <itemOperations>
            <!-- removed for brevity -->
        </itemOperations>
        <collectionOperations>
        <!-- removed for brevity -->
        </collectionOperations>
        <graphql>
            <operation name="item_query"/>
            <operation name="collection_query">
                <attribute name="filters">
                    <attribute>api.checklist_item.filter.search</attribute>
                    <attribute>api.checklist_item.filter.order</attribute>
                </attribute>
            </operation>
            <operation name="create"/>
            <operation name="update"/>
            <operation name="delete"/>
        </graphql>
    </resource>
</resources>
```
```xml
<?xml version="1.0" ?>
<container xmlns="http://symfony.com/schema/dic/services"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
    <services>
        <service id="api.checklist_item.filter.search" parent="api_platform.doctrine.orm.search_filter">
            <argument key="$properties" type="collection">
                <argument key="position">exact</argument>
            </argument>
            <tag name="api_platform.filter"/>
        </service>
        <service id="api.checklist_item.filter.order" parent="api_platform.doctrine.orm.order_filter">
            <argument key="$properties" type="collection">
                <argument key="position">ASC</argument>
            </argument>
            <tag name="api_platform.filter"/>
        </service>
    </services>
</container>
```
```graphql
query checklistItems {
  checklistItems(position: 5, order: [{position: "ASC"}]) {
    edges {
[...]

Issue #2308, opened 2018-11-03
GraphQL : Unable to filter by id or get a single row


alchy58:
Hi,

I'm trying GraphQL with success to replace REST calls inside a react-native app.
It seem's to reduce loading time a lot.

I have a critical problem
With REST, I can request a user by id or filter the collection by id. Not with GraphQL. It seems to be two differents problems.
Maybe it's somehow related to https://github.com/api-platform/core/issues/2190. In that case, is this possible to backport the patch on the 2.2 branch ?

First problem : unable to get a single row by id
Using GraphQL, I obtain 400 HTTP response with incorrect route error  ~~I'm unable to get a stacktrace, nothing in var/log neither than nginx nor php logs. How can I find the stacktrace ?~~

**REST query**
```
curl -X GET "https://my.endpoint.com/users/1" -H  "accept: application/json"
```

**Response**
```
{
  "id": 1,
}
```
 

**GraphQL query**
```
{
  user(id:1) {
    id
  }
}
```
**Response**
```
{
  "errors": [
    {
      "debugMessage": "No route matches \"1\".",
      "message": "Internal server error",
      "category": "internal",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "user"
      ]
    }
  ],
  "data": {
    "user": null
  }
}
```

The second problem is on the filter side, it's functionnal on REST but not with GraphQL : I'm getting all the collection

**REST query**
```
curl -X GET "https://my.endpoint.com/users?id=1" -H  "accept: application/json"
```

**Response**
```
[
  {
    "id": 1,
  }
]
```

**GraphQL query**
```
{
  users(id:1) {
    edges {
      node {
        id
      }
    }
  }
}
```
**Response**
```
{
  "data": {
    "users": {
      "edges": [
        {
          "node": {
            "id": "/users/1"
          }
        },
        {
          "node": {
            "id": "/users/2"
          }
        },
        {
          "node": {
            "id": "/users/3"
          }
        },
        etc.
}
```

nclsHart:
About your first problem, have you try to use IRI instead of id in your graphql query?

For example: 

```graphql
{
  user(id: "/users/1") {
    id
  }
}
```

alchy58:
As you mentionned it, we need to use the IRI instead of the id right

Issue #2190, opened 2018-09-05
Graphql: can't filter entity collection by ids


Siregacvek:
Hello!

Looks like I found a bug. If I add filter to the entity this way:
```
/**
 * @ApiResource()
 * @ApiFilter(SearchFilter::class, properties={"id": "exact"})
 */
class Entity
```
Then I receive the following schema:
`Entity(first: Int, after: String, id: Int, id_list: [Int])`

When I apply filter by id, it's ignoring and returns me the full collection.
Also I believe that the id must string like `/entities/1` instead of integer.

Or I'm doing something wrong?

ArnoudThibaut:
Hello,

This behavior is quite strange but it's caused by the [normalizeValues function](https://github.com/api-platform/core/blob/b15030beb49bd841e2c788e7d7a818c388c199aa/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php#L370) of the SearchFilter.
With REST operations the values passed to the filter aren't typed and are all strings so there is no problem if you pass to this filter values that are in fact numerical string values.
But with graphQl the values are typed.

For me you have two solutions that depend on your needs :
* First is to use the NumericFilter but this filter doesn't allow to pass multiple values so you won't have the `id_list` argument. (I'm working on it)
* Second is to extends the built-in search filter and override the `normalizeValues` function in order to not discard values that aren't strings.

teohhanhui:
Please do not use the `NumericFilter` for ids

teohhanhui:
I think this type conversion should be handled at the GraphQL layer. As far as the filters are concerned, all values are strings.

teohhanhui:
I think the way to fix this is to introduce the `IriFilter` and deprecate using the `SearchFilter` for IRIs / ids.

Related: https://github.com/api-platform/core/issues/319

soyuka:
LMK if the issue remains after #2277

PR #4290, opened 2021-05-22
Fixes filter when filtering with an int

jmontoyaa:
| Q             | A
| ------------- | ---
| Branch       | 2.6
| Tickets       | #1633  #4196 #2190 #2308
| License       | MIT

The error is reported here: #4196 When filtering with an int a notice appear and the filter doesn't work.

`app.NOTICE: Invalid filter ignored {"exception":"[object] (ApiPlatform\\Core\\Exception\\InvalidArgumentException(code: 0): At least one value is required, multiple values should be in \"status[]=firstvalue&status[]=secondvalue\" format at vendor/api-platform/core/src/Bridge/Doctrine/Common/Filter/SearchFilterTrait.php:146)"} []`

The example in #1633 works because the filter is a string (name) but the example here (id) #2190 fails because it's an int.


soyuka:
Can you add a non-regression test? Also as this is a bug fix please target the 2.6 branch.

jmontoyaa:
I added a new PR here #4295 with some tests, closing this one.

PR #4292, opened 2021-05-25
[ApiLoader] Support _format resolving

ahmed-bhs:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no, yes
| New feature?  |no, yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | #There's no ticket
| License       | MIT

Hello there,

Currently, it is not possible to resolve the  default value for the `_format` parameter under a custom operation definition.
This is due to the `ApiLoader` which forces this value to `nul` no matter what value we set in the configuration.
This fix makes it possible to take into account the definition of the `defaults._format` value and have the right value under the request route params.

**Use case:**
Imagine that we have a resource that which is used to export both the data in csv and in xlsx.

resources.yaml:

```
resources:
    App\Entity\Activity\Contract:
        collectionOperations:
            export:
                method: 'GET'
                path: /contracts/export.{_format}
                defaults:
                    _format: "xlsx"
               requirements:
                   _format: "xlsx|csv"
                controller: App\Controller\Api\ExportContractController

```
ExportContractController:

```
final class ExportContractController extends AbstractController
{
    public function __invoke($_format, $data)
    {
    // before the fix, $_format return null
    dd($_format);
    // after the fix, $_format return 'xlsx'
    
    }
}
```

### DIFF
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -128,7 +128,7 @@ final class ApiLoader extends Loader
                     $operation['path'],
                     [
                         '_controller' => $controller,
-                        '_format' => null,
+                        '_format' => !empty($operation['defaults']['_format']) ? $operation['defaults']['_format'] : null,
                         '_stateless' => $operation['stateless'] ?? $resourceMetadata->getAttribute('stateless'),
                         '_api_resource_class' => $operation['resource_class'],
                         '_api_identifiers' => $operation['identifiers'],
@@ -241,7 +241,7 @@ final class ApiLoader extends Loader
             $path,
             [
                 '_controller' => $controller,
-                '_format' => null,
+                '_format' => !empty($operation['defaults']['_format']) ? $operation['defaults']['_format'] : null,
                 '_stateless' => $operation['stateless'],
                 '_api_resource_class' => $resourceClass,
                 '_api_identifiers' => $operation['identifiers'],
--- a/src/Bridge/Symfony/Routing/ApiLoader.php
+++ b/src/Bridge/Symfony/Routing/ApiLoader.php
@@ -128,7 +128,7 @@ final class ApiLoader extends Loader
                     $operation['path'],
                     [
                         '_controller' => $controller,
-                        '_format' => !empty($operation['defaults']['_format']) ? $operation['defaults']['_format'] : null,
+                        '_format' => $operation['defaults']['_format'] ?? null,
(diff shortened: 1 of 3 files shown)


src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -128,7 +128,7 @@ public function load($data, $type = null): RouteCollection
                     $operation['path'],
                     [
                         '_controller' => $controller,
-                        '_format' => null,
+                        '_format' => !empty($operation['defaults']['_format']) ? $operation['defaults']['_format'] : null,
```
alanpoulain, line 131:
```suggestion
                        '_format' => $operation['defaults']['_format'] ?? null,
```

src/Bridge/Symfony/Routing/ApiLoader.php
```diff
@@ -241,7 +241,7 @@ private function addRoute(RouteCollection $routeCollection, string $resourceClas
             $path,
             [
                 '_controller' => $controller,
-                '_format' => null,
+                '_format' => !empty($operation['defaults']['_format']) ? $operation['defaults']['_format'] : null,
```
alanpoulain, line 244:
```suggestion
                '_format' => $operation['defaults']['_format'] ?? null,
```

tests/Bridge/Symfony/Routing/ApiLoaderTest.php
```diff
@@ -318,7 +318,7 @@ private function getRoute(string $path, string $controller, string $resourceClas
             $path,
             [
                 '_controller' => $controller,
-                '_format' => null,
+                '_format' => !empty($extraDefaults['_format']) ? $extraDefaults['_format'] : null,
```
alanpoulain, line 321:
```suggestion
                '_format' => $extraDefaults['_format'] ?? null,
```

tests/Bridge/Symfony/Routing/ApiLoaderTest.php
```diff
@@ -339,7 +339,7 @@ private function getSubresourceRoute(string $path, string $controller, string $r
...
```
alanpoulain, line 342:
```suggestion
                '_format' => $extraDefaults['_format'] ?? null,
```

alanpoulain:
Seems alright to me. Could you add a CHANGELOG entry and if possible a PR for the documentation?

soyuka (approved):

soyuka:
thanks @ahmed-bhs !

PR #4294, opened 2021-05-26
chore(phpstan): doctrine FieldMapping false-positive

soyuka:


guilliamxavier:
Wouldn't it be better to fix the phpDoc upstream (and ignore it here for the time being)?

soyuka:
> Wouldn't it be better to fix the phpDoc upstream (and ignore it here for the time being)?

In fact they did fix it upstream which lead to phpstan saying that this will always be true as the keys are now defined in their doc :).

guilliamxavier:
> In fact they did fix it upstream


Are you referring to [this](https://github.com/doctrine/orm/commit/dc37c2cd2fd22fe8b57ddb1743aa629db98b76e3#diff-732e324167dd49e48b221477c3e0f6d7934f3eb5ec0970dbf1c06f6c7df15398)? It added `options` on `public $fieldMappings` but not `public function getFieldMapping($fieldName)`.

> which led to phpstan saying that this will always be true as the keys are now defined in their doc :).

https://github.com/api-platform/core/runs/2674090639 looks like the opposite: "Call to function array_key_exists() with 'options' and array(list of keys that does *not* include 'options') will always evaluate to *false*."

soyuka:
Indeed I only saw that they added it on `fieldMappings` I missed the function. Did they miss `options`? Anyways, I think that the code here is valid and works just as before, with the benefits of not worrying about their types no?

guilliamxavier:
> Indeed I only saw that they added it on `fieldMappings` I missed the function. Did they miss `options`?

Maybe, and also `columnDefinition`, `declaredField`, `id`, `quoted`, `requireSQLConversion`, `unique` 😕

> Anyways, I think that the code here is valid and works just as before, with the benefits of not worrying about their types no?

Well looking at just the modified lines I would say that there's a potential behavior change if `$fieldMapping['options']['default'] === null`, but I guess that in this case, also `$propertyMetadata->getDefault() === null` anyway?

soyuka:
> Well looking at just the modified lines I would say that there's a potential behavior change if $fieldMapping['options']['default'] === null, but I guess that in this case, also $propertyMetadata->getDefault() === null anyway?

Exactly, I see no harm and we cut the dependency we have with their typings, as you said it's missing many keys ^^.

soyuka:
The mongodb issue has been tracked to https://github.com/doctrine/annotations/pull/412 , can't do much and I don't want to add a conflict on doctrine/annotations yet

PR #4295, opened 2021-05-26
fix: allow search filter to use an int for its value

jmontoyaa:
| Q             | A
| ------------- | ---
| Branch       | 2.6
| Tickets       | #1633  #4196 #2190 #2308
| License       | MIT

Same fix as #4290 but with tests and  in branch 2.6


src/Bridge/Doctrine/Common/Filter/SearchFilterTrait.php
```diff
@@ -136,7 +136,7 @@ protected function getIdFromValue(string $value)
     protected function normalizeValues(array $values, string $property): ?array
     {
         foreach ($values as $key => $value) {
-            if (!\is_int($key) || !\is_string($value)) {
+            if (!\is_int($key) || \is_array($value)) {
```
antograssiot, line 139: What value will not fit the first part of the IF then ? I mean an array will already be validated by the first part of the OR condition isn't it ?

jmontoyaa: First part of the IF it's the $key, second part it's about the $value. If $value is an INT then a notice will appear. I wrote a test about that here: 

https://github.com/api-platform/core/pull/4295/commits/d33159fc1af1c9ac3165a811ffa30c8f8d7cdf7c

features/graphql/collection.feature
```diff
@@ -909,3 +909,65 @@ Feature: GraphQL collection support
     Then the response status code should be 200
     And the response should be in JSON
     And the JSON node "data.fooDummies.collection" should have 1 element
+
+  @createSchema
+  Scenario: Retrieve a collection with a nested collection filtering by age through a GraphQL query
```
alanpoulain, line 914: Wouldn't it be better in `filters.feature` instead?

features/graphql/collection.feature
```diff
@@ -909,3 +909,65 @@ Feature: GraphQL collection support
...
+                  age
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    """
+
```
alanpoulain, line 939: Line to remove.

features/graphql/collection.feature
```diff
@@ -909,3 +909,65 @@ Feature: GraphQL collection support
...
+    Then the JSON node "data.dummies.totalCount" should be equal to 1
+    And the JSON node "data.dummies.edges[0].node.relatedDummies.totalCount" should be equal to 1
+    Then the JSON node "data.dummies.edges[0].node.relatedDummies.edges[0].node.age" should be equal to "31"
```
alanpoulain, line 942:
```suggestion
    And the JSON node "data.dummies.edges[0].node.relatedDummies.edges[0].node.age" should be equal to "31"
```

alanpoulain:
Thank you for the Behat tests! I think it would be nice to have unit ones too. I will handle them.

jmontoyaa:
> Thank you for the Behat tests! I think it would be nice to have unit ones too. I will handle them.

Thanks! I just added fixed your suggestions.

jmontoyaa:
I just fix for the behat test (subresource.feature) that was failing because the age filter was added.

alanpoulain (approved):

alanpoulain:
Thank you @jmontoyaa.

PR #4301, opened 2021-06-01
fix(serializer): collection denormalization from object linked to doctrine entities

nicodmf:
| Q             | A
| ------------- | ---
| Branch?       | current
| Type       | Bug fix
| Tickets       |  no
| License       | MIT

As it, denormalization doesn't find the children class in this case : 

```php
// Class to denormalize, not an entity, just an input class from a operation with messenger=input
class X {
  // id, other props not mentionned here ...
  /** @var Entity[] */
  public $collection;
}

// Object receive in json-ld
['collection'=>[{"@id"=>"/entities/5"}]]
// Object receive in json
['collection'=>[{"id"=>5}]]

```

If $entity has an id like in the example, object is denormalized normally with json ld (```"@id": "iri"```). With json format, the resource_class is absent from context and the object is denormalized without database populate and losing the id value. This fix come from parent class https://github.com/api-platform/core/blob/2.6/src/Serializer/AbstractItemNormalizer.php#L187 which will be call after the populate.

### DIFF
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -51,6 +51,12 @@ class ItemNormalizer extends AbstractItemNormalizer
      */
     public function denormalize($data, $class, $format = null, array $context = [])
     {
+        if (null === $objectToPopulate = $this->extractObjectToPopulate($class, $context, static::OBJECT_TO_POPULATE)) {
+            $normalizedData = $this->prepareForDenormalization($data);
+            $class = $this->getClassDiscriminatorResolvedClass($normalizedData, $class);
+        }
+        $resourceClass = $this->resourceClassResolver->getResourceClass($objectToPopulate, $class);
+        $context['resource_class'] = $resourceClass;
         // Avoid issues with proxies if we populated the object
         if (isset($data['id']) && !isset($context[self::OBJECT_TO_POPULATE])) {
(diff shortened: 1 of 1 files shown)


norkunas:
Why don't you send IRI's for json format?

nicodmf:
to be consistent with response which give not an iri but an numeric id. I just try with iri, same result

nicodmf:
i just see https://github.com/api-platform/core/pull/2326, this fix isn't aware for all cases, neither last PR...

What do you think about ? Json-ld isn't a choice for my cto, so maybe have you an idea why resource class isn't provide (the class is an input in the resource). At least, why the id disapear ?

soyuka:
I need to understand the use case behind it, not sure this is expected behavior

nicodmf:
The use case is the send of many objects which should be received together because each change on the system has important cost. Send objects together permit to validate all changes and to transmit computional operations in one pass.

The object to deserialize looks like

```js
{
  rules: [
   {id: '', period: xx, type: "", values: [ {id, value, populationTarget} ] },
   {id: '', period: xx, type: "", values: [ {id, value, populationTarget} ] }
  ]
}
```

I found a workarround with a dedicated resource and messager=true. But actually, it remains the logger bug which is a NullLogger.

soyuka:
mb related https://github.com/api-platform/core/pull/1645

Issue #3975, opened 2021-01-22
No identifier defined "App\DTO\ActionRequest". You should add #[\ApiPlatform\Core\Annotation\ApiProperty(identifier: true)]" on the property identifying the resource."


sidz:
**API Platform version(s) affected**: 2.6.0

**Description**  
have a simple DTO with custom opreation and custom controller. After update from 2.5.9 to 2.6.0 I've got an error `No identifier defined "App\DTO\ActionRequest". You should add #[\ApiPlatform\Core\Annotation\ApiProperty(identifier: true)]" on the property identifying the resource."`
I am not sure whether I need to add ApiProperty annotation, where and why.

**How to reproduce**  
```
<?php

declare(strict_types=1);

namespace App\DTO;

use ApiPlatform\Core\Annotation\ApiResource;
use Symfony\Component\Validator\Constraints as Assert;

/**
 * @ApiResource(
 *     itemOperations={},
 *     collectionOperations={
 *         "action"={
 *             "method"="POST",
 *             "path"="/action",
 *             "controller"=Action::class,
 *             "openapi_context"={"tags"={"Action"}}
 *         }
 *     }
 * )
 */
final class ActionRequest
{
    /**
     * @Assert\NotBlank()
     */
    public string $prop1;
    /**
     * @Assert\NotBlank()
     */
    public string $prop2;
}
```
**Additional Context**
Code above works fine with Api Platform 2.5.9

maks-rafalko:
Related to https://github.com/api-platform/core/pull/3871

It would be nice to have [some **docs** / upgrade instructions](https://github.com/api-platform/docs/issues/1229) explaining why `id` is required and for what cases. For such cases as above, where we don't have `item` operations and use DTO just for the custom controller in `POST` operation, IMO `id` is not needed.

soyuka:
identifiers are only needed for classes (DTO, POPO) marked with `#[ApiResource]`. `App\DTO\Action` is not the class shown above.

sidz:
@soyuka that is what I get on exactly this class (I've just change class and property names) after update to 2.6.0

maks-rafalko:
@sidz could you please make the following change in your description:

```diff
- No identifier defined "App\DTO\Action"
- No identifier defined "App\DTO\ActionRequest"
```

I guess this is what confused really happens and this is why @soyuka confused

soyuka:
[...]

PR #4303, opened 2021-06-03
Fix api-platform#4037 allow POST operations without identifiers on raw json

### LINKED ISSUE #4037: Identifier should be optional when using only a POST operation
**API Platform version(s) affected**: 2.6.2

**Description**  

POST should allow not having any identifiers #3975

shouze:
| Q             | A
| ------------- | --- 
| Bug fix?    | yes
| New feature | no
| BC Breaks? | no
| Deprecations? | no (fixes bc)
| Tickets | fixes #4037 #3975 
| License | MIT
| Doc PR | TODO

This allows POST operations without itemOperations and without identifiers on raw json :)


soyuka:
Can you add a behat test so that I understand the need behind this?

PR #4304, opened 2021-06-03
fix(openapi): do not throw error with non standard HTTP verb

qboot:
| Q             | A
| ------------- | ---
| Branch       | current stable (2.6)
| License       | MIT

Hello!

I'm trying to use a non standard HTTP verb in my ApiResource operations:
```
collectionOperations={
 *          "test":{
 *              "method": "TEST",
 *              "path": "/something",
 *              "controller": TesterController::class,
 *          },
 *      },
``` 

Everything works great, but when I want to access the Swagger UI (`/docs`), I got an error: 
```
Attempted to call an undefined method named "withTEST" of class "ApiPlatform\Core\OpenApi\Model\PathItem".
```

It seems that we don't check the HTTP verb before creating the related `PathItem` method here: 
```php
$pathItem = $pathItem->{'with' . ucfirst($method)}(new Model\Operation(
```
https://github.com/api-platform/core/blob/2.6/src/OpenApi/Factory/OpenApiFactory.php#L271

OpenApi only supports: `['GET', 'PUT', 'POST', 'DELETE', 'OPTIONS', 'HEAD', 'PATCH', 'TRACE']`.
So I propose this fix: continue if the HTTP verb is unknown, which fix the above error.

Thanks

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -27,6 +27,7 @@ use ApiPlatform\Core\Metadata\Resource\Factory\ResourceNameCollectionFactoryInte
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
 use ApiPlatform\Core\OpenApi\Model;
 use ApiPlatform\Core\OpenApi\Model\ExternalDocumentation;
+use ApiPlatform\Core\OpenApi\Model\PathItem;
 use ApiPlatform\Core\OpenApi\OpenApi;
 use ApiPlatform\Core\OpenApi\Options;
 use ApiPlatform\Core\Operation\Factory\SubresourceOperationFactoryInterface;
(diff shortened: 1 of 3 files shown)


alanpoulain:
I don't think it's a good idea to use a non-standard HTTP verb.
What is your use case?

qboot:
There is a lot more HTTP verb than the few from RFC7231 accepted by OpenApi. It's actually in debate to accept all of them in this comment https://github.com/OAI/OpenAPI-Specification/issues/1747.

IMO it's weird to be able to use a non-standard HTTP verb, without problem, but then be constrained by the documentation's generation.

A same PR have been made in NelmioApiDocBundle: https://github.com/nelmio/NelmioApiDocBundle/pull/1294

lyrixx:
> I don't think it's a good idea to use a non-standard HTTP verb.

But it's all about REST. you can use the verb you want... there is no limit.

And in my use case is the following: an endpoint to test something: you configure some credential (like slack) token, you click on "test" on the interface, and it will emit a slack notification. It could not be a GET one, since there is a body and browser does not support body + GET. And if I use POST, APIP will try to save the new configuration. So `TEST` verb is a good one here.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -144,6 +145,11 @@ private function collectPaths(ResourceMetadata $resourceMetadata, string $resour
             $resourceClass = $operation['resource_class'] ?? $rootResourceClass;
             $path = $this->getPath($resourceShortName, $operationName, $operation, $operationType);
             $method = $resourceMetadata->getTypedOperationAttribute($operationType, $operationName, 'method', 'GET');
+
+            if (!\in_array($method, PathItem::$methods, false)) {
```
soyuka, line 149:
```suggestion
            if (!\in_array($method, PathItem::$methods, true)) {
```

soyuka:
:+1: on this, see comment though

src/OpenApi/Factory/OpenApiFactory.php
qboot: oops, thanks. Fixed!

soyuka:
Thanks @qboot !

PR #4307, opened 2021-06-05
Fix: Improved Doctrine SearchFilter for entities that have an (ORM) identifier that is not just 'id'

VincentChalnot:
Still won't work with multiple identifiers but it's probably too complicated to support in a generic filter.

| Q             | A
| ------------- | ---
| Branch?       | 2.6
| License       | MIT

SearchFilter does not work when searching a related entity that is not identified by a field name 'id'.


VincentChalnot:
@soyuka This fixes the issue I was ranting about on Twitter the other day.

alanpoulain:
Hello,
Could you please fix the unit tests?
Also I don't think we need to keep the "old way" if the new way of retrieving the identifiers is working.

soyuka:
It's not that easy to fix and we're aware of the issue :).

PR #4310, opened 2021-06-08
Add hydra:previous to the hydra view schema properties

coudenysj:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

I don't know if this is intentional, but the `hydra:previous` property is not in the SchemaFactory (rendering the openapi specs incomplete).

### DIFF
--- a/features/openapi/docs.feature
+++ b/features/openapi/docs.feature
@@ -146,6 +146,10 @@ Feature: Documentation support
                             "type": "string",
                             "format": "iri-reference"
                         },
+                        "hydra:previous": {
+                            "type": "string",
+                            "format": "iri-reference"
+                        },
                         "hydra:next": {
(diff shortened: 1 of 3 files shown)


alanpoulain:
Hello,
It should target `2.6` and you should have at least a unit test, but I think it is not intentional. @soyuka WDYT?

coudenysj:
2.6 is the target now and added a unit test.

alanpoulain:
Some Behat tests need to be fixed too.

coudenysj:
Behat fix has been added, I don't think the remaining errors have to do with my changes.

dunglas (approved):

dunglas:
Good catch, thank you @coudenysj!

PR #4313, opened 2021-06-09
fix(ulid): convert Ulid object to binary for ORM search filter

misaert:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

When I want filter a resource by a relation property defined by an Ulid identifier, it doesn't work. This PR adds the conversion to binary before calling the query builder of Doctrine (as explained on the Symfony documentation: https://symfony.com/doc/current/components/uid.html#storing-ulids-in-databases).

Example:
```
Given the entity `Fruit` has a relation `ManyToOne` with the entity `Producer`
And the ID for each entity is Ulid type
And I have one producer with Ulid `01F7RK5K7FE9V0WPT8AGXS2C2J` who has a fruit
When I call the API `GET /api/fruits?producer=/api/producers/01F7RK5K7FE9V0WPT8AGXS2C2J`
Then I want to have one fruit as result
```
Currently, the result is 0 because the query builder uses the string representation that does not correspond to the binary stored in the database.

It is my first contribution. I don't know if I have to add a test for it and how, and I don't know if I have to change the Changelog file right now.

Don't hesitate to help me to improve this PR :slightly_smiling_face:

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
@@ -28,6 +28,8 @@ use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
+use Symfony\Component\Uid\AbstractUid;
+use Symfony\Component\Uid\Ulid;
 
 /**
  * Filter the collection by given properties.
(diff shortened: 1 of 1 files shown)


alanpoulain:
Hello,
Thanks for the PR but we don't really want to handle binary UUID like this.
We tried to have a general approach (https://github.com/api-platform/core/pull/3774) but it has created a lot of issues and has been reverted (https://github.com/api-platform/core/pull/4134).
As I said in this comment (https://github.com/api-platform/core/pull/4134#issuecomment-833664840), we don't want to have a whitelist like you added in the search filter and we recommend to create your own search filter instead.

misaert:
> Hello,
> Thanks for the PR but we don't really want to handle binary UUID like this.
> We tried to have a general approach (#3774) but it has created a lot of issues and has been reverted (#4134).
> As I said in this comment ([#4134 (comment)](https://github.com/api-platform/core/pull/4134#issuecomment-833664840)), we don't want to have a whitelist like you added in the search filter and we recommend to create your own search filter instead.

Hello,

Thanks for the quick answer. I understand. I hope a generic solution will be found in the future. For now, I am following your recommandation with my own search filter on my project.

soyuka:
Closing, we'll attempt something to fix this properly but it'll probably invoke using a new filter or a complete rewrite of the SearchFilter.

PR #4315, opened 2021-06-10
fix(doctrine): denormalization of properties with embeds many which omit target document directive

wuchen90:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| License       | MIT

According to [doctrine/mongodb-odm docs](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/embedded-mapping.html#mixing-document-types), it's possible to embed many without declaring the `target-document` directive.

In api-platform, if we omit this directive, the deserialization won't work.

This fix makes the type guesser of MongoDB return null so other guessers will be used.


alanpoulain:
Hello,
Thanks for this fix, it seems OK to me.
The PR wuld be even better if there was a Behat test, do you think you can provide one?

wuchen90:
> The PR would be even better if there was a Behat test, do you think you can provide one?

Ok, I'll try my best to find my way in this `tests` folder.

wuchen90:
@alanpoulain 
All done.

src/Bridge/Doctrine/MongoDbOdm/PropertyInfo/DoctrineExtractor.php
```diff
@@ -65,6 +65,10 @@ public function getTypes($class, $property, array $context = [])
         if ($metadata->hasAssociation($property)) {
             $class = $metadata->getAssociationTargetClass($property);
 
+            if (!\is_string($class) && null === $class) {
```
wuchen90, line 68: This is some ugly attempt to please PHPStan and I don't know if this will work.
I'm not familiar with PHPStan so if someone knows how to fix it I'm all ears.

This problem here is the interface `$metadata->getAssociationTargetClass` returns a `string` in phpDoc, but implementation of it returns `?string`.

See https://github.com/doctrine/mongodb-odm/blob/ccf26d9e023387e9aae86951186c043b3222fe70/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1797
and https://github.com/doctrine/persistence/blob/79e97bcb1dfe3b242961253d81b8357a04e96ca1/lib/Doctrine/Persistence/Mapping/ClassMetadata.php#L128

alanpoulain: The best way to solve this is to add a PHPDoc on the returned variable to force PHPStan to consider it as `?string`.
You can also launch PHPStan locally (`vendor/bin/phpstan analyze .`) in order to verify if it works.

wuchen90: Great! It works.
Thanks.

alanpoulain (approved):

alanpoulain:
Thank you @wuchen90.

nass59:
Hello, I have a use case where i need to store a document as a ReferenceOne, but it can be different types of documents.
And, since this PR there is an error on ReferenceOne when the `targetDocument` directive is not declared.

`Property "content" on resource "App\Document\SomeDocument" is declared as a subresource, but its type could not be determined`.

So is there a workaround to that ?
Thanks

wuchen90:
Hi @nass59 
When the `targetDocument` directive is not there, the resolution of the object type is delegated to other guessers.
In your case, you need to use a discriminator map to help the internal of Symfony object serializer to guess the type.
See https://symfony.com/doc/current/components/serializer.html#serializing-interfaces-and-abstract-classes.
Hope this will help you.

Issue #4318, opened 2021-06-11
OpenApiNormalizer fails on Entity with `paths` property 


tassilogroeper:
**API Platform version(s) affected**: 2.6

**Description**  
Introduced in `v2.6`, the new core subcomponent OpenApi/Serializer/OpenApiNormalizer.php throws an error when calling `recursiveClean`. Given an ApiResource with a property named `paths`. Visiting the swagger/OpenAPi documentation the error:
`Warning: Undefined array key "paths"`

```php
// in vendor/api-platform/core/src/OpenApi/Serializer/OpenApiNormalizer.php (line 64)
// Side effect of using getPaths(): Paths which itself contains the array
if ('paths' === $key) {
    $value = $data['paths'] = $data['paths']['paths'];
    if ($value) {
        ksort($value);
    }
    unset($data['paths']['paths']);
}
```

**How to reproduce**  
Have an Api resource with a property `paths` and visit the swagger/OpenAPi documentation (`/api/docs` by default).
```php
<?php
// api/src/Entity/Offer.php

namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use Symfony\Component\Serializer\Annotation\Groups;

#[ApiResource(iri: 'http://schema.org/Offer')]
class Offer
{
    // ...

    /**
     * @Groups({"offer_read"})
     */
    private array $paths = [];

    // ...
}
```

**Possible Solution**  
Refactor `vendor/api-platform/core/src/OpenApi/Serializer/OpenApiNormalizer.php` to only apply the array modification once or when needed.
Also possible: Declare `paths` a reserved property word.

**Workaround**
Rename the colliding property.

**Additional Context**  
![FireShot Capture 270 - Warning_ Undefined array key _paths_ (500 Internal Server Error)_ - interactapi lndo site](https://user-images.githubusercontent.com/1324354/121690376-120b5500-cac6-11eb-9b9b-5cfd1f43d59d.png)

pheadeaux:
@soyuka You have been working on these parts, right? Please check the PR.

PR #4319, opened 2021-06-11
fix(openapi): failing recursion on api resources with "paths" key

pheadeaux:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | #4318
| License       | MIT

As described in issue #4318 the function recursiveClean does nest one more time and fails, if an ApiResource has a property named `paths`. Found no tests to reproduce or fail due to this fact.

### DIFF
--- a/src/OpenApi/Serializer/OpenApiNormalizer.php
+++ b/src/OpenApi/Serializer/OpenApiNormalizer.php
@@ -59,7 +59,7 @@ final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsM
             }
 
             // Side effect of using getPaths(): Paths which itself contains the array
-            if ('paths' === $key) {
+            if ('paths' === $key && \is_array($value) && isset($data['paths']['paths'])) {
                 $value = $data['paths'] = $data['paths']['paths'];
                 if ($value) {
                     ksort($value);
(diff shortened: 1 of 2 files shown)


dunglas:
LGTM. Thanks for tracking and fixing this @pheadeaux. Could you add a new unit test to prevent future regressions please?

pheadeaux:
@dunglas done. No need to add a new test, I just updated the existing one to cover the edge case. The Regressen happens when diving into the "Components>Schema" of an OpenApi Object.

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -59,7 +59,7 @@ private function recursiveClean($data): array
             }
 
             // Side effect of using getPaths(): Paths which itself contains the array
-            if ('paths' === $key) {
+            if ('paths' === $key && \is_array($value) && isset($data['paths']['paths'])) {
```
alanpoulain, line 62: Shouldn't the `is_array` use `$data` and not `$value`?

pheadeaux: @alanpoulain granted. But if you look up at [line 57](https://github.com/api-platform/core/pull/4319/files#diff-66a6198b9d63f63ba3a635fa8f23143870ed8c73842fba6b142b1c1a118b4e96R57), the if statement also directly checks the `$value` contents to be an array. This leads us to the question: check for `is_array($value)` at all or just remove it?

pheadeaux:
@dunglas @alanpoulain Can somebody assist with the failing test? I do not see, how PHP-8 and Symfony 5.3 are differently affected by these changes than lets say PHP-7 and earlier versions of Symfony.

src/OpenApi/Serializer/OpenApiNormalizer.php
alanpoulain: I think you can remove it.

dunglas:
The test failure is unrelated, don't bother with it, I'm on it (#4323).

alanpoulain:
We are fixing the PHP 8 / SF 5.3 tests, your PR has no impact on them.

pheadeaux:
@alanpoulain @dunglas actually I have a better idea to solve the underlying problem. Please review #4325

pheadeaux:
closed in favor of  #4325

PR #4325, opened 2021-06-15
Fix openapi normalization

pheadeaux:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | #4318
| License       | MIT

As described in issue #4318 the function recursiveClean does nest one more time and fails, if an ApiResource has a property named `paths`. Updated tests to reproduce the failing parts.

This is version 2, a more general approach, since the logic inside `\ApiPlatform\Core\OpenApi\Serializer\OpenApiNormalizer::recursiveClean` was doing two things:
1. cleaning out `extensionProperties` keys in a recursion and
2. with every recursion, trying to sort the `schemas` and `paths` properties of the OpenApi object

So why not do the sorting in the Object generation directly and fix the double nested `paths`  issue with the normalizing call itself.

Idea is taken from: [Using Callbacks to Serialize Properties with Object Instances](https://symfony.com/doc/current/components/serializer.html#using-callbacks-to-serialize-properties-with-object-instances)

### DIFF
--- a/src/OpenApi/Model/Components.php
+++ b/src/OpenApi/Model/Components.php
@@ -29,6 +29,8 @@ final class Components
 
     public function __construct(\ArrayObject $schemas = null, \ArrayObject $responses = null, \ArrayObject $parameters = null, \ArrayObject $examples = null, \ArrayObject $requestBodies = null, \ArrayObject $headers = null, \ArrayObject $securitySchemes = null, \ArrayObject $links = null, \ArrayObject $callbacks = null)
     {
+        !$schemas ?: $schemas->ksort();
+
         $this->schemas = $schemas;
         $this->responses = $responses;
         $this->parameters = $parameters;
--- a/src/OpenApi/Model/Components.php
+++ b/src/OpenApi/Model/Components.php
@@ -29,7 +29,9 @@ final class Components
 
     public function __construct(\ArrayObject $schemas = null, \ArrayObject $responses = null, \ArrayObject $parameters = null, \ArrayObject $examples = null, \ArrayObject $requestBodies = null, \ArrayObject $headers = null, \ArrayObject $securitySchemes = null, \ArrayObject $links = null, \ArrayObject $callbacks = null)
     {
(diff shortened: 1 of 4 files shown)


pheadeaux:
relates to #4319 4319

src/OpenApi/Model/Components.php
```diff
@@ -29,6 +29,8 @@ final class Components
 
     public function __construct(\ArrayObject $schemas = null, \ArrayObject $responses = null, \ArrayObject $parameters = null, \ArrayObject $examples = null, \ArrayObject $requestBodies = null, \ArrayObject $headers = null, \ArrayObject $securitySchemes = null, \ArrayObject $links = null, \ArrayObject $callbacks = null)
     {
+        !$schemas ?: $schemas->ksort();
```
dunglas, line 32: I would use a standard `if` for readability.

dunglas (changes requested):

src/OpenApi/Model/Components.php
pheadeaux: @dunglas done.

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -38,8 +39,14 @@ public function __construct(NormalizerInterface $decorated)
      */
     public function normalize($object, $format = null, array $context = []): array
     {
+        $pathsCallback = function ($innerObject) {
```
alanpoulain, line 42:
```suggestion
        $pathsCallback = static function ($innerObject) {
```

alanpoulain (approved):
LGTM, could you rebase?

pheadeaux:
> LGTM, could you rebase?

@alanpoulain done

src/OpenApi/Serializer/OpenApiNormalizer.php
pheadeaux: done

alanpoulain:
Thank you @pheadeaux.

PR #4327, opened 2021-06-16
Add a flag to allow an opt-out from AddTagsListener

domgraziano:
| Q             | A
| ------------- | ---
| Tickets       | https://github.com/api-platform/core/issues/3168
| License       | MIT

I noticed that at each request this `ApiPlatform\Core\HttpCache\EventListener\AddTagsListener` adds `Cache-Tags` just before returning the response. For large interconnected entities this can have unexpected consequences and it should really be an opt-in listener in my mind. In my case for example I have https://linkerd.io/ limiting infra service communication headers to 12k. On each get because of this listener I have bigger than 12k headers for specific "big" linked resources ... this happens on GET(s) too! For backward compatibility though I added a flag so that by default is turned on but can be turned off easily under `http_cache` config.

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -585,6 +585,8 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             $loader->load('doctrine_orm_http_cache_purger.xml');
         }
 
+        $loader->load('http_cache_tags.xml');
+
         if ($config['http_cache']['cloudflare'] === true) {
             $iriConverter = new Reference('api_platform.iri_converter');
             $container->register(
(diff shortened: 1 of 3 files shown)


alanpoulain:
You can already turn off the invalidation system:
```yaml
api_platform:
    http_cache:
        invalidation: false
```

domgraziano:
> You can already turn off the invalidation system:
> 
> ```yaml
> api_platform:
>     http_cache:
>         invalidation: false
> ```

thank you! I was aware of that, this is an attempt instead to opt-out from the added cache tags in the response headers, leaving the invalidation enabled without causing bkwrd incompatibility.
Let's say that I want to change the mechanism of how cache tags are stored in the headers (see #3168) sometimes are just too big and I want compress them (for example). So I would not need that subscriber adding all the uris to my headers.

alanpoulain:
So you want to keep the Varnish purger but not the `Cache-Tags` header?
Personally I find it too much fine-grained. Why not removing yourself it in your `DependencyInjection`?

domgraziano:
> So you want to keep the Varnish purger but not the `Cache-Tags` header?
> Personally I find it too much fine-grained. Why not removing yourself it in your `DependencyInjection`?

Ok, I see what you mean ... closing this then

PR #4328, opened 2021-06-17
Fix : the path of the nested property

LoicComparet:
Allows you to filter on nested properties going up the path

| Q             | A
| ------------- | ---
| Tickets       | 
| License       | MIT
| Doc PR        | 

I noticed that filtering on a nested property didn't work: the path of the nested property was null, so we were looking for that nested property in the parent object.

Example : 

{ "vehicle": { "brand": "xxx", "model": "yyy"}}

filter like "vehicle.brand" don't work because path is null instead of "vehicle"


alanpoulain:
LGTM. Could you add a test (PHPUnit and ideally it should be covered by a Behat test too)?

LoicComparet:
I can but not immediately :)

PR #4337, opened 2021-06-23
fix(openapi): Collection fields VS property metadata

guilliamxavier:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | related to https://github.com/api-platform/core/pull/4182#discussion_r612125005
| License       | MIT
| Doc PR        | 

The initial implementation was wrong to pass the *collection* `$propertyMetadata` for each *nested field*, so I fixed it to use a new PropertyMetadata (empty, for lack of anything better)

### DIFF
--- a/src/Core/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
+++ b/src/Core/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
@@ -57,7 +57,7 @@ final class PropertySchemaCollectionRestriction implements PropertySchemaRestric
                 $required[] = $field;
             }
 
-            $restriction['properties'][$field] = $this->mergeConstraintRestrictions($baseConstraint, $propertyMetadata);
+            $restriction['properties'][$field] = $this->mergeConstraintRestrictions($baseConstraint, new ApiProperty());
         }
 
         if ($required) {


src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php
```diff
@@ -57,7 +57,7 @@ public function create(Constraint $constraint, PropertyMetadata $propertyMetadat
                 $required[] = $field;
             }
 
-            $restriction['properties'][$field] = $this->mergeConstraintRestrictions($baseConstraint, $propertyMetadata);
+            $restriction['properties'][$field] = $this->mergeConstraintRestrictions($baseConstraint, new PropertyMetadata());
```
guilliamxavier, line 60: Note: ideally we would extract the nested field type from the Collection property metadata, but I don't know how to do that 😕

alanpoulain: Not sure to understand this change, could you explain what it does please?

guilliamxavier: `$propertyMetadata` corresponds to the Collection property, but `$baseConstraint` corresponds to the nested `$field`.
Related: https://github.com/api-platform/core/pull/4182/commits/be30f17071716e8ee78fbda025106fa970fe20ae

alanpoulain: I think I understand but why using a parameter then? Because we need to "hydrate" the property metadata with the nested field type?

guilliamxavier: Indeed, we could remove the 2nd parameter of https://github.com/api-platform/core/blob/0d8a016d762a9465f12098d3af6905c867e54fa5/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php#L81 but we would still need to have one anyway for https://github.com/api-platform/core/blob/0d8a016d762a9465f12098d3af6905c867e54fa5/src/Bridge/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaCollectionRestriction.php#L88
And I was hoping one could eventually find a way to deduce a `$fieldPropertyMetadata` (at least a `$fieldType` for `new PropertyMetadata($fieldType)`) from `$propertyMetadata`+`$field` and/or `$baseConstraint`

guilliamxavier: (Update: the PR was missing a description, fixed now)

soyuka:
I like this could you rebase plesae?

guilliamxavier:
@soyuka: Sorry for the delay. Now rebased and updated. Thanks

Issue #1776, opened 2018-03-14
[RFC] Use xkey to purge cache


jderusse:
Reading the PR https://github.com/FriendsOfSymfony/FOSHttpCache/pull/409/files on FOSHttpCache, Their is a more efficient way to purge cache. 

Documentation http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys

dunglas:
Very interesting. 👍 to support this feature.
It will probably also impact the design of #1689 (ping @Toflar).
We also need to check how we can make the interface/code compatible with both this and CloudFlare's cache tags.

I think we can deprecate the current system for this new one.

Toflar:
Yeah, I've seen that. So - again - another reason to just ask the purger to add the tags in whatever format it wants to. We can extend that and allow the purger not only to format the iris but also modify the response the way it likes. So in code that would look like this:

```php
class VarnishPurger implements NewPurgerInterface
{
    public function purge(array $iris)
    {
        // same/similar logic as now
    }

    public function tagResponse(Response $response, array $iris)
    {
        $response->headers->set('xkey', implode(',' $iris));
    }
}
```

```php
final class AddTagsListener
{
    private $iriConverter;
    private $purger;

    public function __construct(IriConverterInterface $iriConverter, NewPurgerInterface $purger)
    {
        $this->iriConverter = $iriConverter;
        $this->purger = $purger;
    }

    public function onKernelResponse(FilterResponseEvent $event)
    {
        // same/similar logic as now to get $resources/$iris

        $this->purger->tagResponse($event->getResponse(), $resources);
    }
}
```

It just allows to basically do anything. You can add one header, 5 headers, you can have a `ChainedPurger`, I can have my `Psr6CachePurger` and name/format the tags myself etc. Full flexibility and all we have to do is deprecate the current `PurgerInterface` for a new one. Or just leave it in place and provide an `AdvancedPurgerInterface`.

Toflar:
Any feedback on that? 😄

teohhanhui:
Relevant?

https://www.youtube.com/watch?v=1Uzn31cQUQ8

https://cdn2.hubspot.net/hub/209523/file-366502730-pdf/Cache_Invalidation_Strategies_A_Varnish_Webinar.pdf

dunglas:
@Toflar can we make something independent from the Response class? (I plan to add PSR-7 support at some point)?

[...]

PR #4340, opened 2021-06-25
feat: support xkey

LoicBoursin:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | https://github.com/api-platform/core/issues/1776
| License       | MIT

This PR adds support for xkey in varnish


src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+        $currentHeader = '';
+
+        foreach ($keys as $position => $key) {
+            // Key can be added to header
+            if (\strlen($currentHeader.($position > 0 ? $glue : '').$key) <= $this->maxHeaderLength) {
+                $currentHeader .= ($position > 0 ? $glue : '').$key;
+            } else {
+                if (\strlen($key) > $this->maxHeaderLength) {
+                    // TODO: rename this exception to better represent the error
+                    throw new \Exception(sprintf('Path : %s is too long to fit current max header length (currently set at %s). You must increase max header length.', $key, $this->maxHeaderLength));
```
alanpoulain, line 92:
```suggestion
                    throw new \Exception(sprintf('Path: %s is too long to fit current max header length (currently set to %s). You must increase max header length.', $key, $this->maxHeaderLength));
```
This exception is weird since the header length cannot be modified by the user.

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+    {
+        foreach ($this->chunkKeys($iris, $glue) as $keys) {
+            $this->purgeKeys($keys);
+        }
+    }
+
+    private function purgeKeys(string $keys): void
+    {
+        foreach ($this->clients as $client) {
+            $client->request('PURGE', '', ['headers' => ['ApiPlatform-Purge-Keys' => $keys]]);
```
dunglas, line 68: Shouldn't the header be called `xkey`?

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -40,13 +40,15 @@ final class PurgeHttpCacheListener
     private $resourceClassResolver;
     private $propertyAccessor;
     private $tags = [];
+    private $xKeyPurger;
 
-    public function __construct(PurgerInterface $purger, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
+    public function __construct(PurgerInterface $purger, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, PurgerInterface $xKeyPurger, PropertyAccessorInterface $propertyAccessor = null)
```
LoicBoursin, line 45: I believe we should chain services & I'm not allowed to add another PurgerInterface like that ?

soyuka: yes you can but the argument must be at the last position and `nullable`

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -104,6 +106,7 @@ public function postFlush(): void
         }
 
         $this->purger->purge(array_values($this->tags));
+        $this->xKeyPurger->purge(array_values($this->tags));
```
soyuka, line 109: what if xkeypurger is not present?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -411,6 +411,10 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
                                     ->end()
                                     ->info('To pass options to the client charged with the request.')
                                 ->end()
+                                ->booleanNode('xkey')
+                                    ->defaultFalse()
+                                    ->info('Enable support for xkey (varnish)')
```
soyuka, line 416:
```suggestion
                                    ->info('Enable support for xkey invalidation')
```

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -18,27 +18,32 @@
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
 
 /**
- * Sets the list of resources' IRIs included in this response in the "Cache-Tags" HTTP header.
+ * Sets the list of resources' IRIs included in this response in the "Cache-Tags" and "xkey" HTTP headers.
```
soyuka, line 21:
```suggestion
 * Sets the list of resources' IRIs included in this response in the "Cache-Tags" and/or "xkey" HTTP headers.
```

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $currentHeader = '';
+
+        foreach ($keys as $position => $key) {
+            // Key can be added to header
+            if (\strlen(sprintf('%s%s%s', $currentHeader, $position > 0 ? $glue : '', $key)) <= $this->maxHeaderLength) {
+                $currentHeader .= sprintf('%s%s', $position > 0 ? $glue : '', $key);
+            } else {
+                if (\strlen($key) > $this->maxHeaderLength) {
+                    // TODO: rename this exception to better represent the error
```
soyuka, line 91:
```suggestion
```

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+    public function purge(array $iris)
+    {
+        if (!$iris) {
+            return;
+        }
+
+        $irisChunks = array_chunk($iris, \count($iris));
+
+        foreach ($irisChunks as $irisChunk) {
+            // TODO: glue could be a parameter in a config file
```
soyuka, line 53: let's do it?

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -133,6 +133,7 @@ class ApiPlatformExtensionTest extends TestCase
         'http_cache' => ['invalidation' => [
             'enabled' => true,
             'varnish_urls' => ['test'],
+            'xkey' => false,
```
soyuka, line 136: Is it possible to use `enabled => false` and `xkey => true`?

soyuka (changes requested):

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
LoicBoursin: if enable => false : both xkey and http_tags are disabled.
if enable => true : we can enable xkey and/or http_tags but not disable both

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -40,13 +40,19 @@ final class PurgeHttpCacheListener
     private $resourceClassResolver;
     private $propertyAccessor;
     private $tags = [];
+    private $xKeyPurger;
+    private $xkey;
```
alanpoulain, line 44: It does no seem to be used.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -411,6 +411,18 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->defaultFalse()
+                                    ->info('Enable support for xkey invalidation')
+                                ->end()
+                                ->scalarNode('xkey_glue')
+                                    ->defaultValue(' ')
+                                    ->info('xkey glue between keys')
+                                ->end()
+                                ->booleanNode('http_tags')
+                                    ->defaultTrue()
+                                    ->info('Enable Cache-Tags invalidation')
```
alanpoulain, line 424:
```suggestion
                                    ->info('Enable support for Cache-Tags invalidation.')
```

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_http_cache_purger.xml
```diff
@@ -12,7 +12,10 @@
             <argument type="service" id="api_platform.http_cache.purger" />
             <argument type="service" id="api_platform.iri_converter" />
             <argument type="service" id="api_platform.resource_class_resolver" />
-
+            <argument>null</argument>
```
alanpoulain, line 15: Why the `api_platform.property_accessor` service is not used?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -411,6 +411,18 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
                                     ->end()
                                     ->info('To pass options to the client charged with the request.')
                                 ->end()
+                                ->booleanNode('xkey')
+                                    ->defaultFalse()
+                                    ->info('Enable support for xkey invalidation')
+                                ->end()
+                                ->scalarNode('xkey_glue')
```
alanpoulain, line 418: IMO it would be better to have a `xkey` config attribute, with an `enabled` and `glue` sub-attributes.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -411,6 +411,18 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->defaultValue(' ')
+                                    ->info('xkey glue between keys')
+                                ->end()
+                                ->booleanNode('http_tags')
+                                    ->defaultTrue()
```
alanpoulain, line 423: It would be better to use `canBeDisabled()`.

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -18,27 +18,36 @@
...
  */
 final class AddTagsListener
 {
     private $iriConverter;
+    private $xkey;
+    private $xkeyGlue;
+    private $httpTags;
 
-    public function __construct(IriConverterInterface $iriConverter)
+    public function __construct(IriConverterInterface $iriConverter, bool $xkey = false, string $xkeyGlue = ' ', bool $httpTags = true)
```
alanpoulain, line 41:
```suggestion
    public function __construct(IriConverterInterface $iriConverter, bool $xkeyEnabled = false, string $xkeyGlue = ' ', bool $httpTagsEnabled = true)
```

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        $irisChunks = array_chunk($iris, \count($iris));
+
+        foreach ($irisChunks as $irisChunk) {
+            $this->purgeRequest($irisChunk);
+        }
+    }
+
+    private function purgeRequest(array $iris)
```
alanpoulain, line 59: Why not `purgeIris`?

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+
+            return;
+        }
+
+        $currentHeader = '';
+
+        foreach ($keys as $position => $key) {
+            // Key can be added to header
+            if (\strlen(sprintf('%s%s%s', $currentHeader, $position > 0 ? $this->xkeyGlue : '', $key)) <= $this->maxHeaderLength) {
+                $currentHeader .= sprintf('%s%s', $position > 0 ? $this->xkeyGlue : '', $key);
```
alanpoulain, line 89: IMO it would be better to create the substring before and to calculate the length then.

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+            } else {
```
alanpoulain, line 90: Please use an early return instead.

src/HttpCache/VarnishXKeyPurger.php
```diff
@@ -0,0 +1,104 @@
...
+                if (\strlen($key) > $this->maxHeaderLength) {
+                    throw new \Exception(sprintf('Path: %s is too long to fit current max header length (currently set to %s). You must increase max header length.', $key, $this->maxHeaderLength));
```
alanpoulain, line 92: But how does the user increase it?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka: this is part of the `invalidation` we want to support both as the servers on the invalidation node are used there as well.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka: imo this is overrated as the http tags is only a way of invalidating

src/HttpCache/VarnishXKeyPurger.php
soyuka: there is a configuration for this

src/HttpCache/VarnishXKeyPurger.php
LoicBoursin: Good idea ! I'll change that.

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
LoicBoursin: It is used in the postFlush method L112-118 to determine if tags need to be purged using either/both xkey or/and Cache-Tags

alanpoulain: Renaming it to `$xkeyEnabled` would be clearer then.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
alanpoulain: Not sure I understand what you mean. We are in the `invalidation` part, and there are two modes: `xkey` and `http_tags`.
Each one should have its own configuration.

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
alanpoulain: Yes but using `canBeDisabled()` would allow to add more options in the future.

src/HttpCache/VarnishXKeyPurger.php
alanpoulain: You're right. Maybe it would be nice to tell the user which configuration key it is then?
```suggestion
                    throw new \Exception(sprintf('Path: %s is too long to fit current max header length (currently set to %s). You must increase it in http_cache/invalidation/max_header_length configuration.', $key, $this->maxHeaderLength));
```

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka: does this work with `canBeEnabled`?
```
xkey: false
```

or we need the full 

```
xkey:
  enabled: true
```

src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm_http_cache_purger.xml
soyuka: I'm not sure we should add it here indeed

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
alanpoulain: Yes! https://github.com/symfony/symfony/blob/eb2cae6b4901b2d5ab7d394426c262e4c2c3a4a6/src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php#L304

src/Bridge/Doctrine/EventListener/PurgeHttpCacheListener.php
LoicBoursin: Done.

alanpoulain (approved):

soyuka:
thanks @LoicBoursin !

PR #4341, opened 2021-06-25
fix: Optional expression language on Mercure

djpremier:
| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

Call ```addFunction``` only if has ```Expression Language```.

https://github.com/api-platform/core/pull/4153#discussion_r658715964


alanpoulain:
It would be nice to add an exception if the user uses them anyway.

djpremier:
> It would be nice to add an exception if the user uses them anyway.

Yes, but I had already placed the exception in the previous PR.

https://github.com/api-platform/core/blob/6336f6446722cd6728683c36fb70e0074553296d/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php#L229

alanpoulain (approved):

alanpoulain:
Thanks!

PR #3826, opened 2020-11-12
WIP: Use Accept-Language to select the locale

GregoireHebert:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | ~

Issue #1506, opened 2017-11-21
I18n support (translations, etc)


dkarlovi:
What would be a good way to go forward with internationalization of my resources, asssuming I'm using [Atlantic18/DoctrineExtensions](https://github.com/Atlantic18/DoctrineExtensions) (via [stof/StofDoctrineExtensionsBundle](https://github.com/stof/StofDoctrineExtensionsBundle))?

This was discussed in #127 where @MLKiiwy linked to [his fork](https://github.com/MLKiiwy/DunglasApiBundle/tree/tdd-i18n-support). Does this still look relevant to you (with newer versions/features released), was the direction he was going OK from API Platform's POV?

What would be solid next steps here?

binhle410:
If you want i18n with Doctrine, I suggest using KNP DoctrineBehaviors. It has better support for the translation to be put in a separate row and table.

tahirm:
any update on this?
Im also looking for recommended setup for i18n with api-platform.

pvgnd:
+1

zolex:
+1

beyer-johannes:
+1

teohhanhui:
I don't see how this is specific to API Platform. But it'd be good to have a doc entry for i18n...

alanpoulain:
Closed in favor of: https://github.com/api-platform/docs/issues/600

PR #4342, opened 2021-06-30
feat(wip): resource translation

alanpoulain:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #1506
| License       | MIT
| Doc PR        | TODO

Introduces two interfaces: `TranslatableInterface` and `TranslationInterface`.
It's the user responsibility to implement these interfaces how they want, either by themselves or by using extensions like https://github.com/KnpLabs/DoctrineBehaviors/blob/master/docs/translatable.md.

During the serialization, if the resource uses the `TranslatableInterface`, the attribute value will be taken from the resource translation (represented by `TranslationInterface`), if available for the current locale.

The JSON-LD context is also changed accordingly (see https://json-ld.org/spec/latest/json-ld/#string-internationalization).

TODO:
Manage the retrieval of all translations (resource-wise and client-wise).

To change the locale, ideally, we would need to use the `Accept-Language` header, but this is another PR: https://github.com/api-platform/core/pull/3826.

See also:
- https://github.com/Sylius/Sylius/issues/11412
- https://github.com/Sylius/Sylius/pull/12752


lchrusciel:
Hey Alan,

thanks a lot for your work, looks really good. I cannot wait to test it in Sylius. One question tho. Is it required to have `name` property on a translatable entity? Can we achieve the same result with the `getter` method or some YAML/XML config?

alanpoulain:
Hello @lchrusciel,
Yes of course, API Platform only needs to understand it as a property of your resource, so having only a getter is possible, as well as using the `properties` key in YAML / XML.

SpartakusMd:
Hey @alanpoulain, it seems really nice.

I wanted to mention another approach to implementing translations, that can be seen in https://github.com/Locastic/ApiPlatformTranslationBundle.

The translations there are stored in the `translations` property in the response. For translated properties, the format stays always the same, but the value will be in the requested language. I would say this approach is better in the fact that it doesn't change the format of existing fields in the response.

In the projects we use, the translations are not returned by default and can be requested via the `?groups[]=...` as we use GroupFilter.

For example if you make a get request in french and also retrieve the translations, you'll get the following:
```
{
    "title":"Dummy traduit en français",
    "content":"C'est un dummy !",
    "notTranslatedField": "Not translated",
    "translations": { 
        "en":{
            "title":"Dummy translated in English",
            "content":"It's a dummy!",
            "locale":"en"
        },
        "fr":{
            "title":"Dummy traduit en français",
            "content":"C'est un dummy !",
            "locale":"fr"
        }
    }
}
```

What do you think about this approach?

alanpoulain:
Hello @SpartakusMd.
I know this bundle, its way of doing it is OK but does not follow any standard.
I'm not sure to understand your point, if you request a resource in a specific language, the format of existing fields will not be changed.
With you example, if you send the request `GET /books/1` with the header `Accept-Language: fr`, you will have the following response:
```json
{
    "@context": {
        "@vocab": "http://example.com/docs.jsonld#",
        "@language": "fr"
    },
    "@id": "/books/1",
    "@type": "Book",
    "name": "Nom en français"
}
```

The format is only changed if you request all the translations.

SpartakusMd:
Exactly. The format is only changed if you request all the translations and it becomes something like this:

```
{
    "@context": {
        "@vocab": "http://example.com/docs.jsonld#",
        "@language": "fr"
    },
    "@id": "/books/1",
    "@type": "Book",
    "name": {
        "fr": "Nom en français"
    }
}
```

The field `name` changes from string to object. While I'm ok with it, it was flagged from the mobile teams it's hard for them to implement 2 different formats (there are cases when you need and don't need translations). This is why I was asking about it.

alanpoulain:
OK I understand your point but it will not be changed since the specification defines it like this: https://www.w3.org/TR/json-ld/#example-71-language-map-expressing-a-property-in-three-languages

celorodovalho:
There's no update about this PR?

VincentLanglet:
Hi, are you still interested by this PR ? Do you have time to finish it and resolve the conflict ?
Thanks

Also, now Symfony has his own Translatable interface, does it still make sens to have one in ApiPlatform ?

alanpoulain:
No I think this PR should be closed or taken over.

PR #4345, opened 2021-07-05
feat: OpenApi: Remove a path from Paths

TimoBakx:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT

There is currently no way to remove a path from the generated OpenAPI docs. This PR adds a method to more easily remove a path, for example in a decorator:

Before:
```
    public function __invoke(array $context = []): OpenApi
    {
        $openApi = $this->decorated->__invoke($context);

        $paths = $openApi->getPaths()->getPaths();

        unset(
            $paths['/users'],
            $paths['/users/{id}'],
        );

        $newPaths = new Paths();

        foreach ($paths as $path => $pathItem) {
            $newPaths->addPath($path, $pathItem);
        }

        return $openApi->withPaths($newPaths);
    }
```

After:
```
    public function __invoke(array $context = []): OpenApi
    {
        $openApi = $this->decorated->__invoke($context);

        $openApi->getPaths()->removePath('/users');
        $openApi->getPaths()->removePath('/users/{id}');

        return $openApi;
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 * Exception: Add the ability to customize multiple status codes based on the validation exception (#4017)
 * GraphQL: Fix graphql fetching with Elasticsearch (#4217)
 * ApiLoader: Support `_format` resolving (#4292)
+* OpenAPI: Allow removing paths
 
 ## 2.6.5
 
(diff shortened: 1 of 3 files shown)


soyuka:
this goes against immutability and you should prefer preventing these routes from existing in the first place.

TimoBakx:
@soyuka, thank you for your response. I'm a bit confused about this, though, as the Paths class is not immutable as is, it has a `addPath` method that mutates the list.

If it needs to be immutable, I'd be happy to change this (return a cloned object instead) and do the same with the `removePath()` method.

And there are situations where the route itself is needed, but the documentation is unwanted.

PR #4356, opened 2021-07-13
Fix: Problem to use partial searchFilter with an integer as input.

hogren:
SQL Server does not automatically cast an integer to a string in
the concatenation with + method (...WHERE field1 LIKE :param1 + :param2)

So there is an SQL Server error when we try to enter "1" in a partial filter : 
An exception occurred while executing 'SELECT […] WHERE b0_.label LIKE ('%' + ? + '%') ORDER BY […]' with params [1]:\n\nSQLSTATE [22018, 245]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Conversion failed when converting the varchar value '%' to data type int.

The CONCAT function is not available before the 2012 version.

I so simply force the casting when the searchFilter is not of type exact.

| Q             | A
| ------------- | ---
| Branch?       | current
| Tickets       | not declared
| License       | MIT

About tests, I did not find your test suite. So don’t hesitate to give me the right procedure.

EDIT : After Alan Poulain message, I run tests without any problem.

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
@@ -219,7 +219,9 @@ class SearchFilter extends AbstractContextAwareFilter implements SearchFilterInt
         }
 
         $queryBuilder->andWhere($queryBuilder->expr()->orX(...$ors));
-        array_walk($parameters, [$queryBuilder, 'setParameter']);
+        foreach ($parameters as $value => $name) {
+            $queryBuilder->setParameter($name, (string) $value);
+        }
     }
 
     /**


alanpoulain:
Hello,
You can find the current tests here: https://github.com/api-platform/core/blob/main/tests/Bridge/Doctrine/Orm/Filter/SearchFilterTest.php

hogren:
@alanpoulain  Thanks . No problem with tests :

**$ vendor/bin/simple-phpunit --stop-on-failure -vvv tests/Bridge/Doctrine/Orm/Filter/SearchFilterTest.php**
PHPUnit 9.4.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.0.6
Configuration: C:\Users\flhegron\Documents\programmation\php\api-platform\test2\vendor\api-platform\core\phpunit.xml.dist

Testing ApiPlatform\Core\Tests\Bridge\Doctrine\Orm\Filter\SearchFilterTest
................................................................. 65 / 71 ( 91%)
......                                                            71 / 71 (100%)

Time: 00:11.159, Memory: 78.00 MB

OK (71 tests, 211 assertions)

Legacy deprecation notices (102)

hogren:
@alanpoulain : Hello, just to be sure. Is everything ok for my PR or is there anything I forgot to do ?
Thanks !

Regards,

soyuka:
> The CONCAT function is not available before the 2012 version.

I do think that you should implement your own filter as we don't want to support older version no?

hogren:
@soyuka Sorry, this info was confusing. We have recent version, it’s not the problem.
In fact, the Doctrine Concat use the "+" method (maybe because the concat is not available in old versions). But it add quotes only if the php variable is a string.
With my correction, all concerned php variables are be casting in string.

hogren:
@alanpoulain , @soyuka Hello ! 
Do you have thought about my PR ?
Can I make any improvement ? Or do you need any additional information ?

Thank you !

alessandrofilira:
> @soyuka Sorry, this info was confusing. We have recent version, it’s not the problem. In fact, the Doctrine Concat use the "+" method (maybe because the concat is not available in old versions). But it add quotes only if the php variable is a string. With my correction, all concerned php variables are be casting in string.

Hi @hogren , I used the following workaround: I set filter strategy to `end` and when I call the API I append `%` in the end of value so doctrine convert it in string. So the call is `api/entity?code=34%`.

It's not really good solution but it works!

hogren:
@alessandrofilira Hello, sorry for the long time of silence. I have never seen your answer.
Thank you very much for the workaround, but the goal of my PR is to fix the problem. I already have another workaround.

PR #4353, opened 2021-07-09
feat: add a json schema builder for HAL (based on Hydra)

coudenysj:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT
| Doc PR        | api-platform/docs#... 

Added a JsonSchema builder for HAL specs, so the export commands render correct information.

```
php bin/console api:openapi:export
php bin/console api:openapi:export --yaml
```

PR #4357, opened 2021-07-14
Hal jsonschema

arnedesmedt:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

Hi, this is a replacing MR for https://github.com/api-platform/core/pull/4353
My colleague @coudenysj is now on holiday, so I've written the tests for his MR.

I've also needed to extend the SchemaFactoryInterface, because the decoration pattern is using the following order: Hal -> Hydra -> BaseSchemaFactory, we can't use the addDistinctFormat method only on the BaseSchemaFactory.


src/Hydra/JsonSchema/SchemaFactory.php
```diff
@@ -63,10 +62,7 @@ final class SchemaFactory implements SchemaFactoryInterface
     public function __construct(SchemaFactoryInterface $schemaFactory)
     {
         $this->schemaFactory = $schemaFactory;
-
-        if ($schemaFactory instanceof BaseSchemaFactory) {
```
alanpoulain, line 67: Why not keeping the check?

src/JsonSchema/SchemaFactoryInterface.php
```diff
@@ -26,4 +26,6 @@ interface SchemaFactoryInterface
      * Builds the JSON Schema document corresponding to the given PHP class.
      */
     public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, ?string $operationType = null, ?string $operationName = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema;
+
+    public function addDistinctFormat(string $format): void;
```
alanpoulain, line 30: Adding a method to an existing interface is a BC break, you cannot do it.

tests/Hal/JsonSchema/SchemaFactoryTest.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->assertEquals(str_replace('\\', '.', Dummy::class), $resultSchema->getRootDefinitionKey());
+    }
+
+    public function testHasRootDefinitionKeyBuildSchema(): void
+    {
+        $resultSchema = $this->schemaFactory->buildSchema(Dummy::class);
+        $definitions = $resultSchema->getDefinitions();
+        $rootDefinitionKey = $resultSchema->getRootDefinitionKey();
+
+        $this->assertEquals(str_replace('\\', '.', Dummy::class).'.jsonhal', $rootDefinitionKey);
```
alanpoulain, line 80: This line duplicates a previous test.

alanpoulain:
Hello,
It looks nice! You need to fix PHPStan too.

src/JsonSchema/SchemaFactoryInterface.php
arnedesmedt: The problem here is that we decorating the hydra schema factory with the hal schemafactory. So the schemafactory property in the hal schema factory will never be a "BaseSchemaFactory".

We've got HalSchemaFacotry contains the HydraSchemaFactory contains the BaseSchemaFactory. Since you only could add distinct formats in the baseSchemaFactory, it is impossible for the HalSchemaFactory to add the 'hal' as a distinct format.

Do you've got any tips to add the hal distinct format without extending the schemafactory interface?

alanpoulain: You can keep the public method in the `HydraSchemaFactory` but without adding it to the interface too.

tests/Hal/JsonSchema/SchemaFactoryTest.php
arnedesmedt: I removed the duplicate assert also from the hydra test.

src/Hydra/JsonSchema/SchemaFactory.php
arnedesmedt: I've moved the check now to the public method addDistinctFormat from the hydra schema factory.

src/JsonSchema/SchemaFactoryInterface.php
arnedesmedt: Good point, I did work now with the method_exists function.

arnedesmedt:
@alanpoulain fixed phpstan errors and ready for review. Thx in advance!

src/Hal/JsonSchema/SchemaFactory.php
```diff
@@ -0,0 +1,132 @@
...
+            $schema['required'] = [
+                '_links',
+                '_embedded',
+            ];
+
+            return $schema;
+        }
+
+        return $schema;
+    }
```
alanpoulain, line 131: Shouldn't you add a public `addDistinctFormat` method here too then?
In case this factory would be decorated too in the future?

arnedesmedt: Good point, will add it.

soyuka:
really good feature if you could rebase, thanks!

arnedesmedt:
Did the rebase.

arnedesmedt:
@alanpoulain can this be merged? Or do you expect something else to merge it?

alanpoulain (approved):

alanpoulain:
Thank you @arnedesmedt and @coudenysj!

coudenysj:
🥳

PR #4365, opened 2021-07-22
Fix: Type factory with nested objects for json schema (anyOf)

arnedesmedt:
| Q             | A
| ------------- | ---
| Branch?       | 2.6.5
| License       | MIT

When I did run tests, and I used the `assertMatchesResourceCollectionJsonSchema`, it failed on resources with nested objects where these objects are allowed to be null.

This is generated well for an open api schema, but not for json schema:

See: https://github.com/api-platform/core/blob/main/src/JsonSchema/TypeFactory.php#L165-L181

This should be fixed by this MR.

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -169,6 +169,15 @@ final class TypeFactory implements TypeFactoryInterface
         }
 
         if (\array_key_exists('$ref', $jsonSchema)) {
+            if ($schema && Schema::VERSION_JSON_SCHEMA === $schema->getVersion()) {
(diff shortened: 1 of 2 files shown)


soyuka:
we've created a new class for SchemaFactory on the `main` branch, can you update this accordingly?

arnedesmedt:
Yes, but I will need to do that in a separate MR, I guess? because this is a bugfix I want to merge in branch 2.6.

arnedesmedt:
Hi @soyuka Anything you're expecting from me to merge this one? I've just rebased the branch.

PR #4366, opened 2021-07-26
fix(doctrine): handle inverse side of OneToOne association in Doctrine search filter

carlobeltrame:
| Q             | A
| ------------- | ---
| Branch?       | main because @soyuka told me so, but ideally it should be for the current stable branch
| Tickets       | Same as api-platform/core#2049, but for SearchFilter instead of ExistsFilter
| License       | MIT

The same problem explained in api-platform/core#2049 also happens for the Doctrine SearchFilter. This PR fixes the SearchFilter, similarly to how api-platform/core#2051 fixed the ExistsFilter.

The issue can be reproduced by adding an Entity `Dummy` that has an owning one-to-one relation to another Entity `RelatedOwnedDummy`, and then adding a `SearchFilter` for that relation:

```php
/**
 * @ORM\Entity
 */
#[ApiResource]
#[ApiFilter(SearchFilter::class, properties: ['relatedOwnedDummy'])]
class Dummy {

    /**
     * @ORM\OneToOne(targetEntity="RelatedOwnedDummy", mappedBy="owner", cascade={"persist"})
     */
    public ?RelatedOwnedDummy $relatedOwnedDummy = null;

}
```

Without this fix, Doctrine complains:
"A single-valued association path expression to an inverse side is not supported in DQL queries. Instead of \"o.relatedOwnedDummy\" use an explicit join."

 - [x] Always add tests and ensure they pass.
 - [x] Never break backward compatibility (see https://symfony.com/bc).
 - [x] Bug fixes must be submitted against the current stable version branch.
 - ~Features and deprecations must be submitted against main branch.~
 - ~Legacy code removals go to the main branch.~
 - [ ] Update CHANGELOG.md file.
 - [x] Follow the [Conventional Commits specification](https://www.conventionalcommits.org/).

### DIFF
--- a/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Bridge/Doctrine/Orm/Filter/SearchFilter.php
@@ -143,7 +143,7 @@ class SearchFilter extends AbstractContextAwareFilter implements SearchFilterInt
 
         $associationAlias = $alias;
         $associationField = $field;
-        if ($metadata->isCollectionValuedAssociation($associationField)) {
+        if ($metadata->isCollectionValuedAssociation($associationField) || $metadata->isAssociationInverseSide($field)) {
             $associationAlias = QueryBuilderHelper::addJoinOnce($queryBuilder, $queryNameGenerator, $alias, $associationField);
             $associationField = $associationFieldIdentifier;
         }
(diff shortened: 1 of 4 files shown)


soyuka:
Please rebase against `main` I want to freeze 2.6 from now on.

carlobeltrame:
@soyuka done

alanpoulain (approved):

alanpoulain:
Thank you @carlobeltrame! I will merge `2.6` into `main` later.

Issue #4372, opened 2021-08-02
HAL serializer doesn't respect skip_null_values=false for relations


carlobeltrame:
**API Platform version(s) affected**: 2.6.5, and any other API platform version using Symfony 4.2 (since the [introduction](https://github.com/symfony/symfony/pull/28661) of the `skip_null_values` flag)

**Description**  
The HAL format normalizer does things a little differently from the rest of the normalizers, especially when it comes to normalizing relations. In [ItemNormalizer#populateRelation](https://github.com/api-platform/core/blob/main/src/Hal/Serializer/ItemNormalizer.php#L192), an `empty()` check determines whether to leave out a relation. This also affects optional ToOne relations which can have a null value. In the rest of the framework, the `skip_null_values` flag in the context determines what to do with such null relations. In HAL, this flag is not respected.

I'd be glad to provide a PR to fix this if requested.

**How to reproduce**  
Create two entities like the following:
```php
/**
 * @ORM\Entity
 */
#[ApiResource(normalizationContext: ['skip_null_values' => false])]
class Dummy {
  public ?RelatedDummy $relatedDummy = null;
}

/**
 * @ORM\Entity
 */
#[ApiResource]
class RelatedDummy {
}
```
Activate support for HAL in the config:
```yml
api_platform:
    formats:
        jsonhal: [ 'application/hal+json' ]
        jsonld: [ 'application/ld+json' ]
```
Create a `Dummy` instance in the database without any `RelatedDummy`, and fetch it using the format `Accept: application/hal+json`.
Expected result:
```json
{
  "relatedDummy": null
}
```
Actual result:
```json
{}
```

**Possible Solution**  
Use a check like the following:
```php
if (empty($attributeValue)) {
    $skipNullValues = $context[self::SKIP_NULL_VALUES] ?? $this->defaultContext[self::SKIP_NULL_VALUES] ?? false;
    if (null !== $attributeValue || $skipNullValues) {
        continue;
    }
}
```

dunglas:
The fix you propose looks good to me. Would you like to work on a patch?

carlobeltrame:
I have created a PR #4373 for demonstration, but I have run into problems. The [hyperschema](https://hyperschema.org/mediatypes/hal/) does not allow `null` entries in `_links` and `_embedded`. I have seen many places on the internet where people have been confused by what is and isn't allowed in HAL concerning null values:
[...]

PR #4373, opened 2021-08-03
fix: Respect skip_null_values in the HAL normalizer

carlobeltrame:
| Q             | A
| ------------- | ---
| Branch?       | fix --> 2.6
| Tickets       | #4372
| License       | MIT

RFC: This PR adds support for `skip_null_values: false` with relations to the HAL normalizer. This however deviates from the hyperschema, because now the entries in `_links` and `_embedded` may be null, which isn't really allowed in HAL.

Side note: I have tried to fix all tests that failed due to my changes, but PHP CS Fixer and PHPStan are complaining about some unrelated problems.

### DIFF
--- a/features/hal/collection.feature
+++ b/features/hal/collection.feature
@@ -68,7 +68,10 @@ Feature: HAL Collections support
             "_links": {
               "self": {
                 "href": "/dummies/1"
-              }
+              },
+              "relatedDummy": null,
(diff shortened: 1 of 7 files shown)


soyuka:
Please rebase against `main` I want to freeze 2.6 from now on.

Issue #4368, opened 2021-07-28
[GraphQL] [Syntax Error] line 0, col 65: Error: Unexpected 'WHERE'


Vilsafur:
**API Platform version(s) affected**: 2.6.5

**Description**  
I have a class (A) with a Many-To-Many relationship to another class (B).

A is a multi word name (like TestMultiWord).

I've also a name converter : `serializer.name_converter.camel_case_to_snake_case`

In GraphQL, accessing B from A works correctly, but I've got an error to accessing A from B.

The error given is: `[Syntax Error] line 0, col 65: Error: Unexpected 'WHERE'`

After investigation, Doctrine receives a malformed query: `SELECT o FROM App\Entity\A o WHERE o IN(SELECT WHERE id_a1.id = :id_p1) ORDER BY o.id ASC`

**How to reproduce**  

```php
<?php

namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiResource;
use App\Entity\B;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Serializer\Annotation\Groups;

/**
 * @ApiResource(
 *      normalizationContext={"groups"={"api:a:output"}},
 *      collectionOperations={
 *          "get"
 *      },
 *      itemOperations={
 *          "get"
 *      },
 *      graphql={
 *          "item_query" = {
 *              "normalization_context" = {
 *                  "groups" = {
 *                      "api:output:admin"
 *                  }
 *              }
 *          },
 *          "collection_query" = {
 *              "normalization_context" = {
 *                  "groups" = {
 *                      "api:output:admin"
 *                  }
 *              }
 *          },
 *          "create" = {
 *              "normalization_context" = {
 *                  "groups" = {
 *                      "collection_query"
 *                  }
 *              },
 *              "denormalization_context" = {
 *                  "groups" = {
 *                      "mutation"
 *                  }
 *              }
 *          }
 *      }
 * )
 * @ORM\Entity(repositoryClass="App\Repository\ARepository")
 */
class A
{

    /**
     * Liste des b associées à a
     *
     * @Groups({
     *      "api:output:admin"
     * })
     * @ORM\ManyToMany(
     *     targetEntity="App\Entity\B",
     *     mappedBy="a",
     *     fetch="EXTRA_LAZY"
     * )
     * @ORM\JoinTable(
     *     name="b_a",
[...]

PR #4379, opened 2021-08-12
Fix: graphQl / snake case with main rebase

Vilsafur:
<html>
<body>

Q | A
-- | --
Branch? | main
Tickets | #4368
License | MIT
Doc PR |  

<p style="box-sizing: border-box; margin-bottom: 0px !important; margin-top: 0px; color: rgb(173, 186, 199); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(34, 39, 46); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Fix the root property in GraphQL when a name converter is used.</p>
</body>
</html>

### DIFF
--- a/src/GraphQl/Resolver/Stage/ReadStage.php
+++ b/src/GraphQl/Resolver/Stage/ReadStage.php
@@ -95,7 +95,8 @@ final class ReadStage implements ReadStageInterface
         $source = $context['source'];
         /** @var ResolveInfo $info */
         $info = $context['info'];
-        if (isset($source[$rootProperty = $info->fieldName], $source[ItemNormalizer::ITEM_IDENTIFIERS_KEY], $source[ItemNormalizer::ITEM_RESOURCE_CLASS_KEY])) {
+        $rootProperty = $normalizationContext['denormalizedProperty'] ?? $info->fieldName;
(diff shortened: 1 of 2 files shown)


alanpoulain:
Hello, I don't like the idea of adding a `denormalizedProperty` to the serialization context, it doesn't make sense in GraphQL since we don't know where we are in the query.
Also could you provide a Behat test to understand exactly what the problem is?

Vilsafur:
Hello,

The problem was explain here : #4368.

In ReadStage, you use the fieldName found in the Entity (which is not converted) to pass in the getSubRessource method.

But this method use the given field in GraphQL (which is converted) to identify the subResource.

So you already have the denormalizePropertyName method in the SerializerContextBuilder, you just use it in normalization.
Why not use it in denormalizationContext ?

alanpoulain:
The normalization context is used for normalization, we don't want to use it for something else.

PR #4380, opened 2021-08-13
Throw UnexpectedValueException when value is not valid

lyrixx:
The current behavior is not logic

With the following JSON:
```
{
	"triggers": [
		{
			"type": "url",
			"options": {
				"url": "http://google.fr"
			}
		},
	"priority": 0,
	"description": "description"
}
```

If I put `true` in front of `triggers[0].url`, a `UnexpectedValueException` is raised. 
If I put `true` in front of `priority`, a `InvalidArgumentException` is raised. 

IMHO, the same exception should be thrown in similar situation.


soyuka:
thanks @lyrixx

PR #4381, opened 2021-08-16
PropertyMetadata becomes API Property

soyuka:


GuillaumeSTEIN:
Hi,
I would like to have some context to this PR.
Because I have a deprecation warning :
```
Deprecated: Optional parameter $resourceMetadataFactory declared before required parameter $decorated is implicitly treated as a required parameter in /[...]/vendor/api-platform/core/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php on line 39
```
But I don't know how to fix it. I do have `@ApiProperty` in my project, but I can't find any  `PropertyMetadata`

Thank you for your help

PR #4382, opened 2021-08-17
Add XML and YAML support for new Resource attribute

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/core/projects/10#card-66986300
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1473

- [x] resource/operations: defaults, requirements, options, status, schemes
- [x] properties: defaults, example, builtinTypes, schema
- [x] update doc (xml/yaml)
- [x] test that if a new attribute is added on ApiResource class, it must be added on XML (XmlAdapter)
- [x] test that if a new attribute is added on ApiResource class, it must be added on YAML (YamlAdapter)
- [x] fix tests


src/Metadata/Extractor/XmlExtractor.php
```diff
@@ -0,0 +1,170 @@
...
+            $resourceClass = $this->resolve((string) $resource['class']);
+
+            $this->resources[$resourceClass][] = [
+                'shortName' => $this->phpize($resource, 'shortName', 'string'),
+                'description' => $this->phpize($resource, 'description', 'string'),
+                'uriTemplate' => $this->phpize($resource, 'uriTemplate', 'string'),
+                'types' => $resource['types'] ?? null,
+                'operations' => $this->getOperations($resource, 'operation'),
+                'graphql' => $this->getOperations($resource, 'operation', true),
+                'attributes' => $this->getAttributes($resource, 'attribute') ?: null,
```
soyuka, line 53: attributes got renamed extraProperties

src/Metadata/Extractor/XmlExtractor.php
```diff
@@ -0,0 +1,170 @@
...
+        foreach ($resource->property as $property) {
+            $properties[(string) $property['name']] = [
+                'description' => $this->phpize($property, 'description', 'string'),
+                'readable' => $this->phpize($property, 'readable', 'bool'),
+                'writable' => $this->phpize($property, 'writable', 'bool'),
+                'readableLink' => $this->phpize($property, 'readableLink', 'bool'),
+                'writableLink' => $this->phpize($property, 'writableLink', 'bool'),
+                'required' => $this->phpize($property, 'required', 'bool'),
+                'identifier' => $this->phpize($property, 'identifier', 'bool'),
+                'iri' => $this->phpize($property, 'iri', 'string'),
```
soyuka, line 122: iri got renamed `types` and is now an array

src/Metadata/Extractor/XmlExtractor.php
```diff
@@ -0,0 +1,170 @@
...
+                'attributes' => $this->getAttributes($property, 'attribute'),
+                'subresource' => $property->subresource ? [
```
soyuka, line 124: subresource should not exist anymore

src/Metadata/Extractor/YamlExtractor.php
```diff
@@ -0,0 +1,160 @@
...
+                'description' => $this->phpize($propertyValues, 'description', 'string'),
+                'readable' => $this->phpize($propertyValues, 'readable', 'bool'),
+                'writable' => $this->phpize($propertyValues, 'writable', 'bool'),
+                'readableLink' => $this->phpize($propertyValues, 'readableLink', 'bool'),
+                'writableLink' => $this->phpize($propertyValues, 'writableLink', 'bool'),
+                'required' => $this->phpize($propertyValues, 'required', 'bool'),
+                'identifier' => $this->phpize($propertyValues, 'identifier', 'bool'),
+                'iri' => $this->phpize($propertyValues, 'iri', 'string'),
+                'attributes' => $propertyValues['attributes'] ?? [],
+                'subresource' => $propertyValues['subresource'] ?? null,
```
soyuka, line 127: same comments as above

src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public function create(string $resourceClass): ResourceMetadataCollection
+    {
+        $resourceMetadataCollection = new ResourceMetadataCollection($resourceClass);
+        if ($this->decorated) {
+            $resourceMetadataCollection = $this->decorated->create($resourceClass);
+        }
+
+        if (Comment::class === $resourceClass) {
+            dd(class_exists($resourceClass), $this->extractor->getResources());
```
soyuka, line 60: dd? :D

src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,170 @@
...
+        foreach (get_class_methods($resource) as $methodName) {
+            if (0 !== strpos($methodName, 'get')) {
+                continue;
+            }
+
+            if (!method_exists($operation, $methodName) || $operation->{$methodName}()) {
+                continue;
+            }
+
+            if (!$value = $resource->{$methodName}()) {
```
soyuka, line 161: `[]` `false` and `0` may be considered as valid values use `null ===`

src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
vincentchalamon: Work in progress :wink: 
The Behat tests and the console never go through this point, even if the Comment entity is enabled

src/Metadata/Extractor/AbstractPropertyExtractor.php
```diff
@@ -0,0 +1,127 @@
...
+
+                return (string) $resolved;
+            }
+
+            throw new \RuntimeException(sprintf('The container parameter "%s", used in the resource configuration value "%s", must be a string or numeric, but it is of type %s.', $parameter, $value, \gettype($resolved)));
+        }, $value);
+
+        return str_replace('%%', '%', $escapedValue);
+    }
+}
```
soyuka, line 127: why is this needed ?

src/Metadata/Extractor/XmlResourceExtractor.php
```diff
@@ -0,0 +1,412 @@
...
+
+        $uriVariables = [];
+        foreach ($resource->uriVariables->uriVariable as $data) {
+            $parameterName = (string) $data['parameterName'];
+            if (1 === \count($data->attributes())) {
+                $uriVariables[$parameterName] = $parameterName;
+                continue;
+            }
+
+            if (null !== ($property = $this->phpize($data, 'fromProperty', 'string'))) {
```
soyuka, line 164: does this needs to be `null !==` or can it be `!($property = $this->phpize($data, 'fromProperty', 'string'))` ?

src/Metadata/Extractor/XmlPropertyExtractor.php
```diff
@@ -0,0 +1,142 @@
...
+                continue;
+            }
+
+            $data[] = isset($value->values) ? $this->buildValues($value->values) : (string) $value;
+        }
+
+        return $data;
+    }
+
+    private function buildArrayValue(?\SimpleXMLElement $resource, string $key, $default = null)
```
soyuka, line 107: This method, as well as `phpize` and maybe others are used into XmlPropertyExtractor and XmlResourceExtractor should we use a trait with these to reduce the code base?

src/Metadata/Extractor/YamlResourceExtractor.php
```diff
@@ -0,0 +1,351 @@
...
+            'normalizationContext' => $this->buildAttribute($resource, 'normalizationContext'),
+            'denormalizationContext' => $this->buildAttribute($resource, 'denormalizationContext'),
+            'validationContext' => $this->buildAttribute($resource, 'validationContext'),
+            'filters' => $this->buildAttribute($resource, 'filters'),
+            'order' => $this->buildAttribute($resource, 'order'),
+            'extraProperties' => $this->buildAttribute($resource, 'extraProperties'),
+            'mercure' => $this->buildMercure($resource),
+            'messenger' => $this->buildMessenger($resource),
+        ];
+    }
```
soyuka, line 156: Do you think that we could use the `Metadata/ApiResource` class to build these? I'm thinking of something using reflection:

```
      $refl = new \ReflectionClass(ApiResource::class);
      foreach($refl->getMethods() as $method) {
        if(0 !== strpos($method->getName(), 'get')) {continue;}
        $this->phpize($resource, strtolower(str_replace('get', $method->getName())), $method->getReturnType()); //also replace `?` in the return type
      }
```

we'd probably need to check filters, mercure and messenger. Then `buildAttributes` could be inside `phpize` when the type is an array.

soyuka (commented):
huge work! Few comments.

src/Metadata/Extractor/AbstractPropertyExtractor.php
vincentchalamon: It's from the original [AbstractExtractor](https://github.com/api-platform/core/blob/main/src/Core/Metadata/Extractor/AbstractExtractor.php)

src/Metadata/Extractor/YamlResourceExtractor.php
vincentchalamon: I will still need to filter which property is in the `buildBase` method or not. Those properties can be declared at the resource level, but also overriden by operation, and not all properties can be declared on both sides. That's why I had to filter them here

soyuka (approved):

PR #4392, opened 2021-08-23
feat: add support for security_post_validation

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       |
| License       | MIT
| Doc PR        |

If an object (project) has a relation to a parent (organization), then we want to run the security after the payload is validated, to ensure the project organization exist.

Example:

```
POST /project
{
}
```

Here, I don't want a 403, but a 400 saying the `project` is missing.

ATM, it's doable, with it's a bit an hack: 
```
*              "security_post_denormalize"="not object.getOrganization() or is_granted('PROJECT_ADD', object.getOrganization())",
```

But it's even harder when using typed property (not nullable) because `isset` does not exist in the expression language...

So, IMHO, it's better to have an option to run the security after the validation


dunglas (approved):
lgtm

src/Core/Annotation/ApiResource.php
```diff
@@ -218,6 +222,8 @@ public function __construct(
         ?string $securityMessage = null,
         ?string $securityPostDenormalize = null,
         ?string $securityPostDenormalizeMessage = null,
+        ?string $securityPostValidation = null,
+        ?string $securityPostValidationMessage = null,
```
soyuka, line 226: please add this to `src/Metadata/ApiResource`

lyrixx: ✅

src/GraphQl/Resolver/Stage/SecurityPostValidationStage.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $resourceMetadataCollection = $this->resourceMetadataCollectionFactory->create($resourceClass);
+        $operation = $resourceMetadataCollection->getGraphQlOperation($operationName);
+        $isGranted = $operation->getSecurityPostValidation();
+
+        if (null !== $isGranted && null === $this->resourceAccessChecker) {
+            throw new \LogicException('Cannot check security expression when SecurityBundle is not installed. Try running "composer require symfony/security-bundle".');
+        }
+
+        if (null === $isGranted || $this->resourceAccessChecker->isGranted($resourceClass, (string) $isGranted, $context['extra_variables'])) {
```
norkunas, line 52: `getSecurityPostValidation` returns string|null
```suggestion
        if (null === $isGranted || $this->resourceAccessChecker->isGranted($resourceClass, $isGranted, $context['extra_variables'])) {
```

lyrixx: But it it's null, then the code goes in the very fist condition, so here we know it's a string, isn't ?

norkunas: oh right, misread first condition :)

norkunas: > so here we know it's a string

so cast become useless then :)

lyrixx: ok, I read your diff in the wrong way. Sorry :)

alanpoulain:
Thank you @lyrixx!

PR #4401, opened 2021-08-31
feat(debug): add a debug resource command

samuelbigard:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Type | Feature
| Tickets       | 
| License       | MIT

### DIFF
--- a/src/Bridge/Symfony/Bundle/Command/DebugResourceCommand.php
+++ b/src/Bridge/Symfony/Bundle/Command/DebugResourceCommand.php
@@ -1,0 +1,121 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Bridge\Symfony\Bundle\Command;
+
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\VarDumper\Cloner\ClonerInterface;
+
+final class DebugResourceCommand extends Command
+{
+    protected static $defaultName = 'debug:api-resource';
+
+    private $resourceMetadataCollectionFactory;
+    private $cloner;
+    private $dumper;
+
+    public function __construct(ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory, ClonerInterface $cloner, $dumper)
+    {
+        parent::__construct();
+        $this->resourceMetadataCollectionFactory = $resourceMetadataCollectionFactory;
+        $this->cloner = $cloner;
+        $this->dumper = $dumper;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure(): void
+    {
+        $this
+            ->setDescription('Debug API Platform resources')
+            ->addArgument('class', InputArgument::REQUIRED, 'The class you want to debug');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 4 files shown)


soyuka (approved):

src/Bridge/Symfony/Bundle/Command/DebugResourceCommand.php
```diff
@@ -0,0 +1,114 @@
...
+        );
+
+        $answerOperation = $helper->ask($input, $output, $questionOperation);
+        if ('Debug the resource itself' === $answerOperation) {
+            dump($selectedResource);
+
+            return Command::SUCCESS;
+        }
+
+        dump($resourceCollection->getOperation($answerOperation));
```
dunglas, line 110: This function is usually not available in prod. Maybe should we use the Var Dumper component directly instead?

soyuka: We declared this in the `debug.xml` file which is loaded only in debug mode but I agree. It'll be also easier to test (and tests are missing).

src/Bridge/Symfony/Bundle/Command/DebugResourceCommand.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $answerOperation = $helper->ask($input, $output, $questionOperation);
+        if ('Debug the resource itself' === $answerOperation) {
+            $this->dumper->dump($this->cloner->cloneVar($selectedResource));
+            $output->writeln('Successfully dumped the selected resource');
+
+            return Command::SUCCESS;
+        }
+
+        $this->dumper->dump($this->cloner->cloneVar($resourceCollection->getOperation($answerOperation)));
```
dunglas, line 117: Why do we need to clone the object?

src/Bridge/Symfony/Bundle/Command/DebugResourceCommand.php
```diff
@@ -0,0 +1,122 @@
...
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\VarDumper\Cloner\ClonerInterface;
+use Symfony\Component\VarDumper\Dumper\CliDumper;
+
+class DebugResourceCommand extends Command
```
dunglas, line 25:
```suggestion
final class DebugResourceCommand extends Command
```

dunglas (changes requested):

src/Bridge/Symfony/Bundle/Command/DebugResourceCommand.php
samuelbigard: Because `dump` function needs `Symfony\Component\VarDumper\Cloner\Data` as argument

tests/Bridge/Symfony/Bundle/Command/DebugResourceCommandTest.php
```diff
@@ -0,0 +1,95 @@
...
+        $application->add(new DebugResourceCommand(new AttributesResourceMetadataCollectionFactory(), new VarCloner(), $cliDumper ?? new CliDumper()));
+
+        $command = $application->find('debug:api-resource');
+
+        return new CommandTester($command);
+    }
+
+    public function testDebugResource()
+    {
+        $varDumper = $this->prophesize(CliDumper::class);
```
dunglas, line 46: I think that it's not necessary to mock this class (don't mock what you don't own). You may just instantiate a `CliDumper` dumping to `/dev/null`.

This should do the trick.

```suggestion
        $varDumper = new CliDumper('Windows' === PHP_OS_FAMILY ? 'NUL' : '/dev/null');
```

soyuka: But we want to test that dump is called isn't that alright?

dunglas: We should at least mock the interface instead of a concretion in this case.

dunglas: Another option could be to set `php://memory` as output and to assert that something has been written in it.

dunglas (approved):

soyuka:
Thanks @samuelbigard !

PR #4406, opened 2021-09-01
fix: indexing relations resources IRIs in normalization

cedvan:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/core/issues/3168
| License       | MIT

Hi,

Currently all the IRIs of the resource relations being normalized are indexed in the cache.

However only relationship IRIs that are normalized and nested should be indexed.

This correction allows to considerably lighten the `Cache-Tags` header in the case of a fullREST hypermedia API and thus repair the maximum header size problem, especially with Varnish

Maybe must apply this fix to current 2.6 ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.7.0
 
+* Fix indexing relations resources IRIs in serializer
 * Security: **BC** Fix `ApiProperty` `security` attribute expression being passed a class string for the `object` variable on updates/creates - null is now passed instead if the object is not available (#4184)
 * Security: `ApiProperty` now supports a `security_post_denormalize` attribute, which provides access to the `object` variable for the object being updated/created and `previous_object` for the object before it was updated (#4184)
 * Maker: Add `make:data-provider` and `make :data-persister` commands to generate a data provider / persister (#3850)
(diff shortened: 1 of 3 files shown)


dunglas:
Even if they aren't nested but only referenced, we must invalidate the document containing the reference to handle because the related resource may be deleted.

cedvan:
I understand, but orm relation prevents deleting an entity with a referenced relation, right?

cedvan:
So to delete entity A, I have to update entities X, Y, Z ... which references it to delete the relation with A and then I can delete my entity A. These updates will purge the cache of these entities X, Y, Z ...

cedvan:
What case do I forget that can possibly be different?

cedvan:
Maybe this problem can be effective in non-owner reverse relationships, right?

cedvan:
So maybe we just need to add IRIs for nonn-owner related entities?

cedvan:
Or maybe just change the purge system and add a specific purge when deleting an A entity on all documents from X, Y, Z ... entities that have at least one inverse relationship to A.

Deletion is rare enough to tolerate a complete purge of an entity's documents. Especially since we use a lot of solf delete today

dunglas:
This part isn't coupled with the ORM (you can use it with any data provider, even if you don't use an ORM). By default, we prefer to be as exact as possible and to let users add "risky" (not generic) optimizations in their own code base by replacing or decorating the built-in services.

cedvan:
D'accord. Une autre solution pour résoudre le problème de taille d'en-tête ? J'ai une grande API hypermédia fullREST, et la taille de l'en-tête est vraiment trop grande :/.

dunglas:
Did you try this https://api-platform.com/docs/extra/troubleshooting/#upstream-sent-too-big-header-while-reading-response-header-from-upstream-nginx-502-error?

cedvan:
Yes, but after receive "Error 503 Backend fetch failed" instead of "502"

cedvan:
Ok, I solve 503 problem by add `-p http_resp_size=__k -p http_resp_hdr_len=__k` to varnish command. Maybe necessary add this to documentation ?

PR #4407, opened 2021-09-01
feat: add a collection state provider for Doctrine ORM

dunglas:
Signed-off-by: Kévin Dunglas <dunglas@gmail.com>

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

We have a problem with the legacy wrapper for data providers and persisters: the native providers and persisters must run in the chain after the custom ones provided by the user. However, currently we have a single bridge `LegacyDataProviderState`, that wraps the legacy chain. This means that legacy data providers and persisters always run after the custom ones regardless of their priority.

A possible solution would be to wrap all providers and persisters instead of the chain, and to add to copy the initial priority.


composer.json
```diff
@@ -138,5 +141,6 @@
         "symfony": {
             "require": "^3.4 || ^4.4 || ^5.1"
         }
-    }
+    },
+    "minimum-stability": "dev"
```
soyuka: do not commit this

Issue #4411, opened 2021-09-02
Make sure every ApiResource/ApiProperty property is nullable


soyuka:
The new `ApiPlatform/Metadata/ApiResource` and `ApiPlatform/Metadata/ApiProperty` properties should all be nullable and have null as default which is not the case right now. This makes decoration easier, when a factory runs it should skip nullable values. 

For example: https://github.com/api-platform/core/blob/main/src/Metadata/ApiResource.php#L162.

Zowac:
I'd like to get into it if possible :)

PR #4422, opened 2021-09-06
Allow null attributes in ApiResource, ApiProperty and Operations

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4411 
| License       | MIT

ApiProperty is already ok


dunglas (approved):

mtarld:
We need to handle as well 
- [ ] `ApiPlatform\Metadata\Operation`
- [ ] `ApiPlatform\Metadata\Get`
- [ ] `ApiPlatform\Metadata\Post`
- [ ] `ApiPlatform\Metadata\Put`
- [ ] `ApiPlatform\Metadata\Patch`
- [ ] `ApiPlatform\Metadata\Delete`
- [ ] `ApiPlatform\Metadata\GraphQl\Operation`
- [ ]  `ApiPlatform\Metadata\GraphQl\Query`
- [ ] `ApiPlatform\Metadata\GraphQl\QueryCollection`
- [ ] `ApiPlatform\Metadata\GraphQl\Mutation`
- [ ] `ApiPlatform\Metadata\GraphQl\Subscription`

azjezz (approved):

dunglas (changes requested):
Actually, there is a design issue. The getters can return `null`, but it's not possible to pass `null` to the withers (as in https://github.com/api-platform/core/blob/main/src/Core/Metadata/Resource/ResourceMetadata.php).

`null` is a special value saying "this property hasn't been defined yet". Once it's define, it cannot be reset.

dunglas:
Also, I think that all default values should be `null`, or it's almost useless to make them nullable.

soyuka:
superseeded by https://github.com/api-platform/core/pull/4449 thanks @mtarld  !

PR #4426, opened 2021-09-07
Feat(doctrine): create new writeListener which only use processor

samuelbigard:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Type       |  Feature
| License       | MIT

### DIFF
--- a/src/Symfony/EventListener/WriteListener.php
+++ b/src/Symfony/EventListener/WriteListener.php
@@ -11,15 +11,17 @@
 
 declare(strict_types=1);
 
-namespace ApiPlatform\Core\EventListener;
+namespace ApiPlatform\Symfony\EventListener;
 
 use ApiPlatform\Api\IriConverterInterface;
 use ApiPlatform\Core\Api\ResourceClassResolverInterface;
-use ApiPlatform\Core\DataPersister\DataPersisterInterface;
-use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ToggleableOperationAttributeTrait;
 use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use ApiPlatform\Core\Util\ResourceClassInfoTrait;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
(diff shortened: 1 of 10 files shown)


src/Core/EventListener/WriteListener.php
```diff
@@ -65,6 +65,7 @@ public function __construct($dataPersister, $iriConverter = null, $resourceMetad
 
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->resourceClassResolver = $resourceClassResolver;
+        trigger_deprecation('api-platform/core', '2.7', sprintf('The listener %s is deprecated and will be removed in 3.0', __CLASS__));
```
dunglas, line 68: You should also add a `@deprecated` annotation in the PHPDoc of the class.

src/Symfony/EventListener/WriteListener.php
```diff
@@ -0,0 +1,126 @@
...
+        ) {
+            return;
+        }
+
+        if (!$operation || ($operation->getExtraProperties()['is_legacy_resource_metadata'] ?? false)) {
+            return;
+        }
+
+        if (
+            !$this->resourceMetadataFactory instanceof ResourceMetadataCollectionFactoryInterface &&
```
dunglas, line 80: This type assertion always returns `true` isn't it?

src/Symfony/EventListener/WriteListener.php
```diff
@@ -0,0 +1,126 @@
...
+            (!$attributes['persist']
+                || $this->isOperationAttributeDisabled($attributes, self::OPERATION_ATTRIBUTE_KEY)
+            )) {
+            return;
+        }
+
+        $context = ['operation' => $operation, 'legacy_attributes' => $attributes + ['identifiers' => $operation->getIdentifiers(), 'has_composite_identifier' => $operation->getCompositeIdentifier()]];
+
+        if ($this->processor instanceof ProcessorInterface && !$this->processor->supports($controllerResult, $attributes['identifiers'], $operation->getName(), $context)) {
```
dunglas, line 89: Same here, `$this->processor instanceof ProcessorInterface` is always `true`.

dunglas (changes requested):

soyuka:
Thanks @samuelbigard

PR #4430, opened 2021-09-07
doc(adr): Resource => ApiResource

soyuka:
Update the ADR document with ApiResource instead of Resource as its a PHP reserved word.

### DIFF
--- a/docs/adr/0002-resource-definition.md
+++ b/docs/adr/0002-resource-definition.md
@@ -1,6 +1,6 @@
 # Resource definition
 
-* Status: proposed
+* Status: accepted
 * Deciders: @dunglas @soyuka @vincentchalamon @GregoireHebert
 
 ## Context and Problem Statement
@@ -151,13 +151,19 @@ class User {}
     </tr>
 </table>
 
-To link these operations with identifiers, refer to [Resource Identifiers decision record](./0001-resource-identifiers), for example:
+To link these operations with identifiers, refer to [URI Variables decision record](./0003-uri-variables), for example:
 
 ```php
 <?php
 use Company;
 
-#[Get("/companies/{companyId}/users/{id}", [identifiers=["companyId" => [Company::class, "id"], "id" => [User::class, "id"]]])]
+#[Get(
+    uriTemplate: "/companies/{companyId}/users/{id}", 
+    uriVariables: [
+        "companyId" => ["class" => Company::class, "identifiers" => ["id"], "property" => "user"], 
+        "id" => ["class" => User::class, "identifiers" => ["id"]]
+    ]
+)]
 class User {
   #[ApiProperty(identifier=true)]
   public $id;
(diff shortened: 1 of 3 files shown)


docs/adr/0002-resource-definition.md
```diff
@@ -24,17 +24,17 @@ In API Platform, this resource identifier is also named [IRI (Internationalized
 ```php
 <?php
 
-#[Resource]
+#[ApiResource]
 class Users
 {
-    #[ApiProperty(iri="hydra:member")]
+    #[ApiProperty(types=["hydra:member"])]
```
dunglas, line 30:
```suggestion
    #[ApiProperty(types="hydra:member")]
```

(This works, I already merged the patch, arrays are still supported too)

docs/adr/0002-resource-definition.md
```diff
@@ -24,17 +24,17 @@ In API Platform, this resource identifier is also named [IRI (Internationalized
...
     public User[] $member = [];
```
dunglas, line 31:
```suggestion
    public irable $member = [];
```

We should update the PHPDoc accordingly.

docs/adr/0003-uri-variables.md
```diff
@@ -0,0 +1,71 @@
...
+  public Company $company;
+}
+```
+
+To make this work, API Platform needs to know what property of the class User has the link to Company. 
+
+## Considered Options
+
+* Keeping the tuple but adding a third value which is the property referenced by the parent class.
+* Using a map to allow granularity configuration of the URI variables
```
dunglas, line 30:
```suggestion
* Using a map to allow granularity configuration of the URI variables.
```

docs/adr/0003-uri-variables.md
```diff
@@ -0,0 +1,71 @@
...
+    ]
+]
+```
+
+Where `uriTemplate` keys are the URI template's variable names. Its value is a map where:
+
+- `class` is the PHP FQDN to the class this value belongs to
+- `identifiers` are the properties of the class to which we map the URI variable
+- `composite_identifier` is used to match a single variable to multiple identifiers (`ida=1;idb=2` to `class::ida` and `class::idb`)
+- `property` represents the property that has a link to the next identifier
```
dunglas, line 56:
```suggestion
- `property` represents the property that has a link to the next identifier

As of PHP 8.1, PHP will support [nested attributes](https://wiki.php.net/rfc/new_in_initializers). We'll introduce a proper class as an alternative to the associative array when PHP 8.1 will be released.
```

dunglas (changes requested):

docs/adr/0002-resource-definition.md
norkunas: maybe `iterable` :slightly_smiling_face:

dunglas: Fixed

PR #4443, opened 2021-09-14
feat(datefilter): use static consts

zsapozhnikov:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Use late static binding in the case of inheritance from DateFilter to override parameters' constants

### DIFF
--- a/src/Core/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
+++ b/src/Core/Bridge/Doctrine/Common/Filter/DateFilterTrait.php
@@ -44,10 +44,10 @@ trait DateFilterTrait
                 continue;
             }
 
-            $description += $this->getFilterDescription($property, self::PARAMETER_BEFORE);
-            $description += $this->getFilterDescription($property, self::PARAMETER_STRICTLY_BEFORE);
-            $description += $this->getFilterDescription($property, self::PARAMETER_AFTER);
-            $description += $this->getFilterDescription($property, self::PARAMETER_STRICTLY_AFTER);
+            $description += $this->getFilterDescription($property, static::PARAMETER_BEFORE);
+            $description += $this->getFilterDescription($property, static::PARAMETER_STRICTLY_BEFORE);
+            $description += $this->getFilterDescription($property, static::PARAMETER_AFTER);
+            $description += $this->getFilterDescription($property, static::PARAMETER_STRICTLY_AFTER);
         }
 
         return $description;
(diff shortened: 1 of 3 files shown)


dunglas:
Hi and thanks for contributing!

Could you detail your use case? This will probably have an impact regarding performance, and most of our classes are final anyway.

zsapozhnikov:
Hi, I want to use another parameters like for existing project
```
public const PARAMETER_BEFORE = 'lte';
public const PARAMETER_STRICTLY_BEFORE = 'lt';
public const PARAMETER_AFTER = 'gte';
public const PARAMETER_STRICTLY_AFTER = 'gt';
```
so I extend base class
```
use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\DateFilter as ApiPlatformDateFilter;

class DateFilter extends ApiPlatformDateFilter
{
    public const PARAMETER_BEFORE = 'lte';
    public const PARAMETER_STRICTLY_BEFORE = 'lt';
    public const PARAMETER_AFTER = 'gte';
    public const PARAMETER_STRICTLY_AFTER = 'gt';
}
```

dunglas:
But this class is marked `@final`. You shouldn't extend it but use composition (a decorator) instead.

zsapozhnikov:
sorry, I didn't see it when was on v2.6.5
> use composition (a decorator) instead

it's not clear for me how to use it here

zsapozhnikov:
DateFilter is marked abtract `<service id="api_platform.doctrine.orm.date_filter" class="ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\DateFilter" public="false" abstract="true">`

dunglas (approved):

dunglas:
Why not then

zsapozhnikov:
@dunglas `codecov/patch ` fails or it is not necessary? Anything I need to do or just waiting for merge?

alanpoulain:
The `DateFilter` you are using (`ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\DateFilter`) is marked final though.
I don't think we should encourage inheritance, so I'm :-1: for this.

zsapozhnikov:
@alanpoulain maybe remove `final` as there is no other option to reuse it

alanpoulain:
As @dunglas said, we recommend to decorate our services if you want to change their behavior.

zsapozhnikov:
@alanpoulain I've answered him here
https://github.com/api-platform/core/pull/4443#issuecomment-918911987
the class is abstract. So we can't extend or decorate it for now. Maybe we can remove `final` annotation?

PR #4452, opened 2021-09-16
fix: TraceableChainItemDataProvider bug

kilip:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

This will fix ```TraceableChainItemDataProvider::getItem``` logic, because when the first ```dataProvider``` executed and the result is null, the ```$match``` condition will always be true and the loop will be stopped. This will cause  the other ```dataProvider::getItem()``` will never be executed.

### DIFF
--- a/src/Core/Bridge/Symfony/Bundle/DataProvider/TraceableChainItemDataProvider.php
+++ b/src/Core/Bridge/Symfony/Bundle/DataProvider/TraceableChainItemDataProvider.php
@@ -75,7 +75,9 @@ final class TraceableChainItemDataProvider implements ItemDataProviderInterface,
                 }
 
                 $result = $dataProvider->getItem($resourceClass, $identifier, $operationName, $context);
-                $this->providersResponse[\get_class($dataProvider)] = $match = true;
+                if(null !== $result){
+                    $this->providersResponse[\get_class($dataProvider)] = $match = true;
+                }
             } catch (ResourceClassNotSupportedException $e) {
                 @trigger_error(sprintf('Throwing a "%s" is deprecated in favor of implementing "%s"', \get_class($e), RestrictedDataProviderInterface::class), \E_USER_DEPRECATED);
                 continue;


alanpoulain:
Hello,
A `null` result is considered a result anyway. If you want your data provider to *not* be executed, you need to tell it in the `supports` method.
I'm closing this PR. Feel free to ask questions if needed.

kilip:
Hello @alanpoulain 

Actually I need data provider **to be executed**. And I already define that this in [supports function](https://github.com/kilip/opensid/blob/3561dce7c3041ddf559370704fb6a3d8d54c0c24/src/Application/DataProvider/ConfigItemDataProvider.php#L30-L33)

If you take a closer look at this TraceableChainItemDataProvider class:
https://github.com/api-platform/core/blob/4e22a289e474db454480020ff37d20472668c11c/src/Bridge/Symfony/Bundle/DataProvider/TraceableChainItemDataProvider.php#L48-L84

After the loop reach **Line 76**, regardless the result the loop will be stopped. In debug I have found that after first data provider proceed (ApiPlatform\Core\Bridge\Doctrine\Orm\ItemDataProvider) the loop will stopped, and the other data provider will not be executed. In order to fix this, I need to move my configuration to [```config/services.yaml```](https://github.com/kilip/opensid/blob/3561dce7c3041ddf559370704fb6a3d8d54c0c24/config/services.yaml#L13-L18).

In the attached image you see debug in my editor that I have 3 data provider. But because of that **Line 76**, the other data provider will never be executed.
Screenshot from 2021-09-24 22-24-21

kilip:
Hello @alanpoulain 

Actually I need data provider **to be executed**. And I already define that this in [supports function](https://github.com/kilip/opensid/blob/3561dce7c3041ddf559370704fb6a3d8d54c0c24/src/Application/DataProvider/ConfigItemDataProvider.php#L30-L33)

If you take a closer look at this TraceableChainItemDataProvider class:
https://github.com/api-platform/core/blob/4e22a289e474db454480020ff37d20472668c11c/src/Bridge/Symfony/Bundle/DataProvider/TraceableChainItemDataProvider.php#L48-L84

After the loop reach **Line 76**, regardless the result the loop will be stopped. In debug I have found that after first data provider proceed (ApiPlatform\Core\Bridge\Doctrine\Orm\ItemDataProvider) the loop will stopped, and the other data provider will not be executed. In order to fix this, I need to move my configuration to [```config/services.yaml```](https://github.com/kilip/opensid/blob/3561dce7c3041ddf559370704fb6a3d8d54c0c24/config/services.yaml#L13-L18).

In the attached image you see debug in my editor that I have 3 data provider. But because of that **Line 76**, the other data provider will never be executed.
Screenshot from 2021-09-24 22-24-21

alanpoulain:
Yes, that's exactly how it is wanted. You can also see the "normal" item data provider working like this:
https://github.com/api-platform/core/blob/a9cc84d8dc7a2592d2657e9a0fc92e1097f3a68e/src/DataProvider/ChainItemDataProvider.php#L62
I don't see why it should work differently, maybe you can explain your use case?

kilip:
@alanpoulain 

I want to store all my service configuration in ```src/*/Resources/config.yml``` like [this config](https://github.com/kilip/opensid/blob/79429775c027f09c7fe23005c7f1f1663551eeed/src/Application/Resources/config/application.yaml#L9-L14)

But because of this line those item data provider never executed:
https://github.com/api-platform/core/blob/4e22a289e474db454480020ff37d20472668c11c/src/Bridge/Symfony/Bundle/DataProvider/TraceableChainItemDataProvider.php#L76

So I fix those line like [this](https://github.com/api-platform/core/pull/4452/files). After changing those line my item data provider working as expected, which also a proof that my item data provider in this [service config](https://github.com/kilip/opensid/blob/79429775c027f09c7fe23005c7f1f1663551eeed/src/Application/Resources/config/application.yaml#L9-L14) actually loaded.

In debug I have found that ```TraceableChainItemDataProvider.php``` contains data providers:
1. ```ApiPlatform\Core\Bridge\Doctrine\Orm\ItemDataProvider```
2. ```OpenSID\Application\DataProvider\ConfigItemDataProvider``` (expected data provider to be executed).

Since data providers processing in FIFO, if we don't change line 76, my ```ConfigItemDataProvider::getItem()``` will never be called.

alanpoulain:
You can control the order the data providers are executed by using [decoration priority](https://symfony.com/doc/current/service_container/service_decoration.html#decoration-priority).

PR #4476, opened 2021-09-29
docs: improve uri variables ADR

soyuka:


dunglas (approved):

docs/adr/0003-uri-variables.md
```diff
@@ -31,40 +31,329 @@ To make this work, API Platform needs to know what property of the class User ha
...
 ```
 
-Where `uriTemplate` keys are the URI template's variable names. Its value is a map where:
+Where `uriVariables` keys are the URI template's variable names. Its value is a map where:
 
-- `class` is the PHP FQDN to the class this value belongs to
-- `identifiers` are the properties of the class to which we map the URI variable
-- `composite_identifier` is used to match a single variable to multiple identifiers (`ida=1;idb=2` to `class::ida` and `class::idb`)
-- `property` represents the property that has a link to the next identifier
+- `targetClass` is the PHP FQDN to the class this value belongs to
```
chalasr, line 54:
```suggestion
- `targetClass` is the PHP FQCN of the class this value belongs to
```

soyuka:
superseeded by https://github.com/api-platform/core/pull/4481

docs/adr/0003-uri-variables.md
soyuka: thanks added

PR #4487, opened 2021-10-04
chore: add return type in Configuration.php to support symfony 6

eerison:
| Q             | A
| ------------- | ---
| Branch?       | 2.6 |
| License       | MIT|
| Doc PR.      | n/a|

deprecation message:

```
Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.
```

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -48,7 +48,7 @@ final class Configuration implements ConfigurationInterface
     /**
      * {@inheritdoc}
      */
-    public function getConfigTreeBuilder()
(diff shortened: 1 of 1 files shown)


alanpoulain:
Hello,
We will not do this for API Platform 2.6, only 2.7.
We will not add the return type, only the explicit `@return` annotation to be compatible with old PHP versions.
A lot of methods are impacted, it would be nice to do it everywhere.
I'm closing this in the meantime, thanks anyway.

Issue #4488, opened 2021-10-05
OpenAPI/Swagger does not reflect route prefixes


TheCrealm:
**API Platform version(s) affected**: 2.7.x-dev

**Description**  
I use an `/api` route prefix for api since i started using api platform. On 2.7-x-dev swagger just drops this prefix.

`bin/console debug:router`: 
![image](https://user-images.githubusercontent.com/8281157/135968732-1aeb54c3-0ae2-4dee-9435-b874504cb00d.png)

`Swagger`:
![image](https://user-images.githubusercontent.com/8281157/135968658-b4529653-f8b5-465f-8d4b-7fa02fe53343.png)

**How to reproduce**  
Just use the following route config for api platform:
```
api_platform:
    resource: .
    type: api_platform
    prefix: /api
```

**Possible Solution**  
Somehow utilize the OperationPathResolver like before.

Old getPath function: https://github.com/api-platform/core/blob/2.6/src/OpenApi/Factory/OpenApiFactory.php#L343
New getPath function: https://github.com/api-platform/core/blob/main/src/OpenApi/Factory/OpenApiFactory.php#L327

**Additional Context**  
As i tried its more effort as just use it on the new system (or i am dumb rn). I currently did not figured out how to do a proper fix.

alanpoulain:
Hello,

I've made a fix to manage it correctly, thanks for the report: https://github.com/api-platform/core/pull/4489.

However, using `prefix` is not doable anymore with the new metadata system (API Platform 2.7).

Please use:
```yaml
# api/config/packages/api_platform.yaml
api_platform:
    defaults:
        route_prefix: '/prefix'
```
See also:
- https://api-platform.com/docs/core/configuration/#global-resources-defaults
- https://api-platform.com/docs/core/operations/#prefixing-all-routes-of-all-operations

alanpoulain:
We need to make sure `getRoutePrefix` is not forgotten when calling `getUriTemplate` but I think it's better indeed to do it like this.

alanpoulain:
I'm closing this issue then.

PR #4490, opened 2021-10-07
Fix: Make sure the route prefix is only applied once

GlitchedAxiom:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | #4488 
| License       | MIT
| Doc PR        |

Fix doubled route prefix by only applying the prefix in the ApiLoader.

### DIFF
--- a/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
@@ -90,8 +90,7 @@ final class UriTemplateResourceMetadataCollectionFactory implements ResourceMeta
 
     private function generateUriTemplate(Operation $operation): string
     {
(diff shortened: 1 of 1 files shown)


alanpoulain:
Thanks!

alanpoulain:
IMO the current way is better.
Like @TheCrealm is saying, with the other way, if the URI template is generated with a route prefix, we would have:

_ResourceMetadataCollection_:
_routePrefix_: /prefix
_uriTemplate_: /prefix/dummies

And if the URI template is user-defined:

_ResourceMetadataCollection_:
_routePrefix_: /prefix
_uriTemplate_: /custom-dummies

It doesn't seem right to me, and I'm not sure erasing the routePrefix after being applied to the URI template is a good idea. Metadata should describe what is written in the model, and should not have much logic.

Another possibility would be to have a method in `ResourceMetadataCollection` to concatenate the route prefix and the URI template like `getPrefixedUriTemplate` or `getCompleteUriTemplate`.

Issue #4509, opened 2021-10-11
Failing tests locally without symfony/intl


Chris53897:
**API Platform version(s) affected**: main

The github actions add these packages in CI.
https://github.com/api-platform/core/blob/main/.github/workflows/ci.yml#L73

If you run it locally, you do not have this installed. one test fails.
Is there a reason "symfony/intl": "^3.4 || ^4.4 || ^5.1", is not in `require-dev` ?

**Description**  
1) ApiPlatform\Core\Tests\Bridge\Symfony\Validator\Metadata\Property\ValidatorPropertyMetadataFactoryTest::testCreateWithPropertyWithValidationConstraints
Symfony\Component\Validator\Exception\LogicException: The Intl component is required to use the Bic constraint. Try running "composer require symfony/intl".

/Users/chris/PhpstormProjects/api-platform/core/vendor/symfony/validator/Constraints/Bic.php:58
/Users/chris/PhpstormProjects/api-platform/core/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:977
/Users/chris/PhpstormProjects/api-platform/core/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:719
/Users/chris/PhpstormProjects/api-platform/core/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:376
/Users/chris/PhpstormProjects/api-platform/core/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationReader.php:178
/Users/chris/PhpstormProjects/api-platform/core/vendor/symfony/validator/Mapping/Loader/AnnotationLoader.php:120
/Users/chris/PhpstormProjects/api-platform/core/vendor/symfony/validator/Mapping/Loader/AnnotationLoader.php:60
/Users/chris/PhpstormProjects/api-platform/core/tests/Core/Bridge/Symfony/Validator/Metadata/Property/ValidatorPropertyMetadataFactoryTest.php:224

**How to reproduce**  
run tests locally

dunglas:
I fixed this in 2.6 some days ago. This will land in main when we'll merge 2.6 in main.

PR #4510, opened 2021-10-11
fix: add missing symfony/intl to fix tests

Chris53897:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4509
| License       | MIT

Fix failing test on tests locally


Chris53897:
Can we make the test skipable if symfony/intl is not installed? So test will run in Github action CI and locally without modifying composer manually on each branch?

Issue #4512, opened 2021-10-12
Implement a new cache system natively compatible to Caddy


darkweak:
**Description**  

Hello there 👋,  
I suggest to implement a new cache system called [Souin](https://github.com/darkweak/souin) in addition to the Varnish one.  
The advantages would be that Souin can be used directly as caddy module which replace Nginx. I tried to implement it directly in the api-platform stack and the only one issue I am facing is the mercure and vulcain pushes. I don't know how to handle that, which routes can I exclude (for mercure I could exclude .well-known routes but vulcain is a little bit more tricky).
I handle actually the route/resource invalidation with the bundle [Souin API Platform bundle](https://github.com/darkweak/SouinAPIPlatform) which one implements the PurgerInterface (note that this bundle is currently in a beta state / poc).  
Souin handles your CDN provider invalidation (like Fastly, Akamaï, Cloudflare (next release)) and set the correct HTTP headers to set your cache inside your CDN too (and obviously store the response inside Souin).  

**Example**  

```Caddyfile
{
    order cache before rewrite
    # Debug
    {$DEBUG}
    # HTTP/3 support
    servers {
        protocol {
            experimental_http3
        }
    }
    cache {
        api {
            souin {
                enable true
            }
        }
        regex {
            exclude /test2.*
        }
        headers Content-Type Authorization
        log_level debug
        ttl 1000s
        stale 2000s
    }
}

#...

route {
    cache
    ... other configuration
```

```yaml
# config/api_platform.yaml
api_platform:
        # ...
        http_cache:
                invalidation:
                    enabled: true
                public: true
```

```yaml
# config/souin_api_platform.yaml
souin_api_platform: # minimal to enable the Souin bridge
```

dunglas:
Hi!

It would be great to have native support for Souin. I fail to see what's the issue with Vulcain and Mercure. This shouldn't interfer with an HTTP cache server. Could you be a bit more specific?

darkweak:
Fixed in the https://github.com/darkweak/souin/pull/138 with the optimisations. I hope to merge the previous PR today and release in few days the fully working invalidation from the API Platform system. After that I will add the AddTagsListener and we'll be good.

PR #4540, opened 2021-11-09
Add Souin as available provider to the existing system

darkweak:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4512
| License       | MIT

Add Souin as an available purger provider instead of the Varnish one to be able to use natively Souin as a cache system inside the Caddy reverse-proxy and refactor the AddTagsListener to be able to customise the HTTP header tag and the separator instead of the forced Cache-Tag from Cloudflare.


soyuka:
note that the ci is broken we're trying to fix underlying issues

darkweak:
@soyuka @dunglas this PR is ready for review

tests/Core/Bridge/Symfony/Bundle/DependencyInjection/ConfigurationTest.php
```diff
@@ -195,12 +195,25 @@ private function runDefaultConfigTests(array $doctrineIntegrationsToLoad = ['orm
                     'http_tags' => [
                         'enabled' => true,
                     ],
+                    'purger' => [
+                        'souinBaseHost' => '',
+                        'souinBaseApiPath' => '',
+                        'souinApiSouinPath' => '',
+                        'souinApiAuthenticationPath' => '',
+                        'username' => '',
+                        'password' => '',
```
soyuka, line 204:
```suggestion
                        'souin' => []
```

do we really need the souinBaseApiPath ? Isn't `host` enough? or maybe a dsn

tests/Core/HttpCache/EventListener/AddTagsListenerTest.php
```diff
@@ -37,6 +37,8 @@ class AddTagsListenerTest extends TestCase
 {
     use ProphecyTrait;
 
+    public const DEFAULT_CACHE_TAG = 'Cache-Tag';
```
soyuka, line 40: breaks Imo

features/http_cache/tags.feature
```diff
@@ -17,15 +17,15 @@ Feature: Cache invalidation through HTTP Cache tags
     }
     """
     Then the response status code should be 201
-    And the header "Cache-Tags" should not exist
+    And the header "Cache-Tag" should not exist
```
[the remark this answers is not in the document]
darkweak: It doesn't break the old code because the old code didn't work anymore since mid 2015.

src/Core/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -612,16 +621,32 @@ private function registerHttpCacheConfiguration(ContainerBuilder $container, arr
...
+        switch ($config['http_cache']['invalidation']['type']) {
+            case 'souin':
+                $container->getDefinition('api_platform.http_cache.purger.souin')->setArguments([
+                    $definitions,
+                    $config['http_cache']['invalidation']['purger']['souinBaseHost'],
+                    $config['http_cache']['invalidation']['purger']['souinBaseApiPath'],
+                    $config['http_cache']['invalidation']['purger']['souinApiSouinPath'],
+                    $config['http_cache']['invalidation']['purger']['souinApiAuthenticationPath'],
+                    $config['http_cache']['invalidation']['purger']['username'],
+                    $config['http_cache']['invalidation']['purger']['password'],
```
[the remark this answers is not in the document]
darkweak: I'll reword the parameters to have some generic ones. But the core development won't change imho.

src/Core/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,187 @@
...
+    private function getParametersFromIris(array $iris): string
+    {
+        return implode(self::SEPARATOR, $iris);
+    }
+
+    public function login(): void
+    {
+        if ($this->username && $this->password) {
+            try {
+                $response = $this->clients[0]->request(
```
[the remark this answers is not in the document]
darkweak: We want to login once.

src/Core/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,187 @@
...
+                        continue;
+                    }
+
+                    $this->token = $splitCookie[1] ?? '';
+                }
+            } catch (GuzzleException $e) {
+                $this->logger->warning($e->getMessage());
+            }
+        }
+    }
```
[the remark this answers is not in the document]
darkweak: I can ask for a JWT instead of the credentials, you're right.

tests/Core/HttpCache/EventListener/AddTagsListenerTest.php
darkweak: ??

darkweak:
The tests on the main branch break the PHPUnit ones on PRs. Due to the bb2180f0acc091ce62b414bbc80d76971b300e0c commit.

tests/Core/HttpCache/EventListener/AddTagsListenerTest.php
soyuka: see above

features/http_cache/tags.feature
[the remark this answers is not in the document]
soyuka: You may implement this in a VCL manually and yes it is working .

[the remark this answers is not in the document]
darkweak: As it's configurable from the API Platform configuration, feel free to set the Cache-Tags header instead of the Cache-Tag one. Cloudflare doesn't support Cache-Tags (the one **with** a trailing s) header anymore and only the Cache-Tag (the one **without** the trailing slash).

tests/Core/HttpCache/EventListener/AddTagsListenerTest.php
darkweak: Read the cloudflare documentation.

darkweak:
ping @dunglas to review as we discussed with @soyuka

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,127 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\HttpCache;
+
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\GuzzleException;
```
soyuka, line 17: we shouldn't use guzzle but the symfony http client instead

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -424,7 +424,43 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->info('Cache provider to use (e.g. souin, varnish, whatever)')
+                                ->end()
+                            ->end()
+                        ->end()
+                        ->arrayNode('validation')
+                            ->addDefaultsIfNotSet()
+                            ->info('Enable the tags-based cache validation system.')
+                            ->children()
+                                ->scalarNode('header')
+                                    ->defaultValue('Cache-Tag')
```
soyuka, line 458: should be `Cache-Tags` or it is a BC as we had this as default in 2.6. We can deprecate `Cache-Tags` in favor of `Cache-Tag` in 2.7 if you all agree but as there is no real specification I'm unsure of the correct answer.

tests/HttpCache/EventListener/AddTagsListenerTest.php
```diff
@@ -37,6 +37,8 @@ class AddTagsListenerTest extends TestCase
 {
     use ProphecyTrait;
 
+    public const DEFAULT_CACHE_TAG = 'Cache-Tag';
```
soyuka, line 40: should be `Cache-Tags` see above comment

tests/HttpCache/SouinPurgerTest.php
```diff
@@ -0,0 +1,293 @@
...
+            public function requestAsync($method, $uri, array $options = []): PromiseInterface
+            {
+                throw new LogicException('Not implemented');
+            }
+
+            public function getConfig($option = null)
+            {
+                throw new LogicException('Not implemented');
+            }
+        };
```
soyuka, line 286: I don't really get why the whole interface needs to be implemented

soyuka (changes requested):

src/Symfony/Bundle/DependencyInjection/Configuration.php
darkweak: The cloudflare specification would be the correct answer. Anyway, I'll write a Cloudflare V4 provider and revert the Cache-Tag into Cache-Tags for the existing one.
IMHO keep the Cache-Tags header sucks because it's not maintained by Cloudflare and the "new" specification would be the only one we have to follow.

tests/HttpCache/SouinPurgerTest.php
darkweak: Because I need a ClientInterface.

src/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka: yes I know the only issue is that it breaks current apis so we can't just change it we need to deprecate it first

darkweak:
@soyuka I reverted the Cache-Tag into Cache-Tags. Can you re-review for the last time?
Ty

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -44,18 +44,22 @@ final class AddTagsListener
     private $xkeyEnabled;
     private $xkeyGlue;
     private $httpTagsEnabled;
+    private $header;
+    private $separator;
 
-    public function __construct(IriConverterInterface $iriConverter, ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, bool $xkeyEnabled = false, string $xkeyGlue = ' ', bool $httpTagsEnabled = true)
+    public function __construct(IriConverterInterface $iriConverter, ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, bool $xkeyEnabled = false, string $xkeyGlue = ' ', bool $httpTagsEnabled = true, string $header = 'Cache-Tags', string $separator = ',')
```
dunglas, line 50: Couldn't we do something like this instead (the real implementation will of course be more complicated because we need a BC layer)?

```suggestion
public const VARNISH_HEADER_NAME = 'xkey';
public const VARNISH_SEPARATOR = ' ';
public const SOUIN_HEADER_NAME = 'Cache-Tag';
public const SOUIN_SEPARATOR = ',';

    public function __construct(IriConverterInterface $iriConverter, ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, string $headerName = false, string $separator = ' ') {}
    
    public static function createSouinInstance(IriConverterInterface $iriConverter, ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory): static {};
    public static function createVarnishInstance(IriConverterInterface $iriConverter, ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory): static {};
```

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,126 @@
...
+ */
+class SouinPurger implements PurgerInterface
+{
+    private const MAX_HEADER_SIZE_PER_BATCH = 1500;
+    private const SEPARATOR = ', ';
+    private const SOUIN_COOKIE_NAME = 'souin-authorization-token';
+
+    private $logger;
+
+    // Clients to send cache invalidation
```
dunglas, line 35: You should use the PHPDoc syntax instead for all these comments.

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,126 @@
...
+        string $token
+    ) {
+        $this->clients = $clients;
+        $this->logger = $logger;
+        $this->dsn = $dsn;
+        $this->path = $path;
+        $this->token = $token;
+    }
+
+    private function getParametersFromIris(array $iris): string
```
dunglas, line 62: Is this method really necessary? It's used only one-time.

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,126 @@
...
+    {
+        return implode(self::SEPARATOR, $iris);
+    }
+
+    /**
+     * @param array|string[] $iris
```
dunglas, line 68:
```suggestion
     * @param \Stringable[] $iris
```

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @return string[]
+     */
+    private function getChunkedRegex(array $iris): array
+    {
+        $regex = $this->getParametersFromIris($iris);
+        $batches = [];
+        $separatorLength = \strlen(self::SEPARATOR);
```
dunglas, line 76: Can't this be hardcoded (or stored in another private constant), as the parameter is a constant?

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function purge(array $iris)
+    {
+        if (!$iris) {
+            return;
+        }
```
dunglas, line 120: Useless, can be removed:

```suggestion
```

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -426,6 +426,42 @@ private function addHttpCacheSection(ArrayNodeDefinition $rootNode): void
...
+                                            ->defaultValue('')
+                                            ->info('Purger dsn')
+                                        ->end()
+                                        ->scalarNode('path')
+                                            ->defaultValue('')
+                                            ->info('Purger API path')
+                                        ->end()
+                                        ->scalarNode('token')
+                                            ->defaultValue('')
+                                            ->info('JWT token to use if the provider API if secured')
```
dunglas, line 443: That's weird, shouldn't this be under a `souin` section as it's relevant only in this context. If it's just a bearer token, then the should be renamed to make it explicit that there is no hard dependency on Souin.

dunglas (changes requested):

src/HttpCache/EventListener/AddTagsListener.php
darkweak: I don't think this is a good approach and is more complex than it should be. All additional parameters are not mandatory, they don't break anything. It uses the existing parameters values for the cache header and the separator.

darkweak:
Hi there, what about this?

BeyerJC:
Hey, is there a Chance this feature will be in 2.7 ?

darkweak:
I don't think because they never answered 🙃

darkweak:
@soyuka @dunglas Can I have a review please ? Everything is working as expected with tests. It would be a great add-on on the v3 to deal with the caddy native HTTP cache system.

dunglas:
We're focus on releasing API Platform 3 (which is in beta). We'll review this PR as well as other new features when API Platform 3 will be tagged.

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\HttpCache;
+
+use GuzzleHttp\ClientInterface;
```
maxhelias, line 16: You should use this interface `\Symfony\Contracts\HttpClient\HttpClientInterface` instead

darkweak:
Ready to merge.

src/HttpCache/SouinPurger.php
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * @var string
+     */
+    private $path;
+
+    /**
+     * @param HttpClientInterface[] $clients
+     */
+    public function __construct(
+        array $clients,
```
BeyerJC, line 56: Could we use property Promotion and make them readonly here?
Apip 3.0 requires PHP 8.1

darkweak:
ping @dunglas @soyuka

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -545,21 +544,32 @@ private function registerHttpCacheConfiguration(ContainerBuilder $container, arr
...
-            $definition = new Definition(ScopingHttpClient::class, [new Reference('http_client'), $url, ['base_uri' => $url] + $config['http_cache']['invalidation']['request_options']]);
-            $definition->setFactory([ScopingHttpClient::class, 'forBaseUri']);
 
-            $definitions[] = $definition;
-        }
+        $serviceName = $config['http_cache']['invalidation']['purger'];
+        switch ($serviceName) {
+            case 'api_platform.http_cache.purger.souin':
+                $container->getDefinition($serviceName)->setArguments([
+                    $definitions,
```
maxhelias, line 552: It seems that the definitions here are empty.

darkweak:
Can I have a last review please?

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -562,21 +561,39 @@ private function registerHttpCacheConfiguration(ContainerBuilder $container, arr
         $loader->load('http_cache_purger.xml');
 
         $definitions = [];
-        foreach ($config['http_cache']['invalidation']['varnish_urls'] as $key => $url) {
-            $definition = new Definition(ScopingHttpClient::class, [new Reference('http_client'), $url, ['base_uri' => $url] + $config['http_cache']['invalidation']['request_options']]);
-            $definition->setFactory([ScopingHttpClient::class, 'forBaseUri']);
 
-            $definitions[] = $definition;
-        }
+        $serviceName = $config['http_cache']['invalidation']['purger'];
```
soyuka, line 565: What about having

```
http_cache:
  invalidation:
    souin_urls: 
```

?

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -562,21 +561,39 @@ private function registerHttpCacheConfiguration(ContainerBuilder $container, arr
...
+        switch ($serviceName) {
```
soyuka, line 566: I think we prefer `match` now

src/Symfony/Bundle/Resources/config/http_cache_purger.xml
```diff
@@ -12,12 +12,18 @@
...
+            <argument>%api_platform.http_cache.invalidation.path%</argument>
+        </service>
 
         <service id="api_platform.http_cache.listener.response.add_tags" class="ApiPlatform\HttpCache\EventListener\AddTagsListener">
             <argument type="service" id="api_platform.iri_converter" />
             <argument type="service" id="api_platform.metadata.resource.metadata_collection_factory" on-invalid="null" />
             <argument type="service" id="api_platform.http_cache.purger" on-invalid="null" />
             <tag name="kernel.event_listener" event="kernel.response" method="onKernelResponse" priority="-2" />
         </service>
+
```
soyuka, line 27:
```suggestion
```

tests/Metadata/Extractor/YamlExtractorTest.php
```diff
@@ -471,7 +471,7 @@ public function testInvalidYaml(string $path, string $error): void
     public function getInvalidPaths(): array
     {
         return [
-            [__DIR__.'/yaml/invalid/invalid_resources.yaml', '"resources" setting is expected to be null or an array, string given in "'.__DIR__.'/yaml/invalid/invalid_resources.yaml'.'".'],
+            [__DIR__.'/yaml/invalid/invalid_resources.yaml', '"resources" setting is expected to be null or an array, string given in "'.__DIR__.'/yaml/invalid/invalid_resources.yaml".'],
```
soyuka, line 474: This should've been fixed by now, please revert this chang

soyuka:
Nice! A few comments lefts but we're almost good to merge imo! Thanks @darkweak sorry for the delays!

dkarlovi:
@soyuka note this discussion: https://github.com/darkweak/souin/issues/277#issuecomment-1327339148

It seems Souin will start passing any request as soon as the Authorization header (or Cookie) is present in the request, it doesn't provide (nor plans to provide) any sort of "user context"-like feature FOS HTTP Cache bundle, which IIRC APIP relied on before.

darkweak:
Ready for the merge @soyuka

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
darkweak: The switch case allow us to use varnish by default.

divine:
> @soyuka note this discussion: [darkweak/souin#277 (comment)](https://github.com/darkweak/souin/issues/277#issuecomment-1327339148)
> 
> It seems Souin will start passing any request as soon as the Authorization header (or Cookie) is present in the request, it doesn't provide (nor plans to provide) any sort of "user context"-like feature FOS HTTP Cache bundle, which IIRC APIP relied on before.

Well, then there is no reason that souin should be integrated into APP.

offtopic:

@darkweak would like to consider changing the souin logic?

Thanks!

darkweak:
@divine note this PR https://github.com/darkweak/souin/pull/283

soyuka:
merci @darkweak

soyuka:
I cherry-picked it in my PR just to be a bit less provider specific. souin will have native support from API Platform 3.1 (coming out mid-January)

Issue #4547, opened 2021-11-09
Doctrine json_array type should now be json


soyuka:
Since `json_array` has been deprecated in Doctrine DBAL we should use `json`. In graphql this changes how the type opperates. Change the `json` type in Entity/Dummy and fix the corresponding behat tests.

https://github.com/api-platform/core/blob/main/tests/Fixtures/TestBundle/Entity/Dummy.php#L136-L138 

Note that you should probably also change the `array` type or add the `nullable` PHP type .

+ allow the `doctrine/dbal` at version `3.x` (currently fixed at 2.6 https://github.com/api-platform/core/pull/4544/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R36)

laryjulien:
I have opened a PR to resolve the other issue with doctrine/dbal 3.x

soyuka:
is this fixed @dunglas  ?

dunglas:
Yes: https://dunglas.dev/2022/01/json-columns-and-doctrine-dbal-3-upgrade/

PR #4549, opened 2021-11-09
fix(doctrine): replace Type class by Types

laryjulien:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4547 
| License       | MIT
| Doc PR        |  n/a

In doctrine/dbal 3.0, types constants are now defined in Doctrine\DBAL\Types\Types.php instead of Doctrine/DBAL/Types/Type.php

References:

- https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Types/Type.php
- https://github.com/doctrine/dbal/blob/3.1.x/src/Types/Types.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "doctrine/cache": "^1.11",
         "doctrine/common": "^2.11 || ^3.0",
         "doctrine/data-fixtures": "^1.2.2",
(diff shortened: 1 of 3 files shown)


soyuka:
can you target 2.6 on this?

PR #4562, opened 2021-11-19
feat: link implementation

soyuka:
Remains to be done:
- [x] Remove Elasticsearch support with GraphQL (https://github.com/api-platform/core/commit/a625fe7211f877613ca1e971bbe17f7895401acb)
- [x] Convert ApiProperty annotations to attributes through rectorphp
- [x] Convert ApiFilter annotations to ApiResource filters
- [ ] Fix MongoDB


chalasr:
Is this about hypermedia links as specified by PSR-13? How do they differ? Naming seems confusing :)
I hope we can find a different name (`Relation`?) but if we can't, mentioning the difference in the ADR would probably be helpful.

soyuka:
Not really, here it is about defining how URI variables are mapped inside API Platform. For example the URI template `/book/{id}` also needs to explain that `id` is the property `id` of the Book class. When we're in the case of a subresource `/book/{id}/author/{authorId}` we need to add the information that `authorId` is the property `id` of the class Author and that Book and Author are related via the property `author` of the Book class. 

Notes - uri linksk definition api p

Note that we need all these informations as the relation may be done through another property, here the property `book` of the class Author:

Notes - uri linksk definition api p inverse

The above also works for a graphql query like this:

```
{
  books(id: "/books/2") {
    authors {
      edges {
        node {
          id
        }
      }
    }
  }
}
```

soyuka:
Huge work ! Thank you @alanpoulain @vincentchalamon ! Mongodb is the last bit.

PR #4563, opened 2021-11-20
Remove php-cs-fixer in favor of ECS

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | none
| License       | MIT
| Doc PR        | none

This PR adds [ECS](https://github.com/symplify/easy-coding-standard) instead of PHP-CS-FIXER.
It uses exactly the same rules as previously set in `.php-cs-fixer.dist.php`.
Please note that I am not sure of the rules `PHP71MigrationSet`, `PHP71MigrationRiskySet` and `PHPUnit60MigrationRiskySet` as those are marked as `internal`.
Maybe @matthiasnoback or @TomasVotruba could help us with those choices?
Open to comments and discussions


TomasVotruba:
Thanks for pinging! Great job with such a quick switch :+1: 

May I ask, what motivated you to do the switch just now?

You've probably read this post, but just in case you missed it - here a post with few steps to upgrade from PHP CS FIxer to ECS: https://tomasvotruba.com/blog/2018/06/07/how-to-migrate-from-php-cs-fixer-to-easy-coding-standard/

Few tips from top of my head:

* Use parallel run https://tomasvotruba.com/blog/introducing-up-to-16-times-faster-easy-coding-standard/ - it will be 8-16 faster and much more fun to run locally :)
* Try to avoid any manual fixer registration, as that require maintenance and makes config huge to read - in ECS there are sets like `SetList::PSR_12` that include most of those rules and put your code to based standard

> Please note that I am not sure of the rules PHP71MigrationSet, PHP71MigrationRiskySet and PHPUnit60MigrationRiskySet as those are marked as internal.

* Tokens are very dangerous to use for context, e.g. from tokens you have no idea if the variably is `EntityManager`, `string` or `mixed`. For any PHP package upgrade, there is now Rector with reliable PHPStan type analysis underhood. It should be used instead.

If you have some specific question, ping me at specific line in the PR. I'll try to help :)

Spomky:
Hi @TomasVotruba,

> May I ask, what motivated you to do the switch just now?

I actually bought and started reading your book on [Rector](https://leanpub.com/rector-the-power-of-automated-refactoring) and where ECS is mentioned.
I have started using it for some of my projects and since it is really awesome (especially the PHP config file) I think it is worth using it here.

> Use parallel run

Nice. Adopted!

> Try to avoid any manual fixer registration

The primary intent of this PR is to stick to the current php-cs-fixer configuration file. But if the maintainers agree with it, I will move to ruleset without manual options.

> There was no better tool 5 years ago, but tokens are very dangerous to use for context[...]

OK noted. So I remove these rules in order to avoid potential breaks.
Many thanks for your review and hints!

alanpoulain:
Hello,
Thanks for this PR but I'm not sure we would like to have it in API Platform.
Since we don't use PHP_CS, is there any interest of using it instead of php-cs-fixer?

Spomky:
> Since we don't use PHP_CS, is there any interest of using it instead of php-cs-fixer?

As I wrote few minutes ago,  I think it is worth using it here. But the CS tools used are completely up to you. It's just a matter of taste and habits. It doesn't detract from the quality of API Platform if you don't want to use it.

alanpoulain:
I don't think having the configuration in PHP is a sufficient reason to change.
As you say it's a matter of taste, I understand if you prefer it but we choose to stick to php-cs-fixer for now.
If there are more reasons to change it in the future, maybe we will do it.
Thank you anyway.

TomasVotruba:
The main motivation for people prepared sets beyond bare PSR 12. Just 10 lines in ecs.php and they have most advanced coding standard there is. 

Since last month, the main reason is the parallel run. The current tool now it takes 58 seconds.

@Spomky Out of curiosity, how was ECS fast with parallel in CI?

alanpoulain:
For information, the GHA job currently takes about 1 minute and since it is run in parallel with the tests, the performance of it isn't really a concern for us.

Spomky:
Running using WSL2, 16Go/8Cores allocated

Parallel mode enabled:
>  1418/1418 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
>                                                                                           
>  [OK] No errors found. Great job - your code is shiny in style!                                                         
>                                                                                                                         
> real    0m10.769s
> user    0m0.417s
> sys     0m0.115s

Parallel mode disabled:
>  1418/1418 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
>                                                                                           
>  [OK] No errors found. Great job - your code is shiny in style!                                                         
>                                                                                                                         
> real    0m38.643s
> user    0m38.219s
> sys     0m0.401s

So approx. 4x faster.

TomasVotruba:
@Spomky  Thanks for sharing :+1: 

It should be 8x times faster, as you have 8 cores, right?

> user 0m0.417s

Spomky:
> @Spomky Thanks for sharing 👍
> 
> It should be ~8 times faster, as you have 8 cores, right?
> 
> > user 0m0.417s

Yes it should be, but as it is behind the WSL layer I guess this is not so accurate.

PR #4565, opened 2021-11-22
fix: use `Builder::getAggregation()` instead of `Builder::execute()` when available

azjezz:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A


soyuka:
we need this on 2.6 imo also quite some tests are red would you mind fixing it?

soyuka:
Fixed it thanks @azjezz

Issue #4535, opened 2021-11-05
Additional PHP 8 attributes for easier customization


norkunas:
**Description**  
Now that Symfony 6 will require PHP 8, we could leverage php attributes for `Processor`, `CollectionProvider` and `ItemProvider`.

For now when you have many custom operations and need to configure in what priority processor or providers run, you need to write custom configuration.

I am thinking that it could be nice to go in the same direction as Symfony with `AsEventListener` attribute.

**Example**  
```
class AsStateProcessor
{
    public function __construct(public int $priority = 0) {}
}

class AsCollectionProvider
{
    public function __construct(public int $priority = 0) {}
}

class AsItemProvider
{
    public function __construct(public int $priority = 0) {}
}
```

alanpoulain:
I think it could be nice to have this feature too.
However, only an `AsStateProvider` is necessary, we don't make a distinction between item and collection anymore.

PR #4573, opened 2021-11-25
Add `AsStateProvider` and `AsStateProcessor` attributes

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4535
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/State/AsStateProcessor.php
+++ b/src/State/AsStateProcessor.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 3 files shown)


dunglas:
Couldn't we just use the native [`AsTaggedItem`](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/DependencyInjection/Attribute/AsTaggedItem.php) Symfony attribute?

If it works, we should document this.

norkunas:
> Couldn't we just use the native [`AsTaggedItem`](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/DependencyInjection/Attribute/AsTaggedItem.php) Symfony attribute?
> 
> If it works, we should document this.

But that would be directly tied to Symfony, but if you prefer that way I could add docs then

soyuka:
Isn't `$container->registerForAutoconfiguration(ProcessorInterface::class)` enough?

PR #4576, opened 2021-11-25
use OpenApiNormalizers in SwaggerUiAction

Suvres:
[https://github.com/api-platform/core/issues/4575l](https://github.com/api-platform/core/issues/4575l)


soyuka:
I don't get the use case can you explain ?

PR #4577, opened 2021-11-26
fix alias filter eager

jorissae:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | NA
| License       | MIT
| Doc PR        | NA

When two aliases of a join are very close they conflict with the str_replace, causing an error when applying a filter

because `t.` and `ct.` are similar and `.t` is applied first, the wrong alias is set.
`echo str_replace(['o.', 't.', 'ct.'],['o_a1.', 'transaction_a2.', 'cardTransaction_a3'], 'ct.card').PHP_EOL;`

fix proposal is to use preg_replace instead
`echo preg_replace(['/^o\./', '/^t\./', '/^ct\./'],['o_a1.', 'transaction_a2.', 'cardTransaction_a3'], 'ct.card').PHP_EOL;`

view with @GregoireHebert


soyuka:
can you add a non-regression test please ?

soyuka:
closing in favor of https://github.com/api-platform/core/pull/4592

Issue #4338, opened 2021-06-23
Related resources not serialized into IRI-s, if root resource uses an output DTO


kgilden:
**API Platform version(s) affected**: 2.6.5, but it's likely a much older problem

**Description**  
If a root resource uses an output DTO, then all properties referring to related API resources get serialized to objects instead of their IRI-s.

**How to reproduce**  

```php
/**
 * @ApiResource(output = ParentOutput::class)
 */
class Parent
{
    public int $id;
    public Child $child;
}

class ParentOutput
{
    public int $id;
    public Child $child;
}

/**
 * @ApiResource
 */
class Child
{
    public int $id;
}
```

**Current result**

```sh
curl api.example.com/parents/42
```

```json
{
  "@context": {},
  "@type": "Parent",
  "@id": "/parent/42",
  "id": 42,
  "child": {
    "@context": " ... ",
    "@id": "/children/1",
    "id": 1,
  }
}
```

**Expected result**

```sh
curl api.example.com/parents/42
```

```json
{
  "@context": {},
  "@type": "Parent",
  "@id": "/parent/42",
  "id": 42,
  "child": "/children/1"
}
```

**Possible Solution**  

The problem seems to be caused by [AbstractItemNormalizer](https://github.com/api-platform/core/blob/6336f6446722cd6728683c36fb70e0074553296d/src/Serializer/AbstractItemNormalizer.php). If the root resource **does not** specify an output class, the same normalizer normalizes the attributes, and [getAttributeValue](https://github.com/api-platform/core/blob/6336f6446722cd6728683c36fb70e0074553296d/src/Serializer/AbstractItemNormalizer.php#L609) gets called. If the root resource **does** specify an output class, the attributes are recursively passed to the serializer. The recursion causes the related resource objects to get converted to DTO-s and of course since DTO-s are not resources, they are not converted to IRI-s.

So there should be some kind of a mechanism that serializes the attributes of a DTO the same way as the attributes of a resource class.

**Additional Context**  

#3795 seems to describe the same problem, but due to lack of activity was closed. There is also some discussion about the same problem and workarounds at https://symfonycasts.com/screencast/api-platform-extending/collections-readable-link#comment-5107585662.

[...]

PR #4578, opened 2021-11-27
feat(mercure): Pass 'resource_class' into serialization context

kgilden:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

The `$context` array passed to custom normalizers is different whether a resource is being serialized as part of a request-response lifecycle or whether it's being serialized to publish to the Mercure hub. This makes it difficult to ensure that the same resource gets serialized consistently.

The `resource_class` key defines the root resource getting serialized and it's missing when serializing for the Mercure hub. To not rock the boat too much, I decided just to add this to the serialization context.

### Questions to reviewers

1. Is this perhaps small enough of a feature that it could be merged into current stable (2.6) branch?
2. If documentation update is necessary, which article should I update?

### Future work

In the longer run it might be beneficial to ensure the following.

1. All keys of the context variable declared by API Platform are documented (including their types).
2. All keys of the context variable are the same regardless of whether serializing during a request-response lifecycle or whether serializing to publish a Mercure update. If a context variable makes sense only in one of the two cases, then its type can simply by `T|null`.

*But*, I acknowledge this might be a lot of work and I don't want to burden the maintainers by opening yet another issue.

### Background

Making this PR to make my workaround for #4338 work.

> I myself came up with [a workaround](https://gist.github.com/kgilden/5a0bb420e7b4c8e20e6744e90e79a1f4), which uses the `resource_class` context property to determine what the root is and just use an IRI converter for all other cases. But this falls short, when trying to publish to a Mercure hub, because the context is created differently ([here](https://github.com/api-platform/core/blob/6336f6446722cd6728683c36fb70e0074553296d/src/EventListener/SerializeListener.php) vs [here](https://github.com/api-platform/core/blob/6336f6446722cd6728683c36fb70e0074553296d/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php#L264)).

### DIFF
--- a/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -227,6 +227,9 @@ final class PublishMercureUpdatesListener
         } else {
             $resourceClass = $this->getObjectClass($object);
             $context = $options['normalization_context'] ?? $this->resourceMetadataFactory->create($resourceClass)->getAttribute('normalization_context', []);
+            if (!isset($context['resource_class'])) {
+                $context['resource_class'] = $resourceClass;
+            }
 
             $iri = $options['topics'] ?? $this->iriConverter->getIriFromItem($object, UrlGeneratorInterface::ABS_URL);
             $data = $options['data'] ?? $this->serializer->serialize($object, key($this->formats), $context);
--- a/tests/Bridge/Doctrine/EventListener/PublishMercureUpdatesListenerTest.php
+++ b/tests/Bridge/Doctrine/EventListener/PublishMercureUpdatesListenerTest.php
@@ -93,8 +93,14 @@ class PublishMercureUpdatesListenerTest extends TestCase
         $resourceMetadataFactoryProphecy->create(DummyFriend::class)->willReturn(new ResourceMetadata(null, null, null, null, null, ['mercure' => "['foo', 'bar']"]));
 
         $serializerProphecy = $this->prophesize(SerializerInterface::class);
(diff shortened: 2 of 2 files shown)


src/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -269,7 +269,9 @@ private function publishUpdate($object, array $options, string $type): void
             $data = json_encode(['@id' => $object->id]);
         } else {
             $resourceClass = $this->getObjectClass($object);
-            $context = $options['normalization_context'] ?? $this->resourceMetadataFactory->create($resourceClass)->getOperation()->getNormalizationContext() ?? [];
+            $context = array_replace([
+                'resource_class' => $resourceClass,
+            ], $options['normalization_context'] ?? $this->resourceMetadataFactory->create($resourceClass)->getOperation()->getNormalizationContext() ?? []);
```
soyuka, line 274: can you avoid `array_replace`? Just do something like:

```
  $context['resource_class'] = $resourceClass;
}
```

soyuka:
you can patch this on 2.6 if you want but please also add the fix on `main` if you can thanks

dunglas:
+1 on my side when @soyuka's review will be handled.

I also agree with your other proposals.

kgilden:
Alright, sounds good. I'll try to see if I can make the fixes some time later this week.

kgilden:
@soyuka I missed the train to 2.6.7, but I made the changes. As for adding the fix also to `main`, do you mean create another PR against the `main` branch?

src/Doctrine/EventListener/PublishMercureUpdatesListener.php
kgilden: Done.

PR #4581, opened 2021-11-30
Conflict with symfony/routing 6+

Korbeil:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | N/A
| License       | MIT

When requiring `api-platform/core` on a Symfony 6 project it's breaking because there is no dependencies about `symfony/routing` in the required packages. So we need to conflict with it until compatibility is there.

Issue #4543, opened 2021-11-09
Support Symfony 6.0


dunglas:
Now that we support Symfony 5.4, we must also support Symfony 6.0.

* [ ] Checkout the 2.6 branch
* [ ] Add  allow the `^6.0` version for all Symfony projects
* [ ] Run the test suite
* [ ] Remove the annoying type hints from Symfony's interfaces (open PR and ping me)

leeroynico:
I work on it

Chris53897:
This ticket is not so easy. I will try to help with this.
I think 2 steps are necessary in my point of view. 

1. Build a Pipeline and allow symfony 6 in dependencies to get an overview of how much work is needed to support symfony 6.
2. Decide if APIP 2.x needs support for Symfony 6. Or only APIP 3.x with PHP8. Because you can not make a project compatible with Symfony 6 if you do not require Php >8. At least i do not know a way to do this.


To 1)

The CI Pipeline will not test all symfony 6 packages. Symfony 6.0 needs to be added to the test-matrix on 2 places (this allows the 2 additional packages to be tested against 6.0. [Example Test-Run](https://github.com/Chris53897/core/runs/4194724799?check_suite_focus=true)

Some symfony components are sticked to 5.4 because of dependencies. 
(The link mink-repo is merged now, but the tests are not covering symfony 6.)

Example
`composer why symfony/browser-kit`

...
friends-of-behat/mink-browserkit-driver  v1.5.0    requires     symfony/browser-kit (^4.4|^5.0)

I made a [PR draft](https://github.com/FriendsOfBehat/MinkBrowserKitDriver/pull/10)
and will try to update b).
a) Origin https://github.com/minkphp/driver-testsuite
b) Fork von oberen https://github.com/FriendsOfBehat/driver-testsuite (is used in friends-of-behat/mink-browserkit-driver)

Chris53897:
update:
One blocker is gone. 
Here i wait for a new release.
https://github.com/minkphp/Mink/pull/814

After that "soyuka/contexts" could be updated too.

Now that 6.x is stable the CI Tests should be easier.

PR #4582, opened 2021-12-01
feat: Add symfony 6 compatibility

PierreRebeilleau:
| Q             | A
| ------------- | ---
| Branch?       | 2.6 
| Tickets       | https://github.com/api-platform/core/issues/4543, https://github.com/api-platform/core/pull/4581 
| License       | MIT
| Doc PR        | no 

Test are failing because of some type hinting, waiting for a new release of symfony 6 , solved by https://github.com/symfony/symfony/pull/44331


composer.json
```diff
@@ -18,12 +18,12 @@
         "fig/link-util": "^1.0",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "psr/container": "^1.0",
-        "symfony/http-foundation": "^4.4 || ^5.1",
-        "symfony/http-kernel": "^4.4 || ^5.1",
-        "symfony/property-access": "^3.4.19 || ^4.4 || ^5.1",
+        "symfony/http-foundation": "^4.4 || ^5.1 || ^6.0",
+        "symfony/http-kernel": "^4.4 || ^5.1 || ^6.0",
+        "symfony/property-access": "^3.4.19 || ^4.4 || ^5.1 || ^6.0",
         "symfony/property-info": "^3.4 || ^4.4 || ^5.2.1",
```
dunglas, line 24:
```suggestion
        "symfony/property-info": "^3.4 || ^4.4 || ^5.2.1 || ^6.0",
```

dunglas (changes requested):
Can you also allow the latest versions of the PSR packages please?

pietereggink:
symfony/symfony#44331 is merged, is it possible to execute the quality checks again?

rkeet:
Hi, was looking if I could help out a bit so did some digging through requirements in `composer.json` as I noticed it still has SF 3.4 in there. 

SF 3.4 LTS is now past EOL (since this month). 

Also, SF 3.4 has `"php": "^5.5.9|>=7.0.8"` ([link](https://github.com/symfony/symfony/blob/3.4/composer.json)) for requirements, while ApiPlatform has `">=7.1"` ([link](https://github.com/api-platform/core/blob/3c1f8ece93fc76fb5a3cb91b0f05a161d823deee/composer.json)) (not changed in this PR). Not sure if that would cause issues with compatibility, but maybe a removal of SF 3.4 of supported versions is also an idea for this PR, since you're updating with SF6?

Extra question: is there somewhere I can see/look into/help efforts for a possible v3 and/or minimum PHP 8 / 8.1 requirement?

dunglas:
I would prefer to drop Symfony 3.4/PHP 7.1 in the main branch. But if it's too annoying to keep supporting them, then let's bump these minimal versions.

The main branch contains the 2.7 version, which will be the same code as 3.0 but with a backward compatibility layer for API Platform 2 (that we'll drop before tagging 3.0).

vincentchalamon (approved):

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -125,6 +129,10 @@ protected function setAttributeValue($object, $attribute, $value, $format = null
             $attribute = 'id';
         }
 
+        if ('{}' === $value) {
```
dunglas, line 132: That's weird. Why do we need this?

tests/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtensionTest.php
```diff
@@ -808,49 +808,18 @@ public function testDisabledPaginationViaDefaults()
 
     public function testItLoadsMetadataConfigFileInAlphabeticalOrder()
     {
-        $yamlExtractorDefinition = $this->prophesize(Definition::class);
```
dunglas, line 811: Why is this test removed?

dunglas (changes requested):

src/GraphQl/Serializer/ItemNormalizer.php
PierreRebeilleau: When we dump `dump($context['args']['input']);` in  the DeserializeStage we got this : 
```
array:4 [
  "id" => "/related_dummies/2"
  "symfony" => "laravel"
  "thirdLevel" => "/third_levels/1"
  "embeddedDummy" => "{}"
]
```

dunglas: This looks like a bug in pho-GraphQL. Could you report it upstream and explain why we need this temporary fix in a comment please?

alanpoulain: Input is user-defined, there shouldn't be a fix in the serializer. Maybe the test is bad. Which test is failing?

PierreRebeilleau: https://github.com/api-platform/core/runs/4595101310?check_suite_focus=true
Seems to be this one

alanpoulain: This is strange, we need to dig it, it's probably not the right fix.

alanpoulain:
Thank you everyone.

dunglas:
You rock!

PR #4595, opened 2021-12-13
fix(Pagination): Fix UUID in range filter

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | main (cause I used the new project structure)
| Tickets       | [api-platform/api-platform#2087](https://github.com/api-platform/api-platform/issues/2087)
| License       | MIT
| Doc PR        | none

Turns out the things I was running into were a bug caused by the range filter expecting numbers.
Which doesn't work when you use UUIDs as PK :)


src/Doctrine/Common/Filter/RangeFilterTrait.php
```diff
@@ -124,18 +125,18 @@ private function normalizeBetweenValues(array $values): ?array
     /**
      * Normalize the value.
      *
-     * @return int|float|null
+     * @return int|float|string | null
      */
     private function normalizeValue(string $value, string $operator)
     {
-        if (!is_numeric($value)) {
+        if (!is_numeric($value) && !UuidV6::isValid($value)) {
```
alanpoulain, line 132: We don't want to depend on an external vendor, like Ramsey or even Symfony UID component.
Maybe we should have our own check?

src/Doctrine/Common/Filter/RangeFilterTrait.php
```diff
@@ -124,18 +125,18 @@ private function normalizeBetweenValues(array $values): ?array
...
             $this->getLogger()->notice('Invalid filter ignored', [
                 'exception' => new InvalidArgumentException(sprintf('Invalid value for "[%s]", expected number', $operator)),
             ]);
 
             return null;
         }
 
-        return $value + 0; // coerce $value to the right type.
```
alanpoulain, line 139: You shouldn't change this code.

src/Doctrine/Common/Filter/RangeFilterTrait.php
```diff
@@ -124,18 +125,18 @@ private function normalizeBetweenValues(array $values): ?array
...
      *
-     * @return int|float|null
+     * @return int|float|string | null
      */
     private function normalizeValue(string $value, string $operator)
     {
-        if (!is_numeric($value)) {
+        if (!is_numeric($value) && !UuidV6::isValid($value)) {
             $this->getLogger()->notice('Invalid filter ignored', [
                 'exception' => new InvalidArgumentException(sprintf('Invalid value for "[%s]", expected number', $operator)),
```
alanpoulain, line 134: Exception message needs to be changed.

src/Doctrine/Common/Filter/RangeFilterTrait.php
KDederichs: You can't keep it with uuid's though, I can add a clause to differentiate though

KDederichs:
Ok the behat 7.1 fails cause the test can't init the UUID v6 objects.
I'm not really familiar with behat, is there a way to skip a test if a class is missing?

KDederichs:
Or better why still test for 7.1?
It's long past EOL

Chris53897:
> Or better why still test for 7.1? It's long past EOL

Symfony backwards compatibility promise.
Symfony 4.4  php 7.1

KDederichs:
Ok skipping the test it is then (took me a while to figure out how to do that in behat but I gave it a custom tag that can be removed when 7.1 is dropped

KDederichs:
All green except deprecations 👍

soyuka:
This should be moved to its own filter.

KDederichs:
@soyuka should it though?
You can also have a range on a string imo. But if you want I can close this and make a new PR that introduces something like `UuidRange` or something and then also adds that to the docs.

KDederichs:
Closing in favour of https://github.com/api-platform/core/pull/4689
@soyuka if you'd take a look at that one I'd be grateful :)

PR #4611, opened 2021-12-31
Move last bits of non-legacy code to the new namespace

soyuka:


divine:
Hello,

Any help is needed to move this forward?

Thanks!

PR #4617, opened 2022-01-04
fix(serializer): Pass the child context when denormalizing nested non-resource objects

philipsorst:
Q | A
-- | --
Branch? | 2.6
Bug fix? | yes
New feature? | no
Deprecations? | no
Tickets | n/a
License | MIT
Doc PR | n/a

Currently, the `createChildContext()` method isn't used when denormalizing embedded objects that are not resources. Hence, for example, the `AbstractNormalizer::ATTRIBUTES` won't work for these objects.

This PR fixes that by invoking `createChildContext()` also for nested non-resource objects.

This is closely related to #4258 and #4521 which fixed the same issue when normalizing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## v3.0.4
+
+### Bug fixes
+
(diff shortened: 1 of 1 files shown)


soyuka:
as https://github.com/api-platform/core/pull/4521 broke quite a few implementations I'd like to move this over to the `main` branch with a rebase please.

PR #4634, opened 2022-01-14
fix: leave function to let symfony assert work properly

Nsy:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| License       | MIT

## Current problem this is fixing:

I have an api platform Entity:

```
class Example
{
    #[ORM\Column(type: 'string', length: 255, nullable: false)]
    #[Groups(["write"])]
    #[Assert\NotNull()]
    #[Assert\NotBlank()]
    private $property;
}
```

When I use the POST endpoint /api/examples with a null value on my property
```
{
 "property": null
}
```

The Symfony Asserts are not called/used and instead I have the InvalidArgumentException thrown by the Abstract item normalizer:
```
(400 BAD REQUEST)
{
	"type": "https:\/\/tools.ietf.org\/html\/rfc2616#section-10",
	"title": "An error occurred",
	"detail": "The type of the \"property\" attribute must be \"string\", \"NULL\" given.",
	"trace": [...]
}
```

## Now with this fix

With this fix, the InvalidArgumentException Is not thrown and The Symfony Asserts are called and I have the errors violations message I expected with propertyPath etc:

```
(422 Unprocessables Entity)
{
	"type": "https:\/\/tools.ietf.org\/html\/rfc2616#section-10",
	"title": "An error occurred",
	"detail": "property: Cette valeur ne doit pas être nulle.\nproperty: Cette valeur ne doit pas être vide.",
	"violations": [
		{
			"propertyPath": "property",
			"message": "Cette valeur ne doit pas être nulle.",
			"code": "ad32d13f-c3d4-423b-909a-857b961eb720"
		},
		{
			"propertyPath": "property",
			"message": "Cette valeur ne doit pas être vide.",
			"code": "c1051bb4-d103-4f74-8988-acbcafc7fdc3"
		}
	]
}
```

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -430,6 +430,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
      */
     protected function validateType(string $attribute, Type $type, $value, string $format = null)
     {
+        if (is_null($value)) {
+            return; // Leave the function to let symfony assert work properly
+        }
+
         $builtinType = $type->getBuiltinType();
         if (Type::BUILTIN_TYPE_FLOAT === $builtinType && null !== $format && false !== strpos($format, 'json')) {
             $isValid = \is_float($value) || \is_int($value);


alanpoulain:
The "issue" is more complex than this, it's because we need to denormalize the resource before validating it.
And when using types, it's normal to not be able to deserialize with invalid types and to receive an error.
However an improvement has been made to collect all the denormalization errors in the Serializer component and we should probably use it by default (https://github.com/symfony/symfony/pull/42502).

Nsy:
Thank you for the answer and the Serializer component link.
I was using symfony 5.3. so now i'm upgrading to have the serializer improvements.

I have a question regarding the serializer improvements with api-platform.

I'm not using a custom POST operation (I'm using the default api-platform post) and i have a hard time searching where api-platform is using the serializer improvements for the default operation (and I didn't find any `try catch PartialDenormalizationException`).

This is the exceptionAction I get: https://github.com/api-platform/core/blob/2.6/src/Action/ExceptionAction.php
I failed in finding in the api-platform documentation an explanation about this.

(I don't want to create custom POST/PUT operations Action for each of my Entities. I'm sure there is a clean way to do it properly)

Could you tell me please what is the current best practice / best way to manage this ? 
Thank you again

cedriclombardot:
Hi @alanpoulain, have you got the got way to fix it ?

soyuka:
if you want the property to be denormalized without errors just allow it to be nullable.. Or there's also a flag to disable type enforcements in the normalization context. 
To use a default value in API Platform there's the `defaults` see https://api-platform.com/docs/core/configuration/#global-resources-defaults this will be applied to all resources.

Issue #4614, opened 2022-01-03
syntax error with phpstan starting with 2.6.7 


lazka:
I use phpstan in my project and after the update to 2.6.7 phpstan fails with a "syntax error":

```
 44/44 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ---------------------------------------------------------------------- 
  Line   tests/ApiTest.php                                                     
 ------ ---------------------------------------------------------------------- 
         Internal error: syntax error, unexpected 'static' (T_STATIC)          
         Run PHPStan with --debug option and post the stack trace to:          
         https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md  
 ------ ---------------------------------------------------------------------- 
 ```

The internal error is (I'm using php 7.4):

> Uncaught ParseError: syntax error, unexpected 'static' (T_STATIC) in /home/lazka/my-bundle/vendor/api-platform/core/src/Util/ClientTrait80.php:18

Does anyone know how to work around that?

**Comment by b2p-fred** [comment]
I am currently having the same problem.

**Comment by lazka** [comment]
hm, maybe api-platform should avoid php8-only syntax until php8 is required?

**Comment by norkunas** [comment]
It tries to avoid that https://github.com/api-platform/core/blob/4ec515837cf696788e828a61005594fffc4cf244/src/Util/ClientTrait.php :)

**Comment by lazka** [comment]
yeah, but phpstan parses all referenced files anyway and then fails. And phpstan doesn't provide a way to skip files before parsing, only after parsing :(

**Comment by norkunas** [comment]
How your ApiTest looks like?

**Comment by lazka** [comment]
This is enough to trigger the error:

```php
<?php

declare(strict_types=1);

namespace App\Tests;

use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase;
use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\Client;

class Test extends ApiTestCase
{
    protected function setUp(): void
    {
        parent::setUp();
        static::createClient();
    }
}
```

lazka:
Since https://github.com/api-platform/core/pull/4636#issuecomment-1014414658 says this is a phpstan issue I've opened a bug report with phpstan: https://github.com/phpstan/phpstan/issues/6386

ondrejmirtes:
[...]

PR #4636, opened 2022-01-17
Avoid php8 only syntax when possible

lazka:
| Q             | A
| ------------- | ---
| Branch?       | 2.6 
| Tickets       | #4614 
| License       | MIT

This resulted in phpstan to fail if the active PHP version is <8.0:
Internal error: syntax error, unexpected 'static' (T_STATIC)

In case a project depending on api-platform/core would use ClientTrait directly,
or indirectly, phpstan would try to parse it and fail with the above syntax error.

Move the type declarations into a comment instead.

### DIFF
--- a/src/Util/ClientTrait80.php
+++ b/src/Util/ClientTrait80.php
@@ -15,7 +15,10 @@ namespace ApiPlatform\Core\Util;
 
 trait ClientTrait80
 {
-    public function withOptions(array $options): static
+    /**
+     * @return static
+     */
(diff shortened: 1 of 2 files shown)


norkunas:
But then it's phpstan fault that it cannot understand the code :) I'd suggest for you then to make phpstan happy, override the createClient method and adapt not to use that trait

lazka:
ok, I'll file a bug with phpstan then

Issue #4632, opened 2022-01-13
Cannot use class name for "validationGroups" in ApiResource attribute


mbrodala:
**API Platform version(s) affected**: v2.6.8

**Description**  

When migrating the `@ApiResource` annotation with an `#[ApiResource]` attribute (PHP >= 8.0), the `validationGroups` property does not support a class name anymore. Before this was used to e.g. configure a class implementing the `ValidationGroupsGeneratorInterface` for dynamic validation groups.

Trying to migrate 1:1 results in an error:

```
[TypeError] ApiPlatform\Core\Annotation\ApiResource::__construct(): Argument #42 ($validationGroups) must be of type ?array, string given, called in /var/www/html/src/Entity/<entity>.php on line <line>
```

**How to reproduce**  

Add an `ApiResource` attribute to any class including a validation groups generator class name for `validationGroups`:

```php
#[ApiResource(
    # ...
    validationGroups: MyValidationGroupsGenerator::class,
)]
```

**Possible Solution**  

[The docs suggest](https://api-platform.com/docs/core/validation/#dynamic-validation-groups) to basically bypass the `ApiResource` attribute constructor and directly set the class name via `attributes` instead:

```php
#[ApiResource(
    # ...
    attributes: [
        'validation_groups' => MyValidationGroupsGenerator::class,
    ],
)]
```

From my POV this should have native support, especially since `Validator` did not change here.

alanpoulain:
Thanks for this early testing!
Could you make a PR to allow a string in the metadata?

mbrodala:
@alanpoulain I did a first step in #4640, can you have a look?

soyuka:
Ohh I see, this is possible with API P 2.7 on the main branch though.

PR #4640, opened 2022-01-26
fix: Allow string in ApiResource attribute "validationGroups"

mbrodala:
Technically a union type would be appropriate here but this needs PHP 8

| Q             | A
| ------------- | ---
| Branch?       | 2.6 
| Tickets       | #4632 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1497


mbrodala:
Not sure what to do about the (unrelated?) build errors ... :thinking:

soyuka:
sorry but validation groups is deprecated in favor of a validationContext  in the new metadata and also it is discouraged to use a string here as symfony always works with arrays.

mbrodala:
@soyuka Please have a closer look at #4632 

It's not about allowing a single validation group string VS list of validation group strings. It's about fixing a regression to allow a validation group generator class name again here.

Since I do not know how much longer 2.7/3.0 will take to be released I'd really like to see this fixed in the meantime.

Issue #4643, opened 2022-01-28
[UI] export the SwaggerUIBundle object to allow extensions


b2p-fred:
**Description**  
Considering that:
1. It is allowed to [override the UI template](https://api-platform.com/docs/core/openapi/#overriding-the-ui-template)
2. In the `init-swagger-ui.js` released script, you create an instance of the SwaggerUIBundle object to configure the UI. 

I created a specific Javascript file to manage some extensions in the UI but I cannot get the SwaggerUIBundle object created during the setup... and it would be a non-sens to create a new one as it is still available.

a simple 
```
window.ui = ui;
``` 
at the end of the `window.onload` function would allow to get this object in another script.

A current solution is to replace the `init-swagger-ui.js` distributed script with a modified copy `init-swagger-ui.js` that simply exports the `ui` created object !

dunglas:
+1 on my side. I would call the variable `swaggerUI`.

Do you want to work on a PR?

b2p-fred:
I cloned the repo and I would submit a PR as soon as I have something acceptable :wink:

PR #4363, opened 2021-07-21
Add preserve_empty_objects support

Foxprodev:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | #38192 ?
| License       | MIT
| Doc PR        | 

CollectionNormalizer currently replacing ArrayObject with an empty array

More about main idea discussed here https://github.com/symfony/symfony/pull/28363

Feel free to help me with test case and test location. Thank you :smile:

PR #4644, opened 2022-01-28
#4363 - [UI] export the SwaggerUIBundle object to allow extensions

b2p-fred:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4643 
| License       | MIT
| Doc PR        | n/a 

----
Considering that:
1. It is allowed to [override the UI template](https://api-platform.com/docs/core/openapi/#overriding-the-ui-template)
2. In the `init-swagger-ui.js` released script, the SwaggerUIBundle that configures the UI should be exported to be used in another script file. 

This PR appends a simple 
```
window.swaggerUI = ui;
``` 
at the end of the `window.onload` function to make this object available for another script.

With the same goal, it also exports the `data` object to avoid any other script to parse the JSON one more time -)
```
window.swaggerData = data;
```

### DIFF
--- a/src/Symfony/Bundle/Resources/public/init-swagger-ui.js
+++ b/src/Symfony/Bundle/Resources/public/init-swagger-ui.js
@@ -148,4 +148,8 @@ window.onload = function() {
             }, 10000);
         });
     }
+
+    // Make SwaggerUIBundle and data available for some other scripts
+    window.swaggerUI = ui;
+    window.swaggerData = data;
 };


alanpoulain:
Hello, could you describe a little more why you need to do this? How do you use the exported variables?

b2p-fred:
As described in the issue #4643 I need to use the _SwaggerUIBundle_ object to add some extensions. 

As an example, one can set a pre-authorization header that will be used in the API requests. This is my main intent, adding an extra `<script>` int the Twig template for  doing something like:
```
const ui = window.swaggerUI;
ui.preauthorizeApiKey(jwt, `Bearer 0123.....789`);
```
which will force set an Authorization header with my token.

Anyway, there are several functions that may be invoked on the _SwaggerUIBundle_ object and it is a non sense to create a second instance :wink:

alanpoulain:
OK why not, I didn't know the UI instance had methods like this.
Why the need to export the data object as well? Are you adding extra keys (in the `extraConfiguration` parameter)?

b2p-fred:
The configuration are still fetched from the HTML and parsed in the *init-swagger-ui.js*. 

Exporting the data object is a very small overhead to avoid doing the same job a second time.

soyuka:
I think we should consider having a way to configure the javascript I'm :+1: on this.

Issue #4646, opened 2022-02-01
Support PKCE Authorization Code Grant on Swagger UI


vincentchalamon:
**Description**
API Platform enables an `oauth` option through the Swagger UI OAuth 2.0 support. But when the documentation is exposed on production (which is the case most of the time), the "client_secret" is leaked in the Swagger UI JavaScript code.

According to the [Swagger UI OAuth 2.0 configuration](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/):
> Never use this parameter in your production environment. It exposes cruicial security information. This feature is intended for dev/test environments only.

Alternatively, we could use the PKCE Authorization Code Grant. This would prevent the "client_secret" to be leaked, as far as we don't need the "client" (in this case: Swagger UI) to be fully identified.

**Example**
Add a new config key `usePkceWithAuthorizationCodeGrant` on the bundle configuration and enable it on the `initOAuth` call. This option is available on Swagger UI since 3.24.0.

dunglas:
Shouldn't this be reported directly to Swagger UI? Btw this isn't a prod tool so we respect the spec.

vincentchalamon:
Swagger UI already supports it, thanks to the `usePkceWithAuthorizationCodeGrant` option. But in API Platform, we don't use this option: https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/Resources/public/init-swagger-ui.js#L59-L68

vincentchalamon:
> Btw this isn't a prod tool so we respect the spec.

Does it mean that we should not use the `oauth` API Platform configuration option in production? If so, we should definitively document it and explain why.

dunglas:
+1 to support this option then! Couldn't we find a future-proof way to allow passing any Swagger UI options in the configuration?

PR #4649, opened 2022-02-09
feat: support PKCE on Swagger UI

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4646
| License       | MIT
| Doc PR        | **TODO**

API Platform enables an `oauth` option through the Swagger UI OAuth 2.0 support. But when the documentation is exposed on production (which is the case most of the time), the "client_secret" is leaked in the Swagger UI JavaScript code.

According to the [Swagger UI OAuth 2.0 configuration](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/):
> Never use this parameter in your production environment. It exposes cruicial security information. This feature is intended for dev/test environments only.

Alternatively, we could use the PKCE Authorization Code Grant. This would prevent the "client_secret" to be leaked, as far as we don't need the "client" (in this case: Swagger UI) to be fully identified.

---

- [x] Deprecate `oauth.client_secret`
- [x] Add `oauth.pkce` boolean option (default: true, for security reasons)
- [x] Update SwaggerUI.initOAuth with `usePkceWithAuthorizationCodeGrant`
- [x] Fix tests
- [x] Update CHANGELOG.md file

---

(optional: add a nice picture)
Not related to the current topic, but here is the cutest spider! (sorry for arachnophobes people...)
image


dunglas:
Is it mandatory to deprecate the `client_secret` option? As it's not deprecated in Swagger UI, I suggest reproducing the warning and recommend using PKCE instead.

.github/workflows/ci.yml
```diff
@@ -646,7 +646,7 @@ jobs:
       - name: Require Symfony components
         run: composer require symfony/uid --dev --no-interaction --no-progress --ansi
       - name: Install Elasticsearch-php # there is a version matrix for this package: https://packagist.org/packages/elasticsearch/elasticsearch
-        run: composer require "elasticsearch/elasticsearch:v7.14.0" --no-interaction --no-progress --ansi
+        run: composer require "elasticsearch/elasticsearch:v7.14.0" --no-interaction --no-progress --ansi --with-all-dependencies
```
dunglas, line 649: To do in 2.6 IIUC.

src/Core/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -56,6 +56,7 @@ final class SwaggerUiAction
     private $oauthTokenUrl;
     private $oauthAuthorizationUrl;
     private $oauthScopes;
+    private $oauthPKCE;
```
dunglas, line 59:
```suggestion
    private $oauthPkce;
```

(Same to do everywhere to respect our CS).

src/Core/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -79,9 +80,10 @@ final class SwaggerUiAction
      * @param mixed      $oauthTokenUrl
      * @param mixed      $oauthAuthorizationUrl
      * @param mixed      $oauthScopes
+     * @param mixed      $oauthPKCE
```
dunglas, line 83: bool?

dunglas (changes requested):

.github/workflows/ci.yml
vincentchalamon: The elasticsearch/elasticsearch version supported on 2.6 is not the same and does not conflict with psr/log on 2.6 branch. This issue only occures on the main branch

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -395,6 +395,7 @@ private function registerOAuthConfiguration(ContainerBuilder $container, array $
         $container->setParameter('api_platform.oauth.enabled', $this->isConfigEnabled($container, $config['oauth']));
         $container->setParameter('api_platform.oauth.clientId', $config['oauth']['clientId']);
         $container->setParameter('api_platform.oauth.clientSecret', $config['oauth']['clientSecret']);
+        $container->setParameter('api_platform.oauth.pkce', $config['oauth']['pkce']);
```
dunglas, line 398: IIRC using parameters like this is a deprecated practice since Symfony 3. You can just set the value extracted from the service as a service argument.

dunglas (approved):

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
vincentchalamon: I thought it too, but wanted to be consistent with the other parameters...
Shouldn't we deprecate the other parameters and migrate them to direct injection in 3.0?

dunglas: We can do that yes, but in another PR.

src/Core/Bridge/Symfony/Bundle/Action/SwaggerUiAction.php
```diff
@@ -81,7 +82,7 @@ final class SwaggerUiAction
...
+    public function __construct(ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, $resourceMetadataFactory, NormalizerInterface $normalizer, ?TwigEnvironment $twig, UrlGeneratorInterface $urlGenerator, string $title = '', string $description = '', string $version = '', $formats = [], $oauthEnabled = false, $oauthClientId = '', $oauthClientSecret = '', $oauthType = '', $oauthFlow = '', $oauthTokenUrl = '', $oauthAuthorizationUrl = '', $oauthScopes = [], bool $oauthPkce = false, bool $showWebby = true, bool $swaggerUiEnabled = false, bool $reDocEnabled = false, bool $graphqlEnabled = false, bool $graphiQlEnabled = false, bool $graphQlPlaygroundEnabled = false, array $swaggerVersions = [2, 3], OpenApiSwaggerUiAction $swaggerUiAction = null, $assetPackage = null, array $swaggerUiExtraConfiguration = [])
```
dunglas, line 85: This is a BC break. The new argument must be added at the end of the list.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -401,6 +401,13 @@ private function registerOAuthConfiguration(ContainerBuilder $container, array $
         $container->setParameter('api_platform.oauth.authorizationUrl', $config['oauth']['authorizationUrl']);
         $container->setParameter('api_platform.oauth.refreshUrl', $config['oauth']['refreshUrl']);
         $container->setParameter('api_platform.oauth.scopes', $config['oauth']['scopes']);
+
+        if ($container->hasDefinition('api_platform.swagger.action.ui')) {
+            $container->getDefinition('api_platform.swagger.action.ui')->setArgument('$oauthPkce', $config['oauth']['pkce']);
```
dunglas, line 406: If possible it's better to not use the `'$foo'` syntax as it involved reflection and slows down a bit the build process.

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -260,7 +260,11 @@ private function addOAuthSection(ArrayNodeDefinition $rootNode): void
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('clientId')->defaultValue('')->info('The oauth client id.')->end()
-                        ->scalarNode('clientSecret')->defaultValue('')->info('The oauth client secret.')->end()
+                        ->scalarNode('clientSecret')
+                            ->defaultValue('')
+                            ->info('The oauth client secret. Never use this parameter in your production environment. It exposes crucial security information. This feature is intended for dev/test environments only. Enable `oauth.pkce` instead')
```
dunglas, line 265:
```suggestion
                            ->info('The OAuth client secret. Never use this parameter in your production environment. It exposes crucial security information. This feature is intended for dev/test environments only. Enable "oauth.pkce" instead')
```

dunglas (changes requested):

soyuka (approved):

PR #4655, opened 2022-02-14
docs(adr): refactor state management

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Proposal to simplify the codebase and give more flexibility to users by leveraging the new  "state processor" and "state provider" concept to get rid of our custom Symfony kernel listeners and of the GraphQL resolvers.

### DIFF
--- a/docs/adr/0005-refactor-state-management.md
+++ b/docs/adr/0005-refactor-state-management.md
@@ -1,0 +1,46 @@
+# Refactor State Management
+
+* Status: proposed
+* Deciders: @dunglas, @soyuka, @alanpoulain
+
+## Context and Problem Statement
+
+Since version 2.2, we support both REST and GraphQL API styles out of the box.
+This leads to some code duplication to extract states from request bodies (write requests),
+and to generate resource representations from states retrieved from the data source.
+
+The REST susbystem uses [Symfony-specific kernel event listeners](https://api-platform.com/docs/core/events/#built-in-event-listeners)
+while the GraphQL subsystem relies on [an ad-hoc resolver system](https://api-platform.com/docs/core/graphql/#workflow-of-the-resolvers).
+
+
+Also, while it's possible [to use API Platform as a standalone PHP library](https://api-platform.com/docs/core/bootstrap/),
+this requires writing boilerplate code doing mostly what is done in the Symfony-specific listeners.
+As we're working on integrating API Platform with Laravel, the Laravel package will contain similar code to the one in the Symfony
+event listeners too.
+
+For API Platform 3, we refactored the whole metadata susbsytem to be more flexible, powerful, and covering more use cases.
+This led to the refactoring of the two main interfaces allowing to plug a data source in API Platform: the state provider and the state processor interfaces.
+
+Leveraging these new interfaces, it should be possible to simplify the code base and to remove most code duplication by transforming most of the code currently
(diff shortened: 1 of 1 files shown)


docs/adr/0005-refactor-state-management.md
```diff
@@ -0,0 +1,46 @@
...
+2. Remove the corresponding listeners.
+3. Replace `PlaceholderController` by a controller calling the main `StateProcessor` and `StateProvider` objects directly.
+4. Remove the GraphQL resolvers, call the main `StateProcessor` and StateProvider objects directly.
+
+Consenquently, transforming the raw request body (e.g. the raw JSON document) to a PHP data structure will be the responsibility of a processor.
+The default one will use the Symfony Serializer component to do so, but this will give the opportunity to the user to replace this class by one using another Serializer if necessary.
+This will also allow the user to access the raw body if necessary, and will enable a whole class of optimizations, extra validations (e.g. validating a raw JSON string against a JSON Schema) etc.
+Similarly, transforming PHP data structures into strings to be stored in response bodies will now be the responsibility of a state provider.
```
soyuka, line 37:
```suggestion
Consenquently, transforming the raw request body (e.g. the raw JSON document) to a PHP data structure will be the responsibility of a provider.
The default one will use the Symfony Serializer component to do so, but this will give the opportunity to the user to replace this class by one using another Serializer if necessary.
This will also allow the user to access the raw body if necessary, and will enable a whole class of optimizations, extra validations (e.g. validating a raw JSON string against a JSON Schema) etc.
Similarly, transforming PHP data structures into strings to be stored in response bodies will now be the responsibility of a state processor.
```

Today, providers are used to read data from an URI. The DeserializeListener should be a DeserializeProvider that transforms request body to a PHP data structure.

Today, processors handles the data update or data removal. The SerializeProcessor could transform the resource to a PHP data structure. 

As Martin Fowler writes in the _Patterns of Enterprise Application Architecture_:

> There’s just one controller, so you can easily enhance its behavior at
runtime with decorators [Gang of Four](https://en.wikipedia.org/wiki/Design_Patterns). You can have decorators for
authentication, character encoding, internationalization, and so forth,
and add them using a configuration file or even while the server is
running. ([Alur et al.](http://www.corej2eepatterns.com/InterceptingFilter.htm) describe this approach in detail under the name
Intercepting Filter.)

We could see these decorators as a chain of intercepting filters where:
  - Providers are Pre-Request event listeners
  - Processors are Post-request event listeners

In graphql as well we could use these decorator chains as well:

- `DeserializeStage` - `DeserializeProvider`
- `ReadStage` - `ReadProvider`
- `SerializeStage` - `SerializeProcessor`
- `ValidateStage` - `ValidateProvider`
- `WriteStage` - `WriteProcessor`

- `SecurityPostDenormalizeStage`
- `SecurityPostValidationStage`
- `SecurityStage` - `SecurityProvider`

For security, DeserializeProvider and ValidationProvider will call a decorated SecurityProvider.

soyuka (changes requested):

lchrusciel:
I love that change! An additional benefit of the new architecture will be the increased DX, as this structure will be much nicer to debug/examine with xdebug as it is explicit and straight forward.

Issue #2769, opened 2019-04-27
Doctirne cacheable extensions 


CvekCoding:
I use extensions quite intensive, because I use them to limit users into the proper scopes (e.g. user cant see the data from the other company and even in his company he can see the limited data based on permissions). I noticed, that in many (all?) cases when ApiP prepares query to DB, it goes through all the extensions and apply each one.
May be it's a good idea to create `CacheableExtension` interface with `supports` method to optimize the process?

er1z:
Shouldn't you restrict access using voters?

dunglas:
@er1z for collections it's better (for performance and security) to filter at the DBMS level, and to retrieve only relevant data. It's also the only way to have a working pagination.

@Siregacvek could be a nice improvement indeed, it could take some inspiration from the similar interface of the Symfony Serializer.

Siregacvek:
@er1z I use voters for CUD operations. For Read I use extensions as a recommended way. BTW Symfony voters has no cache too - I feel this could a very good Symfony PR-improvement and may increase performance in big systems.
@dunglas yes, I meant the similar approach like in Symfony Serializer. I'll try to deploy PR if I find a time to create it. Thanks.

soyuka:
closing as old

PR #4664, opened 2022-02-22
feat(extension): add "supports" cacheable extension behavior

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #2769 
| License       | MIT
| Doc PR        | WIP

When retrieving items or collection through a DataProvider, all extensions are applied. These extensions can be dedicated to a resource class and/or operation name.

In order to optimize process, an interface with 2 methods has been added. It is inspired by https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Serializer/Normalizer/CacheableSupportsMethodInterface.php : 
* `supports` to check if the extension should be applied.
* `hasCacheableSupportsMethod` to check if the result of `supports` can be cached by resource name and operation name.

Before implementing tests for each DataProvider and create a doc PR, I would like to know your opinion about this PR.

### DIFF
--- a/src/Core/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
+++ b/src/Core/Bridge/Doctrine/MongoDbOdm/CollectionDataProvider.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\Bridge\Doctrine\MongoDbOdm;
 
 use ApiPlatform\Core\DataProvider\CollectionDataProviderInterface;
 use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
+use ApiPlatform\Core\Extension\CacheableSupportsExtensionTrait;
 use ApiPlatform\Core\Metadata\Resource\Factory\ResourceMetadataFactoryInterface;
 use ApiPlatform\Core\Metadata\Resource\ResourceMetadata;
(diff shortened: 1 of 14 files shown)


dunglas:
Do you have some numbers or a Blackfire profile showing the performance gain?

Also, if we add this, this should be done in the new state providers. Data providers are deprecated in 2.7 and will be removed in 3.0, so let's keep them as-is.

Jean-Beru:
> Do you have some numbers or a Blackfire profile showing the performance gain?
Not really because it depends on how extensions are implemented and their usage. Caching `supports` result will improve performances if this extension does not have to be applied.

So, does thehttps://github.com/api-platform/core/issues/2769 issue still relevant ?

@CvekCoding Do you have some code to share where caching `supports` could improve performances ?

soyuka:
Hi, these are interesting changes but I'd also like to see some performances improvements before adding classes we need to then maintain. Also Extensions are doctrine-specific and can not be added to the root namespace. As @dunglas said, these interfaces are deprecated anyways and if we can prove there's a real benefit we can add them to the `main` branch over the new `Provider` interfaces. 

I'll keep this open pending discussion.

Jean-Beru:
Hi @soyuka,

I updated my PR and made some Blackfire profile by calling `provide` method multiple times but the gain is not terrible. Use cases from #2769 would be interesting.

PR #4670, opened 2022-03-08
SearchFilter : fix MsSQL issue on pure numeric values

tehla:
Fix the following issue: 
```
SQLSTATE[22018]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Conversion failed when converting the varchar value '%' to data type int."
```

| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

### DIFF
--- a/src/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Doctrine/Orm/Filter/SearchFilter.php
@@ -228,7 +228,10 @@ class SearchFilter extends AbstractContextAwareFilter implements SearchFilterInt
         }
 
         $queryBuilder->andWhere($queryBuilder->expr()->orX(...$ors));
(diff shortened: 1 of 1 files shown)


soyuka:
IMO the fix is wrong, Doctrine has a Type implementation that has `convertToPHPValue` which should be used here. Also, we recommend to use the numeric filter to filter numeric values, the search filter is used for strings. If this doesn't work for you please implement your own filter.

tehla:
Thanks, I will take a look to convertToPHPValue.
But IMO, the fix was still relevant, because a number can be a part of a string. And what if I want to filter on a specific number value into a string ? (example : all my documents having the 2022 number in their title)

tehla:
And I should add : the search filter is looking for string into string fields. You MUST be sure that your data is casted as a string. 
42 !== '42'

PR #4695, opened 2022-03-29
feat(http_cache): improve xkey purger implementation

soyuka:
doc: https://github.com/api-platform/docs/pull/1524


src/Core/HttpCache/PurgerInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ * @experimental
+ */
+interface PurgerInterface
+{
+    /**
+     * Purges all responses containing the given resources from the cache.
+     *
+     * @param string[] $iris
+     */
+    public function purge(array $iris);
```
alanpoulain, line 30: Why not:
```suggestion
    public function purge(array $iris): void;
```

Issue #4694, opened 2022-03-29
[BC Break] 2.7 has undocumented bc breaks cause of namespace changes (e.g. by removal of 'Doctrine\Bridge')


robertfausk:
**API Platform version(s) affected**: `2.7`/`dev-main`

**Description**  
As @laryjulien stated in https://github.com/api-platform/core/pull/4556#issuecomment-970048404
> FQCN (fully qualified class name) are **not** updated in CHANGELOG.md

see e.g.:
https://github.com/api-platform/core/commit/10827bb570c44cdee2c1bb7a307a8a18df9b4f00
![image](https://user-images.githubusercontent.com/1651297/160676714-0398ca8c-676d-495f-8d47-86100f93428d.png)

**How to reproduce**  

Implement a class to filter subresources according to solution in https://github.com/api-platform/core/issues/2253#issuecomment-445826669 and update from `2.6.8` to `2.7`/`dev-main`.

This will result in `Type error: App\Doctrine\DefaultOrderExtensionDecorator::__construct(): Argument #1 ($decorated) must be of type ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\OrderExtension, ApiPlatform\Doctrine\Orm\Extension\OrderExtension given`. See example code below.

Changing `use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\OrderExtension;` to `use ApiPlatform\Doctrine\Orm\Extension\OrderExtension;` fixes the error in the example code.

```YAML
#services.yml
    App\Doctrine\DefaultOrderExtensionDecorator:
        decorates: api_platform.doctrine.orm.query_extension.order
        arguments:
            $decorated: '@App\Doctrine\DefaultOrderExtensionDecorator.inner'
            $orderParameterName: '%api_platform.collection.order_parameter_name%'
```
```PHP
<?php
declare(strict_types=1);

namespace App\Doctrine;

use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\ContextAwareQueryCollectionExtensionInterface;
use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\OrderExtension;
use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
use App\Entity\Event;
use Doctrine\ORM\QueryBuilder;

final class DefaultOrderExtensionDecorator implements ContextAwareQueryCollectionExtensionInterface
{
    private OrderExtension $decorated;

    private string $orderParameterName;

[...]

PR #4696, opened 2022-03-31
fix(deprecation): alias before autoload if needed

soyuka:
fixes https://github.com/api-platform/core/issues/4694


src/Core/Action/EntrypointAction.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Core\Action;
+
+class_exists(\ApiPlatform\Action\EntrypointAction::class);
+
+if (false) {
+    class EntrypointAction extends \ApiPlatform\Action\EntrypointAction
```
azjezz, line 19: as mentioned in the internal discussion, since `ApiPlatform\Action\EntrypointAction` is final, the deprecated alias ( `ApiPlatform\Core\Action\EntrypointAction` ) needs to also be declared final so static analysis tools would be able to warn users when they try to extend from it. ( same applies for all other final classes ) 

```suggestion
    final class EntrypointAction extends \ApiPlatform\Action\EntrypointAction
```

PR #4588, opened 2021-12-08
feat(test): add Client::loginUser()

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

Adds a helper method to log a user in the tests:

```php
use ApiPlatform\Symfony\Bundle\Test\ApiTestCase;

class MyTest extends ApiTestCase
{
    public function testLoginUser(): void
    {
        $client = self::createClient();
        $client->loginUser(
            self::getContainer()->get('my_user_repoistory')->findByUsername('kevin')
        );

        $client->request('GET', '/secured_dummies');
        $this->assertResponseIsSuccessful();
    }
}
```

PR #4701, opened 2022-04-03
Remove @internal annotation from Client::getKernelBrowser()

gnito-org:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | none
| License       | MIT
| Doc PR        | none

Remove the @internal annotation from the test Client::getKernelBrowser so that other features of the KernelBrowser, such as user login, can be used without PHPStorm complaining about the usage of an internal method.


soyuka:
we introduced https://github.com/api-platform/core/pull/4588 for this

Issue #4699, opened 2022-04-02
[Bug][v3][Pagination] V3 pagination limits seem to be broken


silverbackdan:
**API Platform version(s) affected**: dev-master - v3.0 (no backwards compatibility)

**Description**  
On dev-master without backwards compatibility pagination seems to be broken.

**How to reproduce**  
I have a behat test running:
```
Scenario Outline: Pagination parameter is configured
    Given there are 120 DummyResourceWithPagination resources
    When I send a "GET" request to "/dummy_resource_with_paginations<postfix>"
    Then the response status code should be 200
    And the JSON node "hydra:member" should have "<total>" elements
    Examples:
      | total  | postfix           |
      | 10     |                   |
      | 20     | ?perPage=20       |
      | 40     | ?perPage=110      |
      | 120    | ?pagination=false |
      | 120    | ?perPage=0        |
```

and the attributes on the resource setup as 
```php
#[ApiResource(
    paginationClientEnabled: true,
    paginationClientItemsPerPage: true,
    paginationItemsPerPage: 10,
    paginationMaximumItemsPerPage: 40
)]
#[Get]
#[GetCollection]
#[ORM\Entity]
class DummyResourceWithPagination
{
    use IdTrait;
}
```

The test for pagination set to 110 by the client should return just 40, but returns 100. perPage 0 is returning 0 results, I think it defaulted to disabling pagination before.. but perhaps I'm mistaken and the only bug is allowing a user to override the max items per page.

**Possible Solution**  
not sure yet.

**Additional Context**  
I have set the config `api_platform.collection.pagination.items_per_page_parameter_name` to `perPage`

silverbackdan:
I'm investigating this still.. It seems the ORM PaginationExtension does not find the request in my test to apply all of the logic and the request is instead passed to `ApiPlatform\State\Pagination\Pagination` - I am just making a dummy GET request in Behat to `/dummy_resource_with_paginations?perPage=110` so something here is a little strange - I've a complicated application so I'm working through to see if there's something else interfering somehow.

silverbackdan:
So it appears no matter what I do with my attributes as shown here:
```
<?php

/*
 * This file is part of the Silverback API Components Bundle Project
 *
 * (c) Daniel West <daniel@silverback.is>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

declare(strict_types=1);

[...]

PR #4702, opened 2022-04-04
fix(metadata): defaults should not override values 

silverbackdan:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4699
| License       | MIT
| Doc PR        | n/a
Fix: https://github.com/api-platform/core/issues/4699


features/hydra/collection.feature
```diff
@@ -273,6 +273,44 @@ Feature: Collections support
     }
     """
 
+  @!mongodb
```
alanpoulain, line 276: Why not MongoDB too?

src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -170,7 +170,11 @@ private function getOperationWithDefaults(ApiResource $resource, $operation): ar
     {
         foreach ($this->defaults['attributes'] as $key => $value) {
             [$key, $value] = $this->getKeyValue($key, $value);
-            if (method_exists($operation, 'get'.ucfirst($key)) && null !== $operation->{'get'.ucfirst($key)}()) {
+
+            $upperProperty = ucfirst($key);
+            $getter = "get$upperProperty";
+
+            if (method_exists($operation, $getter) && null === $operation->{$getter}()) {
```
alanpoulain, line 177: I think just changing the condition would have been alright.

tests/Fixtures/TestBundle/Entity/PaginationEntity.php
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @ORM\Entity
+ */
+#[ApiResource]
+#[Get]
+#[GetCollection(
+    paginationClientItemsPerPage: true,
+    paginationItemsPerPage: 5,
+    paginationMaximumItemsPerPage: 30
+)]
```
alanpoulain, line 32: You cannot use PHP 8 attributes directly in the tests, you need to use annotations like before, they will be automatically converted to PHP attributes by our command (`api:upgrade-resource`).

soyuka: you can you just need to flag them as `@php8` tests :)

features/hydra/collection.feature
soyuka: because it's testing the metadata which doesn't impact mongodb or doctrine in particular. + it reduces time to test

silverbackdan:
Thank you both :)

PR #4711, opened 2022-04-12
test date filter with invalid query

alexislefebvre:
| Q             | A
| ------------- | ---
| Branch?       | I don't know yet
| Tickets       | I don't know yet
| License       | MIT

I wanted to filter data by several ranges of date.

I tried to use `?dateIncludeNullBeforeAndAfter[before][]=2015-04-01`, it doesn't work. This is the expected result since this feature doesn't exist.

The problem is that it threw an exception:

> Type error: Argument 6 passed to ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\DateFilter::addWhere() must be of the type string, array given, called in vendor/api-platform/core/src/Bridge/Doctrine/Orm/Filter/DateFilter.php on line 88 (Behat\Testwork\Call\Exception\FatalThrowableError)

So I added a test to try to replicate this.

Maybe a type check should be added to avoid this exception.

### DIFF
--- a/features/doctrine/date_filter.feature
+++ b/features/doctrine/date_filter.feature
@@ -507,6 +507,14 @@ Feature: Date filter on collections
     And the JSON node "hydra:member[1].dateIncludeNullBeforeAndAfter" should be null
 
   @createSchema
+  Scenario: Get collection filtered by wrong filter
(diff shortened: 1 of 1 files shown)


alexislefebvre:
It crashes with the 2.4 version that I use, but not on the `main` branch, so it looks the issue has been fixed.

PR #4712, opened 2022-04-12
fix(state): operation argument in provide/process

soyuka:
Review the state provider and processor interfaces. Move provider/processor on the Operation.


src/Doctrine/Orm/Metadata/Resource/DoctrineOrmResourceCollectionMetadataFactory.php
```diff
@@ -0,0 +1,100 @@
...
+
+use ApiPlatform\Metadata\AbstractOperation;
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\DeleteOperationInterface;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Persistence\ManagerRegistry;
+
+class DoctrineOrmResourceCollectionMetadataFactory implements ResourceMetadataCollectionFactoryInterface
```
dunglas, line 24:
```suggestion
final class DoctrineOrmResourceCollectionMetadataFactory implements ResourceMetadataCollectionFactoryInterface
```

src/Doctrine/Orm/Metadata/Resource/DoctrineOrmResourceCollectionMetadataFactory.php
```diff
@@ -0,0 +1,100 @@
...
+{
+    private $decorated;
```
dunglas, line 26: missing types?

src/Doctrine/Orm/Metadata/Resource/DoctrineOrmResourceCollectionMetadataFactory.php
```diff
@@ -0,0 +1,100 @@
...
+            $operation = $operation->withProcessor($this->getProcessor($operation));
+        }
+
+        return $operation;
+    }
+
+    private function getProvider(AbstractOperation $operation): string
+    {
+        if ($operation instanceof CollectionOperationInterface) {
+            return 'api_platform.doctrine.orm.state.collection_provider';
```
dunglas, line 86: We should alias these services with the FQCN, and use it here. This will remove the hard coupling with Symfony.

src/Metadata/AbstractOperation.php
```diff
@@ -0,0 +1,776 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata;
+
+abstract class AbstractOperation
```
dunglas, line 16:
```suggestion
abstract class Operation
```

The API will be nicer, and this is consistent with the newest coding convention used in Symfony.

src/State/CallableProcessor.php
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\State;
+
+use ApiPlatform\Core\Exception\RuntimeException;
+use ApiPlatform\Metadata\AbstractOperation;
+use Psr\Container\ContainerInterface;
+
+class CallableProcessor implements ProcessorInterface
```
dunglas, line 20:
```suggestion
final class CallableProcessor implements ProcessorInterface
```

src/State/LegacyDataProviderState.php
```diff
@@ -40,14 +41,14 @@ public function __construct(ResourceMetadataCollectionFactoryInterface $resource
         $this->resourceMetadataCollectionFactory = $resourceMetadataCollectionFactory;
     }
 
-    public function provide(string $resourceClass, array $identifiers = [], ?string $operationName = null, array $context = [])
+    public function provide(AbstractOperation $operation, array $uriVariables = [], array $context = [])
     {
-        $operation = $context['operation'] ?? null;
+        exit('hello');
```
dunglas, line 46: 🔨

src/Symfony/State/ServiceLocatorProcessor.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Symfony\State;
+
+use ApiPlatform\Metadata\AbstractOperation;
+use ApiPlatform\State\ProcessorInterface;
+
+class ServiceLocatorProcessor implements ProcessorInterface
```
dunglas, line 19:
```suggestion
final class ServiceLocatorProcessor implements ProcessorInterface
```

And as this class doesn't depend of Symfony, it could be placed in the top-level namespace.

dunglas (changes requested):

src/Doctrine/Orm/Metadata/Resource/DoctrineOrmResourceCollectionMetadataFactory.php
soyuka: I tried just aliasing and it doesn't work with tags + service locators this was a quick fix I need to use the class as id

src/State/LegacyDataProviderState.php
soyuka: on purpose I need to fix this

src/Metadata/AbstractOperation.php
soyuka: I know right, but we need to rename the current one then

dunglas: Not an issue as this was never released isn't it?

Issue #4628, opened 2022-01-12
[ES] deprecated document type


darthf1:
Hi!

Since `ES 7.0` the providing a documents `_type` field in a request is deprecated.
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-type-field.html

This leads to several deprecating warnings:
![image](https://user-images.githubusercontent.com/17253332/149142795-25f91616-7872-470f-ade0-0f3c395d4a4a.png)

If I read the [schedule](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#_schedule_for_removal_of_mapping_types) correctly:
- ES 5: multiple document types per index allowed
- ES 6: only 1 document type per index is allowed
- ES 7: defining `_types` in a request is deprecated
- ES 8: option will be removed.

Would you accept a PR removing everything related to a documents `_type` altogether? This impacts the version of ES you want to support in the new APIP major version, so I'm not sure what your thoughts are about that.

PR #4720, opened 2022-04-20
fix(elasticsearch): remove deprecated type

darthf1:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | [#... ](https://github.com/api-platform/core/issues/4628)
| License       | MIT

This is just a small change to remove the `type` from the request, as it is deprecated (see issue). This is needed if you want to support Elasticsearch 8.0.

Let me know how you want to continue, and ill do the work and fix any failed tests. 
- Can i put a composer conflict on Elasticsearch < 6.0 (ES 5 is [EOL](https://www.elastic.co/support/eol) since 2019-03-11)
- Can i remove all references to `type` in the [DocumentMetaData](https://github.com/api-platform/core/blob/main/src/Elasticsearch/Metadata/Document/DocumentMetadata.php) and all classes which decorate this class?

### DIFF
--- a/src/Elasticsearch/State/CollectionProvider.php
+++ b/src/Elasticsearch/State/CollectionProvider.php
@@ -75,7 +75,6 @@ final class CollectionProvider implements ProviderInterface
 
         $documents = $this->client->search([
             'index' => $documentMetadata->getIndex(),
-            'type' => $documentMetadata->getType(),
             'body' => $body,
         ]);
 
--- a/src/Elasticsearch/State/ItemProvider.php
+++ b/src/Elasticsearch/State/ItemProvider.php
@@ -54,7 +54,6 @@ final class ItemProvider implements ProviderInterface
         try {
             $document = $this->client->get([
                 'index' => $documentMetadata->getIndex(),
-                'type' => $documentMetadata->getType(),
                 'id' => (string) reset($uriVariables),
             ]);
         } catch (Missing404Exception $e) {


darthf1:
Continuing in https://github.com/api-platform/core/pull/4726

darthf1:
> I have a plan for Elasticsearch any help would be really appreciated. Basically I want to remove this DocumentMetadata in favor of extending the `ApiPlatform\Metadata\Operation` in `ApiPlatform\Elasticsearch\Metadata\Operation`. This would simplify filters a lot.

I'm not yet familiar with the new structure of Api Platform. So that's a no at the moment but ofcourse this can change in the near future :).

> 
> About the Type issue please find a way to know the version of Elasticsearch we're using and add a condition, for example:
> 
> ```
> if (ES_VERSION < 8) {
>   // keep type
> }
> ```

See https://github.com/api-platform/core/pull/4726

> 
> using 7 and 8 in the CI matrix would be awesome!

I'm on gitlab myself, so I cant help you with the github ci unfortunately.

darthf1:
> I have a plan for Elasticsearch any help would be really appreciated. Basically I want to remove this DocumentMetadata in favor of extending the `ApiPlatform\Metadata\Operation` in `ApiPlatform\Elasticsearch\Metadata\Operation`. This would simplify filters a lot.

@soyuka is this this planned? I'm asking because I need additional business logic to return the correct `DocumentMetadata` (the index name is dynamic based on business logic). 

Currently, only the `string $resourceClass` is provided as a parameter for the [create function](https://github.com/api-platform/core/blob/main/src/Elasticsearch/Metadata/Document/Factory/DocumentMetadataFactoryInterface.php#L33), so at this moment I'm injecting the `RequestStack $requestStack` into my `final class DocumentMetadataFactory implements DocumentMetadataFactoryInterface`, but this makes testing very difficult.

So for me, the solution would be to
- either add the `array $context = []` as an additional parameter.
- have the current operation in the function

The first one is the easiest, for which I can do the PR. The second one could be better, but I'm not sure about what you want to do with / how you want to do your solution.

PR #4722, opened 2022-04-20
fix: deprecate data transformers

soyuka:


norkunas:
And what does it fix? :slightly_smiling_face: Imho doesn't make sense in one major to introduce feature and in another remove it

tacman:
Fortunately, data transformers still exist, but the pattern is deprecated.  According to the deprecation message: "use a Provider or a Processor and either use your input or return a new output there."

soyuka:
> And what does it fix? slightly_smiling_face Imho doesn't make sense in one major to introduce feature and in another remove it

Most of the cases where people struggle is related to this pattern. It was wrong from the beginning and instead we advise to handle the case in a Processor (see https://github.com/api-platform/core/discussions/4892#discussioncomment-3529590). 

Input/Output is still available and now every object (anonymous or ApiResource-tagged is considered as an http Resource).

PR #4734, opened 2022-04-28
feat: improve iri converter interface

soyuka:
Removes a method we introduced for "collection" iris, we're back to the original interface:

```php
<?php

/*
 * This file is part of the API Platform project.
 *
 * (c) Kévin Dunglas <dunglas@gmail.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

declare(strict_types=1);

namespace ApiPlatform\Api;

use ApiPlatform\Exception\InvalidArgumentException;
use ApiPlatform\Exception\ItemNotFoundException;
use ApiPlatform\Exception\RuntimeException;

/**
 *
 */
interface IriConverterInterface
{
    /**
     * Retrieves an item from its IRI.
     *
     * @throws InvalidArgumentException
     * @throws ItemNotFoundException
     *
     * @return object
     */
    public function getResourceFromIri(string $iri, array $context = [], ?Operation $operation = null);

    /**
     * Gets the IRI associated with the given item.
     *
     * @param object|class-string $resource
     *
     * @throws InvalidArgumentException
     * @throws RuntimeException
     */
    public function getIriFromResource($resource, int $referenceType = UrlGeneratorInterface::ABS_PATH, ?Operation $operation = null, array $context = []): ?string;
}
```

The interesting diff is available at: https://github.com/api-platform/core/compare/main...soyuka:feat/iri-converter?diff=split#diff-ac4847d931ec39d251d7b28fc138a9c65b7fb2f981fe313d9e373a2c2d6b8860

Basically this means:
- easier to implement
- easier to decorate (in the future we will provide an IriConverter that doesn't use symfony's routing system and one for Skolem IRIs #4731)
- Operation is already used in our Providers and Processors this is in the continuity of the work initialized there
- better vocabulary

### DIFF
--- a/src/Api/IriConverterInterface.php
+++ b/src/Api/IriConverterInterface.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\Api;
 use ApiPlatform\Exception\InvalidArgumentException;
 use ApiPlatform\Exception\ItemNotFoundException;
 use ApiPlatform\Exception\RuntimeException;
+use ApiPlatform\Metadata\Operation;
 
 /**
  * Converts item and resources to IRI and vice versa.
@@ -32,7 +33,7 @@ interface IriConverterInterface
      *
      * @return object
      */
-    public function getItemFromIri(string $iri, array $context = []);
+    public function getItemFromIri(string $iri, array $context = [], ?Operation $operation = null);
 
     /**
      * Gets the IRI associated with the given item.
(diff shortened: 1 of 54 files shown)


src/Api/IriConverterInterface.php
```diff
@@ -42,12 +43,5 @@ public function getItemFromIri(string $iri, array $context = []);
...
      */
-    public function getIriFromItem($item, string $operationName = null, int $referenceType = UrlGeneratorInterface::ABS_PATH, array $context = []): string;
-
-    /**
-     * Gets the IRI associated with the given resource collection.
-     *
-     * @throws InvalidArgumentException
-     */
-    public function getIriFromResourceClass(string $resourceClass, string $operationName = null, int $referenceType = UrlGeneratorInterface::ABS_PATH, array $context = []): string;
+    public function getIriFromItem($item = null, ?Operation $operation = null, int $referenceType = UrlGeneratorInterface::ABS_PATH, array $context = []): ?string;
```
dunglas, line 46: How can `$item` be null?

PR #4741, opened 2022-05-01
fix: json encoder invalid utf8 characters

divine:
Co-Authored-By: AntonioCS <1086401+antoniocs@users.noreply.github.com>

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4416 
| License       | MIT

This is a follow-up for #4416 with the added test.

Test for PHP 7.1 failing because of:

```
JSON_INVALID_UTF8_IGNORE (int)
Ignore invalid UTF-8 characters. Available as of PHP 7.2.0.
```


KDederichs:
You might want to add that conditionally then with PHP_VERSION (https://www.php.net/manual/de/reserved.constants.php)

Issue #4753, opened 2022-05-13
3.0 backward compatibility notes 


soyuka:
While we work on API Platform 3.0 we found out some deprecations were forgotten this issue is to note what we need to do:

- [ ] SchemaFactoryInterface BC
- [ ] deprecate Doctrine FilterInterface and ContextAwareFilterInterface for a single interface

PR #4758, opened 2022-05-16
fix(json-schema): factory backward compatibility layer

soyuka:
link #4753


bendavies:
@soyuka this seems to cause a break between v2.6 and v2.7
specifically
```php
$this->assertMatchesResourceItemJsonSchema(PolicyNotice::class);
```
works on 2.6 but breaks on 2.7 because 2.7 tries to "guess" the operation and can't find one, so creates a new one with the wrong  (empty) serialization context.

https://github.com/api-platform/core/blob/4c16efe1f39c8aadd20e8e08b0b57a4d33034cd8/src/JsonSchema/SchemaFactory.php#L284

in 2.7 i have to do this to fix it:

```php
        $this->assertMatchesResourceItemJsonSchema(PolicyNotice::class, '_api_/policy-notices/{id}.{_format}_get');
``` 

this seems, wrong to me.

can you provide any info on what i should be doing?
https://github.com/api-platform/core/blob/4c16efe1f39c8aadd20e8e08b0b57a4d33034cd8/src/JsonSchema/SchemaFactory.php#L284

PR #4760, opened 2022-05-17
fix(metadata): add read and write to extractor

dannyvw:
Read and write attributes are missing for xml and yaml extractor


dunglas (approved):

soyuka:
I'm thinking that there might me more missing attributes I think that we should just check that everything in `ApiPlatform\Metadata\ApiResource` is available through yaml/xml

PR #4766, opened 2022-05-20
fix(graphql): output creates its own type in TypeBuilder

### LINKED ISSUE #3155: [GraphQL] Setting "output" on a mutation doesn't work
I think the schema is not being generated properly (we still get the "normal" fields, not the fields of the output class / resource).

https://github.com/api-platform/core/blob/e468d98d260e0cf49d63d088fd5678ecab1b8912/src/GraphQl/Type/TypeBuilder.php#L90-L95

I think we might need to either also set the `$resourceMetadata` to the new class, or handle `$wrapData` differently.

### LINKED ISSUE #3647: Output DTO in GraphQl schema
**API Platform version(s) affected**: 2.5.6

**Description**  
If class has `output` property then in schema it will have all the fields of output class (which is ok), but type of output object will be the original class. Is this by design or it's a bug?

**How to reproduce**  
```
/**
 * @ApiResource(
 *     output=Bar::class,
 *     collectionOperations={"get"},
 *     itemOperations={"get"},
 *     graphql={
 *         "item_query",
 *         "collection_query",
 *     }
 * )
 *
 */
class Foo {public $foo}

/**
 * @ApiResource(
 *     collectionOperations={"get"},
 *     itemOperations={"get"},
 *     graphql={
 *         "item_query",
 *         "collection_query",
 *     }
 * )
 *
 */
class Bar {public $bar}
```
Then in the schema the collection of `Foo` objects will have field `$bar` (it's ok), but the type of edge node will be still `Foo`.
```
{
  foos {
    edges {
      node {
        bar
        __typename // here will be Foo
      }
    }
  }
}
```

alanpoulain:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Fixes #3155, fixes #3647
| License       | MIT
| Doc PR        | Explicit the need to have a DataTransformer and not do it in the resolver

### DIFF
--- a/features/graphql/mutation.feature
+++ b/features/graphql/mutation.feature
@@ -819,6 +819,26 @@ Feature: GraphQL mutation support
     And the JSON node "data.testCustomArgumentsDummyCustomMutation.dummyCustomMutation.result" should be equal to "18"
     And the JSON node "data.testCustomArgumentsDummyCustomMutation.clientMutationId" should be equal to "myId"
 
+  Scenario: Execute a custom mutation with output
+    When I send the following GraphQL request:
+    """
+    mutation {
+      testOutputDummyCustomMutation(input: {id: "/dummy_custom_mutations/1", operandA: 9, clientMutationId: "myId"}) {
+        dummyCustomMutation {
+          baz
+          bat
+        }
+        clientMutationId
+      }
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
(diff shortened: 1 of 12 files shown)


soyuka:
This should've targeted 2.7, or maybe even 2.6. I'm not sure that introducing 2 new DataTransformers (implementing a deprecated interface) is a good idea.

lermontex:
@soyuka, Is the main branch no longer associated with version 2.7?

Packagist.org lists this branch as dev-main / 2.7.x-dev

https://s20.directupload.net/images/220521/qq7si96n.png

PR #4767, opened 2022-05-20
fix(metadata): property override when value is set

soyuka:
https://github.com/api-platform/core/issues/4613#issuecomment-1130306564


dannyvw:
@soyuka This fixes it. But now some other test is failing, but maybe i have configured something wrong. I need to check that.

soyuka:
You should use the new ApiProperty attribute instead of legacy ones.

dannyvw:
I use the new XML configuration instead of PHP attributes. XML configuration is not legacy right?

PR #4769, opened 2022-05-21
fix(json-ld): property metadata types and iris

soyuka:


tests/Metadata/Extractor/PropertyMetadataCompatibilityTest.php
```diff
@@ -74,7 +74,7 @@ final class PropertyMetadataCompatibilityTest extends TestCase
         'extraProperties' => [
             'custom_property' => 'Lorem ipsum dolor sit amet',
         ],
-        'iri' => false,
+        'iris' => ['schema.org/totalPrice'],
```
dunglas, line 77: Why is the scheme missing?

dunglas (approved):

PR #4784, opened 2022-06-02
[BUG] Do not rely on api-platform/core attributes classes but on interfaces instead

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | N/A
| License       | MIT
| Doc PR        | **TODO**

In case of a custom GetCollection operation attribute class, the AttributesResourceMetadataCollectionFactory generates the same name as Get operation attribute, which erases the previous operation. api-platform/core should not rely on its own attributes but on interfaces (like CollectionOperationInterface). Same for Post operation, that's why I created a PostOperationInterface.


vincentchalamon:
Closing in favor of https://github.com/soyuka/core/pull/44

PR #4791, opened 2022-06-13
fix(graphql): type builder output

soyuka:
Fix graphql tests.


develth:
seems like there is no possilbity to use `output:` on a graphQL endpoint, or?

Issue #4783, opened 2022-06-01
use AsCommand to 


tacman:
**Description**  

As of Symfony 6.1, Symfony\Component\Console\Attribute\AsCommand attribute is required.

**Example**  
```
bin/console api


2022-06-01T07:36:06-04:00 [info] User Deprecated: Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "ApiPlatform\Core\JsonSchema\Command\JsonSchemaGenerateCommand" class instead.
2022-06-01T07:36:06-04:00 [info] User Deprecated: Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "ApiPlatform\Core\Bridge\Symfony\Bundle\Command\OpenApiCommand" class instead.
2022-06-01T07:36:06-04:00 [info] User Deprecated: Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "ApiPlatform\Core\Bridge\Symfony\Bundle\Command\SwaggerCommand" class instead.
```

The fix is pretty simple, and I'm happy to do it, but am wondering which branch to put it in.

KDederichs:
If you consider it also needs to keep working with older SF versions I don't think it's that simple cause you'd need to switch between the Systems depending on the version.

PR #4795, opened 2022-06-19
Attempt to use AsCommand when available

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | 1.6
| Tickets       | #4783 
| License       | MIT
| Doc PR        | n/a

While reading that ticket I thought about how to solve it and I came up with this bit of PHP trickery.

Using class Aliasing in combination with class exists checks should theoretically switch the command systems as they are available.

The only part that's a bit messy is that I had to remove the `final` from he new Abstract classes in order to inherit them to not duplicate the code.

(Also defining those inline in a way, but I guess I can move them to separate files if desired)


KDederichs:
Ok seems like it's not picking up on the attribute (I assume due to the way that BC is defined?). If someone has experience with that, I'm all ears :)

norkunas:
Why not just overriding getDefaultName method?

KDederichs:
Oh interesting, didn't even think about that cause I was too focused on the Attribute, probably the best solution then!

src/Bridge/Symfony/Bundle/Command/GraphQlExportCommand.php
```diff
@@ -79,4 +77,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         return 0;
     }
+
+    // Use AsCommand Attribute when SF 4.4 and PHP 7 gets dropped
```
dunglas, line 81: In my opinion you can even drop this comment. There is no benefits to use an attribute here and it will be a small BC break. Same for other classes.

src/Bridge/Symfony/Bundle/Command/GraphQlExportCommand.php
```diff
@@ -30,8 +30,6 @@
  */
 class GraphQlExportCommand extends Command
```
dunglas, line 31: Not related but IMHO this class should be final. WDYT @alanpoulain?

dunglas (approved):

src/Bridge/Symfony/Bundle/Command/GraphQlExportCommand.php
KDederichs: Yeah, just thought I'd mention it in case the way commands are registered changes in the future for some reason (or if something needs to be added). I can remove them though

KDederichs:
I learned it also works if you just add the Attribute and a deprecated doc block, so I guess that's preferable @dunglas ?

KDederichs:
I'm fine with either solution tbh, just tell me which one you prefer.
Just learned that that one also works in the Lexik/Jwt repo and thought it might be preferable since it actually uses the "intended" way to declare them

soyuka:
thanks @KDederichs !

PR #4800, opened 2022-06-24
fix: XML/YAML cast values

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | https://github.com/api-platform/core/issues/4613#issuecomment-1155092567
| License       | MIT
| Doc PR        | N/A


src/Metadata/Extractor/ResourceExtractorTrait.php
```diff
@@ -97,21 +97,39 @@ private function buildArgs(SimpleXMLElement $resource): ?array
...
             }
 
-            $data[] = isset($value->values) ? $this->buildValues($value->values) : (string) $value;
+            $data[] = isset($value->values) ? $this->buildValues($value->values) : $this->castValue($value->__toString());
         }
 
         return $data;
     }
+
+    private function castValue($value)
```
alanpoulain, line 115: Could we use something from Symfony which already does it?

vincentchalamon: Good point! [`Symfony\Component\Config\Util\XmlUtils::phpize`](https://github.com/symfony/config/blob/6.1/Util/XmlUtils.php#L210) seems to do the job, except for `'1'`, `1`, `'0'` and `0` values

soyuka:
thanks @vincentchalamon !

PR #4826, opened 2022-07-11
Cleanup backward compatibility configuration

dannyvw:
I think this was forgotten when removing the backwards compatibility layer


dunglas (approved):

alanpoulain:
Thanks Danny!

Issue #3833, opened 2020-11-16
/contexts/{shortName} route should not receive a format


vincentchalamon:
Calling /contexts/Book.jsonld returns a 404. As this route should be only called for JSON-LD, the format is not necessary in the route declaration.

TODO: remove the format from the `api_jsonld_context` route declaration

**Should be fixed in v3**

PR #4844, opened 2022-07-22
Fix api_jsonld_context route declaration

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #3833
| License       | MIT
| Doc PR        | N/A


dunglas (approved):
As it's a bug fix, this should target the 2.6 branch.

dunglas:
We double-checked and it's actually not a bug. It's just a convenient feature in dev (to debug directly in the browser) and this makes the route future-proof with the upcoming YAML-LD format.

Let's merge this in 2.7 as a new feature.

PR #4853, opened 2022-07-29
feat: adjust mapping paths to the SF best practices for Bundles

coldic3:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -

According to the latest changes in Symfony recommendations for the bundles' directory structure ([here](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure)) in this PR I add a few additional paths to mapping for bundles. TBH I have no idea how I should write tests for it but if you have any hints, I'll do it :slightly_smiling_face:

### DIFF
--- a/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -279,12 +279,15 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             $dirname = $bundle['path'];
             foreach (['.yaml', '.yml', '.xml', ''] as $extension) {
                 $paths[] = "$dirname/Resources/config/api_resources$extension";
+                $paths[] = "$dirname/config/api_resources$extension";
             }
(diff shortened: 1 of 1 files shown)


dunglas:
LGTM.

By the way I would also add a directory not related to Doctrine by default, such as `ApiResource` to clearly state that you can (and probably should) use separate classes for public data models and internal ones. cc @api-platform/core-team

alanpoulain:
I'm :+1: for the `ApiResource` directory.

vincentchalamon:
`ApiResource` directory handled here: https://github.com/api-platform/core/pull/4874

vincentchalamon (approved):

vincentchalamon:
Thanks @coldic3 !

rvanlaak:
Bit late, but some more food for thought with regards to using `ApiResource`, it would also be interesting to promote DDD a bit more and stimulate others in using `/src/UserInterface/Resource` or `/src/Presentation/Api/Resource` (or something similar). Promoting that the view model/resource differs from the domain model indeed is a great thin t

Issue #2254, opened 2018-10-14
Support for Enums


ragboyjr:
*edit by soyuka*
With PHP 8.1 enums coming to PHP we want to implement the following into API Platform: 
  - Ability to declare enums as a Resource (see https://github.com/api-platform/core/issues/4349 by @bpolaszek)
  - update the schema factory to implement https://schema.org/Enumeration 
  - update the OpenAPI factory if needed https://swagger.io/docs/specification/data-models/enums/
  - make enums work with GraphQL
  - add a doctrine enum filter?

 
*end edit*

This issue is more for documentation on how to include enums with API Platform and maybe some discussion on how this could potentially be brought into the core.

I've been able to implement enums by doing the following:

1. Create a class which represents the enum, there is a base enum class that looks like this:

    ```php
    abstract class Enum
    {
        protected static $allowedValues = [];
        protected $value;

        public function __construct($value) {
            if (!\in_array($value, static::$allowedValues)) {
                throw new \InvalidArgumentException("Invalid value '${value}' for enum. Allowed values are: " . \implode(', ', static::$allowedValues));
            }

            $this->value = $value;
        }

        public static function getAllValues() {
            return self::$allowedValues;
        }

        public function __invoke() {
            return $this->value;
        }

        public function __toString() {
            return (string) $this->value;
        }
    }
    ```
    where the subclass would look like:
 
    ```php
    final class Direction extends Enum
    {
        const UP = 'up';
        const DOWN = 'down';

        protected static $allowedValues = [self::UP, self::DOWN];

        public static function up(): self {
            return new self(self::UP);
        }
        public static function down(): self {
            return new self(self::DOWN);
        }
    }
    ```
2. Create and Register a custom doctrine type for the enum:
    ```php
    abstract class EnumType extends StringType
    {
        public function convertToDatabaseValue($value, AbstractPlatform $platform) {
            if ($value instanceof Enum) {
                return (string) $value;
            }

            throw new \InvalidArgumentException('Value must be an Enum instance.');
        }

        public function requiresSQLCommentHint(AbstractPlatform $platform) {
            return false;
        }
    }

    final class DirectionEnumType extends EnumType
    {
        public function convertToPHPValue($value, AbstractPlatform $platform) {
            return new Direction($value);
        }

        public function getName() {
            return 'direction_enum';
        }
    }
    ```

    In symfony it would look like:
   ```yaml
    doctrine:
      dbal:
        types:
          channel_enum:
[...]

PR #4857, opened 2022-08-02
Return the data type "string" for the enum PHP class in the JSON Schema document.

ymaillard:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| License       | MIT

Actually, the JSON Schema document specifies the data type "object" for an enum property. According to my interpretation of the specification, it should be a string ( https://swagger.io/docs/specification/data-models/enums/ )

This PR add a test if the class is a BackedEnum class and in this case, return an array ['type'=>'string, 'format'=>'enum']

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -127,6 +127,13 @@ final class TypeFactory implements TypeFactoryInterface
             ];
         }
 
+        if (is_a($className, \BackedEnum::class, true)) {
+            return [
+                'type' => 'string',
+                'format' => 'enum',
+            ];
+        }        
+        
         // Skip if $schema is null (filters only support basic types)
         if (null === $schema) {
             return ['type' => 'string'];


alanpoulain:
I don't think the `enum` format exists though.

ymaillard:
> I don't think the `enum` format exists though.

I read on this page https://swagger.io/docs/specification/data-models/data-types/, 'format' is an open value. By example, I assume 'uuid' format used in the same method is not a value defined by open api specification. So `enum` format is a suggest.

alanpoulain:
`uuid` is a built-in format, see https://json-schema.org/draft/2020-12/json-schema-validation.html#name-resource-identifiers.
I don't think adding a custom format is a good idea, even if the specification allows it, since it will not be understood by clients.

ymaillard:
Ok I understand. Thanks for json-schema reference. I change my PR by removing the 'format' attribute. Is it ok for you ?

alanpoulain:
Yes, also you should add a unit test.

GregoireHebert:
A Backed Enum may be backed by types of int or string, and a given enumeration supports only a single type at a time.
Is there a way to detect the right type?
Plus, some detail have to be taken in consideration.
Please have a look at https://github.com/api-platform/core/issues/2254

GregoireHebert:
https://3v4l.org/dLRMX#v8.1.8 is an approach to find the backed type and associate it accordingly, or fall back into embedded object or resource as proposed in https://github.com/api-platform/core/issues/2254
I wonder how much should part this type detection should be part of `PropertyInfo`

webda2l:
This can be added as well for GraphQl with https://github.com/OLYBe/core/commit/1b63666a7394dcf78ac30f11964d71bbf51bd8ac

alanpoulain:
Superseded https://github.com/api-platform/core/pull/5120.

PR #4860, opened 2022-08-02
feat (ApiTestCase): Assert Item|collection JsonSchema with SerializerContext

GregoireHebert:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | ~
| License       | MIT
| Doc PR        | #todo

When a JsonSchema needs to be tested over specific serialization context, one has to rewrite it's own methods.
This allows to pass the serializationContext to create the right JsonSchema during testing.

### DIFF
--- a/src/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
+++ b/src/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
@@ -105,7 +105,7 @@ trait ApiTestAssertionsTrait
         static::assertThat(self::getHttpResponse()->toArray(false), $constraint, $message);
     }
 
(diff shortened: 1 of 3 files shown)


alanpoulain:
Shouldn't we get the right operation from the resource instead?
By default it will take the serialization context from it.

GregoireHebert:
@alanpoulain That's already the case I reckon :) It's an addition to allow to force the context :)

alanpoulain:
Yes you're right, sorry. Do you have a specific use case then?

GregoireHebert:
Dynamic groups for instance which cannot be extrapolated from the operation only.

alanpoulain:
We should probably wait for `3.0` to be released to merge this then (it will be available in 3.1).

Issue #4848, opened 2022-07-26
The exceptionToStatus mapping feature seems broken


rvanlaak:
**API Platform version(s) affected**: 3.0.0-beta.2

**Description**  

[The `exceptionToStatus` mapping feature ](https://api-platform.com/docs/main/core/errors/)seems broken, when configuring error mapping on an `ApiResource`. The http response code still will be a http 500 error.

**How to reproduce**  
```php
namespace App\Application\ReadModel\Projection;

use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Get;
use App\Application\ReadModel\ProductProvider;
use App\Domain\ReadModel\Product\NotFoundException;

#[ApiResource(
    provider: ProductProvider::class,
    operations: [
        new Get(
            exceptionToStatus: [
                [NotFoundException::class => 404]
            ],
        )
    ]
)]
final class Product
{
    // ...
}
```

**Possible Solution**  

Not sure how I can debug this further, as we get the same http 500 error as on our other endpoint.

**Additional Context**  

* Did do a fresh installation of API Platform v3.0.0-beta.2 on an existing project
* PHP 8.1
* Symfony 6.1

dannyvw:
It looks like that the operation is null here https://github.com/api-platform/core/blob/main/src/Action/ExceptionAction.php#L61

ArnoudThibaut:
I investigated this problem and I think that the `exceptionToStatus` option can't work on operations (and maybe also on Resources, I didn't try it) because of [Symfony ErrorListener](https://github.com/symfony/symfony/blob/f42339552cdeee8aa1e6286c2c97dff1cca3032e/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php#L166). 
The request is duplicated without all the attributes needed to fetch the exception to status mapping

sontungpham:
I had the same problem. After installing a newer version `v3.0.0-rc.1`, it works again.
I write the config in `config/packages/api_platform.yaml` btw

PR #4861, opened 2022-08-02
 Fix exceptionToStatus in ApiResource and Operation

ArnoudThibaut:
| Q             | A
| ------------- | ---
| Tickets       | #4848 
| License       | MIT

I open this PR to discuss about the possible solution to fix the `exceptionToStatus` feature inside ApiResource attribute and operation

Because of [Symfony ErrorListener](https://github.com/symfony/symfony/blob/f42339552cdeee8aa1e6286c2c97dff1cca3032e/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php#L166) the request is duplicated without all the attributes needed to fetch the exception to status mapping.

A possible solution (the one I choose) could be to extend this listener to override the `duplicateRequest` method to also duplicate the attributes needed to get the exception to status mapping but I'm not realy satisfied with this solution so I'm really open to suggestions

### DIFF
--- a/features/main/exception_to_status.feature
+++ b/features/main/exception_to_status.feature
@@ -3,6 +3,7 @@ Feature: Using exception_to_status config
   I can customize the status code returned if the application throws an exception
 
   @createSchema
+  @!mongodb
   Scenario: Configure status code via the operation exceptionToStatus to map custom NotFound error to 404
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "GET" request to "/dummy_exception_to_statuses/123"
@@ -10,6 +11,7 @@ Feature: Using exception_to_status config
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
 
+  @!mongodb
   Scenario: Configure status code via the resource exceptionToStatus to map custom NotFound error to 400
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "PUT" request to "/dummy_exception_to_statuses/123" with body:
@@ -22,6 +24,7 @@ Feature: Using exception_to_status config
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
 
+  @!mongodb
   Scenario: Configure status code via the config file to map FilterValidationException to 400
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "GET" request to "/dummy_exception_to_statuses"
--- a/features/main/exception_to_status.feature
+++ b/features/main/exception_to_status.feature
@@ -1,0 +1,27 @@
+Feature: Using exception_to_status config
+  As an API developer
+  I can customize the status code returned if the application throw an exception
+
+  @createSchema
+  Scenario: Configure status code via the operation exceptionToStatus to map custom NotFound error to 404
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/dummy_exception_to_statuses/123"
+    Then the response status code should be 404
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+
+  @createSchema
+  Scenario: Configure status code via the resource exceptionToStatus to map custom NotFound error to 400
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/dummy_exceptions/123"
+    Then the response status code should be 400
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+
+  @createSchema
+  Scenario: Configure status code via the config file to map FilterValidationException to 400
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/dummy_exception_to_statuses"
+    Then the response status code should be 400
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
--- a/features/main/exception_to_status.feature
+++ b/features/main/exception_to_status.feature
@@ -12,7 +12,7 @@ Feature: Using exception_to_status config
 
   Scenario: Configure status code via the resource exceptionToStatus to map custom NotFound error to 400
     When I add "Content-Type" header equal to "application/ld+json"
-    And I send a "PUT" request to "/dummy_exceptions/123" with body:
+    And I send a "PUT" request to "/dummy_exception_to_statuses/123" with body:
     """
     {
         "name": "black"
--- a/features/main/exception_to_status.feature
+++ b/features/main/exception_to_status.feature
@@ -1,6 +1,6 @@
 Feature: Using exception_to_status config
   As an API developer
-  I can customize the status code returned if the application throw an exception
+  I can customize the status code returned if the application throws an exception
 
   @createSchema
   Scenario: Configure status code via the operation exceptionToStatus to map custom NotFound error to 404
@@ -10,15 +10,18 @@ Feature: Using exception_to_status config
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
 
-  @createSchema
   Scenario: Configure status code via the resource exceptionToStatus to map custom NotFound error to 400
     When I add "Content-Type" header equal to "application/ld+json"
-    And I send a "GET" request to "/dummy_exceptions/123"
+    And I send a "PUT" request to "/dummy_exceptions/123" with body:
+    """
+    {
+        "name": "black"
+    }
+    """
     Then the response status code should be 400
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
 
-  @createSchema
   Scenario: Configure status code via the config file to map FilterValidationException to 400
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "GET" request to "/dummy_exception_to_statuses"
--- a/src/Symfony/Bundle/Resources/config/symfony.xml
+++ b/src/Symfony/Bundle/Resources/config/symfony.xml
@@ -66,13 +66,12 @@
             <tag name="kernel.event_listener" event="kernel.exception" method="onKernelException" />
         </service>
 
-        <service id="api_platform.listener.error" class="ApiPlatform\Symfony\EventListener\ErrorListener" parent="exception_listener">
-        </service>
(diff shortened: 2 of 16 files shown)


alanpoulain:
I don't see a better solution.
Was it working in 2.6?

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Symfony\EventListener;
+
+use Symfony\Component\HttpFoundation\Request;
+
+class ErrorListener extends \Symfony\Component\HttpKernel\EventListener\ErrorListener
```
dunglas, line 9:
```suggestion
final class ErrorListener extends \Symfony\Component\HttpKernel\EventListener\ErrorListener
```

dunglas (approved):

ArnoudThibaut:
> Was it working in 2.6?

I did not find the feature in 2.6

alanpoulain:
Right, it was introduced by this PR: https://github.com/api-platform/core/pull/3519.
Could you target 2.7 then, since it's a bugfix?

ArnoudThibaut:
> Could you target 2.7 then, since it's a bugfix?

It's done

src/Symfony/Bundle/Resources/config/symfony.xml
```diff
@@ -66,11 +66,13 @@
             <tag name="kernel.event_listener" event="kernel.exception" method="onKernelException" />
         </service>
 
+        <service id="api_platform.listener.error" class="ApiPlatform\Symfony\EventListener\ErrorListener" parent="exception_listener">
+        </service>
```
alanpoulain, line 70:
```suggestion
        <service id="api_platform.listener.error" class="ApiPlatform\Symfony\EventListener\ErrorListener" parent="exception_listener" />
```

features/main/exception_to_status.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Using exception_to_status config
+  As an API developer
+  I can customize the status code returned if the application throw an exception
```
alanpoulain, line 3:
```suggestion
  I can customize the status code returned if the application throws an exception
```

features/main/exception_to_status.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  @createSchema
+  Scenario: Configure status code via the operation exceptionToStatus to map custom NotFound error to 404
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "GET" request to "/dummy_exception_to_statuses/123"
+    Then the response status code should be 404
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+
+  @createSchema
```
alanpoulain, line 13: Maybe we can remove some `createSchema` since there are not really useful and time consuming.
I known tests will not be independently anymore, but we already do it everywhere (IMO we should use transactions to speed up tests instead).

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -0,0 +1,26 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Symfony\EventListener;
+
+use Symfony\Component\HttpFoundation\Request;
+
+final class ErrorListener extends \Symfony\Component\HttpKernel\EventListener\ErrorListener
```
alanpoulain, line 17: Maybe you could add a comment on this class to explain why it is useful?

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    protected function duplicateRequest(\Throwable $exception, Request $request): Request
+    {
+        $dup =  parent::duplicateRequest($exception, $request);
```
alanpoulain, line 21:
```suggestion
        $dup = parent::duplicateRequest($exception, $request);
```

tests/Fixtures/TestBundle/Entity/DummyExceptionToStatus.php
```diff
@@ -0,0 +1,114 @@
...
+    ],
+    provider: DummyExceptionToStatusProvider::class
+)]
+#[ApiFilter(RequiredFilter::class)]
+#[ORM\Entity]
+class DummyExceptionToStatus
+{
+    /**
+     * @var int|null The id
+     */
```
alanpoulain, line 46: PHPDoc can be removed.

tests/Fixtures/app/config/config_common.yml
```diff
@@ -107,6 +107,10 @@ services:
         tags:
             -  { name: 'api_platform.state_provider' }
 
+    ApiPlatform\Tests\Fixtures\TestBundle\State\DummyExceptionToStatusProvider:
+        tags:
+            -  { name: 'api_platform.state_provider' }
```
alanpoulain, line 112:
```suggestion
            - { name: 'api_platform.state_provider' }
```

features/main/exception_to_status.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Using exception_to_status config
+  As an API developer
+  I can customize the status code returned if the application throw an exception
+
+  @createSchema
+  Scenario: Configure status code via the operation exceptionToStatus to map custom NotFound error to 404
+    When I add "Content-Type" header equal to "application/ld+json"
```
rvanlaak, line 7: Shouldn't the exception mapping also work for other content types?

alanpoulain: Yes it should, don't worry!
We cannot test every content types, having a functional test for this feature is already a very good thing.

rvanlaak (approved):
Nice, clean implementation!

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -0,0 +1,34 @@
...
+ * This error listener extends the Symfony one in order to add
+ * the `_api_operation` attribute when the request is duplicated.
+ * It will later be used to retrieve the exceptionToStatus from the operation ({@see ExceptionAction}).
+ */
+final class ErrorListener extends SymfonyErrorListener
+{
+    protected function duplicateRequest(\Throwable $exception, Request $request): Request
+    {
+        $dup = parent::duplicateRequest($exception, $request);
+        $dup->attributes->set('_api_operation', $request->attributes->get('_api_operation'));
```
vincentchalamon, line 30: You should check the attribute already exists before adding it with null value:
```suggestion
        if ($request->attributes->has('_api_operation')) {
            $dup->attributes->set('_api_operation', $request->attributes->get('_api_operation'));
        }
```

vincentchalamon (changes requested):

src/Symfony/EventListener/ExceptionListener.php
```diff
@@ -30,11 +28,11 @@ final class ExceptionListener
...
      * @var ErrorListener
      */
-    private $exceptionListener;
+    private $errorListener;
 
-    public function __construct($controller, LoggerInterface $logger = null, $debug = false, ErrorListener $errorListener = null)
+    public function __construct($controller, LoggerInterface $logger = null, $debug = false)
     {
-        $this->exceptionListener = $errorListener ? new ErrorListener($controller, $logger, $debug) : new LegacyExceptionListener($controller, $logger, $debug); // @phpstan-ignore-line
+        $this->errorListener = new ErrorListener($controller, $logger, $debug);
```
chalasr, line 35: Is there anything that prevents injecting the error listener rather than instantiating here? 
```suggestion
        $this->errorListener = $errorListener ?? new LegacyExceptionListener($controller, $logger, $debug); // @phpstan-ignore-line;
```

(+ update of the `api_platform.listener.exception` service definition to give it the `api_platform.listener.error` service as argument)

chalasr (approved):
I think we have no other choice than extending the class+service via inheritance here.
This could probably be revisited on 3.x, in conjunction with making the ErrorListener more cleanly extensible on Symfony's side.

rvanlaak:
> I think we have no other choice than extending the class+service via inheritance here. This could probably be revisited on 3.x, in conjunction with making the ErrorListener more cleanly extensible on Symfony's side.

This approach actually is the right way of doing it. The class gets registered as additional error listener, and basically adds some more data to the response for further processing. As long as we do not set the exception response, the follow-up listener (with a higher priority) will get triggered as well. 

So, maybe we can add an integration test to make sure that the listener thereafter still gets triggered as well?

chalasr:
What I meant is that, as a rule of thumb, decoration should be preferred over inheritance. Especially when it comes to third party code

rvanlaak:
Ah @chalasr fair point, I did not see that it uses extension. My comment solely focussed on the service registration and related event listener priorities.

Is there a reason why the request needs to get duplicated? Is that because the `_api_operation` has to be available at a later moment? What about then just lowering the priority of the error listener to before the other listener that makes use of it? That could then also possibly allow removing the extend, as the duplication then isn't needed anymore.

soyuka:
Please check my last commit:
- fixes a typo in the behat tests that used a wrong resource path (dummy_exceptions instead of dummy_exceptions_to_statuses)
- uses service declaration

Let me know what you think of that.

chalasr (commented):
👍

soyuka:
This is quite ready to be merged but we'll focus on the #4881 first.

soyuka:
Thanks @ArnoudThibaut

dannyvw:
@soyuka This feature is not available in 3.0

soyuka:
You're right, I'll check how to fix this today.

dannyvw:
@soyuka Maybe this merge https://github.com/api-platform/core/pull/4910

rvanlaak:
Looking to https://github.com/api-platform/core/commits/main/src/Symfony/EventListener, it seems that there isn't any commit on main that reverts this change?

rvanlaak:
For reference, feature seems (re)introduced for branch 3.0 via PR https://github.com/api-platform/core/pull/4981

jackbentley:
This broke error handling in v2.7 + Symfony 4.4

The fix implemented by #3285 no longer works because the `error_listener` passed is now always present and never null.

https://github.com/api-platform/core/blob/6b483bc3a2696a908ea4cb68cda0344cfa224146/src/Symfony/Bundle/Resources/config/legacy/api.xml#L95

https://github.com/api-platform/core/blob/6b483bc3a2696a908ea4cb68cda0344cfa224146/src/Symfony/EventListener/ExceptionListener.php#L40

PR #4864, opened 2022-08-03
Remove setPublic for ValidationGroupsGeneratorInterface

dannyvw:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #... 
| License       | MIT

There are a few more 3.0 todos in the code


alanpoulain:
I think it's more complicated than just removing this line.
Apparently, if I follow correctly the discussion [here](https://github.com/api-platform/core/pull/3346), we also need to use a service locator instead of the full container in ApiPlatform\Symfony\Validator\Validator.php.
See this comment in particular: https://github.com/api-platform/core/pull/3346#discussion_r383869931.
Do you want to do it?

dannyvw:
@alanpoulain PR is updated

src/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorPass.php
```diff
@@ -0,0 +1,43 @@
...
+    public function process(ContainerBuilder $container): void
+    {
+        $generators = [];
+        foreach ($container->findTaggedServiceIds('api_platform.validation_groups_generator', true) as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                $generators[$tag['id'] ?? $serviceId] = new Reference($serviceId);
+            }
+        }
+
+        $container->getDefinition('api_platform.validator_locator')->addArgument($generators);
```
alanpoulain, line 41: Can't we try to use `tagged_locator` instead of a compiler pass, like suggested by @dunglas?

dannyvw: That was my first implementation but didn't work because of this error "Failed asserting that exception message 'Invalid definition for service "api_platform.validator_locator": an array of references is expected as first argument when the "container.service_locator" tag is set.' contains '$json must be array or string (JSON array or JSON object)'."

src/Symfony/Validator/Validator.php
```diff
@@ -28,7 +28,7 @@
  */
 class Validator implements ValidatorInterface
 {
-    public function __construct(private readonly SymfonyValidatorInterface $validator, private readonly ?ContainerInterface $container = null)
+    public function __construct(private readonly SymfonyValidatorInterface $validator, private readonly ContainerInterface $locator)
```
alanpoulain, line 31: It's a breaking change, we cannot do this since it was not deprecated beforehand.

dannyvw: Even if the class is marked as final? 

I can revert the null value if needed. Changing name is ok?

src/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorPass.php
alanpoulain: I don't think it is the way to do it.
You don't need to define a new service, just change the configuration of `api_platform.validator`:
```php
        <service id="api_platform.validator" class="ApiPlatform\Symfony\Validator\Validator">
            <argument type="service" id="validator" />
            <argument type="tagged_locator" tag="api_platform.validation_groups_generator" />
        </service>
```

src/Symfony/Validator/Validator.php
alanpoulain: Yes! You should read https://symfony.com/doc/current/contributing/code/bc.html#working-on-symfony-code, we follow the same rules.
In this case, we are changing an argument type for a public method. Even if the class is marked as `final`, it is not allowed.
You can change the name though, it is allowed (but not recommended).

src/Symfony/Bundle/DependencyInjection/Compiler/ValidationGroupsGeneratorPass.php
dannyvw: That works thanks, pr is updated.

alanpoulain (approved):

Issue #4879, opened 2022-08-05
Declare GetCollection operation only on resource


vincentchalamon:
Considering the following resource:
```php
#[ApiResource(graphQlOperations: [], operations: [new GetCollection])]
class Foo
{
    public function __construct(
        #[ApiProperty(identifier: false)]
        private string $id
    ) {}
}
```

The collection is not able to generate an iri for each items cause no _item_ operation is available. The trick should be to add a `new Get` operation with a `NotFoundAction::class` controller...

The cleanest way to handle it would be separate the Collection resource from the Item resource, but let's be real: it's not the best DX. The solution is to automatically add an operation (`NotExposed`) if no item operation is found on the resource, which helps to generate the item iri, and if no identifier is available on the resource use a skolem iri (and don't expose it in Symfony routing).

Version impacted: 2.7

PR #4881, opened 2022-08-05
Add NotExposed operation

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #4879
| License       | MIT
| Doc PRs      | https://github.com/api-platform/docs/pull/1590, https://github.com/api-platform/docs/pull/1592

1. Generating a default NotExposed operation to generate the IRI if the resource has no Get operation available
2. Disabling operation from OpenApi documentation
3. Specifying the operation to use to generate the IRI
4. Handling output on a resource with a GetCollection operation but with no Get operation available
5. Miscellaneous

---

### Description

The following resource:

```php
#[GetCollection] // path will be: /users
class User
```

When I call `GET /users`, I get the following error:
> "Operation \"\" not found for resource \"User\"."

That's because API Platform cannot find any Get operation to generate the IRI for each collection member. A trick currently used is to manually add a Get operation with `read: false, output: false, controller: NotFoundAction::class`, which is not ideal.

### Solution

Detect that this resource has no Get operation, and automatically add a _NotExposed_ operation. This operation could be used to generate an IRI like `/users/{id}` is the resource has an identifier, or a Skolem IRI `/.well-known/genid/{id}` if the resource doesn't have any identifier.

### TODO

- [x] Add NotExposedOperationResourceMetadataCollectionFactory
- [x] Add NotExposedAction
- [x] Add Behat tests
- [x] Add unit tests
- [x] Add documentation about NotExposed auto-generated operation

---

### Description

The following resource:

```php
#[GetCollection]
#[Get(uriTemplate: '/foo')] // I don't want this operation to be exposed on OpenApi documentation
class User
```

For any reason, I want an operation to not be exposed on the OpenApi documentation. It also implies the NotExposed auto-generated operation explained on the previous use case.

Even if it is available in the Symfony routing, it should not be available on the OpenApi documentation.

### Solution

Add a `openapi` boolean option on Operation to disable this operation from the OpenApi documentation.

> Note: in a later version, it should be better to merge `openapiContext` into `openapi` to improve DX and reduce confusion and multiple options.

```php
#[GetCollection]
#[Get(openapi: false, uriTemplate: '/foo')]
class User
```

### TODO

- [x] Add `openapi` boolean option on operation classes
- [x] Ignore operation from OpenApi documentation if `openapi` is strictly false
- [x] Update unit tests
- [x] Add documentation about `openapi` new operation option

---

### Description

The following resource:

```php
#[GetCollection] // path will be: /users
#[Get] // path will be: /users/{id}
#[GetCollection(uriTemplate: '/companies/{companyId}/users', ...)]
#[Get(uriTemplate: '/companies/{companyId}/users/{id}', ...)]
class User
```

When I call `GET /companies/{companyId}/users`, the members IRI will be the first Get operation found, so `/users/{id}`. But it should be `/companies/{companyId}/users/{id}` to be consistent with the operation initially called.

### Solution

Add a `itemUriTemplate` string option on Operation to specify which operation to use to generate the IRI. If this option is not set, then the default behavior is applied (select the first Get operation detected).

This option is only useful for collection operations (GetCollection and Post).

```php
#[GetCollection] // the "itemUriTemplate" option is not used, it will use the first Get operation available
#[Get]
#[GetCollection(itemUriTemplate: '/companies/{companyId}/users/{id}', uriTemplate: '/companies/{companyId}/users', ...)]
#[Get(uriTemplate: '/companies/{companyId}/users/{id}', ...)]
class User
```

### TODO

- [x] Add `itemUriTemplate` option on operation, and use it on `$this->resourceMetadataCollectionFactory->getOperation()` calls if available and from a collection operation context
- [x] Add Behat tests
- [x] Update unit tests
- [x] Add documentation about `itemUriTemplate` option

---

### Description

The following resource:
```php
#[GetCollection(output: UserOutput::class)] // path will be: /users
// A NotExposed operation is automatically generated
```

When I call `GET /users` on a GetCollection with output, API Platform is trying to find a Get operation available on the output resource to generate the IRI, but it is not a resource. The following error occurs:
> "Operation \"\" not found for resource \"User\"."

### Solution

The best solution would be to support ApiResource and ApiProperty on DTO, but it could be done in another PR. Meanwhile, it's better to return a Skolem IRI.

### TODO

- [x] Generate a Skolem IRI
- [x] Add Behat tests

---

## 5. Miscellaneous

- [x] Fix Doctrine errors on PHPStan
- [x] Fix Swagger v2 generation


src/Metadata/NotExposed.php
```diff
@@ -0,0 +1,110 @@
...
+        ?array $requirements = null,
+        ?array $options = null,
+        ?bool $stateless = null,
+        ?string $sunset = null,
+        ?string $acceptPatch = null,
+        $status = null,
+        ?string $host = null,
+        ?array $schemes = null,
+        ?string $condition = null,
+        ?string $controller = NotFoundAction::class,
```
alanpoulain, line 47: Is it useful since the route is not generated anyway?

src/Metadata/Resource/Factory/NotFoundOperationResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,75 @@
...
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+
+/**
+ * Adds a Get operation with NotFoundAction on a resource which only has a GetCollection.
+ * Helps to generate resource iri for items.
+ *
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
+ * @experimental
+ */
+final class NotFoundOperationResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
```
dunglas, line 30:
```suggestion
final class NotExposedOperationResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
```

src/Symfony/Routing/ApiLoader.php
```diff
@@ -107,7 +107,7 @@ public function load($data, $type = null): RouteCollection
 
             foreach ($this->resourceMetadataFactory->create($resourceClass) as $resourceMetadata) {
                 foreach ($resourceMetadata->getOperations() as $operationName => $operation) {
-                    if ($operation->getRouteName()) {
+                    if ($operation->getRouteName() || ($operation instanceof HttpOperation && str_starts_with($operation->getUriTemplate(), '/.well-known/genid/'))) {
```
alanpoulain, line 110: I find this a little ugly, couldn't we have a special parameter in the operation instead?

src/Action/NotExposedAction.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Action;
+
+use ApiPlatform\Exception\NotExposedHttpException;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * An action which always returns HTTP 404 Not Found with an explanation why operation is not exposed.
```
alanpoulain, line 20:
```suggestion
 * An action which always returns HTTP 404 Not Found with an explanation for why the operation is not exposed.
```

src/Action/NotExposedAction.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+final class NotExposedAction
+{
+    public function __invoke(Request $request)
```
alanpoulain, line 24:
```suggestion
    public function __invoke(Request $request): never
```

src/Action/NotExposedAction.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $message = 'This route does not aim to be called.';
+        if ('api_genid' === $request->attributes->get('_route')) {
+            $message = 'This route is not exposed on purpose. It generates an iri for a collection resource without identifier nor item operation.';
```
alanpoulain, line 28:
```suggestion
            $message = 'This route is not exposed on purpose. It generates an IRI for a collection resource without identifier nor item operation.';
```

src/Exception/NotExposedHttpException.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Exception;
+
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+/**
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
+ */
+class NotExposedHttpException extends HttpException
```
alanpoulain, line 21: Why not extending the `NotFoundHttpException`?

src/JsonApi/Serializer/ObjectNormalizer.php
```diff
@@ -102,7 +102,7 @@ public function normalize($object, $format = null, array $context = [])
             ];
         } else {
             $resourceData = [
-                'id' => '/.well-known/genid/'.bin2hex(random_bytes(10)),
+                'id' => '/.well-known/genid/'.spl_object_hash($object),
```
alanpoulain, line 105: Shouldn't we have a trait for this?

src/Exception/NotExposedHttpException.php
vincentchalamon: WDYT @dunglas ?

src/Metadata/NotExposed.php
```diff
@@ -0,0 +1,108 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata;
+
+/**
+ * @internal
```
alanpoulain, line 17: Maybe a little description here?

src/Metadata/NotExposed.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
```
alanpoulain, line 19: I would remove `IS_REPEATABLE` since I don't see any use case.

vincentchalamon: As any operation, it must be repeatable

src/Metadata/Resource/Factory/NotExposedOperationResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,80 @@
...
+namespace ApiPlatform\Metadata\Resource\Factory;
+
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\GraphQl\Operation as GraphQlOperation;
+use ApiPlatform\Metadata\NotExposed;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+
+/**
+ * Adds a Get operation with NotFoundAction on a resource which only has a GetCollection.
```
alanpoulain, line 23:
```suggestion
 * Adds a {@see NotExposed} operation with {@see NotFoundAction} on a resource which only has a GetCollection.
```

src/Metadata/Resource/Factory/NotExposedOperationResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,80 @@
...
+ * Helps to generate resource iri for items.
```
alanpoulain, line 24:
```suggestion
 * This operation helps to generate resource IRI for items.
```

src/Metadata/Resource/Factory/NotExposedOperationResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,80 @@
...
+                }
+
+                // An item operation has been found, nothing to do anymore in this class
+                return $resourceMetadataCollection;
+            }
+        }
+
+        // No item operation has been found on all resources for resource class: generate one on the last resource
+        // Helpful to generate an "@id" for a resource without declaring the Get operation
+        $operation = (new NotExposed())->withResource($resource)->withUriTemplate(); // @phpstan-ignore-line
```
alanpoulain, line 72: Why calling `withUriTemplate`?

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -141,6 +141,11 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
                 continue;
             }
 
+            // Operation ignored from OpenApi
+            if ($operation instanceof HttpOperation && false === $operation->getOpenApi()) {
```
alanpoulain, line 145:
```suggestion
            if ($operation instanceof HttpOperation && false === $operation->getOpenapi()) {
```

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -392,6 +397,11 @@ private function getLinks(ResourceMetadataCollection $resourceMetadataCollection
                     continue;
                 }
 
+                // Operation ignored from OpenApi
+                if ($operation instanceof HttpOperation && false === $operation->getOpenApi()) {
```
alanpoulain, line 401:
```suggestion
                if ($operation instanceof HttpOperation && false === $operation->getOpenapi()) {
```

tests/Fixtures/TestBundle/Model/Chair.php
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Single resource without identifiers and with a single collection.
+ * A NotExposed operation with "routeName: api_genid" is automatically added to this resource.
+ *
+ * @see NotExposedOperationResourceMetadataCollectionFactory::create()
+ *
+ * @author Vincent Chalamon <vincentchalamon@gmail.com>
+ */
+class Chair
```
alanpoulain, line 26: Why not using annotations? It's easier to maintain than YAML.

vincentchalamon: Because this directory seems already loaded by the test configuration and we can't make a difference with v3 configuration vs legacy configuration. Instead, we load different directories depending on the `metadata_backward_compatibility_layer`.

src/Action/NotExposedAction.php
alanpoulain: Not sure if it's possible in 2.7 though.

src/Metadata/Resource/Factory/NotExposedOperationResourceMetadataCollectionFactory.php
vincentchalamon: To don't inherit it from resource if it's declared on it. I'll update the comment in consideration

webda2l:
Hello @vincentchalamon!

From my testing of the latest 2.7 (kudos to the team!) GraphQl will need a fix as well when there is a field with sub collection like this minimal code (videos.collection.xxx) : 
```
{
  collectionPageQueryVideoGroups {
    collection {
      videos {
        collection {
          thumbnailPath
        }
      }
    }
  }
}
```

I get the error:

```
{
  "errors": [
    {
      "debugMessage": "[Semantical Error] line 0, col 41 near 'groups IN (SELECT': Error: Invalid PathExpression. StateFieldPathExpression or SingleValuedAssociationField expected.",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 4,
          "column": 7
        }
      ],
      "path": [
        "collectionPageQueryVideoGroups",
        "collection",
        0,
        "videos"
      ],
```

whereas my video resource is well defined with the Get operation (all my resources are)

```
#[APIP\ApiResource(
    normalizationContext: ['groups' => ['video:read', 'propKey:read', 'propValue:read', 'teacher:read', 'timestamp']],
    order: ['createdAt' => 'DESC'],
    paginationItemsPerPage: 12,
    operations: [new APIP\Get()],
    graphQlOperations: [
        new APIP\GraphQl\QueryCollection(
            name: 'collectionPageQuery',
            resolver: VideoCollectionResolver::class,
            paginationType: 'page'
        ),
        new APIP\GraphQl\QueryCollection(
            name: 'collectionPageQueryRecommended',
            resolver: VideoCollectionResolver::class,
            paginationType: 'page'
        ),
    ],
)]
```

I will post some other feedbacks about this 2.7 in the dedicated issue.

alanpoulain:
@webda2l I don't think it is related to this PR, maybe there is an issue with the generation of the links.
You can add your problem here: https://github.com/api-platform/core/issues/4613
Please add your resources as well (particularly the `video` relation).

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -80,6 +80,30 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
         $this->handleNotFound($operationName, $metadata);
     }
 
+    /**
+     * Find a Get operation which matches the uriTemplate.
+     */
+    public function matchOperation(string $uriTemplate): Operation
```
soyuka, line 86: What about avoiding this new method and instead:

- when `itemUriTemplate` is given, store an `itemOperationName`
- then use `$resourceCollection->getOperation($operation->getItemOperationName())` 

If needed, a user can specify `itemOperationName` instead of the template.

soyuka (commented):
Thanks nice work!

src/Symfony/Routing/ApiLoader.php
```diff
@@ -111,6 +111,10 @@ public function load($data, $type = null): RouteCollection
                         continue;
                     }
 
+                    if ($operation->getUriTemplate() === IriConverter::$skolemUriTemplate) {
```
vincentchalamon:
```suggestion
if (IriConverter::$skolemUriTemplate === $operation->getUriTemplate()) {
```

src/Metadata/Resource/Factory/NotExposedOperationResourceMetadataCollectionFactory.php
```diff
@@ -69,12 +72,14 @@ public function create(string $resourceClass): ResourceMetadataCollection
...
-            $operation = $operation->withRouteName('api_genid');
+        /** @var HttpOperation $operation */
+        $operation = (new NotExposed())->withClass($resource->getClass())->withShortName($resource->getShortName()); // @phpstan-ignore-line $resource is defined if count > 0
+
+        if (!$this->linkFactory->createLinksFromIdentifiers($operation)) {
+            $operation = $operation->withUriTemplate(IriConverter::$skolemUriTemplate);
         }
-        $operations->add(sprintf('_api_%s_get', $resource->getShortName()), $operation)->sort(); // @phpstan-ignore-line
+
+        $operations->add(sprintf('_api_%s_get', $operation->getShortName()), $operation)->sort(); // @phpstan-ignore-line $operations exists
```
vincentchalamon, line 82:
```suggestion
$operations->add(sprintf('_api_%s_get', $operation->getShortName()), $operation)->sort(); // @phpstan-ignore-line $operation exists
```

src/JsonLd/ContextBuilder.php
```diff
@@ -190,7 +192,12 @@ public function getAnonymousResourceContext($object, array $context = [], int $r
         ];
 
         if (!isset($context['iri']) || false !== $context['iri']) {
-            $jsonLdContext['@id'] = $context['iri'] ?? $this->generateSkolemIri($object);
+            // Not using an IriConverter here is deprecated in 2.7, avoid spl_object_hash as it may collide
```
vincentchalamon, line 195: Maybe add a `todo` here to help clear this code in 3.0

src/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -142,10 +142,14 @@
             <argument type="service" id="api_platform.resource_class_resolver" />
             <argument type="service" id="api_platform.metadata.resource.metadata_collection_factory" />
             <argument type="service" id="api_platform.uri_variables.converter" />
-            <argument>null</argument>
+            <argument type="service" id="api_platform.symfony.iri_converter.skolem" on-invalid="null"></argument>
```
vincentchalamon, line 145:
```suggestion
<argument type="service" id="api_platform.symfony.iri_converter.skolem" on-invalid="null" />
```

src/Symfony/Routing/ApiLoader.php
```diff
@@ -111,7 +111,7 @@ public function load($data, $type = null): RouteCollection
                         continue;
                     }
 
-                    if ($operation->getUriTemplate() === IriConverter::$skolemUriTemplate) {
+                    if ($operation->getUriTemplate() === SkolemIriConverter::$skolemUriTemplate) {
```
vincentchalamon, line 114:
```suggestion
                    if (SkolemIriConverter::$skolemUriTemplate === $operation->getUriTemplate()) {
```

src/Symfony/Routing/IriConverter.php
```diff
@@ -136,8 +136,9 @@ public function getIriFromResource($item, int $referenceType = UrlGeneratorInter
         // Custom resources should have the same IRI as requested, it was not the case pre 2.7
         $isLegacyCustomResource = ($operation->getExtraProperties()['is_legacy_resource_metadata'] ?? false) && ($operation->getExtraProperties()['user_defined_uri_template'] ?? false);
 
-        if ($operation instanceof HttpOperation && HttpOperation::METHOD_POST === $operation->getMethod() && ($itemUriTemplate = $operation->getItemUriTemplate())) {
-            $operation = $this->resourceMetadataCollectionFactory->create($resourceClass)->matchOperation($itemUriTemplate);
+        // FIXME: to avoid the method_exists we could create an interface for the Post operation, we can't guarantee that the user extended our ApiPlatform\Metadata\Post
```
vincentchalamon, line 139: Yes we can: the `ApiPlatform\Metadata\Post` is final!

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -61,6 +61,10 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
                 if ($name === $operationName) {
                     return $this->operationCache[$operationName] = $operation;
                 }
+
+                if ($operation->getUriTemplate() === $operationName) {
```
vincentchalamon, line 65:
```suggestion
               if ($operationName === $operation->getUriTemplate()) {
```
Also, the other `if` instances in this loop don't respect the Yoda Style.

PR #4886, opened 2022-08-08
chore: better PHP types

alanpoulain:


src/State/ProcessorInterface.php
```diff
@@ -26,5 +26,5 @@ interface ProcessorInterface
     /**
      * Handle the state.
      */
-    public function process($data, Operation $operation, array $uriVariables = [], array $context = []);
+    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []);
```
chalasr, line 29: Is there any use case for returning something? Or should there be a void return type?

alanpoulain: Yes there is: you return your resource, for instance after a `POST` to obtain the generated identifier.

chalasr: Makes sense, thanks

PR #4888, opened 2022-08-09
fix: deprecated `UpgradeApiResourceCommand::$defaultName`

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | -
| License       | MIT
| Doc PR        | -

Fixes: 

>  * 1x: Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "ApiPlatform\Core\Bridge\Symfony\Bundle\Command\UpgradeApiResourceCommand" class instead.


chalasr:
As discussed this way is more future-proof and inline with what has been done in other community bundles. Opened #4889 to use it across the board

alanpoulain (approved):

PR #4889, opened 2022-08-09
chore: better BC/FC layer for deprecated `Command::$defaultName`

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | -
| License       | MIT
| Doc PR        | -

`Command::setDescription()` method calls can also be removed once Symfony 6.1+ is required

### DIFF
--- a/src/Bridge/Symfony/Bundle/Command/GraphQlExportCommand.php
+++ b/src/Bridge/Symfony/Bundle/Command/GraphQlExportCommand.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Core\Bridge\Symfony\Bundle\Command;
 
 use ApiPlatform\Core\GraphQl\Type\SchemaBuilderInterface;
 use GraphQL\Utils\SchemaPrinter;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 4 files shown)


src/Bridge/Symfony/Bundle/Command/GraphQlExportCommand.php
```diff
@@ -28,8 +29,14 @@
  *
  * @author Alan Poulain <contact@alanpoulain.eu>
  */
+#[AsCommand(name: 'api:json-schema:generate', description: 'Export the GraphQL schema in Schema Definition Language (SDL)')]
 class GraphQlExportCommand extends Command
 {
+    /**
+     * @deprecated To be removed along with Symfony <6.1 compatibility
+     */
+    protected static $defaultName = 'api:json-schema:generate';
```
alanpoulain, line 38:
```suggestion
    protected static $defaultName = 'api:graphql:export';
```

src/Bridge/Symfony/Bundle/Command/OpenApiCommand.php
```diff
@@ -27,8 +28,14 @@
 /**
  * Dumps Open API documentation.
  */
+#[AsCommand(name: 'api:openapi:export', description: 'Dump the Open API documentation')]
```
alanpoulain, line 31: `getDefaultName` to remove?

src/Bridge/Symfony/Bundle/Command/SwaggerCommand.php
```diff
@@ -31,8 +32,14 @@
  *
  * @author Amrouche Hamza <hamza.simperfit@gmail.com>
  */
+#[AsCommand(name: 'api:json-schema:generate', description: 'Dump the Swagger v2 documentation')]
```
alanpoulain, line 35:
```suggestion
#[AsCommand(name: 'api:swagger:export', description: 'Dump the Swagger v2 documentation')]
```

chalasr: 🤦

alanpoulain (approved):

KDederichs:
Just pitching in to say that we had a discussion about the same thing in my original PR which got voted against because of deprecation notifications: https://github.com/api-platform/core/pull/4795

chalasr:
Thanks for the heads up. 

> Wouldn't this just annoy people with deprecations in logs when it's not possible to resolve them?

To answer this one: no, this won't lead to any deprecation notice in logs neither in core nor for end users, as long as the property is not extended (and if it is, that's ok, better get a resolvable notice than a BC break).

PR #4896, opened 2022-08-16
chore: add typehint to ProcessorInterface

alanpoulain:


src/State/ProcessorInterface.php
```diff
@@ -19,12 +19,11 @@
...
  * @author Antoine Bluchet <soyuka@gmail.com>
- * @experimental
  */
 interface ProcessorInterface
 {
     /**
      * Handle the state.
      */
-    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []);
+    public function process(object $data, Operation $operation, array $uriVariables = [], array $context = []): ?object;
```
vincentchalamon, line 28: Please also update phpdoc on 2.7 to be consistent

src/Doctrine/Common/State/RemoveProcessor.php
```diff
@@ -27,14 +27,19 @@ public function __construct(private readonly ManagerRegistry $managerRegistry)
...
     {
         if (!$manager = $this->getManager($data)) {
-            return;
+            return null;
         }
 
         $manager->remove($data);
         $manager->flush();
+
+        return null;
```
vincentchalamon, line 42: Why not returning the original object?

src/Symfony/Messenger/Processor.php
```diff
@@ -31,7 +31,7 @@ public function __construct(MessageBusInterface $messageBus)
         $this->messageBus = $messageBus;
     }
 
-    private function persist(mixed $data, array $context = []): mixed
+    private function persist(object $data, array $context = []): ?object
```
vincentchalamon, line 34: Please also update the `remove` method

tests/Fixtures/TestBundle/State/OperationResourceProcessor.php
```diff
@@ -30,7 +30,7 @@ public function __construct(private readonly ManagerRegistry $managerRegistry)
     {
     }
 
-    private function persist($data, array $context = [])
+    private function persist(object $data, array $context = []): object
```
vincentchalamon, line 33: Please also update the `remove` method

src/Symfony/Messenger/Processor.php
```diff
@@ -54,7 +54,7 @@ private function remove(mixed $data): void
         );
     }
 
-    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []): mixed
+    public function process(object $data, Operation $operation, array $uriVariables = [], array $context = []): ?object
```
vincentchalamon, line 57: To be consistent with `src/Doctrine/Common/State/RemoveProcessor.php`, you should return null in case of DeleteOperationInterface.

tests/Fixtures/TestBundle/State/OperationResourceProcessor.php
```diff
@@ -56,7 +56,7 @@ private function remove($data): void
         $manager->flush();
     }
 
-    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = [])
+    public function process(object $data, Operation $operation, array $uriVariables = [], array $context = []): ?object
```
vincentchalamon, line 59: To be consistent with `src/Doctrine/Common/State/RemoveProcessor.php`, you should return null in case of DeleteOperationInterface.

vincentchalamon (changes requested):

soyuka:
I'm in favor of mixed as it's definitely possible on an RPC route to use arrays, integers, strings, null or any data type in a POST operation according to the HTTP specification. If we use objects only we may end up using objects representing scalar values (for example `NullContent` or `BinaryContent`). Not sure there's a real benefit in doing so.

soyuka:
we decided to keep mixed

PR #4900, opened 2022-08-21
feat(symfony): Deprecate the `$exceptionOnNoToken` parameter in  `ResourceAccessChecker::__construct()`

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | -
| License       | MIT
| Doc PR        | -

Just syncing with symfony/security: https://github.com/symfony/symfony/pull/41965.
This argument is useless with the not-so-new authenticator system, and it must be `false` everywhere in Symfony 6.
The plan is to do the same in 3.0: deprecate the parameter entirely and throw if it's passed as `true`, then remove it in 4.0.
The impact of this deprecation is likely to be small since very few or nobody should be extending this class anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.7.0-rc.3
+
+* Symfony: deprecate the `$exceptionOnNoToken` parameter in `ResourceAccessChecker::__construct()` (#4900)
+
 ## 2.7.0-beta.5
 
 * Serializer: ignore no-operation on SerializeListener (#4828)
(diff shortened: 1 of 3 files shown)


vincentchalamon:
Why not removing this parameter and force its value to `false` in 3.0, as you mentionned in the trigger?:
> It will be the default and only supported value in 3.0

chalasr:
Because the current default value for this constructor param is `true`, changing it to `false` in 2.7 would be a BC break. 
As such, deprecating the parameter itself in 2.7 would not let a chance to the developer to opt-in and adopt the new behavior (`false`). 

So in order to be clean from a BC pov, we need to make it go away in 3 steps as for Symfony:
1- deprecate relying on the current default value to encourage adopting the new behavior (2.7)
2- change the default value + deprecate the parameter entirely (3.0)
3- remove it (4.0)

I'm going to submit the PR for step 2 on 3.0.

chalasr:
Fair enough, PR updated. See #4905 for the removal

soyuka:
thanks @chalasr

PR #4904, opened 2022-08-30
Fix typo “… is … is”

alexislefebvre:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -

I'm not sure about the expected text here, but the current text looks wrong.

### DIFF
--- a/src/Api/FilterInterface.php
+++ b/src/Api/FilterInterface.php
@@ -28,7 +28,7 @@ interface FilterInterface
      *   - type: the type of the filter
      *   - required: if this filter is required
      *   - strategy (optional): the used strategy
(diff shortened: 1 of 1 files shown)


src/Api/FilterInterface.php
```diff
@@ -28,7 +28,7 @@ interface FilterInterface
      *   - type: the type of the filter
      *   - required: if this filter is required
      *   - strategy (optional): the used strategy
-     *   - is_collection (optional): is this filter is collection
+     *   - is_collection (optional): is this filter a collection
```
alanpoulain, line 31:
```suggestion
     *   - is_collection (optional): if this filter is for collection
```

Issue #4911, opened 2022-09-02
`$example` should be `mixed` instead of `string|null`


nesl247:
**API Platform version(s) affected**: 3.0.0-rc2

**Description**  
In https://github.com/api-platform/core/blob/5ef1d612f34dce4cafd0aaec298e346db841c328/src/Metadata/ApiProperty.php#L32 it only supports `string|null`, however, the actual types can be `string|array|int|bool|null` at minimum at least. This is causing issues with `psalm`, even though PHP has no issues, and it works as expected.

soyuka:
example is a literal-string in open api and there's no point in using something else

nesl247:
@soyuka am I misunderstanding where this is used? I got this image from https://swagger.io/docs/specification/adding-examples/


![CleanShot 2022-09-05 at 09 42 42](https://user-images.githubusercontent.com/1037526/188463372-ab0c3b91-9874-42c4-9528-62aa3f68e9b8.png)

juniorb2ss:
@soyuka 🏓

soyuka:
Mhh right this actually refers to https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schema-object and indeed should be a free form object.

PR #4922, opened 2022-09-06
fix: #4911 make ApiProperty::$example `mixed` instead of `string|null`

nesl247:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | #4911 
| License       | MIT
| Doc PR        | -

In `ApiProperty` the `$example` is currently set to `string|null`, however, as discussed in #4911 the actual type can be mixed because OAI supports arrays, objects, etc. as an example. This is already allowed in `openapiContext.example` but this top level property does not allow it in docblock only. This causes an issue with psalm because it detects this inconsistency.


soyuka:
thanks !

Issue #4924, opened 2022-09-07
Memory leak with the normalizer


lyrixx:
**API Platform version(s) affected**: 2.6.8 (and other version too)
**Description**  

I found (the hard way) a memory leak.

APIP [automatically ships](https://github.com/api-platform/core/blob/v2.6.8/src/Serializer/AbstractItemNormalizer.php#L149) the IRI of a document in the `context`.

But Symfony serializes [all the context](https://github.com/symfony/symfony/blob/ef92fb5539f756d0183499f0de05fbcf00969a5f/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L703-L706) to compute the cache key.

So, every time one normalizes an object, a new cache key is generated, thus the `attributesCache` property keeps growing. And it never reset!

**Possible Solution**  

Symfony provides a way to remove some key from the context before computing the cache key.
IMHO, The normalizer should add the `uri` key in the `defaultContext`.

I'm not sure about the consequence, but I think it's safe, because the URI does not change the properties extracted.

WDYT?

lyrixx:
What branch should I target?

soyuka:
I think we should do this on 2.7 instead, 2.6 is in a freeze state and it'd require an hour to update (test suite broken, phpstan and cs updates etc.)f

PR #4925, opened 2022-09-07
[Serializer] Do not use the 'uri' while computing the cache key

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 2.6
| Tickets       | #4924
| License       | MIT
| Doc PR        |

---

BTW, it'll also speed normalization, since the cache will be re-used!


src/Bridge/Symfony/Bundle/Resources/config/api.xml
```diff
@@ -116,6 +116,11 @@
             <argument type="tagged" tag="api_platform.data_transformer" on-invalid="ignore" />
             <argument type="service" id="api_platform.metadata.resource.metadata_factory" on-invalid="ignore" />
             <argument type="service" id="api_platform.security.resource_access_checker" on-invalid="ignore" />
+            <argument type="collection">
+                <argument key="exclude_from_cache_key" type="collection">
+                    <argument>uri</argument>
```
tucksaun, line 121: Should it not be:
```suggestion
                    <argument>iri</argument>
```
?

PR #4928, opened 2022-09-08
Fix Graphql issues

webda2l:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/core/issues/4613#issuecomment-1210583154 & https://github.com/api-platform/core/issues/4613#issuecomment-1210634143
| License       | MIT

This restore the previous behavior of the 2.6


alanpoulain:
Hello,
Thanks for working on this!
Since it's two different issues, could you make two PR?
Also it would be great to have Behat tests to make sure it doesn't break later :slightly_smiling_face:

webda2l:
There are still issues with my use case before going further but yes, I will consider yours feedbacks :)

PR #4940, opened 2022-09-13
chore: warn about mis-configured resources in debug:api-resource command

azjezz:
Signed-off-by: azjezz <azjezz@protonmail.com>

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Symfony/Bundle/Command/DebugResourceCommand.php
+++ b/src/Symfony/Bundle/Command/DebugResourceCommand.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Symfony\Bundle\Command;
 
+use ApiPlatform\Metadata\Extractor\ResourceExtractorInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\QuestionHelper;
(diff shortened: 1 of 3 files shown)


src/Symfony/Bundle/Command/DebugResourceCommand.php
```diff
@@ -24,8 +25,12 @@
 
 final class DebugResourceCommand extends Command
 {
-    public function __construct(private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory, private readonly ClonerInterface $cloner, private $dumper)
-    {
+    public function __construct(
```
alanpoulain, line 28: We follow the style of Symfony: everything on the same line.

soyuka:
I don't get it, the resource metadata factory is already calling the extractor ?

azjezz:
depends on which resource metadata factory is used, attribute resource metadata factory doesn't call the extractor, and will create the resource even if it's not configured ( related to the internal discussion )

soyuka:
We use decoration on our `api_platform.metadata.resource.metadata_collection_factory` service and it should definitely call https://github.com/api-platform/core/blob/main/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php at some point.

azjezz:
> We use decoration on our api_platform.metadata.resource.metadata_collection_factory service and it should definitely call https://github.com/api-platform/core/blob/main/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php at some point.

Yes, but the attribute resource metadata collection factory result won't be empty, hence why the command doesn't fail, or emit a warning.

PR #4945, opened 2022-09-14
Drop dependency on fig/link-util

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | -
| License       | MIT
| Doc PR        | -

To tidy up deps a bit since this is not needed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
     "require": {
         "php": ">=7.1",
         "doctrine/inflector": "^1.0 || ^2.0",
-        "fig/link-util": "^1.0",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "psr/container": "^1.0 || ^2.0",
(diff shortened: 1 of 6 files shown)


dunglas (approved):
1 less dependency 🎉

Could you run PHP CS Fixer and amend the commit message (prefixing it with `chore: ` should make the linter happy) please?
Other issues don't look related.

nicolas-grekas:
> prefixing it with chore: should make the linter happy

Sorry I don't get what this means :grimacing: 
The diff generated by the cs-fixer is wrong: the class *is* used in that file. Dunno what could cause this false positive.

dunglas:
We are following the "conventional commits" spec to automatically generate changelogs. Commit messages have to be prefixed by the type of change. Here it's probably `chore: ` as it's not really a fix nor a new feature.

nicolas-grekas:
prefix added to the commit message

soyuka:
thanks @nicolas-grekas

PR #4951, opened 2022-09-15
fix: AbstractItemNormalizer 

pavel-m10c:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

```php
/** 
 * @var Car[]|null
*/
public ?array $cars;
```

Having `?array` type of member fields in entities breaks serialization when held value is `null` with docblock similar to the above where `Car` is API Resource.


soyuka:
Nice catch, we need to loop on types instead of just adding this. I'll dig into that saturday.

soyuka:
After thinking about this it was never really supported and it's way harder to support. We're going to follow this at

Issue #4966, opened 2022-09-17
`api:upgrade-resource` removes every filters but the last one in each resource file


emmanuelballery:
**API Platform version(s) affected**: 2.7.0

**Description**

Hi! I was expecting to upgrade a project to ApiPlatform 3. I just launched the `api:upgrade-resource` in it and saw in my git diff that every filters but the last ones were removed from my resources. It actually removed 200 `ApiFilter` attributes in my case ...

**How to reproduce**  

```bash
symfony new bookshop-api
cd bookshop-api
composer require api-platform/core "^2"
composer require doctrine
```

Create this entity in `src/Entity/`:

```php
<?php declare(strict_types=1);

namespace App\Entity;

use ApiPlatform\Core\Annotation\ApiFilter;
use ApiPlatform\Core\Annotation\ApiResource;
use ApiPlatform\Doctrine\Orm\Filter\NumericFilter;
use ApiPlatform\Doctrine\Orm\Filter\SearchFilter;
use Doctrine\ORM\Mapping\Column;
use Doctrine\ORM\Mapping\Entity;
use Doctrine\ORM\Mapping\GeneratedValue;
use Doctrine\ORM\Mapping\Id;

#[Entity]
#[ApiResource]
#[ApiFilter(SearchFilter::class, properties: ['name'])]
#[ApiFilter(NumericFilter::class, properties: ['count'])]
class Test
{
    #[Column]
    #[GeneratedValue]
    #[Id]
    public ?int $id = null;

    #[Column]
    public ?string $name = null;

    #[Column]
    public int $count = 0;
}
```

You'll have to install `phpunit` because this package is missing `nikic/php-parser` dependency.

```bash
composer require test
php bin/console api:up
```

The `SearchFilter` disappeared.

![image](https://user-images.githubusercontent.com/1687685/190865920-9e782c3a-c254-4f42-abdf-7b14df6e8abd.png)

soyuka:
Because filters didn't change much (except the namespace) I'd recommend to:
1. fix filters namespaces yourself
2. use the script 

You can also help us fixing the `api:upgrade-resource` command.

Thanks for the report.

emmanuelballery:
👍🏻

Also, `nikic/php-parser` is missing from the dependencies. It may be an edge case but I think it relies on the fact that `phpunit` is generally installed - well, not in my reproducer ...

soyuka:
added some user friendly messages at https://github.com/api-platform/core/pull/4968

PR #4970, opened 2022-09-17
fix(symfony): upgrade command removes filters

emmanuelballery:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #4966
| License       | MIT
| Doc PR        | -

Sorry for the noise.

The `api:upgrade-resource` command is only leaving the last `ApiFilter` attribute of each resource. I did not dig into `nikic/php-parser`but adding an `ApiFilter` removes the previous one.


soyuka:
so cool, thanks!

PR #4983, opened 2022-09-19
chore: commitlint only the Pull Request HEAD

soyuka:


.github/workflows/ci.yml
```diff
@@ -6,18 +6,27 @@ on:
 
 env:
   COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
vincentchalamon, line 9: You should set this env only on the concerned job, not globally

.github/workflows/ci.yml
```diff
@@ -6,18 +6,27 @@ on:
...
+    if: github.event_name == 'pull_request'
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          configFile: .commitlintrc
-      - uses: wagoid/commitlint-github-action@v5
+      - name: Run commitlint
+        run: | 
+            commit=$(gh api \
```
vincentchalamon, line 23: Did you check in `${{ github.event_path }}` var? It's a path of the pull_request JSON data. You may parse it using jq, instead of requesting the API

vincentchalamon (changes requested):

Issue #4975, opened 2022-09-18
Metadata cache uses `ArrayAdapter` instead of APCu


usu:
**API Platform version(s) affected**: 3.0.0

**Description**  
3.0 uses `ArrayAdapter` as Metadata Cache. In 2.7 the default was to utilize APCu when it was installed.

Was this changed by purpose? 

--> If **yes**, I couldn't find this documented. The [docu](https://api-platform.com/docs/core/performance/#enabling-the-metadata-cache) still says:

> If the Symfony Cache component is available (the default in the API Platform distribution), it automatically enables support for the best cache adapter available.

--> If **no**, I think the following commit introduced the issue:
https://github.com/api-platform/core/commit/c76720ebbbd9200ac9cda7dbff03de4b7c132189#

Or do I need to change any config on my side during 3.0 upgrade?

**How to reproduce**  
Upgrade to 3.0
Check symfony profiler for cache hit rate, which is significantly lower than before, as no Metadata is cached between separate requests.

emmanuelballery:
Hi!

`registerCacheConfiguration` in the `ApiPlatformExtension` is indeed breaking the main caching abilities of our apps.

https://github.com/api-platform/core/blob/c76720ebbbd9200ac9cda7dbff03de4b7c132189/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L499-L511

In `prod` we lost the filesystem cache. `ArrayAdapter` is not enough. Removing those lines from `ApiPlatformExtension` is fixing this caching mecanism because `config/metadata/resource.xml` is referencing the `cache.system` (`adapter.filesystem` in prod).

https://github.com/api-platform/core/blob/c76720ebbbd9200ac9cda7dbff03de4b7c132189/src/Symfony/Bundle/Resources/config/metadata/resource.xml#L65-L67

For big apps, `dev` is really slow, as each required metadata is created once for each request, adding seconds to the clock. Opening Swagger UI is not possible (well xdebug stops us at 30 secondes) and exporting the full `openapi` schema is... difficult.

@alanpoulain I'm really sorry to bother you. Since  you worked on this, would you be king enough to take a look when you'll have the time and maybe guide us to help fix this.

Thank you!

acirulis:
[...]

PR #4986, opened 2022-09-20
fix(symfony): metadata cache is broken in dev/prod

emmanuelballery:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4975
| License       | MIT

Hi!

Cache seems broken in 3.0.0. Metadatas are not stored in the filesystem nor in APCu, resulting in poor performances (see https://github.com/api-platform/core/issues/4975#issuecomment-1251668523).

Cache services are provided by each metadata XML file located in `src/Symfony/Bundle/Resources/config/`.

- `api_platform.cache.metadata.property` => `metadata/property.xml`
- `api_platform.cache.metadata.resource` => `metadata/resource_name.xml`
- `api_platform.cache.metadata.resource_collection` => `metadata/resource.xml`
- `api_platform.cache.route_name_resolver` => `metadata/api.xml`
- `api_platform.cache.identifiers_extractor` => seems unused
- `api_platform.elasticsearch.cache.metadata.document` => `elasticsearch.xml`

In `ApiPlatformExtension`, these files are loaded in `registerMetadataConfiguration` but then each cache is overrided in `registerCacheConfiguration` for an `ArrayAdapter` instead of the usual `cache.system`.

### DIFF
--- a/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -486,16 +486,16 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
 
     private function registerCacheConfiguration(ContainerBuilder $container): void
     {
-        if (!$container->hasParameter('kernel.debug') || !$container->getParameter('kernel.debug')) {
+        if ($container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug')) {
+            $container->register('api_platform.cache.metadata.property', ArrayAdapter::class)->addTag('cache.pool');
+            $container->register('api_platform.cache.metadata.resource', ArrayAdapter::class)->addTag('cache.pool');
+            $container->register('api_platform.cache.metadata.resource_collection', ArrayAdapter::class)->addTag('cache.pool');
+            $container->register('api_platform.cache.route_name_resolver', ArrayAdapter::class)->addTag('cache.pool');
+            $container->register('api_platform.cache.identifiers_extractor', ArrayAdapter::class);
+            $container->register('api_platform.elasticsearch.cache.metadata.document', ArrayAdapter::class);
+        } else {
             $container->removeDefinition('api_platform.cache_warmer.cache_pool_clearer');
         }
-
-        $container->register('api_platform.cache.metadata.property', ArrayAdapter::class)->addTag('cache.pool');
-        $container->register('api_platform.cache.metadata.resource', ArrayAdapter::class)->addTag('cache.pool');
-        $container->register('api_platform.cache.metadata.resource_collection', ArrayAdapter::class)->addTag('cache.pool');
-        $container->register('api_platform.cache.route_name_resolver', ArrayAdapter::class)->addTag('cache.pool');
-        $container->register('api_platform.cache.identifiers_extractor', ArrayAdapter::class);
-        $container->register('api_platform.elasticsearch.cache.metadata.document', ArrayAdapter::class);
     }
 
     private function registerDoctrineOrmConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader): void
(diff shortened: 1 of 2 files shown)


soyuka:
Thing is, in development mode we want symfony to clear the cache based on the mtime. Maybe that we should use ArrayAdapter only in development ? I must say I'm not that familiar with how Symfony handles cache.

webda2l:
> Thing is, in development mode we want symfony to clear the cache based on the mtime. Maybe that we should use ArrayAdapter only in development ? I must say I'm not that familiar with how Symfony handles cache.

With an advised cache.yaml like
```
framework:
    cache:
        # Unique name of your app: used to compute stable namespaces for cache keys.
        #prefix_seed: your_vendor_name/app_name

        # The "app" cache stores to the filesystem by default.
        # The data in this cache should persist between deploys.
        # Other options include:

        # Redis
        #app: cache.adapter.redis
        #default_redis_provider: redis://localhost

        # APCu (not recommended with heavy random-write workloads as memory fragmentation can cause perf issues)
        app: cache.adapter.apcu

        # Namespaced pools use the above "app" backend by default
        pools:
            app_cache.pool:
                adapter: cache.app

when@dev:
    framework:
        cache:
            app: cache.adapter.array
```

there shouldn't need of an hardcode ArrayAdapter use I think.

emmanuelballery:
> Thing is, in development mode we want symfony to clear the cache based on the mtime. Maybe that we should use ArrayAdapter only in development ? I must say I'm not that familiar with how Symfony handles cache.

But currently this destroys non debug environment. I updated my PR to only target non debug envs.

For debug env, it hurts a lot. I overrided api platform caches to go to the filesystem and will cache clear when my schema will change.

---

To add to the strangeness, `CachedResourceMetadataCollectionFactory` has a local cache implemented.

It:
- looks for a local cache entry in `$this->localCache`
- then fallbacks to `CacheItemPoolInterface` (meaning an`ArrayAdapter`) which does the same as the first step but in a different array
- fallbacks to creating the real collection

So we store the metadatas in 2 stores for no real benefit.

A clean solution could be to remove this `$this->localCache` and only depend on `CacheItemPoolInterface` then work on a strategy that is ok **per** environment:

- **non debug** cache could rely on `cache.system` (what my commmit does) and be a chain of:
  - ArrayAdapter
  - ApcuAdapter
  - FilesystemAdapter
- **debug** cache could be a chain of:
  - ArrayAdapter
  - (...) something that check `mtime` but based on what files?

soyuka:
> (...) something that check mtime but based on what files?

But symfony already has this on `cache.system` IIRC for dev environments.

emmanuelballery:
Symfony cannot tell that the cache used for `ResourceMetadataCollectionFactoryInterface` has to be reloaded __PER__ api resource. It's all or nothing - except if we create more specific caches.

I'm ok to work on improving the dev cache (probably by testing performances first). Maybe in another PR?

webda2l:
As @emmanuelballery I'll remove the localCache from CachedResourceMetadataCollectionFactory (and so add symfony/cache as dependancy in require, instead current require-dev)
And I'll remove completely the `$this->registerCacheConfiguration($container);` function of ApiPlatformExtension, let SF deal with the default `cache.system` that work well.
Then, it let user override this `cache.system` with their needs, by ex:

```
framework:
    cache:
        # APCu (not recommended with heavy random-write workloads as memory fragmentation can cause perf issues)
        system: cache.adapter.apcu

when@dev:
    framework:
        cache:
            # system: cache.adapter.array           Disabled because lower perf than the default cache.adapter.system
```

A distinct cache 'api_platform' (extends app.system) could be maybe added to the default app & system, to avoid modifying other use of the cache.system like above.

dunglas:
A hitting a local cache in memory will always be faster than calling the cache component. I would keep it (we also follow the same pattern in Symfony components btw).

emmanuelballery:
I did suggest to change things around the cache but the main concern of this PR is that v3 is **not production ready** as is. I tried to route these cache changes in another PR, but failed.

I'd suggest to focus on fixing non debug environments (or prod) here, which my commit does; and I'll happily bench the chain of `ResourceMetadataCollectionFactoryInterface` in another PR to see what could be improved for a better DX.

PR #4990, opened 2022-09-20
fix(symfony): BC layer broken for symfony/console lower than 5.3

deguif:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       |
| License       | MIT
| Doc PR        |

BC layer for symfony/console <= 5.3 is broken

When running `php bin/console` on a project with `symfony/console` lower than `5.3`, it throws this error :
```
Undefined constant Symfony\Component\Console\Input\InputOption::VALUE_NEGATABLE 
```


chalasr:
No longer using `negatable` should imply to define 2 options (for `dry-run` and `no-dry-run`) so technically this is a BC break as it removes an option. But given the nature of this command and because IMHO the negation is not relevant in this case I'm 👍 as-is.

dunglas (approved):

soyuka:
thanks @deguif

PR #4996, opened 2022-09-21
 feat: spec-compliant PUT method

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

Our current implementation of [the `PUT` HTTP method](https://httpwg.org/specs/rfc9110.html#PUT) is not standard-compliant: according to the spec, a `PUT` must create or replace the resource at the requested URL.

However, currently, we don't allow the creation of new resources using `PUT`, and we update the existing resources instead of replacing them: untouched properties are kept instead of being reset to their default values. Basically, our `PUT` behaves as `PATCH`.

The main reason behind that non-conform behavior is [a Doctrine ORM limitation](https://github.com/doctrine/orm/issues/8461).

This patch adds support for spec-compliant `PUT` and provides a workaround for Doctrine ORM.

Making `PUT` spec-compliant is opt-in to preserve backward compatibility, but using the non-conform version will be deprecated (use `PATCH` instead), and removed in API Platform 4.

Allowing resource creation is not enabled by default because it can cause security issues.

TODO:

* [x] Use dedicated attribute properties instead of relying on `extra_properties` 
* [x] Deprecate not setting `standard_put` to `true`
* [ ] Update the recipe to disable `PUT` by default (using `PATCH` is probably a better solution for most common use cases, so we should make `PUT` opt-in)


tests/Fixtures/TestBundle/Entity/StandardPut.php
```diff
@@ -0,0 +1,42 @@
...
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+
+/**
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ */
+#[ApiResource(extraProperties: [
+    'standard_put' => true,
+    'allow_create' => true,
```
soyuka, line 27: can't we have only one flag for this ?

src/Doctrine/Common/State/PersistProcessor.php
```diff
@@ -38,10 +44,32 @@ public function __construct(private readonly ManagerRegistry $managerRegistry)
...
+            !$manager = $this->managerRegistry->getManagerForClass($this->getObjectClass($data))
+        ) {
             return $data;
         }
 
+        // PUT: reset the existing object managed by Doctrine and merge data sent by the user in it
+        // This custom logic is needed because EntityManager::merge() has been deprecated and UPSERT isn't supported:
+        // https://github.com/doctrine/orm/issues/8461#issuecomment-1250233555
+        if (isset($context['read_data'], $context['resource_class']) && $operation instanceof HttpOperation && 'PATCH' !== $operation->getMethod()) {
+            $propertyNames = iterator_to_array($this->propertyNameCollectionFactory->create($context['resource_class'])->getIterator());
```
soyuka, line 58: we can avoid all this, you can retrieve identifiers by looping over `foreach $operation->getUriVariables() as $uriVariable` then `foreach $uriVariable->getIdentifiers() as $property` this way we don't use metadata factories here.

src/Symfony/EventListener/WriteListener.php
```diff
@@ -64,15 +67,21 @@ public function onKernelView(ViewEvent $event): void
...
 
         if (!$operation?->getProcessor()) {
             return;
         }
 
-        $context = ['operation' => $operation, 'resource_class' => $attributes['resource_class'], 'previous_data' => $attributes['previous_data'] ?? null];
+        $context = [
+            'operation' => $operation,
+            'resource_class' => $attributes['resource_class'],
+            'read_data' => $attributes['read_data'] ?? null,
```
soyuka, line 81: is `read_data` really needed?

tests/Fixtures/TestBundle/Entity/StandardPut.php
dunglas: I don't' think so. Even with standard-compliant PUT will be the only remaining behavior, you'll probably don't want to allow resource creation through PUT.

src/Doctrine/Common/State/PersistProcessor.php
dunglas: Awesome.

src/Symfony/EventListener/WriteListener.php
dunglas: Yes because we need a reference to the original fetched data (not a clone). And it is lost in DeserializeListener is the serializer creates a new object.

That being said I didn't try with the clone, but I double it works (I'll check).

dunglas: I've found a hack to get rid of it.

src/Doctrine/Common/State/PersistProcessor.php
dunglas: My refactoring doesn't handle the case where URI variables don't match the Doctrine properties, but I think that it's good enough for now, mainly because it's in sync with what is currently done here: https://github.com/api-platform/core/blob/main/src/Doctrine/Orm/State/ItemProvider.php#L51

At some point, we'll probably want to introduce a utility to map URI variables with internal Doctrine identifiers.

src/Symfony/EventListener/ReadListener.php
```diff
@@ -38,10 +39,12 @@ final class ReadListener
     use OperationRequestInitiatorTrait;
     use UriVariablesResolverTrait;
 
-    public const OPERATION_ATTRIBUTE_KEY = 'read';
-
-    public function __construct(private readonly ProviderInterface $provider, ?ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, private readonly ?SerializerContextBuilderInterface $serializerContextBuilder = null, UriVariablesConverterInterface $uriVariablesConverter = null)
-    {
+    public function __construct(
```
alanpoulain, line 42: I thought we didn't want to have multiple lines in the constructor?

src/Symfony/EventListener/ReadListener.php
```diff
@@ -90,7 +93,13 @@ public function onKernelRequest(RequestEvent $event): void
             throw new NotFoundHttpException('Invalid identifier value or configuration.', $e);
         }
 
-        if (null === $data) {
+        if (
+            null === $data &&
+            (
+                HttpOperation::METHOD_PUT !== $operation->getMethod() ||
```
alanpoulain, line 99: Why using a constant here and not elsewhere?

src/Symfony/EventListener/ReadListener.php
dunglas: Now that we have constructor promotion and the ability to add a comma after the last parameter, I think we should use multiple lines for clarity and to reduce merge conflicts.

alanpoulain: Good to know, it would be great to add it in the CS fixer.

src/Symfony/EventListener/DeserializeListener.php
```diff
@@ -69,7 +69,13 @@ public function onKernelRequest(RequestEvent $event): void
 
         $format = $this->getFormat($request, $operation?->getInputFormats() ?? []);
         $data = $request->attributes->get('data');
-        if (null !== $data) {
+        if (
+            null !== $data &&
+            (
+                'PATCH' === $method ||
+                ('PUT' === $method && !($operation->getExtraProperties()['standard_put'] ?? false))
```
soyuka, line 76: use `HttpOperation` constants for methods

tests/Fixtures/TestBundle/Entity/StandardPut.php
```diff
@@ -0,0 +1,40 @@
...
+use ApiPlatform\Metadata\ApiResource;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+
+/**
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ */
+#[ApiResource(extraProperties: [
+    'standard_put' => true,
```
soyuka, line 25: I find the name confusing

src/Doctrine/Common/State/PersistProcessor.php
```diff
@@ -38,10 +39,39 @@ public function __construct(private readonly ManagerRegistry $managerRegistry)
...
+            !\is_object($data) ||
+            !$manager = $this->managerRegistry->getManagerForClass($class = $this->getObjectClass($data))
+        ) {
             return $data;
         }
 
+        // PUT: reset the existing object managed by Doctrine and merge data sent by the user in it
+        // This custom logic is needed because EntityManager::merge() has been deprecated and UPSERT isn't supported:
+        // https://github.com/doctrine/orm/issues/8461#issuecomment-1250233555
+        if ($operation instanceof HttpOperation && 'PUT' === $operation->getMethod()) {
```
soyuka, line 52:
```suggestion
        if ($operation instanceof HttpOperation && HttpOperation::METHOD_PUT === $operation->getMethod()) {
```

src/Metadata/Put.php
```diff
@@ -92,8 +92,21 @@ public function __construct(
         ?string $name = null,
         $provider = null,
         $processor = null,
-        array $extraProperties = []
+        array $extraProperties = [],
+        private ?bool $allowCreate = null,
```
dunglas, line 96: IMHO the order of these parameters shouldn't be part of the BC promise and users **must** have to use named properties.

soyuka: note that this allows to use `...func_get_args()` to call `parent::__construct` without having to be concerned about `allowCreate`.

PR #5012, opened 2022-09-25
chore(cs): Improve CS

webda2l:
| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

First commit is a must have IMO with some classes having really long line of _construct currently.
Second commit is more optional. I use @PhpCsFixer & @PhpCsFixer:risky fully personally but restricted some rule here to limit changes.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -38,10 +38,11 @@ return (new PhpCsFixer\Config())
         '@PHPUnit84Migration:risky' => true,
         '@Symfony' => true,
         '@Symfony:risky' => true,
+        '@PhpCsFixer' => true,
         'align_multiline_comment' => [
             'comment_type' => 'phpdocs_like',
         ],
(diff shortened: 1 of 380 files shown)


alanpoulain:
Personally I like these changes!

dunglas:
We should apply these changes on 2.7 or we have very bad time merging bug fixes in higher branches...

I like these changes too BUT I'm not sure they are worth the effort (doing the merge between 2.7 and 3.0 will be hard).

alanpoulain:
I think doing it in 2.7 and merging it upstream would be hardest.
Maybe the best course of action would to do it just before the 3.1 release?

dunglas:
That indeed looks like the best timing to do it. So we'll not have to care about 2.7 anymore.

soyuka:
@webda2l API Platform 3.1 is around the corner, I've taken your changes at https://github.com/api-platform/core/pull/5330

PR #5024, opened 2022-09-28
fix: compatibility with PHP 8.2

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Prevents this error:

> Compile Error: Type Traversable|object|array|null contains both object and a class type, which is redundant

Related to https://github.com/php/php-src/issues/9556.
Please note that as described in this issue, this change isn't a BC break and doesn't break existing code even on PHP 8.1.

### DIFF
--- a/src/GraphQl/Resolver/Stage/ReadStage.php
+++ b/src/GraphQl/Resolver/Stage/ReadStage.php
@@ -43,7 +43,7 @@ final class ReadStage implements ReadStageInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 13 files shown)


dunglas:
> LGTM, just wonder if changing the return-type of the ProviderInterface doesn't introduce a BC break...

No, it doesn't: https://github.com/php/php-src/issues/9556#issuecomment-1249453883

Issue #5035, opened 2022-09-29
Skolem IRI and HTTP cacheability


tucksaun:
**API Platform version(s) affected**: 2.6 (at least) and above

**Description**
When using JSON-LD, API Platform automatically exposes IRIs, but it also exposes some IRIs for anonymous resources which are not stable, leading to the body content changing from one answer to the other.
This is not desirable IMHO because it makes debugging and testing harder and also this results in the ETag header value being unstable which bust HTTP caches.

Before 2.7, the IRIs generation mechanism for anonymous resources relies on `spl_object_id` so this could lead to the issue when the overall number or order of PHP objects changes (cache clearing or cache miss for instance).
Since 2.7, the IRI generation relies on `random_bytes` so the IRI for anonymous resources becomes totally unstable preventing any HTTP caching.

Maybe I missed something but trying to override the JSON-LD context is currently of no use and the only workarounds I found are:
- override `ApiPlatform\JsonLd\ContextBuilder` and unset the `@id` key manually,
- or override `ApiPlatform\Symfony\Routing\SkolemIriConverter` to return a constant (empty) string.

**How to reproduce**
Expose an API Resource with a Doctrine embedded or a value object and hit the API using JSON-LD.
You can have a look at https://github.com/tucksaun/api-platform-iri-reproducer.

**Possible Solution**  
IIUC the JSON-LD specs, for blank nodes the `@id` key is only recommended and not mandatory (https://www.w3.org/TR/json-ld11/#identifying-blank-nodes), and is useful essentially if one wants to use node references to them which is not done automatically by API Platform for anonymous resources as far as I know

So one solution could be to allow disabling the addition of Skolem IRI to anonymous resources globally or per subresource.

I even consider that (**if my assumption about `@id` not being strictly required is correct**) the use case where one wants to compare anonymous resources is probably an advanced use case and as such adding IRI to anonymous resources should probably be made an opt-in option?

**Additional Context** 
Produced output (unstable):
```
{
  "@context": "/contexts/Greeting",
  "@id": "/greetings/1",
  "@type": "Greeting",
  "id": 1,
  "name": "Tugdual",
  "options": {
    "@type": "Options",
    "@id": "/.well-known/genid/493e0689a8264f70af65",
    "notify": false
  },
  "foo": {
    "@type": "ValueObject",
    "@id": "/.well-known/genid/659262289b3472742a7a",
    "bar": 42
  }
}
```

headers (unstable):
```
...
[...]

PR #5046, opened 2022-10-03
fix(metadata): no skolem iri by default

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | bug fix (bc break)
| Tickets       | #5035
| License       | MIT

On 3.0 this behavior will be the default one, I'll add it only in 3.0 so that one can disable the skolemIri generation. We break the compatibility here, so let's revert for 2.7.

### DIFF
--- a/src/Metadata/ApiProperty.php
+++ b/src/Metadata/ApiProperty.php
@@ -92,6 +92,7 @@ final class ApiProperty
 
     private $schema;
     private $initializable;
+    private $skolemIri;
 
     /**
      * @var string[]
@@ -146,6 +147,7 @@ final class ApiProperty
         ?array $builtinTypes = null,
         ?array $schema = null,
         ?bool $initializable = null,
+        ?bool $skolemIri = null,
 
         $iris = null,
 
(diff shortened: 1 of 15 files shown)


tucksaun:
Thank you @soyuka for working on a fix for this issue 🙏 

Though, regarding the "fix" for 3.0, may I suggest the following patch:
https://github.com/api-platform/core/compare/3.0...tucksaun:api-platform-core:fix-api-property-iris-anonymous-resources?expand=1

It should remove the need for patching `ApiPlatform\Metadata\ApiProperty` that's why I'm commenting here ;)
If the approach looks viable to you I can open the PR stat.

src/Metadata/ApiProperty.php
```diff
@@ -146,6 +147,7 @@ public function __construct(
         ?array $builtinTypes = null,
         ?array $schema = null,
         ?bool $initializable = null,
+        ?bool $skolemIri = null,
```
soyuka, line 150: > genid 

Si `null` ou `false` (par défaut) => on affiche pas le @id
Si true => id généré

soyuka:
@tucksaun thanks for that I added it to the fix here, `iri` is too confusing however so let's fix this properly by adding a new `genid` option to `ApiProperty`. It'll default to `false` in 2.7 and `true` in 3.0.

tucksaun:
Finally got the chance to give it a go and it seems everything is working properly.

Thank you @soyuka!
And good luck with 3.0 stabilization 🤗

Issue #4897, opened 2022-08-16
When having GraphQL, an exception occurs when Twig is not enabled


rvanlaak:
**API Platform version(s) affected**: 3.0.0-rc.2

## Description  
The API Platform GraphiQL integration seems to have a dependency on the `twig` service, but that leads to an compilation exception when Twig is not enabled for the active environment.

<img width="872" alt="image" src="https://user-images.githubusercontent.com/2707563/184863622-1673a435-0fc2-469a-ad09-2f94f497b627.png">
 
## How to reproduce

* require webonyx/graphql-php to enable GraphQL for api-platform
* but do not enable TwigBundle in bundles.php for the active environment (e.g. `prod`).

## Possible Solution

Do not enable the visual graphiQL playground when Twig service is not available or bundle is not enabled.

## Additional Context

see documentation https://api-platform.com/docs/core/graphql/#graphiql

Our use-case is that we do not want to have Twig enabled on our `prod` environment because we only support json and xml content types.

* Enabling TwigBundle in bundles.php for all environments solves the exception but is not preferable
* Disabling graphiql does not solve the exception:
<img width="655" alt="image" src="https://user-images.githubusercontent.com/2707563/184864562-cb9c792e-4f44-4fdb-b8ad-37884c1b13a9.png">

soyuka:
I think that this was already the case on API P 2.6 no ?

rvanlaak:
Possibly, yes. But won't have the resources to verify this, as we are working on adopting 3.x

soyuka:
I guess

PR #5064, opened 2022-10-13
[GraphQL] Add a clearer error message when TwigBundle is disable but graphql clients aren't

ArnoudThibaut:
| Q             | A
| ------------- | ---
| Branch?       | 3.0 
| Tickets       | https://github.com/api-platform/core/issues/4897 
| License       | MIT

Instead of having an error thrown by the CheckExceptionOnInvalidReferenceBehaviorPass when Twig is not installed or enabled we send a clearer exception and force the user to either disable the GraphQL clients or enable Twig.

If the clients are not enabled we remove the definition as they can't be instantiate and we pass null to the GraphQl action.

Maybe we should not trowing an exception and instead disable silently the the clients if Twig is not enabled but I find better to add an explicit error message.

### DIFF
--- a/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -53,6 +53,7 @@ use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter
 use Symfony\Component\Uid\AbstractUid;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Component\Yaml\Yaml;
+use Twig\Environment;
 
 /**
  * The extension of this bundle.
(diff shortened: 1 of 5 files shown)


ArnoudThibaut:
> nice patch, I think that the phpstan issue is related though

Yes it's totally related but I think it's because of phpsant.neon.dist. Inside we specify to phpstan to use the container of the test env cache and in test the parameter `kernel.bundles` contains the key `TwigBundle` so for phpstan  the isset it useless but that's not true

Do you have any idea on how to fix that without adding a `@phpstan-ignore-next-line` ?

alanpoulain:
In this specific case, I think a `@phpstan-ignore-next-line` with an explanation is in order.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -476,6 +480,15 @@ private function registerGraphQlConfiguration(ContainerBuilder $container, array
 
         $loader->load('graphql.xml');
 
+        // @phpstan-ignore-next-line We ignore the next line because we specify to phpstan to use the container of the test env cache and in test the parameter kernel.bundles always contains the key TwigBundle
```
alanpoulain, line 483:
```suggestion
        // @phpstan-ignore-next-line because PHPStan uses the container of the test env cache and in test the parameter kernel.bundles always contains the key TwigBundle
```

alanpoulain (approved):

alanpoulain:
Thanks @ArnoudThibaut.

PR #5070, opened 2022-10-14
feat(filters): Allow filter by IRI for guid fields in SearchFilter

alexndlm:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | 
| License       | MIT
| Doc PR        | api-platform/docs#1628

We need this PR in `2.7` branch to migrate to `3.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.7.3
+* Filters: Allow filter by IRI for guid fields in SearchFilter
+
 ## 2.7.2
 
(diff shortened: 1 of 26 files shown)


soyuka:
Regarding this, I'm afraid that we won't be merging this, especially on 2.7. You should not change any of our existing features in your tests. Also, we want to refactor the search filter at some point and fix this kind of issues. The thing is, it's quite tricky and potentially breaks lots of user code (we recently had a PR trying to fix a type-related issue that broke everything). Last but not least, this is a really specific use case and you'll be better fixing this in your code for now.

PR #5079, opened 2022-10-21
[Doctrine] SearchFilter - Use abitrary index instead of value

Korbeil:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

When using a search filter with multiple values that have the same lowercase identity, for example:
```
array:2 [▼
  0 => "rayon bleu"
  1 => "Rayon Bleu"
]
```

SearchFilter will lowercase the value and use it as index key, which makes it send only one parameter to Doctrine when it needs two of them.

I choosed to leave the arbitrary array index so we don't have this kind of issues anymore.

### DIFF
--- a/src/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Doctrine/Orm/Filter/SearchFilter.php
@@ -179,7 +179,7 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
         $parameters = [];
         foreach ($values as $key => $value) {
             $keyValueParameter = sprintf('%s_%s', $valueParameter, $key);
-            $parameters[$caseSensitive ? $value : strtolower($value)] = $keyValueParameter;
+            $parameters[] = [$caseSensitive ? $value : strtolower($value), $keyValueParameter];
 
             $ors[] = match ($strategy) {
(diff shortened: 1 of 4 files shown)


welcoMattic (approved):

alanpoulain:
Could you please add unit tests?

Korbeil:
> Could you please add unit tests?

Is that enough ? Not sure what I could add more :thinking:

alanpoulain:
> Is that enough ? Not sure what I could add more thinking

Yes, that's enough :slightly_smiling_face: You need to change the `SearchFilterTest` for MongoDB too (sorry for that, please tell me if you want me to do it).

soyuka:
how was this in 2.6 ? I think that we could fix this on 3.0 instead, 2.7 should have only bc-layer related fixes.

Korbeil:
> how was this in 2.6 ? I think that we could fix this on 3.0 instead, 2.7 should have only bc-layer related fixes. Unless this was wrongly introduced in 2.7 we may fix it here.

I do think this was here since a long time, I'll change the branch to `3.0` then.

soyuka:
Thanks @Korbeil !

Issue #2933, opened 2019-07-12
[BUG 2.4.4] Inheritance abstract class


baudev:
The regression appeared in version 2.4.4. Indeed, the code below is fully functional with version 2.4.3. I've discussed with @teohhanhui on Slack about this issue.

On the image, we can see the inheritance relationship between the different objects. `Teachr` and `Client` inherit from `User` which is an abstract class (codes are below).
There is a relationship between the `User` and `Address` objects. 

![image](https://user-images.githubusercontent.com/29781702/61142886-1c370a00-a4d1-11e9-9b2e-19931f2584c7.png)

`User`: 
```php
/**
 * The User class.
 *
 * @ORM\Entity()
 * @ORM\Table()
 * @InheritanceType("SINGLE_TABLE")
 * @DiscriminatorColumn(name="discr", type="string")
 * @DiscriminatorMap({"user" = "User", "client" = "Client", "teachr" = "Teachr"})
 */
abstract class User implements UserInterface
{

   /** ... */

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\Address", cascade={"persist", "remove"}, mappedBy="owner")
     * @Groups({"read_client", "read_teachr"})
     */
    private $addresses;

}
```

`Client` (same for `Teachr`):
```php
/**
 * Class for Clients.
 *
 * @ApiResource()
 * @ORM\Entity()
 */
class Client extends User
{

}
```

`Address`:
```php
/**
 * Class Address.
 *
 * @ApiResource()
 * @ORM\Entity()
 */
class Address
{

    /** ... */

    /**
     * @var User Client or Teach'r address owner.
     *
     * @ORM\ManyToOne(targetEntity="App\Entity\User", inversedBy="addresses")
     * @Groups({"read_address", "write_address"})
     */
    private $owner;

}
```

---

The error:

```json
{
  "@context": "/contexts/Error",
  "@type": "hydra:Error",
  "hydra:title": "An error occurred",
  "hydra:description": "Cannot instantiate abstract class App\\Entity\\User",
  "trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
      "file": "/var/www/html/vendor/symfony/serializer/Normalizer/AbstractNormalizer.php",
      "line": 452,
      "args": []
    },
    {
      "namespace": "",
      "short_class": "ReflectionClass",
      "class": "ReflectionClass",
      "type": "->",
      "function": "newInstanceArgs",
[...]

PR #5085, opened 2022-10-23
Serializer: don't use abstract class as resource_class on denormalize

ottaviano:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #2933
| License       | MIT
| Doc PR        | 

as we can see in the issue #2933 during the deserialization process the abstract class is taken and used as `resource_class` for the rest of the process. It will crash when it will try to instantiate the object.

I'm doing a quick fix to unlock this point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.7.3
+
+* Serializer: don't use abstract class as resource_class on denormalization
+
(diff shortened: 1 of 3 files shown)


soyuka:
It's not that simple, first it breaks our test for this, then we need to use the abstract class for metadata but the child class for normalization. Also this should be fixed on 3.0.

PR #5087, opened 2022-10-24
chore: adding of stale bot

davy-beauzil:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | 

A stale bot added to remove automatically issues without activity.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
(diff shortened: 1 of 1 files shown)


alanpoulain:
Personnally I'm against it.
We are not enough active to have a stale bot: a lot of valid issue will be closed without us giving any attention to them.

dunglas:
But we don't read them anyway because they are lost in the flow of invalid issues, and most issues are not related to API Platform 3 anyway.

The bot will ask the user it the issue is still there, and confirmed issues (with a label) aren't closed.

soyuka:
Let's try this if we don't like it we can still move backwards.

PR #5095, opened 2022-10-26
Bugfix: Respect input DTO without adding additional id field

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5097 #2736
| License       | MIT
| Doc PR        | -

For all GraphQL mutations except  `create` the `id` field is added automatically. Also for scenarios where an input DTO is used the `id` got added. This PR respects the given input DTO and will not add an additional `id` field. If you need the id your DTO may contain it.

The related Ticker #2736 is not completely solved with this PR but it enables the workaround to have multipe create mutations by using a specifiy input DTO.

Branch: 
- `main`, because not sure if it is a real bugfix or more a tiny new feature.

Note: Failing PHP-cs-fixer is not related to the changes in this PR

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -195,7 +195,7 @@ final class FieldsBuilder implements FieldsBuilderInterface, FieldsBuilderEnumIn
             return $fields;
         }
 
-        if (!$input || 'create' !== $operation->getName()) {
+        if (!$input || ('create' !== $operation->getName() && !$operation->getInput())) {
             $fields['id'] = $idField;
         }
         if ($input && $depth >= 1) {
--- a/tests/GraphQl/Type/FieldsBuilderTest.php
+++ b/tests/GraphQl/Type/FieldsBuilderTest.php
@@ -697,6 +697,31 @@ class FieldsBuilderTest extends TestCase
                     'clientMutationId' => GraphQLType::string(),
                 ],
             ],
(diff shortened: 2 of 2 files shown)


alanpoulain:
It's totally `main`, since it can even be considered as a breaking change.

Why not using an allow list like suggested in the issue?

jotwea:
Thanks @alanpoulain for your quick feedback.

Maybe it was a mistake from my side to promote this issue as a workaround for #2736.
The main issue is #5097, which I created just know to explain the in my opinion wrong behaviour when using DTOs. So when I use DTOs for input I expect that this DTO will be exactly used as an input for the GraphQL Mutation and no `id` should be automatically added. If the mutation needs one it can be specified as a property within the DTO. So this PR is mainly fixing  issue #5097.

But as mentioned it can additionally be considered as a workaround for #2736. But you are right, it does not mainly fix this issue and the propsal using an allow list will be still open. Sorry for misunderstanding.

alanpoulain (approved):

alanpoulain:
Thank you @jotwea.

jotwea:
I guess #5359 also solves the original problem mentioned in #5097 very well.

Your approach differs from mine that it is based on the existence of a `resolver`, mine is based on the existence of an `input DTO`. And you are right, I use a custom resolver as well in my project, so your solutions will work for me as well. The important thing is just that there will be a way to avoid an automatic `id` field.

alanpoulain:
Thanks for the answer.
It will be available in version 3.2.

PR #5096, opened 2022-10-26
fix(symfony): call item data provider for legacy api resource

### LINKED ISSUE #5094: use new IriConverter with legacy ApiResource
We works on updating ApiPlatform to 2.7 with the `metadata_backward_compatibility_layer` flag to `false`.
We have an error `Provider not found on operation "%s"` when we call the function getResourceFromIri on a legacy ApiResource that have a custom DataProvider.

Shouldn't the new IriConverter be able to call either the CallableProvider or the ChainItemDataProvider in 2.7 as a bridge to use legacy ApiResource is still available without the backward compatibility layer?

It's a problem when we add an iri of this resource in a payload of a parent resource.

louismariegaborit:
fix #5094


louismariegaborit:
anyone to help me with behat tests ?

soyuka:
Explained in the issue, this is wrong enable the backward compatibility layer for this to work.

PR #5100, opened 2022-10-27
fix(serializer): empty object as array with supports cache 

soyuka:
reverts #4999 and re-implement it

### DIFF
--- a/src/Hydra/Serializer/CollectionFiltersNormalizer.php
+++ b/src/Hydra/Serializer/CollectionFiltersNormalizer.php
@@ -23,7 +23,6 @@ use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
-use Symfony\Component\Serializer\Serializer;
 
 /**
  * Enhances the result of collection by adding the filters applied on collection.
@@ -60,8 +59,7 @@ final class CollectionFiltersNormalizer implements NormalizerInterface, Normaliz
      */
     public function normalize(mixed $object, string $format = null, array $context = []): array|string|int|float|bool|\ArrayObject|null
     {
-        $preserveEmpty = ($context[Serializer::EMPTY_ARRAY_AS_OBJECT] ?? false) || ($context[AbstractObjectNormalizer::PRESERVE_EMPTY_OBJECTS] ?? false);
-        if ($preserveEmpty && $object instanceof \ArrayObject && !\count($object)) {
+        if (($context[AbstractObjectNormalizer::PRESERVE_EMPTY_OBJECTS] ?? false) && $object instanceof \ArrayObject && !\count($object)) {
             return $object;
         }
 
--- a/src/Hydra/Serializer/CollectionFiltersNormalizer.php
+++ b/src/Hydra/Serializer/CollectionFiltersNormalizer.php
@@ -19,9 +19,11 @@ use ApiPlatform\Api\ResourceClassResolverInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use Psr\Container\ContainerInterface;
 use Symfony\Component\Serializer\Exception\UnexpectedValueException;
+use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
 use Symfony\Component\Serializer\Normalizer\CacheableSupportsMethodInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Serializer\Serializer;
 
 /**
  * Enhances the result of collection by adding the filters applied on collection.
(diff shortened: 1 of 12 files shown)


src/Hydra/Serializer/CollectionNormalizer.php
```diff
@@ -103,6 +108,23 @@ public function normalize(mixed $object, string $format = null, array $context =
...
     public function hasCacheableSupportsMethod(): bool
     {
-        return false;
+        return true;
+    }
+
+    /**
+     * Normalizes a raw collection (not API resources).
+     */
+    protected function normalizeRawCollection(iterable $object, string $format = null, array $context = [])
```
alanpoulain, line 117: Shouldn't we add the typehint?

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -51,7 +51,9 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
 
             foreach ($metadata->getOperations() ?? [] as $name => $operation) {
                 $isCollection = $operation instanceof CollectionOperationInterface;
-                if ('' === $operationName && \in_array($operation->getMethod() ?? HttpOperation::METHOD_GET, [HttpOperation::METHOD_GET, HttpOperation::METHOD_OPTIONS, HttpOperation::METHOD_HEAD], true) && ($forceCollection ? $isCollection : !$isCollection)) {
+                $method = $operation->getMethod() ?? HttpOperation::METHOD_GET;
+                $isGetOperation = HttpOperation::METHOD_GET === $method || HttpOperation::METHOD_OPTIONS === $method || HttpOperation::METHOD_HEAD === $method;
+                if ('' === $operationName && $isGetOperation && ($forceCollection ? $isCollection : !$isCollection)) {
```
alanpoulain, line 56: I'm taking this PR as an opportunity for a long time question I have: is adding an empty string as key in the `operationCache` a good idea?

soyuka: I need to prefix these, but yeah the representation of `null` as string would be an empty string ?

src/Hydra/Serializer/CollectionNormalizer.php
soyuka: right thanks!

src/Hydra/Serializer/CollectionFiltersNormalizer.php
```diff
@@ -58,13 +60,19 @@ public function hasCacheableSupportsMethod(): bool
      */
     public function normalize(mixed $object, string $format = null, array $context = []): array|string|int|float|bool|\ArrayObject|null
     {
-        $data = $this->collectionNormalizer->normalize($object, $format, $context);
-        if (!\is_array($data)) {
-            throw new UnexpectedValueException('Expected data to be an array');
+        $preserveEmpty = ($context[Serializer::EMPTY_ARRAY_AS_OBJECT] ?? false) || ($context[AbstractObjectNormalizer::PRESERVE_EMPTY_OBJECTS] ?? false);
```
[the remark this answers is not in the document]
soyuka: mb you're right

src/Metadata/Resource/ResourceMetadataCollection.php
alanpoulain: I'm not sure to follow: if we ask a `null` `$operationName` to this `getOperation` method, it means we don't know the operation name. It it is cached, the same operation is always returned, even if the other conditions have changed (`$isGetOperation`, `$forceCollection`...). It seems dangerous to me.

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -24,6 +24,8 @@
  */
 final class ResourceMetadataCollection extends \ArrayObject
 {
+    private const GRAPHQL_PREFIX = 'graphql_';
+    private const HTTP_PREFIX = 'http_';
```
dunglas, line 28:
```suggestion
    private const GRAPHQL_PREFIX = 'g_';
    private const HTTP_PREFIX = 'h_';
```

As it's entirely internal, let's save a few bytes.

src/Serializer/AbstractCollectionNormalizer.php
```diff
@@ -82,6 +87,23 @@ public function normalize(mixed $object, string $format = null, array $context =
         return array_merge_recursive($data, $paginationData, $itemsData);
     }
 
+    /**
+     * Normalizes a raw collection (not API resources).
+     */
+    protected function normalizeRawCollection(iterable $object, string $format = null, array $context = []): array|\ArrayObject
+    {
+        if (\is_array($object) && !$object && ($context[Serializer::EMPTY_ARRAY_AS_OBJECT] ?? false)) {
```
dunglas, line 95: The call to `is_array` is probably the slower, the put it at the end.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -511,13 +513,22 @@ protected function getFactoryOptions(array $context): array
...
-            $operation = $context['operation'] ?? $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation($context['operation_name'] ?? null);
-            $options['normalization_groups'] = $operation->getNormalizationContext()['groups'] ?? null;
-            $options['denormalization_groups'] = $operation->getDenormalizationContext()['groups'] ?? null;
-            $options['operation_name'] = $operation->getName();
+        // This is a hot spot
+        if (isset($context['resource_class'])) {
+            $operationCacheKey = $context['resource_class'].($context['operation_name'] ?? '');
+            $operation = $context['operation'] ?? $this->localOperationCache[$operationCacheKey] ?? null;
+
+            if (!isset($this->localOperationCache[$operationCacheKey]) && !$operation && $this->resourceClassResolver->isResourceClass($context['resource_class']) && $this->resourceMetadataCollectionFactory) {
```
dunglas, line 521: Same here, `$this->resourceMetadataCollectionFactory()` is the slowest so should be called last.

src/Metadata/Resource/ResourceMetadataCollection.php
soyuka: Let's fix this in another patch then.

alanpoulain (approved):

PR #4930, opened 2022-09-08
fix(graphql): operation with disabled pagination

webda2l:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | https://github.com/api-platform/core/issues/4613#issuecomment-1210583154 & https://github.com/api-platform/core/issues/4613#issuecomment-1210634143
| License       | MIT
| Doc PR        | api-platform/docs#... 

- [ ] Add tests

PR #5102, opened 2022-10-27
fix(graphql): use right nested operation

alanpoulain:
See https://github.com/api-platform/core/issues/4613#issuecomment-1210634143.

Supersedes https://github.com/api-platform/core/pull/4930.

Fixes https://github.com/api-platform/api-platform/issues/2296.

The principal issue was mainly because the root operation was used instead of the nested resource one.

This PR also improves how GraphQL handles a nested resource when the query operation does not exist.
Instead of having a fallback on the REST operation (and an exception if it does not exist), it creates a "dynamic" resource having the default operations.
This dynamic resource is created using a new `DynamicExtractor` with an `addResource` method, which can take an optional configuration array.

In the future, I think this could be useful to control how a nested resource is handled by GraphQL (for instance by using the `ApiProperty` attribute).

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -17,12 +17,10 @@ use ApiPlatform\Api\ResourceClassResolverInterface;
 use ApiPlatform\Exception\OperationNotFoundException;
 use ApiPlatform\GraphQl\Resolver\Factory\ResolverFactoryInterface;
 use ApiPlatform\GraphQl\Type\Definition\TypeInterface;
+use ApiPlatform\Metadata\Extractor\DynamicResourceExtractorInterface;
 use ApiPlatform\Metadata\GraphQl\Mutation;
 use ApiPlatform\Metadata\GraphQl\Operation;
-use ApiPlatform\Metadata\GraphQl\Query;
-use ApiPlatform\Metadata\GraphQl\QueryCollection;
 use ApiPlatform\Metadata\GraphQl\Subscription;
-use ApiPlatform\Metadata\Operation as AbstractOperation;
 use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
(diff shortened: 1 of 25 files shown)


src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -47,7 +45,7 @@
...
+    public function __construct(private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory, private readonly DynamicResourceExtractorInterface $dynamicResourceExtractor, private readonly ResourceClassResolverInterface $resourceClassResolver, private readonly TypesContainerInterface $typesContainer, private readonly TypeBuilderInterface $typeBuilder, private readonly TypeConverterInterface $typeConverter, private readonly ResolverFactoryInterface $itemResolverFactory, private readonly ResolverFactoryInterface $collectionResolverFactory, private readonly ResolverFactoryInterface $itemMutationResolverFactory, private readonly ResolverFactoryInterface $itemSubscriptionResolverFactory, private readonly ContainerInterface $filterLocator, private readonly Pagination $pagination, private readonly ?NameConverterInterface $nameConverter, private readonly string $nestingSeparator)
```
soyuka, line 48: breaks BC layer

src/GraphQl/Type/TypeBuilder.php
```diff
@@ -210,7 +210,7 @@ public function getNodeInterface(): InterfaceType
     /**
      * {@inheritdoc}
      */
-    public function getResourcePaginatedCollectionType(GraphQLType $resourceType, string $resourceClass, Operation $operation): GraphQLType
+    public function getResourcePaginatedCollectionType(GraphQLType $resourceType, Operation $operation): GraphQLType
```
soyuka, line 213: BC layer is broken

src/GraphQl/Type/TypeBuilderInterface.php
```diff
@@ -43,7 +43,7 @@ public function getNodeInterface(): InterfaceType;
     /**
      * Gets the type of a paginated collection of the given resource type.
      */
-    public function getResourcePaginatedCollectionType(GraphQLType $resourceType, string $resourceClass, Operation $operation): GraphQLType;
+    public function getResourcePaginatedCollectionType(GraphQLType $resourceType, Operation $operation): GraphQLType;
```
soyuka, line 46: We can't change an interface like this right?

src/Metadata/Extractor/DynamicResourceExtractor.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        return $this->dynamicResources;
+    }
+
+    public function addResource(string $resourceClass, array $config = []): string
+    {
+        $dynamicResourceName = $this->getDynamicResourceName($resourceClass);
+
+        $this->dynamicResources[$dynamicResourceName] = [
+            array_merge(['class' => $resourceClass], $config),
```
soyuka, line 40: array_merge is quite slow for this we could just add the class to the config right?

src/Metadata/Extractor/DynamicResourceExtractorInterface.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata\Extractor;
+
+/**
+ * Extracts a dynamic resource (used by GraphQL for nested resources).
+ *
+ * @author Alan Poulain <contact@alanpoulain.eu>
+ */
+interface DynamicResourceExtractorInterface extends ResourceExtractorInterface
```
soyuka, line 21: do we really need to add a new interface? its wanted that a `ResourceExtractorInterface` doesn't have an `addResource` we're using a Factory to prepare resources inside it no?

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -86,6 +88,11 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
         $this->handleNotFound($operationName, $metadata);
     }
 
+    public function isDynamic(): bool
+    {
+        return str_starts_with($this->resourceClass, self::DYNAMIC_RESOURCE_CLASS_PREFIX);
+    }
```
soyuka, line 94: I really don't like this graphql-specific code here, can you explain this?

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -86,6 +88,11 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
         $this->handleNotFound($operationName, $metadata);
     }
 
+    public function isDynamic(): bool
+    {
+        return str_starts_with($this->resourceClass, self::DYNAMIC_RESOURCE_CLASS_PREFIX);
```
soyuka, line 93: imo `resourceClass` should be a `class-string` this breaks it.

src/Metadata/Resource/ResourceMetadataCollection.php
alanpoulain: This is not GraphQL-specific. The aim is to create metadata for a resource which doesn't exist.

src/Metadata/Resource/ResourceMetadataCollection.php
alanpoulain: Sure but we need a way to distinguish a dynamic resource for a normal one.

src/Metadata/Extractor/DynamicResourceExtractorInterface.php
alanpoulain: We need an interface since this method needs to be called from the type builder.

src/Metadata/Extractor/DynamicResourceExtractor.php
alanpoulain: I think it doesn't impact performance but we can.

PR #5104, opened 2022-10-28
fix(patch): Fix PATCH oneToMany

ilicmsreten:
| Q             | A
| ------------- | ---
| Tickets       | #5093

### DIFF
--- a/features/main/patch.feature
+++ b/features/main/patch.feature
@@ -58,3 +58,47 @@ Feature: Sending PATCH requets
       }
     }
     """
+
+  Scenario: Patch the oneToMany relation
+    Given there is PatchOneToManyDummy with "2" related patchOneToManyDummyRelationWithConstructor objects
+    When I add "Content-Type" header equal to "application/ld+json"
+    When I add "Content-Type" header equal to "application/merge-patch+json"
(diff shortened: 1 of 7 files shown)


ilicmsreten:
`CI / Behat (PHP 7.4 lowest)` as I can see that one isn't related to this PR

soyuka:
Patch looks nice. In my opinion this qualifies as a new feature as it was not supported in earlier versions. Wdyt ? In the case of a new feature you should target the `main` branch.

ilicmsreten:
> Patch looks nice. In my opinion this qualifies as a new feature as it was not supported in earlier versions. Wdyt ? In the case of a new feature you should target the `main` branch.

I think this is more BUGFIX, because this PATCH was working in `v2.5` and when I wanted to update `core` to `v2.7` my tests were failing. 

The main reason is that `api-platform/core` didn't have a test to cover these cases.

I didn't have time to test this new `.features`  on the branch `v2.5` , will try to do it today or tomorrow, but if these features works on `v2.5` and not working without this fix on `v2.7` then for me this one is BUGFIX

ilicmsreten:
@soyuka as I said, I tested on v2.5.4 that we use on our project and these `.features` tests are green

ilicmsreten:
@soyuka what's the state of this one

PR #5105, opened 2022-10-31
fixed registration order of routes  and added operation name

WilliamPeralta:
I have adjusted the order of the registration of the routes to avoid that itemOperations conflict with collection operation. For instance:
```
api/books/avaliable
api/books/{id}
```
instead of
```
api/books/{id}
api/books/avaliable
```

I also added the name of the operation that I use a lot in the "doctrine extension" and they are not there anymore than after the migration

| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

### DIFF
--- a/src/Core/Upgrade/UpgradeApiResourceVisitor.php
+++ b/src/Core/Upgrade/UpgradeApiResourceVisitor.php
@@ -130,8 +130,8 @@ final class UpgradeApiResourceVisitor extends NodeVisitorAbstract
 
             $arguments = [];
             $operations = null === $this->resourceAnnotation->itemOperations && null === $this->resourceAnnotation->collectionOperations ? null : array_merge(
-                $this->legacyOperationsToOperations($this->getLegacyOperations()),
-                $this->legacyOperationsToOperations($this->getLegacyOperations(true), true)
+                $this->legacyOperationsToOperations($this->getLegacyOperations(true), true),
+                $this->legacyOperationsToOperations($this->getLegacyOperations())
             );
 
             if (null !== $operations) {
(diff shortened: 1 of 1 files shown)


src/Core/Upgrade/UpgradeApiResourceVisitor.php
```diff
@@ -130,8 +130,8 @@ public function enterNode(Node $node)
 
             $arguments = [];
             $operations = null === $this->resourceAnnotation->itemOperations && null === $this->resourceAnnotation->collectionOperations ? null : array_merge(
-                $this->legacyOperationsToOperations($this->getLegacyOperations()),
-                $this->legacyOperationsToOperations($this->getLegacyOperations(true), true)
+                $this->legacyOperationsToOperations($this->getLegacyOperations(true), true),
+                $this->legacyOperationsToOperations($this->getLegacyOperations())
```
soyuka, line 134: this can't be the default, indeed in API Platform 3 we recommend to put the item GET operation as first one because this is the one that identifies the Resource in the first place.

src/Core/Upgrade/UpgradeApiResourceVisitor.php
```diff
@@ -443,6 +443,9 @@ private function legacyOperationsToOperations($legacyOperations, bool $isCollect
 
             $method = $operation['method'] ?? strtoupper($operationName);
             unset($operation['method']);
+            if(!in_array($operationName, ['get', 'post', 'put', 'patch', 'delete'])) {
+                $operation['name']= $operationName;
```
soyuka, line 447: I'll keep this patch, it's a good idea thanks!

soyuka:
Thanks I continued your work at 5109

src/Core/Upgrade/UpgradeApiResourceVisitor.php
WilliamPeralta: can you kindly tell me how I can avoid that the route of the "itemOperation" take also the customized routes without changing the order? i get a 404 error if i put itemoperation first.

src/Core/Upgrade/UpgradeApiResourceVisitor.php
WilliamPeralta: @soyuka Great, I am very pleased to have been helpful.

PR #5113, opened 2022-11-03
feat: use openapi array to validate filter parameters

norival:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | NA
| License       | MIT
| Doc PR        | NA

The filter validators were only using the 'swagger' entry from the
filter description. It should also use the 'openapi' entry, as swagger
is deprecated.


norival:
It should target v3.0

PR #5114, opened 2022-11-04
feat: use openapi array to validate filter parameters

norival:
The filter validators were only using the 'swagger' entry from the filter description. It should also use the 'openapi' entry, as swagger is deprecated.

| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | NA
| License       | MIT
| Doc PR        | NA

The filter validators were only using the 'swagger' entry from the
filter description. It should also use the 'openapi' entry, as swagger
is deprecated.

### DIFF
--- a/src/Api/QueryParameterValidator/Validator/CheckFilterDeprecationsTrait.php
+++ b/src/Api/QueryParameterValidator/Validator/CheckFilterDeprecationsTrait.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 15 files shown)


src/Api/QueryParameterValidator/Validator/CheckFilterDeprecationsTrait.php
```diff
@@ -0,0 +1,28 @@
...
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Api\QueryParameterValidator\Validator;
+
```
soyuka, line 15: add:

```
/**
 * @internal
 */
 ```

norival: Done!

PR #5174, opened 2022-11-09
fix(graphql): use default nested query operations

alanpoulain:
Instead of relying on a resource metadata factory to retrieve a "dynamic" nested query operation, this PR adds default nested query operations if the resource does not have one.

It simplifies a lot of things and catching the `OperationNotFound` exception is not needed anymore.

Since operations are "nested", they do not appear as top-level queries.

This PR also improves the XML/YAML compatibility.


tests/Fixtures/TestBundle/Resources/config/api_resources_orm/resources.xml
```diff
@@ -15,7 +15,7 @@
         </uriVariables>
 
         <operations>
-            <operation class="ApiPlatform\Metadata\GetCollection" name="custom_operation_name"/>
```
alanpoulain, line 18: It is needed since the IRI converter uses the operation name:
https://github.com/api-platform/core/blob/f1ecc30a38e50536a2a65ae85ef23eb6dc095af3/src/Symfony/Routing/IriConverter.php#L157
In the previous version of the `ExtractorResourceMetadataCollectionFactory`, the name was overridden. The `AttributesResourceMetadataCollectionFactory` does not do that.
Does it mean we cannot customize the operation name @soyuka?

soyuka: yes we can change it, it should not be overriden

src/Metadata/Resource/Factory/OperationDefaultsTrait.php
```diff
@@ -87,6 +104,38 @@ private function addDefaultGraphQlOperations(ApiResource $resource): ApiResource
...
+     */
+    private function completeGraphQlOperations(ApiResource $resource): ApiResource
+    {
+        $graphQlOperations = $resource->getGraphQlOperations();
+
+        $hasQueryOperation = false;
+        $hasQueryCollectionOperation = false;
+        foreach ($graphQlOperations as $operation) {
+            if ($operation instanceof Query && !$operation instanceof QueryCollection) {
+                $hasQueryOperation = true;
```
soyuka, line 120: continue; ?

src/Metadata/Resource/Factory/OperationDefaultsTrait.php
```diff
@@ -87,6 +104,38 @@ private function addDefaultGraphQlOperations(ApiResource $resource): ApiResource
...
+            }
+            if ($operation instanceof QueryCollection) {
+                $hasQueryCollectionOperation = true;
+            }
+        }
+
+        if (!$hasQueryOperation) {
+            $queryOperation = (new Query())->withNested(true);
```
soyuka, line 128: mb comment on Nested or here that this operation is then use when we fetch a query through another query

tests/Fixtures/TestBundle/Resources/config/api_resources_orm/resources.xml
soyuka: but usually the name of the operation is used to define the one inside the router therefore this should work as welll

soyuka:
This is way better! Minor changes but then we can merge this!

src/Metadata/Resource/Factory/OperationDefaultsTrait.php
alanpoulain: Isn't the comment on the method sufficient?

tests/Fixtures/TestBundle/Resources/config/api_resources_orm/resources.xml
alanpoulain: Fixed!

soyuka:
thanks @alanpoulain !

develth:
i currently have the issue, that i query a single item with an id with nested Items.  
As soon as i want to query the nested item, i get an  Error from `api-platform/core/src/Doctrine/Orm/State/ItemProvider.php:73` saying
> More than one result was found for query although one row or none was expected.

Example Query:

```
query parentWithElements($parent: ID!) {
    parent(id: $parent) {
        id
        name
        elements{
            edges{
                node{
                    id
                    children{
                        edges{
                            node{
                                id
                                children{
                                    edges{
                                        node{
                                            id
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

Is that related or a new issue?

PR #5176, opened 2022-11-09
fix(metadata): item uri template with another resource

soyuka:


alanpoulain:
It's hard to understand, could you add some context and a Behat test?

soyuka:
sure thing,added. This is the way we try to recommend to avoid using groups.

features/hydra/item_uri_template.feature
```diff
@@ -129,3 +129,33 @@ Feature: Exposing a collection of objects should use the specified operation to
       }
     }
     """
+
+  Scenario: Get a collection referencing another resource for it's IRI
```
alanpoulain, line 133:
```suggestion
  Scenario: Get a collection referencing another resource for its IRI
```

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -94,6 +94,10 @@ public function normalize(mixed $object, string $format = null, array $context =
             unset($context['operation'], $context['operation_name']);
         }
 
+        if (($operation = $context['operation'] ?? null) && method_exists($operation, 'getItemUriTemplate')) {
+            $context['item_uri_template'] = $operation->getItemUriTemplate();
```
alanpoulain, line 98: No unset?

features/hydra/item_uri_template.feature
```diff
@@ -129,3 +129,33 @@ Feature: Exposing a collection of objects should use the specified operation to
...
     }
     """
+
+  Scenario: Get a collection referencing another resource for its IRI
+    When I add "Content-Type" header equal to "application/json"
+    And I send a "GET" request to "/item_referenced_in_collection"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
```
vincentchalamon, line 139: The following JSON is not a schema. Did you mean `the JSON should be equal to`?

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -83,6 +83,10 @@ public function normalize(mixed $object, string $format = null, array $context =
             $resourceClass = $this->resourceClassResolver->getResourceClass($object, $context['resource_class'] ?? null);
         }
 
+        if (($operation = $context['operation'] ?? null) && method_exists($operation, 'getItemUriTemplate')) {
```
vincentchalamon, line 86: These 3 lines are copied in multiple files, can't we move them in a trait? Something like ContextOperationTrait that helps to build the context properly from an operation (I'm almose sure we have lot of other context behaviors to move on it too).

tests/Metadata/Operation/Factory/OperationMetadataFactoryTest.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $resourceNameCollectionFactory = $this->prophesize(ResourceNameCollectionFactoryInterface::class);
+        $resourceMetadataCollectionFactory = $this->prophesize(ResourceMetadataCollectionFactoryInterface::class);
+        $resourceNameCollectionFactory->create()->willReturn(new ResourceNameCollection(['one']));
+        $resourceMetadataCollectionFactory->create('one')->willReturn(new ResourceMetadataCollection('one', [
+            new ApiResource(operations: [new Get('/one', name: 'one')]),
+        ]));
+
+        $operationMetadata = new OperationMetadataFactory($resourceNameCollectionFactory->reveal(), $resourceMetadataCollectionFactory->reveal());
+        $this->assertEquals(new Get('/one', name: 'one'), $operationMetadata->create('one'));
```
vincentchalamon, line 43: Shouldn't we use `assertSame` instead of `assertEquals`, and store the `new Get` operation in a local variable, to ensure it's not a new one or a copy?

vincentchalamon (approved):
Minor changes, then :+1: 
Nice job!

src/JsonApi/Serializer/ItemNormalizer.php
soyuka: Not a huge fan of adding traits and a dependency between our normalizers.

tests/Metadata/Operation/Factory/OperationMetadataFactoryTest.php
soyuka: they should reference the same object, I'll however reduce the amount of code using a variable.

Issue #5115, opened 2022-11-04
Elasticsearch index discovery is executed when calling an orm entity, and is executed for ALL entities during container build


jannes-io:
**API Platform version(s) affected**: 3.0.2 (but I'm positive that this has always been an issue, our app just grew large enough for this to become an actual blocker during development)

**Description**  
When enabling elasticsearch, all entities do a call to a non-existing elasticsearch index first before going to orm. Creating heaps of unneeded requests.

Additionally, when building the container ALL entities do a call to elasticsearch, if you have a reasonably sized API with a couple hundred doctrine entities, and only a handful of elasticsearch entities, this causes unworkable slowdowns. Each time the container needs to be rebuild (especially annoying on APP_ENV=dev), it takes a minutes on each request.

**How to reproduce** 
Call any non-elasticsearch entity 

manual:
Install api-platform
Enable elasticsearch
Open var/log/{env}.log

repo:
clone [this repo](https://github.com/jannes-io/api-platform-elasticsearch-reproduce)
(using docker to make things simple and consistent)

```
make dev
make shell
php bin/console app:create-index
php bin/console doctrine:database:create
php bin/console doctrine:schema:create
exit
docker logs -f web
```

Now call `GET /api/doctrine_entities` You should be able to see **SIX** calls to `_cat/indices/doctrine_entity` that all fail with an `index_not_found_exception` 

Full log from just 1 request to `GET /api/doctrine_entities`:

```
[2022-11-04T12:28:00.212755+01:00] request.INFO: Matched route "_api_/doctrine_entities.{_format}_get_collection". {"route":"_api_/doctrine_entities.{_format}_get_collection","route_parameters":{"_route":"_api_/doctrine_entities.{_format}_get_collection","_controller":"api_platform.action.placeholder","_format":null,"_stateless":null,"_api_resource_class":"App\\Entity\\DoctrineEntity","_api_operation_name":"_api_/doctrine_entities.{_format}_get_collection"},"request_uri":"http://a.docker/api/doctrine_entities?page=1","method":"GET"} []
[2022-11-04T12:28:00.237725+01:00] app.DEBUG: Request Body [null] []
[...]

PR #5177, opened 2022-11-09
fix(metadata): check if elasticsearch is set to false by user through ApiResource (#5115)

jannes-io:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5115 
| License       | MIT
| Doc PR        | 

If `elasticsearch` is set to `false` by the user forcefully on the `ApiResource` attribute, then don't bother fetching the index anyway. This just causes needless slowdowns during container build.

I would've liked to revert the logic, in other words, only do the `hasIndices` check if `elasticsearch: true`, but that would introduce a BC break, so after this change, projects using a mix of orm & elasticsearch entities can simply mark their orm entities with `elasticsearch: false` and their elasticsearch entities with `elasticsearch: true`, and the index discovery will only be done on those with `elasticsearch: true`.

Not a big deal if your application is small, but once you start getting into triple digits of entities, doing a request for every single entity causes major slowdowns when `APP_ENV` is not set to `prod` (when the operations aren't cached).


soyuka:
nice, you can even target 3.0 as its a bug fix. We're working on improvements for Elasticsearch on 3.1.

kdckrs:
> nice, you can even target the 3.0 branch as its a bug fix.
> 
> side note: we're working on improvements for Elasticsearch on 3.1!

Awesome! Would be epic to get this PR merged as it boosts performance instantly!

jannes-io:
> nice, you can even target the 3.0 branch as its a bug fix.
> 
> side note: we're working on improvements for Elasticsearch on 3.1!

Will do.

jannes-io:
@soyuka moved to 3.0

soyuka:
thanks !

kdckrs:
Epic guys, thanks for the flawless improvements and speed of merge!

Issue #5054, opened 2022-10-05
Impossible to use the security property in ApiProperty attribute


Xusifob:
**API Platform version(s) affected**: 3.0.0

**Description**  
When using the security property on the ApiProperty attribute on an entity property, the _object_ object received, is null

**How to reproduce**  
````php
// src/Entity/Dummy.php 

    #[ApiProperty(
        security: "is_granted('VIEW',object)",
        securityPostDenormalize: "is_granted('MODERATE',object)",
        schema: [
            'type' => 'boolean',
        ])]
    public bool $isModerated = false;
 ````
 
 In my scenario, VIEW should returns true and MODERATE should returns false

**Possible Solution**  
The security attribute should get an access to the object that is created

**Additional Context**  

Here is the data my voters are getting
![image](https://user-images.githubusercontent.com/8103268/194146264-b49c6817-f90b-49ae-b9b5-97c3700a443f.png)

soyuka:
We need a reproducer or more informations about what you're trying to do.

PR #5182, opened 2022-11-12
fix: securityPostDenormalize not working because clone is made after denormalization

LoicBoursin:
| Q             | A
| ------------- | ---
| Branch?      | 3.0
| Tickets       | #5054 
| License       | MIT

Support for `security_post_denormalize` was added in https://github.com/api-platform/core/pull/4184.

At that moment, the clone to populate previousObject was made before denormalization. It worked well.

This [commit](https://github.com/api-platform/core/commit/87b61505d2f719fcf0d8613414d9e6aa436fd61f#diff-d87efa25fbf52904daa06047002652f08645e3ad4a7f4985ffeec5a7c8e39c4f) moved the clone after denormalization.

After upgrading to 3.0, using `security_post_denormalize` on a an `ApiProperty` was not working anymore. Moving the `$previousObject` line to its first position fixed the issue on my project.

I believe [this issue](https://github.com/api-platform/core/issues/5054) will be fixed too.

### DIFF
--- a/features/authorization/deny.feature
+++ b/features/authorization/deny.feature
@@ -87,6 +87,33 @@ Feature: Authorization checking
     And the JSON node "ownerOnlyProperty" should exist
     And the JSON node "ownerOnlyProperty" should not be null
 
+  Scenario: An admin can create a secured resource with properties depending on themselves
+    When I add "Accept" header equal to "application/ld+json"
+    And I add "Content-Type" header equal to "application/ld+json"
+    And I add "Authorization" header equal to "Basic YWRtaW46a2l0dGVu"
+    And I send a "POST" request to "/secured_dummy_with_properties_depending_on_themselves" with body:
+    """
+    {
+        "canUpdateProperty": false,
+        "property": false
+    }
+    """
+    Then the response status code should be 201
+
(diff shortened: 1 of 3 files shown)


soyuka:
Would it be possible to have a non regression test?

soyuka:
nice catch thanks @LoicBoursin !

FPDK:
Thanks for the fix. For anyone still having issues, check your _allow_if_all_abstain_ and _strategy_ [security setting](https://symfony.com/doc/current/security/voters.html#changing-the-access-decision-strategy):

`
security:
   access_decision_manager:
      strategy: priority
      allow_if_all_abstain: true
`

If I have _allow_if_all_abstain_ set to _false_ in my project, one of my voters will produce a false positive of sorts in the Symfony profiler. This is even when the voter's _supports()_ returns _false_, or _voteOnAttribute()_ returns _true_.

soyuka:
@FPDK thanks for that, do you think it could be added to the api platform documentation?

FPDK:
@soyuka If I can find the time, I'll try to test it in a blank project first before I create a PR on the docs :)

PR #5187, opened 2022-11-15
 fix(metadata): _format broken bc promise on #5080 

soyuka:
See https://github.com/api-platform/core/pull/5080 for details

### DIFF
--- a/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
@@ -97,6 +97,7 @@ final class UriTemplateResourceMetadataCollectionFactory implements ResourceMeta
     {
         $uriTemplate = $operation->getUriTemplate() ?? sprintf('/%s', $this->pathSegmentNameGenerator->getSegmentName($operation->getShortName()));
         $uriVariables = $operation->getUriVariables() ?? [];
+        $legacyFormat = null;
 
         if (str_ends_with($uriTemplate, '{._format}') || ($legacyFormat = str_ends_with($uriTemplate, '.{_format}'))) {
             $uriTemplate = substr($uriTemplate, 0, -10);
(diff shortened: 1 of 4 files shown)


src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
```diff
@@ -96,6 +98,15 @@ private function generateUriTemplate(HttpOperation $operation): string
         $uriTemplate = $operation->getUriTemplate() ?? sprintf('/%s', $this->pathSegmentNameGenerator->getSegmentName($operation->getShortName()));
         $uriVariables = $operation->getUriVariables() ?? [];
 
+        if (str_ends_with($uriTemplate, '{._format}') || ($legacyFormat = str_ends_with($uriTemplate, '.{_format}'))) {
+            $uriTemplate = substr($uriTemplate, 0, -10);
+        }
+
+        if ($legacyFormat && ($this->triggerLegacyFormatOnce[$operation->getClass()] ?? true)) {
+            $this->triggerLegacyFormatOnce[$operation->getClass()] = false;
+            trigger_deprecation('api-platform/core', '3.0', sprintf('The special Symfony parameter ".{_format}" in your URI Template is deprecated, use an RFC6570 variable "{.format}" on the class "%s" instead. We will only use the RFC6570 compatible variable in 4.0.', $operation->getClass()));
```
alanpoulain, line 107: Are we using `trigger_deprecation` now? We need to make sure it's not a dev dep then.

alanpoulain, line 107:
```suggestion
            trigger_deprecation('api-platform/core', '3.0', sprintf('The special Symfony parameter ".{_format}" in your URI Template is deprecated, use an RFC6570 variable "{._format}" on the class "%s" instead. We will only use the RFC6570 compatible variable in 4.0.', $operation->getClass()));
```

soyuka: its a dependency of http-foundation

alanpoulain: Sure but it should be a direct dependency now.

Issue #5179, opened 2022-11-10
2.7.3 ignores the `route_prefix` prefix settings


DarkChyper:
Hello,
I found this message in a closed issue named "Issues found on 2.7" because i reproduce this issue with 2.7.3 (SF 4.4.45 PHP 7.4)

`route_prefix` is still ignored when i try to migrate from 2.6 to 2.7 (cannot go to 3.x now).

I don't find any fix for this (but i'm not good to read PR, i could miss it)

edit: it works if i add setting on each resource : 
config/api/resources/myresource.yaml
```
resources:
  My\Resource:
    routePrefix: 'v2'
```
        @soyuka, it seems the latest version ignores the `route_prefix` prefix settings

config/packages/api_platform.yaml

```
api_platform:
...
    metadata_backward_compatibility_layer: false
    openapi:
        backward_compatibility_layer: false
    defaults:
        route_prefix: '/api'
```

`GET /users/{id} Retrieves a User resource.`
Instead of
`GET /api/users/{id} Retrieves a User resource.`

Tested with d876e6638acad4b82887a4f93fd4eca361584d00

_Originally posted by @lermontex in https://github.com/api-platform/core/issues/4613#issuecomment-1126735174_

DarkChyper:
Thanks for the answer ! 

I'll try to provide a reproducer, maybe the issue come from something with symfony 4.4.45 or yaml serializer (we don't use annotation on entities)

vincentchalamon:
Ah ok, it changes some things indeed. I can now reproduce the bug using the following configuration:

```yaml
resources:
    App\Entity\Greeting:
        routePrefix: v2
    App\Entity\Book: ~
```

I only have the bug on API Platform 2.7, not on 3.0

DarkChyper:
ok so it seems that there is no bug with symfony 5.x and api platform 2.7. 
Maybe i have to wait to upgrade the project to 5.x before update api platform on my project ( i have another issue with circular reference on non api resource of sub resource.. )

vincentchalamon:
Should be fixed in https://github.com/api-platform/core/pull/5194. Thanks for reporting this issue

FPDK:
@vincentchalamon Is it intentional that the ApiResource-specific routePrefix will be ignored if the ApiResource only contains graphQlOperations? 

Example:

<pre>
#[ApiResource(
    operations: [],
    routePrefix: '/resourceSpecificPrefix',
    graphQlOperations: [
[...]

PR #5194, opened 2022-11-16
fix(metadata): operations must inherit from resource and defaults

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #5179
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
@@ -26,6 +26,9 @@ use ApiPlatform\Metadata\Post;
 use ApiPlatform\Metadata\Put;
 use ApiPlatform\Metadata\Resource\DeprecationMetadataTrait;
 use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
 
 /**
  * Creates a resource metadata from {@see Resource} extractors (XML, YAML).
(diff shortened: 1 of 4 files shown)


src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
```diff
@@ -205,6 +207,59 @@ private function getOperationWithDefaults(ApiResource $resource, HttpOperation $
...
+    {
+        $extraProperties = [];
+        foreach ($this->defaults as $key => $value) {
+            $upperKey = ucfirst($this->camelCaseToSnakeCaseNameConverter->denormalize($key));
+            $getter = 'get'.$upperKey;
+
+            if (!method_exists($operation, $getter)) {
+                if (!isset($extraProperties[$key])) {
+                    $extraProperties[$key] = $value;
+                }
```
soyuka, line 247: continue;

to avoid the `else`

src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
```diff
@@ -26,6 +26,9 @@
 use ApiPlatform\Metadata\Put;
 use ApiPlatform\Metadata\Resource\DeprecationMetadataTrait;
 use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
```
dunglas, line 31: Can we find a way to remove this dependency? Or at least create an interface to abstract it?

vincentchalamon: It should be done in a separate PR as it's used at least 7 times in traits and classes. It's hardcoded (`new CamelCaseToSnakeCaseNameConverter()`), using an interface would require a dependency injection behavior.

soyuka: I don't think there's a point doing an abstraction for this.

soyuka (approved):

PR #5201, opened 2022-11-18
Don't try to generate iri, if it is already defined in $context

gassan:
| Q             | A
| ------------- | ---
| Branch?       | 2.x
| Tickets       | #... 
| License       | MIT

After delete-item operation I would like to output deleted data, but the normalizer fails on generating `iri` with `PK=null`.
Previously the iri was not generated when passed with in `$context`.


soyuka:
can you commit:
```
fix(serializer): use iri from $context if defined
```

gassan:
done, thanks.

Issue #5203, opened 2022-11-20
Wrong route match if routePrefix is defined


usu:
**API Platform version(s) affected**: 3.0.4

**Description**  
As of now, `routePrefix` is not included in operation name (which also serves as the symfony route name). This can lead to route name conflict and hence to wrong route matches, if multiple resources are defined on the the same resource class.

**How to reproduce**  
1. Specify a resource class with 2 ApiResources, one of them with a `routePrefix`:

```php
#[ApiResource(
    operations: [
        new Get(
            security: 'true'
        ),
        new GetCollection(
            security: 'true'
        ),
)]
#[ApiResource(
    operations: [
        new Get(
            security: 'false'
        ),
        new GetCollection(
            security: 'false'
        ),
    ],
    routePrefix: '/admin'
)]
class User {
}
```

2. Load '/admin/users'

_Expected result:_ 403 Forbidden (matching GetCollection operation of the second ApiResource)

_Received result:_ 200 (matching GetCollection operation of first ApiResource). Confusingly, the individual IRIs are still prefixed with /admin
```json
{
  "_links": {
    "self": {
      "href": "/admin/users"
    },
    "items": [
      {
        "href": "/admin/users/1234"
      },
      {
        "href": "/admin/users/5678"
      }
    ]
  },
  "totalItems": 2
}
```

**Possible Solution**  
Include `routePrefix` in operation name

**Additional context**
- In SwaggerUI, only the /admin/users endpoint are shown (see screenshot below)
- The routes of the first ApiResource seems to overwritten in Symfony router. Loading `/users` leads to a "404 No route found exception"

SwaggerUI:
![Screenshot 2022-11-20 073837](https://user-images.githubusercontent.com/449555/202889721-86cecf5e-7220-47d5-9148-876c9d7a16dd.png)

Route Match Log:
![Screenshot 2022-11-20 074206](https://user-images.githubusercontent.com/449555/202889727-945179e2-54b0-4df7-b20f-8ca21db577a8.png)

usu:
Hi @soyuka. Could this issue be reopened? Not sure if you saw my comment directly in the PR https://github.com/api-platform/core/pull/5208#issuecomment-1328007551, but I think this is not yet resolved.

divine:
@soyuka sorry to bump, this can be closed as PR has been merged.

Thanks!

PR #5208, opened 2022-11-21
fix(metadata): route prefix in the operation name

soyuka:
fixes https://github.com/api-platform/core/issues/5203

### DIFF
--- a/src/Metadata/Operations.php
+++ b/src/Metadata/Operations.php
@@ -26,6 +26,11 @@ final class Operations implements \IteratorAggregate, \Countable
             // When we use an int-indexed array in the constructor, compute priorities
             if (\is_int($operationName)) {
                 $operation = $operation->withPriority($operationName);
+                $operationName = (string) $operationName;
+            }
+
+            if ($operation->getName()) {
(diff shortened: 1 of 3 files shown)


usu:
Hi @soyuka. Thanks for looking into this. I tried to test this with my application, but didn't work:

- Factories are decorated as per the following chain: `OperationNameResourceMetadataCollectionFactory` --> `LinkResourceMetadataCollectionFactory` --> `UriTemplateResourceMetadataCollectionFactory`

- In `UriTemplateResourceMetadataCollectionFactory` on [line 83](https://github.com/api-platform/core/blob/3.0/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php#L83), the operation name is already set (without `routePrefix`)

- In `OperationNameResourceMetadataCollectionFactory` on [line 47](https://github.com/api-platform/core/pull/5208/files#diff-0f46ec91f44bb62e69df79a654abb292176829f09ec91b271fcaf61373682182R47), the loop is aborted, because the name is already set. Hence, the new logic to include `routePrefix` is never executed.

PR #5214, opened 2022-11-22
feat: use phpdoc-parser instead of phpdocumentor

alanpoulain:
PHPStan PHPDoc-Parser is used instead of phpDocumentor because the latter is not actively maintained anymore.


src/Metadata/Extractor/ClassDescriptionExtractorInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata\Extractor;
+
+/**
+ * Guesses the class's human readable description.
+ */
+interface ClassDescriptionExtractorInterface
```
soyuka, line 19: do we really need to introduce a new interface for this?

src/Metadata/Extractor/PhpDocClassExtractor.php
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getLongDescription(string $class, array $context = []): ?string
```
soyuka, line 62: We never need this right?

src/Metadata/Extractor/PhpDocClassExtractor.php
```diff
@@ -0,0 +1,105 @@
...
+            return null;
+        }
+
+        $tokens = new TokenIterator($this->lexer->tokenize($rawDocNode));
+        $phpDocNode = $this->phpDocParser->parse($tokens);
+        $tokens->consumeTokenType(Lexer::TOKEN_END);
+
+        return $this->docBlocks[$class] = $phpDocNode;
+    }
+}
```
soyuka, line 105: I do think that we can avoid this abstraction and use the phpDocParser directly in the PhpDocResourceMetadataCollectionFactory. Indeed the way I see it the metadata factories are the lowest abstraction in our system, having another abstraction adds too much code that we need to then maintain.

src/Metadata/Extractor/ClassDescriptionExtractorInterface.php
alanpoulain: I think it's better, like Symfony does: https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/PropertyInfo/PropertyDescriptionExtractorInterface.php.

src/Metadata/Extractor/PhpDocClassExtractor.php
alanpoulain: I think it can be useful in OpenAPI and GraphQL.

src/Metadata/Extractor/PhpDocClassExtractor.php
alanpoulain: I don't agree. Having this abstraction allows to reuse it elsewhere (see my other PR for GraphQL enum) and it allows to change easily the underlying implementation. Users can even write their own implementation if they have custom needs.

soyuka: my question is do we want to maintain such abstraction? Inside Symfony I'd agree but not inside API Platform.

src/Metadata/Extractor/PhpDocClassExtractor.php
soyuka: let's add this when we need it then we don't want to maintain dead code

src/Metadata/Extractor/ClassDescriptionExtractorInterface.php
soyuka: probably that we should put this code inside Symfony then ?

src/Metadata/Extractor/PhpDocClassExtractor.php
alanpoulain: What is the cost of "maintaining" this class? Having this abstraction is better for a lot of reasons, I don't understand why you are so against it.

src/Metadata/Extractor/ClassDescriptionExtractorInterface.php
alanpoulain: Symfony has no component for it. `PropertyInfo` only cares about properties.

src/Metadata/Extractor/PhpDocClassExtractor.php
alanpoulain: It's mostly to have the same interface as the Symfony one, but alright I will throw instead.

src/Metadata/Extractor/PhpDocClassExtractor.php
soyuka: Let me try to explain my vision on this:

ResourceMetadataFactory is and should be the single source of truth inside of API Platform when it comes to defining resources and their content. Any other abstraction should be `@internal` and not available for the end users to use. It brought problems in the past, and we're still dealing with some of these (thinking about the OperationPathResolver, or the whole filter validation feature and probably others). I'm really trying to simplify the approach for the developers and introducing a class, a new interface, an even here a new concept should really be challenged in every ways. 

Another way of seeing this is that if I'm a developper that wants to extract documentation from a class I want the developper to:
- use phpstan/doc-parser directly
- use an abstraction provided by a library 

It's not, in my opinion, API Platform's responsibility to do this and to provide such an abstraction. I might be wrong though and we probably need more opinions on this subject. This is also why we didn't create an extractor for the initial php doc implementation.

dunglas: I agree with @soyuka, let's keep this internal, it's not the responsibility of the core library to parse PHPdocs and we already provide the relevant extension points: users can just write their own metadata factories.

PR #5216, opened 2022-11-23
fix(metadata): warn on wrong uri variables configuration

### LINKED ISSUE #5184: Problem with custom variables in query / uriTemplate
**API Platform version(s) affected**: 3.0.3

**Description**  
If you set custom `uriTemplate` for an operation with non-identifiers as variables you will get 404 or 500.

**How to reproduce**  
Add operation with template similar to:
```
uriTemplate: '/photos/{uuid}/resize/{width}/{height}',
```
where `{uuid}` is an UUID identifier.

As I can see additional variables are "resolved" as identifiers for the relations in the data provider resulting in query broken or yielding no results at all.

**Possible Solution**  
I have made something I called NoExtraLinkProvider which is basically `ApiPlatform\Doctrine\Orm\State\ItemProvider` with additional code:

```php
if ($operation instanceof HttpOperation && array_key_exists('uuid', $uriVariables)) {
    $uriVariables = ['uuid' => $uriVariables['uuid']];
    $operation = $operation->withUriVariables(['uuid' => $operation->getUriVariables()['uuid']]);
}
```

which removes any uriVariables added automatically by the framework. With help of that I can get the Photo resource by `{uuid}` and pass additional `{with}` and `{height}` to my controller.

But this does not help with collection operations (without `{uuid}`, different resource but similar problem, like `/resources/search/{search}`) and I don't know what to do here.

**Additional Context**  
I did not find how to do this in v3, in v2 it worked without any problems. Maybe I'm doing something wrong (most probably) or maybe there is something missing in this beautiful framework. Please help or point me to the proper place to find help. Thank you.

soyuka:
fixes #5184

### DIFF
--- a/features/main/attribute_resource.feature
+++ b/features/main/attribute_resource.feature
@@ -1,11 +1,11 @@
+@php8
+@!mysql
+@!mongodb
 Feature: Resource attributes
   In order to use the Resource attribute
   As a developer
   I should be able to fetch data from a state provider
 
-  @php8
-  @!mysql
-  @!mongodb
   Scenario: Retrieve a Resource collection
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "GET" request to "/attribute_resources"
@@ -35,9 +35,6 @@ Feature: Resource attributes
     }
     """
 
-  @php8
-  @!mysql
-  @!mongodb
   Scenario: Retrieve the first resource
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "GET" request to "/attribute_resources/1"
@@ -55,9 +52,6 @@ Feature: Resource attributes
     }
     """
 
-  @php8
-  @!mysql
-  @!mongodb
   Scenario: Retrieve the aliased resource
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "GET" request to "/dummy/1/attribute_resources/2"
@@ -77,9 +71,6 @@ Feature: Resource attributes
     }
     """
 
-  @php8
-  @!mysql
-  @!mongodb
   Scenario: Patch the aliased resource
     When I add "Content-Type" header equal to "application/merge-patch+json"
     And I send a "PATCH" request to "/dummy/1/attribute_resources/2" with body:
@@ -101,3 +92,10 @@ Feature: Resource attributes
       "name": "Patched"
     }
     """
+
+  Scenario: Uri variables should be configured properly
+    When I send a "GET" request to "/photos/1/resize/300/100"
+    Then the response status code should be 400
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON node "hydra:description" should be equal to 'Unable to generate an IRI for the item of type "ApiPlatform\Tests\Fixtures\TestBundle\Entity\IncompleteUriVariableConfigured"'
--- a/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
@@ -137,7 +137,9 @@ final class UriTemplateResourceMetadataCollectionFactory implements ResourceMeta
             return $this->normalizeUriVariables($operation);
         }
 
+        $hasUserConfiguredUriVariables = true;
         if (!$operation->getUriVariables()) {
+            $hasUserConfiguredUriVariables = false;
             $operation = $operation->withUriVariables($this->transformLinksToUriVariables($this->linkFactory->createLinksFromIdentifiers($operation)));
         }
 
@@ -164,6 +166,10 @@ final class UriTemplateResourceMetadataCollectionFactory implements ResourceMeta
         $variables = $route->getPathVariables();
 
         if (\count($variables) !== \count($uriVariables)) {
+            if ($hasUserConfiguredUriVariables) {
+                return $operation;
+            }
+
             $newUriVariables = [];
             foreach ($variables as $variable) {
                 if (isset($uriVariables[$variable])) {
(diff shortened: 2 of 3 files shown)


features/main/attribute_resource.feature
```diff
@@ -101,3 +92,10 @@ Feature: Resource attributes
       "name": "Patched"
     }
     """
+
+  Scenario: Uri variables should be configured properly
+    And I send a "GET" request to "/photos/1/resize/300/100"
```
alanpoulain, line 97:
```suggestion
    When I send a "GET" request to "/photos/1/resize/300/100"
```

src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
```diff
@@ -29,7 +31,7 @@ final class UriTemplateResourceMetadataCollectionFactory implements ResourceMeta
 {
     private $triggerLegacyFormatOnce = [];
 
-    public function __construct(private readonly LinkFactoryInterface $linkFactory, private readonly PathSegmentNameGeneratorInterface $pathSegmentNameGenerator, private readonly ?ResourceMetadataCollectionFactoryInterface $decorated = null)
+    public function __construct(private readonly LinkFactoryInterface $linkFactory, private readonly PathSegmentNameGeneratorInterface $pathSegmentNameGenerator, private readonly ?ResourceMetadataCollectionFactoryInterface $decorated = null, private readonly ?LoggerInterface $logger = new NullLogger())
```
alanpoulain, line 34: Shouldn't we inject the logger with a setter instead (see `LoggerAwareTrait`)? I think having a direct dependency is too much for a logger.

src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
```diff
@@ -164,6 +168,12 @@ private function configureUriVariables(ApiResource|HttpOperation $operation): Ap
         $variables = $route->getPathVariables();
 
         if (\count($variables) !== \count($uriVariables)) {
+            if ($hasUserConfiguredUriVariables) {
+                $this->logger?->warning('You have configured less URI variables then present in your URI template. You will need to decorate the IRI converter to avoid IRI generation errors.');
```
alanpoulain, line 172:
```suggestion
                $this->logger?->warning('You have configured less URI variables than present in your URI template. You will need to decorate the IRI converter to avoid IRI generation errors.');
```

alanpoulain: I think the last sentence is not clear for most users.

tests/Fixtures/TestBundle/Entity/UriVariableConfigured.php
```diff
@@ -0,0 +1,34 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Operation;
+
+#[ApiResource(uriTemplate: '/photos/{id}/resize/{width}/{height}', uriVariables: ['id'], provider: [UriVariableConfigured::class, 'provide'])]
+final class UriVariableConfigured
```
alanpoulain, line 20:
```suggestion
class IncompleteUriVariableConfigured
```

tests/Fixtures/TestBundle/Entity/UriVariableConfigured.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function __construct(public string $id)
+    {
+    }
+
+    public static function provide(Operation $operation, array $uriVariables = [], array $context = [])
```
alanpoulain, line 26:
```suggestion
    public static function provide(Operation $operation, array $uriVariables = [], array $context = []): UriVariableConfigured
```

src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
soyuka: we do like this everywhere and its totally optional I don't see the issue

src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
soyuka: I'm wondering what's the best solution here, should I just throw? These use cases are really borderline and should probably just use Symfony. It's definitely wrong to add uriVariables if some are already configured so this needs a fix but whether to throw or to warn idk :|.

src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
alanpoulain: IMO the logger has no value as a direct dependency. It is totally optional and the class can be used without it without losing any benefit. I personally see it as a best practice.

src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php
alanpoulain: Maybe just add a larger explanation? For instance: "this is probably an error in your configuration, but if you need to use URI variables for other uses than as identifiers...".

soyuka: then I should throw, even though I like this to be possible. Note that this will throw later in our IriConverter right now. Maybe that actually the error inside the IriConverter is clear enough?

soyuka:
Let's fix the uri variable thing the IRI converter will be enough user land.

soyuka:
fixed on 2.7 instead #5217

PR #5222, opened 2022-11-23
fix query parameters normalization in createUri

### LINKED ISSUE #5221: Query parameters are incorrectly normalized in IriHelper
**API Platform version(s) affected**: 3.0.4

**Description**  

If you pass array of array in query, the next page got from IriHelper is not correct.

**How to reproduce**  

Call your API endpoint with multi-dimensional array params:

```
GET https://my-api/objects?filters[0][a]=foo&filters[0][b]=bar
```

In PHP we can access `filters[0]['a']` and `filters[0]['b']`

But the next URL from hydra response is something like:

```
https://my-api/objects?page=2&filters[][a]=foo&filters[][b]=bar
```

So in PHP the `filters` structure is altered and `filters[0]['b']` is not defined.

**Possible Solution**  

Remove `preg_replace('/%5B\d+%5D/', '%5B%5D', $query)` from [IriHelper::createIri](https://github.com/api-platform/core/blob/main/src/Util/IriHelper.php#L65).

4rthem:
fixes #5221

### DIFF
--- a/src/Util/IriHelper.php
+++ b/src/Util/IriHelper.php
@@ -61,8 +61,7 @@ final class IriHelper
             $parameters[$pageParameterName] = $page;
         }
 
-        $query = http_build_query($parameters, '', '&', \PHP_QUERY_RFC3986);
-        $parts['query'] = preg_replace('/%5B\d+%5D/', '%5B%5D', $query);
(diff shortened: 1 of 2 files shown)


soyuka:
looks like there are some failures related to this though

Issue #5009, opened 2022-09-24
The elasticsearch/elasticsearch package is required for Elasticsearch support after installation


lkaopremier:
**API Platform version(s) affected**: 3.0

**Description**  
After adding the elasticsearch/elasticsearch package, I got this error "The elasticsearch/elasticsearch package is required for Elasticsearch support.", When analyzing the source code of the version of elasticsearch I added I realized that the namespace declared in the elasticsearch client is different from the namespace used by Api Platform.

**How to reproduce**  
Using version 3.0 of the Api platform and adding the latest version of the elasticsearch/elasticsearch package finally enabled elasticsearch in the configs, [See the official doc](https://api-platform.com/docs/v2.5/core/elasticsearch/#elasticsearch-support).

**Additional Context**  
![Capture d’écran 2022-09-24 à 11 25 23](https://user-images.githubusercontent.com/36473778/192095441-0c5a3355-b8d4-4565-bb72-ed4d11dad913.png)
![Capture d’écran 2022-09-24 à 11 30 33](https://user-images.githubusercontent.com/36473778/192096256-70c56122-a1e5-4f5a-a70b-5a7edd3b2ebf.png)
![Capture d’écran 2022-09-24 à 11 43 01](https://user-images.githubusercontent.com/36473778/192096259-451cafb3-b740-41c3-8104-ae9e98c4eee6.png)

darthf1:
Hi @alanpoulain,

Would you be able to give some pointers, on how to properly support both ES 7 and ES 8? I was planning to migrate to ES 8 in a few weeks so depending on the work I could give it a go at that time.

Current usages of the `Elasticsearch` namespace (in 2.6.8):
![image](https://user-images.githubusercontent.com/17253332/192725562-aed2aa9f-ff0d-4707-8cf6-5b0d433251d6.png)

alanpoulain:
You probably need to use a facade to abstract the client used. You have an example here: https://github.com/api-platform/core/blob/2.7/src/Util/Inflector.php

curry684:
I just upgraded `api-platform/core` to `^3.0` in my composer.json, and accidentally happened to notice that afterwards I got stuck on 3.0.0 without upgrade possibility to 3.0.2. Further investigation got me to this issue, which is all fine and dandy, except that I'm now on 3.0.0, with Elasticsearch 8 deeply embedded in my project. So I should remain stuck on 2.7 for the time being, not 3.0.0, the last version without this conflict.

[...]

PR #5223, opened 2022-11-23
Remove conflict with Elasticsearch used only on dev env

lruozzi9:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | 
| License       | MIT
| Doc PR        | 

The elasticsearch package is required only for the dev environment, but a conflict with versions greater than v8 has been added, so it will block any installation of these versions of the package in other projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "doctrine/mongodb-odm": "^2.2",
         "doctrine/mongodb-odm-bundle": "^4.0",
         "doctrine/orm": "^2.6.4",
-        "elasticsearch/elasticsearch": "^7.11.0",
+        "elasticsearch/elasticsearch": "^7.11.0 <8.0",
         "friends-of-behat/mink-browserkit-driver": "^1.3.1",
         "friends-of-behat/mink-extension": "^2.2",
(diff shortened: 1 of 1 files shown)


soyuka:
conflicts apply to your deps ? can you just remove the conflict and not change our dev dependency ?

lruozzi9:
I think that this repo in dev env with `elasticsearch/elasticsearch` >= 8.0 fails (#5009) so the incompatibility with these versions should remain. But this is a dev dependency, so I can not add `elasticsearch/elasticsearch` > 8.0 in another repo along with `apiplatform/core` ^2.7 due to this conflict.

dunglas:
It's not only a dev dependency, it's also an optional (prod) dependency. If we remove the conflict, like that, our ElasticSearch support will be broken.

The proper fix would be to patch API Platform to support the latest version of this dependency.

lruozzi9:
> It's not only a dev dependency, it's also an optional (prod) dependency.

Ah ok.

> The proper fix would be to patch API Platform to support the latest version of this dependency.

Obviously yes. If possible I will have a look and provide a PR. In the meantime, this is no more needed.

PR #5240, opened 2022-11-29
feat(symfony): wire Symfony JsonEncoder if it exists

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | ~
| License       | MIT
| Doc PR        | ~

While trying to decorate the `Symfony\Component\Serializer\Encoder\JsonEncoder` (service `serializer.json.encoder`) we found out that it is not used by `ApiPlatform\Core\Serializer\JsonEncoder` as a service since the argument is not defined and one is created instead https://github.com/api-platform/core/blob/main/src/Serializer/JsonEncoder.php#L36.
So we currently need to wire the argument ourselves in a compiler pass.

Should I add a test for this?

### DIFF
--- a/src/Symfony/Bundle/Resources/config/jsonld.xml
+++ b/src/Symfony/Bundle/Resources/config/jsonld.xml
@@ -45,6 +45,7 @@
 
         <service id="api_platform.jsonld.encoder" class="ApiPlatform\Serializer\JsonEncoder" public="false">
             <argument>jsonld</argument>
+            <argument type="service" id="serializer.json.encoder" on-invalid="null" />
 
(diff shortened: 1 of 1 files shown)


HeahDude:
>why don't you decorate api_platform.jsonld.encoder? I agree about the patch though can you target 3.0 please?

`api_platform.jsonld.encoder` is not the one we need to decorate as we may use the `json` format one as well, so it should just reuse the json decode logic by injection.

I thought it would be considered as a feature, glad if it can be part of the next patch, even though we are not using 3.0 yet :).

soyuka:
thanks!

PR #5251, opened 2022-12-04
fix(graphql): remove inline styles and add twig blocks to aid overriding templates in strict CSP environments

moismailzai:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

I've added template blocks in the SwaggerUi, Graphiql, and GraphQlPlayground twig templates. I've also removed some inline styles and JavaScript and replaced them with corresponding CSS classes and functions. In one instance, I've moved inline styles to a separate CSS file.

The motivation for these changes is better compatibility with strict Content Security Policy environments. In such environments, it is common to avoid inline scripts and to add a nonce to all script and style tags (https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#nonces). By adding the new template blocks, it becomes possible to extend default templates and override the blocks with scripts and styles.


src/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig
```diff
@@ -73,14 +84,17 @@
...
+            {% if not graphQlEnabled %}<a href="#" onclick="noGraphQlSupport()">GraphiQL</a>{% endif %}
             {% if graphiQlEnabled %}<a href="{{ path('api_graphql_graphiql') }}">GraphiQL</a>{% endif %}
             {% if graphQlPlaygroundEnabled %}<a href="{{ path('api_graphql_graphql_playground') }}">GraphQL Playground</a>{% endif %}
         </div>
     </div>
 </div>
 
 {% block javascript %}
+    <script type="text/javascript">
+        let noGraphQlSupport = () => alert('GraphQL support is not enabled, see https://api-platform.com/docs/core/graphql/');
```
dunglas, line 96: Shouldn't this be put in an external file too?

dunglas (approved):

src/Symfony/Bundle/Resources/views/SwaggerUi/index.html.twig
moismailzai: It absolutely could be, I was just trying to limit the scope of changes as much as I could. In this instance, if I override the `javascript` block, and am using, say the https://github.com/nelmio/NelmioSecurityBundle to handle generating nonces, I can still add a nonce to it. The way it was in-lined before would have made that impossible. That being said, if we move it to it's own file we don't have to allow inline JS period, which is nicer.

moismailzai: added a commit with the click event moved into a handler in the `init-swagger-us.js` file

src/Symfony/Bundle/Resources/public/graphql-playground/index_body.css
```diff
@@ -0,0 +1,468 @@
+html {
```
soyuka, line 1: is this a vendor file ? Could you update the `update-js.sh` script accordingly?

moismailzai: good catch, I didn't realize there was a script that deleted everything in this directory and recopied from node_modules. I moved the CSS file up one level so it's unaffected by the `update-js.sh` script. This seems like the correct approach since it's not actually a vendor file.

moismailzai:
> Can you rename your first commit to match our commit lint rules? also is this considered as a bug fix?

hiya, I renamed the PR to prefix it with `fix(graphql):`. Is that sufficient or should I resubmit a new pr with a squashed commit and commit message?

soyuka:
all good it should do it as I squash form github :) thanks!

PR #5252, opened 2022-12-04
fix(metadata): include routePrefix in default operation name (#5203)

### LINKED ISSUE #5203: Wrong route match if routePrefix is defined

usu:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5203
| License       | MIT

Fixes #5203

### DIFF
--- a/src/Metadata/Resource/Factory/OperationDefaultsTrait.php
+++ b/src/Metadata/Resource/Factory/OperationDefaultsTrait.php
@@ -187,16 +187,28 @@ trait OperationDefaultsTrait
             $operation = $operation->withName($operation->getRouteName());
         }
 
-        $operationName = $operation->getName() ?? sprintf(
-            '_api_%s_%s%s',
-            $operation->getUriTemplate() ?: $operation->getShortName(),
-            strtolower($operation->getMethod() ?? HttpOperation::METHOD_GET),
-            $operation instanceof CollectionOperationInterface ? '_collection' : '',
-        );
+        $path = ($operation->getRoutePrefix() ?? '').($operation->getUriTemplate() ?? '');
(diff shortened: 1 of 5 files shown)


soyuka:
nice thanks!

Issue #4890, opened 2022-08-09
Improve *Context operation options


vincentchalamon:
#4881 added an `openapi` operation option to disable an operation from the OpenApi documentation. This option is in addition to the `openapiContext` which helps to send options to the OpenApi documentation for the dedicated operation.

In order to improve the DX and prevent confusions, those options should be merged in `openapi`, allowing to receive null (default), boolean and array values:

```php
#[Get(openapi: false)] // the operation is not exposed on the OpenApi documentation
#[Get(openapi: null)] // (default value) the operation is exposed on the OpenApi documentation
#[Get(openapi: true)] // (similar to "null") the operation is exposed on the OpenApi documentation
#[Get(openapi: ['foo' => 'bar'])] // the operation is exposed on the OpenApi documentation, and a context is available
```

The same behavior could be done for `hydraContext`.

(ticket from https://github.com/api-platform/core/issues/5209)

See what this person tried in their code: https://github.com/api-platform/core/issues/5191#issue-1450210124 

IMO we should try and support something like this:

```
use ApiPlatform\OpenApi\Model\Operation as OpenApiOperation;
use ApiPlatform\OpenApi\Model\Parameter;

#[Get(
  '/my_openapi_custom_operation', 
  openApi: new OpenApiOperation(
    parameters: [
        new Parameter('myParameter', 'query', 'Enter a value for parameter', false, false, true, ['type' => 'string'])
    ]
)]
```
Would be a way better API to customize the openapi related informations. 

https://github.com/api-platform/core/issues/4890 and https://github.com/api-platform/core/issues/5037 and probably more

vincentchalamon:
Audit code looking for *Context options on ApiResource, Operation and ApiProperty.

The idea behind would be:
- Should we merge some of those options? (e.g.: `jsonldSchemaContext` vs `openapiContext`)
- Deprecate `*Context` option
- Add contextless option accepting false, null and object (e.g.: `openapi: new OpenApiOperation(...)`)

PR #5254, opened 2022-12-07
[OpenApi] Add ApiResource::openapi and deprecate openapiContext

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4890
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1679


src/Metadata/Extractor/XmlResourceExtractor.php
```diff
@@ -156,6 +161,91 @@ private function buildFormats(\SimpleXMLElement $resource, string $key): ?array
...
+        $openapi = $resource->openapi;
+        $data = [];
+        $attributes = $openapi->attributes();
+        foreach ($attributes as $attribute) {
+            $data[$attribute->getName()] = $this->phpize($attributes, 'deprecated', 'deprecated' === $attribute->getName() ? 'bool' : 'string');
+        }
+
+        $data['tags'] = $this->buildArrayValue($resource, 'tag');
+
+        if (isset($openapi->responses->response)) {
```
soyuka, line 183: How does this work?

vincentchalamon: 🤷‍♀️ welcome to the wonderful world of XML

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -269,21 +283,39 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
+
+            $pathItem = $pathItem->{'with'.ucfirst($method)}($openapiOperation
+                ->withOperationId($operationId)
+                ->withResponses($responses)
+                ->withParameters($parameters)
+                ->withRequestBody($requestBody)
+            );
 
             $paths->addPath($path, $pathItem);
         }
```
soyuka, line 321: We should add a deprecation here don't you think? If the openapicontext array has some properties, just `trigger_deprecation`.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -249,14 +255,22 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
+            } elseif ($contextResponses = $operation->getOpenapiContext()['responses'] ?? false) {
+                // TODO Remove this "elseif" in 4.0
                 foreach ($contextResponses as $statusCode => $contextResponse) {
                     $responses[$statusCode] = new Response($contextResponse['description'] ?? '', isset($contextResponse['content']) ? new \ArrayObject($contextResponse['content']) : null, isset($contextResponse['headers']) ? new \ArrayObject($contextResponse['headers']) : null, isset($contextResponse['links']) ? new \ArrayObject($contextResponse['links']) : null);
                 }
             }
 
             $requestBody = null;
-            if ($contextRequestBody = $operation->getOpenapiContext()['requestBody'] ?? false) {
+            if ($contextRequestBody = $operation->getOpenapi()?->getRequestBody() ?: false) {
```
soyuka, line 270: what if getOpenapi is `true` will that work? `$a = true; $a?->test` shows `PHP Warning:  Attempt to read property "test" on bool in php shell code on line 1`.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -158,7 +158,11 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
 
             [$requestMimeTypes, $responseMimeTypes] = $this->getMimeTypes($operation);
 
-            $operationId = $operation->getOpenapiContext()['operationId'] ?? $this->normalizeOperationName($operationName);
+            $operationId = $operation->getOpenapi()?->getOperationId() ?: $this->normalizeOperationName($operationName);
```
soyuka, line 161: we need to handle when `getOpenapi` is `true`, I guess `false` is handled above. Maybe that we could just propagate the `openApiOperation` if it exists to underlying functions?

tests/Fixtures/TestBundle/Resources/config/api_resources/resources.yaml
```diff
@@ -11,8 +11,9 @@ resources:
     ApiPlatform\Tests\Fixtures\TestBundle\Model\DummyAddress:
         operations:
             ApiPlatform\Metadata\GetCollection:
-                openapiContext:
-                    x-visibility: hide
```
soyuka, line 15: we should keep this to handle the deprecation

tests/Metadata/Extractor/ResourceMetadataCompatibilityTest.php
```diff
@@ -301,7 +309,12 @@ final class ResourceMetadataCompatibilityTest extends TestCase
                         'foo' => ['bar' => 'baz'],
                     ],
                     'openapiContext' => [
-                        'bar' => 'baz',
```
soyuka, line 304: same as above

soyuka:
I really like this it's awesome!

src/OpenApi/Factory/OpenApiFactory.php
vincentchalamon: IMHO we should not allow `true` value as it doesn't make any sense. WDYT about trowing an error on the setter (operation, ApiResource) if the value is `true`, or automatically convert it to null?

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -158,7 +161,16 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
 
             [$requestMimeTypes, $responseMimeTypes] = $this->getMimeTypes($operation);
 
-            $operationId = $operation->getOpenapiContext()['operationId'] ?? $this->normalizeOperationName($operationName);
+            $operationId = $hasOpenapi && $openapiOperation->getOperationId() ? $openapiOperation->getOperationId() : $this->normalizeOperationName($operationName);
```
soyuka, line 164: I'd like to find something better to avoid these calls, what about declaring the default operation line 140:

```
if (!$openapiOperation) {
    $openapiOperation = new Model\Operation(operationId: $this->normalizeOperationName($operationName));
}
```

Then you can just use `$openapiOperation->getOperationId();`

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -269,21 +306,44 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
-                $operation->getOpenapiContext()['servers'] ?? null,
-                array_filter($operation->getOpenapiContext() ?? [], static fn ($item): int|false => preg_match('/^x-.*$/i', (string) $item), \ARRAY_FILTER_USE_KEY)
-            ));
+            $openapiOperation = ($hasOpenapi ? $openapiOperation : new Model\Operation())
+                // Defaults
+                ->withTags($hasOpenapi && $openapiOperation->getTags() ? $openapiOperation->getTags() : [$operation->getShortName() ?: $resourceShortName])
+                ->withSummary($hasOpenapi && $openapiOperation->getSummary() ? $openapiOperation->getSummary() : $this->getPathDescription($resourceShortName, $method, $operation instanceof CollectionOperationInterface))
+                ->withDescription($hasOpenapi && $openapiOperation->getDescription() ? $openapiOperation->getDescription() : $this->getPathDescription($resourceShortName, $method, $operation instanceof CollectionOperationInterface))
+                ->withDeprecated($hasOpenapi && $openapiOperation->getDeprecated() ? $openapiOperation->getDeprecated() : (bool) $operation->getDeprecationReason());
```
soyuka, line 314: I'm then not sure about this section, either we try to have sane defaults in the `new Model\Operation` (from the suggestion above). Or we do:

```
if (null === $openapiOperation->getTags()) {
  $openApiOperation->withTags([...]);
}

etc.
```

This means that we should fix our getters in Openapi\Model (which isn't perfect as if we want to reflect a property from the specification as **required** I think it's correct).

src/Documentation/Action/DocumentationAction.php
```diff
@@ -32,6 +32,7 @@ public function __construct(private readonly ResourceNameCollectionFactoryInterf
 
     public function __invoke(Request $request = null): DocumentationInterface
     {
+        $context = [];
```
soyuka, line 35: It should be fixed by now can you remove this from the commit?

phpstan.neon.dist
```diff
@@ -83,5 +83,3 @@ parameters:
 		-
 			message: '#^Property .+ is unused.$#'
 			path: tests/Doctrine/Odm/PropertyInfo/Fixtures/DoctrineDummy.php
-		# Waiting for https://github.com/laminas/laminas-code/pull/150
-		- '#Call to an undefined method ReflectionEnum::.+#'
```
soyuka, line 87: same this should be somewhere else

soyuka:
huge thanks!

jojomi:
I highly welcome this update!

Is there any way to update the code (semi-)automatically? Some Rector task or anything else?

PR #5261, opened 2022-12-12
fix(normalizer): normalize items in related collection with concrete class

usu:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | 
| License       | MIT

Fixes a bug in 3.0, where polymorphic items are normalized with the abstract class in the context instead of with the concrete class.

Given a user, who has a relation `ownedVehicles` to an abstract `Vehicle` resource, the following output would be expected (=behavior in 2.6).

**Expected response for GET request on `/vehicles`** (endpoint of abstract Vehicle class)
```json
{
    "_links": {
        "self": {
            "href": "/vehicles"
        },
        "items": [
            {
                "href": "/cars/1"
            },
            {
                "href": "/bicycles/2"
            }
        ]
    },
    "_embedded": {
        "items": [
            {
                "_links": {
                    "self": {
                        "href": "/cars/1"
                    }
                },
                "name": "I'm a car",
                "motor": "hybrid"
            },
            {
                "_links": {
                    "self": {
                        "href": "/bicycles/2"
                    }
                },
                "name": "I'm a bicycle",
                "mudguard": true
            }
        ]
    }
}
```

**Expected response for GET request on `/user/1`:**
```json
{
    "_links": {
        "self": {
            "href": "/user/1"
        },
        "ownedVehicles": [
            {
                "href": "/cars/1"
            },
            {
                "href": "/bicycles/2"
            }
        ]
    },
    "_embedded": {
        "ownedVehicles": [
            {
                "_links": {
                    "self": {
                        "href": "/cars/1"
                    }
                },
                "name": "I'm a car",
                "motor": "hybrid"
            },
            {
                "_links": {
                    "self": {
                        "href": "/bicycles/2"
                    }
                },
                "name": "I'm a bicycle",
                "mudguard": true
            }
        ]
    }
}
```

However, for the 2nd request (`/user/1`), the following response is received with api-platform 3.0 (wrong self links for embedded items):
```json
{
    "_links": {
        "self": {
            "href": "/user/1"
        },
        "ownedVehicles": [
            {
                "href": "/cars/1"
            },
            {
                "href": "/bicycles/2"
            }
        ]
    },
    "_embedded": {
        "ownedVehicles": [
            {
                "_links": {
                    "self": {
                        "href": "/vehicles/1"
                    }
                },
                "name": "I'm a car",
                "motor": "hybrid"
            },
            {
                "_links": {
                    "self": {
                        "href": "/vehicles/2"
                    }
                },
                "name": "I'm a bicycle",
                "mudguard": true
            }
        ]
    }
}
```

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -628,6 +628,15 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                 throw new UnexpectedValueException('Unexpected non-object element in to-many relation.');
             }
 
+            // update context, if concrete object class deviates from general relation class (e.g. in case of polymorphic resources)
+            $objResourceClass = $this->resourceClassResolver->getResourceClass($obj, $resourceClass);
+            if ($objResourceClass !== $resourceClass) {
+                $context['resource_class'] = $objResourceClass;
+                if ($this->resourceMetadataCollectionFactory) {
+                    $context['operation'] = $this->resourceMetadataCollectionFactory->create($objResourceClass)->getOperation();
+                }
+            }
+
             $value[$index] = $this->normalizeRelation($propertyMetadata, $obj, $resourceClass, $format, $context);
         }
 
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -569,11 +569,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
             $resourceClass = $this->resourceClassResolver->getResourceClass($attributeValue, $className);
             $childContext = $this->createChildContext($context, $attribute, $format);
-            $childContext['resource_class'] = $resourceClass;
-            if ($this->resourceMetadataCollectionFactory) {
-                $childContext['operation'] = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation();
-            }
-            unset($childContext['iri'], $childContext['uri_variables']);
+            unset($childContext['iri'], $childContext['uri_variables'], $childContext['resource_class'], $childContext['operation']);
 
             return $this->normalizeCollectionOfRelations($propertyMetadata, $attributeValue, $resourceClass, $format, $childContext);
         }
@@ -589,7 +585,11 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
             $resourceClass = $this->resourceClassResolver->getResourceClass($attributeValue, $className);
             $childContext = $this->createChildContext($context, $attribute, $format);
-            unset($childContext['iri'], $childContext['uri_variables'], $childContext['resource_class'], $childContext['operation']);
+            $childContext['resource_class'] = $resourceClass;
+            if ($this->resourceMetadataCollectionFactory) {
+                $childContext['operation'] = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation();
+            }
+            unset($childContext['iri'], $childContext['uri_variables']);
 
             return $this->normalizeRelation($propertyMetadata, $attributeValue, $resourceClass, $format, $childContext);
         }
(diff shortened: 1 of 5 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -628,6 +628,15 @@ protected function normalizeCollectionOfRelations(ApiProperty $propertyMetadata,
                 throw new UnexpectedValueException('Unexpected non-object element in to-many relation.');
             }
 
+            // update context, if concrete object class deviates from general relation class (e.g. in case of polymorphic resources)
+            $objResourceClass = $this->resourceClassResolver->getResourceClass($obj, $resourceClass);
+            if ($objResourceClass !== $resourceClass) {
+                $context['resource_class'] = $objResourceClass;
+                if ($this->resourceMetadataCollectionFactory) {
+                    $context['operation'] = $this->resourceMetadataCollectionFactory->create($objResourceClass)->getOperation();
```
soyuka, line 636:
```suggestion
                    $context['operation'] ??= $this->resourceMetadataCollectionFactory?->create($objResourceClass)->getOperation();
```

Would that work to avoid `if`?

tests/Fixtures/TestBundle/Entity/Dummy.php
```diff
@@ -106,6 +106,9 @@ class Dummy
     #[ORM\ManyToMany(targetEntity: RelatedDummy::class)]
     public Collection|iterable $relatedDummies;
 
+    #[ORM\OneToMany(targetEntity: AbstractDummy::class, mappedBy: 'dummy')]
+    public Collection|iterable $abstractDummies;
+
```
soyuka, line 111: we usually try to avoid modifying existing features as it'll probably break functional tests as the answer will be different.

soyuka:
Can you add a functional test inside `features/jsonapi`?

src/Serializer/AbstractItemNormalizer.php
usu: `$context['operation']` is already populated with the wrong operation in https://github.com/api-platform/core/blob/main/src/Serializer/AbstractItemNormalizer.php#L594
So I actually need to override it here with the correct one.

I could remove if from the earlier location entirely and populate `resource_class` and `operation` only within `normalizeCollectionOfRelations`, then this could work

soyuka: Mhh but why isn't that done in the AbstractItemNormalizer? IMO this behavior should be the same accross all our normalizers no?

tests/Fixtures/TestBundle/Entity/Dummy.php
soyuka: you'll have more work fixing every changed response so yeah definitely, especially Dummy is used on way too many use cases.

usu:
> Can you add a functional test inside `features/jsonapi`?

I added functional tests for HAL format, reverted the changes in the dummy entities and found/fixed a bug in my new code.

@soyuka Can you have another look at this? If this looks good, I can stash the commits and fix the commit name, so we have a clean history.

tests/Fixtures/TestBundle/Entity/Dummy.php
usu: Fixed/reverted. I used the `DummyTableInheritance*` entities which were already utilized for exactly this purpose in the functional tests.

src/Serializer/AbstractItemNormalizer.php
usu: > Mhh but why isn't that done in the AbstractItemNormalizer?

This is already in `AbstractItemNormalizer`. Not sure I understand.

> Would that work to avoid if?

The first `if` was actually a really bad idea. As this is in a for-loop, context from one item spilled over to other items in case the if-statement was false (was actually the case in the new behat test I introduced). I now changed it, so `resource_class` and `operation` is populated within the for-loop and always overridden.

Do you feel this is a safe approach? Or better to clone the complete `$context` for every object within the collection?

soyuka: We already mutate this object a lot no problem. It's definitely arguable haha.

soyuka:
Thanks! this looks quite good to me! Don't worry I squash from the github interface and we only take the first comment :).

PR #5213, opened 2022-11-22
refactor(elasticsearch): merge metadata to Operation

davy-beauzil:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | 

Merge elasticsearch metadata in Operation

PR #5264, opened 2022-12-12
Update ElasticsearchProviderResourceMetadataCollectionFactory.php

rapaelector:
by default the elasticsearch is not false but null so it will create confusion with indices and try to check the index as the shortName. But the shortName by default is the model class and it is not true for every cases , even if the index is different of the class and create an issue

| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

### DIFF
--- a/src/Core/Upgrade/UpgradeApiResourceVisitor.php
+++ b/src/Core/Upgrade/UpgradeApiResourceVisitor.php
@@ -443,6 +443,9 @@ final class UpgradeApiResourceVisitor extends NodeVisitorAbstract
 
             $method = $operation['method'] ?? strtoupper($operationName);
             unset($operation['method']);
+            if (!\in_array($operationName, ['get', 'post', 'put', 'patch', 'delete'], true)) {
+                $operation['name'] = $operationName;
+            }
             $operations[] = $this->createOperation($this->getOperationNamespace($method, $isCollection), $operation);
         }
 


rapaelector:
I see that if we custom the index like this : 
`api_platform:
    ....
    mapping:
        paths:
            - '%kernel.project_dir%/src/Model'
    elasticsearch:
        hosts: [ '%env(ELASTICSEARCH_HOST)%' ]
        mapping:
            App\Model\Document:
                index: '%env(APP_ENV)%_hubble_document'
                type: actor
`
we have to specify in the Model the provider and the elasticsearch : 
`#[Get(normalizationContext: ['groups' => ['actor-read']], elasticsearch: false, provider: ItemProvider::class)]`

But why it is not automatic to found the provider and why should I add the elasticsearch as false ?

soyuka:
Right, can you target 3.0 ? We've an ongoing work on elasticsearch here : https://github.com/api-platform/core/pull/5213/files

rapaelector:
> Right, can you target 3.0 ? We've an ongoing work on elasticsearch here : https://github.com/api-platform/core/pull/5213/files

thanks for your reply, yes I’ll do it tomorrow

soyuka:
supersseeded by https://github.com/api-platform/core/pull/5272

soyuka:
There are tests failing over that: https://github.com/api-platform/core/pull/5272 

Indeed, I think there's more to it as if elasticsearch is enabled we want to try and get indices. This is quite bad DX we need to improve it but its not yet a priority. please patch this over to your codebase by overriding the service for now or help me get the tests to work at 5272 thanks!

rapaelector:
thanks for your reply , 
I've sent you this PR that will fix the test : [51](https://github.com/soyuka/core/pull/51)

PR #5266, opened 2022-12-13
fix: do not use api_graphql_graphiql route when graphiQl is disabled because it won’t exist

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

#5251 changed `SwaggerUi/index.html.twig` to always display a link to GraphiQL, but the `api_graphql_graphiql` route does not exist if `graphiQlEnabled` is `false`.

https://github.com/api-platform/core/blob/d0fcd70c34e79e1b072032a847b3a6e94f63db72/src/Symfony/Routing/ApiLoader.php#L141-L145


alanpoulain:
Hello, thanks for this.
Could you add back the line:
```twig
{% if not graphQlEnabled %}<a href="javascript:alert('GraphQL support is not enabled, see https://api-platform.com/docs/core/graphql/')">GraphiQL</a>{% endif %}
```
Or maybe we don't want it anymore, @api-platform/core-team?

MatTheCat:
@alanpoulain indeed missed that one, but it wasn’t consistent with the other link (also some others check the active UI).

MatTheCat:
In that case shouldn’t it be the same for the GraphQL Playground link?

alanpoulain:
> In that case shouldn’t it be the same for the GraphQL Playground link?

You mean checking the `active_ui`? I think you could add it too.

alanpoulain:
GraphiQL is enabled by default, it's not the case for GraphQL Playground, so there is no need to add the alert for it.

MatTheCat:
The default configuration seems to enable both `graphiql` and `graphql_playground`: https://github.com/api-platform/core/blob/d0fcd70c34e79e1b072032a847b3a6e94f63db72/src/Symfony/Bundle/DependencyInjection/Configuration.php#L229-L234

Am I missing something?

MatTheCat:
Okay, PR updated!

alanpoulain:
Thank you @MatTheCat.

soyuka:
this should've been merge on 3.0

MatTheCat:
The bug this PR fixes was introduced on the `main` branch though.

soyuka:
oh right my bad thanks!

Issue #5263, opened 2022-12-12
SchemaFactory generates hydra fields for denormalizationContext


benblub:
**API Platform version(s) affected**: 3.0.4

**Description**  
If I look in the generated schema for denormalizationContext i can see the required fields for '@id' and '@type'. Why do have this fields required for **write** operations? We don't need to pass this fields in a write request. If you used this schema on the client with typescript you need to write workarounds because you not sending this fields in the requests.

This field is coming from `ApiPlatform\Hydra\JsonSchema\SchemaFactory`  -> BASE_PROPS

PR #5267, opened 2022-12-13
fix: do not add linked data for denormalizationContext

benblub:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/core/issues/5263
| License       | MIT
| Doc PR       | N/A

If I look in the generated schema for denormalizationContext i can see the required fields for '@id' and '@type'. Why do have this fields required for write operations? We don't need to pass this fields in a write request. If you used this schema on the client with typescript you need to write workarounds because you not sending this fields in the requests.


benblub:
@soyuka any other changes needed?

PR #5268, opened 2022-12-13
refactor: HTTP cache configuration overhaul

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4512
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1678

By working to solve #4512 on @soyuka’s request I found many oddities which I addressed as well.

I took a pretty long time understanding how the HTTP cache configuration works because the naming really isn’t obvious.

By default, enabling `http_cache.invalidation` will only add tags to the API responses; you need to configure `http_cache.invalidation.varnish_urls` and an optional “purger” to effectively invalidate these tags. Now, the default purger actually leverages [Varnish banning feature](https://www.varnish-software.com/developers/tutorials/ban/), which does not really “purge” anything. It is also the purgers’ job to add tags to API responses, which definitively feels out of place.

As it breaks the SRP I deprecated the `PurgerInterface`. Its `purge` method is replaced by `TagsInvalidatorInterface::invalidate` and its `getResponseHeaders` is replaced by `TagsHeadersProviderInterface::provideHeaders`. This reflects the new configuration:

```yaml
api_platform:
    http_cache:
        tags: # Configure how tags are added to responses. This can be disabled if you don’t use a reverse proxy
            header: # Tags are added in a response header. This is currently the only implementation
                name: # Name of the header. Defaults to “Cache-Tags”
                separator: # String added between tags. Defaults to “,”
            invalidator: # Configure how tags are gonna be invalidated. Only one implementation can be selected. This is disabled by default
                id: # ID of one of your own implementation of “TagsInvalidatorInterface”
                banner: # Send BAN requests
                    header: # Configure the header containing the regex to ban
                        name: # Name of the header. Defaults to “ApiPlatform-Ban-Regex”
                        max_size: # Maximum size of the header. Defaults to 7500 bytes
                    urls: # URLs to send BAN requests to
                    client_options: # Options to add to the HTTP clients sending BAN requests
                purger: # Send PURGE requests
                    header: # Configure the header containing the tags to purge
                        name: # Name of the header. Defaults to “xkey”
                        max_size: # Maximum size of the header. Defaults to 7500 bytes
                        separator: # Tags separator. Defaults to “ ” (space)
                    urls: # URLs to send PURGE requests to
                    client_options: # Options to add to the HTTP clients sending PURGE requests

```

I provided a BC layer so everything should be working without any change.

### DIFF
--- a/src/Doctrine/EventListener/PurgeHttpCacheListener.php
+++ b/src/Doctrine/EventListener/PurgeHttpCacheListener.php
@@ -20,6 +20,7 @@ use ApiPlatform\Exception\InvalidArgumentException;
 use ApiPlatform\Exception\OperationNotFoundException;
 use ApiPlatform\Exception\RuntimeException;
 use ApiPlatform\HttpCache\PurgerInterface;
+use ApiPlatform\HttpCache\TagsInvalidatorInterface;
 use ApiPlatform\Metadata\GetCollection;
 use ApiPlatform\Util\ClassInfoTrait;
 use Doctrine\Common\Util\ClassUtils;
@@ -41,8 +42,12 @@ final class PurgeHttpCacheListener
     private readonly PropertyAccessorInterface $propertyAccessor;
     private array $tags = [];
 
-    public function __construct(private readonly PurgerInterface $purger, private readonly IriConverterInterface $iriConverter, private readonly ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
+    public function __construct(private readonly PurgerInterface|TagsInvalidatorInterface $invalidator, private readonly IriConverterInterface $iriConverter, private readonly ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
     {
+        if ($this->invalidator instanceof PurgerInterface) {
+            trigger_deprecation('api-platform/core', '3.1', 'Passing $invalidator a %s is deprecated, pass a %s instead.', PurgerInterface::class, TagsInvalidatorInterface::class);
+        }
+
         $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
     }
 
@@ -100,7 +105,11 @@ final class PurgeHttpCacheListener
             return;
         }
 
-        $this->purger->purge(array_values($this->tags));
+        if ($this->invalidator instanceof PurgerInterface) {
+            $this->invalidator->purge(array_values($this->tags));
+        } elseif ($this->invalidator instanceof TagsInvalidatorInterface) {
+            $this->invalidator->invalidate($this->tags);
+        }
 
         $this->tags = [];
     }
--- a/src/HttpCache/EventListener/AddTagsListener.php
+++ b/src/HttpCache/EventListener/AddTagsListener.php
@@ -16,6 +16,8 @@ namespace ApiPlatform\HttpCache\EventListener;
 use ApiPlatform\Api\IriConverterInterface;
 use ApiPlatform\Api\UrlGeneratorInterface;
 use ApiPlatform\HttpCache\PurgerInterface;
+use ApiPlatform\HttpCache\TagsHeadersProvider;
+use ApiPlatform\HttpCache\TagsHeadersProviderInterface;
 use ApiPlatform\Metadata\CollectionOperationInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use ApiPlatform\State\UriVariablesResolverTrait;
@@ -24,14 +26,7 @@ use ApiPlatform\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
 
 /**
- * Sets the list of resources' IRIs included in this response in the "Cache-Tags" and/or "xkey" HTTP headers.
- *
- * The "Cache-Tags" is used because it is supported by CloudFlare.
- *
- * @see https://support.cloudflare.com/hc/en-us/articles/206596608-How-to-Purge-Cache-Using-Cache-Tags-Enterprise-only-
- *
- * The "xkey" is used because it is supported by Varnish.
- * @see https://docs.varnish-software.com/varnish-cache-plus/vmods/ykey/
+ * Adds the IRI list of the request’s resources in the response’s headers.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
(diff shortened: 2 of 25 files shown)


dunglas:
I won't do a difference between banners and purgers. It's an implementation detail and they are are many cache invalidation API out there (even Varnish has several ones, depending on if you use the paid version or not, then there are custom APIs for Cloudflare, Fastly, Akamaï...). I would keep "purger" to not have to introduce deprecations unless really necessary.

MatTheCat:
Banners and purgers are different the same way `VarnishXKeyPurger` and `VarnishPurger` are: they are implementations of tags invalidator.

This PR makes them configurable, but they are not configurable the same way so I need to tell them apart.

dunglas:
Both [Varnish xkey and ykey](https://docs.varnish-software.com/varnish-cache-plus/vmods/ykey/),  [Fastly](https://developer.fastly.com/learning/concepts/purging/) and [Cloudflare](https://developers.cloudflare.com/cache/how-to/purge-cache/) use the term "purge". This word is common enough, why should we change
(and introduce deprecations that we annoy users for no clear benefits)?

darkweak:
IMHO that's a huge BC break, unless you keep the both configurations, that coud not be merged before a major release right?
Some of tag-based cache systems are not using `PURGE` method (e.g. cloudflare, because you have to keep in mind that you might have only cloudflare as a cache-tag provider).

MatTheCat:
There are no BC breaks in this PR, only deprecation of the `PurgerInterface` and the `http_cache.invalidation` config node as explained in the description.

API Platform provides two invalidators: one using PURGE, and the other using BAN. This PR makes these configurable but it does not add any.

darkweak:
What about the possible authentication (e.g. API key, basic auth) to invalidate tags on the caching server?

MatTheCat:
This is outside the scope of this PR but I guess you could configure `client_options` (former `request_options`) for that.

src/Doctrine/EventListener/PurgeHttpCacheListener.php
```diff
@@ -100,7 +105,11 @@ public function postFlush(): void
             return;
         }
 
-        $this->purger->purge(array_values($this->tags));
+        if ($this->invalidator instanceof PurgerInterface) {
+            $this->invalidator->purge(array_values($this->tags));
+        } elseif ($this->invalidator instanceof TagsInvalidatorInterface) {
```
soyuka, line 110: we do not want to introduce a new interface

src/HttpCache/EventListener/AddTagsListener.php
```diff
@@ -75,13 +80,9 @@ public function onKernelResponse(ResponseEvent $event): void
...
-        if (!$this->purger) {
-            $response->headers->set('Cache-Tags', implode(',', $resources));
-
-            return;
-        }
-
-        $headers = $this->purger->getResponseHeaders($resources);
+        $headers = $this->headersProvider instanceof PurgerInterface
+            ? $this->headersProvider->getResponseHeaders($resources)
+            : $this->headersProvider->provideHeaders($resources);
```
soyuka, line 85: I'm not sure to understand why this is needed, the purger can hold the responsibility of sending the header he wants. I see a shared responsibility between the Purger and this header as they use often the use especially when using a Surrogate Key implementation, for example [Akamai](https://techdocs.akamai.com/purge-cache/docs/assign-cache-tags) uses (`Edge-Cache-Tag` for both `PURGE` and adding cache tags to resources), the same thing goes for [Souin](https://github.com/darkweak/souin/blob/aca48edc0388010c4498b4f4896e79e0f233eea4/cache/surrogate/specification.md). In varnish, it is your configuration that specifies that and therefore in API Platform it's easier if everything is configured in the same class for our developers IMO.

src/HttpCache/PurgerInterface.php
```diff
@@ -17,6 +17,8 @@
  * Purges resources from the cache.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @deprecated use {@see TagsHeadersProviderInterface} and {@see TagsInvalidatorInterface} instead
```
soyuka, line 21: let's avoid deprecating this new interface, it was discussed pre-3.0 and I don't think the change is justified.

src/HttpCache/TagsInvalidator/Banner.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\HttpCache\TagsInvalidator;
+
+use ApiPlatform\HttpCache\TagsInvalidatorInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class Banner implements TagsInvalidatorInterface
```
soyuka, line 19: The `BAN` invalidation method is quite specific to Varnish and we probably don't want this to be too generic. Indeed, we recommend to use surrogate keys instead as it's less error-prone in its implementation then using regular expressions.

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -305,45 +306,185 @@ private function addSwaggerSection(ArrayNodeDefinition $rootNode): void
...
+                        ->scalarNode('purger')
+                            ->defaultValue('api_platform.http_cache.purger.varnish')
+                            ->info('Specify a varnish purger to use (available values: "api_platform.http_cache.purger.varnish.ban" or "api_platform.http_cache.purger.varnish.xkey").')
+                        ->end()
+                        ->arrayNode('xkey')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->scalarNode('glue')
+                                ->defaultValue(' ')
+                                ->info('xkey glue between keys')
```
soyuka, line 487: I thought that we discussed about the configuration to have less components. Indeed, we want this to be configured using urls and/or scoped http clients (to support auth, default headers and complex approaches). The glue should be configured directly in the service as it is provider-specific. Note that it'd be a valid Symfony configuration to do:

```
services:
  App\MyPurger: 
     parent: 'api_platform.purger'
     arguments: 
         - !tagged api_platform.purger.http_client
         - ', ' // my own glue
         - 'Cache-Tags' // my own header 
```

src/Symfony/Bundle/Resources/config/http_cache_tags.xml
```diff
@@ -5,18 +5,10 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="api_platform.http_cache.purger.varnish" alias="api_platform.http_cache.purger.varnish.ban" public="false" />
```
soyuka, line 8: we can't remove existing services it's a backward compatibility break

tests/Doctrine/EventListener/PurgeHttpCacheListenerTest.php
```diff
@@ -61,8 +61,9 @@ public function testOnFlush(): void
         $toDeleteNoPurge = new DummyNoGetOperation();
         $toDeleteNoPurge->setId(5);
 
-        $purgerProphecy = $this->prophesize(PurgerInterface::class);
-        $purgerProphecy->purge(['/dummies', '/dummies/1', '/dummies/2', '/dummies/3', '/dummies/4'])->shouldBeCalled();
```
soyuka, line 65: you should avoid modifying existing tests as it is often a risk to break current implementations.

src/HttpCache/TagsInvalidator/Banner.php
MatTheCat: Specific to Varnish does not mean it cannot be configurable.

Even if you recommend surrogate keys this invalidator exists so it has to continue existing.

src/HttpCache/EventListener/AddTagsListener.php
MatTheCat: As explained in the PR description, the actual default configuration will not invalidate anything because we want to support adding tags without invalidation. Thus it has to be two different concerns.

Also the header provider can be automatically configured in accordance to the configured invalidator, so it would in fact be simpler.

src/Doctrine/EventListener/PurgeHttpCacheListener.php
MatTheCat: Why?

src/Symfony/Bundle/Resources/config/http_cache_tags.xml
MatTheCat: Indeed.

tests/Doctrine/EventListener/PurgeHttpCacheListenerTest.php
MatTheCat: How am I supposed to test the current implementation? Wouldn’t it trigger deprecations which would make tests fail?

src/HttpCache/TagsInvalidator/Banner.php
darkweak: Only Varnish is using `Ban`-wording. I don't see the benefits of keeping the Banner AND the Purger.

PR #5272, opened 2022-12-16
fix(elasticsearch): check for indices only when enabled

soyuka:
From #5264 but rebased against 3.0


MatTheCat:
Will be superseded by #5213

PR #5275, opened 2022-12-17
feat: better separation of entity class and resource

soyuka:
Related to: https://www.strangebuzz.com/en/blog/replacing-manual-api-endpoints-with-api-platform-3-in-a-symfony-application and https://twitter.com/Pierstoval/status/1581166242659856384


src/Doctrine/Orm/State/CollectionProvider.php
```diff
@@ -42,25 +42,25 @@ public function __construct(private readonly ResourceMetadataCollectionFactoryIn
 
     public function provide(Operation $operation, array $uriVariables = [], array $context = []): iterable
     {
-        $resourceClass = $operation->getClass();
+        $entityClass = $operation->getEntityClass() ?? $operation->getClas();
```
dunglas, line 45:
```suggestion
        $entityClass = $operation->getEntityClass() ?? $operation->getClass();
```

src/Metadata/ApiResource.php
```diff
@@ -147,6 +147,7 @@ public function __construct(
         protected ?array $graphQlOperations = null,
         $provider = null,
         $processor = null,
+        protected ?string $entityClass = null,
```
dunglas, line 150: Shouldn't we create a dedicated attributes in the Doctrine namespace instead of adding Doctrine-specific properties in these class?

soyuka: We already have a bunch of vendor-specific properties (for example ES, mongodb etc). I think that its not a big deal as this class is our interface for developers. Were you thinking about some `Doctrine/Metadata/ApiResource`? but then we need to duplicate every operation?

soyuka:
```
stateOptions: new DoctrineORMOptions(entityClass: MyClass::class)
```

bendavies:
@soyuka confused by this example as there is no custom provider for `ResourceWithSeparatedEntity`. so the `SeparatedEntity` is actually serialized and return in the response, rather than `SeparatedEntity` being converted to `ResourceWithSeparatedEntity`, then serialized.

soyuka:
yes as the fields are exactly the same we don't need any mapping and we just let the serializer handle its thing with SeparatedEntity. Obviously you'd have to use a mapping if that's not the case. A symfony mapper proposal of mine is at https://github.com/symfony/symfony/pull/51741

PR #5277, opened 2022-12-17
fix: apply expected post operation on custom post operation

aplysie:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | 
| License       | MIT
| Doc PR        | 

Context:

This behavior is related to the version 2.7 and when the  `api_platform.metadata_backward_compatibility_layer` is set to `false`.

If you define a resource as follow :

```PHP
/**
 * @ApiResource(
 *     itemOperations={"get"},
 *     collectionOperations={
 *         "subscribe"={
 *             "method"="POST",
 *             "path"="/orders/{orderId}/subscribe",
 *         }
 *     }
 * )
 */
final class Subscription
```
In this case, the metadata will define a `GetCollection` operation instead of the expected `Post` operation.

We don't really know why this specific case was excluded in the condition fixed in this pr.

### DIFF
--- a/src/Metadata/Resource/Factory/LegacyResourceMetadataResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/LegacyResourceMetadataResourceMetadataCollectionFactory.php
@@ -167,7 +167,7 @@ final class LegacyResourceMetadataResourceMetadataCollectionFactory implements R
 
             if (HttpOperation::METHOD_DELETE === $operation['method']) {
                 $newOperation = (new Delete())->withOperation($newOperation);
-            } elseif (HttpOperation::METHOD_POST === $operation['method']) {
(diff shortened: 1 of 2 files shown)


soyuka:
try declaring that as an item operation instead? your patch breaks this definition: https://github.com/api-platform/core/blob/2.7/tests/Fixtures/TestBundle/Entity/Payment.php

aplysie:
> try declaring that as an item operation instead? your patch breaks this definition: https://github.com/api-platform/core/blob/2.7/tests/Fixtures/TestBundle/Entity/Payment.php

Thank you for the feedback. I'm really not sure if i'm in the right path actually. 
Currently it works without the backward compatibility setting. It considers the result of the operation as a GetCollection, the "@iri" is therefore wrong in this specific case.

soyuka:
> try declaring that as an item operation instead?

p4ee5r:
> > try declaring that as an item operation instead?

I was not looking for a solution. Sorry if I could not explain my point better. I'll try later :)

PR #5289, opened 2022-12-22
chore(deprecations): Fix some more deprecations

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Get rid of some more deprecations in the 3.x branch


franmomu:
Looking at deprecations in my app, I've seen the one about using `ArgumentValueResolverInterface` and dumped into this PR.

I think this creates a BC break for people using symfony 6.1 since the [new `ValueResolverInterface` has been added in 6.2](https://github.com/symfony/symfony/pull/47363).

KDederichs:
Mhh yeah seems like it...
Usually they tend to back port stuff like that a version or two but I guess not in this case...
I'll try to take a look if I can fix it in a way that doesn't break BC

Issue #4819, opened 2022-07-06
[RFC] Link Security


KDederichs:
**Description**  
Having the ability to add security checks to the Link Resource would be very convenient for subresources that are linked to a restircted resource.

From what I know it's currently not possible (see https://github.com/api-platform/api-platform/issues/2168) to restrict sub resources that way unless you start building your own logic to extract the entity from the URI using a`QueryCollectionExtensionInterface` which isn't really that great though (also assuming that's a use case that's not THAT uncommon).

From digging around the code a bit it seems the Links are currently archived using query builder joins/conditions so I'm not sure how viable this would actually be using the current system, but it would be nice to have if it somehow could be hooked into Symfonies security systems.

**Example**  
An example would be something like this:
```
#[ApiResource(
    uriTemplate: '/questions/{id}/answer', 
    uriVariables: [
        'id' => new Link(
            fromClass: Question::class,
            fromProperty: 'answer',
            security: "is_granted(SOME_VOTER, object)"
        )
    ], 
    operations: [new Get()]
)]
```

where you can call security operations on the object referenced in the voter.

Eimantas123:
We have almost updated to 2.7 everything and only in the end realised that we can't get uri variables for security. This is very much needed.

PR #5290, opened 2022-12-27
feat(subresource): Link Security

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4819 
| License       | MIT

Here's an idea how Link Security (as described in the issue) might work.
There's still an open issue with choosing the right provider programmatically and mapping the uri variables automatically but I wanted to ask if that's going in the right direction first before investing more work (and making tests)


soyuka:
Hello I need to find time to really understand the need behind that, why don't you put the security on the top level?

Also, not a huge fan of the work done in the `ReadListener` it doesn't look it'll work with all cases. Still I only took a quick look I want to try this.

src/Symfony/EventListener/DenyAccessListener.php
```diff
@@ -85,17 +86,31 @@ private function checkSecurity(Request $request, string $attribute, array $extra
                 $message = $operation->getSecurityMessage();
         }
 
-        if (null === $isGranted) {
```
chalasr, line 88: Why is this change needed?

KDederichs: Originally it'd jump out of there cause if you have no security string you won't need to do anything in the deny listener ($isGranted is probably also the wrong name, since that variable actually contains the expression for checker).

With the change though you could still have security expressions in the `Link`s so you can't just exit out of there. As you can see the check itself also isn't gone, I just moved it to like 94.

KDederichs:
Not stale, still a thing I'd like to see added in some way

usu:
@KDederichs This seems like a very interesting enhancement. Are you still working on this? What do you believe is missing to get this mergeable and is there anything that could be supported with?

KDederichs:
@usu I'd still like it done because I think it's a valid use case.
Technically it's done even, if I were to resolve the merge conflicts since my initial PR did have the whole functionality.

I'm just not sure why it wasn't merged back then, maybe the team doesn't like it or wants some changes? Dunno.
If you can get some API-Platform members to tell you why that's what would be missing :)

soyuka:
Too much work, forgot that PR sorry! I think it's a valid use case. Only thing is that implementation in the ReadListener that I don't like. Please take a look at what we did in the [CreateProvider](https://github.com/api-platform/core/blob/main/src/State/CreateProvider.php#L58). I think it's the same logic where we try to find another resource. When doing so, its needed that the user can specify an operation to get no?

KDederichs:
Hmm yeah I'll try to get it out of there into it's own thing.
I'll probably get on it tomorrow or something and get rid of the merge conflicts and then try to change that

KDederichs:
@soyuka Let me know how you like it now

usu:
@soyuka Is this ready to be merged? Would be awesome to see this in the 3.3 release.

KDederichs:
@soyuka 
Anything missing? :)

Was wondering what the general status of this is

soyuka:
let me review this

src/State/Provider/ReadProvider.php
```diff
@@ -87,6 +90,47 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
...
+                    if (!$securityObjectName) {
+                        continue;
+                    }
+                    $request?->attributes->set($securityObjectName, $relation);
+                } catch (InvalidIdentifierException|InvalidUriVariableException $e) {
+                    throw new NotFoundHttpException('Invalid identifier value or configuration.', $e);
+                }
+            }
+        }
+
```
soyuka, line 133: I think that this may have an impact over performances, it's quite magical and you must be sure to set `parent_uri_template` or to have the correct first operation on the linked resource. Also, by the looks of it I'm wondering if we couldn't just check out the `$data->($uriVariables->getFromProperty()}` instead of calling a read provider on that as it may fetch more then we want no? 

Anyways, the way it could be implemented if calling the provider is needed:

- create a `LinkReadProvider` (or another name not sure ^^) that decorates the `api_platform.state_provider.read` (take example on `api_platform.state_provider.access_checker`)
- this way you can make this `opt-in` add a configuration option that enables link security

src/Symfony/Security/State/AccessCheckerProvider.php
```diff
@@ -75,10 +73,31 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
...
+                }
+
+                $targetResource = $uriVariable->getFromClass() ?? $uriVariable->getToClass();
+
+                if (!$targetResource) {
+                    continue;
+                }
+
+                // We need to add all attributes here again because we do not know the name of the security object
+                $resourceAccessCheckerContext += $request->attributes->all();
```
soyuka, line 93: we gain access to `request` already inside the context can't we work with that?

src/Symfony/Security/State/AccessCheckerProvider.php
```diff
@@ -75,10 +73,31 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
             ];
         }
 
-        if (!$this->resourceAccessChecker->isGranted($operation->getClass(), $isGranted, $resourceAccessCheckerContext)) {
+        if ($isGranted && !$this->resourceAccessChecker->isGranted($operation->getClass(), $isGranted, $resourceAccessCheckerContext)) {
             $operation instanceof GraphQlOperation ? throw new AccessDeniedHttpException($message ?? 'Access Denied.') : throw new AccessDeniedException($message ?? 'Access Denied.');
         }
 
+        if ($operation instanceof HttpOperation && $operation->getUriVariables()) {
```
soyuka, line 80: same thing what about having this feature optional at first, therefore we could put this logic in a provider decorating the `AccessCheckerProvide`. As both the ReadProvider and this part are quite tighten to security I'd leave them into this namespace.

soyuka (changes requested):
Made a few comments, if something is not clear feel free to pm me on the symfony slack.

src/State/Provider/ReadProvider.php
KDederichs: I just remembered why that was there:

If $data is a collection you have no way of actually getting the parent element except trying to fetch it off the DB or other things. So I'd argue using a provider to grab the related field is required.

Sadly that also doesn't work anymore when decorating the read provider, basically making me unable to get the owning object.

Do you by chance know another way I could get that object?

soyuka: yes the `parent_uri_template` is really useful for that. 

> Sadly that also doesn't work anymore when decorating the read provider, basically making me unable to get the owning object. 

You got it by calling `$this->decorated->provide()` no? or else you can always use the `$request->attributes` to keep data during a request.

tests/Symfony/Security/State/LinkAccessCheckerProviderTest.php
```diff
@@ -0,0 +1,89 @@
...
+use ApiPlatform\Metadata\Link;
+use ApiPlatform\State\ProviderInterface;
+use ApiPlatform\Symfony\Security\Exception\AccessDeniedException;
+use ApiPlatform\Symfony\Security\ResourceAccessCheckerInterface;
+use ApiPlatform\Symfony\Security\State\LinkAccessCheckerProvider;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
+
+class LinkAccessCheckerProviderTest extends TestCase
```
soyuka, line 26: :+1: thanks!

soyuka:
Can you add a configuration option and load the `xml` file for these providers only when it's enabled? Thanks.

src/Symfony/Bundle/Resources/config/link_security.xml
```diff
@@ -0,0 +1,20 @@
...
+
+    <services>
+
+        <service id="api_platform.state_provider.read_link" class="ApiPlatform\Symfony\Security\State\LinkedReadProvider" decorates="api_platform.state_provider.read" decoration-priority="499">
+            <argument type="service" id="api_platform.state_provider.read_link.inner" />
+            <argument type="service" id="api_platform.state_provider.locator" />
+            <argument type="service" id="api_platform.metadata.resource.metadata_collection_factory" />
+        </service>
+
+        <service id="api_platform.state_provider.access_checker_linked" class="ApiPlatform\Symfony\Security\State\LinkAccessCheckerProvider" decorates="api_platform.state_provider.read">
```
soyuka, line 15: should this one decorate api_platform.state_provider.access_checker or is it better just after reading the link hence it should be `api_platform.state_provider.read_link`?

src/Symfony/Bundle/Resources/config/link_security.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+
+        <service id="api_platform.state_provider.read_link" class="ApiPlatform\Symfony\Security\State\LinkedReadProvider" decorates="api_platform.state_provider.read" decoration-priority="499">
```
soyuka, line 9: I don't think we need the priority (see next point)

soyuka (commented):
Love it! you also need to add your configuration at https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/DependencyInjection/Configuration.php

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -892,4 +893,11 @@ private function registerInflectorConfiguration(array $config): void
             Inflector::keepLegacyInflector(false);
         }
     }
+
+    private function registerLinkSecurityConfiguration(XmlFileLoader $loader, array $config): void
+    {
+        if ($config['enable_link_security'] ?? true) {
```
soyuka, line 899:
```suggestion
        if ($config['enable_link_security']) {
```

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -111,6 +111,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->booleanNode('enable_docs')->defaultTrue()->info('Enable the docs')->end()
                 ->booleanNode('enable_profiler')->defaultTrue()->info('Enable the data collector and the WebProfilerBundle integration.')->end()
                 ->booleanNode('keep_legacy_inflector')->defaultTrue()->info('Keep doctrine/inflector instead of symfony/string to generate plurals for routes.')->end()
+                ->booleanNode('enable_link_security')->defaultTrue()->info('Enable security for Links (sub resources)')->end()
```
soyuka, line 114:
```suggestion
                ->booleanNode('enable_link_security')->defaultFalse()->info('Enable security for Links (sub resources)')->end()
```

see below reasoning

src/Symfony/Security/State/LinkAccessCheckerProvider.php
```diff
@@ -0,0 +1,75 @@
...
+namespace ApiPlatform\Symfony\Security\State;
+
+use ApiPlatform\Metadata\HttpOperation;
+use ApiPlatform\Metadata\Link;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProviderInterface;
+use ApiPlatform\Symfony\Security\Exception\AccessDeniedException;
+use ApiPlatform\Symfony\Security\ResourceAccessCheckerInterface;
+
+class LinkAccessCheckerProvider implements ProviderInterface
```
soyuka, line 23:
```suggestion

/**
 * class description
 * @experimental
 */
final class LinkAccessCheckerProvider implements ProviderInterface
```

I'd suggest to make these 2 provider experimental until 4.x, then we'll enable the option to `true`. This allows us to make modification until then if needed.

src/Symfony/Security/State/LinkedReadProvider.php
```diff
@@ -0,0 +1,86 @@
...
+use ApiPlatform\Exception\InvalidUriVariableException;
+use ApiPlatform\Metadata\HttpOperation;
+use ApiPlatform\Metadata\Link;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\State\Exception\ProviderNotFoundException;
+use ApiPlatform\State\ProviderInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+class LinkedReadProvider implements ProviderInterface
```
soyuka, line 26: same final + experimental

features/authorization/deny.feature
```diff
@@ -211,6 +211,79 @@ Feature: Authorization checking
     And the response should contain "ownerOnlyProperty"
     And the JSON node "ownerOnlyProperty" should be equal to the string "updated"
 
+  Scenario: An non existing entity should return Not found
```
soyuka, line 214: can you put a tag `@link_security` on this, and then inside `behat.yml.dist` in the `legacy` suite: 

```
      filters:
        tags: '~@postgres&&~@mongodb&&~@elasticsearch&&~@link_security'

```
thanks

soyuka:
thanks! please also add a documentation entry on `api-platform.com/docs`

amacrobert-meq:
I'm looking for YAML configuration of this feature. The docs provide an example using attributes only.

Adding the `security` option as a sibling of a link's `fromClass`/`toProperty`/etc is ignored in YAML. @KDederichs can I request a docs update for YAML and/or XML config? Or if you tell me how to configure it, I'd be happy to make the docs update myself.

KDederichs:
There's no YAML/XML config for this.
You'd have to make a PR for that, it should be fairly simple though? Not completely sure never really dug into how Api Platform loads config values from XML/YAML.
Maybe @soyuka can tell you how much work that'd be

ytilotti:
@amacrobert-meq same issue here! And it's really blocking me. Don't find why is not set in the config load.

PR #5291, opened 2022-12-27
chore(doctrine): use getObjectManager

dannyvw:
use getObjectManager instead of deprecated getEntityManager

| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | #... 
| License       | MIT

### DIFF
--- a/src/Doctrine/EventListener/PublishMercureUpdatesListener.php
+++ b/src/Doctrine/EventListener/PublishMercureUpdatesListener.php
@@ -92,7 +92,7 @@ final class PublishMercureUpdatesListener
     public function onFlush(EventArgs $eventArgs): void
     {
         if ($eventArgs instanceof OrmOnFlushEventArgs) {
-            $uow = $eventArgs->getEntityManager()->getUnitOfWork();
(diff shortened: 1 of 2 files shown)


src/Doctrine/EventListener/PublishMercureUpdatesListener.php
```diff
@@ -92,7 +92,7 @@ public function __construct(ResourceClassResolverInterface $resourceClassResolve
     public function onFlush(EventArgs $eventArgs): void
     {
         if ($eventArgs instanceof OrmOnFlushEventArgs) {
-            $uow = $eventArgs->getEntityManager()->getUnitOfWork();
```
dunglas, line 95: We should test if the `getObjectManager` method exists (and fallback on `getEntityManager` if it doesn't) for backward compatibility with old versions of Doctrine.

That's weird that it's not caught by the CI.

dunglas (changes requested):

src/Doctrine/EventListener/PublishMercureUpdatesListener.php
dannyvw: PR updated

PR #5296, opened 2023-01-02
Draft: feat(normalizer): Add itemsPerPage for json-ld (hydra)

SiebeVE:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | /
| License       | MIT
| Doc PR        | /

Add the hydra:itemsPerPage property to the response of a json ld response.

### DIFF
--- a/features/doctrine/search_filter.feature
+++ b/features/doctrine/search_filter.feature
@@ -69,6 +69,7 @@ Feature: Search filter on collections
           "carBrand": "DummyBrand"
         }
       ],
+      "hydra:itemsPerPage": 3,
       "hydra:totalItems": 1,
       "hydra:view": {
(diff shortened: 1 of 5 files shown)


alanpoulain:
Hello, thanks for this PR however the [Hydra spec](https://www.hydra-cg.com/spec/latest/core/) does not define an `itemsPerPage` property.

SiebeVE:
Mmh, I see, that is unfortunate 😄 .
What do you think of putting it in the `@context`?

soyuka:
I think that you can compute that or probably create your subtype. Anyways we don't really want to implement things into the core that are not part of the specification.

SiebeVE:
Okay, understandably, thanks for the consideration!

Issue #5279, opened 2022-12-19
Elasticsearch


soyuka:
- [ ] create a specific `Elasticsearch\Metadata\Operation`, `Elasticsearch\Metadata\Get`, `Elasticsearch\Metadata\GetCollection` for elasticsearch  (https://github.com/api-platform/core/pull/5213)
- [ ] link the ItemProvider and CollectionProvider when an Elasticsearch operation is found, via a ResourceCollectionMetadataFactory (https://github.com/api-platform/core/pull/5213) 
- [ ] deprecate old elasticsearch mapping via configuration
- [ ] deprecate `ApiPlatform\Elasticsearch\Metadata\Document\DocumentMetadata` and related elasticsearch things
- [ ] check that the old `ApiResource(elasticsearch: true)` still works
- [ ] try to configure Elasticsearch via xml/yaml 
- [ ] check that we can fix some of the reported issues insides [pull requests](https://github.com/api-platform/core/pulls?q=is%3Apr+is%3Aopen+elasticsearch) and [issues](https://github.com/api-platform/core/issues?q=is%3Aissue+is%3Aopen+elasticsearch) with this (especially if we need to add ES specific options). 

For example I think that the actualy logic is wrong in the ES metadata factory (see https://github.com/api-platform/core/pull/5272 and linked issue, https://github.com/api-platform/core/pull/4820)

@davy-beauzil and @MatTheCat

PR #5297, opened 2023-01-02
Refactor $elasticsearch into a new $persistenceMeans option

davy-beauzil:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5279 
| License       | MIT
| Doc PR        | not yet

# Deprecations
- `ApiResource::$elasticsearch` and `Operation::$elasticsearch` were deprecated
- `src/Elasticsearch/Metadata/Document/Factory/*` and `src/Elasticsearch/Metadata/Document/DocumentMetadata` are deprecated
- `elasticsearch.mapping` config option is deprecated

# Feature
- `PersistenceMeansInterface` was created
- `ElasticsearchDocument` was created, which implements `PersistenceMeansInterface`
- `ApiResource::$persistenceMeans` and `Operation::$persistenceMeans` added

### DIFF
--- a/src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchProviderResourceMetadataCollectionFactory.php
+++ b/src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchProviderResourceMetadataCollectionFactory.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Elasticsearch\Metadata\Resource\Factory;
 
 use ApiPlatform\Elasticsearch\State\CollectionProvider;
 use ApiPlatform\Elasticsearch\State\ItemProvider;
+use ApiPlatform\Elasticsearch\State\OptionsInterface;
 use ApiPlatform\Metadata\CollectionOperationInterface;
 use ApiPlatform\Metadata\Operation;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
@@ -26,8 +27,11 @@ use Elasticsearch\Common\Exceptions\NoNodesAvailableException;
 
 final class ElasticsearchProviderResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
 {
-    public function __construct(private readonly Client $client, private readonly ResourceMetadataCollectionFactoryInterface $decorated)
+    public function __construct(private readonly Client $client, private readonly ResourceMetadataCollectionFactoryInterface $decorated, private readonly bool $triggerDeprecation = true)
     {
+        if ($this->triggerDeprecation) {
+            trigger_deprecation('api-platform/core', '3.1', '%s is deprecated and will be removed in v4', self::class);
+        }
     }
 
     /**
@@ -42,6 +46,12 @@ final class ElasticsearchProviderResourceMetadataCollectionFactory implements Re
 
             if ($operations) {
                 foreach ($resourceMetadata->getOperations() as $operationName => $operation) {
+                    if (null !== ($elasticsearch = $operation->getElasticsearch())) {
+                        $solution = $elasticsearch
+                            ? sprintf('Pass an instance of %s to $stateOptions instead', OptionsInterface::class)
+                            : 'You will have to remove it when upgrading to v4';
+                        trigger_deprecation('api-platform/core', '3.1', sprintf('Setting "elasticsearch" in Operation is deprecated. %s', $solution));
+                    }
                     if ($this->hasIndices($operation)) {
                         $operation = $operation->withElasticsearch(true);
                     }
@@ -60,6 +70,12 @@ final class ElasticsearchProviderResourceMetadataCollectionFactory implements Re
 
             if ($graphQlOperations) {
                 foreach ($graphQlOperations as $operationName => $graphQlOperation) {
+                    if (null !== ($elasticsearch = $graphQlOperation->getElasticsearch())) {
+                        $solution = $elasticsearch
+                            ? sprintf('Pass an instance of %s to $stateOptions instead', OptionsInterface::class)
+                            : 'You will have to remove it when upgrading to v4';
+                        trigger_deprecation('api-platform/core', '3.1', sprintf('Setting "elasticsearch" in GraphQlOperation is deprecated. %s', $solution));
+                    }
                     if ($this->hasIndices($graphQlOperation)) {
                         $graphQlOperation = $graphQlOperation->withElasticsearch(true);
                     }
@@ -93,9 +109,7 @@ final class ElasticsearchProviderResourceMetadataCollectionFactory implements Re
             $this->client->cat()->indices(['index' => $index]);
 
             return true;
-        } catch (Missing404Exception) {
-            return false;
-        } catch (NoNodesAvailableException) {
+        } catch (Missing404Exception|NoNodesAvailableException) {
             return false;
         }
     }
--- a/src/Elasticsearch/State/CollectionProvider.php
+++ b/src/Elasticsearch/State/CollectionProvider.php
@@ -14,19 +14,19 @@ declare(strict_types=1);
 namespace ApiPlatform\Elasticsearch\State;
 
 use ApiPlatform\Elasticsearch\Extension\RequestBodySearchCollectionExtensionInterface;
+use ApiPlatform\Elasticsearch\Metadata\Document\DocumentMetadata;
 use ApiPlatform\Elasticsearch\Metadata\Document\Factory\DocumentMetadataFactoryInterface;
 use ApiPlatform\Elasticsearch\Paginator;
 use ApiPlatform\Elasticsearch\Util\ElasticsearchVersion;
 use ApiPlatform\Metadata\Operation;
 use ApiPlatform\State\Pagination\Pagination;
 use ApiPlatform\State\ProviderInterface;
+use ApiPlatform\Util\Inflector;
 use Elasticsearch\Client;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 
 /**
  * Collection provider for Elasticsearch.
- *
- * @experimental
  *
  * @author Baptiste Meyer <baptiste.meyer@gmail.com>
  * @author Vincent Chalamon <vincentchalamon@gmail.com>
@@ -46,7 +46,6 @@ final class CollectionProvider implements ProviderInterface
     public function provide(Operation $operation, array $uriVariables = [], array $context = []): Paginator
     {
         $resourceClass = $operation->getClass();
-        $documentMetadata = $this->documentMetadataFactory->create($resourceClass);
         $body = [];
 
         foreach ($this->collectionExtensions as $collectionExtension) {
@@ -60,13 +59,21 @@ final class CollectionProvider implements ProviderInterface
         $limit = $body['size'] ??= $this->pagination->getLimit($operation, $context);
         $offset = $body['from'] ??= $this->pagination->getOffset($operation, $context);
 
+        $index = Inflector::tableize($operation->getShortName());
+        $options = $operation->getStateOptions() instanceof Options ? $operation->getStateOptions() : new Options(index: $index);
+
+        // TODO: remove in 4.x
+        if ($operation->getElasticsearch() && !$operation->getStateOptions()) {
+            $options = $this->convertDocumentMetadata($this->documentMetadataFactory->create($resourceClass));
+        }
+
         $params = [
-            'index' => $documentMetadata->getIndex(),
+            'index' => $options->getIndex() ?? $index,
             'body' => $body,
         ];
 
-        if (ElasticsearchVersion::supportsMappingType()) {
-            $params['type'] = $documentMetadata->getType();
+        if (null !== $options->getType() && ElasticsearchVersion::supportsMappingType()) {
+            $params['type'] = $options->getType();
         }
 
         $documents = $this->client->search($params);
@@ -79,5 +86,10 @@ final class CollectionProvider implements ProviderInterface
             $offset,
             $context
         );
+    }
+
+    private function convertDocumentMetadata(DocumentMetadata $documentMetadata): Options
+    {
+        return new Options($documentMetadata->getIndex(), $documentMetadata->getType());
     }
 }
--- a/src/Elasticsearch/State/ItemProvider.php
+++ b/src/Elasticsearch/State/ItemProvider.php
@@ -13,19 +13,19 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Elasticsearch\State;
 
+use ApiPlatform\Elasticsearch\Metadata\Document\DocumentMetadata;
 use ApiPlatform\Elasticsearch\Metadata\Document\Factory\DocumentMetadataFactoryInterface;
 use ApiPlatform\Elasticsearch\Serializer\DocumentNormalizer;
 use ApiPlatform\Elasticsearch\Util\ElasticsearchVersion;
 use ApiPlatform\Metadata\Operation;
 use ApiPlatform\State\ProviderInterface;
+use ApiPlatform\Util\Inflector;
 use Elasticsearch\Client;
 use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 
 /**
  * Item provider for Elasticsearch.
- *
- * @experimental
  *
  * @author Baptiste Meyer <baptiste.meyer@gmail.com>
  * @author Vincent Chalamon <vincentchalamon@gmail.com>
@@ -42,16 +42,23 @@ final class ItemProvider implements ProviderInterface
     public function provide(Operation $operation, array $uriVariables = [], array $context = []): ?object
     {
         $resourceClass = $operation->getClass();
-        $documentMetadata = $this->documentMetadataFactory->create($resourceClass);
+        $index = Inflector::tableize($operation->getShortName());
+
+        $options = $operation->getStateOptions() instanceof Options ? $operation->getStateOptions() : new Options(index: $index);
+
+        // TODO: remove in 4.x
+        if ($operation->getElasticsearch() && !$operation->getStateOptions()) {
+            $options = $this->convertDocumentMetadata($this->documentMetadataFactory->create($resourceClass));
+        }
 
         $params = [
             'client' => ['ignore' => 404],
-            'index' => $documentMetadata->getIndex(),
+            'index' => $options->getIndex() ?? $index,
             'id' => (string) reset($uriVariables),
         ];
 
-        if (ElasticsearchVersion::supportsMappingType()) {
-            $params['type'] = $documentMetadata->getType();
+        if (null !== $options->getType() && ElasticsearchVersion::supportsMappingType()) {
+            $params['type'] = $options->getType();
         }
 
         $document = $this->client->get($params);
@@ -65,5 +72,10 @@ final class ItemProvider implements ProviderInterface
         }
 
         return $item;
+    }
+
+    private function convertDocumentMetadata(DocumentMetadata $documentMetadata): Options
+    {
+        return new Options($documentMetadata->getIndex(), $documentMetadata->getType());
     }
 }
--- a/src/Metadata/ApiResource.php
+++ b/src/Metadata/ApiResource.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Metadata;
 
 use ApiPlatform\Metadata\GraphQl\Operation as GraphQlOperation;
 use ApiPlatform\OpenApi\Model\Operation as OpenApiOperation;
+use ApiPlatform\State\OptionsInterface;
 
 /**
  * Resource metadata attribute.
@@ -147,7 +148,8 @@ class ApiResource
         protected ?array $graphQlOperations = null,
         $provider = null,
         $processor = null,
-        protected array $extraProperties = []
+        protected array $extraProperties = [],
+        protected ?OptionsInterface $stateOptions = null,
     ) {
         $this->operations = null === $operations ? null : new Operations($operations);
         $this->provider = $provider;
@@ -616,11 +618,17 @@ class ApiResource
         return $self;
     }
 
+    /**
+     * @deprecated this will be removed in v4
+     */
     public function getElasticsearch(): ?bool
     {
         return $this->elasticsearch;
     }
 
+    /**
+     * @deprecated this will be removed in v4
+     */
     public function withElasticsearch(bool $elasticsearch): self
     {
         $self = clone $this;
@@ -1027,6 +1035,19 @@ class ApiResource
     {
         $self = clone $this;
         $self->extraProperties = $extraProperties;
+
+        return $self;
+    }
+
+    public function getStateOptions(): ?OptionsInterface
+    {
+        return $this->stateOptions;
+    }
+
+    public function withStateOptions(?OptionsInterface $stateOptions): self
+    {
+        $self = clone $this;
+        $self->stateOptions = $stateOptions;
 
         return $self;
     }
--- a/src/Metadata/Extractor/XmlResourceExtractor.php
+++ b/src/Metadata/Extractor/XmlResourceExtractor.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Metadata\Extractor;
 
+use ApiPlatform\Elasticsearch\State\Options;
 use ApiPlatform\Exception\InvalidArgumentException;
 use ApiPlatform\Metadata\GetCollection;
 use ApiPlatform\Metadata\GraphQl\DeleteMutation;
@@ -96,6 +97,7 @@ final class XmlResourceExtractor extends AbstractResourceExtractor
             'paginationViaCursor' => $this->buildPaginationViaCursor($resource),
             'exceptionToStatus' => $this->buildExceptionToStatus($resource),
             'queryParameterValidationEnabled' => $this->phpize($resource, 'queryParameterValidationEnabled', 'bool'),
+            'stateOptions' => $this->buildStateOptions($resource),
         ]);
     }
 
@@ -449,5 +451,22 @@ final class XmlResourceExtractor extends AbstractResourceExtractor
         }
 
         return $data;
+    }
+
+    private function buildStateOptions(\SimpleXMLElement $resource): ?Options
+    {
+        $stateOptions = $resource->stateOptions ?? null;
+        if (!$stateOptions) {
+            return null;
+        }
+        $elasticsearchOptions = $stateOptions->elasticsearchOptions ?? null;
+        if ($elasticsearchOptions) {
+            return new Options(
+                isset($elasticsearchOptions['index']) ? (string) $elasticsearchOptions['index'] : null,
+                isset($elasticsearchOptions['type']) ? (string) $elasticsearchOptions['type'] : null,
+            );
+        }
+
+        return null;
     }
 }
--- a/src/Metadata/Extractor/schema/resources.xsd
+++ b/src/Metadata/Extractor/schema/resources.xsd
@@ -383,6 +383,19 @@
         </xsd:sequence>
     </xsd:complexType>
 
+    <xsd:complexType name="stateOptions">
+        <xsd:choice>
+            <xsd:element ref="elasticsearchOptions"/>
+        </xsd:choice>
+    </xsd:complexType>
+
+    <xsd:element name="elasticsearchOptions">
+        <xsd:complexType>
+            <xsd:attribute name="index" type="xsd:string"/>
+            <xsd:attribute name="type" type="xsd:string"/>
+        </xsd:complexType>
+    </xsd:element>
+
     <xsd:group name="base">
         <xsd:sequence>
             <xsd:element name="denormalizationContext" minOccurs="0" type="sequenceWithValues"/>
@@ -409,6 +422,7 @@
             <xsd:element name="options" minOccurs="0" type="sequenceWithValues"/>
             <xsd:element name="outputFormats" minOccurs="0" type="formats"/>
             <xsd:element name="paginationViaCursor" minOccurs="0" type="paginationViaCursor"/>
+            <xsd:element name="stateOptions" minOccurs="0" type="stateOptions"/>
             <xsd:element name="requirements" minOccurs="0" type="requirements"/>
             <xsd:element name="schemes" minOccurs="0" type="schemes"/>
             <xsd:element name="types" minOccurs="0" type="types"/>
--- a/src/Metadata/GetCollection.php
+++ b/src/Metadata/GetCollection.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Metadata;
 
 use ApiPlatform\OpenApi\Model\Operation as OpenApiOperation;
+use ApiPlatform\State\OptionsInterface;
 
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 final class GetCollection extends HttpOperation implements CollectionOperationInterface
@@ -96,9 +97,10 @@ final class GetCollection extends HttpOperation implements CollectionOperationIn
         $provider = null,
         $processor = null,
         array $extraProperties = [],
-        ?string $itemUriTemplate = null
+        ?string $itemUriTemplate = null,
+        ?OptionsInterface $stateOptions = null,
     ) {
-        parent::__construct(self::METHOD_GET, ...\func_get_args());
+        parent::__construct(self::METHOD_GET, $uriTemplate, $types, $formats, $inputFormats, $outputFormats, $uriVariables, $routePrefix, $routeName, $defaults, $requirements, $options, $stateless, $sunset, $acceptPatch, $status, $host, $schemes, $condition, $controller, $cacheHeaders, $hydraContext, $openapiContext, $openapi, $exceptionToStatus, $queryParameterValidationEnabled, $shortName, $class, $paginationEnabled, $paginationType, $paginationItemsPerPage, $paginationMaximumItemsPerPage, $paginationPartial, $paginationClientEnabled, $paginationClientItemsPerPage, $paginationClientPartial, $paginationFetchJoinCollection, $paginationUseOutputWalkers, $paginationViaCursor, $order, $description, $normalizationContext, $denormalizationContext, $security, $securityMessage, $securityPostDenormalize, $securityPostDenormalizeMessage, $securityPostValidation, $securityPostValidationMessage, $deprecationReason, $filters, $validationContext, $input, $output, $mercure, $messenger, $elasticsearch, $urlGenerationStrategy, $read, $deserialize, $validate, $write, $serialize, $fetchPartial, $forceEager, $priority, $name, $provider, $processor, $extraProperties, $stateOptions);
         $this->itemUriTemplate = $itemUriTemplate;
     }
 
--- a/src/Metadata/GraphQl/DeleteMutation.php
+++ b/src/Metadata/GraphQl/DeleteMutation.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Metadata\GraphQl;
 
 use ApiPlatform\Metadata\DeleteOperationInterface;
+use ApiPlatform\State\OptionsInterface;
 
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 class DeleteMutation extends Mutation implements DeleteOperationInterface
@@ -70,7 +71,8 @@ class DeleteMutation extends Mutation implements DeleteOperationInterface
         ?string $name = null,
         $provider = null,
         $processor = null,
-        array $extraProperties = []
+        array $extraProperties = [],
+        protected ?OptionsInterface $stateOptions = null,
     ) {
         parent::__construct(...\func_get_args());
     }
--- a/src/Metadata/GraphQl/Query.php
+++ b/src/Metadata/GraphQl/Query.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Metadata\GraphQl;
 
+use ApiPlatform\State\OptionsInterface;
+
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 class Query extends Operation
 {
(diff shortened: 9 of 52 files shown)


tests/Fixtures/Elasticsearch/Model/Book.php
```diff
@@ -0,0 +1,41 @@
...
+
+use ApiPlatform\Elasticsearch\Filter\MatchFilter;
+use ApiPlatform\Elasticsearch\Filter\OrderFilter;
+use ApiPlatform\Elasticsearch\Metadata\ElasticsearchDocument;
+use ApiPlatform\Metadata\ApiFilter;
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+#[ApiResource(normalizationContext: ['groups' => ['book:read']], persistenceMeans: new ElasticsearchDocument('book'))]
```
soyuka, line 24: I think that we need to discuss naming on these. We suggested `stateOptions` as it is broader. Why the `persistenceMeans` choice? 

About `ElasticsearchDocument` I'd suggest ElasticsearchStateOptions instead.

tests/Metadata/Extractor/ResourceMetadataCompatibilityTest.php
```diff
@@ -654,4 +663,23 @@ private function withGraphQlOperations(array $values, ?array $fixtures): array
...
+            throw new \InvalidArgumentException('Only one persistence means can be configured at a time.');
+        }
+
+        $configuration = reset($values);
+        switch (key($values)) {
+            case 'elasticsearchDocument':
+                return new ElasticsearchDocument($configuration['index'] ?? null, $configuration['type'] ?? null);
+        }
+
+        throw new \DomainException(sprintf('Unsupported "%s" persistence means.', key($values)));
```
soyuka, line 683: We don't use `DomainException`, `RuntimeException` instead?

src/Metadata/GraphQl/Query.php
```diff
@@ -13,66 +13,71 @@
...
+        protected ?int $priority = null,
+        protected ?string $name = null,
+        /** @var (callable(): mixed)|string|null $provider */
+        protected $provider = null,
+        /** @var (callable(): mixed)|string|null $processor */
+        protected $processor = null,
+        protected array $extraProperties = [],
         protected ?bool $nested = null,
+        protected ?PersistenceMeansInterface $persistenceMeans = null,
     ) {
```
soyuka, line 80: with this change are you sure that the parent constructor gets called? Not sure why we didn't used constructor promotion here but there may be a reason. I'd leave that untouched.

src/Metadata/GraphQl/DeleteMutation.php
```diff
@@ -70,7 +71,8 @@ public function __construct(
         ?string $name = null,
         $provider = null,
         $processor = null,
-        array $extraProperties = []
+        array $extraProperties = [],
+        protected ?PersistenceMeansInterface $persistenceMeans = null,
```
soyuka, line 75: consistency with other changes, why are you not using constructor promotion as you did in the Query object?

src/Metadata/Operation.php
```diff
@@ -90,8 +92,15 @@ public function __construct(
...
         $processor = null,
-        protected array $extraProperties = []
+        protected array $extraProperties = [],
+        protected ?PersistenceMeansInterface $persistenceMeans = null,
     ) {
+        if (null !== $this->elasticsearch) {
+            $solution = $this->elasticsearch
+                ? sprintf('Pass an instance of %s to $persistenceMeans instead', ElasticsearchDocument::class)
+                : 'You will have to remove it when upgrading to v4';
+            trigger_deprecation('api-platform/core', '3.1', sprintf('Setting "elasticsearch" is deprecated. %s', $solution));
```
soyuka, line 102: Please avoid deprecating here and prefer to put this logic only in the deprecated factory.

src/Metadata/ApiResource.php
```diff
@@ -147,8 +148,15 @@ public function __construct(
...
-        protected array $extraProperties = []
+        protected array $extraProperties = [],
+        protected ?PersistenceMeansInterface $persistenceMeans = null,
     ) {
+        if (null !== $this->elasticsearch) {
+            $solution = $this->elasticsearch
+                ? sprintf('Pass an instance of %s to $persistenceMeans instead', ElasticsearchDocument::class)
+                : 'You will have to remove it when upgrading to v4';
+            trigger_deprecation('api-platform/core', '3.1', sprintf('Setting "elasticsearch" is deprecated. %s', $solution));
+        }
```
soyuka, line 159: Please move this deprecation to the deprecated factory.

src/Elasticsearch/State/ItemProvider.php
```diff
@@ -26,7 +26,7 @@
 /**
  * Item provider for Elasticsearch.
  *
- * @experimental
+ * @deprecated
```
soyuka, line 29: I don't think that this provider should be deprecated, the deprecation was about Metadata Factories the provider will be just fine using the new option if available. .

src/Elasticsearch/State/CollectionProvider.php
```diff
@@ -26,7 +26,7 @@
 /**
  * Collection provider for Elasticsearch.
  *
- * @experimental
+ * @deprecated
```
soyuka, line 29: same as for the ItemProvider I think that this code was fine it just needs to use the new option.

src/Elasticsearch/Metadata/ElasticsearchDocument.php
```diff
@@ -0,0 +1,25 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Elasticsearch\Metadata;
+
+use ApiPlatform\Metadata\PersistenceMeansInterface;
+
+class ElasticsearchDocument implements PersistenceMeansInterface
```
soyuka, line 18: What about `ElasticsearchStateOptions` ? See comment below about naming I don't understand the interface name nor why this is a Document.

src/Elasticsearch/Metadata/ElasticsearchDocument.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    public function __construct(
+        public readonly ?string $index = null,
+        public readonly ?string $type = null,
```
soyuka, line 22: This should follow our Metadata architecture and be an Immutable class.

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,94 @@
...
+    public function create(string $resourceClass): ResourceMetadataCollection
+    {
+        $resourceMetadataCollection = $this->decorated->create($resourceClass);
+
+        foreach ($resourceMetadataCollection as $i => $resourceMetadata) {
+            if ($operations = $resourceMetadata->getOperations()) {
+                $resourceMetadata = $resourceMetadata->withOperations(self::configureOperations($operations));
+            }
+
+            // $graphqlOperations and $operations have not same type, we cannot combine function
```
soyuka, line 45:
```suggestion
```

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,94 @@
...
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Operations;
+use ApiPlatform\Metadata\PersistenceMeansInterface;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+
+/**
+ * @internal
+ */
+class ElasticsearchOperationProviderResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
```
soyuka, line 29:
```suggestion
final class ElasticsearchResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
```

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,94 @@
...
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\GraphQl\Operation as GraphQlOperation;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Operations;
+use ApiPlatform\Metadata\PersistenceMeansInterface;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+
+/**
+ * @internal
```
soyuka, line 27: not internal

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,94 @@
...
+    private static function configureOperation(Operation $operation): Operation
+    {
+        if (null !== $operation->getProvider()) {
+            return $operation;
+        }
+        if (!$operation->getPersistenceMeans() instanceof PersistenceMeansInterface) {
+            return $operation;
+        }
+        if (false === $operation->getElasticsearch()) {
+            throw new \LogicException('You cannot set $elasticsearch to false while configuring an elasticsearch document');
```
soyuka, line 89: this is legacy code and should not be called here

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchProviderResourceMetadataCollectionFactory.php
```diff
@@ -26,8 +26,11 @@
 
 final class ElasticsearchProviderResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
 {
-    public function __construct(private readonly Client $client, private readonly ResourceMetadataCollectionFactoryInterface $decorated)
+    public function __construct(private readonly Client $client, private readonly ResourceMetadataCollectionFactoryInterface $decorated, private readonly bool $triggerDeprecation = true)
```
soyuka, line 29: I don't understand how this works, shouldn't we deprecate inside the `create` method instead? And only if the `elasticsearch` flag is used?

src/Metadata/Extractor/XmlResourceExtractor.php
```diff
@@ -450,4 +453,23 @@ private function buildGraphQlOperations(\SimpleXMLElement $resource, array $root
...
+            throw new \DomainException('There can only be one configured persistence means. "resources.xsd" needs to be fixed.');
+        }
+        $elasticsearchDocument = $persistenceMeans->elasticsearchDocument ?? null;
+        if (isset($elasticsearchDocument)) {
+            return new ElasticsearchDocument(
+                isset($elasticsearchDocument['index']) ? (string) $elasticsearchDocument['index'] : null,
+                isset($elasticsearchDocument['type']) ? (string) $elasticsearchDocument['type'] : null,
+            );
+        }
+        throw new \DomainException(sprintf('Unsupported "%s" persistence means. "resources.xsd" needs to be fixed.', $persistenceMeans->children()?->getName()));
```
soyuka, line 473: we're not using domain exceptions probably `RuntimeException` is a better choice

tests/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactoryTest.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    private function getResourceMetadataCollectionFactory(string $resourceClass, Operation $operation): Stub
+    {
+        if ($operation instanceof GraphQlOperation) {
+            $resource = new ApiResource(graphQlOperations: [$operation->getName() => $operation]);
+        } else {
+            $resource = new ApiResource(operations: [$operation->getName() => $operation]);
+        }
+        $metadata = new ResourceMetadataCollection($resourceClass, [$resource]);
+        $decorated = $this->createStub(ResourceMetadataCollectionFactoryInterface::class);
```
soyuka, line 91: we're using `prophecy` in the rest of the code this seems inconsistent with the code base.

soyuka (changes requested):
because of a github ui bug I kinda read from bottom to top so please read all the comments. 

My main issue is about the name of the interface, option and elasticsearch options class.

src/Elasticsearch/State/ElasticsearchItemProvider.php
```diff
@@ -0,0 +1,62 @@
...
+use ApiPlatform\Util\Inflector;
+use Elasticsearch\Client;
+use Elasticsearch\Common\Exceptions\Missing404Exception;
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+/**
+ * @internal
+ */
+final class ElasticsearchItemProvider implements ProviderInterface
```
dunglas, line 30: Same here for the name

src/Elasticsearch/State/ElasticsearchItemProvider.php
```diff
@@ -0,0 +1,62 @@
...
+            'index' => $documentConfiguration->index ?? Inflector::tableize($operation->getShortName()),
+            'id' => (string) reset($uriVariables),
+        ];
+        if (ElasticsearchVersion::supportsMappingType()) {
+            $params['type'] = $documentConfiguration->type;
+        }
+
+        try {
+            $document = $this->client->get($params);
+        } catch (Missing404Exception) {
```
dunglas, line 56: Is there a way to avoid using a try/catch here? Handling exceptions on the hot path hurts performance.

src/Metadata/Extractor/XmlResourceExtractor.php
```diff
@@ -450,4 +453,23 @@ private function buildGraphQlOperations(\SimpleXMLElement $resource, array $root
 
         return $data;
     }
+
+    private function buildPersistenceMeans(\SimpleXMLElement $resource): ?PersistenceMeansInterface
+    {
+        $persistenceMeans = $resource->persistenceMeans ?? null;
+        if (!isset($persistenceMeans)) {
```
dunglas, line 460:
```suggestion
        if ($persistenceMeans ?? false) {
```

src/Metadata/Extractor/XmlResourceExtractor.php
```diff
@@ -450,4 +453,23 @@ private function buildGraphQlOperations(\SimpleXMLElement $resource, array $root
...
+            return null;
+        }
+        if (1 !== \count($persistenceMeans)) {
+            throw new \DomainException('There can only be one configured persistence means. "resources.xsd" needs to be fixed.');
+        }
+        $elasticsearchDocument = $persistenceMeans->elasticsearchDocument ?? null;
+        if (isset($elasticsearchDocument)) {
```
dunglas, line 467:
```suggestion
        if ($elasticsearchDocument ?? false) {
```

dunglas: Use `??` instead of `isset()` everywhere it's possible.

src/Metadata/PersistenceMeansInterface.php
```diff
@@ -0,0 +1,22 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata;
+
+/**
+ * Marker interface to be implemented by value objects encapsulating data needed by {@see ProviderInterface providers}
+ * to retrieve resources from a persistence layer.
+ */
+interface PersistenceMeansInterface
```
dunglas, line 20: IMHO we must not use `persistence` in the name, as we tried to avoid it for v3 (the processor can persist, but also send mails, do a computation...).

Maybe:

```suggestion
namespace ApiPlatform\State;

/**
 * Marker interface to be implemented by value objects encapsulating data needed by {@see ProcessorInterface processors} and {@see ProviderInterface providers}.
 */
interface OptionsInterface
```

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -418,6 +419,7 @@ private function addElasticsearchSection(ArrayNodeDefinition $rootNode): void
                             ->prototype('scalar')->end()
                         ->end()
                         ->arrayNode('mapping')
+                            ->setDeprecated('api-platform/core', '3.1', sprintf('The "%%node%%" option is deprecated. Configure an %s as $persistenceMeans.', ElasticsearchDocument::class))
```
dunglas, line 422: Error? It's the `mapping` option here.

dunglas (changes requested):

tests/Metadata/Extractor/ResourceMetadataCompatibilityTest.php
dunglas: `LogicException` is better IMHO "This kind of exception should lead directly to a fix in your code."

src/Elasticsearch/State/ElasticsearchItemProvider.php
soyuka: no there's not sadly this is because of the es-package

src/Metadata/GraphQl/Query.php
soyuka: I'm unsure, I wish we could have something better to share these constructor options but it is a hard problem. unless if we accept that arguments position can change and that specific options are last?

src/Metadata/PersistenceMeansInterface.php
soyuka: I really like the name for state options

src/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka: The mapping is also deprecated as in fact it's the same thing as what's used in the resource class option...

soyuka:
Let's use this name for options:

```
ApiPlatform\State\OptionsInterface;
```

For the attribute property (ApiResource, Get etc.)
```
stateOptions: OptionsInterface
```

for ES:

```
Elasticsearch\State\Options
```

tests/Metadata/Extractor/ResourceMetadataCompatibilityTest.php
MatTheCat: `DomainException`: “Exception thrown if a value does not adhere to a defined valid data domain.”

This is the case here (and it extends `LogicException`).

src/Metadata/PersistenceMeansInterface.php
MatTheCat: The goal of the interface is described in its PHPDoc. It makes no sense for a provider which sends mail to use it. Is only purpose is about persistence and its name should reflect it.

src/Metadata/Operation.php
MatTheCat: Deprecating a method parameter seems logical to do in the same method. Why are you suggesting otherwise?

src/Metadata/ApiResource.php
MatTheCat: https://github.com/api-platform/core/pull/5297#discussion_r1060352779

src/Elasticsearch/State/ItemProvider.php
MatTheCat: Nope it won’t, because it is using deprecated metadata factories.

src/Elasticsearch/State/CollectionProvider.php
MatTheCat: https://github.com/api-platform/core/pull/5297#discussion_r1060353458

src/Elasticsearch/Metadata/ElasticsearchDocument.php
MatTheCat: The resource is persisted by the means of an elasticsearch document.

src/Elasticsearch/Metadata/ElasticsearchDocument.php
MatTheCat: This is an immutable class.

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
MatTheCat: Do you think this exception should be thrown in `ApiResource` and `Operation`’s constructors?

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchProviderResourceMetadataCollectionFactory.php
MatTheCat: This service will always be instantiated, even if the user did not set `$elasticsearch`. In that case we should not log a deprecation.

src/Metadata/Extractor/XmlResourceExtractor.php
MatTheCat: https://github.com/api-platform/core/pull/5297#discussion_r1060349412

src/Metadata/Extractor/XmlResourceExtractor.php
MatTheCat: Is this for performance reasons?

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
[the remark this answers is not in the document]
MatTheCat: Yes the previous condition is wrong :grimacing:

src/Metadata/Extractor/XmlResourceExtractor.php
davy-beauzil: When I apply this change, I have a phpstorm bug with this message : `[EA] Resolved operands types are not complimentary, while they should be ([\SimpleXMLElement] vs [bool]).`

src/Elasticsearch/State/ElasticsearchItemProvider.php
MatTheCat: Found https://github.com/elastic/elasticsearch-php/blob/7.x/docs/per-request-configuration.asciidoc#ignoring-exceptions. Investigating.

tests/Metadata/Extractor/ResourceMetadataCompatibilityTest.php
dunglas: We don't use it in API Platform and I'm Symfony. We reserve it for the app domain.

src/Elasticsearch/State/ItemProvider.php
dunglas: You can accept both as parameters and deprecate passing the old ones.

src/Metadata/Operation.php
soyuka: tldr: move the deprecation to ElasticsearchProviderResourceMetadataCollectionFactory

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
soyuka: only in the MetadataFactory that uses `getElasticsearch`

src/Metadata/Extractor/schema/resources.xsd
```diff
@@ -409,6 +422,7 @@
             <xsd:element name="options" minOccurs="0" type="sequenceWithValues"/>
             <xsd:element name="outputFormats" minOccurs="0" type="formats"/>
             <xsd:element name="paginationViaCursor" minOccurs="0" type="paginationViaCursor"/>
+            <xsd:element name="persistenceMeans" minOccurs="0" type="persistenceMeans"/>
```
soyuka, line 425:
```suggestion
            <xsd:element name="persistenceMeans" minOccurs="0" maxOccurs="1" type="persistenceMeans"/>
```

soyuka: is the default value so not needed

src/Metadata/Extractor/XmlResourceExtractor.php
dunglas: You can either cast the expression with `(bool)` or just ignore this message. Using `null` instead of `false` can maybe fix the message (and is probably better).

tests/Metadata/Extractor/ResourceMetadataCompatibilityTest.php
dunglas: We don't use `DomainException` in the API Platform code base (its definition is not very clear, see  https://stackoverflow.com/a/1103021/1352334 and the related PHP bug). I suggest keeping `LogicException` for consistency.

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
davy-beauzil: The name of the class already exists

src/Metadata/Extractor/XmlResourceExtractor.php
MatTheCat: `isset` was a remnant from before `$persistenceMeans` assignation :grimacing: 

```php
if (!$persistenceMeans) {
```

should be enough.

src/Metadata/Extractor/XmlResourceExtractor.php
MatTheCat: `isset` was in fact not needed (see https://github.com/api-platform/core/pull/5297#discussion_r1060466524)

tests/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactoryTest.php
MatTheCat: PHPunit’s API is arguably better than Prophecy’s (mocks and stubs are distinguished, and they are recognized as instance of the doubled class eg.) and it does not require an additional dependency.

Do we really want to stick with Prophecy?

src/Metadata/PersistenceMeansInterface.php
dunglas: Why? API Platform doesn't care about what the processor does (except in the bridges obviously).

src/Elasticsearch/Metadata/ElasticsearchDocument.php
```diff
@@ -0,0 +1,35 @@
...
+    ) {
+    }
+
+    public function withIndex(?string $index): self
+    {
+        return new self($index, $this->type);
```
soyuka, line 28: Check how we do in the metadata:

```
    public function getOperations(): ?Operations
    {
        return $this->operations;
    }

    public function withOperations(Operations $operations): self
    {
        $self = clone $this;
        $self->operations = $operations;

        return $self;
    }
``` 

It's easier to evolve if we add fields to the constructor.

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchProviderResourceMetadataCollectionFactory.php
```diff
@@ -42,6 +46,15 @@ public function create(string $resourceClass): ResourceMetadataCollection
...
             if ($operations) {
                 foreach ($resourceMetadata->getOperations() as $operationName => $operation) {
+                    if (null !== ($elasticsearch = $operation->getElasticsearch())) {
+                        $solution = $elasticsearch
+                            ? sprintf('Pass an instance of %s to $persistenceMeans instead', ElasticsearchDocument::class)
+                            : 'You will have to remove it when upgrading to v4';
+                        trigger_deprecation('api-platform/core', '3.1', sprintf('Setting "elasticsearch" in Operation is deprecated. %s', $solution));
+                    }
+                    if (false === $elasticsearch && $operation->getPersistenceMeans() instanceof ElasticsearchDocument) {
+                        throw new \LogicException('You cannot set $elasticsearch to false while configuring an elasticsearch document');
```
soyuka, line 56: IMO the previous deprecation is fine, we can remove this exception and make it work anyways.

src/Metadata/GetCollection.php
```diff
@@ -96,9 +96,10 @@ public function __construct(
...
+        parent::__construct(self::METHOD_GET, $uriTemplate, $types, $formats, $inputFormats, $outputFormats, $uriVariables, $routePrefix, $routeName, $defaults, $requirements, $options, $stateless, $sunset, $acceptPatch, $status, $host, $schemes, $condition, $controller, $cacheHeaders, $hydraContext, $openapiContext, $openapi, $exceptionToStatus, $queryParameterValidationEnabled, $shortName, $class, $paginationEnabled, $paginationType, $paginationItemsPerPage, $paginationMaximumItemsPerPage, $paginationPartial, $paginationClientEnabled, $paginationClientItemsPerPage, $paginationClientPartial, $paginationFetchJoinCollection, $paginationUseOutputWalkers, $paginationViaCursor, $order, $description, $normalizationContext, $denormalizationContext, $security, $securityMessage, $securityPostDenormalize, $securityPostDenormalizeMessage, $securityPostValidation, $securityPostValidationMessage, $deprecationReason, $filters, $validationContext, $input, $output, $mercure, $messenger, $elasticsearch, $urlGenerationStrategy, $read, $deserialize, $validate, $write, $serialize, $fetchPartial, $forceEager, $priority, $name, $provider, $processor, $extraProperties, $persistenceMeans);
```
soyuka, line 102: the previous code is probably still valid here

src/Metadata/GraphQl/Query.php
soyuka: final decision on this is to keep the previous code, we will try to find a better way

src/Metadata/HttpOperation.php
```diff
@@ -146,52 +146,10 @@ public function __construct(
...
-        $this->deserialize = $deserialize;
-        $this->validate = $validate;
-        $this->write = $write;
-        $this->serialize = $serialize;
-        $this->priority = $priority;
-        $this->name = $name;
-        $this->provider = $provider;
-        $this->processor = $processor;
-        $this->extraProperties = $extraProperties;
+        parent::__construct(...\array_slice(\func_get_args(), 26));
```
soyuka, line 152: not sure we want to change this

tests/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactoryTest.php
soyuka: agreed but yes for consistency accross the codebase

src/Symfony/Bundle/Resources/config/elasticsearch.xml
```diff
@@ -102,14 +102,21 @@
...
             <argument type="service" id="api_platform.elasticsearch.client" />
             <argument type="service" id="api_platform.elasticsearch.metadata.document.metadata_factory" />
             <argument type="service" id="serializer" />
 
             <tag name="api_platform.state_provider" priority="-100" />
         </service>
-        <service id="api_platform.state.item_provider" alias="ApiPlatform\Elasticsearch\State\ItemProvider" />
+
+
+        <service id="ApiPlatform\Elasticsearch\State\ElasticsearchItemProvider" public="false">
```
soyuka, line 114: remove this declaration `ItemProvider` is enough

src/Symfony/Bundle/Resources/config/elasticsearch.xml
```diff
@@ -102,14 +102,21 @@
         </service>
         <service id="ApiPlatform\Elasticsearch\Filter\OrderFilter" alias="api_platform.elasticsearch.order_filter" />
 
-        <service id="ApiPlatform\Elasticsearch\State\ItemProvider" class="ApiPlatform\Elasticsearch\State\ItemProvider" public="false">
+        <service id="ApiPlatform\Elasticsearch\State\ItemProvider" public="false">
```
soyuka, line 105: can you do:

```
<service id="api_platform.elasticsearch.state.item_provider" class="ApiPlatform\Elasticsearch\State\ItemProvider"> etc </service>
```
Then alias with the class name?

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,85 @@
...
+
+        return $operations;
+    }
+
+    private function configureOperation(Operation $operation): Operation
+    {
+        if (null !== $operation->getProvider()) {
+            return $operation;
+        }
+        if (!$operation->getStateOptions() instanceof OptionsInterface) {
```
soyuka, line 79: I don't understand this, if we set a state options we don't set the provider?

src/Elasticsearch/State/CollectionProvider.php
```diff
@@ -60,13 +59,20 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
         $limit = $body['size'] ??= $this->pagination->getLimit($operation, $context);
         $offset = $body['from'] ??= $this->pagination->getOffset($operation, $context);
 
+        $options = $operation->getStateOptions();
+        if (null === $options) {
+            $options = $this->convertDocumentMetadata($this->documentMetadataFactory->create($resourceClass));
+        } elseif (!$options instanceof OptionsInterface) {
+            throw new \LogicException(sprintf('Operation‘s state options must be null or an instance of %s, but got %s', OptionsInterface::class, \get_class($options)));
```
soyuka, line 66: Can we remove the `elseif` ? We can move this exception in the MetadataFactory if we really want to catch this. 

I'd like to suggest:

```
$options = $operation->getStateOptions() ?? new ElasticsearchStateOptions(index: Inflector::tableize($operation->getShortName()));

// TODO: remove in 4.x
if ($operation->getElasticsearch() && !$operation->getStateOptions()) {
  $options = $this->convertDocumentMetadata($this->documentMetadataFactory->create($resourceClass));
}
```

src/Elasticsearch/State/CollectionProvider.php
```diff
@@ -60,13 +59,20 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
...
+        }
+
         $params = [
-            'index' => $documentMetadata->getIndex(),
+            'index' => $options->getIndex() ?? Inflector::tableize($operation->getShortName()),
```
soyuka, line 70: see above

src/Elasticsearch/State/ItemProvider.php
```diff
@@ -42,16 +41,22 @@ public function __construct(private readonly Client $client, private readonly Do
     public function provide(Operation $operation, array $uriVariables = [], array $context = []): ?object
     {
         $resourceClass = $operation->getClass();
-        $documentMetadata = $this->documentMetadataFactory->create($resourceClass);
+        $options = $operation->getStateOptions();
+
+        if (null === $options) {
+            $options = $this->convertDocumentMetadata($this->documentMetadataFactory->create($resourceClass));
+        } elseif (!$options instanceof OptionsInterface) {
+            throw new \LogicException(sprintf('Operation‘s options must be null or an instance of %s, but got %s', OptionsInterface::class, \get_class($options)));
```
soyuka, line 49: same as for the CollectionProvider

src/Metadata/GraphQl/Query.php
```diff
@@ -13,66 +13,71 @@
...
+        protected $provider = null,
+        /** @var (callable(): mixed)|string|null $processor */
+        protected $processor = null,
+        protected array $extraProperties = [],
         protected ?bool $nested = null,
+        protected ?OptionsInterface $stateOptions = null,
     ) {
-        parent::__construct(...\func_get_args());
         $this->name = $name ?: 'item_query';
     }
```
soyuka, line 82: can we roll back on the `protected` thing?

src/Metadata/HttpOperation.php
```diff
@@ -146,52 +147,10 @@ public function __construct(
...
     }
```
soyuka, line 155: please rolll back this change

soyuka (changes requested):

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
davy-beauzil: In this code, we set an elasticsearch provider only if `stateOptions` is an instance of `ApiPlatform/Elasticsearch/State/OptionsInterface`, isn't that what we want?

src/Elasticsearch/State/CollectionProvider.php
davy-beauzil: I think we should keep this because `$index` in `ApiPlatform/Elasticsearch/State/Options` can be null. But is it really necessary to keep `$index` nullable ?

src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchOperationProviderResourceMetadataCollectionFactory.php
soyuka: okay yes and this is what I find weird about this implementation instead of a specific Elasticsearch Operation. What is more explicit for the user:

```
use ApiPlatform\Elasticsearch\State\Options;

#[GetCollection(stateOptions: new Options(index: 'test'))
class MyResource {}
```

or


```
use ApiPlatform/Elasticsearch\Metadata\GetCollection;

#[ApiResource(index: 'test')
class MyResource {}
```

src/Elasticsearch/State/CollectionProvider.php
davy-beauzil: I think we should keep the check on instance of  `$options` because at this step of the code, we know only that `$options` is an instance of `ApiPlatform/State/OptionsInterface`, which does not necessarily have `getIndex()`.

soyuka: I understand the concern, alright what about:


```
$options = $operation->getStateOptions() instanceof Options ? $operation->getStateOptions() : new ElasticsearchStateOptions(index: Inflector::tableize($operation->getShortName()));

// TODO: remove in 4.x
if ($operation->getElasticsearch() && !$operation->getStateOptions()) {
  $options = $this->convertDocumentMetadata($this->documentMetadataFactory->create($resourceClass));
}
```

soyuka:
Thanks @davy-beauzil @MatTheCat !

davy-beauzil:
I'm happy to help!

Issue #5269, opened 2022-12-13
Links require same ID type


nesl247:
**API Platform version(s) affected**: 3.0.7

**Description**  
`Link`s require the same property type when following the example here: https://api-platform.com/docs/core/subresources/#company-employees 

If the `id` type is a `string` on `Company`, it will fail if the `id` is type `UUIDv4` for example on `Employee`. The issue is caused because the type is retrieved from the last identifier in the URL, and not for the particular identifier.

**How to reproduce**  
See the link above.

**Possible Solution**  
```
--- /dev/null
+++ ../src/Doctrine/Orm/State/LinksHandlerTrait.php
@@ -118,6 +118,9 @@
             $queryBuilder->join("{$previousAlias}.{$link->getToProperty()}", $joinAlias);
 
             foreach ($identifierProperties as $identifierProperty) {
+                $associationTargetClass = $doctrineClassMetadata->getAssociationMapping($link->getToProperty())['targetEntity'];
+                $doctrineClassMetadata = $manager->getClassMetadata($associationTargetClass);
+
                 $placeholder = $queryNameGenerator->generateParameterName($identifierProperty);
                 $queryBuilder->andWhere("$joinAlias.$identifierProperty = :$placeholder");
                 $queryBuilder->setParameter($placeholder, $this->getIdentifierValue($identifiers, $hasCompositeIdentifiers ? $identifierProperty : null), $doctrineClassMetadata->getTypeOfField($identifierProperty));

```
**Additional Context**  
I didn't submit this as a PR as I'm not sure if there is more to the code that maybe we need to change, but this did solve our issue.

soyuka:
It's a good idea, you'd need to test if `$link->getToProperty()` is an association before making an assumption on that matter. Please open a patch on 3.x and you can add a functional test :).

nesl247:
This isn't stale, I have not had time to work on the related PR.

PR #5304, opened 2023-01-03
fix: links require same id type

nesl247:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | #5269 
| License       | MIT
| Doc PR        | n/a

This PR fixes the issue where if you have multiple links, all of the ID properties must be of the same type (e.g. one is `UUID`, another is `string`).

TODO:
- [ ] Create a functional test case

### DIFF
--- a/src/Doctrine/Orm/State/LinksHandlerTrait.php
+++ b/src/Doctrine/Orm/State/LinksHandlerTrait.php
@@ -120,7 +120,14 @@ trait LinksHandlerTrait
             foreach ($identifierProperties as $identifierProperty) {
                 $placeholder = $queryNameGenerator->generateParameterName($identifierProperty);
                 $queryBuilder->andWhere("$joinAlias.$identifierProperty = :$placeholder");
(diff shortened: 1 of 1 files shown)


src/Doctrine/Orm/State/LinksHandlerTrait.php
```diff
@@ -120,7 +120,14 @@ private function handleLinks(QueryBuilder $queryBuilder, array $identifiers, Que
             foreach ($identifierProperties as $identifierProperty) {
                 $placeholder = $queryNameGenerator->generateParameterName($identifierProperty);
                 $queryBuilder->andWhere("$joinAlias.$identifierProperty = :$placeholder");
-                $queryBuilder->setParameter($placeholder, $this->getIdentifierValue($identifiers, $hasCompositeIdentifiers ? $identifierProperty : null), $doctrineClassMetadata->getTypeOfField($identifierProperty));
+
+                if ($doctrineClassMetadata->hasAssociation($link->getToProperty())) {
```
soyuka, line 124: I think that we really don't want to go there especially here. Please check the `IdentifiersExtractor` class instead? Also can you provide a failing functional test case, I'm pretty sure I can cover that case in a jiffy.

Issue #5310, opened 2023-01-06
Querying a child relation of the same entity type not possible in GraphQL


NicoHaase:
**API Platform version(s) affected**: 3.0.8

**Description**  
In my application, I use a parent/child relation with entities of the same type, like the following:

```php
<?php

declare(strict_types=1);

namespace App\Entity\QualityManual;

use ApiPlatform\Metadata\GraphQl\QueryCollection;
use ApiPlatform\Metadata\GraphQl\Query;
use ApiPlatform\Metadata\ApiResource;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
use Symfony\Component\Serializer\Annotation\Groups;

#[ApiResource(
    graphQlOperations: [
        new Query(name: 'item_query'),
        new QueryCollection(name: 'collection_query'),
    ]
)]
#[ORM\Entity]
class Chapter
{
    #[ORM\Id]
    #[ORM\Column(type: 'guid', unique: true)]
    #[ORM\GeneratedValue(strategy: 'CUSTOM')]
    #[ORM\CustomIdGenerator(class: UuidGenerator::class)]
    private ?string $id = null;

    #[ORM\OneToMany(mappedBy: 'parent', targetEntity: Chapter::class, cascade: ['remove'])]
    #[ORM\OrderBy(['position' => 'ASC'])]
    private Collection $children;

    public function __construct()
    {
        $this->children  = new ArrayCollection();
    }

    public function getId(): ?string
    {
        return $this->id;
    }
}
```

When I try to select a single `Chapter` entity and the number of it's children, the result is not built properly. As far as I can debug this, it's a problem of these lines:

> https://github.com/api-platform/core/blob/e2753faabe77eeb5e58cb76db2d4ac87a86cabab/src/GraphQl/Type/FieldsBuilder.php#L265-L268

Due to the last condition, the `$resourceOperation` is not properly set to a collection operation, but to an item operaion. This results in an error in

> https://github.com/api-platform/core/blob/e2753faabe77eeb5e58cb76db2d4ac87a86cabab/src/GraphQl/Resolver/Factory/CollectionResolverFactory.php#L59-L62

as the resulting collection is **not** a collection, but `null` (even if there **are** entities matching this criteria).

**How to reproduce**  
- use the entity definition as above
[...]

PR #5314, opened 2023-01-09
fix(graphql): use depth for nested resource class operation

NicoHaase:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | #5310
| License       | MIT

Reproducer for the bug reported in #5310

### DIFF
--- a/features/graphql/query.feature
+++ b/features/graphql/query.feature
@@ -68,6 +68,35 @@ Feature: GraphQL query support
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[0].node.name" should be equal to "RelatedOneToManyDummy12"
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[2].node.name" should be equal to "RelatedOneToManyDummy32"
 
+  @createSchema @!mongodb
+  Scenario: Retrieve an item with child relation to the same resource
+    Given there are tree dummies
+    When I send the following GraphQL request:
+    """
+    {
+      treeDummies {
+        edges {
+          node {
+            id
+            children {
+              totalCount
+            }
+          }
+        }
+      }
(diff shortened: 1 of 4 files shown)


alanpoulain:
Hello, thanks for this PR!
Could you use a dedicated resource though?

NicoHaase:
> Hello, thanks for this PR! Could you use a dedicated resource though?

Sorry - this still needs some work, as I'm currently not able to reproduce the problem in an isolated test fase

NicoHaase:
Let's see whether the CI triggers the same problem I could reproduce locally now. Also, I've found how my previous version of this reproducer differs from this one: by reordering the list of graphqloperations on the TreeDummy, I was able to fix the bug. Having QueryCollection first and then Query, no bug is thrown. This also resolves the problem in my own application.

But that looks pretty strange to me. Why should the order of these operations influence how a field is resolved?

alanpoulain:
Hello @NicoHaase, I've made the fix I think is appropriate. What do you think of it?
@develth, does it fix your issue too?

develth:
> Hello @NicoHaase, I've made the fix I think is appropriate. What do you think of it? @develth, does it fix your issue too?

will create a diff patch & test it - thx!

develth:
Your fix was successful for me @alanpoulain !

soyuka (approved):

NicoHaase:
I haven't checked this fix yet, but when the test is no longer failing, I would assume it works :) But what makes me curious is why ordering the operations makes a difference. This is not addressed in this fix

alanpoulain:
It doesn't make a difference anymore since the nested operation will now always be selected depending on the type (collection or not).
Before it was not the case and the first operation on the resource was used.

PR #5265, opened 2022-12-12
fix(graphql): dont add graphql operations when disabled

dannyvw:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | #... 
| License       | MIT

PR #5315, opened 2023-01-09
fix(graphql): pass graphql enabled flag

aazsamir:
| Q             | A
| ------------- | ---
| Branch?       | 3.0 
| Tickets       | #... 
| License       | MIT

In that PR https://github.com/api-platform/core/pull/5265 was added `$graphQlEnabled` flag, but it hasn't been considered in `metadata/yaml.xml` config. It results in always false for that flag (as default). This PR fix that.


soyuka:
thanks!

PR #5324, opened 2023-01-11
chore(deprecation): Only use ValueResolverInterface if it exists

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

As mentioned in #5289 , only use `ValueResolverInterface` if it exists, use the old one otherwise

### DIFF
--- a/src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
+++ b/src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
@@ -18,63 +18,121 @@ use ApiPlatform\Serializer\SerializerContextBuilderInterface;
 use ApiPlatform\Util\OperationRequestInitiatorTrait;
 use ApiPlatform\Util\RequestAttributesExtractor;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
 use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
 use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 
-final class PayloadArgumentResolver implements ValueResolverInterface
-{
-    use OperationRequestInitiatorTrait;
-
-    public function __construct(
-        ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory,
-        private readonly SerializerContextBuilderInterface $serializationContextBuilder
-    ) {
-        $this->resourceMetadataCollectionFactory = $resourceMetadataCollectionFactory;
-    }
-
-    public function supports(Request $request, ArgumentMetadata $argument): bool
+if(class_exists(ValueResolverInterface::class)) {
+    final class PayloadArgumentResolver implements ValueResolverInterface
     {
-        if ($argument->isVariadic()) {
-            return false;
+        use OperationRequestInitiatorTrait;
+
+        public function __construct(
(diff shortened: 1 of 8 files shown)


KDederichs:
Hmm it still seems to load the old one, I'll look into why more tomorrow

src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
```diff
@@ -18,63 +18,120 @@
...
-
-    public function __construct(
-        ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory,
-        private readonly SerializerContextBuilderInterface $serializationContextBuilder
-    ) {
-        $this->resourceMetadataCollectionFactory = $resourceMetadataCollectionFactory;
-    }
-
-    public function supports(Request $request, ArgumentMetadata $argument): bool
+if (class_exists(ValueResolverInterface::class)) {
```
franmomu, line 25:
```suggestion
if (interface_exists(ValueResolverInterface::class)) {
```

KDederichs: Ah... makes sense I guess

src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php
```diff
@@ -18,63 +18,120 @@
...
-
-    public function __construct(
-        ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory,
-        private readonly SerializerContextBuilderInterface $serializationContextBuilder
-    ) {
-        $this->resourceMetadataCollectionFactory = $resourceMetadataCollectionFactory;
-    }
-
-    public function supports(Request $request, ArgumentMetadata $argument): bool
+if (interface_exists(ValueResolverInterface::class)) {
```
soyuka, line 25: I'm not sure we should declare 2 classes in the same file

KDederichs: Yeah I wasn't sure either, I've seen both this and a more split approach before so I just went with this first. If you want I can try and separate it a bit

franmomu: if it's the same code for both (only change of interface), a new interface could be created like:

```php
if (interface_exists(ValueResolverInterface::class)) {
    /** @internal */
    interface CompatibleValueResolverInterface extends ValueResolverInterface
    {
    }
} else {
    /** @internal */
    interface CompatibleValueResolverInterface extends ArgumentValueResolverInterface
    {
    }
}
```
and implement `CompatibleValueResolverInterface` instead of `ValueResolverInterface` or `ArgumentValueResolverInterface`

KDederichs: It's almost the same yeah, the thing that changes can be done with a simple check for the interface as well though

KDederichs:
Hmm looks like there's something else broken in the lowest tests (unrelated to this)
I'll try to get rid of that I guess

soyuka:
whoops the rebase was wrong I did that at https://github.com/api-platform/core/pull/5334 

When rebasing against a branch upstream:

```
git fetch upstream
git rebase upstream/3.0 
```

PR #5327, opened 2023-01-12
Fix `PayloadArgumentResolver::resolve` return type

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

`ArgumentValueResolverInterface::resolve` return type always have been `iterable`.

There is no need to make it more specific, especially because we support PHP ≥ 7.1 and full covariance support was added with PHP 7.4.


soyuka:
thanks @MatTheCat !

PR #5329, opened 2023-01-12
chore(ci): add test against lowest dependencies

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Since the current tests didn't catch the issue fixed with #5324 I thought I'd be good to add some tests that run against the lowest possible dependencies.


KDederichs:
Removed `willdurand/negotiation:2.x` since it's incompatible with PHP 8, `doctrine:3.0` has issues with union types and `symfony/service-contracts:<3` breaks with some annotations

soyuka:
nice, we had that not sure how it got removed, you can use `ci: test against lowest dependencies` as commit message.

KDederichs:
@soyuka changed it

soyuka:
I merged like this and we try to fix lowest in your other PR? 

Thanks!

PR #5332, opened 2023-01-12
Revert "chore(deprecation): Only use ValueResolverInterface if it exists"

soyuka:
Reverts api-platform/core#5324


soyuka:
There was a commit introduced by mistake, let me fix that manually in #5324.

PR #5333, opened 2023-01-12
fix(graphql): usable YAML/XML configuration

alanpoulain:
Fixes https://github.com/api-platform/api-platform/issues/2341.

This PR **could** be considered as a BC break, but since the current GraphQL configuration was not really usable anyway, I think it is not.
For instance, the **only** way to make the current YAML configuration work is to write:
```yaml
Namespace\To\Entity:
  graphQlOperations:
    queries:
      ApiPlatform\Metadata\GraphQl\Query:
        class: Namespace\To\Entity
      ApiPlatform\Metadata\GraphQl\QueryCollection:
        collection: true
        class: Namespace\To\Entity
```
Which is very weird and a lot of users don't know that it can work like this anyway since it is **not documented**.
Now it works like REST:
```yaml
Namespace\To\Entity:
  graphQlOperations:
      ApiPlatform\Metadata\GraphQl\Query: ~
      ApiPlatform\Metadata\GraphQl\QueryCollection: ~
```
Or:
```yaml
Namespace\To\Entity:
  graphQlOperations:
      - class: ApiPlatform\Metadata\GraphQl\Query
      - class: ApiPlatform\Metadata\GraphQl\QueryCollection
```

In XML it is done like this:
```xml
<resources xmlns="https://api-platform.com/schema/metadata/resources-3.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://api-platform.com/schema/metadata/resources-3.0            https://api-platform.com/schema/metadata/resources-3.0.xsd">
<resource class="Namespace\To\Entity">
	<graphQlOperations>
		<graphQlOperation class="ApiPlatform\Metadata\GraphQl\Query"/>
		<graphQlOperation class="ApiPlatform\Metadata\GraphQl\QueryCollection"/>
	</graphQlOperations>
</resource>
</resources>
```

Also this PR moves `paginationViaCursor` to REST only since it is not related to GraphQL (the XML and YAML extractor were correct to add it in "extended").


src/Metadata/HttpOperation.php
```diff
@@ -112,7 +113,6 @@ public function __construct(
         ?bool $paginationClientPartial = null,
         ?bool $paginationFetchJoinCollection = null,
         ?bool $paginationUseOutputWalkers = null,
-        ?array $paginationViaCursor = null,
```
vincentchalamon, line 115: Why changing the position of this parameter?

vincentchalamon:
I would definitively prefer 2 distinct PRs for:
1. fixing the configuration in XML and YAML formats
2. moving out the `paginationViaCursor` option

dunglas:
I agree with Vincent. The `paginationViaCursor` option should target master. The config fix could even target 2.7 (3.0 is fine too). WDYT @soyuka?

src/Metadata/HttpOperation.php
alanpoulain: Because this parameter was in the "abstract operation arguments" group of parameters.
It doesn't make sense to keep it here, and can cause maintenance issue later.

vincentchalamon (approved):

soyuka:
@dunglas I'm afraid that either we fix this in `main` only (and leave XML support on 3.0 quite broken) or we need to change the `resource.xsd` IIUC. We think the risk is quite minimal on breaking. Not sure what do to.

klobastov:
how can I disable the "Query" operation, leaving only "QueryCollection"?

vincentchalamon:
Hi @klobastov,

Did you try the following configuration?:
```php
Namespace\To\Entity:
  graphQlOperations:
      ApiPlatform\Metadata\GraphQl\QueryCollection: ~
```

Issue #5340, opened 2023-01-14
V3.0.9 breaks symfony


frenci8:
Hi, 

I upgraded api-platform/core to the latest version 3.0.9 on my symfony project and I get the following error on all the pages in my website: 

`"Symfony\Component\HttpKernel\Controller\ArgumentResolver\TraceableValueResolver::resolve()" must yield at most one value for non-variadic arguments.
` and in some other pages the error looks like this `An exception has been thrown during the rendering of a template (""Symfony\Component\HttpKernel\Controller\ArgumentResolver\TraceableValueResolver::resolve()" must yield at most one value for non-variadic arguments.").`

I downgraded to v3.0.8 because that one works fine. 

Any ideas? 

Thanks

KDederichs:
What's the SF version? I'd assume 6.1.x?

KDederichs:
Also, could you do me a favor and see if you locally replace this `yield`:
https://github.com/api-platform/core/blob/c39c3e8c1b775926215a750ecef6a03bb569ef40/src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php#L64
with a `return` if that happens to fix it?

LoicBoursin:
> Also, could you do me a favor and see if you locally replace this `yield`:
> 
> https://github.com/api-platform/core/blob/c39c3e8c1b775926215a750ecef6a03bb569ef40/src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php#L64
> 
> 
> with a `return` if that happens to fix it?

@KDederichs I am having the exact same issue in my project and your solution fixes it 👍🏼

frenci8:
> What's the SF version? I'd assume 6.2.x?

yes, 6.2.4 



> Also, could you do me a favor and see if you locally replace this `yield`:
> 
> https://github.com/api-platform/core/blob/c39c3e8c1b775926215a750ecef6a03bb569ef40/src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php#L64
> 
> 
> with a `return` if that happens to fix it?

yes, that fixes it

KDederichs:
yeah I added a PR with that, I'd assume @soyuka might merge it soon and release a new version

KDederichs:
[...]

PR #5342, opened 2023-01-15
fix: fix argument resolver error

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | #5340 
| License       | MIT
| Doc PR        | n/a

Use return instead of yield to fix the issue


soyuka:
we should've catch this, any way we have to add a test?

soyuka:
I'll quick tag this as it makes 3.0.9 quite broken, lmk if you know how we could prevent this from happening,.

PR #5344, opened 2023-01-16
Homogenize operations constructor

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

- Move existing properties’ phpDoc in the class which declares them
- Add array shape to some properties
- Always call parent’s constructor
- Always promote constructors new properties
- Fix some types


src/Metadata/GetCollection.php
```diff
@@ -94,10 +87,81 @@ public function __construct(
         $provider = null,
         $processor = null,
         array $extraProperties = [],
-        ?string $itemUriTemplate = null
+
+        private ?string $itemUriTemplate = null,
```
soyuka, line 91:
```suggestion
        private ?string $itemUriTemplate = null,
```

can this be protected?

src/Metadata/Delete.php
```diff
@@ -91,8 +86,79 @@ public function __construct(
         ?string $name = null,
         $provider = null,
         $processor = null,
-        array $extraProperties = []
+        array $extraProperties = [],
     ) {
-        parent::__construct(self::METHOD_DELETE, ...\func_get_args());
+        parent::__construct(
+            self::METHOD_DELETE,
```
soyuka, line 92: I'd prefer that we use named arguments here as positional ones may break and it'll be harder to maintain. I wrote https://3v4l.org/OFhko as an experiment to do it magically, not sure it's better.

src/Metadata/GraphQl/Mutation.php
```diff
@@ -37,7 +33,7 @@ public function __construct(
         ?bool $paginationClientPartial = null,
         ?bool $paginationFetchJoinCollection = null,
         ?bool $paginationUseOutputWalkers = null,
-        ?bool $paginationViaCursor = null,
+        ?array $paginationViaCursor = null,
```
soyuka, line 36: we need to check https://github.com/api-platform/core/pull/5333 probably work on this target branch instead

src/Metadata/GetCollection.php
MatTheCat: This can be changed but it would have no effect since the class is `final`.

soyuka: mhh right, I forgot that our end attributes where final. Note to self: this needs an interface.

src/Metadata/GraphQl/QueryCollection.php
```diff
@@ -72,22 +68,57 @@ public function __construct(
         $processor = null,
         array $extraProperties = [],
 
-        protected ?bool $nested = null,
+        ?bool $nested = null,
```
soyuka, line 71: note to self: we need an interface for this as well probably

src/Metadata/Delete.php
soyuka: Cleaned up: https://3v4l.org/KSd7t (by @dunglas )

soyuka: optimization:  https://3v4l.org/qqvuK

chalasr: Hard to read and probably resource intensive (reflection), that looks like unnecessary complexity to me. Better avoid such specificity IMHO

src/Metadata/GraphQl/Mutation.php
MatTheCat: Not sure about #5333’s state; I’ll rebase this one when merged.

src/Metadata/Delete.php
dunglas: I'm in favor of having something automated like that. It will prevent a whole class of errors (forgetting to pass a new parameter to the parent's constructor, and this already happened), and will make the maintenance easier.

The overhead shouldn't be significant because it's the reflection is only used one time per type, and only at build time (at runtime, the cache is used).

👍 on my side.

chalasr: > It will prevent a whole class of errors (forgetting to pass a new parameter to the parent's constructor

Can't those be covered by php-cs-fixer and/or static analyzers as part of the CI?

MatTheCat: Also a problem we currently have (see `$paginationViaCursor`) and is not solved by this approach is that type mismatches are only found at runtime.

dunglas: This should be caught when building the cache at compile time.

soyuka: @chalasr not a big deal as this is cached and unless you call `newInstance` attributes are not computed.

Also the problem of `$paginationViaCursor` is something else we need to cover. 

Anyways its either this approach or the manual one where we use named arguments.

MatTheCat: How would you implement it on the `Operation` -> `HttpOperation` -> `Get` hierarchy?

Default values cannot be used because they’re not passed as arguments, so `new Get` would try to pass `null` to `$method` and fail.

src/Metadata/GraphQl/Mutation.php
soyuka: I merged it!

src/Metadata/Delete.php
soyuka: we can actually handle default values: https://3v4l.org/h5FGN

PR #5348, opened 2023-01-18
refactor(metadata): no default graphql operation name

soyuka:


src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -78,6 +78,7 @@ public function getItemQueryFields(string $resourceClass, Operation $operation,
             return [];
         }
 
+        // TODO: it'd be nice to get rid of this hard coded name
```
alanpoulain, line 81: Why not using `instanceof` here?

src/GraphQl/Type/TypeBuilder.php
```diff
@@ -104,27 +107,25 @@ public function getResourceObjectType(?string $resourceClass, ResourceMetadataCo
...
                     // Use a new type for the wrapped object only if there is a specific normalization context for the mutation or the subscription.
                     // If not, use the query type in order to ensure the client cache could be used.
                     $useWrappedType = $queryNormalizationContext !== $mutationNormalizationContext;
 
-                    $wrappedOperationName = $operationName;
+                    $wrappedOperation = $operation;
 
                     if (!$useWrappedType) {
-                        $wrappedOperationName = $operation instanceof Query ? $operationName : 'item_query';
+                        $wrappedOperation = $resourceMetadataCollection->getGraphQlOperation(null, forceCollection: $operation instanceof Query);
```
alanpoulain, line 126: Why using a named argument here and not elsewhere?

src/GraphQl/Type/TypeConverter.php
```diff
@@ -82,10 +82,7 @@ public function convertType(Type $type, bool $input, Operation $rootOperation, s
                         } catch (ResourceClassNotFoundException|OperationNotFoundException) {
                         }
                         /** @var Query $enumOperation */
-                        $enumOperation = (new Query())
-                            ->withClass($type->getClassName())
-                            ->withShortName($operation?->getShortName() ?? (new \ReflectionClass($type->getClassName()))->getShortName())
-                            ->withDescription($operation?->getDescription());
+                        $enumOperation = new Query(name: 'item_query', class: $type->getClassName(), shortName: $operation?->getShortName() ?? (new \ReflectionClass($type->getClassName()))->getShortName(), description: $operation?->getDescription());
```
alanpoulain, line 85: `name` is not necessary, isn't it?

src/GraphQl/Type/TypeBuilder.php
soyuka: I want to start using them I can revert for the codestyle

src/GraphQl/Type/TypeConverter.php
soyuka: we do need a name

src/Metadata/GraphQl/Query.php
```diff
@@ -115,7 +115,7 @@ class: $class,
             fetchPartial: $fetchPartial,
             forceEager: $forceEager,
             priority: $priority,
-            name: $name ?: 'item_query',
+            name: $name,
```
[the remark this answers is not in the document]
soyuka: because we want to remove logic from our operations

src/GraphQl/Type/TypeBuilder.php
alanpoulain: Personally I think it's better like this indeed.

soyuka:
This is done in  https://github.com/api-platform/core/pull/5657

Issue #5361, opened 2023-01-23
Standard compliant PUT not working when enabled in defaults


nickbg321:
**API Platform version(s) affected**: 3.1.0

**Description**  
When standard compliant PUT method is enabled in defaults it still continues to behave like PATCH (i.e. the old behavior).

Setting it on resource level instead seems to work as expected:
```
#[ApiResource(
    shortName: 'Contact',
    operations: [new Get(), new GetCollection(), new Put(), new Post(), new Patch(), new Delete()],
    normalizationContext: ['groups' => [self::READ_GROUP]],
    denormalizationContext: ['groups' => [self::WRITE_GROUP]],
    provider: ContactProvider::class,
    processor: ContactProcessor::class,
    extraProperties: ['standard_put' => true],
)]
final class ContactDto
{
```

**How to reproduce**  
Enable standard compliant PUT in config/packages/api_platform.yaml by setting it in the defaults section:
```
api_platform:
    defaults:
        extra_properties:
            standard_put: true
```

PR #5362, opened 2023-01-24
fix(metadata): defaults extra properties

soyuka:
#5361


src/Metadata/Resource/Factory/OperationDefaultsTrait.php
```diff
@@ -42,7 +42,7 @@ trait OperationDefaultsTrait
 
     private function addGlobalDefaults(ApiResource|Operation $operation): ApiResource|Operation
     {
-        $extraProperties = [];
+        $extraProperties = $this->defaults['extra_properties'] ?? [];
```
alanpoulain, line 45: I don't think `?? []` is needed, but it's nitpicking.

soyuka: yes but because we can't be sure no one is using this trait out of our bundle I think its safer like this

PR #5367, opened 2023-01-25
get enum type correctly from TypesContainer

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #... 
| License       | MIT
| Doc PR        | none

I faced problems in GraphQL queries when the field that should be queried is an Enum. This leads to errors because the Type could not be found.

enum_types

Enums are stored with their name and the suffix `Enum` in the types container. So I added another case to `TypesContainer:get` method and everything works like a charm.

### DIFF
--- a/features/graphql/introspection.feature
+++ b/features/graphql/introspection.feature
@@ -590,7 +590,7 @@ Feature: GraphQL introspection support
     Then the response status code should be 200
     And the response should be in JSON
     And the header "Content-Type" should be equal to "application/json"
-    And the JSON node "data.person.fields[1].type.name" should be equal to "GenderTypeEnum"
+    And the JSON node "data.person.fields[1].type.name" should be equal to "GenderType"
     #And the JSON node "data.person.fields[1].type.description" should be equal to "An enumeration of genders."
     And the JSON node "data.person.fields[1].type.enumValues[0].name" should be equal to "MALE"
     #And the JSON node "data.person.fields[1].type.enumValues[0].description" should be equal to "The male gender."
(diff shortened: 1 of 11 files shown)


alanpoulain:
I don't think it is the right fix, since the enum type is already added here:
https://github.com/api-platform/core/blob/851aa3778ef8a6f959a44b759fc343e9f8cfc8bf/src/GraphQl/Type/TypeBuilder.php#L274
Please add a Behat test showing your issue.

jotwea:
This is exactly the code that sets the enum to the types container. The `enumKey` that is used here is built from the enum name and the suffix `Enum` as i mentioned. So there has to be a similar code on the reading side that will add the `Enum` suffix before the get will be done. Maybe you are right and I have not found the best place to apply this. But I found no other place where to handle this - do you have a hint for me?

I am not used to work with Behat, but will try to add a test for it.

jotwea:
So I tried my best for the moment @alanpoulain and changed the current behat test scenario to demonstrate what happens if the Enum has not an Enum suffix in its name. This is just for demonstrating the behaviour. If you revert my changes in TypesContainer you can see that the mutation tests fail because the gender value is not in the reponse. With them the test is OK.

Of course my last commit needs to be reverted and we have to design a proper test case for this scenario. I am waiting for your response and please give me some advices how you would expect that I should create a proper test scenario. Should we just add another field with an enum to the Person like `Salutation` (values MR and MRS) ?

alanpoulain:
Thanks for the changes, I didn't understand the enum you use had not the `Enum` suffix.
The PR https://github.com/api-platform/core/pull/5369 should solve your issue properly.

alanpoulain:
I'm closing this PR, the patch will be available for 3.1.1. Thank you @jotwea.

PR #5369, opened 2023-01-25
fix(graphql): name and key should be the same for an enum without Enum suffix in class name

alanpoulain:
Supersedes https://github.com/api-platform/core/pull/5367.


jotwea:
Today I applied this fix to my codebase and I am not totally fine with it @alanpoulain. As a side effect all the enum types are also suffixed with `Enum` in the generated graphql schema now. As I use the Apollo Code Generator to create TypeScript code out of the schema this will be a BC now.

There would be an easy way to solve in `TypeBuilder` this by replacing:
```
$enumConfig = [
    'name' => $enumName,
    'values' => $enumCases,
];
```
with
```
$enumConfig = [
    'name' => $operation->getShortName(),
    'values' => $enumCases,
];
```

Do you agree?

alanpoulain:
Why would it be a BC? It was not working before, so it should not considered a BC in my opinion.

PR #5373, opened 2023-01-26
fix(symfony): wrong purger clients type

### LINKED ISSUE #5365: [BUG] v3.1.0 VarnishXKeysPurger getting a RewindableGenerator instead of an array
**API Platform version(s) affected**: 3.1.0

**Description**  
Since updating I now get an error:
```
ApiPlatform\HttpCache\VarnishXKeyPurger::__construct(): Argument #1 ($clients) must be of type array, Symfony\Component\DependencyInjection\Argument\RewindableGenerator given
```

**How to reproduce**  
Just by using the `VarnishXKeys` purger, has there been a change which would result in this? I'm noticing that the service declaration is for a set of tagged services. Was there any update to Symfony which changed this?

soyuka:
fixes #5365

### DIFF
--- a/src/HttpCache/SouinPurger.php
+++ b/src/HttpCache/SouinPurger.php
@@ -29,7 +29,7 @@ class SouinPurger extends SurrogateKeysPurger
     /**
      * @param HttpClientInterface[] $clients
      */
-    public function __construct(array $clients)
+    public function __construct(iterable $clients)
     {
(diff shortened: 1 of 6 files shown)


silverbackdan:
This fix gets all my tests working again as expected, thanks!

soyuka:
I changed the type to `iterable` and added a test should be good (will be out tomorrow)

PR #5375, opened 2023-01-26
fix(metadata): allow custom http status code for put

SpartakusMd:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| License       | MIT

Fixes the possibility to use custom http status codes for the PUT methods. Currently, it returns 201 HTTP_CREATED.


soyuka:
good catch thanks!

PR #5378, opened 2023-01-27
fix(metadata): reader should be nullable

### LINKED ISSUE #5371: Issue with annotation_reader in 2.7 (compatibility_layer: true)
**API Platform version(s) affected**: 2.7.7

I encounter a bug that is similar to https://github.com/api-platform/core/pull/5358 in ApiPlatform 2.7 with `metadata_backward_compatibility_layer: true`.

image

**Description**  

When updating ApiPlatform, the following error is thrown when composer is clearing Symfony's cache.

`The service "api_platform.metadata.resource.name_collection_factory.annotation" has a dependency on a non-existent service "annotation_reader".`

**How to reproduce**  

Just update to ApiPlatform 2.7.7 with `metadata_backward_compatibility_layer: true`.

**Possible Solution**  

Adding `on-invalid="null"` in this file https://github.com/api-platform/core/blob/v2.7.7/src/Symfony/Bundle/Resources/config/legacy/metadata_annotation.xml solve the issue.

Thank you!

louismariegaborit:
fixes #5371


soyuka:
weird that Phpstan didn't catch that

PR #5384, opened 2023-02-01
feat(openapi) : Adding option to addPath to keep custom sort.

matthieu-delisle:
We needed to change the order of path in the Open API auto generated documentation but the ksort was resetting the custom order. I added an option to bypass the ksort so we can keep our sort order.

### DIFF
--- a/src/OpenApi/Model/Paths.php
+++ b/src/OpenApi/Model/Paths.php
@@ -17,11 +17,13 @@ final class Paths
 {
     private array $paths = [];
 
-    public function addPath(string $path, PathItem $pathItem): void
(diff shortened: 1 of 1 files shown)


soyuka:
I'm afraid that there could be an issue with https://github.com/api-platform/core/commit/4f363fc5136221663bfbcde9d36f7fb15fee4261

soyuka:
You can always add a Normalizer that changes the order of the array no?

matthieu-delisle:
I finally used Reflexion Class to change the order of the private "paths" properties since in the Normalizer it is an OpenApi object that doesn't have the setPaths option.

Issue #5381, opened 2023-01-31
Response description is not changable when using openapi


sajixavier:
**API Platform version(s) affected**: 3.1.1

**Description**  

With this new version, it seems impossible to change the response description for a given status.

**How to reproduce**  

Define an ApiResource with a Put operation like this (might be same for other operations)

```
new Metadata\Put(
            status: 200,
            openapi: new Model\Operation(
                summary: 'Test summary',
                responses:[
                    '200' => new Model\Response(
                        description: 'Custom response message'
                    )
                ],
            ),
        )
```
In swagger documentation description is shown as '(Resource name) resource updated'

**Additional Context**  

Symfony: 6.2.5
PHP: 8.1

Byidi:
You can change responses descriptions using `openapiContext['responses'`:

```php
#[Put(
    openapiContext: [
        'responses' =>  [
            200 => [
                'description' => 'Custom response message',
            ]
        ]
    ]
)]
```

soyuka:
We're working on supporting the `openapi`, `openapiContext` is deprecated ;)

HuyNguyen206:
hi @soyuka, regarding api platform, I can't find the document that describe how to use openapi, openapiContext option so It is quite difficult to custom the response, and description in the Swagger documentation. Could you show me how to use these options? Thank you

Zenifica:
same than @HuyNguyen206

monitaurus:
> hi @soyuka, regarding api platform, I can't find the document that describe how to use openapi, openapiContext option so It is quite difficult to custom the response, and description in the Swagger documentation. Could you show me how to use these options? Thank you

It's not a long detailed documentation, but here's a quick guide on how to use `openapi`: [Extend OpenAPI Documentation
](https://api-platform.com/docs/guides/extend-openapi-documentation/)

An interesting part here is that we can manipulate objects rather than array.
It reduces errors and enhances a lot the self discoverability of what we can set.

But, there's this line in the documentation:
> When a field is not specified API Platform will add the missing informations.

It seems valid only for the argument within the Operation level.
In my case, I tried to only set the description of a request body, like this:
```php
    openapi: new Operation(
        requestBody: new RequestBody(
            description: 'Description'
        )
    )
```
The operation's description, the parameters, the responses, ... all is filled up by API Platform.
But as the `requestBody` is a new instance, it replaces the whole request body, including the automatically generated example and schema, even if I wanted to change only the description.
[...]

PR #5393, opened 2023-02-07
fix(openapi): allow overriding of openapi responses

sajixavier:
| Q             | A
| ------------- | ---
| Branch?       | 3.1 
| Tickets       | #5381
| License       | MIT

This fix is for overriding the responses in swagger via openapi

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -279,38 +279,44 @@ final class OpenApiFactory implements OpenApiFactoryInterface
                 }
             }
 
+            $existingResponses = $openapiOperation?->getResponses() ?: [];
             // Create responses
             switch ($method) {
                 case HttpOperation::METHOD_GET:
                     $successStatus = (string) $operation->getStatus() ?: 200;
-                    $responseContent = $this->buildContent($responseMimeTypes, $operationOutputSchemas);
-                    $openapiOperation = $openapiOperation->withResponse($successStatus, new Response(sprintf('%s %s', $resourceShortName, $operation instanceof CollectionOperationInterface ? 'collection' : 'resource'), $responseContent));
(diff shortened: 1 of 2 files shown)


src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -279,38 +279,41 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
             }
 
+            $existingResponses = $openapiOperation?->getResponses() ?: [];
             // Create responses
             switch ($method) {
                 case HttpOperation::METHOD_GET:
                     $successStatus = (string) $operation->getStatus() ?: 200;
                     $responseContent = $this->buildContent($responseMimeTypes, $operationOutputSchemas);
-                    $openapiOperation = $openapiOperation->withResponse($successStatus, new Response(sprintf('%s %s', $resourceShortName, $operation instanceof CollectionOperationInterface ? 'collection' : 'resource'), $responseContent));
+                    $openapiOperation = $openapiOperation->withResponse($successStatus, $this->buildOpenApiResponse($existingResponses, $successStatus, sprintf('%s %s', $resourceShortName, $operation instanceof CollectionOperationInterface ? 'collection' : 'resource'), $responseContent));
```
soyuka, line 288: Actually the rule to apply here is that if a response exists in `$openapiOperation->getResponses()[200]` we should use it instead of building our own response. 
The code here should be:

```
if (!isset($existingResponses[$successStatus])) {
                    $openapiOperation = $openapiOperation->withResponse($successStatus, new Response(sprintf('%s %s', $resourceShortName, $operation instanceof CollectionOperationInterface ? 'collection' : 'resource'), $responseContent));
}
```

This allows the user to create his own responses and we don't alter the user definition, which may cause confusion. 

Obviously using a method for this can reduce the code complexity (especially bellow where we have mutliple statuses).

sajixavier: Yeah it make sense.

soyuka (approved):
Nice thanks! Can you target the `3.1` branch though?

sajixavier:
@soyuka I have targeted to 3.1 branch.

sajixavier:
codecov/project is showing error. I could see the newly added code is covered. Any idea? https://app.codecov.io/gh/api-platform/core/pull/5393

soyuka:
Thanks @sajixavier !

PR #5394, opened 2023-02-08
feat(filter): proposal about uuid validation

jewome62:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | No ticket
| License       | MIT
| Doc PR        | proposal not documented yet

Situation : 
Currently, if you filter on uuid typed field with SearchFilter and the uuid is not valid. You have a DriverException about SQL fail
(tested on Postgresql) 

> An exception occurred while executing a query: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for type uuid: "bad-uuid"

Proposal : 
Catch as the int validation, the bad uuid with `hasValidValues`

So firstly, what do you think about that ?
And secondly, do you have idea to manage Custom UUID type from third dependency (Ramsey or Symfony) for make `DOCTRINE_UUID_TYPE` constant.
And third, I did not see some test currently the  `hasValidValues` behavior. Do you have some functional test about that ? I did not find them.

### DIFF
--- a/src/Doctrine/Common/Filter/SearchFilterTrait.php
+++ b/src/Doctrine/Common/Filter/SearchFilterTrait.php
@@ -18,7 +18,9 @@ use ApiPlatform\Api\IriConverterInterface;
 use ApiPlatform\Doctrine\Common\PropertyHelperTrait;
 use ApiPlatform\Exception\InvalidArgumentException;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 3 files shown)


soyuka:
There's a feature called Query Parameter Validation (https://github.com/api-platform/core/pull/1723) that should help you do exactly that but not a the query level. Maybe it is more appropriate? Sorry there aren't much documentation on the subject and there's an on-going process to improve that part.

Issue #5372, opened 2023-01-26
PUT tries to create new entity even when it already exists if ORM id is hidden in the API


mrossard:
**API Platform version(s) affected**: 3.1.0

**Description**  
Sorry, i'm back with my non-standard uses!

I'm trying to use PUT for create or update on an entity where the ORM ids are purely internal, and other "human-readable" identifiers are used in the API/uriVariables.

When you do that, even if the provider does return the correct record, the Doctrine PersistProcessor tries to persist an entity where the orm id is not set, which causes either duplicate data, or a unique violation if your human readable identifier is set correctly on the database side.

**How to reproduce**  
I can add a reproducer if you think it's useful, but the problem seems pretty straightforward...it's more of a "how should this actually behave?" problem IMO. 

**Possible Solution**  
In ApiPlatform\Doctrine\Common\State\PersistProcessor, around lines 80 to 96 the properties tagged #[ORM\Id] likely shouldn't be overwritten if not made available throught the API.
Another option is obviously to write a custom processor, but shouldn't/couldn't it be standard behaviour?

mrossard:
Another puzzling thing while trying to write a custom processor instead : the contents you get in $context['previous_data'] in the processor gets detached from doctrine's unitofwork somehow, which makes it awkward to use.

soyuka:
what about not creating 2 issues around this and merging it with https://github.com/api-platform/core/issues/5370 ?

mrossard:
It seemed like a different issue to me, and i thought you'd close the other one since allowCreate looked like the right answer - but i can close this one and duplicate its contents there if you like.

soyuka:
Okay sorry then I didn't understood your issue

mrossard:
I'll try to find some time to craft an example then. But after reading more closely what @dunglas says in https://github.com/api-platform/core/pull/4996 , it might actually be what he was refering to when he said "My refactoring doesn't handle the case where URI variables don't match the Doctrine properties"...

soyuka:
yes it's a hard problem if that's the case

mrossard:
I can imagine that! I couldn't get to it yet but i should find some time for an example in the next few days.

mrossard:
A small update : i tried to build a simple example, but if i make it simple i stumble into another problem with the PersistProcessor.
IMO it needs a lot of work still, but as i dig into it i understand how complicated it is too.

[...]

PR #5408, opened 2023-02-16
fix(standard-put) correct retrieval of the updated instance when the ORM id is not in the URI

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | stable
| Tickets       | https://github.com/api-platform/core/issues/5372

When using standard PUT for update, the doctrine PersistProcessor could only find the correct entity if its ORM id was present in the uriVariables. This PR tries to fix this using the previous_data present in the context instead of relying on uriVariables.

Additionnally...that's my first PR, so i'm not sure i did everything right. I included a simple behat test and those pass - phpunit tests were failing out of the box on something openapi-related, i didn't touch those.

### DIFF
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -553,6 +553,31 @@ Feature: Create-Retrieve-Update-Delete
     Then the response status code should be 400
     And the JSON node "hydra:description" should be equal to "Syntax error"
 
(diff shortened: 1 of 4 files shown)


mrossard:
...and that's not the correct version, so i'll do over.

PR #5409, opened 2023-02-16
fix : (standard-put) correct retrieval of the updated object when the ORM id is not in the URI.

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | stable
| Tickets       | https://github.com/api-platform/core/issues/5372

When using standard PUT for update, the doctrine PersistProcessor could only find the correct entity if its ORM id was present in the uriVariables. This PR tries to fix this using the previous_data present in the context instead of relying on uriVariables.

I only included a simple behat test for this, it may be useful to add a phpunit for automatic coverage validation purposes but i wasn't sure how to approach this...and existing behat tests do seem to be covering the added code decently.

### DIFF
--- a/src/Doctrine/Common/State/PersistProcessor.php
+++ b/src/Doctrine/Common/State/PersistProcessor.php
@@ -52,10 +52,34 @@ final class PersistProcessor implements ProcessorInterface
         // https://github.com/doctrine/orm/issues/8461#issuecomment-1250233555
         if ($operation instanceof HttpOperation && HttpOperation::METHOD_PUT === $operation->getMethod() && ($operation->getExtraProperties()['standard_put'] ?? false)) {
             \assert(method_exists($manager, 'getReference'));
-            // TODO: the call to getReference is most likely to fail with complex identifiers
             $newData = $data;
             if (isset($context['previous_data'])) {
-                $newData = 1 === \count($uriVariables) ? $manager->getReference($class, current($uriVariables)) : clone $context['previous_data'];
+                // the correct entity need to be fetched, we need to get the ORM ids from previous data
+                $reflectionProperties = $this->getReflectionProperties($context['previous_data']);
+                $ormIds = [];
+                \assert(method_exists($manager, 'getClassMetadata'));
+                foreach ($manager->getClassMetadata($class)->getIdentifier() as $identifier) {
+                    if (\array_key_exists($identifier, $reflectionProperties)) {
+                        $value = $reflectionProperties[$identifier]->getValue($context['previous_data']);
+                    } else {
+                        // This id is not from current class, let's move up in class hierarchy
+                        $parent = (new \ReflectionObject($data))->getParentClass();
+                        while (!isset($value) && false !== $parent) {
+                            foreach ($parent->getProperties() as $property) {
+                                if ($property->name === $identifier) {
+                                    $value = $property->getValue($context['previous_data']);
+                                }
+                            }
+                            $parent = $parent->getParentClass();
+                        }
+                    }
+                    if (!isset($value)) {
+                        // can this actually happen? the identifier has to exist in the class hierarchy somewhere...
+                        throw new \Exception('identifier not found : '.$identifier);
+                    }
+                    $ormIds[$identifier] = $value;
+                }
+                $newData = $manager->getReference($class, $ormIds);
             }
 
             $identifiers = array_reverse($uriVariables);
@@ -79,7 +103,7 @@ final class PersistProcessor implements ProcessorInterface
             } else {
                 foreach ($reflectionProperties as $propertyName => $reflectionProperty) {
                     // Don't override the property if it's part of the subresource system
-                    if (isset($uriVariables[$propertyName])) {
+                    if (isset($uriVariables[$propertyName]) || isset($ormIds[$propertyName])) {
                         continue;
                     }
 
--- a/tests/Fixtures/TestBundle/Entity/HiddenIdentifierDummy.php
+++ b/tests/Fixtures/TestBundle/Entity/HiddenIdentifierDummy.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\Entity;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Get;
+use ApiPlatform\Metadata\Put;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+#[ApiResource(
+    operations            : [new Get(), new Put(allowCreate: true)],
+    normalizationContext  : ['groups' => ['HiddenIdentifierDummy::out']],
+    denormalizationContext: ['groups' => ['HiddenIdentifierDummy::in']],
+    extraProperties       : [
+        'standard_put' => true,
+    ],
+)]
+#[Entity]
+class HiddenIdentifierDummy
+{
+    #[Id]
+    #[Column]
+    #[ApiProperty(identifier: false)]
+    public ?int $id = null;
+
+    #[Column]
+    #[ApiProperty(identifier: true)]
+    #[Groups(['HiddenIdentifierDummy::out'])]
+    public ?string $visibleId = null;
+
+    #[Column]
+    #[Groups(['HiddenIdentifierDummy::out', 'HiddenIdentifierDummy::in'])]
+    public string $foo = '';
+}
--- a/features/main/crud.feature
+++ b/features/main/crud.feature
@@ -553,6 +553,31 @@ Feature: Create-Retrieve-Update-Delete
     Then the response status code should be 400
     And the JSON node "hydra:description" should be equal to "Syntax error"
 
+  @!mongodb
+  Scenario: Replace an existing resource that doesn't expose its internal identifier
+    Given there is an HiddenIdentifierDummy object
+    When I add "Content-Type" header equal to "application/json"
+    And I send a "PUT" request to "/hidden_identifier_dummies/prettyid" with body:
+    """
+    {
+      "foo": "A nice dummy"
+    }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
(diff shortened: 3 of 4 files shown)


mrossard:
This tries to handle the cases mentionned by @dunglas here : https://github.com/api-platform/core/pull/4996#discussion_r977321985

mrossard:
I hadn't tested with mongodb, which i don't use...and doctrine ODM seems to behave a bit differently. I'll try to adapt my code, but it seems to be a simple change on line 82 : 

$newData = $manager->getReference($class, $ormIds);

would become : 

$newData = $manager->getReference($class, count($ormIds) === 1 ? current($ormIds) : $ormIds);

soyuka:
I'm really not sure that we want this to be part of API Platform, why don't you create your own provider to handle such a case? 
Note we need more opinions on the subject but handling cases like this in the past introduced a lot of complexity. Also, the IdentifierExtractor should do the job of getting the identifier from the class we don't want to write that code again.

mrossard:
Thanks for the feedback! 

I think including this (or a better crafted version of this) has a lot of merit though. In the project where i encountered this problem i did use a custom processor (not provider, but i suppose that's what you meant), but since api platform already has everything it needs to do the job correctly, why not do it?

As far as i can tell, it works as intended for PATCH, so i think it should work for standard PUT too, otherwise the framework is not consistent with itself - and the author of the current version even included a TODO to warn about it not being able to handle everything correctly!


About the IdentifiersExtractor, i'm all for it if it can do this, duplicating code is almost always wrong...but it seemed to me like it "belongs" to the "provider side" of things, not the "processor side". I might be wrong though...?

mrossard:
Any chance someone else wants to look at this so we can at least discuss wether it should be done? 
@dunglas maybe, as the original author of this part?

I have some time currently to work on it if necessary, but it might not last long.

soyuka:
The identifier extractor part is meant to work for providers and processor, its a layer that's just above. Also, when you work on updating something, usually you did call a Provider to fetch the data to be updated.

src/Doctrine/Common/State/PersistProcessor.php
```diff
@@ -52,10 +52,34 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
...
-                $newData = 1 === \count($uriVariables) ? $manager->getReference($class, current($uriVariables)) : clone $context['previous_data'];
+                // the correct entity need to be fetched, we need to get the ORM ids from previous data
+                $reflectionProperties = $this->getReflectionProperties($context['previous_data']);
+                $ormIds = [];
+                \assert(method_exists($manager, 'getClassMetadata'));
+                foreach ($manager->getClassMetadata($class)->getIdentifier() as $identifier) {
+                    if (\array_key_exists($identifier, $reflectionProperties)) {
+                        $value = $reflectionProperties[$identifier]->getValue($context['previous_data']);
+                    } else {
+                        // This id is not from current class, let's move up in class hierarchy
```
soyuka, line 65: This is a behavior I want to remove from API Platform and that is done at https://github.com/api-platform/core/blob/9b4b58ca0113a2645b58a116fe9e4bf200df8aa3/src/Api/IdentifiersExtractor.php#L96. 

See this function is quite wrong already as it is "guessing" that identifiers are at the correct place. The correct way do to this is that in your `resourceClass` you have a property marked as `ApiProperty(identifier: true)`. It can be a ghost property that is just fetching `parent::getId()` but it should be stringable.

Example:

```php
class Foo { protected $id; }

class Bar extends Foo {
  #[ApiProperty(identifier: true)]
  private string $myCustomId;
  
  // must be accessible/readable!
  public function getMyCustomId(): string {
      return parent::$id;
  }
}
```

src/Doctrine/Common/State/PersistProcessor.php
```diff
@@ -52,10 +52,34 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
         // https://github.com/doctrine/orm/issues/8461#issuecomment-1250233555
         if ($operation instanceof HttpOperation && HttpOperation::METHOD_PUT === $operation->getMethod() && ($operation->getExtraProperties()['standard_put'] ?? false)) {
             \assert(method_exists($manager, 'getReference'));
-            // TODO: the call to getReference is most likely to fail with complex identifiers
```
soyuka, line 55: this is a special case, not yet working even in your patch and is about composite identifiers (multiple identifiers)

src/Doctrine/Common/State/PersistProcessor.php
```diff
@@ -52,10 +52,34 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
...
+                foreach ($manager->getClassMetadata($class)->getIdentifier() as $identifier) {
+                    if (\array_key_exists($identifier, $reflectionProperties)) {
+                        $value = $reflectionProperties[$identifier]->getValue($context['previous_data']);
+                    } else {
+                        // This id is not from current class, let's move up in class hierarchy
+                        $parent = (new \ReflectionObject($data))->getParentClass();
+                        while (!isset($value) && false !== $parent) {
+                            foreach ($parent->getProperties() as $property) {
+                                if ($property->name === $identifier) {
+                                    $value = $property->getValue($context['previous_data']);
```
soyuka, line 70: this looks a lot like https://github.com/api-platform/core/blob/9b4b58ca0113a2645b58a116fe9e4bf200df8aa3/src/Api/IdentifiersExtractor.php#L96 we should already have extracted the identifiers and they should be available inside `uriVariables` at this point.

src/Doctrine/Common/State/PersistProcessor.php
```diff
@@ -52,10 +52,34 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
...
+                                }
+                            }
+                            $parent = $parent->getParentClass();
+                        }
+                    }
+                    if (!isset($value)) {
+                        // can this actually happen? the identifier has to exist in the class hierarchy somewhere...
+                        throw new \Exception('identifier not found : '.$identifier);
```
soyuka, line 78: don't its not the responsibility of this class, also we have an `IdentifierNotFoundException`

tests/Fixtures/TestBundle/Entity/HiddenIdentifierDummy.php
```diff
@@ -0,0 +1,49 @@
...
+
+    #[Column]
+    #[ApiProperty(identifier: true)]
+    #[Groups(['HiddenIdentifierDummy::out'])]
+    public ?string $visibleId = null;
+
+    #[Column]
+    #[Groups(['HiddenIdentifierDummy::out', 'HiddenIdentifierDummy::in'])]
+    public string $foo = '';
+}
```
soyuka, line 49: can you explain your use case? I think that this should work already.

soyuka (commented):
sorry was away for some time, I'm back lets discuss this

tests/Fixtures/TestBundle/Entity/HiddenIdentifierDummy.php
mrossard: I'll answer this first since it's faster  : i have a database with generated ids, and "business" ids as strings, which i want to use so API calls use the "business-recognisable" id. 
It lets users call "complex" IRIs (containing 2 or 3 such ids) without having to make multiple calls to get the right "internal" ids first.

And right now (last i tested at least) i can confirm it doesn't work : the orm id is not present in the uriVariables, and has no reason to be there, so this call to getReference() can't work : https://github.com/api-platform/core/blob/9b4b58ca0113a2645b58a116fe9e4bf200df8aa3/src/Doctrine/Common/State/PersistProcessor.php#L58

In my (only semi-confident) opinion uriVariables shouldn't be involved in this part of the framework at all, decoding the uri should be over and done by the time the workflow gets here.

src/Doctrine/Common/State/PersistProcessor.php
mrossard: Ok, i think i get your point here, i'll try to change this so it uses IdentifierExtractor instead. 

`ApiProperty(identifiers: true)`  doesn't seem to apply to the current problem though : the id doctrine needs is not the one used in the URI in my case.

src/Doctrine/Common/State/PersistProcessor.php
mrossard: I haven't tested composite identifiers, but i thought calling getReference with a correct $ormIds array would do the trick, sorry about that. I'll look into this too.

src/Doctrine/Common/State/PersistProcessor.php
mrossard: This does look a lot like what IdentifierExtractor does and likely needs to be replaced, but i disagree on the last part : there's no reason for these values to appear in `uriVariables` in this case, or i'm missing something.

src/Doctrine/Common/State/PersistProcessor.php
mrossard: I suppose that's done in IdentifiersExtractor once again. 
That was a random Exception "just in case", but as i commented i don't see a reason for this to happen anyway.

mrossard:
>  Also, when you work on updating something, usually you did call a Provider to fetch the data to be updated.

Of course i have an associated custom provider for this, but it doesn't have any impact on the problem at hand - it does its job correctly using the "business ids" provided in the URI.

tests/Fixtures/TestBundle/Entity/HiddenIdentifierDummy.php
soyuka: And its a good idea but this means that you have to decorate our processor to resolve the correct identifiers.

Let's take an example following your logic. Please note that I'm using a class that is not a Doctrine Entity as a resource.

```php
namespace App\ApiResource;
use ApiPlatform\Doctrine\Orm\State\Options;

#[ApiResource(stateOptions: new Options(entityClass: SomeEntity::class))]
final class SomeApiResource {
  // name is "id" and gets automatically discovered by API Platform as an identifier, it is your business id
  public string $id; 
  // some API fields
  public string $name;
}
```

Now let's see that `SomeEntity`:

```php

namespace App\Entity;

#[ORM\Entity]
final class SomeEntity {
    #[ORM\Column(type: 'integer')]
    #[ORM\Id]
    #[ORM\GeneratedValue(strategy: 'AUTO')]
    public ?int $id = null;

    // This is our `id` from the SomeApiResource
    #[ORM\Column(type: 'text')]
    public string $myBusinessId;
}
```

We want users to do this kind of separation as it's cleaner. It's written in the [General Design Considerations](https://api-platform.com/docs/core/design/) and is considered best practice. 

Now you can see that we have an issue as API Platform will not be able to transform your `SomeApiResource::id` in the `SomeEntity::myBusinessId`. For that, we recommend to **decorate** our Processor/Providers and fix it yourself. This is exactly what we did by adding this `entityClass` option (see https://github.com/api-platform/core/blob/main/src/Doctrine/Orm/State/ItemProvider.php#L46). 

Maybe that we could add an `entityIdentifier` option that would use the `businessId` instead of the `id` to query stuff? 

I hope that this points you to a better direction then trying to make our processor work on complex resources that are also entities.

Issue #4994, opened 2022-09-21
RFC-7807 compatible hydra error response


nesl247:
**Description**  
> Finally, the server SHOULD provide error descriptions using an [[RFC7807](https://www.hydra-cg.com/spec/latest/core/#bib-rfc7807)] standard by using an application/problem+json response. When doing so, the server also MUST provide an additional header pointing to either the built-in Hydra http://www.w3.org/ns/hydra/error context or any JSON-LD context that maps the terms type, title, detail, status and instance the same way as the standard one.

The existing response is the `Non RFC-7807 compliant error description using raw Hydra JSON-LD representation.` as described here https://www.hydra-cg.com/spec/latest/core/#example-32-non-rfc-7807-compliant-error-description-using-raw-hydra-json-ld-representation.

dunglas:
Indeed. Compatibility with RFC7807 is new in Hydra and we didn't updated our implementation since it has been added. PR welcome!

nesl247:
@dunglas how would you want it to be implemented? Would it be a BC breaking change or would it need to be a new error format `jsonld+problem`? Or would it be more appropriate to be called `hydra` because it's the format described by `hydra` itself, and not `jsonld`?

dunglas:
I think that we can make our current implementation of RFC7807 compatible with Hydra without introducing breaking changes (it's only new keys). Then we'll deprecate the old Hydra error format. WDYT?

nesl247:
I think that would work. Let me take a look and submit a PR for that. If I run into any issues I’ll reply back here.

nesl247:
I'm starting to look into it, and the bigger problem that we need to figure out is the `type` property. This MUST be a `URI` format, which means we need to have some sort of response if someone follows it.

I have the following questions:

* What would the URL prefix you would expect to have (e.g. `/errors`)?
* What do we want the full URL to look like? Do we use `/errors/<ExceptionName>`? Or do we reuse `ErrorCodeSerializableInterface::getErrorCode()` and thus `/errors/<errorCode>`?
* What do we serve at that URL?
    * Do we need to expand `ErrorCodeSerializableInterface` to an `ErrorSerializableInterface` so we can also have a `getErrorTitle()` and `getErrorDescription()`?
        * How do we want to handle translation here?
    * Do we need to support HTML, or just JSON formats?
* Do we want to allow for overriding the URL to something custom, in case there is publicly hosted documentation not provided in the API?

[...]

PR #5433, opened 2023-03-06
feat: error as resources, jsonld errors are now problem-compliant

JacquesDurand:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | #4994 
| License       | MIT

Hello !

This is an attempt to implement [RFC-7807 for hydra error responses](https://www.hydra-cg.com/spec/latest/core/#example-31-rfc-7807-compatible-error-description)

My current (draft) try on the implementation is the following:

- In case `'jsonld' === $config['error_format']`:
    - From the thrown `\Exception`, construct a new kind of exception: `ErrorException`, and pass it to the `ControllerArgumentEvent` (cf the diff on `src/Symfony/EventListener/ErrorListener.php`. The goal of this operation is to progressively break the dependency on Symfony's FlattenException (and then potentially also link stuff from Laravel for instance)
    - From this ErrorException, instantiate a new `Error` ApiResource in `ExceptionAction`, and normalize it in Hydra\ErrorNormalizer. This is mainly for `'@id'` purposes, and also maybe to enable user to create their own Error resources based on this one.
> Note: this case is probably non RFC-7807 compliant, but still enables clearer Error serialization than there currently is

- In case `'jsonproblem' === $config['error_format']`:
    - The same operation concerning the`ErrorException` happens
    - instantiate a `ProblemError` ApiResource instead, which matches the json schema required by the RFC in case the server returns an 'application/problem+json' response.
    - Normalize it for hydra in `Problem\ErrorNormalizer`

BC Layer for now: enable/disable these error serializations via an `extraProperty` field on each resource (false by default), e.g. 
```php
#[ApiResource(extraProperties: ['hydra_errors' => true])]
class MyResource
{
...
}
```

As for now, what remains to be done (if this kind of implementation suits you): 

- [ ] A lot of tests
- [ ] Documenting this feature
- [ ] Add needed headers via `AddLinkListener`
- [ ] Suggestion: creating an `#[ErrorResource]` Attribute to allow users to easily define their own errors with more precise details based on their needs ?

This is still a pretty large work in progress, if I am going in the completely wrong direction i'd be happy to change anything !


src/Action/ExceptionAction.php
```diff
@@ -46,7 +49,7 @@ public function __construct(private readonly SerializerInterface $serializer, pr
     /**
      * Converts an exception to a JSON response.
      */
-    public function __invoke(FlattenException $exception, Request $request): Response
+    public function __invoke(FlattenException|ErrorExceptionInterface $exception, Request $request): Response
```
soyuka, line 52:
```suggestion
    /**
     * @param FlattenException|ErrorExceptionInterface $exception
     */
    public function __invoke($exception, Request $request): Response
```

To avoid breaking if this class is used by users.

src/Problem/Serializer/ErrorNormalizer.php
```diff
@@ -35,7 +37,7 @@ final class ErrorNormalizer implements NormalizerInterface, CacheableSupportsMet
         self::TITLE => 'An error occurred',
     ];
 
-    public function __construct(private readonly bool $debug = false, array $defaultContext = [])
+    public function __construct(private readonly IriConverterInterface $iriConverter, private readonly bool $debug = false, array $defaultContext = [])
```
soyuka, line 40: you can't break this class, the argument must be last position and nullable + add an expect deprecation although I'm quite sure we can handle it with a default IriConverter to avoid the deprecation alltogether.

src/Action/ExceptionAction.php
```diff
@@ -71,7 +74,13 @@ public function __invoke(FlattenException $exception, Request $request): Respons
         $headers['X-Content-Type-Options'] = 'nosniff';
         $headers['X-Frame-Options'] = 'deny';
 
-        return new Response($this->serializer->serialize($exception, $format['key'], ['statusCode' => $statusCode]), $statusCode, $headers);
+        $error = match ($format['key']) {
+            'jsonproblem' => $exception instanceof FlattenException ? $exception : new ProblemError($exception, $request->getPathInfo()),
```
soyuka, line 78: why the `$request->getPathInfo()` ? looks like a security issue

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -34,4 +41,47 @@ protected function duplicateRequest(\Throwable $exception, Request $request): Re
 
         return $dup;
     }
+
+    public function onControllerArguments(ControllerArgumentsEvent $event): void
```
soyuka, line 45: is that code from Symfony? It looks quite heavy could you explain to me why you needed that?

soyuka (commented):
This implementation looks good to me, please keep the `ApiErrorResource` for another PR we just want to be closer to the specification for now.

src/ApiResource/Error.php
```diff
@@ -0,0 +1,59 @@
...
+    private string $title;
+
+    private string $description;
+
+    #[ApiProperty(identifier: true)]
+    private int $statusCode;
+
+    private array $trace;
+
+    public function __construct(ErrorExceptionInterface $exception)
```
alanpoulain, line 32: Use public readonly promoted constructor parameters instead of private with getter properties.

src/ApiResource/Error.php
```diff
@@ -0,0 +1,59 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\ApiResource;
+
+use ApiPlatform\Exception\ErrorExceptionInterface;
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\NotExposed;
+
+#[NotExposed(uriTemplate: '/errors/{statusCode}')]
+class Error
```
alanpoulain, line 21:
```suggestion
final class Error
```

src/Metadata/Resource/Factory/StaticResourceNameCollectionFactory.php
```diff
@@ -0,0 +1,41 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata\Resource\Factory;
+
+use ApiPlatform\Metadata\Resource\ResourceNameCollection;
+
+final class StaticResourceNameCollectionFactory implements ResourceNameCollectionFactoryInterface
```
alanpoulain, line 18: I don't like the "static" wording. Internal? Included?

src/Problem/Serializer/ErrorNormalizer.php
```diff
@@ -45,12 +47,24 @@ public function __construct(private readonly bool $debug = false, array $default
...
+
         $data = [
-            'type' => $context[self::TYPE] ?? $this->defaultContext[self::TYPE],
-            'title' => $context[self::TITLE] ?? $this->defaultContext[self::TITLE],
-            'detail' => $this->getErrorMessage($object, $context, $this->debug),
+            'type' => $object instanceof ProblemError ? $this->iriConverter->getIriFromResource($object, context: $identifiers) : $context[self::TYPE] ?? $this->defaultContext[self::TYPE],
+            'title' => $object instanceof ProblemError ? $object->getTitle() : $context[self::TITLE] ?? $this->defaultContext[self::TITLE],
+            'detail' => $object instanceof ProblemError ? $object->getDetail() : $this->getErrorMessage($object, $context, $this->debug),
+            'instance' => $object instanceof ProblemError ? $object->getInstance() : null,
+            'status' => $object instanceof ProblemError ? $object->getStatus() : $context['statusCode'] ?? $object->getStatusCode(),
```
alanpoulain, line 61: Isn't adding the `status` a BC?

nesl247:
I'm not seeing `type` here, which for me is the biggest feature of RFC 7807.

That brings to me to what I think the big missing piece here, which is how the user is going to populate the `type`, `title`, `detail`, `instance`, etc. My suggestion is to introduce a new interface called `ProblemException`, which mandates the getters. If the exception is already a `ProblemError`, use those values, otherwise convert the exception to a `ProblemError` (extends `\Exception`), and then normalize the `ProblemError`.

src/Metadata/Resource/Factory/StaticResourceNameCollectionFactory.php
soyuka: I think that this will get removed as its not used? It is kind of static as it declares classes within the ResourceNameCollectionFactory.

soyuka:
@nesl247 yes there is one: https://github.com/api-platform/core/pull/5433/files#diff-877c6e433ecaadc292cdac7d6b42f139dc48e86ab831a8fc489e7976a908a630R57

> which is how the user is going to populate the type, title, detail, instance, etc.

We'll implement that in a second PR, as you can see the `ProblemError` is an `ApiResource`. A user that wants to extend this will be able to declare its own `ProblemError` and fill that as he wants. For now I'm not a huge fan of the "routing" (https://github.com/api-platform/core/pull/5433/files#diff-0c4b94869c82da23ed2573571540cf3f511737767c8f8a3462dc4106f6a16421R78) as if you want multiple `ProblemError` resources we'll need to find a way to use the correct one.

src/Action/ExceptionAction.php
JacquesDurand: Is this considered breaking to add a union type ? I'm also not sure how this class could be directly used by users ?
One problem with your suggestion is that in `ErrorListener::onControllerArgument`, I need to check the types of the controller argument (with Reflection) to transform the `\Throwable` in an instance of `ErrorExceptionInterface`.
I am definitely open to other possibilities though

src/Action/ExceptionAction.php
JacquesDurand: You're most probably right, I wasn't too sure of how to hydrate the 'instance' part of the problem+json response, this is just a "filler" part while waiting for a global opinion on this

src/ApiResource/Error.php
JacquesDurand: I was hesitant about this, should we (or not) allow users to extend it for their custom Errors ?

src/Problem/Serializer/ErrorNormalizer.php
JacquesDurand: Yeah I think adding 'status' and 'instance' breaks the current return responses, I should try to add more checks about the existence of the `extraProperties: ['hydra_errors' => true])` for the current request

nesl247:
> @nesl247 yes there is one: [#5433 (files)](https://github.com/api-platform/core/pull/5433/files#diff-877c6e433ecaadc292cdac7d6b42f139dc48e86ab831a8fc489e7976a908a630R57)
> 
> > which is how the user is going to populate the type, title, detail, instance, etc.
> 
> We'll implement that in a second PR, as you can see the `ProblemError` is an `ApiResource`. A user that wants to extend this will be able to declare its own `ProblemError` and fill that as he wants. For now I'm not a huge fan of the "routing" ([#5433 (files)](https://github.com/api-platform/core/pull/5433/files#diff-0c4b94869c82da23ed2573571540cf3f511737767c8f8a3462dc4106f6a16421R78)) as if you want multiple `ProblemError` resources we'll need to find a way to use the correct one.

I missed that because it's not in the `ProblemError` class, which doesn't make sense to me why it isn't.

src/Symfony/EventListener/ErrorListener.php
nesl247: I didn't need to do this to get the original exception in my own implementation.

```--- /dev/null
+++ ../src/Action/ExceptionAction.php
@@ -20,8 +20,10 @@
 use ApiPlatform\Util\OperationRequestInitiatorTrait;
 use ApiPlatform\Util\RequestAttributesExtractor;
 use Symfony\Component\ErrorHandler\Exception\FlattenException;
+use Symfony\Component\HttpFoundation\Exception\RequestExceptionInterface;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 
 /**
@@ -46,12 +48,20 @@
     /**
      * Converts an exception to a JSON response.
      */
-    public function __invoke(FlattenException $exception, Request $request): Response
+    public function __invoke(\Throwable $exception, Request $request): Response
     {
         $operation = $this->initializeOperation($request);
-        $exceptionClass = $exception->getClass();
-        $statusCode = $exception->getStatusCode();
+        $exceptionClass = $exception::class;
+        $statusCode = 500;
+        $headers = [];
 
+        if ($exception instanceof HttpExceptionInterface) {
+            $statusCode = $exception->getStatusCode();
+            $headers = $exception->getHeaders();
+        } elseif ($exception instanceof RequestExceptionInterface) {
+            $statusCode = 400;
+        }
+
         $exceptionToStatus = array_merge(
             $this->exceptionToStatus,
             $operation ? $operation->getExceptionToStatus() ?? [] : $this->getOperationExceptionToStatus($request)
@@ -65,7 +75,6 @@
             }
         }
 
-        $headers = $exception->getHeaders();
         $format = ErrorFormatGuesser::guessErrorFormat($request, $this->errorFormats);
         $headers['Content-Type'] = sprintf('%s; charset=utf-8', $format['value'][0]);
         $headers['X-Content-Type-Options'] = 'nosniff';
```

This was all of the change that I needed to be able to get the actual exception.

JacquesDurand:
> I missed that because it's not in the ProblemError class, which doesn't make sense to me why it isn't.

@nesl247 Imo the `type` is going to be inferred from the `ProblemError` ApiResource itself, e.g. :

```php
#[NotExposed(uriTemplate: '/problems/out-of-credit')]
class OutOfCreditError extends ProblemError
{
}
```
should directly produce a response with
```json
{
    "type": "https://example.com/problems/out-of-credit"
...
}
```
Also this PR is mainly a draft prototype open to discussion, a lot might still change :)

nesl247:
This is why I would scope this to using a `urn` by default, and letting the `URI` be manually set if a developer wants to link to a specific page somewhere. If it was wanted to support hosting this as part of api-platform, then I would do that as a separate feature.

https://gist.github.com/nesl247/3ae6b1f9575e4dcd0960ca4a9016434e - this is a gist of how we implemented it. It could be extended to having a mapper if you need to control third party exceptions.

src/Action/ExceptionAction.php
soyuka: As long as a class  is not marked `@internal` you can't change a public method argument type (https://symfony.com/doc/current/contributing/code/bc.html). If we can't, you need to deprecate this controller and create a new one.

src/Action/ExceptionAction.php
soyuka: Can you show me an example, I'm not sure I understand.

src/ApiResource/Error.php
soyuka: not the resource itself no

soyuka:
Type is also an URI here so no problem on that. About the specification by end users, we want to handle this in another PR but basically you'd need to implement your own `ErrorResource`. 

@JacquesDurand for the "routing" or "mapping" can't an exception be mapped with `ErrorResource`?

nesl247:
> Type is also an URI here so no problem on that. About the specification by end users, we want to handle this in another PR but basically you'd need to implement your own `ErrorResource`.
> 
> @JacquesDurand for the "routing" or "mapping" can't an exception be marked with the `ErrorResource` attribute ?

I still think there is some confusion here @soyuka, but I think @JacquesDurand understands it.

There's 2 parts to the RFC, the response and the documentation endpoint (what type is supposed to point to if not a URN). The first part, which IMO should be the main focus, is the error response. The current implementation conflates the two. The error response is the only thing that really needs to be implemented by api-platform.

Having a default `ErrorResource` IMO will be 99% useless IMO, as the point is for human readable documentation. If the default is just "Unknown error", there is no point to having said documentation. Also as I mentioned earlier, dealing with the human readable documentation aspect of this will bring a host of other problems regarding appearance customization (we need twig, etc.). Also, I would imagine most uses would rather have this link to a knowledge base or something of the sort.

Also, the `ErrorResource` should only be a path and return some sort of human readable documentation. It should not have `status`, `detail`, `instance`, etc, as the type is not supposed to change, as it's an `id` of the "problem" "type", not of the "problem" "instance".

I'm happy to discuss this more in Slack as well if that would be easier. I'm `nesl247` in the `Symfony Devs` Slack community.

src/ApiResource/ProblemError.php
```diff
@@ -0,0 +1,61 @@
...
+
+        return $problem;
+    }
+
+    public function getInstance(): ?string
+    {
+        return $this->instance;
+    }
+
+    public function setInstance(?string $instance): void
```
nesl247, line 47: Wouldn't it be better to make the object immutable either by never allowing it to be changed, or to use `withers` instead?

src/ApiResource/ProblemError.php
```diff
@@ -0,0 +1,61 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\ApiResource;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ErrorResource;
+
+#[ErrorResource(uriTemplate: '/problems/{status}', provider: 'api_platform.state_provider.default_problem_error')]
+class ProblemError extends \Exception
```
soyuka, line 20:
```suggestion
final class ProblemError extends \Exception
```

src/ApiResource/Error.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\ApiResource;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ErrorResource;
+
+#[ErrorResource(uriTemplate: '/errors/{statusCode}', provider: 'api_platform.state_provider.default_error')]
+class Error extends \Exception
```
soyuka, line 20:
```suggestion
final class Error extends \Exception
```

src/Action/ExceptionAction.php
```diff
@@ -51,10 +61,13 @@ public function __invoke(FlattenException $exception, Request $request): Respons
         $operation = $this->initializeOperation($request);
         $exceptionClass = $exception->getClass();
         $statusCode = $exception->getStatusCode();
+        $originalException = $request->attributes->get('exception');
+        $errorOperation = $this->initializeOperationForException($originalException);
```
soyuka, line 65: not a huge fan we can do without that

src/State/DefaultProblemErrorProvider.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): ProblemError
+    {
+        $exception = $context['previous_data'];
+
+        $problem = ProblemError::createFromException($exception, $uriVariables['statusCode']);
+        $identifiers = [
+            'uri_variables' => $problem->status,
```
soyuka, line 32:
```suggestion
            'uri_variables' => ['status' => $problem->status],
```

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -247,6 +249,9 @@ private function registerMetadataConfiguration(ContainerBuilder $container, arra
     {
         [$xmlResources, $yamlResources] = $this->getResourcesToWatch($container, $config);
 
+        // TODO only when hydra
+        $container->setParameter('api_platform.static_resources', $this->getStaticResources());
```
soyuka, line 253: rename that factory to `ClassNameResourceNameCollectionFactory`

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -24,13 +27,34 @@
...
+        $this->resourceMetadataCollectionFactory = $resourceMetadataCollectionFactory;
+    }
+
     protected function duplicateRequest(\Throwable $exception, Request $request): Request
     {
         $dup = parent::duplicateRequest($exception, $request);
+        $operation = $this->initializeOperationForException($exception);
 
         if ($request->attributes->has('_api_operation')) {
             $dup->attributes->set('_api_operation', $request->attributes->get('_api_operation'));
```
soyuka, line 49: should be set to the exception operation

src/Symfony/EventListener/ExceptionListener.php
```diff
@@ -39,6 +39,7 @@ public function onKernelException(ExceptionEvent $event): void
         ) {
             return;
         }
+        $event->allowCustomResponseCode();
```
soyuka, line 42: comment to explain please

src/Symfony/EventListener/ReadListener.php
```diff
@@ -60,6 +65,21 @@ public function onKernelRequest(RequestEvent $event): void
...
+            $operation = $this->initializeOperationForException($exception);
+            if (null === $operation) {
+                $format = ErrorFormatGuesser::guessErrorFormat($request, [
+                    'jsonproblem' => ['mime_types' => ['application/problem+json']],
+                    'jsonld' => ['mime_types' => ['application/ld+json']],
+                ]);
+                $operation = match ($format['key']) {
+                    'jsonproblem' => $this->initializeOperationForException(ProblemError::createFromException($exception, $exception->getCode())),
+                    default => $this->initializeOperationForException(Error::createFromException($exception, $exception->getCode())),
+                };
```
soyuka, line 79: Format negotiation shouldn't be here. Specify the format of each error resource (ProblemError and Error) and let the router do this.

src/Symfony/EventListener/RespondListener.php
```diff
@@ -52,6 +52,9 @@ public function onKernelView(ViewEvent $event): void
         $request = $event->getRequest();
         $operation = $this->initializeOperation($request);
 
+        if ($exception = $request->attributes->get('exception')) {
+            $operation = $request->attributes->get('_exception_operation');
```
soyuka, line 56: if we set `operation` I don't think we need this.

src/Util/OperationExceptionInitiatorTrait.php
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        try {
+            /* @var HttpOperation */
+            return $this->resourceMetadataCollectionFactory->create($exception::class)->getOperation();
+        } catch (OperationNotFoundException|ResourceClassNotFoundException $e) {
+            return null;
+        }
+    }
+}
```
soyuka, line 41: remove this

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -24,14 +37,113 @@
...
+
+        if ($exception instanceof IdentifiedExceptionInterface) {
+            $dup->attributes->set('_identified_exception', $exception);
+        }
+        if (ProblemError::class === $resourceClass) {
+            $dup->attributes->set('_identified_exception', ProblemError::createFromException($exception, $statusCode));
+        }
+        if (Error::class === $resourceClass) {
+            $dup->attributes->set('_identified_exception', Error::createFromException($exception, $statusCode));
         }
```
soyuka, line 107: Pour les exceptions qui ne sont pas des resources, faudra essayer de les retransmettre et d'avoir un provider  generique d'exception ? que faire du content negotiation ? ptet transformer plutot ici en Error (resource exception).

src/Symfony/EventListener/ErrorListener.php
```diff
@@ -24,14 +37,113 @@
...
+            }
+        }
+        if ($exception instanceof HttpExceptionInterface) {
+            $statusCode = $exception->getStatusCode();
+        } elseif ($exception instanceof RequestExceptionInterface) {
+            $statusCode = 400;
+        }
+
+        $statusCode ??= 500;
+
```
soyuka, line 96: 1) extraire les uriVariables ($operation->getUriVariables) comme fait l'**identifiers extractor**
2) mettre ca dans $request->attributes->set($parameterName, $value)

src/ApiResource/Error.php
```diff
@@ -0,0 +1,41 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\ApiResource;
+
+use ApiPlatform\Exception\IdentifiedExceptionInterface;
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ErrorResource;
+
+#[ErrorResource(uriTemplate: '/errors/{statusCode}', provider: 'api_platform.state_provider.default_error')]
```
soyuka, line 20:
```suggestion
#[ErrorResource(uriTemplate: '/errors/{statusCode}', provider: 'api_platform.state_provider.default_error', status: xxx)]
```

src/State/DefaultProblemErrorProvider.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): ProblemError
+    {
+        $exception = $context['previous_data'];
+
+        $problem = ProblemError::createFromException($exception, $uriVariables['status']);
+        $identifiers = [
+            'uri_variables' => ['status' => $problem->getStatus()],
+        ];
+        $problem->setType($this->iriConverter->getIriFromResource($problem, context: $identifiers));
```
soyuka, line 34: isn't that done by the normalizer already?

PR #5434, opened 2023-03-06
feat: api-platform/openapi component split tree

soyuka:


src/OpenApi/Command/OpenApiCommand.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\OpenApi\Command;
```
alanpoulain, line 14: Why in a `Command` namespace? How do we do if we want a Laravel command too?

soyuka: let's handle that when we implement the Laravel command, my guess is that the code will be pretty similar and we shouldn't have 2 classes doing the same thing anyways.

.github/workflows/ci.yml
```diff
@@ -186,6 +186,46 @@ jobs:
...
+      - name: Run OpenApi tests
+        working-directory: src/OpenApi
+        run: |
+            composer update
+            vendor/bin/simple-phpunit
+      - name: Run JsonSchema tests
+        working-directory: src/JsonSchema
+        run: |
+            composer update
+            vendor/bin/simple-phpunit
```
soyuka, line 226: yes we can use a matrix here let's keep that for later

Issue #5303, opened 2023-01-03
[Psalm] ProviderInterface does not allow for arrays to be returned


nesl247:
**API Platform version(s) affected**: 3.0.7

**Description**  
In https://github.com/api-platform/core/blob/3a845f1fb0ee80d6a8dc547a2aa97633fcda0830/src/State/ProviderInterface.php#L23 the type is defined to be an object. Because of this, when running psalm, it complains if a provider returns an array, given that an array is not an object.

```
ERROR: InvalidTemplateParam - src/TestProvider.php:13:31 - Extended template param T expects type object, type array<array-key, mixed> given (see https://psalm.dev/183)
class TestProvider implements ProviderInterface
```

**How to reproduce**  
```
<?php

declare(strict_types=1);

namespace App;

use ApiPlatform\Metadata\Operation;
use ApiPlatform\State\ProviderInterface;

/**
 * @implements ProviderInterface<array>
 */
class TestProvider implements ProviderInterface
{
    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
    {
        return [];
    }
}
```
**Possible Solution**  
Change it to `T of object|array`. I'm not sure if this is incorrect, but I think it's valid, and it does fix the issue.

**Additional Context**

soyuka:
@tyx can I have your insight on this?

soyuka:
related https://github.com/api-platform/core/pull/5443

nesl247:
@soyuka this is still an issue in `v3.3.12` using phpstan.

PR #5443, opened 2023-03-08
fix: return type should be iterable

develth:
... to match for example CollectionProvideres return type on pagination

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/State/ProviderInterface.php
+++ b/src/State/ProviderInterface.php
@@ -29,5 +29,5 @@ interface ProviderInterface
      *
      * @return T|Pagination\PartialPaginatorInterface<T>|iterable<T>|null
      */
-    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null;
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|iterable|null;
 }


src/State/ProviderInterface.php
```diff
@@ -29,5 +29,5 @@ interface ProviderInterface
      *
      * @return T|Pagination\PartialPaginatorInterface<T>|iterable<T>|null
      */
-    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null;
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|iterable|null;
```
soyuka, line 32: I think that this is breaking compatibility

develth: But regarding this it's not compatible

https://github.com/api-platform/core/blob/bde59ba121bcf99f4207a34dd5d09da8a6b2c5cc/src/Doctrine/Odm/State/CollectionProvider.php#L41

develth:
@soyuka 

> FYI object|array is identical to object|iterable even in PHP 8.0 and 8.1. :)

Yes, but IMHO `object|iterable` is not identical to `object|array`. 
And e.g. `QueryResultCollectionExtensionInterface` returns `iterable` as type, which would not match to return type of `object|array|null`

soyuka:
Mhh actually it looks like a PHP issue https://github.com/php/php-src/commit/bb79ef775b91b53f5cba2be81b44d8b65a869e28 

> `Iterable is a built-in compile time type alias for array|Traversable. From its introduction in PHP 7.1.0 and prior to PHP 8.2.0, iterable was a built-in pseudo-type that acted as the aforementioned type alias and can be used as a type declaration. An iterable type can be used in foreach and with yield from within a generator.`

reopening this

soyuka:
We need to find out what PHP's choice is for 8.3 if we should revert or not . but basically `iterable` and `array` are compatible so it doesn't change much

nesl247:
This is still an issue.

nesl247:
@soyuka I didn't see any changes made for this. Were there?

nesl247:
> why is this a problem?

https://github.com/api-platform/core/issues/5303

PR #5444, opened 2023-03-08
fix(serializer): propertyFilter should apply to arrays as well 

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| License       | MIT

`PropertyFilter` does not presently work with arrays of objects which are not `ApiResource`: it just always serializes these objects as null arrays.

FYI I've tested with `ArrayCollection` of objects which are not `ApiResource` and it works like a charm

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -640,6 +640,13 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return $this->serializer->normalize($attributeValue, $format, $childContext);
         }
 
+        if ($type && 'array' === $type->getBuiltinType()) {
+            $childContext = $this->createChildContext($context, $attribute, $format);
+            unset($childContext['iri'], $childContext['uri_variables']);
+
+            return $this->serializer->normalize($attributeValue, $format, $childContext);
+        }
+
         return $this->serializer->normalize($attributeValue, $format, $context);
     }
 
--- a/features/filter/property_filter.feature
+++ b/features/filter/property_filter.feature
@@ -18,3 +18,11 @@ Feature: Set properties to include
     And the JSON node "alias" should be equal to "Alias #0"
     And the JSON node "relatedDummy.name" should be equal to "RelatedDummy #1"
     And the JSON node "relatedDummies" should not exist
+
+  Scenario: Test property filter on not resource relations
+    When I send a "GET" request to "/dummy-with-array-of-objects/1?properties[notResourceObject][]=foo&properties[arrayOfNotResourceObjects][]=bar"
+    Then the JSON node "notResourceObject.foo" should be equal to "foo"
+    And the JSON node "notResourceObject.bar" should not exist
+    And the JSON node "arrayOfNotResourceObjects[0].foo" should not exist
+    And the JSON node "arrayOfNotResourceObjects[0].bar" should be equal to "bar"
+    And the JSON node "id" should not exist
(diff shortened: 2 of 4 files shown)


tests/Behat/DoctrineContext.php
```diff
@@ -13,6 +13,7 @@
 
 namespace ApiPlatform\Tests\Behat;
 
+use ApiPlatform\Tests\Fixtures\NotAResource;
```
soyuka, line 16: can you create another Dto as this one is already used thanks!

tests/Fixtures/TestBundle/Dto/NonResourceClass.php
```diff
@@ -15,7 +15,9 @@
 
 class NonResourceClass implements \Stringable
 {
-    public ?string $foo = null;
+    public function __construct(public ?string $foo = null, public ?string $bar = null)
```
soyuka, line 18: we try not to change an existing fixture as it may have unrelated impact

tests/Fixtures/TestBundle/Entity/DummyWithArrayOfNotResourceObject.php
```diff
@@ -0,0 +1,46 @@
...
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
+    private $id;
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function getNotResourceObject(): NonResourceClass
```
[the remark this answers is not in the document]
nikophil: yes but... with objects from doctrine, it is a little bit more tricky to get an array of objects like this - would you accept if I use a `OneToMany` and I expose the inner array? otherwise we'd need some an  output DTO

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -640,6 +640,13 @@ protected function getAttributeValue(object $object, string $attribute, string $
             return $this->serializer->normalize($attributeValue, $format, $childContext);
         }
 
+        if ($type && $type->getBuiltinType() === 'array') {
+            $childContext = $this->createChildContext($context, $attribute, $format);
+            unset($childContext['iri'], $childContext['uri_variables']);
```
soyuka, line 645: todo: check if we shouldn't remove `operation` also

nikophil: could you explain me what choices lead to `unset()` these contexts?
because we're serializing "non api resource" objects, there would be no need to have the operation available in the context?

soyuka: because it'll re-use the old operation if any when creating an IRI

nikophil: unless I'm misunderstanding something, no IRI should be generated, since we're dealing with non-apiresource objects?

soyuka: Yes they have Skolem IRIs unless disabled by using the `genId: false` attribute.

nikophil:
:tada:

temp:
This MR broke our code in unexpected ways, we now receive "Unable to generate an IRI for the item of type X" in our unit tests.
Our resource "Wheelset" has a property of type array<Wheel>, and Wheel has a property of type WheelManufacturer, which is a resource itself.
The wheel-array is now normalized within the if-block of this MR in AbstractItemNormalizer, and leads deep down in IriConverter and IdentifiersExtractor to the above error. We haven't got a clue what to change. When we comment out the new if-block, everything works as expected. Seems like with the new if-block it tried to generate an IRI, without it it doesn't try to generate an IRI for the nested resource.

PR #5453, opened 2023-03-13
fix: avoid ReflectionEnum usage

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| License       | MIT

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -117,14 +117,16 @@ final class TypeFactory implements TypeFactoryInterface
             ];
         }
         if (!$this->isResourceClass($className) && is_a($className, \BackedEnum::class, true)) {
-            $rEnum = new \ReflectionEnum($className);
-            $enumCases = array_map(static fn (\ReflectionEnumBackedCase $rCase) => $rCase->getBackingValue(), $rEnum->getCases());
+            $enumCases = array_map(static fn (\BackedEnum $enum): string|int => $enum->value, $className::cases());
+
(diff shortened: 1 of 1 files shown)


src/JsonSchema/TypeFactory.php
```diff
@@ -117,14 +117,16 @@ private function getClassType(?string $className, bool $nullable, string $format
             ];
         }
         if (!$this->isResourceClass($className) && is_a($className, \BackedEnum::class, true)) {
-            $rEnum = new \ReflectionEnum($className);
-            $enumCases = array_map(static fn (\ReflectionEnumBackedCase $rCase) => $rCase->getBackingValue(), $rEnum->getCases());
+            $enumCases = array_map(static fn(\BackedEnum $enum): string|int => $enum->value, $className::cases());
+
+            $type = is_string($enumCases[0] ?? '') ? 'string' : 'int';
```
nikophil, line 122: although it is not forbidden in PHP, it would be very weird to have an enum with 0 cases... not sure we should do something special in this case

soyuka: Can't we use `bool` or any kind of type for that matter?

soyuka:
needs php-cs-fixer

nikophil:
linting ok :heavy_check_mark: 

just FYI, here are the prototypes of `try()` and `tryFrom()` in php doc
```php
public static [from](https://www.php.net/manual/fr/backedenum.from.php)(int|string $value): static
public static [tryFrom](https://www.php.net/manual/fr/backedenum.tryfrom.php)(int|string $value): ?static
```

PR #5462, opened 2023-03-15
fix(metadata): add string[] as allowed param type

louismariegaborit:
Fix rule to accept only string in array like as mentioned in doc.


louismariegaborit:
@soyuka PR rebased after target updated

PR #5463, opened 2023-03-15
feat: declare ApiResource inside a symfony controller

soyuka:
After seeing https://github.com/symfony/symfony/pull/49518, we want to provide users a way of using our `ApiResource` within Symfony controllers. We have already all the stack needed to build APIs:

- security
- validation
- data retrieval
- deserialization
- serialization

This new feature allows to declare an `ApiResource` on a Controller argument like this:

```php
final class ApiResourceController
{
    #[Route(path: '/api_resource_controller/{id}', format: 'jsonld')]
    public function index(#[ApiResource(provider: [self::class, 'getData'])] ControllerResource $controllerResource): ControllerResource
    {
        return $controllerResource;
    }

    public static function getData(Operation $operation, array $uriVariables = [], array $context = []): ControllerResource
    {
        $data = new ControllerResource();
        $data->id = $uriVariables['id'];
        $data->name = 'soyuka';

        return $data;
    }
}
```

### DIFF
--- a/features/symfony/controller_api_resource.feature
+++ b/features/symfony/controller_api_resource.feature
@@ -1,0 +1,54 @@
+Feature: ApiResource within Symfony controller
+  In order to use an ApiResource inside a Symfony controller 
+  As a developer
+  I should be able to use the API
+
+  Scenario: Send a GET request
+    When I send a "GET" request to "/api_resource_controller/1"
+    Then print last JSON response
+    Then the response status code should be 200
+    And the response should be in JSON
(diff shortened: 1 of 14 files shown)


soyuka:
needs https://github.com/api-platform/core/pull/4655 and some thoughts that'll be published tomorrow to improve this overall design as right now it is bad imho and doesn't bring anything good.

soyuka:
This needs #4655 as it'll ease this process a lot

soyuka:
I have something else for this see #4655

PR #5465, opened 2023-03-16
Added checking of the allowCreate option to set the status

### LINKED ISSUE #5417: PUT return 201 without ReadListener enabled
**API Platform version(s) affected**: 3.1.2

**Description**  

Since the merge of https://github.com/api-platform/core/pull/4996 we are in a difficult situation without efficient workaround ATM.

In our API and because we are full CQRS, we don't use `ReadListener` for `PUT` operation.
We are ok that we are living on the edge but we are pretty happy with our combination of APIP and our architecture. And latest change in APIP3 were so great for us.

But with those lines added : 
https://github.com/api-platform/core/blob/9b4b58ca0113a2645b58a116fe9e4bf200df8aa3/src/Symfony/EventListener/RespondListener.php#L93-L95

All our `PUT` operation return now `201`... And we have no way to get back the old behavior via the new `standard_put` properties.

**How to reproduce**  

Create a PUT operation without ReadListener

**Possible Solution**  

- We could update manually our `PUT` operations to manually add status: 200. But I should admit I'm scared that we could forget one (now or in the future) that could lead to break someting in prod.
- Is there a world where the lines updated above could rely also to an additionnal condition where we could act on it ? Like allowCreate ?

We are open to any solid solution that would involve to add some custom code in our codebase of course.

Zowac:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5417 
| License       | MIT

Fix #5417

### DIFF
--- a/features/main/custom_put.feature
+++ b/features/main/custom_put.feature
@@ -1,0 +1,29 @@
+Feature: Spec-compliant PUT support
+  As a client software developer
+  I need to be able to create or replace resources using the PUT HTTP method
+
+  @createSchema
+  @!mongodb
+  Scenario: Get a correct status code when updating a resource that is not allowed to read nor to create
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/custom_puts/1" with body:
+    """
+    {
+      "foo": "a",
+      "bar": "b"
+    }
+    """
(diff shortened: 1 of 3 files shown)


tyx:
Thank you !

PR #5469, opened 2023-03-17
fix(jsonSchema): class shortname

Zowac:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| License       | MIT

See https://github.com/api-platform/core/pull/5455#issuecomment-1469699042

### DIFF
--- a/src/Metadata/Resource/ResourceMetadataCollection.php
+++ b/src/Metadata/Resource/ResourceMetadataCollection.php
@@ -89,9 +89,9 @@ final class ResourceMetadataCollection extends \ArrayObject
         }
 
         // Idea:
-        // if ($metadata) {
-        //     return (new class extends HttpOperation {})->withResource($metadata);
-        // }
+//         if ($metadata) {
+//             return (new class extends HttpOperation {})->withResource($metadata);
+//         }
 
         $this->handleNotFound($operationName, $metadata);
     }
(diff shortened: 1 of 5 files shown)


src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -102,7 +102,7 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
     private function handleNotFound(string $operationName, ?ApiResource $metadata): void
     {
         // Hide the FQDN in the exception message if possible
-        $shortName = $metadata ? $metadata->getShortName() : $this->resourceClass;
+        $shortName = $metadata?->getShortName() ? $metadata->getShortName() : $this->resourceClass;
```
soyuka, line 105: Why this change?

soyuka:
Can you squash your commits and rename:

```
fix(jsonschema): find the related operation instead of assuming one
```

src/Metadata/Resource/ResourceMetadataCollection.php
Zowac: `getShortName` can return null, if we don't check the method it happens that we don't have the class that is impacted by the error because the shortName is empty

soyuka:
Thanks!

LoicBoursin:
Thanks @Zowac & @soyuka !! 😍

jamesisaac:
Hi,

This PR seems to have caused a regression in some cases of JsonSchema generation.  I've created a minimal repro here: #5626

Issue #5452, opened 2023-03-11
Union types for property  


syl20b:
**Description**  

Hi,

It seems that only the first property type is checked in ApiPlatform\Serializer\AbstractItemNormalizer::createAndValidateAttributeValue() method.
 
It could be interesting to check the multiples types of each property. 

**How to reproduce**  

Given the following Resource: 

```php
<?php

namespace App\ApiResource;

use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Post;
use App\Dto\FooInput;

#[ApiResource(
    operations: [
        new Post(
            uriTemplate: '/foo',
        ),
    ],
)]
class Foo
{
    public string|int|null $bar;
}
```

When I execute the request with a string 

```
curl -X 'POST' \
  'http://127.0.0.1:8000/api/foo' \
  -H 'accept: application/ld+json' \
  -H 'Content-Type: application/ld+json' \
  -d '{
  "bar": "string"
}'
```

then the response is OK

But when I execute the request with an integer 

```
curl -X 'POST' \
  'http://127.0.0.1:8000/api/foo' \
  -H 'accept: application/ld+json' \
  -H 'Content-Type: application/ld+json' \
  -d '{
  "bar": 1
}'
```

Then I've got the following 400 error:

```json
{
  "@context": "/api/contexts/Error",
  "@type": "hydra:Error",
  "hydra:title": "An error occurred",
  "hydra:description": "The type of the \"bar\" attribute must be \"string\", \"integer\" given.",
  "trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
      "file": "/home/Workspace/api-platform/vendor/symfony/serializer/Exception/NotNormalizableValueException.php",
      "line": 31,
      "args": []
    },
    {
      "namespace": "Symfony\\Component\\Serializer\\Exception",
      "short_class": "NotNormalizableValueException",
      "class": "Symfony\\Component\\Serializer\\Exception\\NotNormalizableValueException",
      "type": "::",
      "function": "createForUnexpectedDataType",
      "file": "/home/Workspace/api-platform/vendor/api-platform/core/src/Serializer/AbstractItemNormalizer.php",
      "line": 467,
      "args": []
    },
    {
      "namespace": "ApiPlatform\\Serializer",
      "short_class": "AbstractItemNormalizer",
      "class": "ApiPlatform\\Serializer\\AbstractItemNormalizer",
      "type": "->",
[...]

PR #5470, opened 2023-03-17
Support union/intersect types

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5452
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -213,17 +213,23 @@ final class FieldsBuilder implements FieldsBuilderInterface, FieldsBuilderEnumIn
                     'denormalization_groups' => $operation->getDenormalizationContext()['groups'] ?? null,
                 ];
                 $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $property, $context);
+                $propertyTypes = $propertyMetadata->getBuiltinTypes();
 
                 if (
-                    null === ($propertyType = $propertyMetadata->getBuiltinTypes()[0] ?? null)
+                    !$propertyTypes
                     || (!$input && false === $propertyMetadata->isReadable())
                     || ($input && $operation instanceof Mutation && false === $propertyMetadata->isWritable())
                 ) {
                     continue;
                 }
 
-                if ($fieldConfiguration = $this->getResourceFieldConfiguration($property, $propertyMetadata->getDescription(), $propertyMetadata->getDeprecationReason(), $propertyType, $resourceClass, $input, $operation, $depth, null !== $propertyMetadata->getSecurity())) {
-                    $fields['id' === $property ? '_id' : $this->normalizePropertyName($property, $resourceClass)] = $fieldConfiguration;
+                // guess union/intersect types: check each type until finding a valid one
+                foreach ($propertyTypes as $propertyType) {
+                    if ($fieldConfiguration = $this->getResourceFieldConfiguration($property, $propertyMetadata->getDescription(), $propertyMetadata->getDeprecationReason(), $propertyType, $resourceClass, $input, $operation, $depth, null !== $propertyMetadata->getSecurity())) {
+                        $fields['id' === $property ? '_id' : $this->normalizePropertyName($property, $resourceClass)] = $fieldConfiguration;
+                        // stop at the first valid type
+                        break;
+                    }
                 }
             }
         }
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -318,44 +318,42 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
             return $jsonldContext['@type'];
         }
 
-        // TODO: 3.0 support multiple types, default value of types will be [] instead of null
-        $type = $propertyMetadata->getBuiltinTypes()[0] ?? null;
-        if (null === $type) {
-            return null;
-        }
+        $types = $propertyMetadata->getBuiltinTypes() ?? [];
 
-        if ($type->isCollection() && null !== $collectionType = $type->getCollectionValueTypes()[0] ?? null) {
-            $type = $collectionType;
-        }
+        foreach ($types as $type) {
+            if ($type->isCollection() && null !== $collectionType = $type->getCollectionValueTypes()[0] ?? null) {
+                $type = $collectionType;
+            }
 
-        switch ($type->getBuiltinType()) {
-            case Type::BUILTIN_TYPE_STRING:
-                return 'xmls:string';
-            case Type::BUILTIN_TYPE_INT:
-                return 'xmls:integer';
-            case Type::BUILTIN_TYPE_FLOAT:
-                return 'xmls:decimal';
-            case Type::BUILTIN_TYPE_BOOL:
-                return 'xmls:boolean';
-            case Type::BUILTIN_TYPE_OBJECT:
-                if (null === $className = $type->getClassName()) {
-                    return null;
-                }
-
-                if (is_a($className, \DateTimeInterface::class, true)) {
-                    return 'xmls:dateTime';
-                }
-
-                if ($this->resourceClassResolver->isResourceClass($className)) {
-                    $resourceMetadata = $this->resourceMetadataFactory->create($className);
-                    $operation = $resourceMetadata->getOperation();
-
-                    if (!$operation instanceof HttpOperation) {
-                        return "#{$operation->getShortName()}";
+            switch ($type->getBuiltinType()) {
+                case Type::BUILTIN_TYPE_STRING:
+                    return 'xmls:string';
+                case Type::BUILTIN_TYPE_INT:
+                    return 'xmls:integer';
+                case Type::BUILTIN_TYPE_FLOAT:
+                    return 'xmls:decimal';
+                case Type::BUILTIN_TYPE_BOOL:
+                    return 'xmls:boolean';
+                case Type::BUILTIN_TYPE_OBJECT:
+                    if (null === $className = $type->getClassName()) {
+                        return null;
                     }
 
-                    return $operation->getTypes()[0] ?? "#{$operation->getShortName()}";
-                }
+                    if (is_a($className, \DateTimeInterface::class, true)) {
+                        return 'xmls:dateTime';
+                    }
+
+                    if ($this->resourceClassResolver->isResourceClass($className)) {
+                        $resourceMetadata = $this->resourceMetadataFactory->create($className);
+                        $operation = $resourceMetadata->getOperation();
+
+                        if (!$operation instanceof HttpOperation) {
+                            return "#{$operation->getShortName()}";
+                        }
+
+                        return $operation->getTypes()[0] ?? "#{$operation->getShortName()}";
+                    }
+            }
         }
 
         return null;
@@ -463,11 +461,13 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
             'domain' => $prefixedShortName,
         ];
 
-        // TODO: 3.0 support multiple types, default value of types will be [] instead of null
-        $type = $propertyMetadata->getBuiltinTypes()[0] ?? null;
+        $types = $propertyMetadata->getBuiltinTypes() ?? [];
 
-        if (null !== $type && !$type->isCollection() && (null !== $className = $type->getClassName()) && $this->resourceClassResolver->isResourceClass($className)) {
-            $propertyData['owl:maxCardinality'] = 1;
+        foreach ($types as $type) {
+            if (!$type->isCollection() && (null !== $className = $type->getClassName()) && $this->resourceClassResolver->isResourceClass($className)) {
+                $propertyData['owl:maxCardinality'] = 1;
+                break;
+            }
         }
 
         $property = [
--- a/src/Hydra/Serializer/DocumentationNormalizer.php
+++ b/src/Hydra/Serializer/DocumentationNormalizer.php
@@ -310,7 +310,7 @@ final class DocumentationNormalizer implements NormalizerInterface, CacheableSup
     /**
      * Gets the range of the property.
      */
-    private function getRange(ApiProperty $propertyMetadata): ?string
+    private function getRange(ApiProperty $propertyMetadata): array|string|null
     {
         $jsonldContext = $propertyMetadata->getJsonldContext();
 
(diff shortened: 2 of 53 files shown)


src/JsonSchema/SchemaFactory.php
```diff
@@ -174,10 +188,9 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
             $propertySchema['externalDocs'] = ['url' => $iri];
         }
 
-        // TODO: 3.0 support multiple types
-        $type = $propertyMetadata->getBuiltinTypes()[0] ?? null;
+        $types = $propertyMetadata->getBuiltinTypes() ?? [];
 
-        if (!isset($propertySchema['default']) && !empty($default = $propertyMetadata->getDefault()) && (null === $type?->getClassName() || !$this->isResourceClass($type->getClassName()))) {
+        if (!isset($propertySchema['default']) && !empty($default = $propertyMetadata->getDefault()) && (!\count($types) || null === ($className = $types[0]->getClassName()) || !$this->isResourceClass($className))) {
```
soyuka, line 193: can't this condition be refactored to be called later when we compute `$valueSchema` and therefore reduce the number of call to the `\count($types)`?

src/JsonSchema/SchemaFactory.php
```diff
@@ -192,8 +205,26 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
         }
 
+        // never override the following keys if at least one is already set (by user or restrictions)
+        if (
+            0 < \count(
+                array_intersect(
+                    ['type', '$ref', 'anyOf', 'allOf', 'oneOf'],
+                    array_keys($propertySchema)
+                )
+            )
```
soyuka, line 215: I wish we could do something a little bit more optimized to avoid 3 function calls and a `count`:

```
if ($propertySchema['type'] ?? $propertySchema['$ref'] ?? $propertySchema['anyOf'] ?? $propertySchema['allOf'] ?? $propertySchema['oneOf'] ?? false) {
```

src/JsonSchema/SchemaFactory.php
```diff
@@ -192,8 +205,26 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
+            || [] === $types
```
soyuka, line 216: this should be first part of the condition as its less resource-consuming.

src/JsonSchema/SchemaFactory.php
```diff
@@ -210,15 +241,37 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
 
-            $valueSchema = $this->typeFactory->getType(new Type($builtinType, $type->isNullable(), $className, $isCollection, $keyType, $valueType), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
+            $valueSchema[] = $this->typeFactory->getType(new Type($builtinType, $type->isNullable(), $className, $isCollection, $keyType, $valueType), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
         }
 
-        if (\array_key_exists('type', $propertySchema) && \array_key_exists('$ref', $valueSchema)) {
-            $propertySchema = new \ArrayObject($propertySchema);
-        } else {
-            $propertySchema = new \ArrayObject($propertySchema + $valueSchema);
+        $valueSchema = array_unique($valueSchema, \SORT_REGULAR);
```
soyuka, line 247: not a huge fan of the call, we could name our schemas and just try to `isset` before setting the value?

src/JsonSchema/SchemaFactory.php
```diff
@@ -210,15 +241,37 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
+
+        // only one builtInType detected (should be "type" or "$ref")
+        if (1 === \count($valueSchema)) {
+            $schema->getDefinitions()[$definitionName]['properties'][$normalizedPropertyName] = new \ArrayObject($propertySchema + $valueSchema[0]);
```
soyuka, line 251: could be refactored to avoid the `count` call

src/JsonSchema/SchemaFactory.php
```diff
@@ -210,15 +241,37 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
+                // cannot detect union/intersect types
+                $composition = 'anyOf';
+                break;
+            case $reflectionType instanceof \ReflectionUnionType:
+                $composition = 'oneOf';
+                break;
+            case $reflectionType instanceof \ReflectionIntersectionType:
+                $composition = 'allOf';
+                break;
+        }
```
soyuka, line 269: as said earlier I think that this needs to be inside the Metadata, I have an even better suggestion. The idea is to create a `PropertySchemaMetadataFactory` that would fill the `$propertySchema` with the appropriate data. This way we could cache that information within property metadata and this class would be simplified a lot.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -59,8 +59,12 @@ public function create(string $resourceClass, string $property, array $options =
 
         $propertyMetadata = $this->transformReadWrite($propertyMetadata, $resourceClass, $property, $normalizationGroups, $denormalizationGroups);
 
-        if (!$this->isResourceClass($resourceClass) && ($builtinType = $propertyMetadata->getBuiltinTypes()[0] ?? null) && $builtinType->isCollection()) {
-            return $propertyMetadata->withReadableLink(true)->withWritableLink(true);
+        if (!$this->isResourceClass($resourceClass) && ($types = $propertyMetadata->getBuiltinTypes())) {
```
soyuka, line 62: you could declare the `$types = $propertyMetadata->getBuiltinTypes()` outside of the `if` as it'll get used later if needed.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -108,41 +112,40 @@ private function transformLinkStatus(ApiProperty $propertyMetadata, array $norma
             return $propertyMetadata;
         }
 
-        // TODO: 3.0 support multiple types, default value of types will be [] instead of null
-        $type = $propertyMetadata->getBuiltinTypes()[0] ?? null;
+        $types = $propertyMetadata->getBuiltinTypes() ?? [];
```
soyuka, line 115: can be avoided as called line 63

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -108,41 +112,40 @@ private function transformLinkStatus(ApiProperty $propertyMetadata, array $norma
...
-            $type->isCollection() &&
-            $collectionValueType = $type->getCollectionValueTypes()[0] ?? null
-        ) {
-            $relatedClass = $collectionValueType->getClassName();
-        } else {
-            $relatedClass = $type->getClassName();
-        }
+            // if property is not a resource relation, don't set link status (as it would have no meaning)
+            if (null === $relatedClass || !$this->isResourceClass($relatedClass)) {
+                return $propertyMetadata;
```
soyuka, line 129: what if another `type` is a resource class? We should first compute every types before making a decision.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -729,122 +732,146 @@ private function createAttributeValue(string $attribute, mixed $value, string $f
...
-        if (null === $value && $type->isNullable()) {
-            return $value;
-        }
+            $collectionValueType = $type->getCollectionValueTypes()[0] ?? null;
 
-        $collectionValueType = $type->getCollectionValueTypes()[0] ?? null;
+            /* From @see AbstractObjectNormalizer::validateAndDenormalize() */
+            // Fix a collection that contains the only one element
+            // This is special to xml format only
+            if ('xml' === $format && null !== $collectionValueType && (!\is_array($value) || !\is_int(key($value)))) {
```
soyuka, line 748: @dunglas should we really support this? We discussed XML is not supporte-able by our normalizers at the symfony live. Maybe that deprecating the support of XML (w/ JSON by default) for 3.2 should be our target?

soyuka (changes requested):
Good start! 

To resume my thoughts I think that we should compute most of the type-inflected schema inside a `JsonSchema\Metadata\TypePropertyMetadataFactory` to reduce the work inside the SchemaFactory. Also move the `ReflectionType` checks there. 
Specifications like JSON-LD, HAL and JSON:API support mutliple types but we stop at the first one. 

Deprecating XML, CSV support from our normalizer is another task but reading these parts of the code reminded me of that so don't bother here. 

I want to suggest that our codebase should not handle types that are scalars **and** objects but we probably need to discuss this with more @api-platform/core-team members.

src/JsonSchema/SchemaFactory.php
vincentchalamon: Could be interesting to do the same for resource, like `SchemaResourceCollectionMetadataFactory`, with code from SchemaFactory::buildSchema.

WDYT?

vincentchalamon: Can't move all code to `SchemaPropertyMetadataFactory` because `$this->typeFactory->getType` requires `$format`, `$serializerContext` and `$schema` which are only available on SchemaFactory.

I suggest to move some code to `SchemaPropertyMetadataFactory`, but keep `$valueSchema` creation on SchemaFactory::buildPropertySchema to complete the property schema.

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -318,44 +318,42 @@ private function getRange(ApiProperty $propertyMetadata): ?string
...
-                return 'xmls:decimal';
-            case Type::BUILTIN_TYPE_BOOL:
-                return 'xmls:boolean';
-            case Type::BUILTIN_TYPE_OBJECT:
-                if (null === $className = $type->getClassName()) {
-                    return null;
-                }
+            switch ($type->getBuiltinType()) {
+                case Type::BUILTIN_TYPE_STRING:
+                    return 'xmls:string';
```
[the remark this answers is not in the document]
vincentchalamon: I couldn't find anything on the Hydra Spec about multiple-types support.

But some people are using `https://schema.org/rangeIncludes` for similar use-cases. WDYT?

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -213,17 +213,23 @@ public function getResourceObjectTypeFields(?string $resourceClass, Operation $o
...
                     continue;
                 }
 
-                if ($fieldConfiguration = $this->getResourceFieldConfiguration($property, $propertyMetadata->getDescription(), $propertyMetadata->getDeprecationReason(), $propertyType, $resourceClass, $input, $operation, $depth, null !== $propertyMetadata->getSecurity())) {
-                    $fields['id' === $property ? '_id' : $this->normalizePropertyName($property, $resourceClass)] = $fieldConfiguration;
+                // guess union/intersect types: check each type until finding a valid one
+                foreach ($propertyTypes as $propertyType) {
+                    if ($fieldConfiguration = $this->getResourceFieldConfiguration($property, $propertyMetadata->getDescription(), $propertyMetadata->getDeprecationReason(), $propertyType, $resourceClass, $input, $operation, $depth, null !== $propertyMetadata->getSecurity())) {
+                        $fields['id' === $property ? '_id' : $this->normalizePropertyName($property, $resourceClass)] = $fieldConfiguration;
+                        // stop at the first valid type
```
soyuka, line 230: @alanpoulain doesn't graphql support multiple types ?

src/Hydra/Serializer/DocumentationNormalizer.php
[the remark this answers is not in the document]
soyuka: just return an array

src/Hydra/Serializer/DocumentationNormalizer.php
```diff
@@ -463,11 +461,13 @@ private function getProperty(ApiProperty $propertyMetadata, string $propertyName
...
 
-        // TODO: 3.0 support multiple types, default value of types will be [] instead of null
-        $type = $propertyMetadata->getBuiltinTypes()[0] ?? null;
+        $types = $propertyMetadata->getBuiltinTypes() ?? [];
 
-        if (null !== $type && !$type->isCollection() && (null !== $className = $type->getClassName()) && $this->resourceClassResolver->isResourceClass($className)) {
-            $propertyData['owl:maxCardinality'] = 1;
+        foreach ($types as $type) {
+            if (!$type->isCollection() && (null !== $className = $type->getClassName()) && $this->resourceClassResolver->isResourceClass($className)) {
+                $propertyData['owl:maxCardinality'] = 1;
```
soyuka, line 468: This doesn't belong to the hydra specification, I suggest to remove it

src/JsonSchema/Metadata/Property/Factory/SchemaPropertyMetadataFactory.php
```diff
@@ -0,0 +1,268 @@
...
+                ], $type);
+            }
+
+            return $this->addNullabilityToTypeDefinition([
+                'type' => 'array',
+                'items' => $this->getType($subType, $readableLink),
+            ], $type);
+        }
+
+        return $this->addNullabilityToTypeDefinition($this->makeBasicType($type, $readableLink), $type);
```
soyuka, line 162: if (!$type->isCollection()) {
      return $this->addNullabilityToTypeDefinition($this->makeBasicType($type, $readableLink), $type);
    }

    $keyType = $type->getCollectionKeyTypes()[0] ?? null;
    $subType = ($type->getCollectionValueTypes()[0] ?? null) ?? new Type($type->getBuiltinType(), false, $type->getClassName(), false);

    if (null !== $keyType && Type::BUILTIN_TYPE_STRING === $keyType->getBuiltinType()) {
      return $this->addNullabilityToTypeDefinition([
          'type' => 'object',
          'additionalProperties' => $this->getType($subType, $readableLink),
      ], $type);
    }

    return $this->addNullabilityToTypeDefinition([
        'type' => 'array',
        'items' => $this->getType($subType, $readableLink),
    ], $type);
```

File: src/JsonSchema/Metadata/Property/Factory/SchemaPropertyMetadataFactory.php [hunk 15]
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    private function makeBasicType(Type $type, ?bool $readableLink = null): array
```
soyuka, line 165:
```suggestion
    private function getJsonSchemaType(Type $type, ?bool $readableLink = null): array
```

or

```suggestion
    private function typeToArray(Type $type, ?bool $readableLink = null): array
```

?

src/JsonSchema/Metadata/Property/Factory/SchemaPropertyMetadataFactory.php
```diff
@@ -0,0 +1,268 @@
...
+            return ['anyOf' => [$jsonSchema, 'type' => 'null']];
+        }
+
+        return [...$jsonSchema, ...[
+            'type' => \is_array($jsonSchema['type'])
+                ? array_merge($jsonSchema['type'], ['null'])
+                : [$jsonSchema['type'], 'null'],
+        ]];
+    }
+}
```
soyuka, line 268: I really like this, this moves away some of the complexity of our TypeFactory and SchemaFactory

src/JsonSchema/SchemaFactory.php
```diff
@@ -154,74 +152,48 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
-                $className = $valueType->getClassName();
+            $className = $valueType?->getClassName();
+            if (null === $className || !$this->isResourceClass($className)) {
+                continue;
             }
 
-            $valueSchema = $this->typeFactory->getType(new Type($builtinType, $type->isNullable(), $className, $isCollection, $keyType, $valueType), $format, $propertyMetadata->isReadableLink(), $serializerContext, $schema);
+            $subSchema = $this->buildSchema($className, $format, Schema::TYPE_OUTPUT, null, $subSchema, $serializerContext, false);
+            $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
+            unset($propertySchema['type']);
```
soyuka, line 192: can you explain why we need to unset this ?

soyuka (commented):
nice job on the metadata and schema factory part! few things left on the hydra support and I haven't read the part about validators / elasticsearch yet. We need to find out what to do about graphql thoug.

src/GraphQl/Type/FieldsBuilder.php
alanpoulain: If it's not an input type, yes. Otherwise no (see https://github.com/graphql/graphql-spec/pull/825).

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -279,32 +279,40 @@ private function getComponents(object $object, ?string $format, array $context):
...
-                'name' => $attribute,
-                'type' => $this->getResourceShortName($className),
-                'cardinality' => $isOne ? 'one' : 'many',
-            ];
+                $components['relationships'][] = $relation;
+                $isRelationship = true;
+            }
 
-            $components['relationships'][] = $relation;
+            // if all types are not relationships, declare it as an attribute
```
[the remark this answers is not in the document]
vincentchalamon: Better something like `Foo|Bar|Baz` where only `Baz` is a resource

src/JsonSchema/SchemaFactory.php
```diff
@@ -42,7 +41,7 @@ final class SchemaFactory implements SchemaFactoryInterface
     public const FORCE_SUBSCHEMA = '_api_subschema_force_readable_link';
     public const OPENAPI_DEFINITION_NAME = 'openapi_definition_name';
 
-    public function __construct(private readonly TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null)
+    public function __construct(ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null)
```
soyuka, line 44: This is a really good improvement imo but we still need to cover BC on that constructor. can we trigger a deprecation on using it and we'll remove it in 4.x

src/JsonSchema/SchemaFactory.php
```diff
@@ -42,8 +41,12 @@ final class SchemaFactory implements SchemaFactoryInterface
     public const FORCE_SUBSCHEMA = '_api_subschema_force_readable_link';
     public const OPENAPI_DEFINITION_NAME = 'openapi_definition_name';
 
-    public function __construct(private readonly TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null)
+    public function __construct(?TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ResourceClassResolverInterface $resourceClassResolver = null)
     {
+        if ($typeFactory) {
```
chalasr, line 46: Don't you want to remove this parameter from the signature at some point e.g. in 4.0? 
This looks like a silent BC break. The layer in place does not allow calling code to adapt before the actual removal given there is no way to make such code ready for the 4.0 constructor (unless using named arguments, not a majority)

PR #5472, opened 2023-03-17
feat: create TraversablePartialPaginator

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

A small PR which adds a `TraversablePartialPaginator` which only implements `PartialPaginatorInterface`

### DIFF
--- a/src/State/Pagination/TraversablePaginator.php
+++ b/src/State/Pagination/TraversablePaginator.php
@@ -13,8 +13,17 @@ declare(strict_types=1);
 
 namespace ApiPlatform\State\Pagination;
 
+/**
+ * @template T of object
+ *
+ * @implements PaginatorInterface<T>
+ * @implements \IteratorAggregate<T>
+ */
 final class TraversablePaginator implements \IteratorAggregate, PaginatorInterface
 {
+    /**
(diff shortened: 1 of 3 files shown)


src/State/Pagination/TraversablePartialPaginator.php
```diff
@@ -0,0 +1,62 @@
...
+        return $this->itemsPerPage;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function count(): int
+    {
+        return iterator_count($this->getIterator());
+    }
```
nikophil, line 53: just wondering what we should count here :thinking: 

I had a problem of `Cannot traverse an already closed generator` when using this, and passed the `$count` as a constructor property. WDYT?

src/State/Pagination/TraversablePaginator.php
```diff
@@ -13,8 +13,17 @@
 
 namespace ApiPlatform\State\Pagination;
 
+/**
+ * @template T of object
+ *
+ * @implements PaginatorInterface<T>
+ * @implements \IteratorAggregate<T>
+ */
```
nikophil, line 21: added some missing generics annotation

Issue #5464, opened 2023-03-15
Exception on POST JSON object with numeric property


alebedev80:
**API Platform version(s) affected**: 3.1.4, 2.7.9
**PHP version**: 8.2.3 
**Description**  
On POST request to endpoint with JSON object where one property is numeric will be exception     

```
"hydra:description": "ApiPlatform\\Serializer\\AbstractItemNormalizer::canAccessAttributePostDenormalize(): Argument #3 ($attribute) must be of type string, int given, called in /srv/app/vendor/api-platform/core/src/Serializer/AbstractItemNormalizer.php on line 227"
```

**How to reproduce**  
1. Set-up API Platform
2. Send request 

```
curl -vk --location 'https://localhost/greetings' \
--header 'Content-Type: application/json' \
--data '{
    "0": 1
}'
```

or

```
curl 'https://demo.api-platform.com/books' -v -H 'content-type: application/ld+json' --data-raw '{"0":"1"}'
```

**Possible Solution**  
As quick fix i locally inserted a row \ApiPlatform\Serializer\AbstractItemNormalizer:227

```
$attribute = (string)$attribute;
```

soyuka:
Book has no `0` property

alebedev80:
@soyuka yes book has no property 0, but code raise exception and server response 500, because what wrong argument type. 
You can't expect that users will send proper data. So please re-open this issue. That's a bug!

soyuka:
I don't understand, what exception do you expect?

alebedev80:
@soyuka i don't expect any exception in that case ;)

soyuka:
I think that this should throw a `400 Bad Request`. Tagging as bug.

alebedev80:
@soyuka i'm not sure that it should throw 400 because on post a json any string property which is 'unepected' per service that property just ignoring. I think behavior should be the same - just ignore that numeric property.

alebedev80:
@soyuka imho if by this request response is 201:

```
curl 'https://demo.api-platform.com/books' -v -H 'content-type: application/ld+json' --data-raw '{"author":"a", "title": "t", "description": "d", "publicationDate": "2023-03-16", "x": "0"}'
```

then by this also should be 201:

```
curl 'https://demo.api-platform.com/books' -v -H 'content-type: application/ld+json' --data-raw '{"author":"a", "title": "t", "description": "d", "publicationDate": "2023-03-16", "0": "0"}'
```

VincentLanglet:
[...]

PR #5474, opened 2023-03-17
fix(normalizer): return 400 HTTP status code when sending bad type pr…

GrimpEreau:
…operty

| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5464 
| License       | MIT
| Doc PR        |


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -222,13 +222,19 @@ public function denormalize(mixed $data, string $class, string $format = null, a
             return $object;
         }
 
+        $options = $this->getFactoryOptions($context);
+        $propertyNames = iterator_to_array($this->propertyNameCollectionFactory->create($resourceClass, $options));
         // Revert attributes that aren't allowed to be changed after a post-denormalize check
         foreach (array_keys($data) as $attribute) {
+            if (!\in_array($attribute, $propertyNames, true)) {
```
deguif, line 229:
```suggestion
            $attribute = $this->nameConverter ? $this->nameConverter->denormalize($attribute) : $attribute;

            if (!\in_array($attribute, $propertyNames, true)) {
```

deguif: Shouldn't the name converter be used to convert the attribute name to object property name?

soyuka:
we will actually fix that at https://github.com/api-platform/core/pull/5495

PR #5489, opened 2023-03-23
fix(openapi): use 3.1 version

soyuka:
Configuration now validates the api keys "key" using the pattern from the OpenAPI 3.1 specification: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json#L282


src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -280,6 +280,10 @@ private function addSwaggerSection(ArrayNodeDefinition $rootNode): void
                         ->end()
                         ->arrayNode('api_keys')
                             ->useAttributeAsKey('key')
+                            ->validate()
+                                ->ifTrue(static fn($v): bool => empty(array_filter(array_keys($v), fn($item) => preg_match('/^[a-zA-Z0-9._-]+$/', $item))))
```
mtarld, line 284:
```suggestion
                                ->ifTrue(static fn($v): bool => !empty(array_filter(array_keys($v), fn($item) => !preg_match('/^[a-zA-Z0-9._-]+$/', $item))))
```

mtarld (approved):

PR #5492, opened 2023-03-23
feat: add the ability to use previous_data object in state processor for update with custom Input DTO

jelovac:
Adding the ability to use 'previous_data' object which is passed via $context to StateProcessor for update when custom Input DTO is provided for the operation. This basically solves the issue which many of us are having due to DataTransformers being deprecated and having to convert DTO to a Resource object in state processor.

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets         | #4916 
| License        | MIT

### DIFF
--- a/src/Symfony/EventListener/DeserializeListener.php
+++ b/src/Symfony/EventListener/DeserializeListener.php
@@ -96,9 +96,31 @@ final class DeserializeListener
             $context[AbstractNormalizer::OBJECT_TO_POPULATE] = $data;
         }
         try {
+            $inputData = $this->serializer->deserialize(
+                $request->getContent(),
+                $context['resource_class'],
+                $format,
+                $context
+            );
+
+            /**
+             * If deserialized request object is an instance of an Input DTO overwrite `previous_data`
+             * with `data` in order to be able to use `previous_data` for updates in State Processor
+             * without the need to load the original object explicitly.
+             */
+            foreach ($operation->getInput() as $input) {
+                if($inputData instanceof $input) {
+                    $request->attributes->set(
+                        'previous_data',
+                        $data,
+                    );
+                    break;
+                }
+            }
+
             $request->attributes->set(
                 'data',
-                $this->serializer->deserialize($request->getContent(), $context['resource_class'], $format, $context)
+                $inputData
             );
         } catch (PartialDenormalizationException $e) {
             $violations = new ConstraintViolationList();


src/Symfony/EventListener/DeserializeListener.php
```diff
@@ -96,9 +96,31 @@ public function onKernelRequest(RequestEvent $event): void
...
+                $format,
+                $context
+            );
+
+            /**
+             * If deserialized request object is an instance of an Input DTO overwrite `previous_data`
+             * with `data` in order to be able to use `previous_data` for updates in State Processor
+             * without the need to load the original object explicitly.
+             */
+            foreach ($operation->getInput() as $input) {
```
soyuka, line 111: Not possible, it's an array because of legacy usage which was `['class' => Foo::class]` but this looks wrong. Also this probably breaks lots of tests

soyuka:
But we do have the `previous_data` available: https://github.com/api-platform/core/blob/main/src/Symfony/EventListener/WriteListener.php#L81 

If you really need this, please use your own listener, or mark your HttpOperation directly on your Input.

jelovac:
> But we do have the `previous_data` available: https://github.com/api-platform/core/blob/main/src/Symfony/EventListener/WriteListener.php#L81
> 
> If you really need this, please use your own listener, or mark your HttpOperation directly on your Input.

Previous data is available indeed. However that object is detached from the Entity Manager. Because it is a clone of the original object. And as a result it is useless when we want to use it for updates. Because detached objects will always cause INSERTS in Doctrine instead of UPDATE.

For the time being I have managed to do it using Event Subscriber like this:

```
<?php

declare(strict_types=1);

namespace App\EventSubscriber;

use ApiPlatform\Metadata\HttpOperation;
use ApiPlatform\Metadata\Patch;
use ApiPlatform\Metadata\Put;
use ApiPlatform\Symfony\EventListener\EventPriorities;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\HttpKernel\Event\RequestEvent;
use Symfony\Component\HttpKernel\Event\ViewEvent;
use Symfony\Component\HttpKernel\KernelEvents;

final class PreloadedDataSubscriber implements EventSubscriberInterface
{
    /**
     * @return array<mixed>
     */
    public static function getSubscribedEvents(): array
    {
        return [
            KernelEvents::REQUEST => [
                [
                    'addPreloadedDataObjectToRequestAttributes',
                    EventPriorities::POST_READ,
                ],
            ],
            KernelEvents::VIEW => [
                [
                    'modifyPreviousDataObjectRequestAttribute',
                    EventPriorities::PRE_WRITE,
                ],
            ],
        ];
    }

    public function addPreloadedDataObjectToRequestAttributes(RequestEvent $requestEvent)
    {
        if ($requestEvent->getRequest()->attributes->has('data')) {
            $requestEvent->getRequest()->attributes->set(
                'preloaded_data',
                $requestEvent->getRequest()->attributes->get('data')
            );
        }
    }

    public function modifyPreviousDataObjectRequestAttribute(ViewEvent $viewEvent)
    {
        $data = $viewEvent->getControllerResult();

        $request = $viewEvent->getRequest();

        if (
            !$request->attributes->has('_api_operation')
            || !$request->attributes->has('preloaded_data')
        ) {
            return;
        }

        $preloadedData = $request->attributes->get('preloaded_data');

        /** @var HttpOperation $operation */
        $operation = $request->attributes->get('_api_operation');

        // For now supporting only PUT / PATCH
        if (
            !$operation instanceof Put
            && !$operation instanceof Patch
        ) {
            return;
        }

        $input = $operation->getInput();

        if (!$input) {
            return;
        }

        $modify = false;
        if (is_array($input)) {
            if (isset($input['class']) && $data instanceof $input['class']) {
                $modify = true;
            } else {
                foreach ($input as $inputType) {
                    if (isset($inputType['class']) && $data instanceof $inputType['class']) {
                        $modify = true;
                        break;
                    } elseif (is_string($inputType) && $data instanceof $inputType) {
                        $modify = true;
                        break;
                    }
                }
            }
        } else {
            if ($data instanceof $input) {
                $modify = true;
            }
        }

        if ($modify) {
            $request->attributes->set('previous_data', $preloadedData);
        }
    }
}
```

PR #5494, opened 2023-03-23
refactor(metadata): replace HttpOperation constants by raw strings

MarvinCourcier:
As spoken with @chalasr there are string hardcoded operation request which could be replaced with the HttpOperation request const

Maybe this isn't what you want but I thought this would be good to replace those.
I'll gladly take any advice.

I don't know to which branch this should be merged so I'd like your advice.

### DIFF
--- a/src/Metadata/HttpOperation.php
+++ b/src/Metadata/HttpOperation.php
@@ -76,7 +76,7 @@ class HttpOperation extends Operation
      * @param string|callable|null   $processor {@see https://api-platform.com/docs/core/state-processors/#state-processors}
      */
     public function __construct(
-        protected string $method = self::METHOD_GET,
+        protected string $method = 'GET',
         protected ?string $uriTemplate = null,
         protected ?array $types = null,
         protected $formats = null,
--- a/src/Doctrine/Common/State/PersistProcessor.php
+++ b/src/Doctrine/Common/State/PersistProcessor.php
@@ -50,7 +50,7 @@ final class PersistProcessor implements ProcessorInterface
         // PUT: reset the existing object managed by Doctrine and merge data sent by the user in it
         // This custom logic is needed because EntityManager::merge() has been deprecated and UPSERT isn't supported:
         // https://github.com/doctrine/orm/issues/8461#issuecomment-1250233555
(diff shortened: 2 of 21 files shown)


soyuka:
Indeed this is a good change, you should change your commit though:

```
refactor(metadata): use HttpOperation constants for methods
```

`chore` is used for cleanup or things you can't categorize.

MarvinCourcier:
> Indeed this is a good change, you should change your commit though:
> 
> ```
> refactor(metadata): use HttpOperation constants for methods
> ```
> 
> `chore` is used for cleanup or things you can't categorize.

Sur thing ! it's done !

soyuka:
@MarvinCourcier the commit should be changed, not the Pull Request title which is unrelated after merging except for github.

dunglas:
Could you do a profile please? IIRC resolving the constants has a runtime cost that can be significant and it's why we do not use them at least on the hot path.

dunglas:
👍 for consistency, but let's use a raw string everywhere then.

src/Metadata/HttpOperation.php
```diff
@@ -18,14 +18,6 @@
 
 class HttpOperation extends Operation
 {
-    public const METHOD_GET = 'GET';
-    public const METHOD_POST = 'POST';
-    public const METHOD_PUT = 'PUT';
-    public const METHOD_PATCH = 'PATCH';
-    public const METHOD_DELETE = 'DELETE';
-    public const METHOD_HEAD = 'HEAD';
-    public const METHOD_OPTIONS = 'OPTIONS';
```
MarvinCourcier, line 27: Since we want to use raw string, I don't see the point of those constant to remain.
So I've deleted them

MarvinCourcier: Let's ask to @dunglas and @chalasr then, what should be done there ?

How do you wish to proceed ?

chalasr: Reasoning is: these constants are for end users, api platform does not reference them for performance reasons. They must stay, and the codebase should not use them at all for consistency as per @dunglas statement.

dunglas: I totally agree with @chalasr

MarvinCourcier: It's done then, I've reverted it back the way it was :)

soyuka:
thanks

PR #5499, opened 2023-03-24
fix(metadata): no deprecation when elasticsearch is null

dannyvw:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #... 
| License       | MIT

Related to https://github.com/api-platform/core/pull/5450. Now fixed for non graphql operations.


soyuka:
Looks like the Behat CI for ES in unhappy witht he changes

dannyvw:
@soyuka The provider is not set anymore https://github.com/api-platform/core/blob/main/src/Elasticsearch/Metadata/Resource/Factory/ElasticsearchProviderResourceMetadataCollectionFactory.php#L63. I could add this in the testcases, but not sure if that is the right way.

soyuka:
thanks @dannyvw !

soyuka:
My bad I reverted it was still red. I cannot reopen, maybe we can discuss this at #5497

dannyvw:
This pipeline is indeed red and it is something else than #5497

PR #5513, opened 2023-03-30
fix(symfony): check operations's parameters

Zowac:
| Q             | A
| ------------- | ---
| Branch?       | 3.1 
| License       | MIT

As discussed with @soyuka, we need to do the verification on the parameters of the operations as well


src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -573,4 +574,11 @@ private function addMakerSection(ArrayNodeDefinition $rootNode): void
                 ->end()
             ->end();
     }
+
+    private function defineDefault($defaultsNode, \ReflectionClass $reflectionClass, $nameConverter)
```
soyuka, line 578: no types?

Zowac: Added, thanks :)

PR #5515, opened 2023-03-31
fix(jsonschema): access related subschema on readableLink

### LINKED ISSUE #5501: [Swagger] Undefined array key "$ref" since 3.1.6
**API Platform version(s) affected**: 3.1.6

**Description**  

Since this new version, it is impossible to acces my swagger documentation via /api 

For instance for a simple entrypoint like this one 

```php
new Metadata\GetCollection(
            routeName: 'api_get_user',
            openapi: new Model\Operation(
                responses: [
                    Response::HTTP_UNAUTHORIZED => new Model\Response(ResponseMessage::UNAUTHORIZED),
                ],
                parameters: [
                    new Model\Parameter(
                        name: 'id',
                        in: 'query',
                        description: 'The User identifier',
                        schema: ['type' => 'object'],
                        style: 'deepObject',
                        explode: true,
                    ),
                    new Model\Parameter(
                        name: 'name|lastname|username',
                        in: 'query',
                        description: 'The User denomination',
                        schema: ['type' => 'object'],
                        style: 'deepObject',
                        explode: true,
                    ),
                ],
            ),
        ),
```

cause 500 error

Capture d’écran 2023-03-27 à 18 12 04

**Additional Context**  

I'm working on MacOs, Php 8.1, Symfony 6.2.7, doctrine-bundle 2.9.
Everything is working fine on 3.1.5

soyuka:
fixes #5501

The locations relation inside BrokenDocs is a Resource (named Related) but its only operation is a NotExposed. Still, serializer groups are set, and therefore it is a "readableLink" so we actually want to compute the schema, even if it's not accessible directly, it is accessible through that relation.

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -24,6 +24,7 @@ use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInter
 use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
 use ApiPlatform\Metadata\ResourceClassResolverInterface;
 use ApiPlatform\Metadata\Util\ResourceClassInfoTrait;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue5501\Related;
 use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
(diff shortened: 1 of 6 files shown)


src/JsonSchema/SchemaFactory.php
```diff
@@ -24,6 +24,7 @@
 use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
 use ApiPlatform\Metadata\ResourceClassResolverInterface;
 use ApiPlatform\Metadata\Util\ResourceClassInfoTrait;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue5501\Related;
```
alanpoulain, line 27: To remove.

PR #5520, opened 2023-04-03
draft: fix(UriVariable): fix string being cast to int

PierreRebeilleau:
| Q             | A
| ------------- | ---
| Branch?       |3.1 
| Tickets       | #5396 
| License       | MIT

When passing multiple UriVariables, it is possible that a string, such as a key, could be cast to an integer (e.g., '73foo'). To avoid any potential issues, a check is performed before the cast as a workaround.

### DIFF
--- a/src/Metadata/Resource/ResourceMetadataCollection.php
+++ b/src/Metadata/Resource/ResourceMetadataCollection.php
@@ -89,9 +89,9 @@ final class ResourceMetadataCollection extends \ArrayObject
         }
 
         // Idea:
-//         if ($metadata) {
-//             return (new class extends HttpOperation {})->withResource($metadata);
-//         }
+        //         if ($metadata) {
+        //             return (new class extends HttpOperation {})->withResource($metadata);
+        //         }
 
         $this->handleNotFound($operationName, $metadata);
     }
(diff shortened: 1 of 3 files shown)


soyuka:
try to reproduce the issue at https://github.com/api-platform/core/issues/5396 first, I don't think that an Integer IntegerUriVariableTransformer should transform something to a string. We probably don't even need to transform it if it's a string in the UriVariablesResolverTrait.

src/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -89,9 +89,9 @@ public function getOperation(?string $operationName = null, bool $forceCollectio
         }
 
         // Idea:
-//         if ($metadata) {
-//             return (new class extends HttpOperation {})->withResource($metadata);
-//         }
+        //         if ($metadata) {
+        //             return (new class extends HttpOperation {})->withResource($metadata);
+        //         }
```
MarvinCourcier, line 94: Hey @PierreRebeilleau, remember me ? ;)

I think this has nothing to do with your PR. So maybe I'm wrong but you might have to revert it.
I actually thought of doing a dedicated PR for this because it affect all the created branch

soyuka: I already did that this needs rebase

PierreRebeilleau: Hey @MarvinCourcier , yes i remember you ! Glad to see you around, ty i'm gonna rebase :)

PR #5533, opened 2023-04-05
fix(hydra): prevent adding Link header when docs is disabled

### LINKED ISSUE #5528: AddLinkHeaderListener always adds a link to docs
**API Platform version(s) affected**: all

**Description**  
[AddLinkHeaderListener.php](https://github.com/api-platform/core/blob/main/src/Hydra/EventListener/AddLinkHeaderListener.php) always adds a link to the api docs, even if enable_docs is false.

**How to reproduce**  

Set enable_docs to false. Your application will fail because the request listener will try to generate a url for a route that does not exist.

**Possible Solution**  

Inject the configuration value and check properly if the documentation exists before generating a link to it.

**Additional Context**

thbil:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5528 
| License       | MIT
| Doc PR        | NA

Hello,

This PR aims to fix #5528.
Should I handle the backward compatibility as explained [here](https://symfony.com/doc/current/contributing/code/bc.html#add-argument-public-method)?

### DIFF
--- a/src/Hydra/EventListener/AddLinkHeaderListener.php
+++ b/src/Hydra/EventListener/AddLinkHeaderListener.php
@@ -29,7 +29,7 @@ final class AddLinkHeaderListener
 {
     use CorsTrait;
 
-    public function __construct(private readonly UrlGeneratorInterface $urlGenerator)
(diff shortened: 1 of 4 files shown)


soyuka:
I think the issue is about openapi only no?

PR #5542, opened 2023-04-12
 fix(serializer): no forced resource class relation 

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | 3.1

Simple bugfix : force_resource_class needs to be reset before trying to normalize children attributes.


soyuka:
merci @mrossard !

PR #5546, opened 2023-04-15
Fix #5545.

### LINKED ISSUE #5545: Guess uri variables with the operation and the data instead of hardcoding id
**API Platform version(s) affected**: 3.*

**Description**

In a situation where these two conditions are true, also alternatively and not necessarily at the same time, the code will break:

1. The IRI has some parameters (ex.: `account/{account}/resource/{resource}/sub_resource/{sub_resource}`
2. The uri parameter of the resource is not `id` (in the above example, the uri parameter is `sub_resource`)

> Unable to generate an IRI for the item of type "App\Entity\ResourceLine""

**Why it breaks**

In `ItemNormalizer:79` there is this todo:

https://github.com/api-platform/core/blob/ccef472cade42b189fa8cb7793e27fad1349e289/src/Serializer/ItemNormalizer.php#L79-L80

As you can see, the context passed to  `$this->iriConverter->getIriFromResource()` (4th parameter) is hardcoded and always refers to the ID of the current denormalizing entity.

This is not correct (and, in fact, there is also the todo to remember this fact) as this could break the code.

**How to reproduce**

This is a code that reproduces this problem (not runnable as it was condensed for the sake of the example).

This is the `Resource` (which, in turn, is a "sub resource" of an `Account`):

```php
#[Entity]
#[Table(name: 'resources')]
#[API\ApiResource(
    ...
    operations: [
        ...
        new API\Put(
            uriTemplate: '/account/{account}/resource/{resource}',
            ...
            uriVariables: [
                'account' => new API\Link(fromClass: Account::class, toProperty: 'account', identifiers: ['id']),
                'resource' => new API\Link(fromClass: self::class, identifiers: ['id']),
            ],
        ),
    ],
)]
class Resource
{
    #[Column(name: 'id', type: Types::INTEGER)]
    #[Id]
    #[GeneratedValue(strategy: 'IDENTITY')]
    #[Groups([
        'resource:read',
        'resource:write',
    ])]
    private int $id;

    #[ManyToOne(targetEntity: Account::class, inversedBy: 'resources')]
    #[JoinColumn(name: 'account', nullable: false)]
    #[Groups(['resource:read', 'resource:write'])]
    private Account $account;

    ...
    
    #[OneToMany(targetEntity: ResourceLine::class, mappedBy: 'resource', cascade: [
        'persist',
        'remove'
    ])]
    #[Groups([
        'resource:read',
        'resource:write'
    ])]
    private Collection $lines;
    
    ...
}
```

This is the `ResourceLine` (which, in turn, is a "sub resource" of an `Account` AND of a `Resource`):

```php
#[Entity]
#[Table(name: 'resources_lines')]
#[API\ApiResource(
    ...
    operations: [
        new API\GetCollection(
            uriTemplate: '/account/{account}/resource/{resource}/resource_lines',
            uriVariables: [
                'account' => new API\Link(fromClass: Account::class, toProperty: 'account', identifiers: ['id']),
                'resource' => new API\Link(fromClass: Resource::class, toProperty: 'resource', identifiers: ['id']),
            ],
        ),
        new API\Get(
            uriTemplate: '/account/{account}/resource/{resource}/resource_lines/{resource_line}',
            uriVariables: [
                'account' => new API\Link(fromClass: Account::class, toProperty: 'account', identifiers: ['id']),
                'resource' => new API\Link(fromClass: Resource::class, toProperty: 'resource', identifiers: ['id']),
                'resource_line' => new API\Link(fromClass: self::class, identifiers: ['id']),
            ],
        ),
    ],
)]
class ResourceLine
{
    #[Id]
    #[GeneratedValue(strategy: 'IDENTITY')]
    #[Column(name: 'id', type: Types::INTEGER)]
    #[Groups(['resource_line:read', 'resource:read'])]
    private int $id;

    #[ManyToOne(targetEntity: Account::class, inversedBy: 'resourcesLines')]
    #[JoinColumn(name: 'account', nullable: false)]
    #[Groups(['resource_line:read'])]
    private Account $account;

    #[ManyToOne(targetEntity: Resource::class, inversedBy: 'lines')]
    #[JoinColumn(name: 'resource', nullable: false)]
    #[Groups(['resource_line:read', 'resource:read'])]
    private Resource $resource;

    ...
}
```

The issue happens when we try to save some `ResourceLine`s through the `Resource` and the array contains also other `ResourceLine`s already saved (so, we mix already present `ResourceLine` and new `ResourceLine`s to add with the HTTP call):

```php

class ResourceTest {

    ...

    public function testPutResourceAddingAResourceLinePassingAlsoAlreadyPresentLines(): void
    {
        $resourceIri = $this->findIriBy(Resource::class, ['id' => '1');

        ...

        $resourcePayloadResponse = static::callEndpoint(
            method: Request::METHOD_GET,
            endpoint: $resourceIri,
        );

        $resourcePayload = $resourcePayloadResponse->toArray();

        // Here we get the already present `ResourceLine`s
        $resourceLines = $resourcePayload['lines'];

        $testData = [
                // The data of the new `ResourceLine` we are adding
        ];

        $resourceLines[] = $testData;

        $response = static::callEndpoint(Request::METHOD_PUT, $resourceIri, [
            'lines' => $resourceLines,
        ]);

        $decodedResponse = $response->toArray(false);
        dd($decodedResponse);
    }

    ...
}
```

This test returns this error:

```console
array:5 [
  "@context" => "/api/contexts/Error"
  "@type" => "hydra:Error"
  "hydra:title" => "An error occurred"
  "hydra:description" => "Unable to generate an IRI for the item of type "App\Entity\ResourceLine""
  "trace" => array:39 [
  ...
```

**Possible Solution**

Resolve the `@todo` pinned in the code.

Aerendir:
| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

Fix #5545


src/Serializer/ItemNormalizer.php
```diff
@@ -76,10 +77,35 @@ private function updateObjectToPopulate(array $data, array &$context): void
...
 
             $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getResourceFromIri($iri, ['fetch_data' => true]);
         }
     }
+
+    private function guessUriVariables(array $data, $operation, array $context): array
+    {
+        if (false === isset($context['uri_variables'])) {
+            // @todo Improve the message and check the exception class is correct
+            throw new InvalidArgumentException('Missing key "uri_variables" from the context.');
```
mrossard, line 91: Wouldn't it be ok to return ['id' => $data['id']] here instead of just bailing out, to mimic the previous incomplete behaviour?

Aerendir: Yes, it is actually a good suggestion 💪 Thank you!

src/Serializer/ItemNormalizer.php
```diff
@@ -76,10 +77,34 @@ private function updateObjectToPopulate(array $data, array &$context): void
...
+            $uriVariables = $this->guessUriVariables($data, $operation, $context);
+            $iri = $this->iriConverter->getIriFromResource($context['resource_class'], UrlGeneratorInterface::ABS_PATH, $operation, ['uri_variables' => $uriVariables]);
 
             $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getResourceFromIri($iri, ['fetch_data' => true]);
         }
     }
+
+    private function guessUriVariables(array $data, $operation, array $context): array
+    {
+        if (false === isset($context['uri_variables'])) {
```
soyuka, line 89:
```suggestion
        if ($context['uri_variables'] ?? false) {
```

src/Serializer/ItemNormalizer.php
```diff
@@ -76,10 +77,34 @@ private function updateObjectToPopulate(array $data, array &$context): void
...
-            // todo: we could guess uri variables with the operation and the data instead of hardcoding id
-            $iri = $this->iriConverter->getIriFromResource($context['resource_class'], UrlGeneratorInterface::ABS_PATH, $operation, ['uri_variables' => ['id' => $data['id']]]);
+            $uriVariables = $this->guessUriVariables($data, $operation, $context);
+            $iri = $this->iriConverter->getIriFromResource($context['resource_class'], UrlGeneratorInterface::ABS_PATH, $operation, ['uri_variables' => $uriVariables]);
 
             $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getResourceFromIri($iri, ['fetch_data' => true]);
         }
     }
+
+    private function guessUriVariables(array $data, $operation, array $context): array
```
soyuka, line 87:
```suggestion
    private function getContextUriVariables(array $data, $operation, array $context): array
```

soyuka (commented):
nice!

soyuka:
Can you target 3.1 as this is a bug fix?

src/Serializer/ItemNormalizer.php
Aerendir: @soyuka , I don't think it will work with this suggestion...

If `$context['uri_variables']` is set and has any value, then it will enter the `if`, doesn't it?

And we don't want this. We want it enter the `if` only if the key `uri_variables` doesn't exist (or is `null`, `false`, etc.).

This suggestion is, in practice, the opposite of what we want to achieve.

WDYT?

PS
Also the tests will fail with this modification.

src/Serializer/ItemNormalizer.php
Aerendir: Changed!

Aerendir:
@soyuka , done: the PR now targets 3.1.

tacman:
Thanks!  Can you document this an provide an example in the official docs?  It's been something I've wanted and worked with subresources in 2.6 but much harder to do in 3.0, in part because there's so few examples.

src/Serializer/ItemNormalizer.php
soyuka: at least `!isset` then?

soyuka:
I'll rebase that thanks!

src/Serializer/ItemNormalizer.php
Aerendir: @soyuka , ok, if you like it more 👍

Aerendir:
@tacman , it is already documented: https://api-platform.com/docs/core/subresources/

This PR simply allow to use an URI like `/companies/{companyId}/employees/{employeeId}` while before you were forced to use one like `/companies/{companyId}/employees/{id}` (like in the example provided in the link).

soyuka:
This will be out friday

Issue #5510, opened 2023-03-30
Using the new entityClass option with a custom provider 


mrossard:
**API Platform version(s) affected**: 3.1.6

**Description**  
I can't get this new option to work as intended while using a custom provider. Either the Get() operation or the GetCollection() is broken.

**How to reproduce**  
https://github.com/mrossard/apip-entityclass

This version gets "The class 'App\\ApiResource\\DetachedResource' was not found in the chain configured namespaces App\\Entity" when you call the Get() operation. Comment the `uriVariables: ['id']`  and you get `Unable to generate an IRI for the item of type \"App\\ApiResource\\DetachedResource\`` on the GetCollection() operation.

**Possible Solution**  
In ApiPlatform\Api\IdentifiersExtractor, if you replace line 136
``` php
 throw new RuntimeException('Not able to retrieve identifiers.');
```
with
``` php
return $this->resolveIdentifierValue($this->propertyAccessor->getValue($item, $property), $property);
```
you seem to get the expected result (on the version without the explicit uriVariables).

DartCZ:
This solution fixing issue #5502

SergeyCherenkov:
If the related entity also has an [ApiResource], then I also get `Unable to generate an IRI for the item of type`. 
https://github.com/api-platform/api-platform/issues/2334#issuecomment-1512618993
This solution works, but it still seems to me that this should not be.

PR #5550, opened 2023-04-18
fix: the stateOptions::entityClass should be used when present while building Links

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | https://github.com/api-platform/core/issues/5510

Currently the LinksHandlerTrait can't cope with stateOptions::entityClass correctly. This PR tries to fix this.
It probably doesn't fix the other issues that benefited from the quickfix mentioned in the issue though, those seem likely separate problems with similar results.


src/Doctrine/Orm/State/LinksHandlerTrait.php
```diff
@@ -53,8 +53,10 @@ private function handleLinks(QueryBuilder $queryBuilder, array $identifiers, Que
             $hasCompositeIdentifiers = 1 < \count($identifierProperties);
 
             if (!$link->getFromProperty() && !$link->getToProperty()) {
-                $doctrineClassMetadata = $manager->getClassMetadata($link->getFromClass());
-                $currentAlias = $link->getFromClass() === $entityClass ? $alias : $queryNameGenerator->generateJoinAlias($alias);
+                // if stateOptions::entityClass is present it should be used. Otherwise it's $link::fromClass
+                $fromClass = $operation->getStateOptions()?->getEntityClass() ?? $link->getFromClass();
```
soyuka, line 57: We should fix that directly in the metadata: https://github.com/api-platform/core/blob/ccef472cade42b189fa8cb7793e27fad1349e289/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php#L227

we forgot to check this option there, the condition should be something like:

https://github.com/api-platform/core/blob/main/src/Metadata/Resource/Factory/UriTemplateResourceMetadataCollectionFactory.php#L186-L188

mrossard: I wasn't sure wether that was to be fixed before the place i needed it or not. 
Aren't there possible side effects if you just replace resourceClass there? Re-checking what's actually in the Operation when i needed seemed "safer".

PR #5569, opened 2023-04-25
fix(serializer): fix denormalizing to non-cloneable objects

### LINKED ISSUE #5561: Error denormalizing to non-cloneable objects
**API Platform version(s) affected**: 2.7.0 through 3.1.8

**Description**  
`ApiPlatform\Serializer\AbstractItemNormalizer::denormalize()` attempts to clone objects without checking if they're actually cloneable, resulting in a fatal error.

https://github.com/api-platform/core/blob/3de661c2d163629cb9097de5a0370612a00f6523/src/Serializer/AbstractItemNormalizer.php#L217

**How to reproduce**  

Attempt to denormalize any object where `object_to_populate` has a `__clone()` function that is `private` or `protected`, like this:

```php
class CannotBeCloned {
    public string $foo;

    private function __clone()
    {
    }
};

$original = new CannotBeCloned();

$itemNormalizer->denormalize(['foo' => 'bar'], CannotBeCloned::class, context: ['resource_class' => CannotBeCloned::class, 'object_to_populate' => $original]);
```

Result:

    Fatal error: Uncaught Error: Call to private CannotBeCloned::__clone() from scope ApiPlatform\Serializer\AbstractItemNormalizer

**Possible Solution**  

Utilize [your `CloneTrait`](https://github.com/api-platform/core/blob/v3.1.8/src/Util/CloneTrait.php) to avoid this fatal error, and then perhaps skip the post-denormalize-revert loop for non-cloneable objects?  Or whatever you feel is most appropriate.

**Additional Context**  

It seems this BC break was introduced in v2.7.0 to support the `security_post_denormalize` feature (#4184).  We discovered this trying to upgrade from 2.6.8 to any newer version.

colinodell:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #5561 
| License       | MIT
| Doc PR        | n/a

Fixes #5561 in the 2.7 branch by using a safer clone approach, and skipping post-normalization attribute revert functionality if the object was never cloned.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -33,6 +33,7 @@ use ApiPlatform\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use ApiPlatform\Symfony\Security\ResourceAccessCheckerInterface;
 use ApiPlatform\Util\ClassInfoTrait;
+use ApiPlatform\Util\CloneTrait;
 use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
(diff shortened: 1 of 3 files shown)


colinodell:
The failing tests seem unrelated to these changes

soyuka:
Nice, thanks a lot, I'll release that by the end of the week.

colinodell:
I think commit https://github.com/api-platform/core/commit/8257741fca0e6e5a9b71096b3965a0a5547e8570 needs to be tagged as v2.7.12?

soyuka:
My bad, just tagged it!

PR #5570, opened 2023-04-26
feat (metadata): added PostCollection operation

joerndyherrn:
| Q             | A
| ------------- | ---
| Branch       | main
| License       | MIT

Implemented a new operation: PostCollection in order to be able to return multiple entities when having a custom POST controller

### DIFF
--- a/src/Metadata/PostCollection.php
+++ b/src/Metadata/PostCollection.php
@@ -1,0 +1,116 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final class PostCollection extends HttpOperation implements CollectionOperationInterface
+{
+    private $itemUriTemplate;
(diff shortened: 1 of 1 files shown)


joerndyherrn:
@soyuka Probably don't need to fix ResourceMetadataCollection as of https://github.com/api-platform/core/pull/5520/files#r1160520085
But won't get a green pipe though - would you mind to check another failed test: https://github.com/api-platform/core/actions/runs/4805382502/jobs/8551742674?pr=5570

soyuka:
It looks weird to me to have a POST that returns a Collection but why not! 
 @dunglas any thoughts?

mrossard:
Isn't this just supposed to be a POST on a custom resource that represents a collection of another type of resource? 
You'd return a single resource, but it would contain a collection of something else, which is pretty simple to do and closer to what a POST should be doing as far as i can see.

joerndyherrn:
@mrossard No, let me explain it with a simple example:
We got a dictionary where we‘d like to add multiple words to it. As response we want a collection of those persisted entities.

mrossard:
> @mrossard No, let me explain it with a simple example: We got a dictionary where we‘d like to add multiple words to it. As response we want a collection of those persisted entities.

Well as far as i know, that's not actually supposed to be done in one single request. What i described earlier was the "best" way I know of doing that while staying close to standards - if you create a specific resource that represents a "group of words", you can POST a "group of words" and get a single "group of words" in return.
That being said it's not great either since it means you're creating subresources on the fly...but batch operations are just not really supported by the standards (or maybe i'm just ignorant in that regard?).

PR #5573, opened 2023-04-26
chore: move api-platform/graphql to its own component

PierreRebeilleau:
| Q             | A
| ------------- | ---
| Branch?       | subtree 
| Tickets       | #... 
| License       | MIT

Move graphql for subtree split


src/GraphQl/Tests/Action/EntrypointActionTest.php
```diff
@@ -0,0 +1,279 @@
+<?php
```
soyuka, line 1: why aren't these files moved from tests/GraphQl to here?

src/Metadata/ApiFilter.php
```diff
@@ -13,7 +13,7 @@
 
 namespace ApiPlatform\Metadata;
 
-use ApiPlatform\Api\FilterInterface as LegacyFilterInterface;
+use ApiPlatform\Metadata\FilterInterface as LegacyFilterInterface;
```
soyuka, line 16: revert

src/Metadata/IdentifiersExtractorInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+interface IdentifiersExtractorInterface
+{
+    /**
+     * Finds identifiers from an Item (object).
+     *
+     * @throws RuntimeException
+     */
+    public function getIdentifiersFromItem(object $item, ?Operation $operation = null, array $context = []): array;
+}
```
soyuka, line 32: conflicts with the Serializer also moving this but :+1: on the new location

src/Metadata/IriConverterInterface.php
```diff
@@ -13,6 +13,7 @@
 
 namespace ApiPlatform\Metadata;
 
+use ApiPlatform\Metadata\UrlGeneratorInterface;
```
soyuka, line 16: not needed if UrlGeneratorInteface we're already at ApiPlatform\Metadata

src/Metadata/Util/ArrayTrait.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata\Util;
+
+trait ArrayTrait
```
soyuka, line 16: this is not internal? can you tell me where its used? We need to cover backward compatibility on the old one and/or probably fix this not being internal but I'm not sure.

src/GraphQl/Tests/Action/EntrypointActionTest.php
soyuka, line 1: tests that you moved here should be removed from the old tests/ directory

Issue #5559, opened 2023-04-21
Unable to update an embedded collection relation with a PATCH operation


jonag:
**API Platform version(s) affected**: 3.1.8

**Description**  

I'm trying to update an embedded relation with a PATCH operation. [Basically this use case](https://github.com/api-platform/core/blob/main/features/main/relation.feature#L368-L397) but with a PATCH instead of a PUT. However my relation is deleted and recreated instead of updated.

**How to reproduce**  

1. Create a resource with a collection. 
2. Try to update an item of this collection with a PATCH operation.

**Possible Solution**  

I tried to debug the issue. I think the issue happens because when the request reach [this line](https://github.com/api-platform/core/blob/main/src/JsonLd/Serializer/ItemNormalizer.php#L142), `$data['@id']` contains the id of my relation but `$context[self::OBJECT_TO_POPULATE]` is set and contains the entire collection instead of just the item.

If I update [this line](https://github.com/api-platform/core/blob/main/src/Serializer/AbstractItemNormalizer.php#L491) with
```php
$childContext = $this->createChildContext($context, $attribute, $format);
$childContext[self::OBJECT_TO_POPULATE] = null;
$values[$index] = $this->denormalizeRelation($attribute, $propertyMetadata, $className, $obj, $format, $childContext);
```
then the request behave as expected, but I'm **really** not sure that this is the right fix.

Please tell me if I can do anything to help fix the issue (or if I'm doing anything wrong).

soyuka:
Hi, merge-patch specification is like this see https://datatracker.ietf.org/doc/html/rfc7386 use put or help us implementing https://datatracker.ietf.org/doc/html/rfc6902 ?

jonag:
Hello, thank you for your answer ! I've read the specification and while it states that the array will be replaced (which I'm totally fine with), it doesn't say that the members composing it should also be replaced (well actually this use case is not documented).

I don't know if my first message was detailed enough so here is exactly what I'm doing:
1. I have [these two resources](https://gist.github.com/jonag/550732ec9e54c6a8af83b5f8b34ab4c7)
2. I create a new Store with:
```bash
curl -X 'POST' \
  'http://localhost:8000/stores' \
  -H 'accept: application/ld+json' \
  -H 'Content-Type: application/ld+json' \
  -d '{
  "name": "My store",
  "pets": [
    {
      "name": "Bernie",
      "age": 2
    },
    {
      "name": "Felix",
      "age": 3
    }
  ]
}'
```
[...]

PR #5577, opened 2023-04-28
fix(jsonapi): fix patch onetomany relation

ilicmsreten:
| Q             | A
| ------------- | ---
| Tickets       | #5559

### DIFF
--- a/features/main/patch.feature
+++ b/features/main/patch.feature
@@ -58,3 +58,47 @@ Feature: Sending PATCH requets
       }
     }
     """
+
+  Scenario: Patch the oneToMany relation
+    Given there is PatchOneToManyDummy with "2" related patchOneToManyDummyRelationWithConstructor objects
+    When I add "Content-Type" header equal to "application/ld+json"
+    When I add "Content-Type" header equal to "application/merge-patch+json"
+    And I send a "PATCH" request to "/patch_one_to_many_dummies/1" with body:
+    """
+    {
+      "relations": [
+        {
+          "id": "/patch_one_to_many_dummy_relation_with_constructors/1",
+          "name": "Relation 1"
+        },
+        {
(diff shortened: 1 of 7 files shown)


features/main/patch.feature
```diff
@@ -58,3 +58,47 @@ Feature: Sending PATCH requets
       }
     }
     """
+
+  Scenario: Patch the oneToMany relation
+    Given there is PatchOneToManyDummy with "2" related patchOneToManyDummyRelationWithConstructor objects
+    When I add "Content-Type" header equal to "application/ld+json"
+    When I add "Content-Type" header equal to "application/merge-patch+json"
```
soyuka, line 65: 2 content types ?

features/main/patch.feature
```diff
@@ -58,3 +58,47 @@ Feature: Sending PATCH requets
...
+    """
+    Then the response status code should be 200
+    And the JSON should be equal to:
+    """
+      {
+        "@context": "/contexts/PatchOneToManyDummy",
+        "@id": "/patch_one_to_many_dummies/1",
+        "@type": "PatchOneToManyDummy",
+        "id": 1,
+        "relations": ["/patch_one_to_many_dummy_relation_with_constructors/1", "/patch_one_to_many_dummy_relation_with_constructors/3"]
```
soyuka, line 88: I need to check https://www.rfc-editor.org/rfc/rfc7386 as I believe that we should always replace the table and that embed relation would not get updated according to spec?

mrossard: I'd expect a POST on /patch_one_to_many_dummy_relation_with_constructors or PUT on /patch_one_to_many_dummy_relation_with_constructors/3 in this example. Creating a subresource with a PATCH on the main resource feels wrong...?

PR #5589, opened 2023-05-16
Fix/composite identifiers 5396

Sarahshr:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5396 
| License       | MIT

This PR allows composite identifiers with differents types

### DIFF
--- a/src/Api/UriVariablesConverter.php
+++ b/src/Api/UriVariablesConverter.php
@@ -63,6 +63,7 @@ final class UriVariablesConverter implements UriVariablesConverterInterface
                 }
             }
         }
+
         return $uriVariables;
     }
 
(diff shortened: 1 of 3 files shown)


dunglas (approved):

soyuka:
mhh breaks https://github.com/api-platform/core/actions/runs/5046456007/jobs/9051992383

src/Api/UriVariablesConverter.php
```diff
@@ -71,7 +73,7 @@ private function getIdentifierTypes(string $resourceClass, array $properties): a
         foreach ($properties as $property) {
             $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $property);
             foreach ($propertyMetadata->getBuiltinTypes() as $type) {
-                $types[] = Type::BUILTIN_TYPE_OBJECT === ($builtinType = $type->getBuiltinType()) ? $type->getClassName() : $builtinType;
+                $types[$property][] = Type::BUILTIN_TYPE_OBJECT === ($builtinType = $type->getBuiltinType()) ? $type->getClassName() : $builtinType;
```
soyuka, line 76: I found the issue, this PR makes an assumption that the parameterName is always in sync with properties, which is not the case as shown here https://github.com/api-platform/core/blob/main/tests/Fixtures/TestBundle/Resources/config/api_resources_odm/resources.xml#L14

PR #5593, opened 2023-05-22
fix(serializer): disable_type_enforcement with null values

### LINKED ISSUE #5584: DISABLE_TYPE_ENFORCEMENT is ignored for non-scalar properties
I'm trying to disable type enforcement so I don't get type errors on nullable=false properties.
Instead, I want to see error messages from Assert\NotBlank, Assert\NotNull, etc.

```
#[ApiResource(
    operations: [
        new Post(
            denormalizationContext: [
                'disable_type_enforcement' => true,
                'collect_denormalization_errors' => true
            ],
        ),
    ],
 )]
class Entity
{
 
    #[ORM\ManyToOne]
    #[Assert\NotBlank()]
    private ?Brand $brand = null;

    public function getBrand(): ?Brand
    {
        return $this->brand;
    }
    
   public function setBrand(?Brand $brand): void
    {
        $this->brand = $brand;
    }

}
```

My request:

```
POST: 
{
  brand: null
}

```
Actual answer:

```
{
   "violations":[
      {
         "propertyPath":"brand",
         "message":"This value should be of type array|string.",
         "code":"0"
      }
}
```

Expected response:

```
{
   "violations":[
      {
         "propertyPath":"brand",
         "message":"This value should not be blank.",
         "code":"0"
      }
}
```

As you can see from the actual answer `DISABLE_TYPE_ENFORCEMENT` does not work for any objects. I have tested it for relations and enum classes (Works for only scalar types).

After researching the problem, I found that in properties with the Object type, we simply do not get to this condition.

https://github.com/api-platform/core/blob/7a1d351b0a37fc7107c548708f4b9e38a9dedc76/src/Serializer/AbstractItemNormalizer.php#L843-L845

What do you think about this? Maybe this was expected behavior for you? 

This issue has also been mentioned here: [#1786](https://github.com/api-platform/api-platform/issues/1786)

soyuka:
fixes #5584

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -753,7 +753,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return $value;
         }
 
-        if (null === $value && $type->isNullable()) {
+        if (null === $value && ($type->isNullable() || ($context[static::DISABLE_TYPE_ENFORCEMENT] ?? false))) {
             return $value;
         }
 
--- a/tests/Fixtures/TestBundle/ApiResource/Issue5584/DtoWithNullValue.php
+++ b/tests/Fixtures/TestBundle/ApiResource/Issue5584/DtoWithNullValue.php
@@ -1,0 +1,23 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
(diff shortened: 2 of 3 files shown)


bizley:
Guys, just to let you know, this change rendered impossible to pass IRI as a value for relation entity when `disable_type_enforcement` is set to true. It was working previously.

soyuka:
@bizley I'm pretty sure we test this, can you give me an example of failing request with its resource? also can you open a new issue? thanks!

bizley:
Sure, will do in a sec.

PR #5601, opened 2023-05-25
fix: allows changing getters and setters without breaks

Sarahshr:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | 
| License       | MIT

This PR tests that no field is added to the resource when renaming the getter and setter of a private property
. (related https://github.com/api-platform/api-platform/issues/2426).


soyuka:
thanks!

Issue #5603, opened 2023-05-26
Cannot PATCH/PUT the resource which contains fields with IRI values


jelovac:
**API Platform version(s) affected**: 3.1.12

**Description**  

Upgraded API Platform from 3.1.9 to 3.1.12.

The issue I am experiencing is that when doing PATCH/PUT request and providing in that request a field which is a relation with IRI value, the value does not get resolved properly.

Basically what happens is the following:

1. Item provider loads for the PATCH/PUT request the resource I am trying to update
2. Next it calls again the Item Provider for all the fields with IRI value in the PATCH payload
3. This happens here:  vendor/api-platform/core/src/Symfony/Routing/IriConverter.php:95
```
        if ($item = $this->provider->provide($operation, $uriVariables, $context)) {
            return $item;
        }

        throw new ItemNotFoundException(sprintf('Item not found for "%s".', $iri));
```
4. The problem is that it doesn't pass the correct ID, instead of the embedded resource ID the ID which is passed to the provider is the initial id of the PUT object, so the above code will always throw an exception.

**How to reproduce**  
1. Create two resources, for example Alpha and Beta
2. Create a field in Beta which is a relation to Alpha
3. Try to create/update Beta
```
PATCH /betas/{id}
{
  "alpha": "/alphas/{id}"
}
```
It will throw 400 error.  And it basically throws ItemNotFoundException due to wrong identifier being passed to state provider.

mdieudonne:
Yeah, I came to the same conclusion there:
https://github.com/api-platform/core/issues/5587

PR #5604, opened 2023-05-29
fix(serializer): put replaces embed collection

### LINKED ISSUE #5587: 3.1.11 - PUT does not work with ManyToMany deep nested relations
**API Platform version(s) affected**: `3.1.11`

**Description**  
`PATCH` does not work with ManyToMany deep nested relations.
But now neither does PUT `standard_put: true` in `3.1.11`

**Possible Solution**  
Revert latest changes made in `ItemNormalizer.php`

soyuka:
fixes #5587 #5603

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -495,6 +495,11 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
         $collectionKeyType = $type->getCollectionKeyTypes()[0] ?? null;
         $collectionKeyBuiltinType = $collectionKeyType?->getBuiltinType();
+        $childContext = $this->createChildContext(['resource_class' => $className] + $context, $attribute, $format);
+        unset($childContext['uri_variables']);
+        if ($this->resourceMetadataCollectionFactory) {
+            $childContext['operation'] = $this->resourceMetadataCollectionFactory->create($className)->getOperation();
+        }
 
         $values = [];
         foreach ($value as $index => $obj) {
@@ -502,7 +507,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                 throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The type of the key "%s" must be "%s", "%s" given.', $index, $collectionKeyBuiltinType, \gettype($index)), $index, [$collectionKeyBuiltinType], ($context['deserialization_path'] ?? false) ? sprintf('key(%s)', $context['deserialization_path']) : null, true);
             }
 
-            $values[$index] = $this->denormalizeRelation($attribute, $propertyMetadata, $className, $obj, $format, $this->createChildContext($context, $attribute, $format));
+            $values[$index] = $this->denormalizeRelation($attribute, $propertyMetadata, $className, $obj, $format, $childContext);
         }
 
         return $values;
@@ -753,7 +758,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             return $value;
         }
 
-        if (null === $value && ($type->isNullable() || ($context[static::DISABLE_TYPE_ENFORCEMENT] ?? false))) {
+        if (null === $value && $type->isNullable() || ($context[static::DISABLE_TYPE_ENFORCEMENT] ?? false)) {
             return $value;
         }
 
@@ -773,7 +778,6 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             && $this->resourceClassResolver->isResourceClass($className)
         ) {
             $resourceClass = $this->resourceClassResolver->getResourceClass(null, $className);
-            $context['resource_class'] = $resourceClass;
 
             return $this->denormalizeCollection($attribute, $propertyMetadata, $type, $resourceClass, $value, $format, $context);
         }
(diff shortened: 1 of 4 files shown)


mdieudonne:
This works for me! Thanks.

jelovac:
Hi, when can we expect a release for this fix?

jonag:
Hello @soyuka, I have updated API Platform to the last version (v3.1.13) and I think that this PR introduced a BC Break.

In order to get arround [this issue](https://github.com/api-platform/core/issues/5559) I wrote a custom Normalizer to unset the `OBJECT_TO_POPULATE` and thus allowing me to update a collection operation while PATCHing a resource.

<details>
  <summary>Here is my custom Normalizer:</summary>

  ```php
  <?php
  
  namespace App\Normalizer;
  
  use ApiPlatform\Metadata\Patch;
  use Doctrine\Common\Collections\Collection;
  use Symfony\Component\DependencyInjection\Attribute\AsDecorator;
  use Symfony\Component\DependencyInjection\Attribute\AutowireDecorated;
  use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
  use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
  use Symfony\Component\Serializer\SerializerAwareInterface;
  
  #[AsDecorator('api_platform.jsonld.normalizer.item')]
  final readonly class AllowPatchEmbeddedCollectionRelationNormalizer implements NormalizerInterface, DenormalizerInterface, SerializerAwareInterface
  {
      public function __construct(
          #[AutowireDecorated] private NormalizerInterface&DenormalizerInterface $decorated,
      ) {
      }
  
      public function supportsNormalization($data, $format = null, array $context = []): bool
      {
          return $this->decorated->supportsNormalization($data, $format);
      }
  
      /**
       * {@inheritdoc}
       */
      public function normalize($object, $format = null, array $context = []): float|array|\ArrayObject|bool|int|string|null
      {
          return $this->decorated->normalize($object, $format, $context);
      }
  
      /**
       * {@inheritdoc}
       */
      public function denormalize($data, $type, $format = null, array $context = []): mixed
      {
          if (
              ($context['operation'] ?? null) instanceof Patch
              && ($context[AbstractNormalizer::OBJECT_TO_POPULATE] ?? null) instanceof Collection
          ) {
              unset($context[AbstractNormalizer::OBJECT_TO_POPULATE]);
          }
  
          return $this->decorated->denormalize($data, $type, $format, $context);
      }
  
      /**
       * {@inheritdoc}
       */
      public function supportsDenormalization($data, $type, $format = null, array $context = []): bool
      {
          return $this->decorated->supportsDenormalization($data, $type, $format);
      }
  
      /**
       * {@inheritdoc}
       */
      public function setSerializer(SerializerInterface $serializer): void
      {
          if ($this->decorated instanceof SerializerAwareInterface) {
              $this->decorated->setSerializer($serializer);
          }
      }
  }
  ```
</details>

However this PR replaces the original operation with a `Get` operation and therefore breaks my workaround.

Please tell me if I should rather open a new issue to discuss it (I can also upload a repo with my failing test if needed).

Aerendir:
@soyuka , I confirm this change is a BC change: it broke my tests where:

- PUT Resource with POST SubResource
- PUT Resource with PUT SubResource
- PUT Resource with DELETE SubResource

More, it may also be wrong: the problem reported by @mdieudonne seems to relate to the name of placeholders: all his placeholders are named "id": this causes a conflict.

Removing the child context removes the conflict, but also breaks my PR #5546 : the child context is required to get the subresources.

So, maybe that simply changing the name of the placeholders, avoiding using the same name for different placeholders, would have resolved the problem reported by @mdieudonne .

Anyway, this PR makes impossible to use, for resources IDs, placeholders that have a different name than "id" (that is exactly what I fixed with my PR - and that was something known as there was a todo comment that remembered to fix it).

WDYT?

soyuka:
I noted both of your comments they don't seem to have the same concerns... I'll handle these asap, please stay on the previous 3.1.12 for now if you can.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -773,7 +778,6 @@ private function createAndValidateAttributeValue(string $attribute, mixed $value
             && $this->resourceClassResolver->isResourceClass($className)
         ) {
             $resourceClass = $this->resourceClassResolver->getResourceClass(null, $className);
-            $context['resource_class'] = $resourceClass;
```
soyuka, line 776: this looks like this should be reverter

soyuka:
@jonag can you use `root_operation` ?

Aerendir:
> I noted both of your comments they don't seem to have the same concerns... I'll handle these asap, please stay on the previous 3.1.12 for now if you can.

Yes, they are two different problems...

jonag:
> @jonag can you use `root_operation` ?

I should find this key in the context passed to the normalizer? Because this key is currently not set when I dump the context on my custom normalizer.

soyuka:
https://github.com/api-platform/core/blob/92a81f024541054b9322e7457b75c721261e14e0/src/Hydra/Serializer/CollectionNormalizer.php#L92 it should be there when using Hydra, I may add this to the SerializerContextBuilder but I'd like for you to try it first if it's possible?

nxtpge:
Hello @jonag and @soyuka, is the fix still ongoing?

jonag:
Hello @nxtpge, if you are talking about the property `root_operation`, I forgot to answer to @soyuka but it's indeed already available in the context passed to the normalizer.

PR #5608, opened 2023-05-31
Fixes #5607, The schema definition for json+hal is inconsistent for c…

### LINKED ISSUE #5607: JSON+HAL SchemaFactory inconsistency with Serialization
**API Platform version(s) affected**: 3.x ( at least, possibly further back )

**Description**  
The schema factory for json+hal is not compatible for collections.

**How to reproduce**  
Just add the helper for testing functional calls to any resource with a collection and try to verify a collection with `json+hal` match (see example below), it won't.

The inconsistency is caused because of;
- https://github.com/api-platform/core/blob/main/src/Hal/Serializer/CollectionNormalizer.php#L94
- https://github.com/api-platform/core/blob/main/src/Hal/JsonSchema/SchemaFactory.php#L79

**Possible Solution**  
https://github.com/api-platform/core/pull/5608

**Additional Context**  

```php
self::assertResponseStatusCodeSame(200);
self::assertMatchesResourceCollectionJsonSchema(CountryApiResource::class, '_api_/countries{._format}_get_collection', 'jsonhal');
```

Returns;

```
_embedded: Object value found, but an array is required
```

wtfzdotnet:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5607 
| License       | MIT
| Doc PR        | N/A

Possibly applies to older versions.

### DIFF
--- a/src/Hal/JsonSchema/SchemaFactory.php
+++ b/src/Hal/JsonSchema/SchemaFactory.php
@@ -76,7 +76,7 @@ final class SchemaFactory implements SchemaFactoryInterface
             $schema['properties'] = [
                 '_embedded' => [
                     'type' => 'array',
-                    'items' => $items,
+                    'item' => $items,
                 ],
                 'totalItems' => [
                     'type' => 'integer',
--- a/tests/Hal/JsonSchema/SchemaFactoryTest.php
+++ b/tests/Hal/JsonSchema/SchemaFactoryTest.php
@@ -119,6 +119,7 @@ class SchemaFactoryTest extends TestCase
         // @noRector
         $this->assertTrue(isset($resultSchema['properties']));
         $this->assertArrayHasKey('_embedded', $resultSchema['properties']);
+        $this->assertArrayHasKey('item', $resultSchema['properties']['_embedded']);
         $this->assertArrayHasKey('totalItems', $resultSchema['properties']);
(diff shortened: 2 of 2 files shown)


soyuka (approved):
Nice, failing tests look unrelated.

src/Hal/JsonSchema/SchemaFactory.php
```diff
@@ -76,7 +76,7 @@ public function buildSchema(string $className, string $format = 'jsonhal', strin
             $schema['properties'] = [
                 '_embedded' => [
                     'type' => 'array',
-                    'items' => $items,
+                    'item' => $items,
```
soyuka, line 79: Actually this schema is wrong see: https://json-schema.org/understanding-json-schema/reference/array.html

it should be something like : 
```json
"_embedded": {
  "type": "object",
  "properties": {
      "item": {"type": "array", "items": $items }
  }
}
```

where items are the json schema of the array items. 

About `item`, indeed it's quite confusing, it should be a name like `contacts` or `dummies`. We'll try to fix this in the coming year but this will definitely break and is quite hard to do right now.

soyuka: can you fix it regarding or isn't there anything to fix?

Issue #5610, opened 2023-06-02
Improve serializer performance with getSupportedTypes


Oipnet:
**Description**  
Symfony 6.3 add getSupportedTypes to improve serializer performance.

**Example**  
Read this article : https://symfony.com/blog/new-in-symfony-6-3-performance-improvements#improved-performance-of-serializer-normalizers-denormalizers

burned42:
Just to have this noted down here, this is not only about improving performance, but since not implementing the `getSupportedTypes` method is deprecated since symfony/serializer 6.3, my deprecation log get flooded with those deprecation messages so much that I had to turn the deprecation log of so it doesn't fill up my disk :D

mvhirsch:
Since #5672 got merged and [3.1.13](https://github.com/api-platform/core/releases/tag/v3.1.13) got released last week, I think we can close this?

Renrhaf:
Excellent thank you ! Deprecation logs are gone after upgrading and adapting our own normalizers :)

PR #5611, opened 2023-06-02
[WIP] Update AbstractConstraintViolationListNormalizer.php

Oipnet:
Add getSupportedTypes method

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5610
| License       | MIT

See the article https://symfony.com/blog/new-in-symfony-6-3-performance-improvements#improved-performance-of-serializer-normalizers-denormalizers

_In Symfony 6.3, we've added a getSupportedTypes(?string $format): array method to normalizers/denormalizers so they can declare the type of objects that they can handle and whether they are cacheable._

Need help for others Normalizers

### DIFF
--- a/src/Serializer/AbstractConstraintViolationListNormalizer.php
+++ b/src/Serializer/AbstractConstraintViolationListNormalizer.php
@@ -45,6 +45,13 @@ abstract class AbstractConstraintViolationListNormalizer implements NormalizerIn
     {
         return static::FORMAT === $format && $data instanceof ConstraintViolationListInterface;
     }
+    
+    public function getSupportedTypes(?string $format)
+    {
+        return [
+            ConstraintViolationListInterface::class => $this->hasCacheableSupportsMethod(),
+        ];
+    }
 
     public function hasCacheableSupportsMethod(): bool
     {


soyuka:
If we're not using the $context in the support method it's quite easy to add in other normalizers, I'm on the symfony slack if you need more informations!

src/Serializer/AbstractConstraintViolationListNormalizer.php
```diff
@@ -45,6 +45,13 @@ public function supportsNormalization(mixed $data, string $format = null, array
     {
         return static::FORMAT === $format && $data instanceof ConstraintViolationListInterface;
     }
+    
+    public function getSupportedTypes(?string $format)
+    {
+        return [
+            ConstraintViolationListInterface::class => $this->hasCacheableSupportsMethod(),
+        ];
+    }
```
MaSpeng, line 54: I would recommend taking the format into account to avoid unwanted side effects and improve the performance event further, how about this one:

```suggestion
    /**
     * @return array<class-string<ConstraintViolationListInterface::class>, bool>|list<mixed>
     */
    public function getSupportedTypes(?string $format)
    {
        return match($format) {
            static::FORMAT => [
                ConstraintViolationListInterface::class => $this->hasCacheableSupportsMethod(),
            ],
            default => [];
        }
    }
```

Also since the `CacheableSupportsMethodInterface` is also deprecated it might be better to avoid right out of the gate and directly change it to:

```suggestion
    /**
     * @return array<class-string<ConstraintViolationListInterface::class>, true>|list<mixed>
     */
    public function getSupportedTypes(?string $format) {
        return match($format) {
            static::FORMAT => [
                ConstraintViolationListInterface::class => true,
            ],
            default => [];
        }
    }
```

MaSpeng (commented):
Thanks for your initiative!

src/Serializer/AbstractConstraintViolationListNormalizer.php
Oipnet: For a moment, I thought of deleting the "hasCacheableSupportsMethod", but I was afraid it would cause backward compatibility problems.

Yes, since the "CacheableSupportsMethodInterface" interface has become deprecated, I suppose that's the wisest thing to do.

MaSpeng: This is not exactly what I meant, in the first iteration we should add the getSupportedTypes method to handle the deprecation from Symfony 6.3.

We cannot directly drop this method and interface as this would break with < Symfony 6.3.

I'am not entirely sure how this deprecation has to be handled inside API Platform to keep the backward compatibility, maybe a new minor version is required which indicates a conflict with Symfony <6.3 to be able to drop the usage of the "CacheableSupportsMethodInterface".

I'am happy to here a suggestion from the maintainers.

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -89,12 +92,25 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
     public function hasCacheableSupportsMethod(): bool
     {
+        trigger_deprecation('symfony/serializer', '6.3', 'The "%s()" method is deprecated, use "getSupportedTypes()" instead.', __METHOD__);
+
         return true;
     }
 
+    public function getSupportedTypes(?string $format)
+    {
+        return [
```
Oipnet, line 106: I'm really not sure about this modification.

I've tried to understand the purpose of the getSupportedTypes method and adapted it to our class.

I'm waiting for your feedback to continue on the subject.

src/Serializer/AbstractCollectionNormalizer.php
```diff
@@ -54,11 +57,28 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
     {
+        trigger_deprecation('symfony/serializer', '6.3', 'The "%s()" method is deprecated, use "getSupportedTypes()" instead.', __METHOD__);
+
         return true;
     }
 
+    public function getSupportedTypes(?string $format)
+    {
+        return match($format) {
+            static::FORMAT => [
```
Oipnet, line 73: I'm really not sure about this modification.

I've tried to understand the purpose of the getSupportedTypes method and adapted it to our class.

I'm waiting for your feedback to continue on the subject.

MaSpeng: See https://github.com/api-platform/core/pull/5611#discussion_r1221524497

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -89,12 +92,25 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
+            HalItemNormalizer::class => true,
+            JsonApiItemNormalizer::class => true,
+            JsonLdItemNormalizer::class => true,
+            ItemNormalizer::class => true,
+        ];
```
MaSpeng, line 111: I think this must return `['object' => false]`.

The description of the `getSupportedTypes` method indicates `a result cannot be cached when it depends on the context or on the data.` which the `AbstractItemNormalizer` does.

Generally, this method should return which objects are supported by the normalizer and these listed classes are normalized on their own which, I think, will never be normalized itself.

src/Serializer/AbstractCollectionNormalizer.php
```diff
@@ -54,11 +57,28 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
+                HalCollectionNormalizer::class => true,
+                HydraCollectionNormalizer::class => true,
+                JsonApiCollectionNormalizer::class => true,
+            ],
+            default => [],
+        };
```
MaSpeng, line 79: I think this must return `['*' => false]`.

The description of the `getSupportedTypes` method indicates `a result cannot be cached when it depends on the context or on the data.` which the `AbstractCollectionNormalizer` does, it verifies that the provided data is iterable.

mrossard:
I'm not sure you're going the right way about this - the supported types should be the ApiResources you want the normalizers to support, not nromalizer instances.

Depending on how symfony does this, it might be done when initializing operations - you'd register the classes you want supported on the right normalizer?

john-dufrene-dev:
Someone have an exemple how to remove this depreciation ? 

Thank in advance,

MaSpeng:
@john-dufrene-dev this is the best example from the [Symfony blog article](https://symfony.com/blog/new-in-symfony-6-3-performance-improvements). The documentation section about [performance and custom normalizers](https://symfony.com/doc/current/serializer/custom_normalizer.html#performance) is still describing the use of the deprecated methods.

This should be a good starting point to remove the deprecation in your own normalizers, for API Platform, at least the item normalizer seems to be a bit more complicated as it discovers the ability to support the provided data by evaluating if it's an API resource.

I think this normalizer has either no catchable supports method at the moment, based on this behavior, or another solution could be to discover all API resources in a compiler pass and pass this knowledge to the normalizer to look up, but I am not sure how this will behave if you add or remove an API Resource during development.

Also, this could still be a solution for a production environment where this will not change over time.

john-dufrene-dev:
@MaSpeng 

Okay thanks for your comment,

But in some project I don't use my own serialiser just default serializer but I have dépréciation.

Just need to wait New release with PR so if I understand ?

Sorry for m'y Bad english

MaSpeng:
@john-dufrene-dev You are right, if you are facing indirect deprecations, from API Platform in this case, you have to wait for a release that fixes those.

john-dufrene-dev:
@MaSpeng 

Thanks for informations :)

src/Serializer/AbstractCollectionNormalizer.php
```diff
@@ -54,11 +57,28 @@ public function supportsNormalization(mixed $data, string $format = null, array
         return static::FORMAT === $format && is_iterable($data);
     }
 
+    /**
+     * @deprecated since Symfony 6.3, use "getSupportedTypes(?string $format)" instead
```
chalasr, line 61: Since API Platform ...

src/Serializer/AbstractCollectionNormalizer.php
```diff
@@ -54,11 +57,28 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
+     */
     public function hasCacheableSupportsMethod(): bool
     {
+        trigger_deprecation('symfony/serializer', '6.3', 'The "%s()" method is deprecated, use "getSupportedTypes()" instead.', __METHOD__);
```
chalasr, line 65:
```suggestion
        trigger_deprecation('api-platform/core', '3.2', 'The "%s()" method is deprecated, use "getSupportedTypes()" instead.', __METHOD__);
```
Same for all notices (using the right package name) as this is API Platform's code

chalasr (changes requested):

chalasr:
Since api-platform supports symfony/serializer 6.1 and higher, the `CacheableSupportsMethodInterface` methods must not be deprecated now as projects running API Platform with Symfony 6.1 wouldn't be able to adopt the new way given it doesn't exist there. We can only add the new method here and revisit when api-platform requires symfony 6.3+.

mrossard:
hey there @Oipnet , any news on that PR? Is it still ongoing?

mrossard:
I tried working a bit on a separate PR for this since there is no revent activity here.

It is still a draft since i haven't looked into testing this correctly, but it seems to do the job for my use...if you're interested i'd gladly take comments : https://github.com/api-platform/core/pull/5672

dunglas:
Closed in favor of #5672. Thanks for having inited the work @Oipnet.

Issue #5612, opened 2023-06-03
Subresources doesn't respect relation field type


divine:
**API Platform version(s) affected**: 3.1.12

**Description**  
When querying relation results in API Platform version 3.1.12, the subresources do not respect the relation field type.

Possibly related to issue #5269.

**How to reproduce**  

Let's say we've followed up [documentation](https://api-platform.com/docs/core/subresources/#answer-to-a-question) and created a similar Question to Answer

/src/Document/Question.php

```php
<?php

namespace App\Document;

use ApiPlatform\Metadata\ApiResource;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;
use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoDB;

#[MongoDB\Document(collection: 'questions')]
#[ApiResource]
class Question
{
    #[MongoDB\Id]
    public $id;

    #[MongoDB\Field(type: 'string')]
    public ?string $text = null;

    #[MongoDB\ReferenceMany(storeAs: 'id', targetDocument: Answer::class, mappedBy: 'question')]
    public Collection|iterable $answers;

    public function __construct()
    {
        $this->answers = new ArrayCollection();
    }

    public function getId()
    {
        return $this->id;
    }

    public function getText(): ?string
    {
        return $this->text;
    }

    public function setText(string $text): self
    {
        $this->text = $text;

        return $this;
    }

    public function getAnswers()
    {
        return $this->answers;
    }

    public function setAnswers($answers): self
    {
        $this->answers = $answers;

        return $this;
    }
}
```

src\Document\Answer.php

```php

<?php

namespace App\Document;

use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Delete;
use ApiPlatform\Metadata\GetCollection;
use ApiPlatform\Metadata\Link;
use ApiPlatform\Metadata\Patch;
use ApiPlatform\Metadata\Post;
use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoDB;

#[MongoDB\Document(collection: 'answers')]
#[ApiResource(
    uriTemplate: '/questions/{id}/answers',
    operations: [
        new GetCollection(),
        new Post(),
        new Patch(),
        new Delete(),
    ],
    uriVariables: [
        'id' => new Link(toProperty: 'question', fromClass: Question::class),
    ],
)]
class Answer
{
    #[MongoDB\Id]
    public $id;

[...]

PR #5613, opened 2023-06-06
fix: subresources odm relation

divine:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #5612
| License       | MIT

This is a bit "dirty" fix for the issue I was having with subresources in #5612

Let me know if this can be improved.

Thanks!


soyuka:
This looks alright, can you add a behat test covering your use case? 

don't worry about `001 Scenario: Update a concrete resource using abstract resource uri # features/main/crud_abstract.feature:108` we fixed that in Symfony.

Issue #4656, opened 2022-02-14
SearchFilter on an association with a Ulid primary key causes an exception


gnito-org:
**API Platform version(s) affected**: 2.6.8

**Description**  

Article has many Tags. Tag has a primary key ``id`` that's defined as a [Symfony Ulid](https://symfony.com/doc/current/components/uid.html#converting-ulids) type.

A search filter on Article to find articles with a specific Tag causes an exception ``An exception occurred while executing a query: SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type uuid: \"01FVQ6SJ6GPCE97X8JPZ6VV5ZX\``. This is on PostgreSQL.

The exception is not thrown when the ``id`` is a Symfony Uuid type.

**How to reproduce**  

See the reproducer at https://github.com/gnito-org/api-platform-ulid-demo.

Create a Tag, associate it with an Article, then do a collection get on Article with the Tag Ulid in the ``tags`` filter field.

Same exception is thrown when valid Ulid values are entered in the ``tags[]`` filter fields.

**Possible Causes**

Api-platform does not supply the third parameter of the ``QueryBuilder::setParameter()`` method for Ulid fields, as decribed by the [Symfony docs](https://symfony.com/doc/current/components/uid.html#storing-ulids-in-databases).

**Possible Solution**

The easy solution is to insert something like the following into [SearchFilter](https://github.com/api-platform/core/blob/119b05b631aae94a922838c6c3ba2f64618210a5/src/Doctrine/Orm/Filter/SearchFilter.php#L150):

```php
    if ('ulid' === $doctrineTypeField){
        foreach ($values as $key => $uLidValue){
            $uLid = new Ulid($uLidValue);
            $values[$key] = $uLid->toRfc4122();
        }
    }
```
That works... however, that makes API Platform dependent on ``symfony/uid``. 

Will it break compatibility with Symfony 4.4?

gnito-org:
Pinging @dunglas

gnito-org:
A work-around for this issue is to create a special filter:

```php
// src/Filter/UlidAssociationFilter
<?php

namespace App\Filter;

use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\SearchFilter;
use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
use Doctrine\ORM\QueryBuilder;
use Symfony\Component\Uid\Ulid;

[...]

PR #5618, opened 2023-06-08
[RFC] Native UID support in search filter

bpolaszek:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | https://github.com/api-platform/api-platform/issues/1958 https://github.com/api-platform/api-platform/issues/2134 https://github.com/api-platform/core/pull/4313 https://github.com/api-platform/core/issues/4656
| License       | MIT

Hello there,

Given:
- The [wide and growing](https://github.com/api-platform/core/issues?q=ulid) adoption of Ulids as primary keys
- The fact that using Ulids instead of integers as primary keys [is encouraged for security and scalability concerns](https://medium.com/@galopintitouan/auto-increment-is-the-devil-using-uuids-in-symfony-and-doctrine-71763721b9a9)
- The usage of `symfony/uid` is natively supported in API-Platform as of https://github.com/api-platform/core/pull/3715
- Docrtrine's `SearchFilter` **currently doesn't work with Ulids** as they're serialized as strings instead of binaries when used in DQL queries
- Current proposal to tackle this is to create a duplicate `SearchFilter`, specifically for handling Ulids, which is tedious and has to be done for every project, without any kind of standardization
- Fixing this seems at first glance **as easy as this PR**, may seem crappy/hacky at first sight but actually means "we support scalars _or_ Uids, just like Symfony does"

I know there have been lots of dicussions, even PRs in the past regarding this, at a time when the UID component was more or less experimental and not a lot of people used it, but I think things have changed since then.
Is it something you could reconsider?

### DIFF
--- a/src/Doctrine/Common/Filter/SearchFilterTrait.php
+++ b/src/Doctrine/Common/Filter/SearchFilterTrait.php
@@ -19,6 +19,7 @@ use ApiPlatform\Doctrine\Common\PropertyHelperTrait;
 use ApiPlatform\Exception\InvalidArgumentException;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\Uid\AbstractUid;
 
 /**
  * Trait for filtering the collection by given properties.
@@ -124,7 +125,9 @@ trait SearchFilterTrait
             $iriConverter = $this->getIriConverter();
             $item = $iriConverter->getResourceFromIri($value, ['fetch_data' => false]);
 
-            return $this->getPropertyAccessor()->getValue($item, 'id');
+            $id = $this->getPropertyAccessor()->getValue($item, 'id');
+
+            return ($id instanceof AbstractUid) ? $id->toBinary() : $id;
         } catch (InvalidArgumentException) {
             // Do nothing, return the raw value
         }


soyuka:
Had to revert #3774 at https://github.com/api-platform/core/pull/4134 

You're patch looks way better as it doesn't change doctrine type system :+1: to merge!

bpolaszek:
That's good news, but there's a final boss to fight: the CI! 😱 
I don't even know where to start! 😅

bpolaszek:
Sorry for acting as a Behat newbie (I am one), but I can't get the test to pass.

It looks like my `UidBasedId` API Resource is not properly registered as `/uid-based-ids` returns 404 - I actually don't get why.

```
  @createSchema
  Scenario: Get collection by ulid 01H2ZS93NBKJW5W4Y01S8TZ43M                                    # features/doctrine/search_filter.feature:721
    Given there is a UidBasedId resource with id "01H2ZS93NBKJW5W4Y01S8TZ43M"                    # ApiPlatform\Tests\Behat\DoctrineContext::thereIsAUidBasedIdResource()
    When I send a "GET" request to "/uid-based-ids?id=/uid-based-ids/01H2ZS93NBKJW5W4Y01S8TZ43M" # Behatch\Context\RestContext::iSendARequestTo()
    Then the response status code should be 200                                                  # Behat\MinkExtension\Context\MinkContext::assertResponseStatus()
      Current response status code is 404, but 200 expected.
```

features/doctrine/search_filter.feature
```diff
@@ -717,6 +717,45 @@ Feature: Search filter on collections
     }
     """
 
+  @createSchema
+  Scenario: Get collection by ulid 01H2ZS93NBKJW5W4Y01S8TZ43M
+    Given there is a UidBasedId resource with id "01H2ZS93NBKJW5W4Y01S8TZ43M"
+    When I send a "GET" request to "/uid-based-ids?id=/uid-based-ids/01H2ZS93NBKJW5W4Y01S8TZ43M"
```
mrossard, line 723: Shouldn't it just be 
```suggestion
    When I send a "GET" request to "/uid-based-ids?id=01H2ZS93NBKJW5W4Y01S8TZ43M"
```
?

soyuka, line 723:
```suggestion
    When I send a "GET" request to "/uid_based_ids?id=/uid-based-ids/01H2ZS93NBKJW5W4Y01S8TZ43M"
```

soyuka (changes requested):
almost there

soyuka:
your commit could be: `feat(doctrine): search filter uid support`

features/doctrine/search_filter.feature
bpolaszek: Thanks for spotting this! I've been using kebab-case for years, I didn't remember it was not the standard 🙃

soyuka:
Yes its a hard problem :/ especially because we don't have a query parameter sanitizer, it'll be better if you knew it was an IRI or not for starters :/. I wish we could have the identifier Type in that method, and we'd have to find out wheter its an IRI or  not before getting its value.
Also, maybe that there should be an IriSearchFilter and a SearchFilter ^^.

soyuka:
also https://github.com/api-platform/core/pull/5623/files#diff-9a51326378c8abdf577e784792a9245eeb219ffc426b09aa1a635bd20fcc51cdR127-R133

mrossard:
Forgot this PR...does this fix your issue? https://github.com/api-platform/core/pull/5771

bpolaszek:
> Forgot this PR...does this fix your issue? #5771

Sorry still didn't take the time to review this - will have a look ASAP!

bpolaszek:
Unfortunately it's not fixing it.
Consider an `AccountMember` resource with an `Account` relation having a `public Ulid $id` as id.

```http
GET /api/account-members.jsonld?account=%2Fapi%2Faccounts%2F01H4K4R4SY577HK9A9XZG4KAGF
```

Capture d’écran 2023-09-08 à 20 13 21


```http
GET /api/account-members.jsonld?account=01H4K4R4SY577HK9A9XZG4KAGF
```
Capture d’écran 2023-09-08 à 20 14 37

The `$id` gets casted as string instead of a binary, and the query returns 0 result.

mrossard:
> Unfortunately it's not fixing it. [{...]

Sorry, i misunderstood your use case - the UID is actually the orm id for you, which i didn't account for in my PR! I guees both are actually relevant, and I even might have to copy what you're doing here in some part of mine.

mrossard:
One thing I'm not sure about : wouldn't using `type: 'symfony_uuid' in you doctrine column definition help? You're using 'ulid' in your test, but I remember having to use symfony_uuid in mine for this to work.

rubobaquero:
Hello @bpolaszek @mrossard @soyuka ,

Do you have any updates about this PR? Facing this problem in a project that is migrating from ids to UUIDs.

mrossard:
> Hello @bpolaszek @mrossard @soyuka ,
> 
> Do you have any updates about this PR? Facing this problem in a project that is migrating from ids to UUIDs.

I haven't been able to work on this for the last few weeks, sorry.

mrossard:
> stumble on #319 today which would quite simplify the code around these things...

that could help make things clearer, but that would break a lot of things, wouldn't it?

rubobaquero:
Thank you very much for your answers and your work @soyuka and @mrossard . Hope we can see those changes soon. Meanwhile, in case some other people end up here after a Google Search as I did, what is the suggested way to implement now a Search Filter on a foreign key that is stored as a binary UUID? Typical case of filtering a Book entity by Author UUID fr example.

Thanks in advance!

bpolaszek:
> Thank you very much for your answers and your work @soyuka and @mrossard . Hope we can see those changes soon. Meanwhile, in case some other people end up here after a Google Search as I did, what is the suggested way to implement now a Search Filter on a foreign key that is stored as a binary UUID? Typical case of filtering a Book entity by Author UUID fr example.
> 
> Thanks in advance!

Here you go:
```php
<?php

declare(strict_types=1);

namespace App\State\Filter;

use ApiPlatform\Doctrine\Orm\Filter\AbstractFilter;
use ApiPlatform\Doctrine\Orm\Util\QueryNameGeneratorInterface;
use Doctrine\ORM\QueryBuilder;
use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
use Symfony\Component\Uid\Ulid;
use Throwable;

use function array_map;
use function is_array;
use function sprintf;
use function str_ends_with;
use function Symfony\Component\String\u;

final class UlidFilter extends AbstractFilter
{
    /**
     * @param string|string[]|null $value
     */
    protected function filterProperty(
        string $property,
        mixed $value,
        QueryBuilder $queryBuilder,
        QueryNameGeneratorInterface $queryNameGenerator,
        string $resourceClass,
        Operation $operation = null,
        array $context = [],
    ): void {
        if (
            null === $value
            || !$this->isPropertyEnabled($property, $resourceClass)
            || !$this->isPropertyMapped($property, $resourceClass, true)
        ) {
            return;
        }

        $alias = $queryBuilder->getRootAliases()[0];
        $valueParameter = ':' . $queryNameGenerator->generateParameterName($property);
        $aliasedField = sprintf('%s.%s', $alias, $property);

        if (is_array($value)) {
            $values = array_map(
                fn (string $value) => $this->convertUlidFromStringToBinary($property, $value),
                $value
            );

            $queryBuilder
                ->andWhere($queryBuilder->expr()->in($aliasedField, $valueParameter))
                ->setParameter($valueParameter, $values);

            return;
        }

        $value = $this->convertUlidFromStringToBinary($property, $value);

        $queryBuilder
            ->andWhere($queryBuilder->expr()->eq($aliasedField, $valueParameter))
            ->setParameter($valueParameter, $value);
    }

    private function convertUlidFromStringToBinary(string $property, string $value): string
    {
        if ('/' === ($value[0] ?? '')) {
            $value = (string) u($value)->afterLast('/');
        }

        try {
            return Ulid::fromString($value)->toBinary();
        } catch (Throwable) {
            throw new BadRequestHttpException("`{$property}`: Unable to parse Ulid `{$value}`");
        }
    }

    /**
     * @return array<string, array<string, mixed>>
     *
     * @codeCoverageIgnore
     */
    {
        if (!$this->properties) {
            return [];
        }

        $description = [];

        foreach ($this->properties as $property => $strategy) {
            if (
                !$this->isPropertyEnabled($property, $resourceClass)
                || !$this->isPropertyMapped($property, $resourceClass, true)
            ) {
                continue;
            }

            /** @var string[] $filterParameterNames */
            $filterParameterNames = [$property, "{$property}[]"];

            foreach ($filterParameterNames as $filterParameterName) {
                $description[$filterParameterName] = [
                    'property' => $property,
                    'type' => 'string',
                    'required' => false,
                    'is_collection' => str_ends_with($filterParameterName, '[]'),
                ];
            }
        }

        return $description;
    }
}
```

Usage:

```php
#[ApiFilter(UlidFilter::class, properties: ['someRelation' => null])]
```

soyuka:
This is awesome, I would like this to be part of API Platform but I'd also like for it not to be added to the Search filter... It's not possible right now I'll work on finding a way to do so.

rubobaquero:
Thanks a lot @bpolaszek !

soyuka:
@bpolaszek can you update this? I would like to merge that in the end.

soyuka:
I tried a quick rebase let's see.

soyuka:
PostgreSQL and MongoDB have issues with the fix, if someone is willing to take a look we'll gladly merge this.

Issue #5605, opened 2023-05-31
SearchFilter shouldn't rely on nested resources having an "id" field


mrossard:
**API Platform version(s) affected**: 3.1.12

**Description**  
When you use a SearchFilter on a nested property that is a resource and doesn't use "id" as its identifier you get "Can't get a way to read the property \"id\" in class XXX".

**How to reproduce**  
Have 2 resources A and B, with A containing a B (both simple ApiResources using entityClass=... in my case)
Use an ApiIdentifier in B that is not "id", ie IRIs are /Bs/{someotherfield}.
Add a searchfilter for the B member on A : 
```php
#[ApiFilter(SearchFilter::class, properties:['b'])
```

**Possible Solution**  
Didn't have time to dig into this yet, but it's happening in SearchFilterTrait line 127 : 
```php
return $this->getPropertyAccessor()->getValue($item, 'id');
```
This should probably be replaced so it at least looks for id in the entityClass when present, or even tries to get the ORM to return the correct id(s) without hard coding it.

j-schumann:
Similar issue here w/ same version. Only difference: Problem occurs when the resourceClass has an identifier *not* named "id", but the association's identifier is indeed "id", as [Doctrine\Orm\Filter\SearchFilter#L126](https://github.com/api-platform/core/blob/3.1/src/Doctrine/Orm/Filter/SearchFilter.php#L126) re-uses the resource's identifier name(s).

https://github.com/api-platform/core/pull/5623 in its current state would fix this.

mrossard:
There are indeed 2 issues here - the hardcoded "id", and then the main resource metada being reused instead of the association's. I'll try to get my fix up to standards so it gets merged.

PR #5623, opened 2023-06-10
fix: search on nested sub-entity that doesn't use "id" as its ORM id

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | https://github.com/api-platform/core/issues/5605

This should fix that use case when ids are not composite. 
Composite IDs would still need to be addressed with a custom filter, but that's even more of a fringe case...and would require quite of lot of work as far as i can see. I still left todo comments in case someone wants to tackle that.

### DIFF
--- a/src/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Doctrine/Orm/Filter/SearchFilter.php
@@ -100,6 +100,7 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
         if ($metadata->hasField($field)) {
             if ('id' === $field) {
                 $values = array_map($this->getIdFromValue(...), $values);
+                // todo: handle composite IDs
             }
 
             if (!$this->hasValidValues($values, $this->getDoctrineFieldType($property, $resourceClass))) {
@@ -121,9 +122,11 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
         }
 
         $values = array_map($this->getIdFromValue(...), $values);
+        // todo: handle composite IDs
 
         $associationResourceClass = $metadata->getAssociationTargetClass($field);
-        $associationFieldIdentifier = $metadata->getIdentifierFieldNames()[0];
+        $associationMetadata = $this->getClassMetadata($associationResourceClass);
+        $associationFieldIdentifier = $associationMetadata->getIdentifierFieldNames()[0];
         $doctrineTypeField = $this->getDoctrineFieldType($associationFieldIdentifier, $associationResourceClass);
 
         if (!$this->hasValidValues($values, $doctrineTypeField)) {
--- a/src/Symfony/Bundle/Resources/config/doctrine_orm.xml
+++ b/src/Symfony/Bundle/Resources/config/doctrine_orm.xml
@@ -152,7 +152,7 @@
             <argument type="service" id="api_platform.iri_converter" />
             <argument type="service" id="api_platform.property_accessor" />
             <argument type="service" id="logger" on-invalid="ignore" />
-            <argument key="$identifiersExtractor" type="service" id="api_platform.identifiers_extractor.cached" on-invalid="ignore" />
+            <argument key="$identifiersExtractor" type="service" id="api_platform.identifiers_extractor" on-invalid="ignore" />
             <argument key="$nameConverter" type="service" id="api_platform.name_converter" on-invalid="ignore" />
         </service>
         <service id="ApiPlatform\Doctrine\Orm\Filter\SearchFilter" alias="api_platform.doctrine.orm.search_filter" />
--- a/tests/Fixtures/TestBundle/ApiResource/Issue5605/MainResource.php
+++ b/tests/Fixtures/TestBundle/ApiResource/Issue5605/MainResource.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue5605;
+
+use ApiPlatform\Doctrine\Orm\Filter\SearchFilter;
+use ApiPlatform\Doctrine\Orm\State\Options;
+use ApiPlatform\Metadata\ApiFilter;
+use ApiPlatform\Metadata\ApiProperty;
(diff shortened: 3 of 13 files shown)


mrossard:
Just noticed https://github.com/api-platform/core/pull/5618 - not much work to merge as far as i see, but both PRs share a goal to be able to use identifiers that are not always "int $id".

src/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -41,7 +41,7 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
 
     public const DOCTRINE_INTEGER_TYPE = Types::INTEGER;
 
-    public function __construct(ManagerRegistry $managerRegistry, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null, IdentifiersExtractorInterface $identifiersExtractor = null, NameConverterInterface $nameConverter = null)
+    public function __construct(ManagerRegistry $managerRegistry, IriConverterInterface $iriConverter, PropertyAccessorInterface $propertyAccessor = null, LoggerInterface $logger = null, array $properties = null, IdentifiersExtractorInterface $identifiersExtractor, NameConverterInterface $nameConverter = null)
```
soyuka, line 44: please revert this needs to be optional

src/Symfony/Bundle/Resources/config/doctrine_orm.xml
```diff
@@ -152,7 +152,7 @@
             <argument type="service" id="api_platform.iri_converter" />
             <argument type="service" id="api_platform.property_accessor" />
             <argument type="service" id="logger" on-invalid="ignore" />
-            <argument key="$identifiersExtractor" type="service" id="api_platform.identifiers_extractor.cached" on-invalid="ignore" />
```
soyuka, line 155: this was correct

tests/Fixtures/TestBundle/State/Issue5605/MainResourceProvider.php
```diff
@@ -0,0 +1,56 @@
...
+namespace ApiPlatform\Tests\Fixtures\TestBundle\State\Issue5605;
+
+use ApiPlatform\Metadata\Get;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProviderInterface;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue5605\MainResource;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue5605\SubResource;
+use ApiPlatform\Tests\Fixtures\TestBundle\Entity\DummyWithSubEntity;
+
+class MainResourceProvider implements ProviderInterface
```
soyuka, line 23: do you really need this? isn't the default provider enough to test?

tests/Fixtures/TestBundle/ApiResource/Issue5605/MainResource.php
```diff
@@ -0,0 +1,41 @@
...
+    provider    : MainResourceProvider::class,
+    stateOptions: new Options(entityClass: DummyWithSubEntity::class)
+)]
+#[ApiFilter(SearchFilter::class, properties: ['subEntity'])]
+class MainResource
+{
+    #[ApiProperty(identifier: true)]
+    public int $id;
+    public string $name;
+    public SubResource $subResource;
```
soyuka, line 40: if you want this to be automatically fetch I think that you need serialization groups so that the EagerLoading extension does its job.

src/Doctrine/Orm/Filter/SearchFilter.php
mrossard: If i leave it optional, calling the identifiersExtractor in the SearchFilterTrait doesn't work - or at least it didn't while running my tests.

tests/Fixtures/TestBundle/State/Issue5605/MainResourceProvider.php
mrossard: I got used to always add custom providers when i use entityClass, but i'll check if it's actually useful here.

mrossard: Just checked - as is if i remove the custom provider i get "Can't get a way to read the property "subResource" in class ...".

And if i rename $subResource to $subEntity so it matches the entities, i get "No resource class found for object of type "ApiPlatform\Tests\Fixtures\TestBundle\Entity\DummySubEntity"."

I don't see an easy way to do this without custom providers.

tests/Fixtures/TestBundle/ApiResource/Issue5605/MainResource.php
mrossard: see above, serialization groups can't really help here - or am i missing something?

src/Symfony/Bundle/Resources/config/doctrine_orm.xml
mrossard: see above, i just get  "Call to a member function getIdentifiersFromItem() on null" if if keep it nullable and that conf.

soyuka: check if its not null before calling the class, if its null just use the old way

mrossard: Is it a backwards compatibility problem? Would injecting it in a setter be ok then?

mrossard: I updated it that way, is it ok or am i missing something?

tests/Fixtures/TestBundle/State/Issue5605/MainResourceProvider.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
+    {
+        if ($operation instanceof Get) {
+            /**
+             * @var DummyWithSubEntity $entity
+             */
+            $entity = $this->itemProvider->provide($operation, $uriVariables, $context);
```
soyuka, line 35: I still want to check why this is needed.

mrossard: I think a simple explanation is that entityClass + gives us the link from the ApiResource to the entity, but nothing really tells us how to convert an entity to a resource, so when the framework gets a subentity it doesn't have a way to know which resource it's supposed to convert it to if you don't do it explicitly in a custom provider.

soyuka:
Thanks @mrossard !

PR #5629, opened 2023-06-13
fix: urlencode "+" sign before processing query parameters

LoicBoursin:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

When sending a request with a date in DATE_ATOM (e.g. `2000-01-01T00:00:00+00:00`) format, the plus sign was stripped and replaced by a space. 
This occurred in tests when sending the unencoded request to a client.

This PR adds tests for the `getQueryString` method and fixes the issue.

Tests are failing but it seems unrelated

### DIFF
--- a/src/Util/RequestParser.php
+++ b/src/Util/RequestParser.php
@@ -79,6 +79,11 @@ final class RequestParser
 
             $keyValuePair = explode('=', $param, 2);
 
+            // Encode "+" sign so it's not decoded to a space afterward.
+            if (isset($keyValuePair[1])) {
+                $keyValuePair[1] = str_replace('+', '%2B', $keyValuePair[1]);
+            }
+
             // GET parameters, that are submitted from a HTML form, encode spaces as "+" by default (as defined in enctype application/x-www-form-urlencoded).
             // PHP also converts "+" to spaces when filling the global _GET or when using the function parse_str. This is why we use urldecode and then normalize to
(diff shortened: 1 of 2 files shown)


soyuka:
But your date should be url encoded no?

LoicBoursin:
When I call the client in tests with an URL, like:

```PHP
$this->client->request(
            'GET',
            \sprintf('api/profiles/%s/classifieds?classifiedUuids[]=%s&classifiedUuids[]=%s&since=%s', UuidV4::v4(), UuidV4::v4(), UuidV4::v4(), $since = (new \DateTime('2021-06-09'))->format(\DATE_ATOM)),
        );
```

I could URLencode the date but it does not make sense to URLencode just the date and not the rest of the query parameters. The rest of the query parameters such as brackets are correctly handled, but my date becomes unparsable at the API Platform level when it goes through that request parser. 

I believe web browsers encode dates/urls before it's processed by API Platform but it's not the case in tests and it would make the development process much more complicated IMO

soyuka:
I really think that this patch is wrong and that you should url encode your values.

dunglas:
@soyuka is right, you must URL-encode the query parameters. With Symfony HttpClient you can use the `query` option to do that: https://symfony.com/doc/current/http_client.html#query-string-parameters

Anyway, thanks for contributing!

Issue #5620, opened 2023-06-08
Upgrade path from v2 to v3


n-valverde:
**API Platform version(s) affected**: 2.7

**Description**  
Hello,

I am currently migrating a big application from v2 to v3, and I am a bit disappointed by how hard the upgrade path is. 
I truly understand the intentions of the v3, but I think the upgrade path from v2 should receive more attention, as there is probably a lot of app still in v2 out there. That leads me to the first point:

* While api-platform follows Symfony's BC promise, it does not follow the same release strategy, i.e the v2.7 is not considered "long-term". 
 
I don't want to bring a debate about LTS, but I do want to highlight that [the documentation clearly states](https://api-platform.com/docs/extra/releases/#maintenance) that 2.7 is maintained for *security fixes*  only, which leaves people like me understanding that nothing will get fixed in 2.7, even if that prevents a smooth upgrade path. 
That's probably not completely true, and I'm assuming maintainers are judging on a case by case basis, but I also came across a few closed issues were the output was: it's 2.7 so won't fix. 
But today, @mtarld gave a nice little talk, where he said you would happily accept contributions to improve the upgrade path 😄 so here I am, still confused but optimistic 😄.

So first question: can we contribute on v2.7 to improve the upgrade path? If yes, can we update the documentation to make it clear the api-platform is not leaving people behind. I'd be happy to contribute, but as I said right now I'm just confused on whether we can contribute to 2.7 or not.

***If the answer of the above is no, no need to read further 🙃***

***The below is a summary of issues I've encountered so far and where I'd be happy to contribute, just don't want to spam new issues***

* `bin/console api:upgrade-resource`

That command is mostly broken, because it **completely** mess up the formatting, and cs-fixer is clearly not enough to fix the mess. In a big application, where the command needs to alter roughly 200 files, that's an issue. **It currently alter source code, while it's supposed to deal with annotations and attributes, but is adding/removing parenthesis in method bodies, adding/removing whitespaces and blanklines litterally everywhere.**

I left a [little fix for that in another closed issue](https://github.com/api-platform/core/issues/4973#issuecomment-1579211095) (not perfect but enough to make the command usable). But that also brings the question of: why not just rector?

* Rector rules

Apart from the above question of why not rector for the upgrade resource command, I am also confused to not find any rector rule out there helping with the namespace updates. There are a lot of namespace changes, some are very simple to fix, others pretty hard, but it would be nice to have some rector rules to help. Did I miss them somewhere?

[...]

PR #5630, opened 2023-06-13
fix: IriConverterInterface DI & deprecation

n-valverde:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | #5620 
| License       | MIT

Hello, proposing a fix for an issue with dependency injection of  `ApiPlatform\Api\IriConverterInterface` & deprecation of `ApiPlatform\Core\Api\IriConverterInterface`. I'm open to discussion if you don't like the proposed solution ofc 🙂 

## The problem:

When BC flag is ON, we receive this deprecation notice
> Since api-platform/core 2.7: The service "ApiPlatform\Core\Bridge\Symfony\Routing\IriConverter" is deprecated, use "ApiPlatform\Symfony\Routing\IriConverter" instead.

There's a few issues with this deprecation imo:

### About DI:
* The message only mentions the concrete class, while user is probably referencing interfaces or service id `api_platform.iri_converter`. Actually both mentioned services does not exist with this id in container, which is confusing.
* Trying to inject the suggested service `ApiPlatform\Symfony\Routing\IriConverter` will crash at compile time because this service does not exist (which is expected)
* Trying to inject the interface implemented by this service `ApiPlatform\Api\IriConverterInterface`, as suggested by the compile time error message (service not existing, try changing the type hint to ...), leads to a type error at *runtime*, due to a misconfiguration (I think) the service referenced by alias `ApiPlatform\Api\IriConverterInterface` does not implement that interface. It is currently referencing `api_platform.iri_converter` which in turns reference `api_platform.iri_converter.legacy` which implements the old `ApiPlatform\Core\Api\IriConverterInterface`
* Trying to inject manually from the service id `api_platform.symfony.iri_converter` leads to another compile time error, because this service has a dependency on non existant services (the new metadata system)

At the end, we cannot get rid of the deprecation as long as BC flag is kept ON, i.e cannot use the new interface, which I think is an issue for a smooth upgrade.

### About deprecation, having the deprecation triggered from constructor has a few drawbacks:
* The deprecation does not show up in userland when running `bin/console debug:container --deprecations`
* The deprecation is triggered when going through the concrete class, while we should be interested in triggering deprecations for *references* to the service, since it can legitimately be used through `ApiPlatform\Core\Api\LegacyIriConverter` (which is now used with BC flag ON with the proposed DI solution below)
* It does not adhere to [Symfony convention](https://symfony.com/doc/current/contributing/code/conventions.html#deprecating-code) (When deprecating a whole class the trigger_deprecation() call should be placed after the use declarations)

### About DI:
Make service `api_platform.symfony.iri_converter` available through `ApiPlatform\Api\IriConverterInterface` when BC flag is ON by using an implementation compatible with old metadata and new interface, e.g `ApiPlatform\Core\Api\LegacyIriConverter`.

### About deprecation:
Mark the old aliases/services as deprected instead of triggering deprecation from constructor (adding a compiler pass to deal with the different versions of `symfony/dependency-injection`). The has the advantage of showing all userland references to the deprecated service when running `bin/console debug:container --deprecations`, and users can easily get rid of the deprecation just by using the new interface

### DIFF
--- a/src/Symfony/Bundle/DependencyInjection/Compiler/DeprecateLegacyIriConverterPass.php
+++ b/src/Symfony/Bundle/DependencyInjection/Compiler/DeprecateLegacyIriConverterPass.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Symfony\Bundle\DependencyInjection\Compiler;
+
+use ApiPlatform\Core\Api\IriConverterInterface;
+use Symfony\Component\Config\Definition\BaseNode;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @internal calls `setDeprecated` method with valid arguments depending on which version of symfony/dependency-injection is used
+ */
+final class DeprecateLegacyIriConverterPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if ($container->hasAlias(IriConverterInterface::class)) {
+            $container
+                ->getAlias(IriConverterInterface::class)
+                ->setDeprecated(...$this->buildDeprecationArgs('Using "%alias_id%" is deprecated since API Platform 2.7. Use "ApiPlatform\Api\IriConverterInterface" instead.'));
+        }
+
+        if ($container->hasDefinition(IriConverterInterface::class)) {
+            $container
+                ->getDefinition(IriConverterInterface::class)
+                ->setDeprecated(...$this->buildDeprecationArgs('Using "%service_id%" is deprecated since API Platform 2.7. Use "ApiPlatform\Api\IriConverterInterface" instead.'));
+        }
+
+        if ($container->hasDefinition('api_platform.iri_converter.legacy')) {
+            $container
+                ->getDefinition('api_platform.iri_converter.legacy')
+                ->setDeprecated(...$this->buildDeprecationArgs('Using "%service_id%" is deprecated since API Platform 2.7. Use "ApiPlatform\Api\IriConverterInterface" instead.'));
+        }
+    }
+
+    private function buildDeprecationArgs(string $message): array
+    {
+        return method_exists(BaseNode::class, 'getDeprecation')
+            ? ['api-platform/core', '2.7', $message]
+            : [$message];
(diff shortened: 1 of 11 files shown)


src/Symfony/Bundle/DependencyInjection/Compiler/DeprecateLegacyIriConverterPass.php
```diff
@@ -0,0 +1,55 @@
...
+
+    private function buildDeprecationArgs(string $message): array
+    {
+        return method_exists(BaseNode::class, 'getDeprecation')
+            ? ['api-platform/core', '2.7', $message]
+            : [$message];
+    }
+}
+
+class_alias(DeprecateLegacyIriConverterPass::class, \ApiPlatform\Core\Bridge\Symfony\Bundle\DependencyInjection\Compiler\DeprecateLegacyIriConverterPass::class);
```
soyuka, line 55:
```suggestion
```

We don't really need this righ?

src/Symfony/Bundle/DependencyInjection/Compiler/DeprecateLegacyIriConverterPass.php
```diff
@@ -0,0 +1,55 @@
...
+            $container
+                ->getDefinition(IriConverterInterface::class)
+                ->setDeprecated(...$this->buildDeprecationArgs('Using "%service_id%" is deprecated since API Platform 2.7. Use "ApiPlatform\Api\IriConverterInterface" instead.'));
+        }
+
+        if ($container->hasDefinition('api_platform.iri_converter.legacy')) {
+            $container
+                ->getDefinition('api_platform.iri_converter.legacy')
+                ->setDeprecated(...$this->buildDeprecationArgs('Using "%service_id%" is deprecated since API Platform 2.7. Use "ApiPlatform\Api\IriConverterInterface" instead.'));
+        }
```
soyuka, line 44: I would not deprecate this service, IMO its here to cover BC in some cases you want to use it.

soyuka: It's weird to introduce a deprecated service no? But indeed this service only helps covering a bc layer so why not! 

> Also by writing this table I'm realizing that ApiPlatform\Core\Api\IriConverterInterface is removed in 2.7 BC OFF, while the service api_platform.iri_converter.legacy still exist for the BC reasons you mentioned, shouldn't we also maintain the alias ApiPlatform\Core\Api\IriConverterInterface to api_platform.iri_converter.legacy in 2.7 BC OFF ?

definitely!

n-valverde:
Hello @soyuka , thanks for the review, we should be good now 🤞 , I guess failing tests are unrelated, let me know if I'm missing anything 🙂

soyuka:
Thank you very much for that improvement, I've a lot of work rn but I'll tag this tomorrow!

PR #5632, opened 2023-06-14
feat: deprecate json support + add openapi format

Sarahshr:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | 
| License       | MIT


dunglas:
I suggest adding a new `docs_format` configuration m allowing to enable formats only for the documentation, in a similar way of what we do for `patch_formats`. This will make explicit that documentation formats are not related to standard resources formats and standard content negotiation but only apply to `/docs`.

Also, I suggest to not deprecate JSON, but disable it by default (in the recipe for 3.x, and by default in 4.x).

Finally, what two OpenAPI formats? `jsonopenapi` for the JSON flavor and `yamlopenapi` for the YAML flavor. Being able to retrieve the YAML description through HTTP is convenient to use some Swagger/OpenAPI tools.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -103,13 +103,20 @@ public function load(array $configs, ContainerBuilder $container): void
...
         $errorFormats = $this->getFormats($config['error_formats']);
+        $docsFormats = $this->getFormats($config['docs_formats']);
 
         // Backward Compatibility layer
         if (isset($formats['jsonapi']) && !isset($patchFormats['jsonapi'])) {
             $patchFormats['jsonapi'] = ['application/vnd.api+json'];
         }
 
-        $this->registerCommonConfiguration($container, $config, $loader, $formats, $patchFormats, $errorFormats);
+        if (!$docsFormats['openapi'] || !isset($docsFormats['json']) !== null) {
```
vincentchalamon, line 113: It's confusing what you're trying to do here

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -163,6 +163,11 @@ public function getConfigTreeBuilder(): TreeBuilder
         $this->addFormatSection($rootNode, 'patch_formats', [
             'json' => ['mime_types' => ['application/merge-patch+json']],
         ]);
+        $this->addFormatSection($rootNode, 'docs_formats', [
+            'graphql' => ['mime_types' => ['application/graphql']],
+            'jsonld' => ['mime_types' => ['application/ld+json']],
+            'openapi' => ['mime_types' => ['application/vnd.openapi+json']],
```
vincentchalamon, line 169: I think `openapi` format should be placed first (as default), then `jsonld`, and finally `html`. `graphql` format should not be available by default on docs

tests/Fixtures/app/config/config_common.yml
```diff
@@ -36,6 +36,10 @@ api_platform:
         html:                          ['text/html']
         graphql:                       ['application/graphql']
         multipart:                     ['multipart/form-data']
+    docs_formats:
+        html:                          ['text/html']
+        jsonld:                        ['application/ld+json']
+        graphql:                       ['application/graphql']
```
vincentchalamon, line 42: IMHO `graphql` is not a valid format for docs, as we already have another tool for it

vincentchalamon (changes requested):

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -103,13 +103,20 @@ public function load(array $configs, ContainerBuilder $container): void
...
         $errorFormats = $this->getFormats($config['error_formats']);
+        $docsFormats = $this->getFormats($config['docs_formats']);
 
         // Backward Compatibility layer
         if (isset($formats['jsonapi']) && !isset($patchFormats['jsonapi'])) {
             $patchFormats['jsonapi'] = ['application/vnd.api+json'];
         }
 
-        $this->registerCommonConfiguration($container, $config, $loader, $formats, $patchFormats, $errorFormats);
+        if (empty($docsFormats) || !isset($docsFormats['openapi'], $docsFormats['json'])) {
```
dunglas, line 113:
```suggestion
        if ($docsFormats || !isset($docsFormats['openapi'], $docsFormats['json'])) {
```
 (Empty is useless)

src/Symfony/Bundle/SwaggerUi/SwaggerUiAction.php
```diff
@@ -33,7 +33,7 @@ final class SwaggerUiAction
...
+    public function __construct(private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly ?TwigEnvironment $twig, private readonly UrlGeneratorInterface $urlGenerator, private readonly NormalizerInterface $normalizer, private readonly OpenApiFactoryInterface $openApiFactory, private readonly Options $openApiOptions, private readonly SwaggerUiContext $swaggerUiContext, private readonly array $formats = [], private readonly array $docsFormats = [], private readonly ?string $oauthClientId = null, private readonly ?string $oauthClientSecret = null, private readonly bool $oauthPkce = false)
```
dunglas, line 36: Isn't this change a BX break?

src/Documentation/Action/DocumentationAction.php
```diff
@@ -27,6 +27,9 @@
  */
 final class DocumentationAction
 {
+    public const OPENAPI_FORMAT = 'openapi';
+    public const JSON_FORMAT = 'json';
```
vincentchalamon, line 31: Those constants should be declared (or maybe already exist) in a normalizer or serializer

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -25,7 +25,8 @@
  */
 final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface
 {
-    public const FORMAT = 'json';
```
vincentchalamon, line 28: Changing this constant name is a BC break. You must keep it

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -72,7 +73,7 @@ private function recursiveClean(array $data): array
      */
     public function supportsNormalization(mixed $data, string $format = null, array $context = []): bool
     {
-        return self::FORMAT === $format && $data instanceof OpenApi;
+        return self::JSON_FORMAT === $format && $data instanceof OpenApi;
```
vincentchalamon, line 76:
```suggestion
        return \in_array($format [self::JSON_FORMAT, self::OPENAPI_FORMAT], true) && $data instanceof OpenApi;
```

vincentchalamon (changes requested):

soyuka (approved):

soyuka:
awesome thanks !

features/openapi/docs.feature
```diff
@@ -319,3 +319,44 @@ Feature: Documentation support
       "nullable":true
     }
     """
+
+  Scenario: Retrieve the json OpenAPI documentation
```
dunglas, line 323:
```suggestion
  Scenario: Retrieve the JSON OpenAPI documentation
```

features/openapi/docs.feature
```diff
@@ -319,3 +319,44 @@ Feature: Documentation support
...
+        "Some_Authorization_Name": {
+            "type": "apiKey",
+            "description": "Value for the Authorization header parameter.",
+            "name": "Authorization",
+            "in": "header"
+        }
+    }
+    """
+
+    Scenario: Retrieve the yaml OpenAPI documentation
```
dunglas, line 358:
```suggestion
    Scenario: Retrieve the YAML OpenAPI documentation
```

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -26,6 +26,8 @@
 final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface
 {
     public const FORMAT = 'json';
+    public const JSONOPENAPI_FORMAT = 'jsonopenapi';
+    public const YAMLOPENAPI = 'yamlopenapi';
```
dunglas, line 30:
```suggestion
    public const YAML_FORMAT = 'yamlopenapi';
```

Same, and `FORMAT` suffix is missing

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -26,6 +26,8 @@
 final class OpenApiNormalizer implements NormalizerInterface, CacheableSupportsMethodInterface
 {
     public const FORMAT = 'json';
+    public const JSONOPENAPI_FORMAT = 'jsonopenapi';
```
dunglas, line 29:
```suggestion
    public const JSON_FORMAT = 'jsonopenapi';
```

(`OPENAPI` is redundant)

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -72,7 +74,7 @@ private function recursiveClean(array $data): array
      */
     public function supportsNormalization(mixed $data, string $format = null, array $context = []): bool
     {
-        return self::FORMAT === $format && $data instanceof OpenApi;
+        return \in_array($format, [self::FORMAT, self::JSONOPENAPI_FORMAT, self::YAMLOPENAPI], true) && $data instanceof OpenApi;
```
dunglas, line 77: As this is on the hot path, I would use the `||` operator instead of `in_array()` (micro-optimization).

src/Serializer/YamlEncoder.php
```diff
@@ -0,0 +1,66 @@
...
+namespace ApiPlatform\Serializer;
+
+use Symfony\Component\Serializer\Encoder\DecoderInterface;
+use Symfony\Component\Serializer\Encoder\EncoderInterface;
+use Symfony\Component\Serializer\Encoder\YamlEncoder as BaseYamlEncoder;
+
+/**
+ * A YAML encoder with appropriate default options to embed the generated document into HTML.
+ */
+final class YamlEncoder implements EncoderInterface, DecoderInterface
```
dunglas, line 23: Why is this class necessary? It looks like it doesn't change the behavior of the decorated class, so using the base class should be good enough.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -103,13 +103,20 @@ public function load(array $configs, ContainerBuilder $container): void
...
         $errorFormats = $this->getFormats($config['error_formats']);
+        $docsFormats = $this->getFormats($config['docs_formats']);
 
         // Backward Compatibility layer
         if (isset($formats['jsonapi']) && !isset($patchFormats['jsonapi'])) {
             $patchFormats['jsonapi'] = ['application/vnd.api+json'];
         }
 
-        $this->registerCommonConfiguration($container, $config, $loader, $formats, $patchFormats, $errorFormats);
+        if (!isset($docsFormats['jsonopenapi'], $docsFormats['json'], $docsFormats['yamlopenapi'])) {
```
dunglas, line 113: Shouldn't we use the default values in the configuration class instead of forcing default values here? What if the user want to disable these formats?

dunglas (changes requested):
This looks great. I just added a few comments but this should be mergeable very soon.

src/Serializer/YamlEncoder.php
Sarahshr: I created it to override supportsEncoding() and supportsDecoding() methods for  'yamlopenapi' format value support.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
Sarahshr: Good question, what do you think @soyuka ?

soyuka: this is only a bc layer when only `json` is set we need to set `jsonopenapi` to make it work no ?

src/Serializer/YamlEncoder.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    // @noRector \Rector\Php81\Rector\Property\ReadOnlyPropertyRector
+    public function __construct(private readonly string $format, private ?BaseYamlEncoder $yamlEncoder = null)
+    {
+        if (null !== $this->yamlEncoder) {
+            return;
+        }
+
+        $this->yamlEncoder = new BaseYamlEncoder();
+    }
```
dunglas, line 33:
```suggestion
    public function __construct(private readonly string $format, private readonly BaseYamlEncoder $yamlEncoder = new BaseYamlEncoder())
    {
    }
```

dunglas (approved):

src/Serializer/YamlEncoder.php
```diff
@@ -0,0 +1,60 @@
...
+use Symfony\Component\Serializer\Encoder\DecoderInterface;
+use Symfony\Component\Serializer\Encoder\EncoderInterface;
+use Symfony\Component\Serializer\Encoder\YamlEncoder as BaseYamlEncoder;
+
+/**
+ * A YAML encoder with appropriate default options to embed the generated document into HTML.
+ */
+final class YamlEncoder implements EncoderInterface, DecoderInterface
+{
+    public function __construct(private readonly string $format, private readonly BaseYamlEncoder $yamlEncoder = new BaseYamlEncoder())
```
dunglas, line 25:
```suggestion
    public function __construct(private readonly string $format, private readonly EncoderInterface $yamlEncoder = new BaseYamlEncoder())
```

This will fix the tests (in dev mode, encoders are decorated to be traceable).

dunglas (changes requested):

src/Serializer/YamlEncoder.php
```diff
@@ -0,0 +1,60 @@
...
+use Symfony\Component\Serializer\Encoder\DecoderInterface;
+use Symfony\Component\Serializer\Encoder\EncoderInterface;
+use Symfony\Component\Serializer\Encoder\YamlEncoder as BaseYamlEncoder;
+
+/**
+ * A YAML encoder with appropriate default options to embed the generated document into HTML.
+ */
+final class YamlEncoder implements EncoderInterface, DecoderInterface
+{
+    public function __construct(private readonly string $format, private readonly EncoderInterface|DecoderInterface $yamlEncoder = new BaseYamlEncoder())
```
dunglas, line 25:
```suggestion
    public function __construct(private readonly string $format, private readonly EncoderInterface&DecoderInterface $yamlEncoder = new BaseYamlEncoder())
```

dunglas, line 25:
```suggestion
    public function __construct(private readonly string $format = 'yamlopenapi', private readonly EncoderInterface|DecoderInterface $yamlEncoder = new BaseYamlEncoder())
```

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
dunglas, line 113: What about this instead? See my other comment below.

```suggestion
        if ((!isset($formats) || isset($formats['json'])) && !isset($docsFormats)) {
```

IMHO this a a better BC layer.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -103,13 +103,20 @@ public function load(array $configs, ContainerBuilder $container): void
...
+            $docsFormats['jsonopenapi'] = ['application/vnd.openapi+json'];
+            $docsFormats['json'] = ['application/json'];
+            $docsFormats['yamlopenapi'] = ['application/vnd.openapi+yaml'];
```
dunglas, line 116:
```suggestion
```

This isn't necessary for BC IIRC.

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -163,6 +163,12 @@ public function getConfigTreeBuilder(): TreeBuilder
         $this->addFormatSection($rootNode, 'patch_formats', [
             'json' => ['mime_types' => ['application/merge-patch+json']],
         ]);
+        $this->addFormatSection($rootNode, 'docs_formats', [
```
dunglas, line 166: We also need to do something about the `formats` section. We should deprecate not setting explicitly `formats` in 3.x and remove `json` from the default values in 4.0.

If `formats` is not set, then we should enable `json` for docs endpoint for BC.

I would remove YAML from the defaults. IMHO it should only be enabled explicitly by the user because it's slower and it increases the attack surface.

The recipe must also be updated to explicitly set these two keys.

dunglas (changes requested):
Also, this should target the main branch as it's a new feature.

soyuka:
Thanks @Sarahshr

Issue #5636, opened 2023-06-19
Replace doctrine/inflector with symfony/string?


micheh:
**Description**  
Currently the package `doctrine/inflector` is used to build the route of a resource. Doctrine updated this package in a patch release to generate a different plural for some words. This means that API Platform built different routes for some entities. The change was later reversed in another patch release, but Doctrine will probably adjust some words again in the next major release. They generally advise to use this package only for one-time text or code generation, where you do not need a stable pluralization: https://github.com/doctrine/inflector/issues/221#issuecomment-1595341265

Maybe it would make sense to replace `doctrine/inflector` with the `symfony/string` component, as the generated routes should be more stable and use more advanced pluralization? There was once a pull request to do this, but it was abandoned: https://github.com/api-platform/core/pull/3580

**Example**  
API Resource name: Permission
Route using `doctrine/inflector:2.0.6`: /api/permissions
Route using `doctrine/inflector:2.0.7`: /api/permission
Route using `doctrine/inflector:2.0.8`: /api/permissions

soyuka:
:+1:  to continue that work!

derrabus:
> Maybe it would make sense to replace `doctrine/inflector` with the `symfony/string` component, as the generated routes should be more stable

Symfony's policy on fixing pluralization rules is not that different from Doctrine's. That being said, switching to Symfony String is probably a good move for API Platform.

PR #5637, opened 2023-06-21
feat: replace doctrine/inflector by symfony/string

Romaixn:
| Q             | A
| ------------- | ---
| Branch       | main
| Tickets       | #5636 
| License       | MIT

Replace `doctrine/inflector` dependance to pluralize and tableize to `symfony/string`

Introduce new config flag `keep_legacy_inflector` to keep the old inflector or to switch to symfony/string by passing the config to `false`

### DIFF
--- a/src/Metadata/composer.json
+++ b/src/Metadata/composer.json
@@ -31,7 +31,8 @@
     "doctrine/inflector": "^2.0",
     "psr/cache": "^3.0",
     "psr/log": "^1.0 || ^2.0 || ^3.0",
-    "symfony/property-info": "^6.1"
+    "symfony/property-info": "^6.1",
+    "symfony/string": "^6.1"
   },
   "require-dev": {
     "phpstan/phpdoc-parser": "^1.16",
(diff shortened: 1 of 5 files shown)


src/Metadata/composer.json
```diff
@@ -31,7 +31,8 @@
     "doctrine/inflector": "^2.0",
     "psr/cache": "^3.0",
     "psr/log": "^1.0 || ^2.0 || ^3.0",
-    "symfony/property-info": "^6.1"
+    "symfony/property-info": "^6.1",
+    "symfony/string": "^6.1"
```
soyuka, line 35: suggest?

dunglas:
Unfortunately I think that we need a configuration option to explicitly choose the Inflector library to use. Existing projects rely on Doctrine Inflector but likely have Symfony String installed (at least as a transient dependency).
With the current implementation, these projects will switch to Symfony Inflector when upgrading, with no way to switch back to Doctrine, and this will likely break apps because some words are pluralized differently.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -790,10 +793,15 @@ private function registerArgumentResolverConfiguration(XmlFileLoader $loader): v
         $loader->load('argument_resolver.xml');
     }
 
-    private function triggerDeprecations(): void
+    private function registerInflectorConfiguration(array $config): void
     {
-        if (!class_exists(UnicodeString::class)) {
-            trigger_deprecation('api-platform/core', '3.2', 'Using doctrine/inflector is deprecated since API Platform 3.2 and will be removed in API Platform 4. Use symfony/string instead. Run "composer require symfony/string".');
+        if ($config['keep_legacy_inflector']) {
+            Inflector::keepLegacyInflector(true);
```
soyuka, line 799:
```suggestion
```

No need to fix this one its deprecated we'll remove it in 4.x and it won't really be used once we have the subtree split

soyuka:
thanks!

PR #4681, opened 2022-03-14
fix: hydra:view example values

divine:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #4354 #4394 #4667
| License       | MIT
| Doc PR        | api-platform/docs#... 

Hi,

I went ahead and checked what's the reason for this issue.

I can confirm that this is a Redoc issue that when property contains format `"iri-reference"`
<details>
  <summary>hydra:view propery</summary>
  
  ```javascript
  {
	"hydra:view": {
		"type": "object",
		"properties": {
			"@id": {
				"type": "string",
				"format": "iri-reference"
			},
			"@type": {
				"type": "string"
			},
			"hydra:first": {
				"type": "string",
				"format": "iri-reference"
			},
			"hydra:last": {
				"type": "string",
				"format": "iri-reference"
			},
			"hydra:previous": {
				"type": "string",
				"format": "iri-reference"
			},
			"hydra:next": {
				"type": "string",
				"format": "iri-reference"
			}
		}
	}
}
  ```
  
</details>

```json
{"type": "string", "format": "iri-reference"}
```

it's basically [returns sample values](https://github.com/Redocly/openapi-sampler/blob/3c10a2311f30b440359c8a3273a499e13e3bcbf1/src/samplers/string.js#L83-L85):

```javascript
function iriReferenceSample() {
  return '../словник';
}
```

If you remove `"iri-reference"` from the property, everything shows correctly.

The solution would be:

1. Remove `iri-reference` from `hydra:view` property (which isn't nice).
2. Add `example` values to `hydra:view`

This PR contains the 2. changes but it's kinda a BC probably.

Let me know if everything looks good.

Thanks!

PR #5638, opened 2023-06-21
fix(JsonSchema): fix reDoc json sample values

ERuban:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | -
| License       | MIT

Actually it fixes the same bug as in that [PR](https://github.com/api-platform/core/pull/4681) but for JSON.

image

### DIFF
--- a/src/JsonSchema/TypeFactory.php
+++ b/src/JsonSchema/TypeFactory.php
@@ -139,7 +139,7 @@ final class TypeFactory implements TypeFactoryInterface
         if (true !== $readableLink && $this->isResourceClass($className)) {
             return [
                 'type' => 'string',
-                'format' => 'iri-reference',
+                'format' => 'iri-template',
             ];
         }
 
--- a/features/openapi/docs.feature
+++ b/features/openapi/docs.feature
@@ -100,7 +100,7 @@ Feature: Documentation support
     """
     {
       "type": "string",
-      "format": "iri-reference"
+      "format": "iri-template"
     }
     """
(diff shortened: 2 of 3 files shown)


ERuban:
@dunglas ?

ERuban:
@soyuka 👋

ERuban:
@soyuka hey, what about this PR? 
Have rebased it to the fresh `3.1`

soyuka:
tests are failing

ERuban:
> tests are failing

@soyuka done. The last one seems not related to my changes.

soyuka:
thanks!

src/JsonSchema/TypeFactory.php
```diff
@@ -139,7 +139,7 @@ private function getClassType(?string $className, bool $nullable, string $format
         if (true !== $readableLink && $this->isResourceClass($className)) {
             return [
                 'type' => 'string',
-                'format' => 'iri-reference',
+                'format' => 'iri-template',
```
soyuka, line 142: I've a problem with that it's not in the specification: https://json-schema.org/understanding-json-schema/reference/string#resource-identifiers isn't redoc wrong?

soyuka:
I reverted that.

dunglas:
@soyuka you're right. We don't generate templates here but references. The patch was wrong (maybe there is a bug in ReDoc, but then ReDoc must be patched, not our code).

soyuka:
I think what we need is to add examples if we really want this feature on docs.

ERuban:
ok, got it. let me try to add examples then

Issue #5642, opened 2023-06-28
Deeply-nested filters don't work with Elasticsearch


colinodell:
**API Platform version(s) affected**: 2.6.8 (and likely earlier) through 3.x

## Description

`MatchFilter` and `TermFilter` do not properly support [multi-level nested fields in Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-nested-query.html#multi-level-nested-query-ex) if the top level is a collection.

### What Works

It seems to work fine if the top level is an **object**.  For example, given this sample document from the docs linked above:

```json
{
  "driver" : {
        "last_name" : "McQueen",
        "vehicle" : [
            {
                "make" : "Powell Motors",
                "model" : "Canyonero"
            },
            {
                "make" : "Miller-Meteor",
                "model" : "Ecto-1"
            }
        ]
    }
}
```

Asking API Platform to filter on `driver.vehicle.model` would generate a valid, working query like this:

```json
{
    "nested": {
        "path": "driver.vehicle",
        "query": {
            "term": {"driver.vehicle.model": "Canyonero"}
        }
    }
}
```

Interestingly, the Elasticsearch docs suggest the query should instead be doubly-nested like this:

```json5
{
    "nested": {
        "path": "driver", // <-- instead of driver.vehicle
        "query": {
            "nested": { // <-- additional nested layer
                "path": "driver.vehicle",
                "query": {
                    "term": {"driver.vehicle.model": "Canyonero"}
                }
            }
        }
    }
}
```

But I'm assuming ES will happily accept either version.  Regardless, everything described above seems to work just fine.

### What's Broken

However, if we modify the structure to have an **array** of `drivers` as the top-level property like this:

```json
{
    "race": "Piston Cup",
    "drivers": [
        {
            "last_name" : "McQueen",
            "vehicle" : [
                {
                    "make" : "Powell Motors",
                    "model" : "Canyonero"
                },
                {
                    "make" : "Miller-Meteor",
                    "model" : "Ecto-1"
                }
            ]
        }
    ]
}
```

Attempting to filter on `drivers.vehicle.model` generates this query:

```json5
{
    "nested": {
        "path": "drivers",
        "query": {
            "term": {"drivers.vehicle.model": "Canyonero"}
        }
    }
}
```

Which fails to match any documents because the `path` is missing the `.vehicle` part.  Manually adding the `.vehicle` part causes the expected documents to appear in the results:

```json5
{
    "nested": {
[...]

PR #5643, opened 2023-06-29
fix(elasticsearch): fix filtering on nested fields

colinodell:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | api-platform/api-platform/issues/1375 and #5642
| License       | MIT
| Doc PR        | n/a

This PR implements the following changes to support filtering on paths like `foo.bar.baz`:

1. If `foo` is an object, `isNestedField()` now returns `true` instead of `false`
2. If `foo` is an object, `getNestedFieldPath()` now returns `"foo.bar"` instead of `null`
3. If `foo` is a collection, `getNestedFieldPath()` now returns `"foo.bar"` instead of just `"foo"` (matching the behavior if `foo` is an object)

As a result, filtering on deeply-nested paths (with 3 or more levels) now produces the following query:

```diff
 {
     "nested": {
-        "path": "foo",
+        "path": "foo.bar",
         "query": {
             "term": {"foo.bar.baz": "qux"}
            }
         }
     }
 }
```

### DIFF
--- a/src/Elasticsearch/Util/FieldDatatypeTrait.php
+++ b/src/Elasticsearch/Util/FieldDatatypeTrait.php
@@ -82,7 +82,9 @@ trait FieldDatatypeTrait
             && null !== ($className = $type->getClassName())
             && $this->resourceClassResolver->isResourceClass($className)
         ) {
-            return $currentProperty;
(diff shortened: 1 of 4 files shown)


soyuka:
nice please target `main` as it's a new feature right?

PR #5646, opened 2023-07-03
Remove the `@id` and `@type` JSON-LD by passing a flag to the resource.

Romaixn:
| Q             | A
| ------------- | ---
| Branch       | main
| License       | MIT

Replace `@id` and `@type` with `id` and `type` respectively, and include the definition in the context by passing a flag in the resource.

Example with the flag set to true in the Book entity:

```
{
  "@context": "/contexts/Book",
  "id": "/books/01891a92-06dc-79ca-9f30-778fbc841f1b",
  "type": "https://schema.org/Book",
  "isbn": "9780840262592",
  "title": "Vero dolorem eos voluptas sed at et.",
  "description": "Temporibus recusandae labore voluptas at quam. Consectetur quia sit omnis aspernatur deleniti ipsum et. Et quo et sequi. Quia dignissimos et enim tempore enim fugit.",
  "author": "Donny Deckow",
  "publicationDate": "1994-08-24T00:00:00+00:00",
  "reviews": [
    {
      "@id": "/reviews/01891a92-132d-7a24-8d81-ee085f11fbac",
      "@type": "https://schema.org/Review",
      "id": "01891a92-132d-7a24-8d81-ee085f11fbac",
      "body": "Dolor sunt animi provident quam aut quia. Impedit reiciendis possimus distinctio aliquam non similique delectus. Fuga fugit aut hic. Voluptatem quia qui unde rerum."
    }
  ]
}
```

```
{"@context":{"@vocab":"https:\/\/localhost\/docs.jsonld#","hydra":"http:\/\/www.w3.org\/ns\/hydra\/core#","id":{"@type":"@id"},"type":{"@type":"@type"},"isbn":"Book\/isbn","title":"Book\/title","description":"Book\/description","author":"Book\/author","publicationDate":"Book\/publicationDate","reviews":"Book\/reviews","cover":"Book\/cover","archivedAt":"Book\/archivedAt"}}
```


src/JsonLd/ContextBuilder.php
```diff
@@ -145,6 +145,15 @@ private function getResourceContextWithShortname(string $resourceClass, int $ref
         $context = $this->getBaseContext($referenceType);
         $propertyContext = $operation ? ['normalization_groups' => $operation->getNormalizationContext()['groups'] ?? null, 'denormalization_groups' => $operation->getDenormalizationContext()['groups'] ?? null] : ['normalization_groups' => [], 'denormalization_groups' => []];
 
+        if($operation->getRemoveJsonLdIdType()) {
```
dunglas, line 148: T
```suggestion
        if($operation->getAliasIdType()) {
```

Nitpicking but technically they are. Or removed but just mapped / aliased using the context.

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -99,8 +99,14 @@ public function normalize(mixed $object, string $format = null, array $context =
         }
 
         if (true === ($context['force_iri_generation'] ?? true) && $iri = $this->iriConverter->getIriFromResource($object, UrlGeneratorInterface::ABS_PATH, $context['operation'] ?? null, $context)) {
+            $operation = $context['operation'] ?? $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation();
             $context['iri'] = $iri;
-            $metadata['@id'] = $iri;
+
+            if($operation->getRemoveJsonLdIdType()) {
```
dunglas, line 105: You could use a ternary instead of a if.

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -117,7 +123,12 @@ public function normalize(mixed $object, string $format = null, array $context =
             if (null === $types) {
                 $types = [$operation->getShortName()];
             }
-            $metadata['@type'] = 1 === \count($types) ? $types[0] : $types;
+
+            if($operation->getRemoveJsonLdIdType()) {
```
dunglas, line 127: You could use a ternary here, and store the previous call to the getter in a variable to prevent calling the method two times on the hot path (method calls are costly in PHP).

dunglas (changes requested):

tests/Fixtures/TestBundle/Entity/DummyAliasIdType.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\Entity;
```
soyuka, line 14: do you really need a doctrine entity ? can you create a simple dto inside ApiResource instead?

src/JsonLd/ContextBuilder.php
```diff
@@ -112,6 +112,7 @@ public function getAnonymousResourceContext(object $object, array $context = [],
         $outputClass = $this->getObjectClass($object);
         $operation = $context['operation'] ?? new Get(shortName: (new \ReflectionClass($outputClass))->getShortName());
         $shortName = $operation->getShortName();
+        $isAliasIdType = $operation->getAliasIdType();
```
soyuka, line 115: I'd like to find a better name for this, like `jsonldAsJson` or `jsonAliases`, or we need to find something ^^

soyuka:
please rebase the branch, it should be green mostly

src/JsonLd/ContextBuilder.php
dunglas: I don't like the current name too, some suggestions: `noSpecialProps`, `cleanOutput`

Romaixn: Indeed, I'm not a fan of the name either. Perhaps `aliasJsonLdKeywords` or `keywordsAsJson` would be more appropriate? As noted in the [specification](https://www.w3.org/TR/json-ld11/#keywords), the terms `@id` / `@type` are referred to as "keywords".

dunglas: I don't like "keywords" because they are other keywords than `@id` and `@type`. As this options is closely related to the existing `jsonLdContext` property (https://api-platform.com/docs/core/extending-jsonld-context/), what about `jsonLdAliases: true`?

Romaixn: `JsonLdAliases` works well for me! It's much more descriptive than `AliasIdType`.

PR #5649, opened 2023-07-04
feat(Serializer): update MissingConstructorArgumentsException message

ERuban:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

Have found that if validation attributes are specified on DTOs or if the `Valid()` constraint is used on not api-resource property (just an entity in example) then Symfony's [AbstractNormalizer](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php) is used and it has a bit different `MissingConstructorArgumentsException` [exception message that also contains the list of missed arguments](https://github.com/symfony/symfony/blob/9b2ac59e8a373e983157f9a57de8ae951cc1a3ba/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php#L414).
But in ApiPlatform this exception has another [message and it contains only the first one of missed arguments in it](https://github.com/api-platform/core/blob/4ef0ef856ced658ac942fd6a2c6f7c5c563078d1/src/Serializer/AbstractItemNormalizer.php#L312).
This way someone, who relies on this message, in example, to convert it to something like validation error in the response, will get different messages.
So I want to suggest to make this message and its parameters consistent with Symfony's one.

The example of the new message:
```
Cannot create an instance of "App\Entity\Book" from serialized data because its constructor requires the following parameters to be present : "$author", "$name".
```

Not sure that it is the BC break, but correct me if it is.

PS: Sorry, have no idea how to make `commitlint` happy, I've tried (seems it always takes the commit message from the first commit) :)

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -338,7 +338,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                 }
             }
 
-            if (\count($missingConstructorArguments) > 0) {
+            if ($missingConstructorArguments) {
                 throw new MissingConstructorArgumentsException(sprintf('Cannot create an instance of "%s" from serialized data because its constructor requires the following parameters to be present : "$%s".', $class, implode('", "$', $missingConstructorArguments)), 0, null, $missingConstructorArguments, $class);
             }
 
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -330,7 +330,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
                     $params[] = $constructorParameter->getDefaultValue();
                 } else {
                     if (!isset($context['not_normalizable_value_exceptions'])) {
-                        $missingConstructorParameters[] = $constructorParameter->getName();
+                        $missingConstructorParameters[] = $constructorParameter->name;
                     }
 
                     $exception = NotNormalizableValueException::createForUnexpectedDataType(sprintf('Failed to create object because the class misses the "%s" property.', $constructorParameter->name), $data, ['unknown'], $context['deserialization_path'] ?? null, true);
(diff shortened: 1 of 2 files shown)


soyuka:
this looks good to me, thanks!

soyuka:
can you rebase? also edit your first message if you want to see the commitlint green :)

ERuban:
@soyuka done

ERuban:
@soyuka 👋

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -309,14 +310,18 @@ protected function instantiateObject(array &$data, string $class, array &$contex
...
-                        throw new MissingConstructorArgumentsException(sprintf('Cannot create an instance of "%s" from serialized data because its constructor requires parameter "%s" to be present.', $class, $constructorParameter->name), 0, null, [$constructorParameter->name]);
+                        $missingConstructorArguments[] = $constructorParameter->name;
                     }
 
                     $exception = NotNormalizableValueException::createForUnexpectedDataType(sprintf('Failed to create object because the class misses the "%s" property.', $constructorParameter->name), $data, ['unknown'], $context['deserialization_path'] ?? null, true);
                     $context['not_normalizable_value_exceptions'][] = $exception;
                 }
             }
 
+            if (\count($missingConstructorArguments) > 0) {
```
dunglas, line 321:
```suggestion
            if ($missingConstructorArguments) {
```

dunglas (changes requested):
Nice one.

Exception messages aren't part of our BC promise, so it's ok.

dunglas:
You have to update the tested message in the failing Behat test.

ERuban:
> You have to update the tested message in the failing Behat test.

oh, sorry, fixed

ERuban:
@dunglas @soyuka 👋

soyuka:
Tests should be green.

ERuban:
@soyuka @dunglas hey, have rebased and seems that ci became better. not sure that red checks are related to my changes. Could you check it please?

ERuban:
@soyuka  have rebased again... CI looks better

soyuka:
thanks, I need to check what branch to target but I think that this is from at least 3.1 I'll handle that!

Issue #5086, opened 2022-10-24
Merge functionality of PurgeHttpCacheListener and PublishMercureUpdatesListener and enchance


silverbackdan:
**Description**  
The responsibility of these two listeners originates from the same initial trigger point. A database entity has been updated so we should clear cache and notify of changes.

The `PurgeHttpCacheListener` has some extended functionality to gather relations as well which the `PublishMercureUpdatesListener` should probably do as well.

My thoughts would be to create a single listener with the extended functionality that the `PurgeHttpCacheListener` has. This can collect all the relevant iris and objects to publish on the `onFlush` and `PostFlush` event.

A simple change to allow a method to be public, like `addTagsFor` method,  would allow any user to also populate any other resource changes outside of the scope of the doctrine database changes and I think a really neat little addition for lots of flexibility.

the `onFlush` could have the detection of which services are enabled and available and then purge caches and publish updates accordingly.

In my case... it may be nice to have 2 iri converters in place in the constructor as well, so one can be replaced for mercure if required.

Thoughts on this update much appreciated before the work. Probably includes BC breaks though so... may need to be approached with caution.

silverbackdan:
Discussed, and too much abstraction. Mercure reflects update and not the API at any given point in time, but cache needs to be in sync with API

PR #5650, opened 2023-07-05
improved cache tagging & purging for relations

usu:
| Q                                                      | A                                                                                                                                                                                                                                                                                    |
| -------------                                          | ---                                                                                                                                                                                                                                                                                  |
| Branch?                                                | 3.1                                                                                                                                                                                                                                                                                  |
| Tickets                                                |
| License                                                | MIT                                                                                                                                                                                                                                                                                  |
| Doc PR                                                 | TBD                                                                                                                                                                                                                                                                                  |

This PR improves the cache tagging logic. The aim is to keep the full exactness (no stale data in the HTTP cache) but drastically reducing the amount of wrongly purged data. I open this PR as draft to receive some early feedback before progressing further and writing tests.

# Problems with current implementation of cache tagging

- Each response contains cache tags for all IRIs referenced. No difference is made between linked entities and embedded entities. However, in many write cases, no purge is needed for linked entities (for example if the change only affects properties within the linked entity).
- For each write operation, the affected item is purged as well as all its related objects. This double-interaction (all related objects included in cache tags + all related objects purged on write) is especially bad for unidirectional relations. A mere property change on `/items/1` would purge all `/items/x`, if they have a unidirectional link to a common entity. (Unidirectional here is not to be understood from a database/doctrine perspective, but from an API/serialization perspective; =property is only serialized from one side of the relationship).

# Changes proposed in this PR

Response change tags:

- pure IRIs are only included in the cache tags for entities that are fully embedded in the response (=requested item + embedded entities / subentities; or items of a collection in case of a collection request)
- For each relation property , 1 cache tag is added in the format "{iri}#{property}" (happy to use another delimiter if # is not a good one). For xToMany relations, this reduces the number of cache tags from N to 1 (helping towards https://github.com/api-platform/core/issues/3168).

For write operations, the following tags are purged:

**Post**

- the collection which was posted on
- all related objects, from the related objects side (=does't purge the related object, if the relation is unidirectional)

**Drop**

- the IRI of the item itself
- the collection the item belonged to
- all related objects, from the related objects side (=does't purge the related object, if the relation is unidirectional)

**Update**

- the IRI of the item itself
- related objects of changed relations only, from the related objects side (=does't purge the related object, if the relation is unidirectional)

# Example

```json
{
  "_links": {
    "self": {
      "href": "/items/1"
    },
    "linkedParent": {
      "href": "/linked_parents/2"
    },
    "linkedChildren": [
      {
        "href": "/linked_children/3"
      }
    ],
    "embeddedParent": {
      "href": "/embedded_parents/4"
    },
    "embeddedChildren": [
      {
        "href": "/embedded_children/5"
      }
    ]
  },
  "_embedded": {
    "embeddedParent": {
      "_links": {
        "self": {
          "href": "/embedded_parents/4"
        }
      },
      "dummyProperty": "",
      "id": 4
    },
    "embeddedChildren": [
      {
        "_links": {
          "self": {
            "href": "/embedded_children/5"
          },
          "embeddedGrandchildren": []
        },
        "_embedded": {
          "embeddedGrandchildren": []
        },
        "dummyProperty": "",
        "id": 5
      }
    ]
  },
  "id": 1
}
```

| Request      | Response cache tags                                                                                                                                                                                               |
| ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| GET /items/1 | /items/1<br>/items/1#linkedParent<br>/items/1#linkedChildren<br>/items/1#embeddedParent<br>/items/1#embeddedChildren<br>/embedded_parents/4<br>/embedded_children/5<br> /embedded_children/5#embeddedGrandchildren |

| Request                                                            | Purged tags <br>(in bold tag which matches the cache tags from above)         | Comment                                                                                                                           |
| ------------------------------------------------------------------ | ----------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| PATCH /linked_parents/2<br>{ stringProperty: "dummy" }             | /linked_parents/2                                                             | Only reponses where /linked_parents/2 is directly embedded are purged.<br><br>The cached reponse of "GET /items/1" is still valid |
| PATCH /linked_children/3<br>{ parent: "/items/10" }                | /linked_children/3<br>**/items/1#linkedChildren**<br>/items/10#linkedChildren | "GET /items/1" is properly purged                                                                                                 |
| POST /linked_children<br>{ parent: "/items/1" }                    | /linked_children<br>**/items/1#linkedChildren**                               | "GET /items/1" is properly purged                                                                                                 |
| PATCH /embedded_parents/4<br>{ dummyProperty: "dummy" }            | **/embedded_parents/4**                                                       | "GET /items/1" is properly purged                                                                                                 |
| POST /items<br>{ embeddedParent: "/embeddedParents/4"}             | /items<br>/embeddedParents/4#children                                         | "GET /items/1" is still valid, because the children property was not serialized on the embeddedParent entity                      |
| POST /embedded_grandchildren<br>{ parent: "/embedded_children/5" } | /embedded_grandchildren<br>**/embedded_children/5#embeddedGrandchildren**     | "GET /items/1" is properly purged                                                                                                 |
|                                                                    |                                                                               |                                                                                                                                   |

Happy to add additional examples, if the logic is not clear.

## manual cache dependencies

For manual getters inside the entity, the logic doesn't know a priori, which dependencies the getter relies on. This can be manually controlled with `extraProperties['cacheDependencies']`.

```php
    #[ApiProperty(readableLink: true, extraProperties: ['cacheDependencies' => ['childrenA', 'childrenB']])]
    #[SerializedName('children')]
    public function getCombinedChildren(): array {
        return array_merge(
            $this->childrenA->getValues(),
            $this->childrenB->getValues()
        );
    }
```

Is this PR will be accepted, we can also transform this into an official property.

# Open points & requested feedback

- Does this approach makes sense? Would like to have a feedback before moving on with writing tests.
- I'm personally working mostly with haljson format. Still need to do some review/checks, if this also works for the other formats.
- Would it makes sense to outsource the actual calculation of the cache tags into seperate event listeners instead of hard-baking it into `AbstractItemNormalizer`? This would it make easier for users to extend or overwrite the cache tag generation with their own logic.


The last point is also related towards the ability to easily extend and/or replace the cache tag system. As of now:
- Tag collection: Is not easily replaceable, as the logic is hardbaked into the Normalizers (the only way to change the logic is to replace the complete Normalizer). It's somewhat extendable (as documented), by adding additional tags to the request.
- Purge operations: Is replaceable by replacing the PurgeHttpCacheListener with an own implementation. Not easily extendable, as all logic to generate and push the purge tags is isolated within `PurgeHttpCacheListener`.

Is there a specific reason, why FOSHttpCacheBundle is not utilized? It has both `$responseTagger->addTags()` and `$cacheManager->invalidateTags()` which makes it really easy, to hook into the cache system from anywhere in the logic (e.g. to invalidate additional tags in the state processor for specific operations).

(A public function `addTagsFor` was for example suggested in https://github.com/api-platform/core/issues/5086, but never implemented.)

### DIFF
--- a/src/Doctrine/EventListener/PurgeHttpCacheListener.php
+++ b/src/Doctrine/EventListener/PurgeHttpCacheListener.php
@@ -41,6 +41,8 @@ final class PurgeHttpCacheListener
     private readonly PropertyAccessorInterface $propertyAccessor;
     private array $tags = [];
 
+    public const IRI_RELATION_DELIMITER = '#';
+
     public function __construct(private readonly PurgerInterface $purger, private readonly IriConverterInterface $iriConverter, private readonly ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null)
     {
         $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
@@ -52,7 +54,7 @@ final class PurgeHttpCacheListener
     public function preUpdate(PreUpdateEventArgs $eventArgs): void
     {
         $object = $eventArgs->getObject();
-        $this->gatherResourceAndItemTags($object, true);
+        $this->addTagForItem($object);
 
         $changeSet = $eventArgs->getEntityChangeSet();
         $objectManager = method_exists($eventArgs, 'getObjectManager') ? $eventArgs->getObjectManager() : $eventArgs->getEntityManager();
@@ -62,9 +64,14 @@ final class PurgeHttpCacheListener
             if (!isset($associationMappings[$key])) {
                 continue;
             }
+            $mappings = $associationMappings[$key];
+            $relatedProperty = $mappings['isOwningSide'] ? $mappings['inversedBy'] : $mappings['mappedBy'];
+            if (!$relatedProperty) {
+                continue;
+            }
 
-            $this->addTagsFor($value[0]);
-            $this->addTagsFor($value[1]);
+            $this->addTagsFor($value[0], $relatedProperty);
+            $this->addTagsFor($value[1], $relatedProperty);
         }
     }
 
@@ -77,17 +84,13 @@ final class PurgeHttpCacheListener
         $uow = $em->getUnitOfWork();
 
         foreach ($uow->getScheduledEntityInsertions() as $entity) {
-            $this->gatherResourceAndItemTags($entity, false);
-            $this->gatherRelationTags($em, $entity);
-        }
-
-        foreach ($uow->getScheduledEntityUpdates() as $entity) {
-            $this->gatherResourceAndItemTags($entity, true);
+            $this->gatherResourceTags($entity);
             $this->gatherRelationTags($em, $entity);
         }
 
         foreach ($uow->getScheduledEntityDeletions() as $entity) {
-            $this->gatherResourceAndItemTags($entity, true);
+            $this->addTagForItem($entity);
+            $this->gatherResourceTags($entity);
             $this->gatherRelationTags($em, $entity);
         }
     }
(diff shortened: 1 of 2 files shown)


usu:
@soyuka Do you mind having a look at this to judge whether this is worth pursuing?

PR #5653, opened 2023-07-07
ci(commitlint): fix resolving the commit message

ERuban:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| License       | MIT

Yeah, 
this PR fixes the issue with `commitlint` in CI I've stacked in my #5638 and #5649 PRs, before - it always takes the message of the first commit in the PR and new commits or force-push with the new commit message didn't help to fix this check.
So now, as you can see with commits in this PR it is fixed. :)


soyuka:
we want to take only the first one as we squash and merge, you should amend to fix it.

Issue #5654, opened 2023-07-07
Documentation custom filter Undefined array key "type" when type is not set


symfonyaml:
**API Platform version(s) affected**: 3.1.12

**Description**  
This bug occurs when creating a custom filter, and I forot to set the `type` key for the description field.
```
Warning: Undefined array key "type"
in vendor/api-platform/core/src/OpenApi/Factory/OpenApiFactory.php line 568
```

**How to reproduce**  
 - Create a custom filter by implementing the `FilterInterface`.
 - Define a property description and omit setting the `type` key.
 ```php
<?php

namespace App\Filter;

use ApiPlatform\Api\FilterInterface;

class MyCustomFilter implements FilterInterface
{
    public function getDescription(string $resourceClass): array
    {
        return [
            'test' => [],
        ];
    }
}
 ```

**Possible Solution**  
We may need to check if the `type` key is set, and if it's not, we do not create a description for this property.

symfonyaml:
PR linked to this ticket : https://github.com/api-platform/core/pull/5655

PR #5655, opened 2023-07-07
fix(openapi): fix custom filter when key "type" is not set

symfonyaml:
| Q             | A
| ------------- | ---
| Branch       | 3.1
| Tickets       | [Issue #5654](https://github.com/api-platform/core/issues/5654)
| License       | MIT

Please read [this issue](https://github.com/api-platform/core/issues/5654) to understand and reproduce.

This is the exception I came across while creating a new custom filter :

bug-5654

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -565,7 +565,7 @@ final class OpenApiFactory implements OpenApiFactoryInterface
             }
 
             foreach ($filter->getDescription($entityClass) as $name => $data) {
-                $schema = $data['schema'] ?? (\in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string']);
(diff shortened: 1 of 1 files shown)


src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -565,7 +565,7 @@ private function getFiltersParameters(CollectionOperationInterface|HttpOperation
             }
 
             foreach ($filter->getDescription($entityClass) as $name => $data) {
-                $schema = $data['schema'] ?? (\in_array($data['type'], Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string']);
+                $schema = $data['schema'] ?? (\in_array($data['type'] ?? false, Type::$builtinTypes, true) ? $this->jsonSchemaTypeFactory->getType(new Type($data['type'], false, null, $data['is_collection'] ?? false)) : ['type' => 'string']);
```
soyuka, line 568: specify a `type` in your filter description instead?

symfonyaml: Sorry for the confusion, but I was just trying to add some DX, instead of having a PHP error when you forget the `type` key.
In that case I fixed, it will return `['type' => 'string']` if there's no type.

PR #5657, opened 2023-07-10
refactor: use provider/processor instead of event listeners

soyuka:
tldr: This patch introduces providers and processors as a replacement for Symfony listeners. 

## API Platform standalone

The idea behind that is to make API Platform work without using Symfony's [HttpKernel](https://symfony.com/doc/current/components/http_kernel.html). This means that to use API Platform, all you need to do is:

```php
$uriVariables = ['id' => 1];
$operation = new Get(uriVariables: ['id'], resourceClass: Book::class);
$body = $this->provider->provide($operation, $uriVariables, $context);
return $this->processor->process($body, $operation, $uriVariables, $context);
```

This allows:
  - to share more code between GraphQl and HTTP (very similar to #2978, we had no provider/processor at that time though) 
  - to make the code more portable (goes in pair with our ongoing subtree split)
  - to be able to use API Platform without the Symfony framework, it was in fact possible but a bit harder to do
  - improve your Developer Experience **a lot**

## Impact on DX and BC-layer

Let me explain this last part. Today API Platform relies on the [HttpKernel](https://symfony.com/doc/current/components/http_kernel.html#creating-an-event-listener) and hooks various event listeners do to work. While doing this, it alters the way Symfony itself works by adding (or mutating) [Request::$attributes](https://github.com/symfony/symfony/blob/9a0f178c5324c6ee7a3d99fe4fb7ef31b454925c/src/Symfony/Component/HttpFoundation/Request.php#L87C2-L87C2). The most notable change compared to a standard Symfony is that, when you're router goes through API Platform your controller can return `mixed` instead of a `Response`:

https://github.com/api-platform/core/blob/60075501a6d15d03d4db26f00715a52cac068977/src/Action/PlaceholderAction.php#L21-L32

Indeed, API Platform will intercept the `mixed` data and create the `Response` for you. This IMO is not so good especially when using API Platform programmatically.

Because we rely on event listeners, we need to make choices for the framework and for example we choose to call Symfony's security before validating the user input. This lead (and still leads) to issues (#5756) as you may want stuff to be made differently. Even within Symfony, http kernel listeners priorities are NOT part of the backward compatibility layer, and therefore everyone will tell you to avoid using listeners if possible, especially avoid relying on vendor listeners.

State providers and processor will help a lot with that. For example in API platform 3.1 we'd have this for our security: 

https://github.com/api-platform/core/blob/60075501a6d15d03d4db26f00715a52cac068977/src/Symfony/Bundle/Resources/config/security.xml#L23-L28

(note that the logic is currently duplicated to work with graphql)

But in API Platform 3.2 we declare 3 different services that just decorate the correct service where it needs to hook security (`read`, `deserialize`, `validate`):

https://github.com/api-platform/core/blob/5138850601fdde88c73030b7cbe5319e8e893673/src/Symfony/Bundle/Resources/config/security.xml#L19-L34

There is a better compatibility layer, you can also do this yourself if you have custom needs without needing an event listener or to merge a new feature. This is, in my opinion, way better to extend API Platform and be less "constraint" then we previously where on every aspect!

This is the result of almost 6 months of code and years of discussions with the core team, many users at conferences or at work so thanks to everyone!


src/Action/EntrypointAction.php
```diff
@@ -23,12 +28,26 @@
  */
 final class EntrypointAction
 {
-    public function __construct(private readonly ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory)
+    public function __construct(
+        private readonly ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory,
+        private readonly ?ProviderInterface $provider = null,
+        private readonly ?ProcessorInterface $processor = null)
```
soyuka, line 34:
```suggestion
        private readonly ?ProcessorInterface $processor = null
    )
```

src/ApiResource/Error.php
```diff
@@ -15,22 +15,22 @@
...
     uriTemplate: '/errors/{status}',
-    provider: 'api_platform.state_provider.default_error',
     types: ['hydra:Error'],
+    openapi: false,
     operations: [
-        new Get(name: '_api_errors_hydra', outputFormats: ['jsonld' => ['application/ld+json']], normalizationContext: ['groups' => ['jsonld'], 'skip_null_values' => true]),
-        new Get(name: '_api_errors_problem', outputFormats: ['json' => ['application/problem+json']], normalizationContext: ['groups' => ['jsonproblem'], 'skip_null_values' => true]),
-        new Get(name: '_api_errors_jsonapi', outputFormats: ['jsonapi' => ['application/vnd.api+json']], normalizationContext: ['groups' => ['jsonapi'], 'skip_null_values' => true], provider: 'api_platform.json_api.state_provider.default_error'),
+        new ErrorOperation(name: '_api_errors_problem', outputFormats: ['json' => ['application/problem+json']], normalizationContext: ['groups' => ['jsonproblem'], 'skip_null_values' => true]),
+        new ErrorOperation(name: '_api_errors_hydra', outputFormats: ['jsonld' => ['application/ld+json']], normalizationContext: ['groups' => ['jsonld'], 'skip_null_values' => true]),
```
soyuka, line 32: `jsonld` should error as `problem+json`

src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            // If authorization has failed for a relation field (e.g. via ApiProperty security), the field is not present in the source: null can be returned directly to ensure the collection isn't in the response.
+            if ($operation && (null === $resourceClass || null === $rootClass || (null !== $source && !\array_key_exists($info->fieldName, $source)))) {
+                return null;
+            }
+
+            // Handles relay nodes
+            if (!$operation) {
+                $operation = new Query();
```
soyuka, line 50: one liner here

src/Symfony/Routing/ApiLoader.php
```diff
@@ -36,7 +36,7 @@ final class ApiLoader extends Loader
...
+    public function __construct(KernelInterface $kernel, private readonly ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly ContainerInterface $container, private readonly array $formats, private readonly array $resourceClassDirectories = [], private readonly bool $graphqlEnabled = false, private readonly bool $entrypointEnabled = true, private readonly bool $docsEnabled = true, private readonly bool $graphiQlEnabled = false, private readonly bool $graphQlPlaygroundEnabled = false, private readonly bool $useSymfonyEvents = true)
```
soyuka, line 39: defaults to true, can use false default system, receipe => false, explicit dans configuration puis en 4 passé à false.

Issue #5665, opened 2023-07-11
Do not apply filters is `filters` option is already set on an operation


vincentchalamon:
**API Platform version(s) affected**: 3.x

**Description**
Considering the following resource:
```php
#[ApiResource(
    types: ['https://schema.org/Review'],
    operations: [
        new GetCollection(),
        new Get(),
    ],
    routePrefix: '/admin',
    security: 'is_granted("ROLE_ADMIN")'
)]
#[ApiResource(
    uriTemplate: '/books/{bookId}/reviews{._format}',
    types: ['https://schema.org/Review'],
    operations: [
        new GetCollection(filters: [], itemUriTemplate: '/books/{bookId}/reviews/{id}{._format}'),
        new NotExposed(
            uriTemplate: '/books/{bookId}/reviews/{id}{._format}',
            uriVariables: [
                'bookId' => new Link(toProperty: 'book', fromClass: Book::class),
                'id' => new Link(fromClass: Review::class),
            ],
        ),
    ],
    uriVariables: [
        'bookId' => new Link(toProperty: 'book', fromClass: Book::class),
    ],
)]
class Review
{
    #[ApiFilter(NumericFilter::class)]
    public ?int $rating = null;
```
The filter on `rating` must be applied only on the first ApiResource (`/admin/reviews`), not the second one. To do so, I set the `filters: []` option to disable all filters. But the FiltersResourceMetadataCollectionFactory erases this configuration.

**Possible Solution**
Do not set filters on FiltersResourceMetadataCollectionFactory is `filters` is not null.

soyuka:
can't you declare the filter on your resource instead of the property? To me a filter on a property is there for every resources, just like every ApiProperty informations.

vincentchalamon:
> can't you declare the filter on your resource instead of the property? To me a filter on a property is there for every resources, just like every ApiProperty informations.

Will do. Do API Platform already supports the following syntax? (if not, do you think it'll be a good feature?):
```php
#[GetCollection(
    filters: [
        new ApiFilter(SearchFilter::class, properties: ['title', 'author']),
    ]
)]
```

jamesisaac:
Relevant discussion: https://github.com/api-platform/core/discussions/5519

[...]

PR #5666, opened 2023-07-11
Do not apply filters if it is already set on the operation

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5665
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Metadata/Resource/Factory/FiltersResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/FiltersResourceMetadataCollectionFactory.php
@@ -51,6 +51,11 @@ final class FiltersResourceMetadataCollectionFactory implements ResourceMetadata
 
         foreach ($resourceMetadataCollection as $i => $resource) {
             foreach ($operations = $resource->getOperations() ?? [] as $operationName => $operation) {
+                if (null !== $operation->getFilters()) {
+                    // filters are already applied on this operation
+                    continue;
+                }
+
                 $operations->add($operationName, $operation->withFilters(array_unique(array_merge($resource->getFilters() ?? [], $operation->getFilters() ?? [], $filters))));
             }
 


src/Metadata/Resource/Factory/FiltersResourceMetadataCollectionFactory.php
```diff
@@ -51,6 +51,11 @@ public function create(string $resourceClass): ResourceMetadataCollection
 
         foreach ($resourceMetadataCollection as $i => $resource) {
             foreach ($operations = $resource->getOperations() ?? [] as $operationName => $operation) {
+                if (null !== $operation->getFilters()) {
+                    // filters are already applied on this operation
+                    continue;
```
mrossard, line 56: Aren't you losing the possibility to add filters at the resource level to an operation that has predefined filters here?
If I understant the current version correctly, it's made to let you have the filters option on the operation and resources filters active at the same ttime

vincentchalamon: I think if the operation already has filters, we should not override it, even from resource filters. For example:
```php
#[ApiResource(
    filters: ['foo', 'bar'],
    uriTemplate: '/foos',
    operations: [
        new GetCollection(), // inherit filters and uriTemplate from resource
        new GetCollection(filters: ['bar', 'baz'], uriTemplate: '/baz'), // override filters and uriTemplate
        new GetCollection(filters: [], uriTemplate: '/no-filters'), // disable filters and override uriTemplate
    ]
)]
```

mrossard: It may be the right way to do this, but since it currently merges both your change would be a BC break wouldn't it? 
I'm just saying this should at least be documented if it's changed this way.

vincentchalamon: Indeed, it will change the behavior and it's a BC break. WDYT @soyuka?

soyuka:
I answered in the issue I don't think that we should change this.

vincentchalamon:
Closing as not the good approach (cf. comments).

PR #5668, opened 2023-07-13
Fix(GraphQL): Add cache_key in item normalizer

xavierleune:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | ¤
| License       | MIT
| Doc PR        | ¤

Hello everyone,
it looks like the localCache in the Serializer is not used for the GraphQL ItemNormalizer, as it's done in the other normalizers.
This small fix reduces the calls to `getAttributes` from the Serializer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ concurrency:
 
 env:
   COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  COVERAGE: '0'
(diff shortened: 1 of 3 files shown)


soyuka:
it looks like there's a security issue on the behat tests, I think that you need to include the operation name in the cache key to be sure that secured operations have their own cache key

xavierleune:
@soyuka thanks for pointing me in that direction. It seems that the operation name is not enough to fix that issue, but I'll dig more from here.

xavierleune:
Got lost in rebase, trying again on a new PR, sorry !

PR #5674, opened 2023-07-17
feat: referencing context via http header link

Sarahshr:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT

Add `context_in_link` operation flag to enable referencing context via header link instead of body response

### DIFF
--- a/features/jsonld/context_in_link.feature
+++ b/features/jsonld/context_in_link.feature
@@ -1,0 +1,18 @@
+Feature: JSON-LD context in link
+  As a client software developer
+  I need to access to a JSON-LD context describing data types in header response
+
+  Scenario: Retrieve a Resource with "context_in_link" flag set to true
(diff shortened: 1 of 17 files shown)


dunglas:
Actually, referencing contexts with a `Link` header is only allowed as an upgrade path for plain JSON documents, but it's explicitly disallowed by the spec for documents already in JSON-LD: https://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld

> Please note that [JSON-LD documents](https://www.w3.org/TR/json-ld/#dfn-json-ld-document) served with the application/ld+json media type MUST have all context information, including references to external contexts, within the body of the document. Contexts linked via a http://www.w3.org/ns/json-ld#context [HTTP Link Header](https://tools.ietf.org/html/rfc8288#section-3) MUST be ignored for such documents.

Sorry for not having noticed this before.

PR #5676, opened 2023-07-18
fix(jsonld): check if docs enabled when generate vocab in context

Romaixn:
| Q             | A
| ------------- | ---
| Branch       | 3.1
| Tickets       | #5528 
| License       | MIT

Check if `enable_docs` option is enabled before adding `@vocab` in JSON-LD context to avoid error route not found.

Transform this : 
```json
{
	"@context": {
		"hydra": "http:\/\/www.w3.org\/ns\/hydra\/core#",
		"@vocab": "https:\/\/localhost\/docs.jsonld#",
		"isbn": "Book\/isbn",
		"title": "Book\/title",
		"description": "Book\/description",
		"author": "Book\/author",
		"publicationDate": "Book\/publicationDate",
		"reviews": "Book\/reviews",
		"cover": "Book\/cover",
		"archivedAt": "Book\/archivedAt"
	}
}
```

To this : 
```json
{
	"@context": {
		"hydra": "http:\/\/www.w3.org\/ns\/hydra\/core#",
		"isbn": "Book\/isbn",
		"title": "Book\/title",
		"description": "Book\/description",
		"author": "Book\/author",
		"publicationDate": "Book\/publicationDate",
		"reviews": "Book\/reviews",
		"cover": "Book\/cover",
		"archivedAt": "Book\/archivedAt"
	}
}
```

### DIFF
--- a/src/JsonLd/ContextBuilder.php
+++ b/src/JsonLd/ContextBuilder.php
@@ -35,7 +35,7 @@ final class ContextBuilder implements AnonymousContextBuilderInterface
 
     public const FORMAT = 'jsonld';
 
-    public function __construct(private readonly ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory, private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly UrlGeneratorInterface $urlGenerator, private readonly ?IriConverterInterface $iriConverter = null, private readonly ?NameConverterInterface $nameConverter = null)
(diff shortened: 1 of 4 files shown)


dunglas:
This doesn't look good to me. RDF types are now wrong. Even if it's considered a best practice to use dereferenceable URI as types, it's not mandatory. RDF parsers will not even try to hit the URL. It's just a namespace.

I would just keep things as-is.

Romaixn:
Alternatively, we could attempt to replace the `@vocab` with something else. The issue is that if the `enable_docs` parameter is set to false, the URL `/contexts` returns an error since the route to the documentation doesn't exist.

Or we could assume that if we disable the documentation, we also disable the `/contexts` route ?

dunglas:
No, contexts must always exist according to the JSON-LD spec.

The error is likely because the docs route doesn't exist in the Symfony router.

IMHO we have two options:

1. Always register the docs route but "manually" throw a `NotFoundHttpException` if `enable_docs` is set to false (then the route will technically exist in the Symfony router
2. Always generate the Hydra "docs" if JSON-LD is enabled, even if `enable_docs` is set to false (but continue to disable the OpenAPI endpoint, Swagger UI etc)

IMHO option 2 is more correct and easier to implement.

WDYT?

Romaixn:
I think you're right; the second option is the best. I will modify my PR accordingly, by enabling only the route to `/docs.jsonld` when `enable_docs` is set to `false`.

PR #5677, opened 2023-07-18
Allow defining controller as a class method (controller: 'Foo::barAction')

gquemener:
| Q             | A
| ------------- | ---
| Tickets       | 
| License       | MIT
| Doc PR        |

Prior to Api Platform 3, it was possible to define operation controller as a class method (using the `class::method` notation).

This seems to not work anymore in version 3, only controller class containing an `__invoke` method seems to be supported.

I have not performed a thorough examination of the code base, however the ApiLoaderTest doesn't present any usage of the mentioned notation.

### DIFF
--- a/src/Symfony/Routing/ApiLoader.php
+++ b/src/Symfony/Routing/ApiLoader.php
@@ -79,8 +79,11 @@ final class ApiLoader extends Loader
                         $path = str_replace('{._format}', '.{_format}', $path);
                     }
 
-                    if (($controller = $operation->getController()) && !$this->container->has($controller)) {
-                        throw new RuntimeException(sprintf('There is no builtin action for the "%s" operation. You need to define the controller yourself.', $operationName));
+                    if ($controller = $operation->getController()) {
+                        $controllerId = explode('::', $controller)[0];
+                        if (!$this->container->has($controllerId)) {
+                            throw new RuntimeException(sprintf('There is no builtin action for the "%s" operation. You need to define the controller yourself.', $operationName));
+                        }
                     }
 
                     $route = new Route(
--- a/src/Symfony/Routing/ApiLoader.php
+++ b/src/Symfony/Routing/ApiLoader.php
@@ -82,7 +82,7 @@ final class ApiLoader extends Loader
                     if ($controller = $operation->getController()) {
                         $controllerId = explode('::', $controller)[0];
                         if (!$this->container->has($controllerId)) {
-                            throw new RuntimeException(sprintf('There is no builtin action for the "%s" operation. You need to define the controller yourself.', $operationName));
+                            throw new RuntimeException(sprintf('Operation "%s" is defining an unknown service as controller "%s". Make sure it is properly registered in the dependency injection container.', $operationName, $controllerId));
                         }
                     }
 
--- a/tests/Symfony/Routing/ApiLoaderTest.php
+++ b/tests/Symfony/Routing/ApiLoaderTest.php
@@ -257,6 +257,20 @@ class ApiLoaderTest extends TestCase
         );
     }
 
+    public function testApiLoaderWithUndefinedControllerService(): void
+    {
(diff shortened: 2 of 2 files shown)


mrossard:
If I'm not mistaken that's already available through something like 
```
provider: [SomeClass::class, 'functionName']
```

mrossard:
> Provider has to implement `ProviderInterface`, don't they? If that so, they won't properly replace a controller (what about controllers that perform mutation?).
> 
> I believe the only valid alternatives as of today are:
> 
> * not to define an API Platform operation for such case, but map the route manually to the controller (using Symfony `#[Route]` for instance)
> * define an invokable controller, and pass it to the `controller` operation attribute
> 
> What do you think?

I mostly think i read too fast, i was sure you were talking abnout providers here :D 

You can use callables the way i was showing on providers and processors, which doesn't require the classes ussed to actually implement the interfaces (see here in the test suite for example : https://github.com/api-platform/core/blob/27f20969f532ebd748c11ad15d8e57366325f4cf/tests/Fixtures/TestBundle/ApiResource/PostWithUriVariables.php#L23 ).
 
That would be an answer for some use cases of custom controllers, but maybe not enough for all. I'm pretty sure everything is doable without custom controllers (the use case would be interesting!), but simple symfony controllers might be less work in some cases. ;)

gquemener:
Thanks for your inputs.

> You can use callables the way i was showing on providers and processors, which doesn't require the classes used to actually implement the interfaces

Well, that's unexpected, I would have expected that providers and processors should always implement the relevant interfaces, but why not!

To contextualize my effort here, I'm working on a code base that was implemented with Api Platform <2.7 and we are on a trail to upgrade to 3.1.

We have first worked to make it compatible with API Platform 3, thanks to the BC layer provided in 2.7.

We are now on the edge to update our version constraint to ^3.1, however still have 60+ operations defining controller as `FQCN::methodName`, which is not supported anymore by API Platform 3.1.

My question is: is it a mistake? Should it still be supported? Some under documented behaviour that didn't survive the work done on API Platform 3 (and in this case that PR solves it, I believe), or was it done on purpose (however I can't find any trace in the changelog, even if I know it's not meant to be an ultra-exhaustive list of all changes) and in that case we will make the necessary changes in our code base?

gquemener:
In case this BC break is intended, I've prepared a documentation PR: https://github.com/api-platform/docs/pull/1784

src/Symfony/Routing/ApiLoader.php
```diff
@@ -79,8 +79,14 @@ public function load(mixed $data, string $type = null): RouteCollection
                         $path = str_replace('{._format}', '.{_format}', $path);
                     }
 
-                    if (($controller = $operation->getController()) && !$this->container->has($controller)) {
-                        throw new RuntimeException(sprintf('There is no builtin action for the "%s" operation. You need to define the controller yourself.', $operationName));
+                    if ($controller = $operation->getController()) {
+                        $controllerId = $controller;
+                        if (str_contains($controller, '::')) {
+                            [$controllerId] = explode('::', $controller);
+                        }
```
dunglas, line 86:
```suggestion
                        $controllerId = explode('::', $controller, 2)[0]
```

dunglas (changes requested):
Supporting again this syntax looks good to me. I just suggested a tiny code simplification.

src/Symfony/Routing/ApiLoader.php
gquemener: I've just applied your suggestion :ok_hand:

gquemener:
Should I target the 3.1 branch, as this could be defined as a bug (BC break)?

gquemener:
And if may ask, would it be possible to release a patch version v3.1.13 when this gets merged?

dunglas:
Yes, please target 3.1!

gquemener:
Reopened against 3.1 branch in https://github.com/api-platform/core/pull/5681

PR #5684, opened 2023-07-21
fix: enable to update all keys in metadata

### LINKED ISSUE #5482: Resource properties configuration in YAML doesn't work
**API Platform version(s) affected**:  3.1.0

**Description**  
I tried to set the openapiContext property in a Yaml file, but it doesn't work. It woks fine in XML, even if I set the resource in a YAML file and the properties in XML. 

**How to reproduce**  
Try this conf [from docs](https://api-platform.com/docs/core/openapi/#using-the-openapi-and-swagger-contexts):
```yaml
properties:
    App\Entity\Product:
        timestamp:
            attributes:
                openapiContext:
                    type: string
                    format: date-time
```
Property's openapiContext configuration is ignored. And then try the same configuration but in XML format, also taken from docs:
```xml
<!-- api/config/api_platform/properties.xml -->

<properties xmlns="https://api-platform.com/schema/metadata/properties-3.0"
           xsi:schemaLocation="https://api-platform.com/schema/metadata/properties-3.0
           https://api-platform.com/schema/metadata/properties-3.0.xsd">
    <property resource="App\Entity\Product" name="timestamp">
        <openapiContext>
            <values>
                <value name="type">string</value>
                <value name="format">date-time</value>
            </values>
        </openapiContext>
    </property>
</properties>
```
And this one works just fine. Resource shows the date-time format for timestamp property in SwaggerUI.

PS: I also tried removing the `attributes` key as I don't see it on YamlPropertyExtractor, but it doesn't work either.

image

ttskch:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #5482
| License       | MIT
| Doc PR        | -

This PR fixes #5482.

Because `ExtractorPropertyMetadataFactory::update()` method was updating only some of the properties of the `ApiProperty` class, so in decorated factories, metadata such as `openapiContext` were not being updated.

### DIFF
--- a/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
@@ -99,6 +99,23 @@ final class ExtractorPropertyMetadataFactory implements PropertyMetadataFactoryI
             'readableLink' => 'is',
             'required' => 'is',
             'identifier' => 'is',
+            'default' => 'get',
+            'example' => 'get',
+            'deprecationReason' => 'get',
+            'fetchable' => 'is',
+            'fetchEager' => 'get',
+            'jsonldContext' => 'get',
+            'openapiContext' => 'get',
+            'jsonSchemaContext' => 'get',
+            'push' => 'get',
+            'security' => 'get',
+            'securityPostDenormalize' => 'get',
+            'types' => 'get',
+            'builtinTypes' => 'get',
+            'schema' => 'get',
+            'initializable' => 'is',
+            'genId' => 'get',
+            'extraProperties' => 'get',
         ];
 
         foreach ($metadataAccessors as $metadataKey => $accessorPrefix) {


src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
```diff
@@ -95,10 +95,27 @@ private function update(ApiProperty $propertyMetadata, array $metadata): ApiProp
             'description' => 'get',
             'readable' => 'is',
             'writable' => 'is',
-            'writableLink' => 'is',
```
dunglas, line 98: Can you don't move existing keys (to ease merge commits) please?

dunglas (approved):

src/Metadata/Property/Factory/ExtractorPropertyMetadataFactory.php
ttskch: @dunglas OK, I've fixed and force-pushed as 624105c. Thanks.

dunglas:
Thanks @ttskch!

PR #5686, opened 2023-07-22
fix(GraphQL): Add cache_key in item normalizer

xavierleune:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | ¤
| License       | MIT
| Doc PR        | ¤

Hello everyone,
it looks like the localCache in the Serializer is not used for the GraphQL ItemNormalizer, as it's done in the other normalizers.
This small fix reduces the calls to `getAttributes` from the Serializer.


xavierleune:
@soyuka following #5668, tests seems ok now, before I add some unit tests, what do you think of this approach ?

soyuka:
Why not, I don't have any other solution in mind for this so let's go with this!

xavierleune:
thanks for the rebase (and the merge) @soyuka !

PR #5697, opened 2023-07-24
ci: allow ramsey/uuid-doctrine

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "phpstan/phpstan-phpunit": "^1.0",
         "phpstan/phpstan-symfony": "^1.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
-        "ramsey/uuid": "^3.7 || ^4.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -53,7 +53,7 @@
         "phpstan/phpstan-symfony": "^1.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
         "ramsey/uuid": "^3.7 || ^4.0",
```
divine, line 55:
```suggestion
        "ramsey/uuid": "^3.9.7 || ^4.0",
```

See @ https://github.com/ramsey/uuid/issues/439

Also `symfony/dependency-injection` needs to be pinned to ^6.1.12 due to bug @ https://github.com/symfony/symfony/pull/49126

Thanks!

dunglas: Thanks, updated!

PR #5703, opened 2023-07-26
fix(openapi): correctly output OpenAPI parameters

ttskch:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR fixes following bug:

When parameters are set in openapiContext for some operations with some filters applied, the filter parameters are lost and only the parameters set in openapiContext are output to OpenAPI.

The cause of this bug was as follows. Even though the pagination and filter parameters are merged at [here](https://github.com/api-platform/core/blob/24133dfecb584e50e26503d852c654fe2224e45d/src/OpenApi/Factory/OpenApiFactory.php#L275-L283), they are reset to openapiContext parameters value at [here](https://github.com/api-platform/core/blob/24133dfecb584e50e26503d852c654fe2224e45d/src/OpenApi/Factory/OpenApiFactory.php#L369-L374).

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -370,6 +370,7 @@ final class OpenApiFactory implements OpenApiFactoryInterface
                         'externalDocs' => new ExternalDocumentation(description: $value['description'] ?? '', url: $value['url'] ?? ''),
                         'requestBody' => new RequestBody(description: $value['description'] ?? '', content: isset($value['content']) ? new \ArrayObject($value['content'] ?? []) : null, required: $value['required'] ?? false),
                         'callbacks' => new \ArrayObject($value ?? []),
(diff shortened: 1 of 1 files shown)


soyuka:
Thanks, may you add a non regression test? Also your commit scope could be: `fix(openapi): ...`

ttskch:
@soyuka 

Sorry, `openapiContext:` is a deprecated feature, and for that it is too much work to modify [OpenApiFactoryTest](https://github.com/api-platform/core/blob/b8ca845a6988b2634a7459519d1657c2ecf063fc/src/OpenApi/Tests/Factory/OpenApiFactoryTest.php) for this PR, so if you can't merge without adding the test, please reject this PR.

I have already confirmed that using `openapi:` instead of `openapiContext:` will not cause this problem.

Thanks.

PR #5711, opened 2023-08-02
feat(graphql): allow to disable the introspection query

epourail:
| Q             | A
| ------------- | ---
| Branch?   | main
| Tickets     | -
| License    | MIT
| Doc PR    | api-platform/docs#1792

For security reason, the introspection query should be disabled to not expose the schema.
This PR allows to enable/disable the introspection query througth the bundle configuration:

```yaml
api_platform:
  ...
  graphql:
    introspection:
      enabled: true # (or false)    
  ...
```

By default, the introspection is enabled to avoid BC and keep the current behavior.

**TODO**:
- [x] update the documentation
- [x] how to test ?


epourail:
What do you think of this feature ? If ok with such improvment, I will update the documentation and try to test.

src/GraphQl/Executor.php
```diff
@@ -24,11 +25,19 @@
...
+    public function __construct(private readonly bool $graphQlIntrospectionEnabled = true)
+    {
+    }
+
     /**
      * {@inheritdoc}
      */
     public function executeQuery(Schema $schema, $source, mixed $rootValue = null, mixed $context = null, array $variableValues = null, string $operationName = null, callable $fieldResolver = null, array $validationRules = null): ExecutionResult
     {
+        $validationRules[] = new DisableIntrospection(
```
alanpoulain, line 37: You need to handle the case where `$validationRules` is `null`.

alanpoulain:
I think it's a nice feature, thanks. I'm not sure you can write functional tests, but unit ones should be fine.

src/GraphQl/Executor.php
epourail: I updated the code to instantiate the rule globally for both reasons:
1) it will not create a new rule (`new DisableIntrospection()`) on each graphql query
2) the configuration yaml-file setup the bundle globally

 ... but  the code is now in the constructor.

@alanpoulain are you ok with that or I keep the previous logic ?

epourail:
Any clue on the phpunit status regarding the unitary tests

tests/GraphQl/ExecutorTest.php
```diff
@@ -0,0 +1,43 @@
...
+ * @author Julien Verger <julien.verger@gmail.com>
+ */
+class ExecutorTest extends TestCase
+{
+    public function testEnableIntrospectionQuery(): void
+    {
+        $executor = new Executor(true);
+
+        $expected = new DisableIntrospection(DisableIntrospection::ENABLED);
+        $this->assertIsObject($executor);
```
alanpoulain, line 31: **nitpick:** It's not needed.

epourail: fixed.

alanpoulain:
Thank you @epourail!

PR #5719, opened 2023-08-04
docs: Explain how to disable the automatic integration with LexikJWTAuthenticationBundle

jonag:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

With #5609 the integration of API Platform with LexikJWTAuthenticationBundle is automatically enabled. However, if we are not using the default configuration of LexikJWTAuthenticationBundle the documented operation is wrong (wrong path, wrong response in my case) and creates a duplicate. Since the behaviour of API Platform changed in a patch version maybe it's a good idea to explain directly in the changelog how to go back to the previous behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 ### Bug fixes
 
 * [0c1c1c36f](https://github.com/api-platform/core/commit/0c1c1c36f55ba96119e6c22ea25fb69c85b20161) fix(symfony): enable API Platform in LexikJWTAuthenticationBundle (#5609)
+
+
+  You can disable this behaviour by setting the configuration key `lexik_jwt_authentication.api_platform.enabled` to `false`
+
 * [146991ba4](https://github.com/api-platform/core/commit/146991ba42014f0141195bc43be5cd875665407c) fix(openapi): merge parameters with deprecated openApiContext (#5703)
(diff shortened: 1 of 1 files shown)


divine:
That behavior was incorrect and a "bug" that was fixed. In my case, it worked fine before (enabled automatically) after the upgrade I found it doesn't work anymore.

Maybe push your docs suggestion on how to disable it to https://github.com/api-platform/docs/blob/3.1/core/jwt.md  or https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/Resources/doc/index.rst#configuration?

Thanks!

soyuka:
we shouldn't prepend if you configured it manually, imo https://github.com/api-platform/core/pull/5609/files needs a patch no?

soyuka:
Merging this for now, don't hesitate to provide a patch if there's something better to be done. Thanks!

jonag:
Detecting if a user has previously manually documented the login endpoint would be quite difficult, so I think it's good as-is.

Enabling the API Platform integration when it's not needed just adds some "noise" in the swagger documentation but doesn't break existing apps, so it should be good.

PR #5721, opened 2023-08-04
fix: perf regression with Symfony 6.3 

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes n/a
| License       | MIT
| Doc PR        | n/a

The switch to the new `NormalizerInterface::getSupportedTypes()` system provided by Symfony in #5672 introduced a performance regression.

Symfony 6.2 + API Platform 3.1.12 (faster) VS Symfony 6.3 + API Platform 3.1.13

image

With this patch fixes the issue: Symfony 6.2 + API Platform 3.11.12 VS Symfony 6.3 + API Platform with this patch

image

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -92,7 +92,6 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     {
         return [
             'object' => true,
-            'native-array' => true,
         ];
     }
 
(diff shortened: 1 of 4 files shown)


dunglas:
Answer to myself: we can because the cache is split per concrete type, so this PR is ready to be merged.

https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Serializer/Serializer.php#L265

Renrhaf:
Thank you @dunglas 👍

PR #5729, opened 2023-08-08
fix(jsonld): skolem uri template may have a _format

soyuka:


soyuka:
needs https://github.com/api-platform/core/pull/5729

Issue #5728, opened 2023-08-08
Fonts in .woff-Format are missing


Chris53897:
**API Platform version(s) affected**: 3.x

**Description**  
The @font-face css-code in fonts are referencing the *.woff Format. But they are not shipped with this repo.
https://github.com/api-platform/core/tree/main/src/Symfony/Bundle/Resources/public/fonts/open-sans

Example https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/Resources/public/fonts/open-sans/700.css#L7

**How to reproduce**  
See links above I just noticed this with running AssetMapper. This will produce warnings for files not found.
symfony console asset-map:compile  

**Possible Solution**  

a) Add .woff Files. 
b) Remove from .css-File (prefered solution. https://caniuse.com/?search=woff2)

PR #5730, opened 2023-08-08
minor: remove reference to .woff fonts-format

Chris53897:
https://github.com/api-platform/core/issues/5728


soyuka:
isn't that a bug fix ?

Chris53897:
yes, it is a minor fix. But only 5% of browsers are effected, and a wrong font is not so dramatic.
I just want to get rid of the warnings.

If you want, i can change the title. I have to look up how to change the commit message.

soyuka:
We want to keep the older format for older  browsers I guess...

Chris53897:
@soyuka ? The files do not exist in the repo.
If you want to add support of the woff format, the files need to be added in the first place.

soyuka:
thanks!

PR #5739, opened 2023-08-09
fix(serializer) Restore the use of placeholders named differently than `id`

### LINKED ISSUE #5736: PR #5604 breaks operations on subresources for PUT operation on Resource when placeholders are named differently than "id"
**API Platform version(s) affected**: 3.1.13

**Description**  
With PR #5546, was introduced the ability to operate on subresources also if the placeholders of the IDs are not named "id".

The PR #5604 actually breaks this feature.

If placeholders are named "id", the code continues to work as there is a fallback in `ItemNormalizer` that returns `$data['id']` if the key `uri_variables` is not set in the `$context`:

https://github.com/api-platform/core/blob/92a81f024541054b9322e7457b75c721261e14e0/src/Serializer/ItemNormalizer.php#L87-L91

But, as the key `uri_variables` is now always removed, the method `ItemNormalizer::getContextUriVariables()` never works as it should.

The PR #5604 was published due to those issues:

- https://github.com/api-platform/core/issues/5587
- https://github.com/api-platform/core/issues/5603

**Reading them behind the lines, they are originates from the same problem: both the Resource and the SubResource have the placeholder named "id".**

This causes the confusion in `ItemNormalizer` that uses the wrong "id" key to retrieve the SubResource.

**How to reproduce**

Create two resources, for example Alpha and Beta
Create a field in Beta which is a relation to Alpha
Try to create/update Beta

**Working on 3.1.12 | NOT working on 3.1.13**

```
PATCH /betas/{beta_id}
{
  "alpha": "/alphas/{alpha_id}"
}
```

**NOT Working on 3.1.12 | (Now) Working on 3.1.13**

```
PATCH /betas/{id}
{
  "alpha": "/alphas/{id}"
}
```

**Possible Solution**  

https://github.com/api-platform/core/pull/5739

Aerendir:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Tickets       | Closes #5736
| License       | MIT

Restore the use of placeholders named differently than `id`.

See #5736 for more details.

- [ ] Verify why `ItemNormalizerTest::testDenormalizeGuessingUriVariables()` din't intercepted the breaking change introduced with #5604 (I need help here, please)

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -513,9 +513,12 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $collectionKeyType = $type->getCollectionKeyTypes()[0] ?? null;
         $collectionKeyBuiltinType = $collectionKeyType?->getBuiltinType();
         $childContext = $this->createChildContext(['resource_class' => $className] + $context, $attribute, $format);
-        unset($childContext['uri_variables']);
-        if ($this->resourceMetadataCollectionFactory) {
-            $childContext['operation'] = $this->resourceMetadataCollectionFactory->create($className)->getOperation();
+        $lastUriVariable = \array_key_exists('uri_variables', $childContext) ? array_key_last($childContext['uri_variables']) : null;
+        if ('id' === $lastUriVariable) {
+            unset($childContext['uri_variables']);
+            if ($this->resourceMetadataCollectionFactory) {
+                $childContext['operation'] = $this->resourceMetadataCollectionFactory->create($className)->getOperation();
+            }
         }
 
         $values = [];


soyuka:
We need to handle this differently, it should just find the correct ids no? I'll try to find out the issue...

Aerendir:
Yes, it should find the correct placeholder, given the resource. But as of what I saw, the array is flat, so, if you have two placeholders with the same name, then the one overwrites the other.

This would happen even if the placeholder were named, for example, `resource_id`.

But, with this change, if it is named `resource_id`, the code may break; if it is named `id`, the code will handle the situation.

I agree that a cleaner solution should be found, but I currently cannot find one: I hope you will as you know the code better than me 💪

Maybe in the mean time this PR may be published and a new patch release issued, just to don't break existing apps that rely on ApiPlatform?

soyuka:
https://github.com/api-platform/core/pull/5743/files#r1291349976

PR #5742, opened 2023-08-11
chore: subtree split (openapi, doctrine, serializer, ramsey-uid bridge, graphql, http-cache)

soyuka:
Note that deprecations will be triggered only on the main branch, mainly this moves things from the `ApiPlatform\Api` namespace to the `ApiPlatform\Metadata` namespace. Exceptions are also moved to `ApiPlatform\Metadata`. This patch introduces a backward compatibility layer to support older namespaces until 4.0.

As tests move to another directory coverage isn't correct anymore I'll try to fix that in another PR.


src/Api/IdentifiersExtractor.php
```diff
@@ -28,6 +28,8 @@
 /**
  * {@inheritdoc}
  *
+ * @deprecated use ApiPlatform\Metadata\IdentifiersExtractor instead
```
chalasr, line 31: This should not go in 3.1 stable as per [semver](https://semver.org/#spec-item-7).

PR #5743, opened 2023-08-11
fix(serializer): use data if no uri_variables provided

### LINKED ISSUE #5736: PR #5604 breaks operations on subresources for PUT operation on Resource when placeholders are named differently than "id"

soyuka:
fixes #5736

### DIFF
--- a/features/main/patch.feature
+++ b/features/main/patch.feature
@@ -58,3 +58,25 @@ Feature: Sending PATCH requets
       }
     }
     """
+
+  Scenario: Patch a relation with uri variables that are not `id`
+    When I add "Content-Type" header equal to "application/merge-patch+json"
+    And I send a "PATCH" request to "/betas/1" with body:
+    """
+      {
+        "alpha": "/alphas/2"
+      }
+    """
+    Then the response should be in JSON
+    And the response status code should be 200
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/Beta",
+      "@id": "/betas/1",
+      "@type": "Beta",
+      "betaId": 1,
+      "alpha": "/alphas/2"
+    }
+    """
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -20,6 +20,7 @@ use ApiPlatform\Exception\InvalidArgumentException;
 use ApiPlatform\Exception\ItemNotFoundException;
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\Exception\OperationNotFoundException;
 use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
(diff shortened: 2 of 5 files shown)


src/Serializer/ItemNormalizer.php
```diff
@@ -86,11 +86,7 @@ private function updateObjectToPopulate(array $data, array &$context): void
 
     private function getContextUriVariables(array $data, $operation, array $context): array
     {
-        if (!isset($context['uri_variables'])) {
-            return ['id' => $data['id']];
-        }
-
-        $uriVariables = $context['uri_variables'];
+        $uriVariables = $context['uri_variables'] ?? $data;
```
soyuka, line 89: @Aerendir can you try this?

features/main/patch.feature
```diff
@@ -58,3 +58,26 @@ Feature: Sending PATCH requets
...
+    And the JSON should be equal to:
+    """
+    {
+      "@context": "/contexts/Beta",
+      "@id": "/betas/1",
+      "@type": "Beta",
+      "betaId": 1,
+      "alpha": "/alphas/2"
+    }
+    """
```
soyuka, line 83: @Aerendir this works right now so I'm not sure what the issue is, are you using merge-patch?

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -899,4 +877,28 @@ private function setValue(object $object, string $attributeName, mixed $value):
...
+            unset($context['operation'], $context['operation_name']);
+            $context['resource_class'] = $resourceClass;
+            if ($this->resourceMetadataCollectionFactory) {
+                $context['operation'] = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation();
+                $context['operation_name'] = $context['operation']?->getName();
+            }
+        }
+
+        return $context;
+    }
```
soyuka, line 903: @jonag can you try this and use `root_operation`?

jonag: Updating to API Platform v3.1.14 and changing my custom normalizer to use the root_operation instead of the operation fixes all my tests! 
Thank you 👏!

src/Serializer/ItemNormalizer.php
Aerendir: Hi @soyuka , I've tried the fix, but it doesn't work...

In `ItemNormalizer::getContextUriVariables()`, there is no key `$context['uri_variables']`, so the variable `$uriVariable` gets the value of `$data`, but `$data` contains the entity and so there will never be the uri variables needed.

I'm going to reproduce in tests what's going on...

features/main/patch.feature
Aerendir: @soyuka , no, I'm using `PUT`s... I'm going to reproduce the scenario...

PR #5753, opened 2023-08-17
chore(HttpCache): outsource cache tag collection into normalizer events

usu:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | 
| License       | MIT
| Doc PR        | 

This PR outsources the collection of cache tags for HTTP cache into individual events triggered from the normalizers instead of hardcoding into the the normalizers. This allows users to override or extend the cache tag logic with their own implementation.

This is related to https://github.com/api-platform/core/pull/5650, where I brought up the idea:

> Would it makes sense to outsource the actual calculation of the cache tags into seperate event listeners instead of hard-baking it into AbstractItemNormalizer? This would it make easier for users to extend or overwrite the cache tag generation with their own logic.

The same events could later be extended for other purposes where we need to trigger side effects during normalization (e.g. to collect push resources). And of course can be used by users for other user specific business logic.

Should this be documented somewhere? And if so, where and how?

### DIFF
--- a/src/HttpCache/EventListener/NormalizerListener.php
+++ b/src/HttpCache/EventListener/NormalizerListener.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\HttpCache\EventListener;
+
+use ApiPlatform\Serializer\NormalizeItemEvent;
+
+/**
+ * Collects cache tags during normalization.
+ *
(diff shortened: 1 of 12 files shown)


soyuka:
Not a huge fan of an event dispatcher, we could just use a callback or some service. I agree that this extension point is missing today...

usu:
> Not a huge fan of an event dispatcher, we could just use a callback or some service. I agree that this extension point is missing today...

Ok. I thought events would give more flexibility to re-use them for other purposes as well - both on api-platform side and on userland (e.g. generating list of resources to push).

But I can also make a proposal using a service instead of events.

usu:
> Not a huge fan of an event dispatcher, we could just use a callback or some service. I agree that this extension point is missing today...

@soyuka Proposal with service instead of events is at https://github.com/api-platform/core/pull/5758

usu:
Superseded by https://github.com/api-platform/core/pull/5758

PR #5756, opened 2023-08-17
Setting object to OBJECT_TO_POPULATE for security expression

### LINKED ISSUE #5755: "object" null in security attribute for canAccessAttribute() on PUT/PATCH
**API Platform version(s) affected**: x3.1

**Description**  

Imagine this property on an `ApiResource`:

```php
class Cake
{
    // ...

    #[ApiProperty(security: 'is_granted("FLAVOR", object)')]
    private ?string $flavor = null;
}
```

During a `GET` operation, the `object` variable is the `Cake` object. Expected! 🥇 
However, during a `PATCH` operation, the `object` variable is `null` during deserialization.

I believe this is a bug, as there IS an object available during deserialization for PUT/PATCH. The current behavior makes it impossible to make `flavor` READable using the security expression without preventing it from always failing to be WRITEable because the `object` is `null` (and your voter probably needs the `object`).

The workaround is bizarre:

```php
class Cake
{
    // ...

    #[ApiProperty(security: 'object === null or is_granted("FLAVOR", object)')]
    private ?string $flavor = null;
}
```

You have to allow the security to pass if the object is `null` so that deserialization works for PUT/PATCH. Then, during serialization, `object` will be populated so the voter will always run.

**How to reproduce**  
Small reproducer! https://github.com/weaverryan/api_platform_null_object_security_reproducer

**Possible Solution**  
Somewhere around https://github.com/api-platform/core/blob/60075501a6d15d03d4db26f00715a52cac068977/src/Serializer/AbstractItemNormalizer.php#L443, if `$object` is null, we look also for `$context['object_to_populate']`.

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | Closes #5755
| License       | MIT
| Doc PR        | Not needed

Hi!

Fixes #5755 - unless I am approaching the problem there incorrectly.

Cheers!

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -441,7 +441,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $security = $propertyMetadata->getSecurity();
         if (null !== $this->resourceAccessChecker && $security) {
             return $this->resourceAccessChecker->isGranted($context['resource_class'], $security, [
-                'object' => $object,
+                'object' => $object ?? $context[self::OBJECT_TO_POPULATE] ?? null,
             ]);
         }
 


soyuka:
See reasoning inside the issue.

PR #5757, opened 2023-08-19
chore(serializer): outsource cache tag collection into separate service

usu:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | Supersedes #5753 
| License       | MIT
| Doc PR        | 

Alternative implementation of #5753 but using a service instead of events


usu:
wrong target branch

PR #5758, opened 2023-08-19
feat(serializer): collect cache tags using a TagCollector

usu:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Supersedes #5753 
| License       | MIT
| Doc PR        | 

Alternative implementation of #5753 but using a service instead of events

### DIFF
--- a/src/JsonApi/Serializer/ItemNormalizer.php
+++ b/src/JsonApi/Serializer/ItemNormalizer.php
@@ -264,18 +264,23 @@ final class ItemNormalizer extends AbstractItemNormalizer
             return $normalizedRelatedObject;
         }
 
-        $data = [
+        $context['data'] = [
             'data' => [
                 'type' => $this->getResourceShortName($resourceClass),
                 'id' => $iri,
             ],
         ];
 
+        $context['data'] = $iri;
+        $context['object'] = $relatedObject;
+        unset($context['property_metadata']);
+        unset($context['api_attribute']);
+
         if ($this->tagCollector) {
-            $this->tagCollector->collect($relatedObject, $format, $context, $iri, $data);
+            $this->tagCollector->collect($context);
         }
 
-        return $data;
+        return $context['data'];
     }
 
     /**
--- a/src/JsonApi/Serializer/ItemNormalizer.php
+++ b/src/JsonApi/Serializer/ItemNormalizer.php
@@ -27,6 +27,7 @@ use ApiPlatform\Metadata\Util\ClassInfoTrait;
 use ApiPlatform\Serializer\AbstractItemNormalizer;
 use ApiPlatform\Serializer\CacheKeyTrait;
 use ApiPlatform\Serializer\ContextTrait;
+use ApiPlatform\Serializer\TagCollectorInterface;
 use ApiPlatform\Symfony\Security\ResourceAccessCheckerInterface;
 use Symfony\Component\ErrorHandler\Exception\FlattenException;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
@@ -55,9 +56,9 @@ final class ItemNormalizer extends AbstractItemNormalizer
 
     private array $componentsCache = [];
 
-    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface|LegacyIriConverterInterface $iriConverter, ResourceClassResolverInterface|LegacyResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, array $defaultContext = [], ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, ResourceAccessCheckerInterface $resourceAccessChecker = null)
+    public function __construct(PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface|LegacyIriConverterInterface $iriConverter, ResourceClassResolverInterface|LegacyResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, array $defaultContext = [], ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, ResourceAccessCheckerInterface $resourceAccessChecker = null, protected ?TagCollectorInterface $tagCollector = null)
     {
-        parent::__construct($propertyNameCollectionFactory, $propertyMetadataFactory, $iriConverter, $resourceClassResolver, $propertyAccessor, $nameConverter, $classMetadataFactory, $defaultContext, $resourceMetadataCollectionFactory, $resourceAccessChecker);
+        parent::__construct($propertyNameCollectionFactory, $propertyMetadataFactory, $iriConverter, $resourceClassResolver, $propertyAccessor, $nameConverter, $classMetadataFactory, $defaultContext, $resourceMetadataCollectionFactory, $resourceAccessChecker, $tagCollector);
     }
 
     /**
@@ -245,7 +246,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
             $iri = $this->iriConverter->getIriFromResource($relatedObject);
             $context['iri'] = $iri;
 
-            if (isset($context['resources'])) {
+            if (!$this->tagCollector && isset($context['resources'])) {
                 $context['resources'][$iri] = $iri;
             }
         }
@@ -263,12 +264,18 @@ final class ItemNormalizer extends AbstractItemNormalizer
             return $normalizedRelatedObject;
         }
 
-        return [
+        $data = [
             'data' => [
                 'type' => $this->getResourceShortName($resourceClass),
                 'id' => $iri,
             ],
         ];
+
+        if ($this->tagCollector) {
+            $this->tagCollector->collect($relatedObject, $format, $context, $iri, $data);
+        }
+
+        return $data;
     }
 
     /**
--- a/src/JsonApi/Serializer/ItemNormalizer.php
+++ b/src/JsonApi/Serializer/ItemNormalizer.php
@@ -271,7 +271,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
             ],
         ];
 
-        $context['data'] = $iri;
+        $context['iri'] = $iri;
         $context['object'] = $relatedObject;
         unset($context['property_metadata']);
         unset($context['api_attribute']);
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -63,7 +63,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
     protected array $localCache = [];
     protected array $localFactoryOptionsCache = [];
 
-    public function __construct(protected PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, protected PropertyMetadataFactoryInterface $propertyMetadataFactory, protected LegacyIriConverterInterface|IriConverterInterface $iriConverter, protected LegacyResourceClassResolverInterface|ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, array $defaultContext = [], ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, protected ?ResourceAccessCheckerInterface $resourceAccessChecker = null)
+    public function __construct(protected PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, protected PropertyMetadataFactoryInterface $propertyMetadataFactory, protected LegacyIriConverterInterface|IriConverterInterface $iriConverter, protected LegacyResourceClassResolverInterface|ResourceClassResolverInterface $resourceClassResolver, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, array $defaultContext = [], ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null, protected ?ResourceAccessCheckerInterface $resourceAccessChecker = null, protected ?TagCollectorInterface $tagCollector = null)
     {
         if (!isset($defaultContext['circular_reference_handler'])) {
             $defaultContext['circular_reference_handler'] = fn ($object): ?string => $this->iriConverter->getIriFromResource($object);
@@ -164,14 +164,22 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $emptyResourceAsIri = $context['api_empty_resource_as_iri'] ?? false;
         unset($context['api_empty_resource_as_iri']);
 
-        if (isset($context['resources'])) {
+        if (!$this->tagCollector && isset($context['resources'])) {
             $context['resources'][$iri] = $iri;
         }
 
         $data = parent::normalize($object, $format, $context);
 
         if ($emptyResourceAsIri && \is_array($data) && 0 === \count($data)) {
+            if ($this->tagCollector) {
+                $this->tagCollector->collect($object, $format, $context, $iri, $iri);
+            }
+
             return $iri;
+        }
+
+        if ($this->tagCollector) {
+            $this->tagCollector->collect($object, $format, $context, $iri, $data);
         }
 
         return $data;
@@ -670,7 +678,13 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
                 $resourceClass = $this->resourceClassResolver->getResourceClass($attributeValue, $className);
 
-                return $this->normalizeCollectionOfRelations($propertyMetadata, $attributeValue, $resourceClass, $format, $childContext);
+                $data = $this->normalizeCollectionOfRelations($propertyMetadata, $attributeValue, $resourceClass, $format, $childContext);
+
+                if ($this->tagCollector) {
+                    $this->tagCollector->collect($object, $format, $context, $context['iri'], $data, $attribute, $propertyMetadata, $type);
+                }
+
+                return $data;
             }
 
             if (
@@ -697,7 +711,13 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
                 $resourceClass = $this->resourceClassResolver->getResourceClass($attributeValue, $className);
 
-                return $this->normalizeRelation($propertyMetadata, $attributeValue, $resourceClass, $format, $childContext);
+                $data = $this->normalizeRelation($propertyMetadata, $attributeValue, $resourceClass, $format, $childContext);
+
+                if ($this->tagCollector) {
+                    $this->tagCollector->collect($object, $format, $context, $context['iri'], $data, $attribute, $propertyMetadata, $type);
+                }
+
+                return $data;
             }
 
             if (!$this->serializer instanceof NormalizerInterface) {
@@ -791,7 +811,9 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
         $iri = $this->iriConverter->getIriFromResource(resource: $relatedObject, context: $context);
 
-        if (isset($context['resources'])) {
+        if ($this->tagCollector) {
+            $this->tagCollector->collect($relatedObject, $format, $context, $iri, $iri);
+        } elseif (isset($context['resources'])) {
             $context['resources'][$iri] = $iri;
         }
 
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -168,18 +168,27 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             $context['resources'][$iri] = $iri;
         }
 
+        $context['object'] = $object;
+        $context['format'] = $format;
+
         $data = parent::normalize($object, $format, $context);
 
+        $context['data'] = $data;
+        unset($context['property_metadata']);
+        unset($context['api_attribute']);
+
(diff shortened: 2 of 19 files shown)


src/HttpCache/TagCollector.php
```diff
@@ -0,0 +1,46 @@
...
+
+use ApiPlatform\Metadata\ApiProperty;
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * Collects cache tags during normalization.
+ *
+ * @author Urban Suppiger <urban@suppiger.net>
+ */
+class TagCollector implements TagCollectorInterface
```
soyuka, line 24: can you explain to me the difference between all these methods? I think that we should be able to simplify this and reduce the number of arguments. I'd suggest:

```
use ApiPlatform\Serializer; // we want to avoid creating a dependency between Serializer and HttpCache, the opposite is fine.

interface IriCollectorInterface {
    public function collect(string $iri, mixed $object, array $context) {}
}
```

Not even sure if `object` matters in the signature and it could be added to the context? It'd be interesting to add the `attribute` that's being normalized if needed although that I'd start without it and if someone needs that we add it afterwards? Or do you already have a use case where that matters? 

If that's the case, we can put everything we need inside our context. I'd not have a `childContext` though, only give the current context it introduces less complexity and the use cases are often not good ones.

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -17,6 +17,7 @@
 use ApiPlatform\Api\ResourceClassResolverInterface;
 use ApiPlatform\Api\UrlGeneratorInterface;
 use ApiPlatform\Exception\ItemNotFoundException;
+use ApiPlatform\HttpCache\TagCollectorInterface;
```
soyuka, line 20: we want to avoid these kind of dependencies, the interface should be within our Serializer namespace, even if the implementation is the one from http cache.

src/HttpCache/TagCollector.php
usu: The various methods (extension points) are needed in order to implement the cache logic proposed under https://github.com/api-platform/core/pull/5650 in userland.

In order to implement this logic, the tag collector service on user side would look something like below (no guarantee for correctness, typing in a text editor only).

```php
<?php

class CustomTagCollector implements TagCollectorInterface
{
    public function collectTagsFromNormalize(mixed $object, string $format = null, array $context = [], string $iri = null): void
    {
        $this->addCacheTagForRessource($context, $iri);
    }

    public function collectTagsFromNormalizeRelation(mixed $object, string $format = null, array $context = [], string $iri = null): void
    {
    }

    public function collectTagsFromGetAttribute(mixed $object, string $format = null, array $context = [], string $iri = null, string $attribute = null, ApiProperty $propertyMetadata = null, Type $type = null, array $childContext = []): void
    {
        $this->addCacheTagsForRelation(array $context, $propertyMetadata)
    }

    private function addCacheTagForRessource(array $context, ?string $iri): void
    {
        if (isset($context['resources']) && isset($iri)) {
            $context['resources'][$iri] = $iri;
        }
    }
    
    private function addCacheTagsForRelation(array $context, ApiProperty $propertyMetadata): void
    {
        // Add cache tag for related collection
        if (isset($context['resources'])) {
            if (isset($propertyMetadata->getExtraProperties()['cacheDependencies'])) {
                foreach ($propertyMetadata->getExtraProperties()['cacheDependencies'] as $dependency) {
                    $cacheTag = $context['iri'].self::IRI_RELATION_DELIMITER.$dependency;
                    $context['resources'][$cacheTag] = $cacheTag;
                }
            } else {
                $cacheTag = $context['iri'].self::IRI_RELATION_DELIMITER.$context['api_attribute'];
                $context['resources'][$cacheTag] = $cacheTag;
            }
        }
    }
    
}
```

If there's only one `collect` method, I'd need to figure out from which point in the normalization process the collect method was called. Might be possible based on the content of the attributes, but not sure whether this leads to robust and understandable code.

With regards to parameters, in order to implement the original behavior of api-platform as well as the logic from https://github.com/api-platform/core/pull/5650 I'd need the following parameters as a minimum: `$context`, `$iri`, `$propertyMetadata`

All the rest (`$object`, `$format`, `$childContext`, `$attribute`, `$type`) can also be removed and we could add in case there's a specific usecase coming up for it. `$type` would be needed though, if anyone wants to differentiate relations to collections (*ToMany) and *ToOne relations.

src/JsonApi/Serializer/ItemNormalizer.php
usu: Got it. Will move to Serializer namespace.

src/HttpCache/TagCollector.php
usu: No prob, fully understand design of the interface is crucial.

For implementation of the current api-platform logic, I could use `$attribute` to ignore calls coming from `getAttribute`, because this parameter is not sent from `normalize` or `normalizeRelation`. I cannot use `$context['api_attribute']`, because this  is set for nested/embedded objects also for calls from `normalize` or `normalizeRelation`.

```php
class TagCollector implements TagCollectorInterface
{
    public function collect(mixed $object, string $format = null, array $context = [], string $iri = null, string $attribute = null, ApiProperty $propertyMetadata = null, Type $type = null, array $childContext = []): void
    {
        if(!$attribute){
            $this->addResourceToContext($context, $iri);
        }
    }

    private function addResourceToContext(array $context, ?string $iri): void
    {
        if (isset($context['resources']) && isset($iri)) {
            $context['resources'][$iri] = $iri;
        }
    }
}
```

Implementing the logic above for #5650 is a bit more tricky. With the current code, I don't see any good means to differentiate between calls from `normalize` and `normalizeRelation`. I could add an additional string attribute to disclose, where the collect call was made from. But honestly, that seems a bit like a disguised event system and I'd rather have a proper event system implemented as in the initial PR #5753.

Alternatively I think I could add the normalized `$data` back as an attribute and then check within collect, if data is an object (fully normalized embedded resource) or only a string (resource normalized as IRI only).

soyuka: You need to find out whether the call is made in a `normalizeRelation` as opposed to a `normalize`? you have `api_sub_level` for that IIRC.

usu: > You need to find out whether the call is made in a `normalizeRelation` as opposed to a `normalize`? 

Yeah, correct. https://github.com/api-platform/core/pull/5650 gives more background on the motivation. But in a nutshell, I'd want to treat fully embedded resources (resource embedded in the response as object) differently from linked resources (only IRI of resource included in response).

For the former, the cache tag is triggered from `normalize`. For latter, the cache tag is triggered from `normalizeRelation` (at least that's how it's implemented now).

> you have `api_sub_level` for that IIRC.

Not sure, this works. When running tags.feature tests, `api_sub_level` seems always to be true.

src/JsonApi/Serializer/ItemNormalizer.php
usu: Done with https://github.com/api-platform/core/pull/5758/commits/720efb82c1ee1cded55a95b853e01a1c0b7c9e44

src/HttpCache/TagCollector.php
usu: > Alternatively I think I could add the normalized `$data` back as an attribute and then check within collect, if data is an object (fully normalized embedded resource) or only a string (resource normalized as IRI only).

I just pushed a proposal implementing this. When passing `$data`, I can check on userland if `$data` is an array/object or a string (IRI) only and hence differentiate embedded objects from linked IRIs.

For JsonAPI this is a bit "trickier", because even for linked data, the [return value is an array](https://github.com/api-platform/core/pull/5758/files#diff-fe1eba7bf34edf60f0eefbfb87fda8fb5aff8f676ab95af88c3ffe276db8a253R241) and not only a string. However, as we pass `$format` as well, this could still be differentiated from a normal fully embedded object.

I had to change the tests, because now the order of tags has changed when waiting for the normalized data. The tags are still the same though.

src/JsonApi/Serializer/ItemNormalizer.php
```diff
@@ -222,10 +223,6 @@ protected function normalizeRelation(ApiProperty $propertyMetadata, ?object $rel
         if (null !== $relatedObject) {
             $iri = $this->iriConverter->getIriFromResource($relatedObject);
             $context['iri'] = $iri;
-
-            if (isset($context['resources'])) {
```
soyuka, line 226: as $tagCollector can be null we need to keep this call for backward compatibility

src/HttpCache/TagCollector.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function collect(mixed $object = null, string $format = null, array $context = [], string $iri = null, mixed $data = null, string $attribute = null, ApiProperty $propertyMetadata = null, Type $type = null): void
```
soyuka, line 27: I think we're closed to something nice, a few things still bothers me. Do we really need `data` and `object` ? I think that this works using only `object`. 
If I think about an IriCollector for cache tags, I don't see the format in its signature, we can put this inside our $context no? I also think that `$propertyMetadata` alone is fine as it should hold the `$type` information. So my signature looks like this: 

```
    public function collect(mixed $object = null, string $iri = null, array $context = [], string $attribute = null, ApiProperty $propertyMetadata = null): void
```

Last but not least, I also think that the `iri` can actually be computed by the tag collector if needed, so we'd end up with something like this (`$context['iri']` should be defined for performances if available): 


```
    public function collect(mixed $object = null, array $context = [], string $attribute = null, ApiProperty $propertyMetadata = null): void
```

Not even sure if we couldn't remove `$attribute` and `$propertyMetadata`, are these really useful? 

You could probably add your use case as functional test so that I see how you use it? You can add your service in our `config_common.yml` inside the TestBundle.

src/JsonApi/Serializer/ItemNormalizer.php
usu: Ah, you mean if someone relies on `$context['resources']` to be populated for other means?

Should such code be marked somehow, so they are removed in later major versions? Or is there no intention to remove in the future?

src/HttpCache/TagCollector.php
usu: > You could probably add your use case as functional test so that I see how you use it? You can add your service in our `config_common.yml` inside the TestBundle.

I tried to integrate this into our own project yesterday. That might help a bit to evaluate some of the use cases:
https://github.com/ecamp/ecamp3/pull/3610/files#diff-92df153f2ab4331ccfd55e92699e78c72a3223ff10a38cb870ad59f769308fe0

Can also try to integrate some of this into the test suit.

For your other comments, will look at them and respond later.

usu: I added a functional test with commit https://github.com/api-platform/core/pull/5758/commits/a404dd7fd2b6d05057ecfea1c7b79687182edd0a

Adding the custom TagCollector to `config_common.yml` obviously breaks the existing tests in `features/http_cache/tags.feature`. Not sure what the best way is to isolate this test and couldn't find a good example to copy from. Do I need to set up a separate behat suite which only adds the tag collector service?

src/JsonApi/Serializer/ItemNormalizer.php
soyuka: I'm not sure if it's really needed to enforce the use of the service as it worked for simpler cases without it

src/HttpCache/TagCollector.php
soyuka: https://github.com/api-platform/core/pull/5758#discussion_r1307370466 should resolve the tests issue no?

src/HttpCache/TagCollector.php
soyuka: Please resolve https://github.com/api-platform/core/pull/5758#discussion_r1307370466 so that we don't change any tests

src/Serializer/TagCollectorInterface.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * Interface for collecting cache tags during normalization.
+ *
+ * @author Urban Suppiger <urban@suppiger.net>
+ */
+interface TagCollectorInterface
+{
+    public function collect(mixed $object = null, string $format = null, array $context = [], string $iri = null, mixed $data = null, string $attribute = null, ApiProperty $propertyMetadata = null, Type $type = null): void;
```
soyuka, line 26: I need to find time to improve that interface, your tests helps a lot thanks!

src/JsonApi/Serializer/ItemNormalizer.php
usu: I think I misunderstood the intention first, but I believe I get it know: You propose that I keep the old code (wrapped in `if(!$this->tagCollector)` and in return remove the `TagCollector.php` implementation from this PR?

I can do that. But I don't think it resolves the breaking tests. As soon as I use a custom tag collector, the output is not the same anymore and hence the old tests break. So I need to isolate the service configuration with the custom tag collector from the old tests.

Test were fine before adding the latest commit https://github.com/api-platform/core/pull/5758/commits/a404dd7fd2b6d05057ecfea1c7b79687182edd0a with the customized tag collector.

soyuka: keep these 3 lines if the TagCollector is null as its optional and we don't want to change these. If the tagCollector is defined, then we use it instead.

usu: Done with commit https://github.com/api-platform/core/pull/5758/commits/c935d8d165f34983f7801f06d5dfe29ac02a0250

usu:
I was struggling a bit to ensure Behat is using the custom tag collector service only for a subset of the tests. The best I could come up with is the latest commit: https://github.com/api-platform/core/pull/5758/commits/889a614e4f7ac838bb932c95e034a13827622acd

It seems not that easy to override a service for a specific Behat scenario. It kind of works, when the kernel is not rebooted between scenarios (there's a whole discussion on this under https://github.com/FriendsOfBehat/SymfonyExtension/issues/149).

With this latest commit, it can at least switch between `TagCollectorDefault` and `TagCollectorCustom`. I couldn't figure out how to remove the service completely (manually setting `api_platform.http_cache.tag_collector` to `null` doesn't work).

The caveat of this is that the old tests had to be adjusted slightly. `TagCollectorDefault` replicates the legacy behavior almost identically, but the sequence of the tags is a bit different.

@soyuka Do you mind reviewing this one more time. Do you know of any better method to enable/disable services for specific scenarios?

tests/Behat/HttpCacheContext.php
```diff
@@ -13,17 +13,31 @@
...
+    {
+    }
+
+    /**
+     * @BeforeScenario @customTagCollector
+     */
+    public function registerCustomTagCollector(BeforeScenarioScope $scope): void
     {
+        $this->disableReboot($scope);
+        $this->driverContainer->set('api_platform.http_cache.tag_collector', new TagCollectorCustom());
```
soyuka, line 40: this is quite awesome I wasn't aware we could do that

soyuka:
I like this, if you could:
- rebase onto main
- reword the commit (I changed the title)

Also about this interface, I suggest to reduce it to:

```php
interface TagCollectorInterface
{
    public function collect(string $iri, array $context = []): void;
```

Can `$iri` be actually `null` but we'd like to collect it anyway?

I definitely think that all these can go to our TagCollector context: 
```
mixed $data = null, string $attribute = null, ApiProperty $propertyMetadata = null, Type $type = null
mixed $object = null, string $format = null, 
```

I'm also okay to add an immutable option object if we don't want an array as $context, but all these are noisy in the function signature no?

LMK what you think

usu:
I will give a try to push everything into $context.

Would you also accept the PR against 3.2 instead of main? Would be nice to start using it right away, as it kind of blocks anything around caching on our application side.

usu:
@soyuka I pushed all the parameters into $context, so nothing else needed for the function signature. Let me know if you believe this looks clean. 

I was a bit woried about any side effects at different places, when touching the $context. But tests seem successful.

usu:
hey @soyuka, just wanted to check if you had the chance to look into this?

soyuka:
I was quite busy with the SymfonyCon + related I'm looking at this asap. It'll be merged onto `main` and out by Feb v3.3 as we can't add features to patch versions. Maybe that I can tag an early beta though!

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -789,9 +822,16 @@ protected function normalizeRelation(ApiProperty $propertyMetadata, ?object $rel
             return $normalizedRelatedObject;
         }
 
-        $iri = $this->iriConverter->getIriFromResource($relatedObject);
+        $context['iri'] = $iri = $this->iriConverter->getIriFromResource($relatedObject);
+
```
soyuka, line 826:
```suggestion
```

src/Serializer/TagCollectorInterface.php
```diff
@@ -0,0 +1,24 @@
...
+namespace ApiPlatform\Serializer;
+
+/**
+ * Interface for collecting cache tags during normalization.
+ *
+ * @author Urban Suppiger <urban@suppiger.net>
+ */
+interface TagCollectorInterface
+{
+    public function collect(array $context = []): void;
```
soyuka, line 23: This is quite nice actually! Can you document the context like this: https://github.com/api-platform/core/blob/00246aeb27f15dbc8eec5f41fa75a6cf7e64bce4/src/State/ProcessorInterface.php#L31

tests/Fixtures/TestBundle/HttpCache/TagCollectorCustom.php
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Urban Suppiger <urban@suppiger.net>
+ */
+class TagCollectorCustom implements TagCollectorInterface
+{
+    public const IRI_RELATION_DELIMITER = '#';
+
+    public function collect(array $context = []): void
+    {
+        $iri = $context['iri'];
```
soyuka, line 31: IRI and object can be null here no? (we probably don't care as it's a test fixture) I'd probably add a comment for readers

src/Symfony/Bundle/Resources/config/http_cache_purger.xml
```diff
@@ -24,5 +24,6 @@
             <argument type="service" id="api_platform.iri_converter" />
             <argument type="service" id="api_platform.http_cache.purger" on-invalid="null" />
         </service>
+
```
soyuka, line 27: can you remove this empty line we didn't touched that file

src/Serializer/Tests/AbstractItemNormalizerTest.php
```diff
@@ -90,6 +90,7 @@ public function testSupportNormalizationAndSupportDenormalization(): void
             [],
             null,
             null,
+            null,
```
soyuka, line 93: all the null added argument should be optional we can remove them right?

src/Serializer/Tests/AbstractItemNormalizerTest.php
```diff
@@ -1536,6 +1558,7 @@ public function testDenormalizeObjectWithNullDisabledTypeEnforcement(): void
...
         ]);
         $normalizer->setSerializer($serializerProphecy->reveal());
```
soyuka, line 1563: Just also just skip this file change from the PR

soyuka (changes requested):
Really nice, this will be a quite nice extension point thanks @usu !! some last few changes and please target the main branch :).

src/Serializer/AbstractItemNormalizer.php
usu: Fixed in https://github.com/api-platform/core/pull/5758/commits/828ecb60ffeab01c34c9b6f922b208832d3a9d3d

src/Serializer/TagCollectorInterface.php
usu: Fixed in https://github.com/api-platform/core/pull/5758/commits/828ecb60ffeab01c34c9b6f922b208832d3a9d3d

src/Symfony/Bundle/Resources/config/http_cache_purger.xml
usu: Fixed in https://github.com/api-platform/core/pull/5758/commits/828ecb60ffeab01c34c9b6f922b208832d3a9d3d

src/Serializer/Tests/AbstractItemNormalizerTest.php
usu: Complete file reverted in https://github.com/api-platform/core/pull/5758/commits/828ecb60ffeab01c34c9b6f922b208832d3a9d3d

tests/Fixtures/TestBundle/HttpCache/TagCollectorCustom.php
usu: Yeah, `iri` and `object` could potentially both be null. I think this should be clearer now with the interface signature documentation.

In this specific implementation (`TagCollectorCustom`), I'm checking further below for this (e.g. `isset($iri)`).

usu:
> Really nice, this will be a quite nice extension point thanks @usu !! some last few changes and please target the main branch :).

All implemented as per comments and rebased against main.

MongoDB tests and cs-fixer is failing. But doesn't seem related to any code I touched.

Also, I had to disable the new tests `tag_collector_service.feature` for Symfony lowest. It seemed the method to disable reboot and override services during boot doesn't play nicely with Symfony 6.1.

usu:
> Maybe that I can tag an early beta though!

That would actually be really nice, then we could already start test & integration with our own application. 😃

tests/Fixtures/app/config/config_common.yml
```diff
@@ -451,3 +451,9 @@ services:
     ApiPlatform\Tests\Fixtures\TestBundle\Serializer\ErrorNormalizer:
         decorates: 'api_platform.problem.normalizer.error'
         arguments: [ '@.inner' ]
+        
+    api_platform.http_cache.tag_collector:
+        class: ApiPlatform\Tests\Fixtures\TestBundle\HttpCache\TagCollectorDefault
+        public: true
+    
+
```
soyuka, line 459:
```suggestion
```

soyuka (approved):

soyuka:
@usu would it be possible to document this feature? BTW I released an alpha version today

usu:
> @usu would it be possible to document this feature? BTW I released an alpha version today

Sure. I can add a chapter to https://github.com/api-platform/docs/blob/main/core/performance.md.
Thanks for releasing the alpha 🚀

soyuka:
@usu I released would love some docs if you can

usu:
Just opened a PR at https://github.com/api-platform/core/pull/6348.
Do you think a guide would be sufficient? Or would you like to see this being mentioned in the normal docs as well?

Issue #5735, opened 2023-08-09
SearchFilter and different identifier


dannyvw:
**API Platform version(s) affected**: 3.1.13

**Description**  

This PR (https://github.com/api-platform/core/pull/5623) breaks our test suite because no where query is added to doctrine.

**How to reproduce**  

The classes below should describe the issue. We are using uuids in API request and internal it uses ids.

```php

use ApiPlatform\Metadata\ApiFilter;
use ApiPlatform\Metadata\ApiProperty;
use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Get;
use ApiPlatform\Metadata\GetCollection;
use ApiPlatform\Doctrine\Orm\Filter\SearchFilter;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Uid\Uuid;

#[ApiResource(
    operations: [
        new GetCollection(),
        new Get(),
    ],
)]
#[ApiFilter(SearchFilter::class, properties: ['groups' => 'exact'])]
#[ORM\Entity]
class User
{
    #[ApiProperty(readable: false, writable: false, identifier: false)]
    #[ORM\Id]
    #[ORM\Column(type: 'integer')]
    #[ORM\GeneratedValue(strategy: 'AUTO')]
    private int $id;

    #[ApiProperty(writable: false, identifier: true)]
    #[ORM\Column(type: 'uuid', unique: true)]
    private Uuid $uuid;

    #[ORM\ManyToMany(targetEntity: Group::class, inversedBy: 'users')]
    #[ORM\JoinColumn(name: 'user_id', referencedColumnName: 'id')]
    #[ORM\InverseJoinColumn(name: 'group_id', referencedColumnName: 'id')]
    private Collection $groups;
  
    public function __construct()
    {
        $this->uuid = Uuid::v4();
        $this->groups = new ArrayCollection();
    }
}

---

use ApiPlatform\Metadata\ApiProperty;
use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Get;
use ApiPlatform\Metadata\GetCollection;
use Symfony\Component\Uid\Uuid;

#[ApiResource(
    operations: [
        new GetCollection(),
        new Get(),
    ],
)]
#[ORM\Entity]
class Group
{
    #[ApiProperty(readable: false, writable: false, identifier: false)]
    #[ORM\Id]
    #[ORM\Column(type: 'integer')]
[...]

PR #5760, opened 2023-08-19
fix: searchfilter with nested custom identifiers

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | Closes [5735](https://github.com/api-platform/core/issues/5735)

This undoes the revert of https://github.com/api-platform/core/pull/5623 , but (as far as i can see) fixes the part that was causing https://github.com/api-platform/core/issues/5735

### DIFF
--- a/src/Doctrine/Orm/Filter/SearchFilter.php
+++ b/src/Doctrine/Orm/Filter/SearchFilter.php
@@ -196,6 +196,7 @@ final class SearchFilter extends AbstractFilter implements SearchFilterInterface
         if ($metadata->hasField($field)) {
             if ('id' === $field) {
                 $values = array_map($this->getIdFromValue(...), $values);
+                // todo: handle composite IDs
             }
 
             if (!$this->hasValidValues($values, $this->getDoctrineFieldType($property, $resourceClass))) {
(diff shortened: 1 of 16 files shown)


src/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -216,13 +217,44 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
-        if (!$this->hasValidValues($values, $doctrineTypeField)) {
+        $values = array_map(function ($value) use ($associationFieldIdentifier, $doctrineTypeField) {
+            if (is_numeric($value)) {
+                return $value;
+            }
+            try {
+                $item = $this->getIriConverter()->getResourceFromIri($value, ['fetch_data' => false]);
+
+                return $this->propertyAccessor->getValue($item, $associationFieldIdentifier);
+            } catch (InvalidArgumentException) {
```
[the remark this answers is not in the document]
mrossard: yep, that's not great for perfs but i can't see another way to check whether what i have is a valid IRI beforehand. That should not change much performance-wise though; I actually copied this logic from SearchFilterTrait::getIdFromValue, which was called before : https://github.com/api-platform/core/blob/b58ec1207da8ba5b669937508a17451af61b0688/src/Doctrine/Common/Filter/SearchFilterTrait.php#L121-L132

src/Doctrine/Orm/Filter/SearchFilter.php
```diff
@@ -216,13 +217,44 @@ protected function filterProperty(string $property, $value, QueryBuilder $queryB
...
+                /*
```
mrossard, line 235: I actually have another version that tries to do that, but i'll leave it for another PR since it needs quite a lot more to work.

soyuka: don't worry let's go for this I think it's just find and maybe we can do better when working on #2400 next year!

soyuka:
could you rebase though? Thanks!

mrossard:
> could you rebase though? Thanks!

Done!

PR #5764, opened 2023-08-21
feat(symfony): add mercure asserts

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | **TODO**

### DIFF
--- a/tests/Fixtures/app/config/config_common.yml
+++ b/tests/Fixtures/app/config/config_common.yml
@@ -1,3 +1,9 @@
+parameters:
+    container.autowiring.strict_mode: true
+    .container.dumper.inline_class_loader: true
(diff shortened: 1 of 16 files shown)


tests/Fixtures/app/config/config_common.yml
```diff
@@ -86,10 +92,6 @@ api_platform:
     mercure:
         include_type: true
 
-parameters:
-    container.autowiring.strict_mode: true
-    .container.dumper.inline_class_loader: true
```
soyuka, line 91: these should be changed in the 3.1 branch no?

vincentchalamon: It is just moved (to line 1), not removed

PR #5770, opened 2023-08-23
feat(metadata): improve CreateProvider

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A (`@internal` and `@experimental`)

Improve CreateProvider with `parent_uri_template` context option, and remove the `new Get`. If no `parent_uri_template` option is provided, get the first operation found as it's already done in core.

- [x] Set CreateProvider as `@internal`.
- [x] Inject CallableProvider instead of Doctrine ItemProvider in CreateProvider
- [x] Retrieve parent operation using `parent_uri_template`, or first operation available by default

### DIFF
--- a/src/State/CreateProvider.php
+++ b/src/State/CreateProvider.php
@@ -14,10 +14,10 @@ declare(strict_types=1);
 namespace ApiPlatform\State;
 
 use ApiPlatform\Exception\RuntimeException;
-use ApiPlatform\Metadata\Get;
 use ApiPlatform\Metadata\HttpOperation;
-use ApiPlatform\Metadata\Link;
 use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
+use ApiPlatform\State\Exception\ProviderNotFoundException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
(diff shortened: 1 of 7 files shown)


src/State/CreateProvider.php
```diff
@@ -31,35 +30,34 @@
...
  *
  * @experimental
+ * @internal
  */
 final class CreateProvider implements ProviderInterface
 {
-    public function __construct(private ProviderInterface $decorated, private ?PropertyAccessorInterface $propertyAccessor = null)
-    {
+    public function __construct(
+        private ProviderInterface $provider,
```
soyuka, line 38:
```suggestion
        private ProviderInterface $decorated,
```

src/State/CreateProvider.php
```diff
@@ -31,35 +30,34 @@
...
-
-        foreach ($operationUriVariables as $parameterName => $value) {
-            if ($key === $parameterName) {
-                $relationUriVariables['id'] = new Link(identifiers: $value->getIdentifiers(), fromClass: $value->getFromClass(), parameterName: $key);
-                continue;
-            }
 
-            $relationUriVariables[$parameterName] = $value;
+        try {
+            $parentOperation = $this->resourceMetadataCollectionFactory->create($relationClass)->getOperation($context['parent_uri_template'] ?? null);
```
soyuka, line 56: we should use the `operationMetadataFactory` instead

vincentchalamon: Not sure what you mean here...

soyuka: oh yeah right now I understand better what this does, and no more weird uri variables guess...

soyuka (approved):
Note: we will probably merge this on main only

PR #5778, opened 2023-08-28
Update ItemNormalizer decorator for ElasticSearch

Renrhaf:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| License       | MIT

Fixing a bug where the ElasticSearch decoration of the ItermNormalizer is returning back a wrong value for the `getSupportedTypes` method.  With the current code, and using the "json" format instead of the "ld+json", the ItemNormalizer will not be called properly when using the ElasticSearch layer on top of API Platform.


soyuka:
see 5769

Renrhaf:
Thanks @soyuka, sorry I did not see the already existing PR

PR #5790, opened 2023-08-31
perf: reset local cache to prevent memory leak if long running http server is used

rmikalkenas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -

Recently started using a RoadRunner as a web server and noticed a small memory leak. 
Once started digging, found out that api-platform has several services where array attributes are used as a cache storages. Unfortunately they are never cleared between requests, resulting to not freeing memory.
With this PR I have implemented symfony's services resetter interface, which is called after each handled request, to clear up these arrays.
Targeting main branch as it looks like a new feature although I;m not sure.. Let me know how such change should be prepared and I will correct my PR

P.S. already opened a similar PR in symfony repository: https://github.com/symfony/symfony/pull/51528

### DIFF
--- a/src/Metadata/Property/Factory/CachedPropertyMetadataFactory.php
+++ b/src/Metadata/Property/Factory/CachedPropertyMetadataFactory.php
@@ -16,13 +16,14 @@ namespace ApiPlatform\Metadata\Property\Factory;
 use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\Util\CachedTrait;
 use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * Caches property metadata.
  *
  * @author Teoh Han Hui <teohhanhui@gmail.com>
  */
-final class CachedPropertyMetadataFactory implements PropertyMetadataFactoryInterface
+final class CachedPropertyMetadataFactory implements PropertyMetadataFactoryInterface, ResetInterface
 {
     use CachedTrait;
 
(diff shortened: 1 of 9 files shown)


dunglas:
I wonder if it's a good idea to do that. This cache is especially useful when using a long-running server such as RoadRunner or FrankenPHP in worker mode: on e populated, it will be reused for subsequent HTTP requests.

I'm curious of why the memory leak. Once all metadata are in cache, no new memory should be allocated.

Maybe could we provide a method to manually clear the cache for advanced use cases (dynamic resource types for instance), but IMHO we shouldn't clear it by default.

rmikalkenas:
> I wonder if it's a good idea to do that. This cache is especially useful when using a long-running server such as RoadRunner or FrankenPHP in worker mode: on e populated, it will be reused for subsequent HTTP requests.
> 
> I'm curious of why the memory leaks. Once all metadata are in cache, no new memory should be allocated.
> 
> Maybe could we provide a method to manually clear the cache for advanced use cases (dynamic resource types for instance), but IMHO we shouldn't clear it by default.

Hi, @dunglas , thanks for a reply. I get the idea behind it, but in my opinion there should be a balance between performance and resources usage (memory in this example).
In a small applications such approach is completely valid and slightly increased memory wont cause any issues. But let's take  large applications with hundreds of API resources - then such implementation might become painful from a memory usage perspective.
Knowing that both FrankenPHP and RoadRunner use multiple worker approach - personally, I would go with shared cache so that each worker wont have it's own (as it is now with localCache). In fact it's already there, all `Cached*` services use `CacheItemPoolInterface` which can be any shared storage.

One idea pops up in my head: maybe implement a new configuration by which you could control whether to use `localCache` or skip it and depend only on PSR-6 :thinking: this would allow you to control performance vs resource usage.

Let me know what you think :v:

soyuka:
if we really want to get rid of this we should enable them on their own metadata factory and add a configuration to disable them, anyways I agree with @dunglas not sure their impact requires such a change... How much are we talking about?

dunglas:
I don't think that a remote cache will be interesting here. Remote cache are order of magnitude slower than local in memory caches. Fetching the data from the remote cache server will probably be slower (or as fast as) recomputing the value.

rmikalkenas:
Closing as resolved differently

PR #5791, opened 2023-09-01
fix: strip down any sql function name

botjaeger:
referencing it here https://github.com/api-platform/api-platform/issues/2507


soyuka:
needs test and should target 3.1

botjaeger:
> needs test and should target 3.1

If I understand it correctly, I updated the PR to point it into branch `3.1`

mrossard:
There may be a need to check where else this can happen - I think i remember having the same issue before with a join condition; i'm not sure it's the same part of the code getting called then.

Anything that tries to rewrite the DQL should be looked at I guess.

botjaeger:
https://github.com/api-platform/core/pull/5794

PR #5794, opened 2023-09-04
Extracting table alias includes any SQL function names

botjaeger:
https://github.com/api-platform/api-platform/issues/2507

### DIFF
--- a/src/Doctrine/Orm/Util/QueryChecker.php
+++ b/src/Doctrine/Orm/Util/QueryChecker.php
@@ -106,7 +106,7 @@ final class QueryChecker
             foreach ($orderBy->getParts() as $part) {
                 if (str_contains((string) $part, '.')) {
                     [$alias] = explode('.', (string) $part);
-                    $alias = str_contains($alias, '(') ? substr($alias, strpos($alias, '(') + 1) : $alias;
+                    $alias = substr($alias, -1);
 
                     $orderByAliases[] = $alias;
                 }


soyuka:
I'd like a non-regression test this looks quite specific.

botjaeger:
The solution doesn't work when the query is something like this

```
->addOrderBy("CASE
                 WHEN $alias.stage = 'NEW_APPLICATION' THEN 0
                 WHEN $alias.stage = 'ARCHIVED' THEN 1
                 WHEN $alias.stage = 'SEND_TO_CLIENT' THEN 2
                 WHEN $alias.stage = 'CLIENT_APPROVED' THEN 3
                 WHEN $alias.stage = 'SCREEN_CANDIDATE' THEN 4
                 WHEN $alias.stage = 'EXAM' THEN 5
                 WHEN $alias.stage = 'INTERVIEW' THEN 6
                 WHEN $alias.stage = 'SEND_OFFER' THEN 7
                 WHEN $alias.stage = 'SIGN_CONTRACT' THEN 8
             END", 'DESC');
```

this will result to 

`The alias \"CASE\n                 WHEN o\" does not exist in the QueryBuilder.`

but I can use the `FIELD` function

botjaeger:
I updated the PR.

This is assuming that all doctrine alias is single character only

mrossard:
> This is assuming that all doctrine alias is single character only

This doesn't seem realistic at all, for example if you use ApiPlatform\Doctrine\Orm\Util\QueryNameGenerator::generateJoinAlias() to get your aliases they won't be 1 character.

botjaeger:
I agree. will close this

PR #5798, opened 2023-09-05
feat: use more descriptive exception messages on IriConverter (#5797)

### LINKED ISSUE #5797: More precise exception message on iri generation
**Description**  
When relations of an API Resource are normalized and API Platform tries to generate an IRI for one of the relations, the exception message is not clear whats going wrong.

**Example**  
*Order* has relation to *customer*, and *customer*'s iri could not be generated due a wrong configuration etc., the exception thrown is: 
> "Unable to generate an IRI for the item of type "App\Order".

Which is really confusing. It should be something like:
> "Unable to generate an IRI for the operation class "App\Order" and the item of type "App\Customer".

kswzr:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #5797 
| License       | MIT
| Doc PR        | none

More descriptive exception messages on IriConverter, to easier find the right place, where you need to check the API Platform definition.


soyuka:
needs cs fixing + phpstan issue

Issue #5793, opened 2023-09-02
API platform 3.2 feedbacks


soyuka:
Use this issue to send feedbacks on the next-gen API Platforms version. 
Thanks for the support!

paullallier:
When I run `composer require api-platform/core v3.2.0-alpha.1 -W`. I get the following:

```
Class ApiPlatform\Tests\Util\ArrayTraitTest located in ./vendor/api-platform/core/src/GraphQl/Tests/Util/ArrayTraitTest.php does not comply with psr-4 autoloading standard. Skipping.
```

I think it should be `namespace ApiPlatform\GraphQl\Tests\Util;` rather than `namespace ApiPlatform\Tests\Util;` (I don't think there's a branch I can PR that to?)

And later, more critically:
```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!  
!!  In DecoratorServicePass.php line 89:
!!                                                                                 
!!    The service "api_platform.graphql.state_provider.validate" has a dependency  
!!     on a non-existent service "api_platform.graphql.state_provider".            
!!                                                                                 
!!  
!!  
Script @auto-scripts was called via post-update-cmd
```

I might be missing a config file change, but I didn't see anything mentioned in the notes.  I'm not intentionally using GraphQL in my project.

soyuka:
Thanks @paullallier I just fixed these on `dev-main` and will tag in a few days.

paullallier:
I grabbed dev-main to have another quick look.  I'm seeing an issue with the Swagger docs, but that might be something specific to the (probably) wrong why I'm defining my entity relations.

Specifically, it doesn't seem happy if the parameter defaults to null (or possibly just if it's nullable)
```
    #[ORM\Column(length: 128, nullable: true)]
    #[Groups(['generic_financial_entity:read_item', 'generic_financial_entity:read_collection', 'generic_financial_entity:write_update'])]
    protected ?string $description = null;
```

In dev-main (f02f379):
<img width="436" alt="Screenshot 2023-09-05 at 14 05 11" src="https://github.com/api-platform/core/assets/42591123/b9ec93d9-73a7-42fa-ba23-6f2b8e2cfa15">

In v3.1.16:
<img width="493" alt="Screenshot 2023-09-05 at 14 05 36" src="https://github.com/api-platform/core/assets/42591123/f5f22183-d6d4-4ef7-b5cc-5941592607f4">


[...]

PR #5810, opened 2023-09-11
fix(jsonschema): allow embed resources

soyuka:
#5793 related ping @paullallier @vincentchalamon

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -126,7 +126,7 @@ final class SchemaFactory implements SchemaFactoryInterface
             $definition['externalDocs'] = ['url' => $operation->getTypes()[0]];
         }
 
-        $options = $this->getFactoryOptions($serializerContext, $validationGroups, $operation instanceof HttpOperation ? $operation : null);
+        $options = ['schema_type' => $type] + $this->getFactoryOptions($serializerContext, $validationGroups, $operation instanceof HttpOperation ? $operation : null);
         foreach ($this->propertyNameCollectionFactory->create($inputOrOutputClass, $options) as $propertyName) {
             $propertyMetadata = $this->propertyMetadataFactory->create($inputOrOutputClass, $propertyName, $options);
             if (!$propertyMetadata->isReadable() && !$propertyMetadata->isWritable()) {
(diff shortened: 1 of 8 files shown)


paullallier:
That looks a lots better! A couple things I still see:

There's something odd about the reference back from TestEntity --> BagOfTests in my example. I don't ever user this in my API, so it might be normal - not sure.
Screenshot 2023-09-12 at 10 44 44

Also, I haven't defined the return structure on errors.  In alpha2, it didn't try to display them but in this branch it is displaying as something undefined.  This might be a feature of the updated SwaggerUI, and it might not be a bad things, since I probably should defined the error return... Though perhaps I shouldn't need to do this explicitly?

In my dev environment, I have the API configured to display on https://main-api (it also displays on https://main-api/docs).  In alpha2, it works.  In this branch, I get raw json (LD, I think) on https://main-api (the /docs version give the expect html).

paullallier:
Oh, and I think that was with e03bb31

src/JsonSchema/SchemaFactory.php
```diff
@@ -194,8 +201,18 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
-            $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
-            // prevent "type" and "anyOf" conflict
+            if ($isCollection) {
+                $propertySchema['items']['$ref'] = $subSchema['$ref'];
+                unset($propertySchema['items']['type']);
+                break;
+            }
+
+            if ($type->isNullable()) {
+                $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
```
vincentchalamon, line 211: IMHO `oneOf` should be more persistent in this case: https://json-schema.org/understanding-json-schema/reference/combining.html#id7
```suggestion
                $propertySchema['oneOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
```

src/JsonSchema/SchemaFactory.php
```diff
@@ -194,8 +201,18 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
+            } else {
+                $propertySchema['oneOf'] = [['$ref' => $subSchema['$ref']]];
```
vincentchalamon, line 213: Useless `oneOf` in this case, `$ref` is enough (cf. https://json-schema.org/understanding-json-schema/structuring.html#ref)

src/JsonSchema/SchemaFactory.php
```diff
@@ -176,13 +182,14 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
 
         // property schema is created in SchemaPropertyMetadataFactory, but it cannot build resource reference ($ref)
         // complete property schema with resource reference ($ref) only if it's related to an object
-
         $version = $schema->getVersion();
         $subSchema = new Schema($version);
         $subSchema->setDefinitions($schema->getDefinitions()); // Populate definitions of the main schema
+        $isCollection = false;
```
vincentchalamon, line 188: Useless variable as it's only used inside the foreach loop

src/JsonSchema/SchemaFactory.php
```diff
@@ -194,8 +201,18 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
             }
 
             $subSchema = $this->buildSchema($className, $format, Schema::TYPE_OUTPUT, null, $subSchema, $serializerContext + [self::FORCE_SUBSCHEMA => true], false);
-            $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
-            // prevent "type" and "anyOf" conflict
+            if ($isCollection) {
```
vincentchalamon, line 204: Useless `$isCollection` var as it's only used inside this foreach loop.
Also, if the first foreach loop enables the `$isCollection` variable but the same loop exists on line 200, the next `$type` will be considered as a collection even if it isn't.
```suggestion
            if ($type->isCollection()) {
```

vincentchalamon (changes requested):

src/JsonSchema/SchemaFactory.php
soyuka: you can probably have null values in a collection with objects?

vincentchalamon: Like `['/books/{id}, null]`? Not sure this is a valid value for us, and is what we want...

PR #5820, opened 2023-09-13
feat(elasticsearch): filtering on nested fields

jonnyeom:
This is a resurrection of https://github.com/api-platform/core/pull/5643

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/api-platform/issues/1375 and https://github.com/api-platform/core/issues/5642
| License       | MIT
| Doc PR        | WIP

This PR implements the following changes to support filtering on paths like foo.bar.baz:

1. If `foo` is an object, `isNestedField()` now returns `true` instead of `false`
2. If `foo` is an object, `getNestedFieldPath()` now returns `"foo.bar"` instead of `null`
3. If `foo` is a collection, `getNestedFieldPath()` now returns `"foo.bar"` instead of just `"foo"` (matching the behavior if `foo` is an object)

As a result, filtering on deeply-nested paths (with 3 or more levels) now produces the following query:

```diff
 {
     "nested": {
-        "path": "foo",
+        "path": "foo.bar",
         "query": {
             "term": {"foo.bar.baz": "qux"}
            }
         }
     }
 }
```

### DIFF
--- a/src/Elasticsearch/Util/FieldDatatypeTrait.php
+++ b/src/Elasticsearch/Util/FieldDatatypeTrait.php
@@ -78,7 +78,9 @@ trait FieldDatatypeTrait
                 && null !== ($className = $type->getClassName())
                 && $this->resourceClassResolver->isResourceClass($className)
             ) {
-                return $currentProperty;
+                $nestedPath = $this->getNestedFieldPath($className, implode('.', $properties));
+
+                return null === $nestedPath ? $currentProperty : "$currentProperty.$nestedPath";
             }
         }
 
--- a/src/Elasticsearch/Util/FieldDatatypeTrait.php
+++ b/src/Elasticsearch/Util/FieldDatatypeTrait.php
@@ -69,7 +69,7 @@ trait FieldDatatypeTrait
             ) {
                 $nestedPath = $this->getNestedFieldPath($nextResourceClass, implode('.', $properties));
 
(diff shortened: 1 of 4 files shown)


soyuka (approved):

soyuka:
Thanks @jonnyeom

soyuka:
I think that this actually breaks elasticsearch test suite my bad.

src/Elasticsearch/Util/FieldDatatypeTrait.php
```diff
@@ -69,7 +69,7 @@ private function getNestedFieldPath(string $resourceClass, string $property): ?s
             ) {
                 $nestedPath = $this->getNestedFieldPath($nextResourceClass, implode('.', $properties));
 
-                return null === $nestedPath ? $nestedPath : "$currentProperty.$nestedPath";
+                return null === $nestedPath ? $currentProperty : "$currentProperty.$nestedPath";
```
soyuka, line 72: @jonnyeom I have to revert this to make https://github.com/api-platform/core/blob/main/features/elasticsearch/match_filter.feature#L120 green any idea? more at https://github.com/api-platform/core/pull/5823

jonnyeom: interesting.
Let me take a deeper look again.

Thanks for heads up!

PR #5835, opened 2023-09-18
feat(elasticsearch): filtering on nested fields

jonnyeom:
Lets try again with added tests!

A better version of [5820](https://github.com/api-platform/core/pull/5820), which was reverted

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | https://github.com/api-platform/api-platform/issues/1375 and https://github.com/api-platform/core/issues/5642
| License       | MIT
| Doc PR        | WIP

It is inline with elasticsearch's usage of nested documents
> The nested type is a specialized version of the object data type that allows arrays of objects to be indexed

see [elasticsearch nested docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/nested.html#nested)

In the case of foo.bar.baz
1. if `bar` is an array of objects, we return `foo.bar`.
2.  if `bar` is not an array, it does not qualify to be a nested property (according to elasticsearch) and we return null.

This allows Filter and Match queries such as 

**GET /object-array-object-string/_search**
```json
{
  "query": {
    "nested": {
      "path": "object.array",
      "query": {
        "bool": {
          "must": [
            { "match": { "object.array.object.property": "search-value" } }
          ]
        }
      }
    }
  }
}
```

### DIFF
--- a/features/elasticsearch/match_filter.feature
+++ b/features/elasticsearch/match_filter.feature
@@ -429,3 +429,54 @@ Feature: Match filter on collections from Elasticsearch
       }
     }
     """
+
+  Scenario: Match filter on a multi-level nested property of text type with new elasticsearch operations
+    When I send a "GET" request to "/books?library.relatedGenres.name=Fiction"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
+    And the JSON should be valid according to this schema:
+    """
+    {
+      "type": "object",
+      "properties": {
+        "@context": {"pattern": "^/contexts/Book$"},
+        "@id": {"pattern": "^/books$"},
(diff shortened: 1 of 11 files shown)


jonnyeom:
> can you fix phpunit tests on elasticsearch? to run them:
> 
> ```
> cd src/Elasticsearch
> composer update
> ./vendor/bin/simple-phpunit
> ```

Yep. currently rethinking the approach.

The resulting nested match query seems to be a bit different from what [elasticsearch has document](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/query-dsl-nested-query.html#multi-level-nested-query-ex)

It still works, im not sure if it is a hidden feature or a bug. So will confirm with them and proceed accordingly.

soyuka (approved):

Renrhaf:
Great ! Thank you @jonnyeom

PR #5846, opened 2023-09-21
docs(subresource): Parser is losing the closing structure on new line

xavierleune:
First of all, thanks @soyuka for your amazing work with the wasm integration and the new guide :heart_eyes: 

It seems that the parser is losing the closing of the string. I guess it's because on a new line, so I hope this will fix it !

image


Romaixn (approved):

soyuka:
thanks @xavierleune !

PR #5848, opened 2023-09-22
exclude error ressources from graphql

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | References #4994
| License       | MIT
| Doc PR        | none

I noticed that PR https://github.com/api-platform/core/pull/5433 introduced side effects on the GraphQL side which lead to an invalid schema.
`{ "errors": [ { "message": "Names must only contain [_a-zA-Z0-9] but \"hydra:title\" does not.", "extensions": {}, "stack": "GraphQLError: Names must only contain [_a-zA-Z0-9] but \"hydra:title\" does not." } ] }`

This PR just excludes the Error resources for graphql
see https://github.com/api-platform/core/issues/4994#issuecomment-1648363615


soyuka:
can't we introduce a graphql error operation instead?

soyuka:
I'd like a little bit more informations about this, also a test covering the issue would be nice not sure if that's the best option

jotwea:
At the moment the Graphql API is broken - the schema is invalid. This PR is meant to be a simple fix that the schema will be valid again.

I do not know much about this error resource and cannot judge if this is relevant for GraphQL and how to implement this properly. Feel free to work on that solution. But still I would vote to merge this PR before as a quick fix.

soyuka:
we need a regression test and to fix serialization groups indeed! I'll look into that.

soyuka:
We investigated and I now understand better what's going on... I'm trying to find a way to test this.

jotwea:
Thx a lot

soyuka:
Thanks @jotwea !

PR #5850, opened 2023-09-25
fix(FilterContext): add xxx[] hydra:search iexact

CorentinFackeure:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       |  [5851](https://github.com/api-platform/core/issues/5851)
| License       | MIT

### DIFF
--- a/src/Doctrine/Common/Filter/SearchFilterInterface.php
+++ b/src/Doctrine/Common/Filter/SearchFilterInterface.php
@@ -37,9 +37,19 @@ interface SearchFilterInterface
     public const STRATEGY_PARTIAL = 'partial';
 
     /**
+     * @var string The value must be contained in the field case-insensitive
+     */
+    public const STRATEGY_IPARTIAL = 'partial';
+
+    /**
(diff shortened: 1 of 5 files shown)


dunglas (approved):
LGTM, but it would be even better with a non-regression test 😃

Thanks for the fix and for your commitment.

vincentchalamon:
Thanks @CorentinFackeure!

src/Doctrine/Common/Filter/SearchFilterInterface.php
```diff
@@ -26,23 +26,48 @@ interface SearchFilterInterface
...
+
     /**
      * @var string The value must be contained in the field
      */
     public const STRATEGY_PARTIAL = 'partial';
 
+    /**
+     * @var string The value must be contained in the field case-insensitive
+     */
+    public const STRATEGY_IPARTIAL = 'partial';
```
bendavies, line 42: shouldn't thisd be `ipartial`?

PR #5853, opened 2023-09-25
feat: improve ApiProperty::security using property name

### LINKED ISSUE #5824: Improve ApiProperty::security using property name
### Description

Currently, [`AbstractItemNormalizer::canAccessAttribute`](https://github.com/api-platform/core/blob/main/src/Serializer/AbstractItemNormalizer.php#L434-L450) controls the access to a property thanks to [`ApiProperty::security`](https://github.com/api-platform/core/blob/main/src/Metadata/ApiProperty.php#L106), thanks to `$resourceAccessChecker->isGranted` with the object.

However, the property name is never injected in the expression language...

### Use Case

I have a matrix of permissions where each role have a read or write access to each property:
> a `collaborator` has a read-only access to `App\Resource\Dummy::foo` property
> a `manager` has a read-and-write access to `App\Resource\Dummy::bar` property

### Proposition

Send `$attribute` as `property` variable in `$extraVariables` while calling `$resourceAccessChecker->isGranted`.

### Example

```php
#[ApiProperty(
    security: 'is_granted("ROLE_COLLABORATOR", property)'
)]
```

```php
// src/Security/CollaboratorVoter.php

final class CollaboratorVoter extends Voter
{
    // ...

    /**
     * @param string $subject the property such as "foo"
     */
    protected function voteOnAttribute(string $attribute, mixed $subject, TokenInterface $token): bool
    {
        // check if user has access to property ($subject as string)
    }
}
```

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 3.3
| Tickets       | Closes #5824
| License       | MIT
| Doc PR        | Inlined

### DIFF
--- a/features/authorization/deny.feature
+++ b/features/authorization/deny.feature
@@ -78,7 +78,7 @@ Feature: Authorization checking
     And I send a "GET" request to "/secured_dummies/2"
     Then the response status code should be 200
 
-  Scenario: A user can see a secured owner-only property on an object they own
+  Scenario: A user can see a secured owner-only property, or accessible property based on voter, on an object they own
     When I add "Accept" header equal to "application/ld+json"
     And I add "Content-Type" header equal to "application/ld+json"
     And I add "Authorization" header equal to "Basic ZHVuZ2xhczprZXZpbg=="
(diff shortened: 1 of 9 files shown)


soyuka (approved):

vincentchalamon:
@soyuka is it ok to merge it?

PR #5862, opened 2023-10-03
fix: only generate jsonld docs when docs disabled

### LINKED ISSUE #5528: AddLinkHeaderListener always adds a link to docs

Romaixn:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | Closes #5528 
| License       | MIT

Only generate jsonld docs when docs disabled to avoid routing error when request to /contexts.

### DIFF
--- a/src/Symfony/Bundle/Resources/config/routing/docs_jsonld.xml
+++ b/src/Symfony/Bundle/Resources/config/routing/docs_jsonld.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://symfony.com/schema/routing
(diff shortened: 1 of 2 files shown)


dunglas:
I wonder if we cannot do that in a cleaner way: for instance, we could set `docs_formats` in the DI Extension to `jsonld` if `enabled_docs` is false, so we can simplify the routing part and always register the route, as it will work only for JSON-LD.

Maybe can we do even better now that we have this `docs_formats` option and deprecate the `enable_docs` config option, as new system is more powerful and cover this old use case.

WDYT?

Issue #5845, opened 2023-09-21
API Platform 3.2


soyuka:
Please report 3.2 issues here: 

```
composer config minimum-stability beta
composer update api-platform/core
composer recipes:update 
```

divine:
Hello!

I'm currently migrating one of the 2.6 projects.

JFYI the migration process is from scratch (install the new 3.2 version and migrate the files one by one I know it might take a longer time but would like to make sure that I'm using the latest features).

1) Serializer error:
Blank APP install visiting `https://localhost` throws an error in profiler:

```php
Uncaught PHP Exception Symfony\Component\Serializer\Exception\UnsupportedFormatException: "Serialization for the format "html" is not supported." at /app/vendor/symfony/serializer/Serializer.php line 134
```

Visiting `https://localhost/docs/` doesn't throw anything.

This is the line problematic line ([source](
 https://github.com/api-platform/api-platform/blob/c7a20ed11b007a81233d6480baa5c81f9a8f7510/api/config/packages/api_platform.yaml#L11)):
```yaml
    docs_formats:
        jsonld: ['application/ld+json']
        jsonopenapi: ['application/vnd.openapi+json']
        html: ['text/html']
```


2) Docker vendor folder isn't copied to container:

Looking at `docker-compose.override.yml` [file](https://github.com/api-platform/api-platform/blob/c7a20ed11b007a81233d6480baa5c81f9a8f7510/docker-compose.override.yml#L16) and commenting out the line:
```yaml
      # If you develop on Mac or Windows you can remove the vendor/ directory
      #  from the bind-mount for better performance by enabling the next line:
      - /app/vendor
```

Doesn't start the container, throwing `frankenphp` error:

```
 Fatal error: Uncaught LogicException: Symfony Runtime is missing. Try running "composer require symfony/runtime". in /app/bin/console:8
| Stack trace:
| #0 {main}
|   thrown in /app/bin/console on line 8
```


That's all for now, continuing the migration and I'll let you any issues I'm having.

Thanks!

vincentchalamon:
@divine I can't reproduce your bug.

Starting from a fresh install of api-platform/api-platform, then starting the project thanks to `docker compose up --build --wait`, going to https://localhost, I don't see any error...

Can you try to clean your local environment just in case you're using an old version of containers?

paullallier:
[...]

PR #5864, opened 2023-10-04
fix(jsonschema): do not override nor complete ApiProperty::schema user value

vincentchalamon:
See https://github.com/api-platform/core/issues/5845#issuecomment-1740478420


vincentchalamon:
Ping @paullallier

paullallier:
Thanks @vincentchalamon.  I think this is right now.  I've been trying to understand a few remaining test failures, but I don't think they are related to this - and this might allow me to mitigate them.

PR #5872, opened 2023-10-06
Add itemUriTemplate to resources.xsd

Wojdylak:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | N/A
| License       | MIT

### DIFF
--- a/src/Metadata/Extractor/schema/resources.xsd
+++ b/src/Metadata/Extractor/schema/resources.xsd
@@ -47,6 +47,7 @@
         <xsd:attribute type="xsd:boolean" name="openapi"/>
         <xsd:attribute type="xsd:positiveInteger" name="priority"/>
         <xsd:attribute type="xsd:string" name="name"/>
+        <xsd:attribute type="xsd:string" name="itemUriTemplate"/>
     </xsd:complexType>
 
     <xsd:complexType name="graphQlOperations">


src/Metadata/Extractor/schema/resources.xsd
```diff
@@ -47,6 +47,7 @@
         <xsd:attribute type="xsd:boolean" name="openapi"/>
         <xsd:attribute type="xsd:positiveInteger" name="priority"/>
         <xsd:attribute type="xsd:string" name="name"/>
+        <xsd:attribute type="xsd:string" name="itemUriTemplate"/>
```
soyuka, line 50: as this is not declared on every operation I'm not sure

Wojdylak: But I got an error when I used `itemUriTemplete`. `Element '{https://api-platform.com/schema/metadata/resources-3.0}operation', attribute 'itemUriTemplate': The attribute 'itemUriTemplate' is not allowed.`

soyuka:
let's merge that for now, I'd like @vincentchalamon insights on this

src/Metadata/Extractor/schema/resources.xsd
vincentchalamon: `itemUriTemplate` is only available on GetCollection and Post operations. But I'm not sure it's possible to handle such difference with XML schema...

A solution would be to allow this attribute on XML and YAML schemas, and add a specific validation on XmlResourceExtractor and YamlResourceExtractor services to ensure this attribute is only set on a GetCollection or Post operation, otherwise throw an error. The XmlExtractorTest and YamlExtractorTest should help to test this behavior.

vincentchalamon: Note: apparently, I already did a similar check in resource extractors :smiley: 
https://github.com/api-platform/core/blob/3.1/src/Metadata/Extractor/XmlResourceExtractor.php#L387-L389
https://github.com/api-platform/core/blob/3.1/src/Metadata/Extractor/YamlResourceExtractor.php#L323-L325

Throwing an exception if the operation does not match could improve the DX instead of just ignoring the key (PR in progress).

PR #5874, opened 2023-10-06
fix(validation): error detail name converted properties

soyuka:
@dannyvw can you try this ?


dannyvw:
After applying this changes I get a lot of `Not Found (Symfony\Component\HttpKernel\Exception\NotFoundHttpException)` errors on different routes.

PR #5890, opened 2023-10-17
4798 validate parameters with multiple values

aegypius:
- fix(validation)!: validate parameters with multiple values
- fixup! fix(validation)!: validate parameters with multiple values


src/Api/QueryParameterValidator/QueryParameterValidator.php
```diff
@@ -72,4 +79,77 @@ public function validateFilters(string $resourceClass, array $resourceFilters, a
...
+    private static function getSeparator(string $collectionFormat): string
+    {
+        return match ($collectionFormat) {
+            'csv' => ',',
+            'ssv' => ' ',
+            'tsv' => '\t',
+            'pipes' => '|',
+            default => throw new \InvalidArgumentException(\sprintf('Unknown collection format %s', $collectionFormat)),
+        };
+    }
```
aegypius, line 140: Same as above, borrowed from `Api\QueryParameterValidator\Validator\ArrayItems`

src/Api/QueryParameterValidator/QueryParameterValidator.php
```diff
@@ -72,4 +79,77 @@ public function validateFilters(string $resourceClass, array $resourceFilters, a
...
+        if (\is_array($value)) {
+            return $value;
+        }
+
+        if (\is_string($value)) {
+            return \explode(self::getSeparator($collectionFormat), $value);
+        }
+
+        return [$value];
+    }
```
aegypius, line 128: Shamelessly borrowed from `Api\QueryParameterValidator\Validator\ArrayItems`

src/Api/QueryParameterValidator/QueryParameterValidator.php
```diff
@@ -60,8 +61,14 @@ public function validateFilters(string $resourceClass, array $resourceFilters, a
             }
 
             foreach ($filter->getDescription($resourceClass) as $name => $data) {
-                foreach ($this->validators as $validator) {
-                    if ($errors = $validator->validate($name, $data, $queryParameters)) {
+                $collectionFormat = $this->getCollectionFormat($data);
+
+                // validate simple values
+                $errorList[] = $this->validate($name, $data, $queryParameters);
```
aegypius, line 67: First validation is expected to allow to validate missing requirements. Non scalar parameters may cause the `Array to string conversion`

src/Api/QueryParameterValidator/QueryParameterValidator.php
```diff
@@ -60,8 +61,14 @@ public function validateFilters(string $resourceClass, array $resourceFilters, a
...
+
+                // manipulate query data to validate each value
+                foreach ($this->iterateValue($name, $queryParameters, $collectionFormat) as $scalarQueryParameters) {
```
aegypius, line 70: We iterate over each splitted value and run all validators against the singular value

src/Api/QueryParameterValidator/QueryParameterValidator.php
```diff
@@ -72,4 +80,76 @@ public function validateFilters(string $resourceClass, array $resourceFilters, a
...
+    {
+        $errorList = [];
+
+        foreach ($this->validators as $validator) {
+            if ($errors = $validator->validate($name, $data, $queryParameters)) {
+                $errorList[] = $errors;
+            }
+        }
+
+        return array_merge(...$errorList);
```
soyuka, line 153: why the array_merge?

src/Api/QueryParameterValidator/QueryParameterValidator.php
```diff
@@ -72,4 +80,76 @@ public function validateFilters(string $resourceClass, array $resourceFilters, a
...
+            static fn (string $key) => $key === $name || "{$key}[]" === $name,
+            \ARRAY_FILTER_USE_KEY
+        );
+
+        foreach ($candidates as $key => $value) {
+            $values = self::getValue($value, $collectionFormat);
+            foreach ($values as $v) {
+                yield [$key => $v];
+            }
+        }
```
soyuka, line 110: can we do the array_filter within the loop to save some iterations

src/Api/QueryParameterValidator/QueryParameterValidator.php
aegypius: It was an attempt to reduce the nesting of array leading to the previous `Array to string conversion`. It fixes unit tests but seems to break elsewhere. 

I started to add more tests but it's not yet completed (hence the "draft" state)

src/Api/QueryParameterValidator/QueryParameterValidator.php
aegypius: You're right an `if ...` statement should do the trick

aegypius:
this PR must be rebased on the new api-platform/parameter-validator component (see https://github.com/api-platform/core/pull/6080)

src/Api/QueryParameterValidator/QueryParameterValidator.php
aegypius: @soyuka most of the tests are passing now, I need to find a way to avoid thoses `array_merge(...$var)` used to flatten errors lists. I think `iterable` might be the solution. WDT ?

VincentLanglet:
Hi, are you still interested by this PR ? If so, could you rebase the branch and solve the conflict ? What else is missing ?
Thanks

aegypius:
@VincentLanglet The validation has significantly change since then. The whole ParameterValidator component is gone, and most of the cases should be resolved now with the uses of Constraints on Parameters.

soyuka:
This actually helped designing complex use cases with parameters, thanks @aegypius !

PR #5897, opened 2023-10-18
 fix(jsonschema): restore type factory usage 

### LINKED ISSUE #5896: Decorating `api_platform.json_schema.type_factory` does not work in 3.2
**API Platform version(s) affected**: 3.2.1

**Description**  
In 3.1 decorating the service `api_platform.json_schema.type_factory` works for customising the json schema type for a custom userland class.

**How to reproduce**  
Here, I decorate the type factory to add support for using [Brick\DateTime\LocalDate](https://github.com/brick/date-time) as a `date`
```php

<?php

declare(strict_types=1);

namespace App\ApiResource;

use ApiPlatform\Metadata\ApiProperty;
use ApiPlatform\Metadata\Get;
use Brick\DateTime\LocalDate;

#[ApiResource(
    operations: [
        new Get(),
    ],
)]
class Foo
{
    #[ApiProperty(readable: false, writable: false, identifier: true)]
    public ?int $id = null;
    public ?LocalDate $expiration;
}

```

```php
<?php

declare(strict_types=1);

namespace App\ApiPlatform;

use ApiPlatform\JsonSchema\Schema;
use ApiPlatform\JsonSchema\TypeFactoryInterface;
use Brick\DateTime\LocalDate;
use Symfony\Component\DependencyInjection\Attribute\AsDecorator;

#[AsDecorator('api_platform.json_schema.type_factory')]
class TypeFactoryDecorator implements TypeFactoryInterface
{
    public function __construct(
        private readonly TypeFactoryInterface $decorated,
    ) {
    }

    /** @phpstan-ignore-next-line  */
    public function getType(Type $type, string $format = 'json', ?bool $readableLink = null, ?array $serializerContext = null, ?Schema $schema = null): array
    {
        /** @phpstan-ignore-next-line  */
        if (\is_a($type->getClassName(), LocalDate::class, true)) {
            return [
                'type' => 'string',
                'format' => 'date',
            ];
        }

        return $this->decorated->getType($type, $format, $readableLink, $serializerContext, $schema);
    }
}

```

in 3.1, the schemas are correct:

image
 
in 3.2, they are not:
image

**Possible Solution**  
Unsure, tried to figure it out but failed.

**Additional Context**

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | fixes #5896
| License       | MIT

We wrongly assumed in #5470 that the TypeFactory would be `@internal` which it is not. We moved the logic related to type computation to our Schema PropertyMetadataFactory so that the schema could get cached. 

This patch restores the TypeFactory behavior.

I think that we should postpone the deprecation of the TypeFactory to 3.3 and provide a way to configure this on non-resources (see #5896) for example:

```php
#[JsonSchema(['type' => 'string', 'format' => 'date'])
class LocalDate {}
```

Overriding the property metadata factory is also an option but requires more code user land.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -26,7 +26,7 @@ parameters:
 		- src/Symfony/Bundle/DependencyInjection/Configuration.php
 		# Templates for Maker
 		- src/Symfony/Maker/Resources/skeleton
-		- src/*/vendor
+		- **vendor**
 	earlyTerminatingMethodCalls:
 		PHPUnit\Framework\Constraint\Constraint:
 			- fail
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -37,7 +37,7 @@ final class SchemaFactory implements SchemaFactoryInterface
 {
     use ResourceClassInfoTrait;
     private array $distinctFormats = [];
-
+    private ?TypeFactoryInterface $typeFactory = null;
     // Edge case where the related resource is not readable (for example: NotExposed) but we have groups to read the whole related object
     public const FORCE_SUBSCHEMA = '_api_subschema_force_readable_link';
     public const OPENAPI_DEFINITION_NAME = 'openapi_definition_name';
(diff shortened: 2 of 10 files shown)


bendavies:
in my example, `LocalDate` lives in `vendor` of course, so i could not annotate it with `JsonSchema` directly

soyuka:
In that case you'd have to overwrite the property metadata factory: 

```
class LocalDatePropertyMetadataFactory implements PropertyMetadataFactoryInterface
{
    public function __construct(private readonly ?PropertyMetadataFactoryInterface $decorated = null)
    {
    }

    public function create(string $resourceClass, string $property, array $options = []): ApiProperty
    {
        $propertyMetadata = $this->decorated->create($resourceClass, $property, $options);
        // check propertyMetadata->getBuiltinTypes();
        return $propertyMetadata->withSchema(['type' => 'string', 'format' => 'date']);
    }
}
```

bendavies:
@soyuka i tried 3.4 and my decorating of `ApiPlatform\JsonSchema\TypeFactory` no longer works. is this a bug or am i meat to decorate something else?

bendavies:
well, it's due to
https://github.com/api-platform/core/pull/6480/files#diff-561f7437fabee8e659b12cf2238e5a495028db4b78fd43ebdcf56681b0cb447eR19

PR #5898, opened 2023-10-18
[typo] fix misnamed, so unused, variable in UriVariablesResolverTrait

paullallier:
I suspect this has no impact since everywhere using $context['uri_variables_map'] falls back on $operation->getUriVariables() anyway.

| Q             | A
| ------------- | ---
| Branch?       | 3.1 (or 3.0 if still getting bugfixes) 
| Tickets       | New issue 
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/State/UriVariablesResolverTrait.php
+++ b/src/State/UriVariablesResolverTrait.php
@@ -52,14 +52,14 @@ trait UriVariablesResolverTrait
 
                 foreach ($currentIdentifiers as $key => $value) {
                     $identifiers[$key] = $value;
-                    $uriVariableMap[$key] = $uriVariableDefinition;
+                    $uriVariablesMap[$key] = $uriVariableDefinition;
                 }
 
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


dunglas (approved):
Thanks! Your patch should improve performance.

paullallier:
I suspect we could remove the call to `$operation->getUriVariables()` in these two functions too, which is probably more of a performance improvement.  I wasn't sure enough to do it though. What do you think?  I'm not certainly enough that `$context['uri_variables_map']` will always exist when these functions are called.

https://github.com/api-platform/core/blob/6b79b6f47fa498bdc6c92b574c176f3897e07ec2/src/Api/UriVariablesConverter.php#L47

https://github.com/api-platform/core/blob/6b79b6f47fa498bdc6c92b574c176f3897e07ec2/src/Metadata/UriVariablesConverter.php#L46

soyuka:
In the UriVariblesConverter we want to get `$operation->getUriVariables()` we can't be sure the $context exists.

PR #5901, opened 2023-10-19
fix(JsonSchema): fix reDoc json sample values

ERuban:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | -
| License       | MIT

Reworking of https://github.com/api-platform/core/pull/5638 PR

### DIFF
--- a/features/openapi/docs.feature
+++ b/features/openapi/docs.feature
@@ -101,7 +101,7 @@ Feature: Documentation support
     {
       "type": "string",
       "format": "iri-reference",
-      "example": "string"
+      "example": "https://example.com/"
     }
     """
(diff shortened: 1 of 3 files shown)


ERuban:
@soyuka hi, but `https://example.com` looks strange for iri also. It should be smth like `/books/{id}` in example

ERuban:
👋

soyuka:
still not good, either put `https://example.com/` or fix that in SwaggerUI... Best would be to even generate the proper IRI in our examples but this is not what the PR is doing.

ERuban:
@soyuka ok, done

soyuka:
thanks, to go further we'll see to implement #6058

PR #5907, opened 2023-10-19
ci: fix PHPUNIT

vasilvestre:
| Q             | A
| ------------- | ---
| Branch?       | 2.7 
| License       | MIT

Left :
- [x] Lexer deprecation on many jobs https://github.com/doctrine/lexer/pull/79
- [ ] In ORM 3.0, the AttributeDriver will report fields for the classes where they are declared. This may uncover invalid mapping configurations. To opt into the new mode also with the AnnotationDriver today, set the "reportFieldsWhereDeclared" constructor parameter to true. (AnnotationDriver.php:80 called by AppKernelTestDebugContainer.php:4566, https://github.com/doctrine/orm/pull/10455, package doctrine/orm)
- [ ] PHP 8.1 dev fails on deprecation, with a lot of annotation things. I'm not sure how to keep a layer for annotation but use attributes on this branch.

friendly ping @soyuka for review and the two things left, after them all PHPUNIT should be green


soyuka:
thanks @vasilvestre !

PR #5911, opened 2023-10-20
fix(doctrine): odm order filter should use a left join on nullable fi…

### LINKED ISSUE #5877: OrderFilter on nullable relation seems broken
**API Platform version(s) affected**: 2.7 and up

**Description**  
When using the MongoDbOdm OrderFilter on a field of a relation where the relation can be null, the results with a nullable relation will be omitted, because the OrderFilters use of addLookupsForNestedProperty doesn't do `->$preserveNullAndEmptyArrays(true)`

Basically like this issue but now for mongo:
[Issue 2109](https://github.com/api-platform/core/issues/2109)

**How to reproduce**  
Imagine two entities: Contract and Customer

In the Customer Document we have this field: 
`@ODM\ReferenceMany(targetDocument=Contract::class, mappedBy="customer")`
`private Collection $contracts;`

`@ODM\Field(type="string")`
 `private string $lastName;`

and in the Contract Document we have this field:

`@ODM\ReferenceOne(targetDocument="Customer::class", nullable=true)`
`protected ?Customer $customer = null;`

now the contracts are filterable with an OrderFilter like this:

`@ApiFilter(OrderFilter::class, properties={"customer.lastName")`
`class Contract...`

Ordering like this will only return results, where the customer is not null. All contracts with a null-customer-relation will be omitted.
`https://localhost.com/contracts?order=lastName`

**Possible Solution**  
In the [PropertyHelperTrait.php](https://github.com/api-platform/core/blob/main/src/Doctrine/Odm/PropertyHelperTrait.php#L63) we add an optional parameter like this:
```diff 
+ protected function addLookupsForNestedProperty(string $property, Builder $aggregationBuilder, string $resourceClass, bool $preserveNullAndEmptyArrays = false): array
- protected function addLookupsForNestedProperty(string $property, Builder $aggregationBuilder, string $resourceClass): array
```

and on line [101](https://github.com/api-platform/core/blob/main/src/Doctrine/Odm/PropertyHelperTrait.php#L101) we add
```diff
$aggregationBuilder->unwind("\$$alias")
+    ->preserveNullAndEmptyArrays($preserveNullAndEmptyArrays);
```

now in the [OrderFilter.php](https://github.com/api-platform/core/blob/49c81194a3e6833f10d135c739776636775b15a5/src/Doctrine/Odm/Filter/OrderFilter.php#L256C1-L256C112) class we add the true flag here
```diff
+ [$matchField] = $this->addLookupsForNestedProperty($property, $aggregationBuilder, $resourceClass, true);
- [$matchField] = $this->addLookupsForNestedProperty($property, $aggregationBuilder, $resourceClass);
```

**Additional Context**  
none

skalrayi:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #5877
| License       | MIT
| Doc PR        | 

The odm order filter omits the whole relation if you want to order by a nullable field of a relation.

### DIFF
--- a/src/Doctrine/Odm/Extension/OrderExtension.php
+++ b/src/Doctrine/Odm/Extension/OrderExtension.php
@@ -64,7 +64,7 @@ final class OrderExtension implements AggregationCollectionExtensionInterface
                 }
 
                 if ($this->isPropertyNested($field, $resourceClass)) {
-                    [$field] = $this->addLookupsForNestedProperty($field, $aggregationBuilder, $resourceClass);
+                    [$field] = $this->addLookupsForNestedProperty($field, $aggregationBuilder, $resourceClass, true);
                 }
                 $aggregationBuilder->sort(
                     $context['mongodb_odm_sort_fields'] = ($context['mongodb_odm_sort_fields'] ?? []) + [$field => $order]
--- a/src/Doctrine/Odm/Filter/OrderFilter.php
+++ b/src/Doctrine/Odm/Filter/OrderFilter.php
@@ -253,7 +253,7 @@ final class OrderFilter extends AbstractFilter implements OrderFilterInterface
         $matchField = $property;
 
         if ($this->isPropertyNested($property, $resourceClass)) {
-            [$matchField] = $this->addLookupsForNestedProperty($property, $aggregationBuilder, $resourceClass);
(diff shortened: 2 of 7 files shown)


soyuka:
tyvm, and with a test <3

Issue #5514, opened 2023-03-31
Symfony framework serializer default_context not passed to JsonLd ItemNormalizer


MaSpeng:
**API Platform version(s) affected**: 3.x.x

**Description**  
While investigating, why the property "__isCloning" from the doctrine entity proxy will always be normalized, I tried to ignore this attribute using the [symfony framework configuration](https://symfony.com/doc/current/reference/configuration/framework.html#default-context) to provide a **default context** to the serializer.

The [SerializerPass](https://github.com/symfony/symfony/blob/e46cc76c95fb52a41e558938425f8efa45a1a6b6/src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php#L72) from the symfony serializer component has the task to pass the configured **default context** to all **normalizers** which are tagged with `serializer.normalizer`.

Currently, only one service with the id `api_platform.serializer.normalizer.item` has the required tag.

As a result, whenever another serializer, besides this one, is used, it is missing the configured **default context**.

This affects the JsonLd `ItemNormalizer` and properly all other normalizers provided by API Platform as well.

**How to reproduce**  
- setup a symfony project containing API Platform
- add an API resource containing at least two properties
- configure `framework.serializer.default_context` with `ignored_attributes` containing one of the two attributes of the API resource, for example
- use the swagger UI to create one instance of the API resource
- introspect the response which is still containing the ignored attribute
- it is also possible to introspect the arguments of `\ApiPlatform\JsonApi\Serializer\ItemNormalizer::__construct` to see that the `defaultContext` array is always empty

**Possible Solution**  
Maybe tagging the other provided normalizer with the tag `serializer.normalizer` is enough to solve the issue.

**Additional Context**

MaSpeng:
Could some maintainers please have a look, so this issue does not get lost?

MaSpeng:
This issue requires classification.

soyuka:
can you provide a patch?

MaSpeng:
Since it is not clear to me why API Platform chose this implementation to serialize API Resources, I am not able to provide an actual patch for it.

My suggestion was a starting point for a possible solution but I am not sure if this might introduce unwanted side effects like the execution of serializers which should not be used in the end.

I was actually hoping that one of the core members who introduced the current concept would be able to take a look and determine if my suggestion could be harmful or not or better actually provide a fix for this issue.

It is also hard for me to actually construct a test case for this issue to ensure no other side effects are appearing.

PR #5920, opened 2023-10-23
fix(Serializer): defaultContext in ItemNormalizer

alli83:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | https://github.com/api-platform/core/issues/5514
| License       | MIT
| Doc PR        | 

Based on https://github.com/symfony/symfony/pull/47637 and taking into account https://github.com/api-platform/core/issues/5514 , this PR aims to consider the `default_context` config option under `serializer` of `FrameworkBundle` when using `ApiPlatform\JsonLd\Serializer\ItemNormalizer`.

Maybe we should check for the other normalizers.

### DIFF
--- a/src/JsonLd/DependencyInjection/NormalizerPass.php
+++ b/src/JsonLd/DependencyInjection/NormalizerPass.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\JsonLd\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class NormalizerPass implements CompilerPassInterface
+{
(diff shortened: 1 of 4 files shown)


src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -45,7 +45,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
 
     public const FORMAT = 'jsonld';
 
-    public function __construct(ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory, PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, PropertyMetadataFactoryInterface $propertyMetadataFactory, IriConverterInterface $iriConverter, ResourceClassResolverInterface $resourceClassResolver, private readonly ContextBuilderInterface $contextBuilder, PropertyAccessorInterface $propertyAccessor = null, NameConverterInterface $nameConverter = null, ClassMetadataFactoryInterface $classMetadataFactory = null, array $defaultContext = [], ResourceAccessCheckerInterface $resourceAccessChecker = null)
```
soyuka, line 48: we can't change the constructor it's a public interface. Maybe that we should provide this from our DI?

alli83: yes, I'll update it

alli83: @soyuka maybe I should add a compiler pass to enforce the use of the default value when encountering an empty string?

PR #5922, opened 2023-10-23
perf(http_cache): don't load every value for a non-resource collection

mwierzbi:
| Q             | A
| ------------- | ---
| Branch?       | 3.1 
| Tickets       | -  
| License       | MIT
| Doc PR        |  -  

**Description**
When a entity has relation to non-resource collection and is flushed, the `PurgeHttpCacheListener` loads each value and checks if it is a resource. This can be a problem with large collections over 10k, it can cause a out-of-memory exception. 

**Solution**
We can check whether the collection contains an entity that is a resource or not before loading each value by checking the class form `$associationMappings[$property]['targetEntity']`


soyuka:
amazing thanks!

PR #5932, opened 2023-10-26
fix: use http exception headers

### LINKED ISSUE #5924: [3.2] Bc break with exception handling

In 3.1, when an exception occurs, Symfony set the _controller (in ErrorListener) to `api_platform.action.exception`
In 3.2 it's `api_platform.symfony.main_controller`.

There is a drawback, because in my exception, I have some headers, and there are not replicated anymore in the final response.

**How to reproduce**  

```
throw new TooManyRequestsHttpException(32);
```

And see the is not "Retry-After" header in the response 

**Possible Solution**  

* revert the change
* **or** add support for header duplication in the MainController

soyuka:
fixes #5924

### DIFF
--- a/src/State/Processor/RespondProcessor.php
+++ b/src/State/Processor/RespondProcessor.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\State\Processor;
 
+use ApiPlatform\Metadata\Exception\HttpExceptionInterface;
 use ApiPlatform\Metadata\HttpOperation;
 use ApiPlatform\Metadata\IriConverterInterface;
 use ApiPlatform\Metadata\Operation;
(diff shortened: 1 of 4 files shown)


src/State/Processor/RespondProcessor.php
```diff
@@ -64,6 +66,11 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
             'X-Frame-Options' => 'deny',
         ];
 
+        $exception = $request->attributes->get('exception');
+        if (($exception instanceof HttpExceptionInterface || $exception instanceof SymfonyHttpExceptionInterface) && $exceptionHeaders = $exception->getHeaders()) {
```
soyuka, line 70: I'm not a huge fan of this right now because of the HttpKernel dependency, I'll see to probably move this in the bridge in the future.

PR #5946, opened 2023-11-04
chore(github): update GitHub Actions

sad270:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        | 

## Update https://github.com/actions/checkout from 3 to 4 
<details><summary>Changelog</summary>

> # Changelog
> 
> ## v4.1.0
> - [Add support for partial checkout filters](https://github.com/actions/checkout/pull/1396)
> 
> ## v4.0.0
> - [Support fetching without the --progress option](https://github.com/actions/checkout/pull/1067)
> - [Update to node20](https://github.com/actions/checkout/pull/1436)

**Source :** https://github.com/actions/checkout/blob/v4.1.1/CHANGELOG.md
**Full Changelog :** https://github.com/actions/checkout/compare/v3.6.0...v4.1.1
</details> 

## Update https://github.com/actions/setup-node from 3 to 4 
<details><summary>Changelog</summary>

> # v4.0.0
> In scope of this release we changed version of node runtime for action from node16 to node20 and updated dependencies in https://github.com/actions/setup-node/pull/866
>
> Besides, release contains such changes as:
> - [Upgrade actions/checkout to v4](https://github.com/actions/setup-node/pull/868)
> - [Update actions/checkout for documentation and yaml](https://github.com/actions/setup-node/pull/876)

**Source :** https://github.com/actions/setup-node/releases/tag/v4.0.0
**Full Changelog :** https://github.com/actions/setup-node/compare/v3.8.2...v4.0.0
</details>

PR #5947, opened 2023-11-04
chore(dependabot): add dependabot

sad270:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | 
| License       | MIT
| Doc PR        | 

Add dependabot to update github actions.

Configuration is comming from https://github.com/api-platform/api-platform/blob/main/.github/dependabot.yml

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly


soyuka:
we're using our own `update-js` script there is no package-lock on this project I think

sad270:
Hello, i'm maybe wrong but `update-js` will update js js packages, in this PR dependabot will update github action `uses`.

sad270:
Hello, i'm maybe wrong but `update-js` will update js js packages, in this PR dependabot will update github action `uses`.

sad270:
In https://github.com/api-platform/core/pull/5946 I did it manually 😅 but if it is automatically updated it will be better ! With dependabot or with another tool.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+
```
dunglas, line 9:
```suggestion
```

dunglas (approved):

dunglas:
Y the way, as we keep the `yarn.lock` file, we may also replace `update-js.js` by Dependabot (in a next PR).

.github/dependabot.yml
sad270: 👍🏻 it's removed !

dunglas:
One last thing: this should target the 3.1 branch

sad270:
> One last thing: this should target the 3.1 branch

@dunglas target changed it's `3.1` now 😄

dunglas:
Thank you!!

PR #5952, opened 2023-11-07
chore: support Symfony 7

endroid:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | l-
| License       | MIT
| Doc PR        | -

Adds support for Symfony 7


endroid:
@dunglas I created a fork with the incorrect branch and made my changes from the main branch so I created a new PR. This one should work.

endroid:
@dunglas Do we wait for the mongo db bundle compatibility issue? API seems to be working correctly here (not using mongodb).

dunglas:
In my opinion there is no need to wait for this deps, but we must skip the tests with MongoDB on Symfony 7 to make the CI green.

endroid:
But if you skip the tests you lose some compatibility information, isn't that a problem? If you want MonboDBBundle to be supported I would keep those tests, even if they fail currently. Or is it really impossible to merge if these jobs fail?

dunglas:
We need our CI green or it's very hard to detect if a patch breaks something new or not. It's okay to skip tests for things that are currently not supported as long as we continue testing these features with supported versions. We'll unskip Mongo tests when their bundle will support Symfony 7. No need to block the upgrade of all API Platform users just for Mongo.

endroid:
I now skip the mentioned tests, however I get some new PHPStan issues which seem unrelated to my changes (not failing on my previous commit).

endroid:
Or could you trigger a re-run?

soyuka:
> Could you trigger a re-run so we can check the status?
done

KDederichs:
Kinda looks like the PHP Matrix broke? PHP 8.1 tests should not install PHP 8.3

endroid:
I see the maintainers already applied all these changes to the 3.2 branch so I will close this one.

PR #5954, opened 2023-11-08
GraphQL: Consider writable attribute for nested objects

jotwea:
| Q             | A
| ------------- | ---
| Branch       | 3.2
| Tickets       | -
| License       | MIT
| Doc PR        | -

We use Doctrine Embeddables in our project. So we also have mutations that contain nested objects. To exclude some properties from beeing modified the `writable: false` option works fine. But this only worked on the rool level. For nested objects this got ignored. This PR is there to fix that and consider `writable: false` also for nested objects.

I invested much in writing behat tests that provide a detailed understanding of the issue.


tests/Fixtures/TestBundle/Entity/FooDummy.php
```diff
@@ -35,11 +37,21 @@ class FooDummy
...
+
     /**
      * @var string The foo name
      */
     #[ORM\Column]
     private $name;
+
+    #[ApiProperty(writable: false)]
+    #[ORM\Column(nullable: true)]
+    public $nonWritableProp;
```
soyuka, line 49: actually instead of setting `writable: false` explicitly we could provide only a getter without setter?

tests/Fixtures/TestBundle/Entity/FooEmbeddable.php
```diff
@@ -0,0 +1,53 @@
...
+class FooEmbeddable
+{
+    /**
+     * @var string The dummy name
+     */
+    #[ApiProperty(identifier: true)]
+    #[ORM\Column(nullable: true)]
+    private ?string $dummyName = null;
+
+    #[ApiProperty(writable: false)]
```
soyuka, line 36: same comment as above

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -220,7 +220,7 @@ public function getResourceObjectTypeFields(?string $resourceClass, Operation $o
                 if (
                     !$propertyTypes
                     || (!$input && false === $propertyMetadata->isReadable())
-                    || ($input && $operation instanceof Mutation && false === $propertyMetadata->isWritable())
+                    || ($input && false === $propertyMetadata->isWritable())
```
soyuka, line 223: I like the change :+1:

tests/Fixtures/TestBundle/Entity/FooDummy.php
```diff
@@ -35,11 +37,21 @@ class FooDummy
...
+
+    #[ApiProperty(readableLink: true, writableLink: true)]
```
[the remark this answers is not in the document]
jotwea: Unfortunatelly they are not `true` by default. If not setting these attributes, there will be a skolem IRI in the JSON instead.

It would be a great behaviour for Embeddables to treat them more like properties than relations. Could you assist in working on that? But I would recommend to tackle this in a separate issue and PR.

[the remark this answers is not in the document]
soyuka: embeddables were hard to handle previously because if they're marked as a resource the behavior was different. Now I think it'd be possible to do this indeed.

soyuka:
the changes in the entity breaks an unrelated test, best would be to create a new entity for this test case (https://github.com/api-platform/core/blob/main/CONTRIBUTING.md#tests)

jotwea:
> the changes in the entity breaks an unrelated test, best would be to create a new entity for this test case (https://github.com/api-platform/core/blob/main/CONTRIBUTING.md#tests)

Fixed it. The other failing tests seem to be already broken in the 3.2 source branch

soyuka:
thanks!

PR #5955, opened 2023-11-08
GraphQL: Support Enum collections

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -

We have a resource with an array property. The content of this array are enum values. This information can be provided with a doc block. Unfortunatelly on the GraphQL API this was visible as an `iterable`. With the changes in this PR it will be a proper "array of enum", same like i.e. "array of string".

The Behat tests should give you an insight into the details.


soyuka (approved):
Nice, could you rebase against main though it looks like there's a commit that doesn't belong to this patch. Commit should be: `feat(graphql): ...`

soyuka:
nice thanks@jotwea!

PR #5956, opened 2023-11-09
Fix issues with extending Yaml -> Yaml resources, additionally fix extending {XML,Yaml} -> {XML,Yaml} resources

senghe:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        |

Hi all! :) I was testing a feature of extending resources by multiple XML/Yaml resource files for the same resource. For XMLs it works from scratch, but there is two issues:

- First, when trying to extend Yaml configuration by another Yaml configuration
- Second, when trying to extend Yaml configuration by XML, and when trying to extend XML by Yaml.

It looks for me as a bug, because as you can see, there is too little changes from me to solve it.

Please let me know, what is your opinion regarding to this topic :vulcan_salute:


senghe:
I see one build run has been cancelled - the red one. Could you rerun it? :pray:

src/Metadata/Extractor/YamlResourceExtractor.php
```diff
@@ -75,14 +75,15 @@ private function buildResources(array $resourcesYaml, string $path): void
                 $resourceYaml = [$resourceYaml];
             }
 
+            $resourcesCount = isset($this->resources[$resourceName]) ? \count($this->resources[$resourceName]) : 0;
             foreach ($resourceYaml as $key => $resourceYamlDatum) {
```
soyuka, line 79: is key always an int @vincentchalamon ?

soyuka:
this looks nice, can you target 3.1? We should add a test somehow

src/Metadata/Extractor/YamlResourceExtractor.php
vincentchalamon: Hmmm not sure that it's always an int. To be sure, it would be better to rely on an int variable declared out of this foreach and incremented accordingly.

senghe:
Hi, to be honest I'm not so familiar with API Platform dev structure, so I think I'm not able to add tests here. But in a free time I'll check whether the index is always integer - as it was discussed above and refactor this.

VincentLanglet:
Hi, the 3.x version is not maintained anymore ; could you update your PR to target 4.2 instead ?
Thanks

PR #5964, opened 2023-11-11
fix(doctrine): OrderFilterTrait.php - iterating over a null value causes error in foreach

thatsLegit:
$properties cannot be null as iterating over null causes an error.
$properties can be null if the result of $this->getProperties() and then $this->getClassMetadata($resourceClass)->getFieldNames() is an empty array.


soyuka:
nice patch, can you target 3.1 and also rebase?

soyuka:
cherry-picked manually thanks

PR #5965, opened 2023-11-12
fix(openapi): api_entrypoint should use "docs_formats" config

paullallier:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Addresses https://github.com/api-platform/core/issues/5845#issuecomment-1748641671
| License       | MIT
| Doc PR        | N/A

I still have a slightly obscure issue where the api_entrypoint is returning a 406 error in some cases (at least when I load the page with curl with "Accept: text/html" headers).  Essentially, its using the formats rather than docs_formats config for that url, and my formats list doesn't include text/html as an option.

There's a test for this currently, I think, but it already passes because the test API has text/html in the config file for both formats and doc_formats, which masks that it's picking the wrong option.  I'm not sure if there's a way to modify the config on a single Behat test?

### DIFF
--- a/features/hydra/docs.feature
+++ b/features/hydra/docs.feature
@@ -4,6 +4,7 @@ Feature: Documentation support
   I need to know Hydra specifications of objects I send and receive
 
   Scenario: Checks that the Link pointing to the Hydra documentation is set
+    When I add "Accept" header equal to "application/ld+json"
     Given I send a "GET" request to "/"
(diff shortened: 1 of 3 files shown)


paullallier:
I've "fixed" some failing Behat test, but I'm  not sure that my fix is valid - perhaps I've changed what's being tested?

Also, I seem to have tests failing on a deprecation - I'm not sure if this is something I broke or not?

paullallier:
Maybe fixes #6010?

paullallier:
Superseded by #6012 - thanks!

Issue #5963, opened 2023-11-10
Error handling with 3.2 is inconsistent & difficult


j-schumann:
**Description**  
With 3.2 there are two different ways in which error responses are created:
1) When the problem occurs while routing or authentication, Symfony's own `ProblemNormalizer` is used.
2) When the problem occurs within ApiPlatforms Processors/Providers an `ApiResource\Error` is created and serialized. (At least when using no deprecated features / the new problem responses)

This leads to multiple problems:
* returned content + headers change for similar errors / same status codes: see #5961
* We can no longer decorate 'api_platform.hydra.normalizer.error' to modify all returned error responses, e.g. to translate the returned error 'detail' / 'hydra:description'.
  * instead we have to create a new Processor to decorate 'api_platform.state_processor.serialize' to modify the `ApiResource\Error` error before it is serialized
  * and modify our existing serializer to instead decorate 'serializer.normalizer.problem' to do the same thing but on an array
* `ApiResource\Error` is immutable, which requires us to create a new instance to modify a single value, e.g. the 'detail'
  * Why are there no setters, and if it is required to be immutable, can we please get `->with...()` clone methods like most of the `ApiPlatform\OpenApi\Model` classes have (withParameters, withTrace, withDescription, ...)? Also the `ORM|ODM|ElasticSearch\State\Options` or the MetaData class(es) have them.

The feature request would be to simplify the modification of the error results, for all error sources together and to better document how this can be done. Currently the docs (https://api-platform.com/docs/core/errors/) neither shows how to use a Provider nor a Serializer to modify the response, and it does not mention the (current) differences in error handling & output.

**Example**  
We want to translate the error messages generated by Symfony's routing / Lexik JWT / Validators / custom exceptions to be parseable by the API-Client.
Currently we have to use:
```    
    $clone = new Error(
        $data->getTitle(),
        $this->translator->trans($data->getDetail(), [], 'validators'),
        $data->getStatus(),
        $data->originalTrace,
        $data->getInstance(),
        $data->getType(),
        $data->getHeaders(),
    );
```
and
```
        $result['detail'] = $this->translator
            ->trans($result['detail'], [], 'validators');

        // currently not available because of #5961 
        //$result['hydra:description'] = $this->translator
        //    ->trans($result['[detail](hydra:description)'], [], 'validators');
```
[...]

PR #5974, opened 2023-11-17
fix: errors bc with rfc_7807_compliant_errors false

soyuka:
# What does this fix: 

- inconsistencies https://github.com/api-platform/core/issues/5963
- backward compatibility of errors
- ways to customize rfc_7807_compliant_errors

## Try this:

composer.json: 
```
"packages: {"api-platform/core": "dev-error-bc"},
"minimum-stability": "dev",
"repositories": [
  { "type": "vcs", "url": "https://github.com/soyuka/core" }
]
```

# TODO: 

- [ ] add a `guide` 

# Documentation:

# Errors 

## Backward compatibility with < 3.1

Use the following configuration: 

```yaml
api_platform:
    defaults:
            rfc_7807_compliant_errors: true
```

This can also be configured on an `ApiResource` or in an `HttpOperation`, for example: 

```php
#[ApiResource(extraProperties: ['rfc_7807_compliant_errors' => true])
```

## Control your exceptions

With `rfc_7807_compliant_errors` a few things happen. First Hydra exception are compatible with the JSON Problem specification. Default exception that are handled by API Platform in JSON will be returned as `application/problem+json`. 

To customize the API Platform response, replace the `api_platform.state.error_provider` with your own provider: 

```php
<?php

namespace App\State;

use ApiPlatform\State\ApiResource\Error;
use ApiPlatform\State\ProviderInterface;
use Symfony\Component\DependencyInjection\Attribute\AsAlias;
use Symfony\Component\DependencyInjection\Attribute\AsTaggedItem;

#[AsAlias('api_platform.state.error_provider')]
#[AsTaggedItem('api_platform.state.error_provider')]
final class ErrorProvider implements ProviderInterface
{
    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
    {
        $request = $context['request'];
        $format = $request->getRequestFormat();
        $exception = $request->attributes->get('exception');

        /** @var \ApiPlatform\Metadata\HttpOperation $operation */
        $status = $operation->getStatus() ?? 500;
        // You don't have to use this, you can use a Response, an array or any object (preferably a resource that API Platform can handle).
        $error = Error::createFromException($exception, $status);

        // care about hiding informations as this can be a security leak
        if ($status >= 500) {
            $error->setDetail('Something went wrong');
        }
        
        return $error;
    }
}
```

```yaml
    api_platform.state.error_provider:
        class: 'App\State\ErrorProvider'
        tags: 
            - key: 'api_platform.state.error_provider'
              name: 'api_platform.state_provider'
```

Note that our validation exception have their own error provider at:

```yaml
api_platform.validator.state.error_provider:
    tags: 
        - key: 'api_platform.validator.state.error_provider'
          name: 'api_platform.state_provider'
```

## Domain exceptions

Another way of having full control over domain exceptions is to create your own Error resource: 

```php
<?php

namespace App\ApiResource;

use ApiPlatform\Metadata\ErrorResource;
use ApiPlatform\Metadata\Exception\ProblemExceptionInterface;

#[ErrorResource()]
class Error extends \Exception implements ProblemExceptionInterface
{
    public function getType(): string
    {
        return 'teapot';
    }

    public function getTitle(): ?string
    {
        return null;
    }

    public function getStatus(): ?int
    {
        return 418;
    }

    public function getDetail(): ?string
    {
        return 'I am teapot';
    }

    public function getInstance(): ?string
    {
        return null;
    }
}
```

We recommend using the `\ApiPlatform\Metadata\Exception\ProblemExceptionInterface` and the `\ApiPlatform\Metadata\Exception\HttpExceptionInterface`. For security reasons we add: `normalizationContext: ['ignored_attributes' => ['trace', 'file', 'line', 'code', 'message', 'traceAsString']]` because you usually don't want these. You can override this context value if you want.

## Exception status code: 

1. we look at `exception_to_status` and take one if there's a match
2. If your exception is a `Symfony\Component\HttpKernel\Exception\HttpExceptionInterface` we get its status.
3. If the exception is a `ApiPlatform\Metadata\Exception\ProblemExceptionInterface` and there is a status we use it
4. Same for `ApiPlatform\Metadata\Exception\HttpExceptionInterface` 
5. We have some defaults: 
  - `Symfony\Component\HttpFoundation\Exception\RequestExceptionInterface` => 400
  - `ApiPlatform\Validator\Exception\ValidationException` => 422
6. the status defined on an `ErrorResource`
7. 500 is the fallback


soyuka:
Failing guides is expected as it doesn't test the code from this branch but from main, my local tests work :)

NikDevPHP:
@soyuka Hello, could you please help, what is the right way to create a ValidationException error from code now ?

    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = []): NotificationTemplate
    {
        Assert::isInstanceOf($data, CreateSitesNotificationTemplate::class);
        Assert::notEmpty($data->siteIds);

        $siteIds = $this->siteRepository->getSiteIdsWithAnyAssignedUsersByIds($data->siteIds);
        Assert::notEmpty($siteIds);

        /** @var list<int> $sitesWithoutUsersIds */
        $sitesWithoutUsersIds = array_diff($data->siteIds, $siteIds);

        if ($sitesWithoutUsersIds !== []) {
            throw new SitesWithoutUsersException($sitesWithoutUsersIds); // <- here I need ValidationException (422 code)
        }

        return $notificationTemplate;
    }

PR #5982, opened 2023-11-20
feat(Serializer): Allow to validate security in input DTOs

voodooism:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT
| Doc PR        | don't think we need it for the feature, but ready to add

Shortly, this pull request introduces security functionality for attributes in input DTOs for the API platform.

I've noticed that the current API platform lacks security functionality for attributes in input DTOs. Although it's possible in ApiResource to check if an attribute is accessible for the current user, like this:

```php
    #[ApiProperty(security: "is_granted('ROLE_ADMIN')")]
    public $adminOnlyProperty;
```
    
Unfortunately, this functionality is not available for input DTOs.

This pull request aims to introduce a feature allowing the checking of user roles for certain properties within input DTOs. In my project, this capability is crucial for controlling access to specific properties based on user roles and enhancing security and data integrity. This enhancement can also be beneficial for other users of the platform who require similar role-based access control for their APIs.

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -182,7 +182,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
      */
     public function supportsDenormalization(mixed $data, string $type, string $format = null, array $context = []): bool
     {
-        if (($context['input']['class'] ?? null) === $type) {
+        if (($context['input']['class'] ?? null) === $type || (($context['from_input'] ?? false) === true && ($context['resource_class'] ?? null) === $type)) {
             return true;
         }
 
@@ -203,6 +203,10 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
 
             unset($context['input'], $context['operation'], $context['operation_name']);
             $context['resource_class'] = $inputClass;
+            $context['from_input'] = true;
+            if (false === isset($context['groups']) || [] === $context['groups']) {
+                $context['groups'][] = '*';
+            }
 
             try {
                 return $this->serializer->denormalize($data, $inputClass, $format, $context);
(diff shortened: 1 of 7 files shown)


voodooism:
@soyuka  @dunglas have a look please

src/Serializer/AbstractItemNormalizer.php
```diff
@@ -203,6 +203,10 @@ public function denormalize(mixed $data, string $class, string $format = null, a
 
             unset($context['input'], $context['operation'], $context['operation_name']);
             $context['resource_class'] = $inputClass;
+            $context['from_input'] = true;
+            if (false === isset($context['groups']) || [] === $context['groups']) {
+                $context['groups'][] = '*';
```
soyuka, line 208: it looks quite dangerous to me to replace groups that are not set as this is not the same behavior

soyuka:
Instead of this we should just consider inputs and outputs as resources so that they take the exact same path as a resource would and therefore you'd get all the features. I need to dig deeper I think there's something easier to be done here.

src/Serializer/AbstractItemNormalizer.php
voodooism: https://github.com/symfony/serializer/blob/6.3/Normalizer/AbstractNormalizer.php#L253

Only because of this check. In case we don't have groups this method will return false instead of list allowed attributes

voodooism:
> Instead of this we should just consider inputs and outputs as resources so that they take the exact same path as a resource would and therefore you'd get all the features. I need to dig deeper I think there's something easier to be done here.

We can, but it looks a bit strange to me. 
We already have an ApiResource and would like to use DTO for it, and the fact we have to consider our DTO as API resources confuses me a lot :(

voodooism:
@soyuka could you please have a look at the previous comment and https://github.com/api-platform/core/pull/5994?

soyuka:
I don't think this is right, actually look at this:

https://github.com/api-platform/core/blob/cbc803cab81bc6acbe74c4db4eba63bc8539aa16/src/Serializer/AbstractItemNormalizer.php#L436-L452 

The call at:
```
        $propertyMetadata = $this->propertyMetadataFactory->create($context['resource_class'], $attribute, $options);
```
 
 can probably just be something like:
 
```
        $propertyMetadata = $this->propertyMetadataFactory->create($inputClass, $attribute, $options);
```

There are helpers to get `getInputClass` look at the code there, also it's either input either output :).

I'd like to see this introduced as a new feature on the main branch. Sorry for the delay.

PR #5992, opened 2023-11-24
fix: missing eager joins on to-one relationships

### LINKED ISSUE #5847: Eager Loading Extension is not joining some *toOne that should be joined
**API Platform version(s) affected**: 3.1.18

**Description**  

In some cases, the EagerLoadingExtension misses some joins it should normally load eagerly. This happens when a *ToOne relationship shares the same `inversedBy` name than in the parent class.

**How to reproduce**  

Given the three ApiResources below :

```php
#[ApiResource(
    operations: [
        new GetCollection(
            normalizationContext: ["groups" => ["employee:get"]],
        )
    ]
)]
#[ORM\Entity]
class Employee
{
   // [...]

    #[ORM\OneToMany(mappedBy: 'employee', targetEntity: Subscription::class)]
    #[Groups(['employee:get'])]
    public Collection $subscriptions;
}
```

```php
#[ApiResource]
#[ORM\Entity]
class Subscription
{
    // [...]

    #[ORM\ManyToOne(inversedBy: 'subscriptions')]
    #[Groups(['employee:get'])]
    public Project $project;
}
```

```php
#[ApiResource]
#[ORM\Entity]
class Project
{
   // [...]

    #[ORM\OneToMany(mappedBy: 'project', targetEntity: Subscription::class)]
    public iterable $subscriptions;
}
```

If you request `GET /employees`, the relation employee.subscriptions is correctly eager fetched but not the relation employee.subscriptions.project. This is due to the below piece of code in the EagerLoadingExtension which skips the eager loading if the inversedBy property has the same name as the parent association. But, in our case, the name is the same but it does not refer to the same class.

```php
// Avoid joining back to the parent that we just came from, but only on *ToOne relations
            if (
                null !== $parentAssociation
                && isset($mapping['inversedBy'])
                && $mapping['inversedBy'] === $parentAssociation
                && $mapping['type'] & ClassMetadata::TO_ONE
            ) {
                continue;
            }
```

**Possible Solution**  

Not completely sure, but checking additionally that the parent class is the same as the target one before skipping the eager loading could fix the issue.

MariusJam:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #5847 
| License       | MIT

In some cases, the EagerLoadingExtension misses some joins it should normally load eagerly. This happens when a *ToOne relationship shares the same inversedBy name than in the parent class. This PR aims to solve this problem and correctly eager join all to-one relationships than should be eager joined.

### DIFF
--- a/features/json/relation.feature
+++ b/features/json/relation.feature
@@ -25,7 +25,8 @@ Feature: JSON relations support
       "badFourthLevel": null,
       "id": 1,
       "level": 3,
-      "test": true
+      "test": true,
+      "relatedDummies": []
     }
     """
 
--- a/src/Doctrine/Orm/Extension/EagerLoadingExtension.php
+++ b/src/Doctrine/Orm/Extension/EagerLoadingExtension.php
@@ -168,6 +168,7 @@ final class EagerLoadingExtension implements QueryCollectionExtensionInterface, 
             if (
                 null !== $parentAssociation
                 && isset($mapping['inversedBy'])
+                && $mapping['sourceEntity'] === $mapping['targetEntity']
                 && $mapping['inversedBy'] === $parentAssociation
                 && $mapping['type'] & ClassMetadata::TO_ONE
             ) {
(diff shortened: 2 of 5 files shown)


soyuka:
nice,  can you target 3.1?

MariusJam:
It's done

soyuka:
thanks @MariusJam

soyuka:
Hi, it happens I have a few tests in error with your patch, I reverted but I think it'd still be a good patch if you can take another look  (ci is green at #6000)

PR #5993, opened 2023-11-24
Bugfix/exclusive properties dont allow booleans for open api 3.1

arnedesmedt:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| License       | MIT

The generated open api spec is currently invalid for open api 3.1. 
exclusiveMinimum and exclusiveMaximum are still used as boolean, while properties minimum and exclusiveMinimum are merged into one property exclusiveMinimum. 

See https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

The only problem I see now, is that users can configure their open api version. 
But I don't know how I can toggle the logic between 3.0 and 3.1.
Any help is appreciated. Thx!


soyuka:
nice thanks!

PR #5495, opened 2023-03-23
Fix SerializedName and name converter

Zowac:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | #4752 
| License       | MIT
| Doc PR        | api-platform/docs#...

PR #5995, opened 2023-11-24
docs(adr): filtering system and query parameters

soyuka:
Introduction of an ADR, few things must be discussed, I joined a few ideas we collected over the past few years. This must be seen as a research into providing better developer experience and more freedom when expressing filters with API Platform. 

Friendly ping: @aegypius, @mrossard, @metaclass-nl, @helyakin, @jdeniau, @bendavies, @bpolaszek 

if you have ideas, things we should think of, or opinions I'd be glad to hear them, thanks!

### DIFF
--- a/src/Doctrine/Common/Filter/UrlPatternToFilterContextTrait.php
+++ b/src/Doctrine/Common/Filter/UrlPatternToFilterContextTrait.php
@@ -1,14 +1,26 @@
 <?php
 
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
 namespace ApiPlatform\Doctrine\Common\Filter;
 
 /**
- * Handles the :property url pattern from ApiPlatform parameters and prepares the `filter` $context
+ * Handles the :property url pattern from ApiPlatform parameters and prepares the `filter` $context.
  *
  * @see ApiPlatform\Doctrine\Orm\Filter\FilterInterface
  * @see ApiPlatform\Metadata\Parameter
  *
  * @author Antoine Bluchet <soyuka@gmail.com>
+ *
  * @internal
  */
 class UrlPatternToFilterContextTrait
--- a/src/Doctrine/Odm/Extension/ParameterExtension.php
+++ b/src/Doctrine/Odm/Extension/ParameterExtension.php
@@ -14,8 +14,6 @@ declare(strict_types=1);
 namespace ApiPlatform\Doctrine\Odm\Extension;
 
 use ApiPlatform\Doctrine\Odm\Filter\FilterInterface;
-use ApiPlatform\Exception\InvalidArgumentException;
-use ApiPlatform\Metadata\HeaderParameterInterface;
 use ApiPlatform\Metadata\Operation;
 use Doctrine\ODM\MongoDB\Aggregation\Builder;
 use Psr\Container\ContainerInterface;
--- a/src/Doctrine/Orm/Extension/ParameterExtension.php
+++ b/src/Doctrine/Orm/Extension/ParameterExtension.php
@@ -15,8 +15,6 @@ namespace ApiPlatform\Doctrine\Orm\Extension;
 
 use ApiPlatform\Doctrine\Orm\Filter\FilterInterface;
 use ApiPlatform\Doctrine\Orm\Util\QueryNameGeneratorInterface;
-use ApiPlatform\Exception\InvalidArgumentException;
-use ApiPlatform\Metadata\HeaderParameterInterface;
 use ApiPlatform\Metadata\Operation;
 use Doctrine\ORM\QueryBuilder;
 use Psr\Container\ContainerInterface;
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -290,7 +290,103 @@ final class FieldsBuilder implements FieldsBuilderInterface, FieldsBuilderEnumIn
             $args[$id]['type'] = $this->typeConverter->resolveType($arg['type']);
         }
 
+        foreach ($operation->getParameters() ?? [] as $parameter) {
+            $key = $parameter->getKey();
+
+            if (str_contains($key, ':property')) {
+                if (!($filterId = $parameter->getFilter()) || !$this->filterLocator->has($filterId)) {
+                    continue;
+                }
+
+                $parsedKey = explode('[:property]', $key);
+                $flattenFields = [];
+                foreach ($this->filterLocator->get($filterId)->getDescription($operation->getClass()) as $key => $value) {
+                    $values = [];
+                    parse_str($key, $values);
+                    if (isset($values[$parsedKey[0]])) {
+                        $values = $values[$parsedKey[0]];
+                    }
+
+                    $name = key($values);
+                    $flattenFields[] = ['name' => $name, 'required' => $value['required'] ?? null, 'description' => $value['description'] ?? null, 'leafs' => $values[$name], 'type' => $value['type'] ?? 'string'];
+                }
+
+                $args[$parsedKey[0]] = $this->parameterToObjectType($flattenFields, $parsedKey[0]);
+                continue;
+            }
+
+            $args[$key] = ['type' => GraphQLType::string()];
+
+            if ($parameter->getRequired()) {
+                $args[$key]['type'] = GraphQLType::nonNull($args[$key]['type']);
+            }
+        }
+
         return $args;
+    }
+
+    /**
+     * Transform the result of a parse_str to a GraphQL object type.
+     *
+     * @param array<array{name: string, required: bool|null, description: string|null, leafs: string|array, type: string}> $flattenFields
+     */
+    private function parameterToObjectType(array $flattenFields, string $name): InputObjectType
+    {
+        $fields = [];
+        foreach ($flattenFields as $field) {
+            $key = $field['name'];
+            $type = $this->getParameterType(\in_array($field['type'], Type::$builtinTypes, true) ? new Type($field['type'], !$field['required']) : new Type('object', !$field['required'], $field['type']));
+
+            if (\is_array($l = $field['leafs'])) {
+                if (0 === key($l)) {
+                    $key = $key;
+                    $type = GraphQLType::listOf($type);
+                } else {
+                    $n = [];
+                    foreach ($field['leafs'] as $l => $value) {
+                        $n[] = ['required' => null, 'name' => $l, 'leafs' => $value, 'type' => 'string', 'description' => null];
+                    }
+
+                    $type = $this->parameterToObjectType($n, $key);
+                    if (isset($fields[$key]) && ($t = $fields[$key]['type']) instanceof InputObjectType) {
+                        $t = $fields[$key]['type'];
+                        $t->config['fields'] = array_merge($t->config['fields'], $type->config['fields']);
+                        $type = $t;
+                    }
+                }
+            }
+
+            if ($field['required']) {
+                $type = GraphQLType::nonNull($type);
+            }
+
+            if (isset($fields[$key])) {
+                if ($type instanceof ListOfType) {
+                    $key .= '_list';
+                }
+            }
+
+            $fields[$key] = ['type' => $type, 'name' => $key];
+        }
+
+        return new InputObjectType(['name' => $name, 'fields' => $fields]);
+    }
+
+    /**
+     * A simplified version of convert type that does not support resources.
+     */
+    private function getParameterType(Type $type): GraphQLType
+    {
+        return match ($type->getBuiltinType()) {
+            Type::BUILTIN_TYPE_BOOL => GraphQLType::boolean(),
+            Type::BUILTIN_TYPE_INT => GraphQLType::int(),
+            Type::BUILTIN_TYPE_FLOAT => GraphQLType::float(),
+            Type::BUILTIN_TYPE_STRING => GraphQLType::string(),
+            Type::BUILTIN_TYPE_ARRAY => GraphQLType::listOf($this->getParameterType($type->getCollectionValueTypes()[0])),
+            Type::BUILTIN_TYPE_ITERABLE => GraphQLType::listOf($this->getParameterType($type->getCollectionValueTypes()[0])),
+            Type::BUILTIN_TYPE_OBJECT => GraphQLType::string(),
+            default => GraphQLType::string(),
+        };
     }
 
     /**
@@ -433,6 +529,7 @@ final class FieldsBuilder implements FieldsBuilderInterface, FieldsBuilderEnumIn
         if (null === $resourceClass) {
             return $args;
         }
+        $d = false;
 
         foreach ($resourceOperation->getFilters() ?? [] as $filterId) {
             if (!$this->filterLocator->has($filterId)) {
@@ -450,9 +547,9 @@ final class FieldsBuilder implements FieldsBuilderInterface, FieldsBuilderEnumIn
                 }
             }
 
-            foreach ($this->filterLocator->get($filterId)->getDescription($entityClass) as $key => $value) {
-                $nullable = isset($value['required']) ? !$value['required'] : true;
-                $filterType = \in_array($value['type'], Type::$builtinTypes, true) ? new Type($value['type'], $nullable) : new Type('object', $nullable, $value['type']);
+            foreach ($this->filterLocator->get($filterId)->getDescription($entityClass) as $key => $description) {
+                $nullable = isset($description['required']) ? !$description['required'] : true;
+                $filterType = \in_array($description['type'], Type::$builtinTypes, true) ? new Type($description['type'], $nullable) : new Type('object', $nullable, $description['type']);
                 $graphqlFilterType = $this->convertType($filterType, false, $resourceOperation, $rootOperation, $resourceClass, $rootResource, $property, $depth);
 
                 if (str_ends_with($key, '[]')) {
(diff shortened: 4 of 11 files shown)


mrossard:
Starting this i great, i've been thinking about that and have some ideas, but couldn't find the time...here are some quick unorganiszed thoughts :

The direction i wanted to go was to first implement an AST-based filter (independant from the actual query language), then specific filters for one or more query language (I started looking at Symfony's Expression Language but anything would work). 
It would then be pretty simple to rewrite the existing filters as "ast builders".

It would need to be broken into quite a lot of smaller parts (further thinking needed here ), for example dealing with individual operations ( = / < > ...), which could provide extension points, allowing devs to add their custom operations...

I think in the new system i'd be better to force explicit enabling of filtering on properties, and if the target property is a resource, only the IRI should be expected (explicitly filtering on resource.id would still be possible after all...) : you would either have `stuff.property = /someResource/1` or `stuff.property.id = 1`, without any ambiguity. 
This might require some work on the BC layer to create a correct AST, but it seems very doable (`stuff.property=string` would likely just need to be transformed into `stuff.property=string or stuff.property.id=string`).

That's about it for my thoughts on it as of right now!

mrossard:
Having the possibility to specify the types / validation method for parameters would be a very nice improvement!

I'd still keep it inside the filter definition though, something that could look like this :
```php
#[ApiFilter(SomeFilter::class, properties:[
        'exposedName'=> new FilterProperty(
            target: 'realPropertyPath',
            type: Type::BUILTIN_TYPE_STRING,
            parser: self:nameParser(...),
            validation: self::nameValidator(...),
        )
    ]
)]
```

soyuka:
We should not have a strong link between a filter (parameter) and a property.

mrossard:
> We should not have a strong link between a filter (parameter) and a property.

well there needs to be a way to specify this some way or other, or we can't build the ORM query, can we? Or am i missing the point?

soyuka:
Because it's locking filters to behave like a parameter is always tight to a property which isn't the case. Something preferable would be: 

```
#[ApiFilter(SomeFilter::class, properties:[
        'exposedName'
    ], filters: [
        new FilterProperty(
            target: 'realPropertyPath',
            type: Type::BUILTIN_TYPE_STRING,
            parser: self:nameParser(...),
            validation: self::nameValidator(...),
        )
    ]
)]
```

mrossard:
> Because it's locking filters to behave like a parameter is always tight to a property which isn't the case. Something preferable would be:
> 
> ```
> #[ApiFilter(SomeFilter::class, properties:[
>         'exposedName'
>     ], filters: [
>         new FilterProperty(
>             target: 'realPropertyPath',
>             type: Type::BUILTIN_TYPE_STRING,
>             parser: self:nameParser(...),
>             validation: self::nameValidator(...),
>         )
>     ]
> )]
> ```

Then where do you see the link between the "exposedName" query parameter and the "realPropertyPath" being made? I'm a bit confused here. 
When you're trying to build the backend query you'll need to know somehow that the "exposedName=someValue" in the query means you have to filter on records that match "someValue"  for "realPropertyPath"...how would you do that?

soyuka:
This is why I have a `property` but also an array `attributes` in my `Parameter` suggestion so that we can add a link, but the link itself is not at the heart of the filter definition. This will help with #5495 for example, which right now is quite hard to fix.

mrossard:
> This is why I have a `property` but also an array `attributes` in my `Parameter` suggestion so that we can add a link, but the link itself is not at the heart of the filter definition. This will help with #5495 for example, which right now is quite hard to fix.

I think I'm actually seing things very differently from what you envision, which explains how i have trouble wrapping my head around the logic of it. 
For example I don't get the need for multiple filters, if the target is some query language. I was seeing just one SearchFilter with multiple query language parsers, since the call would be something like ```GET "/my_resources?query=(stuff=value or otherstuff=whatever)"```.
What i thought we'd need would look more like this :

````php
#[ApiFilter(NewSearchFilter::class,
    properties         : [
        'name' => new FilterProperty(
            target: 'searchableName',
            type  : Type::BUILTIN_TYPE_STRING,
            parser: NameStringSearchFormatter::class
        ),
        'linkedResource' => new FilterProperty(
            target: 'otherResourceCollection',
            type  : Type::BUILTIN_TYPE_STRING,
            parser: IriToOtherResourceConverter::class
        )],
    queryLanguageParser: new ExpressionLanguageParser(),
    parameterName: 'query'
)]
class SomeResource
{
    public int $id;
    public string $name;
    public string $searchableName;
    public array $otherResourceCollection;
}

class NewSearchFilter
{
    function __construct(protected array $properties, protected QueryLanguageParserInterface $queryLanguageParser, protected string $parameterName)
    {
    }

    function filter($property, $value, QueryBuilder $queryBuilder, ...)
    {
        if($property !== $this->parameterName){
            return;
        }
        $ast = $this->queryLanguageParser->parse($value, $this->properties);
        return (new AstFilter())->filter($queryBuilder, $ast);
    }
}
````
The FilterProperty could also have a link property instead of the target, but I feel like it would/will be quite a headache to implement - the correct way to establish the link on the ORM depends on the "state" of the AST (see the code i sent you on slack, typically you're doing simple joins or left joins depending on what happened before).

**Comment by mrossard** [comment]
...and i'm actually only talking about search filter whern you're talking about all filters, which can't help 😅

**Comment by soyuka** [comment]
Not a problem, in fact I want to cover most of the possibilities and doing a search filter like you have in mind should be possible but there are some things that aren't portable in your example. For example `queryLanguageParser`, `parameterName` are things that should belong to your filter `NewSearchFilter` and not to our `ApiFilter`. This makes a huge difference. 

For API Platform core, we have a SearchFilter, but I'd like to improve the current search filter by splitting the logic, mainly to resolve typing issues for example our current search filter does:
- IRI filtering
- id filtering (should work with uids)
- text match (starts with, ends with, case sensitive etc. these are options to apply for each parameter)
- exact match 
- date doesn't work as it should use the date range filter, but why couldn't it be used when we found a date type?

In my opinion, we should find a way to compose with filters working on a single parameter (`search`) or multiple parameters (parameterName equals propertyName). This is why I don't want to correlate a property with a filter, but instead work on a Parameter (or list of parameters) each applying a list of filters.

**Comment by aegypius** [comment]
I think the `queryLanguageParser` name is not appropriate in all cases. A parameter can be found in `path`, `query` are `header` according to OpenAPI Spec the way we parse value can be the same for every location. 

I am still advocating to split the logic between parameter concept and filters.

The nature of the value of the parameter can in some case use a filter or an other, for exemple 
```
?createdAt=2023-11-28
```
can be translated to : `[2023-11-28T00:00:00, 2023-11-28T23:59:59]`  maybe a range filter is more appropriate.

Declaring this type of filter using a Filter centric approche is quite cumbersome.

Does it make sense ?

**Comment by mrossard** [comment]
> Not a problem, in fact I want to cover most of the possibilities and doing a search filter like you have in mind should be possible but there are some things that aren't portable in your example. For example `queryLanguageParser`, `parameterName` are things that should belong to your filter `NewSearchFilter` and not to our `ApiFilter`. This makes a huge difference.

indeed, that was a mistake - i intended something  like ```#[NewSearchfilter(..)]```
 
> For API Platform core, we have a SearchFilter, but I'd like to improve the current search filter by splitting the logic, mainly to resolve typing issues for example our current search filter does:
> 
> * IRI filtering
> * id filtering (should work with uids)
> * text match (starts with, ends with, case sensitive etc. these are options to apply for each parameter)
> * exact match
> * date doesn't work as it should use the date range filter, but why couldn't it be used when we found a date type?

That's where the explicit typing would be very useful, yes, with some kind of parser/transformer specified to transform the text representation from the query into what we actually need (that's what i intended with the "parser" parameter).

Exact match/starts with/... would be different operators in the AST, which would make it fairly easy to have separate logic for them.

> 
> In my opinion, we should find a way to compose with filters working on a single parameter (`search`) or multiple parameters (parameterName equals propertyName). 

If I understand what you mean correctly, the second one is easily translated into the first one - ```/resources?field=value&field2=value2``` is just ```/resources?search=(field=value and field2=value2). It could use the same underlying code without problems

> This is why I don't want to correlate a property with a filter, but instead work on a Parameter (or list of parameters) each applying a list of filters.

Then you'd just add some ```#[Parameter('someparam', filters:[...])``` to your resources? It feels weird to me to separate the declaration of the filters (```#[Get(filters: [new SearchFilter(), new UidFilter()])```) and the definition of what thoses filters can use.

**Comment by mrossard** [comment]
> I think the `queryLanguageParser` name is not appropriate in all cases. A parameter can be found in `path`, `query` are `header` according to OpenAPI Spec the way we parse value can be the same for every location.

the "query" in "queryLanguageParser" was not meant for how the parameter is passed, to specify which "query language" is to be used, if several are supported.

> 
> I am still advocating to split the logic between parameter concept and filters.

I must be missing something...what other uses will the parameters have? If they're only used as filter parameters i'd rather have them tied to the filters that use them, this way one look is enough to know what a filter can do. It feels like better DX to me, but i've been known not to have the most popular opinion on such things... ;)

> 
> The nature of the value of the parameter can in some case use a filter or an other, for exemple
> 
> ```
> ?createdAt=2023-11-28
> ```
> 
> can be translated to : `[2023-11-28T00:00:00, 2023-11-28T23:59:59]` maybe a range filter is more appropriate.
> 
> Declaring this type of filter using a Filter centric approche is quite cumbersome.

To me it should be a specific operation, there is no clear indication as to whether a date without time should match any time for that date, or just that date at 00:00:00, or...?
There are probably multiple ways to deal with that (typing with separate "Date" and "Datetime" types for example, matching a "Date" parameter to a "Datetime" field could trigger a "between 0:00 and 23:59" logic), but i don't see how this impacts the way it is declared...?

File: docs/adr/0006-filters.md [hunk 1]
```diff
@@ -0,0 +1,200 @@
...
+                $operation->withParameter($key, $parameter->withFilter(UuidFilter::class));
+            }
+        }
+
+        return $operation;
+    }
+}
+```
+
+3. Validate parameters through the QueryParameterValidator.
```
aegypius, line 189: We renamed the component here  https://github.com/api-platform/core/pull/6080
```suggestion
3. Validate parameters through the ParameterValidator. 
```

docs/adr/0006-filters.md
```diff
@@ -0,0 +1,253 @@
...
+use ApiPlatform\OpenApi;
+
+final class Parameter {
+    public string $key;
+    public \ArrayObject schema;
+    public array $context;
+    public OpenApi\Parameter $openApi;
+    public string|callable provider(): Operation;
+    // filter service id
+    public string $filter;
```
dunglas, line 113: Maybe could we allow inline filter definitions?

```suggestion
    public string|callable $filter;
```

dunglas (approved):
Good work!

docs/adr/0006-filters.md
soyuka: I thought about that but with which interface? at first I wanted something like `mixed $value, Parameter $parameter, array $context` but it doesn't ally with `FilterInterface`

docs/adr/filter-notes
```diff
@@ -0,0 +1,61 @@
...
+TODO:
+see SerializerFilterContextBuilder: public function apply(Request $request, bool $normalization, array $attributes, array &$context): void;
+maybe something like:
+
+```
+class SerializerFilterInterface {
+    public function getNormalizationContext(...);
+    public function getDenormalizationContext(...);
+}
+```
```
soyuka: oups commited my notes, removing them in the next commit

docs/adr/filter-notes
lyrixx, line 1: missing .md suffix?

docs/adr/0006-filtering-system-and-parameters.md
```diff
@@ -0,0 +1,253 @@
...
+    public function getDescription(string $resourceClass): array;
+}
+```
+
+The idea of this ADR is to find a way to introduce more functionalities to API Platform filters such as: 
+
+- document query parameters for hydra, JSON Schema (OpenAPI being an extension of JSON Schema). 
+- pilot the query parameter validation (current QueryParameterValidator bases itself on the given documentation schema) this is good but lacks flexibility when you need custom validation (created by @jdeniau)
+- compose with filters, which will naturally help creating an or/and filter
+- remove the relation between a query parameter and a property (they may have different names [#5980][pull/5980]), different types, a query parameter can have no link with a property (order filter)
```
dunglas, line 60: We shouldn't totally remove it but make it optional. We could take inspiration of http://www.hydra-cg.com/spec/latest/core/#supported-property-data-source (that we need to implement anyway, so we'll still an optional need a mapping between parameters and properties).

docs/adr/0006-filtering-system-and-parameters.md
```diff
@@ -0,0 +1,253 @@
...
+        return $operation;
+    }
+}
+```
+
+3. Validate parameters through the ParameterValidator.
+
+### Filters
+
+Filters should remain mostly unchanged, the current informations about the `property` to filter should be specified inside a `Parameter`'s `context`.
```
dunglas, line 204: To me properties mapping should remain a first class citizen because of Hydra and because it's the most common use case.

dunglas (changes requested):

src/Metadata/HeaderParameter.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata;
+
+class HeaderParameter extends Parameter implements HeaderParameterInterface
```
vincentchalamon, line 16: Would be safer to declare it as `@experimental` in next minor version (e.g.: 3.4.0), and remove the `@experimental` on the after next minor version (e.g.: 3.5.0).
Same for HeaderParameterInterface, QueryParameter, QueryParameterInterface and `parameters` argument on Metadata/*

docs/adr/0006-filtering-system-and-parameters.md
```diff
@@ -0,0 +1,253 @@
...
+During the `Provider` phase (`RequestEvent::REQUEST`), we could use a `ParameterProvider`:
+
+```php
+/**
+ * Optionnaly transforms request parameters and provides modification to the current Operation.
+ *
+ * @implements ProviderInterface<HttpOperation>
+ */
+interface ParameterProvider extends ProviderInterface {
+    public function provider(Operation $operation, array $uriVariables = [], array $context = []): HttpOperation;
```
vincentchalamon, line 148: Why not `Operation` instead of `HttpOperation`?

src/Metadata/Parameter.php
```diff
@@ -0,0 +1,119 @@
...
+        $self = clone $this;
+        $self->filter = $filter;
+
+        return $self;
+    }
+
+    /**
+     * @param array<string, mixed> $extraProperties
+     */
+    public function withExtraProperties(array $extraProperties = []): static
```
vincentchalamon, line 112: IMO withers should not include default values:
```suggestion
    public function withExtraProperties(array $extraProperties): static
```

src/Metadata/Parameter.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return $this->filter;
+    }
+
+    public function getExtraProperties(): ?array
+    {
+        return $this->extraProperties;
+    }
+
+    public function withKey(?string $key): static
```
vincentchalamon, line 66: How can `key` be null?

src/Metadata/Parameter.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return $this->provider;
+    }
+
+    public function getFilter(): mixed
+    {
+        return $this->filter;
+    }
+
+    public function getExtraProperties(): ?array
```
vincentchalamon, line 61:
```suggestion
    public function getExtraProperties(): array
```

src/Metadata/Parameter.php
```diff
@@ -0,0 +1,119 @@
...
+
+    public function withOpenApi(OpenApi\Model\Parameter $openApi): static
+    {
+        $self = clone $this;
+        $self->openApi = $openApi;
+
+        return $self;
+    }
+
+    public function withProvider(mixed $provider): static
```
vincentchalamon, line 93:
```suggestion
    /**
     * @param ProviderInterface|string $provider
     */
    public function withProvider(mixed $provider): static
```

src/Metadata/Parameter.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        $self = clone $this;
+        $self->provider = $provider;
+
+        return $self;
+    }
+
+    public function withFilter(mixed $filter): static
```
vincentchalamon, line 101:
```suggestion
    /**
     * @param FilterInterface|string $filter
     */
    public function withFilter(mixed $filter): static
```

src/Metadata/Parameter.php
```diff
@@ -0,0 +1,119 @@
...
+
+namespace ApiPlatform\Metadata;
+
+use ApiPlatform\OpenApi;
+use ApiPlatform\State\ProviderInterface;
+
+abstract class Parameter
+{
+    /**
+     * @param array<string, mixed>          $extraProperties
```
vincentchalamon, line 22:
```suggestion
     * @param \ArrayObject<string,mixed>|null $schema
     * @param array<string, mixed>          $extraProperties
```

src/Serializer/Parameter/GroupParameter.php
vincentchalamon, line 1: Empty file!
+ missing interface for consistency with query and header parameters

vincentchalamon (changes requested):

docs/adr/0006-filtering-system-and-parameters.md
soyuka: indeed I used HttpOperation inside the implementation

soyuka, line 148:
```suggestion
    public function provider(HttpOperation $operation, array $uriVariables = [], array $context = []): HttpOperation;
```

PR #6028, opened 2023-12-07
fix: normalize datetime values when used as resource identifier

emmanuel-averty:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes api-platform/api-platform#2444 
| License       | MIT
| Doc PR        | N/A

This PR allows to convert to string any `\DateTimeInterface` contained in resource identifiers.

### DIFF
--- a/src/Metadata/Tests/IdentifiersExtractorTest.php
+++ b/src/Metadata/Tests/IdentifiersExtractorTest.php
@@ -260,10 +260,9 @@ class IdentifiersExtractorTest extends TestCase
                 'id' => (new Link())
                     ->withParameterName('id')
                     ->withFromClass(DummyWithDateTimeIdentifier::class)
-                    ->withIdentifiers(['id'])
+                    ->withIdentifiers(['id']),
             ])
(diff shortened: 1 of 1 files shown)


p-pichet:
Hello. Is this fix is still of actuality ? 

And how we can set the format wanted ?

soyuka:
There should be a https://github.com/api-platform/core/blob/main/src/Api/UriVariableTransformer/DateTimeUriVariableTransformer.php I don't understand why this doesn't already work, use your own transformer for custom formats.

p-pichet:
> There should be a https://github.com/api-platform/core/blob/main/src/Api/UriVariableTransformer/DateTimeUriVariableTransformer.php I don't understand why this doesn't already work, use your own transformer for custom formats.

Hello. Thanks for the comment. 
In my case, it's not to convert an iri to a datetime but when we link an item with the identifier to generate the iri.

soyuka:
we need a failing test case before trying to fix this.

PR #6029, opened 2023-12-08
fix: remove link to ApiDocumentation when doc is disabled

jderusse:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | /
| License       | MIT
| Doc PR        | 

The `HydraLinkProcessor` injects links to ApiDocumentation in the response header even when the doc is disabled.

This PR remove the service (like it was the case with the legacy `AddLinkHeaderListener`) when the doc is not enabled.


soyuka:
thanks @jderusse !

PR #6032, opened 2023-12-09
ci: conflict sebastian/comparator

soyuka:


composer.json
```diff
@@ -95,7 +95,8 @@
         "doctrine/mongodb-odm": "<2.4",
         "doctrine/persistence": "<1.3",
         "symfony/var-exporter" : "<6.1.1",
-        "phpunit/phpunit": "<9.5",
+		"phpunit/phpunit": "<9.5",
```
dunglas, line 98: This should work even without this line.

soyuka: yes mb I was doing many things at once ^^

soyuka: actually we need this for our lowest tests I'll probably move this to the ci though

PR #6038, opened 2023-12-12
GraphQL: Nested Collections

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -

We sometimes use JSON arrays to persist a collection of Value Objects which are not an entity. So this collection should be treated like a property of the parent entity. On the GraphQL queries fetching this collection was not an easy task. We had to create some special resolvers which try to fetch the ID of the parent entity out of the `$context`, load it and determine the collection from there.

So I decided to spend this PR that utilizes the `nested` property of the `ApiProperty` and uses the collection already loaded with the entity when no own resolver is given. I added a Behat test that covers that new scenario.

### DIFF
--- a/features/graphql/query.feature
+++ b/features/graphql/query.feature
@@ -22,7 +22,7 @@ Feature: GraphQL query support
 
   @createSchema
   Scenario: Retrieve an item with different relations to the same resource
-    Given there are 2 multiRelationsDummy objects having each a manyToOneRelation, 2 manyToManyRelations and 3 oneToManyRelations
+    Given there are 2 multiRelationsDummy objects having each a manyToOneRelation, 2 manyToManyRelations, 3 oneToManyRelations and 4 embeddedRelations
     When I send the following GraphQL request:
     """
     {
@@ -49,6 +49,16 @@ Feature: GraphQL query support
             }
           }
         }
+        nestedCollection {
+          name
+        }
+        nestedPaginatedCollection {
+          edges{
+            node {
+              name
+            }
+          }
+        }
       }
     }
     """
@@ -67,6 +77,15 @@ Feature: GraphQL query support
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[1].node.id" should not be null
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[0].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[2].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
+    And the JSON node "data.multiRelationsDummy.nestedCollection[0].name" should be equal to "NestedDummy1"
+    And the JSON node "data.multiRelationsDummy.nestedCollection[1].name" should be equal to "NestedDummy2"
+    And the JSON node "data.multiRelationsDummy.nestedCollection[2].name" should be equal to "NestedDummy3"
+    And the JSON node "data.multiRelationsDummy.nestedCollection[3].name" should be equal to "NestedDummy4"
+    And the JSON node "data.multiRelationsDummy.nestedPaginatedCollection.edges" should have 4 element
+    And the JSON node "data.multiRelationsDummy.nestedPaginatedCollection.edges[0].node.name" should be equal to "NestedPaginatedDummy1"
+    And the JSON node "data.multiRelationsDummy.nestedPaginatedCollection.edges[1].node.name" should be equal to "NestedPaginatedDummy2"
+    And the JSON node "data.multiRelationsDummy.nestedPaginatedCollection.edges[2].node.name" should be equal to "NestedPaginatedDummy3"
+    And the JSON node "data.multiRelationsDummy.nestedPaginatedCollection.edges[3].node.name" should be equal to "NestedPaginatedDummy4"
 
   @createSchema @!mongodb
   Scenario: Retrieve an item with child relation to the same resource
--- a/src/GraphQl/Resolver/Factory/ResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ResolverFactory.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\GraphQl\Resolver\Factory;
 use ApiPlatform\Metadata\GraphQl\Mutation;
 use ApiPlatform\Metadata\GraphQl\Operation;
 use ApiPlatform\Metadata\GraphQl\Query;
+use ApiPlatform\State\Pagination\ArrayPaginator;
 use ApiPlatform\State\ProcessorInterface;
 use ApiPlatform\State\ProviderInterface;
 use GraphQL\Type\Definition\ResolveInfo;
@@ -33,6 +34,11 @@ class ResolverFactory implements ResolverFactoryInterface
         return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operation) {
             // Data already fetched and normalized (field or nested resource)
             if ($body = $source[$info->fieldName] ?? null) {
+                // special treatment for nested resources without a resolver/provider
+                if ($operation instanceof Query && $operation->getNested() && !$operation->getResolver() && !$operation->getProvider()) {
+                    return $this->resolve($source, $args, $info, $rootClass, $operation, new ArrayPaginator($body, 0, \count($body)));
+                }
+
                 return $body;
             }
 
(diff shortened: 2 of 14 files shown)


soyuka (approved):
needs #6032 to run the CI

jotwea:
> needs #6032 to run the CI

Updated the PR. Still some jobs fail because of deprecations, but this is not related to the changes in this PR. Maybe it can be merged anyway

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -110,8 +110,8 @@ public function normalize(mixed $object, string $format = null, array $context =
      */
     protected function normalizeCollectionOfRelations(ApiProperty $propertyMetadata, iterable $attributeValue, string $resourceClass, ?string $format, array $context): array
     {
-        // to-many are handled directly by the GraphQL resolver
-        return [];
+        // to-many are handled directly by the GraphQL resolver, but also check nested attributes for item query
+        return 'item_query' === $context['graphql_operation_name'] && !\array_key_exists('edges', $context['attributes']) ? [...$attributeValue] : [];
```
alanpoulain, line 114: I find it ugly to check the existence of `edges`, can't we find another solution to do it?

alanpoulain:
In my opinion, since it's a collection it should follow the Relay specification and have `edges` and `nodes`.

src/GraphQl/Serializer/ItemNormalizer.php
jotwea: I am with you, was also not happy with this approach. I could not find another way without passing other parameters to this method. So I am open for hints. An alternative I first thought creating and passing the `operation` to this method so we can there also check for `operation->getNested()`. Should I try to do it this way?

alanpoulain: Don't you have the operation in the context? But yes this approach would be better.

jotwea: I found a way to base the check on the operation. This is way more elegant. Thx for your hint.

jotwea:
> In my opinion, since it's a collection it should follow the Relay specification and have `edges` and `nodes`.

Works now. Can be controlled via `paginationEnabled` flag

alanpoulain:
Nice, maybe could you add one case with the pagination enabled too?

jotwea:
> Nice, maybe could you add one case with the pagination enabled too?

Updated the tests. We now have nested collection with and without pagination enabled.

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -110,6 +111,25 @@ public function normalize(mixed $object, string $format = null, array $context =
...
     {
+        if ('item_query' === $context['graphql_operation_name'] && $this->resourceMetadataCollectionFactory) {
+            // check for nested collection
+            $operation = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation('collection_query');
+            if ($operation instanceof Query && $operation->getNested() && !$operation->getResolver() && !$operation->getProvider()) {
+                if (false === $operation->getPaginationEnabled()) {
+                    return [...$attributeValue];
+                }
+
+                $data = ['edges' => []];
```
alanpoulain, line 122: I'm not sure to understand why the serialization of the edges / nodes is done here. It's already done in the `NormalizeProcessor`, why doesn't it pass through it?

alanpoulain: Answer to myself: probably because it's not a resource but a JSON array. However that means we lose what is done in the processor (page information for instance) or we need to duplicate the logic.
IMO we need to find a solution before merging this PR.

jotwea: Maybe we should go one step back and discuss the scope:

As I mentioned in the description of this PR, this feature is to simplify resolving a nested collection and treat them more like **properties** than relations. The consequence was that also the edges / nodes representation was not in my mind, because this is not for properties. I first thought it is easy to have also the edges / nodes support as you demanded and have both opportunities (controlled via `paginationEnabled` flag). But maybe we should not provide that for this case. My proposal would be to revert the latest changes.

Would like to hear your thoughts @alanpoulain.

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -110,6 +111,25 @@ public function normalize(mixed $object, string $format = null, array $context =
      */
     protected function normalizeCollectionOfRelations(ApiProperty $propertyMetadata, iterable $attributeValue, string $resourceClass, ?string $format, array $context): array
     {
+        if ('item_query' === $context['graphql_operation_name'] && $this->resourceMetadataCollectionFactory) {
+            // check for nested collection
+            $operation = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation('collection_query');
```
soyuka, line 116:
```suggestion
            $operation = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation(forceCollection: true);
``` 
mhh maybe we need something to fetch only graphql operations?

features/graphql/query.feature
```diff
@@ -49,6 +49,16 @@ Feature: GraphQL query support
             }
           }
         }
+        nestedCollection {
+          name
+        }
```
soyuka, line 54: you can't fetch a collection without using edges in graphql usually or can you?

jotwea: you can control it for all collections via the `paginationEnabled` flag. With the flag enabled you get edges / nodes structure. With the flag disabled you get the result in a `collection` node

src/GraphQl/Serializer/ItemNormalizer.php
jotwea: using `forceCollection: true` is not identical. It returns the `GetCollection` and not the `QueryCollection`

src/GraphQl/Serializer/ItemNormalizer.php
alanpoulain: I don't see why it should be treated like properties and not like a collection. It should have all the possibilities of a classical relation.

jotwea: Found my mistake. I started creating this feature on 3.2 and added a special case to the `CollectionResolverFactory`. On the main branch the general `ResolverFactory` is already used. So I added the special case also there and voilà: The changes in `ItemNormalizer` could have been reverted. Thx for your support.

src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -45,23 +51,28 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
                 return null;
             }
 
-            // Handles relay nodes
-            $operation ??= new Query();
+            return $this->process($source, $args, $info, $rootClass, $operation, null);
+        };
+    }
+
+    private function process(?array $source, array $args, ResolveInfo $info, string $rootClass = null, Operation $operation = null, mixed $body)
```
alanpoulain, line 58: Resolve instead of process? Since it will use the provider and the processor? Otherwise the PR seems good :slightly_smiling_face:

jotwea: Done

alanpoulain (approved):

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -110,6 +111,14 @@ public function normalize(mixed $object, string $format = null, array $context =
      */
     protected function normalizeCollectionOfRelations(ApiProperty $propertyMetadata, iterable $attributeValue, string $resourceClass, ?string $format, array $context): array
     {
+        if ('item_query' === $context['graphql_operation_name'] && $this->resourceMetadataCollectionFactory) {
```
soyuka, line 114: not a huge fan of this, can we use `$operation instanceof Query && !$operation instanceof CollectionOperationInterface`

soyuka: also put `$this->resourceMetadataCollectionFactory` first

src/GraphQl/Serializer/ItemNormalizer.php
soyuka: Then we need to loop over the resources, I'm willing to add an argument `forceGraphQl: true` to the `getOperation` as I feel this is missing right now.

src/GraphQl/Serializer/ItemNormalizer.php
jotwea: You are right, this is not pretty nice here. We do not have the `operation` yet, it is not in the `$context` - only the 'graphql_operation_name' is already there. Do not know exactly how to change that. Could you assist?

src/GraphQl/Serializer/ItemNormalizer.php
jotwea: Added argument `forceGraphQl`. Works fine

src/GraphQl/Serializer/ItemNormalizer.php
jotwea: Refactored and now works without this ugly check

src/GraphQl/Serializer/ItemNormalizer.php
```diff
@@ -110,6 +111,14 @@ public function normalize(mixed $object, string $format = null, array $context =
...
     protected function normalizeCollectionOfRelations(ApiProperty $propertyMetadata, iterable $attributeValue, string $resourceClass, ?string $format, array $context): array
     {
+        if ($this->resourceMetadataCollectionFactory) {
+            // check for nested collection
+            $operation = $this->resourceMetadataCollectionFactory->create($resourceClass)->getOperation(forceCollection: true, forceGraphQl: true);
+            if ($operation instanceof Query && $operation->getNested() && !$operation->getResolver() && !$operation->getProvider()) {
+                return [...$attributeValue];
+            }
+        }
+
```
soyuka, line 121:
```suggestion
        // check for nested collection
        $operation = $this?->resourceMetadataCollectionFactory->create($resourceClass)->getOperation(forceCollection: true, forceGraphQl: true);
        if ($operation && $operation instanceof Query && $operation->getNested() && !$operation->getResolver() && !$operation->getProvider()) {
            return [...$attributeValue];
        }

```

soyuka:
I'll solve phpstan in another patch thanks @jotwea @alanpoulain !

PR #2594, opened 2019-03-09
Fix test suite compatibility with webonyx/graphql-php v0.13.1

antograssiot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

PR #6050, opened 2023-12-15
Fix: Inconsistency in Context Concatenation with fetch_data

anoziere:
| Q             | A
| ------------- | ---
| Branch?       | current stable version branch for bug fixes
| Tickets       | Closes [#2594](https://github.com/api-platform/api-platform/issues/2594)
| License       | MIT
| Doc PR        | 	N/A

I have noticed an inconsistency in the ItemNormalizer.php file regarding the handling of the $context array in the updateObjectToPopulate method.

In the try block, the context is concatenated with ['fetch_data' => true] when calling getResourceFromIri. However, in the corresponding catch block for LegacyInvalidArgumentException|InvalidArgumentException, the same concatenation with fetch_data is not performed when setting the OBJECT_TO_POPULATE in the context:

```php
// In try block
$context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getResourceFromIri(
    (string) $data['id'], 
    $context + ['fetch_data' => true]
);

// In catch block
$context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getResourceFromIri(
    $iri, 
    ['fetch_data' => true]
);
``` 

so I just added the context in the case of catch.

### DIFF
--- a/src/Serializer/ItemNormalizer.php
+++ b/src/Serializer/ItemNormalizer.php
@@ -76,18 +76,20 @@ class ItemNormalizer extends AbstractItemNormalizer
         try {
             $context[self::OBJECT_TO_POPULATE] = $this->iriConverter->getResourceFromIri((string) $data['id'], $context + ['fetch_data' => true]);
         } catch (InvalidArgumentException) {
-            $operation = $this->resourceMetadataCollectionFactory->create($context['resource_class'])->getOperation();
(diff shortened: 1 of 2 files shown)


soyuka:
I'm really unsure about that, also you're using `json` right?

soyuka:
had to add a test and reword your commit please look at our commit rules inside CONTRIBUTING next time, thanks for the patch!

PR #5505, opened 2023-03-28
Fix invalid OpenAPI schema : missing constraint and empty array 

Zowac:
| Q             | A
| ------------- | ---
| Branch?       | 3.1 
| Tickets       | #5414 #5413  
| License       | MIT

PR #6057, opened 2023-12-19
fix(symfony): do not use metadata when merging schema constraints in Collection constraint

### LINKED ISSUE #5413: PropertySchemaCollectionRestriction generates invalid OpenAPI schema
**API Platform version(s) affected**: main

**Description**  
When `PropertySchemaCollectionRestriction` fails to identify nested constraint restrictions the generated OpenAPI schema might be invalid.

**How to reproduce**  
Given this validation rule:

```
<property name="someArray">
  <constraint name="Collection">
    <option name="fields">
      <value key="someField">
        <constraint name="Optional">
          <constraint name="SomeConstraintWithoutMatchingRestriction"/>
        </constraint>
      </value>
    </option>
  </constraint>
</property>
```

The generated schema will look like this:

```json
"Schema": {
  "type": "object",
  "properties": {
    "someArray": {
      "type": "object",
      "properties": {
        "someField": [],
      },
      "additionalProperties": false,
      "items": {
        "type": "string"
      }
   }
  }
}
```

The issue here is that `someField` is an empty array but should be an empty object.

### LINKED ISSUE #5414: PropertySchemaCollectionRestriction passes incorrect property metadata when merging constraint restrictions
**API Platform version(s) affected**: main

**Description**  
When `PropertySchemaCollectionRestriction` attempts to generate the restrictions for the sub-graph, it passes incorrect property metadata.

**How to reproduce**  
Given a validation rule like this:
```xml
<property name="someArray">
  <constraint name="Collection">
    <option name="fields">
      <value key="someField">
        <constraint name="Optional">
          <constraint name="Length">
             <option name="min">5</option>
          </constraint>
        </constraint>
      </value>
    </option>
  </constraint>
</property>
```

This would be the expected OpenAPI schema:

```json
"Schema": {
  "type": "object",
  "properties": {
    "someArray": {
      "type": "object",
      "properties": {
        "someField": {
          "minLength": 5
        },
      },
      "additionalProperties": false,
      "items": {
        "type": "string"
      }
    }
  }
}
```

Instead this is the schema generated:

```json
```json
"Schema": {
  "type": "object",
  "properties": {
    "someArray": {
      "type": "object",
      "properties": {
        "someField": [],
      },
      "additionalProperties": false,
      "items": {
        "type": "string"
      }
    }
  }
}
```

This is a) invalid because of the empty array but that one is filed with https://github.com/api-platform/core/issues/5413 but b) it misses the length restriction because of this check in `PropertySchemaLengthRestriction::supports()`:

```php
    public function supports(Constraint $constraint, ApiProperty $propertyMetadata): bool
    {
        return $constraint instanceof Length && null !== ($type = $propertyMetadata->getBuiltinTypes()[0] ?? null) && Type::BUILTIN_TYPE_STRING === $type->getBuiltinType();
    }
```

… in combination with `PropertySchemaCollectionRestriction::mergeConstraintRestrictions()`:

```php
    private function mergeConstraintRestrictions(Required|Optional $constraint, ApiProperty $propertyMetadata): array
    {
        $propertyRestrictions = [];
        $nestedConstraints = $constraint->constraints;

        foreach ($nestedConstraints as $nestedConstraint) {
            foreach ($this->restrictionsMetadata as $restrictionMetadata) {
                if ($restrictionMetadata->supports($nestedConstraint, $propertyMetadata) && !empty($nestedConstraintRestriction = $restrictionMetadata->create($nestedConstraint, $propertyMetadata))) {
                    $propertyRestrictions[] = $nestedConstraintRestriction;
                }
            }
        }

        return array_merge([], ...$propertyRestrictions);
    }
```

In `mergeConstraintRestrictions` `$propertyMetadata` is the property metadata of the container element (the array) and not the sub-element.

Since it’s an array, the sub elements are fundamentally unknowable. My hunch is, that the restriction implementations should not check at all for matching types and just assume that the validation rules are operating on the right type.

alanpoulain:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #5414, closes #5413
| License       | MIT
| Doc PR        | N/A

Supersedes https://github.com/api-platform/core/pull/5505.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -82,6 +82,7 @@ return (new PhpCsFixer\Config())
         'no_superfluous_elseif' => true,
         'no_superfluous_phpdoc_tags' => [
             'allow_mixed' => false,
+            'allow_unused_params' => true,
         ],
         'no_unset_cast' => true,
         'no_unset_on_property' => true,
--- a/src/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
+++ b/src/Symfony/Validator/Metadata/Property/Restriction/PropertySchemaChoiceRestriction.php
@@ -52,8 +52,7 @@ final class PropertySchemaChoiceRestriction implements PropertySchemaRestriction
 
         $restriction['type'] = 'array';
 
-        $builtInTypes = $propertyMetadata->getBuiltinTypes() ?? [];
-        $types = array_unique(array_map(fn (Type $type) => Type::BUILTIN_TYPE_STRING === $type->getBuiltinType() ? 'string' : 'number', $builtInTypes));
+        $types = array_values(array_unique(array_map(fn (mixed $choice) => \is_string($choice) ? 'string' : 'number', $choices)));
 
         if ($count = \count($types)) {
             if (1 === $count) {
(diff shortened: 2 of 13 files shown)


.php-cs-fixer.dist.php
```diff
@@ -82,6 +82,7 @@
         'no_superfluous_elseif' => true,
         'no_superfluous_phpdoc_tags' => [
             'allow_mixed' => false,
+            'allow_unused_params' => true,
```
alanpoulain, line 85: Not used in this PR, but is useful when adding a new commented parameter (BC promise).

soyuka (approved):

Issue #4764, opened 2022-05-19
[GraphQL] Make sure partial pagination works with page-based pagination


alanpoulain:
Method `serializePageBasedPaginatedCollection` should also check `PartialPaginatorInterface`.

xavierleune:
Can be closed (see #6068)

PR #6068, opened 2023-12-21
fix(graphql): remove count query if paginationInfo is not requested

xavierleune:
| Q             | A
| ------------- | ---
| Branch?       | current stable version branch for bug fixes
| License       | MIT

This pull request is a draft to fix a bug discovered on the current graphql implementation.

If the following request is sent:

```
{
  nameMetricFirstnameDepartmentOccurrences {
    collection {
      geoEntityDepartment {
        label
      }
    }
  }
}
```
The current implementation executes the count query on the database, to populate `paginationInfo` in the result until this information is dropped by the method `ReferenceExecutor::collectAndExecuteSubFields`.
I would expect the count query to be executed only if the `paginationInfo` is requested, like with the following request:

```
{
  nameMetricFirstnameDepartmentOccurrences {
    paginationInfo {
      itemsPerPage
      lastPage
      totalCount
    }
    collection {
      geoEntityDepartment {
        label
      }
    }
  }
}
```

I'm not quite sure that this is the best way to implement that fix, I open this pull request to start the discussion. There is probably some other implementations I should look at (like cursor based pagination, perhaps the NormalizeProcessor) and add some tests. In the meantime what do you think of this approach ?

Thanks,

### DIFF
--- a/src/GraphQl/Resolver/Stage/SerializeStage.php
+++ b/src/GraphQl/Resolver/Stage/SerializeStage.php
@@ -88,7 +88,7 @@ final class SerializeStage implements SerializeStageInterface
             } else {
                 $data = 'cursor' === $this->pagination->getGraphQlPaginationType($operation) ?
                     $this->serializeCursorBasedPaginatedCollection($itemOrCollection, $normalizationContext, $context) :
-                    $this->serializePageBasedPaginatedCollection($itemOrCollection, $normalizationContext);
+                    $this->serializePageBasedPaginatedCollection($itemOrCollection, $normalizationContext, $context);
             }
         }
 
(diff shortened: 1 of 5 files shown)


xavierleune:
I think this would fix api-platform/api-platform#2310 api-platform/api-platform#1996 and #4764 (after full implementation)

soyuka:
I like this approach, can you add a test?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -41,6 +41,10 @@ public function __construct(private readonly ReadStageInterface $readStage, priv
     public function __invoke(string $resourceClass = null, string $rootClass = null, Operation $operation = null): callable
     {
         return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operation): ?array {
+            if (!isset($info->getFieldSelection()['paginationInfo']) && $operation->getPaginationClientPartial()) {
```
xavierleune, line 44: I'm not sure if that's relevant to check `$operation->getPaginationClientPartial` but at least the behavior will follow the configuration

xavierleune:
@soyuka sure, I think this should do the trick, WDYT ?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
alanpoulain: It doesn't make sense to have `$operation->getPaginationClientPartial`.
If pagination client partial is enabled, the operation already has it.

src/GraphQl/Resolver/Stage/SerializeStage.php
```diff
@@ -175,14 +174,21 @@ private function serializeCursorBasedPaginatedCollection(iterable $collection, a
...
-        $data['paginationInfo']['lastPage'] = $collection->getLastPage();
-        $data['paginationInfo']['itemsPerPage'] = $collection->getItemsPerPage();
+        $data = [
+            'collection' => [],
+            'paginationInfo' => [
+                'itemsPerPage' => $collection->getItemsPerPage(),
+            ],
+        ];
+
+        if ($collection instanceof PaginatorInterface) {
```
alanpoulain, line 188: The serialized data change depending of the context, maybe we should check if it's the case for the GraphQL schema too: if pagination client partial is enabled, the `totalCount` and `lastPage` should be removed from the schema.

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
alanpoulain: It does not make sense to check the `$operation->getPaginationClientPartial`.
GraphQL is always client-wise: if the user has not asked the full pagination info then we can safely enable the partial pagination, even if client partial is not enabled.
Also why checking `paginationInfo` and not the specific full pagination fields (`totalCount`, `lastPage`)?

xavierleune:
Thanks for your feedback @alanpoulain 
As I've now a better understanding of the mechanics here, I think I can go further and offer a better implementation for the partial pagination and solve the bug described in this issue.

I think to control partial pagination in "paged" pagination in graphql, it would make sense to offer a `hasNextPage` in the paginationInfo. It would be a boolean to indicate if there is more results to the current query (so the client can query the next page).

I think this should be done only with a new interface `HasNextPagePaginatorInterface` describing only one method: `public function hasNextPage(): bool`.

So we will have the following `paginationInfo` available for the client:

```
paginationInfo {
    totalCount
    lastPage
    itemsPerPage
    hasNextPage
}
```

Each item may trigger one query to response, only if declared in the client query.

So in the case described in the current issue, no additionnal query would be triggerred:

```
{
  nameMetricFirstnameDepartmentOccurrences {
    collection {
      geoEntityDepartment {
        label
      }
    }
  }
}
```

If paginationInfo.totalCount or paginationInfo.lastPage is declared, the count query would be issued. If `hasNextPage` is declared (and the paginator implements  `HasNextPagePaginatorInterface`) a query to fetch the row +1 would be triggered. A better approach here could be to always fetch itemsPerPage + 1, count the number of rows (and drop the last row if that was an overfetch), but I'm not sure if that's easy or not.

alanpoulain:
Sure why not, maybe this interface could be used for REST as well. IMO it should be done in another PR though.

xavierleune:
> Sure why not, maybe this interface could be used for REST as well. IMO it should be done in another PR though.

I'll go with another PR for that part ^^

On the main topic of this pull request, beside the falling test (because of php-parser 5.0.0), I think we're good to go. @alanpoulain may you take a look please ?

src/GraphQl/Resolver/Factory/CollectionResolverFactory.php
```diff
@@ -49,6 +49,7 @@ public function __invoke(string $resourceClass = null, string $rootClass = null,
             $resolverContext = ['source' => $source, 'args' => $args, 'info' => $info, 'is_collection' => true, 'is_mutation' => false, 'is_subscription' => false];
 
             $collection = ($this->readStage)($resourceClass, $rootClass, $operation, $resolverContext);
+
```
soyuka, line 52:
```suggestion
```

src/GraphQl/Resolver/Stage/SerializeStage.php
```diff
@@ -78,7 +78,6 @@ public function __invoke(object|array|null $itemOrCollection, string $resourceCl
                 $data = $this->normalizer->normalize($itemOrCollection, ItemNormalizer::FORMAT, $normalizationContext);
             }
         }
-
```
soyuka, line 81: can you avoid empty line changes?

src/GraphQl/Resolver/Stage/SerializeStage.php
```diff
@@ -117,53 +116,61 @@ private function serializeCursorBasedPaginatedCollection(iterable $collection, a
             throw new \LogicException(sprintf('Collection returned by the collection data provider must implement %s or %s.', PaginatorInterface::class, PartialPaginatorInterface::class));
         }
 
+        $selection = $context['info']->getFieldSelection(1);
```
soyuka, line 119: are we sure `info` always exists?

src/GraphQl/Resolver/Stage/SerializeStage.php
xavierleune: yep, sorry for that

src/GraphQl/Resolver/Stage/SerializeStage.php
xavierleune: For calls in api-p, in the context of a collection, yes. The [ReadStage](https://github.com/api-platform/core/blob/main/src/GraphQl/Resolver/Stage/ReadStage.php#L78) uses it without any test. Should we enforce it here ?

xavierleune:
Hi @soyuka @alanpoulain do you have a new feedback on this pull request ? Maybe something I should improve before merging it ? Thanks !

soyuka:
can you rebase against 3.2 so we get proper coverage?

xavierleune:
@soyuka It seems that coverage is ok after rebase :+1:

soyuka (approved):

soyuka:
Thanks @xavierleune !

xavierleune:
@alanpoulain following this pull request, I opened the PR #6120 about the hasNextPagePaginatorInterface

Issue #6024, opened 2023-12-06
BC introduced in 3.1.15 when denormalizing a variable in a type than already been denormalized previously


MariusJam:
**API Platform version(s) affected**: ^3.1.15

**Description**  
Hi,

It took me ages to find the problem on this one but a part of our app is broken as soon as we upgrade to ^3.1.15. The BC has been indroduced with this PR : https://github.com/api-platform/core/pull/5663 and is linked to the modifications in src/Serializer/AbstractItemNormalizer.php.

The problem happens in a very specific case: if you try to denormalize manually an item of a type that has been already previously denormalized, the denormalization doesn't hydrate correctly the object. 

**How to reproduce**  

With : 

```php
<?php

namespace App\ApiResource;

use ApiPlatform\Metadata\ApiProperty;
use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\NotExposed;
use ApiPlatform\Metadata\Post;
use App\State\Processor\DummyProcessor;
use Symfony\Component\Serializer\Annotation\Groups;

#[ApiResource(operations: [
    new NotExposed(),
    new Post(denormalizationContext: ['groups' => ['dummy:write']], processor: DummyProcessor::class
    ),
])]
class Dummy
{
    public string $bar;

    #[ApiProperty(writableLink: true)]
    #[Groups(['dummy:write'])]
    public Foo $foo;
}
```

```php
<?php

namespace App\ApiResource;

use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Get;
use Symfony\Component\Serializer\Annotation\Groups;

#[ApiResource(operations: [new Get(normalizationContext: ['groups' => ['foo:read']],)])]
class Foo
{
    #[Groups(['foo:read'])]
    public string $bar;

    #[Groups(['dummy:write', 'foo:read'])]
    public string $name;
}
```

```php
<?php

namespace App\State\Processor;

use ApiPlatform\Metadata\Operation;
use ApiPlatform\State\ProcessorInterface;
use App\ApiResource\Dummy;
use App\ApiResource\Foo;
use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;

class DummyProcessor implements ProcessorInterface
{
    public function __construct(private readonly DenormalizerInterface $denormalizer)
    {
    }

    public function process(mixed $data, Operation $operation, array $uriVariables = [], array $context = [])
    {
        /**
[...]

PR #6071, opened 2023-12-29
fix: operation cache key without operation name

MariusJam:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | Closes https://github.com/api-platform/core/issues/6024
| License       | MIT

Following https://github.com/api-platform/core/pull/5663 , the `$operationCacheKey` generated in \ApiPlatform\Serializer\AbstractItemNormalizer::getFactoryOptions() does not include an operation_name when working on a relation context (e.g. denormalizing a child resource). This leads to `$operationCacheKey` equals to `$context['resource_class']` which is too generic and leads, at its turn, to buggy situations. For instance when manually denormalizing an object of the same class that `$context['resource_class']` but in a complete other context.

More information can be found in https://github.com/api-platform/core/issues/6024

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -580,7 +580,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
             $options['serializer_groups'] = (array) $context[self::GROUPS];
         }
 
-        $operationCacheKey = ($context['resource_class'] ?? '').($context['operation_name'] ?? '').($context['api_normalize'] ?? '');
+        $operationCacheKey = ($context['resource_class'] ?? '').($context['operation_name'] ?? $context['root_operation_name'] ?? '').($context['api_normalize'] ?? '');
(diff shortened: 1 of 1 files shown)


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -580,7 +580,7 @@ protected function getFactoryOptions(array $context): array
             $options['serializer_groups'] = (array) $context[self::GROUPS];
         }
 
-        $operationCacheKey = ($context['resource_class'] ?? '').($context['operation_name'] ?? '').($context['api_normalize'] ?? '');
+        $operationCacheKey = ($context['resource_class'] ?? '').($context['operation_name'] ?? $context['root_operation_name'] ?? '').($context['api_normalize'] ?? '');
```
soyuka, line 583: shouldn't we add `root_operation_name` as part of the key?

soyuka:
That's nice thanks, I've improved the patch with a test at #6073.

PR #6074, opened 2023-12-29
feat(metadata): headers configuration

### LINKED ISSUE #5933: Add `headers` to ApiResource and operations
No reasons we have this only for cache options...

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #5933
| License       | MIT
| Doc PR        | tbd

waiting for #6102


src/Metadata/HttpOperation.php
```diff
@@ -55,6 +55,7 @@ class HttpOperation extends Operation
      *     stale_while_revalidate?: int,
      *     stale-if-error?: int,
      * }|null $cacheHeaders {@see https://api-platform.com/docs/core/performance/#setting-custom-http-cache-headers}
+     * @param array<string, string>|null $headers
```
vincentchalamon, line 58:
```suggestion
     * @param array<string, mixed>|null $headers
```

soyuka: Headers are basically strings (https://httpwg.org/specs/rfc9110.html#fields.values) it looks like the HeaderBag of symfony also only types strings so I should rather enforce a string here no?

PR #6078, opened 2024-01-03
fix: In PHP, requests without content-type will result in an empty content-type, not null

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        | 

In PHP, requests without content-type will result in an empty string, not null. This PR modifies several checks for empty content-type in the codebase. They failed because they previously check for a null value, not an empty string.

Also related: https://github.com/symfony/symfony/pull/53380

### DIFF
--- a/src/State/Provider/ContentNegotiationProvider.php
+++ b/src/State/Provider/ContentNegotiationProvider.php
@@ -97,7 +97,8 @@ final class ContentNegotiationProvider implements ProviderInterface
      */
     private function getInputFormat(HttpOperation $operation, Request $request): ?string
     {
-        if (null === ($contentType = $request->headers->get('CONTENT_TYPE'))) {
+        $contentType = $request->headers->get('CONTENT_TYPE');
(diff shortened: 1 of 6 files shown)


soyuka:
Interesting, is there a way we can add a non-regression test? Also the commit should follow the commitlint conventions.

priyadi:
> Interesting, is there a way we can add a non-regression test?

I experienced the problem when my endpoints that accept empty request body stopped working under `event_listeners_backward_compatibility_layer: false`. That error came from `MainController`. I'll try coming up with a test for that.

The other two are the result of searching the entire code base for `->get('CONTENT_TYPE')`. I did not personally get any error from those.

> Also the commit should follow the commitlint conventions.

It should comply with the commitlint rules now.

priyadi:
Tests for the three changes have been added.

soyuka:
That's awesome thanks!

PR #6081, opened 2024-01-04
setup monorepo isolation rules

aegypius:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Declares isolation rules for splitted components using PHPArkitect (https://github.com/phparkitect/arkitect). 
This allows to track "hidden" dependencies between components while working on the mono-repo.

I tried to declare the rules how they are actually setup to spot the dependencies that forbid for now to use 
individual component instead of api-platform/core. 

When resolved, I think we should move api-platform/core to a "meta" component and use all individual components.


aegypius:
@soyuka I added CI check for interdepencies, should I add current violations in the baseline while we resolve this ?

composer.json
```diff
@@ -53,6 +53,7 @@
         "guzzlehttp/guzzle": "^6.0 || ^7.0",
         "jangregor/phpstan-prophecy": "^1.0",
         "justinrainbow/json-schema": "^5.2.1",
+        "phparkitect/phparkitect": "^0.3.25",
```
dunglas, line 56: Maybe should we use the phar to prevent future issue with direct dev deps?

soyuka:
thanks! We're going to fix errors in another pr.

PR #6084, opened 2024-01-05
GraphQL: Errors with nullish ManyToOne-Relation when using new ResolverFactory

### LINKED ISSUE #6083: GraphQl: Errors with nullish ManyToOne-Relation when using new ResolverFactory

**Description**  
`ItemResolverFactory` and `CollectionResolverFactory` are deprecated and will be removed with 4.0. So I disabled `event_listeners_backward_compatibility_layer` to use the new `ResolverFactory` and see if there are some problems with it. In general this worked fine except for one case. Many-To-One Relations that are allowed to be `null` and are using a custom resolver are behaving different with the new `ResolverFactory`. Formerly the property was just returned as `null` and the resolver got not called. Now the resolver gets called but failes because resolvers are not able to return `null`.

**How to reproduce**  
I created a PR that adds a Behat test for this case.

**Possible Solution**  

**Additional Context**  
I debugged the test case and found the cause, but was not able to fix it properly without breaking other test cases. The checks for already fetched data is different in https://github.com/api-platform/core/blob/main/src/GraphQl/Resolver/Factory/ResolverFactory.php#L36, and https://github.com/api-platform/core/blob/main/src/GraphQl/Resolver/Factory/ItemResolverFactory.php#L48

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #6083 
| License       | MIT
| Doc PR        | -

This PR introduces a new the test case especially for properties and Many-To-One-Relations being `null`. It fixes the issues relevant for 3.2 and additionally provides a fix for the nested collection feature which was recently introduced on the main branch. There is another PR https://github.com/api-platform/core/pull/6092 that contains only the fixes relevant for 3.2.

### DIFF
--- a/src/GraphQl/Resolver/Factory/ResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ResolverFactory.php
@@ -27,14 +27,13 @@ class ResolverFactory implements ResolverFactoryInterface
 {
     public function __construct(
         private readonly ProviderInterface $provider,
-        private readonly ProcessorInterface $processor,
-        private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory
+        private readonly ProcessorInterface $processor
     ) {
     }
 
-    public function __invoke(?string $resourceClass = null, ?string $rootClass = null, ?Operation $operation = null): callable
+    public function __invoke(?string $resourceClass = null, ?string $rootClass = null, ?Operation $operation = null, ?PropertyMetadataFactoryInterface $propertyMetadataFactory = null): callable
     {
-        return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operation) {
+        return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operation, $propertyMetadataFactory) {
             if (\array_key_exists($info->fieldName, $source ?? [])) {
                 $body = $source[$info->fieldName];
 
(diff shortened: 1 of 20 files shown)


src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -32,17 +33,19 @@ public function __construct(
...
-                // special treatment for nested resources without a resolver/provider
-                if ($operation instanceof Query && $operation->getNested() && !$operation->getResolver() && !$operation->getProvider()) {
-                    return $this->resolve($source, $args, $info, $rootClass, $operation, new ArrayPaginator($body, 0, \count($body)));
-                }
-
                 return $body;
             }
 
-            if (null === $resourceClass && \array_key_exists($info->fieldName, $source ?? [])) {
+            if ((null === $resourceClass || !is_iterable($body)) && \array_key_exists($info->fieldName, $source ?? [])) {
```
jotwea, line 48: Meanwhile I found a working solution: this tiny change fixes the issue with `null` values.

soyuka:
you can target the lowest branch, 3.x are merged back

src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -32,17 +32,19 @@ public function __construct(
     public function __invoke(string $resourceClass = null, string $rootClass = null, Operation $operation = null): callable
     {
         return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operation) {
+            // special treatment for nested resources without a resolver/provider
```
soyuka, line 35: mhh this line is from the other PR right?

src/GraphQl/Resolver/Factory/ResolverFactory.php
soyuka: shouldn't we check the `PropertyMetadata` type for this?

src/GraphQl/Resolver/Factory/ResolverFactory.php
jotwea: Yes. It was introduced by other PR. I moved this block one level up, though it also gets called when nested collection is `null`

src/GraphQl/Resolver/Factory/ResolverFactory.php
jotwea: refactored

src/GraphQl/Resolver/Factory/ResolverFactory.php
jotwea: I created another PR based on 3.2 that only contains the bugfix relevant for 3.2. This one here also contains the `null`-fix for the revently introduced "nested collection" feature

soyuka:
thanks cherry-picked

PR #6092, opened 2024-01-09
GraphQL: Errors with nullish ManyToOne-Relation when using new ResolverFactory (for version 3.2)

### LINKED ISSUE #6083: GraphQl: Errors with nullish ManyToOne-Relation when using new ResolverFactory

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #6083 
| License       | MIT
| Doc PR        | -

This PR introduces a new the test case especially for properties and Many-To-One-Relations being `null`. It fixes the issues relevant for 3.2. There is another PR https://github.com/api-platform/core/pull/6084 that additionally provides a fix for the nested collection feature which was recently introduced on the main branch.

### DIFF
--- a/src/GraphQl/Resolver/Factory/ResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ResolverFactory.php
@@ -36,9 +36,9 @@ class ResolverFactory implements ResolverFactoryInterface
                 $body = $source[$info->fieldName];
 
                 $propertyMetadata = $rootClass ? $propertyMetadataFactory?->create($rootClass, $info->fieldName) : null;
-                $propertySchemaType = $propertyMetadata?->getSchema()['type'] ?? null;
+                $type = $propertyMetadata?->getBuiltinTypes()[0] ?? null;
                 // Data already fetched and normalized (field or nested resource)
-                if ($body || null === $resourceClass || $propertyMetadata && 'array' !== $propertySchemaType) {
+                if ($body || null === $resourceClass || ($type && !$type->isCollection())) {
                     return $body;
                 }
             }
--- a/src/GraphQl/Resolver/Factory/ResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ResolverFactory.php
@@ -16,6 +16,7 @@ namespace ApiPlatform\GraphQl\Resolver\Factory;
 use ApiPlatform\Metadata\GraphQl\Mutation;
 use ApiPlatform\Metadata\GraphQl\Operation;
 use ApiPlatform\Metadata\GraphQl\Query;
+use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\State\ProcessorInterface;
 use ApiPlatform\State\ProviderInterface;
 use GraphQL\Type\Definition\ResolveInfo;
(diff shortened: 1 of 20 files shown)


src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -24,20 +25,23 @@ class ResolverFactory implements ResolverFactoryInterface
 {
     public function __construct(
         private readonly ProviderInterface $provider,
-        private readonly ProcessorInterface $processor
+        private readonly ProcessorInterface $processor,
+        private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory
```
soyuka, line 29: this needs to be nullable for backward compatibility

jotwea: Done

src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -24,20 +25,23 @@ class ResolverFactory implements ResolverFactoryInterface
...
-            // Data already fetched and normalized (field or nested resource)
-            if ($body = $source[$info->fieldName] ?? null) {
-                return $body;
-            }
+            if (\array_key_exists($info->fieldName, $source ?? [])) {
+                $body = $source[$info->fieldName];
 
-            if (null === $resourceClass && \array_key_exists($info->fieldName, $source ?? [])) {
-                return $body;
+                $propertyMetadata = $rootClass ? $this->propertyMetadataFactory?->create($rootClass, $info->fieldName) : null;
```
soyuka, line 39: Do you think that we can use the property metadata from https://github.com/api-platform/core/blob/62263d08c9bfbc50484bd444b33e4c3a8a6bf836/src/GraphQl/Type/FieldsBuilder.php#L356 ? Adding it to the above closure this would prevent from calling the property metadata factory here as this is quite a hot spot.

jotwea: yes sure. added it to the closure and dropped it at other place.

src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -28,16 +29,18 @@ public function __construct(
...
-                return $body;
-            }
+        return function (?array $source, array $args, $context, ResolveInfo $info) use ($resourceClass, $rootClass, $operation, $propertyMetadataFactory) {
+            if (\array_key_exists($info->fieldName, $source ?? [])) {
+                $body = $source[$info->fieldName];
 
-            if (null === $resourceClass && \array_key_exists($info->fieldName, $source ?? [])) {
-                return $body;
+                $propertyMetadata = $rootClass ? $propertyMetadataFactory?->create($rootClass, $info->fieldName) : null;
+                $propertySchemaType = $propertyMetadata?->getSchema()['type'] ?? null;
```
soyuka, line 39: shouldn't we use `->getBuiltinTypes`?

src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -28,16 +29,18 @@ public function __construct(
...
+                // Data already fetched and normalized (field or nested resource)
+                if ($body || null === $resourceClass || $propertyMetadata && 'array' !== $propertySchemaType) {
```
soyuka, line 41:
```suggestion
                if ($body || null === $resourceClass || 'array' !== $propertySchemaType) {
```

src/GraphQl/Resolver/Factory/ResolverFactory.php
jotwea: changed it. thx

src/GraphQl/Resolver/Factory/ResolverFactory.php
jotwea: obsolete because `isCollection` of builtInType could have been used.

jotwea:
@soyuka Thx for your latest comments. I updated the PR with the requested changes. The CI still failes for two tests that are not related to my changes. They are already broken on HEAD of 3.2 (https://github.com/api-platform/core/actions/runs/8482337140/job/23241393569). On my local environment they are all passed.

soyuka:
nice thanks! I'll let you now when I merge 3.2 onto main

PR #6095, opened 2024-01-09
feat(metadata): allow \Stringable for security parameters

### LINKED ISSUE #6087: Expressions on security parameter of operations
**Description**  

`security` property of operations attributes only allows `string`. In our project, we want to make this part more statically typed but it's not possible atm. In a lot of parts of Symfony, you can provide either string or [Expression](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/ExpressionLanguage/Expression.php) object for the expressions. Some examples:
- https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Security/Http/Attribute/IsGranted.php#L34
- https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Validator/Constraints/Expression.php#L38
- https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Validator/Constraints/When.php#L26

I think it would make sense to allow it in api-platform attributes as well.
To not couple it with Symfony, [Stringable](https://www.php.net/manual/en/class.stringable.php) could be used since Expression implements Stringable.

No BC break since it would be widening the type of constructor parameter. 

If you approve this change, I could provide the implementation.

**Example**  

Implementation:
```php
final class Get extends HttpOperation
{
    public function __construct(
        //...
        string|\Stringable|null $security = null,
        string $securityMessage = null,
        string|\Stringable|null $securityPostDenormalize = null,
        string $securityPostDenormalizeMessage = null,
        string|\Stringable|null $securityPostValidation = null,
        string $securityPostValidationMessage = null,
        //...
    ) {
        //...
    }
}
```
```php
abstract class Operation
{
    public function getSecurity(): ?string
    {
        return $this->security !== null ? (string)$this->security : null;
    }
}
```

Usage example:
```php
#[\Attribute]
class IsGranted extends Expression
{
    public function __construct(RoleEnum $role)
    {
        parent::__construct(\sprintf("is_granted('%s', object)", $role->value));
    }
}
```
```php
#[ApiResource(
    operations: [
        new Get(
            security: new IsGranted(RoleEnum::ROLE_ADMIN),
        ),
    ],
)]
class SomeResource
{}
```

aurimasrim:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #6087 
| License       | MIT
| Doc PR        | N/A

Allows passing \Stringable objects to security parameters of operations and resources.
This enables users to use expressions in a more statical way.
See https://github.com/api-platform/core/issues/6087

Failing tests seem to be unrelated with changes.


soyuka:
can you target the main branch please?

soyuka:
thanks @aurimasrim

PR #6098, opened 2024-01-10
fix(jsonschema): keep integer and number properties draft 4 compliant

### LINKED ISSUE #6041: json-schema: Generated schema for Range numeric type breaks Draft 4
**API Platform version(s) affected**: 3.2.7

**Description**  
Generated JSON Schema for Range numbers changed from 3.2.6 to 3.2.7. The array copy previously was: 

```php
"position" => ArrayObject {#24994
    storage: array:4 [
    "minimum" => 0
    "exclusiveMinimum" => true
    "externalDocs" => array:1 [
        "url" => "https://schema.org/position"
    ]
    "type" => "integer"
    ]
    flag::STD_PROP_LIST: false
    flag::ARRAY_AS_PROPS: false
    iteratorClass: "ArrayIterator"
}
``` 

In 3.2.7 it is now: 

```php
"position" => ArrayObject {#5413
    storage: array:3 [
        "exclusiveMinimum" => 0
        "externalDocs" => array:1 [
            "url" => "https://schema.org/position"
        ]
        "type" => "integer"
    ]
    flag::STD_PROP_LIST: false
    flag::ARRAY_AS_PROPS: false
    iteratorClass: "ArrayIterator"
}
```

They are both right in their own way. Because in [JSON-Schema docs](https://json-schema.org/understanding-json-schema/reference/numeric#range) under range you can read: 

```
While you can specify both of minimum and exclusiveMinimum or both of maximum and exclusiveMaximum, it doesn't really make sense to do so.
```
There is also a Draft 4 specific interpretation: 

```
In JSON Schema Draft 4, exclusiveMinimum and exclusiveMaximum work differently. There they are boolean values, that indicate whether minimum and maximum are exclusive of the value. For example:

if exclusiveMinimum is false, x ≥ minimum
if exclusiveMinimum is true, x > minimum.

This was changed to have better keyword independence.
```

The [json-schema validator library](https://github.com/justinrainbow/json-schema) used in API Platform only supports Draft 4, so any Range property now breaks schema assertions.

I am not sure if the team considers this a regression or not because I cannot find the reasoning for dropping Draft 4. Depending on this the fix can be different.

**How to reproduce**  

In my case I have a `$position` property in my API Resource that has a `Symfony\Component\Validator\Constraints\Positive` attribute. If you have an API test case that validates response json schema, you can see how it fails between 3.2.6 and 3.2.7.

**Possible Solution**  
- revert the change to drop Draft 4
- keep the change and open an issue over validator library

llupa:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #6041
| License       | MIT

I am opening this PR as a starting point to close the issue linked above. From a quick conversation with @soyuka in Symfony Slack workspace I am introducing a top-level decorator that targets `integer` and `number` properties that have range definition in them. The definition is set to Draft 4 of jsonschema since the assertion library the API Platform comes with does not recognize current Draft 6 defintion. 

I want to add tests but I am not sure what are the preferred tests by the team. There are no behat tests for jsonschema checks directly (they are indirectly assessed in openapi feature files). I thought to add a `KernelTestCase` since the schema will append the extra definitions based on Symfony assertions found in the property. Let me know, happy to add/squash more in this PR.

### DIFF
--- a/src/JsonSchema/BackwardCompatibleSchemaFactory.php
+++ b/src/JsonSchema/BackwardCompatibleSchemaFactory.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\JsonSchema;
+
+use ApiPlatform\Metadata\Operation;
+
+/**
+ * This factory decorates range integer and number properties to keep Draft 4 backward compatibility.
+ *
+ * @see https://github.com/api-platform/core/issues/6041
+ *
+ * @internal
+ */
+final class BackwardCompatibleSchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
+{
+    public const SCHEMA_DRAFT4_VERSION = 'draft_4';
+
+    public function __construct(private readonly SchemaFactoryInterface $decorated)
+    {
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, Operation $operation = null, Schema $schema = null, array $serializerContext = null, bool $forceCollection = false): Schema
+    {
+        $schema = $this->decorated->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
+
+        if (!($serializerContext[self::SCHEMA_DRAFT4_VERSION] ?? false)) {
+            return $schema;
+        }
+
+        foreach ($schema->getDefinitions() as $definition) {
+            foreach ($definition['properties'] ?? [] as $property) {
+                if (isset($property['type']) && \in_array($property['type'], ['integer', 'number'], true)) {
+                    if (isset($property['exclusiveMinimum'])) {
+                        $property['minimum'] = $property['exclusiveMinimum'];
+                        $property['exclusiveMinimum'] = true;
+                    }
+                    if (isset($property['exclusiveMaximum'])) {
+                        $property['maximum'] = $property['exclusiveMaximum'];
+                        $property['exclusiveMaximum'] = true;
+                    }
+                }
+            }
+        }
+
+        return $schema;
+    }
+
+    public function setSchemaFactory(SchemaFactoryInterface $schemaFactory): void
+    {
+        if ($this->decorated instanceof SchemaFactoryAwareInterface) {
+            $this->decorated->setSchemaFactory($schemaFactory);
+        }
+    }
+}
--- a/src/Symfony/Bundle/Resources/config/json_schema.xml
+++ b/src/Symfony/Bundle/Resources/config/json_schema.xml
@@ -36,6 +36,10 @@
             <argument type="service" id="api_platform.json_schema.metadata.property.metadata_factory.schema.inner" />
         </service>
 
+        <service id="api_platform.json_schema.backward_compatible_schema_factory" decorates="api_platform.json_schema.schema_factory" decoration-priority="-2" class="ApiPlatform\JsonSchema\BackwardCompatibleSchemaFactory">
+            <argument type="service" id="api_platform.json_schema.backward_compatible_schema_factory.inner" />
+        </service>
+
     </services>
 
 </container>
--- a/src/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
+++ b/src/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Symfony\Bundle\Test;
 
+use ApiPlatform\JsonSchema\BackwardCompatibleSchemaFactory;
 use ApiPlatform\JsonSchema\Schema;
 use ApiPlatform\JsonSchema\SchemaFactoryInterface;
 use ApiPlatform\Metadata\Get;
@@ -118,7 +119,7 @@ trait ApiTestAssertionsTrait
             $operation = $operationName ? (new GetCollection())->withName($operationName) : new GetCollection();
         }
 
-        $schema = $schemaFactory->buildSchema($resourceClass, $format, Schema::TYPE_OUTPUT, $operation, null, $serializationContext);
+        $schema = $schemaFactory->buildSchema($resourceClass, $format, Schema::TYPE_OUTPUT, $operation, null, ($serializationContext ?? []) + [BackwardCompatibleSchemaFactory::SCHEMA_DRAFT4_VERSION => true]);
 
         static::assertMatchesJsonSchema($schema->getArrayCopy());
     }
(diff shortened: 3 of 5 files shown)


src/JsonSchema/BackwardCompatibleSchemaFactory.php
```diff
@@ -0,0 +1,59 @@
...
+use ApiPlatform\Metadata\Operation;
+
+/**
+ * This factory decorates range integer and number properties to keep Draft 4 backward compatibility.
+ *
+ * @see https://github.com/api-platform/core/issues/6041
+ *
+ * @internal
+ */
+class BackwardCompatibleSchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
```
soyuka, line 25:
```suggestion
final class BackwardCompatibleSchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
```

src/JsonSchema/BackwardCompatibleSchemaFactory.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, Operation $operation = null, Schema $schema = null, array $serializerContext = null, bool $forceCollection = false): Schema
+    {
+        $schema = $this->decorated->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
+
+        foreach ($schema->getDefinitions() as $definition) {
+            foreach ($definition['properties'] as $property) {
+                /** @var \ArrayObject $property */
+                if (\in_array($property['type'], ['integer', 'number'], true)) {
+                    if ($property->offsetExists('exclusiveMinimum')) {
```
soyuka, line 42: why not `isset($property['exclusiveMinimum')`?

src/JsonSchema/BackwardCompatibleSchemaFactory.php
```diff
@@ -0,0 +1,59 @@
...
+                        $property->offsetSet('minimum', $property->offsetGet('exclusiveMinimum'));
```
soyuka, line 43:
```suggestion
                        $property['minimum'] = $property['exclusiveMinimum'];
```

src/JsonSchema/BackwardCompatibleSchemaFactory.php
```diff
@@ -0,0 +1,59 @@
...
+                        $property->offsetSet('exclusiveMinimum', true);
```
soyuka, line 44: same as above just use an array accessor

src/Symfony/Bundle/Resources/config/json_schema.xml
```diff
@@ -36,6 +36,10 @@
             <argument type="service" id="api_platform.json_schema.metadata.property.metadata_factory.schema.inner" />
         </service>
 
+        <service id="api_platform.json_schema.backward_compatible_schema_factory" decorates="api_platform.json_schema.schema_factory" decoration-priority="-2" class="ApiPlatform\JsonSchema\BackwardCompatibleSchemaFactory">
+            <argument type="service" id=".inner" />
```
soyuka, line 40:
```suggestion
            <argument type="service" id="api_platform.json_schema.backward_compatible_schema_factory.inner" />
```

src/JsonSchema/BackwardCompatibleSchemaFactory.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, Operation $operation = null, Schema $schema = null, array $serializerContext = null, bool $forceCollection = false): Schema
+    {
+        $schema = $this->decorated->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
+
+        foreach ($schema->getDefinitions() as $definition) {
```
soyuka, line 38: we need to apply this only when needed, I suggested to add a `DRAFT_VERSION` to the context, not sure if we should rather use `$type` or the `$serializerContext`.

soyuka (commented):
Nice start, we also need to add that draft4 information to the https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/Test/ApiTestAssertionsTrait.php#L121 

indeed, that way it'll continue to work with the current version of the json-schema validator.

src/JsonSchema/BackwardCompatibleSchemaFactory.php
llupa: Honestly, no reason. I will change it to `isset` 👍

src/JsonSchema/BackwardCompatibleSchemaFactory.php
llupa: You mean a flag that overrides this part in `Schema` class? 

```php
public function __construct(private readonly string $version = self::VERSION_JSON_SCHEMA)
{
    parent::__construct(self::VERSION_JSON_SCHEMA === $this->version ? ['$schema' => 'http://json-schema.org/draft-07/schema#'] : []);
}
```

llupa: @soyuka  update:

I chose `$serializerContext` because it is least invasive, but neither really fit. I added some tests too and applied your recommendation from the [comment above](https://github.com/api-platform/core/pull/6098#pullrequestreview-1812879052). 

Let me know if there is anything more to do / change.

src/Symfony/Bundle/Test/ApiTestAssertionsTrait.php
```diff
@@ -111,6 +111,7 @@ public static function assertMatchesJsonSchema(object|array|string $jsonSchema,
     public static function assertMatchesResourceCollectionJsonSchema(string $resourceClass, string $operationName = null, string $format = 'jsonld', array $serializationContext = null): void
     {
         $schemaFactory = self::getSchemaFactory();
+        $serializationContext['backward_compatible_draft_4'] = true;
```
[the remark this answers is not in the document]
llupa: I forgot that `$serializationContext` can be null at times, do you prefer casting it to array before this line, or inline it here?

src/JsonSchema/BackwardCompatibleSchemaFactory.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function buildSchema(string $className, string $format = 'json', string $type = Schema::TYPE_OUTPUT, Operation $operation = null, Schema $schema = null, array $serializerContext = null, bool $forceCollection = false): Schema
+    {
+        $schema = $this->decorated->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
+
+        if (!isset($serializerContext[self::SCHEMA_DRAFT4_VERSION])) {
```
soyuka, line 40:
```suggestion
        if ($serializerContext[self::SCHEMA_DRAFT4_VERSION] ?? false) {
```

soyuka: will fix your issue

llupa: I will add this, but the error is because I am doing `null + []` [example](https://3v4l.org/JjNJ0) in the assertion function

soyuka:
amazing thanks!

PR #6100, opened 2024-01-10
fix(graphql): Embedded nullable relations graphql

Koenstell:
| Q             | A
| ------------- | ---
| Branch?       | `main`
| Tickets       | https://github.com/api-platform/api-platform/issues/2562
| License       | MIT

We have a number of cases where we support creating new entities through a relationship with a 'parent' entity. This works fine, but the prop is always marked as 'required' in the GraphQL types that are being generated. This PR addresses this issue and will look for `#[ApiProperty(required: false)]` or `#[ORM\JoinColumn(nullable: true)]` to mark the input as non-required in the GraphQL type.

### DIFF
--- a/src/GraphQl/Type/FieldsBuilder.php
+++ b/src/GraphQl/Type/FieldsBuilder.php
@@ -47,12 +47,15 @@ use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
  */
 final class FieldsBuilder implements FieldsBuilderInterface, FieldsBuilderEnumInterface
 {
-    private readonly TypeBuilderEnumInterface|TypeBuilderInterface $typeBuilder;
+    private readonly ContextAwareTypeBuilderInterface|TypeBuilderEnumInterface|TypeBuilderInterface $typeBuilder;
 
-    public function __construct(private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory, private readonly ResourceClassResolverInterface $resourceClassResolver, private readonly TypesContainerInterface $typesContainer, TypeBuilderEnumInterface|TypeBuilderInterface $typeBuilder, private readonly TypeConverterInterface $typeConverter, private readonly ResolverFactoryInterface $itemResolverFactory, private readonly ?ResolverFactoryInterface $collectionResolverFactory, private readonly ?ResolverFactoryInterface $itemMutationResolverFactory, private readonly ?ResolverFactoryInterface $itemSubscriptionResolverFactory, private readonly ContainerInterface $filterLocator, private readonly Pagination $pagination, private readonly ?NameConverterInterface $nameConverter, private readonly string $nestingSeparator)
+    public function __construct(private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory, private readonly ResourceClassResolverInterface $resourceClassResolver, private readonly TypesContainerInterface $typesContainer, ContextAwareTypeBuilderInterface|TypeBuilderEnumInterface|TypeBuilderInterface $typeBuilder, private readonly TypeConverterInterface $typeConverter, private readonly ResolverFactoryInterface $itemResolverFactory, private readonly ?ResolverFactoryInterface $collectionResolverFactory, private readonly ?ResolverFactoryInterface $itemMutationResolverFactory, private readonly ?ResolverFactoryInterface $itemSubscriptionResolverFactory, private readonly ContainerInterface $filterLocator, private readonly Pagination $pagination, private readonly ?NameConverterInterface $nameConverter, private readonly string $nestingSeparator)
     {
         if ($typeBuilder instanceof TypeBuilderInterface) {
             @trigger_error(sprintf('$typeBuilder argument of FieldsBuilder implementing "%s" is deprecated since API Platform 3.1. It has to implement "%s" instead.', TypeBuilderInterface::class, TypeBuilderEnumInterface::class), \E_USER_DEPRECATED);
+        }
+        if ($typeBuilder instanceof TypeBuilderEnumInterface) {
+            @trigger_error(sprintf('$typeBuilder argument of TypeConverter implementing "%s" is deprecated since API Platform 3.3. It has to implement "%s" instead.', TypeBuilderEnumInterface::class, ContextAwareTypeBuilderInterface::class), \E_USER_DEPRECATED);
         }
         $this->typeBuilder = $typeBuilder;
     }
--- a/src/GraphQl/Type/TypeBuilder.php
+++ b/src/GraphQl/Type/TypeBuilder.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\GraphQl\Type;
 
 use ApiPlatform\GraphQl\Serializer\ItemNormalizer;
+use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\CollectionOperationInterface;
 use ApiPlatform\Metadata\Exception\OperationNotFoundException;
 use ApiPlatform\Metadata\GraphQl\Mutation;
@@ -36,7 +37,7 @@ use Symfony\Component\PropertyInfo\Type;
  *
  * @author Alan Poulain <contact@alanpoulain.eu>
  */
-final class TypeBuilder implements TypeBuilderInterface, TypeBuilderEnumInterface
+final class TypeBuilder implements ContextAwareTypeBuilderInterface
 {
     private $defaultFieldResolver;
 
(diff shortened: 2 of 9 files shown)


src/GraphQl/Type/TypeBuilder.php
```diff
@@ -48,7 +48,7 @@ public function __construct(private readonly TypesContainerInterface $typesConta
     /**
      * {@inheritdoc}
      */
-    public function getResourceObjectType(?string $resourceClass, ResourceMetadataCollection $resourceMetadataCollection, Operation $operation, bool $input, bool $wrapped = false, int $depth = 0): GraphQLType
+    public function getResourceObjectType(?string $resourceClass, ResourceMetadataCollection $resourceMetadataCollection, Operation $operation, bool $input, bool $wrapped = false, int $depth = 0, bool $required = true): GraphQLType
```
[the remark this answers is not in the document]
Koenstell: Not sure which other things from the `ApiProperty` we would need here?

src/GraphQl/Type/TypeBuilder.php
```diff
@@ -86,8 +86,8 @@ public function getResourceObjectType(?string $resourceClass, ResourceMetadataCo
 
         if ($this->typesContainer->has($shortName)) {
             $resourceObjectType = $this->typesContainer->get($shortName);
-            if (!($resourceObjectType instanceof ObjectType || $resourceObjectType instanceof NonNull)) {
-                throw new \LogicException(sprintf('Expected GraphQL type "%s" to be %s.', $shortName, implode('|', [ObjectType::class, NonNull::class])));
+            if (!($resourceObjectType instanceof ObjectType || $resourceObjectType instanceof NonNull || $resourceObjectType instanceof InputObjectType)) {
```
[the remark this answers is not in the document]
Koenstell: No, they both extend `Type` afaik.

src/GraphQl/Type/TypeBuilder.php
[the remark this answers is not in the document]
soyuka: I'm trying to prevent having 10 arguments on one of our public interfaces, this method has already too many arguments the metadata classes are really helpful to condense information in one argument.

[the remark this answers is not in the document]
Koenstell: That makes sense, let me update that

Koenstell:
I've added a behat test for this, and while doing this I found an edge-case that wasn't taken care of. Namely, if you have 2 references to the same type, it will 'cache' the type definition (including the required/not required attribute). I had to throw a few things around in `TypeBuilder::getResourceObjectType` to accommodate for that. 

Next up I will refactor the required argument and double-check cs-fixer/phpstan.

Koenstell:
@soyuka I've addressed everything I believe. Not sure why the `Behat (PHP8.3) (Symfony lowest)` test is failing, other Behat tests are working and the error doesnt make sense to me?

Please review the changes whenever you can.

src/GraphQl/Type/TypeBuilder.php
```diff
@@ -84,80 +85,84 @@ public function getResourceObjectType(?string $resourceClass, ResourceMetadataCo
             $shortName .= 'Data';
         }
 
-        if ($this->typesContainer->has($shortName)) {
```
soyuka, line 87: quite nice, not sure if reversing this condition was useful (it makes my life harder as the diff is complicated :p) but I think I got it in the end we want to apply the `required` to every resourceObjectType. Maybe we can make this more readable (see next comment)

src/GraphQl/Type/TypeBuilder.php
```diff
@@ -84,80 +85,84 @@ public function getResourceObjectType(?string $resourceClass, ResourceMetadataCo
...
-            $resourceObjectType = $this->typesContainer->get($shortName);
-            if (!($resourceObjectType instanceof ObjectType || $resourceObjectType instanceof NonNull)) {
-                throw new \LogicException(sprintf('Expected GraphQL type "%s" to be %s.', $shortName, implode('|', [ObjectType::class, NonNull::class])));
+        $resourceObjectType = null;
+        if (!$this->typesContainer->has($shortName)) {
+            $ioMetadata = $input ? $operation->getInput() : $operation->getOutput();
```
soyuka, line 90: the content of this `if` is quite long, what about moving it to a private function?

src/GraphQl/Type/TypeBuilderEnumInterface.php
```diff
@@ -31,9 +30,9 @@ interface TypeBuilderEnumInterface
     /**
      * Gets the object type of the given resource.
      *
-     * @return ObjectType|NonNull the object type, possibly wrapped by NonNull
+     * @return GraphQLType the object type, possibly wrapped by NonNull
      */
-    public function getResourceObjectType(?string $resourceClass, ResourceMetadataCollection $resourceMetadataCollection, Operation $operation, bool $input, bool $wrapped = false, int $depth = 0): GraphQLType;
+    public function getResourceObjectType(?string $resourceClass, ResourceMetadataCollection $resourceMetadataCollection, Operation $operation, bool $input, bool $wrapped = false, int $depth = 0, ApiProperty $propertyMetadata = null): GraphQLType;
```
soyuka, line 35: Because of the BC promise (https://symfony.com/doc/current/contributing/code/bc.html) we can not add an argument here. I think that the best here would be to create a new interface. 

Example: 

```php
interface ContextAwareTypeBuilderInterface
{
    /**
     * Gets the object type of the given resource.
     *
     * @param array<string, mixed>&array{input?: bool, wraped?: bool, depth?: int} $context
     *
     * @return GraphQLType the object type, possibly wrapped by NonNull
     */
    public function getResourceObjectType(ResourceMetadataCollection $resourceMetadataCollection, Operation $operation, ?ApiProperty $apiProperty = null, array $context = []): GraphQLType;

    /**
     * Get the interface type of a node.
     */
    public function getNodeInterface(): InterfaceType;

    /**
     * Gets the type of a paginated collection of the given resource type.
     */
    public function getPaginatedCollectionType(GraphQLType $resourceType, Operation $operation): GraphQLType;

    /**
     * Gets the type corresponding to an enum.
     */
    public function getEnumType(Operation $operation): GraphQLType;

    /**
     * Returns true if a type is a collection.
     */
    public function isCollection(Type $type): bool;
}

```

I changed the method to be more future proof, we can always use the `context` to avoid changing (again) this interface... `$resourceClass` is always available through ($operation->getClass()) and `input`, `wrapped` and `depth` are inside the context. 

Then you need to deprecate the `TypeBuilderEnumInterface` and to check inside the `TypeConverter` if the instance is a `ContextAwareTypeBuilderInterface` so we can call the new signature. Sorry I didn't catched that earlier on

soyuka (approved):

src/GraphQl/Type/TypeBuilder.php
Koenstell: Yes, i did this because otherwise the type would set `set` in the typesContainer as `nonNull` in case of a required type. If later in the same entity there would be a different field that refers to the same entity, it would always get this `nonNull` type even if it's NOT required. Bit of an edge-case, but this way we only 'store' the type, and optionally wrap it into a `nonNull` type based on the ApiProperty settings.

src/GraphQl/Type/TypeBuilder.php
Koenstell: Not sure which `if`-statement you're exactly referring to here?

Koenstell:
@soyuka Thanks for the review, i've addressed the deprecating of the `TypeBuilderEnumInterface` please check if it looks good. Some of the tests seem to fail but it feels like it's because of other changes in the `3.2` branch where i branched of from, if not, please let me know.

src/GraphQl/Type/TypeBuilder.php
soyuka: 94-162 right now?

Koenstell: I see, i've updated the code by extracting that part into a private function.

Koenstell:
@soyuka Any chance you can re-review this?

soyuka:
Lowest is green on 3.2 so I think this is related. You need to reproduce the steps of the CI, basically `--prefer-lowest` it looks like the schema changes. Also a rebase of 3.2 should fix `cs`

src/GraphQl/Type/FieldsBuilder.php
```diff
@@ -47,13 +47,16 @@
...
     {
         if ($typeBuilder instanceof TypeBuilderInterface) {
             @trigger_error(sprintf('$typeBuilder argument of FieldsBuilder implementing "%s" is deprecated since API Platform 3.1. It has to implement "%s" instead.', TypeBuilderInterface::class, TypeBuilderEnumInterface::class), \E_USER_DEPRECATED);
         }
+        if ($typeBuilder instanceof TypeBuilderEnumInterface) {
+            @trigger_error(sprintf('$typeBuilder argument of TypeConverter implementing "%s" is deprecated since API Platform 3.3. It has to implement "%s" instead.', TypeBuilderEnumInterface::class, ContextAwareTypeBuilderInterface::class), \E_USER_DEPRECATED);
```
soyuka, line 58: Perfect, but this means that we need to target `main` we wouldn't want to introduce deprecations in a patch release (3.3 will come out soon).

soyuka (approved):
Quite nice, few leftovers, I know the `lowest` test that breaks isn't fun to fix I'm not sure why the schema changed only on the lowest branch.

PR #6103, opened 2024-01-11
fix: better generics support for State\ProcessorInterface

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

In a next PR, the maker command could also be improved to take the name of the provided entity class as an optional parameter.

### DIFF
--- a/src/State/ProcessorInterface.php
+++ b/src/State/ProcessorInterface.php
@@ -14,11 +14,13 @@ declare(strict_types=1);
 namespace ApiPlatform\State;
 
 use ApiPlatform\Metadata\Operation;
+use Symfony\Component\HttpFoundation\Request;
 
 /**
- * Process data: send an email, persist to storage, add to queue etc.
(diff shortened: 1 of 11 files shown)


src/State/ProcessorInterface.php
```diff
@@ -25,8 +25,9 @@
 interface ProcessorInterface
 {
     /**
-     * Handle the state.
+     * Handles the state.
      *
+     * @param T                                                                                                                                                      $data
```
soyuka, line 30: `$data` and the return type may be different can't we have another generic then `T`?

bendavies:
FYI, i think bumping the phpstan level to **at least** 6 (8 ideally) would be a good idea. 
From my own experience, anything below level 6 hides a lot of bugs.

soyuka (approved):

PR #6104, opened 2024-01-11
Exclude IRI from serializer's cache key to avoid cache explosion

amne:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Exclude IRI from serializer's cache key. Having this unique identifier in $context results in private array caches explosion when iterating big result sets and normalizing items (batch processing jobs or exports). 

Trying to serialize with provided $context['cache_key'] would work but this line https://github.com/symfony/symfony/blob/34915f6e16f04537eb18d9d2c303ec375e63cc4b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L708 results in normalized child contexts ignoring the provided 'cache_key' and generating unique cache keys because the IRI is part of the hash.

No tests were added because I'm not really sure how to test for this. The affected cache behavior is in Symfony not here.

### DIFF
--- a/src/JsonApi/Serializer/ItemNormalizer.php
+++ b/src/JsonApi/Serializer/ItemNormalizer.php
@@ -94,6 +94,7 @@ final class ItemNormalizer extends AbstractItemNormalizer
         $context = $this->initContext($resourceClass, $context);
         $iri = $this->iriConverter->getIriFromResource($object, UrlGeneratorInterface::ABS_PATH, $context['operation'] ?? null, $context);
         $context['iri'] = $iri;
+        $context = $this->excludeIriFromCacheKey($context);
         $context['api_normalize'] = true;
 
         if (!isset($context['cache_key'])) {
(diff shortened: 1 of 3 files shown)


amne:
@soyuka I'm curious to find out more about the security impact. Or some leads on where to start. thank you

soyuka:
doesn't:

```
#[ApiResource(normalizationContext: [AbstractObjectNormalizer::EXCLUDE_FROM_CACHE_KEY => ['iri']]);
```

works?

Not having the IRI inside a collection makes that you have the same cache for every item of that collection. When using the `security` option on properties you don't want the same cache.

amne:
> doesn't:
> 
> ```
> #[ApiResource(normalizationContext: [AbstractObjectNormalizer::EXCLUDE_FROM_CACHE_KEY => ['iri']]);
> ```
> 
> works?
> 
> Not having the IRI inside a collection makes that you have the same cache for every item of that collection. When using the `security` option on properties you don't want the same cache.

Adding manually the IRI to cache exclude works, of course.

I didn't dig deep enough in my initial investigation. Looking at features/authorization/deny I can now see more what the implications of this are.

Issue #6072, opened 2023-12-29
GraphQL ReadProvider "Item from read provider should be a nullable object."


nuryagdym:
**API Platform version(s) affected**: 3.2.7

**Description**  
I am getting "**Item from read provider should be a nullable object.**" error when I request a **single** item via GraphQL query if item with that id **does not exist**. When item exists then it is working properly. 

I debugged a little bit and found following infos:
- error is thrown here: https://github.com/api-platform/core/blob/a749fe8494a5b3ca41139a418ed1e29c3e7b33a6/src/GraphQl/State/Provider/ReadProvider.php#L82

- on line 66 of **ReadProvider** `ItemNotFoundException` was thrown and caught here: https://github.com/api-platform/core/blob/a749fe8494a5b3ca41139a418ed1e29c3e7b33a6/src/GraphQl/State/Provider/ReadProvider.php#L68

I am new to GraphQL so I don't know if it is the correct behaviour.
**How to reproduce**  
Try to fetch an item that does not exist via GraphQL API

**Possible Solution**  
Should not it just return **null** if item is not found instead of throwing an exception?

**Additional Context**

Zippovich2:
Same here

soyuka:
please update to latest 3.2. 

we should probably just use the same logic that I see at https://github.com/api-platform/core/blob/a749fe8494a5b3ca41139a418ed1e29c3e7b33a6/src/GraphQl/State/Provider/ReadProvider.php#L71 

I would be curious to see what test fail on this use case. Feel free to patch it.

PR #6118, opened 2024-01-20
fix: return null instead of exception for GraphQL Query operation

Zippovich2:
| Q             | A
| ------------- | ---
| Branch?       | `3.2`
| Tickets       | Closes api-platform/core#6072
| License       | MIT
| Doc PR        | -

After upgrading to 3.2 we noticed a new error `Item from read provider should be a nullable object.` in our projects, when request single resource via GraphQL Query if such resource does not exists with provided id, see api-platform/core#6072.

### DIFF
--- a/src/GraphQl/State/Provider/ReadProvider.php
+++ b/src/GraphQl/State/Provider/ReadProvider.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\GraphQl\State\Provider;
 
-use ApiPlatform\Exception\ItemNotFoundException;
+use ApiPlatform\Metadata\Exception\ItemNotFoundException;
 use ApiPlatform\GraphQl\Resolver\Util\IdentifierTrait;
 use ApiPlatform\GraphQl\Serializer\ItemNormalizer;
 use ApiPlatform\GraphQl\Serializer\SerializerContextBuilderInterface;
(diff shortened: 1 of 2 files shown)


src/GraphQl/State/Provider/ReadProvider.php
```diff
@@ -68,6 +69,10 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
                 $item = null;
             }
 
+            if ($operation instanceof Query && null === $item) {
```
soyuka, line 72: shouldn't it be at line 86 that we check if the `$item` is not null? Also can you add a non regression test?

Zippovich2: I put it above because it looks like condition `!\is_object($item)` on line 86 is related to Subscription and Mutation operations logic and `$item` should be checked on null before it.

Zippovich2: As for test - I did it but unfortunately `ApiPlatform\Exception\ItemNotFoundException` is outside of GraphQL directory and I cannot mock `IriConvertorInterface` to throw this exception. Used `class_alias` to make it work, not sure if it's OK.

rutek:
Hello, do you have any updates on this fix? The issue #6072 contains information that an update to the latest 3.2.x should fix it, but the fix seems to not be merged. :D

PR #6127, opened 2024-01-25
fix(elasticsearch): return type for ItemNormalizer

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Branch?       | 3.2

Decorated normalizer has return type `array|ArrayObject|bool|float|int|null|string \ArrayObject`, and I had issues in test where the returned value was string (item uri)

### DIFF
--- a/src/Elasticsearch/Tests/Serializer/ItemNormalizerTest.php
+++ b/src/Elasticsearch/Tests/Serializer/ItemNormalizerTest.php
@@ -91,6 +91,15 @@ final class ItemNormalizerTest extends TestCase
         self::assertEquals(['foo'], $this->itemNormalizer->normalize($object, 'json', ['groups' => 'foo']));
     }
 
+    public function testNormalizeIntoIri(): void
+    {
+        $iri = '/foobar/1234';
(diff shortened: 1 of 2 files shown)


src/Elasticsearch/Tests/Serializer/ItemNormalizerTest.php
```diff
@@ -91,6 +91,15 @@ public function testNormalize(): void
         self::assertEquals(['foo'], $this->itemNormalizer->normalize($object, 'json', ['groups' => 'foo']));
     }
 
+    public function testNormalizeIntoIri(): void
+    {
+        $iri = '/foobar/1234';
+        $item = new \stdClass();
+        $this->normalizerProphecy->normalize($item, 'json', ['groups' => 'foo'])->willReturn($iri)->shouldBeCalledOnce();
+
+        self::assertEquals($iri, $this->itemNormalizer->normalize($item, 'json', ['groups' => 'foo']));
```
GaryPEGEOT, line 100: @soyuka this fail on current `3.2` branch

PR #6134, opened 2024-01-30
tests: Change visibility from private to protected

thecaliskan:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | -
| License       | MIT
| Doc PR        | -

When an abstract class is used for tests or a custom assertion method is written (like assertJsonContains, assertJsonEquals, assertMatchesJsonSchema etc.), I updated the visibility from private to protected to able access the response.


thecaliskan:
Did you have the opportunity to review @soyuka ?

thecaliskan:
Hi @soyuka 

What is your opinion about MR?

dunglas:
What's the benefit of doing this? As these methods are defined in a trait, you can already change their visibility when importing them if necessary.

thecaliskan:
For example, I will add a new test method like 'assertJsonContains'. When I want to use it this way, I cannot use it in the class I inherited. Because the visibility of the 'self::getHttpResponse()' method is private.

thecaliskan:
I enjoy using API Platform testing methods in more than one project, but I don't like doing a hack like this for all of them.

soyuka:
Please use:

```php
class MyApiTestCase extends ApiTestCase {
    use ApiTestAssertionsTrait { getHttpResponse as protected; }
}
```

PR #6139, opened 2024-02-01
fix(doctrine): fix error when identifier is not 'id'

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Branch?       | 3.2

When your entity identifier is not named id, your `uriVariables` (if not customized) will contains something like `["id" =>"THE_CODE"]`, but Doctine will throw if you send an array as it expect something like  `["nameOfIdentifer" =>"THE_CODE"]`

By using the identifier directly, we let doctrine figure out the rest


soyuka:
There's an `array_key_exists` how do you go through there without any `id`?

GaryPEGEOT:
@soyuka in my case, the class doesn't have any `id` property (identifier is named `sku`). I've added `#[ApiProperty(identifier: true)]` on top of my property, but for some reason identifiers contains `sku` and an imaginary `id` prop. So I did:

```php
 new Get(
            uriVariables: ['id' => new Link('id', fromClass: MyClass::class, identifiers: ['sku'])],
           // ...
        ),
```

PR #6147, opened 2024-02-05
Update AbstractSearchFilter to fix error after upgrading to 3.2

Renrhaf:
After upgrading to API Platform 3.2 I noticed an error that was not there before.
I think it might be related to the wrong exception class used, as you can see in the diff.


soyuka:
should target 3.2 then, thanks.

Renrhaf:
sorry ! thanks for the merge @soyuka

soyuka:
actually this is wrong, the Metadata exception is correct, may you provide me with more informations on this?

Renrhaf:
I have a filter on an ElasticSearch resource, this filter is on the "id" property which is an identifier.

image
image

I thought that the "try catch" in the first screenshot should let go of an "ApiPlatform\Exception\InvalidArgumentException" instead of the metadata one, since one is thrown in the code in the screenshot two.

Let me know if you think it's not right and there is an other way ?

Renrhaf:
@soyuka there seems to be an other PR regarding this topic : https://github.com/api-platform/core/pull/6130

PR #6154, opened 2024-02-11
fix(hydra): remove dependency from ApiPlatform/Api dependency

aegypius:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

This PR purpose is to resolve issue with the splitted version.

`ApiPlatform/Api` is not exported in any components, the `FilterLocatorTrait` cannot be loaded.

This is a quick & dirty workaround that solves the 3.3.0-alpha.1 issue but I think we must find abetter way to solve and provide the filters.

We could describle a FilterRegistryInterface in `ApiPlatform/Metadata` and inject it in the variouscomponents that relies on this feature (Hydra, ParameterValidator) the symfony component could fill an implementation of this registry and inject it.

### DIFF
--- a/src/Hydra/Serializer/CollectionFiltersNormalizer.php
+++ b/src/Hydra/Serializer/CollectionFiltersNormalizer.php
@@ -13,11 +13,10 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Hydra\Serializer;
 
-use ApiPlatform\Api\FilterInterface as LegacyFilterInterface;
-use ApiPlatform\Api\FilterLocatorTrait;
 use ApiPlatform\Api\ResourceClassResolverInterface as LegacyResourceClassResolverInterface;
 use ApiPlatform\Doctrine\Odm\State\Options as ODMOptions;
 use ApiPlatform\Doctrine\Orm\State\Options;
+use ApiPlatform\Metadata\Exception\InvalidArgumentException;
 use ApiPlatform\Metadata\FilterInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use ApiPlatform\Metadata\ResourceClassResolverInterface;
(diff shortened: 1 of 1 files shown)


src/Hydra/Serializer/CollectionFiltersNormalizer.php
```diff
@@ -159,4 +156,30 @@ private function getSearch(string $resourceClass, array $parts, array $filters):
 
         return ['@type' => 'hydra:IriTemplate', 'hydra:template' => sprintf('%s{?%s}', $parts['path'], implode(',', $variables)), 'hydra:variableRepresentation' => 'BasicRepresentation', 'hydra:mapping' => $mapping];
     }
+
+    private ?ContainerInterface $filterLocator = null;
+
+    /**
+     * Sets a filter locator with a backward compatibility.
```
soyuka, line 163: what's the backward compatibility? maybe we can simplify this and actually remove this method?

src/Hydra/Serializer/CollectionFiltersNormalizer.php
```diff
@@ -159,4 +156,30 @@ private function getSearch(string $resourceClass, array $parts, array $filters):
...
+            $this->filterLocator = $filterLocator;
+        } else {
+            throw new InvalidArgumentException(sprintf('The "$filterLocator" argument is expected to be an implementation of the "%s" interface%s.', ContainerInterface::class, $allowNull ? ' or null' : ''));
+        }
+    }
+
+    /**
+     * Gets a filter with a backward compatibility.
+     */
+    private function getFilter(string $filterId): FilterInterface|null
```
soyuka, line 177: it's fine to keep this here imo we don't need the trait

src/Hydra/Serializer/CollectionFiltersNormalizer.php
aegypius: I notice afterwards that I removed the compatibility layer, I will update this to restore the behaviour

aegypius: I might merge both of the trait methods in a single  one

PR #6155, opened 2024-02-12
fix: use short name when generating parameter description

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        | 

OpenApiFactory previously only uses class name when generating parameter description, and did not respect the `shortName` parameter. This PR fixes the problem.

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -265,7 +265,14 @@ final class OpenApiFactory implements OpenApiFactoryInterface
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -265,7 +265,14 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
                     continue;
                 }
 
-                $parameter = new Parameter($parameterName, 'path', (new \ReflectionClass($uriVariable->getFromClass()))->getShortName().' identifier', true, false, false, ['type' => 'string']);
+                $fromClass = $this->resourceMetadataFactory->create($uriVariable->getFromClass());
```
soyuka, line 268: interesting but I think that this is quite a lot of work, probably that this should actually get fixed by #5995

PR #6171, opened 2024-02-20
fix(json-schema): multiple type support (draft4)

### LINKED ISSUE #6041: json-schema: Generated schema for Range numeric type breaks Draft 4

soyuka:
fixes #6041, the `in_array` condition was not working when the `type` is an array (multiple types like `null, number`)

### DIFF
--- a/src/JsonSchema/BackwardCompatibleSchemaFactory.php
+++ b/src/JsonSchema/BackwardCompatibleSchemaFactory.php
@@ -43,7 +43,15 @@ final class BackwardCompatibleSchemaFactory implements SchemaFactoryInterface, S
 
         foreach ($schema->getDefinitions() as $definition) {
             foreach ($definition['properties'] ?? [] as $property) {
-                if (isset($property['type']) && \in_array($property['type'], ['integer', 'number'], true)) {
+                if (!isset($property['type'])) {
+                    continue;
+                }
+
+                foreach ((array) $property['type'] as $type) {
+                    if ('integer' !== $type && 'number' !== $type) {
+                        continue;
+                    }
+
                     if (isset($property['exclusiveMinimum'])) {
                         $property['minimum'] = $property['exclusiveMinimum'];
                         $property['exclusiveMinimum'] = true;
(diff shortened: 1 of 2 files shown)


louismariegaborit:
Tests ok ;) Thanks :)

PR #6174, opened 2024-02-22
fix(graphql): nested collection for mongo

### LINKED ISSUE #6107: GraphQL MongoDB: Nested Collections
https://github.com/api-platform/core/pull/6038 was introduced (@jotwea) but we had mongodb tests disabled. I've skipped the test for now but it'd be nice to check what's going on: 

https://github.com/api-platform/core/blob/main/features/graphql/query.feature#L71 

Note that to run mongodb tests use the docker command + instructions from https://github.com/api-platform/core/blob/main/CONTRIBUTING.md#behat

jotwea:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #6107 
| License       | MIT

https://github.com/api-platform/core/pull/6038 was introduced but we had mongodb tests disabled. Though the new scenario has been disabled temporarily for mongo. I activated it again and updated the mongo test fixtures to represent the same case as the ORM ones do.

### DIFF
--- a/features/graphql/query.feature
+++ b/features/graphql/query.feature
@@ -68,7 +68,6 @@ Feature: GraphQL query support
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[0].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[2].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
 
-  @createSchema @!mongodb
   Scenario: Retrieve embedded collections
     Given there are 2 multiRelationsDummy objects having each a manyToOneRelation, 2 manyToManyRelations, 3 oneToManyRelations and 4 embeddedRelations
     When I send the following GraphQL request:
--- a/features/graphql/query.feature
+++ b/features/graphql/query.feature
@@ -68,6 +68,7 @@ Feature: GraphQL query support
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[0].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[2].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
 
+  @createSchema
   Scenario: Retrieve embedded collections
     Given there are 2 multiRelationsDummy objects having each a manyToOneRelation, 2 manyToManyRelations, 3 oneToManyRelations and 4 embeddedRelations
     When I send the following GraphQL request:
--- a/src/GraphQl/Resolver/Factory/ResolverFactory.php
+++ b/src/GraphQl/Resolver/Factory/ResolverFactory.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\GraphQl\Resolver\Factory;
 
+use ApiPlatform\GraphQl\State\Provider\NoopProvider;
 use ApiPlatform\Metadata\GraphQl\Mutation;
 use ApiPlatform\Metadata\GraphQl\Operation;
 use ApiPlatform\Metadata\GraphQl\Query;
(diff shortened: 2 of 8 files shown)


src/GraphQl/Resolver/Factory/ResolverFactory.php
```diff
@@ -35,7 +35,7 @@ public function __invoke(?string $resourceClass = null, ?string $rootClass = nul
             // Data already fetched and normalized (field or nested resource)
             if ($body = $source[$info->fieldName] ?? null) {
                 // special treatment for nested resources without a resolver/provider
-                if ($operation instanceof Query && $operation->getNested() && !$operation->getResolver() && !$operation->getProvider()) {
+                if ($operation instanceof Query && $operation->getNested() && !$operation->getResolver() && (!$operation->getProvider() || $operation->getProvider() === $resourceClass)) {
```
soyuka, line 38: mhh how is this possible? A provider shouldn't be a resource no?

tests/Fixtures/TestBundle/Document/MultiRelationsNestedPaginated.php
```diff
@@ -17,7 +17,7 @@
 use ApiPlatform\Metadata\GraphQl\QueryCollection;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 
-#[ApiResource(graphQlOperations: [new QueryCollection(nested: true)])]
+#[ApiResource(graphQlOperations: [new QueryCollection(provider: self::class, nested: true)])]
```
soyuka, line 20: you want to force the provider to `null`?

jotwea: Exactly. I found no way how to do that. Setting the provider explicitely to `null` also resulted that the operation had the default provider. So my "trick" was to use the entity itself as a provider. I am open to other ideas or contribution.

jotwea: Like the idea, will change it soon

jotwea: Changed it and work like a charm.Thx

src/GraphQl/Resolver/Factory/ResolverFactory.php
jotwea: Changed it (see comment below)

jotwea:
@soyuka I am done now with the changes you requested. The failing unit tests are not due to my changes (some deprecation notice). From my perspective this PR could be merged.

features/graphql/query.feature
```diff
@@ -68,7 +68,6 @@ Feature: GraphQL query support
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[0].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
     And the JSON node "data.multiRelationsDummy.oneToManyRelations.edges[2].node.name" should match "#RelatedOneToManyDummy(1|3)2#"
 
-  @createSchema @!mongodb
```
soyuka, line 71: why this change? maybe that you want to remove mongodb only?

docs/public/index.php
```diff
@@ -19,7 +19,7 @@
 use ApiPlatform\Playground\Kernel;
 
 if (!($guide = $_SERVER['APP_GUIDE'] ?? $_ENV['APP_GUIDE'] ?? null)) {
-    throw new \RuntimeException('No guide.');
+    throw new RuntimeException('No guide.');
```
soyuka, line 22: can you revert this?

tests/Fixtures/TestBundle/Document/MultiRelationsDummy.php
```diff
@@ -89,7 +89,7 @@ public function addOneToManyRelation(MultiRelationsRelatedDummy $relatedMultiUse
 
     public function getNestedCollection(): Collection
     {
-        return $this->nestedCollection;
+        return $this->nestedCollection->map(fn ($entry) => ['name' => $entry->name]);
```
soyuka, line 92: I'm not fond of this why is this necessary?

jotwea: this mapping function was neccessary because mongo ODM behaves different like ORM does. The ORM Entity provides a Collection with each entry being an associative array. The ODM Document on the other hand provides a Collection with each entry being an instance of `MultiRelationsNested` / `MultiRelationsNestedPaginated`. So my intention here was to align the Testables here to get the same results.

Without this mapper the case for `MultiRelationsNestedPaginated` gets broken in test. If this is not acceptable for you, you may undo this mapping and find another way - I was not able to.

soyuka:
If you can add a documentation for that feature it'd be awesome, thanks!

PR #6178, opened 2024-02-23
fix: description of identifier parameter

Timu57:
I think it is an unexprected behaviour that resource description and summary take the `shortName` parameter into consideration when generating the OpenApi documentation but the identifier parameter description doesn't.

| Q             | A
| ------------- | ---
| Branch?       | current stable version
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -265,7 +265,7 @@ final class OpenApiFactory implements OpenApiFactoryInterface
                     continue;
                 }
 
-                $parameter = new Parameter($parameterName, 'path', (new \ReflectionClass($uriVariable->getFromClass()))->getShortName().' identifier', true, false, false, ['type' => 'string']);
+                $parameter = new Parameter($parameterName, 'path', "$resourceShortName identifier", true, false, false, ['type' => 'string']);
                 if ($this->hasParameter($openapiOperation, $parameter)) {
                     continue;
                 }


priyadi:
This PR has the same goal as mine in https://github.com/api-platform/core/pull/6155 . But I think @soyuka wants to address this in a more fundamental or comprehensive way.

soyuka:
@priyadi as this doesn't execute the `$this->resourceMetadataFactory->create($uriVariable->getFromClass());` I'm more willing into merging this for now then yours I hope that's okay. This is definitely better then the actual code and a good temporary fix.

soyuka:
thanks!

priyadi:
> @priyadi as this doesn't execute the `$this->resourceMetadataFactory->create($uriVariable->getFromClass());` I'm more willing into merging this for now then yours I hope that's okay. This is definitely better then the actual code and a good temporary fix.

This one probably won't fix cases like this:

```php
#[ApiResource(
    shortName: 'User/Review',
    routePrefix: '/user',
    operations: [
        // ...
        new Get(
            uriTemplate: '/books/{bookId}/reviews/{id}',
            uriVariables: [
                'bookId' => new Link(
                    fromClass: BookDto::class,
                )
            ],
        ),
        // ...
    ]
)]
```

But I have no problem as this one should fix the majority of the cases. Should probably test that case first, just to make sure it doesn't generate a wrong documentation.

PR #6179, opened 2024-02-24
fix(serializer): allow POST for `api_allow_update`

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        | 

The documentation says:

> It is also possible to embed a relation in PUT, PATCH and **POST** requests. To enable that feature, set the serialization groups the same way as normalization
>
> ...

But the code currently only allow PUT and PATCH. This PR fixes the problem.

### DIFF
--- a/src/Serializer/SerializerContextBuilder.php
+++ b/src/Serializer/SerializerContextBuilder.php
@@ -95,7 +95,7 @@ final class SerializerContextBuilder implements SerializerContextBuilderInterfac
 
         if (!$normalization) {
             if (!isset($context['api_allow_update'])) {
-                $context['api_allow_update'] = \in_array($method = $request->getMethod(), ['PUT', 'PATCH'], true);
+                $context['api_allow_update'] = \in_array($method = $request->getMethod(), ['PUT', 'PATCH', 'POST'], true);
 
                 if ($context['api_allow_update'] && 'PATCH' === $method) {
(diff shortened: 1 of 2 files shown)


src/Serializer/SerializerContextBuilder.php
```diff
@@ -95,7 +95,7 @@ public function createFromRequest(Request $request, bool $normalization, ?array
 
         if (!$normalization) {
             if (!isset($context['api_allow_update'])) {
-                $context['api_allow_update'] = \in_array($method = $request->getMethod(), ['PUT', 'PATCH'], true);
+                $context['api_allow_update'] = \in_array($method = $request->getMethod(), ['PUT', 'PATCH', 'POST'], true);
```
soyuka, line 98: post should not update I think

priyadi: No, it should not. But it should be able to reference an existing relation, like it is done with PUT and PATCH. Not for updating an existing resource, but for creating one. In which case `api_allow_update` is probably a misnomer.

priyadi: You are right, due to my internal tests, I was under the wrong impression that a plain IRI requires this. But it does not. Closing.

PR #6204, opened 2024-03-05
fix(validation): fix validation exception resource

vincentchalamon:


src/Symfony/Validator/Exception/ValidationException.php
```diff
@@ -27,6 +27,8 @@
  * Thrown when a validation error occurs.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @deprecated since API Platform 3.3, use {@see ValidationException} instead
```
soyuka, line 31:
```suggestion
 * @deprecated since API Platform 3.3, use {@see BaseValidationException} instead
```

PR #5523, opened 2023-04-03
fix: class already declared with preloading

Gwemox:
| Q             | A
| ------------- | ---
| Branch?       | 3.1
| Tickets       | [#... ](https://github.com/api-platform/core/issues/5522)
| License       | MIT
| Doc PR        | 

Error with PHP opacache preload :
`Warning: Cannot declare interface ApiPlatform\Api\FilterInterface, because the name is already in use`

PR #6217, opened 2024-03-13
Fix class alias with preloading

sylfabre:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Tickets       | Closes similar issue to https://github.com/api-platform/core/pull/5523/files
| License       | MIT
| Doc PR        | N/A

https://github.com/api-platform/core/blob/2.7/src/Symfony/Validator/EventListener/ValidationExceptionListener.php suffers from the same issue as https://github.com/api-platform/core/pull/5523/files

=> This PR wraps all calls to `class_alias()` in an if condition to check first if the alias already exists

This is important to us because we want to upgrade first to 2.7 before moving forward to 3.0: we cannot do it right now because preloading breaks with 2.7

### DIFF
--- a/src/Action/EntrypointAction.php
+++ b/src/Action/EntrypointAction.php
@@ -36,4 +36,6 @@ final class EntrypointAction
     }
 }
 
-class_alias(EntrypointAction::class, \ApiPlatform\Core\Action\EntrypointAction::class);
+if (!class_exists(\ApiPlatform\Core\Action\EntrypointAction::class)) {
+    class_alias(EntrypointAction::class, \ApiPlatform\Core\Action\EntrypointAction::class);
+}
--- a/src/Action/ExceptionAction.php
+++ b/src/Action/ExceptionAction.php
@@ -122,4 +122,6 @@ final class ExceptionAction
     }
 }
 
-class_alias(ExceptionAction::class, \ApiPlatform\Core\Action\ExceptionAction::class);
+if (!class_exists(\ApiPlatform\Core\Action\ExceptionAction::class)) {
+    class_alias(ExceptionAction::class, \ApiPlatform\Core\Action\ExceptionAction::class);
+}
--- a/src/Action/NotFoundAction.php
+++ b/src/Action/NotFoundAction.php
@@ -26,4 +26,6 @@ final class NotFoundAction
     }
 }
 
-class_alias(NotFoundAction::class, \ApiPlatform\Core\Action\NotFoundAction::class);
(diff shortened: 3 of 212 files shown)


sylfabre:
Hello @soyuka @dunglas 

My PR caused this critical issue: https://github.com/api-platform/core/issues/6224
Without pointing any finger, I'm wondering why my PR by-passed the CI? 🤔 
Did I miss something somewhere in the process?

dunglas:
Indeed this shouldn’t have been merged.

Should we revert and tag a new release @soyuka?

soyuka:
done

sylfabre:
Thanks @soyuka 

And sorry again for all the trouble. I'll work on another PR with all the previous comments in mind

soyuka:
@sylfabre thing is I don't want to maintain 2.7 and the CI is in a poor state so it's hard to know for sure that we're not breaking anything.  I don't have time to try 2.7 to see if it works... I'm willing to patch this as I think that it helps lots of people but it's also possible that you can't make this patch work (from what I've seen it's harder then it looks).

sylfabre:
@soyuka Yes sure, please focus on the latest version to make it even more exceptional for when we reach it 🚀 

Our initial plan was to work on a fork and send you guys a PR to give back to the community.

This PR was submitted simultaneously with our internal work (including testing) to benefit from the CI.
But, we were not aware you had issues with the CI.
Bottom line: we submitted it too soon and it got merged too soon too!

I suggest the following to move forward:
- We keep our fork for us for now
- Then we make sure internally it works well
- And then reference it in the original issue when we are confident to have a positive impact on the community
=> No extra work for you on 2.7 
=> Anyone can use it as an intermediary step toward a 3.x update if they're willing to use our fork

So far, according to my try this morning, it doesn't look that difficult: I have a version that works with preloading and class autoloader. I'll ask my QA test to validate the change.
We prefer to upgrade little by little rather than making a big bang but I keep your insight about upgrading right to 3.x as a valuable alternative if our way turns out to be too difficult.

soyuka:
I definitely merged too soon but it didn't looked to be as critical as it was. No big deal as you can always use a lower version when it breaks. 

> We keep our fork for us for now 

It works but I think that it'd be best for the community so that everyone can try it, please forward the patches and we should ask everyone that reported the bug in the last 24h to test the patch and I'd be glad to merge and release even though we don't take time to fix our CI. 

tldr: definitely not your fault, please share the work it's awesome!

PR #6223, opened 2024-03-16
fix(jsonschema) don't skip remaining multiple union types

### LINKED ISSUE #6212: Properties with multiple type unions generate incorrect OpenAPI schema

**Description**  
Currently (3.2.16) with the model below the `owner` property will generate the schema:

```
"owner": {
    "anyOf": [
        {
            "$ref": "#/components/schemas/Wren"
        },
        {
            "type": "null"
        }
    ]
}
```

Instead of:
```
"owner": {
    "anyOf": [
        {
            "$ref": "#/components/schemas/Wren"
        },
        {
            "$ref": "#/components/schemas/Robin"
        },
        {
            "type": "null"
        }
    ]
}
```

**How to reproduce**  
```php
#[ApiResource]
#[ORM\Entity]
class Nest
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column]
    private ?int $id = null;

    #[ORM\Column(type: 'bird')]
    private ?Bird $owner;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getOwner(): ?Bird
    {
        return $this->owner;
    }

    public function setOwner(Wren|Robin|null $owner): static
    {
        $this->owner = $owner;

        return $this;
    }
}
```

```php
interface Bird
{
    public function getName(): ?string;

    public function getAge(): ?int;
}
```

```php
final class Robin implements Bird
{
    public ?string $name = null;
    public ?int $age = null;

    public function getName(): ?string
    {
        return $this->name;
    }

    public function getAge(): ?int
    {
        return $this->age;
    }
}
```

```php
final class Wren implements Bird
{
    public ?string $name = null;
    public ?int $age = null;
    public ?int $weight = null;

    public function getName(): ?string
    {
        return $this->name;
    }

    public function getAge(): ?int
    {
        return $this->age;
    }
}
```

**Possible Solution**  
I dug in a bit and it stems from handling in `SchemaFactory::buildPropertySchema`. Simply removing the final `break` [here](https://github.com/api-platform/core/blob/2819d56c80f39a53f2e3d5fd7119566ad96b4075/src/JsonSchema/SchemaFactory.php#L240) make this particular problem disappear but breaks tests. So I hacked up a more robust proof-of-concept fix/patch below that addresses this issue (and passes existing tests). 

```diff
diff --git a/src/JsonSchema/SchemaFactory.php b/src/JsonSchema/SchemaFactory.php
index a128a8968..9a21d47d3 100644
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -196,10 +196,13 @@
         // property schema is created in SchemaPropertyMetadataFactory, but it cannot build resource reference ($ref)
         // complete property schema with resource reference ($ref) only if it's related to an object
         $version = $schema->getVersion();
-        $subSchema = new Schema($version);
-        $subSchema->setDefinitions($schema->getDefinitions()); // Populate definitions of the main schema
+        $refs = [];
+        $isNullable = null;
+
+        foreach ($types as $type) {
+            $subSchema = new Schema($version);
+            $subSchema->setDefinitions($schema->getDefinitions()); // Populate definitions of the main schema
 
-        foreach ($types as $type) {
             // TODO: in 3.3 add trigger_deprecation() as type factories are not used anymore, we moved this logic to SchemaPropertyMetadataFactory so that it gets cached
             if ($typeFromFactory = $this->typeFactory?->getType($type, 'jsonschema', $propertyMetadata->isReadableLink(), $serializerContext)) {
                 $propertySchema = $typeFromFactory;
@@ -230,14 +233,25 @@
                 break;
             }
 
-            if ($type->isNullable()) {
-                $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
-            } else {
-                $propertySchema['$ref'] = $subSchema['$ref'];
+            $isNullable = $isNullable ?? $type->isNullable();
+            $refs[$subSchema['$ref']] = '$ref';
+        }
+
+        if (\count($refs) > 1) {
+            $anyOf = [];
+            foreach (array_keys($refs) as $ref) {
+                $anyOf[] = ['$ref' => $ref];
+            }
+            $propertySchema['anyOf'] = $anyOf;
+
+            if ($isNullable) {
+                $propertySchema['anyOf'][] = ['type' => 'null'];
             }
 
             unset($propertySchema['type']);
-            break;
+        } elseif (\count($refs) === 1) {
+            $propertySchema['$ref'] = array_keys($refs)[0];
+            unset($propertySchema['type']);
         }
 
         $schema->getDefinitions()[$definitionName]['properties'][$normalizedPropertyName] = new \ArrayObject($propertySchema);

```

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #6212 
| License       | MIT

As requested by @soyuka  

For the tests I just imported what I had in a sandbox to create the issue. Happy to modify on request.

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -196,10 +196,13 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
         // property schema is created in SchemaPropertyMetadataFactory, but it cannot build resource reference ($ref)
         // complete property schema with resource reference ($ref) only if it's related to an object
         $version = $schema->getVersion();
-        $subSchema = new Schema($version);
-        $subSchema->setDefinitions($schema->getDefinitions()); // Populate definitions of the main schema
+        $refs = [];
+        $isNullable = null;
 
         foreach ($types as $type) {
+            $subSchema = new Schema($version);
+            $subSchema->setDefinitions($schema->getDefinitions()); // Populate definitions of the main schema
+
             // TODO: in 3.3 add trigger_deprecation() as type factories are not used anymore, we moved this logic to SchemaPropertyMetadataFactory so that it gets cached
             if ($typeFromFactory = $this->typeFactory?->getType($type, 'jsonschema', $propertyMetadata->isReadableLink(), $serializerContext)) {
                 $propertySchema = $typeFromFactory;
@@ -230,14 +233,20 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
                 break;
             }
 
-            if ($type->isNullable()) {
-                $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
-            } else {
-                $propertySchema['$ref'] = $subSchema['$ref'];
-            }
+            $refs[] = ['$ref' => $subSchema['$ref']];
+            $isNullable = $isNullable ?? $type->isNullable();
+        }
 
+        if ($isNullable) {
+            $refs[] = ['type' => 'null'];
+        }
+
+        if (($c = \count($refs)) > 1) {
+            $propertySchema['anyOf'] = $refs;
             unset($propertySchema['type']);
-            break;
+        } elseif (1 === $c) {
+            $propertySchema['$ref'] = $refs[0]['$ref'];
+            unset($propertySchema['type']);
         }
 
         $schema->getDefinitions()[$definitionName]['properties'][$normalizedPropertyName] = new \ArrayObject($propertySchema);
--- a/tests/Fixtures/TestBundle/Entity/Issue6212/Nest.php
+++ b/tests/Fixtures/TestBundle/Entity/Issue6212/Nest.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\Entity\Issue6212;
+
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue6212\Bird;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue6212\Robin;
+use ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue6212\Wren;
+use Doctrine\ORM\Mapping as ORM;
+
+#[ApiResource]
+#[ORM\Entity]
+class Nest
+{
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column]
+    private ?int $id = null;
+
+    #[ORM\Column(type: 'json')]
+    private ?Bird $owner;
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getOwner(): ?Bird
+    {
+        return $this->owner;
+    }
+
+    public function setOwner(Wren|Robin|null $owner): static
+    {
+        $this->owner = $owner;
+
+        return $this;
+    }
+}
--- a/tests/JsonSchema/Command/JsonSchemaGenerateCommandTest.php
+++ b/tests/JsonSchema/Command/JsonSchemaGenerateCommandTest.php
@@ -124,6 +124,22 @@ class JsonSchemaGenerateCommandTest extends KernelTestCase
         ]);
     }
 
+    public function testArraySchemaWithMultipleUnionTypes(): void
+    {
+        $this->tester->run(['command' => 'api:json-schema:generate', 'resource' => 'ApiPlatform\Tests\Fixtures\TestBundle\Entity\Issue6212\Nest', '--type' => 'output']);
+        $result = $this->tester->getDisplay();
+        $json = json_decode($result, associative: true);
+
+        $this->assertEquals($json['definitions']['Nest.jsonld']['properties']['owner']['anyOf'], [
+            ['$ref' => '#/definitions/Wren.jsonld'],
+            ['$ref' => '#/definitions/Robin.jsonld'],
+            ['type' => 'null'],
+        ]);
+
+        $this->assertArrayHasKey('Wren.jsonld', $json['definitions']);
+        $this->assertArrayHasKey('Robin.jsonld', $json['definitions']);
+    }
+
     /**
      * TODO: add deprecation (TypeFactory will be deprecated in api platform 3.3).
      *
--- a/tests/Fixtures/TestBundle/ApiResource/Issue6212/Bird.php
+++ b/tests/Fixtures/TestBundle/ApiResource/Issue6212/Bird.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue6212;
+
+interface Bird extends \JsonSerializable
+{
+    public function getName(): ?string;
+
+    public function getAge(): ?int;
+}
--- a/tests/Fixtures/TestBundle/ApiResource/Issue6212/Robin.php
+++ b/tests/Fixtures/TestBundle/ApiResource/Issue6212/Robin.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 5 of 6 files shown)


tests/Fixtures/TestBundle/Entity/Issue6212/Nest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getOwner(): ?Bird
+    {
+        return $this->owner;
+    }
+
+    public function setOwner(Wren|Robin|null $owner): static
```
vincentchalamon, line 44:
```suggestion
    public function setOwner(?Bird $owner): static
```

tests/JsonSchema/Command/JsonSchemaGenerateCommandTest.php
```diff
@@ -124,6 +124,22 @@ public function testArraySchemaWithReference(): void
...
+        $json = json_decode($result, associative: true);
+
+        $this->assertEquals($json['definitions']['Nest.jsonld']['properties']['owner']['anyOf'], [
+            ['$ref' => '#/definitions/Wren.jsonld'],
+            ['$ref' => '#/definitions/Robin.jsonld'],
+            ['type' => 'null'],
+        ]);
+
+        $this->assertArrayHasKey('Wren.jsonld', $json['definitions']);
+        $this->assertArrayHasKey('Robin.jsonld', $json['definitions']);
```
vincentchalamon, line 140: I think this can be simply moved to `testArraySchemaWithReference`, as `testArraySchemaWithMultipleUnionTypes` is doing approximately the same

vincentchalamon (changes requested):
Minor fixes. Need to fix the CI, then LGTM

tests/Fixtures/TestBundle/Entity/Issue6212/Nest.php
GwendolenLynch: Agreed that in this tiny model that looks wrong, but it is just meant to illustrate the bug. In this case, the concrete classes implement the `Bird` interface, but the `Robin` classes have an extra property. Normally I'd use a second interface but didn't want to overload the example/test :innocent: 

As I understand the flow, the `PropertyAccessor` uses the types on the setter (as a first option), and the additional (non-`null`) properties are currently not properly handled in the schema factory. So it is only the setter in this case using that signature.

tests/JsonSchema/Command/JsonSchemaGenerateCommandTest.php
GwendolenLynch: I'm very willing to make changes, but am a bit confused as `Issue5793` resources don't have multiple type unions … or am I missing the obvious here?

GwendolenLynch:
@vincentchalamon I have fixed the tests that are relevant. The PHPUnit ones are currently an indirect deprecation notice from `doctrine/doctrine-bundle@2.12`, and the Mongo ones are: 
```
TypeError: Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver::__construct(): Argument #2 ($reader) must be of 
type ?Doctrine\Common\Annotations\Reader, Doctrine\ODM\MongoDB\Mapping\Driver\AttributeReader given
```

Can I get away without fixing those? :wink:

vincentchalamon:
> @vincentchalamon I have fixed the code that addresses tests that are relevant. The PHPUnit ones are currently an indirect deprecation notice from `doctrine/doctrine-bundle@2.12`, and the Mongo ones are:
> 
> ```
> TypeError: Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver::__construct(): Argument #2 ($reader) must be of 
> type ?Doctrine\Common\Annotations\Reader, Doctrine\ODM\MongoDB\Mapping\Driver\AttributeReader given
> ```
> 
> Can I get away without fixing those? 😉

Yep don't worry it's already blocked on 3.2 branch too :wink:

src/JsonSchema/SchemaFactory.php
```diff
@@ -230,14 +233,33 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
                 break;
             }
 
-            if ($type->isNullable()) {
-                $propertySchema['anyOf'] = [['$ref' => $subSchema['$ref']], ['type' => 'null']];
-            } else {
-                $propertySchema['$ref'] = $subSchema['$ref'];
+            $isNullable = $isNullable ?? $type->isNullable();
+            $refs[$subSchema['$ref']] = '$ref';
```
vincentchalamon, line 237: Value is never used as you're always calling `array_keys($refs)`. I guess you did it to ensure unicity, so I suggest the following:
```suggestion
            $ref = $subSchema['$ref'];
            if (!\in_array($ref, $refs, true)) {
                $refs[] = $ref;
            }
```
Then you can just use `$refs` instead of `array_keys($refs)`

vincentchalamon: You can also update `$isNullable` as following:
```php
if ($type->isNullable() && !\in_array(['type' => 'null'], $refs, true)) {
    $refs[] = ['type' => 'null'];
}
```

vincentchalamon: With those updates, you won't need `$isNullable` anymore on line 247, and won't need `$anyOf` neither

GwendolenLynch: Pushed the changes. Thank you so much for thinking this though with me, it is already a much better outcome!

src/JsonSchema/SchemaFactory.php
```diff
@@ -230,14 +233,17 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
-            }
+            $refs[] = ['$ref' => $subSchema['$ref']];
+            $isNullable = $isNullable ?? $type->isNullable();
+        }
+
+        if ($isNullable && !\in_array(['type' => 'null'], $refs, true)) {
+            $refs[] = ['type' => 'null'];
+        }
 
+        if (\count($refs) > 0) {
```
vincentchalamon, line 244: What if `\count($refs) === 1`?

vincentchalamon: Same: what if `$refs` is empty?

GwendolenLynch: That's mildly embarrassing, but fixed now. :blush:

src/JsonSchema/SchemaFactory.php
```diff
@@ -230,14 +233,19 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
+            $isNullable = $isNullable ?? $type->isNullable();
+        }
+
+        if ($isNullable) {
+            $refs[] = ['type' => 'null'];
+        }
 
+        if (\count($refs) > 1) {
+            $propertySchema['anyOf'] = $refs;
             unset($propertySchema['type']);
```
vincentchalamon, line 246: If `anyOf` or `$ref` is specified, `type` must be removed.

src/JsonSchema/SchemaFactory.php
```diff
@@ -230,14 +233,19 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
-            break;
+        } elseif (1 === \count($refs)) {
+            $propertySchema = $refs[0];
```
vincentchalamon, line 248: This erases other options in `$propertySchema`, I'm not sure it's safe. We should just add/override `$propertySchema['$ref']` option.

src/JsonSchema/SchemaFactory.php
GwendolenLynch: I'm confused again. :innocent: The `unset()` is removing `type` isn't it?

vincentchalamon: Yep!

vincentchalamon: But you can move this line before (or after) the if to prevent duplication with line 249

GwendolenLynch: Got it! Just to check your meaning though, as we don't want to unset `type` if `\count($refs) > 0`, is this whasat you're thinking?

```php
        if (\count($refs) > 1) {
            $propertySchema['anyOf'] = $refs;
        } elseif (1 === \count($refs)) {
            $propertySchema['$ref'] = $refs[0]['$ref'];
        }
        if (\count($refs) > 0) {
            unset($propertySchema['type']);
        }
```

For extra context, a resource with a `?int $property;` will have an empty `$refs` array, but `$propertySchema['type']` will contain `['integer', 'null']`

vincentchalamon: Ah nice catch! I didn't think about this specific use case.
In this case, indeed I would prefer to keep duplication to avoid useless code complexity!
You can keep your code as it is

vincentchalamon (approved):

src/JsonSchema/SchemaFactory.php
```diff
@@ -230,14 +233,20 @@ private function buildPropertySchema(Schema $schema, string $definitionName, str
...
-            }
+            $refs[] = ['$ref' => $subSchema['$ref']];
+            $isNullable = $isNullable ?? $type->isNullable();
+        }
 
+        if ($isNullable) {
+            $refs[] = ['type' => 'null'];
+        }
+
+        if (\count($refs) > 1) {
```
soyuka, line 244: can we keep the \count in a variable to avoid counting twice?z
```suggestion
        if (($c = \count($refs)) > 1) {
```

use `(1 === $c)` later on

GwendolenLynch: Done!

soyuka:
Thanks! ci failures not related

Issue #5926, opened 2023-10-24
IriConverter reuses unrelated root operation


irmantas:
**API Platform version(s) affected**: 3.2 (upgrade from 2.6)

**Description**  

Hi,

Upgrading from 2.6 to 3.2, do not know if this bug or old 2.6 behavior and won't be fixed. But I believe more people may encounter this if not fixed.
With API platform >2.7 I get an error `Unable to generate an IRI for the item of type` if Resource has a custom collection of resources or value object with other Resource without relation (like subresources) between them.

**How to reproduce**  

```php
<?php

#[ApiResource]
class Post
{
    public int $id;
    public ContentCollection $content;
    public function getContentCollection(): ContentCollection
    {
        return $this->content;
    }
}

// ---

class ContentCollection implements \IteratorAggregate
{
    private array $items;
    public function __construct(ContentItemInterface ...$items)
    {
        $this->items = $items;
    }
    public function getIterator(): \ArrayIterator
    {
        return new \ArrayIterator($this->items);
    }
}

class MediaContentItem implements ContentItemInterface
{
    public string $type;
    public Media $value;
}

//---
#[ApiResource]
class Media
{
    public int $id;
    /* ... */
}

```

**Possible Solution**  

https://github.com/api-platform/core/blob/e10cc5ecf88b98f862222b2e3fea0f4ace94c515/src/Symfony/Routing/IriConverter.php#L130

In my case solution is just to change this line, but I believe it may break other things as well.
```php
if (!$operation || ($operation->getClass() != $resourceClass)) {
```

soyuka:
if it's not a resource it goes through https://github.com/api-platform/core/blob/e10cc5ecf88b98f862222b2e3fea0f4ace94c515/src/Symfony/Routing/IriConverter.php#L122.

I'll test this though.

soyuka:
What was the Request that errored?

soyuka:
Can't reproduce: 
![20231027_14h13m21s_grim](https://github.com/api-platform/core/assets/1321971/a2cdc864-2694-4630-bf6d-0a03d449b6ec)

Used: 
```
#[Get(uriTemplate: 'issue5926', provider: [HasContent::class, 'provide'])]
class HasContent
{
    public int $id;
    public ContentCollection $content;

    public static function provide(Operation $operation, array $uriVariables = []): self
    {
        $y = new Media();
        $y->id = 2;

        $m = new MediaContentItem();
        $m->type = 'audio';
        $m->value = $y;

[...]

PR #6231, opened 2024-03-18
fix(serializer): json non-resource intermittent class (HAL & JSON:API)

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       |3.2
| Tickets       | #5926
| License       | MIT

This is meant as a follow up to #5937, as both the JSON:API & HAL serializers need the context check (c.f. #5937) in the `normalize()` method.

For reference here, without the code change the Behat test will fail with a 400 response that contains a "No resource class found […]" exception emitted from the `ResourceClassResolver`.


soyuka:
Thanks!

PR #6236, opened 2024-03-20
fix(jsonschema): generation of non-LD+JSON distinct schema formats

### LINKED ISSUE #6228: Issues generating multiple schema formats when decorating SchemaFactory

**Description**  
I am working on a PR to introduce a JSON:API schema factory and hit a couple of related bugs that feel better suited to a separate issue/PR combination.

I have working "fixes" laid out below, but would like them reviewed & sanity checked before I create a PR for this.

First problem was that `ApiPlatform\Hydra\JsonSchema\SchemaFactory#addDistinctFormat()` checks for `ApiPlatform\JsonSchema\SchemaFactory`, where as `ApiPlatform\Hal\JsonSchema\SchemaFactory` simply does a `method_exists` call.

No `priority` is set on any of the service decorators, and Hydra's, by chance, is currently the one with the base service  `.inner`. However, my new service took first place and it became impossible to build both schema formats at the same time as the last one "won out". Changing Hydra's method behaviour to be a `method_exists` call fixes this.

The second is just as subtle. The base `ApiPlatform\JsonSchema\SchemaFactory` calls the injected decorator service to enable subschema generation (see #6055). When the Hydra service is the one decorating the base this works fine, but the introduction of another service causes that one to be called instead, meaning Hydra misses out (see #5950)

**How to reproduce**  

```php
final class SchemaFactory implements SchemaFactoryInterface
{
    public function __construct(private readonly SchemaFactoryInterface $schemaFactory)
    {
        $this->addDistinctFormat('jsonapi');
        if ($this->schemaFactory instanceof SchemaFactoryAwareInterface) {
            $this->schemaFactory->setSchemaFactory($this);
        }
    }

    public function buildSchema(string $className, string $format = 'jsonapi', string $type = Schema::TYPE_OUTPUT, ?Operation $operation = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
    {
        $schema = $this->schemaFactory->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);

        if ('jsonapi' !== $format) {
            return $schema;
        }

        // Do schema updates here …

        return $schema;
    }

    public function addDistinctFormat(string $format): void
    {
        // The Hydra schema factory does:
        //  if ($this->schemaFactory instanceof SchemaFactoryAwareInterface) {
        //
        // The HAL one does as below:
        if (method_exists($this->schemaFactory, 'addDistinctFormat')) {
            $this->schemaFactory->addDistinctFormat($format);
        }
    }
}
```

Obligatory service definition :innocent: 
```yaml
    api_platform.jsonapi.json_schema.schema_factory:
        class: 'ApiPlatform\JsonApi\JsonSchema\SchemaFactory'
        decorates: 'api_platform.json_schema.schema_factory'
        arguments:
            - '@api_platform.jsonapi.json_schema.schema_factory.inner'
```

And API Platform config:
```
api_platform:
    formats:
        json: [ 'application/json' ]
        jsonld: [ 'application/ld+json' ]
        jsonhal: [ 'application/hal+json' ]
        jsonapi: [ 'application/vnd.api+json' ]
```

**Possible Solution**  
1. Change Hydra `SchemaFactory#addDistinctFormat()` to use a `method_exists` call
2. Have all `SchemaFactory`s implement `SchemaFactoryAwareInterface` :disappointed: 

Obviously neither of those are particularly elegant, hence this discussion. Help and guidance most welcome!

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #6228
| License       | MIT

I've not figured out a good way to CI test this, so guidance will be gratefully appreciated.

To reproduce the base issue, apply the following to the `3.2` branch and run `vendor/bin/simple-phpunit --filter JsonSchemaGenerateCommandTest::testSubSchemaJsonLd`

```diff
diff --git a/src/Hydra/JsonSchema/SchemaFactory.php b/src/Hydra/JsonSchema/SchemaFactory.php
index b04ba616b8..43478de74d 100644
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -186,7 +186,7 @@ public function buildSchema(string $className, string $format = 'jsonld', string
 
     public function addDistinctFormat(string $format): void
     {
-        if ($this->schemaFactory instanceof BaseSchemaFactory) {
+        if (method_exists($this->schemaFactory, 'addDistinctFormat')) {
             $this->schemaFactory->addDistinctFormat($format);
         }
     }
diff --git a/src/JsonApi/JsonSchema/SchemaFactory.php b/src/JsonApi/JsonSchema/SchemaFactory.php
new file mode 100644
index 0000000000..b9747c281d
--- /dev/null
+++ b/src/JsonApi/JsonSchema/SchemaFactory.php
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ApiPlatform\JsonApi\JsonSchema;
+
+use ApiPlatform\JsonSchema\Schema;
+use ApiPlatform\JsonSchema\SchemaFactoryAwareInterface;
+use ApiPlatform\JsonSchema\SchemaFactoryInterface;
+use ApiPlatform\Metadata\Operation;
+
+final class SchemaFactory implements SchemaFactoryInterface
+{
+    public function __construct(private SchemaFactoryInterface $schemaFactory)
+    {
+        $this->addDistinctFormat('jsonapi');
+        if ($this->schemaFactory instanceof SchemaFactoryAwareInterface) {
+            $this->schemaFactory->setSchemaFactory($this);
+        }
+    }
+
+    public function buildSchema(string $className, string $format = 'jsonapi', string $type = Schema::TYPE_OUTPUT, ?Operation $operation = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
+    {
+        $schema = $this->schemaFactory->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
+
+        if ('jsonapi' !== $format) {
+            return $schema;
+        }
+
+        // Do schema updates here …
+
+        return $schema;
+    }
+
+    public function addDistinctFormat(string $format): void
+    {
+        if (method_exists($this->schemaFactory, 'addDistinctFormat')) {
+            $this->schemaFactory->addDistinctFormat($format);
+        }
+    }
+}
diff --git a/src/Symfony/Bundle/Resources/config/jsonapi.xml b/src/Symfony/Bundle/Resources/config/jsonapi.xml
index 0a84066acc..98f1b08bc6 100644
--- a/src/Symfony/Bundle/Resources/config/jsonapi.xml
+++ b/src/Symfony/Bundle/Resources/config/jsonapi.xml
@@ -5,6 +5,10 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
+        <service id="api_platform.jsonapi.json_schema.schema_factory" class="ApiPlatform\JsonApi\JsonSchema\SchemaFactory" decorates="api_platform.json_schema.schema_factory">
+            <argument type="service" id="api_platform.jsonapi.json_schema.schema_factory.inner" />
+        </service>
+
         <service id="api_platform.jsonapi.encoder" class="ApiPlatform\Serializer\JsonEncoder" public="false">
             <argument>jsonapi</argument>
 
```

### DIFF
--- a/src/Hal/JsonSchema/SchemaFactory.php
+++ b/src/Hal/JsonSchema/SchemaFactory.php
@@ -46,7 +46,6 @@ final class SchemaFactory implements SchemaFactoryInterface
 
     public function __construct(private readonly SchemaFactoryInterface $schemaFactory)
     {
-        $this->addDistinctFormat('jsonhal');
         if ($this->schemaFactory instanceof SchemaFactoryAwareInterface) {
             $this->schemaFactory->setSchemaFactory($this);
         }
@@ -125,12 +124,5 @@ final class SchemaFactory implements SchemaFactoryInterface
         }
 
         return $schema;
-    }
-
-    public function addDistinctFormat(string $format): void
-    {
-        if (method_exists($this->schemaFactory, 'addDistinctFormat')) {
-            $this->schemaFactory->addDistinctFormat($format);
-        }
     }
 }
--- a/src/Hal/JsonSchema/SchemaFactory.php
+++ b/src/Hal/JsonSchema/SchemaFactory.php
@@ -78,8 +78,12 @@ final class SchemaFactory implements SchemaFactoryInterface
             $schema['type'] = 'object';
             $schema['properties'] = [
                 '_embedded' => [
-                    'type' => 'array',
-                    'items' => $items,
+                    'properties' => [
+                        'items' => [
+                            'type' => 'array',
+                            'items' => $items,
+                        ],
+                    ],
                 ],
                 'totalItems' => [
                     'type' => 'integer',
--- a/src/Hal/JsonSchema/SchemaFactory.php
+++ b/src/Hal/JsonSchema/SchemaFactory.php
@@ -78,11 +78,17 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
             $schema['type'] = 'object';
             $schema['properties'] = [
                 '_embedded' => [
-                    'properties' => [
-                        'items' => [
-                            'type' => 'array',
-                            'items' => $items,
+                    'anyOf' => [
+                        [
+                            'type' => 'object',
+                            'properties' => [
+                                'item' => [
+                                    'type' => 'array',
+                                    'items' => $items,
+                                ],
+                            ],
                         ],
+                        ['type' => 'object'],
                     ],
                 ],
                 'totalItems' => [
--- a/src/Hal/JsonSchema/SchemaFactory.php
+++ b/src/Hal/JsonSchema/SchemaFactory.php
@@ -24,7 +24,7 @@ use ApiPlatform\Metadata\Operation;
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Jachim Coudenys <jachimcoudenys@gmail.com>
  */
-final class SchemaFactory implements SchemaFactoryInterface
+final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
 {
     private const HREF_PROP = [
         'href' => [
@@ -128,5 +128,12 @@ final class SchemaFactory implements SchemaFactoryInterface
         }
 
         return $schema;
+    }
+
+    public function setSchemaFactory(SchemaFactoryInterface $schemaFactory): void
+    {
+        if ($this->schemaFactory instanceof SchemaFactoryAwareInterface) {
+            $this->schemaFactory->setSchemaFactory($schemaFactory);
+        }
     }
 }
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -36,14 +36,13 @@ use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
 final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
 {
     use ResourceClassInfoTrait;
-    private array $distinctFormats = [];
     private ?TypeFactoryInterface $typeFactory = null;
     private ?SchemaFactoryInterface $schemaFactory = null;
     // Edge case where the related resource is not readable (for example: NotExposed) but we have groups to read the whole related object
     public const FORCE_SUBSCHEMA = '_api_subschema_force_readable_link';
     public const OPENAPI_DEFINITION_NAME = 'openapi_definition_name';
 
-    public function __construct(?TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ?ResourceClassResolverInterface $resourceClassResolver = null)
+    public function __construct(?TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ?ResourceClassResolverInterface $resourceClassResolver = null, private readonly ?array $distinctFormats = null)
     {
         if ($typeFactory) {
             $this->typeFactory = $typeFactory;
@@ -51,16 +50,6 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
 
         $this->resourceMetadataFactory = $resourceMetadataFactory;
         $this->resourceClassResolver = $resourceClassResolver;
-    }
-
-    /**
-     * When added to the list, the given format will lead to the creation of a new definition.
-     *
-     * @internal
-     */
-    public function addDistinctFormat(string $format): void
-    {
-        $this->distinctFormats[$format] = true;
     }
 
     /**
@@ -267,7 +256,7 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
             $prefix .= '.'.$shortName;
         }
 
-        if (isset($this->distinctFormats[$format])) {
+        if ('json' !== $format && ($this->distinctFormats[$format] ?? false)) {
             // JSON is the default, and so isn't included in the definition name
             $prefix .= '.'.$format;
         }
--- a/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
+++ b/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
@@ -120,6 +120,16 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $patchFormats = $this->getFormats($config['patch_formats']);
         $errorFormats = $this->getFormats($config['error_formats']);
         $docsFormats = $this->getFormats($config['docs_formats']);
+        $jsonSchemaFormats = $config['jsonschema_formats'];
+
+        if (!$jsonSchemaFormats) {
+            foreach (array_keys($formats) as $f) {
+                // Distinct JSON-based formats must have names that start with 'json'
+                if (str_starts_with($f, 'json')) {
+                    $jsonSchemaFormats[$f] = true;
+                }
+            }
+        }
 
         if (!isset($errorFormats['json'])) {
             $errorFormats['json'] = ['application/problem+json', 'application/json'];
@@ -144,7 +154,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
             $docsFormats['jsonopenapi'] = ['application/vnd.openapi+json'];
         }
 
-        $this->registerCommonConfiguration($container, $config, $loader, $formats, $patchFormats, $errorFormats, $docsFormats);
+        $this->registerCommonConfiguration($container, $config, $loader, $formats, $patchFormats, $errorFormats, $docsFormats, $jsonSchemaFormats);
         $this->registerMetadataConfiguration($container, $config, $loader);
         $this->registerOAuthConfiguration($container, $config);
         $this->registerOpenApiConfiguration($container, $config, $loader);
@@ -185,7 +195,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $this->registerInflectorConfiguration($config);
     }
 
-    private function registerCommonConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader, array $formats, array $patchFormats, array $errorFormats, array $docsFormats): void
+    private function registerCommonConfiguration(ContainerBuilder $container, array $config, XmlFileLoader $loader, array $formats, array $patchFormats, array $errorFormats, array $docsFormats, array $jsonSchemaFormats): void
     {
         $loader->load('symfony/events.xml');
         $loader->load('symfony/controller.xml');
@@ -218,6 +228,7 @@ final class ApiPlatformExtension extends Extension implements PrependExtensionIn
         $container->setParameter('api_platform.patch_formats', $patchFormats);
         $container->setParameter('api_platform.error_formats', $errorFormats);
         $container->setParameter('api_platform.docs_formats', $docsFormats);
+        $container->setParameter('api_platform.jsonschema_formats', $jsonSchemaFormats);
         $container->setParameter('api_platform.eager_loading.enabled', $this->isConfigEnabled($container, $config['eager_loading']));
         $container->setParameter('api_platform.eager_loading.max_joins', $config['eager_loading']['max_joins']);
         $container->setParameter('api_platform.eager_loading.fetch_partial', $config['eager_loading']['fetch_partial']);
--- a/src/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -174,6 +174,14 @@ final class Configuration implements ConfigurationInterface
             'jsonproblem' => ['mime_types' => ['application/problem+json']],
             'json' => ['mime_types' => ['application/problem+json', 'application/json']],
         ]);
+        $rootNode
+            ->children()
+                ->arrayNode('jsonschema_formats')
+                    ->scalarPrototype()->end()
+                    ->defaultValue([])
+                    ->info('The JSON formats to compute the JSON Schemas for.')
+                ->end()
+            ->end();
 
         $this->addDefaultsSection($rootNode);
 
--- a/src/Symfony/Bundle/Resources/config/json_schema.xml
+++ b/src/Symfony/Bundle/Resources/config/json_schema.xml
@@ -22,6 +22,7 @@
             <argument type="service" id="api_platform.metadata.property.metadata_factory" />
             <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
             <argument type="service" id="api_platform.resource_class_resolver" />
+            <argument on-invalid="ignore">%api_platform.jsonschema_formats%</argument>
         </service>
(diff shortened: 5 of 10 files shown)


src/Hal/JsonSchema/SchemaFactory.php
```diff
@@ -79,8 +78,21 @@ public function buildSchema(string $className, string $format = 'jsonhal', strin
...
+                                    'type' => 'array',
+                                    'items' => $items,
+                                ],
+                            ],
+                        ],
+                        [
+                            'type' => 'array',
+                            'items' => $items,
+                        ],
+                    ],
```
GwendolenLynch, line 95: I'm not 100% sure if this should be an `anyOf`, or just the `object` … Plus shouldn't `items` the the resource name?

src/JsonSchema/SchemaFactory.php
```diff
@@ -54,13 +53,10 @@ public function __construct(?TypeFactoryInterface $typeFactory, ResourceMetadata
     }
 
     /**
-     * When added to the list, the given format will lead to the creation of a new definition.
-     *
      * @internal
```
soyuka, line 56: this was already `@internal` so you can remove the method alltogether

src/Symfony/Bundle/DependencyInjection/Compiler/SchemaFactoryPass.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container): void
+    {
+        if ($container->hasDefinition('api_platform.json_schema.schema_factory')) {
+            $container->getDefinition('api_platform.json_schema.schema_factory')->setArgument(6, '%api_platform.jsonschema_formats%');
+        }
+    }
+}
```
soyuka, line 35: we don't need a compiler pass for that right?

src/Symfony/Bundle/Resources/config/json_schema.xml
```diff
@@ -22,6 +22,7 @@
             <argument type="service" id="api_platform.metadata.property.metadata_factory" />
             <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
             <argument type="service" id="api_platform.resource_class_resolver" />
+            <argument type="collection" />
```
soyuka, line 25: we can safely inject the parameter here + on-invalid="ignore", it defaults to an empty array anyways in the constructor

src/Hal/JsonSchema/SchemaFactory.php
soyuka: The `items` thing comes from: https://json-schema.org/understanding-json-schema/reference/array#items

But in `hal` `_embedded` is always an object right?

GwendolenLynch: Ah, good to know! Thanks you.

I believe it is, I just looked at `ApiPlatform\Hal\Serializer\CollectionNormalizer::getItemsData()` and it is, I'll adjust the above change.

src/Symfony/Bundle/DependencyInjection/Compiler/SchemaFactoryPass.php
GwendolenLynch: Removed

src/Symfony/Bundle/Resources/config/json_schema.xml
GwendolenLynch: Done. It is a much better idea!

src/Hal/JsonSchema/SchemaFactory.php
```diff
@@ -79,8 +78,15 @@ public function buildSchema(string $className, string $format = 'jsonhal', strin
...
-                    'items' => $items,
+                    [
+                        'type' => 'object',
+                        'properties' => [
+                            'items' => [
+                                'type' => 'array',
+                                'items' => $items,
+                            ],
+                        ],
+                    ],
```
soyuka, line 89: something's bothering me here I need to check that, I don't understand why we put `array` in the first place.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -120,6 +120,9 @@ public function load(array $configs, ContainerBuilder $container): void
         $patchFormats = $this->getFormats($config['patch_formats']);
         $errorFormats = $this->getFormats($config['error_formats']);
         $docsFormats = $this->getFormats($config['docs_formats']);
+        $jsonSchemaFormats = $config['jsonschema_formats']
+            ? $this->getFormats($config['jsonschema_formats'])
+            : array_filter($formats, static fn ($format) => !\in_array($format, ['html', 'multipart', 'xml'], true), \ARRAY_FILTER_USE_KEY);
```
soyuka, line 125: mhh maybe that we could use a list of documentation-allowed formats instead? `jsonld, jsonapi, jsonhal` wdyt?

GwendolenLynch: OK, if it was my code I'd jump at the "documentation-allowed formats" … until I use it on *"that project"* where you need a custom format, and then wonder why it won't fit with my idea of what I want :innocent:  

I don't like this approach, I don't want to defend it, but we do know what defaults we already have that are not "distinct". 100% happy to apply a better solution though.

src/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -174,6 +174,8 @@ public function getConfigTreeBuilder(): TreeBuilder
             'jsonproblem' => ['mime_types' => ['application/problem+json']],
             'json' => ['mime_types' => ['application/problem+json', 'application/json']],
         ]);
+        $this->addFormatSection($rootNode, 'jsonschema_formats', [
+        ]);
```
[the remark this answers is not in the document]
GwendolenLynch: That should be addressed in the `ApiPlatformExtension` change above. Basically, site-configured value *or* a filtered `formats:` value

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
soyuka, line 125: Agree, let's try this (pseudo code, use array_filter if you prefer I don't mind):

```suggestion
        $jsonSchemaFormats = $config['jsonschema_formats'];
        
        if (!$jsonSchemaFormats) {
            foreach (array_keys($formats) as $f) {
              if (str_starts_with('json', $f) { 
                $jsonSchemaFormats[] = $f;
              }
            }
        }
```

We don't need mime types for this so let's get rid on the `getFormats`, and if not defined by the user we want this to be defined for all the supported json formats.

src/Symfony/Bundle/DependencyInjection/Configuration.php
soyuka, line 178:
```suggestion
                ->arrayNode('jsonschema_formats')
                    ->scalarPrototype()->end()
                    ->defaultValue([])
                    ->info('The JSON formats to compute the JSON Schemas for.')
                ->end()
```

src/Hal/JsonSchema/SchemaFactory.php
GwendolenLynch: Introduced in #4357, as was `addDistinctFormat()`

GwendolenLynch:
@soyuka I have also updated more of the tests in `ApiTestCaseTest` to use the formats data provider. I'm a bit nervous about coverage for non-LD+JSON formats.

If you want to suggest other test locations that I can expand on, I'll happily work away on them.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -120,6 +120,16 @@ public function load(array $configs, ContainerBuilder $container): void
         $patchFormats = $this->getFormats($config['patch_formats']);
         $errorFormats = $this->getFormats($config['error_formats']);
         $docsFormats = $this->getFormats($config['docs_formats']);
+        $jsonSchemaFormats = $config['jsonschema_formats'];
```
GwendolenLynch, line 123: During my digging though the codebase I noticed a couple of places where `json_schema` is used over `jsonschema`. Should I change this to add the `_`?

soyuka: I like it like this.

src/Hal/JsonSchema/SchemaFactory.php
soyuka: https://datatracker.ietf.org/doc/html/draft-kelly-json-hal#section-4.1.2 

> It is an object whose property names are link relation types (as defined by [[RFC5988](https://www.rfc-editor.org/info/rfc5988)]) and values are either a Resource Object or an array of Resource Objects.

can you revert this change? I think that our HAL representation always uses an array and we use `items` as key.

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -120,6 +120,16 @@ public function load(array $configs, ContainerBuilder $container): void
...
+
+        if (!$jsonSchemaFormats) {
+            foreach (array_keys($formats) as $f) {
+                // Distinct JSON-based formats must have names that start with 'json'
+                if (str_starts_with($f, 'json') && 'json' !== $f) {
```
soyuka, line 128:
```suggestion
                if (str_starts_with($f, 'json')) {
```

we can keep standard json no?

GwendolenLynch: Yes we do. The goal was to work with the behaviour of `distinctFormats` here:
https://github.com/api-platform/core/blob/4f93952f9201a46416f8d5b21eae440553a86db3/src/JsonSchema/SchemaFactory.php#L259-L262

I'll just shift the logic into the factory, ergo:
```
if ($format !== 'json' && $this->distinctFormats[$format] ?? false) {
```

Otherwise we end up with a bunch of `ResourceName.json` schema components.

src/Hal/JsonSchema/SchemaFactory.php
GwendolenLynch: Doesn't that make the behaviour in `ApiPlatform\Hal\Serializer\CollectionNormalizer` incorrect?

https://github.com/api-platform/core/blob/874e4d6707ecdd48b27ed91b0b66ea5e33691e7c/src/Hal/Serializer/CollectionNormalizer.php#L86-L100

The changes in `ApiTestCaseTest` (testing `jsonld` and `jsonhal` as formats) trigger a failure condition without this change:
```
1) ApiPlatform\Tests\Symfony\Bundle\Test\ApiTestCaseTest::testAssertMatchesResourceCollectionJsonSchema with data set "jsonhal" ('jsonhal', 'application/hal+json')
Failed asserting that Array &0 (
    '_links' => Array &1 (
        'self' => Array &2 (
            'href' => '/resource_interfaces'
        )
        'item' => Array &3 (
            0 => Array &4 (
                'href' => '/resource_interfaces/item1'
            )
            1 => Array &5 (
                'href' => '/resource_interfaces/item2'
            )
        )
    )
    '_embedded' => Array &6 (
        'item' => Array &7 (
            0 => Array &8 (
                '_links' => Array &9 (
                    'self' => Array &10 (
                        'href' => '/resource_interfaces/item1'
                    )
                )
                'foo' => 'item1'
                'fooz' => 'fooz'
            )
            1 => Array &11 (
                '_links' => Array &12 (
                    'self' => Array &13 (
                        'href' => '/resource_interfaces/item2'
                    )
                )
                'foo' => 'item2'
                'fooz' => 'fooz'
            )
        )
    )
) matches the provided JSON Schema.
_embedded: Object value found, but an array is required

2) ApiPlatform\Tests\Symfony\Bundle\Test\ApiTestCaseTest::testAssertMatchesResourceCollectionJsonSchemaKeepSerializationContext with data set "jsonhal" ('jsonhal', 'application/hal+json')
Failed asserting that Array &0 (
    '_links' => Array &1 (
        'self' => Array &2 (
            'href' => '/issue-6146-parents'
        )
        'item' => Array &3 (
            0 => Array &4 (
                'href' => '/issue-6146-parents/1'
            )
        )
    )
    'totalItems' => 1
    'itemsPerPage' => 3
    '_embedded' => Array &5 (
        'item' => Array &6 (
            0 => Array &7 (
                '_links' => Array &8 (
                    'self' => Array &9 (
                        'href' => '/issue-6146-parents/1'
                    )
                )
            )
        )
    )
) matches the provided JSON Schema.
_embedded: Object value found, but an array is required

3) ApiPlatform\Tests\Symfony\Bundle\Test\ApiTestCaseTest::testAssertMatchesResourceItemAndCollectionJsonSchemaOutputWithContext with data set "jsonhal" ('jsonhal', 'application/hal+json')
Failed asserting that Array &0 (
    '_links' => Array &1 (
        'self' => Array &2 (
            'href' => '/users-with-groups'
        )
        'item' => Array &3 (
            0 => Array &4 (
                'href' => '/users/1'
            )
        )
    )
    'totalItems' => 1
    'itemsPerPage' => 3
    '_embedded' => Array &5 (
        'item' => Array &6 (
            0 => Array &7 (
                '_links' => Array &8 (
                    'self' => Array &9 (
                        'href' => '/users/1'
                    )
                )
                'fullname' => 'Grégoire'
            )
        )
    )
) matches the provided JSON Schema.
_embedded: Object value found, but an array is required
```

GwendolenLynch: From https://datatracker.ietf.org/doc/html/draft-kelly-json-hal#section-6

```json
{
  "_links": {
    "self": { "href": "/orders" },
    "next": { "href": "/orders?page=2" },
    "find": { "href": "/orders{?id}", "templated": true }
  },
  "_embedded": {
    "orders": [{
        "_links": {
          "self": { "href": "/orders/123" },
          "basket": { "href": "/baskets/98712" },
          "customer": { "href": "/customers/7809" }
        },
        "total": 30.00,
        "currency": "USD",
        "status": "shipped",
      },{
        "_links": {
          "self": { "href": "/orders/124" },
          "basket": { "href": "/baskets/97213" },
          "customer": { "href": "/customers/12369" }
        },
        "total": 20.00,
        "currency": "USD",
        "status": "processing"
    }]
  },
  "currentlyProcessing": 14,
  "shippedToday": 20
}
```

Which also comes back to a previous question as to the key(s) of `_embedded` … we use `items` but the examples in the draft seem to use the resource name.

soyuka: To be correct the schema should be `anyOf` object or array. We force `item` at https://github.com/api-platform/core/blob/874e4d6707ecdd48b27ed91b0b66ea5e33691e7c/src/Hal/Serializer/CollectionNormalizer.php#L95 therefore we need the key at `item` as I see it.

GwendolenLynch: As in how I did it originally in https://github.com/api-platform/core/commit/97f7dff15db98f7cd0e8510fc828557f35513397?

GwendolenLynch: Ugh, never mind, it clicked what you meant … now I (think) I get you :innocent:

GwendolenLynch: Looking too hard at the same thing for too long and got my brain in a twist.

soyuka: Right now we may have (as this is what we use for our collections):

```
'anyOf' => [
    [
        'type' => 'object',
        'properties' => [
            'item' => [
                'type' => 'array',
                'items' => $items,
            ],
        ],
    ],
    [ 'type' => 'object' ]
],
```

Although the spec says that it can also be

```
  [
        'type' => 'array'
  ],
```

I'm okay with adding it if you want

soyuka:
Thanks @GwendolenLynch ! Feel free to propose the `array` schema in another PR if you think we should add it, I wanted this patch in the next release so I took the liberty of adding back the `anyOf`.

PR #6270, opened 2024-03-30
fix(jsonapi): return empty data array for empty relation

### LINKED ISSUE #6265: JSONApi Normalizer: Empty has many relation should return empty array
Hello,

the item normalizer of api-platform for jsonapi does not comply the specification. If we have an empty has-many relations it should not skip the relationship. Instead it should return an empty array:

```
relation: {
  data: []
}   
```
otherwise the consumer can not distinguish if the relations was not included or if there is simply no relations assigned. At least it is not that simple...

I didn't look too deeply into it yet, but i assume the safest fix would be checking if we have a many-to-may or 1:n relation and in such case return an empty array if value is empty.

https://github.com/api-platform/core/blob/e895b69a04cb52eb6eb030bcc51a1e143dd357cb/src/JsonApi/Serializer/ItemNormalizer.php#L406

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes #6265
| License       | MIT

Additions to the expectations in the `.feature` files are the missing data property arrays mentioned in the spec/issue.

### DIFF
--- a/features/http_cache/tag_collector_service.feature
+++ b/features/http_cache/tag_collector_service.feature
@@ -107,6 +107,9 @@ Feature: Cache invalidation through HTTP Cache tags (custom TagCollector service
                         "type": "RelatedDummy",
                         "id": "/related_dummies/1"
                     }
+                },
+                "related": {
+                    "data": []
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


soyuka:
can we target `main` for this one? it'll potentially break some code userland and as 3.3 is close to be released I think it's safer.

GwendolenLynch:
Updated. I'm tracking main anyway … it's all about JSON:API for me at the moment.

soyuka:
Nice thanks! I never really dig into JSON API nice of you to contribute the missing part!

PR #6250, opened 2024-03-27
feat(jsonschema): JSON:API schema factory

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #3930
| License       | MIT
| Doc PR        | api-platform/docs#... 

~**NOTE:** This depends on recent commits in `3.2` that haven't been merged down to `main` yet. So it is submitted against `3.2` and will be rebased & updated to target `main` after said branch merge.~

PR #6273, opened 2024-03-31
fix(jsonapi): correct data path

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | Related #6250
| License       | MIT

This is a follow-up fix for a mistake I made with #6250 that became apparent when testing more of the new parameter stuff.


src/JsonApi/JsonSchema/SchemaFactory.php
```diff
@@ -119,11 +119,7 @@ public function buildSchema(string $className, string $format = 'jsonapi', strin
             return $schema;
         }
 
-        if ('input' === $type) {
-            return $schema;
-        }
```
GwendolenLynch, line 124: See https://jsonapi.org/format/#crud-updating

soyuka:
thanks!

Issue #6225, opened 2024-03-16
allow_extra_attributes and update nested entity @id


ili101:
**Description**  
As mentioned in documentation https://api-platform.com/docs/core/serialization/#denormalization
```
The following rules apply when denormalizing embedded relations:

    If an @id key is present in the embedded resource, then the object corresponding to the given URI will be retrieved through the state provider. Any changes in the embedded relation will also be applied to that object.
    If no @id key exists, a new object will be created containing state provided in the embedded JSON document.
```
When setting `allow_extra_attributes: false` this results in `Extra attributes are not allowed ("@id" is unknown).`
If using json and not ld+json and sending `id` instead you get the same error on `id`.
P.S. not sure if intentional but for the update to actually work even if extra attributes allowed `@id` needs to also be provided on the parent entity.

**How to reproduce**  
Test example https://github.com/ili101/api-platform/blob/subTest/Test.ps1 (uncomment `allow_extra_attributes` in api_platform.yaml)

**Possible Solution**  
Allow `@id` (or maybe also `@type` `@context`) with `allow_extra_attributes: false` and ld+json.
Allow  `id` with `allow_extra_attributes: false` and json.

Alternatively for more flexible solution, in the Entry ApiProperty or somewhere else add and option to set the base `$allowedAttributes` with default values here:
https://github.com/api-platform/core/blob/5523bf5df93783582bed5591ff35c71e0942a978/src/Serializer/AbstractItemNormalizer.php#L423
Something like:
```php
    #[ApiProperty(
        ExtraAllowedExtraAttributes: ['@id', '@type'],
    )]
```

**Additional Context**  
https://github.com/api-platform/api-platform/issues/2096
https://github.com/api-platform/api-platform/issues/2620
https://stackoverflow.com/questions/70379382/how-to-update-nested-entity-with-api-platform

soyuka:
specify this inside the normalizerContext?

ili101:
> specify this inside the normalizerContext?

Yes, https://github.com/ili101/api-platform-core/commit/7b84701c055f7c2cb972defea220682fdb17a913 then you can:
```php
#[ApiResource(
    denormalizationContext: [
        AbstractNormalizer::GROUPS => ['Main:write'],
        'additional_allowed_attributes' => ['@id', '@type', 'id', '@context'],
    ],
)]
```

soyuka:
can you let me know if my patch works

ili101:
Looks good

PR #6274, opened 2024-03-31
additional_allowed_attributes

ili101:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes https://github.com/api-platform/core/issues/6225
| License       | MIT

Tested it on my project and it works and solves my use case and the id case from https://api-platform.com/docs/core/serialization/#denormalization. But then tried to add to test tests/JsonLd/Serializer/ItemNormalizerTest.php something like:
```php
        $context = [
            'allow_extra_attributes' => false,
            'additional_allowed_attributes' => ['@type', 'extra'],
        ];
        $data = [
            // '@context' => '/contexts/Dummy',
            // '@id' => '/dummies/1988',
            '@type' => 'Dummy',
            'name' => 'hello',
            'extra' => 'hello',
        ];
        $normalizer->denormalize($data, Dummy::class, context: $context);
``` 
But the test fails not sure why. I probably set the test wrong?

### DIFF
--- a/src/Serializer/AbstractItemNormalizer.php
+++ b/src/Serializer/AbstractItemNormalizer.php
@@ -420,7 +420,7 @@ abstract class AbstractItemNormalizer extends AbstractObjectNormalizer
         $options = $this->getFactoryOptions($context);
         $propertyNames = $this->propertyNameCollectionFactory->create($resourceClass, $options);
 
-        $allowedAttributes = [];
+        $allowedAttributes = $context['additional_allowed_attributes'] ?? [];
         foreach ($propertyNames as $propertyName) {
             $propertyMetadata = $this->propertyMetadataFactory->create($resourceClass, $propertyName, $options);
 


src/Serializer/AbstractItemNormalizer.php
```diff
@@ -420,7 +420,7 @@ protected function getAllowedAttributes(string|object $classOrObject, array $con
         $options = $this->getFactoryOptions($context);
         $propertyNames = $this->propertyNameCollectionFactory->create($resourceClass, $options);
 
-        $allowedAttributes = [];
+        $allowedAttributes = $context['additional_allowed_attributes'] ?? [];
```
soyuka, line 423: is this a symfony option?

ili101: No, `allow_extra_attributes` is Symfony, and `additional_allowed_attributes` is something I added.
As I can see Symfony AbstractObjectNormalizer is extended by API-Platform AbstractItemNormalizer where `getAllowedAttributes()` is overeaten. I added there a new `additional_allowed_attributes` option to allow additional attributes when `allow_extra_attributes` is set.
I don't see any bulletin option to do this (maybe I missed it?) and later AbstractItemNormalizer extended again and made final so `getAllowedAttributes()` cannot be modified again by the API-Platform user.

soyuka: Isn't `allow_extra_attributes` working?

ili101: `allow_extra_attributes` is Symfony way of making the system throw a validation error if the API has extra attributes. It works and is needed for `additional_allowed_attributes` to have a meaning.
`additional_allowed_attributes` comes to solve the problem of when you need to whitelist some attributes when using `allow_extra_attributes = false`.

For example, as described in https://api-platform.com/docs/core/serialization/#denormalization
> If an @id key is present in the embedded resource, then the object corresponding to the given URI will be retrieved through the state provider. Any changes in the embedded relation will also be applied to that object.

When trying to do this in combination with `allow_extra_attributes = false` you will get an error that `@id` is not valid as it doesn't exist on the object. to solve this you can add, in addition, the config `'additional_allowed_attributes' => ['@id']` to tell the deserializer that even that extra attributes are not allowed, and the object doesn't have an `@id` property, allow it anyway.

Additionally, this also lets you whitelist other attributes you wish to allow while using `allow_extra_attributes = false`, for example `'additional_allowed_attributes' => ['@id', '@type', 'id', '@context', 'extraParam']`

soyuka: Okay I understand now, so you want `allow_extra_attributes` to `false` but denormalize a JSON-LD document. I think that this should work from scratch, @dunglas any thoughts?

ili101: Possibly. My use case is basically I want to POST a new record `/action` (including sub-records) and then edit the response and POST it again like `/action/1/submit`.
So for the sub-records to work I must have the `@id`s. Additionally, for ease of use, I also like to allow the other stuff so the user will not need to remove it while resubmitting. So in my use-case with this fix I used `'additional_allowed_attributes' => ['@id', '@type', 'id', '@context']`.
So allowing `['@id', '@type', 'id', '@context']` out of the box or adding `'additional_allowed_attributes'` dynamic option, or possibly both so it will work out of the box and be extendable works for me.

ili101: Appreciate it if this (`additional_allowed_attributes`) gets approved or alternatively, if you want to go with a hardcoded solution as @soyuka suggested like allowing `@id` or `['@id', '@type', 'id', '@context']` in src/JsonLd/Serializer/ItemNormalizer.php and allowing `id` in src/JsonApi/Serializer/ItemNormalizer.php in some way.
@dunglas @soyuka

soyuka: I'd like to avoid a new serialization context, can you check https://github.com/api-platform/core/pull/6402 ?

Issue #6210, opened 2024-03-11
Accept application/x-www-form-urlencoded Form Data in custom DeserializeListener is not working anymore


menhir2015:
**API Platform version(s) affected**:3.2.16

**Description**  
i m upgrading my app from symfony from 6.3.8 to 6.4.5 and to apiplatform from 3.2.4 to 3.2.16
most of the stuff is working fine, but i have BC has i have a multipart/form-data that i created following this tuts
https://api-platform.com/docs/core/form-data/,
it was working correctly before the upgrade since long time, but now it give me this error
`The content-type "multipart/form-data; boundary=---------------------------14119891161883577876282016510" is not supported. Supported MIME types are "application/ld+json", "application/json".'`
the error came from
`  "/var/www/vendor/api-platform/core/src/State/Provider/ContentNegotiationProvider.php", line: 48, function: "getInputFormat" `,
but i suppose the problem is somewhere else as i just rebase everything and i only upgrade apiplatform from 3.2.4 to 3.2.16 and leaving symfony to 6.3.8 averything is working fine, any suggestion?

**How to reproduce**  

**Possible Solution**  

**Additional Context**  
The DeserializeListener  is working fine and deserializing all data, beside this if i try to override the headers
`$request->headers->set('Content-Type', 'application/ld+json');`
i receive an error from the symfony serializer
```
    "@id": "/api/errors/400",
    "@type": "hydra:Error",
    "title": "An error occurred",
    "detail": "Syntax error",
    "status": 400,
    "type": "/errors/400",
    "trace": [
        {
            "file": "/var/www/vendor/symfony/serializer/Encoder/JsonEncoder.php",
            "line": 48,
            "function": "decode",
            "class": "Symfony\\Component\\Serializer\\Encoder\\JsonDecode",
            "type": "->"
        },
        {
            "file": "/var/www/vendor/api-platform/core/src/Serializer/JsonEncoder.php",
            "line": 72,
            "function": "decode",
            "class": "Symfony\\Component\\Serializer\\Encoder\\JsonEncoder",
            "type": "->"
        },
```

menhir2015:
I find the trouble 
in my `api_platform.yaml `
i have this line
`event_listeners_backward_compatibility_layer: false`
that come out of a recipe after upgrading
`https://github.com/symfony/recipes/blob/main/api-platform/core/3.2/config/packages/api_platform.yaml`

if i remove it everything is working fine, but it is not clear what this will make
any suggestion?

[...]

Issue #5944, opened 2023-11-03
[Feat] Differentiate permitted request and response format types in configuration


paullallier:
**Description**  
Currently, in v3.2, the config "formats" list seems to be used to define permissible values on the requests "Content-Type" and "Accept" headers - the same list for both.

I'd like to be able to define that the request Content-Type should allow "application/json", but the Accept header does not (I want to always return "application/ld+json").  I think this would need another config options and some changes internally.  I might be able to handle this, but I wanted to check it's worthwhile first.

This applies to v3.2.  I'm not sure if this is a change from the behaviour in v3.1, in which case perhaps this is a bug.  But I suspect it's a potentially a new feature request for v3.3 instead.

**Example**  
Currently we have:
```
    formats:
        json: ['application/json']
        jsonld: ['application/ld+json']
```

I'd propose something like this instead (as a first-pass):
```
    input_formats:
        json: ['application/json']
    output_formats:
        jsonld: ['application/ld+json']
```
(Probably with a deprecation notice on the "formats" config option.)

PR #6275, opened 2024-04-01
feat(openapi): Allow per-endpoint OpenApi request formats to add allowed Content-Types.

### LINKED ISSUE #6141: 3.3.0-alpha1 regression - Can't override formats for a single endpoint
**API Platform version(s) affected**: 3.3.0-alpha1

**Description**  
Almost everything in my project seems to just work in the alpha version - nice!

I have one issue with endpoints which override the project default format types - this worked in 3.2.x.

**How to reproduce**  
I have an endpoint with this definition:
```
        new Post(
            controller: DocumentUploadController::class,
            openapi: new Operation(
                summary: 'Upload a new document',
                description: 'Upload a new document',
                requestBody: new RequestBody(
                    content: new ArrayObject([
                        'multipart/form-data' => [
                            'schema' => [
                                'type' => 'object',
                                'properties' => [
                                    'file' => [
                                        'type' => 'string',
                                        'format' => 'binary'
                                    ]
                                ]
                            ]
                        ]
                    ])
                )
            ),
            normalizationContext: ['groups' => ['document:id_only']],
            validationContext: ['groups' => ['document:write_new', 'media_object_create']],
            deserialize: false,
        ),
```

And project-wide, this in my config:
```
api_platform:
    formats:
        jsonld: ['application/ld+json']
```

This is throwing an error now:
```
The content-type multipart/form-data is not supported. Supported MIME types are application/ld+json.
```

**Additional Context**  
I suspect this is being missed by the test cases because they run with all the possible types in the format config - not sure if we can run one test with a different list, but I'm happy to try to create one if you can give me a few pointers?

paullallier:
| Q             | A
| ------------- | ---
| Branch?       | main (only affects 3.3.0)
| Tickets       | Closes #6141 (may also relate to #6210, which I can't reproduce - it's currently working on 3.2.x for me...)
| License       | MIT
| Doc PR        | I'm happy to update the docs if we need it for this change

I suspect this is non-idiomatic and suboptimal, so I'll leave it as a draft, but would you be willing to accept something like this?

OpenAPI allows endpoint-specific definition of Content-Type:
https://spec.openapis.org/oas/latest.html#special-considerations-for-multipart-content

The UI in API-P displays this correctly - offering a different list of input formats for that one endpoint - and API-P 3.2.x works with this (because it's not rigorously checking input formats in `ApiPlatform\Symfony\EventListener\AddFormatListener`, as far as I can see).

3.3.0 updates AddFormatListener, checks input formats and fails when using one not defined in the global formats list.  In my case, enabling `multipart/form-data` on every endpoint so that I can use it on 2 or 3 endpoints is the wrong solution - it makes the UI show an invalid format on every endpoint (and probably allows API-P to try to process data in the wrong input format too - which will potentially fail further down the line).

This patch pulls the list of formats out of the openapi definition on the request and adds that to the list of accepted formats.  It would probably be more correct to ONLY accept these formats, in place of the global list, but that's breaking some existing tests, so this is a compromise for now.

I think there's a case for separating input and output format definitions (see #5944), but that's a different discussion. 

I haven't added any tests because I think I'd need to stop the tests running with all the formats enabled by default - and I'm not sure how to change that for a single test.  I'm happy to add some if you can give me a pointer how to changes that, just for my tests?

Ultimately, if this isn't acceptable, I'll just patch my own version of 3.3 - the current behaviour is an upgrade-blocker for me.

### DIFF
--- a/src/State/Provider/ContentNegotiationProvider.php
+++ b/src/State/Provider/ContentNegotiationProvider.php
@@ -103,7 +103,7 @@ final class ContentNegotiationProvider implements ProviderInterface
         }
 
         /** @var string $contentType */
-        $formats = $operation->getInputFormats() ?? [];
+        $formats = array_merge($operation->getInputFormats() ?? [], $this->getOpenApiRequestBodyFormats($operation));
         if ($format = $this->getMimeTypeFormat($contentType, $formats)) {
             return $format;
         }
@@ -120,5 +120,20 @@ final class ContentNegotiationProvider implements ProviderInterface
         }
 
         return null;
+    }
+
+    private function getOpenApiRequestBodyFormats(HttpOperation $operation): array
+    {
+        if ($openApi = $operation->getOpenapi()) {
+            if ($openApi->getRequestBody()) {
+                foreach ($openApi->getRequestBody()->getContent() as $mimeType => $format) {
+                    // get the shortname as the mimetype after the slash
+                    $shortName = substr($mimeType, strpos($mimeType, '/') + 1);
+                    $openApiInputFormats[$shortName] = [$mimeType];
+                }
+            }
+        }
+
+        return $openApiInputFormats ?? [];
     }
 }


src/State/Provider/ContentNegotiationProvider.php
```diff
@@ -121,4 +121,19 @@ private function getInputFormat(HttpOperation $operation, Request $request): ?st
 
         return null;
     }
+
+    private function getOpenApiRequestBodyFormats(HttpOperation $operation): array
```
soyuka, line 125: we don't want a link between State and OpenApi, but why can't you add the `multipart` option to this operation only?

PR #6276, opened 2024-04-01
[JsonSchema] Only build schemas which can be serialized / deserialied

GeLoLabs:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | None
| License       | MIT
| Doc PR        | None

Hello,

This PR is a first step for a discussion about a bugfix.

Given the following ApiResource:

```php
#[ApiResource(operations: [new Delete()])]
class MyResource
{
}
```

If I generate the OpenAPI documentation using `php bin/console api:openapi:export`, the `ApiPlatform\OpenApi\Factory\OpenApiFactory` will generate a doc including the components "MyResource" whereas no operation references it (request / response -> no content). 

Futhermore, it exposes all `MyResource` properties which should stay internal as it is not part of our API.

I would expect that the `OpenApiFactory` only builds schemas that operation can deserialize / serialize and this is what this PR does.

Now, to fix my OpenAPI doc, I can just do the following:

```php
#[ApiResource(operations: [new Delete(
    deserialize: false,
    serialize: false,
)])]
class MyResource
{
}
```

WDYT ?

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -80,6 +80,14 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
             return $schema;
         }
 
+        if ('input' === $type && !($operation?->canDeserialize() ?? true)) {
+            return $schema;
+        }
+
+        if ('output' === $type && !($operation?->canSerialize() ?? true)) {
+            return $schema;
+        }
+
         $validationGroups = $operation ? $this->getValidationGroups($operation) : [];
         $version = $schema->getVersion();
         $method = $operation instanceof HttpOperation ? $operation->getMethod() : 'GET';


soyuka:
use `openapi: false`?

GeLoLabs:
@soyuka Using the flag `openapi: false` would remove the operation from the OpenAPI spec. 

What I want is to keep the operation in the OpenAPI spec but skip the associated schema as my `DELETE` operation does not have request body as well as no response body (204). 

Currenlty, it includes the schema of the class tagged as `ApiResource` in the OpenAPI spec.

soyuka:
> Futhermore, it exposes all MyResource properties which should stay internal as it is not part of our API.

This is quite wrong if a class is marked as a resource it is a public API. I like your patch though, is it possible to add a test on that though?

GeLoLabs:
@soyuka I mean it exposes unwanted properties as I use deserialization / serialization groups everywhere on this ApiResource  for all its operations except for the delete operation.  Then, it ends by adding the "root" resource (not specific to a (de)serializer group) with all is properties, something I don't want.

I will complete this PR with some tests.

src/JsonSchema/SchemaFactory.php
```diff
@@ -65,6 +65,14 @@ public function buildSchema(string $className, string $format = 'json', string $
 
         [$operation, $serializerContext, $validationGroups, $inputOrOutputClass] = $metadata;
```
GwendolenLynch, line 66: To save you digging too hard, and easing the rebase, after my last PR these are now assigned here:
https://github.com/api-platform/core/blob/678eb4f8474edc29b2fc806c85aaf203540e23ab/src/JsonSchema/SchemaFactory.php#L62-L70

GeLoLabs:
@soyuka @GwendolenLynch I'm not familiar with the project test suite. Can you point me where / how I can add tests about this feature ?

GeLoLabs:
I just rebased the PR. Can someone helps me with the missing tests ?

soyuka:
This should already be working using `input: false` or `output: false` no?

PR #6277, opened 2024-04-02
fix(jsonapi): add missing "included" schema parts

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | Related #6250
| License       | MIT

Per https://jsonapi.org/format/#fetching-includes which the serializer already handles beautifully.

### DIFF
--- a/src/JsonApi/JsonSchema/SchemaFactory.php
+++ b/src/JsonApi/JsonSchema/SchemaFactory.php
@@ -128,7 +128,7 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
                 return $schema;
             }
 
-            $definitions[$key]['properties'] = $this->buildDefinitionPropertiesSchema($key, $className, $schema, $serializerContext);
+            $definitions[$key]['properties'] = $this->buildDefinitionPropertiesSchema($key, $className, $format, $schema, $serializerContext);
 
             if ($schema->getRootDefinitionKey()) {
                 return $schema;
@@ -166,17 +166,24 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
         }
     }
 
-    private function buildDefinitionPropertiesSchema(string $key, string $className, Schema $schema, ?array $serializerContext): array
+    private function buildDefinitionPropertiesSchema(string $key, string $className, string $format, Schema $schema, ?array $serializerContext): array
     {
         $definitions = $schema->getDefinitions();
         $properties = $definitions[$key]['properties'] ?? [];
 
         $attributes = [];
         $relationships = [];
+        $relatedDefinitions = [];
         foreach ($properties as $propertyName => $property) {
             if ($relation = $this->getRelationship($className, $propertyName, $serializerContext)) {
-                [$isOne, $isMany] = $relation;
+                [$isOne, $hasOperations, $relatedClassName] = $relation;
+                if (false === $hasOperations) {
+                    continue;
+                }
 
+                $relatedDefName = $this->getShortClassName($relatedClassName).('json' === $format ? '' : ".$format");
+                $ref = Schema::VERSION_OPENAPI === $schema->getVersion() ? '#/components/schemas/'.$relatedDefName : '#/definitions/'.$relatedDefName;
+                $relatedDefinitions[$propertyName] = ['$ref' => $ref];
                 if ($isOne) {
                     $relationships[$propertyName]['properties']['data'] = self::RELATION_PROPS;
                     continue;
@@ -197,10 +204,24 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
         $replacement = self::PROPERTY_PROPS;
         $replacement['attributes']['properties'] = $attributes;
 
+        $included = [];
         if (\count($relationships) > 0) {
             $replacement['relationships'] = [
                 'type' => 'object',
                 'properties' => $relationships,
+            ];
+            $included = [
+                'included' => [
+                    'description' => 'Related resources requested via the "include" query parameter.',
+                    'type' => 'array',
+                    'items' => [
+                        'anyOf' => array_values($relatedDefinitions),
+                    ],
+                    'readOnly' => true,
+                    'externalDocs' => [
+                        'url' => 'https://jsonapi.org/format/#fetching-includes',
+                    ],
+                ],
             ];
         }
 
@@ -223,7 +244,7 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
                 'properties' => $replacement,
                 'required' => ['type', 'id'],
             ],
-        ];
+        ] + $included;
     }
 
     private function getRelationship(string $resourceClass, string $property, ?array $serializerContext): ?array
@@ -232,6 +253,7 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
         $types = $propertyMetadata->getBuiltinTypes() ?? [];
         $isRelationship = false;
         $isOne = $isMany = false;
+        $className = $hasOperations = null;
 
         foreach ($types as $type) {
             if ($type->isCollection()) {
@@ -244,8 +266,20 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
                 continue;
             }
             $isRelationship = true;
+            $resourceMetadata = $this->resourceMetadataFactory->create($className);
+            $operation = $resourceMetadata->getOperation();
+            // @see https://github.com/api-platform/core/issues/5501
+            // @see https://github.com/api-platform/core/pull/5722
+            $hasOperations ??= $operation->canRead();
         }
 
-        return $isRelationship ? [$isOne, $isMany] : null;
+        return $isRelationship ? [$isOne, $hasOperations, $className] : null;
+    }
+
+    private function getShortClassName(string $fullyQualifiedName): string
+    {
+        $parts = explode('\\', $fullyQualifiedName);
+
+        return end($parts);
     }
 }
--- a/src/JsonApi/JsonSchema/SchemaFactory.php
+++ b/src/JsonApi/JsonSchema/SchemaFactory.php
@@ -14,11 +14,14 @@ declare(strict_types=1);
 namespace ApiPlatform\JsonApi\JsonSchema;
 
 use ApiPlatform\Api\ResourceClassResolverInterface as LegacyResourceClassResolverInterface;
+use ApiPlatform\JsonSchema\DefinitionNameFactoryInterface;
+use ApiPlatform\JsonSchema\ResourceMetadataTrait;
 use ApiPlatform\JsonSchema\Schema;
 use ApiPlatform\JsonSchema\SchemaFactoryAwareInterface;
 use ApiPlatform\JsonSchema\SchemaFactoryInterface;
 use ApiPlatform\Metadata\Operation;
 use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
+use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use ApiPlatform\Metadata\ResourceClassResolverInterface;
 
 /**
@@ -28,6 +31,7 @@ use ApiPlatform\Metadata\ResourceClassResolverInterface;
  */
 final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
 {
+    use ResourceMetadataTrait;
     private const LINKS_PROPS = [
         'type' => 'object',
         'properties' => [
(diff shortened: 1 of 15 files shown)


src/JsonApi/JsonSchema/SchemaFactory.php
```diff
@@ -244,8 +267,17 @@ private function getRelationship(string $resourceClass, string $property, ?array
                 continue;
             }
             $isRelationship = true;
+            $resourceMetadata = $this->resourceMetadataCollectionFactory->create($className);
+            $hasOperations ??= !$resourceMetadata->getOperation() instanceof NotExposed;
```
soyuka, line 271: I see, not a huge fan of using the `NotExposed` class, especially that the class is final, isn't there something else we could use like `output: false`? I'd like to avoid adding an interface if possible.

src/JsonApi/JsonSchema/SchemaFactory.php
```diff
@@ -166,17 +167,24 @@ public function setSchemaFactory(SchemaFactoryInterface $schemaFactory): void
...
+        $relatedDefinitions = [];
         foreach ($properties as $propertyName => $property) {
             if ($relation = $this->getRelationship($className, $propertyName, $serializerContext)) {
-                [$isOne, $isMany] = $relation;
+                [$isOne, $hasOperations, $relatedClassName] = $relation;
+                if (false === $hasOperations) {
+                    continue;
+                }
 
+                $relatedDefName = $this->getShortClassName($relatedClassName).('json' === $format ? '' : ".$format");
```
soyuka, line 185: Can you return the `Operation` above and use `$operation->getShortName()`?

src/JsonApi/JsonSchema/SchemaFactory.php
GwendolenLynch: `$operation->canRead()` does the trick! :+1:

GwendolenLynch:
@soyuka I have had to proof-of-concept a slightly different approach in a junk commit for review & discussion on where to go from here :sun_with_face: 

**Problem**
Both the base and JSON:API schema factories needs to be able to reliably know the schema component "definition name" for a resource. That requires both of the `getMetadata()` & `buildDefinitionName` methods from `ApiPlatform\JsonSchema\SchemaFactory`.

**Approach**
1. I have moved `buildDefinitionName` and its dependant methods to `ApiPlatform\JsonSchema\DefinitionNameFactory` and making that a container service `api_platform.json_schema.definition_name_factory` that injects into the schema factories as needed.
2. `getMetadata()` and friends I put into `ApiPlatform\JsonSchema\ResourceMetadataTrait`

**Questions**
1. Is this (generally) an acceptable direction in this case?
2. What namespaces would you like those two new files in?

Advice and guidance warmly welcomed!

soyuka (commented):
Nice refactoring actually I think that it's a good idea to have a Trait to share the logic, not sure about adding a new interface though as `buildDefinitionName` is private anyways. Can that work only using the Trait? 

Also don't put another author then yourself (no author phpdoc is also fine) and I think that the trait should be `@internal` so that we can make updates to it in the future without breaking userland.

src/JsonSchema/DefinitionNameFactory.php
```diff
@@ -0,0 +1,78 @@
...
+declare(strict_types=1);
+
+namespace ApiPlatform\JsonSchema;
+
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Util\ResourceClassInfoTrait;
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
+
+/**
+ * @author Kévin Dunglas <dunglas@gmail.com>
```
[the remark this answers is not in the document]
GwendolenLynch: Totally fine with those changes, I just didn't want to take credit for Kévin's work.

src/JsonSchema/SchemaFactory.php
```diff
@@ -35,18 +32,21 @@
...
     public const FORCE_SUBSCHEMA = '_api_subschema_force_readable_link';
     public const OPENAPI_DEFINITION_NAME = 'openapi_definition_name';
 
-    public function __construct(?TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ?ResourceClassResolverInterface $resourceClassResolver = null, private readonly ?array $distinctFormats = null)
+    public function __construct(?TypeFactoryInterface $typeFactory, ResourceMetadataCollectionFactoryInterface $resourceMetadataFactory, private readonly PropertyNameCollectionFactoryInterface $propertyNameCollectionFactory, private readonly PropertyMetadataFactoryInterface $propertyMetadataFactory, private readonly ?NameConverterInterface $nameConverter = null, ?ResourceClassResolverInterface $resourceClassResolver = null, private ?DefinitionNameFactoryInterface $definitionNameFactory = null)
```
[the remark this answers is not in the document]
GwendolenLynch: Derp! I forgot that went into `3.2`

src/JsonApi/JsonSchema/SchemaFactory.php
```diff
@@ -102,22 +106,24 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
...
      * {@inheritdoc}
      */
     public function buildSchema(string $className, string $format = 'jsonapi', string $type = Schema::TYPE_OUTPUT, ?Operation $operation = null, ?Schema $schema = null, ?array $serializerContext = null, bool $forceCollection = false): Schema
     {
-        $schema = $this->schemaFactory->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
         if ('jsonapi' !== $format) {
-            return $schema;
+            return $this->schemaFactory->buildSchema($className, $format, $type, $operation, $schema, $serializerContext, $forceCollection);
         }
+        $schema = $this->schemaFactory->buildSchema($className, $format, $type, $operation, $schema, [], $forceCollection);
```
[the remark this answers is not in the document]
GwendolenLynch: I was hoping you'd bring this up :+1: 

I haven't mapped it out fully yet, but the JSON:API  serializer seems to mostly ignore normalization groups (c.f. `include` query parameter.

So if it is not a `jsonapi` format, just pass it on as-is, else generate the schema without groups. Hadn't gotten past that so far. I got distracted with enums.

GwendolenLynch:
> […]  not sure about adding a new interface though as buildDefinitionName is private anyways. Can that work only using the Trait?

It can. I chose to externalise it as a guess as to what you'd choose … plus the thought that definition name generation becomes implementation override-able.

Your call :+1:

soyuka:
Indeed but as long as you use the shortName the extension point exists no? If you are in a weird case you need to decorate the SchemaFactory.

GwendolenLynch:
As I understand it, it depends on how `formats:` config is set-up.

Assume you have the resource `Foo`.

If you set, for example, one format in the `formats` key:
```yaml
api_platform:
    formats:
        jsonapi:  ['application/vnd.api+json']
```
… then `patch_formats` will default to `application/merge-patch+json` so you will end up with two definitions:
- `Foo` (write operations)
- `Foo.jsonapi` (read operations)

However, if you configure a non `application/merge-patch+json`  in `patch_formats`:
```yaml
api_platform:
    formats:
        jsonapi:  ['application/vnd.api+json']
    patch_formats:
        jsonapi:  ['application/vnd.api+json']
```
… you will end up with one definition:
- `Foo.jsonapi`

Now if you add (de)normalization groups all bets are off. Hence the refactoring in this PR.

I did not know all that a couple of weeks ago, so yeah :innocent:

GwendolenLynch:
One other thing that came to mind with having it separate is unit testing. It makes that part of the process distinctly testable. 

**Disclaimer:** I can be overly obsessive about unit tests and the _"…one thing well"_ principle. :blush:

src/JsonSchema/ResourceMetadataTrait.php
```diff
@@ -0,0 +1,101 @@
...
+{
+    use ResourceClassInfoTrait;
+
+    private function findOutputClass(string $className, string $type, Operation $operation, ?array $serializerContext): ?string
+    {
+        $inputOrOutput = ['class' => $className];
+        $inputOrOutput = Schema::TYPE_OUTPUT === $type ? ($operation->getOutput() ?? $inputOrOutput) : ($operation->getInput() ?? $inputOrOutput);
+        $forceSubschema = $serializerContext[SchemaFactory::FORCE_SUBSCHEMA] ?? false;
+
+        return $forceSubschema ? ($inputOrOutput['class'] ?? $inputOrOutput->class ?? $operation->getClass()) : ($inputOrOutput['class'] ?? $inputOrOutput->class ?? null);
```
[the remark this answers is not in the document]
GwendolenLynch: Yep, unified now.

src/JsonSchema/DefinitionNameFactory.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace ApiPlatform\JsonSchema;
+
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Util\ResourceClassInfoTrait;
+use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;
+
+/**
+ * @internal
+ */
```
soyuka, line 22: let's remove this actually as it's an extension point (I did not understood that at first mb).

src/Symfony/Bundle/Resources/config/json_schema.xml
```diff
@@ -23,6 +23,7 @@
             <argument type="service" id="api_platform.name_converter" on-invalid="ignore" />
             <argument type="service" id="api_platform.resource_class_resolver" />
             <argument on-invalid="ignore">%api_platform.jsonschema_formats%</argument>
+            <argument type="service" id="api_platform.json_schema.definition_name_factory" />
```
soyuka, line 26:
```suggestion
            <argument on-invalid="ignore" type="service" id="api_platform.json_schema.definition_name_factory" />
```

src/JsonSchema/DefinitionNameFactoryInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\JsonSchema;
+
+use ApiPlatform\Metadata\Operation;
+
+interface DefinitionNameFactoryInterface
```
soyuka, line 18: can you add a comment that explain what this does?

src/JsonSchema/DefinitionNameFactoryInterface.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    public function create(string $className, string $format = 'json', ?string $inputOrOutputClass = null, ?Operation $operation = null, array $serializerContext = []): string;
```
soyuka, line 20: would be nice to add a php-doc (`class-string` type for the `$className`, `@return string the definition name`)

src/JsonSchema/ResourceMetadataTrait.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private function findOperation(string $className, string $type, ?Operation $operation, ?array $serializerContext): Operation
+    {
+        if (null === $operation) {
+            $resourceMetadataCollection = $this->resourceMetadataFactory->create($className);
```
soyuka, line 42:
```suggestion
            $resourceMetadataCollection = $this->resourceMetadataFactory->create($className);
```

`resourceMetadataFactory` can be null sometimes, I wonder why phpstan didn't catch that

soyuka (changes requested):
nice, few leftovers

src/JsonSchema/DefinitionNameFactory.php
```diff
@@ -0,0 +1,70 @@
...
+final class DefinitionNameFactory implements DefinitionNameFactoryInterface
+{
+    use ResourceClassInfoTrait;
+
+    public function __construct(private ?array $distinctFormats)
+    {
+    }
+
+    /**
+     * @param class-string $className
```
GwendolenLynch, line 32: Leave this as is or `@inheritdoc`?

src/JsonSchema/ResourceMetadataTrait.php
GwendolenLynch: TBH, I seem to remember having a similar thought and originally just left it with the assumption _"If it isn't broken, don't fix it"_

src/JsonSchema/DefinitionNameFactory.php
soyuka: you can remove inheritdoc is automatic now in most of the tools

src/JsonSchema/DefinitionNameFactory.php
```diff
@@ -0,0 +1,67 @@
...
+     */
```
soyuka, line 30:
```suggestion
```

src/JsonSchema/ResourceMetadataTrait.php
```diff
@@ -0,0 +1,105 @@
...
+
+    private function getSerializerContext(Operation $operation, string $type = Schema::TYPE_OUTPUT): array
+    {
+        return Schema::TYPE_OUTPUT === $type ? ($operation->getNormalizationContext() ?? []) : ($operation->getDenormalizationContext() ?? []);
+    }
+
+    private function getShortClassName(string $fullyQualifiedName): string
+    {
+        $parts = explode('\\', $fullyQualifiedName);
+
```
soyuka, line 102:
```suggestion
```

src/Symfony/Bundle/Resources/config/jsonapi.xml
```diff
@@ -9,6 +9,8 @@
             <argument type="service" id="api_platform.jsonapi.json_schema.schema_factory.inner" />
             <argument type="service" id="api_platform.metadata.property.metadata_factory" />
             <argument type="service" id="api_platform.resource_class_resolver" />
+            <argument type="service" id="api_platform.metadata.resource.metadata_collection_factory" />
+            <argument type="service" id="api_platform.json_schema.definition_name_factory" />
```
soyuka, line 13:
```suggestion
            <argument type="service" id="api_platform.metadata.resource.metadata_collection_factory" on-invalid="ignore" />
            <argument type="service" id="api_platform.json_schema.definition_name_factory" on-invalid="ignore" />
```

soyuka (approved):

GwendolenLynch:
Thanks @soyuka!

PR #6284, opened 2024-04-03
fix(JsonLd): Ignore context resource_class when it is populated with an interface or an abstract class.

### LINKED ISSUE #6251: Interfaced object relations do not resolve JsonLd inheriting resource types
**API Platform version(s) affected**: 3.2.18

**Description**  

If a DTO expose a property with a API Resource but this property is typed using an Interface to allow polymorphism: Api Platform won't be able to resolve the final resource' `types` and will serialize the short className in `@type` for JsonLd format.

```php
interface CommonResource {

}

#[ApiResource( types: [ 'https://schema.org/Thing' ] )]
class MyResource implements CommonResource {

}

class MyDto {
    #[ApiProperty]
    public ?CommonResource $resource = null;
}


$dto = (new MyDto())->resource = new MyResource();
```

Actual output:
```json
{
    "@type": "MyDto",
    "resource": {
        "@type": "MyResource"
    }
}
```

Expected output:
```json
{
    "@type": "MyDto",
    "resource": {
        "@type": "https://schema.org/Thing"
    }
}
```

**Possible cause : \ApiPlatform\JsonLd\Serializer\ItemNormalizer::normalize (line 84)**

I think that's because in `\ApiPlatform\JsonLd\Serializer\ItemNormalizer::normalize`, `$context['resource_class']` is populated with the interface `CommonResource` which is not an `ApiResource`.

```php
$resourceClass = $this->getObjectClass($object); // Here : MyResource
$previousResourceClass = $context['resource_class'] ?? null; // Here: CommonResource
```

So this condition cannot be true (line 84), and resource is normalized as an **anonymous** resource:

```php
if($isResourceClass = $this->resourceClassResolver->isResourceClass($resourceClass) && (null === $previousResourceClass || $this->resourceClassResolver->isResourceClass($previousResourceClass)))
{
    //...
}
```

**Possible solution:**

We should test if `$previousResourceClass` is an interface or abstract class and `$resourceClass` is an instanceof 
`$previousResourceClass` then build context against `$resourceClass`, not `$previousResourceClass`.

ambroisemaupate:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #6251
| License       | MIT

I put `null !== $previousResourceClass && $isResourceClass` condition before testing if it's an abstract or interface in order to avoid using `\ReflectionClass` for every Item normalization. Just for this specific case for performances sake.

### DIFF
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -82,7 +82,20 @@ final class ItemNormalizer extends AbstractItemNormalizer
         // TODO: we should not remove the resource_class in the normalizeRawCollection as we would find out anyway that it's not the same as the requested one
         $previousResourceClass = $context['resource_class'] ?? null;
         $metadata = [];
-        if ($isResourceClass = $this->resourceClassResolver->isResourceClass($resourceClass) && (null === $previousResourceClass || $this->resourceClassResolver->isResourceClass($previousResourceClass))) {
+        $isResourceClass = $this->resourceClassResolver->isResourceClass($resourceClass);
+
+        /*
+         * Ignore context resource class if it's an abstract class or an interface
+         * and the current object is a resource class.
+         */
+        if (null !== $previousResourceClass && $isResourceClass) {
+            $previousResourceClassReflection = new \ReflectionClass($previousResourceClass);
+            if ($previousResourceClassReflection->isAbstract() || $previousResourceClassReflection->isInterface()) {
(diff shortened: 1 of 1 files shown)


soyuka:
Look at the `ItemNormalizerTest`? A functional test would be awesome imo, for this either add a functional test like https://github.com/api-platform/core/blob/main/tests/Functional/Parameters/ParameterTests.php or a behat test.

soyuka:
can you send me the resource and a use case (even curl) so that I add a test?

PR #6292, opened 2024-04-05
fix(symfony): conflict for symfony/framework-bundle versions that break enum serializing

soyuka:


soyuka:
thanks @GwendolenLynch

PR #6296, opened 2024-04-05
feat(validator): parameter validation

soyuka:
Adds constraint to parameters to allow validation:

- [x] add tests
- [x] add a Metadata factory that automatically adds constraints based on OpenAPI and metadata informations

### DIFF
--- a/src/Metadata/Parameter.php
+++ b/src/Metadata/Parameter.php
@@ -15,6 +15,7 @@ namespace ApiPlatform\Metadata;
 
 use ApiPlatform\OpenApi;
 use ApiPlatform\State\ProviderInterface;
+use Symfony\Component\Validator\Constraint;
 
 /**
  * @experimental
@@ -26,6 +27,7 @@ abstract class Parameter
      * @param array<string, mixed>                   $extraProperties
      * @param ProviderInterface|callable|string|null $provider
      * @param FilterInterface|string|null            $filter
+     * @param Constraint|Constraint[]|null           $constraints
      */
     public function __construct(
         protected ?string $key = null,
@@ -37,6 +39,7 @@ abstract class Parameter
         protected ?string $description = null,
         protected ?bool $required = null,
         protected ?int $priority = null,
+        protected Constraint|array|null $constraints = null,
         protected ?array $extraProperties = [],
     ) {
     }
@@ -87,6 +90,14 @@ abstract class Parameter
     public function getPriority(): ?int
     {
         return $this->priority;
+    }
+
+    /**
+     * @return Constraint|Constraint[]|null
+     */
+    public function getConstraints(): Constraint|array|null
+    {
+        return $this->constraints;
     }
 
     /**
@@ -174,6 +185,14 @@ abstract class Parameter
     {
         $self = clone $this;
         $self->required = $required;
+
+        return $self;
+    }
+
+    public function withConstraints(array|Constraint $constraints): static
+    {
+        $self = clone $this;
+        $self->constraints = $constraints;
 
         return $self;
     }
--- a/src/Metadata/Resource/Factory/ParameterResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/ParameterResourceMetadataCollectionFactory.php
@@ -18,9 +18,22 @@ use ApiPlatform\Metadata\HeaderParameterInterface;
 use ApiPlatform\Metadata\Parameter;
 use ApiPlatform\Metadata\Parameters;
 use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
-use ApiPlatform\OpenApi;
+use ApiPlatform\OpenApi\Model\Parameter as OpenApiParameter;
 use ApiPlatform\Serializer\Filter\FilterInterface as SerializerFilterInterface;
 use Psr\Container\ContainerInterface;
+use Symfony\Component\Validator\Constraints\Choice;
+use Symfony\Component\Validator\Constraints\Count;
+use Symfony\Component\Validator\Constraints\DivisibleBy;
+use Symfony\Component\Validator\Constraints\GreaterThan;
+use Symfony\Component\Validator\Constraints\GreaterThanOrEqual;
+use Symfony\Component\Validator\Constraints\Length;
+use Symfony\Component\Validator\Constraints\LessThan;
+use Symfony\Component\Validator\Constraints\LessThanOrEqual;
+use Symfony\Component\Validator\Constraints\NotBlank;
+use Symfony\Component\Validator\Constraints\NotNull;
+use Symfony\Component\Validator\Constraints\Regex;
+use Symfony\Component\Validator\Constraints\Unique;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * Prepares Parameters documentation by reading its filter details and declaring an OpenApi parameter.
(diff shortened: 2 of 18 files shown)


src/Metadata/Parameter.php
```diff
@@ -89,6 +92,14 @@ public function getPriority(): ?int
         return $this->priority;
     }
 
+    /**
+     * @return Constraint|Constraint[]|null
+     */
+    public function getConstraint(): array|Constraint|null
```
vincentchalamon, line 98: I suggest to always manipulate an array of constraints

src/Symfony/Validator/State/ParameterValidatorProvider.php
```diff
@@ -0,0 +1,82 @@
...
+use Symfony\Component\Validator\ConstraintViolation;
+use Symfony\Component\Validator\ConstraintViolationList;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * Validates parameters using the symfony validator.
+ *
+ * @experimental
+ */
+final class ParameterValidatorProvider implements ProviderInterface
```
vincentchalamon, line 28:
```suggestion
final readonly class ParameterValidatorProvider implements ProviderInterface
```

src/Symfony/Validator/State/ParameterValidatorProvider.php
```diff
@@ -0,0 +1,82 @@
...
+{
+    public function __construct(
+        private readonly ProviderInterface $decorated,
+        private readonly ValidatorInterface $validator
```
vincentchalamon, line 32:
```suggestion
        private ProviderInterface $decorated,
        private ValidatorInterface $validator
```

src/Symfony/Validator/State/ParameterValidatorProvider.php
```diff
@@ -0,0 +1,82 @@
...
+    ) {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
+    {
+        $body = $this->decorated->provide($operation, $uriVariables, $context);
+        if (!$context['request'] ?? null) {
```
vincentchalamon, line 39:
```suggestion
        if (!$context['request'] instanceof Request) {
```

src/Symfony/Validator/State/ParameterValidatorProvider.php
```diff
@@ -0,0 +1,82 @@
...
+            return $body;
+        }
+
+        $operation = $context['request']->attributes->get('_api_operation');
+        foreach ($operation->getParameters() as $parameter) {
+            if ($parameter->getRequired() && !\array_key_exists('_api_values', $parameter->getExtraProperties())) {
+                throw new ValidationException(new ConstraintViolationList([new ConstraintViolation(sprintf('Parameter "%s" is required.', $parameter->getKey()), null, [], null, null, null)]));
+            }
```
vincentchalamon, line 47: Can't this constraint be automatically set in `ParameterResourceMetadataCollectionFactory` instead?

vincentchalamon (changes requested):

src/Metadata/Resource/Factory/ParameterResourceMetadataCollectionFactory.php
```diff
@@ -121,6 +142,69 @@ private function setDefaults(string $key, Parameter $parameter, string $resource
...
+
+        if ($openApi['uniqueItems'] ?? false) {
+            $assertions[] = new Unique();
+        }
+
+        if (isset($openApi['enum'])) {
+            $assertions[] = new Choice(choices: $openApi['enum']);
+        }
+
+        if (isset($openApi['enum'])) {
```
michaljusiega, line 200: Duplicated code.

src/Symfony/Validator/State/ParameterValidatorProvider.php
soyuka: Indeed, I'll try a default value to `null` with a NotBlank constraint?

src/Metadata/Parameter.php
soyuka: For DX it's quite nice no? 

```
#[Parameter(constraint: new NotBlank())]
```

src/Symfony/Validator/State/ParameterValidatorProvider.php
vincentchalamon: Maybe just add NotBlank(allowNull: false) if `required` option is set?

src/Metadata/Parameter.php
soyuka: it should rather be `new All(new NotBlank(), new Email())` so what about removing the `array`?

soyuka: actually `All` can't be used like this, we need to keep the array.

vincentchalamon: So `constraints`?
```
#[Parameter(constraints: [new NotBlank(), new Email()])
```

src/Symfony/Validator/State/ParameterValidatorProvider.php
```diff
@@ -0,0 +1,75 @@
...
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProviderInterface;
+use ApiPlatform\Validator\Exception\ValidationException;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Validator\ConstraintViolation;
+use Symfony\Component\Validator\ConstraintViolationList;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * Validates parameters using the symfony validator.
```
dunglas, line 25:
```suggestion
 * Validates parameters using the Symfony validator.
```

src/Metadata/Parameter.php
```diff
@@ -37,6 +39,7 @@ public function __construct(
         protected ?string $description = null,
         protected ?bool $required = null,
         protected ?int $priority = null,
+        protected Constraint|array|null $constraints = null,
```
dunglas, line 42: Shouldn't you cast the passed value to `array` in the constructor?

src/Metadata/Parameter.php
```diff
@@ -178,6 +189,14 @@ public function withRequired(bool $required): static
         return $self;
     }
 
+    public function withConstraints(array|Constraint $constraints): static
+    {
+        $self = clone $this;
+        $self->constraints = $constraints;
```
dunglas, line 195:
```suggestion
        $self->constraints = (array) $constraints;
```

src/Metadata/Resource/Factory/ParameterResourceMetadataCollectionFactory.php
```diff
@@ -121,6 +142,74 @@ private function setDefaults(string $key, Parameter $parameter, string $resource
...
 
+        $schema = $parameter->getSchema() ?? $parameter->getOpenApi()?->getSchema();
+        if (!$parameter->getConstraints()) {
+            $parameter = $this->addSchemaValidation($parameter, $schema, $parameter->getRequired() ?? $description['required'] ?? false, $parameter->getOpenApi());
+        }
+
         return $parameter;
     }
+
+    private function addSchemaValidation(Parameter $parameter, ?array $schema = null, bool $required = false, ?OpenApiParameter $openApi = null): Parameter
```
dunglas, line 153: Couldn't we make the dependency to `symfony/validator` optional, or even better, use inversion of control to move this method in the Symfony bridge (and add an extension point to allow the user to use another validation library)?

src/Metadata/Parameter.php
```diff
@@ -89,6 +92,14 @@ public function getPriority(): ?int
         return $this->priority;
     }
 
+    /**
+     * @return Constraint|Constraint[]|null
```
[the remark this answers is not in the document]
soyuka: mhh why add the FQDN here?

src/Metadata/Parameter.php
```diff
@@ -15,17 +15,19 @@
...
     /**
-     * @param array{type?: string}|null              $schema
-     * @param array<string, mixed>                   $extraProperties
-     * @param ProviderInterface|callable|string|null $provider
-     * @param FilterInterface|string|null            $filter
+     * @param array{type?: string}|null                                 $schema
+     * @param array<string, mixed>                                      $extraProperties
+     * @param ProviderInterface|callable|string|null                    $provider
+     * @param FilterInterface|string|null                               $filter
+     * @param Constraint|\Symfony\Component\Validator\Constraint[]|null $constraints
```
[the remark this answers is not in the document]
dunglas: I was mistaken by this. I thought we were also providing our own `Constraint` class but it's not the case.

So this should be: 

```suggestion
     * @param Constraint|Constraint[]|null $constraints
```

src/Metadata/Resource/Factory/ParameterResourceMetadataCollectionFactory.php
dunglas: Ok, this is premature as we require the Symfony `Constraint` class anyway. Let's keep this as-is for now.

dunglas (approved):
LGTM when docs fix will be merged.

PR #6300, opened 2024-04-08
fix: reorder early return QueryParameterValidateListener

### LINKED ISSUE #6243: ParameterValidator and flag use_symfony_listeners
**API Platform version(s) affected**: 3.3 (3.3.0-alpha.2 and main branch)

**Description**  
An endpoint that throw a exception `ApiPlatform\ParameterValidator\Exception\ValidationException` seems to be not handle by ApiPlatform when `use_symfony_listeners` flag is true. When it's false, exception is correctly handled.

**How to reproduce**  

**Possible Solution**  

**Additional Context**

louismariegaborit:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #6243
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/EventListener/QueryParameterValidateListener.php
+++ b/src/Symfony/EventListener/QueryParameterValidateListener.php
@@ -15,7 +15,7 @@ namespace ApiPlatform\Symfony\EventListener;
 
 use ApiPlatform\Doctrine\Odm\State\Options as ODMOptions;
 use ApiPlatform\Doctrine\Orm\State\Options;
-use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\HttpOperation;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
 use ApiPlatform\ParameterValidator\ParameterValidator;
 use ApiPlatform\State\ProviderInterface;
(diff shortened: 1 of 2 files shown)


src/Symfony/EventListener/QueryParameterValidateListener.php
```diff
@@ -85,6 +71,20 @@ public function onKernelRequest(RequestEvent $event): void
...
+            }
+
+            $this->provider->provide($operation, $request->attributes->get('_api_uri_variables') ?? [], [
+                'request' => $request,
+                'uri_variables' => $request->attributes->get('_api_uri_variables') ?? [],
+                'resource_class' => $operation->getClass(),
+            ]);
+
+            return;
+        }
```
soyuka, line 86: can you add a non reg test in QueryParameterValidateListenerTest

louismariegaborit: Tests added. I did some changes in addition on moved code.  
Could you check if the if block stay good ?

louismariegaborit: I have an error on CI that I don't have in my local env. I work on it this day.

louismariegaborit: Pipeline is ok.
The problem with phpunit tests was the namespace. I didn't see that test class was already exist in Symfony sources.  
For my future contributions, why there are tests in ./tests/Symfony folder and others in ./src/Symfony/Tests ?

soyuka: We moved some tests into their components directory on the main branch, part of our subtree-split work. See: https://github.com/api-platform/core/blob/main/CONTRIBUTING.md#components-tests

soyuka:
nice thanks!

PR #6305, opened 2024-04-11
[RFC] feat(paginator): add `CursorPaginatorInterface`

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | 

This is currently a WIP and an RFC. If maintainers are open to the idea, I will complete the PR. Currently, it is missing HAL supports and the tests.

To improve cursor pagination, I propose the creation of `CursorPaginatorInterface`:

```php
interface CursorPaginatorInterface extends \Countable, \Traversable
{
    public function getNextPageCursor(): ?string;
    public function getPreviousPageCursor(): ?string;
    public function getFirstPageCursor(): ?string;
    public function getLastPageCursor(): ?string;
}
```

A cursor is an opaque identifier embedded in the URL, that indicates a specific page. The format of this identifier is left to the implementation. The framework and the user should not care about this format, they only know that the identifier can be used to locate a page. 

A future `DoctrineCursorPaginatorFactory` might use a Base64-encoded serialized array containing the edge values and sort direction.

If a `ProviderInterface::provide()` returns a `CursorPaginatorInterface`, API Platform will render it like this (for JSON-LD/Hydra):

```json
{
  "@context": "/api/contexts/Book",
  "@id": "/api/books",
  "@type": "hydra:Collection",
  "hydra:member": [
     ...
  ],
  "hydra:view": {
    "@type": "hydra:PartialCollectionView",
    "@id": "/api/books",
    "hydra:first": "/api/books",
    "hydra:last": "/api/books?page=q1YqU7KKjtVRKlCy0jXUUcpRssorzcnRUcpXsjLQUSpRslIqVaoFAA",
    "hydra:next": "/api/books?page=q1YqU7KqVspMUbKyNKjVUSpQsjLRUcpRssorzcnRUcpXsjLQUSpRsgIK1QIA",
    "hydra:previous": "/api/books?page=q1YqU7KqVspMUbIyM6zVUSpQsjLSUcpRssorzcnRUcpXsjLQUSpRslIqVaoFAA"
  }
}
```

It reuses the same `page_parameter_name` as `PartialPaginatorInterface`. My PR changes the type of these `page` parameters from integer to string for this to work. It should still work for `PartialPaginatorInterface`.

Cursor pagination will be decoupled from filtering, and no longer requires filters to work.

In the future, it will be possible to create an adapter from `PaginatorInterface` and `PartialPaginatorInterface` to `CursorPaginatorInterface`. In which case, the page numbers will be treated the same way as cursors. Then, we'll be able to handle all the paginators using a single logic.

In the future, we should also be able to remove many of the GraphQL-related special cases and simplify the code base.

Included in this PR:

- [x] `CursorPaginatorInterface`
- [x] `CursorBasedPartialCollectionViewNormalizer`. I decided to make it separate from `PartialCollectionViewNormalizer` for simplification.
- [x] Change to `OpenApiFactory` so that the `page` query parameters' type is string, not integer.
- [ ] `CursorPagination`
- [ ] Support for HAL

### DIFF
--- a/features/openapi/docs.feature
+++ b/features/openapi/docs.feature
@@ -129,7 +129,7 @@ Feature: Documentation support
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[1].name" should be equal to "page"
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[1].in" should be equal to "query"
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[1].required" should be false
-    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[1].schema.type" should be equal to "integer"
+    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[1].schema.type" should be equal to "string"
 
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[2].name" should be equal to "itemsPerPage"
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters[2].in" should be equal to "query"
--- a/src/Hydra/Serializer/CursorBasedPartialCollectionViewNormalizer.php
+++ b/src/Hydra/Serializer/CursorBasedPartialCollectionViewNormalizer.php
@@ -1,0 +1,131 @@
+<?php
+
+/*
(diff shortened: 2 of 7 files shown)


Eliepse:
Hello,
I think this PR is really interesting.
What is its progress and what contribution could help it progress ?
Thanks :)

PR #6288, opened 2024-04-04
test: Backed enum resource tests

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | #6264 #6279 #6283 
| License       | MIT

@soyuka this is just an updated version of the tests I linked yesterday when discussing #6264 & #6279 … and if I m still wrong please just close this PR. :+1: 

These tests illustrate what *was* working for `json`, `jsonhal`, and `jsonld` formats, but breaks now with the change in https://github.com/symfony/symfony/pull/54315

However, if a fix/workaround is found these tests will still add additional coverage for enum resources.

**Notes:**
- Tests pass when with:
  -  core `3.2` as-of `v3.2.19` (i.e. pre #6283)
  - `symfony/framework-bundle` `<=7.0.5` (i.e. pre https://github.com/symfony/symfony/pull/54315)
- ~Missing `jsonapi` tests due to target branch~ (Derp! serializer !== schema factory)

PR #6309, opened 2024-04-13
feat: BackedEnum resources

### LINKED ISSUE #6298: [3.4] Improving handling of BackedEnums with ApiResource class attributes
To better understand where to go with #6288 I spent some time researching the history of (PHP 8.1+) enums in this library … i.e. what are the correct behavioural expectations, and any missing bits. So this issue is an attempt to bring together a bunch of context to discuss & find a clear path forward.

Right now the key questions are around how to handle `\BackedEnum`s containing the `ApiResource` class attribute:
- Have metadata generation remove operations; *or*
- Add/fix support for serialization & schema generation (see example below)
  - Properties of type `\BackedEnum` should always normalize to the enum `->value` not an iri
  - ~Schema properties of type `\BackedEnum` should always be `{"enum": [value1, value2, value3, ...]}` not the iri type~
  - Enum resource routes return only `name` & `value` properties by default
  - `cases` should probably never be included as an enum item property
  - Automatic registration of item  & collection `GET` providers

One use-case for the latter is where there is a need to provide enum metadata lookup, e.g. in the example below a human friendly description of a Status' integer.

I'm happy to try and cover `json` and `jsonapi`, and @soyuka is already working on [`jsonld`](https://github.com/soyuka/core/tree/feat/enum) … `jsonhal` is probably easy for me to fit in too if needed.

### Background

@soyuka's blog post [API Platform 3.1 is out!](https://soyuka.me/api-platform-3.1-whats-new/)
> Finally after quite some time (this issue was opened in 2018!), we have support for Enums in OpenAPI, REST and GraphQL. Both Doctrine ORM and ODM have support for enums through enumType which is supported by API Platform 3.0. Alan added their support over OpenAPI and GraphQL.

[How to expose Enums with API Platform](https://les-tilleuls.coop/en/blog/how-to-expose-enums-with-api-platform) — a blog post that got attention showing how to expose enums as resources. This is what I have based my assumptions/approach around here. One has to start somewhere.

##### Implementation

- #1727
- #2254
- #4349

##### Misc

- #6185
- #6116
- #6059
- #5928

##### Related to Symfony Changes

- #6264
- #6279

#### PRs

- #5011
- #5120
- #5199
- #6283
- #6288

### Current State Examples

> [!NOTE]
> For simplicity I'm just sticking with `Accept: application/json` here.

#### Enum as Property Value

This works and both the schema and JSON response are as expected.

<details>

<summary>Model</summary>

```php

#[ApiResource]
class Article
{
    public ?int $id = null;
    public ?string $title = null;
    public ?Audit $audit = null;
    public ?Status $status = null;
}

enum Audit: string
{
    case Pending = 'pending';
    case Passed = 'passed';
    case Failed = 'failed';
}

enum Status: int
{
    case DRAFT = 0;
    case PUBLISHED = 1;
    case ARCHIVED = 2;
}
```

</details>

<details>

<summary>Schema</summary>

```json
{
    "Article": {
        "type": "object",
        "description": "",
        "deprecated": false,
        "properties": {
            "id": {
                "readOnly": true,
                "type": "integer"
            },
            "title": {
                "type": "string"
            },
            "audit": {
                "type": "string",
                "enum": [
                    "pending",
                    "passed",
                    "failed"
                ]
            },
            "status": {
                "type": "integer",
                "enum": [0, 1, 2]
            }
        }
    }
}
```

</details>

```shell
curl -X 'GET' 'https://example.com/articles/1' -H 'accept: application/json'
```

```json
{
  "id": 1,
  "title": "Once Upon A Title",
  "audit": "passed",
  "status": 1
}
```

#### Enum Resource as Property Value

<details>

<summary>Model</summary>

```php
#[ApiResource]
#[GetCollection(provider: Audit::class.'::getCases')]
#[Get(provider: Audit::class.'::getCase')]
enum Audit: string
{
    use EnumApiResourceTrait;

    case Pending = 'pending';
    case Passed = 'passed';
    case Failed = 'failed';
}

#[ApiResource]
#[GetCollection(provider: Status::class.'::getCases')]
#[Get(provider: Status::class.'::getCase')]
enum Status: int
{
    use EnumApiResourceTrait;

    case DRAFT = 0;
    case PUBLISHED = 1;
    case ARCHIVED = 2;

    #[ApiProperty]
    public function getDescription(): string
    {
        return match ($this) {
            self::DRAFT => 'Article is not ready for public consumption',
            self::PUBLISHED => 'Article is publicly available',
            self::ARCHIVED => 'Article content is outdated or superseded',
        };
    }
}

trait EnumApiResourceTrait
{
    public function getId(): string|int
    {
        return $this->value;
    }

    public function getValue(): int|string
    {
        return $this->value;
    }

    public static function getCases(): array
    {
        return self::cases();
    }

    public static function getCase(Operation $operation, array $uriVariables): ?BackedEnum
    {
        $id = is_numeric($uriVariables['id']) ? (int) $uriVariables['id'] : $uriVariables['id'];

        return array_reduce(self::cases(), static fn($c, \BackedEnum $case) => $case->name === $id || $case->value === $id ? $case : $c, null);
    }
}
```

</details>

<details>

<summary>Schema</summary>

```json
{
    "Article": {
        "type": "object",
        "description": "",
        "deprecated": false,
        "properties": {
            "id": {
                "readOnly": true,
                "type": "integer"
            },
            "title": {
                "type": "string"
            },
            "audit": {
                "type": "string",
                "format": "iri-reference",
                "example": "https://example.com/"
            },
            "status": {
                "type": "string",
                "format": "iri-reference",
                "example": "https://example.com/"
            }
        }
    },
    "Audit": {
        "type": "object",
        "description": "",
        "deprecated": false,
        "properties": {
            "name": {
                "readOnly": true,
                "type": "string"
            },
            "value": {
                "readOnly": true,
                "allOf": [
                    {
                        "type": "string"
                    },
                    {
                        "type": "integer"
                    }
                ]
            },
            "id": {
                "readOnly": true,
                "anyOf": [
                    {
                        "type": "string"
                    },
                    {
                        "type": "integer"
                    }
                ]
            },
            "cases": {
                "readOnly": true
            }
        }
    },
    "Status": {
        "See above": "Same as Audit"
    }
}
```
</details>

```shell
curl -X 'GET' 'https://example.com/articles/1' -H 'accept: application/json'
```

```json
{
  "id": 1,
  "title": "Once Upon A Title",
  "audit": "/audits/passed",
  "status": "/statuses/1"
}
```

This immediately breaks `Article::status` as it becomes an iri instead. Normal and correct for everything that's **not** an enum.

… and the enum resource routes produce "interesting" responses …

```shell
curl -X 'GET' 'https://example.com/audits/pending' -H 'accept: application/json'
```

<details>

<summary>Response</summary>

```json
{
  "name": "Pending",
  "value": "pending",
  "id": "pending",
  "cases": [
    "/audits/pending",
    {
      "name": "Passed",
      "value": "passed",
      "id": "passed",
      "cases": [
        "/audits/pending",
        "/audits/passed",
        {
          "name": "Failed",
          "value": "failed",
          "id": "failed",
          "cases": [
            "/audits/pending",
            "/audits/passed",
            "/audits/failed"
          ]
        }
      ]
    },
    {
      "name": "Failed",
      "value": "failed",
      "id": "failed",
      "cases": [
        "/audits/pending",
        {
          "name": "Passed",
          "value": "passed",
          "id": "passed",
          "cases": [
            "/audits/pending",
            "/audits/passed",
            "/audits/failed"
          ]
        },
        "/audits/failed"
      ]
    }
  ]
}
```

</details>

```shell
curl -X 'GET' 'https://example.com/statuses/0' -H 'accept: application/json'
```

<details>

<summary>Response</summary>

```json
{
  "name": "DRAFT",
  "value": 0,
  "description": "Article is not ready for public consumption",
  "id": 0,
  "cases": [
    "/statuses/0",
    {
      "name": "PUBLISHED",
      "value": 1,
      "description": "Article is publicly available",
      "id": 1,
      "cases": [
        "/statuses/0",
        "/statuses/1",
        {
          "name": "ARCHIVED",
          "value": 2,
          "description": "Article content is outdated or superseded",
          "id": 2,
          "cases": [
            "/statuses/0",
            "/statuses/1",
            "/statuses/2"
          ]
        }
      ]
    },
    {
      "name": "ARCHIVED",
      "value": 2,
      "description": "Article content is outdated or superseded",
      "id": 2,
      "cases": [
        "/statuses/0",
        {
          "name": "PUBLISHED",
          "value": 1,
          "description": "Article is publicly available",
          "id": 1,
          "cases": [
            "/statuses/0",
            "/statuses/1",
            "/statuses/2"
          ]
        },
        "/statuses/2"
      ]
    }
  ]
}
```

</details>

### LINKED ISSUE #6317: Example documentation is ignored for Enum resources
**API Platform version(s) affected**: 3.2.20

**Description**  

When using a Backed Enum class as an API resource, any `example` value provided in the `ApiProperty` attribute is ignored and does not appear in the OpenAPI documentation.

This is true for `example: 'foo'` and `openapiContext: ['example' => 'foo']`.

On API resources that are Doctrine entities, the supplied example values appear in the OpenAPI documentation.

**How to reproduce**  

https://github.com/gnito-org/problem-replicator-api-platform-enum-example

**Possible Solution**  

**Additional Context**

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #6298, closes #6317
| License       | MIT

Intended for 3.4, so now it is just for review. 

Depends on #6288 as I want to refactor `BackedEnumPlainResourceTest` test methods into those classes.

- Add a provider for `BackedEnum`s
- Identifier is `value` by default, and implementing `function getId()`, etc, works too
- Resource operations will default to only `Get` & `GetCollection` unless otherwise configured 

```php
#[ApiResource]
enum Status: int
{
    case DRAFT = 0;
    case PUBLISHED = 1;
}
```

`GET /statuses`
```json
[
  {
    "name": "DRAFT",
    "value": 0
  },
  {
    "name": "PUBLISHED",
    "value": 1
  }
]
```

`GET /statuses/1`
```json
{
  "name": "PUBLISHED",
  "value": 1
}
```

### DIFF
--- a/src/Metadata/Resource/Factory/BackedEnumResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/BackedEnumResourceMetadataCollectionFactory.php
@@ -24,6 +24,7 @@ use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
 final class BackedEnumResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
 {
     public const PROVIDER = 'api_platform.state_provider.backed_enum';
+
     public function __construct(private readonly ResourceMetadataCollectionFactoryInterface $decorated)
     {
     }
--- a/src/Metadata/Resource/Factory/BackedEnumResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/BackedEnumResourceMetadataCollectionFactory.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Metadata\Resource\Factory;
+
+use ApiPlatform\Metadata\Operations;
+use ApiPlatform\Metadata\Resource\ResourceMetadataCollection;
+
+/**
+ * Triggers resource deprecations.
+ *
+ * @internal
+ */
+final class BackedEnumResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
+{
+    public const PROVIDER = 'api_platform.state_provider.backed_enum';
+    public function __construct(private readonly ResourceMetadataCollectionFactoryInterface $decorated)
+    {
+    }
+
+    public function create(string $resourceClass): ResourceMetadataCollection
+    {
+        $resourceMetadataCollection = $this->decorated->create($resourceClass);
+        if (!is_a($resourceClass, \BackedEnum::class, true)) {
+            return $resourceMetadataCollection;
+        }
+
+        foreach ($resourceMetadataCollection as $i => $resourceMetadata) {
+            $newOperations = [];
+            foreach ($resourceMetadata->getOperations() as $operationName => $operation) {
+                $newOperations[$operationName] = $operation;
+
+                if (null !== $operation->getProvider()) {
+                    continue;
+                }
+
+                $newOperations[$operationName] = $operation->withProvider(self::PROVIDER);
+            }
+
+            $newGraphQlOperations = [];
+            foreach ($resourceMetadata->getGraphQlOperations() as $operationName => $operation) {
+                $newGraphQlOperations[$operationName] = $operation;
+
+                if (null !== $operation->getProvider()) {
+                    continue;
+                }
+
+                $newGraphQlOperations[$operationName] = $operation->withProvider(self::PROVIDER);
+            }
+
+            $resourceMetadataCollection[$i] = $resourceMetadata->withOperations(new Operations($newOperations))->withGraphQlOperations($newGraphQlOperations);
+        }
+
+        return $resourceMetadataCollection;
+    }
+}
--- a/src/Metadata/Resource/Factory/LinkFactory.php
+++ b/src/Metadata/Resource/Factory/LinkFactory.php
@@ -59,6 +59,9 @@ final class LinkFactory implements LinkFactoryInterface, PropertyLinkFactoryInte
 
         $link = (new Link())->withFromClass($resourceClass)->withIdentifiers($identifiers);
         $parameterName = $identifiers[0];
+        if ('value' === $parameterName && enum_exists($resourceClass)) {
+            $parameterName = 'id';
+        }
 
         if (1 < \count($identifiers)) {
             $parameterName = 'id';
@@ -153,6 +156,10 @@ final class LinkFactory implements LinkFactoryInterface, PropertyLinkFactoryInte
 
         if ($hasIdProperty && !$identifiers) {
             return ['id'];
+        }
+
+        if (!$hasIdProperty && !$identifiers && enum_exists($resourceClass)) {
+            return ['value'];
         }
 
         return $identifiers;
--- a/src/State/Provider/BackedEnumProvider.php
+++ b/src/State/Provider/BackedEnumProvider.php
@@ -14,21 +14,18 @@ declare(strict_types=1);
 namespace ApiPlatform\State\Provider;
 
 use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\Exception\RuntimeException;
 use ApiPlatform\Metadata\Operation;
 use ApiPlatform\State\ProviderInterface;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 final class BackedEnumProvider implements ProviderInterface
 {
-    public function __construct(private ProviderInterface $decorated)
-    {
-    }
-
     public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
     {
         $resourceClass = $operation->getClass();
         if (!$resourceClass || !is_a($resourceClass, \BackedEnum::class, true)) {
-            return $this->decorated->provide($operation, $uriVariables, $context);
+            throw new RuntimeException('This resource is not an enum');
         }
 
         if ($operation instanceof CollectionOperationInterface) {
--- a/src/State/Provider/BackedEnumProvider.php
+++ b/src/State/Provider/BackedEnumProvider.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\State\Provider;
+
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProviderInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class BackedEnumProvider implements ProviderInterface
+{
+    public function __construct(private ProviderInterface $decorated)
+    {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
+    {
+        $resourceClass = $operation->getClass();
+        if (!$resourceClass || !is_a($resourceClass, \BackedEnum::class, true)) {
+            return $this->decorated->provide($operation, $uriVariables, $context);
+        }
+
+        if ($operation instanceof CollectionOperationInterface) {
+            return $resourceClass::cases();
+        }
+
+        $id = $uriVariables['id'] ?? null;
+        if (null === $id) {
+            throw new NotFoundHttpException('Not Found');
+        }
+
+        if ($enum = $this->resolveEnum($resourceClass, $id)) {
+            return $enum;
+        }
+
+        throw new NotFoundHttpException('Not Found');
+    }
+
+    /**
+     * @param class-string $resourceClass
+     */
+    private function resolveEnum(string $resourceClass, string|int $id): ?\BackedEnum
+    {
+        $reflectEnum = new \ReflectionEnum($resourceClass);
+        $type = (string) $reflectEnum->getBackingType();
+
+        if ('int' === $type) {
+            if (!is_numeric($id)) {
+                return null;
+            }
+            $enum = $resourceClass::tryFrom((int) $id);
+        } else {
+            $enum = $resourceClass::tryFrom($id);
+        }
+
+        // @deprecated enums will be indexable only by value in 4.0
+        $enum ??= array_reduce($resourceClass::cases(), static fn ($c, \BackedEnum $case) => $id === $case->name ? $case : $c, null);
+
+        return $enum;
+    }
+}
--- a/src/Symfony/Bundle/Resources/config/state/provider.xml
+++ b/src/Symfony/Bundle/Resources/config/state/provider.xml
@@ -25,6 +25,10 @@
             <argument type="service" id="translator" on-invalid="null" />
         </service>
 
+        <service id="api_platform.state_provider.backed_enum" class="ApiPlatform\State\Provider\BackedEnumProvider" decorates="api_platform.state_provider.main" decoration-priority="300">
+            <argument type="service" id="api_platform.state_provider.backed_enum.inner" />
+        </service>
+
         <service id="api_platform.error_listener" class="ApiPlatform\Symfony\EventListener\ErrorListener">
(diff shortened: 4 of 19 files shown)


src/State/Provider/BackedEnumProvider.php
```diff
@@ -0,0 +1,45 @@
...
+final class BackedEnumProvider implements ProviderInterface
+{
+    public function __construct(private readonly ProviderInterface $decorated)
+    {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
+    {
+        $className = $operation->getClass();
+        if (\is_string($className) && enum_exists($className)) {
```
soyuka, line 29: return early `$this->decorated->provide($operation, $uriVariables, $context);` ?

src/State/Provider/BackedEnumProvider.php
```diff
@@ -0,0 +1,45 @@
...
+            if ($operation instanceof CollectionOperationInterface) {
+                return $className::cases();
+            }
+
+            $id = $uriVariables['id'] ?? $uriVariables['value'] ?? null;
```
soyuka, line 34: probably that we can always use `id` here

tests/Functional/BackedEnumPlainResourceTest.php
```diff
@@ -0,0 +1,84 @@
...
+        $identifiers = $result->getIdentifiers();
+
+        $this->assertCount(1, $identifiers);
+        $this->assertNotContains('id', $identifiers);
+        $this->assertContains('value', $identifiers);
+    }
+
+    public function testCollection(): void
+    {
+        self::createClient()->request('GET', '/plain_enum_resources', ['headers' => ['Accept' => 'application/json']]);
```
soyuka, line 51: we should do Json-LD tests instead also?

soyuka (commented):
would love to see json-ld supported, I need to add my commit to this at some point :)

tests/Functional/BackedEnumPlainResourceTest.php
GwendolenLynch: Totally! As luck has it they are in #6288, hence this PR depending on that one.

src/State/Provider/BackedEnumProvider.php
GwendolenLynch: If an enum does not specify an identifier, then the `LinkFactory` marks `value` as the identifier field.

GwendolenLynch: e.g. the `Status` enum example above:
image

soyuka: We can use `id` as uri template and use the `value` in the `identifiers` mapping if needed

src/Metadata/Resource/Factory/LinkFactory.php
```diff
@@ -59,6 +59,9 @@ public function createLinksFromIdentifiers(Metadata $operation): array
 
         $link = (new Link())->withFromClass($resourceClass)->withIdentifiers($identifiers);
         $parameterName = $identifiers[0];
+        if ('value' === $parameterName && enum_exists($resourceClass)) {
+            $parameterName = 'id';
+        }
```
GwendolenLynch, line 64: Does this match what you had in mind @soyuka?

soyuka: Uri variable should be as followed:

```
uriVariables: ['id' => new Link(parameterName: 'id', identifiers: ['value'])]
```

in the metadata directly so that we don't need any condition here right?

dunglas:
I would use URIs in GraphQL too, as we do for existing IDs and relations. WDYT?

src/Metadata/Resource/Factory/LinkFactory.php
GwendolenLynch: Apologies, but I'm confused again :innocent: … I'm probably missing something obvious, but the `Link` is constructed directly above, and the `parameterName` is set at return: `$link->withParameterName($parameterName)` … So is you suggesting how that return array should be shaped? If so what about non-enums?

GwendolenLynch: Which raises another thing about the `createLinksFromIdentifiers` method … If the `$identifier` array is empty there is an early return (lines 56-58), and then there is the `if (1 < \count($identifiers)) {` (lines 66-69) test that looks like dead code, correct?

GwendolenLynch:
@dunglas it must be my week for getting confused. :sun_with_face: 

I'm very happy to make required changes, but could you be so kind as to give a bit more context, please?

gnito-org:
Please look at how I configure a backed enum as an API resource so that each instance has an IRI just like any other API resource. 

https://github.com/gnito-org/problem-replicator-api-platform-enum-example

This is important for machine discovery of APIs and for consistency.

GwendolenLynch:
Thank you @gnito-org, I can reproduce it here.

src/Metadata/Resource/Factory/LinkFactory.php
soyuka: I need to check this into the details, will try to find some time!

soyuka:
@GwendolenLynch I merged #6288 could you rebase I this should fix some of the failing tests right?

src/Metadata/Resource/Factory/LinkFactory.php
soyuka: okay I know understand, basically what I had in bind was

```
            if (enum_exists($operation->getClass())) {
                $operation = $operation->withUriVariables(['id' => new Link(parameterName: 'id', identifiers: ['value'])]);
            } else {
```

directly inside the `UriTemplateResourceMetadataCollectionFactory` class, your version also populates a Link that didn't specified any identifiers as materialized by the test at `testEnumsAreAssignedValuePropertyAsIdentifierByDefault` ! Thanks!

soyuka:
I'll continue working on this tomorrow, hopefully merging it!

src/Symfony/Bundle/Resources/config/state/provider.xml
```diff
@@ -25,6 +25,11 @@
             <argument type="service" id="translator" on-invalid="null" />
         </service>
 
+        <service id="api_platform.state_provider.backed_enum" class="ApiPlatform\State\Provider\BackedEnumProvider">
+            <tag name="api_platform.state_provider" key="ApiPlatform\State\Provider\BackedEnumProvidevr" />
+            <tag name="api_platform.state_provider" key="api_platform.state_provider.backed_enum" />
+        </service>
```
soyuka, line 31: For the record @GwendolenLynch instead of decorating the provider chain, I choose to declare this as an operation provider instead. This will be called inside the ReadProvider: 

https://github.com/api-platform/core/blob/ef4812a688d4cd20877f5bad52d036e80cd43581/src/State/Provider/ReadProvider.php#L75-L79 

It's easier for userland to decorate our enum provider (or any item provider) instead of hooking into the main decoration chain.

tests/Fixtures/TestBundle/ApiResource/Issue6264/BackedEnumStringTrait.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Tests\Fixtures\TestBundle\ApiResource\Issue6264;
+
+use ApiPlatform\Metadata\Operation;
+use Symfony\Component\Serializer\Attribute\Groups;
+
+trait BackedEnumStringTrait
```
soyuka, line 19: For further readers, this was duplicated from the BackedEnumTrait as graphql "guesses" the type from an union:

https://github.com/api-platform/core/blob/ef4812a688d4cd20877f5bad52d036e80cd43581/src/GraphQl/Type/FieldsBuilder.php#L232-L239

I'm not a huge fan, ideally we should provide a fix at some point there, even though when representing resources one should avoid **scalar** union types in my opinion.

soyuka:
Thanks @GwendolenLynch!

src/Metadata/Resource/Factory/BackedEnumResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,67 @@
...
+
+                if (null !== $operation->getProvider()) {
+                    continue;
+                }
+
+                $newOperations[$operationName] = $operation->withProvider(self::PROVIDER);
+            }
+
+            $newGraphQlOperations = [];
+            foreach ($resourceMetadata->getGraphQlOperations() as $operationName => $operation) {
```
hafkenscheid, line 52: This does not work if there are no GraphQlOperations

papppeter:
Hi, @soyuka could you please send me a link to the docs please? 
we would like to use this feature, but could not find the proper documentation.

PR #6314, opened 2024-04-15
fix(graphql): check inheritance in ResolverProvider

jotwea:
| Q             | A
| ------------- | ---
| Branch       | 3.2
| Tickets       | -
| License       | MIT
| Doc PR        | -

With the new `ResolverProvider` some cases with inheritance are broken in our project. It was not a big deal to find out that `ItemResolverFactory` also checked if the resolver provided a subclass of the `resourceClass`. So I added this also to the `ResolverProvider` and everything works like a charm. I am not sure if it should also be extended in `ReadProvider` - there is similar code, but I had no problem with that. Thx for your advice.

### DIFF
--- a/src/GraphQl/State/Provider/ResolverProvider.php
+++ b/src/GraphQl/State/Provider/ResolverProvider.php
@@ -69,7 +69,7 @@ final class ResolverProvider implements ProviderInterface
             return $itemClass;
         }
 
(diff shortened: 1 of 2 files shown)


soyuka:
nice catch, can you run the php-cs-fixer?

soyuka:
thanks @jotwea !

PR #6315, opened 2024-04-15
chore: deprecate ApiPlatform\Api\FilterInterface

louismariegaborit:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | 

The PR #6312 that add deprecated on `ApiPlatform\Api\FilterInterface` trigger deprecation when class extends `ApiPlatform\Metadata\FilterInterface`.  
I suggest to put the alias in the `ApiPlatform\Api\FilterInterface` and leave `ApiPlatform\Metadata\FilterInterface` with the final content.


src/Api/FilterInterface.php
```diff
@@ -13,46 +13,60 @@
...
- * Filters applicable on a resource.
- *
- * @author Kévin Dunglas <dunglas@gmail.com>
- *
- * @deprecated
- */
-interface FilterInterface
-{
+if (interface_exists(\ApiPlatform\Metadata\FilterInterface::class)) {
+    trigger_deprecation('api-platform', '3.3', sprintf('%s is deprecated in favor of %s. This class will be removed in 4.0.', self::class, \ApiPlatform\Metadata\FilterInterface::class));
```
soyuka, line 17:
```suggestion
    trigger_deprecation('api-platform', '3.3', sprintf('%s is deprecated in favor of %s. This class will be removed in 4.0.', \ApiPlatform\Api\FilterInterface::class, \ApiPlatform\Metadata\FilterInterface::class));
```

src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -187,8 +186,6 @@ public function load(array $configs, ContainerBuilder $container): void
 
         $container->registerForAutoconfiguration(FilterInterface::class)
             ->addTag('api_platform.filter');
-        $container->registerForAutoconfiguration(LegacyFilterInterface::class)
```
soyuka, line 190: this will break filters when upgrading no? can't we keep that?

src/Metadata/FilterInterface.php
```diff
@@ -13,59 +13,47 @@
...
+     *     e.g. 'schema' => [
+     *       'type' => 'string',
+     *       'enum' => ['value_1', 'value_2'],
+     *     ]
+     * The description can contain additional data specific to a filter.
+     *
+     * @see \ApiPlatform\OpenApi\Factory\OpenApiFactory::getFiltersParameters
      *
-     * @author Kévin Dunglas <dunglas@gmail.com>
+     * @return array<string, array{property: ?string, type: string, required: bool, description?: string, strategy?: string, is_collection?: bool, swagger?: array<string, mixed>, openapi?: array<string, mixed>|\ApiPlatform\OpenApi\Model\Parameter, schema?: array<string, mixed>}>
```
soyuka, line 56: Mhh my bad I think that `?string` is wrong, I thought that everything was optional but let's keep that. We may need to change the tests you mention, there shouldn't be lots of them right? 

```suggestion
     * @return array<string, array{property: string, type: string, required: bool, description?: string, strategy?: string, is_collection?: bool, swagger?: array<string, mixed>, openapi?: array<string, mixed>|\ApiPlatform\OpenApi\Model\Parameter, schema?: array<string, mixed>}>
```

src/Metadata/Tests/Resource/Factory/ParameterResourceMetadataCollectionFactoryTests.php
```diff
@@ -32,8 +32,18 @@ public function testParameterFactory(): void
...
-                    'everywhere' => ['openapi' => ['allowEmptyValue' => true]],
+                    'hydra' => [
+                        'property' => null,
+                        'type' => 'string',
+                        'required' => false,
+                        'schema' => ['type' => 'foo'],
+                        'openapi' => new Parameter('test', 'query'),
+                    ],
+                    'everywhere' => [
+                        'property' => null,
```
soyuka, line 43:
```suggestion
                        'property' => 'everywhere',
```

src/Metadata/FilterInterface.php
louismariegaborit: Rollback done.

soyuka:
thanks @louismariegaborit

PR #6319, opened 2024-04-16
fix(jsonapi): add support for pagination array parameter name

seb37800:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        | 

This bugfix is related to the jsonapi spec & the handling of pagination parameters in the pagination service.
Following the JSON:API specification around pagination => https://jsonapi.org/format/#fetching-pagination
When i setup api_platform.yaml with
```
api_platform:
    collection:
        pagination:
            page_parameter_name: page[number]
            items_per_page_parameter_name: page[size]
```
And use ApiPlatform\State\Pagination\Pagination in a custom provider, we can't retrieve the `page[number]` & `page[size]` in `$context['filters']`.

Thanks to @GregoireHebert for the help.

In addition this fix that let us specify array in pagination parameter names will fix the generated documentation.

### DIFF
--- a/src/State/Pagination/Pagination.php
+++ b/src/State/Pagination/Pagination.php
@@ -222,16 +222,16 @@ final class Pagination
     /**
      * Extract pagination parameter
      * page[page] => $contextFilters['page'] with default configuration page_parameter_name: page
-     * page[number] => $contextFilters['_page'][number] with configuration page_parameter_name: page[number]
+     * page[number] => $contextFilters['_page'][number] with configuration page_parameter_name: page[number].
      */
     private function extractParameter(array $contextFilters, string $parameterName)
     {
         preg_match_all("/[\w-]+/", $parameterName, $matches);
         foreach ($matches[0] as $i => $key) {
-            if ($i === 0 && $key === 'page' && count($matches[0]) > 1) {
+            if (0 === $i && 'page' === $key && \count($matches[0]) > 1) {
                 $key = '_page';
             }
-            if (\is_array($contextFilters) === false) {
+            if (false === \is_array($contextFilters)) {
                 return $contextFilters;
             }
             if (!\array_key_exists($key, $contextFilters)) {
--- a/src/State/Pagination/Pagination.php
+++ b/src/State/Pagination/Pagination.php
@@ -224,7 +224,7 @@ final class Pagination
      * page[page] => $contextFilters['page'] with default configuration page_parameter_name: page
      * page[number] => $contextFilters['_page'][number] with configuration page_parameter_name: page[number].
      */
-    private function extractParameter(array $contextFilters, string $parameterName)
+    private function extractParameter(array $contextFilters, string $parameterName): mixed
     {
         preg_match_all("/[\w-]+/", $parameterName, $matches);
         foreach ($matches[0] as $i => $key) {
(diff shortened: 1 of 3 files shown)


src/State/Pagination/Pagination.php
```diff
@@ -219,13 +219,38 @@ private function getGraphQlEnabled(?Operation $operation): bool
         return $operation?->getPaginationEnabled() ?? $enabled;
     }
 
+    /**
+     * Extract pagination parameter
+     * page[page] => $contextFilters['page'] with default configuration page_parameter_name: page
+     * page[number] => $contextFilters['_page'][number] with configuration page_parameter_name: page[number].
+     */
+    private function extractParameter(array $contextFilters, string $parameterName)
```
soyuka, line 227: missing return type

src/State/Pagination/Pagination.php
```diff
@@ -219,13 +219,38 @@ private function getGraphQlEnabled(?Operation $operation): bool
         return $operation?->getPaginationEnabled() ?? $enabled;
     }
 
+    /**
+     * Extract pagination parameter
+     * page[page] => $contextFilters['page'] with default configuration page_parameter_name: page
+     * page[number] => $contextFilters['_page'][number] with configuration page_parameter_name: page[number].
```
soyuka, line 225: you mean that you configure `page[number]` on the `page_parameter_name`, then you use `page[number]=1`?

seb37800: Yeah exactly

src/State/Pagination/Pagination.php
seb37800: i've added mixed, but not very top

src/State/Pagination/Pagination.php
GregoireHebert: To add more context to this, JSON:API consider query parameter such as two parameters named `page[offset]` and `page[limit]` really are two separate parameters, instead of a single page parameter in an array form.

[https://jsonapi.org/format/#query-parameters-families](https://jsonapi.org/format/#query-parameters-families)

seb37800: Yeah or `page[number]` & `page[size]` in my implementation if you work with page numbers instead of offset/limit cf https://jsonapi.org/format/#fetching-pagination

src/State/Pagination/Pagination.php
soyuka: why not ?array ?

src/State/Pagination/Pagination.php
soyuka: can you add a function test and a unit test?

src/State/Pagination/Pagination.php
```diff
@@ -219,13 +219,38 @@ private function getGraphQlEnabled(?Operation $operation): bool
         return $operation?->getPaginationEnabled() ?? $enabled;
     }
 
+    /**
+     * Extract pagination parameter
```
soyuka, line 223: The pagination parameter value?

src/State/Pagination/Pagination.php
```diff
@@ -219,13 +219,38 @@ private function getGraphQlEnabled(?Operation $operation): bool
...
+     * page[page] => $contextFilters['page'] with default configuration page_parameter_name: page
+     * page[number] => $contextFilters['_page'][number] with configuration page_parameter_name: page[number].
+     */
+    private function extractParameter(array $contextFilters, string $parameterName): mixed
+    {
+        preg_match_all("/[\w-]+/", $parameterName, $matches);
```
soyuka, line 229: I don't understand the need of this regex here, we know the possible parameter names no? I'd love a test to understand this better.

seb37800:
@soyuka If you've time to check the unit test to better understand the use case.
I don't find a way to make a behat test due to an unknown factor on how to change the default api platform config for this behat test only.

seb37800:
@soyuka @dunglas any news around this PR ?

PR #6320, opened 2024-04-16
fix(jsonapi): exit loop when relationship class is found

GwendolenLynch:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

When using the model I describe in https://github.com/api-platform/core/issues/6212#issuecomment-2019909645, the `$className` variable is overwritten by `null` when an additional (3rd +) type is in the union.


soyuka:
shouldn't we use all the types? I mean in your example it should be one of Species (ref/iri), `null`, `int` right?

GwendolenLynch:
Fair call. Updated.

soyuka:
Do you think you can add a test? I usually add a fixture then uses the `JsonSchemaGenerateCommandTest` to add a functional test

GwendolenLynch:
Tests added. The core remaining issue from #6212 isn't in the schema factory so I just extended the tests I added in #6223.

soyuka:
nice thanks!

PR #6332, opened 2024-04-23
feat(openapi): add error resources schemes

### LINKED ISSUE #6003: Add ErrorResource Schemes to OpenAPI documentation
**Description**  
Currently, we have a way to document error HTTP status codes supported by an Operation with `exceptionToStatus` property, as well as adding our own documentation with `openApi` property.
That much is great, however by default we have limited information about errors in the OpenAPI documentation: an HTTP code, description and "string" as scheme.

With v3.2, we can now represent errors/exceptions as ErrorResource, which should give us the ability to show proper scheme of a response in documentation.

For that we need to have a way to map ErrorResources to specific Operations. Knowing what ErrorResources could be thrown, we can turn ErrorResource classes into OpenAPI documentation.

**Example**  
A way to solve this would be by introducing to Operation class a new property `errors`, that contains a list of ErrorResources used:

```php
use ApiPlatform\Metadata\Get;

#[ApiResource(
    operations: [
        new Get(
            uriTemplate: '/orders/{id}',
            errors: [
                OrderNotFoundError::class,
            ],
        ),
    ],
)]
class Order {}
```

```php
use ApiPlatform\Metadata\ErrorResource;
use ApiPlatform\Metadata\Exception\ProblemExceptionInterface;

#[ErrorResource(status: 404)]
class OrderNotFoundException extends \Exception implements ProblemExceptionInterface
{
    // ...
}
```

JacquesDurand:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #6003 
| License       | MIT
| Doc PR        | api-platform/docs#TODO

Allow linking ErrorResources to ApiResources to automatically generate openapi documentation for errors: 

```php
#[ApiResource(operations: [
    new GetCollection(errors: [MyDomainException::class])
])]
class Greeting
{
...
}
```

with the following Exception (currently needs to be an `ErrorResource` and to implement `ProblemExceptionInterface`): 

```php
#[ErrorResource]
class MyDomainException extends \Exception implements ProblemExceptionInterface
{
    public function getType(): string
    {
        return 'Teapot';
    }
    public function getTitle(): ?string
    {
        return null;
    }
    public function getStatus(): ?int
    {
        return 418;
    }
    public function getDetail(): ?string
    {
        return 'My specific Domain Exception';
    }
    public function getInstance(): ?string
    {
        return null;
    }
}
```

would allow the following: 

image

Do you think it might be an interesting feature ?

There might still be a few things I am not too familiar with, any advice is more than welcome !

### DIFF
--- a/src/Metadata/HttpOperation.php
+++ b/src/Metadata/HttpOperation.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Metadata;
 
+use ApiPlatform\Metadata\Exception\ProblemExceptionInterface;
 use ApiPlatform\OpenApi\Attributes\Webhook;
 use ApiPlatform\OpenApi\Model\Operation as OpenApiOperation;
 use ApiPlatform\State\OptionsInterface;
@@ -73,11 +74,12 @@ class HttpOperation extends Operation
      *     class?: string|null,
      *     name?: string,
      * }|string|false|null $output {@see https://api-platform.com/docs/core/dto/#specifying-an-input-or-an-output-data-representation}
-     * @param string|array|bool|null $mercure   {@see https://api-platform.com/docs/core/mercure}
-     * @param string|bool|null       $messenger {@see https://api-platform.com/docs/core/messenger/#dispatching-a-resource-through-the-message-bus}
-     * @param string|callable|null   $provider  {@see https://api-platform.com/docs/core/state-providers/#state-providers}
-     * @param string|callable|null   $processor {@see https://api-platform.com/docs/core/state-processors/#state-processors}
-     * @param WebLink[]|null         $links
+     * @param string|array|bool|null                              $mercure   {@see https://api-platform.com/docs/core/mercure}
+     * @param string|bool|null                                    $messenger {@see https://api-platform.com/docs/core/messenger/#dispatching-a-resource-through-the-message-bus}
+     * @param string|callable|null                                $provider  {@see https://api-platform.com/docs/core/state-providers/#state-providers}
+     * @param string|callable|null                                $processor {@see https://api-platform.com/docs/core/state-processors/#state-processors}
+     * @param WebLink[]|null                                      $links
+     * @param array<class-string<ProblemExceptionInterface>>|null $errors
      */
     public function __construct(
         protected string $method = 'GET',
@@ -154,6 +156,7 @@ class HttpOperation extends Operation
         protected bool|OpenApiOperation|Webhook|null $openapi = null,
         protected ?array $exceptionToStatus = null,
         protected ?array $links = null,
+        protected ?array $errors = null,
 
         ?string $shortName = null,
         ?string $class = null,
@@ -620,6 +623,22 @@ class HttpOperation extends Operation
     {
         $self = clone $this;
         $self->links = $links;
+
+        return $self;
+    }
+
+    public function getErrors(): ?array
+    {
+        return $this->errors;
+    }
+
+    /**
+     * @param class-string<ProblemExceptionInterface>[] $errors
+     */
+    public function withErrors(array $errors): self
+    {
+        $self = clone $this;
+        $self->errors = $errors;
 
         return $self;
     }
--- a/src/Metadata/Resource/Factory/MetadataCollectionFactoryTrait.php
+++ b/src/Metadata/Resource/Factory/MetadataCollectionFactoryTrait.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace ApiPlatform\Metadata\Resource\Factory;
 
 use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\Exception\InvalidArgumentException;
+use ApiPlatform\Metadata\Exception\ProblemExceptionInterface;
 use ApiPlatform\Metadata\Exception\RuntimeException;
 use ApiPlatform\Metadata\GraphQl\Operation as GraphQlOperation;
 use ApiPlatform\Metadata\HttpOperation;
@@ -88,6 +90,13 @@ trait MetadataCollectionFactoryTrait
                 foreach ($resource->getOperations() ?? new Operations() as $operation) {
                     [$key, $operation] = $this->getOperationWithDefaults($resource, $operation);
                     $operations[$key] = $operation;
+                    if (null !== $operation->getErrors()) {
+                        foreach ($operation->getErrors() as $error) {
+                            if (!is_a($error, ProblemExceptionInterface::class, true)) {
+                                throw new InvalidArgumentException(\sprintf('The error class "%s" does not implement "%s". Did you forget a use statement?', $error, ProblemExceptionInterface::class));
+                            }
+                        }
+                    }
                 }
                 if ($operations) {
                     $resource = $resource->withOperations(new Operations($operations));
--- a/src/Metadata/Resource/Factory/MetadataCollectionFactoryTrait.php
+++ b/src/Metadata/Resource/Factory/MetadataCollectionFactoryTrait.php
@@ -14,8 +14,6 @@ declare(strict_types=1);
 namespace ApiPlatform\Metadata\Resource\Factory;
 
 use ApiPlatform\Metadata\ApiResource;
-use ApiPlatform\Metadata\Exception\InvalidArgumentException;
-use ApiPlatform\Metadata\Exception\ProblemExceptionInterface;
 use ApiPlatform\Metadata\Exception\RuntimeException;
 use ApiPlatform\Metadata\GraphQl\Operation as GraphQlOperation;
 use ApiPlatform\Metadata\HttpOperation;
@@ -90,13 +88,6 @@ trait MetadataCollectionFactoryTrait
                 foreach ($resource->getOperations() ?? new Operations() as $operation) {
                     [$key, $operation] = $this->getOperationWithDefaults($resource, $operation);
                     $operations[$key] = $operation;
-                    if (null !== $operation->getErrors()) {
-                        foreach ($operation->getErrors() as $error) {
-                            if (!is_a($error, ProblemExceptionInterface::class, true)) {
-                                throw new InvalidArgumentException(\sprintf('The error class "%s" does not implement "%s". Did you forget a use statement?', $error, ProblemExceptionInterface::class));
-                            }
-                        }
-                    }
                 }
                 if ($operations) {
                     $resource = $resource->withOperations(new Operations($operations));
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -21,6 +21,7 @@ use ApiPlatform\JsonSchema\TypeFactoryInterface;
 use ApiPlatform\Metadata\ApiResource;
 use ApiPlatform\Metadata\CollectionOperationInterface;
 use ApiPlatform\Metadata\Error;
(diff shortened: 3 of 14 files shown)


soyuka:
This is nuts! Love it! Will review shortly.

src/Metadata/HttpOperation.php
```diff
@@ -203,6 +207,13 @@ public function __construct(
         array|Parameters|null $parameters = null,
         array $extraProperties = [],
     ) {
+        if (null !== $this->errors) {
+            foreach ($this->errors as $error) {
+                if (!(new $error()) instanceof ProblemExceptionInterface) {
```
soyuka, line 212: let's avoid logic inside constructors especially on attributes, if a check is needed it should happen inside metadata factories.

soyuka (changes requested):
small change

danielhomer:
Please don't close this, this is a much needed feature.

JacquesDurand:
I'm so sorry I completely forgot about this, I'll get back to it ASAP

src/Metadata/Resource/Factory/MetadataCollectionFactoryTrait.php
```diff
@@ -88,6 +90,13 @@ private function buildResourceOperations(array $metadataCollection, string $reso
                 foreach ($resource->getOperations() ?? new Operations() as $operation) {
                     [$key, $operation] = $this->getOperationWithDefaults($resource, $operation);
                     $operations[$key] = $operation;
+                    if (null !== $operation->getErrors()) {
+                        foreach ($operation->getErrors() as $error) {
+                            if (!(new $error()) instanceof ProblemExceptionInterface) {
```
soyuka, line 95: `is_a($error, ProblemExceptionInterface::class, true)`

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -269,6 +270,21 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
             $openapiOperation = $openapiOperation->withParameters($openapiParameters);
             $existingResponses = $openapiOperation->getResponses() ?: [];
             $overrideResponses = $operation->getExtraProperties()[self::OVERRIDE_OPENAPI_RESPONSES] ?? $this->openApiOptions->getOverrideResponses();
+            if ($operation instanceof HttpOperation && null !== $operation->getErrors()) {
+                foreach ($operation->getErrors() as $error) {
+                    /** @var ProblemExceptionInterface $exception */
+                    $exception = new $error();
```
soyuka, line 276: Not sure that we should call the constructor here, I'd suggest to use `$operation = $this->resourceMetadataCollectionFactory->create($error::class)->getOperation()` and to read `$operation->getStatus()` although I prefer the `getStatus` on that instance but I'm afraid we can't enforce that:
- the constructor has no arguments
- `getStatus` always exists (or we could've used Reflection->newInstanceWithoutConstructor)

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -269,6 +270,21 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
+
+                    $operationErrorSchemas = [];
+                    foreach ($responseMimeTypes as $operationFormat) {
+                        $operationErrorSchema = $this->jsonSchemaFactory->buildSchema($error, $operationFormat, Schema::TYPE_OUTPUT, null, $schema, null, $forceSchemaCollection);
+                        $operationErrorSchemas[$operationFormat] = $operationErrorSchema;
+                        $this->appendSchemaDefinitions($schemas, $operationErrorSchema->getDefinitions());
+                    }
+
+                    $openapiOperation = $this->buildOpenApiResponse($existingResponses, $exception->getStatus(), $exception->getType(), $openapiOperation, $operation, $responseMimeTypes, $operationErrorSchemas, $resourceMetadataCollection);
```
soyuka, line 285: `type` on problem details is:

> The "type" member is a JSON string containing a URI reference [[URI](https://datatracker.ietf.org/doc/html/rfc3986)] that identifies the problem type. Consumers MUST use the "type" URI (after resolution, if necessary) as the problem type's primary identifier.When this member is not present, its value is assumed to be "about:blank".

https://datatracker.ietf.org/doc/html/rfc9457#name-type

we should probably use something else like `$operation->getDescription()` or `$exception->getDetails()`.

src/OpenApi/Factory/OpenApiFactory.php
JacquesDurand: I might be mistaken, but won't `$this->resourceMetadataCollectionFactory->create($error::class)->getOperation()` throw a `OperationNotFoundException` for an ErrorResource ?
I do agree that the `new $error()` is itself error-prone if the constructor has arguments, I had not thought of this

soyuka: `ErrorResources` declares an ApiResource with an Error operation.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -269,6 +270,21 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
+                    $exception = (new \ReflectionClass($error))->newInstanceWithoutConstructor();
+
+                    $operationErrorSchemas = [];
+                    foreach ($responseMimeTypes as $operationFormat) {
+                        $operationErrorSchema = $this->jsonSchemaFactory->buildSchema($error, $operationFormat, Schema::TYPE_OUTPUT, null, $schema, null, $forceSchemaCollection);
+                        $operationErrorSchemas[$operationFormat] = $operationErrorSchema;
+                        $this->appendSchemaDefinitions($schemas, $operationErrorSchema->getDefinitions());
+                    }
+
+                    $openapiOperation = $this->buildOpenApiResponse($existingResponses, $exception->getStatus(), $exception->getDetail(), $openapiOperation, $operation, $responseMimeTypes, $operationErrorSchemas, $resourceMetadataCollection);
```
soyuka, line 285:
```suggestion
                    $openapiOperation = $this->buildOpenApiResponse($existingResponses, $exception->getStatus(), $exception->getTitle(), $openapiOperation, $operation, $responseMimeTypes, $operationErrorSchemas, $resourceMetadataCollection);
```

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -269,6 +270,21 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
             $openapiOperation = $openapiOperation->withParameters($openapiParameters);
             $existingResponses = $openapiOperation->getResponses() ?: [];
             $overrideResponses = $operation->getExtraProperties()[self::OVERRIDE_OPENAPI_RESPONSES] ?? $this->openApiOptions->getOverrideResponses();
+            if ($operation instanceof HttpOperation && null !== $operation->getErrors()) {
+                foreach ($operation->getErrors() as $error) {
+                    /** @var ProblemExceptionInterface $exception */
+                    $exception = (new \ReflectionClass($error))->newInstanceWithoutConstructor();
```
soyuka, line 276: let's catch ReflectionException and fallback on something else: 

```
$status = null; $description = null;
try {
    $exception = (new \ReflectionClass($error))->newInstanceWithoutConstructor();
    $status = $exception->getStatus();
    $title = $exception->getTitle();
} catch (\ReflectionException) {
}

if (!$status || !$description) {
    // try to get the operation
}
//below

if (!$status || !$description) {
// log that error is misconfigured ?
}

                    $openapiOperation = $this->buildOpenApiResponse($existingResponses, $status, $title, $openapiOperation, $operation, $responseMimeTypes, $operationErrorSchemas, $resourceMetadataCollection);
```

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -843,4 +873,61 @@ private function mergeParameter(Parameter $actual, Parameter $defined): Paramete
...
+        foreach ($errors as $error) {
+            if (!is_a($error, ProblemExceptionInterface::class, true)) {
+                $this->logger?->warning(\sprintf('The error class "%s" does not implement "%s". Did you forget a use statement?', $error, ProblemExceptionInterface::class));
+            }
+
+            $status = null;
+            $description = null;
+            try {
+                /** @var ProblemExceptionInterface */
+                $exception = (new \ReflectionClass($error))->newInstanceWithoutConstructor();
```
dunglas, line 893: `new $error();` should do the trick. No need for reflection here.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -843,4 +873,61 @@ private function mergeParameter(Parameter $actual, Parameter $defined): Paramete
...
+    /**
+     * @param string[]              $errors
+     * @param array<string, string> $responseMimeTypes
+     */
+    private function addOperationErrors(Operation $operation, array $errors, array $responseMimeTypes, ResourceMetadataCollection $resourceMetadataCollection, Schema $schema, \ArrayObject $schemas): Operation
+    {
+        $existingResponses = null;
+        foreach ($errors as $error) {
+            if (!is_a($error, ProblemExceptionInterface::class, true)) {
+                $this->logger?->warning(\sprintf('The error class "%s" does not implement "%s". Did you forget a use statement?', $error, ProblemExceptionInterface::class));
```
dunglas, line 886:
```suggestion
                $this->logger?->warning(\sprintf('The error class "%s" does not implement "%s". Did you forget a use statement?', $error, ProblemExceptionInterface::class));

                continue;
```

But shouldn't we throw an exception instead of logging a warning? This will made it easier to catch this kind of errors during development.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -843,4 +873,61 @@ private function mergeParameter(Parameter $actual, Parameter $defined): Paramete
...
+                $errorOperation = $this->resourceMetadataFactory->create($error)->getOperation();
+                if (!is_a($errorOperation, Error::class)) {
+                    $this->logger?->warning(\sprintf('The error class %s is not an ErrorResource', $error));
+                    continue;
+                }
+                /* @var Error $errorOperation */
+                $status ??= $errorOperation->getStatus();
+                $description ??= $errorOperation->getDescription();
+            } catch (ResourceClassNotFoundException|OperationNotFoundException) {
+                $this->logger?->warning(\sprintf('The error class %s is not an ErrorResource', $error));
```
dunglas, line 909: Same here. It's probably useful to not catch this exception and be sure the developer handles this case.

src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -843,4 +873,61 @@ private function mergeParameter(Parameter $actual, Parameter $defined): Paramete
...
+                continue;
+            }
+
+            if (!$status) {
+                $this->logger?->error(\sprintf(
```
dunglas, line 914: Same here. I would throw.

dunglas (changes requested):

dunglas:
Cool feature!

src/OpenApi/Factory/OpenApiFactory.php
soyuka: I disagree, it's possible that an exception in userland has constructor arguments

dunglas: But then constructing an instance without calling the constructor will likely create an object in an invalid sate, which is worse than an error right?

What if the title is passed as parameter in constructor for instance?

soyuka:
Many thanks @JacquesDurand ! We need some documentation for that awesome feature!

PR #6359, opened 2024-05-07
fix(symfony): no read should not throw on wrong uri variables

### LINKED ISSUE #6358: BC break in 3.3.x: resource_class has been changed for DTOs
**API Platform version(s) affected**: 3.2.2

**Description**  
I guess this is related to https://github.com/api-platform/core/issues/6211. In api-platform/core <=3.3.2, >=3.3.0, error we get is 

> Warning: Undefined array key "resource_class" in src/Serializer/CompanyNormalizer.php

From 3.3.2, error is gone, however `resource_class` that's being passed to normalizer is different form the one before 3.2.x

In 3.2.x `resource_class` is Client::class. In 3.3.x, `resource_class` is Company::class

**How to reproduce**  
```php
#[ApiResource(
    operations: [
        new GetCollection(
            class: Company::class,
            normalizationContext: ['groups' => ['client:list']],
            output: ClientOutput::class,
            provider: CompanyCollectionProvider::class,
        ),
    ],
)]
class Client {}

abstract class CompanyNormalizer implements NormalizerInterface
{
    public function supportsNormalization(mixed $data, ?string $format = null, array $context = []): bool
    {
        return $data instanceof Company && $context['resource_class'] === Client::class;
    }

    public function getSupportedTypes(?string $format): array
    {
        return [Company::class => false];
    }
}
```

Problem above is that `supportsNormalization` returns false since 3.3.x and hence objects are not serialized as expected

**Possible Solution**  

**Additional Context**

soyuka:
fixes #6358

### DIFF
--- a/features/main/input_output.feature
+++ b/features/main/input_output.feature
@@ -1,0 +1,15 @@
+Feature: DTO input and output
+  In order to use a hypermedia API
+  As a client software developer
+  I need to be able to use DTOs on my resources as Input or Output objects.
+
+  Background:
+    Given I add "Accept" header equal to "application/ld+json"
+    And I add "Content-Type" header equal to "application/ld+json"
+
+  @!mongodb
+  Scenario: Fetch a collection of outputs with an entityClass as state option
+    When I send a "GET" request to "/output_and_entity_classes"
+    And the JSON node "hydra:member[0].@type" should be equal to "OutputAndEntityClassEntity"
+
+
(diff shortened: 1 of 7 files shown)


ostrolucky:
Fixes my issue

ostrolucky:
Your last commit fixes issue at https://github.com/api-platform/core/issues/6358#issuecomment-2098514418 too

soyuka:
Nice I'll add some tests thanks for the reports!

PR #2680, opened 2019-04-03
fix: update symfony/serializer required version

jockos:
<!-- Please update this template with something that matches your PR -->
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | /
| License       | MIT

<!-- Bug fixes should be based against the current stable version branch, master is for new features only -->

Currently, the required version of the symfony/serializer is ^4.1 but in api-platform/core 2.4 we use an interface that is available only since Serializer v4.2 :
`final class ReservedAttributeNameConverter implements AdvancedNameConverterInterface`

This creates a nice error like that:
```
Script Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache handling the symfony-scripts event terminated with an exception                       

  [RuntimeException]                                                                                                                                                       
  An error occurred when executing the "'cache:clear --no-warmup'" command:                                                                                                
                                                                                                                                                                           
                                                                                                                                                                           
  Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 2 passed to ApiPlatform\Core\JsonApi\Serializer\ConstraintViolationLi  
  stNormalizer::__construct() must implement interface Symfony\Component\Serializer\NameConverter\NameConverterInterface or be null, instance of ApiPlatform\Core\JsonApi  
  \Serializer\ReservedAttributeNameConverter given, called in /var/www/amms-api/var/cache/dev/ContainerV9modiz/appDevDebugProjectContainer.php on line 6054 in /var/www/a  
  mms-api/vendor/api-platform/core/src/JsonApi/Serializer/ConstraintViolationListNormalizer.php:35                                                                         
  Stack trace:                                                                                                                                                             
  #0 /var/www/amms-api/var/cache/dev/ContainerV9modiz/appDevDebugProjectContainer.php(6054): ApiPlatform\Core\JsonApi\Serializer\ConstraintViolationListNormalizer->__con  
  struct(Object(ApiPlatform\Core\Metadata\Property\Factory\CachedPropertyMetadataFactory), Object(ApiPlatform\Core\JsonApi\Serializer\ReservedAttributeNameConverter))     
  #1 /var/www/amms-api/var/cache/dev/ContainerV9modiz/appDevDebugProjectContainer.php(6333): Con in /var/www/amms-api/vendor/api-platform/core/src/JsonApi/Serializer/Con  
[...]

PR #6374, opened 2024-05-18
feat(openapi): allow optional request body content

monitaurus:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | Closes #2680 
| License       | MIT
| Doc PR        | _already existing ([see doc](https://api-platform.com/docs/guides/extend-openapi-documentation/#:~:text=When%20a%20field%20is%20not%20specified%20API%20Platform%20will%20add%20the%20missing%20informations.))_ 

Allow to generate a default request content or description if not defined in the extension of OpenApi using `Model\RequestBody`.

For example, on a `POST /greetings/test` operation on a `Greeting` entity with a `id` and a `name`, the following:
```php
#[ApiResource(
    mercure: true,
    operations: [
        new Post(
            uriTemplate: '/greetings/test',
            openapi: new Operation(
                requestBody: new RequestBody(
                    description: 'Extended description',
                ),
            ),
        ),
    ],
)]
```

Will generate the following OpenApi, with the default generated content:
image

### DIFF
--- a/src/OpenApi/Factory/OpenApiFactory.php
+++ b/src/OpenApi/Factory/OpenApiFactory.php
@@ -394,17 +394,25 @@ final class OpenApiFactory implements OpenApiFactoryInterface
                 );
                 $openapiOperation = $openapiOperation->withRequestBody(new RequestBody($contextRequestBody['description'] ?? '', new \ArrayObject($contextRequestBody['content']), $contextRequestBody['required'] ?? false));
             } elseif (
-                null === $openapiOperation->getRequestBody() && \in_array($method, ['PATCH', 'PUT', 'POST'], true)
+                \in_array($method, ['PATCH', 'PUT', 'POST'], true)
                 && !(false === ($input = $operation->getInput()) || (\is_array($input) && null === $input['class']))
             ) {
-                $operationInputSchemas = [];
-                foreach ($requestMimeTypes as $operationFormat) {
-                    $operationInputSchema = $this->jsonSchemaFactory->buildSchema($resourceClass, $operationFormat, Schema::TYPE_INPUT, $operation, $schema, null, $forceSchemaCollection);
-                    $operationInputSchemas[$operationFormat] = $operationInputSchema;
-                    $this->appendSchemaDefinitions($schemas, $operationInputSchema->getDefinitions());
+                $content = $openapiOperation->getRequestBody()?->getContent();
+                if (null === $content) {
+                    $operationInputSchemas = [];
+                    foreach ($requestMimeTypes as $operationFormat) {
+                        $operationInputSchema = $this->jsonSchemaFactory->buildSchema($resourceClass, $operationFormat, Schema::TYPE_INPUT, $operation, $schema, null, $forceSchemaCollection);
+                        $operationInputSchemas[$operationFormat] = $operationInputSchema;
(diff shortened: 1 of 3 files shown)


src/OpenApi/Factory/OpenApiFactory.php
```diff
@@ -394,17 +394,25 @@ private function collectPaths(ApiResource $resource, ResourceMetadataCollection
...
+                        $this->appendSchemaDefinitions($schemas, $operationInputSchema->getDefinitions());
+                    }
+                    $content = $this->buildContent($requestMimeTypes, $operationInputSchemas);
                 }
 
-                $openapiOperation = $openapiOperation->withRequestBody(new RequestBody(sprintf('The %s %s resource', 'POST' === $method ? 'new' : 'updated', $resourceShortName), $this->buildContent($requestMimeTypes, $operationInputSchemas), true));
+                $openapiOperation = $openapiOperation->withRequestBody(new RequestBody(
+                    description: $openapiOperation->getRequestBody()?->getDescription() ?? sprintf('The %s %s resource', 'POST' === $method ? 'new' : 'updated', $resourceShortName),
+                    content: $content,
+                    required: $openapiOperation->getRequestBody()?->getRequired() ?? true,
```
soyuka, line 414: to understand this correctly the main change is that `true` is now the default value when there's a request body ?

src/OpenApi/Model/RequestBody.php
```diff
@@ -17,16 +17,16 @@ final class RequestBody
 {
     use ExtensionTrait;
 
-    public function __construct(private string $description = '', private ?\ArrayObject $content = null, private bool $required = false)
+    public function __construct(private ?string $description = null, private ?\ArrayObject $content = null, private bool $required = false)
```
soyuka, line 20: funny enough `content` is Required in the spec: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#request-body-object

src/OpenApi/Model/RequestBody.php
```diff
@@ -17,16 +17,16 @@ final class RequestBody
...
     }
 
-    public function getDescription(): string
+    public function getDescription(): ?string
     {
         return $this->description;
     }
 
-    public function getContent(): \ArrayObject
+    public function getContent(): ?\ArrayObject
```
soyuka, line 29: see my previous comment this was correct but the constructor is not

src/OpenApi/Factory/OpenApiFactory.php
monitaurus: > to understand this correctly the main change is that true is now the default value when there's a request body ?

In the current code, the constructor set `Model\RequestBody`'s `required` to `false` by default. But the `OpenFactory` force the `required` to be `true` if there were no `RequestBody` and if this was a write operation.

Here the PR allow the developer to fine-tune the values as he/she wants.
So it's even possible now to set a `required` to `false` even in a write operation.

src/OpenApi/Model/RequestBody.php
monitaurus: The PR's originated where I wanted to edit the description of an endpoint, but by doing so, I needed to rewrite the whole `content`.

The PR code make the `Model\RequestBody` itself differs from `OpenAPI` own specification.
But the whole `OpenApiFactory` will bring back the compliance while the specification by overwriting the `content` if it's set to `null`.

PR #6375, opened 2024-05-19
fix(validation): ValidationException causes TypeError exception when called with $code=null

KernelMrex:
| Q             | A
| ------------- | ---
| Branch?       | 3.3
| Tickets       | 
| License       | MIT
| Doc PR        | 

After update to 3.3.3, `ValidationException` called in custom validators code may cause issued.

For example: If you previously used `ValidationException` with only 1st string argument, after migration to 3.3.3 backward backward compatibility is broken.

```
Failed asserting that exception of type "TypeError" matches expected exception
"ApiPlatform\Validator\Exception\ValidationException". Message was: "Exception::__construct(): Argument #2 ($code) must be of type int, null given" at
/app/vendor/api-platform/core/src/Validator/Exception/ValidationException.php:84
```


KernelMrex:
May I should add some tests to cover `ValidationException` constructor?

PR #6382, opened 2024-05-21
feat(Serializer) Check in child classes attribute's groups for serialization

antoine-nedelec:
| Q             | A
| ------------- | ---
| Tickets       | Closes #..., closes #... 
| License       | MIT

Original issue created: 
- https://github.com/api-platform/api-platform/issues/2669

Self contained repository for reproduction:
-  https://github.com/antoine-nedelec/api-platform-issue-test

Overall logic: 
- When fetching the groups for a serialization, there might be a group filled only for one child class. and not available in any parent attributes. The entity is then serialized as an IRI, but there could have been specific attributes to be serialized.

Remarks:
- Not sure if this behavior should be limited with a configuration as I did ?
- Name of the attribute to be defined ?


src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -203,6 +204,18 @@ private function getClassSerializerGroups(string $class): array
             $groups[] = $serializerAttributeMetadata->getGroups();
         }
 
+        if (true === $this->normalizeChildClassAttributeGroups) {
+            foreach ($serializerClassMetadata->getReflectionClass()->getAttributes() as $reflectionAttribute) {
+                if (DiscriminatorMap::class !== $reflectionAttribute->getName()) {
```
soyuka, line 209: is discriminatorMap an attribute of the serializer? not sure I understand how this works.

src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php
```diff
@@ -17,6 +17,7 @@
 use ApiPlatform\Metadata\Exception\ResourceClassNotFoundException;
 use ApiPlatform\Metadata\ResourceClassResolverInterface;
 use ApiPlatform\Metadata\Util\ResourceClassInfoTrait;
+use Doctrine\ORM\Mapping\DiscriminatorMap;
```
soyuka, line 20: not a huge fan of using doctrine in this file, it'd be best inside DoctrineOrmPropertyMetadataFactory?

PR #6385, opened 2024-05-23
fix(symfony): make `WriteListener` compatible with `MainController`

ihmels:
While "data" is set in the context in the `MainController`, this was not the case with the `WriteListener`.

I noticed this after updating from v3.2.14 to v3.3.3. Although I had set `event_listeners_backward_compatibility_layer` to `false`, the Symfony events were used again instead of the `MainController`. I had to set `use_symfony_listeners` to `false` so that I could restore the behavior of v3.2.14.

Shouldn't `use_symfony_listeners` be `false` if `event_listeners_backward_compatibility_layer` is also set to `false`?

| Q             | A
| ------------- | ---
| Branch?       | 3.3
| Tickets       | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/EventListener/WriteListener.php
+++ b/src/Symfony/EventListener/WriteListener.php
@@ -110,6 +110,7 @@ final class WriteListener
                 'uri_variables' => $uriVariables,
                 'resource_class' => $operation->getClass(),
                 'previous_data' => false === $operation->canRead() ? null : $request->attributes->get('previous_data'),
+                'data' => false === $operation->canRead() ? null : $request->attributes->get('data'),
             ]);
 
             if ($data) {


src/Symfony/EventListener/WriteListener.php
```diff
@@ -110,6 +110,7 @@ public function onKernelView(ViewEvent $event): void
                 'uri_variables' => $uriVariables,
                 'resource_class' => $operation->getClass(),
                 'previous_data' => false === $operation->canRead() ? null : $request->attributes->get('previous_data'),
+                'data' => false === $operation->canRead() ? null : $request->attributes->get('data'),
```
soyuka, line 113: isn't `data` the same as `$controllerResult`?

soyuka:
> Shouldn't use_symfony_listeners be false if event_listeners_backward_compatibility_layer is also set to false?

Not really as `use_symfony_listeners: false` also breaks some behaviors for example attributes like `read`.

src/Symfony/EventListener/WriteListener.php
ihmels: Yes, when using Symfony events, the `ReadListener` sets `data` in the request attributes. Then the `PayloadArgumentResolver` returns the value of `data`. This value is then passed as first argument to the `PlaceholderAction` which returns that value. This returned value is then set as `ViewEvent::$controllerResult`.

ihmels:
> Not really as `use_symfony_listeners: false` also breaks some behaviors for example attributes like `read`.

But shouldn't the `MainController` be used if `event_listeners_backward_compatibility_layer` is set to `false`?

soyuka:
MainController is never used with listeners, when listeners are enabled we use the PlaceholderAction. Forget the `event_listeners_backward_compatibility_layer` it's not needed anymore, now you can just pick whether:

- `symfony_use_listeners=true` uses PlaceholderAction (or your own controller)
- `symfony_use_listeners=false` uses the MainController

Maybe that I should just unset the `event_listeners_backward_compatibility_layer` flag if `use_symfony_listeners` is set?

PR #6386, opened 2024-05-24
fix(test): canonicalizing json arrays

soyuka:
replaces https://github.com/api-platform/core/pull/5752

@bendavies is that what you had in mind? I'd like to patch this for 3.2 as a bug fix.


bendavies:
LGTM!

PR #6387, opened 2024-05-24
fix(metadata) Added phpdoc to ORM

stephanvierkant:
| Q             | A
| ------------- | ---
| Branch?       |3.3
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

I've added some more phpdoc annotations. Please let me know if I should add more. I've added just some annotations that are relevant for me own use cases.

### DIFF
--- a/src/Doctrine/Orm/Extension/QueryCollectionExtensionInterface.php
+++ b/src/Doctrine/Orm/Extension/QueryCollectionExtensionInterface.php
@@ -25,5 +25,9 @@ use Doctrine\ORM\QueryBuilder;
  */
 interface QueryCollectionExtensionInterface
 {
+    /**
+     * @param class-string         $resourceClass
+     * @param array<string, mixed> $context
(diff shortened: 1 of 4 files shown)


soyuka:
sure but don't forget to run cs-fixer, phpstan and to respect the commit lint rules, thanks.

stephanvierkant:
I'll do! That's why I made it a draft PR, because I didn't run it yet.

stephanvierkant:
@soyuka PHPStan doesn't recognize 'Foo' as a `class-string`. 

I can ignore the error (as false positive), add a `/** @var class-string /*` to it, or use something like 'Foo\Bar' as a test string. What do you want me to do?

stephanvierkant:
Because 'Foo' wasn't recognized as a `class-string`, I've used `Dummy::class` instead.

soyuka:
nice thanks!

PR #6394, opened 2024-05-29
fix(jsonschema): make all required properties optional in PATCH operation with 'json' format

ttskch:
> [!NOTE]
> Is this a new feature? If so, we can close this PR and go to #6395.

| Q             | A
| ------------- | ---
| Branch?       | 3.3
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

In the schema for the request body of `application/merge-patch+json` PATCH operation, all required properties should be output as optional.

> See here: https://datatracker.ietf.org/doc/html/rfc7386#section-3

This PR fixes the SchemaFactory for the `'json'` format to output required properties as optional only for input context of PATCH operations.

For example, for the following entity, the output of the OpenAPI document will change as follows:

```php
#[ApiResource]
class Todo
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column]
    private ?int $id = null;

    #[ORM\Column(length: 255)]
    #[Assert\NotBlank]
    #[Assert\Length(max: 255)]

    #[ORM\Column]
    #[Assert\NotNull]
    private ?bool $done = null;
```

| | Before | After |
| --- | --- | --- |
| **PATCH Operation** | image | image |
| **Schemas** | image |  |

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -18,6 +18,7 @@ use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\CollectionOperationInterface;
 use ApiPlatform\Metadata\HttpOperation;
 use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Patch;
 use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
 use ApiPlatform\Metadata\Property\Factory\PropertyNameCollectionFactoryInterface;
 use ApiPlatform\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface;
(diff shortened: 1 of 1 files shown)


soyuka:
I think that we can make this a bug fix as it won't break any current implementations

soyuka (approved):

soyuka:
Actually we need to revert this as we found it was breaking some implementations but I'll re-introduce this in 3.4.

ttskch:
@soyuka Sorry for this. Of course, after #6476 is merged and 3.3 is merged into 3.4, I can create a new PR for 3.4.

> Actually we need to revert this as we found it was breaking some implementations

But, for reference, could you tell me what the problem was specifically? Or, if there are any related issues, please let me know.

I still think that it is reasonable for required properties to be automatically changed to optional in the schema for JSON Merge Patch.

It is inconvenient for users to have to specify `#[Patch(schema: ['required' => []])]` every time.

soyuka:
Currently some users have required fields on their Patch endpoints. We need a flag to avoid breaking implementations and a way to force keeping required fields. I'll take care of porting this to 3.4.

soyuka:
see for now #6478

PR #6395, opened 2024-05-29
feat(jsonschema): make all required properties optional in PATCH operation with 'json' format

ttskch:
> [!NOTE]
> Is this a bug? If so, we can close this PR and go to #6394.

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

In the schema for the request body of `application/merge-patch+json` PATCH operation, all required properties should be output as optional.

> See here: https://datatracker.ietf.org/doc/html/rfc7386#section-3

This PR fixes the SchemaFactory for the `'json'` format to output required properties as optional only for input context of PATCH operations.

For example, for the following entity, the output of the OpenAPI document will change as follows:

```php
#[ApiResource]
class Todo
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column]
    private ?int $id = null;

    #[ORM\Column(length: 255)]
    #[Assert\NotBlank]
    #[Assert\Length(max: 255)]

    #[ORM\Column]
    #[Assert\NotNull]
    private ?bool $done = null;
```

| | Before | After |
| --- | --- | --- |
| **PATCH Operation** | image | image |
| **Schemas** | image |  |

### DIFF
--- a/src/JsonSchema/SchemaFactory.php
+++ b/src/JsonSchema/SchemaFactory.php
@@ -18,6 +18,7 @@ use ApiPlatform\Metadata\ApiProperty;
 use ApiPlatform\Metadata\CollectionOperationInterface;
 use ApiPlatform\Metadata\HttpOperation;
 use ApiPlatform\Metadata\Operation;
+use ApiPlatform\Metadata\Patch;
 use ApiPlatform\Metadata\Property\Factory\PropertyMetadataFactoryInterface;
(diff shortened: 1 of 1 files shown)


ttskch:
This PR was integrated into #6394.

> refs: https://github.com/api-platform/core/pull/6394#issuecomment-2138917295

PR #6401, opened 2024-05-31
chore(symfony): deprecations

soyuka:


phpstan.neon.dist
```diff
@@ -102,5 +102,3 @@ parameters:
 		- '#Access to undefined constant Symfony\\Component\\HttpKernel\\HttpKernelInterface::MASTER_REQUEST\.#'
 		- '#Attribute class PHPUnit\\Framework\\Attributes\\DataProvider does not exist.#'
 		# Because of a bc-layer in MetadataAwareNameConverter it does not implement AdvancedNameConverterInterface
```
dunglas, line 104: Can be deleted too?

src/Metadata/Util/PropertyInfoToTypeInfoHelper.php
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * A helper about PropertyInfo Type conversion.
+ *
+ * @see https://github.com/mtarld/symfony/commits/backup/chore/deprecate-property-info-type/
+ *
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ *
+ * @internal
+ */
+final class PropertyInfoToTypeInfoHelper
```
dunglas, line 32: I think this is good enough for now, but in the future we should migrate to use the new type class by default and convert it to the legacy type instead of the other way. This will likely improve performance, and will ease the maintenance.

tests/.ignored-deprecations-legacy-events
```diff
@@ -25,3 +25,4 @@
 %ApiPlatform\\Api\\FilterInterface is deprecated in favor of ApiPlatform\\Metadata\\FilterInterface%
 
 %The "Symfony\\Bundle\\MakerBundle\\Maker\\MakeAuthenticator" class is deprecated, use any of the Security\\Make\* commands instead%
+%Since symfony/validator 7.1: Not passing a value for the "requireTld" option to the Url constraint is deprecated. Its default value will change to "true".%
```
dunglas, line 28: nit: we could pass the value explicitly instead?

dunglas (approved):

src/Metadata/Util/PropertyInfoToTypeInfoHelper.php
soyuka: yes but while its experimental this is a good solution

tests/.ignored-deprecations-legacy-events
soyuka: not really:
1. it's only in our tests
2. we need to find a way to know if the option exists
3. the default value is fine for us (as not really used by us)

therefore it won't break when updating

PR #6402, opened 2024-05-31
fix(jsonld): allow @id, @context and @type on denormalization

soyuka:
#6225

### DIFF
--- a/src/JsonLd/Serializer/ItemNormalizer.php
+++ b/src/JsonLd/Serializer/ItemNormalizer.php
@@ -153,4 +153,13 @@ final class ItemNormalizer extends AbstractItemNormalizer
 
         return parent::denormalize($data, $class, $format, $context);
     }
+
+    protected function isAllowedAttribute(object|string $classOrObject, string $attribute, ?string $format = null, array $context = []): bool
+    {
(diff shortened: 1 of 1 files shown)


soyuka:
Hi, the best way to test the branch is to use https://getcomposer.org/doc/05-repositories.md#vcs with the fork and the branch

ili101:
So as I said above https://github.com/api-platform/core/pull/6402#issuecomment-2142494980 this seems not to have any effect.
I don't quite understand Symfony's point of `isAllowedAttribute()` if it's being superseded by `getAllowedAttributes()`.
But this approach I made before works https://github.com/api-platform/core/pull/6274/. So using the same approach but instead of adding a new context option I just did what you did and added the items. branch: https://github.com/ili101/api-platform-core/tree/id (https://github.com/ili101/api-platform-core/commit/7f483074a757a395e378ff5ef8cf963b6e5a5ec1)

And with this it's possible to use the `@id` as [described in the documentation ](https://api-platform.com/docs/core/serialization/#denormalization) while using `allow_extra_attributes: false`

soyuka:
nice, can you open a new PR to replace mine?

soyuka:
superseeded by https://github.com/api-platform/core/pull/6451

PR #6403, opened 2024-05-31
fix(state) handle empty request in read provider

guillaume-sainthillier:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | #6404
| License       | MIT
| Doc PR        | N/A

Hello, this PR let ReadProvider handle nullable request because `RequestParser::getQueryString($request)` expects a request object to work.

We had this bug when using provider in CLI during async export of a GetCollection operation

### DIFF
--- a/src/State/Provider/ReadProvider.php
+++ b/src/State/Provider/ReadProvider.php
@@ -54,7 +54,7 @@ final class ReadProvider implements ProviderInterface
             return null;
         }
 
-        if (null === $filters = $request?->attributes->get('_api_filters')) {
+        if (null === ($filters = $request?->attributes->get('_api_filters')) && $request) {
             $queryString = RequestParser::getQueryString($request);
             $filters = $queryString ? RequestParser::parseRequestParams($queryString) : null;
         }
--- a/src/State/Tests/ReadProviderTest.php
+++ b/src/State/Tests/ReadProviderTest.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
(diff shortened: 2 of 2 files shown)


src/State/Provider/ReadProvider.php
```diff
@@ -54,7 +54,7 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
             return null;
         }
 
-        if (null === $filters = $request?->attributes->get('_api_filters')) {
+        if (null === ($filters = $request?->attributes->get('_api_filters')) && $request) {
```
soyuka, line 57:
```suggestion
        if ($request && null === ($filters = $request->attributes->get('_api_filters'))) {
```

tests/State/Provider/ReadProviderTest.php
```diff
@@ -0,0 +1,46 @@
...
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
```
soyuka, line 12: can you add this to the components test instead? at src/State/Tests

soyuka:
nice catch, wasn't expecting code to be used like this but I see no reason it wouldn't work :)

let me know if you have more issues, commit message is also missing `:` after the fix(state)

guillaume-sainthillier:
> nice catch, wasn't expecting code to be used like this but I see no reason it wouldn't work :)
> 
> let me know if you have more issues, commit message is also missing `:` after the fix(state)

Thank you for quick review!
Indeed we do serialize query filters before turning on async mode, so no request are present :-)

Issue #6407, opened 2024-06-03
Upgrade to Symfony 7.1 seems to be breaking container linting


soyuka:
### Discussed in https://github.com/api-platform/core/discussions/6406

<div type='discussions-op-text'>

<sup>Originally posted by **herndlm** June  3, 2024</sup>
Hi,

not sure yet if this is issue is coming from our app, Symfony or API Platform. I just upgraded from Symfony 7.0 without deprecations to Symfony 7.1. Tests, static analysis and functionality of the app seems fine, but `bin/console lint:container` fails with the following

```
In CheckAliasValidityPass.php line 44:
                                                                                                                                                                                                                                  
  Invalid alias definition: alias "ApiPlatform\Serializer\SerializerContextBuilderInterface" is referencing class "ApiPlatform\OpenApi\Serializer\SerializerContextBuilder" but this class does not implement "ApiPlatform\Seria  
  lizer\SerializerContextBuilderInterface". Because this alias is an interface, "ApiPlatform\OpenApi\Serializer\SerializerContextBuilder" must implement "ApiPlatform\Serializer\SerializerContextBuilderInterface".
```

Any ideas what the issue might be here?</div>

PR #6409, opened 2024-06-04
Lint container

soyuka:
see https://github.com/api-platform/core/issues/6407 

I'm expecting a deprecation issue as this forces me to introduce a deprecated interface back in our tests, let's see.


herndlm:
👍
did happen what you were expecting?

PR #6418, opened 2024-06-06
feat: remove hydra prefix

### LINKED ISSUE #5514: Symfony framework serializer default_context not passed to JsonLd ItemNormalizer

### LINKED ISSUE #6209: How to remove hydra prefix?
Hello everyone,

Is it possible to remove the hydra prefix (hydra:key) so we don't have an invalid character when parsing? (ie. the colon ':')
  
"hydra:member" would be "member"
"hydra:view" would be "view"
etc.

Any idea on how to do that? The serialization doc doesn't seem to cover that part.

### LINKED ISSUE #6415: Remove hydra prefix
**Description**  
`hydra:member` => `member`

+ feature flag to not break

soyuka:
Closes #6209, closes #6415.

Closes #5514.

### DIFF
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ApiPlatform\Hydra\JsonSchema;
 
 use ApiPlatform\JsonLd\ContextBuilder;
+use ApiPlatform\JsonLd\Serializer\HydraPrefixTrait;
 use ApiPlatform\JsonSchema\Schema;
 use ApiPlatform\JsonSchema\SchemaFactoryAwareInterface;
 use ApiPlatform\JsonSchema\SchemaFactoryInterface;
@@ -26,6 +27,7 @@ use ApiPlatform\Metadata\Operation;
  */
 final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
 {
+    use HydraPrefixTrait;
     private const BASE_PROP = [
         'readOnly' => true,
         'type' => 'string',
@@ -57,7 +59,7 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
         ],
     ] + self::BASE_PROPS;
 
-    public function __construct(private readonly SchemaFactoryInterface $schemaFactory)
+    public function __construct(private readonly SchemaFactoryInterface $schemaFactory, private readonly array $defaultContext = [])
     {
         if ($this->schemaFactory instanceof SchemaFactoryAwareInterface) {
             $this->schemaFactory->setSchemaFactory($this);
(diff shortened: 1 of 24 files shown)


src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php
```diff
@@ -231,6 +231,20 @@ private function registerCommonConfiguration(ContainerBuilder $container, array
         // TODO: remove in 4.x
         $container->setParameter('api_platform.event_listeners_backward_compatibility_layer', $config['event_listeners_backward_compatibility_layer']);
 
+        if (null === $config['serializer']['hydra_prefix']) {
+            trigger_deprecation('api-platform/core', '3.3', 'The hydra: prefix will be removed in 4.0 by default, set "api_platform.serializer.defaut_context" to "hydra_prefix: true" to keep the current behavior.');
+        }
```
soyuka, line 236: todo: move this after retrieving the `serializer.default_context`?

src/Hydra/JsonSchema/SchemaFactory.php
```diff
@@ -26,6 +27,9 @@
  */
 final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
 {
+    use HydraPrefixTrait;
+    public const HYDRA_PREFIX = 'hydra:';
+    public const HYDRA_CONTEXT_HAS_PREFIX = 'hydra_prefix';
```
dunglas, line 32: Couldn't we have these constants at only one place for all the component instead of duplicating them in all classes?

src/Hydra/Serializer/HydraPrefixTrait.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Hydra\Serializer;
+
+trait HydraPrefixTrait
+{
+    // Uncomment on API Platform 4
+    // const HYDRA_PREFIX = 'hydra:';
+    // const HYDRA_CONTEXT_HAS_PREFIX = 'hydra_prefix';
```
dunglas, line 20: Ok I see, but even in v4 it's a bit weird to "copy/paste" these constants in each class through a trait instead of having them in a single place.

dunglas (approved):
Great work!! This will simplify a lot usa by developers who don't care about RDF.

src/Hydra/JsonSchema/SchemaFactory.php
soyuka: PHP 8.1 doesn't support constants inside traits...

src/Hydra/Serializer/HydraPrefixTrait.php
soyuka: yes but for the subtree split it's almost mandatory I'll see if I can put  this only in once place

dunglas: Components that haven't a hard dependency on the Hydra component shouldn't depend on this constant (or should use `class_exist()` call).

PR #6422, opened 2024-06-11
fix(doctrine): use null-safe operator when retrieving parameters

Kyzegs:
| Q             | A
| ------------- | ---
| Branch?       | 3.4

The `$operation` argument is nullable, therefore we should always use the null-safe operator when trying to call the `getParameters` method. It may result in the following error otherwise:

```
Call to a member function getParameters() on null
```


soyuka:
thanks rebased at https://github.com/api-platform/core/pull/6423

Issue #6391, opened 2024-05-28
OpenAPI on calculated field add an extra `items` fields even if `openApiContext` define an object


monitaurus:
**API Platform version(s) affected**: 3.2.22, 3.3.2

Tested on 3.2.22 and replicated on 3.3.2

**Description**  
On a calculated field that returns an `array`, if we add a `ApiProperty` to change the type into `object`, the generated OpenAPI does generate a `"type": "object"` but there is also an extra `"items": {"type": "string"}`.

**How to reproduce**  

On a resource, add the following calculated field:
```php
    #[ApiProperty(
        openapiContext: [
            'type' => 'object',
            'properties' => [
                'alpha' => ['type' => 'integer'],
                'beta' => ['type' => 'string'],
            ],
        ]
    )]
    public function getCalculatedField(): array
    {
        return [
            'alpha' => 123,
            'beta' => 'hello',
        ];
    }
```

Then access OpenAPI via Swagger, or generate a openapi json to find the extra `items`.

![swagger_screenshot](https://github.com/api-platform/core/assets/1794571/3f6fa67c-e179-4f2d-b4eb-96c0ba22110d)

**Possible Solution**  

None found at the moment.

**Additional Context**  

Investigation on the factory code:
* the process revolve around `SchemaPropertyMetadataFactory::create`
* the calculated field returns an array, so `$propertyMetadata->getBuiltinTypes()` returns `["array"]` ([view file](https://github.com/api-platform/core/blob/e867d07f59b82d5f1bdca69e096ddf452dd7efc8/src/JsonSchema/Metadata/Property/Factory/SchemaPropertyMetadataFactory.php#L86))
* `array` is considered as a collection of items, and the code define items as `string` by default ([view file](https://github.com/api-platform/core/blob/e867d07f59b82d5f1bdca69e096ddf452dd7efc8/src/JsonSchema/Metadata/Property/Factory/SchemaPropertyMetadataFactory.php#L121))
* and finally `$this->getType` returns a property type as `array` with `items` as `string` ([view file](https://github.com/api-platform/core/blob/e867d07f59b82d5f1bdca69e096ddf452dd7efc8/src/JsonSchema/Metadata/Property/Factory/SchemaPropertyMetadataFactory.php#L133))

The type `object` seems to override the type `array` partially.
[...]

PR #6431, opened 2024-06-15
fix(metadata): wrong schema generated if openapicontext set on array

monitaurus:
| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | Closes https://github.com/api-platform/core/issues/6391
| License       | MIT
| Doc PR        | none

On a calculated field that returns an array, if we add a ApiProperty to change the type into object, the generated OpenAPI does generate a `"type": "object"` but there is also an extra `"items": {"type": "string"}`.


monitaurus:
I don't know if they're needed, but I created new test to cover this bug resolution.

monitaurus:
There's three failing checks, unfortunately I don't know how to fix them, nor if they're linked to this PR.

PR #6433, opened 2024-06-24
feat(doctrine): check property security during search filter

emmanuel-tilleuls:
| Q             | A
| ------------- | ---
| Branch?       | main
| License       | MIT

**TODO**
- [ ] tests
- [ ] add a parameter (where ?) to enable this feature (BC)
- [ ] manage security for nested properties
- [ ] reproduce code in ODM

`SearchFilter` should check property security before adding where condition to the query. In other words : a user should not be able to filter resources by a property he is not allowed to see.

**Use case**
Considering a `Student` resource with following properties :
- `name`
- `anonymizationCode`

Users with «corrector» role don't see `name` property.
A `SearchFilter` can be applied on `name`.

As a corrector I can try to find which student matches each anonymization code by trying : 
`GET /api/students?name=John+Doe`


emmanuel-tilleuls:
This PR is replaced by #6435

PR #6436, opened 2024-06-26
fix(openapi): Yaml openapi export should have numeric keys as string

DartCZ:
###  Description

The OpenAPI specification mandates that all keys in the Responses Object be strings, as outlined here: https://github.com/OAI/OpenAPI-Specification/blob/157a4c81ae537ef793b2bee368bc00d88b461de8/versions/3.0.2.md#patterned-fields-1.

This requirement includes values like '200', which PHP typecasts to integers. Therefore, Yaml should also adhere to this rule.


soyuka:
thanks!

Issue #5934, opened 2023-10-27
Filter Extension broken in 3.2.2


bendavies:
**API Platform version(s) affected**: 3.2.2

**Description**  
ApiFilter annotations have stopped working in 3.2.2.

**How to reproduce**  
Given a. resource:

```php
#[ApiResource(
    shortName: 'MarketingCompany',
    operations: [
        new Get(),
        new GetCollection(),
    ],
    provider: EntityClassDtoStateProvider::class,
    stateOptions: new Options(entityClass: MarketingCompany::class),
)]
#[ApiFilter(SearchFilter::class, properties: [
    'name' => 'ipartial',
])]
class MarketingCompanyApi
{
    #[ApiProperty(readable: false, writable: false, identifier: true)]
    public ?int $id = null;
    public ?string $name;
}
```

Performing a request like:
```
GET /api/marketing-companies?name=foo
```

Will get to here:
https://github.com/api-platform/core/blob/ad2cbe073b1cf0d5589c2ef1e48603e4426286ad/src/Doctrine/Orm/Extension/FilterExtension.php#L54

the `$filterId` is:
`annotated_bind_hq_api_api_resource_marketing_company_api_api_platform_doctrine_orm_filter_search_filter`
but the service name in `ContainerInterface $filterLocator` is:
`annotated_bind_h_q_api_api_resource_marketing_company_api_api_platform_doctrine_orm_filter_search_filter`

the difference is: 
`annotated_bind_hq_` vs
`annotated_bind_h_q_`

**Possible Solution**  
I'm guessing this is down to https://github.com/api-platform/core/pull/5637

**Additional Context**

soyuka:
Thanks for the report. Does `keep_legacy_inflector: true` solves the issue for now?

bendavies:
> Thanks for the report. Does `keep_legacy_inflector: true` solves the issue for now?

yes, it does. 👍

jotwea:
I am currently facing the same issue. Setting `keep_legacy_inflector: true` worked for me for the moment.

gseric:
I think the issue is in the approach used for new Inflector implementation. It's singleton instead of proper service, so it's not configured properly at run time.

AP bundle extension configures it here: https://github.com/api-platform/core/blob/6d15e228611875a3c1a41a14ecf26a38bd67639a/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L986
[...]

PR #6438, opened 2024-06-27
chore: remove deprecated code 

soyuka:


docs/config/packages/framework.yaml
```diff
@@ -7,13 +7,11 @@ api_platform:
     json: ['application/json']
   docs_formats:
     jsonopenapi: ['application/vnd.openapi+json']
-  keep_legacy_inflector: false
```
bendavies, line 10: there still exists a bug which this flag is needed to workaround:
https://github.com/api-platform/core/issues/5934

shall we fix this first, before removing the workaround?

soyuka: https://github.com/api-platform/core/pull/6447 should fix this no?

bendavies: Thanks. Will give it a go

PR #6440, opened 2024-06-28
fix(openapi): parameters can disable openapi

soyuka:


src/OpenApi/Model/PathItem.php
```diff
@@ -184,7 +184,7 @@ public function withServers(?array $servers = null): self
         return $clone;
     }
 
-    public function withParameters(array $parameters): self
+    public function withParameters(?array $parameters = null): self
```
bendavies, line 187: @soyuka you need to change the return type of `getParameters()` to include `null`

soyuka: damn I wish I could have phpstan level7 thanks!

bendavies: Add it! Baseline everything. Fix old code slowly

soyuka: there's a baseline generator? I think we're close to level 7 anyways, it's just the "document array type" rule that is missing in a lot of files.

bendavies: `--generate-baseline` flag!

PR #6447, opened 2024-07-02
feat: inflector as service

soyuka:
Inflector as service


paullallier:
@soyuka:  Not sure if this change requires something in the config files too?  Running 3.4.0-alpha.1 I'm getting a null rather than an inflector in DashPathSegmentNameGenerator

I can't find any note about it or a new recipe. I've tried clearing the cache and dumpautoload.

I can raise an issue if you like.

paullallier:
@soyuka: Sorry - I must not have cleared the cache correctly. It's working now.

PR #6450, opened 2024-07-04
feat: deprecate query parameter validator

soyuka:
The idea here is that we deprecate the QueryParameterValidator in 3.4 in favor of using the new parameter attributes (QueryParameter and HeaderParameter). By doing so, we introduce a backward compatibility flag: 

```
defaults: 
    extra_properties:
        use_legacy_parameter_validator: true
```

This flag is `true` by default in 3.4 and gets removed in 4.0. When using `false` we add parameters for each declared filter and validation will still be effective (according to `$queryParameterValidationEnabled`). A few changes occur though (hence the flag): 

- status code is 422 instead of 400
- there's no more `collectionFormat` support from swagger, indeed this completely changed in OpenAPI 3.1 and we did not yet add the support of `style` validation on parameters (https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#style-values) 
- violation messages are now the ones from Symfony but basically the same validation occurs 

Left tbd: 

- [x] add the `queryParameterValidationEnabled` everywhere (as it also applies to graphql `args`)
- [x] add a deprecation when the `use_legacy_parameter_validator` is not `false`
- [ ] fix tests


src/State/Util/ParameterParserTrait.php
```diff
@@ -23,31 +24,54 @@
...
     {
-        if ($request) {
-            return ($parameter instanceof HeaderParameterInterface ? $request->attributes->get('_api_header_parameters') : $request->attributes->get('_api_query_parameters')) ?? [];
+        $accessors = null;
+        $key = $parameter->getKey();
+        $parsedKey = explode('[:property]', $key);
+        if (isset($parsedKey[0]) && isset($values[$parsedKey[0]])) {
+            $key = $parsedKey[0];
+        } elseif (str_contains($key, '[')) {
+            preg_match_all('/[\w\-_]+/', $key, $matches);
```
deguif, line 53:
```suggestion
            preg_match_all('/[^\[\]]+/', $key, $matches);
```

What do you think about using negated character class, as this would be more bullet proof on complex string (special characters, emojis, ...)
Example which would probably be failing currently: `test😀[a]`, `test.test[a]`, `test@test[a]` ...

PR #6451, opened 2024-07-04
fix(jsonld): allow @id, @context and @type on denormalization 2

ili101:
| Q             | A
| ------------- | ---
| Branch?       | 3.3  
| Tickets       | Closes https://github.com/api-platform/core/issues/6225 
| License       | MIT
| Doc PR        | Feature already documented

Allaw `@id`, `@type`, `@context` with JsonLd. With this it's possible to use the `@id` as [described in the documentation ](https://api-platform.com/docs/core/serialization/#denormalization) while using `allow_extra_attributes: false`.
More info in https://github.com/api-platform/core/pull/6402.

### DIFF
--- a/features/main/standard_put.feature
+++ b/features/main/standard_put.feature
@@ -66,6 +66,20 @@ Feature: Spec-compliant PUT support
     """
     Then the response status code should be 400
 
+  Scenario: Fails to create a new resource when the JSON-LD @id doesn't match the URI
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/standard_puts/7" with body:
+    """
+    {
+      "@id": "/standard_puts/6",
+      "@context": "/contexts/StandardPut",
+      "@type": "StandardPut",
+      "foo": "a",
+      "bar": "b"
+    }
+    """
+    Then the response status code should be 400
+
   Scenario: Replace an existing resource
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "PUT" request to "/standard_puts/5" with body:
--- a/features/main/standard_put.feature
+++ b/features/main/standard_put.feature
@@ -52,6 +52,20 @@ Feature: Spec-compliant PUT support
     }
     """
 
+  Scenario: Fails to create a new resource with the wrong JSON-LD @id
+    When I add "Content-Type" header equal to "application/ld+json"
+    And I send a "PUT" request to "/standard_puts/7" with body:
+    """
+    {
+      "@id": "/dummies/6",
+      "@context": "/contexts/StandardPut",
+      "@type": "StandardPut",
+      "foo": "a",
+      "bar": "b"
+    }
+    """
+    Then the response status code should be 400
+
   Scenario: Replace an existing resource
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "PUT" request to "/standard_puts/5" with body:
--- a/features/main/standard_put.feature
+++ b/features/main/standard_put.feature
@@ -26,7 +26,7 @@ Feature: Spec-compliant PUT support
     }
     """
 
-  Scenario: Create a new resource with json-ld attributes
+  Scenario: Create a new resource with JSON-LD attributes
     When I add "Content-Type" header equal to "application/ld+json"
     And I send a "PUT" request to "/standard_puts/6" with body:
     """
--- a/features/main/standard_put.feature
+++ b/features/main/standard_put.feature
@@ -26,6 +26,32 @@ Feature: Spec-compliant PUT support
     }
     """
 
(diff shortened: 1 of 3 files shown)


soyuka:
this is perfect, would you be able to add a test?

soyuka (approved):

features/main/standard_put.feature
```diff
@@ -26,6 +26,32 @@ Feature: Spec-compliant PUT support
     }
     """
 
+  Scenario: Create a new resource with json-ld attributes
```
dunglas, line 29:
```suggestion
  Scenario: Create a new resource with JSON-LD attributes
```

nit

features/main/standard_put.feature
```diff
@@ -26,6 +26,32 @@ Feature: Spec-compliant PUT support
...
+    {
+      "@context": "/contexts/StandardPut",
+      "@id": "/standard_puts/6",
+      "@type": "StandardPut",
+      "id": 6,
+      "foo": "a",
+      "bar": "b"
+    }
+    """
+
```
dunglas, line 54: Could you also add a test that ensures an error is thrown if the value of the `@id` key doesn't match the URL?

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -148,9 +150,26 @@ public function denormalize(mixed $data, string $class, ?string $format = null,
...
         }
 
         return parent::denormalize($data, $class, $format, $context);
     }
+
+    protected function getAllowedAttributes(string|object $classOrObject, array $context, bool $attributesAsString = false): array|bool
+    {
+        $allowedAttributes = parent::getAllowedAttributes($classOrObject, $context, $attributesAsString);
+        if (\is_array($allowedAttributes) && ($context['api_denormalize'] ?? false)) {
+            $allowedAttributes = array_merge($allowedAttributes, ['@id', '@type', '@context']);
```
dunglas, line 170: Maybe should we allow all special keys (`@set`, `@container`, `@version`, `@nest` etc)?

ili101:
> this is perfect, would you be able to add a test?

I was not available but I see a test was already added.
If it's useful in any way here is a [PHPUnit test](https://github.com/ili101/api-platform/blob/TestId/api/tests/Api/MainsTest.php) ([commit](https://github.com/ili101/api-platform/commit/c19da417399189b1524c29d380715472401fb8ea)) of the full usecase of `writableLink` using `@id` while `allow_extra_attributes: false`.

Side note I see you were discussing:
> Could you also add a test that ensures an error is thrown if the value of the `@id` key doesn't match the URL?

If the [`@id`](https://github.com/ili101/api-platform/blob/c19da417399189b1524c29d380715472401fb8ea/api/tests/Api/MainsTest.php#L41) line on the "Main" object is removed here **or** id is different from the URL, the "subs" array embedded objects will behave unexpectedly (removed or throw a duplicate error). Not sure if a problem or just needs to be documented as the description in the documentation only mentions the need/behavior for the `@id` on the sub embedded objects and not their parent too.

src/JsonLd/Serializer/ItemNormalizer.php
soyuka: done I added all the keywords

src/JsonLd/Serializer/ItemNormalizer.php
```diff
@@ -47,6 +49,29 @@ final class ItemNormalizer extends AbstractItemNormalizer
     use JsonLdContextTrait;
 
     public const FORMAT = 'jsonld';
+    public const JSONLD_KEYWORDS = [
```
dunglas, line 52: This could maybe be an `@internal` enum?

soyuka: too much trouble transforming an enum to an array this is more efficient imo, I could set `private` visibility?

src/Symfony/Routing/IriConverter.php
```diff
@@ -78,6 +78,16 @@ public function getResourceFromIri(string $iri, array $context = [], ?Operation
             throw new InvalidArgumentException(sprintf('No resource associated to "%s".', $iri));
         }
 
+        foreach ($context['uri_variables'] ?? [] as $key => $value) {
+            if (!isset($parameters[$key]) || $parameters[$key] !== (string) $value) {
+                throw new InvalidArgumentException(sprintf('The iri "%s" does not reference the correct resource.', $iri));
+            }
+        }
+
+        if ($operation && $operation->getClass() !== $parameters['_api_resource_class']) {
```
soyuka, line 87: use is_a to make this work with abstract classes

dunglas (approved):

soyuka:
thanks @ili101 !

PR #6477, opened 2024-07-18
fix: remove hydra:next when no items would be returned in the next cursor pagination page

nesl247:
When using cursor pagination, `hydra:next` is currently always set even when it wouldn't return any results.

### DIFF
--- a/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
+++ b/src/Hydra/Serializer/PartialCollectionViewNormalizer.php
@@ -90,6 +90,12 @@ final class PartialCollectionViewNormalizer implements NormalizerInterface, Norm
         if ($isPaginatedWithCursor) {
             $data[$hydraPrefix.'view'] = ['@id' => null, '@type' => $hydraPrefix.'PartialCollectionView'];
 
+            $itemsPerPage = $pageTotalItems = null;
+            if ($object instanceof PartialPaginatorInterface) {
+                $itemsPerPage = $object->getItemsPerPage();
+                $pageTotalItems = (float) \count($object);
+            }
+
             return $this->populateDataWithCursorBasedPagination($data, $parsed, $object, $cursorPaginationAttribute, $operation?->getUrlGenerationStrategy() ?? $this->urlGenerationStrategy, $hydraPrefix, $itemsPerPage, $pageTotalItems);
         }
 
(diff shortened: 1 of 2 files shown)


src/Hydra/Serializer/PartialCollectionViewNormalizer.php
```diff
@@ -168,15 +168,15 @@ private function cursorPaginationFields(array $fields, int $direction, $object):
...
+    private function populateDataWithCursorBasedPagination(array $data, array $parsed, \Traversable $object, ?array $cursorPaginationAttribute, ?int $urlGenerationStrategy, ?float $itemsPerPage, ?float $pageTotalItems): array
     {
         $objects = iterator_to_array($object);
         $firstObject = current($objects);
         $lastObject = end($objects);
 
         $data['hydra:view']['@id'] = IriHelper::createIri($parsed['parts'], $parsed['parameters'], urlGenerationStrategy: $urlGenerationStrategy);
 
-        if (false !== $lastObject && \is_array($cursorPaginationAttribute)) {
+        if (false !== $lastObject && \is_array($cursorPaginationAttribute) && $pageTotalItems >= $itemsPerPage) {
```
soyuka, line 179: when `$pageTotalItems == $itemsPerPage` we want to show hydra:view? Can you test this? I thought that with a cursor we did not know the amount of total items (no count)?

nesl247: We did test it out, we have it live in our production environment. This isn't perfect because it still could return no results on the next page, however, it prevents an empty page from showing hydra:next.

The logic is that with cursor pagination if you no longer have results on a page, the next one won't either because you are doing a less less than or greater than some identifier. Because it's only a comparison like that if it doesn't return enough results we can be certain the next page will return none.

soyuka: would you be able to add a test for this?

nesl247: Sure.

nesl247: Sorry I haven't gotten around to adding a test here. It isn't forgotten though.

nesl247:
It still needs to be resolved. I've not had time to add the proper tests to the PR unfortunately.

VincentLanglet:
Hi, are you still interested by this PR ? Do you have time to resolve the conflict/comment ?
Thanks

soyuka:
I'm not a huge fan of this as it adds more computation then needed. I think that using JSON-LD users would just dereference and see an empty page which is probably fine.

Issue #6483, opened 2024-07-19
3.3.10 [BC break?]: Entities with embedded non-resource subentities fail schema checks after #6366


paullallier:
**API Platform version(s) affected**: 3.3.10, specifically #6366 I think

**Description**  
Upgrading to 3.3.10 breaks schema checks on entities with embedded subentities (especially one with genId: false set).

**How to reproduce**  
I have a lot of instances where I'm using `#[ApiProperty(genId: false)]` to include an array of subentities inside another entity - where I don't want the user to be able to extract the subentities on there own (and there's no endpoint to do so).  For example:

```
class Tax
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column]
    #[Groups(['tax:read', 'tax:id_only', 'account:read_collection', 'account:read_item'])]
    private ?int $id = null; // Returns value from numberId on the API

    #[ORM\Column(length: 128)]
    #[Groups(['tax:read', 'tax:write', 'account:read_collection', 'account:read_item'])]
    private ?string $description = null;

<snip>

    #[ORM\OneToMany(mappedBy: 'componentOf', targetEntity: TaxComponent::class)]
    #[Groups(['tax:read', 'tax:write'])]
    #[ApiProperty(genId: false)]
    protected ?Collection $components;
```

Which give a failure like this when I 'GET' an entity and check it with `self::assertMatchesResourceItemJsonSchema()`:
```
Failed asserting that Array &0 (
    '@context' => '/contexts/Tax'
    '@id' => '/taxes/1622'
    '@type' => 'Tax'
    'id' => 1622
    'description' => 'Consectetur hic optio.'
    'components' => Array &2 (
        0 => Array &3 (
            '@type' => 'TaxComponent'
            'name' => 'Fugit quam sed atque.'
            'ratePercent' => '43.2000'
            'priority' => 7
            'account' => '/accounts/113'
        )
    )
) matches the provided JSON Schema.
components[0].@context: The property @context is required
components[0].@id: The property @id is required
```

Question 1:
While I don't doubt the this statement from the PR is correct:
_"This PR modifies SchemaFactory for JSON-LD to output JSON-LD specific properties such as @context @id @type as REQUIRED."_
Is it compatible with `#[ApiProperty(genId: false)]`?

I can't get the test harness to run for me at the moment, so I can't easily test it, but would the tests involving this entity pass the requirement to have @id and @context on the "notResourceObjects" embedded it it, if those test checked the schema more completely?
[...]

PR #6484, opened 2024-07-20
Revert "feat(jsonschema): make JSON-LD specific properties required"

soyuka:
Reverts api-platform/core#6366


soyuka:
It looks like making these properties required is too much of a change (see #6483) and it also looks like renaming schemas also brings other issues (@vincentchalamon will probably add some cases where it breaks for its usage).

@ttskch would you be able to re-open a PR against 3.4 with this as we definitely should add this? In a perfect world this would've not break but we have too many use cases on API Platform :/.

PR #6485, opened 2024-07-20
feat(jsonschema): make JSON-LD specific properties required only in the schema for output

ttskch:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Tickets       | #6484
| License       | MIT
| Doc PR        | N/A

In JSON Schema for JSON-LD, following things are the truth:

* JSON-LD specific properties such as `@context` `@id` `@type` can be included in the request body. (See #6451)
* JSON-LD specific properties such as `@context` `@id` `@type` always have a value in the response body. (In other words, they are required.)

Therefore, this PR will make the following fixes:

* Make JSON-LD specific properties such as `@context` `@id` `@type` non-readOnly.
* Separate the JSON-LD schema for input and output.
* Only in the schema for output, make JSON-LD specific properties such as `@context` `@id` `@type` required.

Below is a simple example.

```php
#[ApiResource]
class Todo
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column]
    private ?int $id = null;

    #[ORM\Column(length: 255)]
    #[Assert\NotBlank]
    #[Assert\Length(max: 255)]

    #[ORM\Column]
    #[Assert\NotNull]
    private ?bool $done = null;
```

Currently, the following schema is generated for the above API resource.

image

This schema is clearly incorrect because:

* JSON-LD specific properties such as `@context` `@id` `@type` can be included in the request body. (See #6451) **So the must not be readOnly.**
* JSON-LD specific properties such as `@context` `@id` `@type` always have a value in the response body. (**In other words, they are required in the output context.**)

JSON-LD specific properties such as `@context` `@id` `@type` **should be output as required in the OpenAPI document**. However, since the above schema is **shared by both the request body and response body**, if we set these properties to required, they will also be required in the request body.

The fundamental problem is that the above schema is **shared by both the request body and response body**. In the first place, in the current implementation, even though [`ApiPlatform\Hydra\JsonSchema\SchemaFactory` does not add JSON-LD specific properties in the `input` context](https://github.com/api-platform/core/blob/953ec19ed2249cc4a0f67f14174f7fcc9302db39/src/Hydra/JsonSchema/SchemaFactory.php#L77-L79), the schemas for input and output are **generated with the same key**, so the same schema is shared between the input and output contexts. This is clearly a bug.

> See:
> * https://github.com/api-platform/core/blob/953ec19ed2249cc4a0f67f14174f7fcc9302db39/src/OpenApi/Factory/OpenApiFactory.php#L265-L267
> * https://github.com/api-platform/core/blob/953ec19ed2249cc4a0f67f14174f7fcc9302db39/src/OpenApi/Factory/OpenApiFactory.php#L415-L417

Therefore, in this PR, I first modified the schemas for input and output to be generated with different keys.

And I also made JSON-LD specific properties such as `@context` `@id` `@type` non-readOnly (because they can be included in the request body) and made those properties required only in the schema for output (if those properties exist).

This ensures that JSON-LD specific properties are not required in the request body, and those properties are marked as required in the response body.

image

This has the advantage that the type information of API clients automatically generated by [OpenAPI TypeScript](https://openapi-ts.pages.dev/) etc. will be more precise.

```diff
  export interface operations {
      api_todos_post: {
          requestBody: {
-             "application/ld+json": components["schemas"]["Todo.jsonld"];
+             "application/ld+json": components["schemas"]["Todo.jsonld.input"];
          };
          responses: {
              201: {
                  content: {
-                     "application/ld+json": compoennts["schemas"]["Todo.jsonld"];
+                     "application/ld+json": compoennts["schemas"]["Todo.jsonld.output"];
                  };
              };
          };
      };
  };
  
  export interface components {
      schemas: {
+         "Todo.jsonld.input": {
+             "@context"?: string | {
+                 "@vocab": string;
+                 /** @enum {string} */
+                 hydra: "http://www.w3.org/ns/hydra/core#";
+                 [key: string]: unknown;
+             };
+             "@id"?: string;
+             "@type"?: string;
+             readonly id?: number;
+             title: string;
+             description?: string | null;
+             done: boolean;
+         };
-         "Todo.jsonld": {
+         "Todo.jsonld.output": {
-             readonly "@context"?: string | {
+             "@context": string | {
                  "@vocab": string;
                  /** @enum {string} */
                  hydra: "http://www.w3.org/ns/hydra/core#";
                  [key: string]: unknown;
              };
-             readonly "@id"?: string;
+             "@id": string;
-             readonly "@type"?: string;
+             "@type": string;
              readonly id?: number;
              title: string;
              description?: string | null;
              done: boolean;
          };
      };
  };
```

### DIFF
--- a/features/openapi/docs.feature
+++ b/features/openapi/docs.feature
@@ -153,7 +153,7 @@ Feature: Documentation support
     And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.parameters" should have 6 elements
 
     # Subcollection - check schema
-    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.responses.200.content.application/ld+json.schema.properties.hydra:member.items.$ref" should be equal to "#/components/schemas/RelatedToDummyFriend.jsonld-fakemanytomany"
+    And the JSON node "paths./related_dummies/{id}/related_to_dummy_friends.get.responses.200.content.application/ld+json.schema.properties.hydra:member.items.$ref" should be equal to "#/components/schemas/RelatedToDummyFriend.jsonld-fakemanytomany.output"
 
     # Deprecations
     And the JSON node "paths./dummies.get.deprecated" should be false
@@ -165,8 +165,8 @@ Feature: Documentation support
     And the JSON node "paths./deprecated_resources/{id}.patch.deprecated" should be true
 
     # Formats
-    And the OpenAPI class "Dummy.jsonld" exists
-    And the "@id" property exists for the OpenAPI class "Dummy.jsonld"
+    And the OpenAPI class "Dummy.jsonld.output" exists
+    And the "@id" property exists for the OpenAPI class "Dummy.jsonld.output"
     And the JSON node "paths./dummies.get.responses.200.content.application/ld+json" should be equal to:
     """
     {
@@ -176,7 +176,7 @@ Feature: Documentation support
                 "hydra:member": {
                     "type": "array",
                     "items": {
-                        "$ref": "#/components/schemas/Dummy.jsonld"
+                        "$ref": "#/components/schemas/Dummy.jsonld.output"
                     }
                 },
                 "hydra:totalItems": {
--- a/src/Hydra/JsonSchema/SchemaFactory.php
+++ b/src/Hydra/JsonSchema/SchemaFactory.php
@@ -27,7 +27,6 @@ use ApiPlatform\Metadata\Operation;
 final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareInterface
 {
     private const BASE_PROP = [
-        'readOnly' => true,
         'type' => 'string',
     ];
     private const BASE_PROPS = [
@@ -36,7 +35,6 @@ final class SchemaFactory implements SchemaFactoryInterface, SchemaFactoryAwareI
     ];
     private const BASE_ROOT_PROPS = [
         '@context' => [
-            'readOnly' => true,
             'oneOf' => [
                 ['type' => 'string'],
                 [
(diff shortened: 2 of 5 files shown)


ttskch:
@soyuka /cc @paullallier We still need to discuss the following points:

* Which of `@context`, `@id` and `@type` are required and which are not required in the output schema
* About compatibility with `#[ApiProperty(genId: false)]` (See #6483)

paullallier:
Thanks @ttskch.  

I'm not going to claim to know much about the strict requirements of the OpenApi spec and I'm primarily interested in having a functional API for my project.  I suspect the previous iteration of the fix actually didn't break my API itself, just the `assertMatchesResourceItemJsonSchema` assertion - though I'd want that still to work since it's important to me that my CI tests pass (and I don't really want to remove the check entirely).

That's not to say we shouldn't bother trying to be consistent with the OpenApi spec.


I think you're likely to be correct that, generally, for an entity with an embedded array of subresources (say a "Course" entity was an array of "Student" entities enrolled on that course) - the subresources should have an @id, because you can get the details of a single student.  Probably should have an @context too, though I don't know whether it MUST.

My use case is different from this though.  For example, I have an "Invoice" entity with an array of "InvoiceLine" entities embedded into it.  In that case, it's not desirable to me that someone should be able to GET one of the InvoiceLine entities in  isolation from the Invoice and I haven't made an endpoint to get them.  I'm using `genId: false` to suppress generation of a skolem IRI.  I haven't looked at the history, but I suspect that's probably why that genId parameter exists.   In this case, the subresource isn't really a resource, even though - internally - it's an entity object.


A side note, I did also get some failures when checking the schema on errors (which might have been a mistake on my part anyway - but it was passing before). We should potentially check that the problem objects are correctly including @context if that's really needed.

ttskch:
Thank you for the detailed explanation @paullallier. /cc @soyuka

I've created a simple reproduction environment for this issue below.

https://github.com/ttskch/api-platform-core-6485

* There are two entities of [`Todo`](https://github.com/ttskch/api-platform-core-6485/blob/main/src/Entity/Todo.php) "resource" entity and [`Subtask`](https://github.com/ttskch/api-platform-core-6485/blob/main/src/Entity/Subtask.php) "non-resource" entity.
* `Todo` has a collection of `Subtask`

In this situation, I created [a simple test](https://github.com/ttskch/api-platform-core-6485/blob/main/tests/Functional/Api/TodoTest.php) using `assertMatchesResourceItemJsonSchema()` method.

With a42ca67, [this test fails](https://github.com/ttskch/api-platform-core-6485/actions/runs/10029521035/job/27717787464) as you say.

But after applying the fix in 23af961, [this test passes](https://github.com/ttskch/api-platform-core-6485/actions/runs/10029555494/job/27717866829).

From my understanding, in the schema for sub-resources, `@context` is always absent, and `@id` is removed when `genId: false` is set, so only `@type` is required for sub-resources.

In `SchemaFactory::buildSchema()`, [whether the current schema is a sub-schema can be determined by whether the 4th argument is null](https://github.com/api-platform/core/blob/953ec19ed2249cc4a0f67f14174f7fcc9302db39/src/JsonSchema/SchemaFactory.php#L230), so I think this is a reasonable fix.

What do you guys think? Thanks.

soyuka:
Thanks for your work @ttskch! I was on holiday and I'll take time to see how everything fits here, I'll come back to that asap.

soyuka:
So, the main issue here is that we multiply the JSON Schemas for a given resource which is really heavy work for no much gain. I don't have much time (as 3.4/4.0 release is planned for September I have a lot of work right now), but we need to read the JSON schema specification (https://json-schema.org/draft/2020-12/json-schema-core) as I'm sure that there's a way to reference the JSON-LD specific fields. If there's none, I think that it's acceptable to have optional properties for special fields (@id, @type, @context etc.) eventhough they're required on output. WDYT @vincentchalamon ?

soyuka:
We need to use `$id` to have base schemas that each resource schema would extend: https://json-schema.org/draft/2020-12/json-schema-core#id-keyword. This is not so easy to implement but I'll willing to give a shot (after the API Platform con is done ^^).

soyuka:
should be fixed in 4.2

PR #6487, opened 2024-07-21
perf: use XXH3 instead of MD5

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes n/a
| License       | MIT
| Doc PR        | n/a

XXH3 is more than 15 times faster than MD5 (https://xxhash.com/).
Symfony made a similar change: https://github.com/symfony/symfony/pull/47094


theofidry:
Very interesting!

PR #6519, opened 2024-08-18
feat(dx): additional log to help locate a potential missing provider

### LINKED ISSUE #6335: `ProviderNotFoundException` overwrote by misleading `NotFoundHttpException`
**API Platform version(s) affected**: < 3.2.21

**Description**  
When setting `GET` operations, they (may?) require a provider.

When a provider is not found, the exception `ProviderNotFoundException` is caught and the variable `$data` is set to `null`.

Then there is a check that, when the operation is, among the others, a `GET`, throws a `NotFoundHttpException`.

This is completely misleading, leading the developer to do a whole series of tests to understand why things are not working as expected.

To make things even worse, for example, running `bin/console debug:router`, correctly shows the route with the correct operation.

So, a full round of debug, breakpoints, very deep trees of decorators and many more scary things, starts... with countless hours spent going around and around in the code, banging the head against the monitor.... "Why the f*** aren't you working!?🤬"...

In the end, I came to this code:

https://github.com/api-platform/core/blob/6d15e228611875a3c1a41a14ecf26a38bd67639a/src/Symfony/EventListener/ReadListener.php#L130-L145

(it seems to be updated, but the substance didn't change)

In practice, this code overwrites the informative `ProviderNotFoundException` with a very misleading `NotFoundHttpException`.

They also return two very different codes: a misleading `404` versus an informative `500`.

**How to reproduce**  

A controller:

```php
#[AsController]
#[Route(path: SearchRequest::API_ENDPOINT, name: SearchRequest::API_ROUTE_NAME, methods: [Request::METHOD_GET])]
final readonly class SearchController
{
    public function __invoke(
        #[MapQueryString]
        SearchRequest $searchRequest,
        // Account $account,
    ): JsonResponse {
        ...
    }
}
```

A request object set as an ApiPlatform resource:

```php
#[Get(
    /**
     * Documented with a decorator.
     *
     * @see SearchOpenApiGet
     */
    controller: SearchController::class,
    uriTemplate: self::API_ENDPOINT,
)]
class SearchRequest
{
...
}
```

**Possible Solution**  
I think the code should, at least, be changed like this:

```diff
            $data = $this->provider->provide($operation, $uriVariables, $context);
        } catch (InvalidIdentifierException|InvalidUriVariableException $e) {
            throw new NotFoundHttpException('Invalid identifier value or configuration.', $e);
        } catch (ProviderNotFoundException $e) {
            $data = null;
        }

        if (
            null === $data
            && 'POST' !== $operation->getMethod()
            && (
                'PUT' !== $operation->getMethod()
                || ($operation instanceof Put && !($operation->getAllowCreate() ?? false))
            )
        ) {
-            throw new NotFoundHttpException('Not Found');
+            isset($e) ? throw $e : throw new NotFoundHttpException('Not Found');
        }
```

This way, if the problem is a caught `ProviderNotFoundException`, it is not overwrote but, instead, is thrown and the developer knows (s)he need a `StateProvider` to make the route work as expected.

But, maybe this is not the best solution at all: in the end, which is the purpose of throwing a generic `NotFoundHttpException`? The "not found" refers to the result of the found state provider (the state provider didn't find the data it has to return) or the throwing of the `ProviderNotFoundException`.

In the second case, in my opinion, it's better to re-throw the original `ProviderNotFoundException`: this informs the developer (s)he has to implement one.

In the first case, instead, may be better to throw a more informative exception: "The data provider returned no value" or something similar.

In both cases, the problem is with the state provider, not with the route: the route exists: is the state provider that is failing to do its job (or isn't configured at all).

Putting the accent on the state provider, helps the developer look in the right place, not looking for, for example, the configuration of the route.

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | Closes #6335
| License       | MIT

Just adding a simple log to avoid completely hiding the fact that there was a ProviderNotFoundException behind a NotFoundHttpException. 
It could help save time for devs who simply forgot to implement the provider.

### DIFF
--- a/src/State/Provider/ReadProvider.php
+++ b/src/State/Provider/ReadProvider.php
@@ -24,6 +24,7 @@ use ApiPlatform\State\SerializerContextBuilderInterface;
 use ApiPlatform\State\UriVariablesResolverTrait;
 use ApiPlatform\State\Util\OperationRequestInitiatorTrait;
 use ApiPlatform\State\Util\RequestParser;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
@@ -40,6 +41,7 @@ final class ReadProvider implements ProviderInterface
     public function __construct(
         private readonly ProviderInterface $provider,
         private readonly LegacySerializerContextBuilderInterface|SerializerContextBuilderInterface|null $serializerContextBuilder = null,
+        private readonly ?LoggerInterface $logger = null,
     ) {
     }
 
@@ -63,10 +65,12 @@ final class ReadProvider implements ProviderInterface
             $context['filters'] = $filters;
         }
 
+        $resourceClass = $operation->getClass();
+
         if ($this->serializerContextBuilder && $request) {
             // Builtin data providers are able to use the serialization context to automatically add join clauses
             $context += $normalizationContext = $this->serializerContextBuilder->createFromRequest($request, true, [
-                'resource_class' => $operation->getClass(),
+                'resource_class' => $resourceClass,
                 'operation' => $operation,
             ]);
             $request->attributes->set('_api_normalization_context', $normalizationContext);
@@ -75,6 +79,8 @@ final class ReadProvider implements ProviderInterface
         try {
             $data = $this->provider->provide($operation, $uriVariables, $context);
         } catch (ProviderNotFoundException $e) {
+            // In case the dev just forgot to implement it
+            $this->logger?->debug('No provider registered for {resource_class}', ['resource_class' => $resourceClass]);
             $data = null;
         }
 
@@ -85,7 +91,7 @@ final class ReadProvider implements ProviderInterface
                 || ($operation instanceof Put && !($operation->getAllowCreate() ?? false))
             )
         ) {
-            throw new NotFoundHttpException('Not Found');
+            throw new NotFoundHttpException('Not Found', $e ?? null);
         }
 
         $request?->attributes->set('data', $data);
--- a/src/Symfony/Bundle/Resources/config/symfony/events.xml
+++ b/src/Symfony/Bundle/Resources/config/symfony/events.xml
@@ -21,6 +21,7 @@
         <service id="api_platform.state_provider.read" class="ApiPlatform\State\Provider\ReadProvider">
             <argument type="service" id="api_platform.state_provider.locator" />
             <argument type="service" id="api_platform.serializer.context_builder" />
+            <argument key="$logger" type="service" id="logger" on-invalid="null" />
         </service>
 
         <!-- kernel.request priority must be < 8 to be executed after the Firewall -->


mrossard:
Am I missing something? I only added 3 lines to a file, what's the conflict here? :D

dunglas:
@mrossard the conflict is in `use`. You need to rebase.

src/Symfony/EventListener/ReadListener.php
```diff
@@ -50,6 +51,7 @@ public function __construct(
         ?ResourceMetadataCollectionFactoryInterface $resourceMetadataCollectionFactory = null,
         private readonly ?SerializerContextBuilderInterface $serializerContextBuilder = null,
         LegacyUriVariablesConverterInterface|UriVariablesConverterInterface|null $uriVariablesConverter = null,
+        private readonly LoggerInterface $logger
```
dunglas, line 54:
```suggestion
        private readonly ?LoggerInterface $logger = null,
```

(To prevent a BC break).

src/Symfony/EventListener/ReadListener.php
```diff
@@ -142,6 +144,8 @@ public function onKernelRequest(RequestEvent $event): void
                 || ($operation instanceof Put && !($operation->getAllowCreate() ?? false))
             )
         ) {
+            //Log it in debug in case the dev just forgot to implement it
```
dunglas, line 147:
```suggestion
            // In case the dev just forgot to implement it
```

src/Symfony/EventListener/ReadListener.php
```diff
@@ -142,6 +144,8 @@ public function onKernelRequest(RequestEvent $event): void
...
+            $this->logger->debug("No provider found for " . $resourceClass);
```
dunglas, line 148:
```suggestion
            $this->logger->debug('No provider registered for {resource_class}', ['resource_class' => $resourceClass]);
```

dunglas (changes requested):

mrossard:
> @mrossard the conflict is in `use`. You need to rebase.

oh, cool, i'm just dumb then. :D I was sure i had pulled the changes before creating the branch, sorry.

I'll try to look at how it can be added in the profiler/ web debug toolbar too, that's a great point.

src/Symfony/EventListener/ReadListener.php
```diff
@@ -142,6 +144,8 @@ public function onKernelRequest(RequestEvent $event): void
...
             throw new NotFoundHttpException('Not Found');
```
Aerendir, line 149: Maybe it could be useful to also pass the previous exception here 🙏

Aerendir (changes requested):

src/Symfony/EventListener/ReadListener.php
mrossard: I could add the exception message or even the stacktrace, but that seems a bit overkill doesn't it? Having a log saying something like this should be fine, shouldn't it?

`app.DEBUG: No provider registered for App\ApiResource\TestResource {"resource_class":"App\\ApiResource\\TestResource"} []`

src/Symfony/EventListener/ReadListener.php
mrossard: sure, done.

src/Symfony/EventListener/ReadListener.php
mrossard: done.

src/Symfony/EventListener/ReadListener.php
mrossard: done.

mrossard:
I rebased it and added the logger to the api_platform.state_provider.read service definition on the symfony config side. Not sure if there's something similar to do for laravel...?

src/State/Provider/ReadProvider.php
```diff
@@ -84,6 +88,8 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
                 || ($operation instanceof Put && !($operation->getAllowCreate() ?? false))
             )
         ) {
+            // In case the dev just forgot to implement it
+            $this->logger?->debug('No provider registered for {resource_class}', ['resource_class' => $resourceClass]);
```
soyuka, line 92: imo this line should be at line :81

src/State/Provider/ReadProvider.php
```diff
@@ -84,6 +88,8 @@ public function provide(Operation $operation, array $uriVariables = [], array $c
...
             throw new NotFoundHttpException('Not Found');
```
soyuka, line 93: we could also keep the previous exception (ProviderNotFound) and use it here as `previous` argument of `NotFoundHttpException`

src/State/Provider/ReadProvider.php
mrossard: I wasn't sure that was something useful to report for PUT and POST, but that's easy enough to change.

src/State/Provider/ReadProvider.php
mrossard: That's a clear benefit of moving it to line 81. I'll do that.

src/Symfony/EventListener/ReadListener.php
mrossard: sorry, misunderstood what you meant. I'll do that.

Aerendir: @mrossard , thank you! 🙏

Aerendir (approved):
Looks good on my side. Thank you! 🙏

PR #6526, opened 2024-08-20
chore(laravel): code cleanup

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/src/Laravel/Controller/ApiPlatformController.php
+++ b/src/Laravel/Controller/ApiPlatformController.php
@@ -68,11 +68,11 @@ class ApiPlatformController extends Controller
         }
 
         if (null === $operation->canRead()) {
-            $operation = $operation->withRead($operation->getUriVariables() || $request->isMethodSafe());
+            $operation = $operation->withRead($operation->getUriVariables() || $request->isMethodSafe()); // @phpstan-ignore-line
         }
 
         if (null === $operation->canDeserialize()) {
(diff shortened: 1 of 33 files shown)


src/Laravel/Controller/ApiPlatformController.php
```diff
@@ -64,11 +64,11 @@ public function __invoke(Request $request): Response
             $operation = $operation->withValidate(!$request->isMethodSafe() && !$request->isMethod('DELETE'));
         }
 
-        if (null === $operation->canRead() && $operation instanceof HttpOperation) {
+        if ($operation instanceof HttpOperation && null === $operation->canRead()) {
```
soyuka, line 67: maybe we should do this just once at the top of the __invoke function, it's not possible that the operation isn't an instance of HttpOperation.

src/Laravel/Eloquent/State/Options.php
```diff
@@ -15,7 +15,7 @@
 
 use ApiPlatform\State\OptionsInterface;
 
-class Options implements OptionsInterface
+final class Options implements OptionsInterface
```
soyuka, line 18: this should not be final as it's an extension point

src/Laravel/composer.json
```diff
@@ -17,6 +17,7 @@
             "homepage": "https://api-platform.com/community/contributors"
         }
     ],
+    "minimum-stability": "dev",
```
soyuka, line 20:
```suggestion
```

PR #6528, opened 2024-08-20
chore(laravel): only enable JSON-LD by default

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

For improved security (less attack surface by default) and consistency with the Symfony variant.

### DIFF
--- a/src/Laravel/config/api-platform.php
+++ b/src/Laravel/config/api-platform.php
@@ -31,7 +31,7 @@ return [
 
     'formats' => [
         'jsonld' => ['application/ld+json'],
-        'jsonapi' => ['application/vnd.api+json'],
+        //'jsonapi' => ['application/vnd.api+json'],
     ],
 
     'patch_formats' => [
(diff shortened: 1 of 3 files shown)


src/Laravel/config/api-platform.php
```diff
@@ -32,18 +32,18 @@
...
         'json' => ['application/merge-patch+json'],
     ],
 
     'docs_formats' => [
-        'jsonopenapi' => ['application/vnd.openapi+json'],
-        'json' => ['application/json'],
         'jsonld' => ['application/ld+json'],
+        'jsonopenapi' => ['application/vnd.openapi+json'],
         'html' => ['text/html'],
+        //'json' => ['application/json'],
```
soyuka, line 46: let's remove this one completely no? I'm think that this is useful only if you want to reach `/docs.json`

dunglas: Done. I added YAML Open API support instead.

PR #6529, opened 2024-08-20
fix(laravel): Eloquent date and datetime type detection

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

`date` wasn't detected, and immutable and mutable versions were falsely detected.


dunglas:
There is more special types to support (can be done in follow up PRs): https://laravel.com/docs/11.x/eloquent-mutators#attribute-casting

PR #6532, opened 2024-08-21
fix(laravel): snake case props

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Currently, model properties containing an underscore aren't written or read because an unneeded case conversion occurs. This patch fixes the issue.

~~A better fix could be to always convert internal snake_case model props to public camelCase API resources props.~~
Now implemented in this PR.

### DIFF
--- a/src/JsonApi/Tests/Fixtures/CustomConverter.php
+++ b/src/JsonApi/Tests/Fixtures/CustomConverter.php
@@ -15,7 +15,6 @@ namespace ApiPlatform\JsonApi\Tests\Fixtures;
 
 use Symfony\Component\Serializer\NameConverter\AdvancedNameConverterInterface;
 use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
-use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
 /**
  * Custom converter that will only convert a property named "nameConverted"
@@ -23,7 +22,7 @@ use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
  */
 class CustomConverter implements AdvancedNameConverterInterface
 {
-    private NameConverterInterface $nameConverter;
+    private CamelCaseToSnakeCaseNameConverter $nameConverter;
 
     public function __construct()
     {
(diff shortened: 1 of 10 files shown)


src/JsonApi/Tests/Fixtures/CustomConverter.php
```diff
@@ -15,15 +15,14 @@
...
-use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
 
 /**
  * Custom converter that will only convert a property named "nameConverted"
  * with the same logic as Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter.
  */
 class CustomConverter implements AdvancedNameConverterInterface
 {
-    private NameConverterInterface $nameConverter;
+    private CamelCaseToSnakeCaseNameConverter $nameConverter;
```
soyuka, line 25: can we revert this?

src/Laravel/Eloquent/PropertyAccess/PropertyAccessor.php
```diff
@@ -22,9 +23,12 @@
  */
 final class PropertyAccessor implements PropertyAccessorInterface
 {
+    private readonly PropertyAccessorInterface $inner;
+
     public function __construct(
-        private readonly PropertyAccessorInterface $inner,
+        ?PropertyAccessorInterface $inner = null,
```
soyuka, line 29:
```suggestion
        private readonly ?PropertyAccessorInterface $inner =  PropertyAccess::createPropertyAccessor(),
```

soyuka:
Is it possible to add a property to test this at src/Laravel/workbench/app/Models/Book.php ?

src/Laravel/Eloquent/PropertyAccess/PropertyAccessor.php
dunglas: This syntax isn't supported (`expression is not allowed as parameter default value`) and readonly classes aren't available in 8.1.

PR #6538, opened 2024-08-22
fix(laravel): visible and hidden fields support

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

`Model::$visible` is an allow list. If it is defined, only fields explicitly listed in the array must be visible.
This patch fixes the issue and adds some tests for the feature.

According to [the docs](https://laravel.com/docs/11.x/eloquent-serialization#temporarily-modifying-attribute-visibility), it's also possible to hide the field for an instance by using `makeVisible()` and `makeHidden()`. 
I may be wrong, but this feature doesn't seem to be currently supported. This should be fixed in a follow-up PR.

### DIFF
--- a/src/Laravel/Tests/JsonLdTest.php
+++ b/src/Laravel/Tests/JsonLdTest.php
@@ -286,9 +286,10 @@ class JsonLdTest extends TestCase
 
     public function testVisible(): void
     {
-        $response = $this->get('/api/posts/1', ['accept' => 'application/ld+json']);
+        $response = $this->get('/api/posts', ['accept' => 'application/ld+json']);
         $response->assertStatus(200);
         $response->assertHeader('content-type', 'application/ld+json; charset=utf-8');
+        dump($response->json());
         $response->assertJsonMissingPath('internalNote');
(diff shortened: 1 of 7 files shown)


src/Laravel/Tests/JsonLdTest.php
```diff
@@ -269,4 +269,20 @@ public function testSluggable(): void
...
+        $response->assertHeader('content-type', 'application/ld+json; charset=utf-8');
+        $response->assertJsonMissingPath('internalNote');
+    }
+
+    public function testVisible(): void
+    {
+        $response = $this->get('/api/books', ['accept' => 'application/ld+json']);
+        $response->assertStatus(200);
+        $response->assertHeader('content-type', 'application/ld+json; charset=utf-8');
+        $this->assertStringNotContainsString('internalNote', $response->getContent());
```
soyuka, line 286: phpstan isn't happy about this, I guess `getContent` isn't always a string

PR #6541, opened 2024-08-22
fix(laravel): entrypoint serialization

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

The `jsonapi` format weren't working. This patch fixes the issue.

### DIFF
--- a/src/Documentation/Action/EntrypointAction.php
+++ b/src/Documentation/Action/EntrypointAction.php
@@ -37,13 +37,13 @@ final class EntrypointAction
         private readonly ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory,
         private readonly ProviderInterface $provider,
         private readonly ProcessorInterface $processor,
-        private readonly array $documentationFormats = []
+        private readonly array $documentationFormats = [],
     ) {
     }
 
     public function __invoke(Request $request)
     {
-        static::$resourceNameCollection = $this->resourceNameCollectionFactory->create();
+        self::$resourceNameCollection = $this->resourceNameCollectionFactory->create();
         $context = [
             'request' => $request,
             'spec_version' => (string) $request->query->get(LegacyOpenApiNormalizer::SPEC_VERSION),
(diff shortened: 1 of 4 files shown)


src/Documentation/Action/EntrypointAction.php
```diff
@@ -37,13 +37,13 @@ public function __construct(
         private readonly ResourceNameCollectionFactoryInterface $resourceNameCollectionFactory,
```
dunglas, line 37: By the way, this class should probably be undeprecated as it is now used again.

src/Documentation/Action/EntrypointAction.php
```diff
@@ -37,13 +37,13 @@ public function __construct(
...
         private readonly ProviderInterface $provider,
         private readonly ProcessorInterface $processor,
-        private readonly array $documentationFormats = []
+        private readonly array $documentationFormats = [],
```
dunglas, line 40: This is likely a bug in Symfony variant. AFAIU, `formats` should be injected here, not `docs_formats`.

src/Documentation/Action/EntrypointAction.php
soyuka: Shouldn't we just copy that class to the Laravel directory?

src/Documentation/Action/EntrypointAction.php
soyuka: indeed both implementation are not necessary the same

soyuka: Laravel has no differenciation between both formats yet?

src/Documentation/Action/EntrypointAction.php
dunglas: As you wish

src/Documentation/Action/EntrypointAction.php
dunglas: That would be cool to be able to render SwaggerUI when requesting `/api` as we do for the Symfony variant.

PR #6564, opened 2024-08-30
fix(Symfony): canWrite guessing in WriteListener

### LINKED ISSUE #6501: WriteListener is called on Get operations
**API Platform version(s) affected**: 3.3 +

**Description**  

A `Get` operation using a Controller is always calling the `WriteProcessor` after the Controller is called, which causes an "Invalid identifier or value configuration" if you are using another identifier.

**How to reproduce**  

Create a `Get` operation like so : 

```php
new Get(
    uriTemplate: '/something/by_something/{not_id}',
    controller: SomethingController::class,
    output: SomethingOutput::class,
    read: false
)
```

**Possible Solution**  

Sadly I don't come with a simple fix BUT I figured out that the `persist` attribute is `true` after being resolved in `WriteListener`. This is caused by the fact that the Request attribute `_api_persist` is not defined and defined as `true` by default, which is a shame since writing in a `Get` operation looks kind of weird.

**Additional Context**  

This exact setup works perfectly using 3.2.x

Also I figured out that on a simple operation like : 

```php
new Get(
    output: SomethingOutput::class,
    provider: SomethingProvider::class,
)
```

The `WriteProcessor` is also called which makes no sense.

Fabrn:
| Q             | A
| ------------- | ---
| Branch?       | 3.3
| Tickets       | Closes #6501 

As explained in the issue, in version 3.3 : a `Get` operation using a Controller is always calling the `WriteProcessor` after the Controller is called, which causes an "Invalid identifier or value configuration" if you are using another identifier.

Example : 

```php
new Get(
    output: SomethingOutput::class,
    provider: SomethingProvider::class,
)
```

A simple operation like this calls the `WriteProcessor`.

### DIFF
--- a/src/Symfony/EventListener/WriteListener.php
+++ b/src/Symfony/EventListener/WriteListener.php
@@ -90,6 +90,16 @@ final class WriteListener
             return;
         }
 
+        $canWrite = $operation->canWrite();
+
+        if (null === $canWrite) {
+            $canWrite = !$request->isMethodSafe();
+        }
+
+        if (!$canWrite) {
+            return;
+        }
+
         if ($operation && (!$this->processor instanceof CallableProcessor && !$this->iriConverter)) {
             if (null === $operation->canWrite()) {
                 $operation = $operation->withWrite(!$request->isMethodSafe());
--- a/src/Symfony/EventListener/WriteListener.php
+++ b/src/Symfony/EventListener/WriteListener.php
@@ -90,7 +90,7 @@ final class WriteListener
             return;
         }
 
-        $canWrite = $operation->canWrite();
+        $canWrite = $operation?->canWrite();
 
         if (null === $canWrite) {
             $canWrite = !$request->isMethodSafe();


src/Symfony/EventListener/WriteListener.php
```diff
@@ -90,6 +90,16 @@ public function onKernelView(ViewEvent $event): void
...
+        $canWrite = $operation?->canWrite();
+
+        if (null === $canWrite) {
+            $canWrite = !$request->isMethodSafe();
+        }
+
+        if (!$canWrite) {
+            return;
+        }
+
```
soyuka, line 102:
```suggestion
        if (false == ($operation?->canWrite() ?? !$request->isMethodSafe())) {
            return;
        }
```

But shouldn't you put this line 146?

Fabrn: This bit of code seems fine to me 🙂 You could even use `===` instead of `==` here isn't it ? Since the expression between parenthesis will always return a boolean here.

> But shouldn't you put this line 146?

I don't think so, since the issue is that we call the `WriteProcessor` about line 117 like so : 

```php
$data = $this->processor->process($controllerResult, $operation, $uriVariables, [
    'request' => $request,
    'uri_variables' => $uriVariables,
    'resource_class' => $operation->getClass(),
    'previous_data' => false === $operation->canRead() ? null : $request->attributes->get('previous_data'),
]);
```

We must leave the `WriteListener` *before* this call

soyuka: We already do this correctly line 104:

```
            if (null === $operation->canWrite()) {
                $operation = $operation->withWrite(!$request->isMethodSafe());
            }
 ```

Comment (Fabrn) [reply src/Symfony/EventListener/WriteListener.php:102]: Yeah. Holidays and a month went by, I had to setup a little project to remember what was going wrong here 😄 

Initially, I wrote this issue since I get an "Invalid identifier value or configuration." error while running the following operation : 

```php
new Get(
    uriTemplate: '/users/by_identifier/{identifier}',
    controller: GetUserByIdentifier::class,
    read: false,
),
```

I then made my Controller which resolves the User like so : 

```php
#[AsController]
class GetUserByIdentifier
{
    public function __construct(
        private readonly UserRepository $userRepository,
    ) {}

    public function __invoke(string $identifier): User
    {
        return $this->userRepository->findOneBy(['identifiant' => $identifier]);
    }
}
```

The point of using `read: false` here is to prevent resource reading since we aren't using the Doctrine identifier here, and let the Controller do the job. However, after line 104, you get the following : 

```php
$uriVariables = $request->attributes->get('_api_uri_variables') ?? [];

if (!$uriVariables && !$operation instanceof Error && $operation instanceof HttpOperation) {
    try {
        $uriVariables = $this->getOperationUriVariables($operation, $request->attributes->all(), $operation->getClass());
    } catch (InvalidIdentifierException|InvalidUriVariableException $e) {
    }
}
```

This bit of code results in the "Invalid identifier value or configuration." error. If I add the bit of code discussed here, I dodge the URI variables resolving, and the call of the write processor, which allows me to have this working.

The point here is to not call the `ReadListener` with `read: false` because we would have got the issue and it would have been expected. But calling the `WriteListener` is not expected here and should not prevent my read operation from working.

Also, I personally think that we better dodge any more logic if possible. Resolving URI variables and calling the Processor for an operation that should not be handled for this Listener is too much

soyuka: I don't understand why you'd use controllers with API Platform ^^ use `uriVariables: ['identifier']`

Where is the `getOperationUriVariables` code you pasted here? you can also juste use `write: false`.

Fabrn: > I don't understand why you'd use controllers with API Platform ^^ use uriVariables: ['identifier']

I confirm that my simple use case works fine using `uriVariables: ['identifier']`. I had a lack of knowledge about this. 

However, since Api Platform **does not** prevent you from using controllers, they are a part of the bundle and the bundle must adapt and work with it. Since it works fine on 3.2 and always did, I don't know why we should stop supporting it. People with more complex behaviors may rely on controllers to work for their use case. 

Also, if you guys don't want us to use controllers to work anymore, I think it's better to do it on purpose instead of explicitly letting a bug exist and leading to incorrect and confusing behavior.

> Where is the getOperationUriVariables code you pasted here?

Version 3.3, WriteListener :

image

> you can also juste use write: false

I don't agree with the fact that I must use `write: false` in a read operation. It is absurd

soyuka:
Let's start again @Fabrn I'd like to get this fixed, what should we do? Adding `uriVariables` fixed your issue?

Fabrn:
Hello 🙂 

For the moment, we chose to keep the 3.2 version so our previous code would still work as intended, for time saving reasons. Sadly, I cannot confirm if this problem would have been solved for now.

However, even if I would resolve it using `uriVariables`, the thing is that a behavior has been broken on 3.3 and that's why I think it should be resolved. If we tell me that this is intended, well I guess we're fine and there's nothing to be fixed 😄

PR #6566, opened 2024-08-30
test: anonymous bundle class breaks twig extension

soyuka:


tests/PhpUnitBundle.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace ApiPlatform\Tests;
+
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+final class PhpUnitBundle extends Bundle
+{
+    public function shutdown(): void
+    {
+        restore_exception_handler();
```
dunglas, line 22: Can you put this call in our kernel directly?

PR #6570, opened 2024-09-01
fix: remove PUT from default operations

soyuka:


dunglas (approved):
Nice! We'll have to add some docs about the rationale and how to add back `Put` if necessary.

Shine-neko:
@soyuka Why ? :thinking:

dunglas:
@Shine-neko because in most cases users will want to use `PATCH` instead. `PUT` is a bit more niche, and for security reasons, not enabling two operations having (almost) the same use case is likely better. `PUT` is still supported and can be enabled explicitly, or globally by adding it in the default config.

PR #6574, opened 2024-09-02
feat(processor): add missing context when generating iri for content …

joelwurtz:
…location header

| Q             | A
| ------------- | ---
| Branch?       | 3.2
| Tickets       | 
| License       | MIT
| Doc PR        | 

When it generate the iri, it misses the context, so if we need some uri variables from the context to generate the route it will not work (like a object_id not available in the resource)

Without this, it will throw an error that it cannot generate an iri item for this resource

### DIFF
--- a/src/State/Processor/RespondProcessor.php
+++ b/src/State/Processor/RespondProcessor.php
@@ -103,7 +103,7 @@ final class RespondProcessor implements ProcessorInterface
         $status ??= self::METHOD_TO_CODE[$method] ?? 200;
 
         if ($hasData && $this->iriConverter) {
-            $iri = $this->iriConverter->getIriFromResource($originalData);
+            $iri = $this->iriConverter->getIriFromResource($originalData, context: $context);
             $headers['Content-Location'] = $iri;
 
             if ((201 === $status || (300 <= $status && $status < 400)) && 'POST' === $method) {


src/State/Processor/RespondProcessor.php
```diff
@@ -103,7 +103,7 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
         $status ??= self::METHOD_TO_CODE[$method] ?? 200;
 
         if ($hasData && $this->iriConverter) {
-            $iri = $this->iriConverter->getIriFromResource($originalData);
+            $iri = $this->iriConverter->getIriFromResource($originalData, context: $context);
```
soyuka, line 106: I think I already refused a PR like this one, we don't know the nature of `originalData` and URI Variables in the context may not be the right ones anymore. Question is, how is it that we can't extract identifiers from `originalData` here?

joelwurtz: Because in this case identifiers are not in the data but only in the url, something like this : 

```
#[ApiResource(
    operations: [
        new Get(
            uriTemplate: '/realm/{realmId}/data',
            uriVariables: [
                'realmId' => new Link(
                    toProperty: 'realm',
                    fromClass: Realm::class,
                    security: 'is_granted("REALM_GET_DATA", realm)',
                ),
            ],
            provider: RealmDataProvider::class,
        ),
    ],
)]
final class Data {
    public function __construct(public string $name)
    {
    }
}
```

joelwurtz: Which makes this empty cf https://github.com/api-platform/core/blob/main/src/Symfony/Routing/IriConverter.php#L183

soyuka: the `Link` suggests that you're retrieving the `realmId` from the property `realm`? Isn't the identifier in your $originalData?I wish we had a better extension point for https://github.com/api-platform/core/blob/main/src/Symfony/Routing/IriConverter.php#L187 instead of adding the context there as you won't be able to fix the tests with this change.

joelwurtz: No i don't have this property in my entity, AFAIK the link never map the property to the entity ? (We tried to do that in other api calls and it was never linked)

joelwurtz: Another solution would be to be able to configure an IriConverter from a resource, so we can override this behavior per operation / resource ?

PR #6583, opened 2024-09-05
feat: better path sorting for openapi UIs

mrossard:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| License       | MIT

I've noticed for a while that operations sorting as displayed by tools like swagger-ui or scalar could be confusing at first - they are sorted by path, but displayed by tags first.

This PR sorts them by tags, then by path. As a very small optimization (assuming there are more writes than reads) it also triggers the sorting only when the paths are actually used instead of sorting each time a new path is added.

### DIFF
--- a/src/OpenApi/Model/Paths.php
+++ b/src/OpenApi/Model/Paths.php
@@ -20,8 +20,33 @@ final class Paths
     public function addPath(string $path, PathItem $pathItem): void
     {
         $this->paths[$path] = $pathItem;
+    }
 
-        ksort($this->paths);
+    private function comparePathsByKey($keyA, $keyB): int
+    {
+        $a = $this->paths[$keyA];
+        $b = $this->paths[$keyB];
+
+        $tagsA = [
+            ...($a->getGet()?->getTags() ?? []),
+            ...($a->getPost()?->getTags() ?? []),
+            ...($a->getPut()?->getTags() ?? []),
+            ...($a->getDelete()?->getTags() ?? []),
+        ];
+        sort($tagsA);
+
+        $tagsB = [
+            ...($b->getGet()?->getTags() ?? []),
+            ...($b->getPost()?->getTags() ?? []),
+            ...($b->getPut()?->getTags() ?? []),
+            ...($b->getDelete()?->getTags() ?? []),
+        ];
+        sort($tagsB);
+
+        return match(true) {
+            current($tagsA) === current($tagsB) => $keyA <=> $keyB,
+            default => current($tagsA) <=> current($tagsB),
+        };
     }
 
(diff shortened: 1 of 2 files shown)


mrossard:
This probably should target 3.3, used 3.4 without thinking...sorry.

src/OpenApi/Model/Paths.php
```diff
@@ -31,6 +56,9 @@ public function getPath(string $path): ?PathItem
 
     public function getPaths(): array
     {
+        // sort paths by tags, then by path for each tag
+        uksort($this->paths, $this->comparePathsByKey(...));
+
```
soyuka, line 61: maybe that this should be done inside the OpenApi Normalizer instead of the Model class?

mrossard: That's a good point, i wasn't sure where this is used and did it as a quick fix. 

I'll dig a bit further then.

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -39,7 +39,42 @@ public function __construct(private readonly NormalizerInterface $decorated)
      */
     public function normalize(mixed $object, ?string $format = null, array $context = []): array
     {
-        $pathsCallback = static fn ($decoratedObject): array => $decoratedObject instanceof Paths ? $decoratedObject->getPaths() : [];
+        $pathsCallback = static function ($decoratedObject): array {
```
mrossard, line 42: I should probably extract a separate method for this.

src/OpenApi/Serializer/OpenApiNormalizer.php
```diff
@@ -95,4 +95,44 @@ public function hasCacheableSupportsMethod(): bool
...
+
+                // sort paths by tags, then by path for each tag
+                uksort($paths, function ($keyA, $keyB) use ($paths) {
+                    $a = $paths[$keyA];
+                    $b = $paths[$keyB];
+
+                    $tagsA = [
+                        ...($a->getGet()?->getTags() ?? []),
+                        ...($a->getPost()?->getTags() ?? []),
+                        ...($a->getPut()?->getTags() ?? []),
```
soyuka, line 113: it should be `patch` instead as `put` will no longer be there by default in 4.0

mrossard: That's a mistake on my part, I intended to add both. I still leave Put in case it's used, don't I?

mrossard: It's added

soyuka:
Thanks!

PR #6584, opened 2024-09-05
feat(doctrine): allow property aliasing on doctrine filters

Zzareb:
| Q             | A
| ------------- | ---
| Branch?       | main, new feature
| License       | MIT
| Doc PR        | [api-platform/docs](https://github.com/api-platform/docs/pull/1974)

Add a way to alias properties used to query the api filters.

via a simple configuration on the filter declaration, we can replace the parameters such as 

my.relation.nested.superField by a simple superField that will be automatically resolved under the hood.

the documentation generated in the response is up to date with the aliased fields.

### DIFF
--- a/src/Doctrine/Common/Filter/PropertyAliasesFilterTrait.php
+++ b/src/Doctrine/Common/Filter/PropertyAliasesFilterTrait.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Doctrine\Common\Filter;
+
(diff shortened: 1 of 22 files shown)


soyuka:
Interesting, although aliases are made possible using the new `QueryParameter` functionality since API Platform 3.2. I'm wondering how the documentation (hydra) outputs the view template for such a feature? What about OpenAPI does the parameter gets documented?

Zzareb:
> Interesting, although aliases are made possible using the new `QueryParameter` functionality since API Platform 3.2. I'm wondering how the documentation (hydra) outputs the view template for such a feature? What about OpenAPI does the parameter gets documented?

@soyuka 
Thanks for the reply. I’ll search the docs for the QueryParameter you mention ! 
the hydra output returns the aliased property only, might be worthwhile returning both. 

I did not tackle open api documentation for this feature, because it is more a flavour for a dedicated field in an entity than a replacement api-wide for the parameter… might be interesting to have it as well for the given routes, although I’m not sure about the implications in the swagger client etc

SebLevDev:
Any news?
I have an API on v3.4 and each resource is a DTO with different properties from the Doctrine entity.

PR #6585, opened 2024-09-05
fix(laravel): disable GraphQL by default and fix provider

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Prevents a crash when GraphQL isn't installed, which is the default.

### DIFF
--- a/src/Laravel/ApiPlatformProvider.php
+++ b/src/Laravel/ApiPlatformProvider.php
@@ -1090,15 +1090,18 @@ class ApiPlatformProvider extends ServiceProvider
 
         $this->loadViewsFrom(__DIR__.'/resources/views', 'api-platform');
 
-        $fieldsBuilder = $this->app->make(FieldsBuilderEnumInterface::class);
-        $typeBuilder = $this->app->make(ContextAwareTypeBuilderInterface::class);
-        $typeBuilder->setFieldsBuilderLocator(new ServiceLocator(['api_platform.graphql.fields_builder' => $fieldsBuilder]));
+        $config = $this->app['config'];
+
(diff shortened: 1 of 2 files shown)


src/Laravel/ApiPlatformProvider.php
```diff
@@ -1090,15 +1090,18 @@ public function boot(ResourceNameCollectionFactoryInterface $resourceNameCollect
...
-        $fieldsBuilder = $this->app->make(FieldsBuilderEnumInterface::class);
-        $typeBuilder = $this->app->make(ContextAwareTypeBuilderInterface::class);
-        $typeBuilder->setFieldsBuilderLocator(new ServiceLocator(['api_platform.graphql.fields_builder' => $fieldsBuilder]));
+        $config = $this->app['config'];
+
+        if ($config->get('api-platform.graphql.enabled')) {
+            $fieldsBuilder = $this->app->make(FieldsBuilderEnumInterface::class);
+            $typeBuilder = $this->app->make(ContextAwareTypeBuilderInterface::class);
+            $typeBuilder->setFieldsBuilderLocator(new ServiceLocator(['api_platform.graphql.fields_builder' => $fieldsBuilder]));
+        }
```
dunglas, line 1099: Is this intended @soyuka? it looks like `$fieldsBuilder`is unused. It's maybe a remaining of something?

soyuka: we're in the `boot` method, this is to handle a circular dependency between the FieldsBuilder and theTypeBuilder of graphql...

PR #6623, opened 2024-09-18
feat: if a policy matches the name of a model we automatically register it

valentindrdt:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | Closes #..., closes #... 
| License       | MIT

Automatic policy registration

### DIFF
--- a/src/Laravel/Security/ResourceAccessChecker.php
+++ b/src/Laravel/Security/ResourceAccessChecker.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace ApiPlatform\Laravel\Security;
 
+use ApiPlatform\Laravel\Eloquent\Paginator;
 use ApiPlatform\Metadata\ResourceAccessCheckerInterface;
 use Illuminate\Support\Facades\Gate;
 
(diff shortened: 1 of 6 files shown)


src/Laravel/Security/ResourceAccessChecker.php
```diff
@@ -13,13 +13,27 @@
...
+        if (($object = $extraVariables['object']) instanceof Paginator) {
+            // do not deny access if no items are found
+            if (0 === $object->count()) {
+                return true;
+            }
+
+            // we're only checking access to the first item
+            foreach ($object as $obj) {
+                return Gate::allows($expression, $obj);
+            }
```
dunglas, line 34:
```suggestion
            // we're only checking access to the first item
            foreach ($object as $obj) {
                return Gate::allows($expression, $obj);
            }
            
            return true;
```

micro-optim (to be tested)

dunglas (approved):
Very cool!

dunglas:
Couldn't we do something similar for Form Request btw?

src/Laravel/Security/ResourceAccessChecker.php
dunglas, line 34:
```suggestion
        // ugly way to handle collection
        if (($object = $extraVariables['object']) instanceof Paginator) {
            Gate::allows($expression);
```

This should be enough.

valentindrdt: I refactored this part, therefore this code is outdated

PR #6624, opened 2024-09-18
fix: remove "hydra:" remaining prefix

vincentchalamon:
Additionally, [api-platform/admin](https://github.com/api-platform/admin) must be updated because of the following error (with `hydra_prefix: false`):

image

### DIFF
--- a/src/State/ApiResource/Error.php
+++ b/src/State/ApiResource/Error.php
@@ -25,7 +25,6 @@ use Symfony\Component\Serializer\Annotation\SerializedName;
 use Symfony\Component\WebLink\Link;
 
 #[ErrorResource(
-    types: ['hydra:Error'],
     openapi: false,
     uriVariables: ['status'],
     uriTemplate: '/errors/{status}',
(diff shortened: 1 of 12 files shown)


dunglas (approved):

src/State/ApiResource/Error.php
```diff
@@ -98,14 +97,7 @@ public function __construct(
...
-    #[Groups(['jsonld'])]
-    public function getHydraTitle(): ?string
-    {
-        return $this->title;
-    }
-
-    #[SerializedName('hydra:description')]
+    #[SerializedName('description')]
     #[Groups(['jsonld'])]
     public function getHydraDescription(): ?string
```
soyuka, line 102:
```suggestion
    public function getDescription(): ?string
```

+ enlever SerializedName

burned42:
Unfortunately it seems like this introduced a BC break for me with v4.0.3 as even with `hydra_prefix: true`, the hydra prefix is gone now.

For now I brought back the old behavior by downgrading `api-platform/state` and `api-platform/validator` to v4.0.2.

vincentchalamon:
Hi @burned42,

On which schema did you found that the hydra prefix was gone? On a resource schema (with json-ld format), the hydra prefix is still present (cf. [API Platform Demo tests](https://github.com/api-platform/demo/blob/4.0/api/tests/Api/schemas/Book/collection.json#L181-L184)). But on an error schema, the hydra prefix has been removed ([in 4.0.3](https://github.com/api-platform/core/commit/884b0331f1281b8ca5386ddf4e562ba16c3f6d59)). @soyuka can explain further more about this removal.

burned42:
@vincentchalamon yes, that's right. Sorry if I was a bit unspecific here.

I was only talking about the error schema.

vincentchalamon:
@burned42 @soyuka is currently working on a fix: https://github.com/api-platform/core/pull/6704

burned42:
@vincentchalamon @soyuka awesome, thank you :)

burned42:
As mentioned in https://github.com/api-platform/core/issues/6701#issuecomment-2407356349 it seems like the fix included in version 4.0.4 works :+1:

PR #6627, opened 2024-09-19
fix!(state): precise format on content-location

### LINKED ISSUE #6625: Is it possible to have the Content-Location for GetCollections?
**Description**  

Hello, I dunno if that's possible to return the header `Content-Location` with the resource location on GetCollection results.

Context:
In Souin/cache-handler caddy module we have the tag-based invalidation but we use the `Content-Location` HTTP response header to generate an associative array with the given format.
```
[Content-Location header value] => [{Generated cache key from the caddy module for req1}, {Generated cache key from the caddy module for req2}]
```
Example:
```
{
    "/books/1": ["GET-http-localhost-/books/1{VARY}content-type:application/xml", "GET-http-localhost-/books/1", "GET-http-localhost-/books/1{VARY}content-type:application/json"]
}
```

Actually I'm using a kind of hack in the Caddyfile to always have a `Content-Location` returned by the application with the following snippet:
```
	vars initialRequest {uri}
	header Content-Location {http.vars.initialRequest}
```

So it fixes the GetCollection Content-Location issue but generates an additional Content-Location value on GetItem. I know that caddy allows to defer the headers responses but if I use the defer, the middleware doesn't get this header.

**Example**

soyuka:
fixes #6625

### DIFF
--- a/features/hydra/collection.feature
+++ b/features/hydra/collection.feature
@@ -77,6 +77,7 @@ Feature: Collections support
     When I send a "GET" request to "/dummies?page=7"
     Then the response status code should be 200
     And the response should be in JSON
+    And the header "Content-Location" should be equal to "/dummies.jsonld?page=7"
     And the header "Content-Type" should be equal to "application/ld+json; charset=utf-8"
     And the JSON should be valid according to this schema:
     """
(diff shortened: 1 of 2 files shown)


dunglas:
Shouldn't we also append query parameters?

darkweak:
@dunglas the query parameters don't influence the Content-Location IMHO.

dunglas:
query parameters are part of the URI, so it does IMHO. `/collection?page=1` is not the same resource as `/collection?page=42`

PR #6633, opened 2024-09-20
[Laravel] Fix virtual fields metadata

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Laravel/Eloquent/Metadata/ModelMetadata.php
+++ b/src/Laravel/Eloquent/Metadata/ModelMetadata.php
@@ -131,13 +131,14 @@ final class ModelMetadata
                 'name' => $name,
                 'type' => null,
                 'increments' => false,
-                'nullable' => null,
+                'nullable' => true,
                 'default' => null,
                 'unique' => null,
                 'fillable' => $model->isFillable($name),
                 'hidden' => $this->attributeIsHidden($name, $model),
                 'appended' => $model->hasAppended($name),
                 'cast' => $cast,
+                'primary' => $name === $model->getKeyName(),
             ])
             ->values();
     }
--- a/src/Laravel/Eloquent/Metadata/ModelMetadata.php
+++ b/src/Laravel/Eloquent/Metadata/ModelMetadata.php
@@ -109,6 +109,7 @@ final class ModelMetadata
     public function getVirtualAttributes(Model $model, array $columns): Collection
     {
         $class = new \ReflectionClass($model);
+        $columns = collect($columns);
 
         return collect($class->getMethods())
             ->reject(
@@ -126,7 +127,7 @@ final class ModelMetadata
 
                 return [];
             })
-            ->reject(fn ($cast, $name) => collect($columns)->contains('name', $name))
+            ->reject(fn ($cast, $name) => $columns->contains('name', $name))
             ->map(fn ($cast, $name) => [
                 'name' => $name,
                 'type' => null,
--- a/src/Laravel/workbench/app/Models/Book.php
+++ b/src/Laravel/workbench/app/Models/Book.php
@@ -64,7 +64,7 @@ class Book extends Model
     use HasFactory;
     use HasUlids;
 
-    protected $visible = ['name', 'author', 'isbn', 'publication_date'];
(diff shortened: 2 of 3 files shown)


src/Laravel/Eloquent/Metadata/ModelMetadata.php
```diff
@@ -131,13 +131,14 @@ public function getVirtualAttributes(Model $model, array $columns): Collection
...
                 'increments' => false,
-                'nullable' => null,
+                'nullable' => true,
                 'default' => null,
                 'unique' => null,
                 'fillable' => $model->isFillable($name),
                 'hidden' => $this->attributeIsHidden($name, $model),
                 'appended' => $model->hasAppended($name),
                 'cast' => $cast,
+                'primary' => $name === $model->getKeyName(),
```
GromNaN, line 141: The `primary` attribute is accessed for every fields here:
https://github.com/api-platform/core/blob/80fc2d2303f35bccdc6789dc72a205781c3dc3d0/src/Laravel/Eloquent/Metadata/Factory/Property/EloquentPropertyNameCollectionMetadataFactory.php#L55-L56

In MongoDB, "id" ends as a "virtual field" because we have an accessor for it: https://github.com/mongodb/laravel-mongodb/blob/5.0.2/src/Eloquent/DocumentModel.php#L78-L94
Maybe the metadata should be merged when a field is defined both in database schema and as a virtual field.

src/Laravel/Eloquent/Metadata/ModelMetadata.php
```diff
@@ -131,13 +131,14 @@ public function getVirtualAttributes(Model $model, array $columns): Collection
                 'name' => $name,
                 'type' => null,
                 'increments' => false,
-                'nullable' => null,
+                'nullable' => true,
```
GromNaN, line 134: The `nullable` field [must be a boolean](https://github.com/symfony/symfony/blob/8debb252592cf7d6943ac34968413d1a9a7b501d/src/Symfony/Component/PropertyInfo/Type.php#L77), as it is given to the `new Type()` here: https://github.com/api-platform/core/blob/80fc2d2303f35bccdc6789dc72a205781c3dc3d0/src/Laravel/Eloquent/Metadata/Factory/Property/EloquentPropertyMetadataFactory.php#L73-L81

src/Laravel/workbench/app/Models/Book.php
```diff
@@ -70,4 +71,10 @@ public function author(): BelongsTo
     {
         return $this->belongsTo(Author::class);
     }
+
+    // Virtual field
+    public function getTitleAttribute(): string
+    {
+        return Str::title($this->name);
+    }
```
GromNaN, line 79: WIP: Trying to add a test on virtual field.

src/Laravel/Eloquent/Metadata/ModelMetadata.php
toitzi: You might also want to do this on line 83 in the same class and remove the `isColumnPrimaryKey` method all along, imho your solution is cleaner.

soyuka:
can you run cs-fixer (inside the root of the project)? I'd like to merge this to fix virtual props

GromNaN:
I think #6668 does the trick, but I'll leave it to you.

domthomas-dev:
> I think #6668 does the trick, but I'll leave it to you.

No, with 4.0.2, same error.

toitzi:
@domthomas-dev #6668 was not released with 4.0.2, in order to test it you have to apply the patch yourself, i can confirm it is working

domthomas-dev:
> @domthomas-dev #6668 was not released with 4.0.2, in order to test it you have to apply the patch yourself, i can confirm it is working

Ok, but ...how ? :/ (I'm sorry, certainely it's a stupid question )

soyuka:
I'll tag a release Friday :), @domthomas-dev easiest is to install a dev version that points on the 4.0 branch using composer repositories. 

Btw, @GromNaN what's left to merge this?

PR #6637, opened 2024-09-20
Remove unnecessary dependency

maxhelias:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | -
| License       | MIT
| Doc PR        | -


src/Serializer/composer.json
```diff
@@ -25,7 +25,6 @@
         "php": ">=8.2",
         "api-platform/metadata": "^3.4 || ^4.0",
         "api-platform/state": "^3.4 || ^4.0",
-        "doctrine/collections": "^2.1",
```
dunglas, line 28: Good catch. This must be added in `require-dev`.

dunglas (changes requested):

src/Serializer/composer.json
maxhelias: I update when i reopen my laptop

soyuka:
Thanks @maxhelias

PR #6649, opened 2024-09-20
Add installation command, fix config overwrites

### LINKED ISSUE #6645: Configuration Overwrite on each composer update

**Description**  
The configuration file for this package is published within the `laravel-assets` group on [this line](https://github.com/api-platform/laravel/blob/e821ad27072228e4638f3de92d51fc3d15e02c7b/ApiPlatformProvider.php#L1211). However, it should not be part of this group. By default, Laravel republishes assets during each composer update, which is appropriate for assets but not for configuration files, as the config file would be overwritten each time.

Additionally, the package uses this group for asset publishing. While this isn't necessarily incorrect, it’s worth noting that the `laravel-assets` group seems reserved for Laravel packages. According to the official [Laravel documentation](https://laravel.com/docs/11.x/packages#public-assets), it’s recommended to use the `public` group for assets instead. This change is important since Laravel no longer copies vendor files and configs by default, a practice that is discouraged starting with Laravel 11.

**How to reproduce**  
Edit the config file, then run `composer update` — the config file will revert to the original version from the package.

**Possible Solution**  
Remove the config file from the `laravel-assets` group and switch the asset publishing to the `public` group. It would also be beneficial to update the documentation accordingly. This adjustment will prevent the files from being published automatically, which is unnecessary. If users wish to modify these files, they can publish them manually using commands like `php artisan vendor:publish --tag=api-platform-config`

toitzi:
Fix group names for pulishing config and assets.

Closes #6645

| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | Closes #6645 
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1989

### DIFF
--- a/src/Laravel/Console/InstallCommand.php
+++ b/src/Laravel/Console/InstallCommand.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Laravel\Console;
+
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'api-platform:install')]
+class InstallCommand extends Command
+{
+    /**
+     * @var string
+     */
+    protected $signature = 'api-platform:install';
+
+    /**
+     * @var string
+     */
+    protected $description = 'Install all of the API Platform resources';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle(): void
+    {
+        $this->comment('Publishing API Platform Assets...');
+        $this->callSilent('vendor:publish', ['--tag' => 'api-platform-assets']);
+
+        $this->comment('Publishing API Platform Configuration...');
+        $this->callSilent('vendor:publish', ['--tag' => 'api-platform-config']);
+
+        $this->info('API Platform installed successfully.');
+    }
+}
--- a/src/Laravel/ApiPlatformProvider.php
+++ b/src/Laravel/ApiPlatformProvider.php
@@ -1055,6 +1055,10 @@ class ApiPlatformProvider extends ServiceProvider
         if ($this->app['config']->get('api-platform.graphql.enabled')) {
             $this->registerGraphQl($this->app);
         }
+
+        if ($this->app->runningInConsole()) {
+            $this->commands([Console\InstallCommand::class]);
+        }
     }
 
     private function registerGraphQl(Application $app): void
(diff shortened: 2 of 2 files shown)


toitzi:
Note: Some first-party packages, such as Laravel Telescope or Laravel Horizon, provide an artisan command like php artisan horizon:install to publish service provider stubs, configs, and assets. If you're interested, I can include a similar command in this PR and update the documentation accordingly to reflect this feature.

Let me know if this is something you'd like to incorporate!

soyuka:
If I understand correctly the configuration won't be copied on package install right? Indeed I think we should introduce such a command and add it in the composer.json scripts so that's there no other step then `composer require api-platform/laravel`

toitzi:
Yes, while the config won't be copied by default, however this is actually expected behavior for Laravel packages. Even first-party packages like Telescope, Horizon, etc., do not automatically publish configs and assets upon installation. You can see this reflected in their installation instructions ([Telescope](https://laravel.com/docs/11.x/telescope#installation), [Horizon](https://laravel.com/docs/11.x/horizon#installation), [Scout](https://laravel.com/docs/11.x/scout#installation)). The [Laravel documentation on package development](https://laravel.com/docs/11.x/packages#configuration) also outlines this as standard practice in the Laravel ecosystem.

While I'm not deeply familiar with composer post-install/update scripts, they seem to run every time commands like `composer install` or `composer update` are executed, which would risk overwriting configs repeatedly. However, if you prefer handling it that way, I can certainly give it a try.

soyuka:
I was thinking that if we create our own `api-platform:install` command, we could definitely just check if the file exist and not replace it if it does?

Also, I see by using `composer create-project laravel/laravel` that some commands are executed by default:

20240920_19h18m28s_grim

see also https://github.com/laravel/laravel/blob/11.x/composer.json#L32-L47 

I really think that we should have the less steps possible to start a new api platform project, and I agree that replacing the configuration file is not good. @dunglas any opinion on this?

dunglas:
I'll go for the `api-platform:install` command. Let's use Laravel conventions as much as possible.

toitzi:
I added the install command, also updated the documentation PR.

src/Laravel/Console/InstallCommand.php
```diff
@@ -0,0 +1,49 @@
...
+     * @var string
+     */
+    protected $signature = 'api-platform:install';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Install all of the API Platform resources';
```
dunglas, line 34:
```suggestion
    /**
     * @var string
     */
    protected $description = 'Install all of the API Platform resources';
```

src/Laravel/Console/InstallCommand.php
```diff
@@ -0,0 +1,49 @@
...
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'api-platform:install')]
+class InstallCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
```
dunglas, line 26:
```suggestion
    /**
     * @var string
     */
```

src/Laravel/Console/InstallCommand.php
```diff
@@ -0,0 +1,49 @@
...
+    public function handle(): void
+    {
+        $this->comment('Publishing API Platform Assets...');
+        $this->callSilent('vendor:publish', ['--tag' => 'api-platform-assets']);
+
+        $this->comment('Publishing API Platform Configuration...');
+        $this->callSilent('vendor:publish', ['--tag' => 'api-platform-config']);
+
+        $this->info('API Platform scaffolding installed successfully.');
```
dunglas, line 47:
```suggestion
        $this->info('API Platform installed successfully.');
```

dunglas (approved):
LGTM

dunglas:
Thanks!!

PR #6658, opened 2024-09-21
Laravel: Use `Model::qualifyColumn()` instead of `$table.$column`

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | -
| License       | MIT
| Doc PR        | -

The function [`Model::qualifyColumn()`](https://github.com/laravel/framework/blob/abf86b08b08c5ea5c9247877bc8b6ea259805f80/src/Illuminate/Database/Eloquent/Model.php#L586-L593) does exactly the same as the existing code. But for MongoDB, we had to [override this method](https://github.com/mongodb/laravel-mongodb/blob/716d8e166809502a22017b476979678a57d5d91d/src/Eloquent/DocumentModel.php#L564-L567) as the column name (field name) cannot be prefixed with the table name (collection name). So we need this method to be used instead of having the logic hard coded.

For relations, we use the `getQualifiedForeignKeyName` that has the same mechanism for [`BelongsTo`](https://github.com/laravel/framework/blob/abf86b08b08c5ea5c9247877bc8b6ea259805f80/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php#L314) and [`HasOneOrManyThrough `](https://github.com/laravel/framework/blob/abf86b08b08c5ea5c9247877bc8b6ea259805f80/src/Illuminate/Database/Eloquent/Relations/HasOneOrManyThrough.php#L811), but nothing for [`HasOneOrMany`](https://github.com/laravel/framework/blob/abf86b08b08c5ea5c9247877bc8b6ea259805f80/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php#L557) (I don't know why 😞).

### DIFF
--- a/src/Laravel/Eloquent/State/LinksHandler.php
+++ b/src/Laravel/Eloquent/State/LinksHandler.php
@@ -37,19 +37,19 @@ final class LinksHandler implements LinksHandlerInterface
                 $identifier = $uriVariables[$uriVariable];
 
                 if ($to = $link->getToProperty()) {
-                    $builder = $builder->where($builder->getModel()->getTable().'.'.$builder->getModel()->{$to}()->getForeignKeyName(), $identifier);
+                    $builder = $builder->where($builder->getModel()->{$to}()->getQualifiedForeignKeyName(), $identifier);
 
                     continue;
                 }
 
                 if ($from = $link->getFromProperty()) {
                     $relation = $this->application->make($link->getFromClass());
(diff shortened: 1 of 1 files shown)


soyuka:
Awesome thanks @GromNaN ! It's possible that within some filters we're using the same mechanics. With Doctrine we had to duplicate each filter, maybe that with Eloquent we can use the exact same filters?

GromNaN:
> It's possible that within some filters we're using the same mechanics. With Doctrine we had to duplicate each filter, maybe that with Eloquent we can use the exact same filters?

I haven't found any other place where the column name is prefixed.

PR #6659, opened 2024-09-23
fix: passing class as parameter in XML ApiResource's definition

GSadee:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Tickets       | 
| License       | MIT
| Doc PR        | 

In Sylius, where we use API Platform, it is important for us to be able to extend entities and because of this we would like to pass the class as a parameter, which is not possible at the moment due to an error:
```
request.CRITICAL: Exception thrown when handling an exception (Symfony\Component\Config\Exception\LoaderLoadException: The class or interface "%sylius.model.taxon.class%" does not exist in
```

This happens because in the XmlResourceExtractor, the key `class` is set in the ApiResource metadata with a value that does not point to a specific class, but is a parameter key. This is not done in the YamlResourceExtractor, so in order to make the both extractors consistent with each other, I remove the setting of this wrong value. An alternative would be to set the already resolved class via the `resolve` metadata, but I don't think this is needed because the corresponding class is set as a key in the resources array.


src/Metadata/Extractor/XmlResourceExtractor.php
```diff
@@ -61,7 +61,6 @@ protected function extractPath(string $path): void
         foreach ($xml->resource as $resource) {
             $base = $this->buildExtendedBase($resource);
             $this->resources[$this->resolve((string) $resource['class'])][] = array_merge($base, [
-                'class' => $this->phpize($resource, 'class', 'string'),
```
soyuka, line 64: would it be possible to resolve instead in both extractors? A resource without a class is definitely not valid to me.

GSadee: I wondered about this, but in ApiResource the class field does not exist at all if it is not specified here, as far as I have seen, and at the same time the relevant class is indicated as the key, so setting the class here is a kind of unnecessary duplication

soyuka:
thanks!

PR #6660, opened 2024-09-23
fix: overwriting XML ApiResource definition by YAML ApiResource definition

### LINKED ISSUE #6619: YAML ApiResource configuration overwrites XML ApiResource
**API Platform version(s) affected**: 3.3.13

**Description**
In Sylius, we use the API Platform and have all resources configured in XML. However, it is important for us to be able to extend or overwrite these configurations in the end applications. At this point, when using XML configurations in the end application, everything works fine. Unfortunately, using YAML no longer works, because instead of adding a new ApiResource, it overwrites the existing one already in the metadata due to the incrementation from 0 when loading the configuration of each extension.

The problematic line: https://github.com/api-platform/core/blob/3.3/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php#L63

**TL;DR:** there is no way to use both configurations in XML and YAML at the same time, because YAML overwrites XML in unforeseen way.

**How to reproduce**  
Add at the same time the configuration in XML and YAML for the same ApiResource.

**Possible Solution**  
A potential solution would be such a simply change in `ExtractorResourceMetadataCollectionFactory.php`:
```diff
- $resourceMetadataCollection[$i] = $resource;
+ $resourceMetadataCollection[] = $resource;
```

GSadee:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Tickets       | fixes #6619 
| License       | MIT
| Doc PR        | 

In case the same ApiResource is defined in both XML and YAML, this PR makes the ApiResource defined in YAML add itself as another one instead of overwriting the existing one defined in XML.

### DIFF
--- a/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
+++ b/src/Metadata/Resource/Factory/ExtractorResourceMetadataCollectionFactory.php
@@ -53,14 +53,14 @@ final class ExtractorResourceMetadataCollectionFactory implements ResourceMetada
             return $resourceMetadataCollection;
         }
 
-        foreach ($this->buildResources($resources, $resourceClass) as $i => $resource) {
+        foreach ($this->buildResources($resources, $resourceClass) as $resource) {
             foreach ($this->defaults['attributes'] ?? [] as $key => $value) {
                 if (method_exists($resource, 'get'.ucfirst($key)) && !$resource->{'get'.ucfirst($key)}()) {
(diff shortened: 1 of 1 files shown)


soyuka:
Shouldn't your target 3.4 ?

GSadee:
> Shouldn't your target 3.4 ?

Yes, it should, my mistake when opening PR, I changed the base branch as this branch is already rebased with 3.4 👌🏻

soyuka:
Thanks!

PR #6661, opened 2024-09-23
[Laravel]: Fix authorization swagger ui

### LINKED ISSUE #6662: Laravel: Swagger UI Authorization

**Description**  
While API Platform appears to support authorization within Swagger UI, the Laravel module hasn’t fully implemented this functionality yet. As a result, neither OAuth nor API key authentication works within Swagger UI.

**How to reproduce**  
Open the Swagger UI for your API and click the "Authorize" button in the top right. Regardless of your configuration, the authorization window remains empty.

**Possible Solution**  
It seems that the authorization functionality is partially in place, but some configuration values aren’t being passed down to Swagger UI. I’ll submit a PR to address this.

toitzi:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | Closes #6662
| License       | MIT
| Doc PR        | [#1991](https://github.com/api-platform/docs/pull/1991)

### DIFF
--- a/src/Laravel/ApiPlatformProvider.php
+++ b/src/Laravel/ApiPlatformProvider.php
@@ -705,7 +705,33 @@ class ApiPlatformProvider extends ServiceProvider
             /** @var ConfigRepository */
             $config = $app['config'];
 
-            return new Options(title: $config->get('api-platform.title') ?? '');
+            return new Options(
+                title: $config->get('api-platform.title', ''),
+                description: $config->get('api-platform.description', ''),
+                version: $config->get('api-platform.version', ''),
+                oAuthEnabled: $config->get('api-platform.swagger_ui.oauth.enabled', false),
+                oAuthType: $config->get('api-platform.swagger_ui.oauth.type', null),
+                oAuthFlow: $config->get('api-platform.swagger_ui.oauth.flow', null),
+                oAuthTokenUrl: $config->get('api-platform.swagger_ui.oauth.tokenUrl', null),
+                oAuthAuthorizationUrl: $config->get('api-platform.swagger_ui.oauth.authorizationUrl', null),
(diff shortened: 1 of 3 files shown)


toitzi:
@soyuka yes, once im done i will squash commit. 

Different question:
currently bearer tokens are not really supported by api-platform (or at least not fully). There is 2 issues i see when adding bearer token support:

First way / issue, would be adding the option to pass a auth key type via the OpenApiFactory, by adapting [this](https://github.com/api-platform/core/blob/034fcbc73d6b1f64e0fe79cd33264a2dbc401ba3/src/OpenApi/Factory/OpenApiFactory.php#L687) code. It is a non-breaking change (and the way i did it in this PR). But the drawback here is that api-platform already uses `type` in the configuration, which maps to the `in` of the security scheme. Since that is already in use, so i had to come up with a new option `auth_type`. While that works just fine, i can see how this can be confusing to developers / maintainers.

Second way (which removes this confusion a bit) would be to adapt the [Options](https://github.com/api-platform/core/blob/main/src/OpenApi/Options.php) file, with a new property and method (getBearerTokens or smth. like that) and use that additionally in the `OpenApiFactory`. However i would need to append those at the end of the constructor, in order to avoid breaking changes.

What would be the preferred way here?

toitzi:
And also: Is this a feature or a bug? Since it is already somewhat implemented but not really working in laravel - i considered it a bug...if you consider it a feature please let me know (because i think i will also need to merge it to a differen branch then right?)

soyuka:
I see 2 different things here. For Laravel we consider that every existing feature that is missing as a bug fix for 4.x and therefore the `config` patch is fine. For the new option (http auth) I'd see this as a new feature as it changes in Symfony as well. I'd open a new PR to support this.

toitzi:
ok updated to fix only the config, i will make a new PR as feature to support http auth.

soyuka:
Many thanks!

PR #6663, opened 2024-09-23
fix: using parameters in fromClass and toClass uriVariables' options

GSadee:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Tickets       | related to https://github.com/api-platform/core/pull/6659
| License       | MIT
| Doc PR        | 

Currently there is no possibility to pass the class as a parameter in `fromClass` and `toClass` uriVariables' options and this PR adds this missed possibility.

### DIFF
--- a/src/Metadata/Extractor/YamlResourceExtractor.php
+++ b/src/Metadata/Extractor/YamlResourceExtractor.php
@@ -193,13 +193,13 @@ final class YamlResourceExtractor extends AbstractResourceExtractor
                 unset($data[0], $data[1]);
             }
             if (isset($data['fromClass'])) {
-                $uriVariables[$parameterName]['from_class'] = $data['fromClass'];
+                $uriVariables[$parameterName]['from_class'] = $this->resolve($data['fromClass']);
             }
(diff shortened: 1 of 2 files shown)


src/Metadata/Extractor/YamlResourceExtractor.php
```diff
@@ -196,10 +196,10 @@ private function buildUriVariables(array $resource): ?array
                 $uriVariables[$parameterName]['from_class'] = $data['fromClass'];
             }
             if (isset($data['fromProperty'])) {
-                $uriVariables[$parameterName]['from_property'] = $data['fromProperty'];
+                $uriVariables[$parameterName]['from_property'] = $this->resolve($data['fromProperty']);
```
soyuka, line 199: it should be line 196 probably instead of this

soyuka:
May you target 3.4 instead?

GSadee:
> May you target 3.4 instead?

Yes, it should, my mistake when opening PR, I changed the base branch as this branch is already rebased with 3.4 👌🏻

PR #6667, opened 2024-09-23
Fix: route registration of EntrypointController should be after GraphQL's EntrypointController

### LINKED ISSUE #2767: Demo is returning error 500
https://demo.api-platform.com/

Screenshot from 2019-04-27 14-18-19

BurningDog:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | Closes #2767
| License       | MIT

If GraphQL is enabled and API Platform's route prefix is configured to be an empty string instead of the default `/api` then the `/graphql` route is incorrectly handled by `\ApiPlatform\Laravel\Controller\EntrypointController` instead of `ApiPlatform\Laravel\GraphQl\Controller\EntrypointController`.

This is because the routes being handled by the former controller are registered before the GraphQL routes. Instead, they should be registered last.

### DIFF
--- a/src/Laravel/ApiPlatformProvider.php
+++ b/src/Laravel/ApiPlatformProvider.php
@@ -1293,14 +1293,6 @@ class ApiPlatformProvider extends ServiceProvider
         $route->name('api_doc')->middleware(ApiPlatformMiddleware::class);
         $routeCollection->add($route);
 
-        $route = new Route(['GET'], $prefix.'/{index?}{_format?}', function (Request $request, Application $app) {
-            $entrypointAction = $app->make(EntrypointController::class);
-
-            return $entrypointAction->__invoke($request);
-        });
-        $route->where('index', 'index');
-        $route->name('api_entrypoint')->middleware(ApiPlatformMiddleware::class);
-        $routeCollection->add($route);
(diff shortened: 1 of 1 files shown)


BurningDog:
Hmmm...I'm not quite sure why the Behat tests are breaking. At first glance it's tempting to assume that routing is breaking, but I can add `#[QueryParameter(key: 'author', filter: PartialSearchFilter::class)]` to the `Book` model locally and run 

```sh
curl -X 'GET' \
  'http://127.0.0.1:8000/books?page=1&author=Nikki' \
  -H 'accept: application/ld+json'
```

and get a response just fine. 

So perhaps the mechanism for seeding data for the tests is now broken, and so no expected results are returned in the tests...but I've never used Behat and I'm not sure where the data fixtures are.

Advice welcome!

soyuka (approved):

soyuka:
No worries we're using only phpunit for Laravel, instructions to run tests are at [`src/Laravel/CONTRIBUTING.md`](https://github.com/api-platform/core/blob/main/src/Laravel/CONTRIBUTING.md). 

Nice catch on the bug, of course route order has an impact, not sure why we didn't catch this with a route prefix though. I'll add a test later on, thanks!

dunglas (approved):
Good catch!

PR #6668, opened 2024-09-24
fix(laravel): eloquent accessors

### LINKED ISSUE #6644: Error with Laravel Eloquent Accessors

**Description**  

In Laravel, it's possible to define custom Accessors for your models (see [Laravel documentation](https://laravel.com/docs/11.x/eloquent-mutators#defining-an-accessor)). These accessors can either modify an existing property or add new attributes that are not tied to database columns.

Several popular Laravel packages, such as [Spaite's Laravel Translatable](https://spatie.be/docs/laravel-translatable/v6/introduction) utilize this feature to append or manipulate attributes. However, when using such accessors on models defined as `#[ApiResource]`, API Platform throws an error. This occurs because these custom attributes lack a "primary" key in their metadata array, and their "nullable" value is set to `null` instead of a boolean (`true/false`). The issue is exemplified in the screenshot below.

**How to reproduce**  

In any Laravel installation, define a model as `#[ApiResource]` and add the following accessor:

```
protected function test(): Attribute
{
    return Attribute::make(
        get: fn () => 'broken',
    );
}
```

API Platform will then produce an error: `undefined array key "primary"` [on this line](https://github.com/api-platform/laravel/blob/e821ad27072228e4638f3de92d51fc3d15e02c7b/Eloquent/Metadata/Factory/Property/EloquentPropertyNameCollectionMetadataFactory.php#L56)

**Possible Solution**  

At first glance, this might seem like a simple fix (adding an `isset` check on the line mentioned above). However, the error will still occur in the `EloquentPropertyMetadataFactory` [on this line](https://github.com/api-platform/laravel/blob/e821ad27072228e4638f3de92d51fc3d15e02c7b/Eloquent/Metadata/Factory/Property/EloquentPropertyMetadataFactory.php#L80), since `$p['nullable']` is null instead of a `boolean`. A potential fix might involve defaulting it to true (`$p['nullable'] ?? true`), though I'm uncertain if this is a proper solution or just a quick workaround, given that accessors created in this way might not need to be considered in the first place.

EDIT: A potential workaround is to hide these attributes by adding them to the model's `hidden` array. However, the error related to the missing isset check would still remain (though the second error would be resolved). In this case, only the isset check, or something similar, would need to be added. It might also be worth documenting that users may need to hide such attributes manually. Automating this process would be ideal, but I’m unsure if there's a reliable way to differentiate between "real" attributes and those that aren't tied to the database (except maybe for the absence of the `primary` key).

**Additional Context**  
image

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| Tickets       | Closes #6644
| License       | MIT

Fixes https://github.com/api-platform/core/issues/6644 @toitzi may you review?

### DIFF
--- a/src/Laravel/Eloquent/Metadata/Factory/Property/EloquentPropertyMetadataFactory.php
+++ b/src/Laravel/Eloquent/Metadata/Factory/Property/EloquentPropertyMetadataFactory.php
@@ -77,7 +77,7 @@ final class EloquentPropertyMetadataFactory implements PropertyMetadataFactoryIn
                 'collection', 'encrypted:collection' => new Type(Type::BUILTIN_TYPE_ITERABLE, $p['nullable'], Collection::class, true),
                 'encrypted:array' => new Type(Type::BUILTIN_TYPE_ARRAY, $p['nullable']),
                 'encrypted:object' => new Type(Type::BUILTIN_TYPE_OBJECT, $p['nullable']),
-                default => new Type(\in_array($builtinType, Type::$builtinTypes, true) ? $builtinType : Type::BUILTIN_TYPE_STRING, $p['nullable']),
+                default => new Type(\in_array($builtinType, Type::$builtinTypes, true) ? $builtinType : Type::BUILTIN_TYPE_STRING, $p['nullable'] ?? true),
             };
 
             return $propertyMetadata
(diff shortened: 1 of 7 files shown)


toitzi:
@soyuka yes, it does fix the issue, however i think instead of this, fixing the underlaying `Metadata` class (see [this pr](https://github.com/api-platform/core/pull/6633)) might be the better approach here. (Maybe merge those 2 Pr's - his code changes and your unittests)

soyuka:
#6633 is definitely good but we should merge both as we can't know what's in this array for sure I think its best to define defaults values as it triggers a Fatal Error (500) and I'd like to avoid this anyways.

dunglas (approved):

PR #6478, opened 2024-07-19
fix(jsonschema): make all required properties optional in PATCH operation with 'json' format

soyuka:
by @ttskch see #6394

PR #6676, opened 2024-09-26
fix(jsonapi): fixed definition name to allow using the same class names in different namespaces

nightio:
| Q             | A
| ------------- | ---
| Branch?       | 3.3 
| Tickets       | -
| License       | MIT
| Doc PR        | -

It fixes documentation models reference in the case where there are 2 or more classes with the same name but in different namespaces.
Previously the very first class model was created and used in all places in documentation.

Now if the name is the same but namespace is different, a previous one element of a namespace is prepended to the name until the name is unique

### DIFF
--- a/src/JsonSchema/DefinitionNameFactory.php
+++ b/src/JsonSchema/DefinitionNameFactory.php
@@ -21,6 +21,9 @@ final class DefinitionNameFactory implements DefinitionNameFactoryInterface
 {
     use ResourceClassInfoTrait;
 
+    private const GLUE = '.';
+    private array $prefixCache = [];
+
     public function __construct(private ?array $distinctFormats)
     {
     }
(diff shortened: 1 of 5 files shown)


nightio:
@dunglas @soyuka Could you look at that?

soyuka:
just as #6485 and #6478 I'm afraid that changing json schema names may break stuff, therefore this needs a little bit more investigation and I don't have much time right now. I'm adding this to the list of enhancements for the next minor version as I really want to tackle the issue.

nightio:
@soyuka it should not break anything - please look at the tests. The original cases are untouched.

If there are no 2 or more classes with the same name the output is identical.
In case more than one class has the same name just a new definition is added with part of the namespace, the first one looks exactly the same as in the previous implementation.

But it fixes the bug where documentation is generated in a wrong way.

soyuka:
Thanks @nightio !

PR #6683, opened 2024-09-30
fix: parameter provider in a long running http worker

rmikalkenas:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Tickets       | -
| License       | MIT
| Doc PR        | -

In case of using long running http workers (Roadrunner for example), query parameters and in general parameters functionality tends to persist previous request value due to aggressive [caching of resource metadata ](https://github.com/api-platform/core/blob/main/src/Metadata/Resource/Factory/CachedResourceMetadataCollectionFactory.php)

Fixing by clearing extra data before parsing new one

Targeting 3.4, but the issue reproducible at 4.0 as well

### DIFF
--- a/src/State/Provider/ParameterProvider.php
+++ b/src/State/Provider/ParameterProvider.php
@@ -52,6 +52,10 @@ final class ParameterProvider implements ProviderInterface
         $context = ['operation' => $operation] + $context;
         $parameters = $operation->getParameters();
         foreach ($parameters ?? [] as $parameter) {
+            $extraProperties = $parameter->getExtraProperties();
+            unset($extraProperties['_api_values']);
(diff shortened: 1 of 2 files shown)


dunglas (approved):
LGTM but let's wait for @soyuka review before merging.

dunglas:
Actually, shouldn't we also don't cache parameters at all?

soyuka:
Interesting use case, I wish we wouldn't cache this at all actually but we don't really have a way to. I think that having the value inside the Parameter is a good DX feature but indeed having "living" values inside immutable Metadata is not the best. 

@dunglas actually the metadata for Parameters should be cached but values should not. Let's merge this for now as a quick fix.

soyuka:
thanks!

PR #6691, opened 2024-10-03
fix(laravel): fixes to secure a graphql endpoint

soyuka:
As [@toitzi](https://github.com/toitzi) reported GraphQl endpoints had no way to be secured. After a few messages we decided to introduce a new configuration option: 

```
// app/config/api-platform.php
    'routes' => [
        // Global middleware applied to every API Platform routes
        // 'middleware' => []
    ],
```

This gives the ability to plugin in a global auth middleware and to use Policies with GraphQl operations.

### DIFF
--- a/src/Laravel/ApiPlatformProvider.php
+++ b/src/Laravel/ApiPlatformProvider.php
@@ -366,7 +366,8 @@ class ApiPlatformProvider extends ServiceProvider
                                                                     $config->get('api-platform.graphql.enabled'),
                                                                 ),
                                                             )
-                                                        )
+                                                        ),
+                                                        $config->get('api-platform.graphql.enabled')
                                                     )
                                                 )
                                             )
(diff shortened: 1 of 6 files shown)


src/Laravel/ApiPlatformProvider.php
```diff
@@ -936,51 +940,6 @@ public function register(): void
             );
         });
 
-        if ($this->app['config']->get('api-platform.graphql.enabled')) {
```
soyuka, line 939: it's not good to read configuration inside the register method as it may change when `boot` is called

src/Laravel/ApiPlatformProvider.php
```diff
@@ -1130,44 +1129,78 @@ private function registerGraphQl(Application $app): void
...
-                ),
+                $app->make('api_platform.graphql.state_processor.normalize'),
                 $app->make(CallableProcessor::class),
             );
         });
 
         $app->singleton(ResolverFactoryInterface::class, function (Application $app) {
             return new ResolverFactory(
-                $app->make('api_platform.graphql.state_provider'),
+                $app->make('api_platform.graphql.state_provider.access_checker'),
```
soyuka, line 1203: @vinceAmstoutz I splitted the system providers for Laravel!

vinceAmstoutz: @soyuka :tada: Can we prepare documentation now and merge it when your PR will be merged?

PR #6696, opened 2024-10-07
fix: check that api-platform/ramsey-uuid is installed before registering related services

### LINKED ISSUE #6695: ramsey-uuid package required due some external package
**API Platform version(s) affected**: 3.4

**Description**  

We don't use ramsey/uuid in our own code but an external package is installing this package. Due the check inside api-platform/symfony we must install api-platform/ramsey-uuid to not get an error. 

**How to reproduce**  

**Possible Solution**  

Can we improve this check so we can remove this dependency?

**Additional Context**

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | 3.4| Tickets
| Closes #6695
| License       | MIT
| Doc PR        | n/a


soyuka (approved):
we should also suggest the api-platform/ramsey-uuid package inside of ramsey uuid itself inside `src/Symfony/composer.json`

dunglas:
@soyuka done

Issue #6693, opened 2024-10-05
ContextBuilder called twice for normalization context


maxhelias:
**API Platform version(s) affected**: 3.4

**Description**  
During the runtime of a request on my app, I noticed that my ContextBuilder::createFromRequest was called 2 times to create the normalization context : 
- First call here : https://github.com/api-platform/core/blob/fcd9a0725fb9fab942b306e5d3cb0e96226ee63e/src/State/Provider/ReadProvider.php#L71
- Seconde here : https://github.com/api-platform/core/blob/fcd9a0725fb9fab942b306e5d3cb0e96226ee63e/src/State/Processor/SerializeProcessor.php#L55

**How to reproduce**  
Create a ContextBuilder or check out ContextBuilder's built-in 

**Possible Solution**  
I was wondering if we can in the SerializeProcessor, check whether the request attribute `_api_normalization_context` is already computed or not, to avoid a second call.
Normally the context calculation shouldn't change in the same runtime, but I may be missing something.

**Additional Context**  
This may be intentional, in which case, another solution might be considered to leave the possibility of recomputing/caching or not the normalization context at the operation level or elsewhere to leave the choice to the user if a recomputing is necessary

soyuka:
Sometimes `read` is not called (on POST operations for example). But indeed, we may add a check and use the previously computed context.

maxhelias:
Oh yes, thanks, i had in mind the read option on the operation but had forgotten about it.
The legacy system with listeners does the same thing. I'll do a PR soon for both 😉

PR #6700, opened 2024-10-07
fix: prevents recomputing normalizationcontext

maxhelias:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Tickets       | #6693
| License       | MIT
| Doc PR        | -

Add a check and use the previously computed context

### DIFF
--- a/src/Serializer/RequestCacheSerializerContextBuilder.php
+++ b/src/Serializer/RequestCacheSerializerContextBuilder.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ApiPlatform\Serializer;
+
+use ApiPlatform\State\SerializerContextBuilderInterface as StateSerializerContextBuilderInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * {@inheritdoc}
+ *
+ * @author Maxime Hélias <maximehelias16@gmail.com>
+ */
+final class RequestCacheSerializerContextBuilder implements SerializerContextBuilderInterface
+{
+    public function __construct(private readonly SerializerContextBuilderInterface|StateSerializerContextBuilderInterface $decorated)
+    {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createFromRequest(Request $request, bool $normalization, ?array $extractedAttributes = null): array
+    {
+        $attributeKey = $normalization ? '_api_normalization_context' : '_api_denormalization_context';
+
+        if ($request->attributes->has($attributeKey)) {
+            return $request->attributes->get($attributeKey);
+        }
+
+        $context = $this->decorated->createFromRequest($request, $normalization, $extractedAttributes);
+
+        $request->attributes->set($attributeKey, $context);
+
+        return $context;
+    }
+}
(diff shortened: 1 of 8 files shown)


src/State/Processor/SerializeProcessor.php
```diff
@@ -57,10 +57,12 @@ public function process(mixed $data, Operation $operation, array $uriVariables =
             trigger_deprecation('api-platform/core', '3.3', 'The resource class on the router is not the same as the operation\'s class which leads to wrong behaviors. Prefer using "stateOptions" if you need to change the entity class.');
         }
 
-        $serializerContext = $this->serializerContextBuilder->createFromRequest($request, true, [
-            'resource_class' => $class,
-            'operation' => $operation,
-        ]);
+        if (null === ($serializerContext = $request->attributes->get('_api_normalization_context'))) {
```
soyuka, line 60: shouldn't we move the responsibility of this to the `SerializerContextBuilder` directly? It'd avoid duplicating this condition. I'd also do the `set` there.

maxhelias: Do you have in mind a decorator of `api_platform.serializer.context_builder` with a low priority to `get`/`set` the computed context in the attribute request ?
It was certainly a better way to go because the `SerializerContextBuilder` is at the end of the inheritance chain of `api_platform.serializer.context_builder`.

soyuka: Indeed, we could introduce a RequestCacheSerializerContextBuilder that has a negative priority?

maxhelias: I'll take care of it.
What about denormalizationContext? This notion doesn't exist in the request's attributes and I see that we can have 2 calls as well:
- https://github.com/api-platform/core/blob/fbb53e5e35ca0ec3de26ddc7de7ea4d1dda5c20b/src/Symfony/Bundle/ArgumentResolver/PayloadArgumentResolver.php#L78
- and in deserialize layer : https://github.com/api-platform/core/blob/fbb53e5e35ca0ec3de26ddc7de7ea4d1dda5c20b/src/Symfony/EventListener/DeserializeListener.php#L136 or https://github.com/api-platform/core/blob/fbb53e5e35ca0ec3de26ddc7de7ea4d1dda5c20b/src/State/Provider/DeserializeProvider.php#L68

I'm missing information about the execution level of the PayloadArgumentResolver, can you tell me more? Are we adding this support to the `RequestCacheSerializerContextBuilder` as well ?

soyuka:
I thought about this a little bit more and I think that we shouldn't cache the context. A few things may change the context between providers and we can't be sure that the previously cached value is correct. If we do cache, every one that changes the context inside its provider needs to make sure that the `_api_normalization_context` is fresh... We're not even using the `_api_normalization_context` in our code (in DocumentationAction I don't think its useful).

maxhelias:
I understand. 
What is the recommendation in my case ? I have a database query to add a key to the context to activate some normalizer. The result of the query doesn't change during runtime. I can clearly do with 1 query.

- Should i use a locale cache to solve it ? Maybe a mention in the documentation can be added ?
- Or introduce a new concept/interface or something else to determine whether the result of a context builder can be cached ?

soyuka:
sorry for the late answer I missed the comment, as the ContextBuilder is tight to Request I recommend that you do something like this: 

```
// in a middleware, or a Provider decorating one of our system providers (https://api-platform.com/docs/core/extending/#system-providers-and-processors):

$request->attributes->set('my_queried_value', $computed');
```

Then create a SerializerContextBuilder that does: 

```
return ['my_context' => $request->attributes->get('my_queried_value')] + $this->decorated->createFromRequest(...)];
```

maxhelias:
Hi @soyuka,
Sorry for the late reply, I just came back to work on it and it's working well.
Thanks

Issue #6701, opened 2024-10-07
4.0.3 breaks BC on error handling: hydra prefixes missing


j-schumann:
**API Platform version(s) affected**: 4.0.3, 3.4.2

**Description**  
In 4.0.3 + 3.4.2 the "hydra:" prefix was removed from error responses (in https://github.com/api-platform/core/pull/6624/commits/7d23c76fd13e2283af95e9aefc5540a1bc86215a). 
This breaks error handling in our clients in multiple APIP projects as they read `hydra:description`

**How to reproduce**  
APIP works in 4.0.2 -> update api-platform/state|validator to 4.0.3 without any other change -> hydra properties disappear

**Possible Solution**  
a) Revert this "fix", it is a BC break which should not occur in a patch/minor version, especially not without previous deprecation or changelog note (in 4.0.3)
b) Add an option to keep the prefix, as I would expect `serializer: hydra_prefix: true` to behave. This option should be enabled by default (else BC break)

**Additional Context**  
I don't know why this breaking change was introduced in 3.4.2, it was known to be breaking as it was noted in the [Changelog](https://github.com/api-platform/core/blob/v4.0.3/CHANGELOG.md?plain=1#L145):

> Hydra prefix on errors is breaking, read `title` not `hydra:title`. The `hydra_prefix` flag doesn't apply to errors as it provided redundant information (both `hydra:title` and `title` were available)

This note should also have been on the 4.0.3 changelog as this was the first 4.x version that introduced that break. (At least it would have saved me half an hour debugging as nothing in the changelog indicated that the reason could be this version.)

I am not the only one relying on those response fields: https://github.com/api-platform/core/issues/5963#issuecomment-1808497808

soyuka:
Using `rfc_7807_compliant_errors: true` (extra_properties => default) you keep the prefix right? In 4.x this breaks and you don't have these prefixes anymore on errors. We had so much going on that it was hard to think about everything and 4.0.3 has a bug fix as in 4.x these prefixes should've been removed. Indeed, `hydra_prefix` has no incidence on errors, I'll try to find a way to add this behavior it may mitigate the BC break.

soyuka:
I hope my patch fixes your issue, I'll release by Friday

burned42:
I just tested the 4.0.4 release and for me this issue seems to be fixed and in error responses the title and description with hydra prefix are back when using `hydra_prefix: true` :+1: 
Thanks :)

j-schumann:
[...]

PR #6704, opened 2024-10-08
fix(hydra): hydra_prefix on errors

soyuka:
Mitigates #6701

### DIFF
--- a/src/JsonLd/Serializer/ErrorNormalizer.php
+++ b/src/JsonLd/Serializer/ErrorNormalizer.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the API Platform project.
+ *
+ * (c) Kévin Dunglas <dunglas@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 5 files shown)


src/JsonLd/Serializer/ErrorNormalizer.php
```diff
@@ -0,0 +1,58 @@
...
+    use HydraPrefixTrait;
+
+    public function __construct(private readonly NormalizerInterface $inner, private readonly array $defaultContext = [])
+    {
+    }
+
+    public function normalize(mixed $object, ?string $format = null, array $context = []): array|string|int|float|bool|\ArrayObject|null
+    {
+        $normalized = $this->inner->normalize($object, $format, $context);
+        $hydraPrefix = $this->getHydraPrefix($context);
```
dannyvw, line 32: `$hydraPrefix = $this->getHydraPrefix($context + $this->defaultContext);`

soyuka: mhh this is quite weird isn't the default context added via the ContextBuilder anyways? Or maybe it is only through the `api_platform.serializer` option and not `serializer.default_context`?

PR #6710, opened 2024-10-09
fix(hydra): hydra context changed

valentin-dassonville:
| Q             | A
| ------------- | ---
| Branch?       | 4.0
| License       | MIT

PR on admin : https://github.com/api-platform/admin/pull/586
PR on api-doc-parser : https://github.com/api-platform/api-doc-parser/pull/139

Modifying the jsonld context to work with and without the hydra prefix.


valentin-dassonville:
@dunglas Do the values in the `docs.jsonld` fields have to start with a #, given that the urls in the `@context` already end with # ?

PR #6720, opened 2024-10-13
feat(doctrine): allow for comma-separated values in numeric-filter

frankdekker:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | api-platform/docs#2037

Currently the NumericFilter allows for searching for `numeric-string` or `numeric-string[]`:
```
/api/entity?numericValue=5
/api/entity?numericValue[]=5&numericValue[]=6
```

This PR adds the support for comma-separated numeric-values:
```
/api/entity?numericValue=5,6
```

Which allows for much shorter urls when searching on multiple numeric values and if I understand the specification correctly this is a valid format. A snipper from the documentation:

> Query parameters can be primitive values, arrays and objects. OpenAPI 3.0 provides several ways to serialize objects and arrays in the query string.
>
> Arrays can be serialized as:
>
>    `form` – /products?color=blue,green,red or /products?color=blue&color=green, depending on the explode keyword

https://swagger.io/docs/specification/v3_0/describing-parameters/#query-parameters

### DIFF
--- a/src/Doctrine/Common/Filter/NumericFilterTrait.php
+++ b/src/Doctrine/Common/Filter/NumericFilterTrait.php
@@ -81,6 +81,11 @@ trait NumericFilterTrait
 
     protected function normalizeValues($value, string $property): ?array
     {
+        // Allow CSV format for multiple values.
+        if (\is_string($value) && str_contains($value, ',')) {
+            $value = explode(',', $value);
+        }
+
         if (!is_numeric($value) && (!\is_array($value) || !$this->isNumericArray($value))) {
             $this->getLogger()->notice('Invalid filter ignored', [
                 'exception' => new InvalidArgumentException(\sprintf('Invalid numeric value for "%s" property', $property)),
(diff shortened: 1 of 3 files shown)


frankdekker:
I'm a bit lost how to test the NumericFilter. Looking at `src/Doctrine/Orm`, running composer install in this directory and then starting the test will use the api-platform/core classes from vendor instead of using `src/Doctrine/Orm/Filter/NumericFilter.php`

If I run the test with phpunit from the root of the project, the test crashes as the Doctrine driver can't be found.

I could use some tips how to solve this.

soyuka:
I'm really not sure we want to add this to our codebase as all our filters work by having this syntax: 

```
/api/entity?numericValue[]=5&numericValue[]=6
```

With the QueryParameter we tried to improve the tools users can use to re-use filters but in their own way, it should be fairly easy to add (and document) this in your own codebase.

PR #6724, opened 2024-10-14
Patch 2

ViPErCZ:
turn off inserting @id into a schema that is not a resource and has @id autogeneration turned off

cherry pick from version 3.4 > https://github.com/api-platform/core/pull/6716


ViPErCZ:
fail.... i closed this fail PR, sorry

